PR #6862, opened 2019-01-18
Lock when removing disconnect handler

pakrym:
While looking into https://github.com/aspnet/AspNetCore-Internal/issues/1551 I was able to force an AV caused by unsynchronized `m_pHandler` access in `RemoveHandler` method.

May fix the failing test too.


jkotalik (approved):

davidfowl:
@pakrym manager approaval required to merge things. We're stabilizing...

PR #6867, opened 2019-01-19
Remove JsonPatch from the shared framework

natemcmaster:
Part of https://github.com/aspnet/AspNetCore/issues/4260 and https://github.com/aspnet/AspNetCore/issues/3755

This dependency should have been removed when we did https://github.com/aspnet/AspNetCore/pull/5146, but it got missed. Per our design review, this was one of the assemblies we decided should be removed from the shared framework. Users can add this to their projects by adding a PackageReference.

cc @mkArtakMSFT @muratg we should do this for Preview 2 to front load as many breaking changes as we can.

### DIFF
--- a/build/SharedFrameworkOnly.props
+++ b/build/SharedFrameworkOnly.props
@@ -8,7 +8,6 @@
 
   <ItemGroup>
     <!-- Packages to be removed from the shared framework but not done yet due to JSON.net dependency. -->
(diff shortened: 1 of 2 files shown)


pranavkm (approved):
Package is already in the right section in artifacts.props: https://github.com/aspnet/AspNetCore/blob/master/build/artifacts.props#L56

davidfowl (approved):

PR #6869, opened 2019-01-19
Move Kestrel Sockets tests to their own group

Tratcher:
We'll see if this works around the Kestrel binding issues like it did for Identity.
https://github.com/aspnet/AspNetCore/pull/6865


Tratcher:
Success! The Kestrel tests all passed before the build eventually timed out in scaffolding. It doesn't look like it hung either, just that it ran a lot of tests across a lot of TFMs. I recommend we merge it as is.

Tratcher:
@davidfowl It didn't seem like a flaky issue, just that the timeout wasn't long enough.

Tratcher:
Same result, no failures or hangs, it just runs out of time before finishing the scaffolding tests.

davidfowl:
Hmm other PRs failing similarly

natemcmaster (approved):

natemcmaster:
We intentionally wanted to run test assemblies in parallel to see why kind of issues we would hit under stress. I'm fine with this as long as this isn't working around real product issues.

Tratcher:
It's not an AspNetCore product issue but it is likely an SDK or Xunit bug. It's like clear why test groups would affect assembly binding.

Issue #6106, opened 2018-12-23
jQuery GET complex query string not binding correctly


Jimmys20:
### Describe the bug
I'm trying to call a Web API using jQuery.ajax and pass a complex object in the query string. The query string is not binding correctly with the model. The parameters which are formatted in the query string like this:

property[name]=xxx

or this:

array[0][name]=yyy
array[1][name]=zzz

are not binding with equivalent model properties. When I send the same parameters via POST method in the body then they bind correctly with the model. But when using GET you have to format them like this

property.name=xxx
array[0].name=yyy
array[1].name=zzz

for it to work. How do you suggest to solve this problem?

### To Reproduce
I'm attaching a demo app which shows my problem. [demo.zip](https://github.com/aspnet/AspNetCore/files/2706028/demo.zip)

### Expected behavior
I would expect the model to bind correctly with the query string that jQuery.ajax produces. I shouldn't have to build the query string myself, I would like to be able to use the one jQuery produces.

### Additional context
I'm using asp.net core 2.2.

Some old related issues:
[Databinding discrepancy between post/get on complex type](https://github.com/aspnet/Mvc/issues/6361)
[jQuery ajax request with complex data does not work with .net core 1.1 model bindings](https://github.com/aspnet/Mvc/issues/6372)
[Added JQueryQueryStringValueProviderFactory and JQueryQueryStringValueProvider](https://github.com/aspnet/AspNetCore/commit/a952313f1c25651f380e615f286a3f8fbbdee37d)

doominator42:
@Jimmys20 As a workaround before the PR is merged, you can manually add the `JQueryQueryStringValueProviderFactory` in your `Startup.cs`

```
// using Microsoft.AspNetCore.Mvc.ModelBinding

services.AddMvc(options =>
{
	if (!options.ValueProviderFactories.Any(f => f is JQueryQueryStringValueProviderFactory))
		options.ValueProviderFactories.Add(new JQueryQueryStringValueProviderFactory());
});
```

pranavkm:
@Jimmys20, as @doominator42 noted you have to register the `JQueryQueryStringValueProviderFactory` as a value provider in your application startup:

```C#
services.AddMvc(options =>
{
     options.ValueProviderFactories.Add(new JQueryQueryStringValueProviderFactory());
});
```

Closing as the question has been answered.

PR #6893, opened 2019-01-20
Add JQueryQueryStringValueProviderFactory in default MVC options

doominator42:
`JQueryQueryStringValueProviderFactory` is not in the default MVC options, which results in query strings of format `model[property]` to not bind to models.

As the `JQueryFormValueProviderFactory` is added by default, I think the `JQueryQueryStringValueProviderFactory` should also be added by default, for consistency.

Addresses #6106


pranavkm:
@doominator42 thank you for your PR. 

Not registering `JQueryQueryStringValueProviderFactory` by default is intentional. Using the JQuery* value provider factories results in a fairly expensive per-action invocation cost to the pipeline. Passing complex types over the query string is uncommon and we would not want to add the costs of using this value provider to the vast majority of developers who would not benefit from this change. Configuring MVC to use the value provider is fairly trivial, and would be the way to go here.

PR #6911, opened 2019-01-21
Hoist method attributes from the request delegate as metadata.

davidfowl:
- This should allow a more declarative approach to declaring endpoint metadata using the default methods.
- Removed the regions because I was offended :trollface: 

Here's an example of using the Authorize attribute (it's a shame local functions don't support attributes yet):

```C#
using System.Security.Claims;
using System.Threading.Tasks;
using Microsoft.AspNetCore.Authentication;
using Microsoft.AspNetCore.Authorization;
using Microsoft.AspNetCore.Routing;

namespace WebApplication66
{
    public class Startup
    {
        {
            services.AddAuthentication("Cookies")
                    .AddCookie();

            services.AddAuthorization();

            services.AddAuthorizationPolicyEvaluator();
        }

        public void Configure(IApplicationBuilder app, IHostingEnvironment env)
        {
            if (env.IsDevelopment())
            {
            }

            app.UseAuthentication();

            app.UseRouting(routes =>
            {
                routes.MapGet("/", Hello);
                routes.MapGet("/account/login", Login);
                routes.MapGet("/account/logout", Logout);
            });

            app.UseAuthorization();
        }

        [Authorize]
        async Task Hello(HttpContext context)
        {
            await context.Response.WriteAsync($"Hello {context.User.Identity.Name}");
        }

        async Task Login(HttpContext context)
        {
            var claims = new[] {
                        new Claim(ClaimTypes.NameIdentifier, "id"),
                        new Claim(ClaimTypes.Name, "David")
                    };

            var identity = new ClaimsIdentity(claims, "Cookies");
            var user = new ClaimsPrincipal(identity);
            await context.SignInAsync(user);
        }

        async Task Logout(HttpContext context)
        {
            await context.SignOutAsync();
        }
    }
}
```

### DIFF
--- a/src/Http/Routing/src/Builder/EndpointRouteBuilderExtensions.cs
+++ b/src/Http/Routing/src/Builder/EndpointRouteBuilderExtensions.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Reflection;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Routing;
 using Microsoft.AspNetCore.Routing.Patterns;
@@ -21,7 +22,6 @@ namespace Microsoft.AspNetCore.Builder
         private static readonly string[] PutVerb = new[] { "PUT" };
         private static readonly string[] DeleteVerb = new[] { "DELETE" };
 
-        #region MapVerbs
         /// <summary>
         /// Adds a <see cref="RouteEndpoint"/> to the <see cref="IEndpointRouteBuilder"/> that matches HTTP GET requests
         /// for the specified pattern.
@@ -227,9 +227,7 @@ namespace Microsoft.AspNetCore.Builder
 
             return Map(builder, pattern, displayName ?? $"{pattern} HTTP: {string.Join(", ", httpMethods)}", requestDelegate, metadata: resolvedMetadata.ToArray());
         }
-        #endregion
 
-        #region Map
         /// <summary>
         /// Adds a <see cref="RouteEndpoint"/> to the <see cref="IEndpointRouteBuilder"/> that matches HTTP requests
         /// for the specified pattern.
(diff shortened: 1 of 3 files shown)


src/Http/Routing/src/Builder/EndpointRouteBuilderExtensions.cs
```diff
@@ -333,6 +334,12 @@ public static class EndpointRouteBuilderExtensions
                 }
             }
 
+            // Add delegate attributes as metadata
+            foreach (var attribute in requestDelegate.Method.GetCustomAttributes())
```
khellang: What is the `Method` for a lambda expression?

davidfowl: The compiler always generates a method. You just can't apply attributes to it because you don't have a place to put them.

JamesNK:
Cute. I don't hate it, but it is a little unusual.

One thing to think about and test is metadata precedence.

```cs
        app.UseRouting(routes =>
        {
            routes.MapGet("/", Hello);
        });

        [HttpPost]
        async Task Hello(HttpContext context)
        {
            await context.Response.WriteAsync($"Hello {context.User.Identity.Name}");
        }
```

Does `/` accept GET or POST?

davidfowl:
I don't think it's that big of an issue though, it's an edge case. Luckily we have no attributes that implement IHttpMethodMetadata so you can't easily shoot yourself in the foot. What happens today if you supply conflicting metadata on an end point? We don't throw and this is a similar case:

```C#
app.UseRouting(routes =>
{
	routes.MapGet("/", Hello, new HttpMethodMetadata(new[] { "POST" }));
});

async Task Hello(HttpContext context)
{
	await context.Response.WriteAsync($"Hello {context.User.Identity.Name}");
}
```

JamesNK:
> What happens today if you supply conflicting metadata on an end point?

The rule is the last metadata added for the interface takes precedence - https://github.com/aspnet/AspNetCore/blob/e004e5aab29dc26beaafaed49c7aa4b780b6e98f/src/Http/Http.Abstractions/src/Routing/EndpointMetadataCollection.cs#L67-L84

We need to decide what order the metadata gets added. IMO in order of priority:

1. IEndpointRouteBuilder extension method (e.g. MapGet)
2. IEndpointConventionBuilder (e.g. RequireAuthorization)
3. Attributes (e.g. [Authorize])

So in my example - https://github.com/aspnet/AspNetCore/pull/6911#issuecomment-456210352 - the GET from MapGet would take precedence over the attribute.

davidfowl:
I think attributes should be applied before, not after. Attributes are part of the source definition so any explicit metadata should be added later. That way attributes are not an override.

JamesNK:
I agree, that's what I meant. Invert the order of priority for the order that they should be added to the metadata collection.

davidfowl:
🆙 📅

rynowak (dismissed):
use regions

rynowak:
This is dope. I like a lot that we'd make this style of progrmming more capable

JamesNK (approved):
image

PR #6927, opened 2019-01-22
Move Kestrel.Core.Tests to their own test group

Tratcher:
https://github.com/aspnet/AspNetCore-Internal/issues/1699
Experimenting to see if this is the same issue we've been having in 2.1.

Compare to the baseline build https://dnceng.visualstudio.com/public/_build/results?buildId=77668

### DIFF
--- a/src/Servers/Kestrel/Core/test/Microsoft.AspNetCore.Server.Kestrel.Core.Tests.csproj
+++ b/src/Servers/Kestrel/Core/test/Microsoft.AspNetCore.Server.Kestrel.Core.Tests.csproj
@@ -1,8 +1,9 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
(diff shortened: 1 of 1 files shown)


Tratcher:
Success! There are unrelated failures but the Kestrel tests are fixed.

@pakrym that IIS chunked test you disabled in master is failing in 2.2.

halter73 (approved):

halter73:
@natemcmaster Is there anyway to know for sure whether the other test projects are running against the updated Kestrel.Core assembly?

natemcmaster:
> Is there a way to confirm the other test projects are running against the updated Kestrel.Core assembly?

Build locally and check the Kestrel.Core.dll file to see if it matches one from a package or the local build. If it's from a package, add a direct `<Reference>` item to Kestrel.Core into the test .csproj file.

halter73:
Do we have to manually do that for every test project in the repo that might depend on Kestrel.Core?

natemcmaster:
Yes.

Issue #6020, opened 2018-05-30
[Authorize(AuthenticationSchemes = AzureADDefaults.BearerAuthenticationScheme)] generates compiler error CS0182


robinmira:
According to the [documentation](https://docs.microsoft.com/en-us/aspnet/core/security/authorization/limitingidentitybyscheme?view=aspnetcore-2.0&tabs=aspnetcore2x) it should be possible to explicitly specify or override the AuthenticationScheme for a controller with the Authorize attribute.

The following snippet is from the ValuesController.cs file in the base ASP.NET Core Web Application API template with "Work or School accounts" authentication.

`[Route("api/[controller]")]`
`[Authorize(AuthenticationSchemes = AzureADDefaults.BearerAuthenticationScheme)]`
`[ApiController]`
`public class ValuesController : ControllerBase`
`{ ... } `

However, the above code generates compiler error CS0182: An attribute argument must be a constant expression, typeof expression or array creation expression of an attribute parameter type.

A potential fix could be to change the member properties of AzureADDefaults from **static readonly** to **const** as in the [JwtBearerDefaults class](https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication.JwtBearer/JwtBearerDefaults.cs)

`public const string BearerAuthenticationScheme = "AzureADBearer";`

Tratcher:
@muratg recommend fix for 3.0. We're doing something similar for HttpSys windows auth.

muratg:
@Tratcher :+1:  Let's triage this in the meeting today. cc @Eilon

Issue #5884, opened 2018-02-27
Should HttpSysDefaults.AuthenticationScheme be a const string?


withkittens:
Right now it is a `readonly` field and is disallowed to use in a `const` context. Also this is inconsistency with other authentication packages like `JwtBearer`.

### Expected result

This should compile:

```cs
[Authorize(AuthenticationSchemes = HttpSysDefaults.AuthenticationScheme)]
public class FooController : Controller
{
    // ...
}
```

### Actual result

CS0182 compilation error: _An attribute argument must be a constant expression, typeof expression or array creation expression of an attribute parameter type._

PR #6934, opened 2019-01-22
Make auth schemes const

Tratcher:
So you can use them in the Authorize attribute. We already do this for other auth providers.
#5884 HttpSys
#6020 AzureAd

### DIFF
--- a/src/Servers/HttpSys/build.cmd
+++ b/src/Servers/HttpSys/build.cmd
@@ -1,0 +1,3 @@
+@ECHO OFF
+SET RepoRoot=%~dp0..\..\..
+%RepoRoot%\build.cmd -projects %~dp0\**\*.*proj %*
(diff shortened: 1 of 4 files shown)


JamesNK:
Label as a breaking change?

Tratcher:
@JamesNK ack

muratg:
@shirhatti breaking change, so asking for approval from PM side. Thoughts?

Tratcher:
Note these were already approved in security triage.

src/Servers/HttpSys/build.cmd
```diff
@@ -0,0 +1,3 @@
+@ECHO OFF
+SET RepoRoot=%~dp0..\..\..
+%RepoRoot%\build.cmd -projects %~dp0\**\*.*proj %*
```
muratg: Is this file change needed, or accidental?

muratg (approved):
One comment below.

src/Servers/HttpSys/build.cmd
Tratcher: Added for dev convenience, it has no impact on official builds.

PR #6935, opened 2019-01-22
Convert ProjectTemplates to build using ProjectReferences

### LINKED ISSUE #4246: Convert projects in this repo to use ProjectReference
Follow-up to #3598.

The first phase of reducing repos was to merge the source code into a single repo. The second phase is to convert them from isolated folders of code with PackageReference across repo boundaries, to projects which use `ProjectReference`.

We've already started some of this work. See https://github.com/aspnet/AspNetCore/pull/4013 for example and https://github.com/aspnet/AspNetCore/tree/master/src/DataProtection.

As done in #4013, we implement this using `<Reference>` items in .csproj and custom Reference resolution targets. This is done so we can automatically pick the right versions of assemblies to used based on conditions such as, is this a test project? is this branch building a patch or a new minor release? what versions were used in the last stable release? etc.

See also https://github.com/aspnet/specs/blob/master/runtime/design-notes/2018-10-03-mondo-source-org.md for the spec on source code organization.

## Repo groups to convert

- [x] Razor @ajaybhargavb  
- [x] HttpSysServer https://github.com/aspnet/AspNetCore/pull/4335
- [x] Antiforgery - Justin
- [x] IISIntegration https://github.com/aspnet/AspNetCore/pull/4311
- [x] CORS - Justin
- [x] StaticFiles - Justin
- [x] Routing @natemcmaster 
- [x] ResponseCaching -Justin
- [x] Session @JunTaoLuo 
- [x] ServerTests @JunTaoLuo @jkotalik 
- [x] Localization @natemcmaster 
- [x] Security @jkotalik 
- [x] MetaPackages @JunTaoLuo 
- [x] Mvc @ajaybhargavb 
- [x] JavaScriptServices @jkotalik 
- [x] AADIntegration @natemcmaster 
- [x] Identity @JunTaoLuo 
- [x] MvcPrecompilation @JunTaoLuo 
- [x] AzureIntegration @natemcmaster 
- [x] SignalR @natemcmaster 
- [x] Components @natemcmaster 
- [x] MusicStore @jkotalik 
- [x] AuthSamples @jkotalik 
- [ ] Templating @natemcmaster

natemcmaster:
Resolves #4246 

Changes:

* Moved src/Templating/src/* to `src/ProjectTemplates/`
* Stopped building the templates projects as a 'submodule'
* Tests are still skipped -- we will need to address https://github.com/aspnet/AspNetCore/issues/4321 first.


javiercn (approved):
To the degree to which I can attest them, the changes look good to me. So long as you can build a working template from .\build.cmd /t:Pack we should be good. (Having the E2E tests run would be good too).

natemcmaster:
The new command you'll need to run is `src/ProjectTemplates/build.cmd -pack` if you want to product packages for just the templates. I tested on my own and it should still work. The tricky piece is that the templates will expect packages like Mvc.NewtonsoftJson 3.0.0-preview-t000 to exist. I recommend running `$root/build.cmd -pack -all` to produce *all* packages the repo supports.

natemcmaster:
Nevermind. Just learned that we have respin the build 2 or 3 more times to consume partner dependency updates. PR checks are all green, so I'm going to merge this. I'll monitor CI to make sure this doesn't break the build.

muratg (approved):

build/buildorder.props
```diff
@@ -7,6 +7,5 @@
   </ItemDefinitionGroup>
 
   <ItemGroup>
```
ryanbrandenburg: Can we just murder this file now or should we keep it in case we have a temporary need for buildorder building?

build/repo.props
```diff
@@ -125,6 +127,10 @@
                           @(ProjectToBuild);
                           @(ProjectToExclude);" />
 
+        <!--
+          Use caution to avoid deep recursion. If the globbing pattern picks up something which exceeds MAX_PATH,
+          the entire pattern will silently fail to evaluate correctly.
```
ryanbrandenburg: Is there any way we can make it fail loudly?

build/submodules.props
```diff
@@ -36,6 +36,5 @@
   </PropertyGroup>
```
ryanbrandenburg: I assume since `TestsRequiredTheSharedRuntime` is still above that we didn't solve that problem in this PR?

ryanbrandenburg: Otherwise should we consider this file for deletion?

ryanbrandenburg (commented):
I had a couple minor items we might want to file followups for depending, but otherwise this looks awesome.

build/buildorder.props
natemcmaster: Yes, we can murder this now, along with a lot of other code. See https://github.com/aspnet/AspNetCore/pull/6995

build/submodules.props
natemcmaster: https://github.com/aspnet/AspNetCore/issues/4321 is still unresolved. I have one change in the works that addresses part of the problem, but it's a tricky enough challenge that I didn't tackle it completely.

PR #6936, opened 2019-01-22
Improvements to event handling

rynowak:
The changes here make event dispatching (including bind) more
user-friendly by avoiding the need for manual calls to StateHasChnaged.

We also introduce a new type `EventCallback` (and `EventCallback<T>`).
This is a new primitive that is like a super-powered version of a
delegate. When writing a component that accepts delegates as parameters,
consider using `EventCallback` for the following reasons:
- Allows consumer to pass a variety of different delegate signatures
- Does proper event dispatching and error handling

Using `EventCallback` will eliminate most of the remaining cases where a
manual `StateHasChanged` is required when components are passing content
and delegates to each other.

`EventCallback` is inherently async for the reason that this is really
the only way to provide correct error handling.

-----

The fix for this will be two-phase by first creating a set of APIs that
can be targeted by the compiler that has the desired behaviour and then
updating the compiler to target this new infrastructure.


src/Components/Components/src/EventHandlerInvokerFactory.cs
```diff
@@ -0,0 +1,102 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Components
+{
+    public sealed class EventHandlerInvokerFactory
```
rynowak: Still very much a work in progress, but this should give you the gist of it. 

This is an instance (not a static) so it can be targeted by extension methods.

src/Components/Components/src/EventHandlerInvokerFactory.cs
```diff
@@ -0,0 +1,102 @@
...
+            //
+            // We want to make sure to call StateHasChanged on the Outer component instead
+            // of just the Inner component. We can't rely on action.Target to always point
+            // to the outer component because in the case of a non-capturing lambda it won't.
+            if (receiver is IHandleStateChange handler && !object.ReferenceEquals(receiver, action.Target))
+            {
+                return (Action)(() =>
+                {
+                    action();
+                    _ = handler.HandleStateChangeAsync(Task.CompletedTask);
```
rynowak: Note also that this is kinda fire-and-forget. I'm not sure it makes much sense for this method to return a task, because in many many cases we won't be able to ever await it.

src/Components/Components/src/ComponentBase.cs
```diff
@@ -276,41 +308,6 @@ private async Task ReRenderAsyncTask(Task task)
             }
         }
 
-        private async void ContinueAfterLifecycleTask(Task task)
```
javiercn: Glad to see this go away

src/Components/Blazor/Blazor/src/Rendering/WebAssemblyRenderer.cs
```diff
@@ -32,7 +32,10 @@ public WebAssemblyRenderer(IServiceProvider serviceProvider): base(serviceProvid
         }
 
         internal void DispatchBrowserEvent(int componentId, int eventHandlerId, UIEventArgs eventArgs)
-            => DispatchEvent(componentId, eventHandlerId, eventArgs);
+        {
+            // Event dispatching from the browser is fire-and-forget
+            _ = DispatchEventAsync(componentId, eventHandlerId, eventArgs);
```
javiercn: I expect things like error handling to happen at this layer, so either the Renderer does it (on the base class) or we have some code here to at least log to the console or similar (equivalent to what we used to do)

javiercn:
I'm not sure I understand the nature of the change. Is this intended to support something like?
```
<div>
   <Child OnClick="InvokeSomeParentFunction" />
</div>
```
And having StateHasChanged be called on automatically after InvokeSomeParentFunction?

src/Components/Components/src/IHandleStateChange.cs
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Components
+{
+    /// <summary>
+    /// Interface implemented by components that receive notification of state changes.
```
javiercn: This feels confusing/wrong programming model wise. As far as I understand, state is internal to the component and the component is responsible for triggering renders when its state changes and it needs to rerender.

src/Components/Components/src/ComponentBase.cs
```diff
@@ -322,15 +319,18 @@ private static void HandleException(Exception ex)
             Console.Error.WriteLine($"[{ex.GetType().FullName}] {ex.Message}\n{ex.StackTrace}");
         }
 
-        void IHandleEvent.HandleEvent(EventHandlerInvoker binding, UIEventArgs args)
+        Task IHandleStateChange.HandleStateChangeAsync(Task task)
         {
-            var task = binding.Invoke(args);
-            ContinueAfterLifecycleTask(task);
+            var (isAsync, asyncTask) = ProcessLifeCycletask(task);
```
javiercn: I imagine you are doing this to rerender on async work. You should do just that. Given that you also seem to agree that event handling needs to be task returning, we should just flow the errors out of the component into the renderer where they can be handled in a centralized place

src/Components/Components/src/EventHandlerInvokerFactory.cs
javiercn: I think all the above methods need to have Func<Task> based overloads so that we can bind async handlers and flow the task all the way up to the renderer decide what to do

src/Components/Components/src/IHandleStateChange.cs
javiercn: I think I understand now what you are trying to accomplish here. I think this poses the issue that the component might rerender in many cases even if nothing changes, given that it doesn't have a good way to compare its previous state against the current state, and that might introduce performance penalties.

src/Components/Components/src/EventHandlerInvokerFactory.cs
rynowak: Yes, that will be done as well.

rynowak:
> I'm not sure I understand the nature of the change. Is this intended to support something like?
> 
> ```
> <div>
>    <Child OnClick="InvokeSomeParentFunction" />
> </div>
> ```
> And having StateHasChanged be called on automatically after InvokeSomeParentFunction?

Yes, exactly this.

src/Components/Components/src/IHandleStateChange.cs
rynowak: The runtime needs to notify the component **somehow** that an event happened. This is the somehow - it's up to the component whether it wants to rerender or not. By default the answer is yes.

rynowak:
I've rebased this on top of the current bits and incorporated some of the things from @pranavkm 's work so that we're not fighting each other. I assume his thing is going to go in first. 

My next steps are to incorporate feedback from before London and then make this more complete.

src/Components/Components/src/EventHandlerInvoker.cs
```diff
@@ -11,24 +11,30 @@ namespace Microsoft.AspNetCore.Components
     /// </summary>
     public readonly struct EventHandlerInvoker
     {
-        private readonly MulticastDelegate _delegate;
+        public static readonly EventHandlerInvokerFactory Factory = new EventHandlerInvokerFactory();
```
rynowak: @SteveSandersonMS  - what would you think about making this a more user-friendly name and making it part of the programming experience. I'm thinking `Command` and `Command<T>` and keeping the ability to abstract over the async-vs-non.

The reason why you might want to have something like this is to deal with the problem when a component wants to use delegates as parameters. The consumer of the component has to provide delegates with the signature that the component specifies. This is a little clunky in comparison to how event-handlers work which accept a variety of signatures seamlessly. This would be done in such a way that the user of a component doesn't need to know about the details of command, a super-set of what we have today would work. 

The other detail is that if a component accepts a `Command` as a parameter, it allows us to back in the same eventing behaviour that we have with event handlers and bind but without guesswork and conventions.

src/Components/Components/src/IHandleStateChange.cs
```diff
@@ -0,0 +1,25 @@
...
+        /// Notifies the a state change has been triggered.
+        /// </summary>
+        /// <param name="task">
+        /// The <see cref="Task"/> whose asynchronous completion represents the compeletion of the state
+        /// change.
+        /// </param>
+        /// <returns>
+        /// A <see cref="Task"/> that completes once the component has processed the state change.
+        /// </returns>
+        Task HandleStateChangeAsync(Task task);
```
rynowak: Per discussion this is going to be made `void HandleStateChange()`

src/Components/Components/src/Rendering/Renderer.cs
```diff
@@ -536,6 +545,22 @@ private void RemoveEventHandlerIds(ArrayRange<int> eventHandlerIds, Task afterTa
...
+            }
+            catch (Exception ex)
+            {
+                if (!taskToHandle.IsCanceled)
+                {
+                    // Ignore errors due to task cancellations.
+                    HandleException(ex);
+                }
+            }
+        }
```
rynowak: copied from @pranavkm

src/Components/Components/src/Rendering/Renderer.cs
```diff
@@ -329,8 +339,7 @@ public virtual Task InvokeAsync(Func<Task> workItem)
             // This is for example when we run on a system with a single thread, like WebAssembly.
             if (_dispatcher == null)
             {
-                workItem();
-                return Task.CompletedTask;
+                return workItem();
```
rynowak: This was a bug that broke an E2E test. I'm not sure how we didn't detect this somehow.

src/Components/Components/src/Rendering/Renderer.cs
```diff
@@ -267,24 +267,36 @@ private ComponentState AttachAndInitComponent(IComponent component, int parentCo
...
+                Task task = null;
                 try
                 {
                     _isBatchInProgress = true;
-                    GetRequiredComponentState(componentId).DispatchEvent(binding, eventArgs);
+                    task = componentState.DispatchEventAsync(binding, eventArgs);
+                }
+                catch (Exception ex) when (!task.IsCanceled)
+                {
+                    HandleException(ex);
```
SteveSandersonMS: It took me a few minutes to realise what scenario L291 is for. Would something like the following make the flow more obvious, or am I misunderstanding the intended logic?

```cs
catch (Exception synchronousException)
{
    task = Task.FromException(synchronousException);
}
```

... so that the flow always continues to L299, and we don't need any extra special case for `IsCanceled`?

src/Components/Components/test/RendererTest.cs
```diff
@@ -436,12 +436,15 @@ public void CanDispatchEventsToTopLevelComponents()
 
             // Act/Assert: Event can be fired
             var eventArgs = new UIEventArgs();
-            renderer.DispatchEvent(componentId, eventHandlerId, eventArgs);
+            var task = renderer.DispatchEventAsync(componentId, eventHandlerId, eventArgs);
             Assert.Same(eventArgs, receivedArgs);
+            Assert.Equal(TaskStatus.RanToCompletion, task.Status); // Completes synchronously
+
+            await task; // Does not throw
```
SteveSandersonMS: This is a great way to express the flow, showing that the effects are synchronous plus the task itself doesn't fail. 👍

src/Components/Components/test/RendererTest.cs
```diff
@@ -946,7 +1160,7 @@ public void AllRendersTriggeredSynchronouslyDuringEventHandlerAreHandledAsSingle
             Assert.Single(renderer.Batches);
 
             // Act
-            renderer.DispatchEvent(childComponentId, origEventHandlerId, args: null);
+            await renderer.DispatchEventAsync(childComponentId, origEventHandlerId, eventArgs: null);
```
SteveSandersonMS: I think having an `await` here somewhat undermines the idea of this test, which is about what happens synchronously.

Would it be possible *not* to await this task? Or perhaps capture it and put `await capturedTask` at the end of the test.

Ideally it would be preferable to do that for all the tests being updated here, since we explicitly want the effects to remain synchronous.

src/Components/Shared/test/TestRenderer.cs
```diff
@@ -43,21 +40,8 @@ public new Task RenderRootComponentAsync(int componentId)
...
-                var exception = t.Exception.Flatten().InnerException;
-                while (exception is AggregateException e)
-                {
-                    exception = e.InnerException;
-                }
-                ExceptionDispatchInfo.Capture(exception).Throw();
-            }
-        }
+        public new Task DispatchEventAsync(int componentId, int eventHandlerId, UIEventArgs eventArgs)
+            => InvokeAsync(() => base.DispatchEventAsync(componentId, eventHandlerId, eventArgs));
```
SteveSandersonMS: Is it necessary to pass the logic through `InvokeAsync` here? I'm concerned this might make the tests describe something different than how the real renderers would behave, since when dispatching events to the real renderers we don't force it through an async hop.

src/Components/Components/src/ComponentBase.cs
```diff
@@ -157,6 +157,38 @@ protected Task Invoke(Action workItem)
...
+        /// <summary>
+        /// Dispatches a state change notification to the provided component.
+        /// </summary>
+        /// <param name="component">The component instance.</param>
+        /// <param name="invoker">The <see cref="EventHandlerInvoker"/></param>
+        /// <param name="args">The event args.</param>
+        /// <returns>
+        /// A <see cref="Task" /> that asynchronously completes once event processing has completed.
+        /// </returns>
+        public static Task DispatchStateChangeAsync(object component, EventHandlerInvoker invoker, UIEventArgs args)
```
SteveSandersonMS: Could the public versions of `DispatchStateChangeAsync` take an `IComponent component`, instead of `object component`? Seems like that would make the usage more explicit. I don't think there's any scenario where we expect people to dispatch to something that's not a component.

It's fine if there's also an internal overload that takes `object component` so we don't have to put casts and checks into all the `SetValueHandler` calls.

SteveSandersonMS: Or even `IHandleStateChange` instead of `IComponent`.

src/Components/Components/src/BindMethods.cs
```diff
@@ -74,129 +74,187 @@ public static MulticastDelegate GetEventHandlerValue<T>(Func<T, Task> value)
         /// </summary>
         public static Action<UIEventArgs> SetValueHandler(Action<string> setter, string existingValue)
         {
-            return _ => setter((string)((UIChangeEventArgs)_).Value);
+            return _ =>
+            {
+                setter((string)((UIChangeEventArgs)_).Value);
+                ComponentBase.DispatchStateChangeAsync(setter.Target, Task.CompletedTask);
```
SteveSandersonMS: Unless I'm missing something, it looks like this will still suffer from the same problem that killed the approach attempted in https://github.com/aspnet/Blazor/pull/1651, i.e., that `delegate.Target` won't be what we need if it's a capturing lambda.

Is that something you already have on the list to address for this PR?

The plan we previously came up with to fix this was https://github.com/aspnet/Blazor/pull/1651#issuecomment-438036193. It involves changing the codegen for events and bind.

src/Components/Components/src/ComponentBase.cs
SteveSandersonMS: Also the overload that takes `(object component, EventHandlerInvoker invoker, UIEventArgs args)` seems only to be used in one place, and that place could call `invoker.Invoke(args)` itself then use the `(object component, Task task)` overload. So maybe we only need the second overload.

src/Components/Components/src/IHandleStateChange.cs
SteveSandersonMS: I do agree with @javiercn, though @rynowak of course you're right that we need some way to signal that an event happened.

Perhaps we should change the name. Instead of `IHandleStateChange`, something like `IHandleAfterEvent`, since we're not really saying any state has changed, but just that some event has happened.

Actually I'm looking for all the places where we call `HandleStateChangeAsync` with some task other than `Task.CompletedTask`, and there's only one such place: `ComponentBase.cs` L171. That being the case, might it be possible to keep the earlier interface where we expect to be given `EventHandlerInvoker invoker` and `UIEventArgs args`, and then the recipient is the thing responsible for actually calling the `invoker`? If we could do that, it pretty much completely eliminates the risk that this interface gets misused as a way to invoke `StateHasChanged` on arbitrary components outside the intended lifecycle, preserving the encapsulation so that components know they only trigger that themselves (or the system does).

Perhaps there are other cases where we actually have to start with an existing `Task` rather than an `EventHandlerInvoker`. What do you think?

SteveSandersonMS: Oh actually I guess there's always the scenario like in `SetValueHandler` where there's no delegate to pass, but rather we're just supplying notification that an event has already happened. So maybe you can ignore the second half of my suggestions there.

But I do still think `IHandleAfterEvent` might be a better name than `IHandleStateChange`.

src/Components/Components/src/Rendering/ComponentState.cs
```diff
@@ -97,20 +97,6 @@ public void DisposeInBatch(RenderBatchBuilder batchBuilder)
...
-        {
-            if (_component is IHandleEvent handleEventComponent)
-            {
-                return handleEventComponent.HandleEventAsync(binding, eventArgs);
-            }
-            else
-            {
-                throw new InvalidOperationException(
-                    $"The component of type {_component.GetType().FullName} cannot receive " +
-                    $"events because it does not implement {typeof(IHandleEvent).FullName}.");
```
SteveSandersonMS: **Super** pleased to see this nonsense go away.

src/Components/Components/test/RendererTest.cs
```diff
@@ -732,46 +731,15 @@ public async Task CanAsyncDispatchEventsToNestedComponents()
         }
 
         [Fact]
-        public async Task ThrowsIfComponentDoesNotHandleEvents()
```
SteveSandersonMS: Yay for removing this!

src/Components/Shared/test/TestRenderer.cs
```diff
@@ -40,12 +40,17 @@ public new Task RenderRootComponentAsync(int componentId)
...
 
-        public new Task DispatchEventAsync(int componentId, int eventHandlerId, UIEventArgs eventArgs)
-            => InvokeAsync(() => base.DispatchEventAsync(componentId, eventHandlerId, eventArgs));
+        public new Task DispatchEventAsync(int eventHandlerId, UIEventArgs args)
+            => InvokeAsync(() => base.DispatchEventAsync(eventHandlerId, args));
 
         public T InstantiateComponent<T>() where T : IComponent
             => (T)InstantiateComponent(typeof(T));
 
+        public Task WaitForAsyncWork()
```
SteveSandersonMS: This doesn't seem to be used. Can it be removed?

src/Components/Components/src/BindMethods.cs
SteveSandersonMS: OK, I see you're working on addressing this in the commit titled `add event handler infrastructure`.

src/Components/Components/src/EventHandlerInvokerFactory.cs
```diff
@@ -0,0 +1,102 @@
...
+            //
+            //  @* in Inner.cshtml *@
+            //  <button onclick="@OnClick">Click Me!</button>
+            //  @functions {
+            //      [Parameter] Action OnClick { get; set; }
+            //  }
+            //
+            // We want to make sure to call StateHasChanged on the Outer component instead
+            // of just the Inner component. We can't rely on action.Target to always point
+            // to the outer component because in the case of a non-capturing lambda it won't.
```
SteveSandersonMS: Nit: it's the capturing (not non-capturing) case where delegate.Target won't be `Outer`.

src/Components/Components/src/EventHandlerInvoker.cs
SteveSandersonMS: This sounds superb. I was just hitting cases today with the forthcoming `EditForm` component where it's awkward to have an `OnSubmit` parameter because sometimes the consumer might legitimately want to supply any of `Action`, `Action<EditContext>`, `Func<Task>` or `Func<EditContext, Task>`.

If this could be declared by `EditForm` as `[Parameter] Command<EditContext> OnSubmit { get; set; }` and this makes it possible for the consumer to use any of those delegate types, that would be a huge win.

On naming, the name `Command` might be a good fit, or `EventCallback` is also pretty explict since it comes with the implication that it triggers post-event handling logic.

src/Components/Components/src/EventHandlerInvokerFactory.cs
```diff
@@ -0,0 +1,102 @@
...
+            {
+                throw new ArgumentNullException(nameof(receiver));
+            }
+
+            if (action == null)
+            {
+                throw new ArgumentNullException(nameof(action));
+            }
+
+            if (receiver is IHandleStateChange handler && !object.ReferenceEquals(receiver, action.Target))
```
SteveSandersonMS: Ultra-nit: the most common case will be `receiver == action.Target`, so we could swap the order of these conditionals for an incredibly minor perf gain.

src/Components/Components/test/RendererTest.cs
```diff
@@ -730,6 +730,160 @@ public async Task CanAsyncDispatchEventsToNestedComponents()
             Assert.Equal(2, state);
         }
 
+        // The delegate target points to the outer component
+        [Fact]
+        public async Task EventDispatching_DelegateParameter_MethodToDelegateConversion()
```
SteveSandersonMS: Is the point of this test to show that it's unnecessary to use `CreateDelegate` if `action.Target` already points to the component you want to be notified after the event?

If so that's great, but maybe add a comment to that effect, as it's pretty subtle!

SteveSandersonMS:
This is all looking extremely positive. Very excited to finally squash the long-standing events weirdnesses, especially as it makes the forms-and-validation scenarios work properly too.

src/Components/Components/src/Rendering/Renderer.cs
SteveSandersonMS: Actually it probably does still need special handling for `IsCanceled` regardless.

rynowak: I based this on @pranavkm 's pr so I don't want to super-diverge from what's going on there. If we think this needs to change that's fine. 

Would comments make a big difference here? I don't relish the thought of allocating an extra task just for funsies.

src/Components/Components/test/RendererTest.cs
rynowak: Yeah good catch. I was trying to preserve the semantics of this but I misread the test name. I'll review these again.

src/Components/Components/src/Rendering/Renderer.cs
SteveSandersonMS: Sure, a comment would be totally sufficient.

src/Components/Shared/test/TestRenderer.cs
rynowak: Now that we require the use the sync context events have to be dispatched on the sync context. The renderer would already be running on the sync context when `DispatchEventAsync` is called.

So I guess I'm not sure what else this would do? 

Another *very intrusive* approach is a write a test framework that has a sync context, and then just ignore threading in your tests. We have done this a few places, but I don't really relish it here.  When you do this, a different set of test scenarios become hard.

src/Components/Components/src/ComponentBase.cs
rynowak: Sure. I cribbed this from your original PR - but I think the semantics have changed. You originally had this typed as object or `IComponent` so that callers could at least attempt to dispatch a change manually with minimal ceremony.

Another consideration is whether we want to allow AppState objects to implement `IHandleStateChange`.

src/Components/Components/src/BindMethods.cs
rynowak: Right. This isn't the fix for that, but these APIs need to change to preserve the existing semantics. These APIs needs to stay put because we still call them in the editor. It's the simplest thing to do to try and keep these APIs working until we completely transition.

src/Components/Components/src/EventHandlerInvokerFactory.cs
rynowak: @javiercn - note that in come cases these event handlers are fire-and-forget. Nothing awaits them and nothing can.

src/Components/Components/test/RendererTest.cs
rynowak: I've added comments to all of these new tests.

src/Components/Shared/test/TestRenderer.cs
rynowak: done

src/Components/Blazor/Blazor/src/Rendering/WebAssemblyRenderer.cs
rynowak: The error handling is inside `DispatchEventAsync`

src/Components/Components/src/EventHandlerInvokerFactory.cs
rynowak: I think maybe we're just disagreeing on terminology maybe?

** Capturing Lambda**

```html
<MdButton OnClicked="@(() => DoSomething())" />

@functions {
    void DoSomething() { ... }
}
```


**Non-capturing Lambda**

```html
<MdButton OnClicked="@(() => Console.Log("YEET"))" />
```


**Bound Delegate**

```html
<MdButton OnClicked="DoSomething" />

@functions {
    void DoSomething() { ... }
}
```

src/Components/Components/src/ComponentBase.cs
rynowak: I'm not sure I really get what the suggestion is here. But given that we're changing the signature of this method is - I think it's going to be unimportant. Holla back if you still have feedback after I update this.

src/Components/Components/src/IHandleStateChange.cs
rynowak: I'm going to change this to be void and called `IHandleAfterEvent` per feedback.

src/Components/Components/src/ComponentBase.cs
rynowak: I think I'd prefer to leave this internal or let it accept object. There are **a lot** of call sites of this, and it doesn't seem useful to expand the boilerplate into all of those locations.

src/Components/Components/src/IHandleStateChange.cs
rynowak: I ended up changing this back to its original signature. The problem is that if we want the component to return a task, it's helpful for the component to be the one to invoke the delegate. Other solutions (unless we made it synchronous) become *really* complicated.

src/Components/Components/src/EventHandlerInvoker.cs
rynowak: If we want this to support arbitrary T it will require some doing, but shouldn't be that bad. The main that comes up is that it kinda has to be reference type at that point. This might be OK because we can amortize some of the allocations we already have to do - I'm going to have a go at this and we can look at in more detail. 

I want to spend some extra time to really nail this part since I think it nice to have a primitive that we use to events, bind, and arbitrary app functionality like `OnSubmit`.

rynowak:
I've pushed an update after incorporating pranav's change but it's not close to done yet. I'm going deep on a few of the specific details of feedback, mostly the async and command ideas. 

One of the major issues to solve is that most of our current stuff is "fire and forget" in nature, and to address that we need to make everything be async.

src/Components/Components/src/ComponentBase.cs
rynowak: This is gone now, so not super relevant.

rynowak:
@SteveSandersonMS @javiercn @pranavkm - this is no longer work in progress. Please take a look. meanwhile I'll get busy on the compiler changes, but there's no reason to wait on completing this first since it doesn't really regress things.

src/Components/Blazor/Build/test/RenderingRazorIntegrationTest.cs
```diff
@@ -361,26 +362,27 @@ public void SupportsTwoWayBindingForTextboxes()
...
-                    {
-                        Value = "Modified value"
-                    });
-                    Assert.Equal("Modified value", myValueProperty.GetValue(component));
+                    setter = Assert.IsType<Action<UIEventArgs>>(frame.AttributeValue);
                 });
+
+            // Trigger the change event to show it updates the property
+            await renderer.Invoke(() => setter(new UIChangeEventArgs { Value = "Modified value", }));
+            Assert.Equal("Modified value", myValueProperty.GetValue(component));
```
rynowak: Now that I've *improved* some of the bind functionality with polyfills these tests require dispatching to the sync context.

src/Components/Browser.JS/src/package-lock.json
```diff
@@ -1525,7 +1525,8 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true
+          "dev": true,
+          "optional": true
```
rynowak: Will undo this.

src/Components/Components/src/BindMethods.cs
```diff
@@ -69,134 +69,200 @@ public static MulticastDelegate GetEventHandlerValue<T>(Func<T, Task> value)
...
         /// <summary>
         /// Not intended to be used directly.
         /// </summary>
         public static Action<UIEventArgs> SetValueHandler(Action<bool> setter, bool existingValue)
         {
-            return _ => setter((bool)((UIChangeEventArgs)_).Value);
+            return eventArgs =>
+            {
+                setter((bool)((UIChangeEventArgs)eventArgs).Value);
+                _ = DispatchEventAsync(setter.Target, EventCallbackWorkItem.Empty, UIEventArgs.Empty);
```
rynowak: Explained this in a comment elsewhere but for visibility. Yes, this isn't *super* correct. It's a temporary polyfill until we swap over to the new infrastructure. This is needed before of the change to event dispatching in the renderer. If we don't do something like this we're going to *chicken-or-egg* super hard 🐔

src/Components/Components/src/EventCallbackFactory.cs
```diff
@@ -0,0 +1,211 @@
...
+using System;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Components
+{
+    /// <summary>
+    /// A factory for creating <see cref="EventCallback"/> and <see cref="EventCallback{T}"/>
+    /// instances.
+    /// </summary>
+    public sealed class EventCallbackFactory
```
rynowak: Instance class to support extension methods. Users will be able to polyfill their own support for `bind` with arbitrary types once compiler work is done.

src/Components/Components/src/EventCallbackFactoryBinderExtensions.cs
```diff
@@ -0,0 +1,403 @@
...
+    {
+        /// <summary>
+        /// For internal use only.
+        /// </summary>
+        /// <param name="factory"></param>
+        /// <param name="receiver"></param>
+        /// <param name="setter"></param>
+        /// <param name="existingValue"></param>
+        /// <returns></returns>
+        public static EventCallback<UIChangeEventArgs> CreateBinder(
```
rynowak: Replacement for old `BindMethods`

src/Components/Components/src/EventCallbackFactoryUIEventArgsExtensions.cs
```diff
@@ -0,0 +1,446 @@
...
+    {
+        /// <summary>
+        /// Creates an <see cref="EventCallback"/> for the provided <paramref name="receiver"/> and
+        /// <paramref name="callback"/>.
+        /// </summary>
+        /// <param name="factory">The <see cref="EventCallbackFactory"/>.</param>
+        /// <param name="receiver">The event receiver.</param>
+        /// <param name="callback">The event callback.</param>
+        /// <returns>The <see cref="EventCallback"/>.</returns>
+        public static EventCallback<UIEventArgs> Create(this EventCallbackFactory factory, object receiver, Action<UIEventArgs> callback)
```
rynowak: These overloads are needed so that method-group-to-delegate conversion will work well. Yeah, it's fun

src/Components/Components/src/EventCallbackWorkItem.cs
```diff
@@ -0,0 +1,79 @@
...
+using System;
+using System.Reflection;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Components
+{
+    /// <summary>
+    /// Wraps a callback delegate associated with an event.
+    /// </summary>
+    public struct EventCallbackWorkItem
```
rynowak: Suggestions welcome. We want to find an unappealing name for this time so you're likely to use `EventCallback` which is what we want users to find. We could put this in another namespace, but none of the ones we have already are really appropriate. Ultimately we might want to more component lifecycle interfaces elsewhere, and this type would follow.

src/Components/Components/src/RenderTree/RenderTreeBuilder.cs
```diff
@@ -277,6 +277,84 @@ public void AddAttribute(int sequence, string name, MulticastDelegate value)
...
+            {
+                // Since this is a component, we need to preserve the type of the EventCallback, so we have
+                // to box.
+                Append(RenderTreeFrame.Attribute(sequence, name, (object)value));
+            }
+            else if (value.RequiresExplicitReceiver)
+            {
+                // If we need to preserve the receiver - we convert this to an untyped EventCallback. We don't
+                // need to preserve the type of an EventCallback<T> when it's invoked from the DOM.
+                Append(RenderTreeFrame.Attribute(sequence, name, (object)value.AsUntyped()));
```
rynowak: Note: this is a special case. It would be hard to make `RenderTreeDiffBuilder` handle `EventCallback<T>` + DOM events. However we have no such requirement, so I just skip it.

src/Components/Components/src/Rendering/RendererSynchronizationContext.cs
```diff
@@ -47,6 +47,10 @@ public Task Invoke(Action action)
                     action();
                     completion.SetResult(null);
                 }
+                catch (OperationCanceledException)
+                {
+                    completion.SetCanceled();
+                }
```
rynowak: This was a bug in the sync context that I found while unit testing related features. We weren't honouring cancellation and would always return a cancelled task as a faulted task instead.

I unwound some of the test infrastructure that had built up and then found this bug. We should be more careful creating "helpers" instead of writing primitive code.

src/Components/test/testassets/BasicTestApp/EventCallbackTest/EventCallbackCases.cshtml
```diff
@@ -0,0 +1,36 @@
+@*
+    Test cases for using EventCallback with various delegate scenarios that used to be troublesome.
+
+    Currently these cases are **VERBOSE** because we haven't yet landed the compiler support for EventCallback.
+    This will be cleaned up soon, and all of the explicit calls to EventCallback.Factory will go away.
```
rynowak: Right. So before you complain, I already explained what's going on 😆. Once the compiler work is done all of the rest of this code in the file will become idiomatic.

src/Components/test/testassets/BasicTestApp/EventCallbackTest/EventCallbackCases.cshtml
```diff
@@ -0,0 +1,36 @@
...
+<div id="unbound_lambda_strongly_typed">
+    <h3>Passing Capturing Lambda to Strongly Typed Button</h3>
+    <StronglyTypedButton OnClick="@(EventCallback.Factory.Create<UIMouseEventArgs>(this, () => { GC.KeepAlive(this); }))" Text="Unbound Lambda Strongly-Typed" />
+</div>
+<div id="unbound_lambda_child_content">
+    <h3>Passing Child Content</h3>
+    <TemplatedControl>
+        <button onclick="@EventCallback.Factory.Create(this, () => { })">Unbound Lambda Child Content</button>
+    </TemplatedControl>
+</div>
```
rynowak: @SteveSandersonMS anything else you think is worth adding?

src/Components/Components/src/EventHandlerInvokerFactory.cs
SteveSandersonMS: OK yeah. I was distinguishing based on whether the compiler would generate a hidden class to hold captured variables, which it does if and only if you are capturing something *other than* `this`, e.g.:

```html
@foreach(var thing in myCollection)
{
    <MdButton OnClicked="@(() => DoSomething(thing))" />
}
```

In this case, the delegate's `Target` will be an instance of the compiler-generated class instead of the Razor Component.

src/Components/Blazor/Build/test/RenderingRazorIntegrationTest.cs
SteveSandersonMS: But events are still processed synchronously, aren't they? They absolutely have to be in the WebAssembly case, otherwise we'll kill certain APIs that are only allowed synchronously.

Is is still possible to have unit tests that show event processing happens synchronously?

SteveSandersonMS: Is it OK because in the WebAssembly case, dispatched calls always get executed synchronously?

src/Components/Components/src/BindMethods.cs
```diff
@@ -69,134 +69,200 @@ public static MulticastDelegate GetEventHandlerValue<T>(Func<T, Task> value)
...
+        }
+
+        /// <summary>
+        /// Not intended to be used directly.
+        /// </summary>
+        public static EventCallback<T> GetEventHandlerValue<T>(EventCallback<T> value)
+            where T : UIEventArgs
+        {
+            return value;
+        }
```
SteveSandersonMS: I'm assuming the above two are for type checking in the Razor compiler.

src/Components/Components/src/BindMethods.cs
SteveSandersonMS: I haven't completely followed in what sense this isn't correct. Will try to see if the comment elsewhere enlightens me.

src/Components/Components/src/ComponentBase.cs
```diff
@@ -255,9 +255,9 @@ private async Task CallStateHasChangedOnAsyncCompletion(Task task)
             StateHasChanged();
         }
 
-        Task IHandleEvent.HandleEventAsync(EventHandlerInvoker binding, UIEventArgs args)
+        Task IHandleAfterEvent.HandleEventAsync(EventCallbackWorkItem callback, object arg)
         {
-            var task = binding.Invoke(args);
+            var task = callback.InvokeAsync(arg);
```
SteveSandersonMS: Since it actually does invoke the callback itself (instead of just doing stuff *after* the callback was invoked), would the original name `IHandleEvent` actually be a better description?

src/Components/Components/src/EventCallback.cs
```diff
@@ -0,0 +1,115 @@
...
+
+using System;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Components
+{
+    /// <summary>
+    /// A bound event handler delegate.
+    /// </summary>
+    public readonly struct EventCallback
```
SteveSandersonMS: This is such a good feature!

src/Components/Components/src/EventCallbackFactory.cs
SteveSandersonMS: Excellent

src/Components/Components/src/EventCallbackFactoryBinderExtensions.cs
```diff
@@ -0,0 +1,403 @@
...
+            this EventCallbackFactory factory,
+            object receiver,
+            Action<string> setter,
+            string existingValue)
+        {
+            Func<object, string> converter = (obj) => (string)obj;
```
SteveSandersonMS: Does this method get called by generated code at each `bind` call site?

If so, instead of instantiating a new `converter` on each call, can we cache the `Func<object, *>` instances on static fields for reuse?

SteveSandersonMS: Here and in the other `CreateBinder` methods, that is.

src/Components/Components/src/EventCallbackWorkItem.cs
SteveSandersonMS: I think this name sounds unappealing enough as-is.

src/Components/Components/src/Rendering/Renderer.cs
```diff
@@ -338,11 +343,28 @@ internal void AssignEventHandlerId(ref RenderTreeFrame frame)
...
             var id = ++_lastEventHandlerId;
 
-            if (frame.AttributeValue is MulticastDelegate @delegate)
+            if (frame.AttributeValue is EventCallback callback)
+            {
+                // We hit this case when a EventCallback object is produced that needs an explicit receiver.
+                // Common cases for this are "chained bind" or "chained event handler" when a component
+                // accepts a delegate as a parameter and then hooks it up to a DOM event.
+                //
+                // When that happens we intentionally box the Command because we need to hold on to
```
SteveSandersonMS: In the comment, `Command` -> `EventCallback`?

src/Components/Components/test/EventCallbackTest.cs
```diff
@@ -0,0 +1,457 @@
...
+        public async Task EventCallbackOfT_Default()
+        {
+            // Arrange
+            var callback = default(EventCallback<UIEventArgs>);
+
+            // Act & Assert (Does not throw)
+            await callback.InvokeAsync(null);
+        }
+
+
```
SteveSandersonMS: Nit: unnecessary blank line here, and in a few other places in this file. I don't really mind whether it gets fixed. I'm kind of only posting this as proof that I'm actually paying attention.

src/Components/test/testassets/BasicTestApp/EventCallbackTest/EventCallbackCases.cshtml
SteveSandersonMS: Understood!

src/Components/test/testassets/BasicTestApp/EventCallbackTest/EventCallbackCases.cshtml
SteveSandersonMS: When you do `bind-Value` to a component, does the `ValueChanged` invocation now implicitly trigger `StateHasChanged` on the outer component? Not sure whether that is meant to happen automatically given the codegen changes, but if so, would be good to show that too.

SteveSandersonMS (approved):
This looks superb! And a very impressive level of test thoroughness too.

src/Components/Components/src/BindMethods.cs
rynowak: What is incorrect about is the error handling. It calls `DispatchEventAsync` as fire-and-forget.

src/Components/Blazor/Build/test/RenderingRazorIntegrationTest.cs
rynowak: Yes, everything still happens synchronously unless you override `OnHandleEventAsync` and do something async there before calling the delegate.

src/Components/test/testassets/BasicTestApp/EventCallbackTest/EventCallbackCases.cshtml
rynowak: > When you do `bind-Value` to a component, does the `ValueChanged` invocation now implicitly trigger `StateHasChanged` on the outer component?

Using `Action<T>` will always trigger an event on the delegate target (if possible). We'd recommend using `EventCallback<T>` for this case, which will be supported. 

I'll add a test case for this

src/Components/Components/src/BindMethods.cs
rynowak: exactly.

src/Components/Components/src/ComponentBase.cs
rynowak: Sure, but this will be my last iteration of renaming this thing until we do an all-up api review.

src/Components/Components/src/EventCallbackFactoryBinderExtensions.cs
rynowak: If you'd feel better I can move these to static fields so that it's obvious. 

The compiler already does this optimization where possible. There are different preferences for how deal with this problem when you want to be *sure* that the lambda is static, and they all have drawbacks, other than static local functions which requires C# 8.

src/Components/Components/test/EventCallbackTest.cs
rynowak: Cools beans. I might leave this here, Pranav will be overjoyed if he gets the chance to remove it later.

src/Components/test/testassets/BasicTestApp/EventCallbackTest/EventCallbackCases.cshtml
rynowak: Note that this requires the component that is producing the bindable to use `EventCallback<T> ValueChanged`.

rynowak:
Helix has been running now for 2.5 hours - so I'm going to ignore it and just merge this.

PR #6950, opened 2019-01-23
Enable strict mode in roslyn compiler

pakrym:
https://github.com/aspnet/AspNetCore/issues/3645

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -22,6 +22,8 @@
     <NoWarn>$(NoWarn);NU5125</NoWarn>
     <!-- Suppress warnings about using SemVer 2.0. -->
     <NoWarn>$(NoWarn);NU5105</NoWarn>
+    <!-- Enables Strict mode for Roslyn compiler -->
+    <Features>strict</Features>
 
     <!-- Contact email address for NuGet packages and Linux installers. -->
     <MaintainerEmail>nugetaspnet@microsoft.com</MaintainerEmail>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -22,8 +22,6 @@
     <NoWarn>$(NoWarn);NU5125</NoWarn>
     <!-- Suppress warnings about using SemVer 2.0. -->
     <NoWarn>$(NoWarn);NU5105</NoWarn>
-    <!-- Enables Strict mode for Roslyn compiler -->
-    <Features>strict</Features>
 
     <!-- Contact email address for NuGet packages and Linux installers. -->
     <MaintainerEmail>nugetaspnet@microsoft.com</MaintainerEmail>
@@ -63,6 +61,9 @@
 
     <!-- Fixes a common error in targets implementing a NoBuild mode. -->
     <BuildProjectReferences Condition=" '$(NoBuild)' == 'true' ">false</BuildProjectReferences>
+    
+    <!-- Enables Strict mode for Roslyn compiler -->
+    <Features>strict</Features>
   </PropertyGroup>
 
   <!-- Certificate names used for Authenticode signing. -->
--- a/src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/CollectionModelBinder.cs
+++ b/src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/CollectionModelBinder.cs
@@ -448,17 +448,12 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding.Binders
 
         private static IEnumerable<string> GetIndexNamesFromValueProviderResult(ValueProviderResult valueProviderResult)
         {
-            IEnumerable<string> indexNames = null;
-            if (valueProviderResult != null)
+            var indexes = (string[])valueProviderResult;
+            if (indexes != null && indexes.Length > 0)
             {
-                var indexes = (string[])valueProviderResult;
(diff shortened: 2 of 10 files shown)


src/Mvc/test/WebSites/BasicWebSite/Controllers/ContactApiController.cs
```diff
@@ -131,11 +131,6 @@ private class TestModelBinder : IModelBinder
             public Task BindModelAsync(ModelBindingContext bindingContext)
             {
                 var val = bindingContext.ValueProvider.GetValue(bindingContext.ModelName);
-                if (val == null)
-                {
-                    return Task.CompletedTask;
-                }
-
```
rynowak: This is a semantic change.

pakrym: ValueProviderResult is a struct for result would always be false

Tratcher (approved):

src/Mvc/test/WebSites/BasicWebSite/Controllers/ContactApiController.cs
rynowak: This is probably a bug then? /cc @dougbu

dougbu: @rynowak what exactly do you want my thoughts on?

src/Servers/Kestrel/Core/src/Internal/Http/HttpHeaders.Generated.cs
```diff
@@ -27,7 +27,7 @@ public StringValues HeaderCacheControl
         {
             get
             {
-                StringValues value;
+                StringValues value = default;
```
dougbu: This is a generated file but I don't see changes in the generator. Suggest there should be.

pakrym: Generator change is in `src/Servers/Kestrel/tools/CodeGenerator/KnownHeaders.cs`

src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/CollectionModelBinder.cs
```diff
@@ -448,17 +448,12 @@ protected virtual void CopyToModel(object target, IEnumerable<TElement> sourceCo
 
         private static IEnumerable<string> GetIndexNamesFromValueProviderResult(ValueProviderResult valueProviderResult)
         {
-            IEnumerable<string> indexNames = null;
-            if (valueProviderResult != null)
+            var indexes = (string[])valueProviderResult;
+            if (indexes != null && indexes.Length > 0)
```
halter73: Isn't it impossible for `indexes` to be null here? `valueProviderResult` is a struct and the `(string[])` must cast either succeed or throw.

src/Mvc/test/WebSites/BasicWebSite/Controllers/ContactApiController.cs
rynowak: oh nvmt his is in test code. Just ignore then

halter73: Is it OK to remove the if condition entirely (which doesn't change the current behavior) or should be replaced with something like `if (val.Length > 0)`?

halter73 (approved):

src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/CollectionModelBinder.cs
pakrym: Might be `null` for `default(StringValues)`

Directory.Build.props
```diff
@@ -22,6 +22,8 @@
     <NoWarn>$(NoWarn);NU5125</NoWarn>
     <!-- Suppress warnings about using SemVer 2.0. -->
     <NoWarn>$(NoWarn);NU5105</NoWarn>
+    <!-- Enables Strict mode for Roslyn compiler -->
+    <Features>strict</Features>
```
natemcmaster: Nit: can you put this in the propertygroup below that starts with the comment "Compilation options"?

natemcmaster (approved):

src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/CollectionModelBinder.cs
halter73: Yowzers! Return null from an implicit conversion sounds pretty evil. We couldn't even return a static empty array?

benaadams: Could one line it?
```csharp
return (indexes == null || indexes.Length == 0) ? null : indexes;
```

Tratcher: No, that's expected to preserve transitivity. E.g. (string[])(StringValues)null;

PR #6955, opened 2019-01-23
Remove stale links to runtime installers from daily build docs

natemcmaster:
Resolves https://github.com/aspnet/AspNetCore/issues/6509.

We recommend acquiring ASP.NET Core along with a matching SDK at https://github.com/dotnet/core-sdk. Dependency flow should be regularly updating the version of ASP.NET Core bundled in the .NET Core SDK.

### DIFF
--- a/docs/DailyBuilds.md
+++ b/docs/DailyBuilds.md
@@ -5,7 +5,7 @@ Daily builds include the latest source code changes. They are not supported for 
 
 If you want to download the latest daily build and use it in a project, then you need to:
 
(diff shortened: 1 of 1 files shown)


Eilon (approved):
Makes sense.

muratg (approved):

hacst:
The debian dotnet-sdk-latest-x64.deb daily package available from https://github.com/dotnet/core-sdk/blob/master/README.md#installers-and-binaries still depends on ```aspnetcore-runtime-3.0```. It seems like before they went stale these links provided that dependency. Where can I find it now?

Issue #6105, opened 2018-12-23
Replacement token surrounded by [] causes errors


CodeMocker:
It is actually for experimental purpose only. In the production I will not use such a  tricky attribute route.

```
[Route("[controller]")]
public class FooController : Controller
{
    [HttpGet("x[action]y/{i}")]
    public IActionResult WoW(string i)
    {
        return Content(i);
    }
}
```
will match `http://localhost:<number>/Foo/xWoWy/message` url.

Now I want to replace `x` with `[` and `y` with `]` such that the attribute route matches 
 `http://localhost:<number>/Foo/[WoW]/message` url.

Unfortunately the modified code as follows does not compile.
```
[Route("[controller]")]
public class FooController : Controller
{
    [HttpGet("[[[action]]]/{i}")]
    public IActionResult WoW(string i)
    {
        return Content(i);
    }
}
```

with errors saying:

> While processing template '[controller]/[[[action]]]/{xxx}', a replacement value for the token 'action]' could not be found. Available tokens: 'action, controller'. To use a '[' or ']' as a literal string in a route or within a constraint, use '[[' or ']]' instead.

It seems a new bug.

JamesNK:
Hi @CodeMocker

I tested out nested square brackets in ASP.NET 2.1 and I got the same error as you did. It is a bug, but not one new to 2.2. This makes sense because the token replacement logic hasn't changed recently.

AlexejTimonin:
@JamesNK Can I take a stab at this?

JamesNK:
Sure thing.

As a starting point you could use this parsing logic over type names the has to handle nested containers:
https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Serialization/DefaultSerializationBinder.cs#L130-L174

`System.Collections.Generic.Dictionary`2[System.String,System.Collections.Generic.List`1[System.String]]`

PR #6957, opened 2019-01-23
Fix ReplaceTokens bug in AttributeRouteModel

AlexejTimonin:
Add scope to track when the token ends.

Addresses #6105

### DIFF
--- a/src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/ApplicationModels/AttributeRouteModel.cs
+++ b/src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/ApplicationModels/AttributeRouteModel.cs
@@ -231,7 +231,7 @@ namespace Microsoft.AspNetCore.Mvc.ApplicationModels
             var state = TemplateParserState.Plaintext;
 
             int? tokenStart = null;
-            int scope = 0;
+            var scope = 0;
 
             // We'll run the loop one extra time with 'null' to detect the end of the string.
             for (var i = 0; i <= template.Length; i++)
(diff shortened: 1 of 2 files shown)


src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/ApplicationModels/AttributeRouteModel.cs
```diff
@@ -231,6 +231,7 @@ public static string ReplaceTokens(string template, IDictionary<string, string>
             var state = TemplateParserState.Plaintext;
 
             int? tokenStart = null;
+            int scope = 0;
```
JamesNK: nit: var

(it's ASP.NET team coding standard)

JamesNK (changes requested):
Looks good!

I think we should also test for unbalanced nested brackets. Could you add some? For example:
* `"[controller]/[[[action]]/{id}"`
* `"[controller]/[[[action]]]]/{id}"`
* `"[controller]/[[action]]]/{id}"`

I would expect all those situations to throw.

src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/ApplicationModels/AttributeRouteModel.cs
AlexejTimonin: Thank you for the review, I'll add more tests and change to var. Just noticed I've a redundant scope decrement. Sorry for my brain fart..

JamesNK (approved):

JamesNK:
Thanks for fixing this. It will be in 3.0 preview3.

AlexejTimonin:
Was fun! Thanks for guidance!

PR #6959, opened 2019-01-23
Update buildtools to 3.0.0-build-20190123.3

natemcmaster:
Resolves https://github.com/aspnet/AspNetCore-Internal/issues/1707

### DIFF
--- a/korebuild-lock.txt
+++ b/korebuild-lock.txt
@@ -1,2 +1,2 @@
-version:3.0.0-build-20190122.3
-commithash:43e5fa8e3d8b192e573c3c14ae39b43e1d05cd5e
+version:3.0.0-build-20190123.3
+commithash:bfcdf8474c79320b7e5edcdb33d94bbdccb41e32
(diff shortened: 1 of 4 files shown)


korebuild-lock.txt
```diff
@@ -1,2 +1,2 @@
-version:3.0.0-build-20190122.3
-commithash:43e5fa8e3d8b192e573c3c14ae39b43e1d05cd5e
+version:3.0.0-build-20190123.3
```
dougbu: Ick. Is there any way to propagate this version from a single source of truth (preferably this file)? Annoying to need to change multiple locations.

Or, please point me to the script which does the needful 😸

dougbu (approved):

korebuild-lock.txt
natemcmaster: I usually do this:
```
rm korebuild-lock.txt
build.cmd /t:Noop
```
That re-generates korebuild-lock.txt and global.json. Then, I manually update eng/Versions.props to match. The version in eng/Versions.props is only there to make Visual Studio happy. Command-line builds are always in-sync with the version in korebuild-lock.txt.

We could get rid of eng/Versions.props by replacing the PackageReference to Internal.AspNetCore.Sdk with a proper 'MSBuild Sdk', but that's not something I want to tackle just yet.

PR #6977, opened 2019-01-24
Use System.Text.Json for Negotiate and Handshake

BrennanConroy:
Before:

Method |       Mean |    Error |   StdDev |        Op/s |  Gen 0 | Allocated |
------------------------------------------------ |-----------:|---------:|---------:|------------:|-------:|----------:|
WriteResponse_MemoryBufferWriter | 1,365.1 ns | 28.49 ns | 41.76 ns |   732,543.9 | 0.0038 |     456 B |
 ParsingNegotiateResponseMessageSuccessForValid1 | 1,055.5 ns | 27.91 ns | 82.29 ns |   947,388.5 | 0.0458 |    3928 B |
 ParsingNegotiateResponseMessageSuccessForValid2 |   853.0 ns | 20.59 ns | 60.72 ns | 1,172,387.5 | 0.0429 |    3696 B |
 ParsingNegotiateResponseMessageSuccessForValid3 | 1,019.7 ns | 20.33 ns | 52.12 ns |   980,699.7 | 0.0439 |    3776 B |
 ParsingNegotiateResponseMessageSuccessForValid4 | 1,491.3 ns | 29.80 ns | 64.78 ns |   670,546.6 | 0.0477 |    4176 B |
 ParsingNegotiateResponseMessageSuccessForValid5 | 1,826.6 ns | 39.14 ns | 95.26 ns |   547,463.8 | 0.0496 |    4384 B |

After:

Method |       Mean |     Error |     StdDev |        Op/s |  Gen 0 | Allocated |
------------------------------------------------ |-----------:|----------:|-----------:|---------:|------------:|----------:|
WriteResponse_MemoryBufferWriter | 487.2 ns | 8.223 ns | 7.692 ns | 2,052,423.5 | 0.0010 |     144 B |
ParsingNegotiateResponseMessageSuccessForValid1 |   356.4 ns |  7.551 ns |  20.284 ns | 2,806,082.7 | 0.0010 |     120 B |
 ParsingNegotiateResponseMessageSuccessForValid2 |   252.7 ns |  5.078 ns |   7.905 ns | 3,957,506.2 | 0.0014 |     120 B |
 ParsingNegotiateResponseMessageSuccessForValid3 |   411.9 ns |  8.259 ns |  19.468 ns | 2,427,993.3 | 0.0014 |     152 B |
 ParsingNegotiateResponseMessageSuccessForValid4 |   667.2 ns | 14.815 ns |  40.556 ns | 1,498,818.5 | 0.0029 |     272 B |
 ParsingNegotiateResponseMessageSuccessForValid5 | 1,038.2 ns | 51.740 ns | 152.557 ns | 963,224.3 | 0.0048 |     480 B |

cc @ahsonkhan


src/SignalR/common/Http.Connections.Common/src/NegotiateProtocol.cs
```diff
@@ -13,77 +15,166 @@ namespace Microsoft.AspNetCore.Http.Connections
     public static class NegotiateProtocol
     {
         private const string ConnectionIdPropertyName = "connectionId";
+        private static readonly byte[] ConnectionIdPropertyNameBytes = Encoding.UTF8.GetBytes("connectionId");
```
ahsonkhan: Consider re-using the const strings to avoid having to change them in multiple places.
```C#
private static readonly byte[] ConnectionIdPropertyNameBytes = Encoding.UTF8.GetBytes(ConnectionIdPropertyName);
```

ahsonkhan: OR, if performance is really critical, pre-compute the UTF-8 bytes and store them as byte arrays whose values are known at compile-time.

For example, for "url":
```C#
private static ReadOnlySpan<byte> ConnectionIdPropertyNameBytes => new byte [3] {117, 114, 108};
// OR
private static ReadOnlySpan<byte> ConnectionIdPropertyNameBytes => new byte[3] { (byte)'u', (byte)'r', (byte)'l' };
```

src/SignalR/common/Http.Connections.Common/src/NegotiateProtocol.cs
```diff
@@ -247,5 +338,157 @@ private static AvailableTransport ParseAvailableTransport(JsonTextReader reader)
...
+            var availableTransport = new AvailableTransport();
+
+            while (CheckRead(ref reader))
+            {
+                switch (reader.TokenType)
+                {
+                    case JsonTokenType.PropertyName:
+                        var memberName = reader.ValueSpan;
+
+                        if (ArrayEqual(memberName, TransportPropertyNameBytes))
```
davidfowl: Use span.SequenceEqual()

BrennanConroy: Was wondering if there was something for this...

src/SignalR/common/Http.Connections.Common/src/NegotiateProtocol.cs
```diff
@@ -247,5 +338,157 @@ private static AvailableTransport ParseAvailableTransport(JsonTextReader reader)
...
+                while (reader.Read() && depth < reader.CurrentDepth)
+                {
+                }
+            }
+        }
+
+        private static string ReadAsString(ref Utf8JsonReader reader, byte[] propertyName)
+        {
+            reader.Read();
+
```
davidfowl: nit: Remove extra space

src/SignalR/common/Http.Connections.Common/src/NegotiateProtocol.cs
```diff
@@ -13,77 +15,166 @@ namespace Microsoft.AspNetCore.Http.Connections
...
-                        jsonWriter.WritePropertyName(ConnectionIdPropertyName);
-                        jsonWriter.WriteValue(response.ConnectionId);
-                    }
+            writer.WriteEndArray();
+            writer.WriteEndObject();
+
+            writer.Flush(isFinalBlock: true);
+        }
+
+        public static NegotiationResponse ParseResponse(byte[] content)
```
davidfowl: `ReadOnlySpan<byte>`

src/SignalR/clients/csharp/Http.Connections.Client/src/HttpConnection.cs
```diff
@@ -440,10 +440,8 @@ private async Task<NegotiationResponse> NegotiateAsync(Uri url, HttpClient httpC
                     {
                         response.EnsureSuccessStatusCode();
                         NegotiationResponse negotiateResponse;
-                        using (var responseStream = await response.Content.ReadAsStreamAsync())
-                        {
-                            negotiateResponse = NegotiateProtocol.ParseResponse(responseStream);
-                        }
+                        var responseStream = await response.Content.ReadAsByteArrayAsync();
```
davidfowl: nit: responseBuffer

src/SignalR/common/Http.Connections.Common/src/Microsoft.AspNetCore.Http.Connections.Common.csproj
```diff
@@ -17,5 +18,8 @@
     <Reference Include="Microsoft.AspNetCore.Connections.Abstractions" />
     <Reference Include="Newtonsoft.Json" />
     <Reference Include="System.Buffers" />
+    <PackageReference Include="Microsoft.Bcl.Json.Sources" Version="4.6.0-preview.19073.1">
```
davidfowl: We should be multi-targeting here so we don't use this on .NET Core 3.0.

src/SignalR/common/Http.Connections.Common/src/NegotiateProtocol.cs
```diff
@@ -13,77 +15,166 @@ namespace Microsoft.AspNetCore.Http.Connections
...
+
         // Used to detect ASP.NET SignalR Server connection attempt
         private const string ProtocolVersionPropertyName = "ProtocolVersion";
+        private static readonly byte[] ProtocolVersionPropertyNameBytes = Encoding.UTF8.GetBytes("ProtocolVersion");
 
         public static void WriteResponse(NegotiationResponse response, IBufferWriter<byte> output)
         {
-            var textWriter = Utf8BufferTextWriter.Get(output);
-            try
+            var writer = new Utf8JsonWriter(output, new JsonWriterState(new JsonWriterOptions() { SkipValidation = true }));
```
ahsonkhan: nit:
```C#
var writer = new Utf8JsonWriter(output, new JsonWriterState(options: new JsonWriterOptions { SkipValidation = true }));
```

src/SignalR/common/Http.Connections.Common/src/NegotiateProtocol.cs
```diff
@@ -13,77 +15,166 @@ namespace Microsoft.AspNetCore.Http.Connections
...
+            writer.WriteEndObject();
+
+            writer.Flush(isFinalBlock: true);
+        }
+
+        public static NegotiationResponse ParseResponse(byte[] content)
+        {
+            try
+            {
+                var reader = new Utf8JsonReader(content, isFinalBlock: true, state: new JsonReaderState());
```
ahsonkhan: nit:
```C#
var reader = new Utf8JsonReader(content, isFinalBlock: true, state: default);
```

src/SignalR/common/Http.Connections.Common/src/NegotiateProtocol.cs
ahsonkhan: I would be curious to see the performance of using the `JsonDocument` in this case compared to the `Utf8JsonReader`.

src/SignalR/common/Http.Connections.Common/src/NegotiateProtocol.cs
```diff
@@ -247,5 +338,157 @@ private static AvailableTransport ParseAvailableTransport(JsonTextReader reader)
...
+        {
+            if (reader.TokenType == JsonTokenType.PropertyName)
+            {
+                reader.Read();
+            }
+
+            if (reader.TokenType == JsonTokenType.StartObject || reader.TokenType == JsonTokenType.StartArray)
+            {
+                int depth = reader.CurrentDepth;
+                while (reader.Read() && depth < reader.CurrentDepth)
```
ahsonkhan: This might need to be `depth <= reader.CurrentDepth`

At least that's what I observed in https://github.com/dotnet/core-setup/pull/5009/files#diff-02337f8b4f14d01b128b474cc5a5cf2aR98

src/SignalR/common/Http.Connections.Common/src/NegotiateProtocol.cs
```diff
@@ -247,5 +338,157 @@ private static AvailableTransport ParseAvailableTransport(JsonTextReader reader)
...
+                {
+                }
+            }
+        }
+
+        private static string ReadAsString(ref Utf8JsonReader reader, byte[] propertyName)
+        {
+            reader.Read();
+
+            if (reader.TokenType != JsonTokenType.String)
```
ahsonkhan: Do we support "null" values for the strings? In which you likely need to add a check for T`okenType == JsonTokenType.Null`

src/SignalR/common/Http.Connections.Common/src/Microsoft.AspNetCore.Http.Connections.Common.csproj
```diff
@@ -6,6 +6,7 @@
     <RootNamespace>Microsoft.AspNetCore.Http.Connections</RootNamespace>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <IsShippingPackage>true</IsShippingPackage>
+    <NoWarn>3021;1591</NoWarn>
```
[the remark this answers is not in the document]
ahsonkhan: OK. That's fine then. If there was a warning due to the JSON source package, please let me know where so I can fix it.

src/SignalR/common/Http.Connections.Common/src/NegotiateProtocol.cs
ahsonkhan: Yep, sequence equal works here. See https://github.com/aspnet/SignalR/pull/2734/files#diff-0e5591272fc4a9be570908d247813246R118

src/SignalR/common/Http.Connections.Common/src/NegotiateProtocol.cs
```diff
@@ -13,77 +15,166 @@ namespace Microsoft.AspNetCore.Http.Connections
...
+                    AccessToken = accessToken,
+                    AvailableTransports = availableTransports,
+                    Error = error,
+                };
             }
-            finally
+            catch (Exception ex)
             {
-                Utf8BufferTextWriter.Return(textWriter);
+                throw new InvalidDataException("Invalid negotiation response received.", ex);
```
ahsonkhan: Why do we need this catch all exception block now when it wasn't needed previously?

ahsonkhan (approved):
Otherwise, LGTM

src/SignalR/common/Http.Connections.Common/src/NegotiateProtocol.cs
```diff
@@ -247,5 +338,138 @@ private static AvailableTransport ParseAvailableTransport(JsonTextReader reader)
...
+
+        private static void EnsureObjectStart(ref Utf8JsonReader reader)
+        {
+            if (reader.TokenType != JsonTokenType.StartObject)
+            {
+                throw new InvalidDataException($"Unexpected JSON Token Type '{GetTokenString(reader.TokenType)}'. Expected a JSON Object.");
+            }
+        }
+
+        private static string GetTokenString(JsonTokenType tokenType)
```
Tornhoof: Why is this method necessary at all?
The returned special case strings have the same value as the enum names?

Tornhoof:
May I suggest considering special casing the project for netcoreapp3.0 and then using the codegen serializer currently under development and leaving the .netstandard 2.0 path as it is?
Reasons:
- You still have JSON.NET code via the ``Stream`` method code path
- The codegen serializer will most likely have comparable or even faster performance compared to the manually written code, specifically if the order of properties in the JSON in deserialization is not the same as the one in your code (as soon as your ``memberName.SequenceEqual(ErrorPropertyNameBytes)`` misses a branch, you'll be certainly slower than the ilasm integer tricks from Routing)

I understand that you all want to play with your new toys, but I think that code gets a bit out of hand.

davidfowl:
This isn’t about playing with toys. We’re removing JSON.NET from the shared framework so everything that uses it by default needs to be decoupled. This is step 0.

Tornhoof:
> This isn’t about playing with toys. We’re removing JSON.NET from the shared framework so everything that uses it by default needs to be decoupled. This is step 0.

You want to get rid of it in the .net standard 2.0 codepaths too?, that's news to me, but that explains it.

BrennanConroy:
Failing test is because of https://github.com/dotnet/corefx/issues/34632

src/SignalR/common/Http.Connections.Common/src/NegotiateProtocol.cs
halter73: If that's something we decide is worthwhile, we would want to change a bunch of type in Kestrel like HttpProtocol to pre-compute bytes as well.

src/SignalR/common/Http.Connections.Common/src/NegotiateProtocol.cs
BrennanConroy: It was there before, but the git diff makes it hard to see.

src/SignalR/common/Http.Connections.Common/src/NegotiateProtocol.cs
```diff
@@ -5,230 +5,244 @@
...
             catch (Exception ex)
             {
                 throw new InvalidDataException("Invalid negotiation response received.", ex);
             }
         }
 
-        private static AvailableTransport ParseAvailableTransport(JsonTextReader reader)
+        public static NegotiationResponse ParseResponse(Stream content)
+        {
+            var buffer = new byte[content.Length];
```
ahsonkhan: Is it guaranteed for the stream to be seekable here? If so, add `Debug.Assert(content.CanSeek)`? Also, do we need other arg validation like a null check?

BrennanConroy: I was thinking of marking this function as obsolete.

src/SignalR/common/Http.Connections.Common/src/NegotiateProtocol.cs
BrennanConroy: I gave `JsonDocument` a quick try, although I'm definitely no expert on it's usage:

Method |       Mean |    Error |    StdDev |        Op/s |  Gen 0 | Allocated |
------------------------------------------------ |-----------:|---------:|----------:|------------:|-------:|----------:|
ParsingNegotiateResponseMessageSuccessForValid1 | 1,252.6 ns | 24.96 ns | 24.510 ns |   798,312.5 | 0.0057 |     528 B |
ParsingNegotiateResponseMessageSuccessForValid2 |   877.6 ns | 14.30 ns |  9.462 ns | 1,139,434.6 | 0.0038 |     368 B |
ParsingNegotiateResponseMessageSuccessForValid3 | 1,096.4 ns | 21.74 ns | 26.694 ns |   912,048.6 | 0.0057 |     536 B |
ParsingNegotiateResponseMessageSuccessForValid4 | 2,365.0 ns | 46.47 ns | 63.604 ns |   422,835.1 | 0.0076 |     944 B |
ParsingNegotiateResponseMessageSuccessForValid5 | 2,688.2 ns | 52.91 ns | 49.489 ns |   372,001.7 | 0.0114 |    1208 B |

src/SignalR/common/Http.Connections.Common/src/Microsoft.AspNetCore.Http.Connections.Common.csproj
```diff
@@ -2,20 +2,22 @@
...
   <ItemGroup>
-    <Compile Include="$(SignalRSharedSourceRoot)JsonUtils.cs" Link="Internal\JsonUtils.cs" />
     <Compile Include="$(SignalRSharedSourceRoot)Utf8BufferTextWriter.cs" Link="Internal\Utf8BufferTextWriter.cs" />
   </ItemGroup>
 
   <ItemGroup>
     <Reference Include="Microsoft.AspNetCore.Connections.Abstractions" />
-    <Reference Include="Newtonsoft.Json" />
     <Reference Include="System.Buffers" />
+    <PackageReference Condition="'$(TargetFramework)' == 'netstandard2.0'" Include="Microsoft.Bcl.Json.Sources" Version="4.6.0-preview.19073.1">
```
davidfowl: nit: Move the condition last

src/SignalR/common/Shared/SystemTextJsonUtils.cs
```diff
@@ -0,0 +1,80 @@
+using System.IO;
+using System.Text;
+using System.Text.Json;
+
+namespace Microsoft.AspNetCore.Internal
+{
+    internal static class SystemTextJsonUtils
```
davidfowl: Why aren't these extension methods?

src/SignalR/common/Shared/TextMessageParser.cs
```diff
@@ -3,26 +3,47 @@
...
+
+                payload = buffer.Slice(0, index);
+
+                buffer = buffer.Slice(index + 1);
+
+                return true;
             }
+            else
+            {
+                var position = buffer.PositionOf(TextMessageFormatter.RecordSeparator);
```
davidfowl: Maybe split this out into another method?

src/SignalR/common/Shared/TextMessageParser.cs
```diff
@@ -3,26 +3,47 @@
...
 namespace Microsoft.AspNetCore.Internal
 {
     internal static class TextMessageParser
     {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static bool TryParseMessage(ref ReadOnlySequence<byte> buffer, out ReadOnlySequence<byte> payload)
         {
-            var position = buffer.PositionOf(TextMessageFormatter.RecordSeparator);
-            if (position == null)
+            if (buffer.IsSingleSegment)
```
davidfowl: You can do IndexOf on the first segment without checking IsSingleSegment

BrennanConroy:
@ahsonkhan What's with `ValueSpan` vs `ValueSequence`? A brief glance at the source looks like `ValueSequence` is never modified?

src/SignalR/common/Http.Connections.Common/src/NegotiateProtocol.cs
ahsonkhan: Thanks for giving it a try. Do you happen to have a commit with the code change to use `JsonDocument` that I can look at?

ahsonkhan:
> What's with `ValueSpan` vs `ValueSequence`? A brief glance at the source looks like `ValueSequence` is never modified?

Take a look at `Utf8JsonReader.MultiSegment.cs`. The only time `ValueSequence` is set/modified is when your input is within a ReadOnlySequence and your JSON token straddles more than 1 segment (for example a JSON string starts in one segment and ends in another). https://github.com/dotnet/corefx/blob/master/src/System.Text.Json/src/System/Text/Json/Reader/Utf8JsonReader.MultiSegment.cs

In those cases, you shouldn't access the JSON element via `ValueSpan` directly (since it won't always contain the value). Instead, check `HasValueSequence` to see where the value lives.

```C#
ReadOnlySpan<byte> value = reader.HasValueSequence ? reader.ValueSequence.ToArray() : reader.ValueSpan;
```

src/SignalR/common/Shared/SystemTextJsonExtensions.cs
```diff
@@ -0,0 +1,101 @@
...
+            return reader.GetString();
+        }
+
+        public static int? ReadAsInt32(this ref Utf8JsonReader reader, byte[] propertyName)
+        {
+            reader.Read();
+
+            if (reader.TokenType != JsonTokenType.Number)
+            {
+                throw new InvalidDataException($"Expected '{Encoding.UTF8.GetString(propertyName)}' to be of type {JsonTokenType.Number}.");
```
ahsonkhan: This will fail and throw an exception for null values. Poll the TokenType for the null JSON literal, not the ValueSpan.

```C#
        public static int? ReadAsInt32(this ref Utf8JsonReader reader, byte[] propertyName)
        {
            reader.Read();

            if (reader.TokenType == JsonTokenType.Null)
            {
                return null;
            }

            if (reader.TokenType != JsonTokenType.Number)
            {
                throw new InvalidDataException($"Expected '{Encoding.UTF8.GetString(propertyName)}' to be of type {JsonTokenType.Number}.");
            }

            return reader.GetInt32();
        }
```

src/SignalR/common/Shared/SystemTextJsonExtensions.cs
```diff
@@ -0,0 +1,101 @@
...
+            if (reader.TokenType == JsonTokenType.StartObject || reader.TokenType == JsonTokenType.StartArray)
+            {
+                int depth = reader.CurrentDepth;
+                while (reader.Read() && depth < reader.CurrentDepth)
+                {
+                }
+            }
+        }
+
+        public static string ReadAsString(this ref Utf8JsonReader reader, byte[] propertyName)
```
ahsonkhan: Similar to ReadAsInt32, if the string can be null, you need to check for TokenType.Null up front first to return null.

This will fail and throw an exception for null values. Poll the TokenType for the null JSON literal, not the ValueSpan.

```C#
        public static int? ReadAsInt32(this ref Utf8JsonReader reader, byte[] propertyName)
        {
            reader.Read();

            if (reader.TokenType == JsonTokenType.Null)
            {
                return null;
            }

            if (reader.TokenType != JsonTokenType.String)
            {
                throw new InvalidDataException($"Expected '{Encoding.UTF8.GetString(propertyName)}' to be of type {JsonTokenType.String}.");
            }

            return reader.GetString();
        }
```

src/SignalR/common/SignalR.Common/src/Protocol/HandshakeProtocol.cs
```diff
@@ -123,59 +106,51 @@ public static bool TryParseResponseMessage(ref ReadOnlySequence<byte> buffer, ou
...
             {
-                using (var reader = JsonUtils.CreateJsonTextReader(textReader))
+                if (reader.TokenType == JsonTokenType.PropertyName)
                 {
-                    JsonUtils.CheckRead(reader);
-                    JsonUtils.EnsureObjectStart(reader);
-
-                    int? minorVersion = null;
-                    string error = null;
+                    var memberName = reader.ValueSpan;
```
ahsonkhan: ValueSpan won't always have the token (when your input payload is within a multi-segment `ReadOnlySequence`, check `HasValueSequence` and call `ValueSequence.ToArray()` for tokens that straddle more than one segment

src/SignalR/common/SignalR.Common/src/Protocol/HandshakeProtocol.cs
```diff
@@ -192,63 +167,54 @@ public static bool TryParseRequestMessage(ref ReadOnlySequence<byte> buffer, out
...
             {
-                using (var reader = JsonUtils.CreateJsonTextReader(textReader))
+                if (reader.TokenType == JsonTokenType.PropertyName)
                 {
-                    JsonUtils.CheckRead(reader);
-                    JsonUtils.EnsureObjectStart(reader);
-
-                    string protocol = null;
-                    int? protocolVersion = null;
+                    var memberName = reader.ValueSpan;
```
ahsonkhan: same here

src/SignalR/common/SignalR.Common/src/Protocol/HandshakeProtocol.cs
ahsonkhan: >This doesn't matter since we pass in a byte[] correct?

Doesn't look like to me. It is coming from pipe `ReadResult`:
https://github.com/aspnet/AspNetCore/blob/3cd5054eb5d8851b25881cd05cccb931aaa022e6/src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs#L806-L817

TryParseMessage can return a multi-segment ReadOnlySequence:
https://github.com/aspnet/AspNetCore/blob/51be998c3cfe4258eecf30bd01207e3288de51d9/src/SignalR/common/Shared/TextMessageParser.cs#L13

Which will eventually get passed into this `Utf8JsonReader`:
https://github.com/aspnet/AspNetCore/blob/51be998c3cfe4258eecf30bd01207e3288de51d9/src/SignalR/common/SignalR.Common/src/Protocol/HandshakeProtocol.cs#L109

BrennanConroy:
>Take a look at Utf8JsonReader.MultiSegment.cs.

Ah, so it is modified! I'll do the needful to handle `ValueSequence` then.

BrennanConroy:
We do pass in a ReadOnlySequence for Handshakes

src/SignalR/common/Http.Connections.Common/src/NegotiateProtocol.cs
halter73: Even being marked as obsolete, this implementation seems worse than just throwing a NotImplementedException. Even if the Stream is seekable and has a Length, I don't think it's guaranteed the entire content is buffered and ready to go. Stream.Read() could block or return only partial content.

Do we include calls to this method in any of our samples, demos or docs?

JamesNK:
It is always working now because handshakes are so small that they fit in one segment. To fix and test you will need to make each segment tiny.

BrennanConroy:
I know ;)

src/SignalR/common/Http.Connections.Common/src/NegotiateProtocol.cs
BrennanConroy: I'm not sure why this class is public, I'm guessing it was a mistake.

davidfowl: It's not that big of a deal, hopefully it won't matter in the long run. Mark it obselete and we can carry on with our lives.

src/SignalR/common/Shared/SystemTextJsonExtensions.cs
BrennanConroy: I can't think of a situation where we want to accept a null string in the hub protocol.

We explicitly have tests that check for the parsing to fail if null is passed in.

src/SignalR/common/Http.Connections.Common/src/NegotiateProtocol.cs
BrennanConroy: I didn't save it, but it was fairly quick so I can make one for you.

src/SignalR/common/SignalR.Common/src/Protocol/HandshakeProtocol.cs
```diff
@@ -192,63 +167,54 @@ public static bool TryParseRequestMessage(ref ReadOnlySequence<byte> buffer, out
...
             {
-                using (var reader = JsonUtils.CreateJsonTextReader(textReader))
+                if (reader.TokenType == JsonTokenType.PropertyName)
                 {
-                    JsonUtils.CheckRead(reader);
-                    JsonUtils.EnsureObjectStart(reader);
-
-                    string protocol = null;
-                    int? protocolVersion = null;
+                    var memberName = reader.HasValueSequence ? reader.ValueSequence.ToArray() : reader.ValueSpan;
```
JamesNK: @ahsonkhan Have you considered having a helper method for comparing the current value that worries about this for you? It would avoid developers making the mistake of not using ValueSequence, and could do something more efficient than allocating an array when doing the compare.

JamesNK: https://github.com/dotnet/corefx/issues/34938

ahsonkhan: I initially only had `ValueSpan` and would call `ToArray` on the ReadOnlySequence slice which circumvents this usability issue. However, it results in unnecessary allocations for JSON elements the user just wants to read past and not materialize/look into. So, we decided to leave the allocation to the caller since they have the context on when they actually care about the value or not.

> it could do something more efficient than allocating an array when doing the compare.

We could consider providing such a helper for compares. That is interesting feedback :)

The concern I have is, if the compare returns true, the user would have to call ToArray on the ReadOnlySequence anyway, so now we walk the sequence twice (once for the helper to check/compare, and if true, the caller materializing the value to some string/byte[]). We could save on allocations (by potentially copying to an array pool or some custom EqualToSequence method), but I don't know if it helps performance. How often is the search true versus false?

JamesNK: Before:

```cs
var memberName = reader.HasValueSequence ? reader.ValueSequence.ToArray() : reader.ValueSpan;

if (memberName.SequenceEqual(ProtocolPropertyNameBytes))
{
    protocol = reader.ReadAsString(ProtocolPropertyNameBytes);
}
else if (memberName.SequenceEqual(ProtocolVersionPropertyNameBytes))
{
    protocolVersion = reader.ReadAsInt32(ProtocolVersionPropertyNameBytes);
}
else
{
    reader.Skip();
}
```

After:

```cs
if (reader.ValueEquals(ProtocolPropertyNameBytes))
{
    protocol = reader.ReadAsString(ProtocolPropertyNameBytes);
}
else if (reader.ValueEquals(ProtocolVersionPropertyNameBytes))
{
    protocolVersion = reader.ReadAsInt32(ProtocolVersionPropertyNameBytes);
}
else
{
    reader.Skip();
}
```

BrennanConroy: I did this manually with a helper function for `ReadOnlySequence.SequenceEqual` and it helped the multi-segment case but slowed the single-segment case

halter73: So you've already written `ReadOnlySequence.SequenceEqual` and calling that only when HasValueSequence is false slows down the single-segment case significantly? If you were to do something like the following, I hope the single-segment case wouldn't too much of a hit:

```c#
private static bool ValueEquals(in Utf8JsonReader reader, Span<byte> value)
{
    return reader.HasValueSequence ?
        reader.ValueSequence.SequenceEqual(value) :
        reader.ValueSpan.SequenceEqual(value);
}
```

BrennanConroy: I give `ValueEquals` a try, maybe I was doing something dumb.

JamesNK: IMO don't worry about optimizing this in SignalR. These messages are always going to be in one segment. I think this is an improvement for Utf8JsonReader.

halter73: I agree that it's not important for this change. I'm just curious why @BrennanConroy's helper function is slower. I'm thinking maybe it's the Utf8JsonReader parameter getting copied.

BrennanConroy:
Ping, is there anymore actionable feedback here? Or can I get approval(s)?

src/SignalR/common/Shared/SystemTextJsonExtensions.cs
```diff
@@ -0,0 +1,102 @@
...
+                throw new InvalidDataException($"Unexpected JSON Token Type '{GetTokenString(reader.TokenType)}'. Expected a JSON Array.");
+            }
+        }
+
+        // Remove after https://github.com/dotnet/corefx/issues/33295 is done
+        public static void Skip(this ref Utf8JsonReader reader)
+        {
+            if (reader.TokenType == JsonTokenType.PropertyName)
+            {
+                reader.Read();
```
JamesNK: CheckRead?

src/SignalR/common/Shared/SystemTextJsonExtensions.cs
```diff
@@ -0,0 +1,102 @@
...
+            }
+
+            if (reader.TokenType == JsonTokenType.StartObject || reader.TokenType == JsonTokenType.StartArray)
+            {
+                int depth = reader.CurrentDepth;
+                while (reader.Read() && depth < reader.CurrentDepth)
```
JamesNK: CheckRead?

src/SignalR/common/Shared/SystemTextJsonExtensions.cs
```diff
@@ -0,0 +1,102 @@
...
+                {
+                }
+            }
+        }
+
+        public static string ReadAsString(this ref Utf8JsonReader reader, byte[] propertyName)
+        {
+            reader.Read();
```
JamesNK: CheckRead?

src/SignalR/common/Shared/SystemTextJsonExtensions.cs
```diff
@@ -0,0 +1,102 @@
...
+            {
+                throw new InvalidDataException($"Expected '{Encoding.UTF8.GetString(propertyName)}' to be of type {JsonTokenType.String}.");
+            }
+
+            return reader.GetString();
+        }
+
+        public static int? ReadAsInt32(this ref Utf8JsonReader reader, byte[] propertyName)
+        {
+            reader.Read();
```
JamesNK: CheckRead?

JamesNK (approved):
Maybe these should be CheckRead to handle JSON ending unexpectedly. I'll leave to you.

src/SignalR/common/Http.Connections.Common/src/NegotiateProtocol.cs
BrennanConroy: Take a look at https://github.com/aspnet/AspNetCore/compare/brecon/jsonNeg...brecon/jsonDoc?expand=1

This time I noticed the TryGetProperty method, so it was a little bit better.
The `ToArray` in `JsonDocument.Parse` should be avoidable if an overload was added to support `ReadOnlySpan<byte>`.

Side by side numbers:
**With JsonDocument**

Method |       Mean |    Error |   StdDev |        Op/s |  Gen 0 | Allocated |
------------------------------------------------ |-----------:|---------:|---------:|------------:|-------:|----------:|
 ParsingNegotiateResponseMessageSuccessForValid1 | 1,495.7 ns | 31.17 ns | 65.75 ns |   668,580.0 | 0.0019 |     264 B |
 ParsingNegotiateResponseMessageSuccessForValid2 | 1,070.6 ns | 22.78 ns | 43.34 ns |   934,042.2 | 0.0019 |     248 B |
 ParsingNegotiateResponseMessageSuccessForValid3 | 1,314.7 ns | 26.21 ns | 55.86 ns |   760,646.2 | 0.0019 |     304 B |
 ParsingNegotiateResponseMessageSuccessForValid4 | 2,482.5 ns | 49.28 ns | 88.86 ns |   402,823.8 | 0.0038 |     456 B |
 ParsingNegotiateResponseMessageSuccessForValid5 | 2,820.7 ns | 55.10 ns | 99.36 ns |   354,521.8 | 0.0038 |     720 B |

**With Utf8JsonReader**

Method |       Mean |     Error |   StdDev |        Op/s |  Gen 0 | Allocated |
------------------------------------------------ |-----------:|----------:|---------:|------------:|-------:|----------:|
 ParsingNegotiateResponseMessageSuccessForValid1 |   404.7 ns |  8.122 ns | 13.79 ns | 2,470,926.1 | 0.0010 |     120 B |
 ParsingNegotiateResponseMessageSuccessForValid2 |   305.5 ns |  6.168 ns | 15.13 ns | 3,273,484.3 | 0.0010 |     120 B |
 ParsingNegotiateResponseMessageSuccessForValid3 |   461.3 ns |  9.164 ns | 18.09 ns | 2,167,912.7 | 0.0014 |     152 B |
 ParsingNegotiateResponseMessageSuccessForValid4 |   752.4 ns | 14.997 ns | 29.95 ns | 1,329,015.3 | 0.0029 |     272 B |
 ParsingNegotiateResponseMessageSuccessForValid5 | 1,101.3 ns | 22.009 ns | 60.25 ns |   908,003.1 | 0.0038 |     480 B |

ahsonkhan: The `ToArray()` isn't great :(

Can you try copying it to an array pool-rented array instead and share how much that helps?
```C#
// BEFORE:
using (var doc = JsonDocument.Parse(content.ToArray(), readerOptions: default))
```
```C#
// AFTER:
byte[] pooledBuffer = ArrayPool<byte>.Shared.Rent(content.Length);
content.CopyTo(pooledBuffer);

using (var doc = JsonDocument.Parse(pooledBuffer, readerOptions: default))
{
   ...
}
ArrayPool<byte>.Shared.Return(pooledBuffer);

```

I expected a ~2x regression between reader and document, not 3-4x.

ahsonkhan: Better yet, can we change the API signature to accept `ReadOnlyMemory<byte>` instead to avoid having to copy data at all? It looks like the only caller passes in a byte[] anyway:
```C#
public static NegotiationResponse ParseResponse(ReadOnlySpan<byte> content)
```

https://github.com/aspnet/AspNetCore/pull/6977/files#diff-041ab9d757596cece2ef39f92ede7e4dR443

src/SignalR/common/Shared/SystemTextJsonExtensions.cs
ahsonkhan: Add a `Debug.Assert(reader.TokenType != JsonTokenType.Null)` to be explicit?

ahsonkhan (approved):

davidfowl (approved):
LGTM

src/SignalR/common/Http.Connections.Common/src/NegotiateProtocol.cs
ahsonkhan: Thanks for sharing the numbers. They provide useful insight.

PR #4447, opened 2018-12-05
Wait to dispose RequestAborted CTS

halter73:
Partially Addresses #4422

PR #6994, opened 2019-01-24
Revert "Wait to dispose RequestAborted CTS (#4447)"

halter73:
This reverts commit 0622513058eff9a4374a0edca726d2c62959ea1e.

### DIFF
--- a/eng/PatchConfig.props
+++ b/eng/PatchConfig.props
@@ -36,7 +36,6 @@ Later on, this will be checked using this condition:
       Microsoft.AspNetCore.Mvc.Core;
       Microsoft.AspNetCore.Routing;
       Microsoft.AspNetCore.Server.IIS;
-      Microsoft.AspNetCore.Server.Kestrel.Core;
       java:signalr;
     </PackagesInPatch>
   </PropertyGroup>
--- a/src/Servers/Kestrel/Core/test/Http1ConnectionTests.cs
+++ b/src/Servers/Kestrel/Core/test/Http1ConnectionTests.cs
@@ -53,7 +53,11 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Tests
             var connectionFeatures = new FeatureCollection();
             connectionFeatures.Set(Mock.Of<IConnectionLifetimeFeature>());
 
-            _serviceContext = new TestServiceContext();
+            _serviceContext = new TestServiceContext()
+            {
+                Scheduler = PipeScheduler.Inline
+            };
+
             _timeoutControl = new Mock<ITimeoutControl>();
             _http1ConnectionContext = new HttpConnectionContext
             {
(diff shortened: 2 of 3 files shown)


Tratcher (approved):

src/Servers/Kestrel/Core/test/Http1ConnectionTests.cs
```diff
@@ -728,25 +724,6 @@ public async Task RequestAbortedTokenIsResetBeforeLastWriteWithChunkedEncoding()
             Assert.False(_http1Connection.RequestAborted.IsCancellationRequested);
         }
 
-        [Fact]
-        public void RequestAbortedTokenIsFullyUsableAfterCancellation()
```
natemcmaster: Do we want to keep this test?

eng/PatchConfig.props
```diff
@@ -36,7 +36,6 @@ Later on, this will be checked using this condition:
       Microsoft.AspNetCore.Mvc.Core;
       Microsoft.AspNetCore.Routing;
       Microsoft.AspNetCore.Server.IIS;
-      Microsoft.AspNetCore.Server.Kestrel.Core;
```
natemcmaster: Just to confirm - nothing else was added to Kestrel.Core since the 2.2.0 release, right?

dougbu: I double-checked and the only other commits in `src/Servers/Kestrel/Core` since 2.2.0 look like code reorganization and test-only changes. For example, the merge commits from `release/2.1` that affected product code just changed the IVTs in `AssemblyInfo.cs`.

However, it looks like a `release/2.1` change (feb641ddda) that was merged into `release/2.2` after 2.2.0 updated `src/Shared/CertificateGeneration/` significantly. Kestrel Core has a wildcard `<Compile/>` for that directory and uses `CertificateManager`. **Is what feb641ddda merged important for 2.2.2?**

halter73: Nothing changed in Kestrel.Core proper since 2.2.0. Since we're not going to ship Kestrel.Core as part of 2.2.2, I would think that Kestrel.Core's CertificateManager reference is irrelevant.

src/Servers/Kestrel/Core/test/Http1ConnectionTests.cs
halter73: One of the WaitHandle assertions fail now, but I removed it with a comment and re-added the test.

src/Servers/Kestrel/Core/test/Http1ConnectionTests.cs
```diff
@@ -729,14 +729,16 @@ public async Task RequestAbortedTokenIsResetBeforeLastWriteWithChunkedEncoding()
...
+        public void RequestAbortedTokenIsUsableAfterCancellation()
         {
             var originalToken = _http1Connection.RequestAborted;
             var originalRegistration = originalToken.Register(() => { });
 
             _http1Connection.Abort(new ConnectionAbortedException());
 
-            Assert.True(originalToken.WaitHandle.WaitOne(TestConstants.DefaultTimeout));
+            // The following line will throw an ODE because the original CTS backing the token has been diposed.
```
muratg: nit: diposed -> disposed

muratg (approved):

eng/PatchConfig.props
natemcmaster: Yeah, we're okay. That change you're seeing is https://github.com/aspnet/AspNetCore/pull/4801, which moved source code from aspnet/Extensions into this repo.

natemcmaster (approved):

dougbu (approved):

Issue #5149, opened 2018-12-17
UseStartup on GenericWebHostBuilder should fail when Startup.ConfigureServices returns an IServiceProvider


davidfowl:
Right now it's being ignored and it should throw instead.

Program.cs

```C#
using System;
using System.Collections.Generic;
using System.IO;
using System.Linq;
using System.Threading.Tasks;
using Microsoft.AspNetCore;
using Microsoft.AspNetCore.Hosting;
using Microsoft.Extensions.Configuration;
using Microsoft.Extensions.Hosting;
using Microsoft.Extensions.Logging;

namespace WebApplication44
{
    public class Program
    {
        public static void Main(string[] args)
        {
            CreateWebHostBuilder(args).Build().Run();
        }

        public static IHostBuilder CreateWebHostBuilder(string[] args) =>
            Host.CreateDefaultBuilder(args)
                .ConfigureWebHostDefaults(builder =>
                {
                    builder.UseStartup<Startup>();
                });
    }
}
```

Startup.cs

```C#
using System;
using System.Collections.Generic;
using System.Linq;
using System.Threading.Tasks;
using Microsoft.AspNetCore.Builder;
using Microsoft.AspNetCore.Hosting;
using Microsoft.AspNetCore.Http;
using Microsoft.Extensions.DependencyInjection;

namespace WebApplication44
{
    public class Startup
    {
        // This method gets called by the runtime. Use this method to add services to the container.
        // For more information on how to configure your application, visit https://go.microsoft.com/fwlink/?LinkID=398940
        public IServiceProvider ConfigureServices(IServiceCollection services)
        {
            return services.BuildServiceProvider();
        }

        // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
        public void Configure(IApplicationBuilder app, IHostingEnvironment env)
        {
            if (env.IsDevelopment())
            {
                app.UseDeveloperExceptionPage();
            }

            app.Run(async (context) =>
            {
                await context.Response.WriteAsync("Hello World!");
            });
        }
    }
}
```

Here's the code that supposed to do that check.

https://github.com/aspnet/AspNetCore/blob/cc899e2be729364589ba7f4392bce9942e4faca4/src/Hosting/Hosting/src/GenericHost/GenericWebHostBuilder.cs#L245-L249

davidfowl:
cc @Tratcher

buvinghausen:
[...]

PR #6997, opened 2019-01-24
Block generic host Startup.ConfigureServices that return IServiceProvider 

Tratcher:
#5149 `IServiceProvider ConfigureServices(IServiceCollection services)` is one of the patterns not supported by the generic web host as it needs to compose on a single container. Adding an error for discoverability.

@muratg this is a low risk merge.


davidfowl (approved):

JunTaoLuo (approved):

muratg (approved):
Approved, but let's verify with Nate (preview2 build status) before merging tomorrow.

PR #7004, opened 2019-01-25
Harden IISExpress shutdown code

pakrym:
Continuation of https://github.com/aspnet/AspNetCore/pull/6854

`EnumWindows` is more reliable than `GetWindow` and making sure we send the message to the correct window because IISExpress usually has two.


jkotalik (approved):

pakrym:
@muratg this is a test stability fix.

muratg (approved):
Approved, but please watch the build and if it breaks the build be prepared to pull it out.

We're considering a rebuild of preview2 due to a Razor bug and this should not block that build (TBD today).

PR #7028, opened 2019-01-25
No-op Authorization middleware for Razor Pages

pranavkm:
Workaround for https://github.com/aspnet/AspNetCore/issues/7011


src/Security/Authorization/Policy/src/AuthorizationMiddleware.cs
```diff
@@ -45,11 +44,19 @@ public async Task Invoke(HttpContext context)
...
 
+            var endpoint = context.GetEndpoint();
+
+            // Workaround for https://github.com/aspnet/AspNetCore/issues/7011. Do not use the AuthorizationMiddleware for Razor Pages
+            if (endpoint.Metadata.Any(m => m.GetType().FullName == "Microsoft.AspNetCore.Mvc.ApplicationModels.PageRouteMetadata"))
+            {
+                await _next(context);
+                return;
+            }
+
```
rynowak: I guess what I expected to see what for `AuthorizeFilter` (MVC) to ignore `AuthorizationMiddlewareInvokedKey` - so that we always run authorization twice. This should be a little bit slower, but it's is close to what we shipped in preview1 that didn't have this bug.

mkArtakMSFT (approved):

src/Security/Authorization/Policy/src/AuthorizationMiddleware.cs
pranavkm: Summing up offline conversation: Authenticate \ Authorize has side effects and I wanted to avoid running it twice if we could help it. This does work correctly for the particular scenario that is broken. I've added an additional test with controllers to make sure that works correctly too so we've covered all the bases.

rynowak (approved):

pranavkm:
🆙 📅 sorry about the confusion

JamesNK (approved):

mkArtakMSFT:
Merging this as flaky tests are blocking this for third round now.

Issue #6979, opened 2019-01-24
[SignalR JS | Question] Proper webworker detection


dukhanov:
According to the JS code the type of platform which runs client connection is detected by global window object: `const isNode = typeof window === "undefined";`

In case the client is initialized inside of WebWorker (I use `dist/browser/signalr.min.js`) the script execution is failed with an error `Uncaught ReferenceError: window is not defined`

The solution which I made is to add a fake global window object:
```
importScripts(
	'fakeGlobalWindow.js', // global window should be defined for signalR client lib
	'signalr.min.js',
)
```
fakeGlobalWindow.js:
```
const window = self.window = self
```
But it looks like the ugly hack.

Are there any possibilities to change the checking?
https://github.com/aspnet/AspNetCore/blob/master/src/SignalR/clients/ts/signalr/src/HttpConnection.ts#L74
from:
```
const isNode = typeof window === "undefined";
        if (!isNode && typeof WebSocket !== "undefined" && !options.WebSocket) {
            options.WebSocket = WebSocket;
        } else if (isNode && !options.WebSocket) {
            if (WebSocketModule) {
                options.WebSocket = WebSocketModule;
            }
        }
```
to:
```
const isBrowser = typeof window === 'object';
const isWorker = typeof importScripts === 'function';
const isNode = !isBrowser && !isWorker

        if (!isNode && typeof WebSocket !== "undefined" && !options.WebSocket) {
            options.WebSocket = WebSocket;
        } else if (isNode && !options.WebSocket) {
            if (WebSocketModule) {
                options.WebSocket = WebSocketModule;
            }
        }
```
Is there any caveats to use signalR client inside of the webWroker?

bradygaster:
This is something we'd have to try out, as the web worker is something we've not experimented with in this context.

dukhanov:
Added a PR to support webworkers https://github.com/aspnet/AspNetCore/pull/7058

aspnet-hello:
We periodically close 'discussion' issues that have not been updated in a long period of time.

We apologize if this causes any inconvenience. We ask that if you are still encountering an issue, please log a new issue with updated information and we will investigate.

PR #7058, opened 2019-01-28
SignalR JS client add webworker support

dukhanov:
Summary of the changes:
 - Added Platform utils to detect platform type
 - Added additional build for WebWorker
 - Updated the readme file with SignalR WebWorker instructions

Addresses #6979 

### WebWorker

To use the client in a webworker, copy `*.js` files from the `dist/webworker` folder to your script folder include on your webworker using the `importScripts` function. Note that webworker SignalR hub connection supports only absolute path to a SignalR hub.


src/SignalR/clients/ts/signalr/webpack.webworker.config.js
```diff
@@ -0,0 +1,15 @@
...
+    // These are only used in Node environments
+    // so we tell webpack not to pull them in for the browser
+    target: "webworker",
+    platformDist: "webworker",
+    externals: [
+        "websocket",
+        "eventsource",
+        "request"
+    ]
+});
```
halter73: Why do we need this in addition to the main webpack.config.js? It looks like it's just changing the output directory even though the client source code is identical and all the platform checks are done at runtime. What am I missing?

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -382,8 +381,7 @@ export class HttpConnection implements IConnection {
         if (url.lastIndexOf("https://", 0) === 0 || url.lastIndexOf("http://", 0) === 0) {
             return url;
         }
-
-        if (typeof window === "undefined" || !window || !window.document) {
+        if (!Platform.isBrowser || !window.document) {
```
halter73: nit: re-add newline

src/SignalR/clients/ts/signalr/src/Utils.ts
```diff
@@ -23,6 +23,23 @@ export class Arg {
     }
 }
 
+/** @private */
+export class Platform {
```
halter73: 👍

src/SignalR/clients/ts/signalr/src/Utils.ts
```diff
@@ -23,6 +23,23 @@ export class Arg {
...
+    }
+
+    public static get isWebWorker(): boolean {
+        return typeof self === "object" && "importScripts" in self;
+    }
+
+    public static get isNode(): boolean {
+        return !this.isBrowser && !this.isWebWorker;
+    }
+
```
halter73: Nit: remove newline

src/SignalR/clients/ts/signalr/README.md
```diff
@@ -33,6 +37,25 @@ connection.start()
...
+    .withUrl("https://example.com/signalr/chat")
+    .build();
+
+connection.on("send", data => {
+    console.log(data);
+});
+
+connection.start()
+    .then(() => connection.invoke("send", "Hello"));
+
```
halter73: nit

src/SignalR/clients/ts/signalr/webpack.webworker.config.js
halter73: Forgive my ignorance here. Do we actually need a new target and output directory? Why can't we have a single target that works in both a normal browser context and a webworker context?

dukhanov: Because the global window object is not available in WebWorker and if we use build script from `web` inside of webworker the SignalR script execution failed because window object is not available.
More over there are some differences between build target for `webworker` and `web`
https://github.com/webpack/webpack/blob/master/lib/WebpackOptionsApply.js#L98

halter73 (approved):

halter73:
Thanks @dukhanov! If you have time, can you please take a look #7378?

PR #7071, opened 2019-01-28
Remove AspNetCoreModule V1

pakrym:
Fixes: https://github.com/aspnet/AspNetCore/issues/6847


jkotalik:
Is this really a fix 😄 

So a few things we need to consider before merging:

- Customers that upgrade from 2.2 to 3.0 that have web.config files will now be broken if they target ANCM V1 (which is fine as we are in a breaking release). Make sure we document this somewhere.
- Hosts will need to download 2.2 and 3.0 if they want ANCM and latest ANCMV2.

Other than that, I'm fine with doing this. CC @muratg @shirhatti

pakrym:
> Customers that upgrade from 2.2 to 3.0 that have web.config files will now be broken if they target ANCM V1 (which is fine as we are in a breaking release). Make sure we document this somewhere.

https://github.com/aspnet/websdk/issues/473

muratg:
@shirhatti just verifying that you're OK with this in 3.0. Keep in mind that we're keeping V1 in 2.1 and 2.2 so folks can get to it if they really need it.

muratg (approved):

Tratcher:
Breaking change announcement?

pakrym:
https://github.com/aspnet/Announcements/issues/339

PR #7082, opened 2019-01-29
Increase shutdown timeout in Kestrel TestServer

halter73:
Logs on our CI server show that a call to `await Task.Yield()` in ImmediateShutdownAfterOnConnectionAsyncDoesNotCrash can take over 5 seconds during shutdown which can lead to a shutdown timeout and a failed test. This increases the 5 second timeout to 30 seconds for all tests that use TestServer. 30 seconds is the default test timeout in Kestrel.

This change also ensures that `ConnectionContext.Abort(ConnectionAbortedException abortReason)` never gets called with a null abortReason now that the transport logs the message.


src/Servers/Kestrel/Transport.Abstractions/src/Internal/TransportConnection.FeatureCollection.cs
```diff
@@ -91,7 +92,7 @@ CancellationToken IConnectionLifetimeNotificationFeature.ConnectionClosedRequest
             set => ConnectionClosedRequested = value;
         }
 
-        void IConnectionLifetimeFeature.Abort() => Abort(abortReason: null);
+        void IConnectionLifetimeFeature.Abort() => Abort(new ConnectionAbortedException("The connection was aborted by the application via IConnectionLifetimeFeature.Abort()."));
```
jkotalik: This exception string is used twice. Consider adding it to the resx.

src/Servers/Kestrel/test/InMemory.FunctionalTests/ConnectionDispatcherTests.cs
```diff
@@ -0,0 +1,44 @@
...
+
+namespace Microsoft.AspNetCore.Server.Kestrel.InMemory.FunctionalTests
+{
+    public class ConnectionDispatcherTests : TestApplicationErrorLoggerLoggedTest
+    {
+        [Fact]
+        public async Task AppAbortViaConnectionContextIsLogged()
+        {
+            var testContext = new TestServiceContext(LoggerFactory);
+            var connectionDispatcher = new AbortingConnectionDispatcher();
```
jkotalik: Nit: How is this test testing the ConnectionDispatcher? Seems more like you are mocking it and testing TransportConnection?

src/Servers/Kestrel/test/InMemory.FunctionalTests/ResponseTests.cs
```diff
@@ -2285,6 +2286,58 @@ public async Task NoResponseSentWhenConnectionIsClosedByServerBeforeClientFinish
...
+                        "GET / HTTP/1.1",
+                        "Host:",
+                        "",
+                        "");
+                    await connection.ReceiveEnd();
+                }
+                await server.StopAsync();
+            }
+
+            Assert.Single(TestApplicationErrorLogger.Messages.Where(m => m.Message.Contains("The connection was aborted by the application.")));
```
jkotalik: Nit: Can we get these error messages from the resx/core strings directly so we don't need to update two places?

src/Servers/Kestrel/test/InMemory.FunctionalTests/TestTransport/TestServer.cs
```diff
@@ -106,9 +106,12 @@ public InMemoryConnection CreateConnection()
             return new InMemoryConnection(transportConnection);
         }
 
-        public Task StopAsync()
+        public async Task StopAsync()
         {
-            return _host.StopAsync();
+            using (var cts = new CancellationTokenSource(TestConstants.DefaultTimeout))
```
jkotalik: Dispose the cts?

halter73: That's what the using is for.

src/Servers/Kestrel/Transport.Abstractions/src/Internal/TransportConnection.FeatureCollection.cs
halter73: One is for ConnectionContext and the other is for IConnectionLifetimeFeature. ConnectionContext is defined in an assembly that doesn't currently have a resx which is probably the main reason it was left null before.

src/Servers/Kestrel/Core/src/Internal/HttpConnection.cs
```diff
@@ -149,7 +149,7 @@ public async Task ProcessRequestsAsync<TContext>(IHttpApplication<TContext> http
                                     break;
                                 case HttpProtocols.None:
                                     // An error was already logged in SelectProtocol(), but we should close the connection.
-                                    Abort(ex: null);
+                                    Abort(new ConnectionAbortedException(CoreStrings.ProtocolSelectionFailed));
```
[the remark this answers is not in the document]
halter73: In this case, I think it's best to log twice. There's no guarantee debug-level transport logs are even being collected which would be necessary to see the ConnectionAbortedException message logged. The log in SelectProtocol() is an error-level log with exception details which is much more likely to get noticed. Most timeouts similarly get logged by both the transport and the core assembly.

davidfowl:
> Logs on our CI server show that a call to await Task.Yield() in ImmediateShutdownAfterOnConnectionAsyncDoesNotCrash can take over 5 seconds during shutdown which can lead to a shutdown timeout and a failed test. This increases the 5 second timeout to 30 seconds for all tests that use TestServer. 30 seconds is the default test timeout in Kestrel.

Before increasing the timeout can we get a dump when this test fails? I'd like to understand why it takes that long. There's a new attribute you can add to your test to collect a dump on failure

src/Servers/Kestrel/test/InMemory.FunctionalTests/TestTransport/TestServer.cs
jkotalik: Oh yeah I'm blind.

halter73:
@davidfowl I kept the old test exactly as it was with the CollectDump attribute.

jkotalik (approved):

PR #7087, opened 2019-01-29
Fixed copyright element

### LINKED ISSUE #7065: Copyright element in Microsoft.AspNetCore.Blazor.Cli is malformed
Looks like an encoding issue in the nuspec. Due to account restrictions, this is blocking us from uploading this package to nuget.org. We need to fix it so this starts with the &copy; character.

image

Potapy4:
Summary of the changes:
 - Updated copyright symbol
 - Changed file encoding to UTF-8 without BOM

Fixes #7065


natemcmaster:
Thanks @Potapy4!

PR #7091, opened 2019-01-29
Add DisposeAsync support to WebHost and RequestServices

pakrym:
https://github.com/aspnet/Extensions/issues/426

### DIFF
--- a/src/Hosting/Hosting/src/Internal/WebHost.cs
+++ b/src/Hosting/Hosting/src/Internal/WebHost.cs
@@ -343,7 +343,7 @@ namespace Microsoft.AspNetCore.Hosting.Internal
 
         public void Dispose()
         {
-            DisposeAsync().GetAwaiter().GetResult();
+            DisposeAsync().ConfigureAwait(false).GetAwaiter().GetResult();
         }
 
         public async ValueTask DisposeAsync()
@@ -352,7 +352,7 @@ namespace Microsoft.AspNetCore.Hosting.Internal
             {
                 try
                 {
-                    await StopAsync();
+                    await StopAsync().ConfigureAwait(false);
                 }
                 catch (Exception ex)
                 {
@@ -360,11 +360,11 @@ namespace Microsoft.AspNetCore.Hosting.Internal
                 }
             }
 
-            await DisposeServiceProvider(_applicationServices);
-            await DisposeServiceProvider(_hostingServiceProvider);
+            await DisposeServiceProviderAsync(_applicationServices).ConfigureAwait(false);
+            await DisposeServiceProviderAsync(_hostingServiceProvider).ConfigureAwait(false);
         }
 
-        private async ValueTask DisposeServiceProvider(IServiceProvider serviceProvider)
+        private async ValueTask DisposeServiceProviderAsync(IServiceProvider serviceProvider)
         {
             switch (serviceProvider)
             {
--- a/src/Hosting/Hosting/src/Internal/WebHost.cs
+++ b/src/Hosting/Hosting/src/Internal/WebHost.cs
@@ -24,7 +24,7 @@ using Microsoft.Extensions.StackTrace.Sources;
 
 namespace Microsoft.AspNetCore.Hosting.Internal
 {
-    internal class WebHost : IWebHost
+    internal class WebHost : IWebHost, IAsyncDisposable
     {
         private static readonly string DeprecatedServerUrlsKey = "server.urls";
 
@@ -343,11 +343,16 @@ namespace Microsoft.AspNetCore.Hosting.Internal
 
         public void Dispose()
         {
+            DisposeAsync().GetAwaiter().GetResult();
+        }
+
+        public async ValueTask DisposeAsync()
+        {
             if (!_stopped)
             {
                 try
                 {
-                    StopAsync().GetAwaiter().GetResult();
+                    await StopAsync();
                 }
                 catch (Exception ex)
                 {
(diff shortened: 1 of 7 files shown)


src/Http/Http.Abstractions/src/HttpResponse.cs
```diff
@@ -93,6 +98,12 @@ public abstract class HttpResponse
         /// <param name="disposable">The object to be disposed.</param>
         public virtual void RegisterForDispose(IDisposable disposable) => OnCompleted(_disposeDelegate, disposable);
 
+        /// <summary>
+        /// Registers an object for disposal by the host once the request has finished processing.
+        /// </summary>
+        /// <param name="disposable">The object to be disposed.</param>
+        public virtual void RegisterForDispose(IAsyncDisposable disposable) => OnCompleted(_disposeAsyncDelegate, disposable);
```
Tratcher: If the type implements both then the compiler will consider this ambiguous and error. RegisterForDisposeAsync?

src/Http/Http.Abstractions/src/HttpResponse.cs
```diff
@@ -93,6 +98,12 @@ public abstract class HttpResponse
         /// <param name="disposable">The object to be disposed.</param>
         public virtual void RegisterForDispose(IDisposable disposable) => OnCompleted(_disposeDelegate, disposable);
 
+        /// <summary>
+        /// Registers an object for disposal by the host once the request has finished processing.
```
Tratcher: For async dispose

Tratcher (commented):
You still need to update callers of webhost.Dispose

src/Hosting/Hosting/src/Internal/WebHost.cs
```diff
@@ -342,21 +342,39 @@ public async Task StopAsync(CancellationToken cancellationToken = default)
...
+        }
+
+        public async ValueTask DisposeAsync()
         {
             if (!_stopped)
             {
                 try
                 {
-                    StopAsync().GetAwaiter().GetResult();
+                    await StopAsync();
```
davidfowl: This needs a ConfigureAwait(false)

src/Hosting/Hosting/src/Internal/WebHost.cs
```diff
@@ -342,21 +342,39 @@ public async Task StopAsync(CancellationToken cancellationToken = default)
...
                 }
                 catch (Exception ex)
                 {
                     _logger?.ServerShutdownException(ex);
                 }
             }
 
-            (_applicationServices as IDisposable)?.Dispose();
-            (_hostingServiceProvider as IDisposable)?.Dispose();
+            await DisposeServiceProvider(_applicationServices);
```
davidfowl: These name ConfigureAwait(false) as well.

src/Hosting/Hosting/src/Internal/WebHost.cs
```diff
@@ -342,21 +342,39 @@ public async Task StopAsync(CancellationToken cancellationToken = default)
...
+            await DisposeServiceProvider(_hostingServiceProvider);
+        }
+
+        private async ValueTask DisposeServiceProvider(IServiceProvider serviceProvider)
```
davidfowl: nit: DisposeServiceProviderAsync

src/Http/Http.Abstractions/src/HttpResponse.cs
```diff
@@ -22,6 +22,11 @@ public abstract class HttpResponse
             return Task.CompletedTask;
         };
 
+        private static readonly Func<object, Task> _disposeAsyncDelegate = async disposable =>
```
davidfowl: Avoid the state machine. This will happen on every request.

src/Http/Http.Abstractions/src/HttpResponse.cs
davidfowl: Agreed. We should rename this to `RegisterForDisposeAsync` to avoid that ambiguity.

src/Hosting/Hosting/src/Internal/WebHost.cs
```diff
@@ -342,21 +342,39 @@ public async Task StopAsync(CancellationToken cancellationToken = default)
         }
 
         public void Dispose()
```
davidfowl: I know throwing in dispose is wrong but I'd rather do that then do sync over async...

pakrym: Talking about breaking changes, this one would be huge.

davidfowl: It's not that big of a breaking change as it would you to manually dispose this feature which nobody does. What am I missing?

davidfowl:
This is waiting on C# 8

src/Hosting/Hosting/src/Internal/WebHost.cs
pakrym: This is not a feature, this is IWebHost implementation. The feature is below and yeah, it seems we can throw there.

pakrym:
Not waiting for C# 8

src/Http/Http/src/Features/RequestServicesFeature.cs
```diff
@@ -41,11 +42,25 @@ public IServiceProvider RequestServices
...
+                    break;
+            }
+
             _scope = null;
             _requestServices = null;
         }
+
+        public void Dispose()
+        {
+            DisposeAsync().ConfigureAwait(false).GetAwaiter().GetResult();
```
Tratcher: So you want this to throw?

Tratcher (approved):

src/Hosting/Hosting/src/WebHostExtensions.cs
```diff
@@ -131,6 +131,17 @@ private static async Task RunAsync(this IWebHost host, CancellationToken token,
 
                 await host.WaitForTokenShutdownAsync(token);
             }
+            finally
+            {
+                if (host is IAsyncDisposable asyncDisposable)
+                {
+                    await asyncDisposable.DisposeAsync();
```
davidfowl: ConfigureAwait.

src/Http/Http/src/Features/RequestServicesFeature.cs
davidfowl: I'm ok with this not throwing. It should be super rare now since we by default will dispose it async.

PR #7094, opened 2019-01-29
Remove reconnect from C++ Client

BrennanConroy:
Starting to cleanup the client.
Removed all the reconnect stuff because it was a big chunk of code that isn't used and the reconnect tests were causing test crashes.
Cleaned up a bunch of tests to expect the new protocol instead of the old one. We were at ~100 failing tests before now down to ~25.


src/SignalR/clients/cpp/src/signalrclient/connection_impl.cpp
```diff
@@ -40,7 +40,7 @@ namespace signalr
         : m_base_url(url), m_query_string(query_string), m_connection_state(connection_state::disconnected), m_reconnect_delay(2000),
         m_logger(log_writer, trace_level), m_transport(nullptr), m_web_request_factory(std::move(web_request_factory)),
         m_transport_factory(std::move(transport_factory)), m_message_received([](const web::json::value&){}),
-        m_reconnecting([](){}), m_reconnected([](){}), m_disconnected([](){}), m_handshakeReceived(false)
+        m_disconnected([](){}), m_handshakeReceived(false)
     { }
 
     connection_impl::~connection_impl()
```
mikaelm12: The large comment block below has a few references to reconnect.

src/SignalR/clients/cpp/src/signalrclient/connection_impl.cpp
```diff
@@ -40,7 +40,7 @@ namespace signalr
         : m_base_url(url), m_query_string(query_string), m_connection_state(connection_state::disconnected), m_reconnect_delay(2000),
```
mikaelm12: `m_reconnect_delay(2000)` here

mikaelm12 (commented):
There are also other comments that have references to reconnect. Do a global search for the term.

src/SignalR/clients/cpp/src/signalrclient/connection_impl.cpp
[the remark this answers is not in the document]
mikaelm12: And then follow where this flows

src/SignalR/clients/cpp/src/signalrclient/hub_connection_impl.cpp
```diff
@@ -270,27 +244,11 @@ namespace signalr
...
                         std::make_exception_ptr(
-                            signalr_exception(message.at(_XPLATSTR("error")).serialize())));
+                            hub_exception(message.at(_XPLATSTR("error")).serialize())));
                 }
 
                 set_result(json::value::null());
             };
         }
-
-        static utility::string_t adapt_url(const utility::string_t& url, bool use_default_url)
```
halter73: Why was this needed before but not now?

BrennanConroy: I think I removed it's usage a long time ago when I ported the client. It's not referenced anywhere which is why I removed it here.

BrennanConroy:
Ping

src/SignalR/clients/cpp/src/signalrclient/url_builder.cpp
```diff
@@ -80,7 +80,7 @@ namespace signalr
             const utility::string_t& connection_data, const utility::string_t& query_string,
             const utility::string_t& last_message_id = _XPLATSTR(""), const utility::string_t& groups_token = _XPLATSTR(""))
         {
-            _ASSERTE(command == _XPLATSTR("reconnect") || (last_message_id.length() == 0 && groups_token.length() == 0));
```
halter73: There are still a bunch of commented-out reconnect tests in url_builder_tests.cpp. Can we remove the entire url_builder_tests.cpp file? Or at least the reconnect test portions of it?

src/SignalR/clients/cpp/test/signalrclient-e2e-tests/hub_connection_tests.cpp
```diff
@@ -331,17 +236,6 @@ TEST(hub_connection_tests, connection_id_start_stop_start_reconnect)
     ASSERT_NE(hub_conn->get_connection_id(), connection_id);
 
     connection_id = hub_conn->get_connection_id();
-
-    try
-    {
-        hub_conn->send(U("forceReconnect")).get();
```
halter73: It looks like nothing calls HubConnection.ForceReconnect anymore in signalrclient-testhost. Should it be removed?

halter73 (changes requested):

halter73 (approved):
We can take care of the nits in #7160

PR #7099, opened 2019-01-29
Move Cors Functional tests to a separate test group

pranavkm:
Attemp #2 at addressing https://github.com/aspnet/AspNetCore-Internal/issues/1619

Puppetteer will attempt to download chromium as part of running the test. This seems odd, but
it might explain why this might be afflicted with the file descriptor contention issues that
the mondo repo tests encounter.

Moving these out in to a separate test group to see if this helps

Possible fix for https://github.com/aspnet/AspNetCore-Internal/issues/1619

### DIFF
--- a/src/Middleware/CORS/test/FunctionalTests/FunctionalTests.csproj
+++ b/src/Middleware/CORS/test/FunctionalTests/FunctionalTests.csproj
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <TargetFrameworks>netcoreapp2.2;net461</TargetFrameworks>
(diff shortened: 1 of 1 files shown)


pranavkm:
This build - http://aspnetci/viewLog.html?buildId=618439&buildTypeId=Releases_22xPublic_OsxUniverse - does provide clues that the timeout is from `puppeteer.launch` taking over 3 mins. Running the test in isolation they never fail on the same machine.

ajaybhargavb (approved):
Worth a shot

PR #7104, opened 2019-01-29
Remove ANCM V1 leftovers

pakrym:


jkotalik:
Does building the installers locally succeed?

pakrym:
In progress of verifying that.

natemcmaster (approved):

pakrym:
Passed in teamcity http://aspnetci/viewLog.html?buildId=618796&tab=buildResultsDiv&buildTypeId=UniverseCoherence

Issue #4260, opened 2018-11-27
Remove Newtonsoft.Json from the shared framework


natemcmaster:
The following shared-framework libraries pull in Newtonsoft.Json:

* [x] - Microsoft.AspNetCore.Mvc.ViewFeatures
* [x] - Microsoft.Extensions.Configuration.Json
* [x] - Microsoft.Extensions.DependencyModel
* [x] - Microsoft.IdentityModel.JsonWebTokens
* [x] - Microsoft.AspNetCore.SignalR.Protocols.Json
* [x] - Microsoft.AspNetCore.Authentication.OAuth
* [x] - Microsoft.Extensions.Logging.EventSource

These assemblies and dependencies need to be updated to remove their dependency on Newtonsoft.Json.

Tratcher:
What about Microsoft.AspNetCore.Authentication.OAuth?

natemcmaster:
Tracking external issues:

* https://github.com/dotnet/core-setup/issues/3311
* https://github.com/aspnet/Extensions/issues/568
* https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/issues/1060

Internal changes needed:

* @Tratcher - update Microsoft.AspNetCore.Authentication.OAuth
* @rynowak @mkArtakMSFT - we need to update Microsoft.AspNetCore.Mvc.ViewFeatures to make sure it doesn't pull in Newtonsoft.Json.
* Update our dependency on IdentityModel.

natemcmaster:
Also adding @mkArtakMSFT to help with MVC's dependency on Json.net and myself to coordinate with partners on getting Json.net removed.

natemcmaster:
Pushing out to preview3. We are still blocked on https://github.com/dotnet/core-setup/issues/3311 and https://github.com/aspnet/Extensions/issues/568

Tratcher:
Partially unblocked by https://github.com/dotnet/corefx/pull/34485

muratg:
Adding @BrennanConroy for `Microsoft.AspNetCore.SignalR.Protocols.Json`

BrennanConroy:
`JsonDocument` doesn't help us, we need the serializer and deserializer :)

BrennanConroy:
Also, was there ever a discussion/decision for the netstandard issue?

muratg:
@joshfree @ahsonkhan @steveharter?

davidfowl:
[...]

PR #7105, opened 2019-01-29
Replace JObject with JsonDocument in Authentication

Tratcher:
#4260
- "Drop in" replacement with some odd gymnastics sometimes.
- It's not as breaking as I thought it might be for derived OAuth handlers. ClaimActions shield them from most of the changes. They primarily have to call Parse with the new type. @PinpointTownes 
- Users doing advanced claims mapping or inspection via events may be broken.
- Also fixed launch settings so the samples would work again.

### DIFF
--- a/src/Security/Authentication/Core/src/JsonDocumentAuthExtensions.cs
+++ b/src/Security/Authentication/Core/src/JsonDocumentAuthExtensions.cs
@@ -1,0 +1,14 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Text.Json;
+
+namespace Microsoft.AspNetCore.Authentication
+{
+    public static class JsonDocumentAuthExtensions
+    {
+        public static string GetString(this JsonDocument document, string key) =>
+            document.RootElement.TryGetProperty(key, out var property)
+                ? property.ToString() : null;
+    }
+}
--- a/src/Security/Authentication/Core/src/JsonDocumentAuthExtensions.cs
+++ b/src/Security/Authentication/Core/src/JsonDocumentAuthExtensions.cs
@@ -7,9 +7,6 @@ namespace Microsoft.AspNetCore.Authentication
 {
     public static class JsonDocumentAuthExtensions
     {
-        public static string GetString(this JsonDocument document, string key) =>
-            document.RootElement.GetString(key);
-
         public static string GetString(this JsonElement element, string key) =>
             element.TryGetProperty(key, out var property)
                 ? property.ToString() : null;
--- a/src/Security/Authentication/Core/src/JsonDocumentAuthExtensions.cs
+++ b/src/Security/Authentication/Core/src/JsonDocumentAuthExtensions.cs
@@ -8,7 +8,10 @@ namespace Microsoft.AspNetCore.Authentication
     public static class JsonDocumentAuthExtensions
     {
         public static string GetString(this JsonDocument document, string key) =>
-            document.RootElement.TryGetProperty(key, out var property)
+            document.RootElement.GetString(key);
+
+        public static string GetString(this JsonElement element, string key) =>
+            element.TryGetProperty(key, out var property)
                 ? property.ToString() : null;
     }
 }
--- a/src/Security/Authentication/Facebook/src/FacebookHandler.cs
+++ b/src/Security/Authentication/Facebook/src/FacebookHandler.cs
@@ -8,12 +8,12 @@ using System.Security.Claims;
 using System.Security.Cryptography;
 using System.Text;
 using System.Text.Encodings.Web;
+using System.Text.Json;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Authentication.OAuth;
 using Microsoft.AspNetCore.WebUtilities;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
-using Newtonsoft.Json.Linq;
 
 namespace Microsoft.AspNetCore.Authentication.Facebook
 {
@@ -41,15 +41,13 @@ namespace Microsoft.AspNetCore.Authentication.Facebook
                 throw new HttpRequestException($"An error occurred when retrieving Facebook user information ({response.StatusCode}). Please check if the authentication information is correct and the corresponding Facebook Graph API is enabled.");
             }
 
-            var payload = JObject.Parse(await response.Content.ReadAsStringAsync());
-
-            var context = new OAuthCreatingTicketContext(new ClaimsPrincipal(identity), properties, Context, Scheme, Options, Backchannel, tokens, payload);
-            context.RunClaimActions();
-
-            await Events.CreatingTicket(context);
-
-            return new AuthenticationTicket(context.Principal, context.Properties, Scheme.Name);
-
+            using (var payload = JsonDocument.Parse(await response.Content.ReadAsStringAsync()))
+            {
+                var context = new OAuthCreatingTicketContext(new ClaimsPrincipal(identity), properties, Context, Scheme, Options, Backchannel, tokens, payload);
+                context.RunClaimActions();
+                await Events.CreatingTicket(context);
+                return new AuthenticationTicket(context.Principal, context.Properties, Scheme.Name);
+            }
         }
 
         private string GenerateAppSecretProof(string accessToken)
--- a/src/Security/Authentication/Facebook/src/FacebookHandler.cs
+++ b/src/Security/Authentication/Facebook/src/FacebookHandler.cs
@@ -43,7 +43,7 @@ namespace Microsoft.AspNetCore.Authentication.Facebook
 
             using (var payload = JsonDocument.Parse(await response.Content.ReadAsStringAsync()))
             {
-                var context = new OAuthCreatingTicketContext(new ClaimsPrincipal(identity), properties, Context, Scheme, Options, Backchannel, tokens, payload);
+                var context = new OAuthCreatingTicketContext(new ClaimsPrincipal(identity), properties, Context, Scheme, Options, Backchannel, tokens, payload.RootElement);
                 context.RunClaimActions();
                 await Events.CreatingTicket(context);
                 return new AuthenticationTicket(context.Principal, context.Properties, Scheme.Name);
--- a/src/Security/Authentication/Google/src/GoogleHandler.cs
+++ b/src/Security/Authentication/Google/src/GoogleHandler.cs
@@ -7,12 +7,12 @@ using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Security.Claims;
 using System.Text.Encodings.Web;
+using System.Text.Json;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Authentication.OAuth;
 using Microsoft.AspNetCore.WebUtilities;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
-using Newtonsoft.Json.Linq;
 
 namespace Microsoft.AspNetCore.Authentication.Google
 {
@@ -37,13 +37,14 @@ namespace Microsoft.AspNetCore.Authentication.Google
                 throw new HttpRequestException($"An error occurred when retrieving Google user information ({response.StatusCode}). Please check if the authentication information is correct and the corresponding Google+ API is enabled.");
             }
 
-            var payload = JObject.Parse(await response.Content.ReadAsStringAsync());
+            using (var payload = JsonDocument.Parse(await response.Content.ReadAsStringAsync()))
+            {
 
-            var context = new OAuthCreatingTicketContext(new ClaimsPrincipal(identity), properties, Context, Scheme, Options, Backchannel, tokens, payload);
-            context.RunClaimActions();
-
-            await Events.CreatingTicket(context);
-            return new AuthenticationTicket(context.Principal, context.Properties, Scheme.Name);
+                var context = new OAuthCreatingTicketContext(new ClaimsPrincipal(identity), properties, Context, Scheme, Options, Backchannel, tokens, payload);
+                context.RunClaimActions();
+                await Events.CreatingTicket(context);
+                return new AuthenticationTicket(context.Principal, context.Properties, Scheme.Name);
+            }
         }
 
         // TODO: Abstract this properties override pattern into the base class?
--- a/src/Security/Authentication/Google/src/GoogleHandler.cs
+++ b/src/Security/Authentication/Google/src/GoogleHandler.cs
@@ -39,8 +39,7 @@ namespace Microsoft.AspNetCore.Authentication.Google
 
             using (var payload = JsonDocument.Parse(await response.Content.ReadAsStringAsync()))
             {
-
-                var context = new OAuthCreatingTicketContext(new ClaimsPrincipal(identity), properties, Context, Scheme, Options, Backchannel, tokens, payload);
+                var context = new OAuthCreatingTicketContext(new ClaimsPrincipal(identity), properties, Context, Scheme, Options, Backchannel, tokens, payload.RootElement);
                 context.RunClaimActions();
                 await Events.CreatingTicket(context);
                 return new AuthenticationTicket(context.Principal, context.Properties, Scheme.Name);
--- a/src/Security/Authentication/OAuth/src/Events/OAuthCreatingTicketContext.cs
+++ b/src/Security/Authentication/OAuth/src/Events/OAuthCreatingTicketContext.cs
@@ -25,27 +25,6 @@ namespace Microsoft.AspNetCore.Authentication.OAuth
         /// <param name="options">The options used by the authentication middleware.</param>
         /// <param name="backchannel">The HTTP client used by the authentication middleware</param>
         /// <param name="tokens">The tokens returned from the token endpoint.</param>
-        public OAuthCreatingTicketContext(
-            ClaimsPrincipal principal,
-            AuthenticationProperties properties,
-            HttpContext context,
-            AuthenticationScheme scheme,
-            OAuthOptions options,
-            HttpClient backchannel,
-            OAuthTokenResponse tokens)
-            : this(principal, properties, context, scheme, options, backchannel, tokens, user: JsonDocument.Parse("{}"))
-        { }
-
-        /// <summary>
-        /// Initializes a new <see cref="OAuthCreatingTicketContext"/>.
-        /// </summary>
-        /// <param name="principal">The <see cref="ClaimsPrincipal"/>.</param>
-        /// <param name="properties">The <see cref="AuthenticationProperties"/>.</param>
-        /// <param name="context">The HTTP environment.</param>
-        /// <param name="scheme">The authentication scheme.</param>
-        /// <param name="options">The options used by the authentication middleware.</param>
-        /// <param name="backchannel">The HTTP client used by the authentication middleware</param>
-        /// <param name="tokens">The tokens returned from the token endpoint.</param>
         /// <param name="user">The JSON-serialized user.</param>
         public OAuthCreatingTicketContext(
             ClaimsPrincipal principal,
--- a/src/Security/Authentication/OAuth/src/Events/OAuthCreatingTicketContext.cs
+++ b/src/Security/Authentication/OAuth/src/Events/OAuthCreatingTicketContext.cs
@@ -33,7 +33,7 @@ namespace Microsoft.AspNetCore.Authentication.OAuth
             OAuthOptions options,
             HttpClient backchannel,
             OAuthTokenResponse tokens)
-            : this(principal, properties, context, scheme, options, backchannel, tokens, user: JsonDocument.Parse(string.Empty))
+            : this(principal, properties, context, scheme, options, backchannel, tokens, user: JsonDocument.Parse("{}"))
         { }
 
         /// <summary>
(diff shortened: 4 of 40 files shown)


src/Security/Authentication/OAuth/src/JsonKeyClaimAction.cs
```diff
@@ -30,20 +30,27 @@ public JsonKeyClaimAction(string claimType, string valueType, string jsonKey)
...
+                return;
             }
-            else if (value is JArray)
+            if (value.Type == JsonValueType.Array)
             {
-                foreach (var v in value)
+                foreach (var v in value.EnumerateArray())
                 {
-                    AddClaim(v?.ToString(), identity, issuer);
+                    AddClaim(v.ToString(), identity, issuer);
```
JamesNK: What happens if a value in the array is a null token? I think add a test

Tratcher:
@ahsonkhan

Tratcher:
If we get the C# 8 infrastructure working soon then we can use `using var` to clean up all of these new using blocks.

src/Security/Authentication/OAuth/src/Events/OAuthCreatingTicketContext.cs
```diff
@@ -33,7 +33,7 @@ public class OAuthCreatingTicketContext : ResultContext<OAuthOptions>
             OAuthOptions options,
             HttpClient backchannel,
             OAuthTokenResponse tokens)
-            : this(principal, properties, context, scheme, options, backchannel, tokens, user: new JObject())
+            : this(principal, properties, context, scheme, options, backchannel, tokens, user: JsonDocument.Parse(string.Empty))
```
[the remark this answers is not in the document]
JamesNK: FYI `JsonDocument.Parse(string.Empty)` and `new JObject()` aren't necessarily equivalent.

`new JObject()` = an empty JSON object

`JsonDocument.Parse("{}")` would be closer. I don't know if it matters in this case.

src/Security/Authentication/OAuth/src/JsonKeyClaimAction.cs
Tratcher: Added a test.

src/Security/Authentication/OAuth/src/Events/OAuthCreatingTicketContext.cs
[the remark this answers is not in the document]
ahsonkhan: `JsonDocument.Parse(string.Empty)`

I don't think this will work.

An empty json string is considered invalid and hence we throw an exception:
https://github.com/dotnet/corefx/blob/5197c8dc78adcb130a08944724f1b49c3765a4b1/src/System.Text.Json/tests/JsonDocumentTests.cs#L1204

Also, given that JsonDocument is readonly, I don't know if it is useful to create an empty JsonDocument.

`JsonDocument.Parse("{}")` <= this will work, but it is odd to see.

cc @bartonjs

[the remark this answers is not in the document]
Tratcher: Changed to `"{}"`.  Having an empty document placeholder can make usage simpler by being able to exclude a lot of null checks.

src/Security/Authentication/Core/src/JsonDocumentAuthExtensions.cs
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Text.Json;
+
+namespace Microsoft.AspNetCore.Authentication
+{
+    public static class JsonDocumentAuthExtensions
```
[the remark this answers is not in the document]
PinpointTownes: This extension looks so basic that it seems strange it's not directly part of `JsonDocument`. Is such an API planned?

src/Security/Authentication/Facebook/src/FacebookHandler.cs
```diff
@@ -41,15 +41,13 @@ protected override async Task<AuthenticationTicket> CreateTicketAsync(ClaimsIden
...
-            var payload = JObject.Parse(await response.Content.ReadAsStringAsync());
-
-            var context = new OAuthCreatingTicketContext(new ClaimsPrincipal(identity), properties, Context, Scheme, Options, Backchannel, tokens, payload);
-            context.RunClaimActions();
-
-            await Events.CreatingTicket(context);
-
-            return new AuthenticationTicket(context.Principal, context.Properties, Scheme.Name);
-
+            using (var payload = JsonDocument.Parse(await response.Content.ReadAsStringAsync()))
```
PinpointTownes: Out of curiosity, do you happen to know why `JsonDocument` implements `IDisposable`?

PinpointTownes: (at some point, we may want to move away from `ReadAsStringAsync()` in both the "official" and aspnet-contrib providers. I'm even surprised @davidfowl didn't cringe about the fact it's allocatey as hell :smile:)

src/Security/Authentication/test/ClaimActionTests.cs
```diff
@@ -17,10 +14,7 @@ public class ClaimActionTests
         [Fact]
         public void CanMapSingleValueUserDataToClaim()
         {
-            var userData = new JObject
-            {
-                ["name"] = "test"
-            };
+            var userData = JsonDocument.Parse("{ \"name\": \"test\" }");
```
PinpointTownes: That's sad we have to do that. Is a "mutable" `JsonDocument` API planned for future versions of the new JSON stack?

kevinchalet (commented):
@Tratcher thanks for tagging, it's much appreciated :+1: 

The new JSON stack looks promising and it's good to know we now have a `JObject`-like API, which is much more convenient than directly using a low-level JSON reader.

When you have a moment, you may want to take a look at the aspnet-contrib providers to ensure we have everything we need. Some of our providers use some nasty hacks to work around non-standard implementations, like re-creating `JObject` from query string parameters (e.g StackExchange) or extracting the user profile from the token response (e.g Myob).

src/Security/Authentication/Core/src/JsonDocumentAuthExtensions.cs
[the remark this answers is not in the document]
Tratcher: Yes, we use it across multiple product assemblies, samples, tests, etc., and it's something our customers will want for easy migration.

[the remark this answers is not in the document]
Tratcher: I did keep the namespace constrained to only auth users though.

src/Security/Authentication/Facebook/src/FacebookHandler.cs
Tratcher: It's disposable because it uses pooled buffers.

Sticking with ReadAsStringAsync might be a good idea. I know it's a bit less efficient but it does handle a bunch of decoding issues for us. The new json reader is hardcoded to UTF8.

src/Security/Authentication/Core/src/JsonDocumentAuthExtensions.cs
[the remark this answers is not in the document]
pranavkm: It feels like the wrong place for the extension to exists. If it's general purpose enough that your code and user code would benefit from it, we should move it down-level. @ahsonkhan could we consider adding this as an extension in System.Text.Json?

src/Security/Authentication/Facebook/src/FacebookHandler.cs
davidfowl: Why would we keep using ReadAsStringAsync ?

ahsonkhan (commented):
Given `JsonDocument` is disposable, we should consider who ends up disposing it, and consider passing `JsonElement` around that points to the sub-payload rather than the whole `JsonDocument`. Granted there is a usability burden here, but its a side effect of having a high-performance implementation which emphasizes on low allocations and performance.

src/Security/Authentication/Facebook/src/FacebookHandler.cs
davidfowl: https://github.com/dotnet/corefx/blob/28e216680a990ce4a58798aa0d2026c9d3e1f148/src/System.Net.Http/src/System/Net/Http/HttpContent.cs#L189-L245

How about this,  if the encoding is UTF8, use the fast path (don't go byte[] -> string -> JsonDocument), if it's not UTF8, in those 2% cases, then use ReadAsStringAsync.

davidfowl: PS: The auth stack could do with some microbenchmarks

src/Security/Authentication/Core/src/JsonDocumentAuthExtensions.cs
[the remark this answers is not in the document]
ahsonkhan: See my comment here: https://github.com/aspnet/AspNetCore/pull/7105#discussion_r252091383
I am not sure why such an extension method exists. And if the goal is primarily to create a custom "null" string behavior, it doesn't make sense to provide an in-box implementation for that.

`JsonElement.GetString()` already returns null when the token type is null.

src/Security/Authentication/Facebook/src/FacebookHandler.cs
ahsonkhan: Yep, see https://github.com/aspnet/AspNetCore/pull/7105#discussion_r252088444

src/Security/Authentication/Core/src/JsonDocumentAuthExtensions.cs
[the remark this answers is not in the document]
Tratcher: It's not special casing null strings, it's special casing missing properties and avoids the `KeyNotFoundException : The given key was not present in the dictionary.` thrown by GetProperty for missing elements. The extension is used 8 times in product code and 5 in samples, and I expect 3rd party OAuth implementations to use it as well. @ahsonkhan I can see why you wouldn't add this directly to JsonDocument, but it suits our scenarios well.

Consider the example here:
https://github.com/aspnet/AspNetCore/pull/7105/files#diff-a3f5673336ed9c217610a732aec6cdcfL14
This extension gives us a drop in replacement for prior functionality. The TryGetProperty pattern encapsulated by the extension is x3 as long.

src/Security/Authentication/Facebook/src/FacebookHandler.cs
Tratcher: ReadAsStringAsync deals with charsets and BOMs for us. ReadBufferAsString is exactly the method we don't want to re-implement, it would be useful if it were public.

This code path is not sufficiently perf sensitive to warrant a fast and slow path. Correctness is far more important and easier to maintain with a single code path.

davidfowl: @Tratcher We need performance tests for auth, we have none today and we know it performs poorly. I buy that it isn't on the hot path but lets not get lazy with our performance work. It's a tiny set of code to make it do a better thing in the 90% case. We can always fallback in the other scenarios.

Tratcher: Yes we need perf tests for auth that runs on most requests like Cookie & Bearer. OAuth and OIDC are much lower priority when the results are cached for weeks.

src/Security/Authentication/Google/src/GoogleHandler.cs
```diff
@@ -37,13 +37,14 @@ public GoogleHandler(IOptionsMonitor<GoogleOptions> options, ILoggerFactory logg
                 throw new HttpRequestException($"An error occurred when retrieving Google user information ({response.StatusCode}). Please check if the authentication information is correct and the corresponding Google+ API is enabled.");
             }
 
-            var payload = JObject.Parse(await response.Content.ReadAsStringAsync());
-
-            var context = new OAuthCreatingTicketContext(new ClaimsPrincipal(identity), properties, Context, Scheme, Options, Backchannel, tokens, payload);
-            context.RunClaimActions();
+            using (var payload = JsonDocument.Parse(await response.Content.ReadAsStringAsync()))
```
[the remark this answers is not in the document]
Tratcher: See above https://github.com/aspnet/AspNetCore/pull/7105#discussion_r252065241

src/Security/Authentication/OAuth/src/JsonKeyClaimAction.cs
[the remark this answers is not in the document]
Tratcher: We don't have grantees about the contents, but our end result needs to be a string. AddClaim handles null or empty values.

src/Security/Authentication/Core/src/JsonDocumentAuthExtensions.cs
```diff
@@ -0,0 +1,14 @@
...
+    {
+        public static string GetString(this JsonDocument document, string key) =>
```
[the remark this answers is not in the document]
Tratcher: I've added one for JsonElement, but we still have usages for JsonDocument since much of the time we're looking for top level properties.

src/Security/Authentication/Core/src/JsonDocumentAuthExtensions.cs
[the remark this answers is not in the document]
ahsonkhan: > This extension gives us a drop in replacement for prior functionality.

But will likely result in quite slow perf since the extension is on `JsonDocument`.

In your example, it might make more sense for the `response` to be a `JsonElement`, not `JsonDocument`.

```C#
// Before:
AccessToken = response.Value<string>("access_token");
```

```C#
// After:
AccessToken = response.GetString("access_token");
```

I understand the usability concern. We currently do not have an API that searches the `JsonElement` for a property and returns the .NET type value back to you in one call. **We can consider adding an API similar to `Value<T>` to help alleviate some of the concern here. Thanks for the feedback :)**

I would appreciate it if you can file an issue in corefx to track this. Otherwise, I will do so.

One thing we should consider is making sure we are consistent around "null" or key not found behavior.

```C#
var sample = JObject.Parse("{\"message\": 42}");
JToken token = sample["foo"]; // Why shouldn't this throw?
string value = sample.Value<string>("foo"); // And if the above throws, this returning null might be confusing
```

src/Security/Authentication/Core/src/JsonDocumentAuthExtensions.cs
```diff
@@ -0,0 +1,17 @@
...
+            document.RootElement.GetString(key);
+
+        public static string GetString(this JsonElement element, string key) =>
+            element.TryGetProperty(key, out var property)
+                ? property.ToString() : null;
```
ahsonkhan: `GetString()` not `ToString()`

If the `JsonElement` is not of `Type` `JsonValueType.String`, the results of `ToString()` are likely not what you want here (unless I am mistaken).

ToString() returns the entire sub-object for object/arrays and returns "true", "false" for bools, and "" for nulls.

src/Security/Authentication/OAuth/src/JsonKeyClaimAction.cs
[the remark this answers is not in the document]
ahsonkhan: Could you have nested objects/arrays in the array? Can it be bool literals like true/false or numbers?

src/Security/Authentication/Core/src/JsonDocumentAuthExtensions.cs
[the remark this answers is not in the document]
Tratcher: After I switched to JsonElement most of the remaining usage is in samples. I can remove the JsonDocument overload.

src/Security/Authentication/OAuth/src/MapAllClaimsAction.cs
```diff
@@ -17,24 +17,24 @@ public MapAllClaimsAction() : base("All", ClaimValueTypes.String)
...
         {
             if (userData == null)
             {
                 return;
             }
-            foreach (var pair in userData)
+            foreach (var pair in userData.RootElement.EnumerateObject())
             {
-                var claimValue = userData.TryGetValue(pair.Key, out var value) ? value.ToString() : null;
+                var claimValue = userData.GetString(pair.Name);
```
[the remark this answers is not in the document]
ahsonkhan: I am curious to know, what's preventing you from doing the following:
```C#
var claimValue = pair.Value.GetString();
```

Currently, you are enumerating the payload twice when you call your `GetString()` extension method

src/Security/Authentication/OAuth/src/OAuthHandler.cs
```diff
@@ -177,7 +178,7 @@ protected virtual async Task<OAuthTokenResponse> ExchangeCodeAsync(string code,
             var response = await Backchannel.SendAsync(requestMessage, Context.RequestAborted);
             if (response.IsSuccessStatusCode)
             {
-                var payload = JObject.Parse(await response.Content.ReadAsStringAsync());
+                var payload = JsonDocument.Parse(await response.Content.ReadAsStringAsync());
```
ahsonkhan: Who disposes this?

ahsonkhan: Oh, `OAuthTokenResponse` is now disposable. I see. So the caller must remember to dispose it. K :)

src/Security/Authentication/OpenIdConnect/samples/OpenIdConnectSample/Startup.cs
```diff
@@ -211,8 +211,8 @@ await context.SignOutAsync(OpenIdConnectDefaults.AuthenticationScheme, new Authe
                     using (var payload = JsonDocument.Parse(await tokenResponse.Content.ReadAsStringAsync()))
                     {
                         // Persist the new acess token
-                        props.UpdateTokenValue("access_token", payload.GetString("access_token"));
-                        props.UpdateTokenValue("refresh_token", payload.GetString("refresh_token"));
+                        props.UpdateTokenValue("access_token", payload.RootElement.GetString("access_token"));
```
ahsonkhan: nit: cache `RootElement` in a local wherever its references 3+ times.

ahsonkhan (approved):
Other than `ToString()` usage and some redundant enumeration, LGTM!

src/Security/Authentication/Core/src/JsonDocumentAuthExtensions.cs
Tratcher: No, `.ToString()` is the equivalent of `.Value<string>()`. The destination for these values are represented as strings, the underlying data type is not significant. Yes it would be odd to get a full object but the caller usually knows their expected structure well enough to avoid that (or wants it to happen).

ahsonkhan: Got it. Thanks for clarifying :)

src/Security/Authentication/OAuth/src/MapAllClaimsAction.cs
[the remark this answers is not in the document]
Tratcher: Yeah, this code looks like it was strange/broken before the conversion too. I'll fix it up.

src/Security/Authentication/Core/src/JsonDocumentAuthExtensions.cs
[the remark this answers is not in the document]
Tratcher: Filed https://github.com/dotnet/corefx/issues/34997

Note AspNetCore has banned indexers that throw key not found exceptions, they're a significant usability barrier. All indexers return null or defaults for missing values. If that's not appropriate for a type then that type does not get an indexer.

ahsonkhan (approved):

Tratcher:
@HaoK rebased and ready for a final review.

src/Security/Authentication/OpenIdConnect/src/OpenIdConnectHandler.cs
```diff
@@ -853,42 +856,46 @@ protected virtual async Task<OpenIdConnectMessage> RedeemAuthorizationCodeAsync(
...
             if (contentType.MediaType.Equals("application/json", StringComparison.OrdinalIgnoreCase))
             {
-                user = JObject.Parse(userInfoResponse);
+                user = JsonDocument.Parse(userInfoResponse);
             }
             else if (contentType.MediaType.Equals("application/jwt", StringComparison.OrdinalIgnoreCase))
             {
                 var userInfoEndpointJwt = new JwtSecurityToken(userInfoResponse);
-                user = JObject.FromObject(userInfoEndpointJwt.Payload);
+                user = JsonDocument.Parse(userInfoEndpointJwt.Payload.SerializeToJson());
```
davidfowl: This is kinda gross. This isn't a hot path either right?  

cc @ahsonkhan Some potential API feedback here.

davidfowl: Still, its is just a crapping thing to have to do.

src/Security/Authentication/OpenIdConnect/src/OpenIdConnectHandler.cs
```diff
@@ -853,42 +856,46 @@ protected virtual async Task<OpenIdConnectMessage> RedeemAuthorizationCodeAsync(
...
-            if (userInformationReceivedContext.Result != null)
-            {
-                return userInformationReceivedContext.Result;
-            }
-            principal = userInformationReceivedContext.Principal;
-            properties = userInformationReceivedContext.Properties;
-            user = userInformationReceivedContext.User;
-
-            Options.ProtocolValidator.ValidateUserInfoResponse(new OpenIdConnectProtocolValidationContext()
+            using (user)
```
HaoK: This new using is unfortunate but I guess we can't avoid it :/

HaoK (approved):

src/Security/Authentication/OpenIdConnect/src/OpenIdConnectHandler.cs
Tratcher: Most of them can get squashed with the new C# 8 using syntax.

PR #7110, opened 2019-01-29
Expose response PipeWriter in Kestrel

jkotalik:
For https://github.com/aspnet/AspNetCore/issues/4351 and does many other things too that are required for this change.

- Exposes the PipeWriter as directly as possible from Kestrel (for both Http1 and Http2)
  - Pay careful attention to Http1 chunking logic as that actually required work.
- Changes most tests to start using the PipeWriter

Things left to do:
- Add tests for setting response body/ response pipe and how they override each other.
- Perf tests (micro benchmarks and before/after with benchmarks).

Have fun everyone 😄


jkotalik (commented):
Some feedback for myself and things to figure out.

jkotalik:
Looks like swapping WriteAsync broke stuff too. I think they are issues with StreamPipeWriter so I'll investigate and create a separate PR.

JunTaoLuo:
👀

src/Http/Http.Abstractions/src/Extensions/HttpResponseWritingExtensions.cs
```diff
@@ -61,7 +61,7 @@ public static Task WriteAsync(this HttpResponse response, string text, Encoding
             }
 
             byte[] data = encoding.GetBytes(text);
-            return response.Body.WriteAsync(data, 0, data.Length, cancellationToken);
+            return response.BodyPipe.WriteAsync(new Memory<byte>(data, 0, data.Length), cancellationToken).AsTask();
```
davidfowl: A couple of things:
- This ends up creating a temporary buffer
- It allocates a new task

We need to do a few things here:
- Use GetMemory() and write directly into it
- Don't use `AsTask()` unless the ValueTask returned from FlushAsync Isn't completed.
  - Actually, make a local function that does the await, instead of using AsTask.

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
```diff
@@ -188,5 +277,56 @@ public Task Write100ContinueAsync()
...
+                    // If the current chunk of memory isn't completely utilized, we need to copy the contents forwards.
+                    // This occurs if someone uses less than 255 bytes of the current Memory segment.
+                    // Therefore, we need to copy it forwards by either 1 or 2 bytes (depending on number of bytes)
+                    _currentChunkMemory.Slice(BeginChunkLengthMax, _advancedBytesForChunk).CopyTo(_currentChunkMemory.Slice(bytesWritten));
+                }
+
+                writer.Write(_currentChunkMemory.Slice(bytesWritten, _advancedBytesForChunk).Span);
+                writer.WriteEndChunkBytes();
+                writer.Commit();
+                _advancedBytesForChunk = 0;
```
Tratcher: _currentChunkMemory needs to be sliced / advanced or dropped.

Tratcher (dismissed):

jkotalik:
Figured out why all response compression tests are failing. We are now double flushing the stream for ResponseCompression; once from StreamPipeWriter.FlushAsync, and another from BodyWrapperStream.FinishCompressionAsync (which calls Dispose), which causes extra compression frames to be written to the response. 

To fix this, either we need a way to either make the StreamPipeWriter not call FlushAsync on the underlying stream and only write the buffered contents to the stream (which may straight up break scenarios), or we need to fix the tests to expect a longer length (which isn't desired as it means response compression is less optimized).

src/Http/Http/src/StreamPipeWriter.cs
```diff
@@ -265,7 +265,7 @@ private void AddSegment(int sizeHint = 0)
             }
 
             // Get a new buffer using the minimum segment size, unless the size hint is larger than a single segment.
-            _currentSegmentOwner = _pool.Rent(Math.Max(_minimumSegmentSize, sizeHint));
+            _currentSegmentOwner = _pool.Rent(Math.Min(Math.Max(_minimumSegmentSize, sizeHint), _pool.MaxBufferSize));
```
halter73: Are there regression tests for these StreamPipeWriter fixes?

jkotalik: I'm making separate PRs for the StreamPipeWriter. I'll make a PR shortly for that fix 😄

src/Http/Http/src/WriteOnlyPipeStream.cs
```diff
@@ -60,6 +60,8 @@ public override int ReadTimeout
             set => throw new NotSupportedException();
         }
 
+        public PipeWriter InnerPipeWriter => _pipeWriter;
```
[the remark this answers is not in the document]
jkotalik: Yes unused, but I think from an API standpoint it makes sense to have as we have the same for the StreamPipeWriter. I can add the same to the ReadOnlyPipeStream too.

src/Http/Http/src/StreamPipeWriter.cs
```diff
@@ -93,7 +93,7 @@ public override Memory<byte> GetMemory(int sizeHint = 0)
         {
             EnsureCapacity(sizeHint);
 
-            return _currentSegment;
+            return _currentSegment.Slice(_position);
```
jkotalik: https://github.com/aspnet/AspNetCore/pull/7113/files

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
```diff
@@ -1338,5 +1260,148 @@ protected void ReportApplicationError(Exception ex)
...
+            if (_canHaveBody)
+            {
+                if (_autoChunk)
+                {
+                    if (data.Length == 0)
+                    {
+                        return await FlushAsyncInternal(cancellationToken);
+                    }
+
+                    return await WriteChunkedAsync(data.Span, cancellationToken);
```
halter73: Since you're now passing an autoChunk param to IHttpOutputProducer, can we do all of our chunking logic there instead of having it split between Http1OutputProducer and HttpProtocol.

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
```diff
@@ -153,20 +229,33 @@ public void Abort(ConnectionAbortedException error)
...
-        public Task Write100ContinueAsync()
+        public ValueTask<FlushResult> Write100ContinueAsync()
         {
             return WriteAsync(_continueBytes.Span);
         }
 
-        private Task WriteAsync(
+        public void Complete(Exception exception = null)
+        {
+            // TODO What to do with exception.
```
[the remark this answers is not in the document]
halter73: I would treat it as closely as possible to an application throwing an uncaught exception as possible. If the response hasn't started, that means a 500 response. Otherwise, log and and abort the underlying connection.

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
```diff
@@ -188,5 +277,56 @@ public Task Write100ContinueAsync()
...
+            {
+                // First time calling GetMemory
+                _currentChunkMemory = _pipeWriter.GetMemory(sizeHint);
+            }
+
+            var memoryMaxLength = _currentChunkMemory.Length - BeginChunkLengthMax - EndChunkLength;
+            if (_advancedBytesForChunk == memoryMaxLength)
+            {
+                // Chunk is completely written, commit it to the pipe so GetMemory will return a new chunk of memory.
+                CommitChunkToPipe();
```
halter73: Does this mean we can implicitly commit what should be uncommitted bytes in PipeWriter.GetMemory/Span()?

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
```diff
@@ -188,5 +277,56 @@ public Task Write100ContinueAsync()
...
+
+            if (_advancedBytesForChunk > 0)
+            {
+                var bytesWritten = writer.WriteBeginChunkBytes(_advancedBytesForChunk);
+                if (bytesWritten < BeginChunkLengthMax)
+                {
+                    // If the current chunk of memory isn't completely utilized, we need to copy the contents forwards.
+                    // This occurs if someone uses less than 255 bytes of the current Memory segment.
+                    // Therefore, we need to copy it forwards by either 1 or 2 bytes (depending on number of bytes)
+                    _currentChunkMemory.Slice(BeginChunkLengthMax, _advancedBytesForChunk).CopyTo(_currentChunkMemory.Slice(bytesWritten));
```
halter73: Shouldn't this fill the 1 or 2 bytes to the left to '0's? I think we should modify WriteBeginChunkBytes to always write 5 bytes right-justify the chunked prefix anyway.

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.FeatureCollection.cs
```diff
@@ -193,6 +190,64 @@ bool IHttpBodyControlFeature.AllowSynchronousIO
             }
         }
 
+        PipeWriter IResponseBodyPipeFeature.ResponseBodyPipe
+        {
+            get
+            {
+                if (_userSetPipeWriter != null)
```
halter73: You could remove this and just move the condition to the next if.

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.FeatureCollection.cs
```diff
@@ -193,6 +190,64 @@ bool IHttpBodyControlFeature.AllowSynchronousIO
...
+                }
+
+                if (!object.ReferenceEquals(_cachedResponseBodyStream, ResponseBody))
+                {
+                    var responsePipeWriter = new StreamPipeWriter(ResponseBody);
+                    ResponsePipeWriter = responsePipeWriter;
+                    _cachedResponseBodyStream = ResponseBody;
+                    OnCompleted((rpw) =>
+                    {
+                        ((StreamPipeWriter)rpw).Dispose();
```
halter73: Why not just do this in HttpProtocol.Reset()?

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
jkotalik: Do clients handle that preceding '0's correctly? That's my big concern with doing that.

Tratcher: I don't recommend leading 0's, it would be new behavior and we don't know if clients would handle it correctly.

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
jkotalik: Agreed. To clarify, you want all the chunking logic in Http1OutputProducer or HttpProtocol?

src/Http/Http/src/StreamPipeWriter.cs
jkotalik: https://github.com/aspnet/AspNetCore/pull/7143

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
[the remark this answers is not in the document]
jkotalik: So some combination of https://github.com/aspnet/AspNetCore/blob/20ee67b0cec738e99ccf4007876a729083804982/src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs#L585-L586 and making sure this logic isn't called twice?

src/Http/Http.Abstractions/src/Extensions/HttpResponseWritingExtensions.cs
jkotalik: If I use GetMemory, I would need to `await StartAsync`.

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
```diff
@@ -188,5 +277,56 @@ public Task Write100ContinueAsync()
...
+            // 4089.
+
+            if (_currentChunkMemory.Length == 0)
+            {
+                // First time calling GetMemory
+                _currentChunkMemory = _pipeWriter.GetMemory(sizeHint);
+            }
+
+            var memoryMaxLength = _currentChunkMemory.Length - BeginChunkLengthMax - EndChunkLength;
+            if (_advancedBytesForChunk == memoryMaxLength)
```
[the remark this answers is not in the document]
jkotalik: I'll make it const 1024 for now. We can consider making it modifiable, but I doubt we would change it.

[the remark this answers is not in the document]
Tratcher: How about Min(1024, sizeHint)?

src/Servers/Kestrel/Core/src/Internal/Http/ChunkWriter.cs
```diff
@@ -44,13 +44,14 @@ public static int BeginChunkBytes(int dataCount, Span<byte> span)
             return count;
         }
 
-        internal static void WriteBeginChunkBytes(this ref BufferWriter<PipeWriter> start, int dataCount)
+        internal static int WriteBeginChunkBytes(this ref BufferWriter<PipeWriter> start, int dataCount)
         {
             // 10 bytes is max length + \r\n
             start.Ensure(10);
```
halter73: If the max length of the chunk prefix is actually 5 bytes, shouldn't we change this to Ensure(5)?

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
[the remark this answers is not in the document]
halter73: SetBadRequestState is used to set 4XX responses for invalid requests. I think callin

```C#
ReportApplicationError(ex);

if (HasResponseStarted)
{
    ApplicationAbort();
}
```

I would probably move ApplicationAbort() from HttpProtocol.FeatureCollection.cs to HttpProtocol.cs as part of this.

src/Servers/Kestrel/Core/src/Internal/Http/ChunkWriter.cs
jkotalik: I'd need to update the ChunkWriter tests, but as of now I believe you are correct.

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
jkotalik: I think it's probably best to use a bool for tracking what the current segment is rather than setting it to default accordingly.

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
halter73: @pakrym

That's not how Pipe's normally work, right? I just double checked the source, and Advance only advances the write head not the read head. You have to flush or complete the PipeWriter to commit data.

Code that writes to the buffer in multiple segments and then goes back to change something in a previous segment (just like we were considering doing with the transport pipe to efficiently chunk responses) would break.

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.FeatureCollection.cs
```diff
@@ -193,6 +190,64 @@ bool IHttpBodyControlFeature.AllowSynchronousIO
             }
         }
 
+        PipeWriter IResponseBodyPipeFeature.ResponseBodyPipe
+        {
+            get
+            {
+                if (_userSetPipeWriter != null)
+                {
+                    return _userSetPipeWriter;
```
halter73: Where is this getting reset back to null? You should add a regression test similar to RequestTests.StreamsAreNotPersistedAcrossRequests but for Pipes.

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
pakrym: > Code that writes to the buffer in multiple segments and then goes back to change something in a previous segment (just like we were considering doing with the transport pipe to efficiently chunk responses) would break.

Didn't we decide that this is an unsupported scenario?

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.FeatureCollection.cs
halter73: I would rather we didn't use a public OnCompleted method for this because intermingling StreamPipeWriter.Dispose() with user code seems like a recipe for problems down the road. If you were to dispose the StreamPipeWriter here, you should at least set ResponseBodyPipe back to the previous value.

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.FeatureCollection.cs
```diff
@@ -193,6 +190,64 @@ bool IHttpBodyControlFeature.AllowSynchronousIO
...
+        }
+
+
+        Stream IHttpResponseFeature.Body
+        {
+            get
+            {
+                if (_userSetResponseBody != null)
+                {
+                    return _userSetResponseBody;
```
halter73: I do find it kinda weird how any attempt to set the property, even back to it's original value, breaks the automatic stream-to-pipe-to-stream adaptation. For example.

```C#
var originalBody = context.Response.Body;
var ms = new MemoryStream()
try
{
    context.Response.Body = new MemoryStream();
    await next();
    ms.Position = 0;
    await ms.CopyToAsync(originalBody);
}
finally
{
    context.Response.Body = originalBody;
}

// Later...

context.Response.BodyPipe = new FooPipeWriter(context.Response.BodyPipe);

// Even later... The following uses the last user-set response stream,
// "originalBody", not an adapted FooPipeWriter :(

await context.Response.Body.WriteAsync(data, 0, data.Length); 
```=

File: src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs (hunk 8 above)
Comment (halter73) [reply src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs]: That behavior could be pretty useful. If at all possible, I'd like to support it.

Comment (davidfowl) [reply src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs]: Yes! Come to the dark side. We need to support this 😄

File: src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs [hunk 16]
```diff
@@ -322,7 +327,7 @@ private void CommitChunkToPipe()
                     _currentChunkMemory.Slice(BeginChunkLengthMax, _advancedBytesForChunk).CopyTo(_currentChunkMemory.Slice(bytesWritten));
                 }
 
-                writer.Write(_currentChunkMemory.Slice(bytesWritten, _advancedBytesForChunk).Span);
+                writer.Advance(_advancedBytesForChunk);
```
halter73: 😌

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.FeatureCollection.cs
davidfowl: +1

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
pakrym: I don't understand, we were in the same meeting where we decided that's not a good idea without giving a consumer of IBufferWriter a way to notify other components that memory is preserved.


Even with pipe implementation holding onto Memory is safe only until flush is called which is often not controlled by IBufferWriter consumer. Like in the following sample:

``` C#
var writer = new SomeLenghtPrefixSerializer(response.BodyPipe);
foreach (var o in objects)
{
   writer.WriteObject(o);
   // Is this safe to do or is writer still holding onto memory?
   await response.BodyPipe.FlushAsync();
}
```

Getting this kind of assumption wrong can lead to severe consequences - considering that we reuse blocks response intended for one client can be sent to another one.

Another reason is that IBufferWriter implementation can be transforming bytes synchronously (encryption/compression) after they were advanced.

src/Http/Http.Abstractions/src/Extensions/HttpResponseWritingExtensions.cs
davidfowl: You would need to call StartAsync, see if it is completed, if it isn't then await it.

src/Http/Http.Abstractions/src/Extensions/HttpResponseWritingExtensions.cs
```diff
@@ -60,8 +61,42 @@ public static Task WriteAsync(this HttpResponse response, string text, Encoding
...
+            }
+
+            return FlushTaskAwaited(flushTask);
+        }
+
+        private static void WriteDataToPipe(PipeWriter responsePipeWriter, string text, Encoding encoding)
+        {
             byte[] data = encoding.GetBytes(text);
-            return response.Body.WriteAsync(data, 0, data.Length, cancellationToken);
+            var memory = responsePipeWriter.GetMemory(data.Length);
```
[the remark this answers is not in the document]
davidfowl: Yes this isn’t right, we need to encode directly into the buffer

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.FeatureCollection.cs
jkotalik: I don't see how/why that would work. This is for when the user sets the ResponseBodyPipe.

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.FeatureCollection.cs
jkotalik: Few things:
- Because we are creating the StreamPipeWriters/WriteOnlyPipeStream, we need to handle disposing them. That means if multiple wrappers are created, we need to keep a list of them to dispose in Reset.
- I wanted to keep the solution the same between the default ResponsePipeFeature and Kestrel's implementation, but as we know exactly when the object will be dispose in Kestrel, we should take advantage of it. 
- If we Dispose on Reset, I don't think we need to bother with setting the ResponseBodyPipe to the previous value.

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.FeatureCollection.cs
jkotalik: Bleh... yeah you are right. That isn't great behavior (though very nested). The problem I'm trying to avoid is when the user sets both the response body and response pipe in the same middleware, for example:

```C#
context.Response.Body = new MemoryStream();
context.Response.BodyPipe = new FooPipeWriter();
// Body is now a StreamPipeWriter
```

I don't see a great way to solve both of these problems.

src/Http/Http.Abstractions/src/Extensions/HttpResponseWritingExtensions.cs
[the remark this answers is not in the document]
jkotalik: Didn't know that was possible. Glad it is.

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
```diff
@@ -58,28 +70,91 @@ public Task WriteDataAsync(ReadOnlySpan<byte> buffer, CancellationToken cancella
...
+            }
+        }
+
+        public void CancelPendingFlush()
+        {
+            _pipeWriter.CancelPendingFlush();
+        }
+
+        // This method is for chunked http responses
+        public ValueTask<FlushResult> WriteChunkAsync(ReadOnlySpan<byte> buffer, CancellationToken cancellationToken)
```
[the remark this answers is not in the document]
jkotalik: Me being scared on my first pass to break things 😄

src/Http/Http.Abstractions/src/Extensions/HttpResponseWritingExtensions.cs
jkotalik: @davidfowl for future reference, why did this create a temporary buffer before?

src/Http/Http.Abstractions/src/Extensions/HttpResponseWritingExtensions.cs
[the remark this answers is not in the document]
jkotalik: Spoke offline, we are going to optimize this in a separate PR and for now call WriteAsync with an allocation. If we are encoding directly to the buffer, we need to do something similar to IBufferWriter.Write(). It shouldn't be hard. https://github.com/dotnet/corefx/blob/a10890f4ffe0fadf090c922578ba0e606ebdd16c/src/System.Memory/src/System/Buffers/BuffersExtensions.cs#L111

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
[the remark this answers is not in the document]
jkotalik: Looking back now it makes sense why I did it. One path is for when someone calls WriteAsync directly, which translates to WriteChunkAsync. The other is for GetMemory/Advance, which needs to do buffer moving and track advanced bytes.

src/Servers/Kestrel/Core/src/Internal/Http/ChunkWriter.cs
jkotalik: Actually I take that back, the max length can still be 10. If someone calls response.Body.WriteAsync(new byte[1000000]); this will still be written as one chunk. The chunk prefix restriction only applies to GetMemory/Advance.

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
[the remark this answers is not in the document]
jkotalik: Alright, I'll try that and see how it looks. @davidfowl I'll test this in a console app to see how visible the exception is too.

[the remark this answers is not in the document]
jkotalik: @halter73 If someone called Complete() without an exception, we probably shouldn't treat this as the application error.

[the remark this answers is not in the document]
halter73: Absolutely agreed. I was only talking about the exceptional case.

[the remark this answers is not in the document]
Tratcher: Do not call the OnCompleted callbacks early, only ProduceEnd.

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.FeatureCollection.cs
halter73: I'm not sure there is a great way to solve both problems. Preferring the user-set stream/pipe is important for the reason you highlighted, but it causes the kind of issue I highlighted.

That's why I'd prefer no magical getter/setter. It's way too hard to reason about.

src/Http/Http.Abstractions/src/Extensions/HttpResponseWritingExtensions.cs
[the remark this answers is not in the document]
Tratcher: Just revert HttpResponseWritingExtensions if we're going to do it right later.

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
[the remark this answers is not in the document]
jkotalik: I don't think we can call ProduceEnd as that writes to the response. Complete is sync and we can't flush/write t all. Also, I'd rather just let it be handled by the app func returning and handling it in ProcessRequests.

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.FeatureCollection.cs
jkotalik: Let's chat about this tomorrow. I think I prefer your suggestion more than mine, but let's make sure we don't miss anything.

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.FeatureCollection.cs
jkotalik: I have a prototype of having a list of disposable objects here. Tell me if you like the idea, and I'll clean it up if it works.

src/Http/Http.Abstractions/src/Extensions/HttpResponseWritingExtensions.cs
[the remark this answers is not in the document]
jkotalik: Gah now I'm stuck in an awkward middleground here. Guess I'll have to revert all of the test fixes to response compression 😢

jkotalik:
🆙 📅 . There is still some crust around the new Disposal logic I added, but I wanted to make sure it's a good idea before cleaning up.

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
Tratcher: `_currentChunkMemoryUpdated = false;`?

src/Servers/Kestrel/test/InMemory.FunctionalTests/Http2/Http2StreamTests.cs
```diff
@@ -2468,5 +2469,718 @@ public async Task WriteAsync_CancellationTokenTriggeredDueToFlowControl_SendRST(
...
+                new KeyValuePair<string, string>(HeaderNames.Method, "GET"),
+                new KeyValuePair<string, string>(HeaderNames.Path, "/"),
+                new KeyValuePair<string, string>(HeaderNames.Scheme, "http"),
+            };
+            await InitializeConnectionAsync(async httpContext =>
+            {
+                var response = httpContext.Response;
+                await response.StartAsync();
+
+                var memory = response.BodyPipe.GetMemory(5000); // This will return 4096
```
Tratcher: Assert that?

src/Http/Http.Abstractions/src/Extensions/HttpResponseWritingExtensions.cs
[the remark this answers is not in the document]
jkotalik: Went ahead and made it good: https://github.com/aspnet/AspNetCore/pull/7110/files#diff-6f529bca1b5cb1bf98dbd8b32d27969dR64

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.FeatureCollection.cs
jkotalik: I'd also say this is a non-blocker for getting this PR in.

jkotalik:
I think this is in a pretty good spot. Few topics that we still need to discuss:
- How to handle user setting BodyPipe and automatic creation
- Http2OutputProducer.Complete() - Should we call Complete on the DataPipe?
- Performance - I'm seeing a performance decrease with the InMemoryTransportBenchmark (~5%) when calling response.Body.WriteAsync(). I don't see any performance decrease with the Benchmarks UI though. 

Besides that, feel free to really start nit picking at things. I think my test coverage is good, but if there are new/interesting scenarios that I missed, please inform me.

src/Servers/Kestrel/Core/src/Internal/Http2/Http2OutputProducer.cs
```diff
@@ -299,5 +299,11 @@ ValueTask<FlushResult> IHttpOutputProducer.WriteChunkAsync(ReadOnlySpan<byte> da
         {
             throw new NotImplementedException();
         }
+
+        public void Complete()
+        {
+            // This can't complete the pipe today because headers NEED to be added as part of the protocol
+            //_dataPipe.Writer.Complete();
```
davidfowl: Why is this commented out? Is there a bug somewhere?

halter73: I also don't get why the _dataPipe.Writer cannot be completed here. Completing it doesn't prevent normal response headers or trailers from being written AFAIK.

jkotalik: Wouldn't WriteResponseHeaders need to be called before Complete? Otherwise we won't get normal response headers. 

Maybe we can throw if Complete is called before StartAsync too?

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
```diff
@@ -58,28 +76,116 @@ public Task WriteDataAsync(ReadOnlySpan<byte> buffer, CancellationToken cancella
...
-        public Task FlushAsync(CancellationToken cancellationToken = default)
+        public ValueTask<FlushResult> FlushAsync(CancellationToken cancellationToken = default)
         {
             return WriteAsync(Constants.EmptyData, cancellationToken);
         }
 
-        public Task WriteChunkAsync(ReadOnlySpan<byte> buffer, CancellationToken cancellationToken)
+        public Memory<byte> GetMemory(int sizeHint = 0)
+        {
+            if (_completed)
```
Tratcher: Why can you still call GetMemory after Complete? You can't Advance. Let's throw and see what breaks.

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
```diff
@@ -58,28 +76,116 @@ public Task WriteDataAsync(ReadOnlySpan<byte> buffer, CancellationToken cancella
...
+            {
+                return _pipeWriter.GetMemory(sizeHint).Span;
+            }
+        }
+
+        public void Advance(int bytes)
+        {
+            if (_completed)
+            {
+                return;
```
Tratcher: Throw?

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
jkotalik: Okay! 😄

jkotalik: Few issues:
- Abort calls PipeWriter.Complete(), which makes the pipe "completed". Should this cause WriteAsync to throw still? Can Abort be called due to the client disconnecting?

Tratcher: More likely Abort shouldn't call Complete, it should call CancelPendingRead.

JamesNK:
`Grpc.AspNetCore.Server` is using BodyPipe exclusively, and it has fairly robust functional tests of various HTTP/2 scenarios: unary requests/client streaming/server streaming/duplex streaming.

Consider testing the new PipeWriter (and eventually PipeReader) implementation with Grpc.AspNetCore.Server and see whether the unit tests all pass.

https://github.com/grpc/grpc-dotnet

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.FeatureCollection.cs
jkotalik: Based on offline feedback, this will always wrap whatever was set last.

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
jkotalik: Going to let this Noop still because PW.Complete can still be called by Abort. May revisit in a separate PR to distinguish calling Abort vs BodyPipe.Complete().

halter73: Can you file an issue to track splitting Http1OutputProducer.Disable/Dispose()?

jkotalik: Sure. I did some of it already here https://github.com/aspnet/AspNetCore/pull/7110/files#diff-cf42d74c2bbea44bc90926ce3dcee751R243

jkotalik: What else do you think I need to do?

src/Http/Http.Abstractions/src/Extensions/HttpResponseWritingExtensions.cs
```diff
@@ -60,8 +61,73 @@ public static Task WriteAsync(this HttpResponse response, string text, Encoding
...
+                if (!startAsyncTask.IsCompletedSuccessfully)
+                {
+                    return StartAndWriteAsyncAwaited(response, text, encoding, cancellationToken, startAsyncTask);
+                }
+            }
+
+            Write(response, text, encoding);
+
+            var flushAsyncTask = response.BodyPipe.FlushAsync(cancellationToken);
+            if (flushAsyncTask.IsCompleted)
```
davidfowl: This should be IsCompletedSuccessfully

jkotalik:
Verified with GRPC and things look good 😄

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
```diff
@@ -117,19 +222,30 @@ public void WriteResponseHeaders(int statusCode, string reasonPhrase, HttpRespon
                 writer.Commit();
 
                 _unflushedBytes += writer.BytesCommitted;
+                _autoChunk = autoChunk;
             }
         }
 
         public void Dispose()
```
halter73: Make sure this is still being called.

src/Servers/Kestrel/Core/src/Internal/Http2/Http2FrameWriter.cs
```diff
@@ -312,12 +312,14 @@ private void WriteDataUnsynchronized(int streamId, ReadOnlySequence<byte> data,
...
-        private async Task WriteDataAsync(int streamId, StreamOutputFlowControl flowControl, ReadOnlySequence<byte> data, long dataLength, bool endStream)
+        private async ValueTask<FlushResult> WriteDataAsync(int streamId, StreamOutputFlowControl flowControl, ReadOnlySequence<byte> data, long dataLength, bool endStream)
         {
+            FlushResult flushResult = default;
+
             while (dataLength > 0)
             {
                 OutputFlowControlAwaitable availabilityAwaitable;
-                var writeTask = Task.CompletedTask;
+                var writeTask = new ValueTask<FlushResult>(new FlushResult());
```
halter73: nit: `default(ValueTask<FlushResult>)`

halter73 (approved):

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.FeatureCollection.cs
```diff
@@ -193,6 +191,54 @@ bool IHttpBodyControlFeature.AllowSynchronousIO
             }
         }
 
+        PipeWriter IResponseBodyPipeFeature.ResponseBodyPipe
+        {
+            get
+            {
+                if (!object.ReferenceEquals(_cachedResponseBodyStream, ResponseBody))
```
halter73: Just assign greedily like @davidfowl suggested. With this logic, if you set the body Stream first without then the body Pipe second without any calls to the getters in-between and then afterwards get the body Pipe, you'll get a Pipe wrapping the body Stream that was assigned first instead of the  body Pipe set after the body Stream was set.

We should have a test that assigns to both body properties and verifies the last assignment wins.

jkotalik: Agreed. Done and added tests.

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
jkotalik: It is called here https://github.com/aspnet/AspNetCore/blob/7d52a0c9e3da57839e227f814ea73d31351578f8/src/Servers/Kestrel/Core/src/Internal/Http/Http1Connection.cs#L79

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.FeatureCollection.cs
```diff
@@ -195,48 +195,48 @@ PipeWriter IResponseBodyPipeFeature.ResponseBodyPipe
...
 
-        Stream IHttpResponseFeature.Body
-        {
-            get
-            {
                 if (!object.ReferenceEquals(_cachedResponsePipeWriter, ResponsePipeWriter))
                 {
                     var responseBody = new WriteOnlyPipeStream(ResponsePipeWriter);
                     ResponseBody = responseBody;
-                    _cachedResponsePipeWriter = ResponsePipeWriter;
```
Tratcher: _cachedResponsePipeWriter is never set

jkotalik:
Apparently I merged something that didn't pass validation... Let me go fix that.

src/Servers/Kestrel/test/FunctionalTests/ResponseTests.cs
```diff
@@ -297,7 +298,7 @@ public async Task WritingToConnectionAfterUnobservedCloseTriggersRequestAbortedT
                 {
                     for (var i = 0; i < 1000; i++)
                     {
-                        await context.Response.Body.WriteAsync(scratchBuffer, 0, scratchBuffer.Length, context.RequestAborted);
```
Tratcher: So we lost a bunch of coverage for Response.Body?

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
```diff
@@ -64,6 +63,12 @@ public abstract partial class HttpProtocol : IDefaultHttpContextContainer, IHttp
         protected string _methodText = null;
         private string _scheme = null;
 
+        protected PipeWriter _userSetPipeWriter;
+        protected Stream _userSetResponseBody;
```
halter73: The _userSet... fields are no longer used.

Tratcher (approved):

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.FeatureCollection.cs
```diff
@@ -193,6 +192,50 @@ bool IHttpBodyControlFeature.AllowSynchronousIO
...
+        {
+            get
+            {
+                return ResponsePipeWriter;
+            }
+            set
+            {
+                ResponsePipeWriter = value;
+
+                if (!object.ReferenceEquals(_cachedResponsePipeWriter, ResponsePipeWriter))
```
halter73: I wouldn't even bother doing this ReferenceEquals check. Just create the WriteOnlyPipeStream no matter what.

If you really want to be more efficient in the unlikely case someone sets ResponsePipeWriter to its current value, you could put the following at the top of the setter for the same effect:

```c#
if (object.ReferenceEquals(value, ResponsePipeWriter) return;
```

But I don't think it's worth even wasting one line of code optimizing this scenario.

src/Servers/Kestrel/test/FunctionalTests/ResponseTests.cs
halter73: Yeah. This does lose some coverage for possible regressions in Response.Body. Fortunately, the stream implementation itself is very stable, and most of the logic being tested is in HttpProtocol or one of the OutputProduces which is used by both the Pipe and the Stream.

We could do something similar to what I did for HTTPS and run a bunch of tests twice: once with the body Stream and once with the BodyPipe, but I don't think it's worth it to do that en masse right now. If in the future, we do find ways in with the Stream and Pipe behavior diverge, we should add regression tests for both to ensure they both behave as expected.

jkotalik: Yes we did. I'll quickly add some regression tests.

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.FeatureCollection.cs
jkotalik: Yeah, I kind of realized this optimization is so irrelevant that we should just ignore it.

benaadams:
1 error on `AspNetCore-helix-test` but only on `Centos.7.Amd64.Open`

```
[xUnit.net 00:00:02.41]     Microsoft.AspNetCore.Components.Test.RendererTest.CanTriggerEventHandlerDisposedInEarlierPendingBatchAsync [FAIL]
Failed   Microsoft.AspNetCore.Components.Test.RendererTest.CanTriggerEventHandlerDisposedInEarlierPendingBatchAsync
Error Message:
 Assert.Throws() Failure
Expected: typeof(System.ArgumentException)
Actual:   (No exception was thrown)
Stack Trace:
   at Microsoft.AspNetCore.Components.Test.RendererTest.CanTriggerEventHandlerDisposedInEarlierPendingBatchAsync() 
in /_/src/Components/Components/test/RendererTest.cs:line 1364
```

jkotalik:
Fortunately Helix tests aren't required yet 😄

PR #7113, opened 2019-01-29
Call Slice in GetMemory

jkotalik:
For #7031

### DIFF
--- a/src/Http/Http/test/StreamPipeWriterTests.cs
+++ b/src/Http/Http/test/StreamPipeWriterTests.cs
@@ -351,27 +351,41 @@ namespace System.IO.Pipelines.Tests
         }
 
         [Fact]
-        public void GetMemorySlicesCorrectly()
+        public async Task GetMemorySlicesCorrectly()
         {
             var expectedString = "abcdef";
             var memory = Writer.GetMemory();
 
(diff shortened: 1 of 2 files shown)


Tratcher (approved):

src/Http/Http/test/StreamPipeWriterTests.cs
```diff
@@ -316,6 +316,30 @@ public async Task UseStreamThenPipeToWriteMultipleTimes()
             Assert.Equal(expectedString, ReadAsString());
         }
 
+        [Fact]
+        public void GetMemorySlicesCorrectly()
+        {
+            var expectedString = "abcdef";
+            var memory = Writer.GetMemory();
+
+            Encoding.ASCII.GetBytes(expectedString).CopyTo(memory);
```
pakrym: I'm not a fan of these unit tests, in my opinion they are testing the wrong thing. 

Why not check that correct data is written to the stream when calling GetMemory() - Advance() multiple times?

jkotalik:
@pakrym updated.

pakrym:
LGTM

davidfowl (approved):

JamesNK:
:shipit:

PR #7114, opened 2019-01-29
Change site extension build pool to dotnet-external-temp

pakrym:
Might fix signing

### DIFF
--- a/.azure/pipelines/site-extensions.yml
+++ b/.azure/pipelines/site-extensions.yml
@@ -5,6 +5,8 @@ trigger:
 
 phases:
 - phase: SiteExtensions
+  pool:
+    name: dotnet-internal-temp
   variables:
     CI: true
     DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true


.azure/pipelines/site-extensions.yml
```diff
@@ -5,6 +5,8 @@ trigger:
 
 phases:
 - phase: SiteExtensions
+  pool:
+    name: dotnet-external-temp
```
natemcmaster: This needs to vary based on which project we build in. See https://github.com/dotnet/arcade/blob/master/Documentation/ChoosingAMachinePool.md and

https://github.com/aspnet/AspNetCore/blob/ef1fc9e5c74b93ae45b8f4b36a4af0ea8bce9b43/.azure/pipelines/jobs/default-build.yml#L89-L92

natemcmaster (changes requested):

.azure/pipelines/site-extensions.yml
pakrym: This is never built in public.

pakrym: Fixed the name,

PR #7115, opened 2019-01-30
Make StreamPipeWriter.Complete throw if there is unflushed data

jkotalik:
For https://github.com/aspnet/AspNetCore/issues/7102

### DIFF
--- a/src/Http/Http/src/StreamPipeWriter.cs
+++ b/src/Http/Http/src/StreamPipeWriter.cs
@@ -119,6 +119,7 @@ namespace System.IO.Pipelines
             }
 
             _isCompleted = true;
+
             if (exception != null)
             {
                 _exceptionInfo = ExceptionDispatchInfo.Capture(exception);
@@ -135,6 +136,12 @@ namespace System.IO.Pipelines
             }
 
             _currentSegmentOwner?.Dispose();
+
+            // We still want to cleanup segments before throwing an exception.
+            if (_bytesWritten > 0)
+            {
+                ThrowHelper.ThrowInvalidOperationException_DataNotAllFlushed();
+            }
         }
 
         /// <inheritdoc />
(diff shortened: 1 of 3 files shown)


src/Http/Http/src/StreamPipeWriter.cs
```diff
@@ -135,6 +136,12 @@ public override void Complete(Exception exception = null)
             }
 
             _currentSegmentOwner?.Dispose();
+
+            // We still want to cleanup segments before throwing an exception.
```
davidfowl: Why? Why doesn't it just trigger the reader? That's what happens in the Pipe impl

Tratcher: There is no reader, it would be a sync Write on the stream.

davidfowl: Ah yes. Maybe this should just be IAsyncDuspoable

Tratcher: The disposal timing would be problematic. Right now it's not disposed until OnCompleted. Because this is created dynamically nothing in pipeline knows to flush or dispose it before the response exits.

People must call FlushAsync after writing data or else that data will never be sent. At least this won't be flaky, it will always behave the same way and people should discover that during initial development.

src/Http/Http/src/StreamPipeWriter.cs
```diff
@@ -135,6 +136,12 @@ public override void Complete(Exception exception = null)
...
+            if (_bytesWritten > 0)
```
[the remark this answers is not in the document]
jkotalik: After talking with @pakrym , the exception passed in to Complete should only be used by the reader. So the exception passed in should be ignored.

jkotalik:
Updated. Few changes
- Made GetMemory/GetSpan throw if they are completed or pass in an invalid size
- Don't throw the exception completed by the writer (but still check if the exception is there when deciding to throw in Complete).

Tratcher (approved):

PR #7128, opened 2019-01-30
[MVC] Normalize dependencies under the _dependencies folder in the solution

javiercn:
There were two folders on the solution _dependencies and dependencies. With the dependencies split across both folders. I've moved all to _dependencies and got rid of dependencies (_dependencies always appears as the first element in the solution and never between the main projects)


ajaybhargavb (commented):
Sorry I'm not sure I understand what's going on here. Why the change? Where is the _dependencies folder? cc @pranavkm

ajaybhargavb (approved):

PR #7136, opened 2019-01-30
[Components][Fixes #6953] Prevent multiple threads from entering ProcessRenderQueue

### LINKED ISSUE #6953: [Test failure] RendererTest.CanRenderAsyncComponentsWithMultipleAsyncChildren
https://dev.azure.com/dnceng/public/_build/results?buildId=78506

```
 [xUnit.net 00:00:02.02]     Microsoft.AspNetCore.Components.Test.RendererTest.CanRenderAsyncComponentsWithMultipleAsyncChildren [FAIL]
2019-01-23T17:10:22.2646129Z   Failed   Microsoft.AspNetCore.Components.Test.RendererTest.CanRenderAsyncComponentsWithMultipleAsyncChildren
2019-01-23T17:10:22.2699250Z   Error Message:
2019-01-23T17:10:22.2754711Z    System.NullReferenceException : Object reference not set to an instance of an object.
2019-01-23T17:10:22.2977333Z   Stack Trace:
2019-01-23T17:10:22.3046592Z      at Microsoft.AspNetCore.Components.Rendering.Renderer.RenderInExistingBatch(RenderQueueEntry renderQueueEntry) in /_/src/Components/Components/src/Rendering/Renderer.cs:line 439
2019-01-23T17:10:22.3133950Z      at Microsoft.AspNetCore.Components.Rendering.Renderer.ProcessRenderQueue() in /_/src/Components/Components/src/Rendering/Renderer.cs:line 394
2019-01-23T17:10:22.3328157Z      at Microsoft.AspNetCore.Components.Rendering.Renderer.AddToRenderQueue(Int32 componentId, RenderFragment renderFragment) in /_/src/Components/Components/src/Rendering/Renderer.cs:line 370
2019-01-23T17:10:22.3481592Z      at Microsoft.AspNetCore.Components.RenderHandle.Render(RenderFragment renderFragment) in /_/src/Components/Components/src/RenderHandle.cs:line 43
2019-01-23T17:10:22.3744799Z      at Microsoft.AspNetCore.Components.ComponentBase.StateHasChanged() in /_/src/Components/Components/src/ComponentBase.cs:line 111
2019-01-23T17:10:22.3792890Z      at Microsoft.AspNetCore.Components.ComponentBase.RunInitAndSetParameters() in /_/src/Components/Components/src/ComponentBase.cs:line 231
2019-01-23T17:10:22.3920211Z      at Microsoft.AspNetCore.Components.Rendering.Renderer.ProcessAsynchronousWork() in /_/src/Components/Components/src/Rendering/Renderer.cs:line 198
2019-01-23T17:10:22.3957322Z      at Microsoft.AspNetCore.Components.Rendering.Renderer.RenderRootComponentAsync(Int32 componentId, ParameterCollection initialParameters) in /_/src/Components/Components/src/Rendering/Renderer.cs:line 185
2019-01-23T17:10:22.4243861Z      at Microsoft.AspNetCore.Components.Test.RendererTest.CanRenderAsyncComponentsWithMultipleAsyncChildren() in /_/src/Components/Components/test/RendererTest.cs:line 399
2019-01-23T17:10:22.4282433Z   --- End of stack trace from previous location where exception was thrown ---
```

/cc @SteveSandersonMS

javiercn:
We were seeing some errors in the CI on tests dealing with async rendering. I was able to repro them by running the tests on a tight loop. I believe the main cause was that two threads were entering ProcessRenderQueue simultaneously, one doing synchronous work and one or more threads coming in from async callbacks.

I added a lock to control the access to the method so that only one thread can be processing the queue at a time.

### DIFF
--- a/src/Components/Components/src/Rendering/Renderer.cs
+++ b/src/Components/Components/src/Rendering/Renderer.cs
@@ -365,10 +365,7 @@ namespace Microsoft.AspNetCore.Components.Rendering
                     new RenderQueueEntry(componentState, renderFragment));
             }
 
-            if (!_isBatchInProgress)
-            {
-                ProcessRenderQueue();
-            }
+            ProcessRenderQueue();
         }
 
         private ComponentState GetRequiredComponentState(int componentId)
@@ -383,7 +380,18 @@ namespace Microsoft.AspNetCore.Components.Rendering
 
         private void ProcessRenderQueue()
         {
-            _isBatchInProgress = true;
+            lock (_asyncWorkLock)
+            {
+                if (!_isBatchInProgress)
+                {
+                    _isBatchInProgress = true;
+                }
+                else
+                {
+                    return;
+                }
+            }
+
             var updateDisplayTask = Task.CompletedTask;
 
             try
@@ -402,7 +410,10 @@ namespace Microsoft.AspNetCore.Components.Rendering
             {
                 RemoveEventHandlerIds(_batchBuilder.DisposedEventHandlerIds.ToRange(), updateDisplayTask);
                 _batchBuilder.Clear();
-                _isBatchInProgress = false;
+                lock (_asyncWorkLock)
+                {
+                    _isBatchInProgress = false;
+                }
             }
         }
 


javiercn:
I ran the tests 100 times in a loop and they all passed

src/Components/Components/src/Rendering/Renderer.cs
```diff
@@ -383,7 +380,18 @@ private ComponentState GetOptionalComponentState(int componentId)
 
         private void ProcessRenderQueue()
         {
-            _isBatchInProgress = true;
+            lock (_asyncWorkLock)
```
pranavkm: Would something like Interlocked.Exchange work better here? See the example here: https://docs.microsoft.com/en-us/dotnet/api/system.threading.interlocked.exchange?view=netframework-4.7.2#System_Threading_Interlocked_Exchange_System_Int32__System_Int32_

javiercn: I'm not too concerned about this lock being contested, and we already use it on other places.

pranavkm: Right, but the sample uses a int value to do basically the same pattern that's being done here. Seems simpler to use the right tool here, no?

javiercn:
@SteveSandersonMS Yeah, I thought I described it on the description, but I'll try to explain it again:
A possible example is, imagine a parent and a child component:
I'll label the threads A,B,C and so on.
1. Parent renders on A.
2. Child renders on A.
3. Child starts some async work as part of SetParametersAsync
4. Parent renders again for some reason and changes Child parameters (all this in A)
5. SetParametersAsync gets called again on A which calls StateHasChanged and goes back to the renderer.
6. The callback from 3 comes back in thread B and calls StateHasChanged
Note that the body from StateHasChanged (below) doesn't protect correctly against this as its not a critical section and two threads can read _hasPendingQueuedRender and continue before setting it later on (hence both end up calling _renderHandle.Render(_renderFragment) triggering additional renders).
_renderHandle.Render(_renderFragment) ends up at some point in AddToRenderQueue where a similar situation happens. We check the value before we set it, and that needs to be done atomically as there are two threads in this code path.
```csharp
            if (_hasPendingQueuedRender)
            {
                return;
            }

            if (_hasNeverRendered || ShouldRender())
            {
                _hasPendingQueuedRender = true;

                try
                {
                    _renderHandle.Render(_renderFragment);
                }
                catch
                {
                    _hasPendingQueuedRender = false;
                    throw;
                }
            }
        }
```

The reason we do it this way is so that we don't have to do anything in ComponentBase (as the renderer needs to handle this for any IComponent implementation, people starting threads, etc)

We could also do something in StateHasChanged to prevent it from triggering multiple renders simultaneously, but that is in addition to this.

src/Components/Components/src/Rendering/Renderer.cs
javiercn: It feels like premature optimization to me, I would rather have this represented as a bool than as an int for the sake of using a "faster" locking primitive. (Which we also don't know if it works in WebAssembly).

pranavkm (approved):

SteveSandersonMS:
@javiercn The sync context is meant to ensure all continuations for a given `RemoteRenderer` occur in series, not simultaneously. I know people can *try* to issue renders from off the sync context, but we do verify that calls are on the sync context before queueing each render. Similarly, for `WebAssemblyRenderer` the scenario can't occur because there's only one thread under wasm anyway.

Is it perhaps that the test code isn't setting up a sync context, and hence doesn't represent the real usage scenario? Or am I mistaken in what flows are possible? Is the issue specific to `TestRenderer` or `HtmlRenderer` maybe?

My concern about adding a `lock` is having to reason about whether this can ever lead to a deadlock, with a child triggering a re-render for its parent, either with current rendering rules or whatever we change them to in the future. We hoped to avoid this possibility by using the sync context to make locking unnecessary.

javiercn:
@SteveSandersonMS 
* The test code is definitely not setting up the sync context. 
* The HtmlRenderer doesn't have a sync context.

I think we have two options:
* We can introduce the sync context as a fundamental primitive of the renderer (Renderer would have a sync context and would set it up appropriately on all of its async calls), enforcing the behavior that we expect.
* We introduce some tactical locking so that the render works fine even in the absence of a sync context (and the sync context becomes additional guarantees on some scenarios).

I think in the end what we want to achieve is the ability for things to work in a natural way with the user and any renderer implementer having to reason about the execution model.

If we don't lock we need to use a sync context on Renderer.

javiercn:
Closing this PR as we changed it for https://github.com/aspnet/AspNetCore/commit/f456e3d153039986e62d969b1c9d63aa301daf65

PR #7152, opened 2019-01-31
Reduce number of SauceLabs browsers to 3

mikaelm12:
Looked into retries but it doesn't seem like there is a nice way to do that. I can explore that more though. 
I was to see if this configuration stabilizes our daily tests. If so we can spit groups of browsers into different build tasks.


mikaelm12:
Want to test if the flakiness is SauceLabs related. recently we're only seeing issues on Safari. I want to test reducing the number of browsers, but keeping Safari to be clear. https://github.com/aspnet/AspNetCore-Internal/issues/1711

BrennanConroy (approved):
Ok, approved since we'll be adding these back

PR #7154, opened 2019-01-31
Increase Server Timeout for Cleaning up Connections

mikaelm12:
Issue: https://github.com/aspnet/AspNetCore/issues/6797
Upping the timeout from 5 seconds to 15. We also talked about doing 30 instead.

### DIFF
--- a/src/SignalR/common/Http.Connections/src/HttpConnectionDispatcherOptions.cs
+++ b/src/SignalR/common/Http.Connections/src/HttpConnectionDispatcherOptions.cs
@@ -29,7 +29,6 @@ namespace Microsoft.AspNetCore.Http.Connections
             TransportMaxBufferSize = DefaultPipeBufferSize;
             ApplicationMaxBufferSize = DefaultPipeBufferSize;
             DisconnectTimeout = DefaultDisconnectTimeout;
-
         }
 
         /// <summary>
--- a/src/SignalR/common/Http.Connections/src/HttpConnectionDispatcherOptions.cs
+++ b/src/SignalR/common/Http.Connections/src/HttpConnectionDispatcherOptions.cs
@@ -15,6 +15,8 @@ namespace Microsoft.AspNetCore.Http.Connections
         // There maybe the opportunity for performance gains by tuning this default.
         private const int DefaultPipeBufferSize = 32768;
 
+        private const int DefaultDisconnectTimeout = 15;
+
         /// <summary>
         /// Initializes a new instance of the <see cref="HttpConnectionDispatcherOptions"/> class.
         /// </summary>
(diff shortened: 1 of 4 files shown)


davidfowl:
We need to make it configurable first, then we can change the default setting to 15 seconds.

src/SignalR/common/Http.Connections/src/HttpConnectionDispatcherOptions.cs
```diff
@@ -57,5 +60,10 @@ public HttpConnectionDispatcherOptions()
         /// Gets or sets the maximum buffer size of the application writer.
         /// </summary>
         public long ApplicationMaxBufferSize { get; set; }
+
+        /// <summary>
+        /// Gets or sets the allowed duration in seconds that a client can not send messages before being removed by the server.
```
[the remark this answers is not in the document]
mikaelm12: Hm, Should this take a TimeSpan?

src/SignalR/common/Http.Connections/src/Internal/HttpConnectionDispatcher.cs
```diff
@@ -779,9 +779,9 @@ private async Task<HttpConnectionContext> GetOrCreateConnectionAsync(HttpContext
 
         private HttpConnectionContext CreateConnection(HttpConnectionDispatcherOptions options)
         {
+            _manager.DisconnectTimeout = options.DisconnectTimeout;
```
halter73: You cannot make the disconnect timeout a per-endpoint option and then store it in the singleton HttpConnectionManager. This will cause the HttpConnectionManager to enforce the most recent connection's disconnect timeout for *all* connections including those that might belong to a differently-configured endpoint.

halter73 (changes requested):

src/SignalR/common/Http.Connections/src/HttpConnectionDispatcherOptions.cs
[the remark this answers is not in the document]
halter73: Yes.

src/SignalR/common/Http.Connections/src/Internal/HttpConnectionDispatcher.cs
mikaelm12: Good catch. Hmm what would be the best way to do this then

mikaelm12: I'll make an HttpConnectionManagerOption type and configure it via an `addSignalR` overload.

mikaelm12:
So the new API for making this timeout configurable could be something like 
```csharp   
  public static ISignalRServerBuilder AddHttpConnectionManagerOptions(this ISignalRServerBuilder signalrBuilder, Action<HttpConnectionManagerOptions> configure)
```

This feels a bit weird since we don't really expose http/connection options at the SignalR layer. Wanted to get some thoughts on it

halter73:
I think I'd go with the following overload instead:

```C#
public static IServiceCollection AddConnections(this IServiceCollection services, Action<ConnectionOptions> configure)
```

mikaelm12:
Closing in favor of https://github.com/aspnet/AspNetCore/pull/7340

PR #7160, opened 2019-01-31
SignalR C++ tests passing and part of build

BrennanConroy:
Went through all failing tests and fixed them or removed them if they were no longer relevant.

Also added the client to the build. @halter73 and @mikaelm12 it might be a good idea to pull down the branch and run build.cmd from the signalr root to see if you have the required VS settings.


src/Servers/IIS/AspNetCoreModuleV2/CommonLibTests/CommonLibTests.vcxproj
```diff
@@ -193,5 +193,7 @@
       </AdditionalDependencies>
     </Lib>
   </ItemDefinitionGroup>
-  <Import Project=".\NativeTests.targets" />
+  <PropertyGroup>
+    <IsTestProject>true</IsTestProject>
```
BrennanConroy: cc @pakrym

src/Servers/IIS/AspNetCoreModuleV2/CommonLibTests/NativeTests.targets
```diff
@@ -1,8 +0,0 @@
-<Project>
-  <PropertyGroup>
-    <VsTestConsole>$(VCIDEInstallDir)..\CommonExtensions\Microsoft\TestWindow\vstest.console.exe</VsTestConsole>
```
natemcmaster: Does this need to move into Cpp.Common.targets?

eng/targets/Cpp.Common.targets
```diff
@@ -12,4 +12,7 @@
   <Target Name="Pack" />
   <Target Name="Restore" />
   <Target Name="ResolveNuGetPackageAssets" />
+  <Target Name="Test" DependsOnTargets="Build" Condition="'$(IsTestProject)' == 'true'" >
```
natemcmaster:
```suggestion
  <Target Name="Test" Condition="'$(IsTestProject)' == 'true'" >
```
I'd recommend removing this target dependency so `build.cmd -nobuild -test` will work (cref #7137)

src/Servers/IIS/AspNetCoreModuleV2/CommonLibTests/NativeTests.targets
BrennanConroy: It wasn't being used. At least we don't think it was.

src/Servers/IIS/AspNetCoreModuleV2/CommonLibTests/CommonLibTests.vcxproj
pakrym: 👍

natemcmaster (approved):

halter73:
I ran "./build.cmd" in the src\SignalR directory, and it only compiled without running tests. I had to explicitly pass a "-test" flag to run the tests. Should we run tests by default now?

BrennanConroy:
@natemcmaster It looks like the `<ImportGroup>` has run before the nuget packages are restored. Do you have any recommendations on how to fix that?
Line ref: https://github.com/aspnet/AspNetCore/pull/7160/files#diff-36a000fe432a66617010b714fa6a4d13L106

src/SignalR/clients/cpp/src/signalrclient/Build/VS/signalrclient.vcxproj
```diff
@@ -17,16 +17,16 @@
   <Import Project="..\..\..\..\Build\Config.Definitions.props" />
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.Default.props" />
   <PropertyGroup Label="Configuration" Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
-    <PlatformToolset>v141</PlatformToolset>
+    <PlatformToolset>v140</PlatformToolset>
```
jkotalik: Why downgrade from 141 to 140?

natemcmaster:
> It looks like the <ImportGroup> has run before the nuget packages are restored. Do you have any recommendations on how to fix that?

The problem is your projects are using a really old version of NuGet and packages.config. It's been a while since I used them, but think these targets work under the expectation that you will have executed `NuGet.exe` on the .sln before you build. You might try looking into using newer version of NuGet and `PackageReference` instead.

src/SignalR/clients/cpp/src/signalrclient/Build/VS/signalrclient.vcxproj
natemcmaster: Would be preferable to keep the toolset version consistent across all cpp projects in the repo. Maybe we should enforce this via eng/targets/Cpp.Common{props.targets} ?

BrennanConroy: Hopefully a temporary change. If you look at the C++ packages we restore, they are meant to run with v140. I also thought it was causing the build issues, turns out it was other things.

src/SignalR/clients/cpp/src/signalrclient/Build/VS/signalrclient.vcxproj
```diff
@@ -17,16 +17,16 @@
   <Import Project="..\..\..\..\Build\Config.Definitions.props" />
```
jkotalik: Can you do me a big favor and make WindowsTargetPlatformVersion = 10.0.17134.0 everywhere (including here)? This PR I merged from 2.1 did some but I forgot to update more when merging from 2.1 to 2.2. https://github.com/aspnet/AspNetCore/commit/80bb14c3ef423cea4b970cc69ff273056c277428

jkotalik: You should just be able to find replace for 10.0.15063.0

src/SignalR/clients/cpp/test/signalrclienttests/internal_hub_proxy_tests.cpp
```diff
@@ -65,37 +65,6 @@ TEST(on, cannot_register_handler_if_connection_not_in_disconnected_state)
     }
 }
 
-//TEST(invoke_event, invoke_event_invokes_event_and_passes_arguments)
```
halter73: Why are we removing this test and the one below?

BrennanConroy: invoke_event was some internal hub_proxy thing for invoking the `.on` handler of a hub_connection. We don't have hub_proxy.

halter73: I see. You should probably rename this class to something other than internal_hub_proxy_tests.cpp. I'd also clean up signalrclient/CMakeLists.txt as it still includes hub_proxy.cpp and internal_hub_proxy.cpp.

halter73: Also signalrclientdll.vcxproj.

BrennanConroy:
Ping!

halter73 (approved):

PR #7165, opened 2019-01-31
Improve Components error handling

pranavkm:
* Change event handlers IHandleEvent, IHandleAfterEvent to be async.
* Return faulted tasks to Renderer instead of handling exceptions in ComponentBase
* Use ILogger in RemoteRenderer, and log to console in WebAssemblyRenderer

Fixes https://github.com/aspnet/AspNetCore/issues/4964

### DIFF
--- a/src/Components/Blazor/Blazor/src/Rendering/WebAssemblyRenderer.cs
+++ b/src/Components/Blazor/Blazor/src/Rendering/WebAssemblyRenderer.cs
@@ -23,7 +23,7 @@ namespace Microsoft.AspNetCore.Blazor.Rendering
         /// Constructs an instance of <see cref="WebAssemblyRenderer"/>.
         /// </summary>
         /// <param name="serviceProvider">The <see cref="IServiceProvider"/> to use when initializing components.</param>
-        public WebAssemblyRenderer(IServiceProvider serviceProvider): base(serviceProvider)
+        public WebAssemblyRenderer(IServiceProvider serviceProvider) : base(serviceProvider)
         {
             // The browser renderer registers and unregisters itself with the static
             // registry. This works well with the WebAssembly runtime, and is simple for the
@@ -109,7 +109,7 @@ namespace Microsoft.AspNetCore.Blazor.Rendering
                 foreach (var innerException in aggregateException.Flatten().InnerExceptions)
                 {
                     Console.Error.WriteLine(innerException);
-    }
+                }
             }
             else
             {
--- a/src/Components/Blazor/Blazor/src/Rendering/WebAssemblyRenderer.cs
+++ b/src/Components/Blazor/Blazor/src/Rendering/WebAssemblyRenderer.cs
@@ -1,14 +1,13 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
+using System.Threading.Tasks;
 using Microsoft.AspNetCore.Components;
 using Microsoft.AspNetCore.Components.Browser;
 using Microsoft.AspNetCore.Components.Rendering;
 using Microsoft.JSInterop;
 using Mono.WebAssembly.Interop;
-using System;
-using System.Threading;
-using System.Threading.Tasks;
 
 namespace Microsoft.AspNetCore.Blazor.Rendering
 {
@@ -92,6 +91,25 @@ namespace Microsoft.AspNetCore.Blazor.Rendering
                 // This renderer is not used for server-side Blazor.
                 throw new NotImplementedException($"{nameof(WebAssemblyRenderer)} is supported only with in-process JS runtimes.");
             }
+        }
+
+        /// <inheritdoc />
+        protected override bool HandleException(int componentId, IComponent component, Exception exception)
+        {
+            Console.Error.WriteLine($"Unhandled exception executing component {component.GetType().FullName} ({componentId})");
+            if (exception is AggregateException aggregateException)
+            {
+                foreach (var innerException in aggregateException.Flatten().InnerExceptions)
+                {
+                    Console.Error.WriteLine(innerException);
+    }
+            }
+            else
+            {
+                Console.Error.WriteLine(exception);
+            }
+
+            return true;
         }
     }
 }
--- a/src/Components/Blazor/Blazor/src/Rendering/WebAssemblyRenderer.cs
+++ b/src/Components/Blazor/Blazor/src/Rendering/WebAssemblyRenderer.cs
@@ -37,11 +37,13 @@ namespace Microsoft.AspNetCore.Blazor.Rendering
         /// </summary>
         /// <typeparam name="TComponent">The type of the component.</typeparam>
         /// <param name="domElementSelector">A CSS selector that uniquely identifies a DOM element.</param>
-        public void AddComponent<TComponent>(string domElementSelector)
-            where TComponent: IComponent
-        {
-            AddComponent(typeof(TComponent), domElementSelector);
-        }
+        /// <returns>A <see cref="Task"/> that represents the asynchronous rendering of the added component.</returns>
+        /// <remarks>
+        /// Callers of this method may choose to ignore the returned <see cref="Task"/> if they do not
+        /// want to await the rendering of the added component.
+        /// </remarks>
+        public Task AddComponentAsync<TComponent>(string domElementSelector) where TComponent : IComponent
+            => AddComponentAsync(typeof(TComponent), domElementSelector);
 
         /// <summary>
         /// Associates the <see cref="IComponent"/> with the <see cref="WebAssemblyRenderer"/>,
@@ -49,7 +51,12 @@ namespace Microsoft.AspNetCore.Blazor.Rendering
         /// </summary>
         /// <param name="componentType">The type of the component.</param>
         /// <param name="domElementSelector">A CSS selector that uniquely identifies a DOM element.</param>
-        public async void AddComponent(Type componentType, string domElementSelector)
+        /// <returns>A <see cref="Task"/> that represents the asynchronous rendering of the added component.</returns>
+        /// <remarks>
+        /// Callers of this method may choose to ignore the returned <see cref="Task"/> if they do not
+        /// want to await the rendering of the added component.
+        /// </remarks>
+        public Task AddComponentAsync(Type componentType, string domElementSelector)
         {
             var component = InstantiateComponent(componentType);
             var componentId = AssignRootComponentId(component);
@@ -65,8 +72,7 @@ namespace Microsoft.AspNetCore.Blazor.Rendering
                 domElementSelector,
                 componentId);
 
-            // Dispatch the rendering. In the most common cases, this will finish synchronously
-            await RenderRootComponentAsync(componentId);
+            return RenderRootComponentAsync(componentId);
         }
 
         /// <inheritdoc />
--- a/src/Components/Blazor/Blazor/src/Rendering/WebAssemblyRenderer.cs
+++ b/src/Components/Blazor/Blazor/src/Rendering/WebAssemblyRenderer.cs
@@ -94,9 +94,9 @@ namespace Microsoft.AspNetCore.Blazor.Rendering
         }
 
         /// <inheritdoc />
-        protected override bool HandleException(int componentId, IComponent component, Exception exception)
+        protected override bool HandleException(Exception exception)
         {
-            Console.Error.WriteLine($"Unhandled exception executing component {component.GetType().FullName} ({componentId})");
+            Console.Error.WriteLine($"Unhandled exception rendering component:");
             if (exception is AggregateException aggregateException)
             {
                 foreach (var innerException in aggregateException.Flatten().InnerExceptions)
--- a/src/Components/Blazor/Blazor/src/Rendering/WebAssemblyRenderer.cs
+++ b/src/Components/Blazor/Blazor/src/Rendering/WebAssemblyRenderer.cs
@@ -49,7 +49,7 @@ namespace Microsoft.AspNetCore.Blazor.Rendering
         /// </summary>
         /// <param name="componentType">The type of the component.</param>
         /// <param name="domElementSelector">A CSS selector that uniquely identifies a DOM element.</param>
-        public void AddComponent(Type componentType, string domElementSelector)
+        public async void AddComponent(Type componentType, string domElementSelector)
         {
             var component = InstantiateComponent(componentType);
             var componentId = AssignRootComponentId(component);
@@ -65,7 +65,8 @@ namespace Microsoft.AspNetCore.Blazor.Rendering
                 domElementSelector,
                 componentId);
 
-            RenderRootComponent(componentId);
+            // Dispatch the rendering. In the most common cases, this will finish synchronously
+            await RenderRootComponentAsync(componentId);
         }
 
         /// <inheritdoc />
@@ -94,7 +95,7 @@ namespace Microsoft.AspNetCore.Blazor.Rendering
         }
 
         /// <inheritdoc />
-        protected override bool HandleException(Exception exception)
+        protected override void HandleException(Exception exception)
         {
             Console.Error.WriteLine($"Unhandled exception rendering component:");
             if (exception is AggregateException aggregateException)
@@ -108,8 +109,6 @@ namespace Microsoft.AspNetCore.Blazor.Rendering
             {
                 Console.Error.WriteLine(exception);
             }
-
-            return true;
         }
     }
 }
--- a/src/Components/Components/src/ComponentBase.cs
+++ b/src/Components/Components/src/ComponentBase.cs
@@ -203,7 +203,16 @@ namespace Microsoft.AspNetCore.Components
                 // async work is to be performed.
                 StateHasChanged();
 
-                await ProcessLifecycleTaskAsync(task);
+                try
+                {
+                    await task;
+                }
+                catch when (task.IsCanceled)
+                {
+                    // Ignore exceptions from task cancelletions.
+                }
+
+                // Don't call StateHasChanged here. CallOnParametersSetAsync should handle that for us.
             }
 
             await CallOnParametersSetAsync();
@@ -224,11 +233,11 @@ namespace Microsoft.AspNetCore.Components
             StateHasChanged();
 
             return shouldAwaitTask ?
-                ProcessLifecycleTaskAsync(task) :
+                CallStateHasChangedOnAsyncCompletion(task) :
                 Task.CompletedTask;
         }
 
-        private async Task ProcessLifecycleTaskAsync(Task task)
+        private async Task CallStateHasChangedOnAsyncCompletion(Task task)
         {
             try
             {
@@ -236,7 +245,7 @@ namespace Microsoft.AspNetCore.Components
             }
             catch when (task.IsCanceled)
             {
-                // Ignore task cancelation but don't bother issuing a state change.
+                // Ignore exceptions from task cancelletions, but don't bother issuing a state change.
                 return;
             }
 
@@ -255,7 +264,7 @@ namespace Microsoft.AspNetCore.Components
             StateHasChanged();
 
             return shouldAwaitTask ?
-                ProcessLifecycleTaskAsync(task) :
+                CallStateHasChangedOnAsyncCompletion(task) :
                 Task.CompletedTask;
         }
 
--- a/src/Components/Components/src/ComponentBase.cs
+++ b/src/Components/Components/src/ComponentBase.cs
@@ -185,7 +185,7 @@ namespace Microsoft.AspNetCore.Components
             }
             else
             {
-                return SetParametersAsyncCore();
+                return CallOnParametersSetAsync();
             }
         }
 
@@ -206,14 +206,14 @@ namespace Microsoft.AspNetCore.Components
                 await ProcessLifecycleTaskAsync(task);
             }
 
-            await SetParametersAsyncCore();
+            await CallOnParametersSetAsync();
         }
 
-        private Task SetParametersAsyncCore()
+        private Task CallOnParametersSetAsync()
         {
             OnParametersSet();
             var task = OnParametersSetAsync();
-            // If no aync work is to be performed, i.e. the task has already ran to completion
+            // If no async work is to be performed, i.e. the task has already ran to completion
             // or was canceled by the time we got to inspect it, avoid going async and re-invoking
             // StateHasChanged at the culmination of the async work.
             var shouldAwaitTask = task.Status != TaskStatus.RanToCompletion &&
@@ -243,14 +243,20 @@ namespace Microsoft.AspNetCore.Components
             StateHasChanged();
         }
 
-        async Task IHandleEvent.HandleEventAsync(EventHandlerInvoker binding, UIEventArgs args)
+        Task IHandleEvent.HandleEventAsync(EventHandlerInvoker binding, UIEventArgs args)
         {
-            await binding.Invoke(args);
+            var task = binding.Invoke(args);
+            var shouldAwaitTask = task.Status != TaskStatus.RanToCompletion &&
+                task.Status != TaskStatus.Canceled;
 
             // After each event, we synchronously re-render (unless !ShouldRender())
             // This just saves the developer the trouble of putting "StateHasChanged();"
             // at the end of every event callback.
             StateHasChanged();
+
+            return shouldAwaitTask ?
+                ProcessLifecycleTaskAsync(task) :
+                Task.CompletedTask;
         }
 
         Task IHandleAfterRender.OnAfterRenderAsync()
--- a/src/Components/Components/src/ComponentBase.cs
+++ b/src/Components/Components/src/ComponentBase.cs
@@ -32,7 +32,7 @@ namespace Microsoft.AspNetCore.Components
 
         private readonly RenderFragment _renderFragment;
         private RenderHandle _renderHandle;
-        private bool _hasCalledInit;
+        private bool _initialized;
         private bool _hasNeverRendered = true;
         private bool _hasPendingQueuedRender;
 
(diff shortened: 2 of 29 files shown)


src/Components/Blazor/Blazor/src/Rendering/WebAssemblyRenderer.cs
```diff
@@ -31,9 +31,6 @@ public WebAssemblyRenderer(IServiceProvider serviceProvider): base(serviceProvid
             _webAssemblyRendererId = RendererRegistry.Current.Add(this);
         }
 
-        internal void DispatchBrowserEvent(int componentId, int eventHandlerId, UIEventArgs eventArgs)
```
pranavkm: Unused

src/Components/Blazor/Blazor/src/Rendering/WebAssemblyRenderer.cs
```diff
@@ -95,5 +92,24 @@ protected override Task UpdateDisplayAsync(in RenderBatch batch)
                 throw new NotImplementedException($"{nameof(WebAssemblyRenderer)} is supported only with in-process JS runtimes.");
             }
         }
+
+        /// <inheritdoc />
+        protected override bool HandleException(int componentId, IComponent component, Exception exception)
```
pranavkm: We could in theory leave it unhandled in the blazor case. You get a slightly shorter stack trace if you handle it rather than have the runtime catch it.

src/Components/Components/src/ComponentBase.cs
```diff
@@ -177,188 +177,93 @@ void IComponent.Configure(RenderHandle renderHandle)
         public virtual Task SetParametersAsync(ParameterCollection parameters)
         {
             parameters.SetParameterProperties(this);
-            if (!_hasCalledInit)
+            if (!_initialized)
```
pranavkm: Might be useful to diff this file side-by-side.

src/Components/Components/src/Rendering/ComponentState.cs
```diff
@@ -16,22 +15,13 @@ namespace Microsoft.AspNetCore.Components.Rendering
...
-        private readonly IComponent _component;
         private readonly Renderer _renderer;
         private readonly IReadOnlyList<CascadingParameterState> _cascadingParameters;
         private readonly bool _hasAnyCascadingParameterSubscriptions;
-        private RenderTreeBuilder _renderTreeBuilderCurrent;
         private RenderTreeBuilder _renderTreeBuilderPrevious;
         private ArrayBuilder<RenderTreeFrame> _latestDirectParametersSnapshot; // Lazily instantiated
         private bool _componentWasDisposed;
 
-        public int ComponentId => _componentId;
```
pranavkm: VS tells me to use auto properties instead

src/Components/Components/src/Rendering/Renderer.cs
```diff
@@ -209,8 +170,19 @@ private async Task ProcessAsynchronousWork()
...
                 // new work might be added before we check again as a result of waiting for all
                 // the child components to finish executing SetParametersAsync
-                await pendingWork;
-            };
+
+                try
+                {
+                    await pendingWork;
+                }
+                catch when (pendingWork.IsCanceled || HandleException(componentState.ComponentId, componentState.Component, pendingWork.Exception))
```
pranavkm: Using `TaskCancelledException` doesn't always work. For instance, the exception thrown from cancelling a cancellation token results in the context of a Task results in the task transitioning to Canceled, but you still get the `OperationCancelledException`. Inspecting `IsCanceled` seems to be the reliable way to go

src/Components/Components/src/Rendering/Renderer.cs
```diff
@@ -423,14 +406,39 @@ private bool TryDequeueRenderQueueEntry(out RenderQueueEntry entry)
...
+                    // The component might be rendered and disposed in the same batch (if its parent
+                    // was rendered later in the batch, and removed the child from the tree).
+                    batch = batch ?? new List<Task>();
+                    batch.Add(NotifyRenderCompletedAsync(componentState));
+                }
+            }
+
+            if (batch != null)
+            {
+                await Task.WhenAll(batch);
```
pranavkm: Mimics the previous behavior where `NotifyRenderCompleted` was an `async void` method

src/Components/Blazor/Blazor/src/Rendering/WebAssemblyRenderer.cs
SteveSandersonMS: Well spotted! It used to be the way that .Browser.JS triggered event handlers, but looks like we moved to a different code path when we redid the JS interop.

src/Components/Components/src/Rendering/Renderer.cs
```diff
@@ -84,61 +85,21 @@ protected void RenderRootComponent(int componentId)
         /// </summary>
         /// <param name="componentId">The ID returned by <see cref="AssignRootComponentId(IComponent)"/>.</param>
         /// <param name="initialParameters">The <see cref="ParameterCollection"/>with the initial parameters to use for rendering.</param>
-        protected void RenderRootComponent(int componentId, ParameterCollection initialParameters)
+        protected async void RenderRootComponent(int componentId, ParameterCollection initialParameters)
```
SteveSandersonMS: Does this method really serve any purpose now? Seems like we could drop it and just make RenderRootComponentAsync be the only one.

src/Components/Blazor/Blazor/src/Rendering/WebAssemblyRenderer.cs
javiercn: Lol, I just killed this too as part of my changes

src/Components/Components/src/ComponentBase.cs
```diff
@@ -177,188 +177,93 @@ void IComponent.Configure(RenderHandle renderHandle)
...
-            {
-                await parametersTask;
+                await ProcessLifecycleTaskAsync(task);
             }
+
+            await SetParametersAsyncCore();
         }
 
-        private (bool isAsync, Task asyncTask) ProcessLifeCycletask(Task task)
+        private Task SetParametersAsyncCore()
```
SteveSandersonMS: Could we rename this to `CallOnParametersSetAsync`? The current name `SetParametersAsyncCore` is super misleading, because it doesn't actually set the parameters - it just notifies the component that the parameters *were* set earlier.

src/Components/Components/src/Rendering/Renderer.cs
javiercn: @SteveSandersonMS I'm fine with it, so would the recommendation for people be to simply not await the returned task if they don't want to wait for the entire render?

src/Components/Components/src/ComponentBase.cs
```diff
@@ -32,7 +32,7 @@ public abstract class ComponentBase : IComponent, IHandleEvent, IHandleAfterRend
```
SteveSandersonMS: The simplifications in this file look great.

@javiercn, are you happy that these changes preserve the behavior of the lifecycle async flow you designed for prerendering?

src/Components/Components/test/RendererTest.cs
```diff
@@ -431,12 +433,12 @@ public void CanDispatchEventsToTopLevelComponents()
 
             // Act/Assert: Event can be fired
             var eventArgs = new UIEventArgs();
-            renderer.DispatchEvent(componentId, eventHandlerId, eventArgs);
+            await renderer.DispatchEventAsync(componentId, eventHandlerId, eventArgs);
```
SteveSandersonMS: I'm not certain we really should be awaiting here, or in any of the other tests. That would be a significant change.

Up until now, we've always guaranteed that the event dispatcher and initial re-render is synchronous. It returns a Task now for error-handling purposes, but that shouldn't change the contract that event dispatching and re-render is synchronous.

I think we should explicitly *not* be awaiting so that we can observe the synchronous effects.

Or if you prefer to await in most tests but just add a few tests to show that the dispatch and rerender actually occur synchronously (including for nested components), that would be fine with me too.

src/Components/Components/test/RendererTest.cs
```diff
@@ -1362,28 +1352,260 @@ public async Task CanTriggerEventHandlerDisposedInEarlierPendingBatch()
...
+            var result = Assert.Single(renderer.HandledExceptions);
+            Assert.Equal(componentId, result.componentId);
+            var aggregateException = Assert.IsType<AggregateException>(result.exception);
+            Assert.Equal(2, aggregateException.InnerExceptions.Count);
+            Assert.Contains(exception1, aggregateException.InnerExceptions);
+            Assert.Contains(exception2, aggregateException.InnerExceptions);
+        }
+
+        [Fact]
+        public async Task ExceptionsThrownFromHandleAfter_AreHandled()
```
SteveSandersonMS: Typo: `HandleAfter` -> `HandleAfterRender`

src/Components/Server/src/DependencyInjection/RazorComponentsServiceCollectionExtensions.cs
```diff
@@ -94,6 +94,8 @@ public static class RazorComponentsServiceCollectionExtensions
 
         private static void AddStandardRazorComponentsServices(IServiceCollection services)
         {
+            services.AddLogging();
```
SteveSandersonMS: Is this safe to do even if the developer is explicitly adding their own logging service elsewhere in Startup.cs? I'm not super familiar with the internals of this.

src/Components/Server/src/Microsoft.AspNetCore.Components.Server.csproj
```diff
@@ -19,6 +19,7 @@
     <Reference Include="Microsoft.AspNetCore.StaticFiles" />
     <Reference Include="Microsoft.AspNetCore.SpaServices.Extensions" />
     <Reference Include="Microsoft.Extensions.FileProviders.Embedded" />
+    <Reference Include="Microsoft.Extensions.TypeNameHelper.Sources" PrivateAssets="All" />
```
SteveSandersonMS: I'm curious about what this is for.

src/Components/Components/src/Rendering/Renderer.cs
SteveSandersonMS: @javiercn I guess so. There's no extra advantage in providing an `async void` wrapper around an `async Task` method, as far as I can tell.

src/Components/Components/src/ComponentBase.cs
```diff
@@ -177,188 +177,93 @@ void IComponent.Configure(RenderHandle renderHandle)
         public virtual Task SetParametersAsync(ParameterCollection parameters)
```
javiercn: I need to look at this in more detail, I'm not clear if more renders are being introduced.

src/Components/Components/test/RendererTest.cs
SteveSandersonMS: In case you're wondering why I think this matters, there are various scenarios in client-side browser code where the behavior is different depending on whether the effects happen synchronously after a user action (such as a click) or if they are async. For example, the "open popup" or "go full screen" APIs are allowed synchronously after a click, but disallowed asynchronously.

Maybe you wouldn't invoke either of those APIs during rendering, but to avoid us having to reason about those fine details on a per-API basis, it's much more reasonable if we guarantee synchrony as much as possible. Asynchrony is hard to reason about and nearly impossible to track down if it starts leaking in.

src/Components/Components/src/ComponentBase.cs
```diff
@@ -177,188 +177,93 @@ void IComponent.Configure(RenderHandle renderHandle)
...
-            Console.Error.WriteLine($"[{ex.GetType().FullName}] {ex.Message}\n{ex.StackTrace}");
+            StateHasChanged();
         }
 
-        void IHandleEvent.HandleEvent(EventHandlerInvoker binding, UIEventArgs args)
+        async Task IHandleEvent.HandleEventAsync(EventHandlerInvoker binding, UIEventArgs args)
         {
-            var task = binding.Invoke(args);
-            ContinueAfterLifecycleTask(task);
+            await binding.Invoke(args);
```
javiercn: This is not correct. We want to trigger a render after calling binding.Invoke has finished executing the sync part and another optional render after the binding has completed executing in case it does actual async work.

This is similar to what we expect happen on SetParametersAsync.
A single render after all sync code has run
An additional render every time we switch between async/sync

src/Components/Components/src/Rendering/ComponentState.cs
javiercn: @pranavkm Can you not do random cleanups :cry: It distracts from the main point

src/Components/Components/src/ComponentBase.cs
SteveSandersonMS: Ah yes!

@pranavkm Would be awesome if we could also have a unit test to represent the flow mentioned by @javiercn here.

src/Components/Components/src/ComponentBase.cs
javiercn: Agree

src/Components/Components/src/ComponentBase.cs
```diff
@@ -177,188 +177,93 @@ void IComponent.Configure(RenderHandle renderHandle)
...
-            if (task == null)
-            {
-                throw new ArgumentNullException(nameof(task));
-            }
+            OnParametersSet();
+            var task = OnParametersSetAsync();
+            // If no aync work is to be performed, i.e. the task has already ran to completion
+            // or was canceled by the time we got to inspect it, avoid going async and re-invoking
+            // StateHasChanged at the culmination of the async work.
+            var shouldAwaitTask = task.Status != TaskStatus.RanToCompletion &&
```
javiercn: Does this handle task.FromException(e)?

src/Components/Server/src/DependencyInjection/RazorComponentsServiceCollectionExtensions.cs
pranavkm: It should be, but I don't see MVC or SignalR doing this. I'll move this to the test app.

javiercn: It's likely they don't do because it's already added by something else they need like DataProtection. In any case, it's safe to do this, if it's not idempotent it's a bug in the method implementation. That is the intended behavior. 

You call what you need and it will noop if called more than once.

src/Components/Server/src/Microsoft.AspNetCore.Components.Server.csproj
pranavkm: It was for pretty printing the component name. But now that we no longer include it, this reference is gone.

src/Components/Components/test/RendererTest.cs
pranavkm: >  I think we should explicitly not be awaiting so that we can observe the synchronous effects.

I think that would require https://github.com/aspnet/AspNetCore/pull/7240/files to go in first

pranavkm:
I addressed changes to the sources, but I'm going to hold off on test changes (which currently don't compile) until we resolve https://github.com/aspnet/AspNetCore/pull/7240/files. There's a @javiercn's PR sets up the tests correctly, and I would need to rebase my changes on top of it to make the correct assertions.

SteveSandersonMS:
@pranavkm OK, sounds good. Please let us know when this is ready for a final review.

pranavkm:
🆙 📅

src/Components/Blazor/Blazor/src/Rendering/WebAssemblyRenderer.cs
```diff
@@ -50,7 +49,7 @@ public void AddComponent<TComponent>(string domElementSelector)
         /// </summary>
         /// <param name="componentType">The type of the component.</param>
         /// <param name="domElementSelector">A CSS selector that uniquely identifies a DOM element.</param>
-        public void AddComponent(Type componentType, string domElementSelector)
+        public async void AddComponent(Type componentType, string domElementSelector)
```
javiercn: @SteveSandersonMS I think we should simply remove all sync versions of this. As we spoke about in the past, if you don't want to handle errors nor wait for it to finish, don't await the Task

src/Components/Components/src/Rendering/Renderer.cs
```diff
@@ -240,8 +179,18 @@ private async Task ProcessAsynchronousWork()
...
+                try
+                {
+                    await pendingWork;
+                }
+                catch when (!pendingWork.IsCanceled)
+                {
+                    // await will unwrap an AggregateException and throw exactly one inner exception.
+                    // Using pendingWork.Exception gives us access to all of the exception
+                    HandleException(pendingWork.Exception);
+                }
```
javiercn: This try-catch block probably needs to be done before the Task.WhenAll so that we report on the individual exception.

src/Components/Components/src/Rendering/Renderer.cs
```diff
@@ -373,12 +330,12 @@ internal void AddToPendingTasks(Task task)
...
                     break;
                 case TaskStatus.Faulted:
-                    // We want to throw immediately if the task failed synchronously instead of
+                    // We want to immediately handle exceptions if the task failed synchronously instead of
                     // waiting for it to throw later. This can happen if the task is produced by
                     // an 'async' state machine (the ones generated using async/await) where even
                     // the synchronous exceptions will get captured and converted into a faulted
                     // task.
-                    ExceptionDispatchInfo.Capture(task.Exception.InnerException).Throw();
+                    HandleException(task.Exception.GetBaseException());
```
javiercn: Can you explain to me the difference between these two? I thought you had to do ExceptionDispatchInfo.Capture to preserve the original callstack

src/Components/Components/src/Rendering/Renderer.cs
```diff
@@ -482,14 +439,53 @@ private void ProcessRenderQueue()
...
+            }
+
+            if (batch != null)
+            {
+                var aggregateTask = Task.WhenAll(batch);
+                try
+                {
+                    await aggregateTask;
+                }
+                catch when (!aggregateTask.IsCanceled)
```
javiercn: I think you should do this per individual task, not on the Task.WhenAll as you want to handle exceptions from individual components

src/Components/Components/src/Rendering/HtmlRenderer.cs
```diff
@@ -43,12 +44,12 @@ protected override Task UpdateDisplayAsync(in RenderBatch renderBatch)
         /// Renders a component into a sequence of <see cref="string"/> fragments that represent the textual representation
         /// of the HTML produced by the component.
         /// </summary>
-        /// <typeparam name="T">The type of the <see cref="IComponent"/>.</typeparam>
+        /// <typeparam name="TComponent">The type of the <see cref="IComponent"/>.</typeparam>
         /// <param name="initialParameters">A <see cref="ParameterCollection"/> with the initial parameters to render the component.</param>
         /// <returns>A sequence of <see cref="string"/> fragments that represent the HTML text of the component.</returns>
-        public IEnumerable<string> RenderComponent<T>(ParameterCollection initialParameters) where T : IComponent
```
javiercn: I'm all for killing the sync versions of these. @SteveSandersonMS any objection? (We can file a separate issue for this and do it in one go)

src/Components/Components/src/ComponentBase.cs
```diff
@@ -177,188 +177,99 @@ void IComponent.Configure(RenderHandle renderHandle)
...
-            var setParametersIsAsync = false;
-            (setParametersIsAsync, parametersTask) = ProcessLifeCycletask(OnParametersSetAsync());
-            // We always call StateHasChanged here as we want to trigger a rerender after OnParametersSet and
-            // the synchronous part of OnParametersSetAsync has run, triggering another re-render in case there
-            // is additional async work.
-            StateHasChanged();
-            if (setParametersIsAsync)
-            {
-                await parametersTask;
+                await ProcessLifecycleTaskAsync(task);
```
javiercn: This introduces an extra StateHasChanged event. We want StateHasChanged to be called right before every "true" async operation and after the sum of the sync operations and the "sync" part of an async operation runs.

For example:
OnInit
OnInitAsync (Completely synchronous)
OnParametersSet()
OnParametersSetAsync (Completely synchronous)
StateHasChanged()

OnInit
OnInitAsync (Completely synchronous)
OnParametersSet()
OnParametersSetAsync (Sync)
StateHasChanged()
OnParametersSetAsync (Continuation)
StateHasChanged()

src/Components/Components/src/ComponentBase.cs
```diff
@@ -177,188 +177,99 @@ void IComponent.Configure(RenderHandle renderHandle)
...
+            // the synchronous part of OnParametersSetAsync has run.
+            StateHasChanged();
+
+            return shouldAwaitTask ?
+                ProcessLifecycleTaskAsync(task) :
+                Task.CompletedTask;
         }
 
-        private async Task ReRenderAsyncTask(Task task)
+        private async Task ProcessLifecycleTaskAsync(Task task)
```
javiercn: Can we find a better name for this? At this point is basically awaiting an async task, swallowing cancellations and calling StateHasChanged. 

Could we call this something like CallStateHasChangedOnAsyncCompletion or something like that?

src/Components/Blazor/Blazor/src/Rendering/WebAssemblyRenderer.cs
SteveSandersonMS: Sounds good to me.

src/Components/Components/src/ComponentBase.cs
SteveSandersonMS: @javiercn Please let us know if you have had a chance to do this.

src/Components/Components/src/Rendering/HtmlRenderer.cs
SteveSandersonMS: Totally agreed.

src/Components/Components/test/RendererTest.cs
```diff
@@ -464,12 +496,12 @@ public void CanDispatchTypedEventsToTopLevelComponents()
 
             // Act/Assert: Event can be fired
             var eventArgs = new UIMouseEventArgs();
-            renderer.DispatchEvent(componentId, eventHandlerId, eventArgs);
+            await renderer.DispatchEventAsync(componentId, eventHandlerId, eventArgs);
```
SteveSandersonMS: As with the comment above, can we avoid the `await`s in these tests so they are making assertions about synchronous behavior?

SteveSandersonMS:
@pranavkm This is looking great. I appreciate it's ended up being quite complex and fiddly, so thanks for continuing to track down all the details. @javiercn has made some good comments so I'll look forward to seeing the final version of the PR.

javiercn:
@pranavkm Let me know when you need me to look at this again

pranavkm (commented):
🆙 📅

src/Components/Components/src/ComponentBase.cs
```diff
@@ -224,19 +233,19 @@ private Task CallOnParametersSetAsync()
...
         }
 
-        private async Task ProcessLifecycleTaskAsync(Task task)
+        private async Task CallStateHasChangedOnAsyncCompletion(Task task)
         {
             try
             {
                 await task;
             }
             catch when (task.IsCanceled)
```
javiercn: Nit: Is there something specific you can catch in here instead of using task.IsCancelled? It's just weird to read a catch block with a when clause instead of an exception. It's not immediate how the exception thrown and the condition are related.

src/Components/Components/src/Rendering/Renderer.cs
```diff
@@ -150,6 +150,10 @@ protected async Task RenderRootComponentAsync(int componentId, ParameterCollecti
             try
             {
                 await ProcessAsynchronousWork();
+                if (_pendingTasks.Count != 0)
+                {
+                    throw new Exception();
+                }
```
SteveSandersonMS: Is this new code block meant to be here? I'm guessing we wouldn't really want to throw a completely blank `Exception`. Doesn't the `Debug.Assert` on L157 handle this adequately already?

SteveSandersonMS (approved):
This is looking solid to me.

@javiercn Could you give sign-off before @pranavkm merges? I know you've had some pretty detailed involvement in the final changes here.

src/Components/Components/src/Rendering/Renderer.cs
pranavkm: Debugging something and forgot to remove it. `Debug.Asserts` just result in xunit crashing the test process and I had a bug here at some point

src/Components/Components/src/ComponentBase.cs
pranavkm: We get a mix of two different exceptions depending on how the Task was cancelled. Throwing a `OperationCanceledException` (`cancelationToken.ThrowIfCancellationRequested()`) does this. The `Task` becomes cancelled, but the exception continues to be `OperationCancelledException`. Alternatively awaiting a cancelled task `Task.FromCanceled` results in a `TaskCancellationException` which also has the same state. Observing the state of the task seems like less work \ easier to grock.

src/Components/Components/src/Rendering/Renderer.cs
```diff
@@ -329,8 +272,7 @@ public virtual Task InvokeAsync(Func<Task> workItem)
             // This is for example when we run on a system with a single thread, like WebAssembly.
             if (_dispatcher == null)
             {
-                workItem();
-                return Task.CompletedTask;
+                return workItem();
```
pranavkm: @javiercn noticed a blazor test fail and this turned out to be the culprit. `workItem()` returns a `Task` and we weren't awaiting it as part of `InvokeAsync`.

src/Components/Components/src/Rendering/Renderer.cs
```diff
@@ -482,15 +428,45 @@ private void ProcessRenderQueue()
             }
         }
 
-        private void InvokeRenderCompletedCalls(ArrayRange<RenderTreeDiff> updatedComponents)
+        private Task InvokeRenderCompletedCalls(ArrayRange<RenderTreeDiff> updatedComponents)
```
pranavkm: Changed this from `async void` to `Task` returning which we ignore. This follows the same pattern we would tell users to do, plus avoids any state machine if everything is synchronous.

pranavkm (commented):
🆙 📅

src/Components/Components/src/ComponentBase.cs
pranavkm: See https://docs.microsoft.com/en-us/dotnet/standard/parallel-programming/task-cancellation. The thing is a bit involved.

javiercn: Can you add a comment? I'm happy if we have a comment so that people can understand it.

pranavkm: Done

pranavkm:
🆙 📅

javiercn (approved):
Great improvement!

src/Components/Components/src/IHandleAfterRender.cs
```diff
@@ -11,6 +13,7 @@ public interface IHandleAfterRender
         /// <summary>
         /// Notifies the component that it has been rendered.
         /// </summary>
-        void OnAfterRender();
+        /// <returns>A <see cref="Task"/> that represents the asynchronous event handling operation.</returns>
```
rynowak: This looks like a bad copy-paste

pranavkm: I couldn't think of anything else to say here so went with the boiler plate. Suggestions?

rynowak: `asynchronous event handling operation` -> `asynchronous rendering operation`

pranavkm:
Thanks for looking through the PR!

chucker:
Hi,

I saw someone struggle with a dependency injection error in an RC project, where adding an `@inject` statement to their component caused a blank page to render. No error was logged to the browser console or the VS output window.

Will this PR improve this behavior, or should I file a separate issue?

pranavkm:
A new issue with steps to reproduce the error would be great.

PR #7168, opened 2019-01-31
Fix platform benchmark

benaadams:
Resolves https://github.com/aspnet/Benchmarks/issues/612

/cc @sebastienros


sebastienros:
Don't think it will be sufficient, I am working on it, and also fixing the other projects.

PR #7169, opened 2019-01-31
Add readonly modifier to readonly structs

JamesNK:
Started with ObjectMethodExecutor and noticed many other structs could be made readonly.

Please review and check that these won't break anything. If a type can't be made readonly then I'll update the struct with a comment explaining why not.


src/Mvc/samples/MvcSandbox/Startup.cs
```diff
@@ -67,8 +67,6 @@ public void Configure(IApplicationBuilder app)
                     });
 
                 builder.MapApplication();
-
-                builder.MapHealthChecks("/healthz");
```
JamesNK: MapHealthChecks here was a test example. Removed because it isn't needed anymore and it conflict with the real extension method.

halter73:
Did you use an analyzer to find the candidate structs, or did you do this manually?

JamesNK:
Manually. I searched for " struct " and then eye-balled any that weren't already readonly

pranavkm:
Would this PR benefit from using the `in` modifier?

pranavkm:
Mvc, Components, JsonPatch looks :shipit: for me. Somebody more familiar with the other pieces should sign off. That said, if it compiles, my guess is this should be good to go.

JamesNK:
> Would this PR benefit from using the in modifier?

Is there code that could benefit from `in`? Sure. Should it be part of this PR? IMO no 😋

halter73:
I think we should add the `in` modifier in this PR so we can hopefully get a perf improvement passing around large structs. The guidance is to "apply the in modifier to readonly struct parameters larger than System.IntPtr.Size."

https://docs.microsoft.com/en-us/dotnet/csharp/write-safe-efficient-code#apply-the-in-modifier-to-readonly-struct-parameters-larger-than-systemintptrsize

Tratcher (approved):

JamesNK:
@pranavkm @halter73 I've created an issue for that piece of work - https://github.com/aspnet/AspNetCore/issues/7180

This PR is already 25 files across many projects. I don't want to turn this PR into a giant monster.

pranavkm (approved):

halter73 (approved):

JunTaoLuo (approved):

PR #7170, opened 2019-01-31
Attempt to install chromium as part of restore instead of running the…

pranavkm:
… CORS tests

One more go at https://github.com/aspnet/AspNetCore-Internal/issues/1619


pranavkm:
Based on the comment here: https://github.com/GoogleChrome/puppeteer/issues/1602#issuecomment-355545915. Explicitly calling `npm install puppeteer` installs the bundled chromium version. This should help differentiate between actual test failures and networking issues (or possibly resolve them)

sebastienros (approved):

PR #7171, opened 2019-01-31
Fixing benchmark apps

sebastienros:
They should be ignored from the default build logic.

/cc @benaadams


src/Mvc/benchmarkapps/Directory.Build.targets
```diff
@@ -0,0 +1,2 @@
+<Project>
+</Project>
```
dougbu: Add a comment to the empty files about _not_ importing the root `Directory.Build.*` files

dougbu (changes requested):

build/repo.props
```diff
@@ -57,6 +57,14 @@
                       $(RepositoryRoot)src\ProjectTemplates\Web.ProjectTemplates\content\**\*.csproj;
                       $(RepositoryRoot)src\ProjectTemplates\Web.Spa.ProjectTemplates\content\**\*.csproj;
                       " />
+
+    <!-- Exclude the benchmarks because they use <PackageReference>. -->
+    <ProjectToExclude Include="
```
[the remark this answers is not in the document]
natemcmaster: No, this is a restriction we have given ourselves. See https://github.com/aspnet/AspNetCore/blob/master/docs/ReferenceResolution.md for the thinking on why we decided to do this.

src/Servers/Kestrel/perf/PlatformBenchmarks/Directory.Build.props
```diff
@@ -0,0 +1,5 @@
+<Project>
```
jkotalik: @sebastienros does this need to be done for the IIS projects too? https://github.com/aspnet/AspNetCore/tree/master/src/Servers/IIS/IIS/benchmarks/IIS.Performance

src/Mvc/benchmarkapps/Directory.Build.targets
sebastienros: Done

dougbu (changes requested):

src/Mvc/test/Microsoft.AspNetCore.Mvc.FunctionalTests/Microsoft.AspNetCore.Mvc.FunctionalTests.csproj
```diff
@@ -26,8 +26,6 @@
 
   <ItemGroup>
     <Reference Include="Microsoft.AspNetCore.Mvc.Testing" />
-    <ProjectReference Include="..\..\benchmarkapps\BasicApi\BasicApi.csproj" />
-    <ProjectReference Include="..\..\benchmarkapps\BasicViews\BasicViews.csproj" />
```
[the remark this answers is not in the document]
benaadams: Currently not detecting that they are broken https://github.com/aspnet/Benchmarks/issues/612

dougbu (approved):
Discussed offline. Will lose building the benchmark apps in CI as  well as local builds from root. Will also lose the CI tests of the MVC benchmark apps. But, all of this is temporary -- @sebastienros has new automation planned that will catch issues when the apps are run on benchmarks server.

PR #7173, opened 2019-01-31
Add longer timeout for dotnet-watch test

BrennanConroy:
@muratg Test only change.

Also, realized the retry doesn't work for this test because the DOTNET_WATCH_ITERATION count can increase twice if it retries.

Fixes https://github.com/aspnet/AspNetCore-Internal/issues/1696

### DIFF
--- a/build/RepositoryBuild.targets
+++ b/build/RepositoryBuild.targets
@@ -20,7 +20,7 @@
         <AdditionalProperties>
           RepositoryToBuild=%(RepositoryBuildOrder.Identity);
           BuildRepositoryRoot=$([MSBuild]::NormalizeDirectory(%(RepositoryBuildOrder.RootPath)));
-          SkipTests=%(RepositoryBuildOrder.SkipTest)
+          SkipTests=%(RepositoryBuildOrder.SkipTests)
         </AdditionalProperties>
       </BatchedRepository>
     </ItemGroup>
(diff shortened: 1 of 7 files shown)


muratg (commented):
Per @natemcmaster there's another fix in 3.0, and the timeout would not actually work. Let's see if we can find it.

natemcmaster:
This is the flaky test fix I was trying to find that I don't think made it into 2.2: https://github.com/aspnet/DotNetTools/commit/a24b4ee4598b425d9e8b2f041be508c32eb41cf5

BrennanConroy:
So it's not related

muratg (approved):

natemcmaster:
If it fixes it, :shipit:

BrennanConroy:
Am I the only one hitting build timeouts in our 2.2 builds? @natemcmaster @Eilon Should we increase the AzDo timeouts or is there a better solution?

natemcmaster:
Based on a cursory inspection of the log, I think we may be running into known issues which deadlock EFCore tests on AzDO. Increasing the timeout would not help.

I would be in favor of skipping EFCore tests completely when building it as a submodule of AspNetCore. The EFCore repo already run tests on all pull requests to 2.x servicing branches.

@ajcvickers @Eilon  would that be okay with you?

Eilon:
If that logic is correct, I'm fine with it.

muratg (approved):

build/RepositoryBuild.targets
```diff
@@ -56,7 +57,8 @@
       StopOnFirstFailure="false"
       Targets="_TestRepository"
       Properties="BuildNumber=$(BuildNumber);IsFinalBuild=$(IsFinalBuild);Configuration=$(Configuration);_NoBuildRepos=$(_NoBuildRepos)"
-      ContinueOnError="true">
+      ContinueOnError="true"
+      Condition=" @(_BatchedTestRepo.SkipTests) != 'true' ">
```
BrennanConroy: @natemcmaster This doesn't seem to do anything

natemcmaster:
```suggestion
      Condition=" '%(_BatchedTestRepo.SkipTests)' != 'true' ">
```

Might be due to wrong syntax...

BrennanConroy: Oh, the quotes...

BrennanConroy:
@natemcmaster FYI I moved the msbuild check and it finally worked.

natemcmaster:
🎉 thanks for figuring this one out.

PR #7179, opened 2019-02-01
Handle PermanentRedirects in PrerenderTagHelpers

### LINKED ISSUE #6477: Permanent redirect (301) is not supported in the PrerenderTagHelper
I was trying to do a 301 redirect from our JavaScript application using SpaServices prerendering. While I was trying to set the RedirectUrl with StatusCode 301 in our JavaScript server-renderer, a 302 redirect was still happening. 

### Investigation
I started looking into the source and I found out that the PrerenderTagHelper is only using the StatusCode from the result if it's not a redirect. If a RedirectUrl was set, the redirect is always a 302 redirect.

### Solution
The solution seems to add a check to the **PrerenderTagHelper** that checks if the result contains a 301 StatusCode, in that case do a permanent redirect. This is also how it's already being done for example in **SpaPrerenderingExtensions**.

So instead of this:
`
if (!string.IsNullOrEmpty(result.RedirectUrl))
{
ViewContext.HttpContext.Response.Redirect(result.RedirectUrl);
return;
}`

It has to be like this:
`if (!string.IsNullOrEmpty(result.RedirectUrl))
{
var permanentRedirect = result?.StatusCode == 301;
ViewContext.HttpContext.Response.Redirect(result.RedirectUrl, permanentRedirect);
return;
}`

ryanbrandenburg:
Fixes #6477.

### DIFF
--- a/src/Middleware/SpaServices/src/Prerendering/PrerenderTagHelper.cs
+++ b/src/Middleware/SpaServices/src/Prerendering/PrerenderTagHelper.cs
@@ -31,11 +31,11 @@ namespace Microsoft.AspNetCore.SpaServices.Prerendering
         /// <param name="serviceProvider">The <see cref="IServiceProvider"/>.</param>
         public PrerenderTagHelper(IServiceProvider serviceProvider)
         {
-            var hostEnv = (IHostingEnvironment) serviceProvider.GetService(typeof(IHostingEnvironment));
-            _nodeServices = (INodeServices) serviceProvider.GetService(typeof(INodeServices)) ?? _fallbackNodeServices;
+            var hostEnv = (IHostingEnvironment)serviceProvider.GetService(typeof(IHostingEnvironment));
+            _nodeServices = (INodeServices)serviceProvider.GetService(typeof(INodeServices)) ?? _fallbackNodeServices;
             _applicationBasePath = hostEnv.ContentRootPath;
-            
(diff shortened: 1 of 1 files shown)


src/Middleware/SpaServices/src/Prerendering/PrerenderTagHelper.cs
```diff
@@ -102,7 +102,8 @@ public override async Task ProcessAsync(TagHelperContext context, TagHelperOutpu
             if (!string.IsNullOrEmpty(result.RedirectUrl))
             {
                 // It's a redirection
-                ViewContext.HttpContext.Response.Redirect(result.RedirectUrl);
+                var permanentRedirect = renderResult.StatusCode.GetValueOrDefault() == 301;
```
SteveSandersonMS: This doesn't look right. `renderResult` isn't defined anywhere in the file. Should it be `result`?

SteveSandersonMS (changes requested):

src/Middleware/SpaServices/src/Prerendering/PrerenderTagHelper.cs
ryanbrandenburg: Yes.

ryanbrandenburg:
:up::date:

SteveSandersonMS (approved):
Looks great!

PR #7198, opened 2019-02-01
Prkrishn/merge 2.2

pranavkm:


pranavkm:
https://github.com/aspnet/AspNetCore/pull/7161 is borked

Tratcher (approved):

src/Middleware/CORS/test/FunctionalTests/CORS.FunctionalTests.npmproj
```diff
@@ -12,5 +12,10 @@
     <IsTestProject>false</IsTestProject>
   </PropertyGroup>
 
+  <Target Name="InstallPuppeteer" BeforeTargets="Restore">
+    <!-- Explicitly install puppeteer. This will install the bundled Chromium as part of restore instead of as part of execution -->
+    <Exec Command="npm install puppeteer" />
```
natemcmaster: Any reason not to put this into the package.json file instead? That would give us consistent install results and reduce churn on package-lock.json.

pranavkm: Sure. I wasn't aware we could do that 😁:

```js
scripts: { "install": "npm install puppeteer" }
```?

pranavkm: I'll fix it up in master

natemcmaster: Actually, I meant you should list puppeteer in devDependencies so you don't need to explicitly `npm install` this package.

pranavkm: It is listed. However, an explicit install is required to download chromium as part of `npm install` as opposed to as part of running. See https://github.com/GoogleChrome/puppeteer/issues/1602#issuecomment-355545915

natemcmaster: That's strange. I don't see mention of this in the docs, and that particular comment makes it seem like the problem was the user invoking `yarn` incorrectly.

Related note: I looked at the [docs for puppeteer](https://github.com/GoogleChrome/puppeteer). Would it make sense to use puppeteer-core instead? Most dev machines have Chrome already. And we could add a step to install the full puppeteer package instead on CI if AzDO agents don't have it.

pranavkm: Running the code (`puppeteer.launch`) installs chromium, however it adds an element of unreliability to the test execution. Moving it to restore seemed like one fewer things to diagnose during test failures. We generally haven't had issues with it until more recently where tests on TeamCity OSX seem to be timing out (3 mins) when I guess it's attempting to download chromium.

With regards to Chrome, they do recommend using the bundled version of Chromium: https://github.com/GoogleChrome/puppeteer#q-why-doesnt-puppeteer-vxxx-work-with-chromium-vyyy. The installed file gets cached under node_modules, so devs typically wouldn't see a lot of noise from it.

PR #7199, opened 2019-02-01
Only publish artifacts on successful builds (unless specified otherwise)

natemcmaster:
This resolves an error that happened in this build: https://dev.azure.com/dnceng/internal/_build/results?buildId=85382

The build failed in Linux_x64 due to network timeout from Docker, but the Linux_x64_Packages artifact was published anyways without running the afterBuild step that filters out duplicate packages. As a result, the deployment had duplicate copies of Microsoft.AspNetCore.App (one from the Linux build, the other from the Windows build).

### DIFF
--- a/.azure/pipelines/ci.yml
+++ b/.azure/pipelines/ci.yml
@@ -73,6 +73,7 @@ jobs:
       path: artifacts/installers/
     - name: Windows_Logs
       path: artifacts/logs/
+      publishOnError: true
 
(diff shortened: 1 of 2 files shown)


dougbu (approved):

natemcmaster:
Updated to include a fix for missing codesignatures on Microsoft.AspNetCore.AzureAppServices.SiteExtension

cc @pakrym

PR #7210, opened 2019-02-02
Handle null attributes in generated delegates

davidfowl:
Found while implementing my custom framework on top of endpoint routing

### DIFF
--- a/src/Http/Routing/src/Builder/EndpointRouteBuilderExtensions.cs
+++ b/src/Http/Routing/src/Builder/EndpointRouteBuilderExtensions.cs
@@ -327,9 +327,15 @@ namespace Microsoft.AspNetCore.Builder
             };
 
             // Add delegate attributes as metadata
-            foreach (var attribute in requestDelegate.Method.GetCustomAttributes())
+            var attributes = requestDelegate.Method.GetCustomAttributes();
+
+            // This can be null if the delegate is a dynamic method or compiled from an expression tree
+            if (attributes != null)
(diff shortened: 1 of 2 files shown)


src/Http/Routing/src/Builder/EndpointRouteBuilderExtensions.cs
```diff
@@ -327,9 +327,15 @@ public static class EndpointRouteBuilderExtensions
             };
 
             // Add delegate attributes as metadata
-            foreach (var attribute in requestDelegate.Method.GetCustomAttributes())
+            var attributes = requestDelegate.Method.GetCustomAttributes();
```
JamesNK: Could `Method` ever be null? Is it worth `requestDelegate.Method?.GetCustomAttributes()`

JamesNK (approved):

src/Http/Routing/src/Builder/EndpointRouteBuilderExtensions.cs
JamesNK: Is this exception from `Method` ever something to worry about? https://docs.microsoft.com/en-us/dotnet/api/system.delegate.method?view=netcore-2.2#exceptions

davidfowl: Hmm, I'm not sure what combination of things would throw that exceptions. I'll see if I can reproduce it.

Issue #7052, opened 2019-01-28
Model binding infinite recursion or infinite loop with complex collections


dougbu:
### Describe the bug
Collections of complex types may bind unconditionally until `int.MaxValue` elements are reached. This relates to #4939 but involves built-in model binders.

Similarly, nested types may recursively bind until the stack is exhausted. This may relate to #6616.

### To Reproduce
1. Using the current version of ASP.NET Core i.e. `master`
2. Create a simple model such as
    ``` c#
    private class LocationInfo
    {
        [FromHeader]
        public string GpsCoordinates { get; set; }

        public int Zipcode { get; set; }
    }
    ```
3. Add an action to a controller or a handler to a page with a `List<LocationInfo>` argument
4. Post a request containing a `` header to the action
5. Experience extremely slow response (as the collection fills up `int.MaxValue` items

### Expected behavior
Expect model binding to detect the problem and `throw` an `Exception` informing the developer.

### Additional context
Beyond `[FromHeader]`, the problem may occur with `[FromServices]` or a custom greedy model binder. It's possible other models can cause infinite recursion or extremely large collections in model binding.

dougbu:
Hit this while working on #6793. That fix does not cover all the cases (obviously).

We need reasonable limits e.g. `MvcOptions.MaxModelBindingCollectionSize` and `MvcOptions.MaxModelBindingRecursionDepth` to put the breaks on these cases and inform the developer. Suggest the new options should default to match `FormReader.DefaultValueCountLimit` and `MvcOptions.MaxValidationDepth`, respectively.

dougbu:
Self-assigning since I started working on this after hitting the problem while doing #6793.

dougbu:
3e0c75187c

PR #7214, opened 2019-02-02
Place limits on model binding collection size and recursion depth

dougbu:
- #7052
- add MvcOptions.MaxModelBindingCollectionSize` and `MvcOptions.MaxModelBindingRecursionDepth`

nits:
- update syntax of a few `Resources.Designer.cs` files (I ran `/t:resx` on Mvc.sln)
- take VS suggestions in a few test classes


src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/DefaultModelBindingContext.cs
```diff
@@ -184,6 +188,26 @@ public override ModelBindingResult Result
             }
         }
 
+        private int MaxModelBindingRecursionDepth
+        {
+            get
+            {
+                if (_maxModelBindingRecursionDepth == 0)
```
rynowak: suggestion - use `int?` to make the sentinel value clear

src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/CollectionModelBinder.cs
```diff
@@ -319,8 +358,12 @@ private Task<CollectionResult> BindComplexCollection(ModelBindingContext binding
             else
             {
                 indexNamesIsFinite = false;
-                indexNames = Enumerable.Range(0, int.MaxValue)
-                                       .Select(i => i.ToString(CultureInfo.InvariantCulture));
+                var limit = _maxModelBindingCollectionSize == int.MaxValue ?
+                    int.MaxValue :
```
rynowak: suggestion - use `int?` to make the sentinel value clear

src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/DefaultModelBindingContext.cs
```diff
@@ -184,6 +188,26 @@ public override ModelBindingResult Result
...
+                {
+                    var mvcOptions = ActionContext?.HttpContext?.RequestServices?.GetService<IOptions<MvcOptions>>();
+                    if (mvcOptions == null)
+                    {
+                        // Ignore incomplete initialization. This must be a test scenario.
+                        return 32;
```
rynowak: Should this be a constant on `MvcOptions`?

src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/CollectionModelBinder.cs
```diff
@@ -22,14 +25,17 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding.Binders
     public class CollectionModelBinder<TElement> : ICollectionModelBinder
     {
         private static readonly IValueProvider EmptyValueProvider = new CompositeValueProvider();
+        private readonly int _maxModelBindingCollectionSize = FormReader.DefaultValueCountLimit;
```
rynowak: should this be a constant on `MvcOptions`?

src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/DefaultModelBindingContext.cs
```diff
@@ -299,6 +323,21 @@ public override NestedScope EnterNestedScope()
         {
             _stack.Push(_state);
 
+            // Would this new scope (which isn't in _stack) exceed the allowed recursion depth? That is, has the model
+            // binding system already nested MaxModelBindingRecursionDepth binders?
+            if (_stack.Count >= MaxModelBindingRecursionDepth)
+            {
+                // Find the root of this deeply-nested model.
+                var states = _stack.ToArray();
```
rynowak: This is why stack is the worst data structure in .net

src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/MvcOptions.cs
```diff
@@ -209,6 +213,74 @@ public int MaxModelValidationErrors
...
+        /// <remarks>
+        /// <para>
+        /// When binding a collection, some element binders may succeed unconditionally and model binding may run out
+        /// of memory. This limit constrains such unbounded collection growth.
+        /// </para>
+        /// <para>
+        /// This limit does not <em>correct</em> the bound model. The <see cref="InvalidOperationException"/> instead
+        /// informs the developer of an issue in their model or model binder.
+        /// </para>
+        /// <para>
```
rynowak: I think it's worth mentioning explicitly that this is a safeguard against incorrectly written model binders or problematic models, we expect the developer to fix their code. You hint at this, but I'd like it to be more explict.

src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/MvcOptions.cs
```diff
@@ -209,6 +213,74 @@ public int MaxModelValidationErrors
...
+        /// This limit does not apply to collections of simple types. When
+        /// <see cref="CollectionModelBinder{TElement}"/> relies entirely on <see cref="IValueProvider"/>s, it cannot
+        /// create collections larger than the available data.
+        /// </para>
+        /// </remarks>
+        /// <value>The default value is <c>1024</c>, matching <see cref="FormReader.DefaultValueCountLimit"/>.</value>
```
rynowak: It would be good to see guidance that the developer can set this to `int.MaxValue` to effectively remove the limit.

src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/MvcOptions.cs
```diff
@@ -209,6 +213,74 @@ public int MaxModelValidationErrors
...
+        /// <para>
+        /// For models that are very deep or infinitely recursive, model binding may result in stack overflow. This
+        /// limit constrains such recursion.
+        /// </para>
+        /// <para>
+        /// This limit does not <em>correct</em> the bound model. The <see cref="InvalidOperationException"/> instead
+        /// informs the developer of an issue in their model.
+        /// </para>
+        /// </remarks>
+        /// <value>The default value is <c>32</c>, matching the default <see cref="MaxValidationDepth"/> value.</value>
```
rynowak: I have similar comments to the above

src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/Resources.resx
```diff
@@ -493,4 +493,12 @@
...
     <value>'{0}' requires a reference to '{1}'. Configure your application by adding a reference to the '{1}' package and calling '{2}.{3}' inside the call to '{4}' in the application startup code.</value>
   </data>
+  <data name="ModelBinding_ExceededMaxModelBindingCollectionSize" xml:space="preserve">
+    <value>Collection bound to '{0}' exceeded {1}.{2} ({3}). Remove properties that bind unconditionally from {4}.</value>
+    <comment>{0} is model name, {1} is MvcOptions, {2} is MaxModelBindingCollectionSize, {3} is option value, {4} is affected type.</comment>
+  </data>
+  <data name="ModelBinding_ExceededMaxModelBindingRecursionDepth" xml:space="preserve">
+    <value>Model binding system exceeded {0}.{1} ({2}). Reduce the potential nesting of {3}.</value>
+    <comment>{0} is MvcOptions, {1} is MaxModelBindingRecursionDepth, {2} is option value, {3} is (loopy or deeply nested) top-level model type.</comment>
+  </data>
```
rynowak: These error message should call out that you should go to `MvcOptions.xyz` and understand this more. I know you mention the existing of these properties, but there's good documentation there that we can refer to.

src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/Resources.resx
```diff
@@ -493,4 +493,12 @@
   <data name="ReferenceToNewtonsoftJsonRequired" xml:space="preserve">
     <value>'{0}' requires a reference to '{1}'. Configure your application by adding a reference to the '{1}' package and calling '{2}.{3}' inside the call to '{4}' in the application startup code.</value>
   </data>
+  <data name="ModelBinding_ExceededMaxModelBindingCollectionSize" xml:space="preserve">
+    <value>Collection bound to '{0}' exceeded {1}.{2} ({3}). Remove properties that bind unconditionally from {4}.</value>
```
rynowak: I think this is a little more terse than is helpful. In particular `Remove properties that bind unconditionally from {4}.` - I don't think that's going to be very clear to anyone that see this error. I'd suggest something like:

> This is an internal limit that is intended to prevent unbounded growth with an problematic model binder or model type. This error can be cause by writing a model like ....

src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/CollectionModelBinder.cs
```diff
@@ -359,6 +402,25 @@ private Task<CollectionResult> BindComplexCollection(ModelBindingContext binding
...
+            {
+                // Look for a non-empty name. Both ModelName and OriginalModelName may be empty at the top level.
+                var name = string.IsNullOrEmpty(bindingContext.ModelName) ?
+                    (string.IsNullOrEmpty(bindingContext.OriginalModelName) &&
+                        bindingContext.ModelMetadata.MetadataKind != ModelMetadataKind.Type ?
+                        bindingContext.ModelMetadata.Name :
+                        bindingContext.OriginalModelName) : // This name may unfortunately be empty.
+                    bindingContext.ModelName;
+
+                throw new InvalidOperationException(Resources.FormatModelBinding_ExceededMaxModelBindingCollectionSize(
```
rynowak: What are your thoughts on making this a 400 vs a 500?

I considered this, and I thought that a 500 is better since we think this is a problem with the code you wrote (models and model binders). 500 is going to make more noise and will be harder to just brush off.

src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/CollectionModelBinder.cs
dougbu: Sure. `internal` or `public`?

src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/CollectionModelBinder.cs
dougbu: This isn't a sentinel. True that people can set insanely large limits and turn things off effectively. But, &hellip;
- ` _maxModelBindingCollectionSize + 1` would overflow if this wasn't a ternary expression.
- always need a value for the `Range(...)` expression just below

src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/CollectionModelBinder.cs
dougbu: `400` doesn't sound right to me for this case. Leaving as-is.

src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/DefaultModelBindingContext.cs
dougbu: :ok:

src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/DefaultModelBindingContext.cs
dougbu: :ok: `internal` or `public`?

src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/MvcOptions.cs
dougbu: I'll revise the doc comments and resources in the next iteration

src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/ArrayModelBinderProvider.cs
```diff
@@ -23,15 +24,17 @@ public IModelBinder GetBinder(ModelBinderProviderContext context)
...
 
-                var binderType = typeof(ArrayModelBinder<>).MakeGenericType(elementType);
                 var loggerFactory = context.Services.GetRequiredService<ILoggerFactory>();
+                var mvcOptions = context.Services.GetRequiredService<IOptions<MvcOptions>>().Value;
                 return (IModelBinder)Activator.CreateInstance(
                     binderType,
                     elementBinder,
                     loggerFactory,
-                    true /* allowValidatingTopLevelNodes */);
+                    true /* allowValidatingTopLevelNodes */,
```
rynowak: you can just use named arguments. It's 2019 😆

src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/Binders/CollectionModelBinder.cs
rynowak: internal

src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/DefaultModelBindingContext.cs
rynowak: internal

src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/ModelBinding/DefaultModelBindingContext.cs
```diff
@@ -184,6 +188,26 @@ public override ModelBindingResult Result
             }
         }
 
+        private int MaxModelBindingRecursionDepth
+        {
+            get
+            {
+                if (_maxModelBindingRecursionDepth == 0)
+                {
+                    var mvcOptions = ActionContext?.HttpContext?.RequestServices?.GetService<IOptions<MvcOptions>>();
```
rynowak: This is pretty ugly. Would it be reasonable for the calling code to set this value instead of service locatoring?

dougbu (commented):
:up::date: to address PR comments and get with the new MVC file layout

rynowak (approved):

PR #7224, opened 2019-02-03
Add CancelInvocation support to MsgPack in TS client

BrennanConroy:
Fixes https://github.com/aspnet/AspNetCore/issues/7157

### DIFF
--- a/src/SignalR/clients/ts/FunctionalTests/ts/HubConnectionTests.ts
+++ b/src/SignalR/clients/ts/FunctionalTests/ts/HubConnectionTests.ts
@@ -184,6 +184,39 @@ describe("hubConnection", () => {
                 });
             });
 
+            it("can stream server method and cancel stream", (done) => {
+                const hubConnection = getConnectionBuilder(transportType)
+                    .withHubProtocol(protocol)
+                    .build();
+
+                hubConnection.onclose((error) => {
+                    expect(error).toBe(undefined);
+                    done();
+                });
+
+                hubConnection.on("StreamCanceled", () => {
+                    hubConnection.stop();
+                });
+
+                hubConnection.start().then(() => {
+                    const subscription = hubConnection.stream<string>("InfiniteStream").subscribe({
+                        complete() {
+                        },
+                        error(err) {
+                            fail(err);
+                            hubConnection.stop();
+                        },
+                        next() {
+                        },
+                    });
+
+                    subscription.dispose();
+                }).catch((e) => {
+                    fail(e);
+                    done();
+                });
+            });
+
             it("rethrows an exception from the server when invoking", (done) => {
                 const errorMessage = "An unexpected error occurred invoking 'ThrowException' on the server. InvalidOperationException: An error occurred.";
                 const hubConnection = getConnectionBuilder(transportType)
--- a/src/SignalR/clients/ts/signalr-protocol-msgpack/tests/MessagePackHubProtocol.test.ts
+++ b/src/SignalR/clients/ts/signalr-protocol-msgpack/tests/MessagePackHubProtocol.test.ts
@@ -202,4 +202,19 @@ describe("MessagePackHubProtocol", () => {
         const buffer = new MessagePackHubProtocol().writeMessage({ type: MessageType.Ping });
         expect(new Uint8Array(buffer)).toEqual(payload);
     });
+
+    it("can write cancel message", () => {
+        const payload = new Uint8Array([
+            0x07, // length prefix
+            0x93, // message array length = 1 (fixarray)
+            0x05, // type = 6 = Ping (fixnum)
+            0x80, // headers
+            0xa3, // invocationID = string length 3
+            0x61, // a
+            0x62, // b
+            0x63, // c
+        ]);
+        const buffer = new MessagePackHubProtocol().writeMessage({ type: MessageType.CancelInvocation, invocationId: "abc" });
+        expect(new Uint8Array(buffer)).toEqual(payload);
+    });
 });
--- a/src/SignalR/clients/ts/signalr-protocol-msgpack/tests/MessagePackHubProtocol.test.ts
+++ b/src/SignalR/clients/ts/signalr-protocol-msgpack/tests/MessagePackHubProtocol.test.ts
@@ -207,7 +207,7 @@ describe("MessagePackHubProtocol", () => {
         const payload = new Uint8Array([
             0x07, // length prefix
             0x93, // message array length = 1 (fixarray)
-            0x05, // type = 6 = Ping (fixnum)
+            0x05, // type = 5 = CancelInvocation (fixnum)
             0x80, // headers
             0xa3, // invocationID = string length 3
             0x61, // a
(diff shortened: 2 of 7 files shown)


src/SignalR/clients/ts/signalr-protocol-msgpack/tests/MessagePackHubProtocol.test.ts
```diff
@@ -202,4 +202,19 @@ describe("MessagePackHubProtocol", () => {
         const buffer = new MessagePackHubProtocol().writeMessage({ type: MessageType.Ping });
         expect(new Uint8Array(buffer)).toEqual(payload);
     });
+
+    it("can write cancel message", () => {
+        const payload = new Uint8Array([
+            0x07, // length prefix
+            0x93, // message array length = 1 (fixarray)
+            0x05, // type = 6 = Ping (fixnum)
```
JamesNK:
```suggestion
            0x05, // type = 5 = CancelInvocation (fixnum)
```

JamesNK (commented):
Add a TS functional test that uses msgpack with cancel.

BrennanConroy:
Unfortunately there is no good way in the functional tests to test that it actually did anything. Unless I do some ugly stuff with using the `IHubContext<THub>` to send a message to the client when the streaming cancellation occurs.

src/SignalR/clients/ts/signalr/tests/HubConnection.test.ts
```diff
@@ -1075,6 +1075,8 @@ describe("HubConnection", () => {
                     // Observer should no longer receive messages
                     expect(observer.itemsReceived).toEqual([1]);
 
+                    // Close message sent asynchronously so we need to wait
+                    await delay(50);
```
halter73: Do you want flaky tests? Because this is how you get flaky tests.

You need a much higher timeout (like 30 seconds) and a way to complete early if the close message arrives before the timeout.

src/SignalR/clients/ts/FunctionalTests/ts/HubConnectionTests.ts
```diff
@@ -184,6 +184,39 @@ describe("hubConnection", () => {
...
+                    expect(error).toBe(undefined);
+                    done();
+                });
+
+                hubConnection.on("StreamCanceled", () => {
+                    hubConnection.stop();
+                });
+
+                hubConnection.start().then(() => {
+                    const subscription = hubConnection.stream<string>("InfiniteStream").subscribe({
```
halter73: Who is the stream actually getting cancelled in this test?

BrennanConroy: `subscription.dispose`

halter73:
> Unfortunately there is no good way in the functional tests to test that it actually did anything. Unless I do some ugly stuff with using the IHubContext<THub> to send a message to the client when the streaming cancellation occurs.

Is the "can stream server method and cancel stream" test not verifying this change actually did something?

src/SignalR/clients/ts/signalr/tests/HubConnection.test.ts
BrennanConroy: I can do the setInterval pattern
```
setInterval(() => {
    //check value we're waiting for and dispose interval if found or took too long
}, 10);
```

BrennanConroy:
It is, but I had to do the `IHubContext` stuff to make it verify the stream was closed.

halter73:
Fair. It wasn't *that* ugly.

BrennanConroy:
>It wasn't that ugly.

Opinions

BrennanConroy:
:up: 📅

src/SignalR/clients/ts/signalr/tests/Utils.ts
```diff
@@ -13,9 +13,16 @@ export function registerUnhandledRejectionHandler(): void {
...
 
-export function delay(durationInMilliseconds: number): Promise<void> {
+export function delayUntil(timeoutInMilliseconds: number, condition?: () => boolean): Promise<void> {
     const source = new PromiseSource<void>();
-    setTimeout(() => source.resolve(), durationInMilliseconds);
+    let timeWait: number = 0;
+    const interval = setInterval(() => {
+        timeWait += 10;
+        if ((condition && condition() === true) || timeoutInMilliseconds <= timeWait) {
+            source.resolve();
```
halter73: Shouldn't this fail if `timeoutInMilliseconds <= timeWait` and there was a condition that never succeeded?

src/SignalR/clients/ts/signalr/tests/HubConnection.test.ts
```diff
@@ -66,7 +66,7 @@ describe("HubConnection", () => {
 
                 try {
                     await hubConnection.start();
-                    await delay(500);
+                    await delayUntil(500);
```
halter73: Why did you increase this delay?

src/SignalR/clients/ts/signalr/src/HubConnection.ts
```diff
@@ -156,13 +156,16 @@ export class HubConnection {
...
+        let promiseQueue: Promise<void>;
         const subject = new Subject<T>();
         subject.cancelCallback = () => {
             const cancelInvocation: CancelInvocationMessage = this.createCancelInvocation(invocationDescriptor.invocationId);
 
             delete this.callbacks[invocationDescriptor.invocationId];
 
-            return this.sendWithProtocol(cancelInvocation);
+            return promiseQueue.then(() => {
+                return this.sendWithProtocol(cancelInvocation);
```
halter73: I assume this failed before if you canceled the stream too early. Do we have any regression tests the promiseQueue part of this change?

src/SignalR/clients/ts/signalr/tests/HubConnection.test.ts
BrennanConroy: It went from 500 to 500 :trollface:

src/SignalR/clients/ts/signalr/src/HubConnection.ts
BrennanConroy: The promiseQueue doesn't change with this PR, I just use it somewhere I didn't before.
The test change needed was https://github.com/aspnet/AspNetCore/pull/7224/files#diff-e53c0bba1752f4416da08f9f40b4ac92R1078 to wait for a message since it now goes async.

halter73: I know that the promiseQueue existed before. Are you saying that without this change, the "automatically sends multiple pings" would fail? Do you think that's a reliable regression test for this?

BrennanConroy: The "stream can be canceled" test would fail because TS isn't threaded, so the cancel message wouldn't be sent before getting to the `expect(connection.sentData.length).toBe(3);` check. Now the cancel callback waits until the stream has finished being started before allowing the cancel to be sent.

I think the test is good.

src/SignalR/clients/ts/signalr/tests/Utils.ts
BrennanConroy: Yeah, that makes sense.
Done

halter73 (approved):

PR #7227, opened 2019-02-04
Use Pipelines in PlaintextApp sample

benaadams:
Guess it would be a regression until https://github.com/aspnet/AspNetCore/pull/7110 is merged?

/cc @davidfowl  @jkotalik

 @halter73 @pakrym @Tratcher

### DIFF
--- a/src/Servers/Kestrel/samples/PlaintextApp/Startup.cs
+++ b/src/Servers/Kestrel/samples/PlaintextApp/Startup.cs
@@ -5,6 +5,7 @@ using System;
 using System.IO;
 using System.IO.Pipelines;
 using System.Net;
+using System.Runtime.CompilerServices;
 using System.Text;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Builder;
@@ -31,22 +32,7 @@ namespace PlaintextApp
             response.ContentType = "text/plain";
             response.ContentLength = payload.Length;
 
-            var vt = response.BodyPipe.WriteAsync(payload);
-            if (vt.IsCompletedSuccessfully)
-            {
-                // Signal consumption to the IValueTaskSource
-                vt.GetAwaiter().GetResult();
-                return Task.CompletedTask;
-            }
-            else
-            {
-                return AwaitResult(vt);
-            }
-
-            async Task AwaitResult(ValueTask<FlushResult> flushResult)
-            {
-                await flushResult;
-            }
+            return response.BodyPipe.WriteAsync(payload).GetAsTask();
         }
 
         public static Task Main(string[] args)
@@ -61,6 +47,24 @@ namespace PlaintextApp
                 .Build();
 
             return host.RunAsync();
+        }
+    }
+
+    internal static class ValueTaskExtensions
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static Task GetAsTask<T>(this in ValueTask<T> valueTask)
+        {
+            if (valueTask.IsCompletedSuccessfully)
+            {
+                // Signal consumption to the IValueTaskSource
+                valueTask.GetAwaiter().GetResult();
+                return Task.CompletedTask;
+            }
+            else
+            {
+                return valueTask.AsTask();
+            }
         }
     }
 }
--- a/src/Servers/Kestrel/samples/PlaintextApp/Startup.cs
+++ b/src/Servers/Kestrel/samples/PlaintextApp/Startup.cs
@@ -1,7 +1,9 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.IO;
+using System.IO.Pipelines;
 using System.Net;
 using System.Text;
 using System.Threading.Tasks;
@@ -16,16 +18,24 @@ namespace PlaintextApp
 
         public void Configure(IApplicationBuilder app)
         {
-            app.Run((httpContext) =>
+            app.Run(async (httpContext) =>
             {
                 var response = httpContext.Response;
                 response.StatusCode = 200;
                 response.ContentType = "text/plain";
+                response.ContentLength = _helloWorldBytes.Length;
 
-                var helloWorld = _helloWorldBytes;
-                response.ContentLength = helloWorld.Length;
-                return response.Body.WriteAsync(helloWorld, 0, helloWorld.Length);
+                var pipe = response.BodyPipe;
+                Write(pipe, _helloWorldBytes);
+                await pipe.FlushAsync();
             });
+        }
+
+        private static void Write(PipeWriter pipe, byte[] payload)
+        {
+            var span = pipe.GetSpan(sizeHint: payload.Length);
+            payload.CopyTo(span);
+            pipe.Advance(payload.Length);
         }
 
         public static Task Main(string[] args)
--- a/src/Servers/Kestrel/samples/PlaintextApp/Startup.cs
+++ b/src/Servers/Kestrel/samples/PlaintextApp/Startup.cs
@@ -1,16 +1,13 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
 using System.IO;
-using System.IO.Pipelines;
 using System.Net;
 using System.Runtime.CompilerServices;
 using System.Text;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
-using Microsoft.AspNetCore.Http;
 
 namespace PlaintextApp
 {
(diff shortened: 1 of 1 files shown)


src/Servers/Kestrel/samples/PlaintextApp/Startup.cs
```diff
@@ -16,18 +18,26 @@ public class Startup
...
                 response.StatusCode = 200;
                 response.ContentType = "text/plain";
+                response.ContentLength = _helloWorldBytes.Length;
 
-                var helloWorld = _helloWorldBytes;
-                response.ContentLength = helloWorld.Length;
-                return response.Body.WriteAsync(helloWorld, 0, helloWorld.Length);
+                var pipe = response.BodyPipe;
+                Write(pipe, _helloWorldBytes);
+                await pipe.FlushAsync();
```
JamesNK: FlushAsync returns a `ValueTask<T>` right? Could `ValueTask.IsCompletedSuccessfully` be tested so that the async state machine is only created when its false? Or does the nature of pipelines mean it always be false?

benaadams: Could do; technically you should then also call `GetAwaiter().GetResult()` on the ValueTask to communicate consumption back to the `IValueTaskSource` 

Not sure it matters for pipelines; but people might use it as reference.

pakrym: Why would a state machine be created for a completed task?

This feels like a slight over-optimization for a sample.

benaadams: Because it forces the method to be marked `async` which then always becomes a statemachine.

> This feels like a slight over-optimization for a sample.

Its the go to sample for people that want to test plaintext without working out what's going on in  aspnet/benchmarks; as that's 100 scenerios in one app; with lots of config switching and conditional includes.

pakrym: What about doing result.AsTask()?

benaadams: Is a result task, it would allocate a `Task<FlushResult>`. 

Thought about adding an extension method called `.AsNonGenericTask`; but that seemed a terrible name.

pakrym: I see. I would wait for other opinions, personally I still think sample should be simple.

jkotalik:
@benaadams agreed, we should probably avoid merging this PR until #7110 is in.

src/Servers/Kestrel/samples/PlaintextApp/Startup.cs
```diff
@@ -16,16 +19,34 @@ public class Startup
...
+            var vt = response.BodyPipe.WriteAsync(payload);
+            if (vt.IsCompletedSuccessfully)
+            {
+                // Signal consumption to the IValueTaskSource
+                vt.GetAwaiter().GetResult();
+                return Task.CompletedTask;
+            }
+            else
+            {
+                return AwaitResult(vt);
```
halter73: Is this better than `return vt.AsTask()`?

pakrym: See the thread above.

src/Servers/Kestrel/samples/PlaintextApp/Startup.cs
halter73: I still don't understand the desire fore `AsNonGenericTask()`. So what if `AsTask()` returns `Task<FlushResult>` instead of a `Task` when it is a `Task`? It's not like it has to allocate more to implicitly convert to a `Task`.

Are you just worried about the larger object size? Are you sure the local `AwaitResult` function really more efficient than `AsTask()`?

src/Servers/Kestrel/samples/PlaintextApp/Startup.cs
halter73: Just saw that. I still don't see how `AsTask()` returning a `Task<FlushResult>` proves it's less efficient.

pakrym: It'll have to allocate in completed case.

pakrym: If we replace the entire  `if (vt.IsCompletedSuccessfully) {} else {} ` check with `vt.AsTask()`

src/Servers/Kestrel/samples/PlaintextApp/Startup.cs
benaadams: Its a chunky ol'block of code https://github.com/dotnet/coreclr/blob/9aa68da7d3dcd24761d03c4f41720194ab3638b8/src/System.Private.CoreLib/shared/System/Threading/Tasks/ValueTask.cs#L522-L645 but it doesn't make it much nicer.

What I wanted to do with `AsNonGenericTask()` is remove the whole lot including `IsCompletedSucessfully`; so it would just be:

```csharp
return response.BodyPipe.WriteAsync(payload).AsNonGenericTask();
```
And that would handle the optimization to `Task.CompletedTask`

halter73: Say what you will about GetTaskForValueTaskSource(), but at least it doesn't introduce an async state machine like the local AwaitResult() function, so it's still not obvious to me that AwaitResult() is more efficient than AsTask().

The desire for AsNonGenericTask makes a lot more sense to me now though. Not having to write the IsCompletedSucessfully logic ourselves would be nice. I'm still hoping that someday the awaited pattern we use all over the place will become a relic of the past because it's no longer necessary.

src/Servers/Kestrel/samples/PlaintextApp/Startup.cs
halter73: That'd be awesome if we could replace the whole if/else with AsTask(), but that has nothing to do with whether the current AsTask() is better than AwaitResult().

src/Servers/Kestrel/samples/PlaintextApp/Startup.cs
JamesNK: > This feels like a slight over-optimization for a sample.

Yeah, for a sample the original code was fine. I was coming at this code from the perspective that it is going to be run in benchmarks. It will still be useful to measure if this is faster is the plaintext benchmark.

src/Servers/Kestrel/samples/PlaintextApp/Startup.cs
pakrym: I agree that we should prefer AsTask until proven otherwise.

davidfowl: Don't use AsTask, it'll allocate.

pakrym: AwaitResult allocates as well.

davidfowl: True, it might be the same. It is the slow path so maybe it doesn't matter as much.

src/Servers/Kestrel/samples/PlaintextApp/Startup.cs
```diff
@@ -16,16 +20,19 @@ public class Startup
 
         public void Configure(IApplicationBuilder app)
         {
-            app.Run((httpContext) =>
```
davidfowl: I prefer to have the code inline.

benaadams: Done

src/Servers/Kestrel/samples/PlaintextApp/Startup.cs
```diff
@@ -42,4 +44,22 @@ public static Task Main(string[] args)
             return host.RunAsync();
         }
     }
+
+    internal static class ValueTaskExtensions
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static Task GetAsTask<T>(this in ValueTask<T> valueTask)
```
davidfowl: No need to be an open generic?

benaadams: Changed

src/Servers/Kestrel/samples/PlaintextApp/Startup.cs
```diff
@@ -42,4 +45,22 @@ public static Task Main(string[] args)
             return host.RunAsync();
         }
     }
+
+    internal static class ValueTaskExtensions
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static Task GetAsTask(this in ValueTask<FlushResult> valueTask)
```
Tratcher: Seems like this could be submitted as an improvement to AsTask?

benaadams: Its converting a result type `ValueTask<T>` task-like to a non-result type `Task`

`AsTask()` would covert it to `Task<T>` would need a different name to work with overloading?

davidfowl: Right, this would need to go into pipelines itself.

davidfowl (approved):

halter73 (approved):

anurse:
Any reason this wasn't merged... 2 months ago 😱? Looks approved and builds are green...

davidfowl:
Forgot, or the perf tests weren’t running

anurse:
Are we good to merge or is there something we're waiting on?

davidfowl (approved):

Issue #4067, opened 2018-11-16
Pull request builds run out of disk space in Azure Pipelines


natemcmaster:
PRs against the master branch currently fail because the Hosted VS2017 image has limited disk space

>  System.IO.IOException : There is not enough space on the disk

Options:

* Switch to a new queue. I tried `dotnet-external-temp`, but this doesn't have VS C++ tooling installed
* Work with @dotnet/dnceng to create a new pool
* Don't run tests on PRs?

cc @chcosta @Chrisboh - in case you have other ideas for options available to us.

chcosta:
Which "VS C++" tooling is missing?  I thought we installed all of the VS workloads on the dotnet-external-temp pool.  

@riarenas

riarenas:
`dotnet-external-temp pool is supposed to have a full install of vs2017. we've had some issues with some of the optional components for the different workloads not being there, but we should be able to add any that are missing. Can you tell what exactly is missing from them?

riarenas:
We also have a special pool: AspNetCore-VS2017 where the components in the pool were specifically tailored for ASPNet, is there something missing from that pool?

natemcmaster:
AspNetCore-VS2017 is only for internal builds.

We use VSWhere to search for a VS installation with these required workloads. When we do so, vswhere says there is not compatible installation.

* Microsoft.VisualStudio.Component.VSSDK
* Microsoft.VisualStudio.Component.VC.Tools.x86.x64
* Microsoft.VisualStudio.ComponentGroup.NativeDesktop.Win81
* Microsoft.VisualStudio.Component.VC.ATL
* Microsoft.VisualStudio.Component.Windows10SDK.15063.Desktop

We're also looking for VS >= 15.8.

natemcmaster:
AspNetCore-VS2017 is half-complete and currently unused. We got it mostly ready to use before we punted all AzDO migration work to post-2.2. The things we had left to figure out where JDK and Code signing.

At this point, we could probably kill that queue entirely and use dotnet-internal-temp.

natemcmaster:
At this point, space issues should no longer affect Windows.

For 2.x, we need to switch to new Linux queues, which I believe should be available now.

natemcmaster:
@dougbu let's chat about this one. The code change should be minimal, but requires context to know _which_ line to change.

dougbu:
> For 2.x, we need to switch to new Linux queues, which I believe should be available now.

Now using new queues in 2.x branches.

PR #7228, opened 2019-02-04
Kill processes after completing build and before publishing test results or artifacts

dougbu:
- aspnet/AspNetCore-Internal#1735
- port aspnet/Extensions@e2bb8ac1a0 fix into this repo
  - change `ci-official.yml` and `default-build.yml` instead of `azure-pipelines.yml`


natemcmaster:
We should do this in ci.yml and delete ci-official.yml from 2.x branches. It's not used and I removed that file from the master branch in favor of ci.yml.

dougbu:
> We should do this in ci.yml

`ci.yml` doesn't even contain `buildScript` settings at the moment. Alright to remove `ci-official.yml` and stick with my changes to the `default-build.yml` template?

ryanbrandenburg:
The OOM error on linux seems unrelated. @natemcmaster is that tracked somewhere? I seem to remember you mentioning something similar.

ryanbrandenburg (commented):
The code seems fine. I'll be happy when @natemcmaster is happy.

dougbu:
> The OOM error on linux seems unrelated.

Agreed, I copied from the wrong tab. The bug related to this PR is #4067.

natemcmaster:
> remove ci-official.yml and stick with my changes to the default-build.yml template?

Yes sounds good to me

natemcmaster (approved):

eng/scripts/KillProcesses.ps1
```diff
@@ -0,0 +1,16 @@
...
+taskkill /T /F /IM dotnet.exe
+taskkill /T /F /IM testhost.exe
+taskkill /T /F /IM iisexpress.exe
+taskkill /T /F /IM iisexpresstray.exe
+taskkill /T /F /IM w3wp.exe
+taskkill /T /F /IM msbuild.exe
+taskkill /T /F /IM vbcscompiler.exe
+taskkill /T /F /IM git.exe
+taskkill /T /F /IM vctip.exe
+taskkill /T /F /IM chrome.exe
```
natemcmaster: Can you add `h2spec.exe`? We've had issues with this on occasion.

eng/scripts/KillProcesses.ps1
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env pwsh -c
```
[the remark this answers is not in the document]
dougbu: :ok: you've convinced me 😺 I'll remove this and do `chmod -x` on the file.

eng/scripts/KillProcesses.ps1
dougbu: Sure

dougbu (commented):
Once this goes through, I'll run through the aspnet/Extensions branches to make `KillProcesses.ps1` consistent.

PR #7230, opened 2019-02-04
Improve the error experience in Startup

davidfowl:
- When using reflection to invoke the Configure and ConfigureServices, the exception assistant doesn't properly show the exception even when it happens in user code. Even throw we try to re-throw the TargetInvocationException, it doesn't work super well as the callstack is deep in ExceptonDispatchInfo.Throw.
- Instead, we create a delegate that calls into both Configure and ConfigureServices to avoid using reflection.

Here's what this looks like in VS with an exception today:

image

```
   at WebApplication93.Startup.Configure(IApplicationBuilder app, IHostingEnvironment env) in C:\Users\david\source\repos\WebApplication93\WebApplication93\Startup.cs:line 24
   at Microsoft.AspNetCore.Hosting.ConventionBasedStartup.Configure(IApplicationBuilder app)
   at Microsoft.AspNetCore.HostFilteringStartupFilter.<>c__DisplayClass0_0.<Configure>b__0(IApplicationBuilder app)
   at Microsoft.AspNetCore.Hosting.Internal.WebHost.BuildApplication()
   at Microsoft.AspNetCore.Hosting.Internal.WebHost.StartAsync(CancellationToken cancellationToken)
   at Microsoft.AspNetCore.Hosting.WebHostExtensions.RunAsync(IWebHost host, CancellationToken token, String shutdownMessage)
   at Microsoft.AspNetCore.Hosting.WebHostExtensions.RunAsync(IWebHost host, CancellationToken token)
   at Microsoft.AspNetCore.Hosting.WebHostExtensions.Run(IWebHost host)
   at WebApplication93.Program.Main(String[] args) in C:\Users\david\source\repos\WebApplication93\WebApplication93\Program.cs:line 21
```

After this change:

image

PS: It turns out since Configure is called on an async code path, this doesn't help for exceptions throw there. ConfigureServices works great though.

### DIFF
--- a/src/Hosting/Hosting/src/Internal/ConfigureBuilder.cs
+++ b/src/Hosting/Hosting/src/Internal/ConfigureBuilder.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Linq.Expressions;
 using System.Reflection;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.Extensions.DependencyInjection;
@@ -25,9 +26,14 @@ namespace Microsoft.AspNetCore.Hosting.Internal
             // without the hassle of manually creating a scope.
             using (var scope = builder.ApplicationServices.CreateScope())
             {
+                var instanceArg = Expression.Parameter(typeof(object));
+                var argsArray = Expression.Parameter(typeof(object[]));
+
                 var serviceProvider = scope.ServiceProvider;
                 var parameterInfos = MethodInfo.GetParameters();
                 var parameters = new object[parameterInfos.Length];
+                var arguments = new Expression[parameterInfos.Length];
+
                 for (var index = 0; index < parameterInfos.Length; index++)
                 {
                     var parameterInfo = parameterInfos[index];
@@ -51,8 +57,24 @@ namespace Microsoft.AspNetCore.Hosting.Internal
                                 MethodInfo.DeclaringType.FullName), ex);
                         }
                     }
+
+                    // (ParameterType)args[index]
+                    arguments[index] = Expression.Convert(Expression.ArrayIndex(argsArray, Expression.Constant(index)), parameterInfo.ParameterType);
                 }
-                MethodInfo.Invoke(instance, parameters);
+
+                // We're going to build a dynamic method to invoke the Configure method
+                // void Configure(object instance, object[] args)
+                // {
+                //     ((Startup)instance).Configure((ParameterType)args[1..n])
+                // }
+
+                // ((Startup)instance)
+                var instanceExpr = Expression.Convert(instanceArg, instance.GetType());
+
+                // (Startup)instance).Configure(...)
+                var bodyExpr = Expression.Call(instanceExpr, MethodInfo, arguments);
+                var invoker = Expression.Lambda<Action<object, object[]>>(bodyExpr, instanceArg, argsArray).Compile();
+                invoker.Invoke(instance, parameters);
             }
         }
     }
--- a/src/Hosting/Hosting/src/Startup/ConventionBasedStartup.cs
+++ b/src/Hosting/Hosting/src/Startup/ConventionBasedStartup.cs
@@ -18,39 +18,15 @@ namespace Microsoft.AspNetCore.Hosting
         {
             _methods = methods;
         }
-        
+
         public void Configure(IApplicationBuilder app)
         {
-            try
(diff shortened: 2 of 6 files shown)


src/Hosting/Hosting/src/Internal/ConfigureBuilder.cs
```diff
@@ -51,9 +57,25 @@ private void Invoke(object instance, IApplicationBuilder builder)
...
+                // {
+                //     ((Startup)instance).Configure((ParameterType)args[1..n])
+                // }
+
+                // ((Startup)instance)
+                var instanceExpr = Expression.Convert(instanceArg, instance.GetType());
+
+                // (Startup)instance).Configure(...)
+                var bodyExpr = Expression.Call(instanceExpr, MethodInfo, arguments);
+                var invoker = Expression.Lambda<Action<object, object[]>>(bodyExpr, instanceArg, argsArray).Compile();
```
davidfowl: I can avoid expression compilation in the default template cases:
- Startup.Configure(IApplicationBuilder)
- Startup.Configure(IApplicationBuilder, IHostingEnvironment)

src/Hosting/Hosting/test/StartupManagerTests.cs
```diff
@@ -499,8 +499,7 @@ public void StartupLoaderCanLoadByTypeWithEnvironment()
             var app = new ApplicationBuilder(services);
             app.ApplicationServices = startup.ConfigureServicesDelegate(serviceCollection);
 
-            var ex = Assert.Throws<TargetInvocationException>(() => startup.ConfigureDelegate(app));
-            Assert.IsAssignableFrom<InvalidOperationException>(ex.InnerException);
+            var ex = Assert.Throws<InvalidOperationException>(() => startup.ConfigureDelegate(app));
```
davidfowl: This is a nice side effect of the direct call.

src/Hosting/Hosting/src/Startup/ConventionBasedStartup.cs
```diff
@@ -18,39 +18,15 @@ public ConventionBasedStartup(StartupMethods methods)
...
 
         public IServiceProvider ConfigureServices(IServiceCollection services)
         {
-            try
-            {
-                return _methods.ConfigureServicesDelegate(services);
-            }
-            catch (Exception ex)
-            {
-                if (ex is TargetInvocationException)
```
Tratcher: Isn't the TargetInvocationException still possible for ConfigureServices when you fall back to MethodInfo.Invoke?

pakrym:
This is unfortunate to have to do this. Do you think we should open an issue for VS debugger?

Tratcher (commented):
I don't follow the Expression generation, but the rest looks OK.

src/Hosting/Hosting/test/WebHostBuilderTests.cs
```diff
@@ -139,6 +139,30 @@ public async Task StartupConfigureThrows_Fallback(IWebHostBuilder builder)
...
+                Assert.Equal("Exception from Configure", exception.Message);
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(DefaultWebHostBuilders))]
+        public void StartupConfigureServicesThrowsDoesNotThrowTargetInvocationException(IWebHostBuilder builder)
+        {
+            var server = new TestServer();
+            var hostBuilder = builder.UseServer(server).UseStartup<StartupConfigureServicesThrows>();
```
pakrym: This hits `void` code path, add another one that hits `IServiceProvider` one.

pakrym (approved):
Improvement in customer experience is hard to argue.

davidfowl:
> This is unfortunate to have to do this. Do you think we should open an issue for VS debugger?

For async exceptions, I started a thread and it turns out it's hard to do without deeper changes but the relevant parties are talking about that.

For general target invocation exceptions, now I think we should just try to avoid them where possible.

src/Hosting/Hosting/test/WebHostBuilderTests.cs
davidfowl: The other tests (StartupManagerTests) hit all code paths (because it was exploding before I got it right).

pakrym:
> For general target invocation exceptions, now I think we should just try to avoid them where possible.

Customers love calling `.Invoke`, fixing this would be a huge net win for diagnosability in the framework in general.

davidfowl:
I'm no longer a fan of this change 😢

PR #7231, opened 2019-02-04
Update Azure queues for Linux and MacOS builds

dougbu:
- #4067

### DIFF
--- a/.azure/pipelines/jobs/default-build.yml
+++ b/.azure/pipelines/jobs/default-build.yml
@@ -83,6 +83,7 @@ jobs:
         name: dnceng-linux-internal-temp
       ${{ if ne(variables['System.TeamProject'], 'internal') }}:
         name: dnceng-linux-external-temp
+      vmImage: Linux_Ubuntu_16.04
     ${{ if and(eq(parameters.poolName, ''), eq(parameters.agentOs, 'Windows')) }}:
       ${{ if eq(variables['System.TeamProject'], 'internal') }}:
         name: dotnet-internal-temp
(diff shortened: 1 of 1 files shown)


natemcmaster:
Looks like the Linux queues are missing unzip. We'll need that to bootstrap our build.

dougbu:
> Looks like the Linux queues are missing unzip. We'll need that to bootstrap our build.

Don't see a task that installs such basics and I can't get details for the pool or a list of available images. Let me know where I should be looking.

natemcmaster:
The machine images are defined here: https://dev.azure.com/dnceng/internal/_git/dotnet-helix-machines. We might need to add unzip to the images that are used for agents in this queue. The dnceng team should be able to review and merge changes.

dougbu:
Chatted with someone from dnceng and filed dotnet/core-eng#5144 because they preferred to fix the problem themselves. Unfortunately they're now pushing back.

dougbu:
@natemcmaster if you're available, please review

.azure/pipelines/jobs/default-build.yml
```diff
@@ -73,11 +73,17 @@ jobs:
     ${{ if ne(parameters.poolName, '') }}:
       name: ${{ parameters.poolName }}
     ${{ if and(eq(parameters.poolName, ''), eq(parameters.agentOs, 'macOS')) }}:
-      name: Hosted macOS
+      ${{ if eq(variables['System.TeamProject'], 'internal') }}:
+        name: Hosted Mac Internal
```
natemcmaster: Hosted macOS is fine for internal builds. Not sure what this queue offers, and it may not be worth switching.

natemcmaster (approved):

ryanbrandenburg (approved):

PR #7240, opened 2019-02-04
[Components] Async issues alternative

javiercn:
Lifts the synchronization context from RemoteRenderer into Renderer


src/Components/Server/src/Circuits/CircuitHost.cs
```diff
@@ -52,23 +52,21 @@ public static void SetCurrentCircuitHost(CircuitHost circuitHost)
...
             JSRuntime = jsRuntime ?? throw new ArgumentNullException(nameof(jsRuntime));
-            SynchronizationContext = synchronizationContext ?? throw new ArgumentNullException(nameof(synchronizationContext));
 
             Services = scope.ServiceProvider;
 
             Circuit = new Circuit(this);
 
             Renderer.UnhandledException += Renderer_UnhandledException;
-            SynchronizationContext.UnhandledException += SynchronizationContext_UnhandledException;
+            Renderer.UnhandledSynchronizationException += SynchronizationContext_UnhandledException;
```
javiercn: Temporary until we have the improved error handling mechanism from @pranavkm

src/Components/Server/src/Circuits/CircuitHost.cs
```diff
@@ -85,11 +83,9 @@ public static void SetCurrentCircuitHost(CircuitHost circuitHost)
 
         public IServiceProvider Services { get; }
 
-        public CircuitSynchronizationContext SynchronizationContext { get; }
-
         public async Task InitializeAsync()
         {
-            await SynchronizationContext.Invoke(() =>
+            await Renderer.Invoke(() =>
```
javiercn: Just use Renderer.Invoke as it will internally call SynchronizationContext.Invoke and that way there's no need for the SyncContext to be explicit

src/Components/Browser/src/RendererRegistryEventDispatcher.cs
```diff
@@ -21,10 +21,10 @@ public static class RendererRegistryEventDispatcher
         {
             var eventArgs = ParseEventArgsJson(eventDescriptor.EventArgsType, eventArgsJson);
             var renderer = RendererRegistry.Current.Find(eventDescriptor.BrowserRendererId);
-            renderer.DispatchEvent(
+            renderer.Invoke(() => renderer.DispatchEvent(
```
javiercn: Every different implementation calls dispatch and renderrootcomponent through invoke in order to set the proper SynchronizationContext

src/Components/Components/src/Rendering/HtmlRenderer.cs
```diff
@@ -58,7 +58,11 @@ protected override Task UpdateDisplayAsync(in RenderBatch renderBatch)
         /// <returns>A sequence of <see cref="string"/> fragments that represent the HTML text of the component.</returns>
         private IEnumerable<string> RenderComponent(Type componentType, ParameterCollection initialParameters)
         {
-            var frames = CreateInitialRender(componentType, initialParameters);
+            ArrayRange<RenderTreeFrame> frames = default;
+            Invoke(() =>
+            {
+                frames = CreateInitialRender(componentType, initialParameters);
```
javiercn: This is hacky as I was just doing a proof of concept.
It can be easily fixed by breaking CreateInitialRender and calling GetRenderTreeFrames outside of CreateInitialRender

src/Components/Components/src/Rendering/Renderer.cs
```diff
@@ -20,18 +21,24 @@ public abstract class Renderer : IDisposable
         private readonly Dictionary<int, ComponentState> _componentStateById = new Dictionary<int, ComponentState>();
         private readonly RenderBatchBuilder _batchBuilder = new RenderBatchBuilder();
         private readonly Dictionary<int, EventHandlerInvoker> _eventBindings = new Dictionary<int, EventHandlerInvoker>();
+        private RendererSynchronizationContext _synchronizationContext = new RendererSynchronizationContext();
```
javiercn: We probably want to receive this on the constructor and RendererSynchronizationContext might need to be public

src/Components/Components/src/Rendering/Renderer.cs
```diff
@@ -20,18 +21,24 @@ public abstract class Renderer : IDisposable
...
         private int _lastEventHandlerId = 0;
         private List<Task> _pendingTasks;
 
-        // We need to introduce locking as we don't know if we are executing
-        // under a synchronization context that limits the ammount of concurrency
-        // that can happen when async callbacks are executed.
-        // As a result, we have to protect the _pendingTask list and the
-        // _batchBuilder render queue from concurrent modifications.
-        private object _asyncWorkLock = new object();
+        public event UnhandledExceptionEventHandler UnhandledSynchronizationException
```
javiercn: This is just temporary until we have @pranavkm improved error handling

src/Components/Components/test/CascadingParameterStateTest.cs
```diff
@@ -385,7 +386,7 @@ static CascadingValue<T> CreateCascadingValueComponent<T>(T value, string name =
                 supplierParams.Add("Name", name);
             }
 
-            supplier.SetParameters(supplierParams);
+            renderer.Invoke(() => supplier.SetParameters(supplierParams));
```
javiercn: All code that triggers renders needs to execute in the context of the render, but that doesn't change the behavior of the code with regards to synchronicity (that's why we don't have to await things here)

src/Components/Components/test/CascadingParameterTest.cs
```diff
@@ -312,16 +312,16 @@ public void CascadingValueThrowsIfFixedFlagChangesToTrue()
                 builder.CloseComponent();
             });
             renderer.AssignRootComponentId(component);
-            component.TriggerRender();
+            var _ = renderer.Invoke(() => component.TriggerRender());
 
             // Act/Assert
             isFixed = true;
-            var ex = Assert.Throws<InvalidOperationException>(() => component.TriggerRender());
+            var ex = await Assert.ThrowsAsync<InvalidOperationException>(() => renderer.Invoke(() => component.TriggerRender()));
```
javiercn: This is just because xUnit complains about Throw with Func<Task>. I will replace this for void RunHelper(Func<Task> t) so that we can maintain `Assert.Throw(() => renderer.Invoke(() => component.TriggerRender())`

src/Components/Browser/src/RendererRegistryEventDispatcher.cs
davidfowl: Closures, everywhere...

src/Components/Server/src/Circuits/CircuitHost.cs
```diff
@@ -113,10 +109,10 @@ public async void BeginInvokeDotNetFromJS(string callId, string assemblyName, st
 
             try
             {
-                await SynchronizationContext.Invoke(() =>
+                await Renderer.Invoke(() =>
```
SteveSandersonMS: I find it odd that the renderer "owns" the synchronization context, and the `CircuitHost` (which creates the renderer) calls into the renderer to transfer operations onto that sync context. It seemed more natural that the `CircuitHost` created and owned the sync context.

Perhaps a more flexible arrangement would be for the `CircuitHost` to continue to create and own the sync context, but also pass it as a constructor param to the `RemoteRenderer`, so the renderer can also use it in its `Invoke` implementation.

For `HtmlRenderer`, it could still be responsible for instantiating its own sync context if it wanted, and pass it to its own base constructor.

For `WebAssemblyRenderer`, there's no reason to be instantiating a sync context at all and it would be good to be able to skip that. We could even make it possible for the linker to drop any synccontext-related code if we didn't put sync context stuff in the base `Renderer`, but rather put it in a derived `SynchronizedRenderer` that `HtmlRenderer` and `RemoteRenderer` both inherit from. What do you think?

This way, we aren't locked to the idea that there is exactly one renderer per circuit, or that each renderer has a distinct sync context. We'd have the flexibility to change this around without affecting the base renderer.

src/Components/Components/src/Rendering/Renderer.cs
SteveSandersonMS: I agree - it would be great for this to be a constructor param, as per other comments. That would increase the flexibility a lot, e.g., being able to pass in a sync context owned by the CircuitHost.

Not sure why it would need to be public but you've probably thought this through more than me!

src/Components/Browser/src/RendererRegistryEventDispatcher.cs
SteveSandersonMS: We might not need to make this change if the sync context continued to be owned by the circuit host.

For `RemoteRenderer`, events are always dispatched by the circuit host, so that could continue to be responsible for dispatching to the correct context.

For `HtmlRenderer`, there are no events. For `WebAssemblyRenderer`, there's no sync context. So there aren't any other cases where this is needed, and it's (very slightly!) advantageous not to put `WebAssemblyRenderer` through additional indirection.

src/Components/Components/src/Rendering/Renderer.cs
SteveSandersonMS: Sounds good.

src/Components/Components/src/Rendering/Renderer.cs
SteveSandersonMS: Also then for `WebAssemblyRenderer`, assuming it's not passing in a `RendererSynchronizationContext`, that would allow the linker to strip out `RendererSynchronizationContext` and any dependency.

For `WebAssemblyRenderer` in particular, it would be good if it could have a `null` sync context so we don't accidentally try to `Invoke` on it in an asynchronous way (rather than using `SynchronizationContext.Invoke`, which calls into `ThreadPool.QueueUserWorkItem`), or if that makes the code messy with special cases, consider defining some interface like `ISynchronizationContextDispatcher` that can have a no-op passthrough implementation for the `WebAssemblyRenderer` case.

SteveSandersonMS (commented):
This totally looks on the right lines to me!

A few thoughts posted about how we could try to separate out ownership of the sync context from the base renderer a bit, especially to avoid mixing these concerns into `WebAssemblyRenderer`, but I don't think doing that would change your code very much. And it sounds like you were already thinking of this, considering your comment about receiving the sync context as a constructor parameter.

src/Components/Browser/src/RendererRegistryEventDispatcher.cs
javiercn: I don't feel it's terrible to have 1 closure per event. I don't think it's going to impact the performance in any measurable way, but I'll think about alternatives. When we get threads on WebAssembly we will have to do this anyways.

src/Components/Components/src/Rendering/Renderer.cs
javiercn: I'll figure something out along those lines

src/Components/Server/src/Circuits/CircuitHost.cs
javiercn: Passing the renderer on the constructor seems fine. I would say though that the circuit has the responsibility of creating the sync context but that the renderer should be the actual owner for the synccontext.

In my mind the renderer represents the UI loop (initial render (renderrootcomponent), dispatched interactions (dispatch event) and external event sources (invoke and invoke async). The synchronization context is the abstraction it uses to ensure that everything happens in a single-threaded way.

Each "HostingEnvironment" is responsible to set the appropriate context (circuit, etc) before calling into the renderer to dispatch any of the above events through Invoke.

javiercn:
@SteveSandersonMS I updated the code with what we talked about, anything left here is cleanup which I'm going to do right away.

I checked with the linker and it doesn't remove anything, but I think it's because the way the linker works. We will need to add a descriptor to the assembly for the linker that tells it to remove the pieces we don't care about (the extra constructor with the dispatcher.

I'm not sure if the linker would be smart enough to realize that _dispatcher is always going to be null in that case and to simplify the code, but that's not something I'm super concerned about.

src/Components/Browser/src/RendererRegistryEventDispatcher.cs
javiercn: I got rid of this one

src/Components/Components/src/Rendering/HtmlRenderer.cs
javiercn: This got cleaned up

src/Components/Server/src/Circuits/CircuitHost.cs
javiercn: I passed in the synccontext as a parameter, but kept using it through the renderer. Introduced an IDispatcher interface for this.

src/Components/Components/src/Rendering/Renderer.cs
javiercn: I created an IDispatcher interface, we don't set it up on the web assembly case (hence I removed the closure for dispatch) we simply check if the IDispatcher is a SynchronizationContext and throw if that's the case.

I also provided a CreateDefaultDispatcher static method so that it can be created externally (and the RendererSynchronizationContext becomes an implementation detail).

We need to do work for the linker to strip things in the WASM case, but it's totally doable.

src/Components/Components/test/CascadingParameterStateTest.cs
javiercn: I removed all these by putting them inside the testcomponent and testrenderer, it's the same, just cleaner. I take care of asserting that everything happens synchronously and unwrap and rethrow exceptions as appropriate

src/Components/Components/test/CascadingParameterTest.cs
javiercn: This is fixed now by doing renderer.Invoke inside TriggerRender

javiercn:
@SteveSandersonMS Except for minor cleanups this is good to go.

I want to do a separate pass to cleanup the Renderer APIs after @pranavkm is done with his changes.

src/Components/Components/src/Rendering/HtmlRenderer.cs
```diff
@@ -26,7 +26,8 @@ public class HtmlRenderer : Renderer
         /// </summary>
         /// <param name="serviceProvider">The <see cref="IServiceProvider"/> to use to instantiate components.</param>
         /// <param name="htmlEncoder">A <see cref="Func{T, TResult}"/> that will HTML encode the given string.</param>
-        public HtmlRenderer(IServiceProvider serviceProvider, Func<string, string> htmlEncoder) : base(serviceProvider)
+        public HtmlRenderer(IServiceProvider serviceProvider, Func<string, string> htmlEncoder, IDispatcher dispatcher)
```
SteveSandersonMS: Nit: Looks like we're missing a `<param>` in XML docs for `dispatcher`.

src/Components/Components/src/Rendering/Renderer.cs
```diff
@@ -42,6 +57,22 @@ public Renderer(IServiceProvider serviceProvider)
             _componentFactory = new ComponentFactory(serviceProvider);
         }
 
+        /// <summary>
+        /// Constructs an instance of <see cref="Renderer"/>.
+        /// </summary>
+        /// <param name="serviceProvider">The <see cref="IServiceProvider"/> to be used when initialising components.</param>
```
SteveSandersonMS: I feel bad about saying this, but usually we spell `initialising` in the American way (`initializing`).

src/Components/Components/src/Rendering/Renderer.cs
```diff
@@ -42,6 +57,22 @@ public Renderer(IServiceProvider serviceProvider)
...
+        public Renderer(IServiceProvider serviceProvider, IDispatcher dispatcher)
+        {
+            _componentFactory = new ComponentFactory(serviceProvider);
```
SteveSandersonMS: Could we chain this constructor to the single-parameter one to avoid repeating this line? I know it's not a lot of code to repeat, but I think it would make the intent clearer.

i.e.,

```
public Renderer(IServiceProvider serviceProvider, IDispatcher dispatcher) : this(serviceProvider)
{
    _dispatcher = dispatcher;
}
```

javiercn: I didn't chain it so that the linker could strip this one completely away. I can do so for now, but I think we'll end up changing it back.

src/Components/Components/src/Rendering/Renderer.cs
```diff
@@ -266,9 +296,23 @@ public void DispatchEvent(int componentId, int eventHandlerId, UIEventArgs event
         /// <param name="workItem">The work item to execute.</param>
         public virtual Task Invoke(Action workItem)
         {
-            // Base renderer has nothing to dispatch to, so execute directly
-            workItem();
-            return Task.CompletedTask;
+            if (!(_dispatcher is SynchronizationContext synchronizationContext))
```
SteveSandersonMS: I don't understand this line. Why should an `IDispatcher` also be a `SynchronizationContext`? I would have thought it's none of `Renderer`'s business what the underlying concrete implementation of the `IDispatcher` is.

Is this line only intended for the WebAssembly case? In that case, could we change it just to `if (_dispatcher is null)`?

src/Components/Components/src/Rendering/Renderer.cs
```diff
@@ -266,9 +296,23 @@ public void DispatchEvent(int componentId, int eventHandlerId, UIEventArgs event
...
+            {
+                workItem();
+                return Task.CompletedTask;
+            }
+
+            if (SynchronizationContext.Current == synchronizationContext)
```
SteveSandersonMS: OK I sort of see what's happening. We need the dispatcher to be a `SynchronizationContext` so that we can meaningfully compare it with `SynchronizationContext.Current`.

But does this imply we shouldn't really have an `IDispatcher` interface? With this logic, it's entirely meaningless to implement `IDispatcher` *except* when also inheriting from `SynchronizationContext`. So would it be better to eliminate `IDispatcher` and just use `SynchronizationContext` as the type for `_dispatcher`? Or maybe some concreate class derived from `SynchronizationContext`?

src/Components/Components/src/Rendering/Renderer.cs
SteveSandersonMS: Update: see comment below.

src/Components/Components/src/Rendering/Renderer.cs
```diff
@@ -359,18 +415,30 @@ protected internal virtual void AddToRenderQueue(int componentId, RenderFragment
...
             }
         }
 
+        private void EnsureSynchronizationContext()
+        {
+            // Render operations are not thread-safe, so they need to be serialized.
+            // Plus, any other logic that mutates state accessed during rendering also
+            // needs not to run concurrently with rendering so should be dispatched to
+            // the renderer's sync context.
+            if (_dispatcher is SynchronizationContext synchronizationContext && SynchronizationContext.Current != synchronizationContext)
```
SteveSandersonMS: Also here maybe this could be simiplified to `if (_dispatcher != null)`

src/Components/Components/src/Rendering/Renderer.cs
```diff
@@ -5,6 +5,7 @@
 using System.Collections.Generic;
```
SteveSandersonMS: It's great to see all the locking disappear from this file!

src/Components/Components/src/Rendering/Renderer.cs
SteveSandersonMS: I don't follow. Wouldn't the linker want to remove the 2-param one in the WebAssembly case? The fact that it calls the 1-param one wouldn't stop the linker from removing it. Am I misunderstanding what gets called?

src/Components/Components/src/Rendering/RendererSynchronizationContext.cs
```diff
@@ -6,10 +6,10 @@
 using System.Threading;
 using System.Threading.Tasks;
 
-namespace Microsoft.AspNetCore.Components.Server.Circuits
+namespace Microsoft.AspNetCore.Components.Rendering
 {
     [DebuggerDisplay("{_state,nq}")]
-    internal class CircuitSynchronizationContext : SynchronizationContext
+    internal class RendererSynchronizationContext : SynchronizationContext, IDispatcher
```
SteveSandersonMS: Now that we're passing the dispatcher into the renderer as a constructor argument, I'm curious about whether it would be possible to keep `RendererSynchronizationContext` inside `Microsoft.AspNetCore.Components.Server`, rather than migrating it to `Microsoft.AspNetCore.Components`.

If we could, that would match the layering pretty cleanly, eliminating it entirely from the WebAssembly case even without the linker.

src/Components/Components/test/RendererTest.cs
```diff
@@ -1340,7 +1325,7 @@ public async Task CanTriggerEventHandlerDisposedInEarlierPendingBatch()
             EventComponent component = null;
             Action<UIEventArgs> eventHandler = null;
 
-            eventHandler = _ =>
+            eventHandler = ___ =>
```
SteveSandersonMS: Is this a typo? I've no idea what this change is for :)

src/Components/Components/src/Rendering/Renderer.cs
javiercn: You could potentially have a dispatcher that is not a sync context, with the limitation that all async work should be done through dispatcher.Invoke and dispatcher.InvokeAsync. In real life, I think its most likely that it's also going to be a synchronization context, however I feel bad about offering that interface to users.

Neither WPF nor WinForms do so (they have alternatives like Application.Current.Dispatcher) I think it's just fine if we internally handle the fact that is a synchronization context, part of the reason is done this way is because it's been lifted up from remoterenderer and I didn't want to change the semantics if not needed.

I'll rewrite some things to make the implementation clearer and add some comments

src/Components/Components/test/Rendering/HtmlRendererTests.cs
```diff
@@ -374,15 +376,30 @@ public void RenderComponent_ElementRefsNoops()
...
+            Assert.True(task.IsCompleted);
+            if (task.IsCompletedSuccessfully)
+            {
+                return task.Result;
+            }
+            else
+            {
+                ExceptionDispatchInfo.Capture(task.Exception).Throw();
+                // We never get here
+                return null;
```
SteveSandersonMS: Could this line be a `throw new InvalidOperationException("This line should never be hit");` instead then?

SteveSandersonMS (approved):
Great, thanks for the updates!

src/Components/Components/src/Rendering/Renderer.cs
javiercn: i read this wrong, as discussed offline, I will change it

src/Components/Components/src/Rendering/RendererSynchronizationContext.cs
javiercn: As discussed offline, we want to avoid MVC depending on the server-side bits if possible.

PR #7244, opened 2019-02-04
More logs in forwardinghandler.cpp

pakrym:
We've seen some flakiness in the area but logs I've added were not enough to pinpoint the exact location, add more.


src/Servers/IIS/AspNetCoreModuleV2/OutOfProcessRequestHandler/forwardinghandler.cpp
```diff
@@ -124,34 +124,30 @@ FORWARDING_HANDLER::ExecuteRequestHandler()
...
 
     if (m_pApplication == NULL)
     {
-        hr = E_INVALIDARG;
-        goto Failure;
+        FAILURE(E_INVALIDARG);
     }
 
     hr = m_pApplication->GetProcess(&pServerProcess);
     if (FAILED_LOG(hr))
```
jkotalik: Nit: this is redundant now.

jkotalik (approved):

PR #7248, opened 2019-02-04
Remove obsolete publishing targets and tasks

natemcmaster:
We no longer need to maintain these publishing targets and MSBuild tasks. The release pipeline takes care of these steps for us. (See https://dev.azure.com/dnceng/internal/_release?definitionId=31&view=mine&_a=releases).

The only part of Publish.csproj that I preserved was the step which generates the build asset manifests. These are published to Maestro to flow dependencies automatically to upstream repos.


build/Maestro/Maestro.csproj
```diff
@@ -0,0 +1,49 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.DotNet.Maestro.Tasks" Version="1.1.0-beta.19065.2" />
+  </ItemGroup>
+
+  <Target Name="GetFilesToPublish">
+    <ItemGroup>
+      <PackageToPublish Include="$(ArtifactsShippingPackagesDir)*.symbols.nupkg" IsSymbolsPackage="true" />
+      <PackageToPublish Include="$(ArtifactsShippingPackagesDir)*.nupkg" Exclude="@(PackageToPublish)" />
```
dougbu: Excluding `$(ArtifactsShippingPackagesDir)*.symbols.nupkg` would be more obvious.

build/Maestro/Maestro.csproj
```diff
@@ -0,0 +1,49 @@
...
+
+    <!--
+      'Internal' packages are used to transfer bits to partner teams, and should not be used by customers.
+      Publishing these can be disabled.
+    -->
+    <ItemGroup>
+      <PackageToPublish Include="$(ArtifactsNonShippingPackagesDir)*.symbols.nupkg" IsSymbolsPackage="true">
+        <ManifestArtifactData>NonShipping=true</ManifestArtifactData>
+      </PackageToPublish>
+      <PackageToPublish Include="$(ArtifactsNonShippingPackagesDir)*.nupkg" Exclude="@(PackageToPublish)">
```
dougbu: Not positive but won't this remove everything added above? In any case, excluding `$(ArtifactsNonShippingPackagesDir)*.symbols.nupkg` would be more obvious.

dougbu (changes requested):
Where was `$(PublishToAzureFeed)` set and shouldn't that be updated too?

build/Maestro/Maestro.csproj
dougbu: Oops, should have said "and switching the two lines".

natemcmaster: :up: :date:

This file was the result of deleting obsolete code from publish.csproj and renaming the file. Now that I'm looking at this, I see I can delete the `IsSymbolsPackage` attribute too. There should be no reason anymore to single out the `.symbols.nupkg` files.

dougbu (approved):
Will the created blobs still contain our `npm` packages and so on?

natemcmaster:
> Will the created blobs still contain our npm packages and so on?

Yes. That's part of "Public Azure Blob Storage" deployment phase in https://dev.azure.com/dnceng/internal/_release?definitionId=31&view=mine&_a=releases.

PR #7249, opened 2019-02-04
Remove misleading websocket error log

jkotalik:
Noticed in https://dev.azure.com/dnceng/public/_build/results?buildId=86501

### DIFF
--- a/src/Servers/IIS/AspNetCoreModuleV2/OutOfProcessRequestHandler/outprocessapplication.cpp
+++ b/src/Servers/IIS/AspNetCoreModuleV2/OutOfProcessRequestHandler/outprocessapplication.cpp
@@ -87,7 +87,7 @@ OUT_OF_PROCESS_APPLICATION::SetWebsocketStatus(
     PCWSTR pszTempWebsocketValue;
     DWORD cbLength;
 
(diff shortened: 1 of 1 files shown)


src/Servers/IIS/AspNetCoreModuleV2/OutOfProcessRequestHandler/outprocessapplication.cpp
```diff
@@ -87,7 +87,7 @@ OUT_OF_PROCESS_APPLICATION::SetWebsocketStatus(
     PCWSTR pszTempWebsocketValue;
     DWORD cbLength;
 
-    if (FAILED_LOG(pHttpContext->GetServerVariable("WEBSOCKET_VERSION", &pszTempWebsocketValue, &cbLength)))
+    if (pHttpContext->GetServerVariable("WEBSOCKET_VERSION", &pszTempWebsocketValue, &cbLength))
```
pakrym: This at least needs FAILED() to work correctly. But it's better to handle error code for not found result and  keep FAILED_LOG in other cases.

pakrym: It's `ERROR_INVALID_INDEX`

pakrym (approved):

PR #7252, opened 2019-02-04
Add support for suspending ANCM startup

pakrym:
Uses the `ASPNETCORE_STARTUP_SUSPEND_EVENT` environment variable to get an event name.

Fixes: https://github.com/aspnet/AspNetCore/issues/6972

### DIFF
--- a/src/Servers/IIS/AspNetCoreModuleV2/AspNetCore/applicationinfo.cpp
+++ b/src/Servers/IIS/AspNetCoreModuleV2/AspNetCore/applicationinfo.cpp
@@ -146,6 +146,13 @@ APPLICATION_INFO::TryCreateApplication(IHttpContext& pHttpContext, const ShimOpt
         }
         else
         {
+            auto const suspendedEventName = startupEvent.value() + L"_suspended";
+
+            HandleWrapper<NullHandleTraits> suspendedEventHandle = OpenEvent(EVENT_MODIFY_STATE, false, suspendedEventName.c_str());
+            if (suspendedEventHandle != nullptr)
+            {
+                LOG_LAST_ERROR_IF(!SetEvent(suspendedEventHandle));
+            }
             LOG_LAST_ERROR_IF(WaitForSingleObject(eventHandle, INFINITE) != WAIT_OBJECT_0);
         }
 
--- a/src/Servers/IIS/AspNetCoreModuleV2/AspNetCore/applicationinfo.cpp
+++ b/src/Servers/IIS/AspNetCoreModuleV2/AspNetCore/applicationinfo.cpp
@@ -133,6 +133,23 @@ APPLICATION_INFO::CreateApplication(IHttpContext& pHttpContext)
 HRESULT
 APPLICATION_INFO::TryCreateApplication(IHttpContext& pHttpContext, const ShimOptions& options)
 {
+    const auto startupEvent = Environment::GetEnvironmentVariableValue(L"ASPNETCORE_STARTUP_SUSPEND_EVENT");
+    if (startupEvent.has_value())
+    {
+        LOG_INFOF(L"Startup suspend event %ls", startupEvent.value().c_str());
+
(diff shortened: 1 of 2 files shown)


src/Servers/IIS/AspNetCoreModuleV2/AspNetCore/applicationinfo.cpp
```diff
@@ -133,6 +133,23 @@ APPLICATION_INFO::CreateApplication(IHttpContext& pHttpContext)
...
+
+        HandleWrapper<NullHandleTraits> eventHandle = OpenEvent(SYNCHRONIZE, false, startupEvent.value().c_str());
+
+        if (eventHandle == nullptr)
+        {
+            LOG_INFOF(L"Unable to open startup suspend event");
+        }
+        else
+        {
+            LOG_LAST_ERROR_IF(WaitForSingleObject(eventHandle, INFINITE) != WAIT_OBJECT_0);
```
jkotalik: Thoughts on having a timeout here? What are the downsides if this hangs forever?

pakrym: I don't think there are any downsides, that's what ANCM was literally asked to do.

src/Servers/IIS/IIS/test/Common.FunctionalTests/Inprocess/StartupTests.cs
```diff
@@ -520,6 +521,40 @@ public async Task SetCurrentDirectoryHandlerSettingWorks()
...
+            deploymentParameters.EnvironmentVariables["ASPNETCORE_STARTUP_SUSPEND_EVENT"] = "ANCM_TestEvent";
+
+            var eventWaitHandle = new  EventWaitHandle(false, EventResetMode.ManualReset, "ANCM_TestEvent");
+
+            var deploymentResult = await DeployAsync(deploymentParameters);
+
+            var request = deploymentResult.HttpClient.GetAsync("/HelloWorld");
+
+            // didn't figure out a better way to check that ANCM is waiting to start
+            var applicationDll = Path.Combine(deploymentResult.ContentRoot, "InProcessWebSite.dll");
```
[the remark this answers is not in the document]
jkotalik: I don't have anything better. This is what I did for checking if a file was in use, which is similar to this. https://github.com/aspnet/AspNetCore/blob/master/src/Servers/IIS/IIS/test/testassets/OutOfProcessWebSite/Startup.cs#L71

pakrym:
:up::date: improved the test.

src/Servers/IIS/AspNetCoreModuleV2/AspNetCore/applicationinfo.cpp
```diff
@@ -133,6 +133,30 @@ APPLICATION_INFO::CreateApplication(IHttpContext& pHttpContext)
...
+
+        if (eventHandle == nullptr)
+        {
+            LOG_INFOF(L"Unable to open startup suspend event");
+        }
+        else
+        {
+            auto const suspendedEventName = startupEvent.value() + L"_suspended";
+
+            HandleWrapper<NullHandleTraits> suspendedEventHandle = OpenEvent(EVENT_MODIFY_STATE, false, suspendedEventName.c_str());
```
jkotalik: Is this event only used for testing?

jkotalik: Feels like we shouldn't open an event if it purely used for testing.

pakrym: Why not? Makes for better testing, caught a test bug on IIS. And I don't think it hurts anything.

jkotalik (approved):

BillHiebert:
Note that the way it currently works is that there are two events. One ANCM waits on before processing the request, and another ANCM signals to indicate it is going to process a request.

pakrym:
Why is the second event required? Isn't request processing going to happen right after the first event is signalled?

PR #7259, opened 2019-02-04
Re-enable h2spec tests with more diagnostics

Tratcher:
https://github.com/aspnet/AspNetCore-internal/issues/1720 These tests were timing out or hanging. I've added more diagnostics to see if we can figure out why. I won't merge yet, I'll run them in a loop on this PR until we can get more info.

### DIFF
--- a/src/Servers/Kestrel/test/Interop.FunctionalTests/H2SpecCommands.cs
+++ b/src/Servers/Kestrel/test/Interop.FunctionalTests/H2SpecCommands.cs
@@ -10,6 +10,7 @@ using System.Runtime.InteropServices;
 using System.Threading.Tasks;
 using System.Xml;
 using Microsoft.Extensions.Logging;
+using Xunit;
 
 namespace Interop.FunctionalTests
 {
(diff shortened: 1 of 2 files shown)


Tratcher:
Hmm, it's not hanging or timing out anymore. I don't see how my changes would have fixed it. Might as well review while I run it a few more times.

src/Servers/Kestrel/test/Interop.FunctionalTests/H2SpecCommands.cs
```diff
@@ -169,31 +170,51 @@ private static bool IsTestLine(string line, out string testNumber, out string de
...
 
-            using (var process = Process.Start(processOptions))
-            {
-                var dataTask = process.StandardOutput.ReadToEndAsync();
+                Assert.True(process.Start());
+                process.BeginOutputReadLine(); // Starts OutputDataReceived
+                process.BeginErrorReadLine(); // Starts ErrorDataReceived
 
-                if (await Task.WhenAny(dataTask, Task.Delay(TimeSpan.FromSeconds(TimeoutSeconds * 2))) != dataTask)
+                if (!process.WaitForExit(1000 * TimeoutSeconds))
```
halter73: It's probably better to use the Process.Exited event and a TCS instead of blocking using Process.WaitForExit().

halter73 (approved):
Let's see what happens.

PR #7276, opened 2019-02-05
Blazor 0.8.0 fixes

SteveSandersonMS:
This PR contains three fixes that are about to be shipped in Blazor 0.8.0. They are already in `stevesa/release-blazor-0.8.0`, but they need to go in `master` too, because these ones are meant to be permanent.

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -144,7 +144,7 @@
     <SystemIdentityModelTokensJwtPackageVersion>5.3.0</SystemIdentityModelTokensJwtPackageVersion>
     <WindowsAzureStoragePackageVersion>8.1.4</WindowsAzureStoragePackageVersion>
     <!-- Dependencies for Blazor. -->
-    <MicrosoftAspNetCoreBlazorMonoPackageVersion>0.8.0-preview-20190125.1</MicrosoftAspNetCoreBlazorMonoPackageVersion>
+    <MicrosoftAspNetCoreBlazorMonoPackageVersion>0.8.0-preview-20190204.1</MicrosoftAspNetCoreBlazorMonoPackageVersion>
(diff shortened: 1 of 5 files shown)


eng/Versions.props
```diff
@@ -144,7 +144,7 @@
     <SystemIdentityModelTokensJwtPackageVersion>5.3.0</SystemIdentityModelTokensJwtPackageVersion>
     <WindowsAzureStoragePackageVersion>8.1.4</WindowsAzureStoragePackageVersion>
     <!-- Dependencies for Blazor. -->
-    <MicrosoftAspNetCoreBlazorMonoPackageVersion>0.8.0-preview-20190125.1</MicrosoftAspNetCoreBlazorMonoPackageVersion>
+    <MicrosoftAspNetCoreBlazorMonoPackageVersion>0.8.0-preview-20190204.1</MicrosoftAspNetCoreBlazorMonoPackageVersion>
```
pranavkm: Out of curiosity, where does this package get built? We should figure out a way to move it to some place we can get automated signing

pranavkm (approved):

eng/Versions.props
SteveSandersonMS: The sources are in https://github.com/aspnet/blazor, and it gets built in the blazor-ci official build pipeline. It already gets signed there.

natemcmaster (approved):

natemcmaster:
>  LINK : fatal error LNK1104: cannot open file 'F:\vsagent\2\s\src\Installers\Windows\AspNetCoreModule-Setup\IIS-Setup\iisca\lib\Release\iisca.lib' 

I think this was resolved by a changed Pavel made early today. You might need to merge master into this PR to get the fix.

PR #7278, opened 2019-02-05
Add a TypeForward for JsonResult

pranavkm:
Fixes https://github.com/aspnet/AspNetCore/issues/7220


natemcmaster (approved):

davidfowl:
Do you need to do it this way? Would TypeForwardFrom work? Or nah?

pranavkm:
I thought `TypeForwardedFrom` was informational only and not used for type resolution.

davidfowl:
Ah I see.

davidfowl:
@pranavkm how did you test this? Should we reference the 2.1 package in one of our tests?

pranavkm:
I added a reference to the Swashbuckle in the MvcSandbox and saw it fail as soon as I referenced `JsonResult`. Using the helper method on `Controller` worked fine. Once the type forward was in place, the error was resolved.

> Should we reference the 2.1 package in one of our tests?

Would help to test it against a real app (something that references the framework, like MusicStore), but it's really unlikely this is a scenario we would regress.

davidfowl (approved):

Issue #5855, opened 2014-12-01
Expose HttpRequestInfoTypeSslProtocol


Tratcher:
HTTP_REQUEST_V2.pRequestInfo points to an array of HTTP_REQUEST_INFO in the request. One of the new items is HttpRequestInfoTypeSslProtocol, where the pInfo field is a HTTP_SSL_PROTOCOL_INFO structure as outlined below. We haven't found the constant definitions yet to interpret the data though...

``` C
typedef struct _HTTP_SSL_PROTOCOL_INFO  
{  
    ULONG Protocol;  
    ULONG CipherType;  
    ULONG CipherStrength;  
    ULONG HashType;  
    ULONG HashStrength;  
    ULONG KeyExchangeType;  
    ULONG KeyExchangeStrength;
}
```

PR #7284, opened 2019-02-05
Implement ITlsHandshakeFeature for HttpSys

Tratcher:
#5855 Implement ITlsHandshakeFeature for HttpSys like we already did for Kestrel. Had to pull in the shared Connections dependency to get it. It's all easy mapping except for SslProtocols.

I also found a way to run the HTTPS tests by imitating IIS Express's appid and re-using its cert bindings.

@shirhatti FYI


davidfowl:
Why is this important for preview3? That issue was filed in 2014 and it wasn't even triaged for 3.0. What's the deal?

Tratcher:
Yes @shirhatti triaged it into 3.0. 

It looked like a good snow day project I could finish.

src/Servers/HttpSys/src/RequestProcessing/Request.cs
```diff
@@ -232,6 +238,59 @@ private AspNetCore.HttpSys.Internal.SocketAddress LocalEndPoint
...
+        private void GetTlsHandshakeResults()
+        {
+            var handshake = _nativeRequestContext.GetTlsHandshake();
+
+            Protocol = handshake.Protocol;
+            // https://docs.microsoft.com/en-us/windows/desktop/api/schannel/ns-schannel-_secpkgcontext_connectioninfo
+            // restore client/server bits so the result maps exaclty on SslProtocols constants
+            // Compare to https://referencesource.microsoft.com/#System/net/System/Net/SecureProtocols/_SslState.cs,8905d1bf17729de3
+#pragma warning disable CS0618 // Type or member is obsolete
+            if ((Protocol & SslProtocols.Ssl2) != 0)
```
jkotalik: From what I can tell this is what is doing.
Ssl2 == 12, which is 1100 in binary. If the Protocol either has the 3rd or 4th bit set, set both bits. 

From refsource, it says
```
// restore client/server bits so the result maps exaclty on published constants
```

so for each ssl state, there is both a client and server set bit? 

No matter what this deserves a comment as a majority of the time, enums are not multibyte.

src/Servers/HttpSys/src/RequestProcessing/Request.cs
```diff
@@ -232,6 +238,59 @@ private AspNetCore.HttpSys.Internal.SocketAddress LocalEndPoint
...
+
+        public int HashStrength { get; private set; }
+
+        public ExchangeAlgorithmType KeyExchangeAlgorithm { get; private set; }
+
+        public int KeyExchangeStrength { get; private set; }
+
+        private void GetTlsHandshakeResults()
+        {
+            var handshake = _nativeRequestContext.GetTlsHandshake();
```
jkotalik: Can the handshake be null here? Probably a good check to have anyways event if it can't be null.

Tratcher:
@jkotalik yeah we'll deal with IIS separately. Aren't the IIS values already available via server variables?

src/Servers/HttpSys/test/FunctionalTests/Utilities.cs
```diff
@@ -148,17 +151,38 @@ internal static IServer CreateDynamicHttpServer(string basePath, out string root
...
+                    var server = CreatePump();
+                    server.Features.Get<IServerAddressesFeature>().Addresses.Add(baseAddress);
+                    configureOptions(server.Listener.Options);
+                    try
+                    {
+                        server.StartAsync(new DummyApplication(app), CancellationToken.None).Wait();
+                        return server;
+                    }
+                    catch (HttpSysException)
+                    {
```
jkotalik: Log informational here?

src/Servers/HttpSys/test/FunctionalTests/Utilities.cs
```diff
@@ -148,17 +151,38 @@ internal static IServer CreateDynamicHttpServer(string basePath, out string root
...
         {
-            var server = CreatePump();
-            server.Features.Get<IServerAddressesFeature>().Addresses.Add(UrlPrefix.Create(scheme, host, port, path).ToString());
-            server.StartAsync(new DummyApplication(app), CancellationToken.None).Wait();
-            return server;
+            lock (PortLock)
+            {
+                while (NextHttpsPort < MaxHttpsPort)
+                {
+
```
jkotalik: nit newline.

src/Servers/HttpSys/src/RequestProcessing/Request.cs
Tratcher: The OS considers client and server TLS as different enum values (shown in the connectioninfo link above). .NET SslProtocols choose to combine those for some reason. I'll clarify the comment.

src/Servers/HttpSys/src/RequestProcessing/Request.cs
Tratcher: No, it's a struct

src/Servers/HttpSys/test/FunctionalTests/Utilities.cs
Tratcher: No, we're expecting port conflict errors, it's Debug at best. These are also not Logged tests, there's nowhere to log to.

Tratcher:
Updated

src/Servers/HttpSys/src/RequestProcessing/Request.cs
```diff
@@ -232,6 +238,60 @@ private AspNetCore.HttpSys.Internal.SocketAddress LocalEndPoint
...
+            {
+                Protocol |= SslProtocols.Tls;
+            }
+            if ((Protocol & SslProtocols.Tls11) != 0)
+            {
+                Protocol |= SslProtocols.Tls11;
+            }
+            if ((Protocol & SslProtocols.Tls12) != 0)
+            {
+                Protocol |= SslProtocols.Tls12;
```
Tratcher: @shirhatti when is Tls 1.3 coming in?

jkotalik (approved):

PR #7290, opened 2019-02-05
Consume Mono.WebAssembly.Interop from Extensions

pranavkm:


eng/Version.Details.xml
```diff
@@ -281,6 +281,10 @@
       <Uri>https://github.com/dotnet/corefx</Uri>
       <Sha>9e074e4a3f3626fa953f36ab79e3cd3e6db1c9de</Sha>
     </Dependency>
+    <Dependency Name="Mono.WebAssembly.Interop" Version="3.0.0-preview.19105.3">
```
pranavkm: I still don't know how to darc.

eng/Versions.props
```diff
@@ -93,6 +93,9 @@
     <MicrosoftExtensionsValueStopwatchSourcesPackageVersion>3.0.0-preview.19105.4</MicrosoftExtensionsValueStopwatchSourcesPackageVersion>
     <MicrosoftExtensionsWebEncodersPackageVersion>3.0.0-preview.19105.4</MicrosoftExtensionsWebEncodersPackageVersion>
     <MicrosoftJSInteropPackageVersion>3.0.0-preview.19105.4</MicrosoftJSInteropPackageVersion>
+    <!-- When updating this, ensure you also update src/Components/Browser.JS/src/package.json to reference the corresponding version of @dotnet/jsinterop -->
```
natemcmaster: The bot doesn't know how to do this.

eng/Version.Details.xml
natemcmaster: Looks right.

eng/Versions.props
pranavkm: That's fine. I think we've been ignoring the package.json so far. cc @SteveSandersonMS if this is super important for preview4

SteveSandersonMS: The original reason for the comment is that, unless we consume the equivalent versions of the NuGet and NPM packages, there's no guarantee that the .NET and JS sides are going to understand each other at runtime. As such, it's not really correct to update the NuGet side automatically if we can't also update the NPM side automatically.

However, in practice this is a very slow-moving dependency. I don't think the NPM side has changed at all since we first took the dependency, and the .NET side has only changed due to build system changes (not due to runtime feature changes).

So, I think we're generally going to be OK with letting the NuGet dependency update automatically and only updating the NPM dependency manually if we have reason to believe it's needed.

I think **the most important thing is that we get Components E2E tests running in Azure Pipelines.** Once we have that, we get a very strong reassurance that the .NET and JS sides of interop are working together properly. Of course, this is a separate piece of work from this PR which I have on my backlog.

SteveSandersonMS (approved):
Thanks, @pranavkm!

PR #7292, opened 2019-02-05
Use same code for Json Platform as TE

benaadams:
Serialize struct rather than anonymous class

/cc @sebastienros


sebastienros:
Make sense that we follow what we run on TE. Only concern is that it changes what we continuously measure, so any difference should be ignored when we look at the charts.

sebastienros (approved):

benaadams:
TE is different as it also has the database scenarios etc; but this at least means its executing the same method when it gets there.

Had a PR into old Kestrel that brought them completely into line, but there was some debate about where the code should live, so it never happened.

anurse:
@benaadams I'm doing a little PR housekeeping. This PR looks OK and approved but has a conflict. You up for resolving that?

Issue #7280, opened 2019-02-05
Convert to Arcade SDK


natemcmaster:
This repo is the last hold-out on using KoreBuild. We should convert it to use the Arcade SDK.

**Benefits**

* In the long run, reduces cost of maintaining build tools.
* Makes our repo look like the others which reduces friction in moving between repo
* Hopefully this will lead to improvements in Arcade which other teams will benefit from

**Challenges**

I've peered into my crystal ball 🔮 and have identified some differences that the implementer should be aware of:

* [x] Arcade does not support a concept of "repo level targets". See build/repo.targets for what we currently do in our build which needs to find an new, unspecified home in the Arcade world.
* [x] Arcade slightly changes the order in which packaging and tests run. Some of our tests, like template tests, depend on the existence of the .nupkg files
* [x] Arcade has different conventions for detecting test projects. See https://github.com/dotnet/arcade/issues/1646
* [ ] Arcade does not have support doing a 'dry run' of benchmarks projects during the 'Test' phase. See https://github.com/aspnet/BuildTools/blob/803fd176f1ae147aa8af08e49e76ffda7221a4cd/files/KoreBuild/modules/benchmarks/module.targets#L14-L37 for the korebuild implementation
* Arcade has many defaults which do not align with what aspnet does. We will need to carefully vet build outputs to make sure our product packages, binaries, installers, etc. do not substantially change as result of ripping our KoreBuild. (cref https://github.com/dotnet/arcade/issues/1912). Some particular areas of concern where Arcade's defaults may be different:
  * [x] Symbols and symbols packages
  * [x] Strong name keys
  * [x] Authenticode certs
  * [x] Localization tooling and resx code generation - tracked in #11054 
  * [x] Xunit dependency versions
  * [x] .NET Framework reference assemblies - global install vs package (affects MVC and runtime compilation)
  * [x] The location of `obj/` and `bin/` directories
  * [x] Settings for projects which produce a VSIX
* [x] Custom project types not yet supported in KoreBuild. This repo has project files which are not Microsoft.NET.Sdk projects. E.g. .npmproj, .javaproj, .zipproj, and others. Arcade does not support these yet
* Toolset acquisition
  * [x] Arcade will put tools, such as dotnet.exe, in a different locations.
[...]

PR #7005, opened 2019-01-25
Make Visual Studio 2019 a prerequisite to building this repo

natemcmaster:
We're preparing to consume some changes soon which will require Visual Studio 2019 (C# 8, updated .NET Core 3 SDKs, etc.)

Changes:

- Updated documentation in aspnet/AspNetCore
- Updated InstallVisualStudio.ps1 scripts
- Updated .sln files to require VS >= 16
- Updated CI config to use agents with VS 2019 installed

PR #7293, opened 2019-02-05
Install the .NET Core SDK into the repo root instead of UserProfile and 'install' copy of AspNetCore shared framework

natemcmaster:
Changes:
* This moves the default location of the .NET Core SDK installation into `$repoRoot/.dotnet`. This location was already in use for CI builds. 
* Update the build step for Microsoft.AspNetCore.App to install the shared framework into the local copy of the .NET Core SDK

This is required to workaround several limitations in the way the .NET Core SDK works and allow our tests to use shared frameworks and targeting packs.

* https://github.com/dotnet/core-setup/issues/4809
* https://github.com/dotnet/sdk/issues/2910

 
## Recommendations

- Use the “startvs.cmd” script to launch Visual Studio. This will set required environment variables to make VS happier than if you just double click the .sln file.

- Use “activate.sh/ps1” if you want to run “dotnet build”, “dotnet test” and others. These will set required environment variables, including PATH.
 - I recommend removing %USERPROFILE%/.dotnet to your PATH variable if you had added it manually before. This will no longer match what build tools will install.
 
## FYI
This also deprecates support for the DOTNET_HOME environment variable. This was a variable our team added to build tools years ago, but it was never officially adopted by the .NET Core SDK.
 
## Pro-tip
`git clean -xfd -e .dotnet/` preserves the folder so you don’t have to re-download the SDK again.

### DIFF
--- a/activate.ps1
+++ b/activate.ps1
@@ -1,0 +1,57 @@
+#
+# This file must be used by invoking ". .\activate.ps1" from the command line.
+# You cannot run it directly.
+# To exit from the environment this creates, execute the 'deactivate' function.
+#
+
+function deactivate ([switch]$init) {
+
+    # reset old environment variables
+    if (Test-Path variable:_OLD_PATH) {
+        $env:PATH = $_OLD_PATH
+        Remove-Item variable:_OLD_PATH
+    }
+
+    if (test-path function:_old_prompt) {
+        Set-Item Function:prompt -Value $function:_old_prompt -ea ignore
+        remove-item function:_old_prompt
+    }
+
+    Remove-Item env:DOTNET_ROOT -ea ignore
+    Remove-Item env:DOTNET_MULTILEVEL_LOOKUP -ea ignore
+    if (-not $init) {
+        # Remove the deactivate function
+        Remove-Item function:deactivate
+    }
+}
+
+# Cleanup the environment
+deactivate -init
+
+$_OLD_PATH = $env:PATH
+# Tell dotnet where to find itself
+$env:DOTNET_ROOT = "$PSScriptRoot\.dotnet\x64"
+# Tell dotnet not to look beyond the DOTNET_ROOT folder for more dotnet things
+$env:DOTNET_MULTILEVEL_LOOKUP = 0
+# Put dotnet first on PATH
+$env:PATH = "${env:DOTNET_ROOT};${env:PATH}"
+
+# Set the shell prompt
+if (-not $env:DISABLE_CUSTOM_PROMPT) {
+    $function:_old_prompt = $function:prompt
+    function dotnet_prompt {
+        # Add a prefix to the current prompt, but don't discard it.
+        write-host "($( split-path $PSScriptRoot -leaf )) " -nonewline
+        & $function:_old_prompt
+    }
+
+    Set-Item Function:prompt -Value $function:dotnet_prompt -ea ignore
+}
+
+Write-Host -f Magenta "Enabled the .NET Core environment. Execute 'deactivate' to exit."
+if (-not (Test-Path "${env:DOTNET_ROOT}\dotnet.exe")) {
+    Write-Host -f Yellow ".NET Core has not been installed yet. Run $PSScriptRoot\restore.cmd to install it."
+}
+else {
+    Write-Host "dotnet = ${env:DOTNET_ROOT}\dotnet.exe"
+}
--- a/build.ps1
+++ b/build.ps1
@@ -234,12 +234,7 @@ if (Test-Path $ConfigFile) {
     }
 }
 
-$DotNetHome = if ($env:DOTNET_HOME) { $env:DOTNET_HOME } `
-    elseif ($CI) { Join-Path $PSScriptRoot '.dotnet' } `
-    elseif ($env:USERPROFILE) { Join-Path $env:USERPROFILE '.dotnet'} `
-    elseif ($env:HOME) {Join-Path $env:HOME '.dotnet'}`
-    else { Join-Path $PSScriptRoot '.dotnet'}
-
+$DotNetHome = Join-Path $PSScriptRoot '.dotnet'
 $env:DOTNET_HOME = $DotNetHome
 
 # Execute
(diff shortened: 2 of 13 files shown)


src/ProjectTemplates/test/ProjectTemplates.Tests.csproj
```diff
@@ -4,9 +4,6 @@
     <DefineConstants>$(DefineConstants);XPLAT</DefineConstants>
     <!-- WebDriver is not strong-named, so this test project cannot be strong named either. -->
     <SignAssembly>false</SignAssembly>
-
-    <!-- Workaround until https://github.com/aspnet/AspNetCore/issues/4321 is resolved. -->
-    <SkipTests Condition="'$(RunTemplateTests)' != 'true'">true</SkipTests>
```
natemcmaster: @ryanbrandenburg I added to this PR. It passes on all but one test for me. If I can't figure it out, I'm going to revert this so tests are still skipped in templating.

```
 Failed   Templates.Test.RazorComponentsTemplateTest.RazorComponentsTemplateWorks

error : NETSDK1061: The project was restored using Microsoft.AspNetCore.App version 3.0.0-preview-18572-0018, but with current settings, version 3.0.0-alpha1-
  10663 would be used instead. To resolve this issue, make sure the same settings are used for restore and for subsequent operations such as build or publish. Typically thi
  s issue can occur if the RuntimeIdentifier property is set during build or publish but not during restore. For more information, see https://aka.ms/dotnet-runtime-patch-s
  election.

```

ryanbrandenburg (approved):
One suggestion/question, but nothing we should hold the PR for, let's light this candle!

ryanbrandenburg:
If you just want to keep the tests skipped I'll fix Templating separately. We do some hacks in those tests which make it likely that our hacks need to change rather than there actually being something wrong with this PR.

activate.ps1
```diff
@@ -0,0 +1,57 @@
...
+        Remove-Item function:deactivate
+    }
+}
+
+# Cleanup the environment
+deactivate -init
+
+$_OLD_PATH = $env:PATH
+# Tell dotnet where to find itself
+$env:DOTNET_ROOT = "$PSScriptRoot\.dotnet\x64"
```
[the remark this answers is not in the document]
natemcmaster: I guess we could restore all env vars we set, but I didn't really have that scenario in mind when I wrote it. Do you set this variable often?

src/ProjectTemplates/test/ProjectTemplates.Tests.csproj
natemcmaster: Looks like there was some workaround code in the test that was unnecessary. I've removed it and the test passed locally.

activate.ps1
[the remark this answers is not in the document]
ryanbrandenburg: I have it set on both my laptop and deskptop, but now that you mention it I couldn't really tell you why. Never mind then.

natemcmaster:
@ryanbrandenburg It looks like some of the templates tests are still broken on CI, but only windows. They error is the same - something about selenium. I'm going to re-skip them on Windows. We can use https://github.com/aspnet/AspNetCore/issues/4321 to follow up on this.

activate.ps1
```diff
@@ -0,0 +1,57 @@
...
+        $env:PATH = $_OLD_PATH
+        Remove-Item variable:_OLD_PATH
+    }
+
+    if (test-path function:_old_prompt) {
+        Set-Item Function:prompt -Value $function:_old_prompt -ea ignore
+        remove-item function:_old_prompt
+    }
+
+    Remove-Item env:DOTNET_ROOT -ea ignore
```
dougbu: Does everyone have PowerShell 3.0 or later? `ignore` wasn't supported in earlier releases and 3.0 was only an add-on for Windows before Windows 8 and Windows Server 2012.

build.ps1
```diff
@@ -234,12 +234,7 @@ if (Test-Path $ConfigFile) {
...
 }
 
-$DotNetHome = if ($env:DOTNET_HOME) { $env:DOTNET_HOME } `
-    elseif ($CI) { Join-Path $PSScriptRoot '.dotnet' } `
-    elseif ($env:USERPROFILE) { Join-Path $env:USERPROFILE '.dotnet'} `
-    elseif ($env:HOME) {Join-Path $env:HOME '.dotnet'}`
-    else { Join-Path $PSScriptRoot '.dotnet'}
-
+$DotNetHome = Join-Path $PSScriptRoot '.dotnet'
 $env:DOTNET_HOME = $DotNetHome
```
dougbu: Why do we still need both `$env:DOTNET_ROOT` and `$env:DOTNET_HOME`?  I realize they have different values on Windows but don't `dotnet` and our build scripts do the **Right Thing:tm:** with one of these?

Put another way, could we be consistent and use only one of these?

dougbu (approved):

ryanbrandenburg:
Failures is already tracked [here](https://github.com/aspnet/AspNetCore/issues/6597). Retrying.

build.ps1
natemcmaster: DOTNET_HOME is a thing our team invented years ago (like 1.0 days), but the SDK team decided not to adopt. We'll get rid of it in #7280. KoreBuild uses this env var to pick the location of where SDKs and build tools are installed.

There is a difference on Windows. Because we support x64 and x86 SDKs, DOTNET_HOME=$repoRoot/.dotnet, DOTNET_ROOT=$DOTNET_HOME/$ARCH

activate.ps1
natemcmaster: We require PowerShell 5 or newer to build this repo, so I'm not too worried about this: https://github.com/aspnet/AspNetCore/blob/94a4293cce735b3756661ef5a475c4944f57d64d/build.ps1#L1

Also, with #7005, we're also going to require Windows 10.

natemcmaster:
Ok, mail send to the team. I've updated this PR description with the same information for external contributors looking at this change.

PR #7309, opened 2019-02-06
Use Win32 as native platform

pakrym:
Try to address https://github.com/aspnet/AspNetCore-Internal/issues/1768

We fixed similar issues in master by preventing double building caused by using x86 as a native project platform.


jkotalik (approved):
If it works!

pakrym:
Build succeeded and tests ran past IIS tests before getting cancelled because of the timeout.

PR #7334, opened 2019-02-07
 Remove ANCMV1 from everywhere

pakrym:


Tratcher (approved):

jkotalik (approved):

jkotalik:
Can you also remove ANCM here? https://github.com/aspnet/AspNetCore/blob/a8f1b034d8b8c4c0fc341f2a4e39fa54af28294c/src/Servers/IIS/IIS/test/IIS.BackwardsCompatibility.FunctionalTests/IIS.BackwardsCompatibility.FunctionalTests.csproj#L27

pakrym:
Good point.

pakrym:
@Tratcher FYI, this is a much larger change now.

Tratcher:
Was there any plan to keep back compat tests? Or do we rely on the patch branches for that?

pakrym:
Back compat of what with what?

pakrym:
Talked offline, back compat in @Tratcher question is unrelated to this.

PR #7337, opened 2019-02-07
Return Completed ReadResult after stream returns 0 bytes

jkotalik:
For https://github.com/aspnet/AspNetCore/issues/7329. This is a combination of 2 bugs:
1. StreamPipeReader.ReadAsync shouldn't call Stream.ReadAsync if the stream previously return 0 bytes. This PR solves this issue. 
Note: most implementation of Stream continue to return 0 bytes from ReadAsync after the stream is completed.
2. ResponseStream.ReadAsync should probably not complete the PipeReader https://github.com/aspnet/AspNetCore/blob/master/src/Hosting/TestHost/src/ResponseStream.cs#L118. We don't do this in ReadOnlyPipeStream today. This is a problem because calling ReadAsync after it return 0 bytes will throw an exception rather than returning 0 bytes.

### DIFF
--- a/src/Http/Http/src/StreamPipeReader.cs
+++ b/src/Http/Http/src/StreamPipeReader.cs
@@ -208,6 +208,11 @@ namespace System.IO.Pipelines
         public override async ValueTask<ReadResult> ReadAsync(CancellationToken cancellationToken = default)
         {
             // TODO ReadyAsync needs to throw if there are overlapping reads.
+            if (_isWriterCompleted)
+            {
+                return new ReadResult(buffer: default, isCanceled: false, isCompleted: true);
+            }
+
             ThrowIfCompleted();
 
             // PERF: store InternalTokenSource locally to avoid querying it twice (which acquires a lock)
--- a/src/Http/Http/test/StreamPipeReaderTests.cs
+++ b/src/Http/Http/test/StreamPipeReaderTests.cs
@@ -431,7 +431,7 @@ namespace System.IO.Pipelines.Tests
         [Fact]
         public async Task AsyncReadWorks()
         {
-            MemoryStream = new AsyncStream();
+            Stream = new AsyncStream();
             CreateReader();
 
             WriteByteArray(2000);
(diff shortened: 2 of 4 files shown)


src/Http/Http/src/StreamPipeReader.cs
```diff
@@ -208,6 +208,11 @@ public override void OnWriterCompleted(Action<Exception, object> callback, objec
         public override async ValueTask<ReadResult> ReadAsync(CancellationToken cancellationToken = default)
         {
             // TODO ReadyAsync needs to throw if there are overlapping reads.
+            if (_isWriterCompleted)
```
davidfowl: This bool name is wrong though. Is _readingCompleted.

src/Http/Http/test/StreamPipeReaderTests.cs
```diff
@@ -661,5 +672,28 @@ public override async ValueTask<int> ReadAsync(Memory<byte> buffer, Cancellation
             }
 #endif
         }
+
+        private class ThrowAfterZeroByteReadStream : MemoryStream
+        {
+            private bool throwOnNextCallToRead;
```
JamesNK: nit: underscore

JamesNK (approved):

davidfowl (approved):

src/Http/Http/test/StreamPipeReaderTests.cs
```diff
@@ -599,6 +599,17 @@ public async Task ReadAsyncWithEmptyDataCompletesStream()
...
         }
 
+        [Fact]
+        public async Task ReadAsyncAfterReceivingCompletedReadResultDoesNotThrow()
+        {
+            Stream = new ThrowAfterZeroByteReadStream();
+            Reader = new StreamPipeReader(Stream);
+            var readResult = await Reader.ReadAsync();
+
+            readResult = await Reader.ReadAsync();
```
Tratcher: Assert zero bytes read.

Tratcher (approved):

davidfowl (approved):

davidfowl:
@jkotalik played with things a bit and found this https://github.com/aspnet/AspNetCore/issues/7448

PR #7338, opened 2019-02-07
Use the 'correct' windows platform version

BrennanConroy:
Fixes https://github.com/aspnet/AspNetCore-Internal/issues/1780

### DIFF
--- a/src/SignalR/clients/cpp/src/signalrclient/Build/VS/signalrclient.vcxproj
+++ b/src/SignalR/clients/cpp/src/signalrclient/Build/VS/signalrclient.vcxproj
@@ -12,7 +12,7 @@
     <OutDir>$(OutDir)lib\</OutDir>
     <IntDir>$(Configuration)\lib\</IntDir>
     <ConfigurationType>StaticLibrary</ConfigurationType>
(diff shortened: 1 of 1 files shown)


jkotalik (approved):

src/SignalR/clients/cpp/src/signalrclient/Build/VS/signalrclient.vcxproj
```diff
@@ -12,7 +12,7 @@
     <OutDir>$(OutDir)lib\</OutDir>
     <IntDir>$(Configuration)\lib\</IntDir>
     <ConfigurationType>StaticLibrary</ConfigurationType>
-    <WindowsTargetPlatformVersion>10.0.17763.0</WindowsTargetPlatformVersion>
+    <WindowsTargetPlatformVersion>10.0.17134.0</WindowsTargetPlatformVersion>
```
Eilon: Yet another "off by 0.0.629.0" error... so typical!

BrennanConroy: There are two problems in programming...

PR #7339, opened 2019-02-07
 Unset reserved bit when setting 31 bit uint values

JunTaoLuo:
Addresses https://github.com/aspnet/AspNetCore/issues/6974. cc @JamesNK

### DIFF
--- a/src/Servers/Kestrel/Core/src/Internal/Http2/Http2FrameWriter.cs
+++ b/src/Servers/Kestrel/Core/src/Internal/Http2/Http2FrameWriter.cs
@@ -402,7 +402,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
                 _outgoingFrame.PrepareWindowUpdate(streamId, sizeIncrement);
                 WriteHeaderUnsynchronized();
                 var buffer = _outputWriter.GetSpan(4);
-                Bitshifter.WriteUInt31BigEndian(buffer, (uint)sizeIncrement);
+                BinaryPrimitives.WriteUInt32BigEndian(buffer, (uint)sizeIncrement);
                 _outputWriter.Advance(4);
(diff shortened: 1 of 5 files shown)


src/Servers/Kestrel/Core/src/Internal/Http2/Http2FrameWriter.cs
```diff
@@ -402,7 +402,7 @@ public Task WriteWindowUpdateAsync(int streamId, int sizeIncrement)
                 _outgoingFrame.PrepareWindowUpdate(streamId, sizeIncrement);
                 WriteHeaderUnsynchronized();
                 var buffer = _outputWriter.GetSpan(4);
-                Bitshifter.WriteUInt31BigEndian(buffer, (uint)sizeIncrement);
+                BinaryPrimitives.WriteUInt32BigEndian(buffer, (uint)sizeIncrement);
```
Tratcher: If you change this you need to ensure sizeIncrement is still only 31 bits so the reserve bit never gets set accidentally.

JamesNK: Might as well still have WriteUInt31BigEndian helper method that checks the value fits in 31 bits and calls WriteUInt32BigEndian internally.

Tratcher: Add a bool for preserveHighBit? We wouldn't preserve it, but it would clarify the intent.

JunTaoLuo: There are cases where this would need to be preserved, like the exclusive bit in priority frame. Though I'm not sure if we'll be implementing that. I'll add an overload to specify preservation.

Tratcher (approved):

PR #7340, opened 2019-02-07
Make server timeout configurable

mikaelm12:
Issue: https://github.com/aspnet/AspNetCore/issues/6797

Making the server side disconnect timeout configurable.

### DIFF
--- a/src/SignalR/common/Http.Connections/src/ConnectionsDependencyInjectionExtensions.cs
+++ b/src/SignalR/common/Http.Connections/src/ConnectionsDependencyInjectionExtensions.cs
@@ -23,6 +23,7 @@ namespace Microsoft.Extensions.DependencyInjection
         {
             services.AddRouting();
             services.AddAuthorizationPolicyEvaluator();
+            services.TryAddEnumerable(ServiceDescriptor.Singleton<IConfigureOptions<ConnectionOptions>, ConnectionOptionsSetup>());
             services.TryAddSingleton<HttpConnectionDispatcher>();
             services.TryAddSingleton<HttpConnectionManager>();
             return services;
@@ -36,13 +37,8 @@ namespace Microsoft.Extensions.DependencyInjection
         /// <returns>The same instance of the <see cref="IServiceCollection"/> for chaining.</returns>
         public static IServiceCollection AddConnections(this IServiceCollection services, Action<ConnectionOptions> options)
         {
-            services.AddRouting();
-            services.AddAuthorizationPolicyEvaluator();
-            services.TryAddSingleton<IConfigureOptions<ConnectionOptions>, ConnectionOptionsSetup>();
-            services.TryAddSingleton<HttpConnectionDispatcher>();
-            services.TryAddSingleton<HttpConnectionManager>();
-            services.Configure(options);
-            return services;
+            return services.Configure(options)
+                .AddConnections();
         }
     }
 }
--- a/src/SignalR/common/Http.Connections/src/ConnectionsDependencyInjectionExtensions.cs
+++ b/src/SignalR/common/Http.Connections/src/ConnectionsDependencyInjectionExtensions.cs
@@ -41,6 +41,7 @@ namespace Microsoft.Extensions.DependencyInjection
             services.TryAddSingleton<IConfigureOptions<ConnectionOptions>, ConnectionOptionsSetup>();
             services.TryAddSingleton<HttpConnectionDispatcher>();
             services.TryAddSingleton<HttpConnectionManager>();
+            services.Configure(options);
             return services;
         }
     }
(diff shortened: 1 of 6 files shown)


src/SignalR/common/Http.Connections/src/Internal/HttpConnectionManager.cs
```diff
@@ -31,12 +32,14 @@ public partial class HttpConnectionManager
         private readonly ILogger<HttpConnectionManager> _logger;
         private readonly ILogger<HttpConnectionContext> _connectionLogger;
 
-        public HttpConnectionManager(ILoggerFactory loggerFactory, IApplicationLifetime appLifetime)
+        public TimeSpan DisconnectTimeout { get; set; }
+
+        public HttpConnectionManager(ILoggerFactory loggerFactory, IApplicationLifetime appLifetime, IOptions<ConnectionOptions> connectionOptions)
```
davidfowl: This is a breaking change but I'm not sure how much we care. Add another constructor until we decide it's OK to break here?

src/SignalR/common/Http.Connections/src/Internal/HttpConnectionManager.cs
```diff
@@ -34,17 +34,23 @@ public partial class HttpConnectionManager
...
             _nextHeartbeat = new TimerAwaitable(_heartbeatTickRate, _heartbeatTickRate);
-            DisconnectTimeout = connectionOptions.Value.DisconnectTimeout ?? ConnectionOptionsSetup.DefaultDisconectTimeout;
+            DisconnectTimeout = ConnectionOptionsSetup.DefaultDisconectTimeout;
             // Register these last as the callbacks could run immediately
             appLifetime.ApplicationStarted.Register(() => Start());
             appLifetime.ApplicationStopping.Register(() => CloseConnections());
         }
 
+        public HttpConnectionManager(ILoggerFactory loggerFactory, IApplicationLifetime appLifetime, IOptions<ConnectionOptions> connectionOptions)
+            :this(loggerFactory, appLifetime)
```
davidfowl: nit: formatting... 😄

src/SignalR/common/Http.Connections/src/Internal/HttpConnectionManager.cs
```diff
@@ -31,17 +32,25 @@ public partial class HttpConnectionManager
...
             _connectionLogger = loggerFactory.CreateLogger<HttpConnectionContext>();
             _nextHeartbeat = new TimerAwaitable(_heartbeatTickRate, _heartbeatTickRate);
-
+            DisconnectTimeout = ConnectionOptionsSetup.DefaultDisconectTimeout;
             // Register these last as the callbacks could run immediately
             appLifetime.ApplicationStarted.Register(() => Start());
             appLifetime.ApplicationStopping.Register(() => CloseConnections());
         }
 
+        public HttpConnectionManager(ILoggerFactory loggerFactory, IApplicationLifetime appLifetime, IOptions<ConnectionOptions> connectionOptions)
```
davidfowl: Usually, the one with the most argument does the actual work and the ones with less arguments just delegate to that one.

src/SignalR/common/Http.Connections/test/HttpConnectionDispatcherTests.cs
```diff
@@ -439,7 +440,7 @@ public async Task TransportEndingGracefullyWaitsOnApplicationLongPolling()
                     await task.OrTimeout();
 
                     // We've been gone longer than the expiration time
-                    connection.LastSeenUtc = DateTime.UtcNow.Subtract(TimeSpan.FromSeconds(10));
+                    connection.LastSeenUtc = DateTime.UtcNow.Subtract(TimeSpan.FromSeconds(20));
```
BrennanConroy: We should probably explicitly set the DisconnectTimeout in this test

src/SignalR/common/Http.Connections/src/ConnectionsDependencyInjectionExtensions.cs
```diff
@@ -20,9 +23,22 @@ public static IServiceCollection AddConnections(this IServiceCollection services
...
             return services;
         }
+
+        /// <summary>
+        /// Adds required services for ASP.NET Core Connection Handlers to the specified <see cref="IServiceCollection" />.
+        /// </summary>
+        /// <param name="services">The <see cref="IServiceCollection" /> to add services to.</param>
+        /// <param name="options">A callback to configure  <see cref="ConnectionOptions" /></param>
+        /// <returns>The same instance of the <see cref="IServiceCollection"/> for chaining.</returns>
+        public static IServiceCollection AddConnections(this IServiceCollection services, Action<ConnectionOptions> options)
```
BrennanConroy: What happens if you call the AddConnections overload with options after AddSignalR? And if you call the overload before AddSignalR?

mikaelm12: I tried the variations on the SignalRSamples project and it still works, propagating the passed in values and uses the default properly.

mikaelm12:
🆙 📅

davidfowl (approved):

PR #7343, opened 2019-02-07
Handle IAsyncDisposable scoped objects

davidfowl:
- We make a scope today around hub invocations, with IAsyncDisposable now implemented in the DI container, we need to support IServiceScope being IAsyncDisposable and IDisposable

https://github.com/aspnet/Extensions/commit/a38846cfa13a1e32a5aae90e54dbb71bef20a7db

### DIFF
--- a/src/SignalR/server/Core/src/Internal/AsyncDisposableExtensions.cs
+++ b/src/SignalR/server/Core/src/Internal/AsyncDisposableExtensions.cs
@@ -1,3 +1,6 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
 using System;
 using System.Collections.Generic;
 using System.Text;
--- a/src/SignalR/server/Core/src/Internal/AsyncDisposableExtensions.cs
+++ b/src/SignalR/server/Core/src/Internal/AsyncDisposableExtensions.cs
@@ -1,0 +1,21 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.SignalR.Internal
+{
+    internal static class AsyncDisposableExtensions
+    {
+        // Does a light up check to see if a type is IAsyncDisposable and calls DisposeAsync if it is
+        public static ValueTask DisposeAsync(this IDisposable disposable)
+        {
+            if (disposable is IAsyncDisposable asyncDisposable)
+            {
+                return asyncDisposable.DisposeAsync();
+            }
+            disposable.Dispose();
+            return default;
+        }
+    }
+}
--- a/src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs
+++ b/src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs
@@ -59,15 +59,7 @@ namespace Microsoft.AspNetCore.SignalR.Internal
             }
             finally
             {
-                if (scope is IAsyncDisposable asyncDisposable)
-                {
-                    // If the container supports IAsyncDisposable then dispose it asynchronously
-                    await asyncDisposable.DisposeAsync();
-                }
-                else
-                {
-                    scope.Dispose();
-                }
+                await scope.DisposeAsync();
             }
         }
 
@@ -93,15 +85,7 @@ namespace Microsoft.AspNetCore.SignalR.Internal
             }
             finally
             {
-                if (scope is IAsyncDisposable asyncDisposable)
-                {
-                    // If the container supports IAsyncDisposable then dispose it asynchronously
-                    await asyncDisposable.DisposeAsync();
-                }
-                else
-                {
-                    scope.Dispose();
-                }
+                await scope.DisposeAsync();
             }
         }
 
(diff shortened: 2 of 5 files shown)


src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs
```diff
@@ -53,12 +57,28 @@ public override async Task OnConnectedAsync(HubConnectionContext connection)
                     hubActivator.Release(hub);
                 }
             }
+            finally
```
DAllanCarr: Seems like alot of rinse repeat with the dispose pattern, could scope to method and inline.

src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs
```diff
@@ -342,17 +358,18 @@ async Task ExecuteInvocation()
...
         }
 
-        private void CleanupInvocation(HubConnectionContext connection, HubMethodInvocationMessage hubMessage, IHubActivator<THub> hubActivator,
+        private async Task CleanupInvocation(HubConnectionContext connection, HubMethodInvocationMessage hubMessage, IHubActivator<THub> hubActivator,
             THub hub, IServiceScope scope)
         {
             hubActivator?.Release(hub);
-            scope.Dispose();
 
+            await scope.DisposeAsync();
```
BrennanConroy: I think it's safe to return the task at the end of the function so no statemachine.

```
var scopeDisposeTask = scope.DisposeAsync();
// ...
return scopeDisposeTask;
```

src/SignalR/server/Core/src/Internal/AsyncDisposableExtensions.cs
```diff
@@ -0,0 +1,21 @@
+using System;
```
BrennanConroy: copyright

src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs
davidfowl: Why?

BrennanConroy: To get rid of the statemachine

davidfowl: It doesn't matter here. Not even a little bit

src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs
davidfowl: If you mean make a helper method that takes a lambda then that would have worse performance than just doing it inline.

DAllanCarr: Extension method is sweet

BrennanConroy (approved):

src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs
davidfowl: OK did it 👍

src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs
```diff
@@ -53,12 +57,20 @@ public override async Task OnConnectedAsync(HubConnectionContext connection)
                     hubActivator.Release(hub);
                 }
             }
+            finally
+            {
+                await scope.DisposeAsync();
```
halter73: Will this be converted back to `async using`?

src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs
```diff
@@ -365,6 +378,10 @@ async Task ExecuteInvocation()
                     catch (KeyNotFoundException) { }
                 }
             }
+
+            hubActivator?.Release(hub);
```
halter73: Why did you move `hubActivator?.Release(hub);` to the end? Is it important?

src/SignalR/server/SignalR/test/HubConnectionHandlerTests.cs
```diff
@@ -54,6 +54,37 @@ public async Task HubsAreDisposed()
...
+
+                    var result = (await client.InvokeAsync(nameof(HubWithAsyncDisposable.Test)).OrTimeout());
+                    Assert.NotNull(result);
+
+                    // kill the connection
+                    client.Dispose();
+
+                    await connectionHandlerTask;
+
+                    Assert.Equal(3, trackDispose.DisposeCount);
```
halter73: Is this three because the HubConnectionHandler happened to resolve and dispose HubWithAsyncDisposable 3 times? That's MockBehavior.Strict levels of finicky 😆

src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs
davidfowl: No, I just wanted the Release and Dispose to be together.

src/SignalR/server/SignalR/test/HubConnectionHandlerTests.cs
BrennanConroy: 1. OnConnectedAsync
1. HubWithAsyncDisposable.Test
1. OnDisconnectedAsync

src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs
davidfowl: Not unless we make a wrapper struct. Current plan is to leave it as is.

src/SignalR/server/SignalR/test/HubConnectionHandlerTests.cs
halter73: I know *why* the hub was resolved 3 times, but if we consider using the same hub for OnConnectedAsync and OnDisconnectedAsync is a bug, we should have a test that specifically tests that instead of making other tests unnecessarily strict. Not a big deal though, we'll just have one more test to change if we change this behavior.

PR #7350, opened 2019-02-07
Add xml doc comments for BufferWriter struct

AArnott:
I borrowed this code for another project and had to reverse-engineer how this struct worked so I could use it efficiently. I added the missing xml doc comments for myself, and to share with others.

### DIFF
--- a/src/Servers/Kestrel/Core/src/Internal/BufferWriter.cs
+++ b/src/Servers/Kestrel/Core/src/Internal/BufferWriter.cs
@@ -5,13 +5,38 @@ using System.Runtime.CompilerServices;
 
 namespace System.Buffers
 {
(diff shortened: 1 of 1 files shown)


davidfowl:
Doc comments on an internal type? Sure 😄

AArnott:
I find xml doc comments extremely helpful on all internal types and members, whether I wrote them or (even more so when) others write them. It saves me the effort of studying the implementation details of the method to figure out how to call it. Particularly so on a large project when it's likely that some developers in the code base aren't intimately familiar with the private details of every type in the project. :)

davidfowl:
@AArnott sure you did the work so I have no complaints 😄 .

davidfowl (approved):

PR #7352, opened 2019-02-07
Skip code signing job if dependencies aren't available

dougbu:


natemcmaster:
Hmm, the expanded condition doesn't look right. Have you tested this with a fake failure to make sure it works? (You could test by pushing directly to our internal repo and triggering a build that intentionally fails the xplat build step)

image

dougbu:
I suspect the CI pipeline didn't bother to expand the second part of the condition because signing wasn't enabled. Using the official build, I failed the MacOS job (`exit 23`) and signing was skipped correctly:
image
Other jobs ran fine:
image

natemcmaster (approved):

PR #7353, opened 2019-02-07
Reorganize helix files

HaoK:
@natemcmaster


src/Servers/Kestrel/test/Sockets.BindTests/Sockets.BindTests.csproj
```diff
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <Compile Include="..\..\..\..\..\test\xunit\SkipOnHelixAttribute.cs" />
+    <Compile Include="..\..\..\..\shared\test\SkipOnHelixAttribute.cs" />
```
natemcmaster:
```suggestion
    <Compile Include="$(SharedSourceRoot)test\SkipOnHelixAttribute.cs" />
```

Figuring out the `..`'s can be annoying.

Also, FYI paths are case-sensitive on Linux and I think it's a capital S in "Shared"

natemcmaster (approved):

PR #7354, opened 2019-02-07
Remove extraneous call to GetMemory in BufferWriter

AArnott:
The BufferWriter called GetMemory(count) and dropped the result, only to then call GetSpan(). This moves the count argument to GetSpan, and drops the call to GetMemory.


AArnott:
CC: @davidfowl @natemcmaster

pakrym (approved):

halter73 (approved):

PR #7355, opened 2019-02-07
Produce a nuget package with just reference assemblies for AspNetCore.App

natemcmaster:
This implements the first part of https://github.com/aspnet/AspNetCore/issues/6501.

To be done in a separate PR, if necessary. Still unclear on the details of native installers...

* Debian/Windows/Rpm installers
* Add zip's maybe?

cref https://github.com/dotnet/designs/pull/50


src/Framework/ref/Microsoft.AspNetCore.App.Ref.csproj
```diff
@@ -0,0 +1,89 @@
...
+    <TargetFramework>netcoreapp$(AspNetCoreMajorVersion).$(AspNetCoreMinorVersion)</TargetFramework>
+    <IsShippingPackage>true</IsShippingPackage>
+
+    <PackageDescription>This package provides reference assemblies for compiling an ASP.NET Core application. This package is an internal implementation of the .NET Core SDK and is not meant to be used as a normal PackageReference.</PackageDescription>
+    <PackageTags>aspnetcore;targeting-pack</PackageTags>
+    <SuppressDependenciesWhenPacking>true</SuppressDependenciesWhenPacking>
+
+    <!-- Subject to change: see https://github.com/dotnet/designs/pull/50 -->
+    <PackageType>TargetingPack</PackageType>
+    <RefAssemblyPackagePath>ref/$(TargetFramework)/</RefAssemblyPackagePath>
```
natemcmaster: @nguerrera what was the result of your battle with NuGet on this one? I had to step out early from that meeting on Monday.

natemcmaster:
Btw, is this package supposed to include docxml, too?

dsplaisted:
> Btw, is this package supposed to include docxml, too?

@natemcmaster Yes, it should (if you want intellisense to show ASP.NET API documentation :-) )

src/Framework/ref/Microsoft.AspNetCore.App.Ref.csproj
```diff
@@ -0,0 +1,89 @@
...
+  </PropertyGroup>
+
+  <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk" />
+  <Import Project="..\Microsoft.AspNetCore.App.props" />
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp$(AspNetCoreMajorVersion).$(AspNetCoreMinorVersion)</TargetFramework>
+    <IsShippingPackage>true</IsShippingPackage>
+
+    <PackageDescription>This package provides reference assemblies for compiling an ASP.NET Core application. This package is an internal implementation of the .NET Core SDK and is not meant to be used as a normal PackageReference.</PackageDescription>
```
dagood: Maybe this should also mention documentation and other design-time assets?

For comparison, this is what I have currently for Microsoft.NETCore.App.Ref:

> A set of .NET APIs that are included in the default .NET Core application model. Contains reference assemblies, documentation, and other design-time assets.

The first sentence is Microsoft.NETCore.App's existing description, second sentence is based on the design doc definition, "reference assemblies, documentation, and other design-time assets".

I see I'm missing the internal implementation warning--filed https://github.com/dotnet/core-setup/issues/5107 to add that.

dagood (commented):
In Core-Setup I continued with custom packaging infra (which is shared with CoreFX), so I can't really comment on the infra. Good to see where this is and what it takes with what I assume are the SDK packaging targets, though.

src/Framework/ref/Microsoft.AspNetCore.App.Ref.csproj
```diff
@@ -0,0 +1,89 @@
...
+
+  <PropertyGroup>
+    <OutputInRepoRoot>true</OutputInRepoRoot>
+  </PropertyGroup>
+
+  <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk" />
+  <Import Project="..\Microsoft.AspNetCore.App.props" />
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp$(AspNetCoreMajorVersion).$(AspNetCoreMinorVersion)</TargetFramework>
```
pakrym: Should we define this at some lower level (root Directory.props maybe)? Seems like something that would be very commonly used across the repo.

src/Framework/ref/Microsoft.AspNetCore.App.Ref.csproj
```diff
@@ -0,0 +1,89 @@
...
+    <CopyBuildOutputToOutputDirectory>false</CopyBuildOutputToOutputDirectory>
+    <CopyOutputSymbolsToOutputDirectory>false</CopyOutputSymbolsToOutputDirectory>
+    <GenerateDependencyFile>false</GenerateDependencyFile>
+
+    <!-- This project should not be referenced via the `<Reference>` impementation. -->
+    <IsProjectReferenceProvider>false</IsProjectReferenceProvider>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Reference Include="@(Dependency)" />
```
pakrym: Where is `Dependency` coming from?

natemcmaster: `..\Microsoft.AspNetCore.App.props` (see https://github.com/aspnet/AspNetCore/blob/master/src/Framework/Microsoft.AspNetCore.App.props)

src/Framework/ref/Microsoft.AspNetCore.App.Ref.csproj
natemcmaster: Probably. Updated to add a property named DefaultNetCoreTargetFramework. This could be used in a lot of places, but for now I'll just update the only two places currently using `netcoreapp$(AspNetCoreMajorVersion).$(AspNetCoreMinorVersion)`

natemcmaster:
Updated for include xmldocs for all assemblies except the two from dotnet/core-setup (see https://github.com/dotnet/core-setup/pull/5109)

src/Mvc/src/Microsoft.AspNetCore.Mvc/Microsoft.AspNetCore.Mvc.csproj
```diff
@@ -9,7 +9,6 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <Reference Include="Microsoft.AspNetCore.Mvc.Analyzers" PrivateAssets="None" />
```
natemcmaster: FYI @pranavkm  - this reference shouldn't be used here. Found it as a result of trying to figure out why analyzers were in the reference API surface for Microsoft.AspNetCore.App.

src/Framework/ref/Microsoft.AspNetCore.App.Ref.csproj
natemcmaster: I don't really care either way, just wanted to know which to use to make this actually work with the SDK. Is the SDK being updated soon to look for `data/`? Last I checked, the FrameworkReference resolution still looks for `ref/<TFM>/`

nguerrera: Anyway, I think we can follow up with a PR to these two lines when we have the decision.

natemcmaster: Ok, let me and @pakrym know when a decision is reached. I'll leave as `ref/<tfm>` for now cuz I want to test this with the SDK to find the holes in this implementation.

natemcmaster:
Alright, I think this is ready to merge. Had to to react to changes merged earlier today which breaks XML docs generation.

pakrym (approved):

dsplaisted:
@natemcmaster @pakrym Can you let me know the version of the Ref package that's published to the internal feed, so I can start using it in the SDK?

natemcmaster:
The first build with this package was 3.0.0-preview-19107-09.

dsplaisted:
@natemcmaster When testing the new package, I'm getting a compile error with the following code:

```c#
        public static IHostBuilder CreateHostBuilder(string[] args) =>
            Host.CreateDefaultBuilder(args)
                .ConfigureWebHostDefaults(webBuilder =>
                {
                    webBuilder.UseStartup<Startup>();
                });
```

The error is: `error CS0103: The name 'Host' does not exist in the current context`

Is this a mismatch between the targeting pack I'm using and the template version for `dotnet new mvc`?

natemcmaster:
This is very strange. Microsoft.AspNetCore.App.Ref is missing the assembly with this API. Let me debug further to see why `ResolveReferences` isn't putting this into the item group I'm using to create the package.

natemcmaster:
Found the problem. I'll send a PR today.

PR #7361, opened 2019-02-07
Enable verbose h2spec logs

Tratcher:
https://github.com/aspnet/AspNetCore-Internal/issues/1786
Experementing with verbose client logs to see if they include anything more useful.


halter73 (approved):

Tratcher:
@HaoK how do I get the test logs from the Helix run?

HaoK:
Right now what you see in the helix run logs on mission control is all we have.  We can certainly come up with some mechanism for tests to copy logs somewhere central so we can see them after, and add it to the bootstrappers, see https://github.com/aspnet/AspNetCore/blob/master/eng/helix/vstest/runtests.cmd

We have full control so we can add some additional commands that copy logs to some well known location

PR #7373, opened 2019-02-08
Update helix sdk and try workaround for warning

HaoK:
Try workaround suggested for https://github.com/dotnet/arcade/issues/1605

Also pickup fix for parallel uploads https://github.com/dotnet/arcade/issues/1483 to see if it improves helix run time

### DIFF
--- a/eng/helix/Directory.Build.props
+++ b/eng/helix/Directory.Build.props
@@ -1,0 +1,2 @@
+<Project>
+</Project>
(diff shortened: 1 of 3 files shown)


natemcmaster (approved):

HaoK:
@jonfortescue the empty stub directory.build props/targets resulted in a build error, so I'm backing them out of this PR

HaoK:
Ok looks like the regression/change is actually from helix sdk itself as opposed to the stub files, as just updating the sdk causes the helix work items to no longer be found

Issue #7326, opened 2019-02-06
Remove obsolete APIs from Razor


mkArtakMSFT:
See #3654 for details.

#### (55) src\Mvc\src\Microsoft.AspNetCore.Mvc.Razor\Compilation\CompiledViewDescriptor.cs
Links: <a href="https://github.com/aspnet/AspNetCore/blob/master/src/Mvc/src/Microsoft.AspNetCore.Mvc.Razor/Compilation/CompiledViewDescriptor.cs#L89">[master]</a> | <a href="https://github.com/aspnet/AspNetCore/blob/release/2.2/src/Mvc/src/Microsoft.AspNetCore.Mvc.Razor/Compilation/CompiledViewDescriptor.cs#L89">[release/2.2]</a>
[...]

PR #7374, opened 2019-02-08
[MVC] Remove obsolete APIs

NTaylorMullen:
- Removed ViewAttribute Property
- Removed RazorViewAttribute
- Removed ViewsFeatureProvider
- Removed PageArgumentBinder and its internal implementation DefaultPageArgumentBinder.
- Removed RazorPageAttribute
- Removed corresponding test classes/methods for all the above.
- Reacted to class/member changes in dependencies.

Addresses #7326


src/Mvc/src/Microsoft.AspNetCore.Mvc.Razor/Compilation/CompiledViewDescriptor.cs
```diff
@@ -37,59 +37,11 @@ public CompiledViewDescriptor(RazorCompiledItem item)
...
         }
 
-#pragma warning disable CS0618// Type or member is obsolete
-        /// <summary>
-        /// Creates a new <see cref="CompiledViewDescriptor"/>. At least one of <paramref name="attribute"/> or
-        /// <paramref name="item"/> must be non-<c>null</c>.
-        /// </summary>
-        /// <param name="item">The <see cref="RazorCompiledItem"/>.</param>
-        /// <param name="attribute">The <see cref="RazorViewAttribute"/>.</param>
-        public CompiledViewDescriptor(RazorCompiledItem item, RazorViewAttribute attribute)
```
rynowak: This was **just** marked as obsolete. I'm not sure why we'd remove it in 3.0 since it's not costing us anything to keep around.

src/Mvc/src/Microsoft.AspNetCore.Mvc.Razor/Compilation/RazorViewAttribute.cs
```diff
@@ -1,28 +0,0 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
-using System;
-
-namespace Microsoft.AspNetCore.Mvc.Razor.Compilation
-{
-    [Obsolete("This attribute has been superseded by RazorCompiledItem and will not be used by the runtime.")]
-    [AttributeUsage(AttributeTargets.Assembly, AllowMultiple = true)]
-    public class RazorViewAttribute : Attribute
```
rynowak: This will cause using anything compiled against 2.0 to fail with a `TypeLoadException`. This was also **just** marked obsolete

rynowak: I think we need to take a lot more care with removing obsolete things that appear in user-code and build outputs of user code.

src/Mvc/src/Microsoft.AspNetCore.Mvc.Razor/Compilation/ViewsFeatureProvider.cs
```diff
@@ -1,111 +0,0 @@
...
-using Microsoft.AspNetCore.Mvc.ApplicationParts;
-using Microsoft.Extensions.Primitives;
-
-namespace Microsoft.AspNetCore.Mvc.Razor.Compilation
-{
-    /// <summary>
-    /// An <see cref="IApplicationFeatureProvider{TFeature}"/> for <see cref="ViewsFeature"/>.
-    /// </summary>
-    [Obsolete("This type is obsolete and will be removed in a future version. See " + nameof(IRazorCompiledItemProvider) + " for alternatives.")]
-    public class ViewsFeatureProvider : IApplicationFeatureProvider<ViewsFeature>
```
rynowak: For contrast, this is random old extensibility that is a good thing to remove since it will rarely appear in user-code.

dougbu (approved):
Once @rynowak's happy, I'm happy

src/Mvc/src/Microsoft.AspNetCore.Mvc.Razor/Compilation/RazorViewAttribute.cs
NTaylorMullen: Ah, I took the presence of the release/2.2 link above this specific item verbatim in the issue. Figured it was obsoleted there. Either way I think you have a fair point about the type of complications this would cause to remove an API like this.

src/Mvc/src/Microsoft.AspNetCore.Mvc.Razor/Compilation/ViewsFeatureProvider.cs
NTaylorMullen: One concern I do have, is if I remove this but don't remove the `RazorViewAttribute` and a user doesn't heed the obsolete warnings they'll get a hard to resolve, significant behavior change.

pranavkm: > and a user doesn't heed the obsolete warnings they'll get a hard to resolve

We can add it to the migration notes as something to be aware of. If we get lots of user feedback (I'd think it's fairly unlikely given this attribute comes from Precompilation), we can always change the runtime to throw.

NTaylorMullen: I'm still not convinced since i'd rather break hard or not at all then break soft when it comes to these things (can have migration guidance with either hard or soft breaks). I'll leave it as removed though since you both feel differently 😄 

@danroth27 is there a place where we're collecting migration guidance?

NTaylorMullen:
🆙 📅

NTaylorMullen:
ping

rynowak (approved):

PR #7375, opened 2019-02-08
Small fixes to SignalR service extension methods

JamesNK:
Borrowing stuff from SignalR configuration and I noticed some little things.

### DIFF
--- a/src/SignalR/server/SignalR/src/SignalRDependencyInjectionExtensions.cs
+++ b/src/SignalR/server/SignalR/src/SignalRDependencyInjectionExtensions.cs
@@ -24,6 +24,11 @@ namespace Microsoft.Extensions.DependencyInjection
         /// <returns>The same instance of the <see cref="ISignalRServerBuilder"/> for chaining.</returns>
         public static ISignalRServerBuilder AddHubOptions<THub>(this ISignalRServerBuilder signalrBuilder, Action<HubOptions<THub>> configure) where THub : Hub
         {
+            if (signalrBuilder == null)
+            {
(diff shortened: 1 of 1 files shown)


davidfowl (approved):

BrennanConroy (approved):

src/SignalR/server/SignalR/src/SignalRDependencyInjectionExtensions.cs
```diff
@@ -48,10 +58,15 @@ public static ISignalRServerBuilder AddSignalR(this IServiceCollection services)
         /// Adds SignalR services to the specified <see cref="IServiceCollection" />.
         /// </summary>
         /// <param name="services">The <see cref="IServiceCollection" /> to add services to.</param>
-        /// <param name="configure">An <see cref="Action{MvcOptions}"/> to configure the provided <see cref="HubOptions"/>.</param>
+        /// <param name="configure">An <see cref="Action{HubOptions}"/> to configure the provided <see cref="HubOptions"/>.</param>
```
mikaelm12: Good catch. While you're here can you also update the param doc for the action on `AddHubOptions` ?

mikaelm12 (approved):

src/SignalR/server/SignalR/src/SignalRDependencyInjectionExtensions.cs
JamesNK: It doesn't have the same mistake. Referencing generics in xmldoc is difficult.

PR #7376, opened 2019-02-08
Make Dispose not throw in StreamPipeWriter

jkotalik:
Follow up to https://github.com/aspnet/AspNetCore/issues/7102

### DIFF
--- a/src/Http/Http/src/StreamPipeWriter.cs
+++ b/src/Http/Http/src/StreamPipeWriter.cs
@@ -129,20 +129,7 @@ namespace System.IO.Pipelines
                 return;
             }
 
-            _isCompleted = true;
-
-            _internalTokenSource?.Dispose();
-
-            if (_completedSegments != null)
-            {
-                foreach (var segment in _completedSegments)
-                {
-                    segment.Return();
-                }
-            }
-
-            _currentSegmentOwner?.Dispose();
-
+            Dispose();
             // We still want to cleanup segments before throwing an exception.
             if (_bytesWritten > 0 && exception == null)
             {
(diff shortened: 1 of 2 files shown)


davidfowl:
The usual pattern for a PipeWriter is to call Complete on the writer when you are done with it and nobody else is going to write to the `PipeWriter`. To make the API more usable in ASP.NET Core, we don't require calling Complete and it's the job of the pipeline to clean up the memory when the request is over. I also thing that Dispose and Complete are synonymous (we didn't make pipe implement IDisposable since the Pipe could be reset but that might have been a mistake in hindsight). Overall, that makes things easier but we have to decide what to do with buffered data:

1. Dispose/Complete can throw if there is buffered data.
2. Dispose/Complete can discard buffered data.
3. Dispose/Complete can Write and Flush unbuffered data
    - Implement IAsyncDisposable and Write and Flush unbuffered

I think we should do 3 and always flush the unbuffered data in Dispose and do it asynchronously now that we have IAsyncDisposable.

Tratcher:
By the time dispose would be called it's too late to flush the data. At best we can log a usage error.

davidfowl:
> By the time dispose would be called it's too late to flush the data. At best we can log a usage error.

Then lets go with this PR (no logging). We can discuss how we could log but throwing is a non starter.

Tratcher:
Since Complete isn't required I still expect people to make the mistake where they buffer data but forget to flush it. At least they'll know immediately on the client side that the response was incomplete, but tracking down why might take some work.

Tratcher (approved):

PR #7392, opened 2019-02-08
Account for transitive shared framework dependencies separately

natemcmaster:
Put transitive external dependencies of the shared framework in a separate category, and don't reference them directly unless we are building a patch. This will help us find changes to dependencies, such as the removal of JSON.NET or possible changes to Crypto.Xml.

FYI @dsplaisted - this also fixes missing Microsoft.Extensions.Hosting.dll from the reference assembly package.


natemcmaster:
Helix tests are failing, but only for Fedora 28. Opened https://github.com/aspnet/AspNetCore/issues/7398. I'm going to disable this queue because it seems like a problem with the machine, not our tests. Tests are otherwise passing:

https://mc.dot.net/#/user/aspnetcore/pr~2Faspnet~2Faspnetcore/ci/20190208.14
image

natemcmaster:
=> https://github.com/aspnet/AspNetCore/pull/7399

pakrym (approved):

PR #7394, opened 2019-02-08
Update package branding to 3.0.0-preview3

natemcmaster:
Per latest 3.0 release plan, we'll start including the preview number in the package branding.

### DIFF
--- a/version.props
+++ b/version.props
@@ -3,8 +3,9 @@
     <AspNetCoreMajorVersion>3</AspNetCoreMajorVersion>
     <AspNetCoreMinorVersion>0</AspNetCoreMinorVersion>
     <AspNetCorePatchVersion>0</AspNetCorePatchVersion>
-    <PreReleaseLabel>preview</PreReleaseLabel>
-    <PreReleaseBrandingLabel>Preview</PreReleaseBrandingLabel>
+    <PreReleasePreviewNumber>3</PreReleasePreviewNumber>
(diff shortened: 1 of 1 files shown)


version.props
```diff
@@ -3,8 +3,9 @@
     <AspNetCoreMajorVersion>3</AspNetCoreMajorVersion>
     <AspNetCoreMinorVersion>0</AspNetCoreMinorVersion>
     <AspNetCorePatchVersion>0</AspNetCorePatchVersion>
-    <PreReleaseLabel>preview</PreReleaseLabel>
-    <PreReleaseBrandingLabel>Preview</PreReleaseBrandingLabel>
+    <PreReleasePreviewNumber>3</PreReleasePreviewNumber>
+    <PreReleaseLabel>preview$(PreReleasePreviewNumber)</PreReleaseLabel>
+    <PreReleaseBrandingLabel>Preview $(PreReleasePreviewNumber)</PreReleaseBrandingLabel>
```
dougbu: Curious: Why use a variable here and not in the other repos? Is this because we have both `$(PreReleaseLabel)` and `$(PreReleaseBrandingLabel)` here?

dougbu (approved):

version.props
natemcmaster: Yes. One fewer thing that can get out of sync.

natemcmaster: PreReleaseBrandingLabel is for the "human friendly" version strings in installers, like "ASP.NET Core 3.0.0 Preview 3"

PR #7399, opened 2019-02-08
Disable helix tests on Fedora.28 and 27

natemcmaster:
cref https://github.com/aspnet/AspNetCore/issues/7398

Something is wrong with the machines. All tests fail. Disabling to allow time for investigation without blocking PR checks.

### DIFF
--- a/eng/helix/helix.proj
+++ b/eng/helix/helix.proj
@@ -44,8 +44,8 @@
       Debian.8.Amd64.Open;
       Debian.9.Amd64.Open;
       Fedora.27.Amd64.Open;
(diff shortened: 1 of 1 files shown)


HaoK (approved):

HaoK:
yeah not sure what happened since this queue was fine a few days ago

natemcmaster:
Ok, what is going on. Now all Fedora 27 builds are failing too. These were fine an hour ago.

HaoK:
The good news is now our helix runs might be less than an hour with 2 less queues :)

PR #7402, opened 2019-02-08
Fix build break in Identity tests

ajcvickers:
By configuring the service provider warning to log instead of throw


HaoK (approved):

HaoK:
I think we had a bug laying around talking about these warnings at some point too

ajcvickers:
Yah, we do. I'm going to take this breaking change back to triage on the EF side. I think we'll break more people than we might have thought.

PR #7404, opened 2019-02-08
[Backport] Add CancelInvocation support to MsgPack in TS client

BrennanConroy:
Backport of https://github.com/aspnet/AspNetCore/pull/7224
Issue https://github.com/aspnet/AspNetCore/issues/7157

#### Description

The @aspnet/signalr-protocols-msgpack addition to the Javascript client would throw when sending a new message that was added to the SignalR protocol in the 2.2 release. This change adds the serializing for that new message so it no longer throws for correct behavior.

#### Customer Impact

Customers will no longer get exceptions for using this correct behavior.

#### Regression?

None, this is a new feature that was added and was missed getting added to the @aspnet/signalr-protocols-msgpack npm package.

#### Risk

Very low, adding a serialized response on the client when it would incorrectly throw before.

cc @muratg

### DIFF
--- a/src/SignalR/clients/ts/FunctionalTests/TestHub.cs
+++ b/src/SignalR/clients/ts/FunctionalTests/TestHub.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Reactive.Linq;
+using System.Threading;
 using System.Threading.Channels;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http.Connections;
@@ -20,6 +21,13 @@ namespace FunctionalTests
 
     public class TestHub : Hub
     {
+        private readonly IHubContext<TestHub> _context;
+
+        public TestHub(IHubContext<TestHub> context)
+        {
+            _context = context;
+        }
+
         public string Echo(string message)
         {
             return message;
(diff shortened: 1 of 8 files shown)


muratg:
@Eilon could you bar-check this one please?

Eilon:
1. Why is it already marked servicing-approved?
2. Does this only affect the npm package? This doesn't ship in any NUPKG or installers/templates right? If so, this seems reasonable to me and we should get shiproom approval.

mikaelm12:
>1. Why is it already marked servicing-approved?

I'm pretty sure that was an off by one. I just updated it to consider.

> 2. Does this only affect the npm package? This doesn't ship in any NUPKG or installers/templates right?

This is correct.

Eilon:
Cool. Can send shiproom mail for it. BTW I recommend updating the shiproom template to indicate it ships as OOB because it's an NPM package, not NuGet.

BrennanConroy:
>indicate it ships as OOB

We try not to do that and build/ship at the same time the nupkgs do.

Eilon:
Yeah I get we ship as part of the build, and I don't want to do a one-off, but it's less scary when it's not in a NUPKG 😄

muratg:
Thanks! (It was my mistake to mislabel this.)

I'll get this to the shiproom tomorrow.

halter73 (approved):

vivmishra:
Approved for 2.2.3

muratg (approved):
Approved in shiproom. Let's get this in asap (before EOD please.)

mikaelm12 (approved):

muratg:
@dougbu FYI this is in.

PR #7407, opened 2019-02-09
Made changes to stream <-> pipe adapters

davidfowl:
- Use the array pool by default when the shared memory pool is specified for both the StreamPipeReader and StreamPipeWriter
- Support allocating unpooled memory if the StreamPipeWriter is asked for memory outside of the max pool size

### DIFF
--- a/src/Http/Http/src/BufferSegment.cs
+++ b/src/Http/Http/src/BufferSegment.cs
@@ -9,7 +9,7 @@ namespace System.IO.Pipelines
 {
     internal sealed class BufferSegment : ReadOnlySequenceSegment<byte>
     {
-        private IMemoryOwner<byte> _memoryOwner;
+        private object _memoryOwner;
         private BufferSegment _next;
         private int _end;
 
@@ -46,12 +46,40 @@ namespace System.IO.Pipelines
             }
         }
 
+        public void SetMemory(object memoryOwner)
+        {
+            if (memoryOwner is IMemoryOwner<byte> owner)
+            {
+                SetMemory(owner);
+            }
+            else if (memoryOwner is byte[] array)
+            {
+                SetMemory(array);
+            }
+            else
+            {
+                Debug.Fail("Unexpected memoryOwner");
+            }
+        }
+
         public void SetMemory(IMemoryOwner<byte> memoryOwner)
         {
             _memoryOwner = memoryOwner;
 
-            AvailableMemory = _memoryOwner.Memory;
+            SetUnownedMemory(memoryOwner.Memory);
+        }
 
+        public void SetMemory(byte[] arrayPoolBuffer)
+        {
+            _memoryOwner = arrayPoolBuffer;
+
+            SetUnownedMemory(arrayPoolBuffer);
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void SetUnownedMemory(Memory<byte> memory)
+        {
+            AvailableMemory = memory;
             RunningIndex = 0;
             End = 0;
             NextSegment = null;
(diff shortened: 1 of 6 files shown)


src/Http/Http/test/StreamPipeWriterTests.cs
```diff
@@ -320,6 +320,24 @@ public void CallCompleteWithoutFlushAndException_DoesNotThrowInvalidOperationExc
...
+            var memory = Writer.GetMemory(Pool.MaxBufferSize);
+
+            Assert.Equal(Pool.MaxBufferSize, memory.Length);
+            Assert.Equal(1, Pool.GetRentCount());
+        }
+
+        [Fact]
+        public void GetMemoryBiggerThanPoolSizeAllocatesUnpooledArray()
+        {
+            var memory = Writer.GetMemory(Pool.MaxBufferSize + 1);
```
jkotalik: What is the default size of this memory? If it's Int.Max, consider recreating the pool with a different MaxBufferSize.

jkotalik (approved):

src/Http/Http/test/StreamPipeWriterTests.cs
davidfowl: It'll OOM.

src/Http/Http/src/BufferSegment.cs
```diff
@@ -46,25 +46,62 @@ public BufferSegment NextSegment
...
+
+        public void SetMemory(byte[] arrayPoolBuffer)
+        {
+            _memoryOwner = arrayPoolBuffer;
 
+            SetUnownedMemory(arrayPoolBuffer);
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void SetUnownedMemory(Memory<byte> memory)
```
pakrym: private.

src/Http/Http/src/StreamPipeReader.cs
```diff
@@ -60,9 +60,9 @@ public StreamPipeReader(Stream readingStream, StreamPipeReaderOptions options)
                 throw new ArgumentOutOfRangeException(nameof(options.MinimumReadThreshold));
             }
 
-            _minimumSegmentSize = options.MinimumSegmentSize;
             _minimumReadThreshold = Math.Min(options.MinimumReadThreshold, options.MinimumSegmentSize);
-            _pool = options.MemoryPool;
+            _pool = options.MemoryPool == MemoryPool<byte>.Shared ? null : options.MemoryPool;
+            _bufferSize = _pool == null ? options.MinimumSegmentSize : Math.Min(options.MinimumSegmentSize, _pool.MaxBufferSize);
```
pakrym: Rename public property too?

src/Http/Http/src/StreamPipeWriter.cs
```diff
@@ -289,31 +303,44 @@ public void Dispose()
                 }
             }
 
-            _currentSegmentOwner?.Dispose();
+            DisposeOwner(_currentSegmentOwner);
+        }
+
+        private static void DisposeOwner(object owner)
```
pakrym: Helper method?

pakrym: To share with BufferSegment

src/Http/Http/src/StreamPipeReader.cs
pakrym: Should we throw in the case where MinimumSegmentSize is larger then MaxBufferSize?

src/Http/Http/src/BufferSegment.cs
davidfowl: When this type moves to System.IO.Pipelines, this will be public so I rather not have them diverge too much.

src/Http/Http/src/StreamPipeWriter.cs
davidfowl: Same comment as above. Once we move this over to corefx then the merging will happen.

src/Http/Http/src/StreamPipeReader.cs
davidfowl: Yep, I plan to make all of the corefx changes in here in another PR. I think these types need a bit more bake time in this repo first.

PR #7414, opened 2019-02-09
Extend WebApplicationFactory<T>

martincostello:
This pull request extends `WebApplicationFactory<T>` based on some experiences I've had with it while updating some applications to use ASP.NET Core 3.0 using preview 2 from ASP.NET Core 2.2.

First, it adds a new property, `public virtual IServiceProvider Services`. This is to support the scenario where the test application wants access to the `IServiceProvider` of the hosted service.

In 2.x, this was possible through the access chain `factory.Server.Host.Services`, however in 3.0 if the application is using `IHostBuilder` this access throws the following exception:

```
System.InvalidOperationException:
  The TestServer constructor was not called with a IWebHostBuilder so IWebHost is not available.
```

As the `IHost` vs. `IWebHost` implementation detail is hidden, rather than expose two different properties for the possible hosts and make the caller choose, instead the new property returns the appropriate `IServiceProvider` based on the host that is in use.

The property is `virtual` so that derived instances leveraging the base functionality to use a different host can override the property to change the `IServiceProvider`. An example of this can be seen in this PR for an application of my own where I build on top of `WebApplicationFactory<T>` to host the server on an actual HTTP port for use for browser-based UI tests: https://github.com/martincostello/alexa-london-travel-site/pull/252

Secondly, this PR adds a new method `protected string GetContentRootPath()` that exposes the existing logic to get the content root path for the application being tested. Previously this was private, so if a derived type also needed access to the value for some reason, it had to duplicate the code for itself.

I also considered exposing `EnsureServer()` in some manner so that derived types don't need to "hack" the services being started up by calling either `CreateDefaultClient()` or `Server` so they can ensure the server is running before doing other custom setup actions. I wasn't sure exactly what to call that, so I've left it for now, but if you have any suggestions on the best way to do that, I'd also like to do that either in this PR or a different one.

### DIFF
--- a/src/Mvc/Mvc.Testing/ref/Microsoft.AspNetCore.Mvc.Testing.netcoreapp3.0.cs
+++ b/src/Mvc/Mvc.Testing/ref/Microsoft.AspNetCore.Mvc.Testing.netcoreapp3.0.cs
@@ -26,6 +26,7 @@ namespace Microsoft.AspNetCore.Mvc.Testing
         public Microsoft.AspNetCore.Mvc.Testing.WebApplicationFactoryClientOptions ClientOptions { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
         public System.Collections.Generic.IReadOnlyList<Microsoft.AspNetCore.Mvc.Testing.WebApplicationFactory<TEntryPoint>> Factories { get { throw null; } }
         public Microsoft.AspNetCore.TestHost.TestServer Server { get { throw null; } }
+        public virtual System.IServiceProvider Services { get { throw null; } }
         protected virtual void ConfigureClient(System.Net.Http.HttpClient client) { }
         protected virtual void ConfigureWebHost(Microsoft.AspNetCore.Hosting.IWebHostBuilder builder) { }
         public System.Net.Http.HttpClient CreateClient() { throw null; }
--- a/src/Mvc/Mvc.Testing/src/WebApplicationFactory.cs
+++ b/src/Mvc/Mvc.Testing/src/WebApplicationFactory.cs
@@ -80,6 +80,18 @@ namespace Microsoft.AspNetCore.Mvc.Testing
         }
 
         /// <summary>
+        /// Gets the <see cref="IServiceProvider"/> created by the server associated with this <see cref="WebApplicationFactory{TEntryPoint}"/>.
+        /// </summary>
+        public virtual IServiceProvider Services
+        {
+            get
+            {
+                EnsureServer();
+                return _host?.Services ?? _server.Host.Services;
+            }
+        }
+
+        /// <summary>
         /// Gets the <see cref="IReadOnlyList{WebApplicationFactory}"/> of factories created from this factory
         /// by further customizing the <see cref="IWebHostBuilder"/> when calling 
         /// <see cref="WebApplicationFactory{TEntryPoint}.WithWebHostBuilder(Action{IWebHostBuilder})"/>.
--- a/src/Mvc/test/Mvc.FunctionalTests/TestingInfrastructureInheritanceTests.cs
+++ b/src/Mvc/test/Mvc.FunctionalTests/TestingInfrastructureInheritanceTests.cs
@@ -52,6 +52,17 @@ namespace Microsoft.AspNetCore.Mvc.FunctionalTests
             Assert.False(factory.CreateWebHostBuilderCalled);
         }
 
+        [Fact]
+        public void TestingInfrastructure_GenericHost_WithWithHostBuilderHasServices()
+        {
+            // Act
+            var factory = new CustomizedFactory<GenericHostWebSite.Startup>();
+
+            // Assert
+            Assert.NotNull(factory.Services);
+            Assert.NotNull(factory.Services.GetService(typeof(IConfiguration)));
+        }
+
         private class CustomizedFactory<TEntryPoint> : WebApplicationFactory<TEntryPoint> where TEntryPoint : class
         {
             public bool GetTestAssembliesCalled { get; private set; }
--- a/src/Mvc/test/Mvc.FunctionalTests/TestingInfrastructureInheritanceTests.cs
+++ b/src/Mvc/test/Mvc.FunctionalTests/TestingInfrastructureInheritanceTests.cs
@@ -7,6 +7,7 @@ using System.Reflection;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Mvc.Testing;
 using Microsoft.AspNetCore.TestHost;
+using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Hosting;
 using Xunit;
 
--- a/src/Mvc/test/Mvc.FunctionalTests/TestingInfrastructureTests.cs
+++ b/src/Mvc/test/Mvc.FunctionalTests/TestingInfrastructureTests.cs
@@ -14,6 +14,7 @@ using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Mvc.Testing;
 using Microsoft.AspNetCore.Mvc.Testing.Handlers;
 using Microsoft.AspNetCore.TestHost;
+using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using RazorPagesClassLibrary;
 using Xunit;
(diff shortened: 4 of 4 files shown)


src/Mvc/src/Microsoft.AspNetCore.Mvc.Testing/WebApplicationFactory.cs
```diff
@@ -103,6 +115,40 @@ public TestServer Server
         public WebApplicationFactory<TEntryPoint> WithWebHostBuilder(Action<IWebHostBuilder> configuration) =>
             WithWebHostBuilderCore(configuration);
 
+        /// <summary>
+        /// Gets the content root path for the <see cref="WebApplicationFactory{TEntryPoint}"/>, if known.
+        /// </summary>
+        /// <returns>
+        /// The path of the content root path to use, if known; otherwise <see langword="null"/>.
+        /// </returns>
+        protected string GetContentRootPath()
```
javiercn: Can you provide a specific use-case for this? I don't want to add public APIs if we don't have to as these are more things to support/maintain.

Could you not get the ContentRoot from the IHost/IWebHost directly?

martincostello: The default is being calculated incorrectly in the use case where I'm deriving from `WAF<T>` to host an HTTP server on a real port for Selenium UI tests, so there's no point reaching in for it, because it's wrong. This mean that the application was failing to find various static assets, like JS files, so the tests were failing.

I needed to manually "fix" it, and that involved having to duplicate this code so I could re-use it:

https://github.com/martincostello/alexa-london-travel-site/blob/e9af0ce3e70c7cbb0d047a276db42491ccec89e5/tests/LondonTravel.Site.Tests/Integration/TestServerFixture.cs#L83

It's possible that there's a separate issue either here or in my code that means the root is being set/calculated/overridden incorrectly in the first place, but it seemed "better" to be able to just leverage the code that was already there (but privately), rather than having to copy-paste as I've had to do with 2.2 and 3.0 preview 2.

javiercn: @martincostello The WAF is meant to be used specifically for in-memory integration tests. 
I want to make sure you understand that there is currently no support for standing up a real server using the web application factory. 

I want to be very careful in exposing the guts of the WAF if there is no good reason for it, as each time we add public API it has a cost for us and limits our options in the future. The WAF was designed to enable in-memory integration tests and trying to use it for E2E tests I believe doesn't have a lot of benefits compared to copying the same approach we follow in other places (in our actual E2E tests).

If you are trying to do E2E tests with Selenium we have other things that might help in this area, although nothing is official, you can follow the same approach. See
https://github.com/aspnet/AspNetCore/tree/master/src/Hosting/Server.IntegrationTesting
https://github.com/aspnet/AspNetCore/tree/master/src/Middleware/CORS/test/FunctionalTests
https://github.com/aspnet/AspNetCore/blob/master/src/ProjectTemplates/test/Infrastructure/SeleniumServerFixture.cs

I would be interested in getting a better feeling on what specifically made you want to use the WebApplicationFactory along with Selenium and see if there is a way we can factor those things out for ease of use instead of exposing the WAF guts.

javiercn:
> In 2.x, this was possible through the access chain `factory.Server.Host.Services`, however in 3.0 if the application is using `IHostBuilder` this access throws the following exception:
> 
> ```
> System.InvalidOperationException:
>   The TestServer constructor was not called with a IWebHostBuilder so IWebHost is not available.
> ```

@Tratcher Is this a bug on TestServer? Is there an alternative to reach the service collection from testserver when using IHost?

src/Mvc/src/Microsoft.AspNetCore.Mvc.Testing/WebApplicationFactory.cs
martincostello: The main motivation was because WAF contains a lot of boilerplate to make sure that Razor compilation references work correctly for self-hosted tests on a non-deployed application.

Previously trying to do this without leveraging WAF just meant that the application failed to find the references when Razor views were executed, which broke the application.

As Selenium tests to do UI automation (where buttons need to be clicked, etc.), an HTTP port has to be exposed so that the browser can actually interact with the application. This sort of testing isn't possible in memory.

In memory works great for APIs and very basic HTML applications, but it doesn't work for more complex HTML-based applications where you want to do as much testing in the inner loop as possible without also having to maintain a lot of extra boilerplate to make it work.

WAF happens to contain most of this boilerplate, but due to the design being through inheritance and protected methods, rather than generic callable public APIs, the only way to leverage it at the moment to to create derived types and then customise the behaviour in the derived type.

This PR has come about from trying to port existing web applications using this approach to use 3.0 in advance of its release where they didn't work properly anymore because of the hosting changes.

martincostello: If it's useful to you, having done some refactoring in my master branch already to make relevant updates for 2.2.x to keep the diff size down, a branch with the changes I've had to make to one of my applications that illustrates the workarounds I've used with preview 2 to keep things working with 3.0 is here: https://github.com/martincostello/alexa-london-travel-site/pull/255/files

Tratcher:
@javiercn it's an expected change when we moved to generic host. It wasn't clear how many people used that code path so we hadn't worked on mitigating it yet.

martincostello:
Any further thoughts on this so I can update the branch for the code rearrangement in master since it was opened?

javiercn:
@Tratcher Could we get the ServiceProvider fixed?

@martincostello Based on what you mention it seems to me that what you are most interested in is the ability to find the content root for the app. I would suggest a couple of things. If you are writing E2E tests I suggest you avoid using WAF all together as it doesn't buy you much.

For setting up the app for tests I would simply create a small base class or small fixture targeted for your app specifically. As per the content root, I would simply reference it directly (relative to the app) or using
`builder.UseSolutionRelativeContentRoot()` to make it more straightforward.

WAF does things in a very complicated way because it aims to work out of the box and doesn't have any knowledge of the app. You don't have that restriction so my recommendation would be to keep it simple.

As an aside, I've been working lately on some E2E testing infrastructure changes for our own E2E tests and there is something that would probably would work better for you and that includes support for finding out the content roots without guessing.

https://github.com/aspnet/AspNetCore/pull/8188/files?utf8=%E2%9C%93&diff=split#diff-c9e49b86b2dd11680a0d4a3fdc22e802R33

https://github.com/aspnet/AspNetCore/pull/8188/files?utf8=%E2%9C%93&diff=split#diff-26eb065f5048f1cf055179dceb6df3a4R40

With that, your fixture simply becomes

```csharp
public class ServerFactory 
{
   public ServerFactory()
   {
        var builder = Program.CreateHostBuilder(args);
       builder.UseContentRoot(FindProject(typeof(Program).Assembly.FullName));
       _host = builder.Build();
   }

   public string FindProject(string projectName)
   {
           return typeof(ServerFactory).Assembly.GetCustomAttributes<AssemblyMetadataAttribute>() 
               .Single(m => m.Key == $"TestAssemblyApplication[{projectName}]")).Value;
   }

   public void Dispose()
   {
        _host.Stop();
        _host.Dispose();
   }
}
```

Would that work for you?

martincostello:
@javiercn Thanks - I'll have a play around with something like that in my 3.0 branch once preview 3 drops and see how I get on.

Tratcher:
@javiercn this fix in this PR (https://github.com/aspnet/AspNetCore/pull/7414/files#diff-5f39437e8fa98005e4490f30c05e6b1eR86) is correct. It's up to you if you want to take the GetContentRootPath change as well.

javiercn:
@Tratcher This breaks existing customers code. I would like to avoid forcing customers to change their code if we can. Why not handle it in TestServer directly?

Tratcher:
It doesn't break existing code, users have to opt into the new IHostBuilder. Asking them to make a few other changes is not unreasonable. 

> this was possible through the access chain factory.Server.Host.Services

The Host property type is IWebHostBuilder and changing that to IHostBuilder would break existing apps. The new Services property will work for both scenarios.

martincostello:
If I rebase this onto `master` and just change the PR to be e1b0564767f6a1dee683bb65a7dc8b455c58ccda, would this be merged?

javiercn:
Yes, and thanks for following through!

martincostello:
Cool, thanks - I'll update the PR later today.

javiercn:
Will merge once the build is happy

javiercn (approved):

martincostello:
Looks like the code check build has failed - is there something I need to update because I've added a new public member?

javiercn:
Ah yes. You need to follow the instructions in /docs to update the reference assembly I believe.

javiercn:
https://github.com/aspnet/AspNetCore/blob/master/docs/ReferenceAssemblies.md#when-changing-public-api

martincostello:
Thanks - I'll fix that now.

javiercn:
Ah, nevermind. It must have been the compilation error. You can ignore the instructions above.

martincostello:
Looks like it was both! I'm just waiting for Visual Studio Enterprise 2019 to install so the generation tool works, as it doesn't seem to like 2019 Professional being installed.

javiercn:
There is a script to make sure your VS has all the prerequisites installed. It should work with professional afaik.

martincostello:
I'm guessing I'm missing a few prerequisites (I'd guess C++ stuff for the IIS hosting module), but it's installing them into a new VS Enterprise installation instead of adding them to VS Professional.

image

image

javiercn:
I see. Well, the script will do the right thing I think.

javiercn:
If for some reason it doesn't work, I'll figure out if I can run the tool and put a commit on top of the PR or create a patch you can apply.

SteveSandersonMS:
@martincostello I'm pretty sure you didn't mean to delete those `.csproj` files in your last PR, but of course this can't be merged until that is reverted 😄

javiercn:
lol. That is true. Thanks @SteveSandersonMS for catching that!

martincostello:
I misread the instructions and ran `dotnet msbuild /t:GenerateReferenceSource` from `src/Mvc` instead of `src/Mvc/Mvc.Testing/src`, and the result of that included the deletes.

I'll go back a commit and just commit the relevant file.

javiercn:
That looks better

martincostello:
Looks OK now apart from two (hopefully) unrelated test failures.

martincostello:
Got there eventually 😎

javiercn (approved):

javiercn:
Merged. Thanks for following through!

PR #7415, opened 2019-02-09
 Don't write empty data for Flush

benaadams:
Can skip getting Span etc

Reuse BufferWriter for `WriteCurrentMemoryToPipeWriter`

Ignore whitespace change https://github.com/aspnet/AspNetCore/pull/7415/files?w=1

Plaintext pipelined; left current; right PR:

image

/cc @jkotalik @davidfowl

### DIFF
--- a/src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
+++ b/src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
@@ -96,7 +96,50 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 
         public ValueTask<FlushResult> FlushAsync(CancellationToken cancellationToken = default)
         {
-            return WriteAsync(Constants.EmptyData, cancellationToken);
+            lock (_contextLock)
+            {
+                if (_pipeWriterCompleted)
+                {
+                    return default;
+                }
+
+                if (_autoChunk)
+                {
+                    if (_advancedBytesForChunk > 0)
+                    {
+                        // If there is data that was chunked before flushing (ex someone did GetMemory->Advance->FlushAsync)
+                        // make sure to write whatever was advanced first
+                        return FlushAsyncChunked(this, cancellationToken);
+                    }
+                    else
+                    {
+                        // If there is an empty write, we still need to update the current chunk
+                        _currentChunkMemoryUpdated = false;
+                    }
+                }
+
+                var bytesWritten = _unflushedBytes;
+                _unflushedBytes = 0;
+
+                return _flusher.FlushAsync(_minResponseDataRateFeature.MinDataRate, bytesWritten, this, cancellationToken);
+            }
+
+            ValueTask<FlushResult> FlushAsyncChunked(Http1OutputProducer producer, CancellationToken token)
+            {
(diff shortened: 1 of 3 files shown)


benaadams:
Broke the `ConnectionNotClosedWhenClientSatisfiesMinimumDataRateGivenXxx` tests

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
```diff
@@ -96,9 +96,53 @@ public ValueTask<FlushResult> WriteStreamSuffixAsync()
...
+                writer.Commit();
+
+                var bytesWritten = producer._unflushedBytes + writer.BytesCommitted;
+                producer._unflushedBytes = 0;
+
+                return producer._flusher.FlushAsync(producer._minResponseDataRateFeature.MinDataRate, bytesWritten, producer, token);
+            }
         }
 
+
```
benaadams: nit empty line

benaadams:
Unrelated ci error

IIS.Tests.Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests.ClientDisconnectTests.ReaderThrowsCancelledException
```
Error message
Assert.IsType() Failure
Expected: System.OperationCanceledException
Actual: Microsoft.AspNetCore.Connections.ConnectionResetException

Stack trace
   at Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests.ClientDisconnectTests.ReaderThrowsCancelledException() in /_/src/Servers/IIS/IIS/test/IIS.Tests/ClientDisconnectTests.cs:line 206
```

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
```diff
@@ -183,19 +226,35 @@ public ValueTask<FlushResult> WriteChunkAsync(ReadOnlySpan<byte> buffer, Cancell
...
+                        writer.WriteEndChunkBytes();
+                    }
+
+                    writer.Commit();
                     _unflushedBytes += writer.BytesCommitted;
                 }
+                else
+                {
+                    // If there is an empty write, we still need to update the current chunk
+                    _currentChunkMemoryUpdated = false;
```
jkotalik: _currentChunkMemoryUpdated is always false by the end of this method. I'd do the following:
- remove setting _currentChunkMemoryUpdated = false in WriteCurrentMemoryToPipeWriter()
- set _currentChunkMemoryUpdated  to false in WriteChunkAsync and FlushAsync

jkotalik (approved):
Changes look good besides a nit.

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
benaadams: WriteChunkAsync calls FlushAsync, so only need to do it there?

jkotalik: Yeah I think you are right.

benaadams: Done

PR #7422, opened 2019-02-10
Use C#7.3's ReadOnlySpan<byte> optimization for static data

benaadams:
Better write-up than the roslyn issue  https://vcsjones.com/2019/02/01/csharp-readonly-span-bytes-static/

/cc @vcsjones

### DIFF
--- a/src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
+++ b/src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
@@ -16,9 +16,13 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
     public class Http1OutputProducer : IHttpOutputProducer, IHttpOutputAborter, IDisposable
     {
         // Use C#7.3's ReadOnlySpan<byte> optimization for static data https://vcsjones.com/2019/02/01/csharp-readonly-span-bytes-static/
+        // "HTTP/1.1 100 Continue\r\n\r\n"
         private static ReadOnlySpan<byte> ContinueBytes => new byte[] { (byte)'H', (byte)'T', (byte)'T', (byte)'P', (byte)'/', (byte)'1', (byte)'.', (byte)'1', (byte)' ', (byte)'1', (byte)'0', (byte)'0', (byte)' ', (byte)'C', (byte)'o', (byte)'n', (byte)'t', (byte)'i', (byte)'n', (byte)'u', (byte)'e', (byte)'\r', (byte)'\n', (byte)'\r', (byte)'\n' };
+        // "HTTP/1.1 "
         private static ReadOnlySpan<byte> HttpVersion11Bytes => new byte[] { (byte)'H', (byte)'T', (byte)'T', (byte)'P', (byte)'/', (byte)'1', (byte)'.', (byte)'1', (byte)' ' };
+        // "\r\n\r\n"
         private static ReadOnlySpan<byte> EndHeadersBytes => new byte[] { (byte)'\r', (byte)'\n', (byte)'\r', (byte)'\n' };
+        // "0\r\n\r\n"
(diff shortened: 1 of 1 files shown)


davidfowl (approved):

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
```diff
@@ -16,10 +15,11 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 {
     public class Http1OutputProducer : IHttpOutputProducer, IHttpOutputAborter, IDisposable
     {
-        private static readonly ReadOnlyMemory<byte> _continueBytes = new ReadOnlyMemory<byte>(Encoding.ASCII.GetBytes("HTTP/1.1 100 Continue\r\n\r\n"));
-        private static readonly byte[] _bytesHttpVersion11 = Encoding.ASCII.GetBytes("HTTP/1.1 ");
-        private static readonly byte[] _bytesEndHeaders = Encoding.ASCII.GetBytes("\r\n\r\n");
-        private static readonly ReadOnlyMemory<byte> _endChunkedResponseBytes = new ReadOnlyMemory<byte>(Encoding.ASCII.GetBytes("0\r\n\r\n"));
+        // Use C#7.3's ReadOnlySpan<byte> optimization for static data https://vcsjones.com/2019/02/01/csharp-readonly-span-bytes-static/
+        private static ReadOnlySpan<byte> ContinueBytes => new byte[] { (byte)'H', (byte)'T', (byte)'T', (byte)'P', (byte)'/', (byte)'1', (byte)'.', (byte)'1', (byte)' ', (byte)'1', (byte)'0', (byte)'0', (byte)' ', (byte)'C', (byte)'o', (byte)'n', (byte)'t', (byte)'i', (byte)'n', (byte)'u', (byte)'e', (byte)'\r', (byte)'\n', (byte)'\r', (byte)'\n' };
```
davidfowl: I'd put the text in a comment about since these are so hard to read

benaadams: Added comments

BrennanConroy:
@davidfowl Should we do this for SignalR too?

davidfowl:
@BrennanConroy yep. How many static arrays do we have? It would be nice to autogen them but if we don't have enough then maybe it doesn't matter.

BrennanConroy:
https://github.com/aspnet/AspNetCore/blob/master/src/SignalR/common/Http.Connections.Common/src/NegotiateProtocol.cs#L17
https://github.com/aspnet/AspNetCore/blob/master/src/SignalR/common/SignalR.Common/src/Protocol/HandshakeProtocol.cs#L21
And we'll have more when we get System.Text.Json protocol https://github.com/aspnet/AspNetCore/blob/master/src/SignalR/common/Protocols.NewtonsoftJson/src/Protocol/NewtonsoftJsonHubProtocol.cs#L24

benaadams:
Ideally could get the C# compiler to also recognise this pattern:
```csharp
static ReadOnlySpan<byte> ConnectionIdPropertyNameBytes => Encoding.UTF8.GetBytes("connectionId");
```
As its pretty impenetrable to cast each char to byte individually

benaadams:
Added comment https://github.com/dotnet/roslyn/pull/24621#issuecomment-462202923

benaadams:
Added comment https://github.com/dotnet/roslyn/pull/24621#issuecomment-462202923 and raised issue https://github.com/dotnet/csharplang/issues/2212

jkotalik (approved):

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
Tratcher: Line wraps?

PR #7424, opened 2019-02-10
Remove unnecessary Flush

benaadams:
When the body has been written to the headers are flushed so it doesn't need a second flush at end; so set `_requestProcessingStatus = RequestProcessingStatus.HeadersFlushed`

Currently handled for `_autoChunk` but not for `ContentLength`

image

/cc @jkotalik @davidfowl

### DIFF
--- a/src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
+++ b/src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
@@ -794,9 +794,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
         }
 
         [MethodImpl(MethodImplOptions.NoInlining)]
-        private ValueTask<FlushResult> FlushAsyncInternal(CancellationToken cancellationToken)
+        private Task FlushAsyncInternal()
         {
-            return Output.FlushAsync(cancellationToken);
+            return Output.FlushAsync(default).AsTask();
         }
 
         private void VerifyAndUpdateWrite(int count)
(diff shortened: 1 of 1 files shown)


benaadams:
Probably can use the same BufferWriter for first write and headers https://github.com/aspnet/AspNetCore/issues/7429

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
```diff
@@ -1011,6 +995,7 @@ private Task WriteSuffix()
 
             if (!HasFlushedHeaders)
             {
+                _requestProcessingStatus = RequestProcessingStatus.HeadersFlushed;
                 return FlushAsyncInternal(default).AsTask();
```
jkotalik: Remove FlushAsyncInternal entirely and just call Output.FlushAsync(ct);

jkotalik (approved):

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
benaadams: Done for all the other places; changed this to 
```csharp
return FlushAsyncInternal();
```
Then had `FlushAsyncInternal` do the `.AsTask()` since its no-inline so it doesn't pull the `ValueTask<FlushResult>` into this stackframe as it should rarely be called e.g.
```csharp
private Task FlushAsyncInternal()
{
    return Output.FlushAsync(default).AsTask();
}
```

Issue #7418, opened 2019-02-09
Blazor onpointerdown event gives error


Lupusa87:
### Describe the bug
When onpointerdown event fires I get error.

### To Reproduce
1. In Blazor 0.8.0 standalone app I dynamically create element, register pointer event;
2. `builder.AddAttribute(k++, "onpointerdown", OnPointerDown);`
3. `public void OnPointerDown(UIPointerEventArgs e) {};`
4. When I click on element I get this error:
System.ArgumentException: Object of type 'System.Int64' cannot be converted to type 'System.String'.

### Expected behavior
It should invoke OnPointerDown method with proper UIPointerEventArgs.

### Screenshots
![untitled](https://user-images.githubusercontent.com/40779557/52525497-a6581980-2c78-11e9-8fc4-358647df2bdb.png)

### Additional context
I guess it happens with all pointer events.

Lupusa87:
[EventHandlers.cs](https://github.com/aspnet/AspNetCore/blob/master/src/Components/Components/src/EventHandlers.cs#L72) is showing:

`[EventHandler("pointerdown", typeof(UIPointerEventArgs))]`

not **on**pointerdown


but if I do `builder.AddAttribute(k++, "pointerdown", OnPointerDown);` and click element nothing happens.

So another question - which is correct pointerdown or **on**pointerdown?

Lupusa87:
I found where is problem.

According to [this](https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/pointerId) PointerID is number.

"pointerEvent.pointerId - The pointer event's unique identifier number."


[Here](https://github.com/aspnet/AspNetCore/blob/5a70f5312fb57fc3788e5af56a99e7b43761e195/src/Components/Browser.JS/src/Rendering/EventForDotNet.ts#L303) in EventForDotNet.ts we have also number.

```
interface UIPointerEventArgs extends UIMouseEventArgs {
  pointerId: number;
  width: number;
  height: number;
  pressure: number;
  tiltX: number;
  tiltY: number;
  pointerType: string;
  isPrimary: boolean;
}
```

But [here](https://github.com/aspnet/AspNetCore/blob/5a70f5312fb57fc3788e5af56a99e7b43761e195/src/Components/Components/src/UIEventArgs.cs#L333) we have string.

```
public class UIPointerEventArgs : UIMouseEventArgs
    {
        /// <summary>
        /// A unique identifier for the pointer causing the event.
        /// </summary>
[...]

PR #7426, opened 2019-02-10
PointerID type string replaced to long

Lupusa87:
Summary of the changes
 - PointerID should be long but was string
 - It was giving error - System.ArgumentException: Object of type 'System.Int64' cannot be converted to type 'System.String'.

Addresses #7418

### DIFF
--- a/src/Components/Components/src/UIEventArgs.cs
+++ b/src/Components/Components/src/UIEventArgs.cs
@@ -330,7 +330,7 @@ namespace Microsoft.AspNetCore.Components
         /// <summary>
         /// A unique identifier for the pointer causing the event.
         /// </summary>
-        public string PointerId { get; set; }
+        public long PointerId { get; set; }
 
         /// <summary>
         /// The width (magnitude on the X axis), in CSS pixels, of the contact geometry of the pointer.


Lupusa87:
There is updated only one word, does it take so long to review this?

SteveSandersonMS:
Sorry, but someone does have to find the time to review what scenarios this affects and whether the fix is correct, and whether it might possibly introduce new failure cases. The fact that the diff is small doesn’t avoid this!

We appreciate the contribution of course.

javiercn:
@SteveSandersonMS I'm probably not the best person to review this as I'm not super familiar with the area, but I'm happy to learn.

[EventHandlers.cs](https://github.com/aspnet/AspNetCore/blob/master/src/Components/Components/src/EventHandlers.cs#L72) is showing:

`[EventHandler("pointerdown", typeof(UIPointerEventArgs))]`

not **on**pointerdown

but if I do `builder.AddAttribute(k++, "pointerdown", OnPointerDown);` and click element nothing happens.

**which is correct pointerdown or **on**pointerdown?**
My understanding is that pointerdown is correct, and the issue is just about the type

**Do we have any test for pointer events?**

Seems like we don't. @SteveSandersonMS If you are ok with the change (I am) we can get this in. I'll file an issue to go and add E2E tests for all our events as I believe this is something we should have.  Thoughts?

javiercn (approved):

SteveSandersonMS (approved):
Thanks for investigating. It looks like we have several issues:

1. Some of the `[EventHandler]` entries starting at https://github.com/aspnet/AspNetCore/blob/master/src/Components/Components/src/EventHandlers.cs#L72 are incorrect, because they lack the necessary `on` prefix from the event name. This means you can't register pointer events from normal Razor code and have to do the `RenderTreeBuilder` workaround that @Lupusa87 is using in the repro code.
2. The type of `PointerID` is defined as `string` but should be a numeric type (`long` would be fine, though so would `int`).
3. We don't have E2E tests for all possible event types. To some extent this is deliberate because it's very expensive to create E2E tests for all possible event types (and Selenium can't necessarily simulate all of them, such as network state changes), however the above failures show that perhaps we should cover more of the cases.

This PR addresses item 2.

@javiercn I agree with your summary, so please do merge this if you're happy. If you are also happy to file issues for the other points that's great (or if you prefer me to do so please just say).

javiercn:
> 1. Some of the `[EventHandler]` entries starting at https://github.com/aspnet/AspNetCore/blob/master/src/Components/Components/src/EventHandlers.cs#L72 are incorrect, because they lack the necessary `on` prefix from the event name. This means you can't register pointer events from normal Razor code and have to do the `RenderTreeBuilder` workaround that @Lupusa87 is using in the repro code.

Filed https://github.com/aspnet/AspNetCore/issues/7570

> 2. The type of `PointerID` is defined as `string` but should be a numeric type (`long` would be fine, though so would `int`).

Will be fixed by this PR. So I'm merging.

> 3. We don't have E2E tests for all possible event types. To some extent this is deliberate because it's very expensive to create E2E tests for all possible event types (and Selenium can't necessarily simulate all of them, such as network state changes), however the above failures show that perhaps we should cover more of the cases.

Filed https://github.com/aspnet/AspNetCore/issues/7566

Lupusa87:
Thank you for merging this and for following great actions.

PR #7431, opened 2019-02-10
Make Hosting.sln compile in vs

davidfowl:
Added projects until it compiled


Tratcher:
2017 or 2019? This isn't supposed to be necessary in 2019.

Tratcher (approved):

davidfowl:
It's required for both, it doesn't compile in any version i tried.

PR #7432, opened 2019-02-10
Use MemoryWriter for Platform

benaadams:
Cache Memory across the awaits rather than going back to the PipeWriter until the Memory runs out


davidfowl (changes requested):
Closing in favor of https://github.com/dotnet/corefx/pull/35216

Issue #5055, opened 2018-12-14
Generating URL with page parameter fails with new endpoint routing (regression from 2.1)


poke:
This appears to be a regression due to the new endpoint routing system in 2.2. As per aspnet/Mvc#7876, using parameter names `action` and `page` is somewhat difficult, however using a `page` parameter inside an MVC action or using an `action` parameter inside of a Razor page does work fine. At least it did.

With endpoint routing activated (through the MVC compatibility switch `CompatibilityVersion.Version_2_2`), generating an action link that has a `page` parameter will now fail **iff** Razor pages exist in the same project.

#### Without Razor pages
1. Create new 2.2 application from standard MVC template.
2. Add an action to the `HomeController`:
    ```c#
    public IActionResult Test(int page)
    {
        return Json(new
        {
            Url = Url.Action(nameof(Test), new { page = 123 }),
            Page = page,
        });
    }
    ```
3. Launch the application and open `http://localhost:5000/Home/Test?page=4`
4. Observe the result which shows a generated URL `/Home/Test?page=123`.

#### Add Razor pages
5. Now, create a Razor page at `Pages/Foo.cshtml` with the contents `@page`.
6. Launch the application and open the previous URL again.
7. Observe that the generated URL is now `null`.

#### Disable endpoint routing
8. Disable endpoint routing by changing the `AddMvc()` call in `Startup` to set the compatibility level switch back to 2.1:
    ```c#
    services.AddMvc().SetCompatibilityVersion(CompatibilityVersion.Version_2_1);
    ```
9. Launch the application and open the previous URL again.
10. Observe that the URL is generated properly again.

---

As you can see, endpoint routing breaks the URL generation here although the target URL exists just fine and although it also can be used properly the whole time (note that the JSON response always includes the passed `page` value). The mere presence of Razor pages should not make the URL generation treat any `page` parameter in a special way here.

nickalbrecht:
I ran into this yesterday/today. Thanks for taking the time to flesh out the details and submit the bug @poke. My short term solution was to rename my action parameters to something like `pageNum` instead of `page`, But I'd prefer to see this get fixed so it wasn't so aggressive with regards to preventing the use of `page` as an action parameter when trying to generate URLs.

WistfulWolf:
Just encountered this issue in .NET Core 2.1. UrlHelper.Link when adding a parameter with name "page" completely ignores it. Tested on clean MVC project with an API controller.

Api Controller class:
```
[...]

PR #7437, opened 2019-02-10
Fix #4212 area ambient value page->controller

### LINKED ISSUE #4212: Link generation behaviour change: page to controller in area
Linking from a page in an area to a controller in the same area, without specifying the area is changing.

* Succeeded in 2.1
* Succeeded in 2.2 (unintentionally)
* Fails in 3.0 (current behaviour)

We could make 3.0 succeed with a moderate amount of additional work. The question is whether anyone is doing this and if it is important to customers.

This issue is something to evaluate near the end of 3.0 to make a decision.

Affected test: PagesInAreas_CanGenerateLinksToControllersAndPages
```cs
<a asp-controller="Home" asp-action="Index">Link to area action</a>
```

PR that is changing this - https://github.com/aspnet/Routing/pull/914

rynowak:
This change enhances our ambient value logic to also deal with required
values. In 2.2 we introduced a 'required values' semantic to allow route
values to appear "to the left" of a route pattern for the purpose of
ambient values copying. This is a complicated way of saying "when you
like to a different endpoint then discard the ambient values".

What we didn't consider is that some ambient values are special (like
area). So basically, we'll allow an ambient value to be used if it's
part of the required values - even if we've already decided to discard
the ambient values.

This is a pretty surgical fix and only affected the desired scenario
based on tests.

-----

I also removed an optimization that I think is broken. I put an earlier
optimization in place that attempted to count ambient values as they
were "seen" to try and avoid some extra copying. This copying loop has a
cost even if it no-ops which is what I was trying to prevent.

Unfortunately since we added 'required values' - it's now possible for
an ambient value to be double-counted, which makes this optimization
incorrect.

### DIFF
--- a/src/Http/Routing/src/Template/TemplateBinder.cs
+++ b/src/Http/Routing/src/Template/TemplateBinder.cs
@@ -118,10 +118,9 @@ namespace Microsoft.AspNetCore.Routing.Template
             var slots = new KeyValuePair<string, object>[_slots.Length];
             Array.Copy(_slots, 0, slots, 0, slots.Length);
 
-            // Keeping track of the number of 'values' and 'ambient values' we've processed can be used to avoid doing
+            // Keeping track of the number of 'values' we've processed can be used to avoid doing
             // some expensive 'merge' operations later.
             var valueProcessedCount = 0;
-            var ambientValueProcessedCount = 0;
 
             // Start by copying all of the values out of the 'values' and into the slots. There's no success
             // case where we *don't* use all of the 'values' so there's no reason not to do this up front
@@ -167,11 +166,6 @@ namespace Microsoft.AspNetCore.Routing.Template
                     if (ambientValues == null || !ambientValues.TryGetValue(key, out var ambientValue))
                     {
                         ambientValue = null;
-                    }
-                    else
-                    {
-                        // Track the count of processed ambient values - this allows a fast path later.
-                        ambientValueProcessedCount++;
                     }
 
                     // For now, only check ambient values with required values that don't have a parameter
@@ -226,15 +220,10 @@ namespace Microsoft.AspNetCore.Routing.Template
 
                 var parameter = parameters[i];
 
+                // We are copying **all** ambient values
                 if (copyAmbientValues)
                 {
                     hasAmbientValue = ambientValues != null && ambientValues.TryGetValue(key, out ambientValue);
-                    if (hasAmbientValue)
-                    {
-                        // Track the count of processed ambient values - this allows a fast path later.
-                        ambientValueProcessedCount++;
-                    }
-
                     if (hasExplicitValue && hasAmbientValue && !RoutePartsEqual(ambientValue, value))
                     {
                         // Stop copying current values when we find one that doesn't match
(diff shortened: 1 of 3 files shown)


src/Http/Routing/src/Template/TemplateBinder.cs
```diff
@@ -118,10 +118,9 @@ public TemplateValuesResult GetValues(RouteValueDictionary ambientValues, RouteV
             var slots = new KeyValuePair<string, object>[_slots.Length];
             Array.Copy(_slots, 0, slots, 0, slots.Length);
 
-            // Keeping track of the number of 'values' and 'ambient values' we've processed can be used to avoid doing
+            // Keeping track of the number of 'values' we've processed can be used to avoid doing
             // some expensive 'merge' operations later.
             var valueProcessedCount = 0;
-            var ambientValueProcessedCount = 0;
```
rynowak: Removed all usage of this because we can't count this accurately in practice.

src/Mvc/test/Microsoft.AspNetCore.Mvc.FunctionalTests/RazorPagesWithBasePathTest.cs
```diff
@@ -323,7 +323,7 @@ public async Task PagesInAreas_CanGenerateLinksToControllersAndPages()
             var expected =
 @"<a href=""/Accounts/Manage/RenderPartials"">Link inside area</a>
 <a href=""/Products/List/old/20"">Link to external area</a>
-<a href="""">Link to area action</a>
+<a href=""/Accounts"">Link to area action</a>
```
rynowak: This was the original regression.

src/Http/Routing/src/Template/TemplateBinder.cs
```diff
@@ -226,15 +220,10 @@ public TemplateValuesResult GetValues(RouteValueDictionary ambientValues, RouteV
 
                 var parameter = parameters[i];
 
+                // We are copying **all** ambient values OR
```
pranavkm: OR?

JamesNK:
https://github.com/aspnet/AspNetCore/issues/5055 is also a link generation issue with required values. Are they the same issue?

rynowak:
> #5055 is also a link generation issue with required values. Are they the same issue?

Nope. That issue is much much much more frightening.

src/Http/Routing/src/Template/TemplateBinder.cs
```diff
@@ -261,6 +250,29 @@ public TemplateValuesResult GetValues(RouteValueDictionary ambientValues, RouteV
                     }
                 }
 
+                // This might an ambient value that matches a required value. We want to use these even if we're
```
JamesNK: Clean up comment

JamesNK (approved):

src/Http/Routing/src/Template/TemplateBinder.cs
rynowak: Sorry can you make a suggestion? I don't get what you're asking me to change.

JamesNK:
```suggestion
                // This might be an ambient value that matches a required value. We want to use these even if we're
```

PR #7440, opened 2019-02-11
Remove PlatformBenchmarks from sln

davidfowl:
Since I like solutions to compile when I'm devving 😄


Tratcher (approved):

benaadams:
Should the project be moved to aspnet/Benchmarks with the other benchmarks so it doesn't have an uncompilable .proj here due to the restrictions applied to it? (e.g. no PackageRefeneces)

/cc @sebastienros

halter73 (approved):

PR #7441, opened 2019-02-11
Fix #5055 use of page parameter

### LINKED ISSUE #5055: Generating URL with page parameter fails with new endpoint routing (regression from 2.1)

rynowak:
This change introduces some new heuristics to make it easier to use
'page' or other resevered parameter names as parameters in URL
generation.

--

The main change here is to allow the link generation tree to *ignore* a
value passed in to URL generation when it conflicts with an endpoint's
required values.

The main concern of this feature area is "how do we tell whether you are
linking to an action or a page?". Routing attempts to do the right thing
will requiring very little from the user in terms of expressing intent.

In this case, we try to tell the difference between an attempt to generate
a link to an action due to the presence of the 'action' parameter and
absence of the 'page' parameter. This obviously doesn't work when you
want to use 'page' as a non-reserved parameter in an action. The same
case occurs for pages, but users are already used to the idea that
'action' is a reserved word in MVC.

We can loosen this restriction when the value that's supplied for 'page'
is known not to be any existing value of the 'page' route value.

This approach seems somewhat reasonable but has many of the problems inherent
to this area. When it fails (the value you want to use for 'page' causes
a conflict) - it's going to be esoteric and hard to understand.

### DIFF
--- a/src/Http/Routing/src/Internal/LinkGenerationDecisionTree.cs
+++ b/src/Http/Routing/src/Internal/LinkGenerationDecisionTree.cs
@@ -21,12 +21,29 @@ namespace Microsoft.AspNetCore.Routing.Internal
         private static readonly RouteValueDictionary EmptyAmbientValues = new RouteValueDictionary();
 
         private readonly DecisionTreeNode<OutboundMatch> _root;
+        private readonly Dictionary<string, HashSet<object>> _knownValues;
 
         public LinkGenerationDecisionTree(IReadOnlyList<OutboundMatch> entries)
         {
             _root = DecisionTreeBuilder<OutboundMatch>.GenerateTree(
                 entries,
                 new OutboundMatchClassifier());
+
+            _knownValues = new Dictionary<string, HashSet<object>>(StringComparer.OrdinalIgnoreCase);
+            for (var i = 0; i < entries.Count; i++)
+            {
+                var entry = entries[i];
+                foreach (var kvp in entry.Entry.RequiredLinkValues)
+                {
+                    if (!_knownValues.TryGetValue(kvp.Key, out var values))
+                    {
+                        values = new HashSet<object>(RouteValueEqualityComparer.Default);
+                        _knownValues.Add(kvp.Key, values);
+                    }
+
+                    values.Add(kvp.Value ?? string.Empty);
+                }
+            }
         }
 
         public IList<OutboundMatchResult> GetMatches(RouteValueDictionary values, RouteValueDictionary ambientValues)
@@ -92,6 +109,19 @@ namespace Microsoft.AspNetCore.Routing.Internal
                     if (criterion.Branches.TryGetValue(value ?? string.Empty, out var branch))
                     {
                         Walk(results, values, ambientValues, branch, isFallbackPath);
+                    }
+                    else
+                    {
+                        // If an explicitly specified value doesn't match any branch, then speculatively walk the
+                        // "null" path if the value doesn't match any known value.
+                        //
+                        // This can happen when linking from a page <-> action. We want to be
+                        // able to use "page" and "action" as normal route parameters.
+                        var knownValues = _knownValues[key];
+                        if (!knownValues.Contains(value ?? string.Empty) && criterion.Branches.TryGetValue(string.Empty, out branch))
+                        {
+                            Walk(results, values, ambientValues, branch, isFallbackPath: true);
+                        }
                     }
                 }
                 else
--- a/src/Http/Routing/test/UnitTests/Internal/LinkGenerationDecisionTreeTest.cs
+++ b/src/Http/Routing/test/UnitTests/Internal/LinkGenerationDecisionTreeTest.cs
@@ -341,6 +341,385 @@ namespace Microsoft.AspNetCore.Routing.Internal.Routing
         }
 
         [Fact]
+        public void GetMatches_ControllersWithArea_AllValuesExplicit()
+        {
+            // Arrange
+            var entries = new List<OutboundMatch>();
+
+            var entry1 = CreateMatch(new { controller = "Store", action = "Buy", area = (string)null, });
+            entry1.Entry.RouteTemplate = TemplateParser.Parse("a");
+            entries.Add(entry1);
+
+            var entry2 = CreateMatch(new { controller = "Store", action = "Buy", area = "Admin" });
+            entry2.Entry.RouteTemplate = TemplateParser.Parse("b");
+            entries.Add(entry2);
+
+            var tree = new LinkGenerationDecisionTree(entries);
+
+            var context = CreateContext(new { controller = "Store", action = "Buy", area = "Admin" });
+
+            // Act
+            var matches = tree.GetMatches(context.Values, context.AmbientValues).Select(m => m.Match).ToList();
+
+            // Assert
+            // Assert
+            Assert.Collection(
+                matches,
+                m => { Assert.Same(entry2, m); });
+        }
+
+        [Fact]
+        public void GetMatches_ControllersWithArea_SomeValuesAmbient()
+        {
+            // Arrange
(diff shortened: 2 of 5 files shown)


src/Http/Routing/src/Internal/LinkGenerationDecisionTree.cs
```diff
@@ -91,7 +108,20 @@ public IList<OutboundMatchResult> GetMatches(RouteValueDictionary values, RouteV
...
+                        // If an explicitly specified value doesn't match any branch, then speculatively walk the
+                        // "null" path if the value doesn't match any known value.
+                        //
+                        // This can happen when linking from a page <-> action. We want to be
+                        // able to use "page" and "action" as normal route parameters.
+                        var knownValues = _knownValues[key];
+                        if (!knownValues.Contains(value ?? string.Empty) && criterion.Branches.TryGetValue(string.Empty, out branch))
+                        {
+                            Walk(results, values, ambientValues, branch, quality + 0.01m);
+                        }
```
rynowak: This is the gist of the fix. Each approach has problems. This kind of feature as at-odds with the algorithmic requirements of routing so it's going to require esoteric heuristics to improve this.

=====

A few things we could try:

1. (in this pr) check 'page' against known values to prevent ambiguity
1. allow the tree to ignore 'page' and rely on ordering - changes the behavior when you have a link fail to generate
1. remove 'page' from required values of an action

I went with approach once because I don't like the drawbacks of 2. I don't want to impact the majority cases by changing how it behaves when you fail to generate a link when linking from a page to an action.

Imagine you're trying to like an action: `{ action = "Buy", controller = "Store" }` with template `/Store/Buy/{productId?}`. you are on page `/Foo/Bar` and you supply the route values `{ action = "Buy", controller = "Store" }` - but no productId. If we did option 2, then both your action *and your current page* would be candidates. Since you failed to supply productId, the action would fail to generate, and you'd get `/Foo/Bar?action=Buy&controller=Store` instead.

This is why I did option 1. 

Option 3 is totally busto because it makes ambient value invalidation stop working when you like from a page to an action. Yuck.

src/Http/Routing/src/Internal/LinkGenerationDecisionTree.cs
```diff
@@ -139,6 +169,12 @@ private class OutboundMatchResultComparer : IComparer<OutboundMatchResult>
 
             public int Compare(OutboundMatchResult x, OutboundMatchResult y)
             {
+                if (x.Quality != y.Quality)
+                {
+                    // Reversed because higher is better
+                    return y.Quality.CompareTo(x.Quality);
+                }
```
rynowak: I made this change while pursuing option 2, I don't think it's needed for anything so I could revert it once we decide what to do.

src/Http/Routing/test/UnitTests/Internal/LinkGenerationDecisionTreeTest.cs
```diff
@@ -340,6 +340,385 @@ public void SelectMultipleEntries_BothMatch_OrderedByTemplate()
...
+
+            var context = CreateContext(new { page = "/Store/Buy", action = "Index", }, new { controller = "Home", action = "Index", page = "16", });
+
+            // Act
+            var matches = tree.GetMatches(context.Values, context.AmbientValues).Select(m => m.Match).ToList();
+
+            // Assert
+            Assert.Empty(matches);
+        }
+
```
rynowak: Note: I didn't make any attempt to address the problem of writing a conventional route that uses `{page}` as a parameter. I'm not sure this can really be solved.

src/Http/Routing/src/Internal/LinkGenerationDecisionTree.cs
```diff
@@ -91,7 +108,20 @@ public IList<OutboundMatchResult> GetMatches(RouteValueDictionary values, RouteV
                 {
                     if (criterion.Branches.TryGetValue(value ?? string.Empty, out var branch))
                     {
-                        Walk(results, values, ambientValues, branch, isFallbackPath);
+                        Walk(results, values, ambientValues, branch, quality + 1m);
```
JamesNK: Change the quality modifier values to const fields

src/Http/Routing/src/Internal/LinkGenerationDecisionTree.cs
JamesNK: Makes sense. I think I tried option 3 before realizing I was out of my depth 😄

JamesNK (commented):
Looks good to me.

Add the original issue's scenario as a functional test

src/Http/Routing/src/Internal/LinkGenerationDecisionTree.cs
rynowak: obsolete, they are gone completely

rynowak:
@JamesNK - resolved your feedback, can you take a quick look?

src/Http/Routing/src/Internal/LinkGenerationDecisionTree.cs
```diff
@@ -35,7 +52,7 @@ public IList<OutboundMatchResult> GetMatches(RouteValueDictionary values, RouteV
             if (_root.Matches.Count > 0 || _root.Criteria.Count > 0)
             {
                 var results = new List<OutboundMatchResult>();
-                Walk(results, values, ambientValues ?? EmptyAmbientValues, _root, isFallbackPath: false);
+                Walk(results, values, ambientValues ?? EmptyAmbientValues, _root, isFallbackpath: false);
```
JamesNK: nit: wasn't it more correct before?

JamesNK (approved):

src/Http/Routing/src/Internal/LinkGenerationDecisionTree.cs
rynowak: Oh, this is showing up as a rename. I'll fix it rq.

dougbu (commented):
I don't see anything in this PR that I'm a code owner for. Am I missing something?

rynowak:
@dougbu I dunno mang 🤷‍♂️

rynowak:
@mkArtakMSFT this should be ready now

Issue #7416, opened 2019-02-09
<summary> description of the BlazorOptions class in Razor Components is wrong...


enetstudio:
In the <summary> tag for the BlazorOptions class 
<see cref="ComponentsApplicationBuilderExtensions.UseBlazor(IApplicationBuilder, BlazorOptions)"/>

should be 
<see cref="BlazorApplicationBuilderExtensions.UseBlazor(IApplicationBuilder, BlazorOptions)"/>

Note: I've previously posted issues about similar mistakes, to which the Asp.Net team responded with the suggestion that I should post a Pull request. Unfortunately, I can't post Pull requests. You guys should do it yourselves.

enetstudio:
In the `<summary>` tag for the BlazorOptions class
 `<see cref="ComponentsApplicationBuilderExtensions.UseBlazor(IApplicationBuilder, BlazorOptions)"/>`

should be
`<see cref="BlazorApplicationBuilderExtensions.UseBlazor(IApplicationBuilder, BlazorOptions)"/>`
Note: I've previously posted issues about similar mistakes, to which the Asp.Net team responded with the suggestion that I should post a Pull request. Unfortunately, I can't post Pull requests. You guys should do it yourselves.

mkArtakMSFT:
Thanks for contacting us, @enetstudio.
Sure, we'll fix it.

PR #7443, opened 2019-02-11
Fixed a type reference in a doc comment.

mkArtakMSFT:
Addresses #7416

### DIFF
--- a/src/Components/Server/src/BlazorOptions.cs
+++ b/src/Components/Server/src/BlazorOptions.cs
@@ -4,7 +4,7 @@
 namespace Microsoft.AspNetCore.Builder
 {
     /// <summary>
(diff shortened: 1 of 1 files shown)


javiercn (changes requested):
Thanks for your contribution.
I believe this is on components.server, so it’s likely that the thing incorrectly renamed is the UseBlazor.

We have a separate work item to do names and namespace cleanups as part of productizing components on the server.

I think we won’t be taking this change.

SteveSandersonMS (approved):
The change looks correct to me.

@javiercn You're right that it's weird this Blazor-specific stuff is in the Components.Server project. It should be moved to Blazor.Server. But nonetheless the XML doc comment *should* have the change in this PR, as it correctly represents the current state of the code, *and* the future state after we move `BlazorOptions` and `BlazorApplicationBuilderExtensions` into the Blazor.Server project.

javiercn (approved):
@SteveSandersonMS Is likely right, I just saw Blazor on the .Server and thought it was a missing rename.

PR #7445, opened 2019-02-11
Introduce dynamic endpoints and fix #7011

### LINKED ISSUE #7011: Authorization broken in Razor Pages
### Describe the bug
The Authorize attribute no longer works in Razor Pages applications.

### To Reproduce
1. Make a new Razor Pages project with individual auth
2. Add an `[Authorize]` attribute on the Index page
3. Running the application doesn't redirect to the login page.

### Expected behavior
The page should redirect to the Login page.

### Additional context
This is using build `3.0.0-preview-19073-0424`

Seems like the authorize attribute isn't being exposed as authorize metadata.

cc @rynowak @pranavkm @JamesNK

rynowak:
Introduces a pattern for matcher policy that replaces endpoints during matching.

This allows you to uses the matcher's URL processing, but then run arbitrary other code to choose your own endpoint.


src/Http/Routing/src/IDynamicEndpointMetadata.cs
```diff
@@ -0,0 +1,34 @@
...
+    /// and thus cannot have its characteristics cached.
+    /// </para>
+    /// <para>
+    /// Using dynamic endpoints can be useful because the default matcher implementation does not
+    /// supply extensibility for how URLs are processed. Routing implementations that have dynamic
+    /// behavior can apply their dynamic logic after URL processing, by replacing a endpoints as
+    /// part of a <see cref="CandidateSet"/>.
+    /// </para>
+    /// </remarks>
+    public interface IDynamicEndpointMetadata
```
rynowak: I didn't update any of our policies to respect this. it can be done as needed in the future.

src/Mvc/src/Microsoft.AspNetCore.Mvc.RazorPages/Infrastructure/PageLoaderMatcherPolicy.cs
```diff
@@ -0,0 +1,167 @@
...
+                ref var candidate = ref candidates[i];
+                var endpoint = (RouteEndpoint)candidate.Endpoint;
+
+                var page = endpoint.Metadata.GetMetadata<PageActionDescriptor>();
+                if (page != null)
+                {
+                    var compiled = _loader.Load(page);
+
+                    // This is wrong - we don't want to allocate this per-request. #YOLO
+                    RequestDelegate requestDelegate = (c) =>
```
rynowak: Imagine that line 88-159 are hidden away somewhere else and you get the main idea.

src/Http/Routing/src/IDynamicEndpointMetadata.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        /// <summary>
+        /// Returns a value that indicates whether the associated endpoint has dynamic matching
+        /// behavior.
+        /// </summary>
+        bool IsDynamic { get; }
+    }
```
rynowak: I also considered another subclass of `Endpoint` and you can into trouble. You end up wanting to subclass `RouteEndpoint` instead.

Then the next problem you run into is that `Endpoint`s are **data** - but making a runtime decision about what to do always requires **services**. Putting your services into your data is always a mistake that will cause pain in the future.

From this logic, I ended up thinking of dynamic endpoints as a pattern instead of infrastructure.

rynowak: Imagine that we also used this same pattern to build something like:

```
routes.MapControllers();
routes.MapDynamicControllerRoute("{*path}", async (context, values) =>
{
    // Write your CMS here
   
    return someEndpoint;
});
```

The idea is that we give you a services that you can call to do something like MVC's action selection, and you can return the endpoint that you want (after doing your stuff). This is the pattern that I've seen used most often with `IRouter`.

src/Http/Routing/src/Matching/CandidateSet.cs
```diff
@@ -281,6 +281,57 @@ public void SetValidity(int index, bool value)
...
+        /// <param name="endpoint">
+        /// The <see cref="Endpoint"/> to replace the original <see cref="Endpoint"/> at
+        /// the <paramref name="index"/>. If <paramref name="endpoint"/> the candidate will be marked
+        /// as invalid.
+        /// </param>
+        /// <param name="values">
+        /// The <see cref="RouteValueDictionary"/> to replace the original <see cref="RouteValueDictionary"/> at
+        /// the <paramref name="index"/>.
+        /// </param>
+        public void ReplaceEndpoint(int index, Endpoint endpoint, RouteValueDictionary values)
```
rynowak: I felt that it made most since for this to be a 1:1 replacement - there's no ability for a policy to inject 8 additional endpoints. We can add that in the future if someone really needs it.

JamesNK: Would a non 1:1 replacement be useful for static files? I think that we should try and create a solution that can be used in the future to solve this issue - https://github.com/aspnet/AspNetCore/issues/4215

Background: I updated AuthSamples to use the AuthZ middleware. Being able to map static files in a location to endpoints (or a single dynamic endpoint), and then put AuthZ metadata on the endpoints would have made the sample really clean.

rynowak: > Would a non 1:1 replacement be useful for static files?

No I would imagine static files would be 1:1. So you register one *dynamic* endpoint - and then **run code** to figure out what endpoint you actually want to return - but in the case of static files it would always be one because static files would *unambiguously* determine what file matched.

The use case for multiple endpoints would be so that you can collaborate with another policy that needs to run after yours. Suppose you wanted to filter your results by HTTP method.

rynowak:
@JamesNK @pranavkm - Updated and added some stuff.

So once I got all of this working, I realised that I had a problem because use the invoker factory from page loader creates a DI cycle. This is just a general thing we should try to avoid - ideally MVC doesn't need to construct the entire runtime just to do action discovery.

To resolve this I had to disentangle a few things... ahem removing `MvcActionInvokerProvider` which was a total hack. The nice thing is that we now only interact with `IActionContextAccessor` in one place instead of 3.

src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/Routing/ActionEndpointFactory.cs
```diff
@@ -9,32 +9,25 @@
...
 {
     internal class ActionEndpointFactory
     {
         private readonly RoutePatternTransformer _routePatternTransformer;
-        private readonly MvcEndpointInvokerFactory _invokerFactory;
 
-        public ActionEndpointFactory(
-            RoutePatternTransformer routePatternTransformer,
-            MvcEndpointInvokerFactory invokerFactory)
+        public ActionEndpointFactory(RoutePatternTransformer routePatternTransformer)
```
rynowak: This is a lot better now because we don't need to create a bunch of state/services wicked early. There's an extra DI lookup on the hot path, but if that's a problem we can mitigate that in a few ways.

src/Mvc/src/Microsoft.AspNetCore.Mvc.RazorPages/CompiledPageActionDescriptor.cs
```diff
@@ -59,5 +60,10 @@ public CompiledPageActionDescriptor(PageActionDescriptor actionDescriptor)
         /// Gets or sets the <see cref="TypeInfo"/> of the page.
         /// </summary>
         public TypeInfo PageTypeInfo { get; set; }
+
+        /// <summary>
+        /// Gets or sets the associated <see cref="Endpoint"/> of this page.
+        /// </summary>
+        public Endpoint Endpoint { get; set; }
```
rynowak: The compiled page AD now always has an endpoint as well. This isn't super fun but it seems like the most reasonable thing to do.

src/Mvc/src/Microsoft.AspNetCore.Mvc.RazorPages/Infrastructure/PageActionInvokerProvider.cs
```diff
@@ -57,6 +59,45 @@ internal class PageActionInvokerProvider : IActionInvokerProvider
...
+            IModelMetadataProvider modelMetadataProvider,
+            IModelBinderFactory modelBinderFactory,
+            ITempDataDictionaryFactory tempDataFactory,
+            IOptions<MvcOptions> mvcOptions,
+            IOptions<HtmlHelperOptions> htmlHelperOptions,
+            IPageHandlerMethodSelector selector,
+            DiagnosticListener diagnosticListener,
+            ILoggerFactory loggerFactory,
+            IActionResultTypeMapper mapper,
+            IActionContextAccessor actionContextAccessor)
```
rynowak: yeah I know. Fun stuff

src/Mvc/test/Microsoft.AspNetCore.Mvc.Core.Test/Routing/ActionEndpointFactoryTest.cs
```diff
@@ -32,112 +32,13 @@ public ActionEndpointFactoryTest()
...
         }
 
         internal ActionEndpointFactory Factory { get; }
 
-        internal Mock<IActionInvokerFactory> InvokerFactory { get; }
-
         internal IServiceProvider Services { get; }
 
-        [Fact]
-        public async Task AddEndpoints_AttributeRouted_UsesActionInvoker()
```
rynowak: Obsolete

src/Http/Routing/src/IDynamicEndpointMetadata.cs
pranavkm: `IDynamicEndpointMetadata.IsDynamic => false;` :trollface: Is the presence of the interface not sufficient?

src/Http/Routing/test/UnitTests/MatcherPolicyTest.cs
```diff
@@ -0,0 +1,92 @@
...
+            }
+
+            public bool IsDynamic { get; }
+        }
+
+        private class TestMatcherPolicy : MatcherPolicy
+        {
+            public override int Order => throw new System.NotImplementedException();
+
+            public new static bool ContainsDynamicEndpoints(IReadOnlyList<Endpoint> endpoints)
```
JamesNK: What is this for? Why not call `MatcherPolicy.ContainsDynamicEndpoints(endpoints)` in tests?

src/Http/Routing/src/Matching/MatcherPolicy.cs
```diff
@@ -24,5 +27,30 @@ public abstract class MatcherPolicy
...
         /// </summary>
         public abstract int Order { get; }
+
+        /// <summary>
+        /// Returns a value that indicates whether the provided <paramref name="endpoints"/> contains
+        /// one or more dynamic endpoints.
+        /// </summary>
+        /// <param name="endpoints">The set of endpoints.</param>
+        /// <returns><c>true</c> if a dynamic endpoint is found; otherwise returns <c>false</c>.</returns>
+        protected static bool ContainsDynamicEndpoints(IReadOnlyList<Endpoint> endpoints)
```
JamesNK: Who would use this? It looks like it is only called from tests at the moment.

src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/Routing/ActionEndpointFactory.cs
JamesNK: What is the perf impact of `var invokerFactory = context.RequestServices.GetRequiredService<IActionInvokerFactory>();`? Can we measure and fix it as part of this PR or create a followup issue?

src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/Infrastructure/ActionContextAccessor.cs
```diff
@@ -7,12 +7,23 @@ namespace Microsoft.AspNetCore.Mvc.Infrastructure
 {
     public class ActionContextAccessor : IActionContextAccessor
     {
+        internal static IActionContextAccessor Null = new NullActionContextAccessor();
```
JamesNK: readonly

src/Mvc/src/Microsoft.AspNetCore.Mvc.RazorPages/Infrastructure/PageLoaderMatcherPolicy.cs
```diff
@@ -0,0 +1,83 @@
...
+            
+            for (var i = 0; i < candidates.Count; i++)
+            {
+                ref var candidate = ref candidates[i];
+                var endpoint = (RouteEndpoint)candidate.Endpoint;
+
+                var page = endpoint.Metadata.GetMetadata<PageActionDescriptor>();
+                if (page != null)
+                {
+                    var compiled = _loader.Load(page);
```
pranavkm: Interesting. We could use this to fix - https://github.com/aspnet/AspNetCore/pull/6848 - specifically for endpoint routing

src/Mvc/src/Microsoft.AspNetCore.Mvc.RazorPages/Infrastructure/PageActionDescriptorProvider.cs
```diff
@@ -106,6 +107,10 @@ private void AddActionDescriptors(IList<ActionDescriptor> actions, PageRouteMode
                     descriptor.RouteValues.Add("page", model.ViewEnginePath);
                 }
 
+                // Mark all pages as a "dynamic endpoint" - this is how we deal with the compilation of pages
+                // in endpoint routing.
+                descriptor.EndpointMetadata.Add(new DynamicEndpointMetadata());
```
JamesNK: Is this tested for and used anywhere? ContainsDynamicEndpoints uses it but ContainsDynamicEndpoints doesn't appear to be used in any runtime code.

JamesNK: I guess PageLoaderMatcherPolicy is making the assumption that anything with PageActionDescriptor is implicitly dynamic.

src/Mvc/src/Microsoft.AspNetCore.Mvc.RazorPages/Infrastructure/PageLoaderMatcherPolicy.cs
pranavkm: A bit of a stretch because we don't need to do anything, but this might let us not use `IActionDescriptorChangeProvider` with endpoint routing. Essentially `IPageLoader.Load` will reflect the cache coherent version of the page. We could use this to replace the candidate rather than rely on `IADChangeProvider`

src/Security/Authorization/Policy/src/AuthorizationMiddleware.cs
```diff
@@ -46,13 +46,6 @@ public async Task Invoke(HttpContext context)
 
             var endpoint = context.GetEndpoint();
 
-            // Workaround for https://github.com/aspnet/AspNetCore/issues/7011. Do not use the AuthorizationMiddleware for Razor Pages
```
pranavkm: 👏

src/Http/Routing/src/Matching/CandidateSet.cs
pranavkm: Not familiar with the code, but are there any concurrency concerns here if this is invoked by multiple invocations of matcher policy?

src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/Routing/ActionEndpointFactory.cs
```diff
@@ -191,10 +184,14 @@ private static (RoutePattern resolvedRoutePattern, IDictionary<string, string> r
...
                 var routeData = context.GetRouteData();
-
                 var actionContext = new ActionContext(context, routeData, action);
 
-                var invoker = _invokerFactory.CreateInvoker(actionContext);
+                // We don't want to close over the invokerFactory in ActionEndpointFactory as
+                // that creates cycles in DI. Since we're creating this delegate at startup time
+                // we don't want to create all of the things we use at runtime until the action
+                // actually matches.
+                var invokerFactory = context.RequestServices.GetRequiredService<IActionInvokerFactory>();
```
pranavkm: Presumably you could cache this inside the delegate.

src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/Routing/ActionEndpointFactory.cs
rynowak: I'll make the change to amortize this.

src/Http/Routing/test/UnitTests/MatcherPolicyTest.cs
rynowak: Visibility bro.

src/Http/Routing/src/Matching/MatcherPolicy.cs
rynowak: A matcher policy that wants to know if there's any dynamic behaviour inside its `AppliesToEndpoints` method. Since you and @pranavkm are both stuck on this I'll use it just to make y'all happy.

src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/Infrastructure/ActionContextAccessor.cs
rynowak: nice

src/Mvc/src/Microsoft.AspNetCore.Mvc.RazorPages/Infrastructure/PageActionDescriptorProvider.cs
rynowak: I added it to make you guys happy. This is here so **other** policies can know that we have dynamic behaviour. Right now there are no such other policies.

src/Mvc/src/Microsoft.AspNetCore.Mvc.RazorPages/Infrastructure/PageLoaderMatcherPolicy.cs
rynowak: I don't super-get your comment. but yes. You can do the async

rynowak: > A bit of a stretch because we don't need to do anything, but this might let us not use `IActionDescriptorChangeProvider` with endpoint routing.

That lets us add and remove as well as change though.

src/Http/Routing/src/Matching/CandidateSet.cs
rynowak: `CandidateSet` is per-request mutable state

src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/Routing/ActionEndpointFactory.cs
rynowak: dids't

rynowak:
@JamesNK @pranavkm updated for your reviewing pleasure

rynowak:
@JamesNK  ping

src/Http/Routing/src/Matching/MatcherPolicy.cs
```diff
@@ -44,7 +44,7 @@ protected static bool ContainsDynamicEndpoints(IReadOnlyList<Endpoint> endpoints
             for (var i = 0; i < endpoints.Count; i++)
             {
                 var metadata = endpoints[i].Metadata.GetMetadata<IDynamicEndpointMetadata>();
-                if (metadata.IsDynamic)
+                if (metadata?.IsDynamic == true)
```
JamesNK: Personally I prefer `if (metadata?.IsDynamic ?? false)` more but whateves

src/Mvc/src/Microsoft.AspNetCore.Mvc.Core/Routing/ActionEndpointFactory.cs
```diff
@@ -188,13 +181,27 @@ private static (RoutePattern resolvedRoutePattern, IDictionary<string, string> r
...
+            IActionInvokerFactory invokerFactory = null;
+
             RequestDelegate requestDelegate = (context) =>
             {
                 var routeData = context.GetRouteData();
-
                 var actionContext = new ActionContext(context, routeData, action);
 
-                var invoker = _invokerFactory.CreateInvoker(actionContext);
+                if (invokerFactory == null)
```
JamesNK: 👍

src/Mvc/src/Microsoft.AspNetCore.Mvc.RazorPages/CompiledPageActionDescriptor.cs
JamesNK: Is it possible that the same compiled page could be referenced by multiple route endpoints? I am concern about what happens if they are setup with different metadata and how that would effect the endpoint here.

Or is it not a problem because different `CompiledPageActionDescriptor` instances are used for each?

rynowak: No, that is not possible. Pages always use attribute routing, so if you have two routes you always have to action descriptors.

Conventional routes + controllers allow the same ad to be used multiple times. 

A good question!

JamesNK (approved):

pranavkm (approved):

rynowak:
Thanks chaps!

src/Http/Routing/src/Matching/MatcherPolicy.cs
rynowak: #SledGang

Issue #4927, opened 2018-10-10
Improve filenames and class names in code generation


dougbu:
#### Is this a Bug or Feature request?
Bug that's part of #4896

#### Steps to reproduce (preferably a link to a GitHub repo with a repro project):
1. Create new console project
2. Add reference to manually-built NSwag.MSBuild.CodeGeneration package (relying on the Microsoft.Extensions.ApiDescription.Design package)
3. Add `<ServiceUriReference Include="https://docs.oracle.com/en/cloud/paas/content-cloud/rest-api-documents/swagger.json" ... />` to the .csproj file
4. `dotnet build`
##### Expected
Successful build
##### Actual
Lots of compilation errors in the generated Docs.oracle.com_en_cloud_paas_content-cloud_rest-api-documents_swaggerClient.cs file

#### Description of the problem
Default filenames and class names may contain invalid characters. In this case, the chosen class name contains both dashes and periods (Docs.oracle.com_en_cloud_paas_content-cloud_rest-api-documents_swaggerClient).

It's also possible for the package to generate invalid filenames from a URI containing, say, angle brackets or double-quotes.

#### Version of `Microsoft.AspNetCore.Mvc` or `Microsoft.AspNetCore.App` or `Microsoft.AspNetCore.All`
See the [features/client.code.generation](https://github.com/aspnet/Mvc/tree/feature/client.code.generation) branch. That work is intended for the 2.2 Preview 3 milestone.

dougbu:
8d6d300bfc

PR #7447, opened 2019-02-11
Improve service reference filenames and class names

dougbu:
- #4927
- fully-sanitize class names and filenames
- default metadata in sequence [URI or project&document name ->] `%(DocumentPath)` -> `%(OutputPath)` -> `%(ClassName)`
  - if user sets metadata explicitly, the override affects defaults later in the sequence
- separate some nested validations and defaulting steps
  - provide default `%(DocumentName}` even if `%(DocumentPath}` is set explicitly
  - validate URI is absolute even if `%(DocumentPath}` is set explicitly

other:
- don't write out an empty Open API / Swagger file

nits:
- do not use default `%(DocumentName}` in default `%(DocumentPath)` for `<ServiceProjectReference>` items
- do not use empty URI path or query string in default `%(DocumentPath)` for `<ServiceUriReference>` items

### DIFF
--- a/src/Mvc/Extensions.ApiDescription.Design/src/CSharpIdentifier.cs
+++ b/src/Mvc/Extensions.ApiDescription.Design/src/CSharpIdentifier.cs
@@ -1,0 +1,55 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Globalization;
+using System.Text;
+
+// Copied from
+// https://github.com/aspnet/AspNetCore-Tooling/blob/master/src/Razor/src/Microsoft.AspNetCore.Razor.Language/CSharpIdentifier.cs
+namespace Microsoft.Extensions.ApiDescription.Tasks
+{
+    internal static class CSharpIdentifier
+    {
+        // CSharp Spec §2.4.2
+        private static bool IsIdentifierStart(char character)
+        {
+            return char.IsLetter(character) ||
+                character == '_' ||
+                CharUnicodeInfo.GetUnicodeCategory(character) == UnicodeCategory.LetterNumber;
+        }
+
+        public static bool IsIdentifierPart(char character)
+        {
+            return char.IsDigit(character) ||
+                   IsIdentifierStart(character) ||
+                   IsIdentifierPartByUnicodeCategory(character);
+        }
+
+        private static bool IsIdentifierPartByUnicodeCategory(char character)
+        {
+            var category = CharUnicodeInfo.GetUnicodeCategory(character);
+
+            return category == UnicodeCategory.NonSpacingMark || // Mn
+                category == UnicodeCategory.SpacingCombiningMark || // Mc
+                category == UnicodeCategory.ConnectorPunctuation || // Pc
+                category == UnicodeCategory.Format; // Cf
+        }
+
+        public static string SanitizeIdentifier(string inputName)
+        {
+            if (!IsIdentifierStart(inputName[0]) && IsIdentifierPart(inputName[0]))
+            {
+                inputName = "_" + inputName;
+            }
+
+            var builder = new StringBuilder(inputName.Length);
+            for (var i = 0; i < inputName.Length; i++)
+            {
+                var ch = inputName[i];
+                builder.Append(IsIdentifierPart(ch) ? ch : '_');
+            }
+
+            return builder.ToString();
+        }
+    }
+}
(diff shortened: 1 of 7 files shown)


src/Mvc/src/Microsoft.Extensions.ApiDescription.Design/GetFileReferenceMetadata.cs
```diff
@@ -17,6 +19,29 @@ public class GetFileReferenceMetadata : Task
     {
         private const string TypeScriptLanguageName = "TypeScript";
 
+        // Ignore Unicode escape sequences because, though they may escape valid class name characters,
+        // backslash is not a valid filename character.
+        private static readonly HashSet<UnicodeCategory> ValidClassNameCharacterCategories =
```
pranavkm: Razor has a fairly straightforward implementation for producing a sanitized identifier. Could we copy-pasta the type just so that it's easier to patch them all if we were ever to patch one? https://github.com/aspnet/AspNetCore-Tooling/blob/master/src/Razor/src/Microsoft.AspNetCore.Razor.Language/CSharpIdentifier.cs

src/Mvc/src/Microsoft.Extensions.ApiDescription.Design/GetProjectReferenceMetadata.cs
```diff
@@ -53,29 +57,69 @@ public override bool Execute()
...
+                                ".",
+                                sanitizedDocumentName.Split(InvalidFilenameStrings, StringSplitOptions.None));
+                        }
+
+                        documentPath = $"{projectFilename}_{sanitizedDocumentName}";
+
+                        // Possible the document path already ends with .json. Don't duplicate that or a final period.
+                        if (!documentPath.EndsWith(".json", StringComparison.OrdinalIgnoreCase))
+                        {
+                            if (documentPath.EndsWith(".", StringComparison.Ordinal))
```
pranavkm: This seems like a super weird scenario to address 🤷‍♀️

pranavkm (approved):
Seems ok as is. Would slightly prefer using the Razor type for better shareability.

src/Mvc/src/Microsoft.Extensions.ApiDescription.Design/GetFileReferenceMetadata.cs
dougbu: Why does that class use such a limited selection of `UnicodeCategories`s (four)?

pranavkm: Not super sure. Think @NTaylorMullen authored it

NTaylorMullen: Nope, lol those categories are from wayyyy back when. AspNetWebStack 😄

NTaylorMullen (commented):
This is quite a bit of added code without any tests. Any particular reason?

src/Mvc/src/Microsoft.Extensions.ApiDescription.Design/GetFileReferenceMetadata.cs
dougbu: If we incorporate the rest of the categories, I'd be willing to copy the code 😃

dougbu:
> This is quite a bit of added code without any tests. Any particular reason?

This feature has some engineering dept and the testing has been manual to-date. But, it's an issue that's on our radar!

src/Mvc/src/Microsoft.Extensions.ApiDescription.Design/GetFileReferenceMetadata.cs
dougbu: Actually, I'd probably just reuse the updated class with the appropriate `<Compile />` item in the project.

dougbu: And, I'll file the appropriate bug sometime this evening.

dougbu: > Actually, I'd probably just reuse the updated class with the appropriate `<Compile />` item in the project.

I misread the URI and see `CSharpIdentifier` isn't in aspnet/AspNetCore. Will have to dupe the class ☹️

dougbu: Double-checked the `char.IsDigit(...)` and `char.IsLetter(...)` documentation. Nothing is missing given the broadness of those two calls. No issue on the AspNetCore-Tooling side.

PR #7467, opened 2019-02-11
Remove capture of "this" from local function

halter73:
@benaadams Was it important that "producer" was an argument instead of capturing "this"? The method generated by the compiler for the capture version doesn't look bad other than not being static.


jkotalik (approved):

benaadams (approved):

benaadams:
Was originally async, and then capturing is allocating :(

PR #7468, opened 2019-02-11
AsTask => GetAsTask()

jkotalik:
Working on improving perf.
cc @benaadams


davidfowl (approved):

benaadams (approved):

halter73 (approved):

BrennanConroy:
Could we move that code somewhere more common? I'm sure we can make use of it in SignalR.

davidfowl:
> Could we move that code somewhere more common? I'm sure we can make use of it in SignalR.

Copy it into SignalR, it isn't enough code to warrant a package. I do think maybe this could go into CoreFx in pipelines though.

PR #7470, opened 2019-02-11
Fixup handshake logic and layering

BrennanConroy:
Part of https://github.com/aspnet/AspNetCore/issues/5301

Moved the handshake and receive loop out of the connection layer into the hub_connection layer. Also, fixed up a few issues such as waiting for handshake in `start`

### DIFF
--- a/src/SignalR/clients/cpp/src/signalrclient/hub_connection_impl.cpp
+++ b/src/SignalR/clients/cpp/src/signalrclient/hub_connection_impl.cpp
@@ -54,12 +54,25 @@ namespace signalr
         // weak_ptr prevents a circular dependency leading to memory leak and other problems
         auto weak_hub_connection = std::weak_ptr<hub_connection_impl>(this_hub_connection);
 
-        m_connection->set_message_received_string([weak_hub_connection](const utility::string_t& message)
+        m_connection->set_message_received([weak_hub_connection](const utility::string_t& message)
         {
             auto connection = weak_hub_connection.lock();
             if (connection)
             {
                 connection->process_message(message);
+            }
+        });
+
+        m_connection->set_disconnected([weak_hub_connection]()
+        {
+            auto connection = weak_hub_connection.lock();
+            if (connection)
+            {
+                connection->m_handshakeTask.set_exception(signalr_exception(_XPLATSTR("connection closed while handshake was in progress.")));
+                if (connection->m_disconnected)
+                {
+                    connection->m_disconnected();
+                }
             }
         });
     }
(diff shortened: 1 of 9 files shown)


src/SignalR/clients/cpp/src/signalrclient/hub_connection_impl.cpp
```diff
@@ -89,7 +102,49 @@ namespace signalr
...
+        auto weak_connection = m_connection->weak_from_this();
+        return m_connection->start()
+            .then([weak_connection, this](pplx::task<void> startTask)
+            {
+                startTask.get();
+                auto connection = weak_connection.lock();
+                if (!connection)
+                {
+                    // The connection has been destructed, we should just return and avoid any unobserved task exceptions
+                    return pplx::task_from_result();
```
halter73: If an invalid handshake response results in a faulted task, so should a connection being deconstructed during start.

BrennanConroy: The difference is that this would only happen if someone loses the reference to the hub_connection, which I believe would crash the process due to an unhandled exception in a task

src/SignalR/clients/cpp/src/signalrclient/hub_connection_impl.cpp
```diff
@@ -54,14 +54,27 @@ namespace signalr
...
             }
         });
+
+        m_connection->set_disconnected([weak_hub_connection]()
+        {
+            auto connection = weak_hub_connection.lock();
+            if (connection)
+            {
+                connection->m_handshakeTask.set_exception(signalr_exception(_XPLATSTR("connection closed while handshake was in progress.")));
+                if (connection->m_disconnected)
```
halter73: You don't need to null check m_disconnected. Just add `m_disconnected([](){})` to the hub_connection_impl initializer list like it is in connection_impl.cpp.

halter73 (changes requested):

halter73 (approved):

src/SignalR/clients/cpp/src/signalrclient/hub_connection_impl.cpp
halter73: > I believe would crash the process due to an unhandled exception in a task

I don't think I understand. Can you demonstrate the problem?

Callers to start() are going to need to handle faulted task resulting from handshake failures anyway. I would just return a faulted task in this case.

BrennanConroy: Ok fair, even if the user loses the ref to `hub_connection` if they call a method that returns a task they should observe the task
```c++
pplx::task startTask;
// scope
{
    hub_connection connection;
    startTask = connection.start();
}
startTask.get();
```

halter73: `startTask.get();` could throw even if the hub_connection wasn't destructed due to server issues, network failure, etc... I cannot even begin imagine why you wouldn't want it to throw when hub_connection was destructed.

PR #7471, opened 2019-02-11
Update CORS package.json and friends

rynowak:
NPM (latest node) is sad about our current package.json and stuff - this updates for every time I restore/build and it's gotten annoying. 

Looks like just a bunch of version updates, any reason why we shouldn't take this?

### DIFF
--- a/src/Middleware/CORS/test/FunctionalTests/package.json
+++ b/src/Middleware/CORS/test/FunctionalTests/package.json
@@ -2,7 +2,7 @@
   "devDependencies": {
     "jest": "^23.6.0",
     "merge": "^1.2.1",
-    "puppeteer": "^1.8.0"
+    "puppeteer": "^1.12.2"
   },
   "dependencies": {},
   "scripts": {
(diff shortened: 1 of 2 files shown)


src/Middleware/CORS/test/FunctionalTests/package.json
```diff
@@ -2,7 +2,7 @@
   "devDependencies": {
     "jest": "^23.6.0",
     "merge": "^1.2.1",
-    "puppeteer": "^1.8.0"
+    "puppeteer": "^1.12.2"
```
rynowak: NPM did this for me.

pranavkm: cc @Eilon in case we need to notify LCA.

pranavkm (approved):

src/Middleware/CORS/test/FunctionalTests/package.json
Eilon: Component Governance will pick it up. Good to go.

PR #7473, opened 2019-02-11
Update metadata for the Blazor VSIX

danroth27:
Fixes https://github.com/aspnet/AspNetCore/issues/7472

### DIFF
--- a/src/Components/Blazor/BlazorExtension/src/source.extension.vsixmanifest
+++ b/src/Components/Blazor/BlazorExtension/src/source.extension.vsixmanifest
@@ -2,8 +2,8 @@
 <PackageManifest Version="2.0.0" xmlns="http://schemas.microsoft.com/developer/vsx-schema/2011" xmlns:d="http://schemas.microsoft.com/developer/vsx-schema-design/2011">
     <Metadata>
         <Identity Id="Microsoft.VisualStudio.BlazorExtension" Version="|%CurrentProject%;GetBuildVersion|" Language="en-US" Publisher="Microsoft" />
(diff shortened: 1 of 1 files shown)


src/Components/Blazor/BlazorExtension/src/source.extension.vsixmanifest
```diff
@@ -2,8 +2,8 @@
 <PackageManifest Version="2.0.0" xmlns="http://schemas.microsoft.com/developer/vsx-schema/2011" xmlns:d="http://schemas.microsoft.com/developer/vsx-schema-design/2011">
     <Metadata>
         <Identity Id="Microsoft.VisualStudio.BlazorExtension" Version="|%CurrentProject%;GetBuildVersion|" Language="en-US" Publisher="Microsoft" />
-        <DisplayName>ASP.NET Core Blazor Language Services</DisplayName>
-        <Description xml:space="preserve">Provides Visual Studio support for ASP.NET Core Blazor</Description>
+        <DisplayName>Blazor</DisplayName>
+        <Description xml:space="preserve">Provides Visual Studio support for Blazor</Description>
```
rynowak: Suggestion `Blazor Templates` and `Provides templates for Blazor` or similar?

danroth27: I prefer to keep it generic. This is the VSIX you need for all things Blazor. Also, I think the shorter display name "Blazor" looks more official.

rynowak: I think it makes it look less official 😁  but still better.

rynowak (approved):

PR #7475, opened 2019-02-11
Generate a nuget package which contains the runtime pack for Microsoft.AspNetCore.App

natemcmaster:
Part of https://github.com/aspnet/AspNetCore/issues/6501

Changes:

* Create a new package  - Microsoft.AspNetCore.App.Runtime.${rid} - which contains crossgen-ed versions of shared framework assemblies.
* Add a task which can generate the .deps.json file for the shared framework. Unlike the existing task we have, this does not rely on consuming a .deps.json file generated by first running a restore on packages.

TODO (in a separate PR)
 * Update the layout of this package to match what the SDK will expect to use
 * Stop producing the "runtime.*.Microsoft.AspNetCore.App" package. This will be retired in favor of the package added in this PR.
 * Where do platform manifests and package conflict data go?

cref https://github.com/dotnet/designs/pull/50

cc @dsplaisted @nguerrera @dagood


pakrym:
> Add a task which can generate the .deps.json file for the shared framework. Unlike the existing task we have, this does not rely on consuming a .deps.json file generated by first running a restore on packages.

Why?

natemcmaster:
> Why?

The SDK's .deps.json generation task requires a project.assets.json file, and it doesn't handle ProjectReference or native assets. I wrote the task to workaround these limitations. This will mean we can stop producing packages at all for assemblies that are shared framework only, like Mvc.dll.

dsplaisted:
@natemcmaster If you want to put a locally-built copy of these packages on a share somewhere, I can test consuming them in the SDK.

src/Framework/src/Microsoft.AspNetCore.App.Runtime.csproj
```diff
@@ -0,0 +1,285 @@
...
+
+This package is an internal implementation of the .NET Core SDK and is not meant to be used as a normal PackageReference.
+    </PackageDescription>
+    <PackageTags>aspnetcore;shared-framework</PackageTags>
+    <SuppressDependenciesWhenPacking>true</SuppressDependenciesWhenPacking>
+
+    <!-- Subject to change: see https://github.com/dotnet/designs/pull/50 -->
+    <PackageType>RuntimePack</PackageType>
+    <!-- NuGet appends target framework to this value. Example: runtimes/win-x64/nativeassets/netcoreapp3.0/ -->
+    <BuildOutputTargetFolder>runtimes/$(RuntimeIdentifier)/nativeassets/</BuildOutputTargetFolder>
```
natemcmaster: Totally guessed on what this should be. @dsplaisted what does the SDK currently expect for runtime packs?

dsplaisted: Is this the path to assets within the package?  Currently the SDK looks in `runtimes/<rid>/lib/netcoreapp3.0` for managed assets and `runtimes/<rid>/native` for native assets.

Feel free to stay with that layout for now.  Alternatively, you can just put them into `data/native` and `data/managed` (which is closer to what I'd like to eventually land on) and I'll update the SDK to consume them from there.

There does need to be a way to distinguish between the native and managed assets though, as they are written differently to the deps file.

build/tasks/FileUtilities.cs
```diff
@@ -0,0 +1,46 @@
...
+        }
+
+        static readonly HashSet<string> s_assemblyExtensions = new HashSet<string>(new[] { ".dll", ".exe", ".winmd" }, StringComparer.OrdinalIgnoreCase);
+        public static Version TryGetAssemblyVersion(string sourcePath)
+        {
+            var extension = Path.GetExtension(sourcePath);
+
+            return s_assemblyExtensions.Contains(extension) ? GetAssemblyVersion(sourcePath) : null;
+        }
+        private static Version GetAssemblyVersion(string sourcePath)
```
pakrym: nit: formatting

build/tasks/GenerateSharedFrameworkDepsFile.cs
```diff
@@ -0,0 +1,114 @@
...
+        [Required]
+        public string TargetFramework { get; set; }
+
+        [Required]
+        public string FrameworkName { get; set; }
+
+        [Required]
+        public string FrameworkVersion { get; set; }
+
+
```
pakrym: nit: formatting

src/Framework/src/Microsoft.AspNetCore.App.Runtime.csproj
```diff
@@ -0,0 +1,285 @@
...
+    <CrossGenSymbolsType>CreatePerfMap</CrossGenSymbolsType>
+    <CrossGenSymbolsType Condition="'$(TargetOsName)' == 'win'">CreatePDB</CrossGenSymbolsType>
+
+    <LibPrefix Condition=" '$(TargetOsName)' != 'win' ">lib</LibPrefix>
+    <LibExtension>.so</LibExtension>
+    <LibExtension Condition=" '$(TargetOsName)' == 'win' ">.dll</LibExtension>
+    <LibExtension Condition=" '$(TargetOsName)' == 'osx' ">.dylib</LibExtension>
+    <ExeExtension Condition=" '$(TargetOsName)' == 'win' ">.exe</ExeExtension>
+    <!-- 3B = semicolon in ASCII -->
+    <PathSeparator Condition="'$(PathSeparator)' == ''">:</PathSeparator>
```
pakrym: Do you think this would be reused anywhere? Maybe move it to somewhere common?

src/Framework/src/Microsoft.AspNetCore.App.Runtime.csproj
```diff
@@ -0,0 +1,285 @@
...
+
+    <RuntimePackageRoot>$(NuGetPackageRoot)/runtime.$(RuntimeIdentifier).microsoft.netcore.app/$(MicrosoftNETCoreAppPackageVersion)/</RuntimePackageRoot>
+    <CrossgenToolPath>$(RuntimePackageRoot)tools/$(CrossgenToolPackagePath)</CrossgenToolPath>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Reference Include="@(Dependency);@(ExternalDependency)" />
+    <ProjectReference Condition="'$(BuildIisNativeProjects)' == 'true' AND '$(BuildNative)' != 'false'" Include="$(RepositoryRoot)src\Servers\IIS\AspNetCoreModuleV2\InProcessRequestHandler\InProcessRequestHandler.vcxproj">
+      <SetPlatform>Platform=$(TargetArchitecture)</SetPlatform>
+      <SetPlatform Condition="'$(TargetArchitecture)' == 'x86'">Platform=Win32</SetPlatform>
```
pakrym: Slightly unrelated: Might be good to define `TargetNativeArchitecture` and `NativePlatform` so we can use it in native project references without conditions.

src/Framework/src/Microsoft.AspNetCore.App.Runtime.csproj
```diff
@@ -0,0 +1,285 @@
...
+    <CrossgenToolPackagePath Condition=" '$(TargetRuntimeIdentifier)' == 'linux-arm64' ">x64_arm64\$(CrossgenToolPackagePath)</CrossgenToolPackagePath>
+    <CrossgenToolPackagePath Condition=" '$(TargetRuntimeIdentifier)' == 'win-arm' ">x86_arm\$(CrossgenToolPackagePath)</CrossgenToolPackagePath>
+
+    <RuntimePackageRoot>$(NuGetPackageRoot)/runtime.$(RuntimeIdentifier).microsoft.netcore.app/$(MicrosoftNETCoreAppPackageVersion)/</RuntimePackageRoot>
+    <CrossgenToolPath>$(RuntimePackageRoot)tools/$(CrossgenToolPackagePath)</CrossgenToolPath>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Reference Include="@(Dependency);@(ExternalDependency)" />
+    <ProjectReference Condition="'$(BuildIisNativeProjects)' == 'true' AND '$(BuildNative)' != 'false'" Include="$(RepositoryRoot)src\Servers\IIS\AspNetCoreModuleV2\InProcessRequestHandler\InProcessRequestHandler.vcxproj">
```
pakrym: Why not reference IISReferenceResolution tasks from here and use NativeReference?

pakrym (approved):

src/Framework/src/Microsoft.AspNetCore.App.Runtime.csproj
natemcmaster: IIS resolution ends up copying both x86 and x64 into output, and puts it under `$(outdir)/$(arch)/`. I want to reference one bitness and have it in `$(outDir)`.

src/Framework/src/Microsoft.AspNetCore.App.Runtime.csproj
natemcmaster: I think it should be done too, but it's not trivial. Let's use https://github.com/aspnet/AspNetCore/issues/7196 to follow up on this.

src/Framework/src/Microsoft.AspNetCore.App.Runtime.csproj
natemcmaster: It's only used in this one place. If we had a more general need for it, yes, we could move it to a common file, but for now, YAGNI.

src/Framework/src/Microsoft.AspNetCore.App.Runtime.csproj
natemcmaster: I'll update to `runtimes/<rid>/lib/<tfm>` and `runtimes/<rid>/native` before merging.

src/Framework/src/Microsoft.AspNetCore.App.Runtime.csproj
pakrym: You are wrong about both platforms part, in master it references only one platform https://github.com/aspnet/AspNetCore/blob/master/eng/targets/ResolveIisReferences.targets#L13. Output location can be changed by setting LinkBase.

natemcmaster: You're right, looks like we have two `NativeProjectReference`, each with a platform. I still need slightly different properties though, like ReferenceOutputAssembly.

src/Framework/src/Microsoft.AspNetCore.App.Runtime.csproj
pakrym: I'll let you chose here. I'm just worried about supporting all these slight modifications of one thing in many places.

natemcmaster:
Build failure in PR:

```
LINK : fatal error LNK1104: cannot open file 'Release\tracelog.obj' [F:\vsagent\2\s\src\Servers\IIS\AspNetCoreModuleV2\IISLib\IISLib.vcxproj]
```
https://dev.azure.com/dnceng/public/_build/results?buildId=93376&view=logs&jobId=f31c9f97-4411-58e7-49ac-fc73f645e6b6&taskId=6eab48b9-b4ce-58d4-7c0f-2d2ef2bbc0bd&lineStart=3479&lineEnd=3479&colStart=1&colEnd=142

Any ideas @jkotalik @pakrym? Is this something you see often? I periodically see issues like this with vcxproj. I think it's a race condition because retrying the build often makes it go away.

pakrym:
I haven't seen this failure before but it's very likely a race because  `Release\tracelog.obj` is not under platform specific directory and we are building two platforms in parallel.

src/Framework/src/Microsoft.AspNetCore.App.Runtime.csproj
pakrym: Commend here was intended for https://github.com/aspnet/AspNetCore/pull/7475#discussion_r256009498 thread.

natemcmaster: I'd like to leave this as is. Once we have confirmed that this new package is sufficient to replace the `runtime.<rid>` versions, we probably don't need to have as many NativeProjectReference's. I'd like to move most of our tests on to using the shared framework instead of having app-local versions of ANCM.

natemcmaster:
More flaky C++ build failures:
```
f:\vsagent\2\s\src\servers\iis\aspnetcoremodulev2\commonlib\stdafx.cpp : fatal error C1041: cannot open program database 'F:\vsagent\2\s\src\Servers\IIS\AspNetCoreModuleV2\CommonLib\Release\CommonLib.pdb'; if multiple CL.EXE write to the same .PDB file, please use /FS [F:\vsagent\2\s\src\Servers\IIS\AspNetCoreModuleV2\CommonLib\CommonLib.vcxproj]
FileTracker : error FTK1011: could not create the new file tracking log file: F:\vsagent\2\s\src\Servers\IIS\AspNetCoreModuleV2\CommonLib\Release\CommonLib.tlog\CL.write.1.tlog. The file exists. [F:\vsagent\2\s\src\Servers\IIS\AspNetCoreModuleV2\CommonLib\CommonLib.vcxproj]
```

pakrym:
I think you are double building in this PR. It's slightly flaky but not this much.

natemcmaster:
I don't think it's a double build of the same architecture. The binlogs show MSBuild doing the right thing and de-duplicates the build request. The errors show problems with files shared between x86 and x64 builds.

natemcmaster:
Everything passed on retry except Helix. I'm looking into it more...

natemcmaster:
It looks like Helix problems were unrelated to this change, but surfaced as a result of adding a new project. I've made some tweaks to the ways Helix builds add a dependency to xunit.runner.console (cc @HaoK )

natemcmaster:
CI checks passed, and after running build 5 times, I haven't been able to get a repro of the flaky CI problems. If we run into this again, we can revert and investigate.

I opened a bug for the flaky Helix test failure, but it appears unrelated to this change, so merging now.

PR #7476, opened 2019-02-11
Fix returning buffered data after stream is drained

### LINKED ISSUE #7448: StreamPipeReader should return the buffered data even after the Stream is done
### Describe the bug
The StreamPipeReader should return the buffered data from the stream even after the underlying Stream has been drained. 

### To Reproduce

```C#
using System;
using System.Buffers;
using System.Linq;
using System.Text;
using System.Threading.Tasks;

namespace WebApplication101
{
    public class Startup
    {
        // This method gets called by the runtime. Use this method to add services to the container.
        // For more information on how to configure your application, visit https://go.microsoft.com/fwlink/?LinkID=398940
        {
        }

        // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
        public void Configure(IApplicationBuilder app, IHostingEnvironment env)
        {
            if (env.IsDevelopment())
            {
            }

            app.Use(async (context, next) =>
            {
                if (context.Request.ContentLength is null)
                {
                    await next();
                    return;
                }

                await context.Request.BufferAsync();

                await next();
            });

            app.UseRouting(routes =>
            {
                routes.MapGet("/", async context =>
                {
                    await context.Response.WriteAsync("Hello World!");
                });

                routes.MapPost("/api/values", async context =>
                {
                    var result = await context.Request.BodyPipe.ReadAsync();
                    var buffer = result.Buffer;

                    var value = Encoding.UTF8.GetString(buffer.ToArray());

                    await context.Response.WriteAsync(value);
                });
            });
        }
    }

    public static class RequestBufferingExtensions
    {
        public static async Task BufferAsync(this HttpRequest request)
        {
            // Buffer the entire request body into the pipe
            while (true)
            {
                var result = await request.BodyPipe.ReadAsync(request.HttpContext.RequestAborted);
                var buffer = result.Buffer;

                if (result.IsCompleted)
                {
                    break;
                }

                request.BodyPipe.AdvanceTo(buffer.Start, buffer.End);
            }
        }
    }
}
```

The above code tries to buffer all of the pipe's contents middleware before the post body is processed. 

### Expected behavior
I expected to get the entire post body in /api/values, instead I get a 0 byte read.

### Screenshots
image

Notice the buffered bytes is non-zero but the buffer returned is empty.

### Additional context
latest build

davidfowl:
- Change TryRead to return the buffer if the Stream is completed

Fixes #7448

### DIFF
--- a/src/Http/Http/src/StreamPipeReader.cs
+++ b/src/Http/Http/src/StreamPipeReader.cs
@@ -21,7 +21,7 @@ namespace System.IO.Pipelines
 
         private CancellationTokenSource _internalTokenSource;
         private bool _isReaderCompleted;
-        private bool _isWriterCompleted;
+        private bool _isStreamCompleted;
         private ExceptionDispatchInfo _exceptionInfo;
 
         private BufferSegment _readHead;
@@ -206,11 +206,6 @@ namespace System.IO.Pipelines
         public override async ValueTask<ReadResult> ReadAsync(CancellationToken cancellationToken = default)
         {
             // TODO ReadyAsync needs to throw if there are overlapping reads.
-            if (_isWriterCompleted)
-            {
-                return new ReadResult(buffer: default, isCanceled: false, isCompleted: true);
-            }
-
             ThrowIfCompleted();
 
             // PERF: store InternalTokenSource locally to avoid querying it twice (which acquires a lock)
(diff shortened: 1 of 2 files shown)


src/Http/Http/src/StreamPipeReader.cs
```diff
@@ -251,7 +251,7 @@ public override async ValueTask<ReadResult> ReadAsync(CancellationToken cancella
 
                     if (length == 0)
                     {
-                        _isWriterCompleted = true;
+                        _isStreamCompleted = true;
```
jkotalik: nit: _isInnerStreamCompleted?

jkotalik (approved):

src/Http/Http/src/StreamPipeReader.cs
davidfowl: What other stream is there?

jkotalik: fine.

pakrym (approved):

Issue #6817, opened 2019-01-17
Background Worker template


Tratcher:
A new template for generic host without web host.

No websdk.

Include a Hello World IHostedService

Include health checks (non-http)?

@glennc 's POC https://github.com/glennc/worker

Tratcher:
Deployment environments:
- Kubernetties
- Windows service (+installer)
- Daemon

Tratcher:
@glennc since this will only be using Extensions should it get an optional net4x TFM? Not sure if our template infrastructure supports that.

Tratcher:
@glennc moving to preview4 to track template upgrades for https://github.com/aspnet/AspNetCore/issues/6817#issuecomment-455328857

anurse:
The template is there. We have issues filed for the remaining work we want to explore.

PR #7478, opened 2019-02-12
Get worker template showing up in VS

Tratcher:
#6817
image

### DIFF
--- a/src/ProjectTemplates/Web.ProjectTemplates/content/Worker-CSharp/.template.config/vs-2017.3.host.json
+++ b/src/ProjectTemplates/Web.ProjectTemplates/content/Worker-CSharp/.template.config/vs-2017.3.host.json
@@ -3,37 +3,19 @@
   "name": {
     "text": "Worker Service",
     "package": "{0CD94836-1526-4E85-87D3-FB5274C5AFC9}",
-    "id": "1011"
+    "id": "1025"
   },
   "description": {
     "text": "An empty project template for creating a worker service.",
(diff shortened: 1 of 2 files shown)


davidfowl (approved):
LGTM

src/ProjectTemplates/Web.ProjectTemplates/content/Worker-CSharp/.template.config/vs-2017.3.host.json
```diff
@@ -3,37 +3,19 @@
...
-    "id": "1012"
+    "id": "1026"
   },
   "order": 100,
   "icon": "vs-2017.3/Empty.png",
   "learnMoreLink": "https://go.microsoft.com/fwlink/?linkid=2072778",
   "uiFilters": [ "oneaspnet" ],
-  "supportsDocker": true,
+  "supportsDocker": false,
   "legacyTemplateIdentity": "Microsoft.NetCore.CSharp.EmptyWorker",
```
phenning: By the way, you can kill the legacyTemplateIdentity, it was pretty much there to provide consistent telemetry when we swapped moved to the template engine based templates.

phenning (approved):
Other then my comment about legacyTemplateIdentity, LGTM from an authoring perspective.

glennc (approved):

glennc:
Merge this now. But I want to look at DockerSupport once a build goes through. Looking at the Dockerfile we have now it looks like it should just work.

PR #7480, opened 2019-02-12
Copy Kestrel perf changes

BrennanConroy:
See https://github.com/aspnet/AspNetCore/pull/7468 and https://github.com/aspnet/AspNetCore/pull/7422

### DIFF
--- a/src/SignalR/clients/csharp/Client.Core/src/Microsoft.AspNetCore.SignalR.Client.Core.csproj
+++ b/src/SignalR/clients/csharp/Client.Core/src/Microsoft.AspNetCore.SignalR.Client.Core.csproj
@@ -13,6 +13,7 @@
     <Compile Include="$(SignalRSharedSourceRoot)PipeWriterStream.cs" Link="PipeWriterStream.cs" />
     <Compile Include="$(SignalRSharedSourceRoot)ReflectionHelper.cs" Link="ReflectionHelper.cs" />
     <Compile Include="$(SignalRSharedSourceRoot)TimerAwaitable.cs" Link="Internal\TimerAwaitable.cs" />
+    <Compile Include="$(SignalRSharedSourceRoot)ValueTaskExtensions.cs" Link="Internal\ValueTaskExtensions.cs" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/SignalR/common/Shared/PipeWriterStream.cs
+++ b/src/SignalR/common/Shared/PipeWriterStream.cs
@@ -6,6 +6,7 @@ using System.Buffers;
 using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Internal;
 
 namespace System.IO.Pipelines
 {
@@ -57,7 +58,7 @@ namespace System.IO.Pipelines
 
         public override Task WriteAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
         {
-            return WriteCoreAsync(buffer.AsMemory(offset, count), cancellationToken).AsTask();
+            return WriteCoreAsync(buffer.AsMemory(offset, count), cancellationToken).GetAsTask();
         }
 
 #if NETCOREAPP3_0
--- a/src/SignalR/common/Shared/PipeWriterStream.cs
+++ b/src/SignalR/common/Shared/PipeWriterStream.cs
@@ -58,7 +58,7 @@ namespace System.IO.Pipelines
 
         public override Task WriteAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
         {
-            return WriteCoreAsync(buffer.AsMemory(offset, count), cancellationToken).GetAsTask();
+            return WriteCoreAsync(buffer.AsMemory(offset, count), cancellationToken).AsTask();
         }
 
 #if NETCOREAPP3_0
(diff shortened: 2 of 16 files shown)


src/SignalR/common/Shared/PipeWriterStream.cs
```diff
@@ -6,6 +6,7 @@
 using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Internal;
 
 namespace System.IO.Pipelines
 {
```
jkotalik: Can this class be WriteOnlyPipeStream? https://github.com/aspnet/AspNetCore/blob/master/src/Http/Http/src/WriteOnlyPipeStream.cs

src/SignalR/common/Shared/ValueTaskExtensions.cs
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.IO.Pipelines;
+using System.Runtime.CompilerServices;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Internal
```
jkotalik: Not a fan of this namespace. Is this appropriate?

src/SignalR/common/Shared/ValueTaskExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+            else
+            {
+                return valueTask.AsTask();
+            }
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static Task GetAsTask(this in ValueTask valueTask)
+        {
+            if (valueTask.IsCompletedSuccessfully)
```
davidfowl: AsTask is already optimized for ValueTask => https://github.com/dotnet/coreclr/blob/master/src/System.Private.CoreLib/shared/System/Threading/Tasks/ValueTask.cs#L170.

src/SignalR/clients/csharp/Client.Core/src/Microsoft.AspNetCore.SignalR.Client.Core.csproj
```diff
@@ -13,6 +13,7 @@
     <Compile Include="$(SignalRSharedSourceRoot)PipeWriterStream.cs" Link="PipeWriterStream.cs" />
     <Compile Include="$(SignalRSharedSourceRoot)ReflectionHelper.cs" Link="ReflectionHelper.cs" />
     <Compile Include="$(SignalRSharedSourceRoot)TimerAwaitable.cs" Link="Internal\TimerAwaitable.cs" />
+    <Compile Include="$(SignalRSharedSourceRoot)ValueTaskExtensions.cs" Link="Internal\ValueTaskExtensions.cs" />
```
JamesNK: IMO instead of placing this code under SignalR's shared source directory, it should be under the repos shared source directory: `src/Shared` -> https://github.com/aspnet/AspNetCore/tree/master/src/Shared

That way Kestrel and SignalR can use the same copy of code. There is no overhead of creating a source package because we mono repo now.

src/SignalR/common/Shared/ValueTaskExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+using System.IO.Pipelines;
+using System.Runtime.CompilerServices;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Internal
+{
+    internal static class ValueTaskExtensions
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static Task GetAsTask(this in ValueTask<FlushResult> valueTask)
```
JamesNK: Add a comment that this will avoid the allocation of `AsTask` when run in sync because it does not convert to `Task<T>`

ycrumeyrolle (commented):
You can apply the same optimization to the SignalR client in the class [ServerSentEventsMessageParser](https://github.com/aspnet/AspNetCore/blob/master/src/SignalR/clients/csharp/Http.Connections.Client/src/Internal/ServerSentEventsMessageParser.cs) with the fields `_dataPrefix ` and `_sseLineEnding `.

src/SignalR/clients/csharp/Client.Core/src/Microsoft.AspNetCore.SignalR.Client.Core.csproj
BrennanConroy: @davidfowl

davidfowl: Do it in another PR. I don't care that much.

src/SignalR/common/Shared/PipeWriterStream.cs
BrennanConroy: You made it netcoreapp3.0 only, we need it for netstandard2.0

src/SignalR/common/Shared/ValueTaskExtensions.cs
halter73: Why was this marked as resolved?

BrennanConroy: Because GitHub has bugs and this comment is for a previous commit on code that no longer exists

davidfowl (approved):

halter73 (approved):

PR #7489, opened 2019-02-12
Fix endpoint routing flaky test on build server in 2.2

JamesNK:
Addresses https://github.com/aspnet/AspNetCore-Internal/issues/1710

Note: In 2.2 branch. Test only change.

### DIFF
--- a/src/Http/Routing/test/UnitTests/Matching/DfaMatcherBuilderTest.cs
+++ b/src/Http/Routing/test/UnitTests/Matching/DfaMatcherBuilderTest.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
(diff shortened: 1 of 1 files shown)


src/Http/Routing/test/UnitTests/Matching/DfaMatcherBuilderTest.cs
```diff
@@ -231,7 +231,9 @@ public void BuildDfaTree_MultipleEndpoint_LiteralDifferentCase()
             Assert.Null(root.Parameters);
 
             var next = Assert.Single(root.Literals);
-            Assert.Equal("a", next.Key);
+            // Ignore case in 2.2 because EndpointComparer.ComparePattern does not specify a culture and running in the wrong culture
+            // will change the order of results. This causes 'A' to be used instead of 'a'
+            Assert.Equal("a", next.Key, ignoreCase: true);
```
rynowak: Is this really limited to just **this** line/comparison?

JamesNK: I'm not completely sure because I can't replicate it on my machine. If it isn't then we'll start to get new failures later in the test.

rynowak: Ah I see from looking at a bit more context. This might be enough. 

It looks like the `b1` and `b2` is already sort order independent as well.

rynowak (approved):

PR #7504, opened 2019-02-12
Replace custom targets for generating assembly attributes and source control info by using SDK features

natemcmaster:
Part of https://github.com/aspnet/BuildTools/issues/618 and preparation for #7280 

Changes:
* This disables the sourcelink and attribute generation targets from Internal.AspNetCore.Sdk
* Replace source linke with Microsoft.SourceLink.{GitHub, Vsts.Git} packages
* Replace Internal.AspNetCore.Sdk task which generates file version by using the values already computed in version.props
* Move some properties from Directory.Build.props to CSharp.Common.props if they only apply to C# projects
* Port some settings from Internal.AspNetCore.Sdk into Directory.Build.props. Except for some build tasks still used, we're almost ready to completely remove this package.
* add workaround for https://github.com/aspnet/AspNetCore/issues/7503


SteveSandersonMS (approved):
Components changes look good.

version.props
```diff
@@ -8,17 +8,18 @@
...
     <AspNetCoreModuleVersionMinor>$(AspNetCoreMinorVersion)</AspNetCoreModuleVersionMinor>
     <AspNetCoreModuleVersionRevision>$(AspNetCorePatchVersion)</AspNetCoreModuleVersionRevision>
 
     <OfficialBuildId Condition="'$(OfficialBuildId)' == ''">$(BUILD_BUILDNUMBER)</OfficialBuildId>
-    <OfficialBuildId Condition="'$(OfficialBuildId)' == '' AND '$(TEAMCITY_VERSION)' != ''">$([System.DateTime]::Now.ToString('yyyyMMdd')).$(BUILD_NUMBER)</OfficialBuildId>
+    <OfficialBuildId Condition="'$(OfficialBuildId)' == '' AND '$(IsOfficialBuild)' == 'true'">$([System.DateTime]::Now.ToString('yyyyMMdd')).1</OfficialBuildId>
   </PropertyGroup>
 
-  <PropertyGroup Condition=" '$(OfficialBuildId)' != '' AND '$(BuildNumberSuffix)' == '' ">
+  <PropertyGroup Condition=" '$(OfficialBuildId)' != '' ">
```
dougbu: Does this mean we can only build `-t000` packages locally now? Ugh

Directory.Build.props
```diff
@@ -156,6 +175,7 @@
   <Import Project="eng\ProjectReferences.props" />
   <Import Project="eng\targets\Cpp.Common.props"  Condition="'$(MSBuildProjectExtension)' == '.vcxproj'" />
   <Import Project="eng\targets\CSharp.Common.props"  Condition="'$(MSBuildProjectExtension)' == '.csproj'" />
+  <Import Project="eng\targets\FSharp.Common.props"  Condition="'$(MSBuildProjectExtension)' == '.fsproj'" />
```
dougbu: Why bother import an empty file?

Directory.Build.props
```diff
@@ -42,20 +33,45 @@
...
     <RepositoryUrl>https://github.com/aspnet/AspNetCore</RepositoryUrl>
     <RepositoryType>git</RepositoryType>
     <SharedSourceRoot>$(MSBuildThisFileDirectory)src\Shared\</SharedSourceRoot>
     <GoogleTestSubmoduleRoot>$(RepositoryRoot)src\submodules\googletest\</GoogleTestSubmoduleRoot>
+    <!-- Embed source files that are not tracked by the source control manager in the PDB. -->
+    <EmbedUntrackedSources>true</EmbedUntrackedSources>
+    <!-- Additional assembly attributes are already configured to include the source revision ID. -->
+    <IncludeSourceRevisionInInformationalVersion>false</IncludeSourceRevisionInInformationalVersion>
+
+    <DisableSourceLink Condition="'$(MSBuildProjectExtension)' == '.zipproj' OR '$(MSBuildProjectExtension)' == '.debproj' OR '$(MSBuildProjectExtension)' == '.rpmproj'">true</DisableSourceLink>
```
dougbu: Seems odd to me that `*.proj`, `*.javaproj`, `*.npmproj`, `*.pkgproj`, `*.shfxproj` , and `*.vcxproj` aren't included in this condition. Why not?

dougbu (approved):
Questions, questions&hellip;

version.props
natemcmaster: You can still override whatever you want, e.g. `build.cmd /p:IsOfficialBuild=true`, `build.cmd /p:BuildNumberSuffix=1234`, etc.  This change is mean to remove the part of this code that uses TeamCity environment variables.

Directory.Build.props
natemcmaster: The project types listed were excluded because adding the PackageReference causes NuGet to fail because they don't import Microsoft.Common.targets. We probably could disable source link on javaproj/npmproj, but I left them on enabled because they weren't causing issues.

Directory.Build.props
natemcmaster: will remove. Left over from earlier draft of this PR

PR #7508, opened 2019-02-12
Enable docker for the worker template

Tratcher:
#6817 The basic scenario works fine with the defaults.

Also, Phil said we didn't need `legacyTemplateIdentity` for new templates.


JunTaoLuo:
Not sure I have context on this so removing myself as a reviewer

davidfowl (approved):

PR #7510, opened 2019-02-12
 Add targets to generate the list of shared framework assemblies from project property

natemcmaster:
This will help later on when implementing reference assemblies, `FrameworkReference`, and to add checks which assert things references are used correctly. (see prototype in https://github.com/natemcmaster/AspNetCore/compare/vs2019...local-fx)

Changes:
* Add support for a property, `IsAspNetCoreApp`, in the .csproj file of assemblies which are part of the shared framework. 
* Remove unused dependencies
* Remove reference which have become part of 'netcoreapp3.0'

Reviewer: Split into several commits to make this easier to review


Directory.Build.props
```diff
@@ -154,6 +154,7 @@
   <Import Project="eng\Dependencies.props" />
   <Import Project="eng\PatchConfig.props" />
   <Import Project="eng\ProjectReferences.props" />
+  <Import Project="src\Framework\Microsoft.AspNetCore.App.props" />
```
pakrym: Not the greatest location for common properties. Why not keep it in eng?

src/Framework/Microsoft.AspNetCore.App.props
```diff
@@ -152,7 +79,26 @@
...
+    <ExternalAspNetCoreAppReference Include="@(_TransitiveExternalAspNetCoreAppReference)" />
+  </ItemGroup>
+
+  <!--
+    These compilation references are necessary to compile ns2.0 assemblies in the shared framework, but the runtime assemblies are part of netcoreapp3.0.
+    The preferred solution to reference problems with shared framework assemblies is to reference netcoreapp3.0 and not use System.* references,
+    but in some edge cases, this is not reasonable.
+  -->
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+    <_FrameworkProvidedReference Include="System.Buffers" />
```
pakrym: I don't completely understand this. Is this just to avoid producing errors about referencing `System.Buffers` ?

pakrym: Can we flow this information via ExternalAspNetCoreAppReference somehow?

Directory.Build.props
natemcmaster: Wasn't really thinking about that - i just left the file where it was. I'll move to eng/...that's a better place for it.

eng/targets/ResolveReferences.targets
```diff
@@ -75,6 +80,27 @@
     <Reference Remove="@(Reference)" />
   </ItemGroup>
 
+  <ItemDefinitionGroup>
+    <Reference>
+      <SharedSources>false</SharedSources>
```
pakrym: Can this be conditional on reference name? Or can we define this on `LatestPackageReference` definition level?

These packages are never going to be consumed in any other way so it would be nice to avoid declaring SharedSources on every reference

src/Framework/Microsoft.AspNetCore.App.props
natemcmaster: Correct. A few signalR assemblies target netstandard2.0 and need the System.Buffers reference, but this dependency is knocked out in netcoreapp3.0. I wanted to provide a way for us to require explicit action to put new assemblies into the shared framework.

pakrym (approved):
Minor comments

eng/targets/ResolveReferences.targets
natemcmaster: I would love to do this, but I don't know of a simple way in MSBuild to write an expression that does this: `Reference.Where(i => i.Identity.EndsWith(".Sources")).SetMetadata('PrivateAssets', 'All')`. We might be able to do this as an MSBuild task, but every other approach I've tried is expensive computationally.

src/Framework/Microsoft.AspNetCore.App.props
natemcmaster: I updated the name and comments to clarify that this is a "compilation only reference". This list should be fairly stable since netstandard2.0 is set in stone. The only thing which might change is what is in netcoreapp3.0.

eng/targets/ResolveReferences.targets
pakrym: You can set SharedSource metadata in https://github.com/aspnet/AspNetCore/pull/7510/files#diff-51f57d74bf38501ecf1b2a82e10577d7R121. Should be enough for the purposes you are using it for.

natemcmaster: That's where I would put it, but this isn't valid MSBuild syntax...
```
      <Reference>
        <PrivateAssets Condition="%(Identity.EndsWith('.Sources'))">All</PrivateAssets>
      </Reference>
```
Let me play with some MSBuild intrinsic functions and see if I can find a workaround.

PR #7511, opened 2019-02-12
Ignore unexpected stream items and completions from client

BrennanConroy:
Fixes https://github.com/aspnet/AspNetCore/issues/7069

Added a note to https://github.com/aspnet/AspNetCore/issues/7070 about sending a completion for unexpected stream items.


src/SignalR/server/SignalR/test/HubConnectionHandlerTests.cs
```diff
@@ -3087,37 +3058,24 @@ bool ExpectedErrors(WriteContext writeContext)
                     var simpleCompletion = Assert.IsType<CompletionMessage>(result);
                     Assert.Null(simpleCompletion.Result);
 
-                    // This will log an error on the server as the hub method has completed and will complete all associated streams
+                    // This will log a warning on the server as the hub method has completed and will complete all associated streams
                     await client.SendHubMessageAsync(new StreamItemMessage("id", "error!")).OrTimeout();
```
BrennanConroy: These 2 tests will be updated when https://github.com/aspnet/AspNetCore/issues/7070 is worked on

src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs
```diff
@@ -151,15 +155,34 @@ private Task ProcessStreamBindingFailure(HubConnectionContext connection, Stream
...
 
             return Task.CompletedTask;
         }
 
         private Task ProcessStreamItem(HubConnectionContext connection, StreamItemMessage message)
         {
-            Log.ReceivedStreamItem(_logger, message);
-            return connection.StreamTracker.ProcessItem(message);
+            var processTask = connection.StreamTracker.ProcessItem(message);
+            // Review: This is kind of ugly and relies on the fact that the exception will be thrown from the synchronous code path
```
davidfowl: Why are we doing this?

BrennanConroy: Currently if the client sends a stream item after the stream was closed we kill the connection. This is to prevent killing the connection.

src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs
```diff
@@ -357,12 +380,7 @@ async Task ExecuteInvocation()
             {
                 foreach (var stream in hubMessage.StreamIds)
                 {
-                    try
-                    {
-                        connection.StreamTracker.Complete(CompletionMessage.Empty(stream));
-                    }
-                    // ignore failures, it means the client already completed the streams
```
halter73: Might want to keep the comment

src/SignalR/server/SignalR/test/HubConnectionHandlerTests.cs
```diff
@@ -2887,13 +2887,7 @@ public async Task UploadStreamItemInvalidType()
...
         public async Task UploadStreamItemInvalidId()
         {
-            bool ExpectedErrors(WriteContext writeContext)
-            {
-                return writeContext.LoggerName == "Microsoft.AspNetCore.SignalR.HubConnectionHandler" &&
-                       writeContext.EventId.Name == "ErrorProcessingRequest";
-            }
-
-            using (StartVerifiableLog(ExpectedErrors))
+            using (StartVerifiableLog())
```
halter73: Does this test verify "StreamItemMessage received unexpectedly." got logged?

src/SignalR/server/SignalR/test/HubConnectionHandlerTests.cs
```diff
@@ -2906,24 +2900,16 @@ bool ExpectedErrors(WriteContext writeContext)
...
         public async Task UploadStreamCompleteInvalidId()
         {
-            bool ExpectedErrors(WriteContext writeContext)
-            {
-                return writeContext.LoggerName == "Microsoft.AspNetCore.SignalR.HubConnectionHandler" &&
-                       writeContext.EventId.Name == "ErrorProcessingRequest";
-            }
-
-            using (StartVerifiableLog(ExpectedErrors))
+            using (StartVerifiableLog())
```
halter73: ditto about checking logs

src/SignalR/server/SignalR/test/HubConnectionHandlerTests.cs
```diff
@@ -3087,37 +3058,24 @@ bool ExpectedErrors(WriteContext writeContext)
...
+
                     // Shut down
                     client.Dispose();
 
                     await connectionHandlerTask.OrTimeout();
                 }
             }
-
-            // Check that the stream has been completed by noting the existance of an error
-            Assert.True(errorLogged);
```
halter73: Can we assert the expected warning is logged?

src/SignalR/server/SignalR/test/HubConnectionHandlerTests.cs
```diff
@@ -3134,18 +3092,18 @@ bool ExpectedErrors(WriteContext writeContext)
...
```
halter73: ditto about verifying the warning

src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs
halter73: We should not be using async vs. sync exceptions for control flow. Even `Task<<Task>>` with inner vs outer exceptions for control is slightly better. Or maybe `bool StreamTracker.TryProcessItem(out Task)`.

halter73 (changes requested):

src/SignalR/server/SignalR/test/HubConnectionHandlerTests.cs
BrennanConroy: No, the verifiable log stuff checks that no unexpected `Error` logs appear in the test.

These tests will be better once we add the stream completion, because we'll verify that the completion was sent from the server on failure.

halter73: It's good that you're verifying there are no unexpected errors, but you should also verify "StreamItemMessage received unexpectedly." was logged so you know you are hitting the code path you are trying to test.

BrennanConroy: We know we're hitting the code path because the connection isn't being closed anymore.

We generally don't test log messages, I did it before because I couldn't think of anything better.

halter73: I understand not wanting to test logs. If you're not careful, it can definitely make tests unnecessarily fragile.

That being said, this PR description claims "We should instead ignore (and log) any stream activity after it has been closed", and I think if that's a major goal of this PR, we should verify it.

The tests these are replacing included lines like `Assert.True(errorLogged);` anyway, so it's not like we weren't testing the logs before. The old tests were also verifying the exception message matched exactly what was expected. Sharing resource strings with the test project could help make the test less fragile.

src/SignalR/common/Protocols.NewtonsoftJson/src/Protocol/NewtonsoftJsonHubProtocol.cs
```diff
@@ -348,12 +348,12 @@ private HubMessage ParseMessage(Utf8BufferTextReader textReader, IInvocationBind
                     case HubProtocolConstants.StreamItemMessageType:
                         if (itemToken != null)
                         {
-                            var returnType = binder.GetReturnType(invocationId);
                             try
                             {
-                                item = itemToken.ToObject(returnType, PayloadSerializer);
+                                var itemType = binder.GetStreamItemType(invocationId);
```
halter73: Do we have any tests that cover the case when this throws a KeyNotFoundException?

BrennanConroy: The `UploadStreamItemInvalidId` test is doing this, I'll take a look at our json protocols tests and add one if it's missing

halter73 (approved):

PR #7519, opened 2019-02-13
Fix “must comprised of”

jnm2:
The first commit fixes a grammatical error.
The second commit avoids the newer sense of the word ‘comprise’:

> Although it has been in use since the late 18th century, sense 2 is still attacked as wrong. Why it has been singled out is not clear, but until comparatively recent times it was found chiefly in scientific or technical writing rather than belles lettres. Our current evidence shows a slight shift in usage: sense 2 is somewhat more frequent in recent literary use than the earlier senses. You should be aware, however, that if you use sense 2 you may be subject to criticism for doing so, and you may want to choose a safer synonym such as _compose_ or _make up_.

https://www.merriam-webster.com/dictionary/comprise#usage-1-anchor

### DIFF
--- a/src/DataProtection/DataProtection/src/KeyManagement/IKeyEscrowSink.cs
+++ b/src/DataProtection/DataProtection/src/KeyManagement/IKeyEscrowSink.cs
@@ -21,7 +21,7 @@ namespace Microsoft.AspNetCore.DataProtection.KeyManagement
         /// Stores the given key material to the escrow service.
         /// </summary>
         /// <param name="keyId">The id of the key being persisted to escrow.</param>
-        /// <param name="element">The unencrypted XML element that comprises the key material.</param>
+        /// <param name="element">The unencrypted key material in XML form.</param>
         void Store(Guid keyId, XElement element);
     }
 }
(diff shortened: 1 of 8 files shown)


jnm2:
Oops, deleted the wrong branch in my fork. Sorry!

Eilon:
Hmm, does this change improve readability in a substantial way? Grammar/usage "rules" are typically descriptive, not prescriptive, so the question is: what wording would most people be able to mostly understand, as opposed to "does a rule exist somewhere that says this is right/wrong." I don't know the answer to that, though.

jnm2:
Yes, great point! Descriptivism is the only realistic or useful philosophy in my opinion. I'm in favor of whatever is the most readable or professional from your point of view. 

Would you prefer I leave the 'is comprised of' or find another wording, such as 'A combination of in the case of the enum flags?

Eilon:
#grammarnerds

I think the one that sounds most off to me is https://github.com/aspnet/AspNetCore/pull/7519/files#diff-bf4ec3339b253002eb8be7272b5eaea7 .

jnm2:
How about, "Gets or sets the minimum number of unique characters which a password must contain"?

Eilon:
Yeah that sounds great. Simpler English is good!

jnm2:
Okay, what do you think of the new changes?

Eilon (approved):
Looks great, thank you!

PR #7521, opened 2019-02-13
Make update.sh and restore.sh executable

JunTaoLuo:
Chmod +x so this script is executable

### DIFF
--- a/restore.sh
+++ b/restore.sh
@@ -3,5 +3,4 @@
 set -euo pipefail
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-repo_root="$DIR/../.."
-"$repo_root/build.sh" --all --restore --no-build "$@"
+"$DIR/build.sh" --all --restore --no-build "$@"


restore.sh
```diff
@@ -3,5 +3,4 @@
 set -euo pipefail
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-repo_root="$DIR/../.."
-"$repo_root/build.sh" --all --restore --no-build "$@"
+"$DIR/build.sh" --all --restore --no-build "$@"
```
JunTaoLuo: @natemcmaster Though this works, the script is different from the one in aspnet/Extensions any idea which one we should be using?

dougbu (dismissed):
Undo changes to `activate.sh`. That file must be sourced or it won't do anything useful.

restore.sh
natemcmaster: aspnet/Extensions's script is different because it is using a different build system, Arcade. We haven't converted this repo yet. (#7280)

natemcmaster (changes requested):
Can you revert the change to activate.sh? It should still be 644 (no executable permissions) because it can only function correctly if you run `source activate.sh`

JunTaoLuo:
Undid the change to activate.sh

natemcmaster (approved):

natemcmaster:
Thanks! Apparently no one has used this script in the last few weeks. Looks like I broke it a while ago.

PR #7525, opened 2019-02-13
FormFeature -Added exception for invalid content-disposition header 

JulijaRamoskiene:
Fix for: https://github.com/aspnet/AspNetCore/issues/7497
Added user-friendly exception when content-disposition header value is not valid.

### DIFF
--- a/src/Http/Http/src/Features/FormFeature.cs
+++ b/src/Http/Http/src/Features/FormFeature.cs
@@ -177,7 +177,7 @@ namespace Microsoft.AspNetCore.Http.Features
                         // Parse the content disposition here and pass it further to avoid reparsings
                         if (!ContentDispositionHeaderValue.TryParse(section.ContentDisposition, out var contentDisposition))
                         {
-                            throw new InvalidDataException("Form section has invalid Content-Disposition value.");
+                            throw new InvalidDataException("Form section has invalid Content-Disposition value: " + section.ContentDisposition);
                         }
 
                         if (contentDisposition.IsFileDisposition())
--- a/src/Http/Http/src/Features/FormFeature.cs
+++ b/src/Http/Http/src/Features/FormFeature.cs
@@ -176,7 +176,10 @@ namespace Microsoft.AspNetCore.Http.Features
                     {
                         // Parse the content disposition here and pass it further to avoid reparsings
                         ContentDispositionHeaderValue contentDisposition;
-                        ContentDispositionHeaderValue.TryParse(section.ContentDisposition, out contentDisposition);
+                        if (!ContentDispositionHeaderValue.TryParse (section.ContentDisposition, out contentDisposition))
+                        {
+                            throw new InvalidDataException ("Form section has invalid Content-Disposition value.");
+                        }
 
                         if (contentDisposition.IsFileDisposition())
                         {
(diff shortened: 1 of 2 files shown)


src/Http/Http/src/Features/FormFeature.cs
```diff
@@ -175,8 +175,10 @@ private async Task<IFormCollection> InnerReadFormAsync(CancellationToken cancell
                     while (section != null)
                     {
                         // Parse the content disposition here and pass it further to avoid reparsings
-                        ContentDispositionHeaderValue contentDisposition;
-                        ContentDispositionHeaderValue.TryParse(section.ContentDisposition, out contentDisposition);
+                        if (!ContentDispositionHeaderValue.TryParse (section.ContentDisposition, out var contentDisposition))
```
Tratcher:
```suggestion
                        if (!ContentDispositionHeaderValue.TryParse(section.ContentDisposition, out var contentDisposition))
```

src/Http/Http/src/Features/FormFeature.cs
```diff
@@ -175,8 +175,10 @@ private async Task<IFormCollection> InnerReadFormAsync(CancellationToken cancell
...
+                        {
+                            throw new InvalidDataException ("Form section has invalid Content-Disposition value.");
```
Tratcher:
```suggestion
                            throw new InvalidDataException("Form section has invalid Content-Disposition value.");
```

Tratcher (approved):

natemcmaster:
Hmm, not sure why it can't find this pipeline: https://dev.azure.com/dnceng/public/_build/index?definitionId=304&_a=completed

src/Http/Http/src/Features/FormFeature.cs
```diff
@@ -175,8 +175,10 @@ private async Task<IFormCollection> InnerReadFormAsync(CancellationToken cancell
                     while (section != null)
                     {
                         // Parse the content disposition here and pass it further to avoid reparsings
-                        ContentDispositionHeaderValue contentDisposition;
-                        ContentDispositionHeaderValue.TryParse(section.ContentDisposition, out contentDisposition);
+                        if (!ContentDispositionHeaderValue.TryParse(section.ContentDisposition, out var contentDisposition))
+                        {
+                            throw new InvalidDataException("Form section has invalid Content-Disposition value.");
```
davidfowl: nit: It should show the raw value in the exception message

JulijaRamoskiene: It would mean that untrusted (user provided) data is reflected in error message. It is not good security practice. Do you really want to do that here?

davidfowl: Yes, it should be in there. Exceptions don't make it back to the client at all in a proper application as they themselves have more revealing information (like the stack trace, line numbers etc).

JulijaRamoskiene: I added change since there are already many such exceptions with raw data. 
However, I think it would have been nice to keep InvalidDataException messages safe (defense in depth). 
I would have preferred  to return generic user friendly exception message and having raw value logged.

Tratcher:
Thanks

PR #7539, opened 2019-02-13
Automatically set 'PrivateAssets=All' for .Sources references

natemcmaster:
Follow-up to #7510 

When a `<reference>` ends in the name `.Sources`, set PrivateAssets=All (exclude from generated nuspec) and IncludeAssets=ContentFiles (only consume content files, not .dll's)

### DIFF
--- a/eng/targets/ResolveReferences.targets
+++ b/eng/targets/ResolveReferences.targets
@@ -87,12 +87,6 @@
     <Reference Remove="@(Reference)" />
   </ItemGroup>
 
-  <ItemDefinitionGroup>
-    <Reference>
-      <SharedSources>false</SharedSources>
-    </Reference>
-  </ItemDefinitionGroup>
-
   <!--
     This target resolves remaining Referene items to Packages, if possible. If not, they are left as Reference items fo the SDK to resolve.
     This target helps ensure projects within the shared framework do no unintentionally add new references,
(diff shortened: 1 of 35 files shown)


eng/targets/ResolveReferences.targets
```diff
@@ -119,8 +117,8 @@
 
       <!-- Ensure only content asset are consumed from .Sources packages -->
       <Reference>
-        <IncludeAssets Condition="'%(SharedSources)' == 'true'">ContentFiles</IncludeAssets>
-        <PrivateAssets Condition="'%(SharedSources)' == 'true'">All</PrivateAssets>
+        <IncludeAssets Condition="$([System.String]::new('%(Identity)').EndsWith('.Sources'))">ContentFiles</IncludeAssets>
```
pakrym: What about setting `SharedSources` metadata conditionally on `Condition="$([System.String]::new('%(Identity)').EndsWith('.Sources'))` and then using it in all other places.

pakrym (approved):

eng/targets/ResolveReferences.targets
natemcmaster: that would have worked too.

pakrym: Would allow setting `SharedSources` for references that do not end in `.Sources` and keep the logic in one place.

natemcmaster: updated.

pakrym (approved):

PR #7540, opened 2019-02-13
Touch up component detection

dougbu:
- run only on success of previous steps


natemcmaster (approved):

Eilon:
Why not always run?

dougbu:
> Why not always run?

Seems wasteful and the artifacts the detection looks at likely won't exist if the previous step fails.

PR #7551, opened 2019-02-13
More C++ Client cleanup

BrennanConroy:
`invoke_json` => `invoke`
`invoke_voud` => `send`

Ran the C++ Core guidelines analysis on the code and fixed up some of the easier issues it flagged.

### DIFF
--- a/src/SignalR/clients/cpp/src/signalrclient/connection_impl.cpp
+++ b/src/SignalR/clients/cpp/src/signalrclient/connection_impl.cpp
@@ -4,7 +4,6 @@
 #include "stdafx.h"
 #include <thread>
 #include <algorithm>
-#include "cpprest/asyncrt_utils.h"
 #include "constants.h"
 #include "connection_impl.h"
 #include "request_sender.h"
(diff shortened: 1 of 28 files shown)


src/SignalR/clients/cpp/src/signalrclient/connection_impl.cpp
```diff
@@ -39,7 +38,7 @@ namespace signalr
         std::unique_ptr<web_request_factory> web_request_factory, std::unique_ptr<transport_factory> transport_factory)
         : m_base_url(url), m_query_string(query_string), m_connection_state(connection_state::disconnected), m_logger(log_writer, trace_level),
         m_transport(nullptr), m_web_request_factory(std::move(web_request_factory)), m_transport_factory(std::move(transport_factory)),
-        m_message_received([](const utility::string_t&){}), m_disconnected([](){})
+        m_message_received([](const utility::string_t&) noexcept {}), m_disconnected([]() noexcept {})
```
halter73: Is this the result of following [this](https://docs.microsoft.com/en-us/visualstudio/code-quality/using-the-cpp-core-guidelines-checkers?view=vs-2017) guide?

Are we basically putting noexcept on every method/ctor/callback we can? This wouldn't virally force all callbacks to use the noexcept specifier, right? I hope adding noexcept is never breaking.

halter73 (approved):

src/SignalR/clients/cpp/src/signalrclient/connection_impl.cpp
BrennanConroy: Yes.

This would not force callbacks to use noexcept.

PR #7555, opened 2019-02-13
Add package conflict manifest data to the targeting pack

natemcmaster:
Example file this generates:

[Microsoft.AspNetCore.App.PackageOverrides.txt](https://github.com/aspnet/AspNetCore/files/2862555/Microsoft.AspNetCore.App.PackageOverrides.txt)

Part of https://github.com/aspnet/AspNetCore/issues/6501

cc @nguerrera @dsplaisted @dagood

### DIFF
--- a/eng/targets/ResolveReferences.targets
+++ b/eng/targets/ResolveReferences.targets
@@ -84,16 +84,14 @@
     <!--
       Turn Reference items into a ProjectReference when UseProjectReferences is true.
       Order matters. This comes before package resolution because projects should be used when possible instead of packages.
-
-      ProjectReferenceByAssemblyName can be used by project targets to determine which references were resolved to projects.
     -->
-    <ProjectReferenceByAssemblyName Condition="'$(UseProjectReferences)' == 'true'"
+    <_ProjectReferenceByAssemblyName Condition="'$(UseProjectReferences)' == 'true'"
       Include="@(ProjectReferenceProvider)"
       Exclude="@(_UnusedProjectReferenceProvider)" />
 
-    <ProjectReference Include="@(ProjectReferenceByAssemblyName->'%(ProjectPath)')" />
+    <ProjectReference Include="@(_ProjectReferenceByAssemblyName->'%(ProjectPath)')" />
 
-    <Reference Remove="@(ProjectReferenceByAssemblyName)" />
+    <Reference Remove="@(_ProjectReferenceByAssemblyName)" />
 
     <!-- Use _ReferenceTemp to workaround issues in Visual Studio which causes a conflict between Reference, packages, and projects. -->
     <_ReferenceTemp Include="@(Reference)" />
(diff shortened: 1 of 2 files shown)


pakrym (approved):

nguerrera:
File looks great. Filed https://github.com/dotnet/sdk/issues/2937 to consume it.

natemcmaster:
Per @pakrym's suggestion, switched this to use ReferencePath instead. This should better align with what RAR does.

Result:
[Microsoft.AspNetCore.App.PackageOverrides.txt](https://github.com/aspnet/AspNetCore/files/2862646/Microsoft.AspNetCore.App.PackageOverrides.txt)

nguerrera:
Doesn't this version assume all package references are 1:1 with assemblies? Is that guaranteed by anything?

natemcmaster:
It only makes that assumption for assemblies which build in this repo. And for this subset of stuff, yes, package id == assembly name.

Worth nothing - this list includes package IDs for assemblies we shipped as OOB packages in 2.x, but have dropped in 3.0. For instance, `Microsoft.AspNetCore.Mvc.Razor|3.0.0-preview3-t000`. We haven't, and never plan to, ship Microsoft.AspNetCore.Mvc.Razor for 3.0.

nguerrera:
I'm seeing stuff that is not in the repo in the generated file. Isn't it the *project* references that are in this repo?

nguerrera:
```
Newtonsoft.Json|12.0.1
System.IO.Pipelines|4.6.0-preview.19106.2
System.Runtime.CompilerServices.Unsafe|4.6.0-preview.19106.2
System.Security.AccessControl|4.6.0-preview.19106.2
System.Security.Cryptography.Cng|4.6.0-preview.19106.2
System.Security.Cryptography.Xml|4.6.0-preview.19106.2
System.Security.Permissions|4.6.0-preview.19106.2
System.Security.Principal.Windows|4.6.0-preview.19106.2
System.Text.Encodings.Web|4.6.0-preview.19106.2
```

nguerrera:
Unless I'm misunderstanding, this would generate dupes if a package had more than one assembly:
```
 <_AspNetCoreAppPackageOverrides Include="@(ReferencePath->'%(NuGetPackageId)|%(NuGetPackageVersion)')" Condition=" '%(ReferencePath.NuGetPackageId)' != 'Microsoft.NETCore.App' AND '%(ReferencePath.NuGetSourceType)' == 'Package' " />
```

nguerrera:
Not sure there are any such packages today or if dupes will break anything, but it seems like something I'd question if I ever had to revisit this code.

nguerrera:
> Worth noting - this list includes package IDs for assemblies we shipped as OOB packages in 2.x, but have dropped in 3.0. For instance, `Microsoft.AspNetCore.Mvc.Razor|3.0.0-preview3-t000`. We haven't, and never plan to, ship Microsoft.AspNetCore.Mvc.Razor for 3.0.

I think this is fine. This says that anything less than that version is overridden by this package, which is true.

dsplaisted:
@natemcmaster We're thinking that data files for targeting and runtime packs should go in the `data/` folder of the package.  So I'd suggest that the path for this file should be `data/PackageOverrides.txt`.

natemcmaster:
> I'm seeing stuff that is not in the repo in the generated file.

That's intentional. We include some assemblies like System.IO.Pipelines in the aspnet targeting pack. They are built by corefx, but for reasons, they didn't want it in microsoft.netcore.app.

nguerrera:
My point is that it looks like you are assuming things outside this repo have 1 package:1 assembly. I doubt that will change in practice. But that was my point.

natemcmaster:
Makes sense. So far, this assumption has been correct for all assemblies currently in the targeting pack. We don't really have a good way to test this, so it could break at some point 😬

natemcmaster:
Per your request, I've updated the PR to put the manifest in `data/PackageOverrides.txt`

PR #7558, opened 2019-02-13
Update benchmarks.yml

sebastienros:
Trying to make benchmarks build def to work again

### DIFF
--- a/.azure/pipelines/benchmarks.yml
+++ b/.azure/pipelines/benchmarks.yml
@@ -1,8 +1,5 @@
 # This configuration builds the repository and runs benchmarks
 
-# Don't run CI for this config
-trigger: none
-
 jobs:
 - template: jobs/default-build.yml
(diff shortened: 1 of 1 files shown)


.azure/pipelines/benchmarks.yml
```diff
@@ -1,8 +1,5 @@
 # This configuration builds the repository and runs benchmarks
 
-# Don't run CI for this config
-trigger: none
```
natemcmaster: The unfortunate consequence of this will be that this build definition will run for all pushes to master, not just PRs.

You might try this instead:

```yml
trigger: none

pr:
- '*'
```

sebastienros: It shouldn't because we disabled it from the UI. It worked as expected before adding this line. But I can try.

natemcmaster (approved):

PR #7574, opened 2019-02-14
[Components] Introduce ComponentEnvironment

javiercn:
* Component environment is an abstraction that hosts the primitives that
  are specific to a given rendering environment.
* Rendering environments (Remote, Prerrender, Local) are responsible for
  creating an initalizing the ComponentEnvironment.
* The ComponentEnvironment needs to be available on DI across environments.
* It is a scoped service on server-side environments and a singleton
  service on the web assembly case.
* If a service requires the ComponentEnvironment it can't be a singleton
  (at least on server environments)
* We might choose to make the ComponentEnvironment on the WebAssembly
  case a singleton too.

### DIFF
--- a/src/Components/Blazor/Blazor/src/Hosting/WebAssemblyHostBuilder.cs
+++ b/src/Components/Blazor/Blazor/src/Hosting/WebAssemblyHostBuilder.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Net.Http;
 using Microsoft.AspNetCore.Blazor.Services;
+using Microsoft.AspNetCore.Components.Environment;
 using Microsoft.AspNetCore.Components.Services;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.JSInterop;
@@ -91,7 +92,9 @@ namespace Microsoft.AspNetCore.Blazor.Hosting
             services.AddSingleton<IJSRuntime, MonoWebAssemblyJSRuntime>();
 
             services.AddSingleton<IUriHelper>(WebAssemblyUriHelper.Instance);
-            services.AddSingleton<HttpClient>(s =>
+            services.AddSingleton<ComponentEnvironment, WebAssemblyComponentEnvironment>();
+
+            services.AddSingleton(s =>
             {
                 // Creating the URI helper needs to wait until the JS Runtime is initialized, so defer it.
                 var uriHelper = s.GetRequiredService<IUriHelper>();
@@ -108,6 +111,16 @@ namespace Microsoft.AspNetCore.Blazor.Hosting
 
             var builder = _serviceProviderFactory.CreateBuilder(services);
             _appServices = _serviceProviderFactory.CreateServiceProvider(builder);
+        }
+
+        private class WebAssemblyComponentEnvironment : ComponentEnvironment
+        {
+            public WebAssemblyComponentEnvironment(IJSRuntime jsRuntime, IUriHelper uriHelper)
+            {
+                Name = Local;
+                JSRuntime = jsRuntime;
+                UriHelper = uriHelper;
+            }
         }
     }
 }
--- a/src/Components/Components/src/Environment/ComponentEnvironment.cs
+++ b/src/Components/Components/src/Environment/ComponentEnvironment.cs
@@ -1,0 +1,44 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Components.Services;
+using Microsoft.JSInterop;
+
+namespace Microsoft.AspNetCore.Components.Environment
+{
+    /// <summary>
+    /// Provides information an environment specific features about the environment the components are running on.
+    /// </summary>
+    public class ComponentEnvironment
+    {
+        /// <summary>
+        /// Name of associated with the <see cref="ComponentEnvironment"/> when components are being remotely rendered on the server.
+        /// </summary>
+        public const string Remote = nameof(Remote);
+
+        /// <summary>
+        /// Name of associated with the <see cref="ComponentEnvironment"/> when components are being prerrendered on the server.
+        /// </summary>
+        public const string Prerrender = nameof(Prerrender);
+
+        /// <summary>
+        /// Name of associated with the <see cref="ComponentEnvironment"/> when components are being locally rendered on the client.
+        /// </summary>
+        public const string Local = nameof(Local);
+
+        /// <summary>
+        /// Gets or sets identifier associated with the <see cref="ComponentEnvironment"/> the components are running on.
+        /// </summary>
+        public string Name { get; set; }
+
+        /// <summary>
+        /// Gets or sets the <see cref="IJSRuntime"/> for this <see cref="ComponentEnvironment"/>.
+        /// </summary>
+        public IJSRuntime JSRuntime { get; set; }
+
+        /// <summary>
+        /// Gets or sets the <see cref="IUriHelper"/> for this <see cref="ComponentEnvironment"/>.
+        /// </summary>
+        public IUriHelper UriHelper { get; set; }
+    }
+}
--- a/src/Components/Server/src/Circuits/DefaultCircuitFactory.cs
+++ b/src/Components/Server/src/Circuits/DefaultCircuitFactory.cs
@@ -5,6 +5,7 @@ using System;
 using System.Linq;
 using Microsoft.AspNetCore.Components.Browser;
 using Microsoft.AspNetCore.Components.Browser.Rendering;
+using Microsoft.AspNetCore.Components.Environment;
 using Microsoft.AspNetCore.Components.Rendering;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.SignalR;
@@ -44,13 +45,15 @@ namespace Microsoft.AspNetCore.Components.Server.Circuits
             }
 
             var scope = _scopeFactory.CreateScope();
-            var jsRuntime = new RemoteJSRuntime(client);
+            var environment = scope.ServiceProvider.GetRequiredService<ComponentEnvironment>();
+            InitializeEnvironment(environment, client);
+
             var rendererRegistry = new RendererRegistry();
             var dispatcher = Renderer.CreateDefaultDispatcher();
             var renderer = new RemoteRenderer(
                 scope.ServiceProvider,
                 rendererRegistry,
-                jsRuntime,
+                environment.JSRuntime,
                 client,
                 dispatcher,
                 _loggerFactory.CreateLogger<RemoteRenderer>());
@@ -65,14 +68,23 @@ namespace Microsoft.AspNetCore.Components.Server.Circuits
                 rendererRegistry,
                 renderer,
                 config,
-                jsRuntime,
+                environment.JSRuntime,
                 circuitHandlers);
 
             // Initialize per-circuit data that services need
-            (circuitHost.Services.GetRequiredService<IJSRuntimeAccessor>() as DefaultJSRuntimeAccessor).JSRuntime = jsRuntime;
+#pragma warning disable CS0618 // Type or member is obsolete
+            (circuitHost.Services.GetRequiredService<IJSRuntimeAccessor>() as DefaultJSRuntimeAccessor).JSRuntime = environment.JSRuntime;
+#pragma warning restore CS0618 // Type or member is obsolete
             (circuitHost.Services.GetRequiredService<ICircuitAccessor>() as DefaultCircuitAccessor).Circuit = circuitHost.Circuit;
 
             return circuitHost;
+        }
+
+        private static void InitializeEnvironment(ComponentEnvironment environment, IClientProxy client)
+        {
+            environment.Name = ComponentEnvironment.Remote;
+            environment.JSRuntime = new RemoteJSRuntime(client);
+            environment.UriHelper = new RemoteUriHelper(environment.JSRuntime);
         }
     }
 }
--- a/src/Components/test/E2ETest/Infrastructure/SeleniumStandaloneServer.cs
+++ b/src/Components/test/E2ETest/Infrastructure/SeleniumStandaloneServer.cs
@@ -123,7 +123,7 @@ namespace Microsoft.AspNetCore.Components.E2ETest.Infrastructure
                     output = builder.ToString();
                 }
 
-                throw new InvalidOperationException($"Failed to start selenium sever. {Environment.NewLine}{output}", ex.GetBaseException());
+                throw new InvalidOperationException($"Failed to start selenium sever. {System.Environment.NewLine}{output}", ex.GetBaseException());
             }
         }
 
--- a/src/Mvc/src/Microsoft.AspNetCore.Mvc.ViewFeatures/HtmlHelperComponentExtensions.cs
+++ b/src/Mvc/src/Microsoft.AspNetCore.Mvc.ViewFeatures/HtmlHelperComponentExtensions.cs
@@ -1,11 +1,11 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Collections.Generic;
-using System.IO;
-using System.Text.Encodings.Web;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Components;
+using Microsoft.AspNetCore.Components.Environment;
 using Microsoft.AspNetCore.Components.Rendering;
 using Microsoft.AspNetCore.Html;
 using Microsoft.AspNetCore.Mvc.Rendering;
@@ -27,7 +27,7 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures
         {
             if (htmlHelper == null)
             {
-                throw new System.ArgumentNullException(nameof(htmlHelper));
+                throw new ArgumentNullException(nameof(htmlHelper));
             }
 
             return htmlHelper.RenderComponentAsync<TComponent>(null);
@@ -46,11 +46,19 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures
         {
             if (htmlHelper == null)
             {
-                throw new System.ArgumentNullException(nameof(htmlHelper));
+                throw new ArgumentNullException(nameof(htmlHelper));
             }
 
-            var serviceProvider = htmlHelper.ViewContext.HttpContext.RequestServices;
-            var encoder = serviceProvider.GetRequiredService<HtmlEncoder>();
+            var httpContext = htmlHelper.ViewContext.HttpContext;
+            var serviceProvider = httpContext.RequestServices;
+            var prerrenderContext = serviceProvider.GetRequiredService<MvcPrerrenderingContext>();
+            var encoder = prerrenderContext.Encoder;
+            prerrenderContext.Environment.Name = ComponentEnvironment.Prerrender;
+            // Add our custom JSRuntime that throws upon invocation. We might need an option here to not throw or log a warning
+            // message instead.
+            prerrenderContext.Environment.JSRuntime = new UnsupportedJavaScriptRuntime();
+            prerrenderContext.Environment.UriHelper = new HttpUriHelper(httpContext);
+
             var dispatcher = Renderer.CreateDefaultDispatcher();
             using (var htmlRenderer = new HtmlRenderer(serviceProvider, encoder.Encode, dispatcher))
             {
@@ -58,26 +66,7 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures
                     parameters == null ?
                         ParameterCollection.Empty :
                         ParameterCollection.FromDictionary(HtmlHelper.ObjectToDictionary(parameters))));
-
                 return new ComponentHtmlContent(result);
-            }
-        }
-
-        private class ComponentHtmlContent : IHtmlContent
-        {
-            private readonly IEnumerable<string> _componentResult;
-
-            public ComponentHtmlContent(IEnumerable<string> componentResult)
-            {
-                _componentResult = componentResult;
-            }
-
-            public void WriteTo(TextWriter writer, HtmlEncoder encoder)
-            {
-                foreach (var element in _componentResult)
-                {
-                    writer.Write(element);
-                }
             }
         }
     }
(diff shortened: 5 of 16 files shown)


src/Components/Components/src/Environment/ComponentEnvironment.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// <summary>
+        /// Gets or sets the <see cref="IJSRuntime"/> for this <see cref="ComponentEnvironment"/>.
+        /// </summary>
+        public IJSRuntime JSRuntime { get; set; }
+
+        /// <summary>
+        /// Gets or sets the <see cref="IUriHelper"/> for this <see cref="ComponentEnvironment"/>.
+        /// </summary>
+        public IUriHelper UriHelper { get; set; }
+    }
```
javiercn: We might tighten up this in the future and add some extension methods to make some things easier. For example:
```csharp
environment.IsPrerrendering()
```

src/Components/Blazor/Blazor/src/Hosting/WebAssemblyHostBuilder.cs
```diff
@@ -91,7 +92,9 @@ private void CreateServiceProvider()
             services.AddSingleton<IJSRuntime, MonoWebAssemblyJSRuntime>();
 
             services.AddSingleton<IUriHelper>(WebAssemblyUriHelper.Instance);
-            services.AddSingleton<HttpClient>(s =>
+            services.AddSingleton<ComponentEnvironment, WebAssemblyComponentEnvironment>();
```
javiercn: We might want to make this Scoped for consistency with server-side scenarios, but it doesn't really make a difference.

src/Components/Server/src/Circuits/DefaultCircuitFactory.cs
```diff
@@ -65,14 +68,23 @@ public override CircuitHost CreateCircuitHost(HttpContext httpContext, IClientPr
...
                 renderer,
                 config,
-                jsRuntime,
+                environment.JSRuntime,
                 circuitHandlers);
 
             // Initialize per-circuit data that services need
-            (circuitHost.Services.GetRequiredService<IJSRuntimeAccessor>() as DefaultJSRuntimeAccessor).JSRuntime = jsRuntime;
+#pragma warning disable CS0618 // Type or member is obsolete
+            (circuitHost.Services.GetRequiredService<IJSRuntimeAccessor>() as DefaultJSRuntimeAccessor).JSRuntime = environment.JSRuntime;
```
javiercn: Can we kill this?

src/Components/Components/src/Environment/ComponentEnvironment.cs
```diff
@@ -0,0 +1,44 @@
...
+
+using Microsoft.AspNetCore.Components.Services;
+using Microsoft.JSInterop;
+
+namespace Microsoft.AspNetCore.Components.Environment
+{
+    /// <summary>
+    /// Provides information an environment specific features about the environment the components are running on.
+    /// </summary>
+    public class ComponentEnvironment
```
javiercn: We might want to extend this with some sort of FeatureCollection like functionality so that different environments can save here their own custom abstractions. Like for example the Circuit on the case of the Remote rendering scenario.

src/Mvc/src/Microsoft.AspNetCore.Mvc.ViewFeatures/HtmlHelperComponentExtensions.cs
```diff
@@ -46,39 +46,28 @@ public static class HtmlHelperComponentExtensions
...
             {
-                throw new System.ArgumentNullException(nameof(htmlHelper));
+                throw new ArgumentNullException(nameof(htmlHelper));
             }
 
-            var serviceProvider = htmlHelper.ViewContext.HttpContext.RequestServices;
-            var encoder = serviceProvider.GetRequiredService<HtmlEncoder>();
+            var httpContext = htmlHelper.ViewContext.HttpContext;
+            var serviceProvider = httpContext.RequestServices;
+            var prerrenderContext = serviceProvider.GetRequiredService<MvcPrerrenderingContext>();
```
javiercn: MvcPrerrenderingContext This is just a container to get all the services in one go.

src/Mvc/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Infrastructure/HttpUriHelper.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        private string GetContextBaseUri()
+        {
+            var request = _context.Request;
+            return UriHelper.BuildAbsolute(request.Scheme, request.Host, request.PathBase);
+        }
+
+        protected override void NavigateToCore(string uri, bool forceLoad)
+        {
+            // For now throw.
```
javiercn: I'm specifically avoiding dealing with redirects as of right now so that we can discuss how to better handle this post preview3.

src/Mvc/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Infrastructure/MvcPrerrenderingContext.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            // It's safe to initialize the environment here as this is a scoped service, equivalent to ComponentEnvironment
+            InitializeEnvironment(environment);
+            Environment = environment;
+            Encoder = encoder;
+        }
+
+        private void InitializeEnvironment(ComponentEnvironment environment)
+        {
+            environment.Name = ComponentEnvironment.Prerrender;
```
javiercn: This is ok as both services are scoped.

src/Mvc/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Infrastructure/UnsupportedJavaScriptRuntime.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Threading.Tasks;
+using Microsoft.JSInterop;
+
+namespace Microsoft.AspNetCore.Mvc.ViewFeatures
+{
+    internal class UnsupportedJavaScriptRuntime : IJSRuntime
+        {
+            public Task<T> InvokeAsync<T>(string identifier, params object[] args)
+            {
+                throw new InvalidOperationException("JavaScript interoperability is not supported in the current environment.");
```
javiercn: We throw because there's no better thing to do and this way at least you get a chance to see the issue early (hopefully during development).

Writing functional in memory tests for prerrendering should be pretty trivial with WebApplicationFactory and MVC.

src/Components/Server/src/Circuits/DefaultCircuitFactory.cs
```diff
@@ -44,13 +45,15 @@ public override CircuitHost CreateCircuitHost(HttpContext httpContext, IClientPr
             }
 
             var scope = _scopeFactory.CreateScope();
-            var jsRuntime = new RemoteJSRuntime(client);
+            var environment = scope.ServiceProvider.GetRequiredService<ComponentEnvironment>();
+            InitializeEnvironment(environment, client);
```
pranavkm: It's really unusual to see state on a service being mutated after initialization. If the intent is to tell components where they're being rendered, should it be on the `Renderer`?

src/Components/Components/src/Environment/ComponentEnvironment.cs
rynowak: We have DI for that though. I don't think that it's needs to be that complicated

src/Components/Components/src/Environment/ComponentEnvironment.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        /// <summary>
+        /// Gets or sets the <see cref="IJSRuntime"/> for this <see cref="ComponentEnvironment"/>.
+        /// </summary>
+        public IJSRuntime JSRuntime { get; set; }
+
+        /// <summary>
+        /// Gets or sets the <see cref="IUriHelper"/> for this <see cref="ComponentEnvironment"/>.
+        /// </summary>
+        public IUriHelper UriHelper { get; set; }
```
rynowak: I really don't think we should put services available from DI onto a context object. Anyone who needs these things should get them from DI.

src/Components/Server/src/Circuits/DefaultCircuitFactory.cs
```diff
@@ -44,13 +45,15 @@ public override CircuitHost CreateCircuitHost(HttpContext httpContext, IClientPr
...
+
             var rendererRegistry = new RendererRegistry();
             var dispatcher = Renderer.CreateDefaultDispatcher();
             var renderer = new RemoteRenderer(
                 scope.ServiceProvider,
                 rendererRegistry,
-                jsRuntime,
+                environment.JSRuntime,
```
rynowak: Yeah this is not right. The circuit host is the owner of this state.

rynowak:
I started reviewing this and adding comments, but I'd really like to talk about the goals and scenarios for this first.

> * Component environment is an abstraction that hosts the primitives that
  are specific to a given rendering environment.

I don't understand what this accomplishes in practice. We have DI for that purpose and it's working well for providing access already.

src/Components/Components/src/Environment/ComponentEnvironment.cs
javiercn: That's a fair point.

src/Components/Server/src/Circuits/DefaultCircuitFactory.cs
javiercn: The circuit is the one that sets up the environment a few lines above. It's still the owner.

src/Components/Components/src/Environment/ComponentEnvironment.cs
javiercn: You can still get them through DI, but we need a way to provide different implementations whether you are rendering components on a prerrendering environment or on a non prerrendered environment, as they require different behaviors.

javiercn:
> I started reviewing this and adding comments, but I'd really like to talk about the goals and scenarios for this first.
> 
> > * Component environment is an abstraction that hosts the primitives that
> >   are specific to a given rendering environment.
> 
> I don't understand what this accomplishes in practice. We have DI for that purpose and it's working well for providing access already.

* We want all services for components to be available through DI
  * IUriHeper
  * IJSRuntime
* We need different implementations of some of those services provided 
  through DI depending on whether we are prerrendering or not as they need to provide different answers to problems.
* We want to be able to use components from MVC without extra ceremony. 
  * You can create an existing MVC/Razor pages application.
  * Create a component in a .razor file and use it from your page or view and that should just work.

Using the component environment achieves this by:
* Making it responsible for providing the abstractions that are environment specific.
* Making the "component host" responsible for initializing the environment before starting the rendering process with the primitives specific to that environment (e.g prerrendering).
* Providing direct access to the environment specific abstractions though DI by way of retrieving them from the host.

src/Components/Blazor/Blazor/src/Hosting/WebAssemblyHostBuilder.cs
SteveSandersonMS: Is it necessary to remove the `HttpClient` service?

Will this alone stop people from putting `@inject HttpClient Http` in their components? We have this all over our samples and tutorials and definitely don't want to break it if at all possible. Can there be an `HttpClient` service that gets the value from the `ComponentEnvironment`?

src/Components/Components/src/Environment/ComponentEnvironment.cs
```diff
@@ -0,0 +1,44 @@
...
+
+namespace Microsoft.AspNetCore.Components.Environment
+{
+    /// <summary>
+    /// Provides information an environment specific features about the environment the components are running on.
+    /// </summary>
+    public class ComponentEnvironment
+    {
+        /// <summary>
+        /// Name of associated with the <see cref="ComponentEnvironment"/> when components are being remotely rendered on the server.
```
SteveSandersonMS: Typo: `Name of associated with the`?

Also it would be good not to specify `on the server`, because we'd also use this one in Electron.

SteveSandersonMS: Same typo in other comments below.

src/Components/Components/src/Environment/ComponentEnvironment.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// </summary>
+        public const string Remote = nameof(Remote);
+
+        /// <summary>
+        /// Name of associated with the <see cref="ComponentEnvironment"/> when components are being prerrendered on the server.
+        /// </summary>
+        public const string Prerrender = nameof(Prerrender);
```
SteveSandersonMS: Typo: `Prerrender` should be `Prerender`

src/Components/Components/src/Environment/ComponentEnvironment.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        /// <summary>
+        /// Name of associated with the <see cref="ComponentEnvironment"/> when components are being locally rendered on the client.
```
SteveSandersonMS: Don't need to specify `on the client`. It's enough to say `being rendered locally`, since theoretically you could do local rendering anywhere if you really wanted.

src/Components/Components/src/Environment/ComponentEnvironment.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// </summary>
+        public const string Local = nameof(Local);
+
+        /// <summary>
+        /// Gets or sets identifier associated with the <see cref="ComponentEnvironment"/> the components are running on.
+        /// </summary>
+        public string Name { get; set; }
```
SteveSandersonMS: I'm curious about why these properties are publicly writable. When someone obtains an instance of this, is there a use case for mutating it?

src/Components/Components/src/Environment/ComponentEnvironment.cs
SteveSandersonMS: If they are defined as non-singletons (presumably, scoped), can't the DI system conditionally return different values depending on whether we're in a prerendering context or not?

SteveSandersonMS: BTW I've nothing against the existence of `ComponentEnvironment` itself. It's certainly a nice way for components to find out whether they are currently prerendering or not. If we could make other services injectable independently of it, that would be great.

src/Components/Blazor/Blazor/src/Hosting/WebAssemblyHostBuilder.cs
SteveSandersonMS: OK I see it looks like you are registering services that do get `IJSRuntime` (etc.) from the component environment, so it's not breaking for those :)

However maybe it is breaking for `HttpClient`. Is it problematic to retain this service? Do we need to add it to the remote/prerendering cases too for consistency?

src/Components/test/E2ETest/Infrastructure/SeleniumStandaloneServer.cs
```diff
@@ -123,7 +123,7 @@ void LogOutput(object sender, DataReceivedEventArgs e)
                     output = builder.ToString();
                 }
 
-                throw new InvalidOperationException($"Failed to start selenium sever. {Environment.NewLine}{output}", ex.GetBaseException());
+                throw new InvalidOperationException($"Failed to start selenium sever. {System.Environment.NewLine}{output}", ex.GetBaseException());
```
SteveSandersonMS: The fact that it's necessary to qualify all uses of `Environment` makes me wonder if this is going to bother customers, too. Should we consider renaming the `Microsoft.AspNetCore.Components.Environment` namespace?

src/Mvc/src/Microsoft.AspNetCore.Mvc.ViewFeatures/HtmlHelperComponentExtensions.cs
```diff
@@ -46,39 +46,28 @@ public static class HtmlHelperComponentExtensions
...
+            var encoder = prerrenderContext.Encoder;
+            prerrenderContext.Environment.Name = ComponentEnvironment.Prerrender;
+            // Add our custom JSRuntime that throws upon invocation. We might need an option here to not throw or log a warning
+            // message instead.
```
SteveSandersonMS: At least for now, I'd be strongly in favour of throwing like you have here. Silently ignoring JS interop calls would almost always produce unhelpful behavior. People who are doing JS interop in combination with prerendering need to think through how they want to change their application behavior/output in the prerendering case, and the exception forces them to do that.

src/Mvc/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Infrastructure/HttpUriHelper.cs
SteveSandersonMS: > The bigger (not easily solvable problem) is that the response might have already
> started (for example when the page/view has already been flushed)

Is that really possible? Am I right in thinking it's only possible *after* the initial async render? I know that components can include arbitrary async logic that runs on a separate thread (or non-awaited continuation) and tries to trigger a redirect after the page has been rendered. But if we impose the rule "we only listen to your attempts to navigate while the initial async render is in progress", we can be sure it hasn't started writing to the response, right? Because we don't even start transforming the rendertree into HTML until the initial async render is done.

src/Mvc/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Infrastructure/HttpUriHelper.cs
```diff
@@ -0,0 +1,71 @@
...
+            // We don't have a good way to do redirects on a prerrendering context.
+            // Currently there's no good way to stop the rendering process. We can fix
+            // this by passing in a cancellation token to RenderRootComponentAsync and
+            // having that checked before we await on every ProcessAsynchronousWork iteration
```
SteveSandersonMS: Cancelling on redirection would be a nice-to-have, but even if we don't do that, it would still be possible to send the redirection response and then ignore the subsequent completion of async prerendering.

src/Mvc/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Infrastructure/HttpUriHelper.cs
```diff
@@ -0,0 +1,71 @@
...
+            // and do _context.Response.Redirect(uri)
+            // The bigger (not easily solvable problem) is that the response might have already
+            // started (for example when the page/view has already been flushed) which will mean
+            // that the behavior of this feature is not reliable, and not easily to diagnose as
+            // the developer only notices it when this happens at runtime; something we don't really want.
+            throw new InvalidOperationException(
+                "Redirects are not supported on a prerrendering environment.");
+
+            // We need to do something here about the hash to prevent infinite redirects
+            // as the hash portion of a url doesn't get sent to the browser.
```
SteveSandersonMS: Typo: `get sent to the browser` should be `get sent to the server`

src/Mvc/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Infrastructure/HttpUriHelper.cs
```diff
@@ -0,0 +1,71 @@
...
+            // This means that the hash won't be respected for prerrendering scenarios.
+            // There are several involved things that we could do. Bake the knowledge into
+            // the IUriHelper abstraction so that we can produce a special link where we
+            // encode the hash as a query stirng parameter.
+            // We would preprocess this on the server to set the appropriate url.
+            // For example: 
+            // CurrentUrl = "http://www.localhost/base/path?query=value#top
+            // Redirect = "http://www.localhost/base/path?query=value#first
```
SteveSandersonMS: I'm not sure how this scenario would occur in realistic use cases. We're not doing hash-based routing by default. Anyone who tries to use URL hashes in routing will have to deal with the fact that the server doesn't see it, just like they do in (say) MVC actions.

src/Components/Blazor/Blazor/src/Hosting/WebAssemblyHostBuilder.cs
javiercn: i didn't remove httpclient. The diff is doing weird things I believe.

src/Mvc/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Infrastructure/HttpUriHelper.cs
javiercn: But this can happen inside an MVC application where you could have already called FlushAsync to start sending the head to the page (to speed-up script loading) for example.

We might be Ok by checking response.started or something similar, but I want to think about it before baking it in

src/Components/Server/src/Circuits/DefaultCircuitFactory.cs
javiercn: No, we could do it all in services, but some of these things require access to the httpcontext

src/Mvc/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Infrastructure/UnsupportedJavaScriptRuntime.cs
SteveSandersonMS: For extra clarity, could the message say something like:

```
JavaScript interop calls cannot be issued during server-side prerendering, because the page has not yet loaded in the browser. Prerendered components must wrap any JavaScript interop calls in conditional logic to ensure those interop calls are not attempted during prerendering.
```

Maybe rename the type to `PrerenderingUnsupportedJavaScriptRuntime` if that helps with having such a specific error message.

src/Mvc/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Infrastructure/HttpUriHelper.cs
SteveSandersonMS: Oh, that scenario sounds fine. The rules around when you can issue redirections in MVC are well established (i.e., not after you've started flushing output), so this wouldn't be any different. As long as it only fails when the developer has explicitly taken steps to start flushing the output, I think we're fine.

rynowak:
> * We need different implementations of some of those services provided through DI depending on whether we are prerrendering or not as they need to provide different answers to problems.

Thanks for clarifying, that's the kind of information I was looking 

So it sounds like the challenge here is that these services need to behave differently at runtime depending on whether or not you're prerendering. This sounds like it's an implementation detail though no? We don't really want the developer to be concerned with *what* `IUrlHelper` they are getting, but rather they just ask for it and it appears.

I'm reacting strongly and negatively to the `ComponentEnvironment` as a public type with accessors for other services on it. We have already proved out that DI works well for components, I don't want to change the guidance. It sounds to me like `ComponentEnvironment` is designed to inform application about what *host* they are running in, and if you take away the services part it does that well.

I'd like to consider other solutions for resolving the behaviour of services - I think what's being proposed here leaks an implementation detail into the user experience. 


It's **dumb** but might work perfectly... what if server implementations of `IUrlHelper` and `IJsRuntime` started with an `if (environment.IsPrerendering())` - is that a totally bogus approach? Basically I'm asking you to take my on the journey, what if we start with the simplest solution and fix problems with it until it's good?

javiercn:
Closing this PR as it has been superseeded

PR #7577, opened 2019-02-14
Update corefx and core-setup

pakrym:
The manually aligned set of versions.


natemcmaster (changes requested):
Our master branch is still building for preview3. This would move us forward to preview4 builds. Can you pick an older version?

pakrym:
The fix I wanted to take is not in preview3.

PR #7583, opened 2019-02-14
Log before releasing streams and firing tokens

pakrym:
Might fix: https://github.com/aspnet/AspNetCore-Internal/issues/1659

I think writes complete too fast so we don't observe when the message gets logged


src/Servers/IIS/IIS/test/IIS.Tests/ClientDisconnectTests.cs
```diff
@@ -209,9 +209,9 @@ public async Task ReaderThrowsCancelledException()
                 {
                     Assert.IsType<OperationCanceledException>(exception);
                 }
-                catch (Exception e)
+                catch (Exception)
                 {
-                    Logger.LogError(e, "Unexpected exception");
+                    Logger.LogError(exception, "Unexpected exception type");
```
pakrym: I was logging the wrong thing here. `e` is `XUnitAssertException`

jkotalik (approved):

PR #7599, opened 2019-02-15
Remove uses of JSRuntime.Current

pranavkm:

### DIFF
--- a/src/Components/Blazor/Blazor/src/Hosting/WebAssemblyHostBuilder.cs
+++ b/src/Components/Blazor/Blazor/src/Hosting/WebAssemblyHostBuilder.cs
@@ -8,7 +8,6 @@ using Microsoft.AspNetCore.Blazor.Services;
 using Microsoft.AspNetCore.Components.Services;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.JSInterop;
-using Mono.WebAssembly.Interop;
 
 namespace Microsoft.AspNetCore.Blazor.Hosting
 {
@@ -88,10 +87,10 @@ namespace Microsoft.AspNetCore.Blazor.Hosting
             var services = new ServiceCollection();
             services.AddSingleton(_BrowserHostBuilderContext);
             services.AddSingleton<IWebAssemblyHost, WebAssemblyHost>();
-            services.AddSingleton<IJSRuntime, MonoWebAssemblyJSRuntime>();
+            services.AddSingleton<IJSRuntime>(WebAssemblyJSRuntime.Current);
 
             services.AddSingleton<IUriHelper>(WebAssemblyUriHelper.Instance);
-            services.AddSingleton<HttpClient>(s =>
+            services.AddSingleton(s =>
             {
                 // Creating the URI helper needs to wait until the JS Runtime is initialized, so defer it.
                 var uriHelper = s.GetRequiredService<IUriHelper>();
--- a/src/Components/Blazor/Blazor/src/Hosting/WebAssemblyHostBuilder.cs
+++ b/src/Components/Blazor/Blazor/src/Hosting/WebAssemblyHostBuilder.cs
@@ -87,10 +87,10 @@ namespace Microsoft.AspNetCore.Blazor.Hosting
             var services = new ServiceCollection();
             services.AddSingleton(_BrowserHostBuilderContext);
             services.AddSingleton<IWebAssemblyHost, WebAssemblyHost>();
-            services.AddSingleton<IJSRuntime>(WebAssemblyJSRuntime.Current);
+            services.AddSingleton<IJSRuntime>(WebAssemblyJSRuntime.Instance);
 
             services.AddSingleton<IUriHelper>(WebAssemblyUriHelper.Instance);
-            services.AddSingleton(s =>
+            services.AddSingleton<HttpClient>(s =>
             {
                 // Creating the URI helper needs to wait until the JS Runtime is initialized, so defer it.
                 var uriHelper = s.GetRequiredService<IUriHelper>();
--- a/src/Components/Blazor/Blazor/src/Http/WebAssemblyHttpMessageHandler.cs
+++ b/src/Components/Blazor/Blazor/src/Http/WebAssemblyHttpMessageHandler.cs
@@ -1,8 +1,6 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using Microsoft.JSInterop;
-using Mono.WebAssembly.Interop;
 using System;
 using System.Collections.Generic;
 using System.Linq;
@@ -10,6 +8,8 @@ using System.Net;
 using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Blazor.Services;
+using Microsoft.JSInterop;
 
 namespace Microsoft.AspNetCore.Blazor.Http
 {
@@ -25,10 +25,10 @@ namespace Microsoft.AspNetCore.Blazor.Http
         public static FetchCredentialsOption DefaultCredentials { get; set; }
             = FetchCredentialsOption.SameOrigin;
 
-        static object _idLock = new object();
-        static int _nextRequestId = 0;
-        static IDictionary<int, TaskCompletionSource<HttpResponseMessage>> _pendingRequests
+        private static readonly object _idLock = new object();
+        private static readonly IDictionary<int, TaskCompletionSource<HttpResponseMessage>> _pendingRequests
             = new Dictionary<int, TaskCompletionSource<HttpResponseMessage>>();
+        private static int _nextRequestId = 0;
 
         /// <summary>
         /// The name of a well-known property that can be added to <see cref="HttpRequestMessage.Properties"/>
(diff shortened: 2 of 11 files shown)


davidfowl (approved):
Approving based on the title

src/Components/Blazor/Blazor/src/Hosting/WebAssemblyHostBuilder.cs
```diff
@@ -88,10 +87,10 @@ private void CreateServiceProvider()
             var services = new ServiceCollection();
             services.AddSingleton(_BrowserHostBuilderContext);
             services.AddSingleton<IWebAssemblyHost, WebAssemblyHost>();
-            services.AddSingleton<IJSRuntime, MonoWebAssemblyJSRuntime>();
+            services.AddSingleton<IJSRuntime>(WebAssemblyJSRuntime.Current);
 
             services.AddSingleton<IUriHelper>(WebAssemblyUriHelper.Instance);
-            services.AddSingleton<HttpClient>(s =>
```
rynowak: Is there a way to just generally turn off this rule for cleanup? I don't like simplifying things like this (referring to DI) and many places in unit tests where we're testing delegates and type inference.

src/Components/Blazor/Blazor/src/Http/WebAssemblyHttpMessageHandler.cs
```diff
@@ -25,10 +25,10 @@ public class WebAssemblyHttpMessageHandler : HttpMessageHandler
         public static FetchCredentialsOption DefaultCredentials { get; set; }
             = FetchCredentialsOption.SameOrigin;
 
-        static object _idLock = new object();
-        static int _nextRequestId = 0;
-        static IDictionary<int, TaskCompletionSource<HttpResponseMessage>> _pendingRequests
+        private static readonly object _idLock = new object();
+        private static readonly IDictionary<int, TaskCompletionSource<HttpResponseMessage>> _pendingRequests
```
rynowak: Do we need `IDictionary<>` here? This is strictly worse than using `Dictionary` directly (inlining and such).

src/Components/Blazor/Blazor/src/Http/WebAssemblyHttpMessageHandler.cs
```diff
@@ -64,29 +64,22 @@ public class WebAssemblyHttpMessageHandler : HttpMessageHandler
...
-                    "Blazor._internal.http.sendAsync",
-                    id,
-                    request.Content == null ? null : await request.Content.ReadAsByteArrayAsync(),
-                    Json.Serialize(options));
-            }
-            else
-            {
-                throw new NotImplementedException("WebAssemblyHttpMessageHandler only supports running under Mono WebAssembly.");
-            }
+            WebAssemblyJSRuntime.Current.InvokeUnmarshalled<int, byte[], string, object>(
```
rynowak: Should we be passing this in instead of relying on a static? I can't think of a case where this would go wrong for wasm - but in general if we're telling others they can't have a static then we have to figure it out as well.

src/Components/Blazor/Blazor/src/Http/WebAssemblyHttpMessageHandler.cs
```diff
@@ -113,10 +106,18 @@ private string[][] GetHeadersAsStringArray(HttpRequestMessage request)
...
+        /// this method is referenced by client code and must persist.
+        /// </remarks>
+#pragma warning disable IDE0051 // Remove unused private members
+        private static byte[] AllocateArray(string length) => new byte[int.Parse(length)];
+#pragma warning restore IDE0051 // Remove unused private members
+
+        private string[][] GetHeadersAsStringArray(HttpRequestMessage request)
+            => (from header in request.Headers.Concat(request.Content?.Headers ?? Enumerable.Empty<KeyValuePair<string, IEnumerable<string>>>())
+                from headerValue in header.Value // There can be more than one value for each name
+                select new[] { header.Key, headerValue }).ToArray();
```
rynowak: eww linq

src/Components/Blazor/Blazor/src/Rendering/WebAssemblyRenderer.cs
```diff
@@ -66,7 +66,8 @@ public Task AddComponentAsync(Type componentType, string domElementSelector)
             // (otherwise the logged exception will come from UpdateDisplayAsync instead of here)
             // When implementing support for out-of-process runtimes, we'll need to call this
             // asynchronously and ensure we surface any exceptions correctly.
-            ((IJSInProcessRuntime)JSRuntime.Current).Invoke<object>(
+
+            WebAssemblyJSRuntime.Current.Invoke<object>(
```
rynowak: Should we get this from services or constructor?

src/Components/Blazor/Blazor/src/Http/WebAssemblyHttpMessageHandler.cs
SteveSandersonMS: I think this one was my fault. I used to do this everywhere.

src/Components/Blazor/Blazor/src/Http/WebAssemblyHttpMessageHandler.cs
SteveSandersonMS: The static is safe in the WebAssembly case, so there isn't necessarily a problem with us using it as an internal implementation detail for code paths that definitely only occur on WebAssembly.

src/Components/Blazor/Blazor/src/Http/WebAssemblyHttpMessageHandler.cs
SteveSandersonMS: That was me again. Sometimes that syntax just feels right. I mean, usually it doesn't, but *sometimes*...

src/Components/Blazor/Blazor/src/Services/WebAssemblyJSRuntime.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Mono.WebAssembly.Interop;
+
+namespace Microsoft.AspNetCore.Blazor.Services
+{
+    internal static class WebAssemblyJSRuntime
+    {
+        public static readonly MonoWebAssemblyJSRuntime Current = new MonoWebAssemblyJSRuntime();
```
SteveSandersonMS: It's a super-minor point, but `Instance` is a more idiomatic name for singletons (which this now is).

src/Components/Blazor/Blazor/src/Services/WebAssemblyUriHelper.cs
```diff
@@ -52,7 +52,7 @@ protected override void NavigateToCore(string uri, bool forceLoad)
                 throw new ArgumentNullException(nameof(uri));
             }
 
-            ((IJSInProcessRuntime)JSRuntime.Current).Invoke<object>(Interop.NavigateTo, uri, forceLoad);
+            WebAssemblyJSRuntime.Current.Invoke<object>(Interop.NavigateTo, uri, forceLoad);
```
SteveSandersonMS: This is pretty nice. I like how much more explicit it is.

src/Components/Blazor/Templates/src/content/BlazorLibrary-CSharp/ExampleJsInterop.cs
```diff
@@ -5,10 +5,10 @@ namespace BlazorLibrary_CSharp
 {
     public class ExampleJsInterop
     {
-        public static Task<string> Prompt(string message)
+        public static Task<string> Prompt(IJSRuntime jsRuntime, string message)
         {
             // Implemented in exampleJsInterop.js
-            return JSRuntime.Current.InvokeAsync<string>(
+            return jsRuntime.InvokeAsync<string>(
```
SteveSandersonMS: 👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍

SteveSandersonMS (approved):
There's some small CR feedback, but I'll assume you'll choose how to react to it!

rynowak (approved):
:shipit: from me since we're happy leaving the static.

src/Components/Blazor/Blazor/src/Hosting/WebAssemblyHostBuilder.cs
pranavkm: Yeah, I was a bit hesitant about doing this. It doesn't improve readibility and is very easy to break

src/Components/Blazor/Blazor/src/Http/WebAssemblyHttpMessageHandler.cs
pranavkm: This specific case requires having an instance of `MonoWebAssemblyJSRuntime` not one of the abstractions such as `JSInProcessRuntimeBase` \ `IJSInProcessRuntime`. Relying on the service would mean we would have to downcast and throw \ read an optional service which just seems a bit icky. This seems a little cleaner in comparison.

src/Components/Blazor/Blazor/src/Rendering/WebAssemblyRenderer.cs
pranavkm: Spoke about this offline. It's kinda iffy because we really need an instance of `MonoWebAssemblyJSRuntime` to be passed in to the ctor. It doesn't help with unit testing much and we can clean it up in the future if it turns out to be a problem.

Issue #4757, opened 2018-02-21
Create request body pipe reader by decorating the connection pipe reader


halter73:
This should be more efficient than pumping request body data into a new pipe.

halter73:
A reference to the original PR seems to have been lost in the issue transfer.

https://github.com/aspnet/KestrelHttpServer/pull/2345

Issue #4697, opened 2018-11-05
Kestrel gRPC support: provide more efficient APIs for reading request and response bodies


jtattermusch:
One of the possible ways how to integrate with gRPC is to write a generic handler that reads http request's body (`httpContext.Request.Body`) and parses them into grpc request messages  and writes grpc response messages into `httpContext.Response.Body`.  This works pretty well (see prototype in https://github.com/jtattermusch/grpc-aspnetcore-experiments), but both request body and response body are instances of `System.IO.Stream`, which works but isn't particularly efficient (e.g. copying of the data is required, it's hard to use with memory pooling etc.) and will probably become a performance bottleneck in the future.
Is there a way to access request and response bodies e.g. in terms of the new System.IO.Pipelines API?

Context: https://github.com/grpc/grpc/issues/15139

jkotalik:
The base layer for this is done with #7603 and https://github.com/aspnet/AspNetCore/pull/7110. Counting this as done.

Issue #4771, opened 2017-09-07
Put request trailers in a separate collection


cesarbs:
Right now we merge them with request headers. They should go into their own collection. There should also be some way (likely a `Task`) to allow an application to wait until trailers have been received.

Tratcher:
Bumping for triage, this needs to happen in 3.0. At a minimum we need to stop putting 1.x headers into the request header collection, it's a safety concern, even if it means they aren't available.

anurse:
Work:
* Remove the trailers from the main headers collection (breaking change)
* Add `IHttpTrailersFeature` and do the needful.

Tratcher:
Announcement: https://github.com/aspnet/AspNetCore/issues/10519

Tratcher:
Covered in unit tests. Almost no clients support request trailers so I'm skipping manual tests.

PR #7603, opened 2019-02-15
Implement Kestrel Request PipeReader

jkotalik:
For #4757 and #4697.

This PR is the second half of exposing pipes in Kestrel. It does a few main things:
- Makes Content-Length requests directly read from the connection pipe rather than from a request pipe
- Moves RequestPipe logic (PumpAsync) into the Chunked message body.
- Changes Streams => BodyControl and makes all body control calls be done through the HttpRequestPipeReader/HttpResponsePipeWriter

I would consider this PR semi-WIP. I think the overall design is solid, but there is some code duplication I can remove. I also want to add some more tests for Http2 message bodies and tests for ConsumeAsync. Finally, I think I'll need to add some benchmarks for Content-Length request reading and maybe add a few benchmark/inmemory scenarios too. Also, after scanning this PR, there is a lotttt of commented out code 🐋 . 

cc/ @benaadams @shirhatti @JamesNK as I know you guys care 😄


jkotalik:
For some reason I didn't have the option to make a draft PR 😢

src/Servers/Kestrel/Core/src/Internal/Http/ForChunkedEncodingMessageBody.cs
```diff
@@ -0,0 +1,670 @@
...
+using Microsoft.AspNetCore.Connections;
+using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure;
+using Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.Internal;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
+{
+    /// <summary>
+    ///   http://tools.ietf.org/html/rfc2616#section-3.6.1
+    /// </summary>
+    public class ForChunkedEncoding : Http1MessageBody
```
jkotalik: This class has very little changed from the original implementation. It now contains the RequestBodyPipe rather than MessageBody and continues to call PumpAsync().

src/Servers/Kestrel/Core/src/Internal/Http/ForChunkedEncodingMessageBody.cs
```diff
@@ -0,0 +1,670 @@
...
+            TryStart();
+
+            while (true)
+            {
+                _readResult = await StartTimingReadAsync(cancellationToken);
+                var readableBuffer = _readResult.Buffer;
+                var readableBufferLength = readableBuffer.Length;
+                StopTimingRead(readableBufferLength);
+
+                if (_readResult.IsCanceled)
```
jkotalik: I don't believe I need the _userCanceled check here as there is still a RequestPipe. @halter73 you mentioned that the ConnectionPipe can be canceled at practically any point, but I don't believe it applies to the RequestPipe. I think I can remove the bool here and if IsCanceled, break out of the loop. 

Actually, I believe I can remove the while loop entirely if that's the case.

src/Servers/Kestrel/Core/src/Internal/Http/ForChunkedEncodingMessageBody.cs
```diff
@@ -0,0 +1,670 @@
...
+                pool: context.MemoryPool,
+                readerScheduler: context.ServiceContext.Scheduler,
+                writerScheduler: PipeScheduler.Inline,
+                pauseWriterThreshold: 1,
+                resumeWriterThreshold: 1,
+                useSynchronizationContext: false,
+                minimumSegmentSize: KestrelMemoryPool.MinimumSegmentSize
+            ));
+
+        private async Task PumpAsync()
```
jkotalik: From here to https://github.com/aspnet/AspNetCore/pull/7603/files#diff-74fe00cb9040031782f9df9558963000R551 is unchanged from MessageBody before.

src/Servers/Kestrel/Core/src/Internal/Http/ForContentLengthMessageBody.cs
```diff
@@ -0,0 +1,302 @@
...
+            : base(context)
+        {
+            RequestKeepAlive = keepAlive;
+            _contentLength = contentLength;
+            _inputLength = _contentLength;
+        }
+
+        public override async ValueTask<ReadResult> ReadAsync(CancellationToken cancellationToken = default)
+        {
+            if (_completed)
```
jkotalik: This call is spicy. Particularly pay close attention to it.

src/Servers/Kestrel/Core/src/Internal/Http/ForContentLengthMessageBody.cs
```diff
@@ -0,0 +1,302 @@
...
+            {
+                TryStop();
+            }
+
+            return _readResult;
+        }
+
+        public override bool TryRead(out ReadResult readResult)
+        {
+            if (_completed)
```
jkotalik: I'll consolidate a lot of these checks between ReadAsync and TryRead into helpers.

src/Servers/Kestrel/Core/src/Internal/Http/ForUpgradeMessageBody.cs
```diff
@@ -0,0 +1,82 @@
...
+        public override void AdvanceTo(SequencePosition consumed, SequencePosition examined)
+        {
+            _context.Input.AdvanceTo(consumed, examined);
+        }
+
+        public override void Complete(Exception exception)
+        {
+            // Don't call Connection.Complete.
+            _context.ReportApplicationError(exception);
+            _completed = true;
```
jkotalik: I don't believe we should still complete the Connection pipe if the user calls Complete (just like we did with the request side).

src/Servers/Kestrel/Core/src/Internal/Http/Http1MessageBody.cs
```diff
@@ -1,223 +1,18 @@
...
-
 namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 {
     public abstract class Http1MessageBody : MessageBody
     {
-        private readonly Http1Connection _context;
-
-        private volatile bool _canceled;
-        private Task _pumpTask;
+        protected readonly Http1Connection _context;
```
jkotalik: This now only contains the "For" method as the other message bodies are in different classes. I may restore them here if the diff is too nasty.

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.FeatureCollection.cs
```diff
@@ -94,8 +95,39 @@ IHeaderDictionary IHttpRequestFeature.Headers
 
         Stream IHttpRequestFeature.Body
         {
-            get => RequestBody;
-            set => RequestBody = value;
+            get
```
jkotalik: Same logic as the ResponseBody/PipeWriter

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
```diff
@@ -652,9 +650,6 @@ private async Task ProcessRequests<TContext>(IHttpApplication<TContext> applicat
 
                 if (HasStartedConsumingRequestBody)
                 {
-                    RequestBodyPipe.Reader.Complete();
-
-                    // Wait for Http1MessageBody.PumpAsync() to call RequestBodyPipe.Writer.Complete().
                     await messageBody.StopAsync();
```
jkotalik: StopAsync now calls PipeReader.Complete() as they need to do different things whether there is a connection pipe or request pipe.

src/Servers/Kestrel/Core/src/Internal/Http/HttpRequestStream.cs
```diff
@@ -3,217 +3,65 @@
...
+        private ValueTask<int> ReadAsyncInternal(Memory<byte> buffer, CancellationToken cancellationToken)
         {
             try
             {
-                await _body.CopyToAsync(destination, cancellationToken);
+                return base.ReadAsync(buffer, cancellationToken);
             }
             catch (ConnectionAbortedException ex)
             {
                 throw new TaskCanceledException("The request was aborted", ex);
```
jkotalik: I had to change a few tests because this exception was not being caught. Definitely need to fix. I'm fairly sure it's due to new async state machines being introduced.

jkotalik (commented):
Some commentary to guide feedback.

src/Http/Http/src/ReadOnlyPipeStream.cs
```diff
@@ -165,6 +165,7 @@ private async ValueTask<int> ReadAsyncInternal(Memory<byte> buffer, Cancellation
                 var readableBuffer = result.Buffer;
                 var readableBufferLength = readableBuffer.Length;
 
+                // TODO make this throw if the result is canceled
```
davidfowl: Bug filed for this?

src/Servers/Kestrel/Core/src/Internal/Http/ForContentLengthMessageBody.cs
```diff
@@ -0,0 +1,302 @@
...
+                {
+                    AdvanceTo(readResult.Buffer.End);
+
+                    if (readResult.IsCompleted)
+                    {
+                        return Task.CompletedTask;
+                    }
+                }
+            }
+            catch (OperationCanceledException)
```
davidfowl: This can be removed now.

src/Http/Http/src/ReadOnlyPipeStream.cs
jkotalik: https://github.com/aspnet/AspNetCore/pull/7618

src/Servers/Kestrel/Core/src/Internal/Http/ForContentLengthMessageBody.cs
```diff
@@ -0,0 +1,306 @@
...
+                {
+                    break;
+                }
+            }
+
+            // handle cases where we send more data than the content length
+            if (_readResult.Buffer.Length > _inputLength)
+            {
+                _readResult = new ReadResult(_readResult.Buffer.Slice(0, _inputLength), _readResult.IsCanceled, isCompleted: true);
+
```
davidfowl: nit: Space

src/Servers/Kestrel/Core/src/Internal/Http/ForContentLengthMessageBody.cs
```diff
@@ -0,0 +1,306 @@
...
+            }
+            else if (_readResult.Buffer.Length == _inputLength)
```
davidfowl: Is this an optimization? I figure you would handle this in AdvanceTo (since _inputLength  is decremented there and there's an earlier check for _inputLength  == 0 in there)

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
```diff
@@ -652,9 +650,6 @@ private async Task ProcessRequests<TContext>(IHttpApplication<TContext> applicat
 
                 if (HasStartedConsumingRequestBody)
                 {
-                    RequestBodyPipe.Reader.Complete();
```
davidfowl: 👏

src/Servers/Kestrel/Core/src/Internal/Http2/Http2MessageBody.cs
```diff
@@ -1,15 +1,18 @@
...
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http;
-using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
 {
     public class Http2MessageBody : MessageBody
     {
         private readonly Http2Stream _context;
+        private ReadResult _previousReadResult;
```
davidfowl: nit: Rename to _readResult

src/Servers/Kestrel/Core/src/Internal/Http/ForContentLengthMessageBody.cs
jkotalik: Yeah an optimization. It makes it so people don't need to call ReadAsync/TryRead to get a second read result that would just be IsCompleted=true.

src/Servers/Kestrel/Core/src/Internal/Http/Http1ChunkedEncodingMessageBody.cs
```diff
@@ -87,6 +87,7 @@ public override async ValueTask<ReadResult> ReadAsync(CancellationToken cancella
         public override void Complete(Exception exception)
         {
             _requestBodyPipe.Reader.Complete(exception);
+            _context.ReportApplicationError(exception);
```
halter73: This will produce an application error log even if `exception` is null.

jkotalik: I changed ReportApplicationError to return if exception is null.

src/Servers/Kestrel/Core/src/Internal/Http/HttpRequestStream.cs
halter73: It's the await removal that does this. Both ConnectionAbortedException and TaskCanceledExceptions are OperationCanceledExceptions, so if we have to break, there's a workaround, but let's not break.

src/Servers/Kestrel/Core/src/Internal/Http/Http1ContentLengthMessageBody.cs
```diff
@@ -0,0 +1,295 @@
...
+                    // response is written after observing the unexpected end of request content instead of just
+                    // closing the connection without a response as expected.
+                    _context.OnInputOrOutputCompleted();
+
+                    BadHttpRequestException.Throw(RequestRejectionReason.UnexpectedEndOfRequestContent);
+                }
+
+                if (readableBufferLength > 0)
+                {
+                    break;
```
halter73: I would always reloop if there is a non-user cancellation.

src/Servers/Kestrel/Core/src/Internal/Http/Http1ContentLengthMessageBody.cs
```diff
@@ -0,0 +1,295 @@
...
+
+                if (_context.RequestTimedOut)
+                {
+                    Debug.Assert(_readResult.IsCanceled);
+                    BadHttpRequestException.Throw(RequestRejectionReason.RequestBodyTimeout);
+                }
+
+                if (_readResult.IsCanceled)
+                {
+                    if (Interlocked.CompareExchange(ref _userCanceled, 0, 1) == 1)
```
halter73: Nit:

```suggestion
                    if (Interlocked.Exchange(ref _userCanceled, 0) == 1)
```

src/Servers/Kestrel/Core/src/Internal/Http/MessageBody.cs
```diff
@@ -231,49 +154,5 @@ protected void AddAndCheckConsumedBytes(long consumedBytes)
                 BadHttpRequestException.Throw(RequestRejectionReason.RequestBodyTooLarge);
             }
         }
-
-        private ValueTask<ReadResult> StartTimingReadAsync(CancellationToken cancellationToken)
-        {
-            var readAwaitable = _context.RequestBodyPipe.Reader.ReadAsync(cancellationToken);
```
halter73: Start/StopTimingReadAsync is now basically duplicated 3 times. I would try to consolidate this. I know the line above to get the `ValueTask<ReadResult>` will be different in each implementation, but that can be fixed by calling an abstract method.

src/Servers/Kestrel/Core/src/Internal/Http/Http1ContentLengthMessageBody.cs
```diff
@@ -0,0 +1,295 @@
...
+            TryStart();
+
+            // The while(true) loop is required because the Http1 connection calls CancelPendingRead to unblock
+            // the call to StartTimingReadAsync to check if the request timed out.
+            // However, if the user called CancelPendingRead, we want that to return a canceled ReadResult
+            // We internally track an int for that.
+            while (true)
+            {
+                // This isn't great. The issue is that TryRead can get a canceled read result
+                // which is unknown to StartTimingReadAsync.
```
halter73: I don't think I understand this. Shouldn't TryRead check IsCanceled and throw a BadHttpRequestException too?

src/Servers/Kestrel/Core/src/Internal/Http/Http1ContentLengthMessageBody.cs
```diff
@@ -0,0 +1,295 @@
...
+            {
+                readResult = new ReadResult(default, isCanceled: false, isCompleted: true);
+                return true;
+            }
+
+            TryStart();
+
+            var boolResult = _context.Input.TryRead(out _readResult);
+
+            CreateReadResultFromConnectionReadResult();
```
halter73: Should this run if TryRead returns false?

src/Servers/Kestrel/Core/src/Internal/Http/Http1ContentLengthMessageBody.cs
```diff
@@ -0,0 +1,295 @@
...
+                {
+                    // OnInputOrOutputCompleted() is an idempotent method that closes the connection. Sometimes
+                    // input completion is observed here before the Input.OnWriterCompleted() callback is fired,
+                    // so we call OnInputOrOutputCompleted() now to prevent a race in our tests where a 400
+                    // response is written after observing the unexpected end of request content instead of just
+                    // closing the connection without a response as expected.
+                    _context.OnInputOrOutputCompleted();
+
+                    BadHttpRequestException.Throw(RequestRejectionReason.UnexpectedEndOfRequestContent);
+                }
```
halter73: It would be nice if this along with the other shared ReadResult validation were moved into shared methods instead of duplicated.

src/Servers/Kestrel/Core/src/Internal/Http/Http1ContentLengthMessageBody.cs
```diff
@@ -0,0 +1,295 @@
...
+
+        public override void Complete(Exception exception)
+        {
+            _context.ReportApplicationError(exception);
+            _completed = true;
+        }
+
+        public override void OnWriterCompleted(Action<Exception, object> callback, object state)
+        {
+            // TODO make this work with ContentLength.
```
halter73: File an issue for this.

src/Servers/Kestrel/Core/src/Internal/Http/Http1ContentLengthMessageBody.cs
```diff
@@ -0,0 +1,295 @@
...
+            _context.Input.CancelPendingRead();
+        }
+
+        protected override Task OnStopAsync()
+        {
+            Complete(null);
+            return Task.CompletedTask;
+        }
+
+        protected override Task OnConsumeAsync()
```
halter73: There's a lot of duplication in the OnConsumeAsync paths. Try to consolidate as much as possible. Virtual dispatch won't hurt perf too bad, and it could save us from making only a partial fix in the future.

src/Servers/Kestrel/Core/src/Internal/Http/HttpRequestStream.cs
```diff
@@ -3,217 +3,65 @@
...
-            {
-                return await _body.ReadAsync(buffer, cancellationToken);
-            }
-            catch (ConnectionAbortedException ex)
-            {
-                throw new TaskCanceledException("The request was aborted", ex);
-            }
-        }
-
-        public override Task CopyToAsync(Stream destination, int bufferSize, CancellationToken cancellationToken)
```
halter73: We should continue to override CopyToAsync in order to avoid the large intermediate buffer allocations of the default implementation.

halter73: Never mind. I see it in the ReadOnlyPipeStream base class now.

src/Servers/Kestrel/Core/src/Internal/Http2/Http2Stream.cs
```diff
@@ -55,7 +57,7 @@ public Http2Stream(Http2StreamContext context)
                 this,
                 context.ServiceContext.Log);
 
-            RequestBodyPipe = CreateRequestBodyPipe(context.ServerPeerSettings.InitialWindowSize);
+            RequestPipe = CreateRequestBodyPipe(context.ServerPeerSettings.InitialWindowSize);
```
halter73: Did you really need to change this name?

halter73 (approved):
On the right track. I want so see better code sharing before I approve this.

src/Servers/Kestrel/Core/src/Internal/Http/Http1ContentLengthMessageBody.cs
davidfowl: Try* API shouldn’t throw. Makes them unusable, so we should avoid it where possible

jkotalik: Yeah I was trying to avoid throwing in Try methods. However, TryRead throws internally in pipes, so we should consider fixing that in pipes.

jkotalik:
Updated. I think I saw one test fail with timing (think there is a different exception message), I’ll take a look at it later today/tomorrow.

benaadams:
```
#### Conflicting files
`src/Servers/Kestrel/test/InMemory.FunctionalTests/RequestTests.cs`
```

src/Servers/Kestrel/Core/src/Internal/Http/Http1ContentLengthMessageBody.cs
halter73: > Try* API shouldn’t throw. Makes them unusable, so we should avoid it where possible

Like @jkotalik points out, TryRead is a little special. Errors also need to be raised from TryRead when the writer completes with an error. TryRead should only return false when the PipeReader is still active but there's no data available to currently read.

JamesNK: A little more info about the TryXXX pattern: TryXXX doesn't mean the method should never throw an exception. It means it shouldn't throw an exception for a certain use-case.

For example, int.TryParse will return false if the string cannot be parsed to an integer. But it will throw an ArgumentNullException if you give it a null value.

davidfowl: > TryRead should only return false when the PipeReader is still active but there's no data available to currently read.

Sure, it has to throw *sometimes* in those rare cases where the input is invalid. I don't know that I agree with the above.

src/Servers/Kestrel/Core/src/Internal/Http/Http1ContentLengthMessageBody.cs
jkotalik: I will once I merge this PR.

src/Servers/Kestrel/Core/src/Internal/Http/Http1MessageBody.cs
halter73: You might as well delete the class then. You can leave a static class for the For method, but otherwise all it's doing is setting the _context field in the ctor.

halter73: Never mind. I see you put more shared logic here. 👍 Carry on.

src/Servers/Kestrel/Core/src/Internal/Http/Http1ChunkedEncodingMessageBody.cs
```diff
@@ -0,0 +1,545 @@
...
+            {
+                TryStop();
+            }
+
+            return _readResult;
+        }
+
+        public override void Complete(Exception exception)
+        {
+            _requestBodyPipe.Reader.Complete(exception);
```
Tratcher: This exception will get re-thrown for the writer, no? Why would we want that?

src/Servers/Kestrel/Core/src/Internal/Http/Http1ChunkedEncodingMessageBody.cs
```diff
@@ -0,0 +1,545 @@
...
+                    return false;
+                }
+
+                readableBuffer = readableBuffer.Slice(consumed);
+            }
+
+            // _consumedBytes aren't tracked for trailer headers, since headers have separate limits.
+            if (_mode == Mode.TrailerHeaders)
+            {
+                if (_context.TakeMessageHeaders(readableBuffer, out consumed, out examined))
```
Tratcher: We really need to stop putting trailers into the normal header collection.

halter73: #4771 is currently in the backlog

src/Servers/Kestrel/Core/src/Internal/Http/Http1ChunkedEncodingMessageBody.cs
halter73: Good call. I don't think we complete a PipeReader with an exception anywhere else in Kestrel. As is, PumpAsync would wind up completing the writer side with the same exception that's passed in here, but that shouldn't be important since no one is reading at that point.

davidfowl (approved):
This PR looks really good. Make sure you get it into the preview3 branch

src/Servers/Kestrel/Core/src/Internal/Http/Http1ChunkedEncodingMessageBody.cs
jkotalik: Sounds good, I'll remove the Complete here and just report the application exception.

jkotalik:
@Eilon we would like this in for preview3. This completes the work required to expose the HttpRequest body as a PipeReader. It's something that is a major part of the preview3 milestone.

JamesNK:
Introduction of the pipe writer had some breaking changes for grpc-dotnet. I worry about putting the reader in at the last minute. And we've just published the package we'll ship with the preview 3 templates - https://github.com/grpc/grpc-dotnet/releases/tag/v0.1.19-pre1

At the very least grpc-dotnet will need to be retested after this change. Rerun the unit and functional tests, and smoke testing that each of the samples work.

davidfowl:
@JamesNK the reader needs to go in and even though this change is pretty big, we shouldn't hold it hostage as nothing is using the pipe reader as yet and the entire repository has passed, bugs in the PipeReader aren't a big deal the problem is more upstack components making use of existing Stream APIs (that needs to continue working). Unlike the GRPC scenario, using the new APIs and finding bugs are the reasons previews exist. 

That said, I think this needs to go into preview3 so we can flesh out more of the use cases on top.

Eilon:
So this change should have no negative impact in Preview 3 for the gRPC work or other things?

jkotalik:
@JamesNK I'll run this change against gRPC right now.

jkotalik:
> So this change should have no negative impact in Preview 3 for the gRPC work or other things?

Yes, but I will verify now by running it against gRPC.

jkotalik:
Rebased on the wrong branch, will fix shortly.

JamesNK:
> Yes, but I will verify now by running it against gRPC.

It's important to test the 3 clients under examples/clients run without issue. The functional tests use TestServer which I'm guessing will still use the stream adapters.

davidfowl:
@Eilon can we make this PR contingent on no gRPC regressions? I would be good to merge this today.

Eilon:
@davidfowl contingent not just on some one-off gRPC testing, but on us asserting that the risk is low for all cases (preview 3 works just fine now, I don't want to break anything because we rushed).

JamesNK (approved):

jkotalik:
@Eilon tested against all HTTP2 clients in gRPC. This is a low risk change for gRPC because it barely changed anything with HTTP2 . See: https://github.com/aspnet/AspNetCore/pull/7603/files#diff-82d8f6d18b1c89f77b834073ac706d54R4 for the changes to HTTP2 . This effectively just exposes the underlying HTTP2 stream pipe.

src/Servers/Kestrel/Core/src/Internal/Http/Http1ChunkedEncodingMessageBody.cs
jkotalik: Misread, I should keep the call to complete, but remove the passed in exception.

Eilon:
Approved for preview 3.

jkotalik:
All required checks passed.

PR #7610, opened 2019-02-15
Removed comments

### LINKED ISSUE #7579: Remove comment about potential routing changes from the App.cshtml file
The *App.cshtml* file has the following comment about potential updates to routing:

```
<!--
    Configuring this here is temporary. Later we'll move the app config
    into Program.cs, and it won't be necessary to specify AppAssembly.
-->
<Router AppAssembly="typeof(Program).Assembly" />
```

The language here was speculative and the change here is still to be worked out. Let's just remove this comment and track whatever routing enhancements we want to do with #5489.

Potapy4:
Fixes #7579


SteveSandersonMS (approved):
Thanks @Potapy4!

Issue #5471, opened 2018-02-07
Runtime Culture and Localization


grahamehorner:
Please consider browser runtime cultural and allow switching of culture with in the Blazor environment

SteveSandersonMS:
When doing this, be sure to check cases around Data Annotations validation and the built-in parsing validators in the `Input*` components to be sure they localize conveniently.

DamianEdwards:
Is this issue intended to cover just the WASM side of the world, or localization in the Razor Components server-side features in 3.0 too?

SteveSandersonMS:
@DamianEdwards Both. Currently this issue is quite high-level and all-encompassing, but will get broken down into several sub-issues as we start work on it.

grahamehorner:
IMHO and just for my two pennies worth @SteveSandersonMS ;) as a developer and a multi-lingual application Razor Component circuits would need some mechanism to send the users browser culture to the server; to allow the rendering component(s) to identifier the correct resources to use and/or having routing determine any culture specifics; so for a starter for 10 a smaller task of the bigger story IMHO would be to possibly update the component circuit implementation to provide browser agent string and culture with in the messages sent to the signalr hub.

NOTE: do you think a user should be able to switch culture of the browser without causing issues/re-login etc. ? or should a user session be forced into a fixed culture ?

Franklin89:
I have experienced a strange behavior (in my opinion strange) as I wanted to try to implement some sort of localization (client side blazor). If I set the `CultureInfo` in the startup of the client app:

```
public void Configure(IComponentsApplicationBuilder app)
{
    System.Threading.Thread.CurrentThread.CurrentCulture = new CultureInfo("de-CH");
    System.Threading.Thread.CurrentThread.CurrentUICulture = new CultureInfo("de-CH");

    app.AddComponent<App>("app");
}
```

and then render a input with a `DateTime` field the input gets rendered in the correct format according to the culture. If I then change the date inside the input and the state gets updated the date gets displayed in the default format. I then tried the same with a string property.

```
public string Birthdate
{
    get
    {
        Console.WriteLine(System.Threading.Thread.CurrentThread.CurrentCulture);
        return Birthdate.ToShortDateString();
    }
    set
    {
        Birthdate = DateTime.Parse(value);
    }
}
```

The culture that gets written to the console first shows the correct culture. After changing the value the culture is empty. So it seems to me that the culture gets cleared somehow.

RemiBou:
[...]

PR #7614, opened 2019-02-15
Components: Forms and validation

SteveSandersonMS:
OK, this is a pretty massive PR, so @rynowak / @javiercn / @pranavkm - would any of you like me to give you a 30-minute tour of the changes and new features? Or if you prefer just to dive into the code that's completely fine with me too.

**Features**

 * `EditContext` - this tracks metadata about an edit process (e.g., what's been modified, current validation messages, etc.). It doesn't do any actual validation itself, but rather has an API designed to let external systems hook into it. External things can get notification about when fields are edited so they can choose to validate if they want.
 * `EditForm` - this creates an `EditContext` for you and cascades it down to contents. It also provides extra events (`OnValidSubmit`, `OnInvalidSubmit`, though you can still use `OnSubmit` directly if you want to trigger the validation yourself based on your own lifecycle design).
 * `DataAnnotationsValidator` - this attaches data annotations support to an `EditContext`
 * `ValidationSummary` - like the one in MVC
 * Built-in input components (`InputText`, `InputCheckbox`, `InputSelect`, etc.). These provide default behavior of validating on edit and changing their CSS class to reflect field state. Some of them have useful parsing logic (e.g., `InputDate` and `InputNumber` handle unparseable values gracefully by registering them as validation errors). The relevant ones also support nullability of the target field (e.g., `int?`).

For usage examples, see the new test cases in the E2E BasicTestApp:

 * SimpleValidationComponent
 * TypicalValidationComponent
 * NotifyPropertyChangedValidationComponent

**Still to do for preview 3**

 - ~Make the build pass. Need to check with @natemcmaster what's needed to deal with the platform manifest error.~ Should be OK now.
 - ~Rebase on `master`. This is tricky because in the last few days, the repo stopped taking a dependency on DataAnnotations, so I need to put that back. And first I need approval to join the darc group, etc.~ Done
 - ~E2E tests. The test cases themselves are implemented - I just haven't written the Selenium automation for them yet.~ Done
 - ~Maybe implement a `ValidationMessage` component, which would be trivial. Or developers can implement it themseles. By design, these sorts of things don't require priviledged access to any internal APIs.~ Implemented.
 - ~Remove the manual usages of `*Expression` from the test cases once this enhancement gets merged into Razor: https://github.com/aspnet/AspNetCore-Tooling/pull/213~ Not waiting for this. 

**Further enhancements planned for after preview 3**

 - Async validation. I have a design in mind, but there's only so much can get done in time for Preview 3, and it would be good to get some customer feedback before trying to do everything.
 - Maybe some way to configure a default validation system, so you don't have to put `<DataAnnotationsValidator>` in every `EditForm` if that's what you're using.
 - Maybe some way to override the default CSS classes on the built-in Input* components
 - Ideally (though nontrivial), support for splatting attributes on Razor Components. Then we could make the built-in input controls automatically pass through all extra attributes, and developers who subclass them can still get that attribute passthrough behavior. Until we do something like this, some basic use cases like setting a `maxlength` on `InputText` just aren't supported (or at least, you have to subclass `InputText` and put that feature on your subclass, which is inconvenient).


SteveSandersonMS:
@natemcmaster Is it OK for me to put back the dependency on `System.ComponentModel.Annotations` that was removed from this repo in the last few days? We are still depending on it transitively via `Microsoft.Extensions.Options`, so it seems like there's no reason why it shouldn't be referenced from shared framework assemblies. If I understand correctly, I need to use `darc` to do this, somehow.

natemcmaster:
@SteveSandersonMS See https://github.com/aspnet/AspNetCore/blob/master/docs/ReferenceResolution.md#example-adding-a-new-dependency.

I removed it because projects don't need this `<Reference>` if they target `netcoreapp3.0`. But it looks like this PR adds the reference in a netstandard2.0 project, so yes, a direct dependency on `System.ComponentModel.Annotations` is okay to put back.

src/Components/Components/src/Forms/InputComponents/InputBase.cs
```diff
@@ -0,0 +1,189 @@
...
+        /// <returns>True if the value could be parsed; otherwise false.</returns>
+        protected virtual bool TryParseValueFromString(string value, out T result, out string validationErrorMessage)
+            => throw new NotImplementedException($"Components that inherit from {nameof(InputBase<T>)} must override " +
+                $"{nameof(TryParseValueFromString)} in order to use {nameof(CurrentValueAsString)}.");
+
+        /// <summary>
+        /// Gets a string that indicates the status of the field being edited. This will include
+        /// some combination of "modified", "valid", or "invalid", depending on the status of the field.
+        /// </summary>
+        protected string FieldClass
```
Andrzej-W: Is this a good name for the property? It doesn't correspond to the comment above.

src/Components/Components/src/Forms/InputComponents/InputDate.cs
```diff
@@ -0,0 +1,108 @@
...
+{
+    /// <summary>
+    /// An input component for editing date values.
+    /// Supported types are <see cref="DateTime"/> and <see cref="DateTimeOffset"/>.
+    /// </summary>
+    public class InputDate<T> : InputBase<T>
+    {
+        const string dateFormat = "yyyy-MM-dd"; // Compatible with HTML date inputs
+
+        [Parameter] string ParsingErrorMessage { get; set; } = "The {0} field must be a date.";
```
Andrzej-W: I would like to have a chance to localize messages like this and don't pass them as parameters each time.

src/Components/Components/src/Forms/InputComponents/InputNumber.cs
```diff
@@ -0,0 +1,173 @@
...
+            else
+            {
+                result = default;
+                return false;
+            }
+        }
+
+        static bool TryParseFloat(string value, out T result)
+        {
+            var success = float.TryParse(value, out var parsedValue);
```
Andrzej-W: Probably you should parse float/double/decimal values with InvariantCulture. I have Polish Windows and Polish culture set in browsers. In browser I see comma as decimal separator but field values in Blazor 0.7 have dot. Example code:
<input type="number" bind="@number" /> @number
@functions
{ string number = "1.9"; }
Notice that input type is set to "number" and contrary to InputDate you don't add this attribute in InputNumber.

src/Components/Components/src/Forms/InputComponents/InputNumber.cs
```diff
@@ -0,0 +1,173 @@
...
+        }
+
+        [Parameter] string ParsingErrorMessage { get; set; } = "The {0} field must be a number.";
+
+        /// <inheritdoc />
+        protected override void BuildRenderTree(RenderTreeBuilder builder)
+        {
+            base.BuildRenderTree(builder);
+            builder.OpenElement(0, "input");
+            builder.AddAttribute(1, "class", FieldClass);
```
Andrzej-W: Why there is no
```c#
builder.AddAttribute(1, "type", "number");`
```

src/Components/Components/src/Forms/EditContext.cs
```diff
@@ -0,0 +1,185 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.AspNetCore.Components.Forms
+{
+    /* Async validation plan
```
rynowak: ewww block comment. That's like the `async void` of comments.

src/Components/Components/src/Forms/EditContext.cs
```diff
@@ -0,0 +1,185 @@
...
+     *   can cancel them if it wants, and can still issue ValidationResultsChanged notifications when
+     *   each task completes. So a UI can determine whether to show "pending" state on a per-field
+     *   and per-form basis, and will re-render as each field's results arrive.
+     * - Note: it's unclear why we'd need WhenAllValidationTasks(FieldIdentifier) (i.e., per-field),
+     *   since you wouldn't "await" this to get per-field updates (rather, you'd use ValidationResultsChanged).
+     *   Maybe WhenAllValidationTasks can be private, and only called by ValidateAsync. We just expose
+     *   public HasPendingValidationTasks (per-field and per-edit-context).
+     * Will implement this shortly after getting more of the system in place, assuming it still
+     * appears to be the correct design.
+     */
```
rynowak: Should this go into an issue instead of a comment?

rynowak: FYI I'm ignoring the contents of this for now.

src/Components/Components/src/Forms/EditContext.cs
```diff
@@ -0,0 +1,185 @@
...
+
+    /// <summary>
+    /// Holds metadata related to a data editing process, such as flags to indicate which
+    /// fields have been modified and the current set of validation messages.
+    /// </summary>
+    public class EditContext
```
rynowak: sealed ?

src/Components/Components/src/Forms/EditContext.cs
```diff
@@ -0,0 +1,185 @@
...
+        /// <summary>
+        /// Constructs an instance of <see cref="EditContext"/>.
+        /// </summary>
+        /// <param name="model">The model object for the <see cref="EditContext"/>. This object should hold the data being edited, for example as a set of properties.</param>
+        public EditContext(object model)
+        {
+            // The only reason we disallow null is because you'd almost always want one, and if you
+            // really don't, you can pass an empty object then ignore it. Ensuring it's nonnull
+            // simplifies things for all consumers of EditContext.
+            Model = model ?? throw new ArgumentNullException(nameof(model));
```
rynowak: I think it's also a really good starting point to say that our infrastructure will not attempt to make things *automatically null-safe*. MVC does this and it causes much wailing and gnashing of teeth.

src/Components/Components/src/Forms/EditContext.cs
```diff
@@ -0,0 +1,185 @@
...
+
+        /// <summary>
+        /// An event that is raised when validation is requested.
+        /// </summary>
+        public event EventHandler OnValidationRequested;
+
+        /// <summary>
+        /// An event that is raised when validation state has changed.
+        /// </summary>
+        public event EventHandler OnValidationStateChanged;
```
rynowak: Consider whether these things should have explicit event-args types. Defining an event-args type name is free if its empty, and lets us add more data in the future if we need it. 

I haven't totally thought through whether this is important but it comes to mind every time I see an event used.

src/Components/Components/src/Forms/EditContext.cs
```diff
@@ -0,0 +1,185 @@
...
+            GetFieldState(fieldIdentifier, ensureExists: true).IsModified = true;
+            OnFieldChanged?.Invoke(this, fieldIdentifier);
+        }
+
+        /// <summary>
+        /// Signals that some aspect of validation state has changed.
+        /// </summary>
+        public void NotifyValidationStateChanged()
+        {
+            OnValidationStateChanged?.Invoke(this, null);
```
rynowak: `EventArgs.Empty` instead of null

src/Components/Components/src/Forms/EditContext.cs
```diff
@@ -0,0 +1,185 @@
...
+        }
+
+        /// <summary>
+        /// Clears any modification flag that may be tracked for the specified field.
+        /// </summary>
+        /// <param name="fieldIdentifier">Identifies the field whose modification flag (if any) should be cleared.</param>
+        public void MarkAsUnmodified(FieldIdentifier fieldIdentifier)
+        {
+            if (_fieldStates.TryGetValue(fieldIdentifier, out var state))
```
rynowak: Is there a reason why we have an `if` here and `NotifyFieldChanged` doesn't? What should be the behaviour of these methods when a field can't be found?

src/Components/Components/src/Forms/EditContext.cs
```diff
@@ -0,0 +1,185 @@
...
+        /// <summary>
+        /// Gets the current validation messages across all fields.
+        ///
+        /// This method does not perform validation itself. It only returns messages determined by previous validation actions.
+        /// </summary>
+        /// <returns>The current validation messages.</returns>
+        public IEnumerable<string> GetValidationMessages()
+            // Since we're only enumerating the fields for which we have a non-null state, the cost of this grows
+            // based on how many fields have been modified or have associated validation messages
+            => _fieldStates.Values.SelectMany(state => state.GetValidationMessages());
```
rynowak: Note: all of these cases of linq allocate - a `foreach` doesn't (for `Dictionary<>.ValueCollection`)

rynowak: I'm not laser-focused on the perf of stuff yet, just trying to build a mental model about what tradeoffs are possible.

src/Components/Components/src/Forms/EditContextExpressionExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq.Expressions;
+
+namespace Microsoft.AspNetCore.Components.Forms
+{
+    /// <summary>
+    /// Provides extension methods to simplify using <see cref="EditContext"/> with expressions.
+    /// </summary>
+    public static class EditContextExpressionExtensions
```
rynowak: Is there a really good reason to make these extension methods? Would anything go wrong if they were just defined on the `EditContext`?

src/Components/Components/src/Forms/EditContextDataAnnotationsExtensions.cs
```diff
@@ -0,0 +1,101 @@
...
+
+            // Perform object-level validation on request
+            editContext.OnValidationRequested +=
+                (sender, eventArgs) => ValidateModel((EditContext)sender, messages);
+
+            // Perform per-field validation on each field edit
+            editContext.OnFieldChanged +=
+                (sender, fieldIdentifier) => ValidateField(editContext, messages, fieldIdentifier);
+
+            return editContext;
```
rynowak: This is a very non-C# way of wiring something up, and it's also really abstraction-light which is probably fine.

The only concern that I have about this is that if these events have mixed use (used by validation providers, used by components/apps) that becomes really hard to reason about due to ordering. I don't have any really big concerns about this at this point. I think all of the other approaches to how something like data annotations gets wired up have their own drawbacks as well.

src/Components/Components/src/Forms/EditContextExpressionExtensions.cs
rynowak: It seems like the idea to is just to keep these separated based on how critical they are to the core system's layering.

src/Components/Components/src/Forms/EditContextFieldClassExtensions.cs
```diff
@@ -0,0 +1,46 @@
...
+        public static string FieldClass(this EditContext editContext, FieldIdentifier fieldIdentifier)
+        {
+            var isValid = !editContext.GetValidationMessages(fieldIdentifier).Any();
+            if (editContext.IsModified(fieldIdentifier))
+            {
+                return isValid ? "modified valid" : "modified invalid";
+            }
+            else
+            {
+                return isValid ? "valid" : "invalid";
```
rynowak: If we want this to be extensible these could be properties on the `EditContext`.

src/Components/Components/src/Forms/EditForm.cs
```diff
@@ -0,0 +1,150 @@
...
+using Microsoft.AspNetCore.Components.RenderTree;
+using System;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Components.Forms
+{
+    /// <summary>
+    /// Renders a form element that cascades an <see cref="EditContext"/> to descendants.
+    /// </summary>
+    public class EditForm : ComponentBase
```
rynowak: Is this useful/meaningful to subclass?

src/Components/Components/src/Forms/EditForm.cs
```diff
@@ -0,0 +1,150 @@
...
+        /// A callback that will be invoked when the form is submitted and the
+        /// <see cref="EditContext"/> is determined to be valid.
+        /// </summary>
+        [Parameter] Func<EditContext, Task> OnValidSubmit { get; set; }
+
+        /// <summary>
+        /// A callback that will be invoked when the form is submitted and the
+        /// <see cref="EditContext"/> is determined to be invalid.
+        /// </summary>
+        [Parameter] Func<EditContext, Task> OnInvalidSubmit { get; set; }
```
rynowak: I'm assuming that these become `EventCallback<EditContext>` 😀

src/Components/Components/src/Forms/EditForm.cs
```diff
@@ -0,0 +1,150 @@
...
+                throw new InvalidOperationException($"When supplying an {nameof(OnSubmit)} parameter to " +
+                    $"{nameof(EditForm)}, do not also supply {nameof(OnValidSubmit)} or {nameof(OnInvalidSubmit)}.");
+            }
+
+            // Update _fixedEditContext if we don't have one yet, or if they are supplying a
+            // potentially new EditContext, or if they are supplying a different Model
+            if (_fixedEditContext == null || EditContext != null || Model != _fixedEditContext.Model)
+            {
+                _fixedEditContext = EditContext ?? new EditContext(Model);
+            }
```
rynowak: I'm not super clear on what this is for - the comment says what it does, but not why it's important.

rynowak: Looks like it's explained below.

src/Components/Components/src/Forms/FieldIdentifier.cs
```diff
@@ -0,0 +1,111 @@
...
+    /// </summary>
+    public readonly struct FieldIdentifier
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="FieldIdentifier"/> structure.
+        /// </summary>
+        /// <param name="accessor">An expression that identifies an object member.</param>
+        public static FieldIdentifier Create<T>(Expression<Func<T>> accessor)
+        {
+            ParseAccessor(accessor, out var model, out var fieldName);
```
rynowak: should probably null check the accessor

src/Components/Components/src/Forms/FieldIdentifier.cs
```diff
@@ -0,0 +1,111 @@
...
+using System;
+using System.Linq.Expressions;
+
+namespace Microsoft.AspNetCore.Components.Forms
+{
+    /// <summary>
+    /// Uniquely identifies a single field that can be edited. This may correspond to a property on a
+    /// model object, or can be any other named value.
+    /// </summary>
+    public readonly struct FieldIdentifier
```
rynowak: note: since this is 2x pointer and readonly, it's a good optimization to use `in` anywhere it's a parameter to reduce copying.

src/Components/Components/src/Forms/FieldIdentifier.cs
```diff
@@ -0,0 +1,111 @@
...
+            if (accessorBody is UnaryExpression unaryExpression
+                && unaryExpression.NodeType == ExpressionType.Convert
+                && unaryExpression.Type == typeof(object))
+            {
+                accessorBody = unaryExpression.Operand;
+            }
+
+            if (!(accessorBody is MemberExpression memberExpression))
+            {
+                throw new ArgumentException("The accessor is not supported because its body is not a MemberExpression");
```
rynowak: I'd expect that this is going to see some more clarity. It makes sense to me but I speak compiler.

src/Components/Components/src/Forms/FieldIdentifier.cs
```diff
@@ -0,0 +1,111 @@
...
+                    break;
+                case MemberExpression nestedMemberExpression:
+                    // It would be great to cache this somehow, but it's unclear there's a reasonable way to do
+                    // so, given that it embeds captured values such as "this". We could consider special-casing
+                    // for "() => something.Member" and building a cache keyed by "something.GetType()" with values
+                    // of type Func<object, object> so we can cheaply map from "something" to "something.Member".
+                    var modelLambda = Expression.Lambda(nestedMemberExpression);
+                    var modelLambdaCompiled = (Func<object>)modelLambda.Compile();
+                    model = modelLambdaCompiled();
+                    break;
```
rynowak: You could also do this using reflection directly which would probably be faster and safer. It's possible that `.Compile()` causes unbounded memory growth, but we also use it in MVC so #YOLO

src/Components/Components/src/Forms/FieldIdentifier.cs
```diff
@@ -0,0 +1,111 @@
...
+                default:
+                    // An error message that might help us work out what extra expression types need to be supported
+                    throw new InvalidOperationException($"The accessor is not supported because the model value cannot be parsed from it. Expression: '{memberExpression.Expression}', type: '{memberExpression.Expression.GetType().FullName}'");
```
rynowak: The other thing that's going to come up immediately is indexers. Other than that, I'm not sure MVC supports much else.

src/Components/Components/src/Forms/InputComponents/InputBase.cs
```diff
@@ -0,0 +1,189 @@
...
+
+        /// <summary>
+        /// Gets the associated <see cref="Microsoft.AspNetCore.Components.Forms.EditContext"/>.
+        /// </summary>
+        protected EditContext EditContext { get; private set; }
+
+        /// <summary>
+        /// Gets the <see cref="FieldIdentifier"/> for the bound value.
+        /// </summary>
+        protected FieldIdentifier FieldIdentifier { get; private set; }
```
rynowak: This is dope.

src/Components/Components/src/Forms/InputComponents/InputBase.cs
```diff
@@ -0,0 +1,189 @@
...
+        /// Parses a string to create an instance of <typeparamref name="T"/>. Derived classes can override this to change how
+        /// <see cref="CurrentValueAsString"/> interprets incoming values.
+        /// </summary>
+        /// <param name="value">The string value to be parsed.</param>
+        /// <param name="result">An instance of <typeparamref name="T"/>.</param>
+        /// <param name="validationErrorMessage">If the value could not be parsed, provides a validation error message.</param>
+        /// <returns>True if the value could be parsed; otherwise false.</returns>
+        protected virtual bool TryParseValueFromString(string value, out T result, out string validationErrorMessage)
+            => throw new NotImplementedException($"Components that inherit from {nameof(InputBase<T>)} must override " +
+                $"{nameof(TryParseValueFromString)} in order to use {nameof(CurrentValueAsString)}.");
```
rynowak: I'll take a look a the other places where this gets overridden, but a possible enhancement here would be to use `IConvertable` or `TypeConverter` to handle this.

I expect using `TypeConverter` might require more linker stuff.

src/Components/Components/src/Forms/InputComponents/InputBase.cs
```diff
@@ -0,0 +1,189 @@
...
+
+                // We don't support changing EditContext because it's messy to be clearing up state and event
+                // handlers for the previous one, and there's no strong use case. If a strong use case
+                // emerges, we can consider changing this.
+                throw new InvalidOperationException($"{GetType()} does not support changing the " +
+                    $"{nameof(Forms.EditContext)} dynamically.");
+            }
+
+            // For derived components, retain the usual lifecycle with OnInit/OnParametersSet/etc.
+            return base.SetParametersAsync(ParameterCollection.Empty);
```
rynowak: Why is this better to put after our init logic? fragile base class problem? 

It looks like a gotcha that we have to pass empty here.

src/Components/Components/src/Forms/InputComponents/InputCheckbox.cs
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Components.RenderTree;
+
+namespace Microsoft.AspNetCore.Components.Forms
+{
+    /* This is exactly equivalent to a .razor file containing:
```
rynowak: :(

src/Components/Components/src/Forms/InputComponents/InputCheckbox.cs
```diff
@@ -0,0 +1,34 @@
...
+     *
+     *    @inherits InputBase<bool>
+     *    <input type="checkbox" bind="@CurrentValue" class="@CssClass" />
+     *
+     * The only reason it's not implemented as a .razor file is that we don't presently have the ability to compile those
+     * files within this project. Developers building their own input components should use Razor syntax.
+     */
```
rynowak: Do you know the reason this doesn't work? Or we just haven't tried yet.

src/Components/Components/src/Forms/InputComponents/InputBase.cs
rynowak: Also, is it really reasonable to do things this way? Is checkbox the only case where we don't need to parse?

src/Components/Components/src/Forms/InputComponents/InputDate.cs
```diff
@@ -0,0 +1,108 @@
...
+        /// <inheritdoc />
+        protected override void BuildRenderTree(RenderTreeBuilder builder)
+        {
+            base.BuildRenderTree(builder);
+            builder.OpenElement(0, "input");
+            builder.AddAttribute(1, "type", "date");
+            builder.AddAttribute(2, "class", FieldClass);
+            builder.AddAttribute(3, "value", BindMethods.GetValue(CurrentValueAsString));
+            builder.AddAttribute(4, "onchange", BindMethods.SetValueHandler(__value => CurrentValueAsString = __value, CurrentValueAsString));
+            builder.CloseElement();
```
rynowak: Since these are implemented in C# it is totally possible for us to support splatting right? (not asking for you to add it to this PR)

src/Components/Components/src/Forms/InputComponents/InputDate.cs
```diff
@@ -0,0 +1,108 @@
...
+            else
+            {
+                validationErrorMessage = string.Format(ParsingErrorMessage, FieldIdentifier.FieldName);
+                return false;
+            }
+        }
+
+        static bool TryParseDateTime(string value, out T result)
+        {
+            var success = DateTime.TryParse(value, out var parsedValue);
```
rynowak: As noted in a bug that was filed recently, we're going to have to review all of these for correct usage of CultureInfo. (not asking you to solve it in this PR).

src/Components/Components/src/Forms/InputComponents/InputNumber.cs
```diff
@@ -0,0 +1,173 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Components.RenderTree;
+using System;
+
+namespace Microsoft.AspNetCore.Components.Forms
+{
+    /// <summary>
+    /// An input component for editing numeric values.
+    /// Supported numeric types are <see cref="short"/>, <see cref="int"/>, <see cref="long"/>, <see cref="float"/>, <see cref="double"/>, <see cref="decimal"/>.
```
rynowak: note: we don't have overloads of bind for `short` (because there's no real reason to use it)

src/Components/Components/src/Forms/InputComponents/InputSelect.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            if (typeof(T) == typeof(string))
+            {
+                result = (T)(object)value;
+                validationErrorMessage = null;
+                return true;
+            }
+            else if (typeof(T).IsEnum)
+            {
+                // There's no non-generic Enum.TryParse (https://github.com/dotnet/corefx/issues/692)
```
rynowak: Yeah, this really makes us sad in MVC too.

src/Components/Components/src/Forms/ValidationMessageStore.cs
```diff
@@ -0,0 +1,105 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Linq.Expressions;
+
+namespace Microsoft.AspNetCore.Components.Forms
+{
+    /// <summary>
+    /// Holds validation messages for an <see cref="EditContext"/>.
+    /// </summary>
+    public class ValidationMessageStore
```
rynowak: sealed?

src/Components/Components/src/Forms/ValidationMessageStore.cs
```diff
@@ -0,0 +1,105 @@
...
+            => GetOrCreateMessagesListForField(fieldIdentifier).AddRange(messages);
+
+        /// <summary>
+        /// Gets the validation messages within this <see cref="ValidationMessageStore"/> for the specified field.
+        ///
+        /// To get the validation messages across all validation message stores, use <see cref="EditContext.GetValidationMessages(FieldIdentifier)"/> instead
+        /// </summary>
+        /// <param name="fieldIdentifier">The identifier for the field.</param>
+        /// <returns>The validation messages for the specified field within this <see cref="ValidationMessageStore"/>.</returns>
+        public IEnumerable<string> this[FieldIdentifier fieldIdentifier]
```
rynowak: Another thing we could consider for these APIs is using `IReadOnlyList<>`. It's the non-allocating big brother of `IEnumerable<>`

src/Components/Components/src/Forms/ValidationMessageStore.cs
```diff
@@ -0,0 +1,105 @@
...
+        /// <returns>The validation messages for the specified field within this <see cref="ValidationMessageStore"/>.</returns>
+        public IEnumerable<string> this[Expression<Func<object>> accessor]
+            => this[FieldIdentifier.Create(accessor)];
+
+        /// <summary>
+        /// Removes all messages within this <see cref="ValidationMessageStore"/>.
+        /// </summary>
+        public void Clear()
+        {
+            foreach (var fieldIdentifier in _messages.Keys)
```
rynowak: I notice we're using `Keys` here (and `Values` in `EditContext`). These do allocate (cached), so it's nice to avoid them if it's just inside the implementation and not exposed to users.

src/Components/Components/src/Forms/ValidationSummary.cs
```diff
@@ -0,0 +1,95 @@
...
+    /*
+     * Note: there's no reason why developers strictly need to use this. It's equally valid to
+     * put a @foreach(var message in context.GetValidationMessages()) { ... } inside a form.
+     * This component is for convenience only, plus it implements a few small perf optimizations.
+     */
+
+    /// <summary>
+    /// Displays a list of validation messages from a cascaded <see cref="EditContext"/>.
+    /// </summary>
+    public class ValidationSummary : ComponentBase, IDisposable
```
rynowak: I have some cool thoughts about this and other control-likes, but I'm refraining because I think this is all definitely good enough for us to give to users. 

One burning thought that I have is whether things like this should be sealed. I'm not sure that we know the implications of trying to write component classes that both do rendering and support inheritance well.

src/Components/Components/src/Forms/ValidationSummary.cs
```diff
@@ -0,0 +1,95 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Components.RenderTree;
+using System;
+
+namespace Microsoft.AspNetCore.Components.Forms
```
rynowak: Do these control-likes belong in the `.Browser` assembly since these are coupled to HTML? I guess NavLink is in this assembly. Not urgent if we do want to change it.

src/Components/Components/test/Forms/EditContextDataAnnotationsExtensionsTest.cs
```diff
@@ -0,0 +1,171 @@
...
+{
+    public class EditContextDataAnnotationsExtensionsTest
+    {
+        [Fact]
+        public void CannotUseNullEditContext()
+        {
+            var editContext = (EditContext)null;
+            var ex = Assert.Throws<ArgumentNullException>(() => editContext.AddDataAnnotationsValidation());
+            Assert.Equal("editContext", ex.ParamName);
+        }
```
rynowak: I'm fine with us not writing these kinds of tests unless we think it's going to be especially error prone. We're going to have nullability support soon

src/Components/Components/test/Forms/EditContextDataAnnotationsExtensionsTest.cs
```diff
@@ -0,0 +1,171 @@
...
+        class TestModel
+        {
+            [Required] public string RequiredString { get; set; }
+
+            [Range(1, 100)] public int IntFrom1To100 { get; set; }
+
+#pragma warning disable 649
+            [Required] public string ThisWillNotBeValidatedBecauseItIsAField;
+            [Required] string ThisWillNotBeValidatedBecauseItIsPrivate { get; set; }
+            [Required] internal string ThisWillNotBeValidatedBecauseItIsInternal { get; set; }
```
rynowak: BTW putting data annotations messages in tests is kinda a pit of failure. You have to do work to make sure the tests only run in a single locale, and we can also break if a different platform defines the messages differently.

src/Components/Components/test/Forms/EditContextTest.cs
```diff
@@ -0,0 +1,241 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Components.Forms;
+using System;
+using System.Linq;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Components.Tests.Forms
```
rynowak: nit: remove `.Tests` from these namespaces and change the `BaseNamespace` of the project to match the product-side. This makes it trivial to write unit tests.

src/Components/Components/test/Forms/FieldIdentifierTest.cs
```diff
@@ -0,0 +1,158 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Components.Forms;
```
rynowak: nit: all of your stuff seems to sort with non-`System.` first but we use `System.` first. I thought we defined an editorconfig for this already..

src/Components/Components/test/Forms/InputBaseTest.cs
```diff
@@ -0,0 +1,423 @@
...
+        public async Task ThrowsIfNoValueExpressionIsSupplied()
+        {
+            // Arrange
+            var model = new TestModel();
+            var rootComponent = new TestInputHostComponent<string, TestInputComponent<string>> { EditContext = new EditContext(model) };
+
+            // Act/Assert
+            var ex = await Assert.ThrowsAsync<InvalidOperationException>(() => RenderAndGetTestInputComponentAsync(rootComponent));
+            Assert.Contains($"{typeof(TestInputComponent<string>)} requires a value for the 'ValueExpression' parameter. Normally this is provided automatically when using 'bind-Value'.", ex.Message);
+        }
```
rynowak: 👍

src/Components/test/testassets/BasicTestApp/FormsTest/NotifyPropertyChangedValidationComponent.cshtml
```diff
@@ -0,0 +1,105 @@
...
+<p>
+    There's no requirement for models to implement INotifyPropertyChanged, but if they do,
+    you can easily wire that up to the EditContext. Then you have no need to use the built-in
+    Input* components - you can instead bind to regular HTML elements and still get modification
+    notifications. This provides more flexibility in how the UI is rendered, at the cost of
+    more complexity and boilerplate in your model classes.
+</p>
+<p>
+    This example also shows that you don't strictly have to use EditForm. You can manually
+    cascade an EditContext to the components that integrate with it.
```
rynowak: I think it's really nice that you're able to write this sort of test because it demonstrates that design principles we were trying to apply have worked out in the end.

src/Components/test/testassets/BasicTestApp/FormsTest/NotifyPropertyChangedValidationComponent.cshtml
```diff
@@ -0,0 +1,105 @@
...
+    protected override void OnInit()
+    {
+        editContext = new EditContext(person).AddDataAnnotationsValidation();
+
+        // Wire up INotifyPropertyChanged to the EditContext
+        person.PropertyChanged += (sender, eventArgs) =>
+        {
+            var fieldIdentifier = new FieldIdentifier(sender, eventArgs.PropertyName);
+            editContext.NotifyFieldChanged(fieldIdentifier);
+        };
```
rynowak: I think it's OK to have this be manual for now. If we see people doing it in this release we could consider baking it in.

src/Components/test/testassets/BasicTestApp/FormsTest/NotifyPropertyChangedValidationComponent.cshtml
```diff
@@ -0,0 +1,105 @@
...
+</p>
+
+<form onsubmit="@HandleSubmit">
+    <p>
+        User name:
+        <input bind="@person.UserName" class="@editContext.FieldClass(() => person.UserName)" />
+    </p>
+    <p>
+        Accept terms:
+        <input type="checkbox" bind="@person.AcceptsTerms" class="@editContext.FieldClass(() => person.AcceptsTerms)" />
```
rynowak: This is the kind of place where having an extensible directive-attribute kind of system would be useful. You could imagine someone writing their own little helper to pull out the class and using that attribute instead.

src/Components/test/testassets/BasicTestApp/FormsTest/TypicalValidationComponent.cshtml
```diff
@@ -0,0 +1,83 @@
...
+
+    List<string> submissionLog = new List<string>(); // So we can assert about the callbacks
+
+    Task HandleValidSubmitAsync(EditContext editContext)
+    {
+        submissionLog.Add("OnValidSubmit");
+        StateHasChanged(); // Temporary workaround for event routing bug
+        return Task.CompletedTask;
+    }
+}
```
rynowak: I'm really really happy with how this is going to look once the compiler support is merged. This is what scaffolding is going to generate, it's great to see it come together.

rynowak (commented):
I think this looks amazing! Most of my feedback here is about minor details of the components or how the core pieces of this fit together. I think the primary scenarios have been absolutely nailed

SteveSandersonMS:
> a direct dependency on `System.ComponentModel.Annotations` is okay to put back.

@natemcmaster Thanks. Besides the instructions you mentioned, it was also necessary to add `System.ComponentModel.Annotations` to `SharedFramework.External.props`, since `Microsoft.AspNetCore.Components` is in the shared framework. I'm assuming that follows from your comment (and the fact that we already do have a transitive dependency on it in the shared framework), but if you have any concerns about that please let me know.

src/Components/Components/src/Forms/InputComponents/InputDate.cs
SteveSandersonMS: Fair request. Localisation is tracked as #5471, so I've added a note there about this.

In the meantime, if you want to change your `ParsingErrorMessage` values centrally, you can subclass `InputDate` and set a different `ParsingErrorMessage` in `OnInit`.

src/Components/Components/src/Forms/InputComponents/InputNumber.cs
SteveSandersonMS: Good point. I've changed it to `InvariantCulture`. We'll review further for better localizability when we do #5471.

src/Components/Components/src/Forms/InputComponents/InputNumber.cs
SteveSandersonMS: Good point. Originally it was because by default `number` only allows whole numbers, and this component is meant to accept `float`/`double`/`decimal` too. But I've read up on the spec and now conditionally emit `step="any"` for the non-integral types.

src/Components/Components/src/Forms/EditContext.cs
SteveSandersonMS: Fair point. Filed https://github.com/aspnet/AspNetCore/issues/7680 and removed the comment.

src/Components/Components/src/Forms/EditContext.cs
SteveSandersonMS: I'm curious about the reasoning, but in any case this is removed now.

src/Components/Components/src/Forms/EditContext.cs
SteveSandersonMS: OK, sounds good for future extensibility. I've done it for all of them. It's not free in the `OnFieldChanged` case (1 extra allocation per event) but this isn't per-render any anything ultra perf-critical like that.

src/Components/Components/src/Forms/EditContext.cs
SteveSandersonMS: Yes, it's because:

 1. We want "mark a field unmodified when it's already unmodified" to be a no-op (not an error)
 2. The `EditContext` stores the field states sparsely, so there isn't necessarily any field state tracked for a given field. Especially not if it's already "unmodified", which is the default state.

> What should be the behaviour of these methods when a field can't be found?

That's not really a meaningful case. All possible field identifiers can be "found", because EditContext will track the states for any `FieldIdentifier` value you give to it. The "if not found" checks are just internal implementation details because we track the states sparsely.

I've added a comment summarizing this.

src/Components/Components/src/Forms/EditContextExpressionExtensions.cs
SteveSandersonMS: It's a combination of:

 * Making it easier to reason about the correctness of the core system (specifically, its inability to be put into an invalid state by any given API call). Anything expressed as extension methods is safe because it doesn't change the true API surface.
 * Proving that certain types of extensibility are possible by external code. For example, the DataAnnotations stuff being extension methods is to ensure we're not relying on any private APIs. (I know we could still make the mistake of relying on an `internal`, but that one thing is easier to spot than if everything was in one file).

TBH with the `EditContextExpressionExtensions` cases it doesn't really matter very much, but I'm inclined to leave it as-is unless there are specific objections.

src/Components/Components/src/Forms/EditContextDataAnnotationsExtensions.cs
SteveSandersonMS: Agreed. I *hope* people stick within the one-way flow of update propagation, because then ordering is largely not a concern. But I recognize people might do really dubious things like having a validation mechanism that varies its behavior based on validation messages registered by a different validation system.

I've never really believed in the idea of solving ordering issues by having an `Order` integer - it only scales so far, before you can't manage it either (e.g., validation systems registered by multiple independent NuGet packages that need to be ordered). `Order` only works internally within a single, sufficiently small codebase.

Let's see how it goes.

src/Components/Components/src/Forms/EditForm.cs
SteveSandersonMS: I recognize that inheritance is a very sketchy way of extending or reusing functionality in general, but it is pretty convenient and well-performing in the case of these forms-and-input components. You can subclass `EditForm` to change how it renders (e.g., to put your own special attributes on the `<form>` tag). It's more useful still in the case of the `Input*` components - the base classes are designed around the idea that you'd subclass them to customize rendering.

src/Components/Components/src/Forms/EditContextFieldClassExtensions.cs
SteveSandersonMS: Agreed. A likely direction for the future.

src/Components/Components/src/Forms/EditForm.cs
SteveSandersonMS: Yes! Done.

src/Components/Components/src/Forms/EditContext.cs
SteveSandersonMS: Good point. I've switched this to `yield return` now, so as to retain the ability to short-circuit from the caller's side.

src/Components/Components/src/Forms/FieldIdentifier.cs
SteveSandersonMS: OK, I've done this.

However TBH I don't actually know where the perf tradeoff falls in the WebAssembly case. Mono WebAssembly is like 32-bit .NET, though WebAssembly itself has 64-bit int types. Whether this is faster or slower on Mono WebAssembly may depend on whether they have an optimization to copy-on-pass sufficiently small structs as a single 64-bit value.

src/Components/Components/src/Forms/FieldIdentifier.cs
SteveSandersonMS: The message is only designed so that people can paste it in GitHub issues and we can easily see which parts of the logic would have to be updated to support a given expression type. I don't expect people to follow it beyond that.

src/Components/Components/src/Forms/FieldIdentifier.cs
SteveSandersonMS: Sounds like a good future enhancement. I'd rather not try to predict all the use cases for the first preview.

src/Components/Components/src/Forms/FieldIdentifier.cs
SteveSandersonMS: Indexers aren't supported because they we can't identify them as a field name (we'd need to encode the indexer key in the field name somehow). Maybe one day we might add some extension along those lines, but for now I'd rather not try to solve that.

src/Components/Components/src/Forms/InputComponents/InputBase.cs
SteveSandersonMS: OK, I agree it would be more conventional to make `TryParseValueFromString` an `abstract` method. Done.

src/Components/Components/src/Forms/InputComponents/InputBase.cs
SteveSandersonMS: It's because the base class will call the `OnInit` (etc.) methods, and we want that to happen *after* we've run the subclass's logic to populate `EditContext` etc., in case those lifecycle methods try to access `EditContext` etc.

src/Components/Components/src/Forms/InputComponents/InputCheckbox.cs
SteveSandersonMS: We have to change the `Sdk` etc. AFAIK it can be done, but I am trying to keep this change scoped.

src/Components/Components/src/Forms/InputComponents/InputDate.cs
SteveSandersonMS: Yes, we can implement it in the built-in rendering, but that leads to a cliff edge for anyone who wants to subclass (say) `InputText` to set up custom rendering. They would lose the spatting, until we implement a built-in system for them to use. So until we do that I'd rather not have the feature in the base class.

src/Components/Components/src/Forms/InputComponents/InputNumber.cs
SteveSandersonMS: Removed `short`

src/Components/Components/src/Forms/ValidationMessageStore.cs
SteveSandersonMS: Yes, done.

src/Components/Components/src/Forms/ValidationMessageStore.cs
SteveSandersonMS: We could, but it doesn't help developers much in this case. There isn't much use case in reading back the messages from a single store. Normally you only want to read back all the messages for a given field across all stores in the `EditContext`, which has to be `IEnumerable<string>` because it dynamically combines messages from all stores (and we don't want to build lists on each evaluation).

src/Components/Components/src/Forms/ValidationMessageStore.cs
SteveSandersonMS: Good point. Removed those usages.

src/Components/Components/src/Forms/ValidationSummary.cs
SteveSandersonMS: Currently I'm thinking of subclassing as a very convenient way to customize rendering. Maybe this will turn out to be a mistake. I'd like to see if the usages justify it first. Let me know if you think you really do anticipate it to be a mistake!

src/Components/test/testassets/BasicTestApp/FormsTest/NotifyPropertyChangedValidationComponent.cshtml
SteveSandersonMS: OK, it's good to include this scenario in the considerations then.

src/Components/test/testassets/BasicTestApp/FormsTest/NotifyPropertyChangedValidationComponent.cshtml
SteveSandersonMS: Yes. I shied away from baking it in because (1) it forces another dependency on `System.ComponentModel` and (2) so far we've consistently said about `INotifyPropertyChange`/`INotifyDataErrorInfo`/etc that we're not taking an opinion on them but rather are designing APIs to be convenient to integrate them (and other things) with.

But if this does become a popular option we definitely should consider that.

src/Components/Components/src/Forms/ValidationSummary.cs
SteveSandersonMS: I agree 100%, and in fact the same thing occurred to me over the weekend, and I was planning to do it.

However after trying it, I think it's better if we don't do so right now. The problem is that you then become forced to add `@addTagHelper *, Microsoft.AspNetCore.Components.Browser` in all the projects and Razor component libraries where you want to use those components. We can't auto-add those tag helpers in the default compilation, because that breaks the whole point of the layering. And `@addTagHelper` is extremely non-discoverable in the case where you don't even know where a given component comes from (how do you know to add `Microsoft.AspNetCore.Components.Browser` specifically?).

So, once we replace `@addTagHelper` with `@using`, I think we should move the `Input*` and `NavLink` components to `.Browser`.

src/Components/Components/test/Forms/EditContextDataAnnotationsExtensionsTest.cs
SteveSandersonMS: Yeah, I felt a bit bad about doing it the first time.

Anyway now I've put in explicit messages for all the rules we assert about, so it's independent of locale/platform.

src/Components/Components/test/Forms/EditContextTest.cs
SteveSandersonMS: Done

src/Components/Components/test/Forms/FieldIdentifierTest.cs
SteveSandersonMS: I've never deliberately configured anything about this. I'm just using VS2019 in its out-of-box config. Maybe VS doesn't care about the `.editorconfig`.

I've reordered the namespaces in a bunch of files, hopefully all the new ones.

src/Components/Components/src/Forms/InputComponents/InputBase.cs
```diff
@@ -0,0 +1,206 @@
...
+
+        /// <summary>
+        /// Gets or sets a callback that updates the bound value.
+        /// </summary>
+        [Parameter] Action<T> ValueChanged { get; set; }
+
+        /// <summary>
+        /// Gets or sets an expression that identifies the bound value.
+        /// </summary>
+        [Parameter] Expression<Func<T>> ValueExpression { get; set; }
```
RemiBou: Why do you need both ValueExpression and Value ? it seems redundant. Can you also add some commonly used attributes like "placeholder" or "name"

SteveSandersonMS: It's much cheaper to evaluate `Value` than `ValueExpression`, so components should only accept a `ValueExpression` if they truly need it.

SteveSandersonMS: > Can you also add some commonly used attributes like "placeholder" or "name"

This is covered by the desire to pass through arbitrary attributes, which in turn is waiting for support for a "splatting" syntax within the Razor compiler. I've added `class` and `id` because they are so essential; people who want others will have to subclass the built-in `Input*` components in the short term or wait for build-in attribute passthrough in a later preview.

src/Components/Components/src/Forms/EditContextDataAnnotationsExtensions.cs
```diff
@@ -0,0 +1,101 @@
...
+                messages.Clear(fieldIdentifier);
+                messages.AddRange(fieldIdentifier, results.Select(result => result.ErrorMessage));
+
+                // We have to notify even if there were no messages before and are still no messages now,
+                // because the "state" that changed might be the completion of some async validation task
+                editContext.NotifyValidationStateChanged();
+            }
+        }
+
+        private static bool TryGetValidatableProperty(in FieldIdentifier fieldIdentifier, out PropertyInfo propertyInfo)
```
DAllanCarr: Roll on static locals

SteveSandersonMS:
@rynowak OK, this should be pretty much final now. I still plan to put in a `ValidationMessage` component, but it will be trivial and almost identical to `ValidationSummary`.

The only remaining unknown is how to get the build to succeed (as in, how to make it accept the new reference to `System.ComponentModel.Annotations). I've emailed @natemcmaster with a question about that, but in the meantime this should be ready for a final review now.

eng/Version.Details.xml
```diff
@@ -293,6 +293,14 @@
       <Uri>https://github.com/dotnet/corefx</Uri>
       <Sha>0abec4390b30fdda97dc496594f9b1f9c9b20e17</Sha>
     </Dependency>
+    <Dependency Name="System.ComponentModel" Version="4.3.0">
+      <Uri>https://github.com/dotnet/corefx</Uri>
+      <Sha>000000</Sha>
```
mkArtakMSFT: How will these be corrected?

src/Components/Components/src/Forms/EditForm.cs
```diff
@@ -0,0 +1,139 @@
...
+            // optimizing for the common case where _fixedEditContext never changes.
+            builder.OpenRegion(_fixedEditContext.GetHashCode());
+
+            builder.OpenElement(0, "form");
+            builder.AddAttribute(1, "onsubmit", _handleSubmitDelegate);
+            builder.OpenComponent<CascadingValue<EditContext>>(2);
+            builder.AddAttribute(3, "IsFixed", true);
+            builder.AddAttribute(4, "Value", _fixedEditContext);
+            builder.AddAttribute(5, RenderTreeBuilder.ChildContent, ChildContent?.Invoke(_fixedEditContext));
+            builder.CloseComponent();
```
mkArtakMSFT: I don't see a `method` attribute available in here. Don't we plan to support `post` method?

eng/Version.Details.xml
SteveSandersonMS: According to https://github.com/aspnet/AspNetCore/blob/master/docs/ReferenceResolution.md#example-adding-a-new-dependency, the `Sha` values don't matter and eventually get updated by the system automatically. `000000` is the suggested value to use when adding a new dependency.

src/Components/Components/src/Forms/EditForm.cs
SteveSandersonMS: The important use case is forms that are handled via an `onsubmit` handler, not ones that are directly posted to a server. As such, the `method` isn't applicable in normal use cases.

We are planning to allow arbitrary attribute pass-through at some point (a.k.a. "splatting"), at which point people could set a `method` or any other attribute if they wanted. We probably wouldn't want to put in any special-case support for `method` before then, because the use case is not major.

src/Components/Components/src/Forms/EditContext.cs
rynowak: Ok great, that's a really good reason for the current approach.

src/Components/Components/src/Forms/EditContextExpressionExtensions.cs
rynowak: No, I think these are all fine reasons.

src/Components/Components/src/Forms/FieldIdentifier.cs
rynowak: I think we should be concerned about this in the sense that we need to provide guidance for it.

Imagine you're building a form where users can enter multiple emergency contacts, each one with a name and phone number. How would we tell someone to do this?

src/Components/Components/src/Forms/FieldIdentifier.cs
rynowak: I can appreciate that but I think we can do better. It's really common for people to put in some random object where this construct is used in MVC (`() => new Person()').

What if said something like the following:
```
The provided expression contains a {accessorBody.GetType().Name} which is not supported. FieldIdentifier only support simple member accessors (fields, properties) of the object passed into the EditForm.
```

src/Components/Components/src/Forms/InputComponents/InputDate.cs
rynowak: Fair enough

src/Components/Components/src/Forms/FieldIdentifier.cs
SteveSandersonMS: This should be supported (haven't explicitly verified, but will do):

```
@(() => something[someindex].SomeProperty)
```

That will create a FieldIdentifier with model=`something[someindex]` and fieldname=`SomeProperty`.

What's not supported is this:

```
@(() => something[someindex])
```

... and can't be within this model, because there's no value we can give to "fieldname".

Do you think this limitation is reasonable?

> Imagine you're building a form where users can enter multiple emergency contacts, each one with a name and phone number. How would we tell someone to do this?

In the natural way:

```
@foreach (var contact in person.Contacts)
{
    <InputText bind="contact.Phone" /> (etc)
}
```

or even, if your model requires it:

```
@foreach (var contactKey in person.Contacts.Keys)
{
    <InputText bind="person.Contacts[contactKey].Phone" /> (etc)
}
```

src/Components/Components/src/Forms/ValidationSummary.cs
rynowak: Ok, let's try this and see what happens. 

My expectation is that we'll get feedback from users who say **both**:

1. Please add feature XYZ to the base component because I want it
2. I don't want to subclass because it forces me to copy-paste your code and I don't want that either

However, I don't have a good answer for how we'd actually make this person happy. What you've done here makes it technically possible for this user to solve the problem themselves so that might be the best we can do.

src/Components/Components/src/Forms/ValidationSummary.cs
rynowak: Alright great. I totally agree and hadn't thought about it in that level of detail yet, but I like your plan.

src/Components/Components/test/Forms/FieldIdentifierTest.cs
rynowak: VS is supposed to use `.editorconfig` for this, and yes we do use a non-default setting.

src/Components/Components/src/Forms/ValidationSummary.cs
SteveSandersonMS: My hope is that, in the long term, it will be sufficient to have basic rendering of the standard HTML element corresponding to the input type (which we already have) with passthrough of arbitrary attributes (which we don't yet have). In the cases where someone wants to customize rendering beyond "pass through some more attributes", it's on them to either:

 * Subclass
 * Or, wrap the built-in input in some other component of their own

src/Components/Components/src/Forms/FieldIdentifier.cs
rynowak: > Do you think this limitation is reasonable?

Sure, I think that's eminently reasonable. The one place that users will feel sad about it is if if they try to define a `List<string>` or similar

rynowak (approved):
:shipit: - this is dope. I'm really excited to try this out!

eng/Versions.props
```diff
@@ -25,6 +25,8 @@
     <MicrosoftBclJsonSourcesPackageVersion>4.6.0-preview.19109.6</MicrosoftBclJsonSourcesPackageVersion>
     <MicrosoftCSharpPackageVersion>4.6.0-preview.19109.6</MicrosoftCSharpPackageVersion>
     <MicrosoftWin32RegistryPackageVersion>4.6.0-preview.19109.6</MicrosoftWin32RegistryPackageVersion>
+    <SystemComponentModelPackageVersion>4.3.0</SystemComponentModelPackageVersion>
```
natemcmaster: The corefx team is no longer producing nighties of this assembly. This should be moved down into the "manual" section (look for the comment that says "Stable dotnet/corefx packages no longer updated for .NET Core 3")

eng/Version.Details.xml
natemcmaster: This entry can be removed since we will not be flowing dependency updates automatically. This package is obsolete for .NET Core 3 and no longer builds new versions.

natemcmaster (changes requested):

eng/Version.Details.xml
```diff
@@ -293,6 +293,14 @@
...
+    </Dependency>
+    <Dependency Name="System.ComponentModel.Annotations" Version="4.6.0-preview.19109.6">
+      <Uri>https://github.com/dotnet/corefx</Uri>
+      <Sha>000000</Sha>
```
natemcmaster:
```suggestion
      <Sha>0abec4390b30fdda97dc496594f9b1f9c9b20e17</Sha>
```

natemcmaster: Since we know the sha, let's use it. 000000 was meant to be a suggestion if you _don't_ know the right sha to use.

eng/Version.Details.xml
SteveSandersonMS: Thanks for the tip. Have removed this entry.

eng/Versions.props
SteveSandersonMS: Done

eng/Version.Details.xml
SteveSandersonMS: Thanks - done.

SteveSandersonMS:
@natemcmaster Changed as requested. Please leave a 5☆ review if satisfactory :)

natemcmaster (approved):
⭐️ ⭐️ ⭐️ ⭐️ ⭐️

src/Components/Components/src/Forms/FieldIdentifier.cs
SteveSandersonMS: OK, that sounds good. Done. Also I found that the expression parser could easily be generalised significantly (or rather, it was a bug that it wasn't already as general as expected), so did that plus tests to demonstrate broader cases.

SteveSandersonMS: > of the object passed into the EditForm

`of an object` (can be any object; doesn't have to be editForm.Model).

rynowak:
> Command 'Go' is not supported by Azure Pipelines.
> 
> Supported commands
>      help:
>           Get descriptions, examples and documentation about supported commands
>           Example: help "command_name"
>      run:
>           Run all pipelines or a specific pipeline for this repository using a comment. Use
>           this command by itself to trigger all related pipelines, or specify a pipeline
>           to run.
>           Example: "run" or "run pipeline_name"
> 
> See [additional documentation](https://go.microsoft.com/fwlink/?linkid=2056279).

boom

eng/SharedFramework.External.props
```diff
@@ -90,6 +90,7 @@
   -->
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
     <_CompilationOnlyReference Include="System.Buffers" />
+    <_CompilationOnlyReference Include="System.ComponentModel.Annotations" />
```
dougbu: Why is this here when System.ComponentModel.DataAnnotations is not?

SteveSandersonMS: `System.ComponentModel.Annotations` is the name of the package that contains the `System.ComponentModel.DataAnnotations` types.

PR #7618, opened 2019-02-15
Make reads that return canceled ReadResult throw

jkotalik:
Noticed in https://github.com/aspnet/AspNetCore/pull/7603


davidfowl (approved):
LGTM

Tratcher (approved):

PR #7620, opened 2019-02-15
Disable remote symbol loading for dump files in CI

pakrym:
Should make builds faster


jkotalik (approved):

Tratcher (approved):

Tratcher:
40 minutes faster for that build step! It's no longer the slowest. :-)

pakrym:
I've kept symbol loading just decreased max dump count and enabled the cache.

Issue #7546, opened 2019-02-13
2.1.8 downgraded some assemblies from 2.1.7


natemcmaster:
The 2.1.8 shared framework contains some assemblies older than the 2.1.7 framework. This is a result of incorrect data in the baseline file. For instance: 
https://github.com/aspnet/AspNetCore/blob/4cceccd5689ebf7712c03789df536cbb4cfc2efc/eng/Baseline.xml#L123

The actual latest stable release of Microsoft.AspNetCore is 2.1.7. This can cause runtime exceptions such as https://github.com/aspnet/AspNetCore/issues/3503#issuecomment-463134356 if users target 2.1.7 but rollforward to newer shared frameworks.

cc @dougbu @mikaelm12 @Eilon

Eilon:
Oops. Are there any good workarounds to this? Add explicit reference to the correct version? (Downgrading to older patch is not a "good" workaround.)

natemcmaster:
There are several workarounds. 

1. add `<PackageReference Include="Microsoft.AspNetCore" Version="2.1.7" />` (NB: this is NOT the .App package)
2. Remove version from AspNetCore.App `<PackageReference Include="Microsoft.AspNetCore" />`
3. Update AspNetCore.App: `<PackageReference Include="Microsoft.AspNetCore.App" Version="2.1.8" />`

natemcmaster:
To help us catch this in the future, I think we should consider adding a unit test which reads this baseline and checks that it matches the latest 2.1.\* on nuget.org. The SDK team did this for a similar kind of baseline file problem.

muratg:
This looks like a patch candidate for sure. 

@dougbu Would you be the right person to fix this? Could you prepare a PR and ping me on it, so that we can bring this to shiproom tomorrow? This can be a tactical PR just to fix the issue, but we definitely need to add the test @natemcmaster specifies above to prevent this from happening again.

dougbu:
I'd appreciate it if @mikaelm12 could take a look at this first. @mikaelm12 is that :ok: with you?

But, I'll double-check the current 2.1.9 branding.

mikaelm12:
Yeah, I can do it.

dougbu:
Problem remains in current `release/2.1` branch. I'll fix that up and leave the test to @mikaelm12

dougbu:
See aspnet/AspNetCore#7556

muratg:
Is the following related to this issue?

[...]

PR #7627, opened 2019-02-15
Add baseline version test

mikaelm12:
Issue: https://github.com/aspnet/AspNetCore/issues/7546

### DIFF
--- a/test/SharedFx.UnitTests/SharedFx.UnitTests.csproj
+++ b/test/SharedFx.UnitTests/SharedFx.UnitTests.csproj
@@ -38,6 +38,7 @@
     <PackageReference Include="xunit" Version="$(XunitPackageVersion)" />
     <PackageReference Include="xunit.analyzers" Version="$(XunitAnalyzersPackageVersion)" />
     <PackageReference Include="xunit.runner.visualstudio" Version="$(XunitRunnerVisualstudioPackageVersion)" />
+    <PackageReference Include="McMaster.Extensions.CommandLineUtils" Version="2.2.2" />
   </ItemGroup>
 
 </Project>
--- a/test/SharedFx.UnitTests/SharedFx.UnitTests.csproj
+++ b/test/SharedFx.UnitTests/SharedFx.UnitTests.csproj
@@ -26,6 +26,10 @@
       <_Parameter1>MicrosoftNETCoreAppPackageVersion</_Parameter1>
       <_Parameter2>$(RuntimeFrameworkVersion)</_Parameter2>
     </AssemblyAttribute>
+    <AssemblyAttribute Include="Microsoft.AspNetCore.TestData">
+      <_Parameter1>PreviousAspNetCoreReleaseVersion</_Parameter1>
+      <_Parameter2>$(PreviousAspNetCoreReleaseVersion)</_Parameter2>
+    </AssemblyAttribute>
   </ItemGroup>
 
   <ItemGroup>
--- a/test/SharedFx.UnitTests/SharedFxTests.cs
+++ b/test/SharedFx.UnitTests/SharedFxTests.cs
@@ -38,7 +38,7 @@ namespace Microsoft.AspNetCore
 
             var zipPath = Path.Combine(AppContext.BaseDirectory, zipName);
 
-            if (!Directory.Exists("unzipped"))
+            if (!Directory.Exists(zipPath + "unzipped"))
             {
                 ZipFile.ExtractToDirectory(zipPath, "unzipped");
             }
--- a/test/SharedFx.UnitTests/SharedFxTests.cs
+++ b/test/SharedFx.UnitTests/SharedFxTests.cs
@@ -26,7 +26,6 @@ namespace Microsoft.AspNetCore
             var url = new Uri($"https://dotnetcli.blob.core.windows.net/dotnet/aspnetcore/Runtime/" + previousVersion + "/aspnetcore-runtime-internal-" + previousVersion + "-win-x64.zip");
             var zipName = "assemblies.zip";
             var nugetAssemblyVersions = new Dictionary<string, Version>();
-
             var root = TestData.GetDotNetRoot();
             var dir = Path.Combine(root, "shared", config.Name, config.Version);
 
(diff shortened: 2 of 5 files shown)


natemcmaster (approved):
:shipit: once you have approval from @Eilon or @muratg, too.

This change is a test change only and gives us coverage which would have caught #7546

test/SharedFx.UnitTests/SharedFxTests.cs
```diff
@@ -1,14 +1,76 @@
...
+        {
+            var previousVersion = TestData.GetPreviousAspNetCoreReleaseVersion();
+            var url = $"https://dotnetcli.blob.core.windows.net/dotnet/aspnetcore/Runtime/" + previousVersion + "/aspnetcore-runtime-internal-" + previousVersion + "-win-x64.zip";
+            var zipName = "assemblies.zip";
+            var nugetAssemblyVersions = new Dictionary<string, Version>();
+
+            var root = TestData.GetDotNetRoot();
+            var dir = Path.Combine(root, "shared", config.Name, config.Version);
+
+            using (var testClient = new WebClient())
```
Eilon: This definitely needs retry logic. Blob storage can be very finicky.

Here's an amazing RetryHelper: https://github.com/aspnet/AspNetCore-Internal/blob/master/Infrastructure/src/TriageBuildFailures/RetryHelpers.cs

And here's an example of how it's used:
https://github.com/aspnet/AspNetCore-Internal/blob/018295b06b95ff24e016cb34151785fd30eb06b7/Infrastructure/src/TriageBuildFailures/GitHub/GitHubClientWrapper.cs#L58

test/SharedFx.UnitTests/SharedFxTests.cs
```diff
@@ -1,14 +1,76 @@
...
+            {
+                testClient.DownloadFile(url, zipName);
+            }
+
+            var zipPath = Path.Combine(AppContext.BaseDirectory, zipName);
+
+            if (!Directory.Exists("unzipped"))
```
Eilon: Should this create a new, random folder somewhere? What if there's already junk there from a previous failed run?

Eilon: Also, where is the "current directory" when this is called? Looks like "current directory" and "BaseDirectory" are used but are they necessarily the same?

test/SharedFx.UnitTests/SharedFxTests.cs
```diff
@@ -1,14 +1,76 @@
...
+            {
+                ZipFile.ExtractToDirectory(zipPath, "unzipped");
+            }
+
+            var nugetAssembliesPath = Path.Combine(AppContext.BaseDirectory, "unzipped", "shared", config.Name, previousVersion);
+
+            string[] files = Directory.GetFiles(nugetAssembliesPath, "*.dll");
```
Eilon: `var`

test/SharedFx.UnitTests/SharedFxTests.cs
```diff
@@ -1,14 +1,76 @@
...
+            foreach (string file in files)
```
Eilon: More `var`

test/SharedFx.UnitTests/SharedFxTests.cs
```diff
@@ -1,14 +1,76 @@
...
+            {
+                try
+                {
+                    var assemblyVersion = AssemblyName.GetAssemblyName(file)?.Version;
+                    var splitPath = file.Split('\\');
+                    var dllName = splitPath[splitPath.Length - 1];
```
Eilon: Is this trying to do this? https://docs.microsoft.com/en-us/dotnet/api/system.io.path.getfilename?view=netframework-4.7.2

test/SharedFx.UnitTests/SharedFxTests.cs
```diff
@@ -1,14 +1,76 @@
...
+            }
+
+            var nugetAssembliesPath = Path.Combine(AppContext.BaseDirectory, "unzipped", "shared", config.Name, previousVersion);
+
+            string[] files = Directory.GetFiles(nugetAssembliesPath, "*.dll");
+            foreach (string file in files)
+            {
+                try
+                {
+                    var assemblyVersion = AssemblyName.GetAssemblyName(file)?.Version;
```
Eilon: Is the `?` needed here? If so, why?

Eilon (changes requested):

test/SharedFx.UnitTests/SharedFxTests.cs
```diff
@@ -1,14 +1,76 @@
...
+            files = Directory.GetFiles(dir, "*.dll");
+
+            Assert.All(files, file =>
+            {
+                try
+                {
+                    var localAssemblyVersion = AssemblyName.GetAssemblyName(file)?.Version;
+                    var splitPath = file.Split('\\');
+                    var dllName = splitPath[splitPath.Length - 1];
+                    Assert.True(nugetAssemblyVersions.ContainsKey(dllName), $"Expected {dllName} to be in the downloaded dlls");
```
Eilon: Oh and BTW just checking, are there are no xUnit asserts for these two asserts here? Usually `Assert.True` is frowned upon, unless there's no other reasonable choice.

Eilon (approved):
Looks good. Please send TELL MODE email to shiproom alias with FYI (specify it's for 2.1). I sent you an email with info.

test/SharedFx.UnitTests/SharedFx.UnitTests.csproj
```diff
@@ -34,6 +38,7 @@
     <PackageReference Include="xunit" Version="$(XunitPackageVersion)" />
     <PackageReference Include="xunit.analyzers" Version="$(XunitAnalyzersPackageVersion)" />
     <PackageReference Include="xunit.runner.visualstudio" Version="$(XunitRunnerVisualstudioPackageVersion)" />
+    <PackageReference Include="McMaster.Extensions.CommandLineUtils" Version="2.2.2" />
```
dougbu: I know this is test-only but have we used @natemcmaster's private packages elsewhere?

test/SharedFx.UnitTests/SharedFxTests.cs
dougbu: `Assert.Contains(nugetAssemblyVersions.Keys, ...)` should work here. The only supposed loss would be the custom error message when it fails.

dougbu (approved):

test/SharedFx.UnitTests/SharedFx.UnitTests.csproj
mikaelm12: I mimicked here what we did in the internal repo.

test/SharedFx.UnitTests/SharedFxTests.cs
mikaelm12: We could change to use the `Assert.Contains` api. The error message becomes slightly less clear. I'm indifferent.

PR #7629, opened 2019-02-15
ITlsHandshakeFeature requires Win8+

Tratcher:
https://github.com/aspnet/AspNetCore-Internal/issues/1671#issuecomment-463305888

### DIFF
--- a/src/Servers/HttpSys/test/FunctionalTests/HttpsTests.cs
+++ b/src/Servers/HttpSys/test/FunctionalTests/HttpsTests.cs
@@ -102,9 +102,32 @@ namespace Microsoft.AspNetCore.Server.HttpSys
         }
 
         [ConditionalFact]
+        [OSDontSkipCondition(OperatingSystems.Windows, WindowsVersions.Win7, WindowsVersions.Win2008R2)]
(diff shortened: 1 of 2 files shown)


src/Servers/HttpSys/test/FunctionalTests/HttpsTests.cs
```diff
@@ -102,6 +102,30 @@ public async Task Https_ClientCertRequested_ClientCertPresent()
...
+                try
+                {
+                    var tlsFeature = httpContext.Features.Get<ITlsHandshakeFeature>();
+                    Assert.Null(tlsFeature);
+                }
+                catch (Exception ex)
+                {
+                    return httpContext.Response.WriteAsync(ex.ToString());
+                }
+                return Task.FromResult(0);
```
jkotalik: Task.CompletedTask?

jkotalik (approved):

pakrym (approved):

src/Servers/HttpSys/test/FunctionalTests/HttpsTests.cs
pakrym: just use await

PR #7646, opened 2019-02-16
Remove extra await/state machine in FlushAsync

jkotalik:
Small improvement to FlushAsync for WriteOnlyPipeStream.


src/Http/Http/src/WriteOnlyPipeStream.cs
```diff
@@ -81,9 +81,17 @@ public override void Flush()
         }
 
         /// <inheritdoc />
-        public override async Task FlushAsync(CancellationToken cancellationToken)
+        public override Task FlushAsync(CancellationToken cancellationToken)
         {
-            await InnerPipeWriter.FlushAsync(cancellationToken);
+            var task = InnerPipeWriter.FlushAsync(cancellationToken);
```
benaadams: Can this use the `.GetAsTask()` extension instead?

JamesNK: Place in a shared file in https://github.com/aspnet/AspNetCore/tree/master/src/Shared and reuse here, servers, and SignalR

JamesNK (changes requested):

src/Http/Http/src/WriteOnlyPipeStream.cs
jkotalik: Yeah I think it deserves to be shared now. I will once I’m not OOF.

benaadams (approved):

davidfowl:
Why aren’t these files linked instead of included separately?

jkotalik:
> Why aren’t these files linked instead of included separately

Clarify?

JamesNK (approved):

halter73 (approved):

jkotalik:
All failures have already been filed.

PR #7657, opened 2019-02-16
Align host behavior with respect to exceptions

davidfowl:
- Don't catch errors in IHostedService.StartAsync
- Only catch errors when executing StopAsync but rethrow an aggregate
- Updated the tests

Fixes https://github.com/aspnet/AspNetCore/issues/5900

### DIFF
--- a/src/Hosting/Hosting/src/Internal/HostedServiceExecutor.cs
+++ b/src/Hosting/Hosting/src/Internal/HostedServiceExecutor.cs
@@ -28,7 +28,7 @@ namespace Microsoft.AspNetCore.Hosting.Internal
 
         public Task StopAsync(CancellationToken token)
         {
-            return ExecuteAsync(service => service.StopAsync(token), throwOnfirstFailure: false);
+            return ExecuteAsync(service => service.StopAsync(token), throwOnFirstFailure: false);
         }
 
         private async Task ExecuteAsync(Func<IHostedService, Task> callback, bool throwOnFirstFailure = true)
@@ -43,7 +43,7 @@ namespace Microsoft.AspNetCore.Hosting.Internal
                 }
                 catch (Exception ex)
                 {
-                    if (throwOnfirstFailure)
+                    if (throwOnFirstFailure)
                     {
                         throw;
                     }
--- a/src/Hosting/Hosting/src/Internal/HostedServiceExecutor.cs
+++ b/src/Hosting/Hosting/src/Internal/HostedServiceExecutor.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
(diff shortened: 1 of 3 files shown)


src/Hosting/Hosting/src/Internal/HostedServiceExecutor.cs
```diff
@@ -21,31 +21,17 @@ public HostedServiceExecutor(ILogger<HostedServiceExecutor> logger, IEnumerable<
...
-            }
-            catch (Exception ex)
-            {
-                _logger.ApplicationError(LoggerEventIds.HostedServiceStopException, "An error occurred stopping the application", ex);
-            }
+            return ExecuteAsync(service => service.StopAsync(token), throwOnfirstFailure: false);
         }
 
-        private async Task ExecuteAsync(Func<IHostedService, Task> callback)
+        private async Task ExecuteAsync(Func<IHostedService, Task> callback, bool throwOnfirstFailure = true)
```
nil4:
```suggestion
        private async Task ExecuteAsync(Func<IHostedService, Task> callback, bool throwOnFirstFailure = true)
```

src/Hosting/Hosting/src/Internal/HostedServiceExecutor.cs
```diff
@@ -21,31 +21,17 @@ public HostedServiceExecutor(ILogger<HostedServiceExecutor> logger, IEnumerable<
...
-            }
-            catch (Exception ex)
-            {
-                _logger.ApplicationError(LoggerEventIds.HostedServiceStopException, "An error occurred stopping the application", ex);
-            }
+            return ExecuteAsync(service => service.StopAsync(token), throwOnFirstFailure: false);
         }
 
-        private async Task ExecuteAsync(Func<IHostedService, Task> callback)
+        private async Task ExecuteAsync(Func<IHostedService, Task> callback, bool throwOnFirstFailure = true)
```
JunTaoLuo: What happened to overloads instead of default parameters?

src/Hosting/Hosting/src/Internal/HostedServiceExecutor.cs
```diff
@@ -21,31 +21,17 @@ public HostedServiceExecutor(ILogger<HostedServiceExecutor> logger, IEnumerable<
...
-        public async Task StartAsync(CancellationToken token)
+        public Task StartAsync(CancellationToken token)
         {
-            try
-            {
-                await ExecuteAsync(service => service.StartAsync(token));
-            }
-            catch (Exception ex)
-            {
-                _logger.ApplicationError(LoggerEventIds.HostedServiceStartException, "An error occurred starting the application", ex);
```
JunTaoLuo: nit: Hmm maybe remove these LoggerEventIds? It'll leave gaps but I prefer that over obsolete Ids.

JunTaoLuo (approved):
Changes look fine. A couple of nits.

src/Hosting/Hosting/src/Internal/HostedServiceExecutor.cs
davidfowl: This thing is private, why complicate it with overloads?

src/Hosting/Hosting/src/Internal/HostedServiceExecutor.cs
davidfowl: Will do that as a follow up because the CI is slow AF.

PR #7658, opened 2019-02-16
Allow immediately completing Flushes from Dispose to complete

benaadams:
Dispose of a TextWrtier calls Flush unconditionally even if there is nothing to Flush, so we need to guard against immediately completing Flushes vs sync Flushes when sync IO is disallowed


Tratcher:
TextWriter.DisposeAsync is the real solution to this.

Issue #2442, opened 2018-01-01
Add support for .webmanifest file extension


aspnet-hello:
_From @madskristensen on Friday, October 27, 2017 3:57:16 PM_

This is used in Progressive Web Apps (PWAs) for the manifest JSON file. 

The content type should be `application/manifest+json`

_Copied from original issue: aspnet/StaticFiles#219_

aspnet-hello:
_From @muratg on Friday, October 27, 2017 4:00:02 PM_

cc @glennc

aspnet-hello:
_From @Tratcher on Friday, October 27, 2017 4:13:03 PM_

Is there a spec reference?

aspnet-hello:
_From @madskristensen on Friday, October 27, 2017 4:24:22 PM_

<https://w3c.github.io/manifest/#media-type-registration> shows the content type `application/manifest+json` and a little further down in the same chapter the `.webmanifest` file extension.

Sorry, I should have added this when I opened the bug

aspnet-hello:
_From @RehanSaeed on Monday, October 30, 2017 2:59:39 AM_

Just saw this issue, I already raised https://github.com/aspnet/StaticFiles/pull/221 to fix this.

muratg:
Backlogging this.

cc @glennc FYI.

willdean:
There's an additional imperative for this now, because the usual work-around of adding the required mapping to a `StaticFilesOptions` object and passing it to `UseStaticFiles`  then prevents (for reasons I don't properly understand) Blazor Server from initialising its static file service properly.   So you can serve static .webmanifest files, or you can have Blazor Server, but you can't easily do both.

This seems to be because if you pass `StaticFileOptions` to `UseStaticFiles`, the framework stops calling `ConfigureStaticFilesOptions`, which then means that Blazor Server doesn't work (because blazor.server.js isn't served).   

Could we please just have the mapping entry?  Someone even filed a one-line PR for this over two years ago #221 - if just seemed to get lost somehow.

Tratcher:
@danroth27 @anurse We're getting more hits on this. See the linked PR #7664.

Tratcher:
It's documented by w3c, but not registered with IANA.
https://w3c.github.io/manifest/#media-type-registration
https://www.iana.org/assignments/media-types/media-types.xhtml

danroth27:
@SteveSandersonMS How are we handling the manifest media type currently in the new PWA support for Blazor WebAssembly?

[...]

PR #7664, opened 2019-02-17
Support .webmanifest static files.

scottsauber:
Add support for static files with .webmanifest extensions which are commonly used for Progressive Web Apps - https://developer.mozilla.org/en-US/docs/Web/Manifest

Addresses #2442

### DIFF
--- a/src/Middleware/StaticFiles/src/FileExtensionContentTypeProvider.cs
+++ b/src/Middleware/StaticFiles/src/FileExtensionContentTypeProvider.cs
@@ -339,6 +339,7 @@ namespace Microsoft.AspNetCore.StaticFiles
                 { ".wcm", "application/vnd.ms-works" },
                 { ".wdb", "application/vnd.ms-works" },
                 { ".webm", "video/webm" },
+                { ".webmanifest", "application/manifest+json" },
                 { ".webp", "image/webp" },
                 { ".wks", "application/vnd.ms-works" },
                 { ".wm", "video/x-ms-wm" },


scottsauber:
@shirhatti - anything else you need from me on this?

Tratcher:
FYI: https://github.com/aspnet/AspNetCore/blob/a9d3f9c22b16e16f017a36a6e8ae72902c2a698d/src/Middleware/StaticFiles/src/FileExtensionContentTypeProvider.cs#L14-L17

scottsauber:
@tratcher - ok so do you want me to just close this PR? No sense in leaving it open if there is no desire to merge it.

huysentruitw:
@Tratcher @shirhatti  this seems to be a problem for us when we enable a service worker in an Angular 8.2. Currently we need to use this work-around:

```csharp
FileExtensionContentTypeProvider provider = new FileExtensionContentTypeProvider();
provider.Mappings[".webmanifest"] = "application/manifest+json";

app.UseStaticFiles(new StaticFileOptions
{
    ContentTypeProvider = provider,
});
```

Since more and more PWA's are being created, it seems a very logical step forward to be able to host this content type by default (so nobody has to waste time searching for this solution).

Thanks for considering.

stephenlautier:
But why not support it if its still being used? Unless its obsolete, which doesnt seems like it - it will make our lives much easier by this small change

huysentruitw:
Setting up a new project and here we are again... What do we miss to have this type in the framework? Do we need to create an official project template for PWA's?

When starting from the ASP.NET Core + Angular template, converting it to a PWA is as simple as executing `ng add @angular/pwa --project ClientApp` and then you're up for hours of googling why it doesn't work in your ASP.NET Core host until you come up at this issue... We can do better, right?

image

PR #7668, opened 2019-02-17
Don't hardcode fixed size for GetMemory in ChunkedTests

benaadams:
Pass the length from GetMemory to the receiver so it knows what chunk size to expect.

Should fix the ChunksWithGetMemoryWithInitialFlushWorks issue in the dep updates https://github.com/aspnet/AspNetCore/pull/7640

/cc @jkotalik @Tratcher

### DIFF
--- a/src/Servers/Kestrel/test/InMemory.FunctionalTests/ChunkedResponseTests.cs
+++ b/src/Servers/Kestrel/test/InMemory.FunctionalTests/ChunkedResponseTests.cs
@@ -469,6 +469,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.InMemory.FunctionalTests
         [Fact]
         public async Task ChunksWithGetMemoryLargeWriteBeforeFirstFlush()
         {
+            var length = new IntAsRef();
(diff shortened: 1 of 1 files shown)


Tratcher:
Superseded by https://github.com/aspnet/AspNetCore/pull/7671

src/Servers/Kestrel/test/InMemory.FunctionalTests/ChunkedResponseTests.cs
```diff
@@ -517,6 +526,8 @@ public async Task ChunksWithGetMemoryLargeWriteBeforeFirstFlush()
         [Fact]
         public async Task ChunksWithGetMemoryWithInitialFlushWorks()
         {
+            var length = new IntAsRef();
```
pakrym: @benaadams, why did you need this? Isn't local capturing enough?

benaadams: Didn't trust it :) Was just being explicit that that was the intent

Issue #7541, opened 2019-02-13
SignalR / Support ServiceWorkers


unsafecode:
### Is your feature request related to a problem? Please describe.
Looks like the SignalR Javascript bundle does NOT support ServiceWorkers, due to lacking of XMLHttpRequest object in SW scope.

### Describe the solution you'd like
Provide an additional HttpClient implementation that uses the `fetch` primitive instead.

### Describe alternatives you've considered
No alternatives :(

### Additional context
N/A

bradygaster:
cc @davidfowl this seems similar to an issue @halter73 merged in #7058. Thoughts on this?

davidfowl:
It’s not the same thing service workers an web workers are different and though there might be some overlap in the fixed I’m not sure this covers everything. It seems like we’ll need to switch to fetch when available as part of supporting service workers. 

I’m not 100% sure about using SignalR in service workers though @anurse did some investigation here

unsafecode:
@bradygaster @davidfowl I made some changes and it works now. How can I submit a PR?

davidfowl:
Sure that would be great!

halter73:
> How can I submit a PR?

https://github.com/aspnet/AspNetCore/blob/master/CONTRIBUTING.md#submitting-a-pull-request

@unsafecode Thanks for looking into this. The reason we just added web worker support is because an external contribute submitted PR.

unsafecode:
PR submitted at #7672

davidfowl:
@unsafecode While the PR is good and we're likely going to merge it because fetch support is awesome (we punted it here https://github.com/aspnet/SignalR/issues/1240) but I don't understand the use case for signalr inside of a service worker? How do you plan to use it?

unsafecode:
@davidfowl We're build an enterprise-wide PWA that:
1) Uses Saga-based distributed transactions on microservices-oriented application, so we needed a way to notify the client when operations actually complete
2) We wanted to re-use the same approach to send "push notification" without WebPush. And SignalR in SW enables to trigger Notifications.

davidfowl:
Can you describe how SignalR fits in? Doesn't the service worker die on idle? Isn't the point to avoid a long running connection in the first place? Service Workers are about intermittent connectivity while SignalR/WebSockets are about long running connections. Here's a quote from one of the many blog posts on this topic:

[...]

PR #7378, opened 2019-02-08
 Add SignalR web worker sample and test

halter73:
Thanks to @dukhanov for adding the web worker support with #7058.

PR #7672, opened 2019-02-17
Add fetch-based HTTP client

unsafecode:
- Provide a new `FetchHttpClient `to use `fetch `primitive and allow usage within ServiceWorker scope
- Extend `DefaultHttpClient `to leverage when client when needed

Addresses #7541 (in this specific format)

### DIFF
--- a/src/SignalR/clients/ts/signalr/src/FetchHttpClient.ts
+++ b/src/SignalR/clients/ts/signalr/src/FetchHttpClient.ts
@@ -98,7 +98,7 @@ export class FetchHttpClient extends HttpClient {
           });
         })
         .catch((error) => {
-          if (error.message === "timemout") {
+          if (error.message === "timeout") {
             this.logger.log(LogLevel.Warning, `Timeout from HTTP request.`);
             reject(new TimeoutError());
           }
--- a/src/SignalR/clients/ts/signalr/src/FetchHttpClient.ts
+++ b/src/SignalR/clients/ts/signalr/src/FetchHttpClient.ts
@@ -1,0 +1,123 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+import { AbortError, HttpError, TimeoutError } from "./Errors";
+import { HttpClient, HttpRequest, HttpResponse } from "./HttpClient";
+import { ILogger, LogLevel } from "./ILogger";
+
+export class FetchHttpClient extends HttpClient {
+  private readonly logger: ILogger;
+
+  public constructor(logger: ILogger) {
+    super();
+    this.logger = logger;
+  }
+
+  /** @inheritDoc */
+  public send(request: HttpRequest): Promise<HttpResponse> {
+    // Check that abort was not signaled before calling send
+    if (request.abortSignal && request.abortSignal.aborted) {
+      return Promise.reject(new AbortError());
+    }
+
+    if (!request.method) {
+      return Promise.reject(new Error("No method defined."));
+    }
+    if (!request.url) {
+      return Promise.reject(new Error("No url defined."));
+    }
+
+    return new Promise<HttpResponse>((resolve, reject) => {
+      // https://developers.google.com/web/updates/2017/09/abortable-fetch
+      const controller = new AbortController();
+      const signal = controller.signal;
+
+      const requestInfo: RequestInit = {
+        body: request.content || "",
+        credentials: "include",
+        headers: {
+          "Content-Type": "text/plain;charset=UTF-8",
+          "X-Requested-With": "fetch",
+          ...request.headers,
+        },
+        method: request.method!,
+        signal,
+      };
+
+      if (request.abortSignal) {
+        request.abortSignal.onabort = () => {
+          controller.abort();
+          reject(new AbortError());
+        };
+      }
+
+      timeout(request.timeout || 10000, fetch(request.url!, requestInfo))
+        .then((response: Response) => {
+          if (!response.ok) {
+            throw new Error(`${response.status}: ${response.statusText}.`);
+          }
+          return response;
+        })
+        .then((response) => {
+          if (request.abortSignal) {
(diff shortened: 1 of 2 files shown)


src/SignalR/clients/ts/signalr/src/FetchHttpClient.ts
```diff
@@ -0,0 +1,123 @@
...
+                response.statusText,
+                payload,
+              ),
+            );
+          }).catch((error) => {
+            reject(new HttpError(response.statusText, response.status));
+          });
+        })
+        .catch((error) => {
+          if (error.message === "timemout") {
```
campersau:
```suggestion
          if (error.message === "timeout") {
```

src/SignalR/clients/ts/signalr/src/FetchHttpClient.ts
```diff
@@ -0,0 +1,123 @@
...
+
+      if (request.abortSignal) {
+        request.abortSignal.onabort = () => {
+          controller.abort();
+          reject(new AbortError());
+        };
+      }
+
+      timeout(request.timeout || 10000, fetch(request.url!, requestInfo))
+        .then((response: Response) => {
```
davidfowl: I’d rather have set timeout cancel the request rather than have the promises race for completion (also see 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/rac)e

src/SignalR/clients/ts/signalr/src/FetchHttpClient.ts
```diff
@@ -0,0 +1,123 @@
...
+    return new Promise<HttpResponse>((resolve, reject) => {
+      // https://developers.google.com/web/updates/2017/09/abortable-fetch
+      const controller = new AbortController();
+      const signal = controller.signal;
+
+      const requestInfo: RequestInit = {
+        body: request.content || "",
+        credentials: "include",
+        headers: {
+          "Content-Type": "text/plain;charset=UTF-8",
```
davidfowl: @Tratcher we need to update the auth code to handle this case as well. Fetch doesn’t set the headers we expect and of course we can set it but it’s clear that it’s not the right way to check

src/SignalR/clients/ts/signalr/src/FetchHttpClient.ts
```diff
@@ -0,0 +1,123 @@
...
+              content = response.text();
+              break;
+          }
+
+          content.then((payload) => {
+            resolve(
+              new HttpResponse(
+                response.status,
+                response.statusText,
+                payload,
```
davidfowl: I dunno if it’s problematic that we lose the responseType here, I need to look at the calling code.

src/SignalR/clients/ts/signalr/src/FetchHttpClient.ts
Tratcher: What else can auth check?

davidfowl: A header that doesn’t assume XMLHttRequest something with a better semantic name that means don’t redirect. In essence, it’s a feature of the cookie auth instead of Ajax detection. X-AspNetCore-DisableCookieAuthRedirect

src/SignalR/clients/ts/signalr/src/FetchHttpClient.ts
```diff
@@ -0,0 +1,123 @@
...
+      // https://developers.google.com/web/updates/2017/09/abortable-fetch
+      const controller = new AbortController();
+      const signal = controller.signal;
+
+      const requestInfo: RequestInit = {
+        body: request.content || "",
+        credentials: "include",
+        headers: {
+          "Content-Type": "text/plain;charset=UTF-8",
+          "X-Requested-With": "fetch",
```
davidfowl: You may want to change this to match this  https://github.com/aspnet/AspNetCore/blob/4364246081da5ae4c26f417a006dd6ff77563063/src/SignalR/clients/ts/signalr/src/NodeHttpClient.ts#L38 (see the above discussion and the comment in the referenced source)

src/SignalR/clients/ts/signalr/src/FetchHttpClient.ts
```diff
@@ -0,0 +1,123 @@
...
+      };
+
+      if (request.abortSignal) {
+        request.abortSignal.onabort = () => {
+          controller.abort();
+          reject(new AbortError());
+        };
+      }
+
+      timeout(request.timeout || 10000, fetch(request.url!, requestInfo))
```
davidfowl: Can't you use the async/await syntax here instead of using promises directly?

halter73:
Thanks for the PR! Can you please add a Service Workers sample and test for this similar to what we did for Web Workers with #7378?

BrennanConroy:
@unsafecode Are you planning on continuing with this PR or would you like us to take over when we have time to finish it up?

unsafecode:
I'll try to get this done next week

brazierdavid:
I built this from source to try it out and got an error (more of a warning treated as error)

anurse:
Given that we don't really think there's a scenario for SignalR in **Service** Workers at this time (**Web** Workers make sense but Service Workers are shutdown regularly and persistent connections will be broken by that) I don't think we're going to take this. Apologies for any confusion on the issue. There was some confusion between **Service** workers and **Web** workers on the issue, sorry about that.

We do think that migrating to `fetch` makes sense, but we're not prepared to do that yet (since we need to have a polyfill for browsers that don't support it).

anurse:
Re-opening. We'll use this to resolve https://github.com/aspnet/AspNetCore/issues/9444 instead.

anurse:
Update: we're OK taking this, but we do need the feedback we posted addressed. @unsafecode if you're still interested in merging this, take a look at the feedback and let us know!

anurse:
Closing this PR as we haven't heard from you since our last set of feedback. If you're still interested in contributing this change, take a look at the feedback and reopen when you're able to address it and rebase on the latest version of the repo!

PR #7675, opened 2019-02-18
Add negotiation response support

BrennanConroy:
Including support for the Azure SignalR Server (ASRS)!

Fixes https://github.com/aspnet/AspNetCore/issues/7623 too


src/SignalR/clients/cpp/src/signalrclient/connection_impl.cpp
```diff
@@ -84,94 +84,139 @@ namespace signalr
             m_message_id = m_groups_token = m_connection_id = _XPLATSTR("");
         }
 
+        return start_negotiate(m_base_url, 0);
+    }
+
+    pplx::task<void> connection_impl::start_negotiate(const web::uri& url, int redirect_count)
+    {
+        if (redirect_count >= 100)
```
halter73: Nit: Make 100 a constant. Maybe call it MAX_NEGOTIATE_REDIRECTS and put it in constants.h

src/SignalR/clients/cpp/src/signalrclient/connection_impl.cpp
```diff
@@ -84,94 +84,139 @@ namespace signalr
...
+                    foundWebsockets = true;
+                    break;
                 }
-                connection->m_connection_id = std::move(negotiation_response.connection_id);
+            }
 
-                // TODO: check available transports
+            if (!foundWebsockets)
+            {
+                return pplx::task_from_exception<void>(signalr_exception(_XPLATSTR("WebSockets is the only supported transport currently")));
```
halter73: Nit: "WebSockets is the only supported transport currently" -> "The server does not support WebSockets which is currently the only transport supported by this client."

src/SignalR/clients/cpp/src/signalrclient/connection_impl.cpp
```diff
@@ -84,94 +84,139 @@ namespace signalr
...
-                    return pplx::task_from_exception<negotiation_response>(_XPLATSTR("connection no longer exists"));
+                    auto headers = connection->m_signalr_client_config.get_http_headers();
+                    headers.add(_XPLATSTR("Authorization"), _XPLATSTR("Bearer ") + negotiation_response.accessToken);
+                    connection->m_signalr_client_config.set_http_headers(headers);
                 }
-                return request_sender::negotiate(*connection->m_web_request_factory, connection->m_base_url,
-                    connection->m_query_string, connection->m_signalr_client_config);
-            }, m_disconnect_cts.get_token())
-            .then([weak_connection](negotiation_response negotiation_response)
+                return connection->start_negotiate(negotiation_response.url, redirect_count + 1);
```
halter73: The redirect query string should replace m_query_string until the client is restarted. Either that, or remove the query string as something that can configured independently of the url like the other new clients.

src/SignalR/clients/cpp/src/signalrclient/connection_impl.cpp
```diff
@@ -84,94 +84,139 @@ namespace signalr
...
+                return pplx::task_from_exception<void>(signalr_exception(negotiation_response.error));
+            }
+
+            if (!negotiation_response.url.empty())
+            {
+                if (!negotiation_response.accessToken.empty())
                 {
-                    return pplx::task_from_exception<negotiation_response>(_XPLATSTR("connection no longer exists"));
+                    auto headers = connection->m_signalr_client_config.get_http_headers();
+                    headers.add(_XPLATSTR("Authorization"), _XPLATSTR("Bearer ") + negotiation_response.accessToken);
```
halter73: Does this replace the previous Authorization header, or just add to it? Does this get cleared when the client stops?

src/SignalR/clients/cpp/src/signalrclient/connection_impl.cpp
```diff
@@ -84,94 +84,139 @@ namespace signalr
...
+                        .append(translate_connection_state(connection->get_connection_state())));
 
-                    connection->m_transport = nullptr;
-                    connection->change_state(connection_state::disconnected);
-                    connection->m_start_completed_event.set();
-                    start_tce.set_exception(std::current_exception());
+                    _ASSERTE(false);
                 }
 
                 return pplx::task_from_result();
```
halter73: Why return a completed task here only to immediately continue into the following then block? Couldn't you just put the following lines directly here instead and save yourself all the then block boiler plate to reacquire the connection from the weak reference and so forth below?

```
connection->m_start_completed_event.set();
start_tce.set();
```

halter73 (changes requested):

src/SignalR/clients/cpp/src/signalrclient/connection_impl.cpp
mikaelm12: Maybe prefix this with the regular message of "No compatible transports found on the server"

src/SignalR/clients/cpp/src/signalrclient/connection_impl.cpp
BrennanConroy: Nope, if you look closely the `.then` is actually for a bigger scope and will handle cleanup for failures from earlier parts of the negotiate logic.

src/SignalR/clients/cpp/test/signalrclienttests/connection_impl_tests.cpp
```diff
@@ -307,6 +342,193 @@ TEST(connection_impl_start, start_fails_if_connect_request_times_out)
...
+        connection->start().get();
+        ASSERT_TRUE(false); // exception not thrown
+    }
+    catch (const signalr_exception & e)
+    {
+        ASSERT_STREQ("WebSockets is the only supported transport currently", e.what());
+    }
+}
+
+TEST(connection_impl_start, negotiate_follows_redirect)
```
mikaelm12: Add a test that checks that the query string is preserved

src/SignalR/clients/cpp/test/signalrclienttests/connection_impl_tests.cpp
```diff
@@ -285,16 +285,51 @@ TEST(connection_impl_start, start_fails_if_negotiate_request_fails)
...
+        connection->start().get();
+        ASSERT_TRUE(false); // exception not thrown
+    }
+    catch (const signalr_exception & e)
+    {
+        ASSERT_STREQ("bad negotiate", e.what());
+    }
+}
+
+TEST(connection_impl_start, start_fails_if_negotiate_response_does_not_have_websockets)
```
mikaelm12: Add a test for an invalid negotiate response.

mikaelm12: And with an empty transport list.

src/SignalR/clients/cpp/src/signalrclient/request_sender.cpp
```diff
@@ -20,11 +20,45 @@ namespace signalr
...
                 auto negotiation_response_json = web::json::value::parse(body);
-                negotiation_response response
+
+                negotiation_response response;
+
+                if (negotiation_response_json.has_field(_XPLATSTR("error")))
+                {
+                    response.error = negotiation_response_json[_XPLATSTR("error")].as_string();
+                    return std::move(response);
+                }
```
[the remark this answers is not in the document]
BrennanConroy: Sure, no harm in adding it

src/SignalR/clients/cpp/src/signalrclient/connection_impl.cpp
BrennanConroy: `.add` was wrong, good catch. I now replace the header. When the client stops the config will be overwritten when it starts again.

src/SignalR/clients/cpp/src/signalrclient/connection_impl.cpp
BrennanConroy: Removed the query string entirely 👍

src/SignalR/clients/cpp/src/signalrclient/connection_impl.cpp
```diff
@@ -84,94 +83,138 @@ namespace signalr
...
         pplx::task_from_result()
-            .then([weak_connection]()
+            .then([weak_connection, url]()
+        {
+            auto connection = weak_connection.lock();
+            if (!connection)
             {
-                auto connection = weak_connection.lock();
-                if (!connection)
+                return pplx::task_from_exception<negotiation_response>(_XPLATSTR("connection no longer exists"));
```
mikaelm12: Capitalization and periods for the exception messages

BrennanConroy: I'll do an exception/log message pass in a different PR as there are a lot that need to be updated.

halter73 (approved):

Issue #6980, opened 2019-01-24
Support for SignalR  C++ client: "aren't currently supported and may never be".


stukselbax:
### What to do with old C++ clients that used ASP.NET SignalR client?
We have a C++ application which communicates with ASP.NET SignalR server. We are in phase of migrating our server solutions to AspNetCore, and one of issue is that the state of  C++ client isn't clear.

[On docs:](https://docs.microsoft.com/en-us/aspnet/core/signalr/supported-platforms?view=aspnetcore-2.1#unsupported-clients)

> The following clients are available but are experimental or unofficial. They aren't currently supported and may never be.
> C++ client

[On github:](https://github.com/aspnet/AspNetCore/tree/master/src/SignalR/clients/cpp)

> There are no plans to ship this client at this time.

It would be great if the cpp client will be officially supported by AspNetCore team. Actually we didn't grab and try current C++ client code due to above reason.

As for now we have to support both ASP.NET SignalR and new AspNetCore SignalR servers.

mikaelm12:
We currently have the ASP.NET Core SignalR C++ client as  part of our 3.0 milestone https://github.com/aspnet/AspNetCore/issues/5301. 👍

bradygaster:
Reiterating what @mikaelm12 said - it is our goal to ship this soon. if you need more info let us know, for now i will close this as we've discussed our plans.

wegylexy:
I think the C++ client is getting a wrong `connection_id`. It doesn't match with the one server side. The server one seems correct.

PR #7681, opened 2019-02-18
Fix mismatch of connection ID between SignalR C++ client and hub

wegylexy:
Summary of the changes
 - Append negotiated connection ID to connection URL

Addresses #6980

### DIFF
--- a/src/SignalR/clients/cpp/src/signalrclient/connection_impl.cpp
+++ b/src/SignalR/clients/cpp/src/signalrclient/connection_impl.cpp
@@ -254,7 +254,11 @@ namespace signalr
     pplx::task<void> connection_impl::send_connect_request(const std::shared_ptr<transport>& transport, const pplx::task_completion_event<void>& connect_request_tce)
     {
         auto logger = m_logger;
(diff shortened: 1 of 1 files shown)


BrennanConroy:
Thanks for fixing this!

Would it be possible for you to add a test or two?

wegylexy:
I lack experience in writing automated tests for client. It'd require a method in the hub to return the actual connection ID so the client can compare, or another way for connections without hubs. And it needs a server running to test the client.

BrennanConroy:
We currently only have unit tests for the C++ client. You could try it out if you'd like by looking at some of the current tests and copying one that's similar to the functionality we want to test. https://github.com/aspnet/AspNetCore/blob/master/src/SignalR/clients/cpp/test/signalrclienttests/connection_impl_tests.cpp#L53

Or I can do it later if you'd like 😄

wegylexy:
I'll let you do it.

BrennanConroy (approved):

BrennanConroy:
Thanks for the PR!

Issue #5388, opened 2018-11-19
Add Func<T1...Tn, Task> extension methods to HubConnectionExtensions


AlkisFortuneFish:
The Action<T1...Tn> extension methods in HubConnectionExtension that wrap around the base Func<object[], object, Task> On method are very handy. It would be very useful to have Func<T1...Tn, Task> extension methods, allowing async delegates to be registered using the higher level API.

PR #7700, opened 2019-02-19
Add support for Async On Delegates

mikaelm12:
Fixes: #5388 
Add On overloads that take async handlers. It's actually one less layer of indirection since the current handlers just return a completed task to match the default ` Func<object[], Task>`api.

### DIFF
--- a/src/SignalR/clients/csharp/Client.Core/src/HubConnectionExtensions.cs
+++ b/src/SignalR/clients/csharp/Client.Core/src/HubConnectionExtensions.cs
@@ -242,5 +242,211 @@ namespace Microsoft.AspNetCore.SignalR.Client
                 return currentHandler(parameters);
             }, handler);
         }
+
+
+        /// <summary>
+        /// Registers a handler that will be invoked when the hub method with the specified method name is invoked.
+        /// </summary>
+        /// <param name="hubConnection">The hub connection.</param>
+        /// <param name="methodName">The name of the hub method to define.</param>
+        /// <param name="handler">The handler that will be raised when the hub method is invoked.</param>
+        /// <returns>A subscription that can be disposed to unsubscribe from the hub method.</returns>
+        public static IDisposable On(this HubConnection hubConnection, string methodName, Func<Task> handler)
+        {
+            if (hubConnection == null)
+            {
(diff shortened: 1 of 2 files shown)


davidfowl (dismissed):
I think these need to be called OnAsync. Pretty sure the compiler will get confused during overload resolution...

mikaelm12:
The overloads are resolved unambiguously because the return value in the delegate.  They can all be called `On` without any conflicts.

BrennanConroy (approved):

halter73 (approved):

mikaelm12:
@davidfowl Everything seems to work in C# 6,5,4 and 3  I'm going to go ahead and merge this.

rynowak:
> The overloads are resolved unambiguously because the return value in the delegate. They can all be called `On` without any conflicts.

BTW the issue with this usually occurs while you're typing something out. If you have an error inside the lambda you will see some other wierd behavior because it prefers the `Func<T, Task>` overload to `Action<T>` in that case. It's only really bad if they take different parameter types, because then the experience is totally broken.

mikaelm12:
> this usually occurs while you're typing something out

Do you think this warrants changing the name of the overloads?

rynowak:
> Do you think this warrants changing the name of the overloads?

It depends. If they take the same parameter types it's probably fine. If they take different parameter types its pretty much game over.  We did this in health chexxxx because there are a variety of signatures. https://github.com/aspnet/Extensions/blob/master/src/HealthChecks/HealthChecks/src/DependencyInjection/HealthChecksBuilderDelegateExtensions.cs

halter73:
> If they take different parameter types its pretty much game over.

@rynowak Can you give us an example of this that confuses overload resolution? It looks like the current `Func<Task>` On() overloads in this PR are resolve without issue.

I would prefer just to keep the name On() without the Async if possible. It's not like it returns a Task. Would this really prevent us from adding more overloads in the future?

mikaelm12:
Discussed with the potential experience issues @rynowak IRL. I think this is okay to more forward. If we get feedback on this in Preview 3 we can change the names to `OnAsync`, but for the time being keeping them as `On` seems like the best option.

PR #7706, opened 2019-02-19
(nit) Move enumerating extra headers to separate method

benaadams:
(nit) Only pay the stack zeroing costs (40 bytes) for the 2 x enumerators if used and trims 658 bytes from the method taking it to 71 bytes keeping the data cache hotter.

Before
```asm
; Assembly listing for method HttpResponseHeaders:CopyTo(byref):this
; Lcl frame size = 200

G_M60361_IG01:
       push     rdi
       push     rsi
       push     rbp               ; 2 extra reg pushes to stack
       push     rbx               ;
       sub      rsp, 200
       vzeroupper 
       mov      rsi, rcx
       lea      rdi, [rsp+28H]
       mov      ecx, 40           ; zero 40 bytes of stack
       xor      rax, rax          ;
       rep stosd                  ;
       mov      rcx, rsi
       mov      rdi, rcx
       mov      rsi, rdx

G_M60361_IG02:
       mov      rcx, rdi
       mov      rdx, rsi
       call     HttpResponseHeaders:CopyToFast(byref):this
       cmp      gword ptr [rdi+8], 0
       je       G_M60361_IG23
       ;...
       
; Total bytes of code 729, prolog size 34 for method HttpResponseHeaders:CopyTo(byref):this
```

After
```asm
; Assembly listing for method HttpResponseHeaders:CopyTo(byref):this
; Lcl frame size = 40

G_M60302_IG01:
       push     rdi
       push     rsi
       sub      rsp, 40
       mov      rsi, rcx
       mov      rdi, rdx

G_M60302_IG02:
       mov      rcx, rsi
       mov      rdx, rdi
       call     HttpResponseHeaders:CopyToFast(byref):this
       mov      rcx, gword ptr [rsi+8]
       test     rcx, rcx
       je       SHORT G_M60302_IG04
       ;...
       
; Total bytes of code 71, prolog size 6 for method HttpResponseHeaders:CopyTo(byref):this
```

### DIFF
--- a/src/Servers/Kestrel/Core/src/Internal/Http/HttpResponseHeaders.cs
+++ b/src/Servers/Kestrel/Core/src/Internal/Http/HttpResponseHeaders.cs
@@ -35,10 +35,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             if (extraHeaders != null && extraHeaders.Count > 0)
             {
                 // Only reserve stack space for the enumartors if there are extra headers
-                CopyExtraHeaders(extraHeaders, ref buffer);
+                CopyExtraHeaders(ref buffer, extraHeaders);
             }
 
-            void CopyExtraHeaders(Dictionary<string, StringValues> headers, ref BufferWriter<PipeWriter> buf)
+            static void CopyExtraHeaders(ref BufferWriter<PipeWriter> buffer, Dictionary<string, StringValues> headers)
             {
                 foreach (var kv in headers)
                 {
@@ -46,10 +46,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                     {
                         if (value != null)
                         {
-                            buf.Write(_CrLf);
-                            buf.WriteAsciiNoValidation(kv.Key);
-                            buf.Write(_colonSpace);
-                            buf.WriteAsciiNoValidation(value);
+                            buffer.Write(_CrLf);
+                            buffer.WriteAsciiNoValidation(kv.Key);
+                            buffer.Write(_colonSpace);
+                            buffer.WriteAsciiNoValidation(value);
                         }
                     }
                 }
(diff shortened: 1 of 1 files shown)


src/Servers/Kestrel/Core/src/Internal/Http/HttpResponseHeaders.cs
```diff
@@ -30,19 +30,27 @@ public Enumerator GetEnumerator()
         internal void CopyTo(ref BufferWriter<PipeWriter> buffer)
         {
             CopyToFast(ref buffer);
-            if (MaybeUnknown != null)
+
+            var extraHeaders = MaybeUnknown;
```
jkotalik: Pardon my ignorance, but why is it better to check a local variable here rather than the property?

jkotalik (approved):
Seems fine besides a small comment.

src/Servers/Kestrel/Core/src/Internal/Http/HttpResponseHeaders.cs
benaadams: As we are using it 3 times (`is null`, `.Count`, `.GetEnumerator`) while the Jit *may* read it once into a register; its just to encourage that to happen and it use a CPU register rather than main memory for the reference.

We then pass it as a parameter to `CopyExtraHeaders` and make that method static, so it keeps it in a register rather than looking it up again from main memory (e.g. `this.MaybeUnknown`).

benaadams: Note if there was a lot of stack and registers in use by the method (as there were previously) then this wouldn't be an optimization as it would just copy the value to stack instead (as there would be no registers free)

benaadams:
Rebased

benaadams:
CI issues https://github.com/aspnet/AspNetCore/issues/7839 and https://github.com/aspnet/AspNetCore/issues/7867

benaadams:
Rebased

benaadams:
Moved new method to local function (also to retrigger CI)

anurse:
Build failure seems to be unrelated (still working on knocking those down ;)). Filed https://github.com/aspnet/AspNetCore/issues/8126. I'll try a re-spin. I presume you rebased recently @benaadams? Looks like it from the GitHub history.

benaadams:
>  I presume you rebased recently

Yeah , this was the rebase on current

benaadams:
Might have to retrigger Windows x64 manually; doesn't seem to want to restart via the pipeline comment?

anurse:
Looks like it *is* re-running, according to Azure Pipelines, despite GitHub's check UI not updating.

benaadams:
Rebased

JamesNK (approved):

anurse:
@jkotalik if this is good to go, let's merge it this morning for preview 4.

PR #7725, opened 2019-02-19
Replace IHostingEnvironment with IWebHostEnvironment

Tratcher:
This is the second half of https://github.com/aspnet/Extensions/issues/966

IHostingEnvironment is replaced by IWebHostEnvironment
IApplicationEnvironment is replaced by Extensions.Hosting.IHostApplicationEnvironment

Due to the size and timing of this PR nits in samples or tests are unlikely to be addressed.

Also, there are quite a few public constructors that were modified. These are primarily invoked by DI and there is little end-user impact. I'm not keeping the old constructors as the old types are not convertible to the new types. Removing the old constructors seems preferable to obsoleting them with an error.


BrennanConroy (approved):

JunTaoLuo (approved):

JunTaoLuo:
:shipit:  When tests pass.

src/Hosting/Hosting/src/Internal/HostingEnvironmentExtensions.cs
```diff
@@ -9,7 +9,62 @@ namespace Microsoft.AspNetCore.Hosting.Internal
 {
     public static class HostingEnvironmentExtensions
     {
+#pragma warning disable CS0618 // Type or member is obsolete
```
davidfowl: I dunno why this was made public in the first place.

tebeco:
dumb question but why not align interface name ?
`IXxxHostEnvironment` or `IHostYyyEnvironment`

when people are going to find `IHostApplicationEnvironment` it will probably NOT be straightforward to think that you actually have to backspace back to `I` to re-type `Web`

that’s a really naive question there are probably tons of better reason than doing
`IWebHostEnvironment`
`IApplicationHostEnvironment`

or
`IHostWebEnvironnent`
`IHostApplicationEnvironment`

Tratcher:
The new types are:
`IHostEnvironment`
`IWebHostEnvironment`
`IHostApplicationLifetime`

`IHostWebEnvironment` doesn't convey any more meaning.

`IApplicationHostLifetime` sounds slightly better than `IHostApplicationLifetime`. @davidfowl @glennc?

Tratcher:
Announcement posted. https://github.com/aspnet/AspNetCore/issues/7749

PR #7730, opened 2019-02-20
Update Razor SDK and resolve issues

rynowak:

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <!--
 
   This file is used by automation to update Versions.props and may be used for other purposes, such as
@@ -9,21 +9,21 @@
 -->
 <Dependencies>
   <ProductDependencies>
-    <Dependency Name="Microsoft.AspNetCore.Razor.Language" Version="3.0.0-preview-19107-04">
+    <Dependency Name="Microsoft.AspNetCore.Razor.Language" Version="3.0.0-preview3.19119.5">
       <Uri>https://github.com/aspnet/AspNetCore-Tooling</Uri>
-      <Sha>408b123a6454841884c1d840fb853eec8de12c10</Sha>
+      <Sha>182b2c81a5a3c2d5219cf20205e49dbe86823e18</Sha>
     </Dependency>
-    <Dependency Name="Microsoft.AspNetCore.Mvc.Razor.Extensions" Version="3.0.0-preview-19107-04">
+    <Dependency Name="Microsoft.AspNetCore.Mvc.Razor.Extensions" Version="3.0.0-preview3.19119.5">
       <Uri>https://github.com/aspnet/AspNetCore-Tooling</Uri>
-      <Sha>408b123a6454841884c1d840fb853eec8de12c10</Sha>
+      <Sha>182b2c81a5a3c2d5219cf20205e49dbe86823e18</Sha>
     </Dependency>
-    <Dependency Name="Microsoft.CodeAnalysis.Razor" Version="3.0.0-preview-19107-04">
+    <Dependency Name="Microsoft.CodeAnalysis.Razor" Version="3.0.0-preview3.19119.5">
       <Uri>https://github.com/aspnet/AspNetCore-Tooling</Uri>
-      <Sha>408b123a6454841884c1d840fb853eec8de12c10</Sha>
+      <Sha>182b2c81a5a3c2d5219cf20205e49dbe86823e18</Sha>
     </Dependency>
-    <Dependency Name="Microsoft.NET.Sdk.Razor" Version="3.0.0-preview-19107-04">
+    <Dependency Name="Microsoft.NET.Sdk.Razor" Version="3.0.0-preview3.19119.5">
       <Uri>https://github.com/aspnet/AspNetCore-Tooling</Uri>
-      <Sha>408b123a6454841884c1d840fb853eec8de12c10</Sha>
+      <Sha>182b2c81a5a3c2d5219cf20205e49dbe86823e18</Sha>
     </Dependency>
     <Dependency Name="dotnet-ef" Version="3.0.0-preview.19080.1">
       <Uri>https://github.com/aspnet/EntityFrameworkCore</Uri>
--- a/src/Components/Blazor/Build/test/BindRazorIntegrationTest.cs
+++ b/src/Components/Blazor/Build/test/BindRazorIntegrationTest.cs
@@ -91,7 +91,7 @@ namespace Test
                 frames,
                 frame => AssertFrame.Component(frame, "Test.MyComponent", 3, 0),
                 frame => AssertFrame.Attribute(frame, "Value", 42, 1),
-                frame => AssertFrame.Attribute(frame, "ValueChanged", typeof(Action<UIEventArgs>), 2));
+                frame => AssertFrame.Attribute(frame, "ValueChanged", typeof(EventCallback<UIChangeEventArgs>), 2));
         }
 
         [Fact]
@@ -167,7 +167,7 @@ namespace Test
                 frames,
                 frame => AssertFrame.Component(frame, "Test.MyComponent", 3, 0),
                 frame => AssertFrame.Attribute(frame, "Value", 42, 1),
-                frame => AssertFrame.Attribute(frame, "OnChanged", typeof(Action<UIEventArgs>), 2));
+                frame => AssertFrame.Attribute(frame, "OnChanged", typeof(EventCallback<UIChangeEventArgs>), 2));
         }
 
         [Fact]
@@ -201,7 +201,7 @@ namespace Test
                 frames,
                 frame => AssertFrame.Element(frame, "div", 3, 0),
                 frame => AssertFrame.Attribute(frame, "myvalue", "hi", 1),
-                frame => AssertFrame.Attribute(frame, "myevent", typeof(Action<UIEventArgs>), 2));
+                frame => AssertFrame.Attribute(frame, "myevent", typeof(EventCallback), 2));
         }
 
         [Fact]
@@ -235,7 +235,7 @@ namespace Test
                 frames,
                 frame => AssertFrame.Element(frame, "div", 3, 0),
                 frame => AssertFrame.Attribute(frame, "myvalue", "hi", 1),
-                frame => AssertFrame.Attribute(frame, "myevent", typeof(Action<UIEventArgs>), 2));
+                frame => AssertFrame.Attribute(frame, "myevent", typeof(EventCallback), 2));
         }
 
         [Fact]
@@ -292,7 +292,7 @@ namespace Test
                 frames,
                 frame => AssertFrame.Element(frame, "input", 3, 0),
                 frame => AssertFrame.Attribute(frame, "value", "42", 1),
-                frame => AssertFrame.Attribute(frame, "onchange", typeof(Action<UIEventArgs>), 2));
+                frame => AssertFrame.Attribute(frame, "onchange", typeof(EventCallback), 2));
         }
 
         [Fact]
@@ -315,7 +315,7 @@ namespace Test
                 frame => AssertFrame.Element(frame, "input", 4, 0),
                 frame => AssertFrame.Attribute(frame, "type", "text", 1),
                 frame => AssertFrame.Attribute(frame, "value", new DateTime(2018, 1, 1).ToString("MM/dd/yyyy"), 2),
-                frame => AssertFrame.Attribute(frame, "onchange", typeof(Action<UIEventArgs>), 3));
+                frame => AssertFrame.Attribute(frame, "onchange", typeof(EventCallback), 3));
         }
 
         [Fact]
@@ -340,7 +340,7 @@ namespace Test
                 frame => AssertFrame.Element(frame, "input", 4, 0),
                 frame => AssertFrame.Attribute(frame, "type", "text", 1),
                 frame => AssertFrame.Attribute(frame, "value", new DateTime(2018, 1, 1).ToString("MM/dd/yyyy"), 2),
-                frame => AssertFrame.Attribute(frame, "onchange", typeof(Action<UIEventArgs>), 3));
+                frame => AssertFrame.Attribute(frame, "onchange", typeof(EventCallback), 3));
         }
 
         [Fact]
@@ -363,7 +363,7 @@ namespace Test
                 frame => AssertFrame.Element(frame, "input", 4, 0),
                 frame => AssertFrame.Attribute(frame, "type", "text", 1),
                 frame => AssertFrame.Attribute(frame, "value", "42", 2),
-                frame => AssertFrame.Attribute(frame, "onchange", typeof(Action<UIEventArgs>), 3));
+                frame => AssertFrame.Attribute(frame, "onchange", typeof(EventCallback), 3));
         }
 
         [Fact]
@@ -385,7 +385,7 @@ namespace Test
                 frames,
                 frame => AssertFrame.Element(frame, "input", 3, 0),
                 frame => AssertFrame.Attribute(frame, "type", "checkbox", 1),
-                frame => AssertFrame.Attribute(frame, "onchange", typeof(Action<UIEventArgs>), 3));
+                frame => AssertFrame.Attribute(frame, "onchange", typeof(EventCallback), 3));
         }
 
         [Fact]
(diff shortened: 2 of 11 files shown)


eng/Version.Details.xml
```diff
@@ -9,19 +9,19 @@
 -->
 <Dependencies>
   <ProductDependencies>
-    <Dependency Name="Microsoft.AspNetCore.Razor.Language" Version="3.0.0-preview-19107-04">
+    <Dependency Name="Microsoft.AspNetCore.Razor.Language" Version="3.0.0-preview3.19119.4">
       <Uri>https://github.com/aspnet/AspNetCore-Tooling</Uri>
       <Sha>408b123a6454841884c1d840fb853eec8de12c10</Sha>
```
rynowak: @natemcmaster - does it matter what I do with the SHA?

src/Components/Blazor/Build/test/BindRazorIntegrationTest.cs
```diff
@@ -201,7 +201,7 @@ public static class BindAttributes
                 frames,
                 frame => AssertFrame.Element(frame, "div", 3, 0),
                 frame => AssertFrame.Attribute(frame, "myvalue", "hi", 1),
-                frame => AssertFrame.Attribute(frame, "myevent", typeof(Action<UIEventArgs>), 2));
+                frame => AssertFrame.Attribute(frame, "myevent", typeof(EventCallback), 2));
```
rynowak: @SteveSandersonMS - the types that get placed here depend on whether the delegate has a relevant `.Target` or not and whether bind is applied to a component or element. We already discussed this in detail, but I wanted to point it out for reviewing purposes.

src/Components/Blazor/Build/test/RenderingRazorIntegrationTest.cs
```diff
@@ -365,19 +365,19 @@ public async Task SupportsTwoWayBindingForTextboxes()
             var renderer = new TestRenderer();
 
             // Assert
-            Action<UIEventArgs> setter = null;
+            EventCallback setter = default;
```
rynowak: Same deal. The thing we actually put in the frame depends on the details.

src/Components/Components/src/EventCallbackFactory.cs
```diff
@@ -135,6 +135,23 @@ public string Create<T>(object receiver, string callback)
...
         }
 
+        /// <summary>
+        /// Returns the provided <paramref name="callback"/>. For internal framework use only.
+        /// </summary>
+        /// <param name="receiver"></param>
+        /// <param name="callback"></param>
+        /// <returns></returns>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public EventCallback<T> Create<T>(object receiver, EventCallback callback)
```
rynowak: Found which porting this. This allows you to assign an `EventCallback` to an `EventCallback<T>`. This is useful if you want to pass an eventcallback through 3 layers of components - and yes we had a test that hit this 😆

src/Components/Components/src/RuntimeHelpers.cs
```diff
@@ -17,5 +20,23 @@ public static class RuntimeHelpers
...
+
+        /// <summary>
+        /// Not intended for use by application code.
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="value"></param>
+        /// <returns></returns>
+        public static T TypeCheck<T>(Func<Task> value) => default(T);
+
+        public static T TypeCheck<T, U>(Action<U> value) => default(T);
```
rynowak: Needs docs.

eng/Version.Details.xml
natemcmaster: Yes. The bot and some of our scripts use this. If it gets out of sync, it can cause problems. Use `darc` to change these files on your own.

```
darc update-dependencies --channel '.NET Core 3 Release' --source-repo AspNetCore-Tooling
```

src/Components/Components/src/RuntimeHelpers.cs
```diff
@@ -17,5 +20,23 @@ public static class RuntimeHelpers
...
         /// <returns></returns>
         public static T TypeCheck<T>(T value) => value;
+
+        /// <summary>
+        /// Not intended for use by application code.
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="value"></param>
+        /// <returns></returns>
+        public static T TypeCheck<T>(Action value) => default(T);
```
rynowak: OK. Time for explanation - there's a problem here that we won't be able to fix in preview 3 (tooling only). This only concerns what happens in the editor, since that code is not getting updated.

Suppose you have the following component:

```html
@* MyButton.razor *@

<button onclick=@OnClick>Hi!</button>
@functions {
    [Parameter] EventCallback<UIMouseEventArgs> OnClick { get; set; }
}
```

and usage:
```html
<MyButton OnClick="@(() => { ... }) />
```

When the **OLD** compiler that runs in VS see this usage, it will generate code like:
```C#
builder.OpenComponent<MyButton>(0);
builder.AddAttribute(1, "OnClick", RuntimeHelpers.TypeCheck<EventCallback<UIMouseEventArgs>>(() => { ... }));
builder.CloseComponent();
```

The **OLD** compiler that runs in VS doesn't know about `EventCallback<>` and thus doesn't know how to generate a call to the `EventCallback.Factory.Create<>(...)` method, it does the old thing when you're editing. You might see a problem with this, because an `Action` is not an `EventCallback<UIMouseEventArgs>`.

Well, this is the reason why I added `TypeCheck<T>(Action value)` - it's a hack for preview 3 that will subvert the type check if you're using an `Action`. This will apply to a bunch of cases as well, so basically a bunch of code that wouldn't work at runtime will now compile. (I plan to update this to do a cast, so it will fail at runtime if needed).

I did a similar thing for `Func<Task>`. 

**now again** - I want to reiterate that this is only a problem for the editor, when means you get error squigglies. Everything works great when you build because the compiler is updated.

----

Now we get to the real problem. We would want to make a case like `Action<UIMouseEventArgs>` or `Func<UIMouseEventArgs, Task>` work as well. Well now we have a problem,

The compiler wrote:
```C#
RuntimeHelpers.TypeCheck<EventCallback<UIMouseEventArgs>>(() => { ... })
```

We want to define a method that accepts an `Action<U>` where `U` is substituted for `UIMouseEventArgs` in `EventCallback<UIMouseEventArgs>`.... and we're stuck.

This won't work because we can't change the method to `TypeCheck<T, U>`. C# doesn't allow you to partially specify type-arguments. We can't change what the compiler generates because that's in VS.

---

**TLDR;** So to reiterate, in the case you have a component that accepts an `EventCallback<T>` parameter and you want set that parameter to an `Action<T>` or `Func<T, Task>` this will cause a spurious error in the editor. 

Using `bind-...` on a component, or doing anything to an element has no problem because it doesn't do the same type checking.

src/Components/test/testassets/BasicTestApp/CounterComponent.cshtml
```diff
@@ -1,6 +1,6 @@
 <h1>Counter</h1>
 <p>Current count: @currentCount</p>
-<p><button onclick="@((handleClicks ? (Action)IncrementCount : null))">Click me</button></p>
+<p><button onclick="@((handleClicks ? (Action)IncrementCount : () => { }))">Click me</button></p>
```
rynowak: @SteveSandersonMS - something else to discuss here - this was failing because `EventCallback.Factory.Create` doesn't accept a null delegate. What do you think should be the behaviour of a null-event-handler, should it trigger `HandleEventAsync`?

src/Components/Components/src/RuntimeHelpers.cs
rynowak: I'm going to try implicit conversion as well and see if that helps. I'm not sure you're allowed an implicit conversion from a delegate type.

rynowak:
There's currently one other broken test that I'm investigating beyond what I mentioned here.

eng/Version.Details.xml
rynowak: Done.

src/Components/test/testassets/BasicTestApp/CascadingValueTest/CascadingValueReceiveFixedByInterface.cshtml
```diff
@@ -11,7 +11,9 @@
 <p>
     @(nameof(CascadingValueReceiveFixedByInterface)) render count:
     <strong id="receive-by-interface-num-renders">@numRenders</strong>
-    <button id="decrement-count" onclick=@Ancestor.DecrementCount>Decrement</button>
+
+    @* Prevent rerendering this component when the button is clicked. *@
+    <button id="decrement-count" onclick=@(EventCallback.Factory.Create(Ancestor, Ancestor.DecrementCount))>Decrement</button>
```
rynowak: I'm guessing this is the best thing to do based on the original intent of the request. Our new and improved logic makes you use workarounds if you really want to prevent rendering.

dougbu (commented):
eng/ files look fine 😸

src/Components/Blazor/Build/test/RenderingRazorIntegrationTest.cs
```diff
@@ -365,19 +365,19 @@ public async Task SupportsTwoWayBindingForTextboxes()
...
                 frame =>
                 {
                     AssertFrame.Attribute(frame, "onchange", 2);
-                    setter = Assert.IsType<Action<UIEventArgs>>(frame.AttributeValue);
+                    setter = Assert.IsType<EventCallback>(frame.AttributeValue);
                 });
 
             // Trigger the change event to show it updates the property
-            await renderer.Invoke(() => setter(new UIChangeEventArgs { Value = "Modified value", }));
+            await renderer.InvokeAsync(() => setter.InvokeAsync(new UIChangeEventArgs { Value = "Modified value", }));
```
SteveSandersonMS: Is the new `await` for convenience, or is the synchronous outcome actually different?

Assuming we haven't changed the event handling to now only take effect synchronously, it would be good to have at least one test for binding that is really explicit about what happens synchronously vs asynchronously (if there isn't already one).

src/Components/test/testassets/BasicTestApp/CascadingValueTest/CascadingValueReceiveFixedByInterface.cshtml
SteveSandersonMS: Yes, OK. I guess it's a bit of an odd test now it's covering a scenario that wouldn't normally exist.

It could be expressed more clearly by not trying to stop the rendering, but instead having another component that is also wired to the same cascading value, and observing that it re-renders when the cascading value changed.

Assuming you don't want to refactor the test (though please go ahead if you do want!), perhaps the commend could be tweaked to the following to keep a record of the intent:

```
@* Prevent the usual rerendering this component when the button is clicked, so that the test code can observe it re-renders because the cascading value changes. *@
```

src/Components/test/testassets/BasicTestApp/CounterComponent.cshtml
SteveSandersonMS: Hmm, I'd imagine it would be common to supply a null value for an event handler if you're using delegate types instead of `EventCallback`, e.g.:

```
<button onclick=@DoSomething>Click me</button>
@functions {
    [Parameter] Action DoSomething { get; set; }
}
```

In this case, it's inconvenient to deal with absence of a value for the `DoSomething` parameter (you'd either wrap `DoSomething` in your own method, or use `onclick=@(() => DoSomething?.Invoke())`, which is pretty laborious and nonobvious to non-C#-experts). Previously it was a no-op if the value was null, but now it's an error. I think the no-op is maybe more useful.

Do you think it's enough to tell people "remember to use `EventCallback`, or if you don't, you are responsible for null checking"?

src/Components/test/testassets/BasicTestApp/CascadingValueTest/CascadingValueReceiveFixedByInterface.cshtml
SteveSandersonMS: Hang on, I'm confused. I thought that `onclick=@Ancestor.DecrementCount` would be exactly equivalent to `onclick=@(EventCallback.Factory.Create(Ancestor, Ancestor.DecrementCount))`, since `Ancestor.DecrementCount.Target == Ancestor`, right? Both formats would result in `Ancestor` being the recipient of the event, wouldn't they?

I must be misunderstanding, but could you clarify why it's different?

src/Components/Components/src/RuntimeHelpers.cs
SteveSandersonMS: > So to reiterate, in the case you have a component that accepts an `EventCallback<T>` parameter and you want set that parameter to an `Action<T>` or `Func<T, Task>` this will cause a spurious error in the editor.

That's unfortunate. So will have to be a known issue for the preview 3 release, then?

rynowak: > That's unfortunate. So will have to be a known issue for the preview 3 release, then?

I don't see an alternative unless people have other suggestions. We can't change what the editor does without updating VS, and we can't update VS. We are limited to fixes that change the runtime. 

BTW the workaround for the user is to use `EventCallback.Factory.Create` for this case - or just ignore the error from the editor because everything works in the actual build.

src/Components/test/testassets/BasicTestApp/CascadingValueTest/CascadingValueReceiveFixedByInterface.cshtml
rynowak: Because of the new compiler feature, this will generate:

```
EventCallback.Factory.Create(this, Ancestor.DecrementCount) 
```

In this case `this` will win out over the delegate target. If you expected different lets dicuss this in detail.

SteveSandersonMS: Ah, I see. In our earlier iterations of this new event dispatching technique, we looked at the supplied `delegate.Target`, and if it was `IHandleEvent`, we let it win over `this`.

If our objective in general is "the event goes to whichever component declares the handler", then I would have expected this one to go to `Ancestor`. I'm trying to think of scenarios where you'd reference a methodgroup on another component instance and *not* expect that other component to receive the event, but can't think of any. Can you think of any cases where you'd want `this` to win over `delegate.Target` when `delegate.Target` is itself a component? If there are any such cases then the existing logic is probably right.

This is quite an edge case, so even if we wanted to change this, I think it would be totally OK to do it as a follow-up that goes in after preview 3.

src/Components/Components/src/RuntimeHelpers.cs
SteveSandersonMS: Understood. We can't update everything simultaneously. So I guess it's a known issue when we ship, but will go away when people update to VS 16 Preview 4?

rynowak: no, VS is done at this point. It has to go into 16.1

natemcmaster (approved):
Signing off on the eng/ stuff only. Didn't review the product changes.

src/Components/test/testassets/BasicTestApp/CounterComponent.cshtml
rynowak: Yeah that's what I wondered as well. I'll fix this and add some tests.

src/Components/test/testassets/BasicTestApp/CascadingValueTest/CascadingValueReceiveFixedByInterface.cshtml
rynowak: Ok, let's do your version of this for now. I think you're right and I'd rather ship in preview3 the best version of this.

src/Components/Blazor/Build/test/RenderingRazorIntegrationTest.cs
rynowak: It's for thoroughness in tests. I'll update these to verify that it completed synchronously.

SteveSandersonMS (approved):
We discussed the remaining changes.

src/Components/Components/src/RuntimeHelpers.cs
rynowak: After discussion of this, the plan is to add more special-casing for overloads of  `TypeCheck` so that the editing experience is nice.

src/Components/test/testassets/BasicTestApp/FormsTest/TypicalValidationComponent.cshtml
```diff
@@ -1,42 +1,42 @@
 @using System.ComponentModel.DataAnnotations
 @using Microsoft.AspNetCore.Components.Forms
 
-<EditForm Model="@person" OnValidSubmit="@(EventCallback.Factory.Create<EditContext>(this, HandleValidSubmit))">
+<EditForm Model="@person" OnValidSubmit="@HandleValidSubmit">
```
SteveSandersonMS: 😍

src/Components/test/testassets/BasicTestApp/FormsTest/TypicalValidationComponent.cshtml
```diff
@@ -1,42 +1,42 @@
...
     <DataAnnotationsValidator />
 
     <p class="name">
-        Name: <InputText bind-Value="@person.Name" ValueExpression="@(() => person.Name)" />
+        Name: <InputText bind-Value="@person.Name" />
```
SteveSandersonMS: Thanks for updating these!

rynowak:
Now the hard part. Getting the CI to pass.

PR #7753, opened 2019-02-20
Follow-up to VS2019 conversion

natemcmaster:
* Set VS version in file to 16.0.0.0
* Fix for restore failure on Microsoft.AspNetCore.App.Runtime.csproj


Tratcher (approved):

Tratcher:
16.0.28606.0 also works.

natemcmaster:
Let's use 16.0.0.0 to avoid blocking people if earlier VS previews are still working fine for them. We strongly encourage updating anyways.

Issue #5077, opened 2018-12-14
Token replacement in @page directive?


metoule:
I'm trying to create a custom route for my Razor page, via the `@page` directive.
When using the `RouteAttribute` on a controller, we have access to the following [token replacements](https://docs.microsoft.com/en-us/aspnet/core/mvc/controllers/routing?view=aspnetcore-2.1#token-replacement-in-route-templates-controller-action-area): 

> The tokens [action], [area], and [controller] will be replaced with the values of the action name, area name, and controller name from the action where the route is defined.

However, such replacement doesn't seem to occur with the `@page` directive:

`@page "/[area]/{name}"` 

doesn't work.

Is there anything approaching this for Razor pages?

__Background__

I'm trying to migrate the following controller to Razor pages:
```
[Route("[controller]")]
public class MyFeatureController : Controller
{
	[HttpGet("")]
	public IActionResult Index()
	{
		return View();
	}
	
	[HttpGet("{name}")]
	public IActionResult GetProduct(string name)
	{
		return View("Product", name);
	}
}
```

NTaylorMullen:
> @NTaylorMullen I would like to try and tackle this one. Is there anything "not so obvious" I should know before I dig in?

Sounds like a great plan, thanks @AlexejTimonin, I haven't looked at this area in quite a while, @pranavkm anything "not so obvious" you can think of?

pranavkm:
I think there you have to be careful around [outbound parameter transformation](https://github.com/aspnet/AspNetCore/blob/master/src/Mvc/src/Microsoft.AspNetCore.Mvc.RazorPages/Infrastructure/PageActionDescriptorProvider.cs#L113-L140) to ensure replacement doesn't happen twice. Beside that, it should likely be fairly straightforward. The functional tests should ensure your changes work with both flavors of routing.

metoule:
That seemed pretty straightforward, so closing it comes as a surprise. Do you think you can provide a bit more info about why it doesn't align with your long term plans?

PR #7757, opened 2019-02-20
Token replacement in page directive

AlexejTimonin:
Add functionality to replace tokens.

There is more tests left to write. But before I dig in any further I want
to be sure that I'm on the right track here with the token replacement.
@pranavkm how does the implementation look like?

Addresses #5077

### DIFF
--- a/src/Mvc/Mvc.RazorPages/src/Infrastructure/PageActionDescriptorProvider.cs
+++ b/src/Mvc/Mvc.RazorPages/src/Infrastructure/PageActionDescriptorProvider.cs
@@ -145,7 +145,7 @@ namespace Microsoft.AspNetCore.Mvc.RazorPages.Infrastructure
                 template = AttributeRouteModel.ReplaceTokens(
                     template, 
                     descriptor.RouteValues, 
-                    model.RouteParameterTransformer).Replace("//", "/");
+                    model.RouteParameterTransformer);
             }
 
             return AttributeRouteModel.CombineTemplates(pageRoute, template);
--- a/src/Mvc/Mvc.RazorPages/src/Infrastructure/PageActionDescriptorProvider.cs
+++ b/src/Mvc/Mvc.RazorPages/src/Infrastructure/PageActionDescriptorProvider.cs
@@ -124,7 +124,16 @@ namespace Microsoft.AspNetCore.Mvc.RazorPages.Infrastructure
                 // Selector does not have expected metadata
                 // This selector was likely configured by AddPageRouteModelConvention
                 // Use the existing explicitly configured template
-                return selectorModel.AttributeRouteModel.Template;
+                if (selectorModel.AttributeRouteModel.Template != null)
+                {
+                    return AttributeRouteModel.ReplaceTokens(
+                        selectorModel.AttributeRouteModel.Template,
+                        model.RouteValues);
+                }
+                else
+                {
+                    return selectorModel.AttributeRouteModel.Template;
+                }
             }
 
             var pageRoute = pageRouteMetadata.PageRoute;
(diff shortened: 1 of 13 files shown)


pranavkm:
Sorry, I missed your earlier PR. I'll have a look

src/Mvc/Mvc.RazorPages/src/Infrastructure/PageActionDescriptorProvider.cs
```diff
@@ -132,16 +127,28 @@ private static string TransformPageRoute(PageRouteModel model, SelectorModel sel
                 return selectorModel.AttributeRouteModel.Template;
```
pranavkm: Should it work in this case?

AlexejTimonin:
> 
> 
> Sorry, I missed your earlier PR. I'll have a look

Nothing to be sorry about, no harm done :) Thank you for taking your time!

src/Mvc/Mvc.RazorPages/src/Infrastructure/PageActionDescriptorProvider.cs
```diff
@@ -132,16 +127,28 @@ private static string TransformPageRoute(PageRouteModel model, SelectorModel sel
...
             }
 
-            var transformedPageRoute = string.Join("/", segments);
+            var template = pageRouteMetadata.RouteTemplate;
+            if (template != null)
+            { 
+                template = AttributeRouteModel.ReplaceTokens(
+                    template, 
+                    descriptor.RouteValues, 
+                    model.RouteParameterTransformer).Replace("//", "/");
```
pranavkm: Why is the `Replace` needed?

pranavkm (commented):
Looks neat. cc @rynowak

src/Mvc/Mvc.RazorPages/src/Infrastructure/PageActionDescriptorProvider.cs
AlexejTimonin: The page value in route dictionary contains "/" prefix. This does not play well with `AttributeRouteModel.ReplaceTokens(...)` because the internals of the method grabs the token, retrieves the value from route dictionary and appends it to transformed string. 

For example given route values:
```
{
    "area" : "admin",
    "page": "/myPage"
}
```
and template `[area]/[page]` the `AttributeRouteModel.ReplaceTokens(...)` method will return `admin//myPage`

Would a comment be enough in this case? Or should I investigate on how to achieve it inside `AttributeRouteModel.ReplaceTokens(...)`, maybe even create a new method just for page tokens because the method has no way of knowing if route values are for controller or page.

pranavkm: Might be worthwhile to consider changing the existing `ReplaceTokens` method. I think it would be ok to special case slashes to address this one case.

AlexejTimonin:
Sorry for review requested spam, rebased my 30 days old branch.

src/Mvc/Mvc.RazorPages/src/Infrastructure/PageActionDescriptorProvider.cs
AlexejTimonin: Probably should, it would enable devs to create templates with tokens in their custom configured routing model for specific page. I will dig into it :)

AlexejTimonin:
@dotnet-bot test this please

dougbu:
@AlexejTimonin who do you think _actually_ should be reviewing this PR. Just @pranavkm and @rynowak?

AlexejTimonin:
> @AlexejTimonin who do you think _actually_ should be reviewing this PR. Just @pranavkm and @rynowak?

@dougbu Yes, sorry for inconvenience.

PR #7762, opened 2019-02-20
Make Client Proxies an abstract class

halter73:
Fixes https://github.com/aspnet/AspNetCore/issues/5382

### DIFF
--- a/src/SignalR/server/Core/src/ClientProxyExtensions.cs
+++ b/src/SignalR/server/Core/src/ClientProxyExtensions.cs
@@ -8,73 +8,73 @@ using System.Threading.Tasks;
 namespace Microsoft.AspNetCore.SignalR
 {
     /// <summary>
-    /// Extension methods for <see cref="IClientProxy"/>.
+    /// Extension methods for <see cref="ClientProxy"/>.
     /// </summary>
     public static class ClientProxyExtensions
     {
         /// <summary>
-        /// Invokes a method on the connection(s) represented by the <see cref="IClientProxy"/> instance.
+        /// Invokes a method on the connection(s) represented by the <see cref="ClientProxy"/> instance.
         /// Does not wait for a response from the receiver.
         /// </summary>
-        /// <param name="clientProxy">The <see cref="IClientProxy"/></param>
+        /// <param name="clientProxy">The <see cref="ClientProxy"/></param>
         /// <param name="method">The name of the method to invoke.</param>
         /// <param name="cancellationToken">The token to monitor for cancellation requests. The default value is <see cref="CancellationToken.None" />.</param>
         /// <returns>A <see cref="Task"/> that represents the asynchronous invoke.</returns>
-        public static Task SendAsync(this IClientProxy clientProxy, string method, CancellationToken cancellationToken = default)
+        public static Task SendAsync(this ClientProxy clientProxy, string method, CancellationToken cancellationToken = default)
         {
             return clientProxy.SendCoreAsync(method, Array.Empty<object>(), cancellationToken);
         }
 
         /// <summary>
-        /// Invokes a method on the connection(s) represented by the <see cref="IClientProxy"/> instance.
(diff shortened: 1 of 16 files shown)


mikaelm12 (approved):

mikaelm12:
Was getting some build errors before about not implementing abstract members but it seemed like you were overriding them correctly. Let's see what the new build says

mikaelm12:
Obsolete is causing failures in components now.
ex `Server/src/Circuits/RemoteRenderer.cs(25,26): error CS0618: 'IClientProxy' is obsolete: 'Use the ClientProxy abstract class instead.'`

src/SignalR/server/Core/src/IClientProxy.cs
```diff
@@ -9,6 +10,7 @@ namespace Microsoft.AspNetCore.SignalR
     /// <summary>
     /// A proxy abstraction for invoking hub methods.
     /// </summary>
+    [Obsolete("Use the ClientProxy abstract class instead.")]
```
davidfowl: This isn’t right. There’s no need to obsolete this, let’s just leave it.

src/SignalR/server/Core/src/ClientProxyExtensions.cs
```diff
@@ -8,90 +8,90 @@
...
+        /// Invokes a method on the connection(s) represented by the <see cref="ClientProxy"/> instance.
         /// Does not wait for a response from the receiver.
         /// </summary>
-        /// <param name="clientProxy">The <see cref="IClientProxy"/></param>
+        /// <param name="clientProxy">The <see cref="ClientProxy"/></param>
         /// <param name="method">The name of the method to invoke.</param>
         /// <param name="cancellationToken">The token to monitor for cancellation requests. The default value is <see cref="CancellationToken.None" />.</param>
         /// <returns>A <see cref="Task"/> that represents the asynchronous invoke.</returns>
-        public static Task SendAsync(this IClientProxy clientProxy, string method, CancellationToken cancellationToken = default)
+        public static Task SendAsync(this ClientProxy clientProxy, string method, CancellationToken cancellationToken = default)
```
halter73: REVIEW: Even though this PR now keeps IClientProxy around, parts of the change like this still make it technically breaking. Since the only real IClientProxy implementation that i know of is ClientProxy, this should only break tests using mocks.

We could keep the IClientProxy extension methods, but if we end up updating the abstract base class we won't be able to use the new members in the IClientProxy extension methods. I think it's better to rip off the bandaid right way and make the breaking change now.

BrennanConroy: Take a look at https://github.com/aspnet/AspNetCore/pull/7395/files/1ba0c149e2093be5c953f8ed129adab503ce2420#diff-e04c9b7b5f6ed4cef369701a5b03ff2c

halter73: That's true. Any IClientProxy decorators would also break.

davidfowl: I'm not sure we need to break any of these things right now.

halter73: The point of this PR is to make any necessary breaking changes in a major release. If we don't change this now, we risk needing to break in a future minor release if any of these extension methods need any new members/overloads added to ClientProxy.

At that point, I don't think it would be a simple as adding new ClientProxy-based extension methods with the same names. This is not only because it breaks binary compatibility, but also because the compiler wouldn't know whether to choose the IClientProxy or ClientProxy extension methods. We'd have to remove the IClientProxy extension methods at that point, or (if we are really serious about not breaking anyone) deprecate and burn all the old extension method names and come up with new ones.

halter73:
Talked to @davidfowl offline. Turns out we want to transition `IHubClients<T>` to an abstract base class, not `IClientProxy`.

PR #7763, opened 2019-02-20
Reenable some helix tests

HaoK:
Publishes the src files the tests is using so they get included as part of the helix payload

@ajcvickers @natemcmaster

### DIFF
--- a/src/Identity/test/Identity.Test/IdentityUIScriptsTest.cs
+++ b/src/Identity/test/Identity.Test/IdentityUIScriptsTest.cs
@@ -38,8 +38,7 @@ namespace Microsoft.AspNetCore.Identity.Test
             }
         }
 
-        [ConditionalTheory]
-        [SkipOnHelix] // https://github.com/aspnet/AspNetCore/issues/6549
+        [Theory]
         [MemberData(nameof(ScriptWithIntegrityData))]
         public async Task IdentityUI_ScriptTags_SubresourceIntegrityCheck(ScriptTag scriptTag)
         {
@@ -68,8 +67,7 @@ namespace Microsoft.AspNetCore.Identity.Test
             }
         }
 
-        [ConditionalTheory]
-        [SkipOnHelix] // https://github.com/aspnet/AspNetCore/issues/6549
+        [Theory]
         [MemberData(nameof(ScriptWithFallbackSrcData))]
         public async Task IdentityUI_ScriptTags_FallbackSourceContent_Matches_CDNContent(ScriptTag scriptTag)
         {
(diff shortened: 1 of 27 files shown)


dougbu (approved):
LGTM but please get Helix passing before you submit

HaoK:
I think the dotnet runtime version upgrade (not part of this PR) might be causing some of the failures (it def exposed the issue with the env vars being setup wrong in the runtests.cmd)

HaoK:
@dougbu can you take a quick look at a similar fix applied to the mvc analyzer test projects?

dougbu:
> @dougbu can you take a quick look at a similar fix applied to the mvc analyzer test projects?

Are you asking for a review of a PR? If so, yes. If you need something else, I don't have much time at the moment.

HaoK:
Yes just rereview the PR, I added some changes to the MVC api test projects similar to the identity UI project, its just an additional publish + helix branch around where to find the test files

dougbu (approved):
Still looks good to me though CI builds and tests disagree.

HaoK:
@pakrym looks like the logging test errors are consistent on these tests now, I guess I will skip them for now, if you wanted to take a look at the logged tests that are failing with Null Ref, its these:

https://mc.dot.net/#/user/aspnetcore/pr~2Faspnet~2Faspnetcore/ci/20190221.68

HaoK:
@pakrym finally things are looking good again with your logging fix, I put it in Directory.Build.targets for now since I don't have the rest of your refactoring in this PR.

This is the best run I've seen in a while (passing ci and windows tests are all green in helix-test) so I'm thinking this is a candidate for squash/merge

src/Identity/test/Identity.Test/Microsoft.AspNetCore.Identity.Test.csproj
```diff
@@ -7,6 +7,8 @@
   <ItemGroup>
     <Compile Include="$(IdentityTestSharedSourceRoot)**\*.cs" />
     <Compile Include="$(SharedSourceRoot)test\SkipOnHelixAttribute.cs" />
+    <Content Include="..\..\UI\src\Areas\Identity\Pages\**\*.*" CopyToPublishDirectory="PreserveNewest" LinkBase="UI\src\Areas\Identity\Pages" />
```
pakrym: Use HelixContent item group?

src/Identity/test/Identity.Test/IdentityUIScriptsTest.cs
```diff
@@ -148,13 +146,17 @@ private static List<ScriptTag> GetScriptTags(string cshtmlFile)
         private static string GetSolutionDir()
         {
             var dir = new DirectoryInfo(AppContext.BaseDirectory);
-            while (dir != null)
+            // On helix we use the published copy
+            if (!string.Equals(Environment.GetEnvironmentVariable("helix"), "true", StringComparison.OrdinalIgnoreCase))
```
pakrym: I would rather prefer we had a shared way of doing this (looking for the test base directory). Rather then copying the snippet around.

pakrym (approved):
Pending nits.

src/Identity/test/Identity.Test/IdentityUIScriptsTest.cs
HaoK: Do you mean add a test helper method for this that both the helix skip attribute and actual tests can use to check if we are running on helix?

pakrym: Add a way to find the test asset root. Or at minimum to detect running in helix.

HaoK: I don't feel comfortable generalizing a single test, I can add the helix detection now since the HelixSkip has the same logic.  Once I fix a few more tests and see a clear pattern we can look into a generalized shared GetSolutionDir.

HaoK: @pakrym added a static helper method for now which encapsulates the Environment helix check

Issue #7765, opened 2019-02-20
Samples do not work with IIS InProcess


Tratcher:
Since InProcess became the default mode many of the samples in this repo have stopped working with `500.0 - ANCM In-Process Handler Load Failure`. Adding `<AspNetCoreHostingModel>OutOfProccess</AspNetCoreHostingModel>` to their csproj makes them work again.

Some of these projects don't have a reference to `Microsoft.AspNetCore.Server.IIS` nor call `UseIIS()`, but adding those is not enough.

The NativeIISSample is the only one I've seen that still works and that's because it heavily customized the launch settings.

Ideally any customizations needed to make InProcess work in this repo would be consolidated to some shared source dependency we could reference from all of the samples.

@pakrym @jkotalik

Tratcher:
Some reference samples:
- Security.sln' CookieSample
- Middleware.sln's SessionSample, StaticFileSample

When fixing up the dependencies most samples can move from individual references to pulling in the Microsoft.AspNetCore metapackage as there's no more repo layering to worry about. I've already done that for some of the Security samples.

Eilon:
Let's make the samples use out-of-proc because it's the simplest option.

jkotalik:
Boooooooo!

jkotalik:
I don't believe it would be much work at all. It _should_ just be adding an additional props file to each sample.

Eilon:
@jkotalik - from my understanding during triage it was quite a bit more work to get the native assets to work as well. But either way, not as simple as using in-proc, and for no clear gain.

PR #7767, opened 2019-02-20
Fix up Security samples

Tratcher:
#7765 None of these samples have been working since InProcess became the default model. This mitigates the issue until we can come up with a standard solution.


jkotalik:
How long have the samples not been working? Was this due to us taking in a new websdk?

Anyways, can we just fix samples everywhere with this PR?

Tratcher:
Not sure on the timing, I only noticed it this week when updating them.

This is only a mitigation for samples I commonly use during development, I'm not pushing it to all samples. #7765 covers the real fix for all samples.

pakrym (approved):

PR #7776, opened 2019-02-21
Create Linux installers for the aspnetcore targeting pack

natemcmaster:
Resolves https://github.com/aspnet/AspNetCore/issues/6501

This creates 3 new build outputs:

* aspnetcore-targeting-pack-$(version).tar.gz
* aspnetcore-targeting-pack-$(version)-linux-x64.deb
* aspnetcore-targeting-pack-$(version)-linux-x64.rpm

Other changes:
* Make RPM packaging consistent with other installers. Vendor == Microsoft Corporation (cc @Eilon)
* Add shared properties for building the targeting pack (or not building it in a servicing build)

cc @dagood @nguerrera @dsplaisted


natemcmaster:
@Eilon - are we in 'ask mode' for preview3? I've retargeted this PR to release/3.0-preview3 b/c I believe this is required to complete the SDK's RPM/DEB story for Preview 3

dagood:
> I believe this is required to complete the SDK's RPM/DEB story for Preview 3

We don't have Linux or MacOS installers for the .NET Core targeting pack for preview 3, and it's not in progress yet for 4. I have been working with the understanding that there isn't anything blocked on these until 3.1 comes along--similar to the targeting pack bundle Windows installers (nice EXE installers wrapping the MSIs).

@nguerrera @dsplaisted can you help clear this up?

dsplaisted:
@dagood @natemcmaster  No need to rush to get these installers for Preview 3 at this point.  We do need them for 3.0, and would like them for preview 4.

The reason we need these is that the SDK package needs to depend on them, which is analogous to bundling the MSIs in the Windows installer.

natemcmaster:
Ok, thanks for clarification. I've retargeted this PR to master (preview4) since we missed the cutoff for P3.

eng/scripts/KillProcesses.ps1
```diff
@@ -21,6 +21,10 @@ _kill git.exe
 _kill vctip.exe
 _kill chrome.exe
 _kill h2spec.exe
-iisreset /restart
+_kill WerFault.exe
+if (Get-Command iisreset -ErrorAction ignore) {
```
dougbu: Good idea. Fortunately, the command only writes to `stdout` when process lacks admin rights.

dougbu (changes requested):
Many of the renamed files are losing their `exec` bit. I doubt that was intentional. If I'm correct, please do the `git update-index --chmod=-x {path}` fun.

Otherwise looks good. I didn't delve into all of the Debian or RPM-specific updates, assuming they came from somewhere else e.g. recommendations. If there's something in there that needs detailed review, please let me know.

natemcmaster:
> Many of the renamed files are losing their `exec` bit.

Intentional for some of the files, like the plaintext ones. I've restored +x permissions on .py and .bats files.

Changes to existing rpm/deb code was done to share common targets/props between multiple .deb and .rpm projects. They mostly build the same way -- take a folder of stuff and put it in a package, but with slightly different names and dependencies.

Still investigating why build fails on CI but passes locally. Added more binlog output so I can investigate

natemcmaster:
:up: :date: figure out the problem was that AspNeCore.App.Ref.csproj was not getting built. IsPackable was set to 'false' due to new conventions about ref/ folders (cc @pakrym ) CI builds are looking good now.

src/Installers/Debian/Directory.Build.props
```diff
@@ -0,0 +1,23 @@
+<Project>
+  <PropertyGroup>
+    <OutputInRepoRoot>true</OutputInRepoRoot>
+  </PropertyGroup>
+
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory)..\, Directory.Build.props))\Directory.Build.props" />
+
+  <UsingTask Condition=" '$(BuildToolsTaskAssembly)' != '' AND Exists('$(BuildToolsTaskAssembly)') " TaskName="Microsoft.AspNetCore.BuildTools.GenerateFileFromTemplate" AssemblyFile="$(BuildToolsTaskAssembly)" />
```
pakrym: Why isn't this defined in some common location?

src/Installers/Debian/Runtime/Debian.Runtime.debproj
```diff
@@ -0,0 +1,56 @@
...
+    </ProjectReference>
+  </ItemGroup>
+
+  <Import Project="$(MSBuildToolsPath)\Microsoft.Common.targets" />
+
+  <PropertyGroup>
+    <DebBuildDependsOn>$(DebBuildDependsOn);LayoutSharedFramework</DebBuildDependsOn>
+  </PropertyGroup>
+
+  <Target Name="LayoutSharedFramework" DependsOnTargets="PrepareForBuild;ResolveProjectReferences">
```
pakrym: This is completely duplicated in between both installer projects. Worth to move to common targets?

pakrym (approved):

src/Installers/Debian/Directory.Build.props
natemcmaster: Only the Linux installer projects use these tasks directly.

src/Installers/Debian/Runtime/Debian.Runtime.debproj
natemcmaster: Figured it wasn't worth de-duplicating because it will be deleted in #7832 anyways.

dougbu (approved):

PR #7794, opened 2019-02-21
Targeting pack and shared framework fixups

natemcmaster:
* Fix package type to match spec ('DotnetPlatform')
* Fix missing metadata in MSI installer
* Fix incorrect target order which caused duplicate assemblies in netcore.app and aspnetcore.app (resolves https://github.com/aspnet/AspNetCore-Internal/issues/1784)

### DIFF
--- a/src/Framework/test/SharedFxTests.cs
+++ b/src/Framework/test/SharedFxTests.cs
@@ -25,6 +25,33 @@ namespace Microsoft.AspNetCore
         }
 
         [Fact]
+        public void SharedFrameworkContainsExpectedFiles()
+        {
+            var actualAssemblies = Directory.GetFiles(TestData.GetTestDataValue("RuntimeAssetsOutputPath"), "*.dll")
+                .Select(Path.GetFileNameWithoutExtension)
+                .ToHashSet();
+            var expectedAssemblies = TestData.GetSharedFxDependencies()
+                .Split(';', StringSplitOptions.RemoveEmptyEntries)
+                .ToHashSet();
+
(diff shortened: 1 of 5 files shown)


pakrym (approved):

src/Framework/test/SharedFxTests.cs
```diff
@@ -24,6 +24,33 @@ public SharedFxTests(ITestOutputHelper output)
...
+        {
+            var actualAssemblies = Directory.GetFiles(TestData.GetTestDataValue("RuntimeAssetsOutputPath"), "*.dll")
+                .Select(Path.GetFileNameWithoutExtension)
+                .ToHashSet();
+            var expectedAssemblies = TestData.GetSharedFxDependencies()
+                .Split(';', StringSplitOptions.RemoveEmptyEntries)
+                .ToHashSet();
+
+            _output.WriteLine("==== actual assemblies ====");
+            _output.WriteLine(string.Join('\n', actualAssemblies));
```
Eilon: Just curious - for the sake of printing these out, would it make sense to sort them by some arbitrary string sort? That would make the debug output easier for humans to read (I realize it doesn't technically matter).

Eilon (approved):
One small test suggestion. But otherwise go ahead and merge.

src/Framework/test/SharedFxTests.cs
natemcmaster: Yeah, we could order by name. I'll add that when I merge this back to master (i have green CI checks right now, and don't want to wait 90 minutes to fix this up.)

Issue #6975, opened 2019-01-24
Resolve public API that exposes on 'pubternal' API


natemcmaster:
I tried enabling the 'pubternal' analyzer on this entire project and found 12 cases of public API which depends on pubternal. What should we do about these?

Options:
* Remove the API (breaking change)
* Suppress the analyzer error
* Rename API to be public (breaking change)

area-mvc: @mkArtakMSFT @JamesNK 
```
src/Features/JsonPatch/src/Adapters/AdapterFactory.cs(16,24): error PUB0001: Pubternal type ('Microsoft.AspNetCore.JsonPatch.Internal.IAdapter') usage in public API [src/Features/JsonPatch/src/Microsoft.AspNetCore.JsonPatch.csproj]
src/Features/JsonPatch/src/Adapters/IAdapterFactory.cs(20,9): error PUB0001: Pubternal type ('Microsoft.AspNetCore.JsonPatch.Internal.IAdapter') usage in public API [src/Features/JsonPatch/src/Microsoft.AspNetCore.JsonPatch.csproj]
src/Http/Routing/src/Template/TemplateBinder.cs(45,13): error PUB0001: Pubternal type ('Microsoft.AspNetCore.Routing.Internal.UriBuildingContext') usage in public API [src/Http/Routing/src/Microsoft.AspNetCore.Routing.csproj]
src/Http/Routing/src/Template/TemplateBinder.cs(65,13): error PUB0001: Pubternal type ('Microsoft.AspNetCore.Routing.Internal.UriBuildingContext') usage in public API [src/Http/Routing/src/Microsoft.AspNetCore.Routing.csproj]
src/Http/Routing/src/Tree/TreeRouteBuilder.cs(43,13): error PUB0001: Pubternal type ('Microsoft.AspNetCore.Routing.Internal.UriBuildingContext') usage in public API [src/Http/Routing/src/Microsoft.AspNetCore.Routing.csproj]
src/Http/Routing/src/Tree/TreeRouteBuilder.cs(63,13): error PUB0001: Pubternal type ('Microsoft.AspNetCore.Routing.Internal.UriBuildingContext') usage in public API [src/Http/Routing/src/Microsoft.AspNetCore.Routing.csproj]
[...]

PR #7800, opened 2019-02-21
Make ConventionBasedStartup pubternal

Tratcher:
#6975 ConventionBasedStartup consumes pubternal types and was never meant for public use.

### DIFF
--- a/src/Hosting/Hosting/ref/Microsoft.AspNetCore.Hosting.netcoreapp3.0.cs
+++ b/src/Hosting/Hosting/ref/Microsoft.AspNetCore.Hosting.netcoreapp3.0.cs
@@ -3,12 +3,6 @@
 
 namespace Microsoft.AspNetCore.Hosting
 {
-    public partial class ConventionBasedStartup : Microsoft.AspNetCore.Hosting.IStartup
-    {
-        public ConventionBasedStartup(Microsoft.AspNetCore.Hosting.Internal.StartupMethods methods) { }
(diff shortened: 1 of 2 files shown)


natemcmaster:
As long as we're making a breaking change, can this be made actual-internal, not just pubternal?

Adding @muratg to also review

Tratcher:
The decision on pupternals in this area was to leave them alone. Only MVC ended up doing the big pupternal purge.

natemcmaster:
> The decision on pupternals in this area was to leave them alone

Sure, but this isn't leaving an API alone, it's making a public API into a pubternal API. It's a breaking change due to namespace changes. I'll let @muratg evaluate if making this breaking change is worth it. 

What I was trying to suggest is that _if_ we're okay making a breaking change, is there a technical reason we can't make it actual-internal? I think we should avoid increasing the number of pubternal APIs

muratg:
Unfortunately the proposal of getting out of "pubternal" namespaces (and making things properly internal) was pushed back heavily. 

In this case, I think it's probably fine to do a breaking-change. Unless you guys think this is a heavily used.

natemcmaster:
Hard to say. GitHub search results don’t turn up much, other than people copying our samples and code, and I’m not aware of any other data on specific API usage. I suppose we could use complaint-based-metrics to measure the importance of the API.

Tratcher:
Note to self: rebase on the ref assembly change

muratg (approved):

PR #7804, opened 2019-02-21
Fix SHA calculations

ryanbrandenburg:
Fixes aspnet/AspNetCore-Internal#1833.


natemcmaster (approved):
:shipit: Get manager approval too as this targets release/2.2. Likely won't need shiproom approval because it's a test-only change.

cc @mkArtakMSFT @Eilon

ryanbrandenburg:
Didn't occur to me that I'd need approval since it's a test only change.

ryanbrandenburg:
Got approval from both @mkArtakMSFT and @Eilon offline.

PR #7808, opened 2019-02-21
Avoid port conflicts for HttpSys

Tratcher:
https://github.com/aspnet/AspNetCore-Internal/issues/1795

GetNextPort doesn't cover the case where netsh has a reservation for a specific port but it's not open right now. Kestrel can re-use those ports bug HttpSys can't since it's not the reservation owner.

The easiest way to check for a conflict is to try it. This is about how it's done in HttpSys's own functional tests. https://github.com/aspnet/AspNetCore/blob/4451a804477f7f3d4d0ed89d56c51332d8c3ee25/src/Servers/HttpSys/test/FunctionalTests/Utilities.cs#L125

### DIFF
--- a/src/Hosting/Server.IntegrationTesting/src/Common/TestPortHelper.cs
+++ b/src/Hosting/Server.IntegrationTesting/src/Common/TestPortHelper.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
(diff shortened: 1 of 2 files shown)


Tratcher:
*ping* Note AspNetCore-ci passed but did not report all of it's status.

halter73 (approved):

src/Hosting/Server.IntegrationTesting/src/Common/TestPortHelper.cs
```diff
@@ -56,5 +56,34 @@ public static int GetNextSSLPort()
...
+                        server.Stop();
+                        return port;
+                    }
+                    catch (HttpListenerException)
+                    {
+                    }
+                }
+            }
+            NextPort = BasePort;
+            throw new Exception("Failed to locate a free port.");
```
halter73: Any idea how long this would take to fail?

jkotalik (approved):
Eeek whoops didn't see this.

src/Hosting/Server.IntegrationTesting/src/Common/TestPortHelper.cs
Tratcher: The individual failures are instant. You could burn through the whole port range in under a minute if it really came to that. It's never come up in HttpSys.

PR #7817, opened 2019-02-21
Detect Classic SignalR Server

mikaelm12:
Fixes: https://github.com/aspnet/AspNetCore/issues/7781
Detect a user trying to connect to a classic SignalR Server with a Core client and guide them in the right direction.
Related: https://github.com/aspnet/AspNetCore/issues/7705

### DIFF
--- a/src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/NegotiateResponse.java
+++ b/src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/NegotiateResponse.java
@@ -28,6 +28,10 @@ class NegotiateResponse {
                     case "error":
                         this.error = reader.nextString();
                         break;
+                    case "ProtocolVersion":
+                        this.error = "Detected a connection attempt to an ASP.NET SignalR Server. This client only supports connecting to an ASP.NET Core SignalR Server. See https://aka.ms/signalr-core-differences for details.";
+                        reader.skipValue();
+                        break;
                     case "url":
                         this.redirectUrl = reader.nextString();
                         break;
@@ -69,7 +73,6 @@ class NegotiateResponse {
                         break;
                 }
             } while (reader.hasNext());
-
             reader.endObject();
             reader.close();
         } catch (IOException ex) {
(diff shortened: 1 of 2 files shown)


src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/NegotiateResponse.java
```diff
@@ -28,6 +28,10 @@ public NegotiateResponse(String negotiatePayload) {
                     case "error":
                         this.error = reader.nextString();
                         break;
+                    case "ProtocolVersion":
+                        this.error = "Detected a connection attempt to an ASP.NET SignalR Server. This client only supports connecting to an ASP.NET Core SignalR Server. See https://aka.ms/signalr-core-differences for details.";
+                        reader.skipValue();
```
BrennanConroy: should we just return immediately here?

src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/NegotiateResponse.java
```diff
@@ -28,6 +28,10 @@ public NegotiateResponse(String negotiatePayload) {
                     case "error":
                         this.error = reader.nextString();
                         break;
+                    case "ProtocolVersion":
+                        this.error = "Detected a connection attempt to an ASP.NET SignalR Server. This client only supports connecting to an ASP.NET Core SignalR Server. See https://aka.ms/signalr-core-differences for details.";
```
anurse: "Detected a connection attempt" sounds a little awkward and I think we can simplify a bit.

```suggestion
                        this.error = "Detected an ASP.NET SignalR Server. This client only supports connecting to an ASP.NET Core SignalR Server. See https://aka.ms/signalr-core-differences for details.";
```

analogrelay (approved):
👍

src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/NegotiateResponse.java
anurse: That would be reasonable. There's no need to finish parsing, the rest is garbage anyway.

mikaelm12: 👍

PR #7818, opened 2019-02-21
Change build to use yarn instead of npm 🧶

natemcmaster:
Resolves https://github.com/aspnet/AspNetCore/issues/7563 

This PR has two commits. I definitely recommend reviewing the first commit separately because the result of running `yarn import` (converts package-lock.json to yarn.lock) changed like a bajillion lines of code.


analogrelay (commented):
Looks OK to me but I don't have the context of all the build system changes that have happened over the past couple of months to feel confident signing off completely :)

ryanbrandenburg (approved):
I'm not super familiar with yarn but this looks good.

ryanbrandenburg:
Looks like you need to regenerate your ref files and fix a reference to Yarn.msbuild?

src/SignalR/clients/ts/signalr/README.md
```diff
@@ -4,6 +4,7 @@ JavaScript and TypeScript clients for SignalR for ASP.NET Core
 
 ```bash
 npm install @aspnet/signalr
+# or "yarn add @aspnet/signalr"
```
BrennanConroy: Whats with using `#` here?

src/SignalR/clients/ts/signalr-protocol-msgpack/README.md
```diff
@@ -4,6 +4,7 @@ MsgPack support for SignalR for ASP.NET Core
 
 ```bash
 npm install @aspnet/signalr-protocol-msgpack
+# or "yarn add @aspnet/signalr-protocol-msgpack"
```
BrennanConroy: And here

BrennanConroy (approved):

dougbu (approved):
LGTM once builds succeed

To confirm: Builds will do the needful and `yarn` is not a prerequisite. Correct?

src/SignalR/clients/ts/signalr/README.md
natemcmaster: It's a code comment. Users should run either npm install or yarn add, but not both. Open to suggestions for making that more obvious

BrennanConroy:
```bash
npm install @aspnet/signalr
```
or
```bash
yarn add @aspnet/signalr
```

Could work

SteveSandersonMS (approved):
https://github.com/natemcmaster/Yarn.MSBuild says that it installs Yarn for you so you don't need to install it globally. Does that mean it always prefers the local installation even if you have a global one?

If so, great! It would definitely be preferable for the repo to control the version of Yarn we're using, so we don't get any inconsistencies based on which versions people have installed globally.

If it ever does use the globally-installed version, maybe that's something we could look at changing if it turns out to cause problems.

natemcmaster:
> Does that mean it always prefers the local installation even if you have a global one?

Yes, by default Yarn.MSBuild uses a local version of yarn. Yarn.MSBuild 1.12.3 will use yarn 1.12.3.

That said, I've just noticed the latest stable yarn is 1.13.0. I'll update Yarn.MSBuild and update this PR to use the latest verison of yarn.

natemcmaster:
> @dougbu: To confirm: Builds will do the needful and `yarn` is not a prerequisite. Correct?

Confirmed. Machine prerequisites have not changed. You still only need NodeJS. Yarn.MSBuild takes care of acquiring yarn. You only need to install yarn to your machine if you want to make changes to one of the npm packages. (`npm i -g yarn` is the easiest way to do that.)

src/SignalR/clients/ts/signalr/README.md
natemcmaster: ok, updated the readme

PR #7830, opened 2019-02-21
Reenable helix payload

HaoK:
See if we can skip a few tests instead of disabling the entire project on helix


HaoK:
Changes look good adding these to https://github.com/aspnet/AspNetCore/pull/7763

Issue #4257, opened 2018-11-27
3.0 Preview 1 packages require adding a FrameworkReference to the project


natemcmaster:
Depends on https://github.com/NuGet/Home/issues/7342.
Follow up to https://github.com/aspnet/AspNetCore/pull/4178

The packages we ship which reference the shared framework should be built using `FrameworkReference`. Once https://github.com/NuGet/Home/issues/7342 is resolved, the metadata for this item will be preserved in the nuspec, which should ensure consumers get all the necessary runtime settings and compiler references.

**Workaround**
Consumers using our AspNetCore packages, such as Microsoft.AspNetCore.NodeServices, need to make sure they also have `<FrameworkReference Include="Microsoft.AspNetCore.App" />`. This will be implicitly added with Microsoft.NET.Sdk.Web, but not other MSBuild SDKs. 

**Changes required**

1. We need to change our own packages to build with `FrameworkReference`. (Related work #4246)
2. Update our SDK in BuildTools to support using `FrameworkReference`

natemcmaster:
Prototype in progress: https://github.com/aspnet/AspNetCore/compare/natemcmaster:local-fx

Current challenges:

* I get compiler errors due to duplicate APIs in the same namespace. I think this is the result of having InternalsVisibleTo and duplicate copies of API due to shared source.
* Visual Studio go-to-definition is broken when navigating from something outside the shared framework into a shared framework assembly.
* Working around https://github.com/dotnet/core-setup/issues/4809 - the only way to make tests run against the shared framework is to install the shared framework into DOTNET_ROOT. My workaround is to make DOTNET_ROOT = "$repoRoot/.dotnet/$(arch)/".
* Requires VS 2019 and a Preview 2 SDK (blocked on https://github.com/aspnet/AspNetCore/pull/7005).

cc @pakrym @ryanbrandenburg

natemcmaster:
@Pilchie I think this is still worth addressing, but this is a challenging issue to address. The challenges I described in the previous comment are still unresolved. We would probably need help from the teams which own VS, the runtime host, and the SDK to completely resolve this. It's extremely unlikely this will land in 3.0. 3.1 or 5.0 is a more appropriate timeframe for this.

[...]

PR #7832, opened 2019-02-22
Deprecate the Microsoft.AspNetCore.App metapackage in favor of targeting and runtime packs

### LINKED ISSUE #6501: Create targeting pack for ASP.NET Core
For .NET Core 3.0, we expect to separate targeting packs, which contain the assets needed at build time when targeting .NET Core 3.0 (for example, reference assemblies and intellisense documentation) from "runtime packs", which will include the platform-specific runtime components for .NET Core 3.  See dotnet/cli#10006, and [dotnet/designs#50](https://github.com/dotnet/designs/pull/50/files?short_path=ebf265d#diff-ebf265d0476041dd8cab82fa24ab5b0e)

We will have targeting pack for each shared framework (.NET Core, ASP.NET Core, and WindowsDesktop), as well as for .NET Standard.  This issue tracks the work required for the ASP.NET Core shared framework.

The work required includes:

- A new NuGet package including the targeting pack assets.  This should include the reference assemblies, intellisense documentation, and platform manifest file.  The current straw man proposal for the package ID is `Microsoft.NETCore.App.Ref`
- Installers for each supported installer format (msi, deb, pkg, and rpm), which lay out the targeting pack files under `<DOTNET_ROOT>/packs/<PackageId>/version`.  These installers will be bundled or depended on by the corresponding .NET Core SDK installers.

Some of the details (such as package layout, names, etc) are still [under design](https://github.com/dotnet/designs/pull/50/files?short_path=ebf265d#diff-ebf265d0476041dd8cab82fa24ab5b0e) and may change slightly.

natemcmaster:
Resolves #6501 

This is all about how we build and redistribute aspnetcore. It doesn't change the way users reference aspnetcore.

Changes
* Stop producing Microsoft.AspNetCore.App, and runtime.$(rid).Microsoft.AspNetCore.App
* Generate the shared framework without using 'NuGet' restore
* Put the platform manifest into the targeting pack (data/Microsoft.AspNetCore.App.PlatformManifest.txt)
* Removes the need to generate packages so we can figure out what belongs in the shared framework
* Create well-known, shared intermediate directories that installers can use to bundle content
 
cc @nguerrera @dagood @dsplaisted


dsplaisted:
We need to do https://github.com/dotnet/cli/issues/10442 before you stop producing the runtime.$(rid).Microsoft.AspNetCore.App packages

natemcmaster:
I see https://github.com/dotnet/cli/issues/10442 is marked as `<blocked>`. Do you have a rough ETA?

nguerrera:
I think the `blocked` was because we didn't have the new packages so that is now unblocked. Also, we don't need the file list to unblock this because we kept nuget folder conventions.

natemcmaster:
> we don't need the file list

Which file list, the platform manifest?

Directory.Build.targets
```diff
@@ -8,6 +8,11 @@
     -->
     <IsShippingPackage Condition=" '$(IsShippingPackage)' == '' ">false</IsShippingPackage>
 
+    <!--
+      By default, assemblies which are only in the Microsoft.AspNetCore.App shared framework are not available as NuGet packages.
+    -->
+    <IsPackable Condition="'$(IsAspNetCoreApp)' == 'true' AND '$(IsShippingPackage)' != 'true'">false</IsPackable>
```
dougbu: Should be `OR`

docs/ProjectProperties.md
```diff
@@ -6,4 +6,4 @@ In addition to the standard set of MSBuild properties supported by Microsoft.NET
 Property name      | Meaning
 -------------------|--------------------------------------------------------------------------------------------
 IsShippingPackage  | When set to `true`, the package produced by from project is intended for use by customers. Defaults to  `false`, which means the package is intended for internal use only by Microsoft teams.
-IsAspNetCoreApp    | Set to `true` when the assembly is part of the [Microsoft.AspNetCore.App shared framework](./SharedFramework.md)
\ No newline at end of file
+IsAspNetCoreApp    | Set to `true` when the assembly is part of the [Microsoft.AspNetCore.App shared framework](./SharedFramework.md) and is not available as a NuGet package (unless IsShippingPackage is also set to `true`.)
```
dougbu: Property needs a new name if you really mean "and is not &hellip;". Isn't "not available as a NuGet package" true for all shared framework components these days?

IOW property name makes sense only so long it means the property ships in the shared framework (no caveats).

dougbu (approved):
One bit is confusing me but, otherwise (once unblocked and builds pass), :shipit:

Directory.Build.targets
natemcmaster: `AND` is deliberate. This ensures we don't produce packages for shared-framework-only assemblies like Mvc.Core, but still do for shared-framework-and-package assemblies like SignalR

docs/ProjectProperties.md
natemcmaster: > Isn't "not available as a NuGet package" true for all shared framework components these days?

Nope. Here are the exceptions: https://github.com/aspnet/AspNetCore/blob/f19e8da1daf2808f3bbb870141304b73e76a547a/eng/SharedFramework.Local.props#L10-L17

pakrym (approved):

Directory.Build.targets
natemcmaster: Decide to back this out. After closer inspection of the nupkg's produced by the build, we need to keep producing the intermediate nupkg's for now. We can tackle this again when we do #4257

pakrym:
cc @anurse

natemcmaster:
After this merges, anything still using a package reference to Microsoft.AspNetCore.App will stop getting updates. The new packages, Microsoft.AspNetCore.App.Ref and Microsoft.AspNetCore.App.Runtime.$(rid) should not be referenced directly...they're not mean to be used as a `<PackageReference>`. Instead, consumers should use `<FrameworkReference>`.

@nguerrera @dsplaisted - if we merge this to aspnet today, will you be able to make changes to react to this in dotnet/core-sdk? I took a look, and it seems the repo is still using Microsoft.AspNetCore.App to flow dependency updates and find the list of RIDs the aspnetcore runtime supports. 

cc @sebastienros - benchmarking infrastructure

davidfowl:
Are the sdk changes  already in to consume these changes?

natemcmaster:
No, that's why I made the previous comment. Not sure if we're expected to
do this or if the SDK team is already on it.

>

davidfowl:
Ah then we shouldn’t stop producing the old shared framework until that’s the case.

dsplaisted:
> Not sure if we're expected to do this or if the SDK team is already on it.

We will be updating the SDK to consume the updated packages

natemcmaster:
I'd like to merge this before master diverges too much from this PR. Got tired of waiting for the SDK, so here you go: https://github.com/dotnet/core-sdk/pull/754

sebastienros (approved):

davidfowl:
@natemcmaster This change is problematic. Do libraries authors build packages that use ASP.NET Core during the 3.0 time frame? This will affect our GRPC development experience if the ref pack can't be used as a substitute.

natemcmaster:
Quoting myself from above

> This is all about how we build and redistribute aspnetcore. It doesn't change the way users reference aspnetcore.

Keep using `<FrameworkReference>`

dsplaisted:
@davidfowl Libraries should not have a package dependency on Microsoft.AspNetCore.App.  We need https://github.com/NuGet/Home/issues/7342 and https://github.com/dotnet/cli/issues/10666 for packages to be able to declare dependencies on the AspNetCore shared framework.  Until then, apps that reference libraries that depend on ASP.NET Core will need to have a direct FrameworkReference to Microsoft.AspNetCore.App.

davidfowl:
I see, libraries just use the FrameworkReference for compile assets and those won't persist. OK. I see that's already the case for GRPC https://www.nuget.org/packages/Grpc.AspNetCore.Server/0.1.19-pre1.

nguerrera:
Aside: GRPC's dependency on Microsoft.NETCore.Platforms is a bug (fixed in preview3) and can lead to downgrades.

PR #7833, opened 2019-02-22
Clarify ExternalLoginInfo displayName doc comment

### LINKED ISSUE #6712: Inconsistent DisplayName CTOR/property documented intent within ExternalLoginInfo and UserLoginInfo
In [ExternalLoginInfo.cs](https://github.com/aspnet/AspNetCore/blob/65ff6e951a29ac026e6528c203f590244948910c/src/Identity/Core/src/ExternalLoginInfo.cs) it says `displayName` is:

> The display name for this user provided by the login provider.

However, in [UserLoginInfo.cs](https://github.com/aspnet/AspNetCore/blob/65ff6e951a29ac026e6528c203f590244948910c/src/Identity/Extensions.Core/src/UserLoginInfo.cs) that parameter is assigned to `ProviderDisplayName`, which it says is:

> The display name for the provider.

The ctor description also diverges in the base class.

However, I don't really see why the ProviderDisplayName is stored with each user, when it is bound to the provider, not the user. Maybe the intent was originally to store the user's Profile.DisplayName from the provider in that field?

HaoK:
Fixes https://github.com/aspnet/AspNetCore/issues/6712 

@blowdart

### DIFF
--- a/src/Identity/Core/src/ExternalLoginInfo.cs
+++ b/src/Identity/Core/src/ExternalLoginInfo.cs
@@ -18,7 +18,7 @@ namespace Microsoft.AspNetCore.Identity
         /// <param name="principal">The <see cref="ClaimsPrincipal"/> to associate with this login.</param>
         /// <param name="loginProvider">The provider associated with this login information.</param>
         /// <param name="providerKey">The unique identifier for this user provided by the login provider.</param>
-        /// <param name="displayName">The display name for this user provided by the login provider.</param>
+        /// <param name="displayName">The display name for the login provider.</param>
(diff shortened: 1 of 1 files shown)


blowdart (approved):
LGTM

osagga:
@HaoK #6712 states that the doc comment is not correct in both `ExternalLoginInfo` and `UserLoginInfo`, is there a reason that this PR only addresses the issue in `ExternalLoginInfo`? Shouldn't a fix to the issue also address the same issue here
https://github.com/dotnet/aspnetcore/blob/9939b3358c886b725ac122b63bfcbf0d665d1d21/src/Identity/Extensions.Core/src/UserLoginInfo.cs#L16

PR #7834, opened 2019-02-22
Remove web::uri usage from public API

BrennanConroy:
Also removed a couple redundant tests and enabled some of the disabled ones since they should be fixed.

Deleted some old unused code I found too.

And fixed a flaky build issue with copying a dll

### DIFF
--- a/src/SignalR/clients/cpp/test/signalrclienttests/connection_impl_tests.cpp
+++ b/src/SignalR/clients/cpp/test/signalrclienttests/connection_impl_tests.cpp
@@ -111,6 +111,28 @@ TEST(connection_impl_start, connection_state_is_disconnected_when_connection_can
     ASSERT_EQ(connection->get_connection_state(), connection_state::disconnected);
 }
 
+TEST(connection_impl_start, throws_for_invalid_uri)
+{
+    std::shared_ptr<log_writer> writer(std::make_shared<memory_log_writer>());
+
+    auto websocket_client = create_test_websocket_client(
+        /* receive function */ []() { return pplx::task_from_result(std::string("{ }\x1e")); });
+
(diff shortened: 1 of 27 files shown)


halter73:
Did web::uri do any validation? If so, what happens now when I try to start the connection with a uri string would be an invalid web::uri? For example, does create_web_request start throwing?

BrennanConroy:
>does the http_client ctor start throwing?

Yes, I believe this will throw now, as it converts the string into a `uri` and that does check for a valid uri.
I'll check for a test with an invalid uri, and add one if it's missing.

src/SignalR/clients/cpp/test/signalrclienttests/connection_impl_tests.cpp
```diff
@@ -111,6 +111,28 @@ TEST(connection_impl_start, connection_state_is_disconnected_when_connection_can
...
+    auto connection = connection_impl::create(_XPLATSTR("/bad_uri?a=b"), trace_level::errors, writer, create_test_web_request_factory(), std::make_unique<test_transport_factory>(websocket_client));
+
+    try
+    {
+        connection->start().get();
+        ASSERT_TRUE(false);
+    }
+    catch (...)
+    {
+        // We shouldn't check the exact exception as it would be specific to the http library being used
```
halter73: What is the exception with the default http library? Does it make it clear the URI is invalid?

halter73 (approved):

src/SignalR/clients/cpp/test/signalrclienttests/connection_impl_tests.cpp
BrennanConroy: "provided uri is invalid: :1	 ÃƒÂ¤ bad_uri&a=b"

Issue #7841, opened 2019-02-22
Fix SignalRSamples by reacting to #7751


halter73:
https://github.com/aspnet/AspNetCore/pull/7751 broke SignalRSamples

```
C:\dev\aspnet\AspNetCore\src\SignalR\samples\SignalRSamples> dotnet run
Application startup exception: System.InvalidOperationException: The CORS protocol does not allow specifying a wildcard (any) origin and credentials at the same time. Configure the CORS policy by listing individual origins if credentials needs to be supported.
   at Microsoft.AspNetCore.Cors.Infrastructure.CorsPolicyBuilder.Build() in C:\dev\aspnet\AspNetCore\src\Middleware\CORS\src\Infrastructure\CorsPolicyBuilder.cs:line 232
   at Microsoft.AspNetCore.Cors.Infrastructure.CorsOptions.AddPolicy(String name, Action`1 configurePolicy) in C:\dev\aspnet\AspNetCore\src\Middleware\CORS\src\Infrastructure\CorsOptions.cs:line 101
   at SignalRSamples.Startup.<>c.<ConfigureServices>b__0_1(CorsOptions o) in C:\dev\aspnet\AspNetCore\src\SignalR\samples\SignalRSamples\Startup.cs:line 38
   at Microsoft.Extensions.Options.ConfigureNamedOptions`1.Configure(String name, TOptions options)
   at Microsoft.Extensions.Options.OptionsFactory`1.Create(String name)
   at Microsoft.Extensions.Options.OptionsManager`1.<>c__DisplayClass5_0.<Get>b__0()
   at System.Lazy`1.ViaFactory(LazyThreadSafetyMode mode)
   at System.Lazy`1.ExecutionAndPublication(LazyHelper executionAndPublication, Boolean useDefaultConstructor)
   at System.Lazy`1.CreateValue()
   at System.Lazy`1.get_Value()
   at Microsoft.Extensions.Options.OptionsCache`1.GetOrAdd(String name, Func`1 createOptions)
   at Microsoft.Extensions.Options.OptionsManager`1.Get(String name)
   at Microsoft.Extensions.Options.OptionsManager`1.get_Value()
   at Microsoft.AspNetCore.Cors.Infrastructure.CorsService..ctor(IOptions`1 options, ILoggerFactory loggerFactory) in C:\dev\aspnet\AspNetCore\src\Middleware\CORS\src\Infrastructure\CorsService.cs:line 41
[...]

PR #7751, opened 2019-02-20
Throw if CORS policy is configured to allow credentials and any origin

pranavkm:
Fixes https://github.com/aspnet/AspNetCore/issues/3106

Changes per https://github.com/aspnet/AspNetCore/issues/3106#issuecomment-429091009

PR #7842, opened 2019-02-22
React to #7751

halter73:
See #7841 for why we need to react to #7751

### DIFF
--- a/src/SignalR/samples/SignalRSamples/Startup.cs
+++ b/src/SignalR/samples/SignalRSamples/Startup.cs
@@ -4,7 +4,6 @@
 using System;
 using System.IO;
 using System.Reflection;
-using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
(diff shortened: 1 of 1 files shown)


src/SignalR/samples/SignalRSamples/Startup.cs
```diff
@@ -32,17 +31,6 @@ public void ConfigureServices(IServiceCollection services)
...
-            services.AddCors(o =>
-            {
-                o.AddPolicy("Everything", p =>
-                {
-                    p.AllowAnyHeader()
-                     .AllowAnyMethod()
-                     .AllowAnyOrigin()
-                     .AllowCredentials();
-                });
-            });
```
halter73: I don't know what was using this before, but now this causes the app to crash on startup, so removing it is at least better.

If we can find out how CORS is supposed to be used in the SignalRSamples app, we can redo the CORS configuration in a way that's still allowed.

BrennanConroy:
Need to remove the `UseCors(...)` call too.

halter73:
@BrennanConroy Good catch. Now approve it 😄

BrennanConroy (approved):

Issue #7708, opened 2019-02-19
Blazor new Lifecycle method executed after references got filled


Lupusa87:
### Is your feature request related to a problem? Please describe.
As we know component/element references are getting filled when component is rendered, we 
 use OnAfterRender to do some actions with referenced component/element.

OnAfterRender gets executed each time component re-renders and in most cases we don't need to do some action with referenced component/element each time.

To avoid this we are forced to do it only first time OnAfterRender is called.
[example](https://github.com/Lupusa87/BlazorScrollbar/blob/master/BlazorScrollbar/Pages/Index_Logic.cs#L58)

```
public CompBlazorScrollbar CompBlazorScrollbar1;
bool FirstLoad = true;

protected override void OnAfterRender()
        {
            if (FirstLoad)
            {
                FirstLoad = false;

                CompBlazorScrollbar1.OnPositionChange += OnPositionChanged;

                base.OnAfterRender();
            }  
        }
```
This technique solves problem but this method gets executed each time on re-render and extra work is made to check FirstLoad value also having this extra variable in code.

### Describe the solution you'd like
If it is possible, would be great to have separated lifecycle method which will execute once as OnInit but with filled component/element references.

It is not feature we can't live without but I think will be helpful and convenient to have it.

mkArtakMSFT:
Thanks for contacting us, @Lupusa87.
@SteveSandersonMS thoughts? I agree that a separate lifecycle event (maybe named `OnAfterFirstRender`) would be useful.

SteveSandersonMS:
Yes, either `OnAfterFirstRender` or `OnAfterRender(bool isFirstRender)` seem like they would be useful.

javiercn:
I'm gonna let @SteveSandersonMS chip in here, but based on previous discussions for similar things (JSInterop) we want to limit the amount of lifecycle methods that we include. As you mention it is already possible to achieve this and we are not sure about how common this scenario is for application developers (vs for example library authors), so we want to be conservative here.

We can always add lifecycle methods later, but taking them away is harder.

@SteveSandersonMS I'm wondering:
Is it possible to change a razor component base class from the .razor file?
This way you could create your own custom lifecycle methods based on ours by implementing the above pattern in a base class (that extends ComponentBase). and simply use that across your components

Lupusa87:
@javiercn thank you,

I understand what you say and it makes lot of sense.

[...]

PR #7843, opened 2019-02-22
Add new lifecycle method OnAfterFirstRender

Lupusa87:
Summary of the changes
 - Added OnAfterFirstRender and OnAfterFirstRenderAsync methods
 - On first render will be called OnAfterFirstRender and OnAfterRender methods both.

Addresses #7708


DNF-SaS:
Wondering why this is not approved.

Lupusa87:
Team will do it if they decide, there is different possition about this.

PR #7856, opened 2019-02-22
Avoid UI flickering after prerendering

SteveSandersonMS:
Prerendering is a great feature.

But one minor quirk I noticed was that the UI went through a brief -- probably < 50ms -- flash of whiteness when replacing the prerendered elements with client-rendered (interactive) elements.

This PR tweaks the client-side rendering mechanism so there is no longer any async gap between clearing and replacing the content, so the browser never repaints the screen in an intermediate state. It feels slicker.

### DIFF
--- a/src/Components/Browser.JS/src/Rendering/BrowserRenderer.ts
+++ b/src/Components/Browser.JS/src/Rendering/BrowserRenderer.ts
@@ -7,6 +7,7 @@ const selectValuePropname = '_blazorSelectValue';
 const sharedTemplateElemForParsing = document.createElement('template');
 const sharedSvgElemForParsing = document.createElementNS('http://www.w3.org/2000/svg', 'g');
 const preventDefaultEvents: { [eventType: string]: boolean } = { submit: true };
+const rootComponentsPendingFirstRender: { [componentId: number]: Element } = {};
 
 export class BrowserRenderer {
   private eventDelegator: EventDelegator;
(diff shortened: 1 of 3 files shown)


javiercn (approved):
This is nice and pretty straightforward. It works the way I would expect.

Quick question, (cuz I'm not that familiar with the DOM side of things) when we do a client render, do we keep the main root component tag? (Like `<app></app>`) or do we replace the node with the result from rendering the component (which typically would be the contents of the component)

SteveSandersonMS:
> when we do a client render, do we keep the main root component tag?

Yes, we leave it there. The developer put that `<app>` element in their source code, and they might have CSS rules like:

```css
app { color: red }
```

... so it would be pretty unhelpful to be removing the `<app>` element for them.

This also simplifies things for prerendering and the prerendered-to-live transition, doesn't it? It means we know for sure there's exactly one top-level element, so on the client we know which prerendered elements are the ones we should replace.

javiercn:
> This also simplifies things for prerendering and the prerendered-to-live transition, doesn't it? It means we know for sure there's exactly one top-level element, so on the client we know which prerendered elements are the ones we should replace.

Yeah, I was asking because I have some ideas on how that will end up working, this helps!

rynowak (approved):
I don't have much familiarity with the code, but I agree that it's a good idea!

Issue #7648, opened 2019-02-16
Extract property AutoSaveChanges into new interface


michaeldz:
### Problem
I need to execute multiply actions related with UserStore into scope of one transaction. Easiest way to do it:
``` 
myUserStore.AutoSaveChanges = false; 
// manipulate with users ...
await dbContext.SaveChangesAsync();
```

So I need to use AutoSaveChanges in my service/code. But user store is injected into my code thru DE as `IUserStore<MyUser>`.

AutoSaveChanges is defined into base generic class `UserStore<TUser, TRole, TContext, TKey, TUserClaim, TUserRole, TUserLogin, TUserToken, TRoleClaim>`. 

It is very inconvenient to cast the store object to base (or derived) type, especially if my service or code more or less abstract. I mean if my service/code should works with custom implementation of UserStore that unknown. In this case I should cast to base generic UserStore<...>

Solution with reflection like this 
```
var userStoreType = _userStore.GetType();
var genericUserStoreType = typeof(UserStore<,,,,,,,,>);
var t = userStoreType;

while (t != null)
{
    if (t.IsGenericType && t.GetGenericTypeDefinition() == genericUserStoreType)
    {
        userStoreType.InvokeMember(nameof(UserStore.AutoSaveChanges), BindingFlags.SetProperty, null, _userStore, new object[] { false });
    }
    t = t.BaseType;
}
```
doesn't look perfect and optimal.

### Solution
Add new interface IEFStore (perhaps more convenient/right name is required):
```
public interface IEFStore
{
    bool AutoSaveChanges { get; set; }
}
```

and implement it in:
`UserStore<TUser, TRole, TContext, TKey, TUserClaim, TUserRole, TUserLogin, TUserToken, TRoleClaim>`
`UserOnlyStore<TUser, TContext, TKey, TUserClaim, TUserLogin, TUserToken>`
`RoleStore<TRole, TContext, TKey, TUserRole, TRoleClaim>` 

my code will look:
```
if(_userStore is IEFStore efStore)
{
    efStore.AutoSaveChanges = false;
}
```

blowdart:
This is not something we can do easily, if at all, without a grander redesign, so I'm sorry, but no.

michaeldz:
Looks strange, probably misunderstanding ...
Adding the interface **IEFStore** to the store class:
```
[...]

PR #7869, opened 2019-02-22
Extract property AutoSaveChanges into new interface (#7648)

michaeldz:
Just extract AutoSaveChanges property in new interface. See #7648 for details.

### DIFF
--- a/src/Identity/EntityFrameworkCore/src/IAutoSaveChanges.cs
+++ b/src/Identity/EntityFrameworkCore/src/IAutoSaveChanges.cs
@@ -1,0 +1,9 @@
+using System;
+
+namespace Microsoft.AspNetCore.Identity.EntityFrameworkCore
+{
+    public interface IAutoSaveChanges
+    {
+        bool AutoSaveChanges { get; set; }
+    }
+}
--- a/src/Identity/EntityFrameworkCore/src/RoleStore.cs
+++ b/src/Identity/EntityFrameworkCore/src/RoleStore.cs
@@ -75,7 +75,8 @@ namespace Microsoft.AspNetCore.Identity.EntityFrameworkCore
     /// <typeparam name="TRoleClaim">The type of the class representing a role claim.</typeparam>
     public class RoleStore<TRole, TContext, TKey, TUserRole, TRoleClaim> :
         IQueryableRoleStore<TRole>,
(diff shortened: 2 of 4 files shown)


HaoK:
Issue was closed https://github.com/aspnet/AspNetCore/issues/7648

michaeldz:
@HaoK Hao could you explain little bit more what's wrong with this change?

HaoK:
If you are trying to enable AutoSaveChanges in your app, just derive from the appropriate UserStore class and set it to true.  We aren't going to adding an interface for this

michaeldz:
@HaoK In my case it doesn't work. Because it is like a "3rd party component" .. all what I know - there is some store that is UserStore, but I don't know which class is it exactly, and of course I cannot derive it.
So there are two solution - use complex and not fast reflection or change AspNet because open-source ... I hope.

michaeldz:
@HaoK Probably I explained my issue not well. Let's try one more time. I'm developing a framework. This framework **extends** base functionality of ASP.Net Identity. It adds couple of new things - providers, services etc.
It should be applicable for other applications. We do not want to limit users in customization - they can provide custom user, roles and other things that they usually do with regular Identity framework.
This approach works well. There is only one issue. We require to apply changes for service-level operations in scope of single transaction for related entities. 
In Identity framework there is good mechanism - the AutoSaveChanges property.
But as I mentioned early because there is no interface - just property of base User/Role Store - I cannot use this mechanism.
Could you provide workaround for implement this functionality?

Thank you.

HaoK:
You can try doing some kind of reflection/base type inspection to set AutoSaveChanges, we do that ourselves to do things like determine what's the appropriate store class to put in DI based on the user/dbcontext types.

See https://github.com/aspnet/AspNetCore/blob/master/src/Identity/EntityFrameworkCore/src/IdentityEntityFrameworkBuilderExtensions.cs#L33

michaeldz:
@HaoK You be able to do it more or less easy because you have user (or role) type. You can receive the types from the builder, but in my case I don't have the types, unfortunately. This is problem.

michaeldz:
@HaoK So am I right that there is no solution for my case?

HaoK:
Not really, to do anything with identity you need to know the user type for the app, once you have the user type you can do reflection against the store just like we do

PR #7873, opened 2019-02-22
Restore Microsoft.VisualStudio.Web.CodeGeneration.Design references in templates

dougbu:

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -76,6 +76,7 @@
     <MicrosoftExtensionsWebEncodersPackageVersion>2.1.1</MicrosoftExtensionsWebEncodersPackageVersion>
 
     <!-- Packages that come from aspnet/Scaffolding. (Used only in the 2.1.x package archives.) -->
+    <!-- See src/Templating/build/dependencies.props for the versions used in templates. -->
     <MicrosoftVisualStudioWebCodeGenerationContractsPackageVersion>2.1.7</MicrosoftVisualStudioWebCodeGenerationContractsPackageVersion>
     <MicrosoftVisualStudioWebCodeGenerationCorePackageVersion>2.1.7</MicrosoftVisualStudioWebCodeGenerationCorePackageVersion>
     <MicrosoftVisualStudioWebCodeGenerationDesignPackageVersion>2.1.7</MicrosoftVisualStudioWebCodeGenerationDesignPackageVersion>
--- a/src/Templating/src/Microsoft.DotNet.Web.ProjectTemplates/RazorPagesWeb-CSharp.csproj.in
+++ b/src/Templating/src/Microsoft.DotNet.Web.ProjectTemplates/RazorPagesWeb-CSharp.csproj.in
@@ -22,6 +22,7 @@
     <PackageReference Include="Microsoft.AspNetCore.Razor.Design" Version="${MicrosoftAspNetCoreRazorDesignPackageVersion}" PrivateAssets="All" />
     <PackageReference Include="Microsoft.EntityFrameworkCore.Sqlite" Version="${MicrosoftEntityFrameworkCoreSqlitePackageVersion}" Condition=" '$(IndividualLocalAuth)' == 'True' AND '$(UseLocalDB)' != 'True'" />
     <PackageReference Include="Microsoft.VisualStudio.Web.BrowserLink" Version="${MicrosoftVisualStudioWebBrowserLinkPackageVersion}" Condition="'$(UseBrowserLink)' == 'True'" />
+    <PackageReference Include="Microsoft.VisualStudio.Web.CodeGeneration.Design" Version="${MicrosoftVisualStudioWebCodeGenerationDesignPackageVersion}" PrivateAssets="All" Condition="'$(IndividualAuth)' == 'True'" />
   </ItemGroup>
   <ItemGroup Condition="'$(TargetFrameworkOverride)' != ''">
     <PackageReference Include="Microsoft.AspNetCore" Version="${MicrosoftAspNetCorePackageVersion}" />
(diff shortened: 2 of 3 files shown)


build/dependencies.props
```diff
@@ -76,6 +76,7 @@
     <MicrosoftExtensionsWebEncodersPackageVersion>2.1.1</MicrosoftExtensionsWebEncodersPackageVersion>
 
     <!-- Packages that come from aspnet/Scaffolding. (Used only in the 2.1.x package archives.) -->
+    <!-- See src/Templating/build/dependencies.props for the versions used in templates. -->
```
dougbu: @ryanbrandenburg those versions are a bit older e.g. `$(MicrosoftVisualStudioWebCodeGenerationCorePackageVersion)` is `2.1.3-rtm-30913` instead of the `2.1.7` used here. Should we update that version in the release branches?

ryanbrandenburg: I think I'm confused about where the `2.1.3-rtm-30913` version is coming from, so I don't know if we should update. Could you go into more detail?

dougbu:
@mkArtakMSFT this change should be necessary in `2.2.3` but it will affect users if we ship with the current templates in 2.1.9.

@ryanbrandenburg are we tracking the issue anywhere besides aspnet/AspNetCore-Internal#1759? And what symptoms would users experience? Is the workaround to add the missing `<ProjectReference>` in new projects -- or something more complicated?

build/dependencies.props
dougbu: The `2.1.3-rtm-30913` version comes from `src/Templating/build/dependencies.props`. That file is the one the templating projects use, not this one.

ryanbrandenburg: Ah, I missed that this was a PR against 2.1. I would assume that we should do 2.1.7 unless someone has reasons not to, but that just comes from a general desire to keep versions aligned rather than from specific knowledge.

dougbu:
@mkArtakMSFT / @vijayrkn should we use the 2.1.7 version of Microsoft.VisualStudio.Web.CodeGeneration.Design in the 2.1.9 release?
- @ryanbrandenburg and I aren't familiar with improvements since 2.1.3-rtm-30913 and can't say whether they're worth inclusion in the next patch release.

@ryanbrandenburg breaking up my [earlier questions](https://github.com/aspnet/AspNetCore/pull/7873#issuecomment-466515454):
> - are we tracking the issue anywhere besides [aspnet/AspNetCore-Internal#1759](https://github.com/aspnet/AspNetCore-Internal/issues/1759)?
> - what symptoms would users experience?
> - is the workaround to add the missing `<ProjectReference>` in new projects -- or something more complicated?

vijayrkn:
Yes you can use 2.1.7 version of Microsoft.VisualStudio.Web.CodeGeneration.Design in 2.1.9 release.

dougbu:
@vijayrkn "can" or "should"? The bar is fairly high in this branch and more information would help.

vijayrkn:
2.1.8 servicing shipped with 2.1.7 code generation package. 2.19 should ship with the same version as 2.1.8.

dougbu:
@vijayrkn I misspoke: Our `src/Templating/build/dependencies.props` file is only used in local builds. Official builds picked up version `2.1.7` of the code generation package from ProdCon. No need to change that `dependencies.props` file at all.

Example generated project using the 2.1.504 SDK and other 2.1.8 bits:
``` xml
<Project Sdk="Microsoft.NET.Sdk.Web">

  <PropertyGroup>
    <TargetFramework>netcoreapp2.1</TargetFramework>
    <UserSecretsId>aspnet-test-8895F3F0-B132-4C7E-A14F-93D2FA068321</UserSecretsId>
  </PropertyGroup>

  <ItemGroup>
    <None Update="app.db" CopyToOutputDirectory="PreserveNewest" />
  </ItemGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.AspNetCore.App" />
    <PackageReference Include="Microsoft.AspNetCore.Razor.Design" Version="2.1.2" PrivateAssets="All" />
    <PackageReference Include="Microsoft.EntityFrameworkCore.Sqlite" Version="2.1.8" />
    <PackageReference Include="Microsoft.VisualStudio.Web.CodeGeneration.Design" Version="2.1.7" PrivateAssets="All" />
  </ItemGroup>

</Project>
```

dougbu:
### Description
As described in aspnet/AspNetCore-Internal#1759, the templates projected for our 2.1.9 release lack references to the Microsoft.VisualStudio.Web.CodeGeneration.Design package. This affects at least the `mvc` and `razor` templates with individual auth enabled.

### Customer impact
@vijayrkn or @ryanbrandenburg what's the exact error users would see when attempting to use scaffolding without this package reference?

#### Workaround
Users may resolve the issue by adding the missing package reference:
``` xml
<PackageReference Include="Microsoft.VisualStudio.Web.CodeGeneration.Design" Version="2.1.7" PrivateAssets="All" />
```

### Regression?
Yes. Earlier 2.1.x releases did not have this issue.

ryanbrandenburg:
> @ryanbrandenburg breaking up my [earlier questions](https://github.com/aspnet/AspNetCore/pull/7873#issuecomment-466515454):
> 
> > * are we tracking the issue anywhere besides [aspnet/AspNetCore-Internal#1759](https://github.com/aspnet/AspNetCore-Internal/issues/1759)?
> > * what symptoms would users experience?
> > * is the workaround to add the missing `<ProjectReference>` in new projects -- or something more complicated?

1. No, that's the only place this is being tracked.
2. I don't know the answer to that one, would have to investigate, but hopefully @vijayrkn just knows offhand?
3. Yes, that should fix it.

dougbu:
My [comment above](https://github.com/aspnet/AspNetCore/pull/7873#issuecomment-466515454) contained a typo. I've edited it there as well.
> @mkArtakMSFT this change should **not** be necessary in `2.2.3` but it will affect users if we ship with the current templates in 2.1.9.

(I didn't change the template files as I merged this commit into `release/2.2` and `master`.)

dougbu:
Closing. Real way to address this will be test-only. Coming soon&hellip;

PR #7887, opened 2019-02-23
Elide bounds check for MethodToString

benaadams:
Use array length for check

### DIFF
--- a/src/Servers/Kestrel/Core/src/Internal/Infrastructure/HttpUtilities.cs
+++ b/src/Servers/Kestrel/Core/src/Internal/Infrastructure/HttpUtilities.cs
@@ -410,10 +410,11 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure
         }
         public static string MethodToString(HttpMethod method)
         {
-            int methodIndex = (int)method;
-            if (methodIndex >= 0 && methodIndex <= 8)
+            var methodIndex = (int)method;
+            var methodNames = _methodNames;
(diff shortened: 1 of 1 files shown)


src/Servers/Kestrel/Core/src/Internal/Infrastructure/HttpUtilities.cs
```diff
@@ -411,7 +411,7 @@ public static string VersionToString(HttpVersion httpVersion)
         public static string MethodToString(HttpMethod method)
         {
             int methodIndex = (int)method;
-            if (methodIndex >= 0 && methodIndex <= 8)
+            if ((uint)methodIndex < (uint)_methodNames.Length)
```
benaadams: @AndyAyersMS should I take this `static readonly` array to a local? (thinking re: https://github.com/dotnet/coreclr/issues/21951)

AndyAyersMS: Probably so -- otherwise I would expect the to will redo the bounds check, since it will think `_methodNames` might change.

benaadams: 👍

pakrym (approved):

benaadams:
Rebased

benaadams:
Rebased to retrigger CI

pakrym:
Helix failure: https://github.com/aspnet/AspNetCore/issues/8164

Issue #7486, opened 2019-02-12
Add doc comments for SameSiteMode enum (it has none)


Eilon:
Need to do it in two places:

1. https://github.com/aspnet/AspNetCore/blob/master/src/Http/Headers/src/SameSiteMode.cs
2. https://github.com/aspnet/AspNetCore/blob/master/src/Http/Http.Features/src/SameSiteMode.cs

PR #7907, opened 2019-02-25
Adding comments

natery2000:
Adding comments for SameSiteMode enum values

Addresses #7486

### DIFF
--- a/src/Http/Headers/src/SameSiteMode.cs
+++ b/src/Http/Headers/src/SameSiteMode.cs
@@ -3,10 +3,16 @@
 
 namespace Microsoft.Net.Http.Headers
 {
-    // RFC Draft: https://tools.ietf.org/html/draft-ietf-httpbis-cookie-same-site-00
+    /// <summary>
+    /// Determines whether to send a cookies on "same-site" or "cross-site" requests
+    /// RFC Draft: https://tools.ietf.org/html/draft-ietf-httpbis-cookie-same-site-00
+    /// </summary>
+    /// <remarks>
+    /// This mirrors Microsoft.AspNetCore.Http.SameSiteMode
+    /// </remarks>
     public enum SameSiteMode
     {
-        /// <summary>The cookie will not be sent along with "same-site" requests, with "cross-site" top-level navigations</summary>
+        /// <summary>The cookie will not be sent along with "same-site" requests or with "cross-site" top-level navigations</summary>
         None = 0,
         /// <summary>The cookie will be sent with "same-site" requests, and with "cross-site" top-level navigations</summary>
         Lax,
--- a/src/Http/Headers/src/SameSiteMode.cs
+++ b/src/Http/Headers/src/SameSiteMode.cs
@@ -4,19 +4,17 @@
 namespace Microsoft.Net.Http.Headers
 {
     /// <summary>
-    /// Determines whether to send a cookies on "same-site" or "cross-site" requests
+    /// Determines whether to send a cookies on "same-site" or "cross-site" requests.
     /// RFC Draft: https://tools.ietf.org/html/draft-ietf-httpbis-cookie-same-site-00
     /// </summary>
-    /// <remarks>
-    /// This mirrors Microsoft.AspNetCore.Http.SameSiteMode
-    /// </remarks>
+    // This mirrors Microsoft.AspNetCore.Http.SameSiteMode
     public enum SameSiteMode
     {
-        /// <summary>The cookie will not be sent along with "same-site" requests or with "cross-site" top-level navigations</summary>
+        /// <summary>The cookie will not be sent along with "same-site" requests or with "cross-site" top-level navigations.</summary>
         None = 0,
-        /// <summary>The cookie will be sent with "same-site" requests, and with "cross-site" top-level navigations</summary>
+        /// <summary>The cookie will be sent with "same-site" requests, and with "cross-site" top-level navigations.</summary>
         Lax,
-        /// <summary>The cookie will only be sent along with "same-site" requests</summary>
+        /// <summary>The cookie will only be sent along with "same-site" requests.</summary>
         Strict
     }
 }
--- a/src/Http/Headers/src/SameSiteMode.cs
+++ b/src/Http/Headers/src/SameSiteMode.cs
@@ -4,17 +4,17 @@
 namespace Microsoft.Net.Http.Headers
 {
     /// <summary>
-    /// Determines whether to send a cookies on "same-site" or "cross-site" requests.
+    /// Indicates if the client should include a cookie on "same-site" or "cross-site" requests.
     /// RFC Draft: https://tools.ietf.org/html/draft-ietf-httpbis-cookie-same-site-00
     /// </summary>
     // This mirrors Microsoft.AspNetCore.Http.SameSiteMode
     public enum SameSiteMode
     {
-        /// <summary>The cookie will not be sent along with "same-site" requests or with "cross-site" top-level navigations.</summary>
+        /// <summary>No SameSite field will be set, the client should follow its default cookie policy.</summary>
         None = 0,
-        /// <summary>The cookie will be sent with "same-site" requests, and with "cross-site" top-level navigations.</summary>
+        /// <summary>Indicates the client should send the cookie with "same-site" requests, and with "cross-site" top-level navigations.</summary>
         Lax,
-        /// <summary>The cookie will only be sent along with "same-site" requests.</summary>
+        /// <summary>Indicates the client should only send the cookie with "same-site" requests.</summary>
         Strict
     }
 }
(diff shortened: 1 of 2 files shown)


src/Http/Headers/src/SameSiteMode.cs
```diff
@@ -1,13 +1,16 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 namespace Microsoft.Net.Http.Headers
 {
     // RFC Draft: https://tools.ietf.org/html/draft-ietf-httpbis-cookie-same-site-00
     public enum SameSiteMode
```
Eilon: If you're in the mood, it would also be nice to have a doc comment for the type itself. Including a link to the RFC would be useful!

Eilon:
@natery2000 thanks for the PR! The lack of doc comments on this enum has been a recent complaint.

src/Http/Headers/src/SameSiteMode.cs
natery2000: I can include all of those changes as well. Sorry for being new, but could you help me find those docs?

Eilon: Hi @natery2000 sorry I'm not sure I understand. I'm suggesting adding a new triple-slash (`///`) doc for the `public enum` section as well. So, much like the `///` docs you already added (thank you!), it's possible to add top-level docs for the type itself. And I'm suggesting deleting the current `//` comment, and making it a `///` comment with the same RFC link that's currently there. Hopefully that makes sense 😄

natery2000: Ah yes, that makes a lot of sense. I was thinking there was a formal external documentation that this needed to be included in. I will work on this tonight, thank you very much for the help.

Eilon: Thanks again!

src/Http/Headers/src/SameSiteMode.cs
```diff
@@ -3,10 +3,16 @@
 
 namespace Microsoft.Net.Http.Headers
 {
-    // RFC Draft: https://tools.ietf.org/html/draft-ietf-httpbis-cookie-same-site-00
+    /// <summary>
+    /// Determines whether to send a cookies on "same-site" or "cross-site" requests
+    /// RFC Draft: https://tools.ietf.org/html/draft-ietf-httpbis-cookie-same-site-00
+    /// </summary>
+    /// <remarks>
+    /// This mirrors Microsoft.AspNetCore.Http.SameSiteMode
```
Eilon: I think the remarks can be omitted because they're really just "for us" - so maybe leave those as a `//` comment?

src/Http/Headers/src/SameSiteMode.cs
```diff
@@ -3,10 +3,16 @@
 
 namespace Microsoft.Net.Http.Headers
 {
-    // RFC Draft: https://tools.ietf.org/html/draft-ietf-httpbis-cookie-same-site-00
+    /// <summary>
+    /// Determines whether to send a cookies on "same-site" or "cross-site" requests
```
Eilon: Add a period at the end of this line. It'll get formatted as all in one line anyway (like HTML).

src/Http/Http.Features/src/SameSiteMode.cs
```diff
@@ -3,11 +3,16 @@
 
 namespace Microsoft.AspNetCore.Http
 {
-    // RFC Draft: https://tools.ietf.org/html/draft-ietf-httpbis-cookie-same-site-00
-    // This mirrors Microsoft.Net.Http.Headers.SameSiteMode
+    /// <summary>
+    /// Determines whether to send a cookies on "same-site" or "cross-site" requests
+    /// RFC Draft: https://tools.ietf.org/html/draft-ietf-httpbis-cookie-same-site-00
+    /// </summary>
+    /// <remarks>
```
Eilon: See 2 comments above.

Eilon (changes requested):
Looking very good! Just a couple of small suggestions.

src/Http/Headers/src/SameSiteMode.cs
natery2000: Should the single enum value level comments get this same period?

Eilon: Good catch, yes they do, thanks for fixing.

Eilon (approved):

Eilon:
Looks good to me!

@Tratcher - can you do a quick once-over and merge when ready?

src/Http/Http.Features/src/SameSiteMode.cs
```diff
@@ -1,14 +1,20 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 namespace Microsoft.AspNetCore.Http
 {
-    // RFC Draft: https://tools.ietf.org/html/draft-ietf-httpbis-cookie-same-site-00
+    /// <summary>
+    /// Determines whether to send a cookies on "same-site" or "cross-site" requests.
```
Tratcher: This phrasing is a bit misleading. This enum is used when generating response cookies and controls adding said fields to the cookie. It does not control the servers behavior regarding cookies. The client is the one that reacts to the values and decides if it should return the cookie on future requests.

```suggestion
    /// Used to set the SameSite field on response cookies to indicate if those cookies should be included by the client on future "same-site" or "cross-site" requests.
```

src/Http/Http.Features/src/SameSiteMode.cs
```diff
@@ -1,14 +1,20 @@
...
+    /// RFC Draft: https://tools.ietf.org/html/draft-ietf-httpbis-cookie-same-site-00
+    /// </summary>
     // This mirrors Microsoft.Net.Http.Headers.SameSiteMode
     public enum SameSiteMode
     {
+        /// <summary>The cookie will not be sent along with "same-site" requests or with "cross-site" top-level navigations.</summary>
```
Tratcher:
```suggestion
        /// <summary>No SameSite field will be set, the client should follow its default cookie policy.</summary>
```

src/Http/Http.Features/src/SameSiteMode.cs
```diff
@@ -1,14 +1,20 @@
...
         None = 0,
+        /// <summary>The cookie will be sent with "same-site" requests, and with "cross-site" top-level navigations.</summary>
```
Tratcher:
```suggestion
        /// <summary>Indicates the client should send the cookie with "same-site" requests, and with "cross-site" top-level navigations.</summary>
```

src/Http/Http.Features/src/SameSiteMode.cs
```diff
@@ -1,14 +1,20 @@
...
         Lax,
+        /// <summary>The cookie will only be sent along with "same-site" requests.</summary>
```
Tratcher:
```suggestion
        /// <summary>Indicates the client should only send the cookie with "same-site" requests.</summary>
```

src/Http/Headers/src/SameSiteMode.cs
```diff
@@ -1,13 +1,20 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 namespace Microsoft.Net.Http.Headers
 {
-    // RFC Draft: https://tools.ietf.org/html/draft-ietf-httpbis-cookie-same-site-00
+    /// <summary>
+    /// Determines whether to send a cookies on "same-site" or "cross-site" requests.
```
Tratcher: The context for this type is slightly different from the other one (see my comments for the other one), it's for the header parser that could be used by a client or server. However it's still only ever used with response cookies, these fields aren't included with requests.

```suggestion
    /// Indicates if the client should include a cookie on "same-site" or "cross-site" requests.
```

src/Http/Headers/src/SameSiteMode.cs
```diff
@@ -1,13 +1,20 @@
...
+    /// RFC Draft: https://tools.ietf.org/html/draft-ietf-httpbis-cookie-same-site-00
+    /// </summary>
+    // This mirrors Microsoft.AspNetCore.Http.SameSiteMode
     public enum SameSiteMode
     {
+        /// <summary>The cookie will not be sent along with "same-site" requests or with "cross-site" top-level navigations.</summary>
```
Tratcher: See my comments from the other type.

Tratcher (approved):

natery2000:
Are these seemingly random failing tests anything of concern?

JamesNK:
They're unrelated flakey tests. I'll re-run the build and hopefully you get a pass and can merge.

Tratcher:
Thanks.

PR #7914, opened 2019-02-25
[Templates] Imports and CSS tweaks for forms/validation

SteveSandersonMS:
Following @rynowak's app-building/verification exercise, this PR:

* Adds `Microsoft.AspNetCore.Components.Forms` to the default `_ViewImports.cshtml` in the Components/Blazor templates
* Adds initial CSS styling for the build-in forms/validation classes so they have a visual effect

These changes are extremely low-risk. They affect the template only, and only in trivial ways. The value of this change is that it cleans up the developer experience around forms and validation, which is one of the key new Components features for Preview 3.


SteveSandersonMS:
@mkArtakMSFT Do you know if we're still able to take PRs like this one? It's not that anything's badly broken without it, it's just that this improves the developer experience substantially and is very low risk (just CSS and `@using` changes).

cc @danroth27

rynowak (approved):
:shipit: from me - if we can't take it for preview3 then we should take this to master

Eilon (approved):
Approved for preview3.

PR #7915, opened 2019-02-25
Make InputBase use EventCallback for ValueChanged

SteveSandersonMS:
... so that the host component gets re-rendered automatically after each value change (as also happens when binding to DOM elements).

This is another fit-and-finish issue discovered by @rynowak during his app-building exercise.

### DIFF
--- a/src/Components/Components/src/Forms/InputComponents/InputBase.cs
+++ b/src/Components/Components/src/Forms/InputComponents/InputBase.cs
@@ -42,7 +42,7 @@ namespace Microsoft.AspNetCore.Components.Forms
         /// <summary>
         /// Gets or sets a callback that updates the bound value.
         /// </summary>
-        [Parameter] Action<T> ValueChanged { get; set; }
+        [Parameter] EventCallback<T> ValueChanged { get; set; }
 
         /// <summary>
         /// Gets or sets an expression that identifies the bound value.
@@ -71,7 +71,7 @@ namespace Microsoft.AspNetCore.Components.Forms
                 if (hasChanged)
                 {
                     Value = value;
-                    ValueChanged?.Invoke(value);
+                    _ = ValueChanged.InvokeAsync(value);
                     EditContext.NotifyFieldChanged(FieldIdentifier);
                 }
             }
--- a/src/Components/test/E2ETest/Tests/FormsTest.cs
+++ b/src/Components/test/E2ETest/Tests/FormsTest.cs
@@ -317,12 +317,12 @@ namespace Microsoft.AspNetCore.Components.E2ETest.Tests
         }
 
         [Fact]
-        public async Task InputComponentsCauseContainerToRerenderOnChange()
+        public void InputComponentsCauseContainerToRerenderOnChange()
         {
             var appElement = MountTestComponent<TypicalValidationComponent>();
             var ticketClassInput = new SelectElement(appElement.FindElement(By.ClassName("ticket-class")).FindElement(By.TagName("select")));
             var selectedTicketClassDisplay = appElement.FindElement(By.Id("selected-ticket-class"));
-            var select = ticketClassInput.WrappedElement;
+            var messagesAccessor = CreateValidationMessagesAccessor(appElement);
 
             // Shows initial state
             WaitAssert.Equal("Economy", () => selectedTicketClassDisplay.Text);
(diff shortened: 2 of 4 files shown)


SteveSandersonMS:
Again, I don't know whether it's still possible to include items like this for preview 3, but here's the PR anyway.

The value of including this is that is makes the new built-in input components (`InputText`, `InputSelect`, etc.) consistent with regular DOM elements in terms of what gets re-rendered after each edit. Without this PR, edits in the input components won't automatically re-render the container component, which is strange and can be very inconvenient.

src/Components/test/E2ETest/Tests/FormsTest.cs
```diff
@@ -316,6 +316,27 @@ public void ValidationMessageDisplaysMessagesForField()
...
+            // Shows initial state
+            WaitAssert.Equal("Economy", () => selectedTicketClassDisplay.Text);
+
+            // Refreshes on edit
+            ticketClassInput.SelectByValue("Premium");
+            WaitAssert.Equal("Premium", () => selectedTicketClassDisplay.Text);
+
+            // Leaves previous value unchanged if new entry is unparseable
+            ticketClassInput.SelectByText("(select)");
+            await Task.Delay(500); // Not expecting any UI change, so wait a moment to see if one happens
```
davidfowl: No better way to test this?

SteveSandersonMS: The inherent asynchrony of the server-side execution model makes it tricky for tests to assert "something shouldn't change". It's easy to assert when something should change within a given time span - that's what the `WaitAssert` calls do - but proving a negative is harder.

We *could* add some extra complications to the code under test for cases like this, whereby we try to trigger a secondary effect that we can wait for before running the "nothing changed" assertion. But it's a very unnatural way to write test code.

There are few cases where we really need "nothing changed" assertions in the Components E2E tests, so I haven't regarded it as a major issue. The same test cases also run automatically in the client-side (Blazor) runtime which is fully synchronous, so the assertion is guaranteed to be applied to the post-updates state anyway, meaning that we'd almost definitely catch any cases where the asynchony was hiding a bug.

If we later decide that we really need to solve this in a better way and not rely on waiting like this, I'd prefer to do it once in some central place, by having the test app somehow expose information about the number of renders, so the test code can wait until that number increments.

rynowak: The E2E tests have a bunch of *wait until* loops and so on. 

In general waiting for someone not to happen is just problematic since it will fail to validate what you want if slow.

src/Components/Components/src/Forms/InputComponents/InputBase.cs
```diff
@@ -71,7 +71,7 @@ protected T CurrentValue
                 if (hasChanged)
                 {
                     Value = value;
-                    ValueChanged?.Invoke(value);
+                    _ = ValueChanged.InvokeAsync(value);
```
rynowak: Because this is fire and forget, it's possible for no one to see the error if one happens. 

I've considered in the past whether we should have a `HandleError(Task)` method on `ComponentBase` in case you find yourself in one of these situations. 

It doesn't seem in scope for this PR? but thoughts in general?

src/Components/test/E2ETest/Tests/FormsTest.cs
rynowak: You could add validation and look for the `invalid` class. Not sure if that's a good fit or not for what's being tested.

rynowak: I guess you have validation. So you could wait for the select to have `invalid` and then look at `selectedTicketClassDisplay`

src/Components/Components/src/Forms/InputComponents/InputBase.cs
davidfowl: Yes we should avoid unobserved tasks in general.

SteveSandersonMS: Yes, I also agree we should explicitly handle this somehow, and the `ComponentBase.HandleError` approach sounds promising.

That's well out of scope for this PR, of course!

Also it's not a scenario that's going to affect conventional uses of `bind-*`, since with those, the "setter" code we generate is always synchronous anyway. You couldn't return a `Task` to the system even if you wanted to.

src/Components/test/E2ETest/Tests/FormsTest.cs
SteveSandersonMS: Great point. Done.

rynowak (approved):

Eilon (approved):
Approved for preview3.

PR #7919, opened 2019-02-25
Remove VisualStudioExtension as a required workload

natemcmaster:
As a result of changes @pranavkm made recently to the blazor VSIX build, this VS workload is not required anymore. This unblockes us to build with the BuildTools SKU.


pranavkm (approved):
As long as it works!

dougbu (approved):
Can this go in before we update to use the new MSBuild++ queues?

PR #7921, opened 2019-02-25
HeaderPropagation: propagate incoming request headers to outgoing HTTP requests

alefranz:
This PR introduces a `MessageHandler` for the `HttpClientFactory` to forward headers received in the current MVC request.

It is a port of https://github.com/aspnet/Extensions/pull/1099 which 
Addresses https://github.com/aspnet/Extensions/issues/526

### Main feedback on the original PR
@natemcmaster : it can't be added to https://github.com/aspnet/Extensions as it introduce a circulr dependency
@rynowak has suggested to move this to this repo as middleware.
@Tratcher it is not handling concurrency in accessing the Context if multiple HttpClient requests are sent in parallel
@Tratcher @rynowak it should not need to be based on the HttpClientFactory but it could be a `DelegatingHandler` and a `Middleware` to initialize the state.

### Current Plan
- Collect feedback on this PR from relevant stake holders (should I open a issue as well?)
- In the meantime, I'll work on the `DelegatingHandler` with a initializer middleware, locking the access to the Context if needed, similar to https://github.com/Microsoft/ApplicationInsights-aspnetcore/issues/373

### Original description

I believe it is a common use case which deserves to be included in the `Extensions` namespace, so I'd like to help to land this feature. Its main use case is probably to track distributed transaction which requires the ability to pass through a transaction identifier as well as generating a new one when not present.
The code is loosely based on this [gist](https://gist.github.com/davidfowl/c34633f1ddc519f030a1c0c5abe8e867) from @davidfowl with some extra features and a different behaviour as, by default, this PR doesn't add the header if it's already present in the `HttpRequestMessage`.

Features:

* Forward header from current request MVC context to outgoing `HttpClient` requests, allowing to use a different header name. The header is forwarded only if present and all values are used.
* Always forward the header to the `HttpRequestMessage` (`AlwaysAdd=true`) or only when the header is not already present (default).
* Define a default value(s) to be added to the `HttpRequestMessage` if the incoming request does not contain the header
* Define a default value(s) generator to be used to add the header value to the `HttpRequestMessage` if the incoming request does not contain the header. The generator can rely on the `HttpRequestMessage` or `HttpContext` to decide the value. _(note: in my mind the use case is to generate a new GUID/identifier for a transaction chain if it's not in the incoming request, I'm not sure if having access to the request and context add any value)_

_The [tests](https://github.com/alefranz/Extensions/blob/8d75c26b5167adc6730152836d631ea5202689ba/src/HttpClientFactory/HeaderPropagation/test/HeaderPropagation.Tests/HeaderPropagationTest.cs) should clarify the behaviour if needed._

It definitely needs some polishing on the definition of the configuration as well as probably removing some of the different behaviours, however I believe opening a PR is the best way to move forward the discussion and shape the behaviour.
Probably the discussion on the requirements should stay on the [issue](https://github.com/aspnet/Extensions/issues/526) to keep the focus of the discussion here on the implementation.

Please Note:

* I didn't have explicit feedback to open the PR following my [comment](https://github.com/aspnet/Extensions/issues/526#issuecomment-459548541) on the issue, however the issue if marked as `up-for-grabs`
* Part of this PR is based on this [gist](https://gist.github.com/davidfowl/c34633f1ddc519f030a1c0c5abe8e867) from @davidfowl , I'm not sure what is the correct way to give attribution.

I really appreciate you taking the time to review this PR.

Thanks,
Alessio


src/Middleware/HeaderPropagation/src/DependencyInjection/HeaderPropagationExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+            services.Configure(configure);
+            services.TryAddEnumerable(ServiceDescriptor.Singleton<IHttpMessageHandlerBuilderFilter, HeaderPropagationMessageHandlerBuilderFilter>());
+            return services;
+        }
+
+        public static IHttpClientBuilder AddHeaderPropagation(this IHttpClientBuilder builder, Action<HeaderPropagationOptions> configure)
+        {
+            builder.Services.AddHttpContextAccessor();
+            builder.Services.Configure(configure);
+            builder.AddHttpMessageHandler((sp) =>
```
Tratcher: Shouldn't this use the type overload? `AddHttpMessageHandler<HeaderPropagationMessageHandler>();`

src/Middleware/HeaderPropagation/src/HeaderPropagationMessageHandler.cs
```diff
@@ -0,0 +1,54 @@
...
+            _contextAccessor = contextAccessor;
+        }
+
+        protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+        {
+            if (_contextAccessor.HttpContext != null)
+            {
+                foreach (var header in _options.Headers)
+                {
+                    if (!_contextAccessor.HttpContext.Request.Headers.TryGetValue(header.InputName, out var values)
```
Tratcher: This didn't fix the concurrency issue, these headers were supposed to be captured in an AspNetCore middleware.

src/Middleware/HeaderPropagation/src/Microsoft.AspNetCore.Http.HeaderPropagation.csproj
```diff
@@ -0,0 +1,24 @@
...
+    </Description>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <IsAspNetCoreApp>true</IsAspNetCoreApp>
+    <NoWarn>$(NoWarn);CS1591</NoWarn>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <PackageTags>aspnetcore;httpclient</PackageTags>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <InternalsVisibleTo Include="Microsoft.AspNetCore.Http.HeaderPropagation.Tests" />
```
Tratcher: Our middleware packages all drop the `.Http`.

src/Middleware/HeaderPropagation/src/DependencyInjection/HeaderPropagationExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+using Microsoft.Extensions.Http.HeaderPropagation;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    public static class HeaderPropagationExtensions
+    {
+        public static IServiceCollection AddHeaderPropagation(this IServiceCollection services, Action<HeaderPropagationOptions> configure)
+        {
+            services.AddHttpContextAccessor();
```
Tratcher: IHttpContextAccessor shouldn't be needed of you're going to properly avoid the concurrency issue.

Tratcher (changes requested):

src/Middleware/HeaderPropagation/src/HeaderPropagationMessageHandler.cs
alefranz: Sure, I'll implement the middleware.
Apologies if it wasn't clear in the description, I had just moved the PR as-is.

src/Middleware/HeaderPropagation/src/DependencyInjection/HeaderPropagationExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+            services.Configure(configure);
+            services.TryAddEnumerable(ServiceDescriptor.Singleton<IHttpMessageHandlerBuilderFilter, HeaderPropagationMessageHandlerBuilderFilter>());
```
rynowak: I'm not certain that we should plug this into all places in HttpClient factory by default. I think it would be reasonable to take out the handler builder filter and then let users plug in the handler themselves.

The problems I see with the current approach:
- it's really hard to take out something that's "on by default" for one http client configuration
- this approach doesn't let the app developer configure the ordering of message handlers, which can be important

alefranz: Yes agree! the default registration is so confusing that was also throwing me off in the tests as it was also registering on the test client 😅

alefranz:
I've introduced the Middleware to solve the concurrency issue.
It's also not tightly couple with the `HttpClientFactory`, although it still include an extension method to easily add the `DelegeatingHandler` to the client setup in the factory, as an opt-in for each client.
The configuration, however, is shared for all the clients. This is to simplify the behavior of the middleware, as it needs to know all the possible headers to be forwarded.

The `Middleware` take the `input` headers and put those in the `state` with the name defined as `output`, eventually generating/using the default value when not present.
The `DelegatingHandler` simply add the headers from the state to the outgoing `HttpRequestMessage` headers, eventually skipping the one already present based on the value of `AlwaysAdd`.

I'm not convinced about the configuration... the behavior it's probably not really straightforward and can be simplified.

src/Middleware/HeaderPropagation/src/DependencyInjection/HeaderPropagationExtensions.cs
```diff
@@ -0,0 +1,38 @@
+using System;
```
rynowak: All new files will need the copyright notice (copy from other examples).

src/Middleware/HeaderPropagation/src/DependencyInjection/HeaderPropagationExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+using Microsoft.AspNetCore.Builder;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.AspNetCore.HeaderPropagation
+{
+    public static class HeaderPropagationExtensions
```
rynowak: We'll want docs on all new public API (once design is finished).

src/Middleware/HeaderPropagation/src/HeaderPropagationMiddleware.cs
```diff
@@ -0,0 +1,60 @@
...
+                throw new ArgumentNullException(nameof(options));
+            }
+            _options = options.Value;
+
+            _state = state ?? throw new ArgumentNullException(nameof(state));
+        }
+
+        public Task Invoke(HttpContext context)
+        {
+            if (context != null)
```
rynowak: context won't ever be null here legally, this should be an exception (or just ignored)

src/Middleware/HeaderPropagation/src/HeaderPropagationMiddleware.cs
```diff
@@ -0,0 +1,60 @@
...
+            {
+                foreach (var header in _options.Headers)
+                {
+                    if (!context.Request.Headers.TryGetValue(header.InputName, out var values)
```
rynowak: I think this logic would be clearer if you reversed this condition, and didn't nest `if`s. It took me a few passes through to be sure that it was doing what you wanted.

src/Middleware/HeaderPropagation/src/HeaderPropagationMessageHandler.cs
```diff
@@ -0,0 +1,43 @@
...
+            _state = state ?? throw new ArgumentNullException(nameof(state));
+        }
+
+        protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request,
+            CancellationToken cancellationToken)
+        {
+            foreach (var header in _options.Headers)
+            {
+                if (_state.Headers.TryGetValue(header.OutputName, out var values) &&
+                    !StringValues.IsNullOrEmpty(values) &&
```
rynowak: @Tratcher - can you think of a case where a *present but empty* header is meaningful? I see the checks for null or empty and I wonder...

src/Middleware/HeaderPropagation/src/DependencyInjection/HeaderPropagationExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+                var state = services.GetRequiredService<HeaderPropagationState>();
+                return new HeaderPropagationMessageHandler(services.GetRequiredService<IOptions<HeaderPropagationOptions>>(), state);
+            });
+
+            return builder;
+        }
+
+        public static IApplicationBuilder UseHeaderPropagation(this IApplicationBuilder builder)
+        {
+            return builder.UseMiddleware<HeaderPropagationMiddleware>();
```
rynowak: This should check for `HeaderPropagationState` to be registered in services and throw a friendly error reminding the user to call `AddHeaderPropagation`. This is something we're starting to do for all middleware to make mistakes more figure-out-able.

src/Middleware/HeaderPropagation/src/DependencyInjection/HeaderPropagationExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+
+            return services;
+        }
+
+        public static IHttpClientBuilder AddHeaderPropagation(this IHttpClientBuilder builder)
+        {
+            builder.Services.TryAddSingleton<HeaderPropagationState>();
+            builder.Services.TryAddTransient<HeaderPropagationMessageHandler>();
+ 
+            builder.AddHttpMessageHandler(services =>
```
rynowak: Is there a reason why this isn't using `AddHttpMessageHandler<HeaderPropagationMessageHandler>();`?

src/Middleware/HeaderPropagation/src/HeaderPropagationEntry.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace Microsoft.AspNetCore.HeaderPropagation
+{
+    public class HeaderPropagationEntry
+    {
+        public string InputName { get; set; }
+        public string OutputName { get; set; }
+        public StringValues DefaultValues { get; set; }
+        public Func<HttpContext, StringValues> DefaultValuesGenerator { get; set; }
+        public bool AlwaysAdd { get; set; }
```
rynowak: Do you have a use case in mind for `AlwaysAdd`?  Usually you have a global default (propagate this header) and allow local overrides (specifying the header explicitly). Why do you think this is needed?

src/Middleware/HeaderPropagation/src/HeaderPropagationEntry.cs
```diff
@@ -0,0 +1,15 @@
+using System;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNetCore.HeaderPropagation
+{
+    public class HeaderPropagationEntry
+    {
+        public string InputName { get; set; }
+        public string OutputName { get; set; }
```
rynowak: Suggestion for names `InboundHeaderName`, `OutboundHeaderName`

src/Middleware/HeaderPropagation/src/HeaderPropagationState.cs
```diff
@@ -0,0 +1,19 @@
+using System.Collections.Generic;
+using System.Threading;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNetCore.HeaderPropagation
+{
+    public class HeaderPropagationState
+    {
+        private static AsyncLocal<Dictionary<string, StringValues>> _headers { get; } = new AsyncLocal<Dictionary<string, StringValues>>();
```
rynowak: There's no real need for this to be a property.

src/Middleware/HeaderPropagation/src/HeaderPropagationState.cs
```diff
@@ -0,0 +1,19 @@
...
+
+        public Dictionary<string, StringValues> Headers
```
rynowak: @Tratcher - is there another type we want to use for this or is this the best collection type for headers?

rynowak: I was thinking about `IHeaderDictionary` but it's not obvious we want to use that here.

src/Middleware/HeaderPropagation/src/HeaderPropagationState.cs
```diff
@@ -0,0 +1,19 @@
...
+        {
+            get
+            {
+                return _headers.Value ?? (_headers.Value = new Dictionary<string, StringValues>());
```
rynowak: Comparer? @Tratcher do we use ignore case?

src/Middleware/HeaderPropagation/src/HeaderPropagationOptions.cs
```diff
@@ -0,0 +1,9 @@
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.HeaderPropagation
+{
+    public class HeaderPropagationOptions
+    {
+        public IList<HeaderPropagationEntry> Headers { get; set; } = new List<HeaderPropagationEntry>();
+    }
+}
```
rynowak: The drawback of using a list is that you can express duplicates (same outbound name). I'm not sure that it's really a problem that needs to be solved.

src/Middleware/HeaderPropagation/src/HeaderPropagationState.cs
```diff
@@ -0,0 +1,19 @@
+using System.Collections.Generic;
+using System.Threading;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNetCore.HeaderPropagation
+{
+    public class HeaderPropagationState
```
rynowak: Since this is a public type I'd like to consider different names for this. Open to suggestions, my ideas:

`HeaderPropagationValues`

`PropagatedHeaders`

`PropagatedHeaderValues`.

src/Middleware/HeaderPropagation/test/HeaderPropagationIntegrationTest.cs
```diff
@@ -0,0 +1,94 @@
...
+            Assert.Equal(new[] { "test" }, handler.Headers.GetValues("out"));
+        }
+
+        private IWebHostBuilder CreateBuilder(Action<HeaderPropagationOptions> configure, HttpMessageHandler primaryHandler)
+        {
+            return new WebHostBuilder()
+                .Configure(app =>
+                {
+                    app.UseHeaderPropagation();
+                    app.Map("", x => x.UseMiddleware<SimpleMiddleware>());
```
rynowak: Does map here really do anything?

src/Middleware/HeaderPropagation/test/HeaderPropagationMessageHandlerTest.cs
```diff
@@ -0,0 +1,152 @@
...
+    {
+        public HeaderPropagationMessageHandlerTest()
+        {
+            Handler = new SimpleHandler();
+
+            State = new HeaderPropagationState();
+            Options = new HeaderPropagationOptions();
+
+            var headerPropagationMessageHandler =
+                new HeaderPropagationMessageHandler(new OptionsWrapper<HeaderPropagationOptions>(Options), State)
```
rynowak: You can also use `Options.Create`, it's a little less verbose`

src/Middleware/HeaderPropagation/test/HeaderPropagationMessageHandlerTest.cs
```diff
@@ -0,0 +1,152 @@
...
+
+            protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request,
+                CancellationToken cancellationToken)
+            {
+                Headers = request.Headers;
+                return Task.FromResult(new HttpResponseMessage());
+            }
+        }
+    }
+}
```
rynowak: Good coverage here 👍

src/Middleware/HeaderPropagation/src/HeaderPropagationMiddleware.cs
```diff
@@ -0,0 +1,60 @@
...
+                        {
+                            values = header.DefaultValues;
+                        }
+                        else
+                        {
+                            continue;
+                        }
+                    }
+
+                    var outputName = !string.IsNullOrEmpty(header.OutputName) ? header.OutputName : header.InputName;
```
rynowak: It doesn't look like the message handler handles this case. It will try to use `OutputName` regardless.

src/Middleware/HeaderPropagation/test/HeaderPropagationMiddlewareTest.cs
```diff
@@ -0,0 +1,184 @@
...
+                DefaultValuesGenerator = ctx => StringValues.Empty
+            });
+
+            // Act
+            await Middleware.Invoke(Context);
+
+            // Assert
+            Assert.DoesNotContain("out", State.Headers.Keys);
+        }
+    }
```
rynowak: Good coverage here as well.

rynowak (commented):
Looks like this is on track! Some minor feedback.

rynowak:
@stevejgordon HYPE

src/Middleware/HeaderPropagation/src/HeaderPropagationMessageHandler.cs
Tratcher: The only case I've seen is the Host header. You're required to have one but it may be empty.

Tratcher:
Wasn't there some earlier discussion about collecting many headers up front but choosing in the outgoing handler which headers applied to that scenario? Either wanting to customize it per outgoing request or at least being able to have multiple handlers with different outgoing headers? 

The current design frontloads everything in the collecting middleware, there doesn't seem to be a way to customize it later.

ericsampson:
Really glad this is getting worked, thanks @alefranz ! Is this going to be available in 2.1/2.2, or just 3.0? Is there a way to access the value of a generated header when handling the incoming request (like if you want to log it/stash it somewhere)?

ericsampson:
Is there going to be a way to wire up one of the incoming (or generated if missing) headers to be the TraceIdentifier? At a higher level, is the type of functionality currently available in something like @stevejgordon's https://github.com/stevejgordon/CorrelationId package going to be in scope here (or elsewhere)? It seems like a natural fit. Thanks everyone!

alefranz:
Apologies everyone for the lack of progress. I'll get back to this tonight.
@rynowak @Tratcher I've really appreciated your valuable feedback. I will address it.

@ericsampson yes, a scenario like the one from the library is already supported in the current form. You will be able to setup the middleware to capture the TraceIdentifier and have it automatically added to outgoing request. A GUID for new requests can then be generated using the `DefaultValuesGenerator`. Am i missing something else?

@Tratcher yes it has been mentioned, and yes I believe it should be the case. However I wasn't sure on the best approach to handle the configuration for it and I would like to know your point of view. As the configuration object will be one, it means it can be customized for each client that use `AddHeaderPropagation`, and the overload with an action to "Configure" is not the correct approach as it will modify the shared configuration. Should I provide an overload to accept a configuration object to use for the particular client/handler (eventually with a configurationbuilder to simplify it)? However that would mean it would be only possible to configure it in code, not through the standard configuration infrastructure. Is there a different pattern to follow for this scenario?
Also as it would be nice to keep the configuration at minimum and avoid possible errors. Setting up the headers in the middleware and then in the handler specify which ones to use is probably the easier solution but offer a bit of duplication, which maybe can be alleviated making the default behaviour to pass all the headers, so the "allow list" is opt-in. The alternative approach I wanted in the first place was was to define the list of headers when setting up the clients and then having the middleware pulling all this configurations to decide which headers to collect. This allow to specify the config in the client/handler, where it feels more natural, as well as defining different outgoing header names per client however it makes the implementation less clean and it creates ambiguity on the defaultvalues generation (per outgoing request or shared). Sorry I'm just thinking out loud, not sure if it makes sense.

@rynowak about AlwaysAdd, in the current implementation it's a awkward and not really fulfilling it's purpose. I need to rethink that. The scenario I had in mind was to track a chain of requests (so adding a new id to the list of incoming Ids) or the `tracestate` in the [W3C `trace-context` draft spec](https://w3c.github.io/trace-context/) where you have to chain with identifier other vendors identifiers. I think a better place to handle this would be in the middleware to have something to control the behavior to generate a new value even when one is in the incoming requests, as I believe in this scenarios it is a better approach to generate a new id that will be used by all the outgoing HTTP calls as part of that requests. Anyway this will not necessary be the desired behavior for everyone, as well as there are other aspects, like if the new value should be appended or put in front, as well if it should be multiple heaver values or a single value comma separated (as it is required for `tracestate`). Probably a flexible solution to handle all the possible scenario will increase complexity without bringing much values, so I'm tempted to leave this out for now (as it can eventually be added later as a non breaking change) eventually adding some extensibility points to be able to inject a behavior for the specific scenario. What do you think?

Tratcher:
@rynowak do you have a pattern for using named options with handlers added to the HttpClientFactory?

ericsampson:
@alefranz , the CorrelationId library has the ability to overwrite the default TraceIdentifier with the value of a given header from the incoming message, that's what I was wondering about. Also, I'm still curious if it's going to be possible to access the value from the DefaultValuesGenerator in the controller that handled the incoming message?

src/Middleware/HeaderPropagation/test/HeaderPropagationIntegrationTest.cs
alefranz: Removed!

src/Middleware/HeaderPropagation/src/HeaderPropagationState.cs
alefranz: I was tempted to use somethingHeaders (like `PropagatedHeaders` you suggested) but then `PropagatedHeaders.Headers` feel awkward. Renaming the `Dictionary` to `Values` is not ideal either as it has a `Values` property :)

Anyway I don't have a particular preference, what's your pick? and what should be the property name?

src/Middleware/HeaderPropagation/src/HeaderPropagationMiddleware.cs
alefranz: What do you think now? Open for suggestions

src/Middleware/HeaderPropagation/src/HeaderPropagationState.cs
alefranz: Should this property be internal?

src/Middleware/HeaderPropagation/src/HeaderPropagationState.cs
alefranz: true, corrected! thanks!

src/Middleware/HeaderPropagation/src/HeaderPropagationEntry.cs
alefranz: see https://github.com/aspnet/AspNetCore/pull/7921#issuecomment-470457767

src/Middleware/HeaderPropagation/src/DependencyInjection/HeaderPropagationExtensions.cs
alefranz: I've copyed the message from https://github.com/aspnet/AspNetCore/blob/85e2147ff037cc9950249fbb9d095ad47ec4184a/src/Http/Routing/src/Resources.resx#L199.
Is there a better approach?

src/Middleware/HeaderPropagation/src/HeaderPropagationMiddleware.cs
alefranz: Removed check

src/Middleware/HeaderPropagation/src/DependencyInjection/HeaderPropagationExtensions.cs
alefranz: fixed

src/Middleware/HeaderPropagation/src/HeaderPropagationState.cs
alefranz: Done

src/Middleware/HeaderPropagation/test/HeaderPropagationMessageHandlerTest.cs
alefranz: ok

src/Middleware/HeaderPropagation/src/HeaderPropagationMiddleware.cs
alefranz: You are right, I fixed it doing a change:
I removed the logic in the middleware and saved in the "state" using the input name, then moved the logic in the handler.
This is because to address the ability to specify the headers to be propagated for each client, there will be a part of config needed for the `DelegatingHandler` and I believe it will be the right place to also specify the output name (so calls to different services can use different header names).

src/Middleware/HeaderPropagation/src/DependencyInjection/HeaderPropagationExtensions.cs
```diff
@@ -0,0 +1,90 @@
...
+                throw new ArgumentNullException(nameof(configureOptions));
+            }
+
+            services.Configure(configureOptions);
+            services.AddHeaderPropagation();
+
+            return services;
+        }
+
+        public static IHttpClientBuilder AddHeaderPropagation(this IHttpClientBuilder builder)
```
rynowak: Docs here.

src/Middleware/HeaderPropagation/src/HeaderPropagationEntry.cs
rynowak: > I'm tempted to leave this out for now (as it can eventually be added later as a non breaking change) eventually adding some extensibility points to be able to inject a behavior for the specific scenario. What do you think?

Yeah, I'd suggest leaving this out. Users will ask how to do it if it's important.

Another suggestion that I have for this (and other needs) - replace `DefaultValuesGenerator` with `ValueFactory` - if it's set, then you just always call it to get the output value. I think that solves all of the cases you were interested in, and also can fill the niche that it currently does.

src/Middleware/HeaderPropagation/src/HeaderPropagationState.cs
rynowak: I'd pick `HeaderPropagationValues` of those.

rynowak: I think `Headers` is fine for the name

src/Middleware/HeaderPropagation/src/HeaderPropagationState.cs
rynowak: No, I think this is fine.

src/Middleware/HeaderPropagation/src/HeaderPropagationMiddleware.cs
rynowak: Looks good

src/Middleware/HeaderPropagation/src/HeaderPropagationMessageHandler.cs
rynowak: OK, well that's one we deffo don't care about propagating.

src/Middleware/HeaderPropagation/src/DependencyInjection/HeaderPropagationExtensions.cs
rynowak: Loooks good

src/Middleware/HeaderPropagation/src/HeaderPropagationState.cs
```diff
@@ -0,0 +1,29 @@
...
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNetCore.HeaderPropagation
+{
+    /// <summary>
+    /// Contains the headers values for the <see cref="HeaderPropagationMiddleware"/>.
+    /// </summary>
+    public class HeaderPropagationState
+    {
+        private static AsyncLocal<Dictionary<string, StringValues>> _headers = new AsyncLocal<Dictionary<string, StringValues>>();
```
rynowak: readonly too

rynowak:
> Wasn't there some earlier discussion about collecting many headers up front but choosing in the outgoing handler which headers applied to that scenario? Either wanting to customize it per outgoing request or at least being able to have multiple handlers with different outgoing headers?

Does it make sense to try and lock down what's currently in this PR and then send another to consider scenarios like this?


> do you have a pattern for using named options with handlers added to the HttpClientFactory?

We don't have a specific pattern, but you get access to services when creating a handler, so things like this can be built easily.

src/Middleware/HeaderPropagation/src/HeaderPropagationState.cs
alefranz: done

src/Middleware/HeaderPropagation/src/DependencyInjection/HeaderPropagationExtensions.cs
alefranz: done

src/Middleware/HeaderPropagation/src/HeaderPropagationOptions.cs
alefranz: I had decided to change it to a dictionary, as overriding arrays in config is always a bit messy. Also the ordering doesn't add much value imo. However I used the input name as key so it doesn't address the potential issue you raised.
The reason why I used the input name is that the input name is the one relevant in the middleware. The output name is only used by the Handler, is optional and I believe it should be moved to be moved to the handler configuration as part as handling the ability to specify which handler to include per client (in a following PR).

src/Middleware/HeaderPropagation/src/HeaderPropagationState.cs
alefranz: done

src/Middleware/HeaderPropagation/src/HeaderPropagationEntry.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// <summary>
+        /// Gets or sets the value factory to be used.
+        /// It gets as input the inbound header name for this entry as defined in the <see cref="HeaderPropagationOptions"/>
+        /// and the <see cref="HttpContext"/> of the current request.
+        /// </summary>
+        /// <remarks>
+        /// When present, the factory is the only method used to set the value.
+        /// To not add the header, return <see cref="StringValues.Empty"/>.
+        /// </remarks>
+        public Func<string, HttpContext, StringValues> ValueFactory { get; set; }
```
alefranz: @rynowak In changing this to a ValueFactory, I've also added the inbound header name passed to the value factory, what do you think?
It simplify reusing the same value factory for multiple headers, but it's not really a strong argument.

alefranz:
from my side, I'm happy to try to get this PR locked down and then raise it a new one to add the ability to specify the headers to include on a per client basis.

rynowak:
@alefranz - sorry for the delay on this. It's MVP summit this week so the team is pretty busy. I'll take a look at this again in detail as soon as I can.

alefranz:
@ryonwak thank you for the update. Enjoy the summit :)

rynowak:
FYI - you're running into this build error:

> One or more errors occurred. (One or more errors occurred. (The process cannot access the file 'E:\A\_work\1\s\src\Components\test\E2ETest\node_modules\selenium-standalone\.selenium\selenium-server\3.141.5-server.jar' because it is being used by another process.)) (The process cannot access the file 'E:\A\_work\1\s\src\Components\test\E2ETest\node_modules\selenium-standalone\.selenium\selenium-server\3.141.5-server.jar' because it is being used by another process.)

That's unrelated to your changes, and can be fixed by rebasing on `origin/master`

src/Middleware/HeaderPropagation/src/DependencyInjection/HeaderPropagationExtensions.cs
rynowak: Still looking for docs here. Something basic like `Provides extension methods for using header propagation.` would suffice.

src/Middleware/HeaderPropagation/src/DependencyInjection/HeaderPropagationExtensions.cs
```diff
@@ -0,0 +1,95 @@
...
+            services.TryAddSingleton<HeaderPropagationValues>();
+
+            return services;
+        }
+
+        /// <summary>
+        /// Adds services required for propagating headers to a <see cref="HttpClient"/>.
+        /// </summary>
+        /// <param name="services">The <see cref="IServiceCollection"/> to add the services to.</param>
+        /// <param name="configureOptions">The <see cref="HeaderPropagationOptions"/> to configure the middleware with.</param>
```
rynowak:
```suggestion
        /// <param name="configureOptions">A delegate used to configure the <see cref="HeaderPropagationOptions"/>.</param>
```

src/Middleware/HeaderPropagation/src/DependencyInjection/HeaderPropagationExtensions.cs
```diff
@@ -0,0 +1,95 @@
...
+            }
+
+            if (app.ApplicationServices.GetService<HeaderPropagationValues>() == null)
+            {
+                throw new InvalidOperationException(_unableToFindServices);
+            }
+
+            return app.UseMiddleware<HeaderPropagationMiddleware>();
+        }
+    }
```
rynowak: These methods should be split up into separate classes. One reason for this is that they need to go into different namespaces. We want `Add` and `Use` methods to be in scope by default in startup code.


I would expect to see the following:

`Microsoft.Extensions.DependencyInjection.HeaderPropagationServiceCollectionExtensions` -> `AddHeaderPropagation`
`Microsoft.Extensions.DependencyInjection.HeaderPropagationHttpClientBuilderExtensions` -> `AddHeaderPropagation`
`Microsoft.AspNetCore.Builder.HeaderPropagationApplicationBuilderExtensions` -> `UseHeaderPropagation`

src/Middleware/HeaderPropagation/src/HeaderPropagationEntry.cs
```diff
@@ -0,0 +1,37 @@
...
+using System;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNetCore.HeaderPropagation
+{
+    /// <summary>
+    /// Define the configuration of a header for the <see cref="HeaderPropagationMiddleware"/>.
+    /// </summary>
+
```
rynowak: nit: remove blank line

rynowak:
I'm going through and looking at minor details like doc comments and such now, hopefully this is the last round 👍

src/Middleware/HeaderPropagation/src/DependencyInjection/HeaderPropagationExtensions.cs
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        /// <summary>
+        /// Adds a message handler for propagating headers collected by the <see cref="HeaderPropagationMiddleware"/> to a outgoing request.
+        /// </summary>
+        /// <param name="builder">The <see cref="IHttpClientBuilder"/> to add the message handler to.</param>
+        /// <returns>The <see cref="IHttpClientBuilder"/> so that additional calls can be chained.</returns>
+        public static IHttpClientBuilder AddHeaderPropagation(this IHttpClientBuilder builder)
+        {
+            builder.Services.TryAddSingleton<HeaderPropagationValues>();
```
rynowak: null check builder

rynowak: Should this call `builder.Services.AddHeaderPropagation` instead?

src/Middleware/HeaderPropagation/src/DependencyInjection/HeaderPropagationExtensions.cs
```diff
@@ -0,0 +1,95 @@
...
+            builder.Services.TryAddTransient<HeaderPropagationMessageHandler>();
```
rynowak: I think this could be moved to `services.AddHeaderPropagation` - it is a service after all, and It can be used without the factory.

src/Middleware/HeaderPropagation/src/HeaderPropagationEntry.cs
```diff
@@ -0,0 +1,37 @@
...
+    public class HeaderPropagationEntry
+    {
+        /// <summary>
+        /// Gets or sets the name of the header to be used by the <see cref="HeaderPropagationMessageHandler"/> fot the outbound http requests.
```
rynowak:
```suggestion
        /// Gets or sets the name of the header to be used by the <see cref="HeaderPropagationMessageHandler"/> for the outbound http requests.
```

src/Middleware/HeaderPropagation/src/HeaderPropagationEntry.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// </summary>
+        public string OutboundHeaderName { get; set; }
+
+        /// <summary>
+        /// Gets or sets the default values to be used when the header in the incoming request is missing or empty.
+        /// </summary>
```
rynowak: It would be nice if the docs explain how this interacts with `ValueFactory`.

src/Middleware/HeaderPropagation/src/HeaderPropagationEntry.cs
rynowak: I think this is fine.

src/Middleware/HeaderPropagation/src/HeaderPropagationEntry.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        /// <summary>
+        /// Gets or sets the value factory to be used.
+        /// It gets as input the inbound header name for this entry as defined in the <see cref="HeaderPropagationOptions"/>
+        /// and the <see cref="HttpContext"/> of the current request.
+        /// </summary>
+        /// <remarks>
+        /// When present, the factory is the only method used to set the value.
+        /// To not add the header, return <see cref="StringValues.Empty"/>.
+        /// </remarks>
```
rynowak: docs should cover what happens when don't have a value factory.

src/Middleware/HeaderPropagation/src/HeaderPropagationEntry.cs
```diff
@@ -0,0 +1,37 @@
...
+    /// Define the configuration of a header for the <see cref="HeaderPropagationMiddleware"/>.
+    /// </summary>
+
+    public class HeaderPropagationEntry
+    {
+        /// <summary>
+        /// Gets or sets the name of the header to be used by the <see cref="HeaderPropagationMessageHandler"/> fot the outbound http requests.
+        /// </summary>
+        public string OutboundHeaderName { get; set; }
+
```
rynowak: I think somewhere in the docs it would be nice to have a comment like: 

```
If ValueFactory is null and DefaultValue is null then....
If ValueFactory is non-null then...
If ValueFactory is null and DefaultValues is non-null then...
```

Basically, somewhere we should document how users should think about the precedence of these options.

src/Middleware/HeaderPropagation/src/HeaderPropagationMessageHandler.cs
```diff
@@ -0,0 +1,50 @@
...
+{
+    /// <summary>
+    /// A message handler for propagating headers collected by the <see cref="HeaderPropagationMiddleware"/> to a outgoing request.
+    /// </summary>
+    public class HeaderPropagationMessageHandler : DelegatingHandler
+    {
+        private readonly HeaderPropagationValues _values;
+        private readonly HeaderPropagationOptions _options;
+
+        public HeaderPropagationMessageHandler(IOptions<HeaderPropagationOptions> options, HeaderPropagationValues values)
```
rynowak: docs here (even if it doesn't explain much)

src/Middleware/HeaderPropagation/src/HeaderPropagationMessageHandler.cs
```diff
@@ -0,0 +1,50 @@
...
+            {
+                throw new ArgumentNullException(nameof(options));
+            }
+
+            _options = options.Value;
+
+            _values = values ?? throw new ArgumentNullException(nameof(values));
+        }
+
+        protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
```
rynowak: docs here (even if it doesn't explain much)

src/Middleware/HeaderPropagation/src/HeaderPropagationEntry.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// <summary>
+        /// Gets or sets the default values to be used when the header in the incoming request is missing or empty.
+        /// </summary>
+        public StringValues DefaultValues { get; set; }
```
rynowak: I noticied that this is `DefaultValueS` and the factory is `ValueFactory` (singular). What would you think about calling this `DefaultValue`?

src/Middleware/HeaderPropagation/src/HeaderPropagationMessageHandler.cs
```diff
@@ -0,0 +1,50 @@
...
+        {
+            foreach ((var headerName, var entry) in _options.Headers)
+            {
+                var outputName = !string.IsNullOrEmpty(entry?.OutboundHeaderName) ? entry.OutboundHeaderName : headerName;
```
rynowak: nit: consider reversing this condition to simplifiy it.

src/Middleware/HeaderPropagation/src/HeaderPropagationMiddleware.cs
```diff
@@ -0,0 +1,66 @@
...
+    /// <summary>
+    /// A Middleware for propagating headers to a <see cref="HttpClient"/>.
+    /// </summary>
+    public class HeaderPropagationMiddleware
+    {
+        private readonly RequestDelegate _next;
+        private readonly HeaderPropagationOptions _options;
+        private readonly HeaderPropagationValues _values;
+
+        public HeaderPropagationMiddleware(RequestDelegate next, IOptions<HeaderPropagationOptions> options, HeaderPropagationValues values)
```
rynowak: docs

src/Middleware/HeaderPropagation/src/HeaderPropagationMiddleware.cs
```diff
@@ -0,0 +1,66 @@
...
+            if (options == null)
+            {
+                throw new ArgumentNullException(nameof(options));
+            }
+            _options = options.Value;
+
+            _values = values ?? throw new ArgumentNullException(nameof(values));
+        }
+
+        public Task Invoke(HttpContext context)
```
rynowak: docs

src/Middleware/HeaderPropagation/src/HeaderPropagationMiddleware.cs
```diff
@@ -0,0 +1,66 @@
...
+                    _values.Headers.TryAdd(headerName, values);
+                }
+            }
+
+            return _next.Invoke(context);
+        }
+
+        private static StringValues GetValues(string headerName, HeaderPropagationEntry entry, HttpContext context)
+        {
+            if (entry?.ValueFactory != null)
```
rynowak: What are the cases where entry could be null?

 Are you just hedging against someone adding a null entry to `_options.Headers`? I think it would be better to let this crash with an nullref.

src/Middleware/HeaderPropagation/src/HeaderPropagationValues.cs
```diff
@@ -0,0 +1,29 @@
...
+    /// Contains the headers values for the <see cref="HeaderPropagationMiddleware"/>.
+    /// </summary>
+    public class HeaderPropagationValues
+    {
+        private readonly static AsyncLocal<Dictionary<string, StringValues>> _headers = new AsyncLocal<Dictionary<string, StringValues>>();
+
+        /// <summary>
+        /// Gets the headers values collected by the <see cref="HeaderPropagationMiddleware"/> from the current request that can be propagated.
+        /// </summary>
+        public Dictionary<string, StringValues> Headers
```
rynowak:
```suggestion
        public IDictionary<string, StringValues> Headers
```

rynowak (commented):
This is a looking really good. Just a few small comments. @Tratcher - do you have any other feedback on this?

src/Middleware/HeaderPropagation/src/DependencyInjection/HeaderPropagationExtensions.cs
alefranz: Makes sense

src/Middleware/HeaderPropagation/src/DependencyInjection/HeaderPropagationExtensions.cs
alefranz: I've done the change, however I'm not 100% convinced as maybe in future you have services not needed by the HttpClient directly?
Also with this change it gives the impression that `HeaderPropagationServiceCollectionExtensions.AddHeaderPropagation` doesn't need to be called (as long as you use this in at least a client builder).
Is this what we want? or the recommended usage would be to always call `HeaderPropagationServiceCollectionExtensions.AddHeaderPropagation`?

src/Middleware/HeaderPropagation/src/DependencyInjection/HeaderPropagationExtensions.cs
alefranz: ok

eng/SharedFramework.Local.props
```diff
@@ -56,6 +56,7 @@
     <AspNetCoreAppReference Include="Microsoft.AspNetCore.Cors" />
     <AspNetCoreAppReference Include="Microsoft.AspNetCore.Diagnostics.Abstractions" />
     <AspNetCoreAppReference Include="Microsoft.AspNetCore.Diagnostics" />
+    <AspNetCoreAppReference Include="Microsoft.AspNetCore.HeaderPropagation" />
```
anurse: Can you remove this from the shared framework? We're generally very very cautious about what we put in there since it's very difficult to remove things later on. We're trying to keep it to things that every single app will use (moving forward at least). 

We can still ship it as an official package as part of our release, it just wouldn't be included in the shared framework.

@dougbu Do you know if we need to do anything special to make sure this package is in the set of packages that will end up on NuGet? If we remove it from the shared framework, I want to make sure we know it will end up landing on nuget.org

```suggestion
```

src/Middleware/HeaderPropagation/src/HeaderPropagationMiddleware.cs
alefranz: I think this could be a valid scenario, now that the header inbound name is the key of the dictionary and not in the entry.
A null configuration entry will correspond to using the same name outbound and having no factory nor default value, see last 2 tests: https://github.com/aspnet/AspNetCore/pull/7921/files#diff-e52877730c5607887b50f212d490c111R189

I guess I'll have to document this, unless you think an entry should be mandatory (eventually with all fields as null)

src/Middleware/HeaderPropagation/src/HeaderPropagationMiddleware.cs
alefranz: I haven't added documentation as I haven't seen it documented in other middlewares and I think it makes sense given it should never been instantiated directly or called `Invoke`, event if they are both public.

eng/SharedFramework.Local.props
dougbu: I don't remember exactly. Removing the `<AspNetCoreAppReference />` item may be sufficient. If not, there should be a project property you can set&hellip;

src/Middleware/HeaderPropagation/src/HeaderPropagationMessageHandler.cs
alefranz: done

src/Middleware/HeaderPropagation/src/HeaderPropagationEntry.cs
alefranz: yes, it makes sense.

eng/SharedFramework.Local.props
alefranz: I've removed `IsAspNetCoreApp` ans set `IsShippingPackage` https://github.com/aspnet/AspNetCore/pull/7921/commits/61df6957ab65b9b2bb9e4082cd0d4c2ea1ef767f.
Looking at https://github.com/aspnet/AspNetCore/blob/master/Directory.Build.targets#L5-L17 this should be the correct approach and executing `dotnet pack` locally it create the package in the `Shipping` folder.

Can you please confirm @dougbu ?

src/Middleware/HeaderPropagation/src/HeaderPropagationMessageHandler.cs
alefranz: done

src/Middleware/HeaderPropagation/src/HeaderPropagationMessageHandler.cs
alefranz: done

src/Middleware/HeaderPropagation/src/DependencyInjection/HeaderPropagationExtensions.cs
alefranz: I can do this, but it doesn't look like a common practice across the repo.

alefranz:
Hi @rynowak , thank you for the review and apologies for the delay on my side :)
I should have addressed the feedback.
Let me know what you think about the documentation, specifically the parts I skipped (as it doesn't seems common practice looking at other projects in the solution) and the attempt to explain the logic at https://github.com/aspnet/AspNetCore/pull/7921/files#diff-ceea95213282add8708d3dc7be38d089.

I'll have a look tomorrow at why it's not passing on the CI

Tratcher (approved):

src/Middleware/HeaderPropagation/src/HeaderPropagationMiddleware.cs
rynowak: I'm happy to leave this for now. When this gets merged I'm going to spend a little time playing with it.

src/Middleware/HeaderPropagation/src/DependencyInjection/HeaderPropagationExtensions.cs
rynowak: Well the reason that I commented here is that this used to duplicate the work of another method. Now that I think about it, it makes sense that you'd always want to call `services.AddHeaderPropagation` because that's where you configure stuff.

rynowak (approved):

rynowak:
Looks good from my POV - I want to get a :shipit: from @dougbu and @anurse before we merge this bad boy!

analogrelay (approved):

eng/SharedFramework.Local.props
dougbu: Seems to be working fine :smiley_cat: And, the CI build logs contain the following which also looks right:
```
  Successfully created package 'F:\workspace\_work\1\s\artifacts\packages\Release\Shipping\Microsoft.AspNetCore.HeaderPropagation.3.0.0-preview4-19177-46.nupkg'.
  Successfully created package 'F:\workspace\_work\1\s\artifacts\packages\Release\Shipping\Microsoft.AspNetCore.HeaderPropagation.3.0.0-preview4-19177-46.symbols.nupkg'.
```

Unfortunately, this is a fork and so artifacts aren't uploaded. Can't double-check that the shared framework bundles _lack_ Microsoft.AspNetCore.HeaderPropagation.dll until this is merged. That's also fine because we can look at what gets published after this is merged.

dougbu (approved):
Didn't look at the C# files at all but the rest looks good

rynowak:
@alefranz - could you rebase? after that I think we're ready to go.

alefranz:
Thanks everyone!

alefranz:
ci tests on windows are failing, but it looks unrelated
```
 [xUnit.net 00:04:51.46]     Templates.Test.RazorPagesTemplateTest.RazorPagesTemplate_IndividualAuthImplAsync(useLocalDB: True) [FAIL]
  Failed   Templates.Test.RazorPagesTemplateTest.RazorPagesTemplate_IndividualAuthImplAsync(useLocalDB: True)
  Error Message:
   System.Net.Http.HttpRequestException : The SSL connection could not be established, see inner exception.
  ---- System.IO.IOException : Authentication failed because the remote party has closed the transport stream.
```

dougbu:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=137434

aspnet-hello:

I've triaged the above build.

rynowak:
Merged ! Thanks @alefranz  and everyone else who helped review.

alefranz:
@rynowak What is the policy about back-porting this kind of packages to 2.2?

rynowak:
@alefranz - we don't backport features to a,ready shipped releases. 

If you want a version of this you can use with 2.2, I would suggest grabbing the code and building it from source - it should be straightforward since this package doesn't have tendrils anywhere else.. @Eilon will correct me if I'm wrong but our license allows you to copy as much code as you want (with attribution).

Eilon:
Yup, what @rynowak said.

ericsampson:
@rynowak I didn't get a chance to review the entire changeset closely, but is it implemented to reduce allocations as much as practical?

ericsampson:
Nice work @alefranz :)

rynowak:
>  I didn't get a chance to review the entire changeset closely, but is it implemented to reduce allocations as much as practical?

Our work is never over. If you think it can be improved, collect some data and send a PR 👍

ericsampson:
> > I didn't get a chance to review the entire changeset closely, but is it implemented to reduce allocations as much as practical?
> 
> Our work is never over. If you think it can be improved, collect some data and send a PR 👍
 
👍 Maybe I'll see if I can nerd-snipe @benaadams into taking a peek first :) Cheers all

src/Middleware/HeaderPropagation/src/HeaderPropagationMessageHandler.cs
```diff
@@ -0,0 +1,67 @@
...
+        {
+            foreach ((var headerName, var entry) in _options.Headers)
+            {
+                var outputName = string.IsNullOrEmpty(entry?.OutboundHeaderName) ? headerName : entry.OutboundHeaderName;
+
+                if (!request.Headers.Contains(outputName) &&
+                    _values.Headers.TryGetValue(headerName, out var values) &&
+                    !StringValues.IsNullOrEmpty(values))
+                {
+                    request.Headers.TryAddWithoutValidation(outputName, (string[])values);
```
benaadams: The cast `(string[])values` will cause an array allocation for most headers (which are only one)

Better would be to check the count and then switch on overload e.g.
```csharp
var count = values.Count;
if (count == 1)
{
    request.Headers.TryAddWithoutValidation(outputName, (string)values);
}
else if (count > 1)
{
    request.Headers.TryAddWithoutValidation(outputName, (string[])values);
}
```

benaadams: Also HttpClient has an odd behaviour where some headers are rejected from the Headers collection and instead need to be added to the Content.Headers collection (if there is a content); so might need to be more like:
```csharp
var hasContent = request.Content != null;

// ...

var count = values.Count;
if (count == 1)
{
    var value = (string)values;
    if (!request.Headers.TryAddWithoutValidation(outputName, value) && hasContent)
    {
        request.Content.Headers.TryAddWithoutValidation(outputName, value);
    }
} 
else ...
```
from: https://github.com/aspnet/Benchmarks/blob/7b1a5e986f06ae79270b13866e701cc6cb5bb395/src/Proxy/ProxyExtensions.cs#L31-L34

alefranz: Thanks @benaadams ! 
I'll raise a PR to address this by the end of the week.

PR #7933, opened 2019-02-26
Remove bad IsCompleted Check

jkotalik:
Test only issue. There is a check for IsCompleted that is super flaky... had a brain fart here.

### DIFF
--- a/src/Servers/Kestrel/Core/src/Internal/Http2/Http2Stream.cs
+++ b/src/Servers/Kestrel/Core/src/Internal/Http2/Http2Stream.cs
@@ -371,18 +371,13 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
 
                         // If the stream is completed go ahead and call RequestBodyPipe.Writer.Complete().
                         // Data will still be available to the reader.
-                        if (endStream)
-                        {
-                            OnEndStreamReceived();
-                        }
-                        else
+                        if (!endStream)
                         {
                             var flushTask = RequestBodyPipe.Writer.FlushAsync();
                             // It shouldn't be possible for the RequestBodyPipe to fill up an return an incomplete task if
                             // _inputFlowControl.Advance() didn't throw.
                             Debug.Assert(flushTask.IsCompleted);
                         }
-                        return Task.CompletedTask;
                     }
                 }
             }
--- a/src/Servers/Kestrel/Core/src/Internal/Http2/Http2Stream.cs
+++ b/src/Servers/Kestrel/Core/src/Internal/Http2/Http2Stream.cs
@@ -369,6 +369,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
                             RequestBodyPipe.Writer.Write(segment.Span);
                         }
 
+                        // If the stream is completed go ahead and call RequestBodyPipe.Writer.Complete().
+                        // Data will still be available to the reader.
                         if (endStream)
                         {
                             OnEndStreamReceived();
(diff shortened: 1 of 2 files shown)


jkotalik:
For https://github.com/aspnet/AspNetCore-Internal/issues/1873

src/Servers/Kestrel/test/InMemory.FunctionalTests/Http2/Http2StreamTests.cs
```diff
@@ -1022,7 +1022,6 @@ public async Task ContentLength_Received_ReadViaPipes()
             {
                 var readResult = await context.Request.BodyPipe.ReadAsync();
                 Assert.Equal(12, readResult.Buffer.Length);
-                Assert.True(readResult.IsCompleted);
```
halter73: This seems like a perf regression at the very least. Why isn't the ReadResult completed if the entire body has been read?

jkotalik: Good point. I'll investigate more thoroughly.

halter73: I just realized this was a new test. That makes it not as bad. It'd be nice if this wasn't racy though. It could certainly hurt request body consumption perf.

jkotalik: Yeah. I didn't think it was racy either 🙃

jkotalik: Looked into it. This code path is inherently racy, depending on when we call Complete on the Writer. In Http2Stream, we call complete in two places, OnRequestProcessingEnded and OnEndStreamReceived. These calls will happen either before, during or after the call to ReadAsync, depending on how quickly we call ReadAsync. 

As potentially an optimization for Http2, we can Complete the pipe when we received Content-Length amount of data. I don't believe I see us doing that today.

jkotalik:
Test failure: https://github.com/aspnet/AspNetCore-Internal/issues/1879. Will investigate once I have time.

halter73:
> As potentially an optimization for Http2, we can Complete the pipe when we received Content-Length amount of data. I don't believe I see us doing that today.

If this isn't any more difficult than deflakifying the tests, I definitely recommend it.

jkotalik:
@halter73 @Tratcher updated to optimize content length requests for Http2. Let's get this in sooner than later as the test is very flaky.

src/Servers/Kestrel/Core/src/Internal/Http2/Http2Stream.cs
```diff
@@ -368,11 +368,21 @@ public Task OnDataAsync(Http2Frame dataFrame, ReadOnlySequence<byte> payload)
...
-                        var flushTask = RequestBodyPipe.Writer.FlushAsync();
 
-                        // It shouldn't be possible for the RequestBodyPipe to fill up an return an incomplete task if
-                        // _inputFlowControl.Advance() didn't throw.
-                        Debug.Assert(flushTask.IsCompleted);
+                        // If the stream is completed go ahead and call RequestBodyPipe.Writer.Complete().
+                        // Data will still be available to the reader.
+                        if (endStream)
+                        {
+                            OnEndStreamReceived();
```
Tratcher: Isn't OnEndStreamReceived already being called somewhere else?

Tratcher: Nevermind, I see it bellow.

Tratcher: Rather than calling OnEndStreamReceived in two places, change this to `if (!endStream)` for flush.

Tratcher (approved):

halter73 (approved):

jkotalik:
Mac agent having issuse 😢 @dougbu @ryanbrandenburg thoughts?

benaadams:
> Mac agent having issuse 😢

Yeah its in a bad way https://github.com/aspnet/AspNetCore/issues/7978

PR #7936, opened 2019-02-26
Retry on SauceLabs failure

mikaelm12:
Adding a simple retry for when our SauceLabs tests fail.
Turns out we also had a small mistake in this logic where we were are running the browser tests locally even when we specifically don't want to.

### DIFF
--- a/src/SignalR/clients/ts/FunctionalTests/SignalR.Npm.FunctionalTests.npmproj
+++ b/src/SignalR/clients/ts/FunctionalTests/SignalR.Npm.FunctionalTests.npmproj
@@ -25,11 +25,8 @@
   </Target>
 
   <Target Name="RunBrowserTests">
-    <Message Text="Running JavaScript client Browser tests" Importance="high" />
-    <Yarn Command="run test:inner -- --no-color --configuration $(Configuration)" WorkingDirectory="$(RepositoryRoot)src/SignalR/clients/ts/FunctionalTests" />
-    <Message Text="Running JavaScript tests" Importance="high" />
-
     <!-- Skip the "inner" test run when we're running DailyTests -->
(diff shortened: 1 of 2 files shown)


anurse:
Looks like unrelated stuff. Summoning RaaS.

anurse:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=106559

aspnet-hello:

I've triaged the above build.

anurse:
Retrying

anurse:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=106559

mikaelm12:
@anurse  This needs to re-based first. We wanted to do this initially but we held off. No need to triage these failures yet.

src/SignalR/clients/ts/FunctionalTests/SignalR.Npm.FunctionalTests.npmproj
```diff
@@ -39,11 +36,23 @@
...
+          ContinueOnError="true"
           WorkingDirectory="$(RepositoryRoot)src/SignalR/clients/ts/FunctionalTests"
-          IgnoreStandardErrorWarningFormat="true" />
+          IgnoreStandardErrorWarningFormat="true">
+      <Output TaskParameter="ExitCode" PropertyName="ErrorCode"/>
+    </Exec>
+    <Message Importance="high" Text="$(ErrorCode)"/>
+
+    <Message Condition="'$(ErrorCode)' == '1'" Text="First run failed with $(ErrorCode) -- Rerunning sauce labs tests" Importance="high" />
+    <Exec Command="npm run test:sauce -- $(_TestSauceArgs)"
```
halter73: So this will only retry a maximum of once? Also would it make sense to try again if ErrorCode is anything other than 0?

PR #7938, opened 2019-02-26
Make page load async when using endpoint routing

pranavkm:
Fixes https://github.com/aspnet/AspNetCore/issues/8016

### DIFF
--- a/src/Mvc/Mvc.RazorPages/src/Infrastructure/DefaultPageLoader.cs
+++ b/src/Mvc/Mvc.RazorPages/src/Infrastructure/DefaultPageLoader.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
@@ -10,34 +11,35 @@ using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc.ApplicationModels;
 using Microsoft.AspNetCore.Mvc.Filters;
+using Microsoft.AspNetCore.Mvc.Infrastructure;
 using Microsoft.AspNetCore.Mvc.Razor.Compilation;
 using Microsoft.AspNetCore.Mvc.Routing;
-using Microsoft.Extensions.Caching.Memory;
 using Microsoft.Extensions.Options;
 
 namespace Microsoft.AspNetCore.Mvc.RazorPages.Infrastructure
 {
-    internal class DefaultPageLoader : PageLoaderBase
+    internal class DefaultPageLoader : PageLoader
     {
+        private readonly IActionDescriptorCollectionProvider _collectionProvider;
         private readonly IPageApplicationModelProvider[] _applicationModelProviders;
         private readonly IViewCompilerProvider _viewCompilerProvider;
         private readonly ActionEndpointFactory _endpointFactory;
         private readonly PageConventionCollection _conventions;
         private readonly FilterCollection _globalFilters;
-        private readonly MemoryCache _memoryCache;
+        private volatile InnerCache _currentCache;
 
         public DefaultPageLoader(
+            IActionDescriptorCollectionProvider actionDescriptorCollectionProvider,
             IEnumerable<IPageApplicationModelProvider> applicationModelProviders,
             IViewCompilerProvider viewCompilerProvider,
             ActionEndpointFactory endpointFactory,
             IOptions<RazorPagesOptions> pageOptions,
             IOptions<MvcOptions> mvcOptions)
         {
+            _collectionProvider = actionDescriptorCollectionProvider;
             _applicationModelProviders = applicationModelProviders
                 .OrderBy(p => p.Order)
                 .ToArray();
-
-            _memoryCache = new MemoryCache(new MemoryCacheOptions());
 
             _viewCompilerProvider = viewCompilerProvider;
             _endpointFactory = endpointFactory;
@@ -47,31 +49,42 @@ namespace Microsoft.AspNetCore.Mvc.RazorPages.Infrastructure
 
         private IViewCompiler Compiler => _viewCompilerProvider.GetCompiler();
 
-        public async override ValueTask<CompiledPageActionDescriptor> LoadAsync(PageActionDescriptor actionDescriptor)
+        private ConcurrentDictionary<PageActionDescriptor, Task<CompiledPageActionDescriptor>> CurrentCache
+        {
+            get
+            {
+                var current = _currentCache;
+                var actionDescriptors = _collectionProvider.ActionDescriptors;
+
+                if (current == null || current.Version != actionDescriptors.Version)
+                {
+                    current = new InnerCache(actionDescriptors.Version);
+                    _currentCache = current;
+                }
+
+                return current.Entries;
+            }
+        }
+
+        public override Task<CompiledPageActionDescriptor> LoadAsync(PageActionDescriptor actionDescriptor)
         {
             if (actionDescriptor == null)
             {
                 throw new ArgumentNullException(nameof(actionDescriptor));
             }
 
-            if (_memoryCache.TryGetValue(actionDescriptor, out CompiledPageActionDescriptor compiledDescriptor))
+            var cache = CurrentCache;
+            if (cache.TryGetValue(actionDescriptor, out var compiledDescriptorTask))
             {
-                return compiledDescriptor;
+                return compiledDescriptorTask;
             }
 
-            var viewDescriptor = await Compiler.CompileAsync(actionDescriptor.RelativePath);
-            var compiledPageActionDescriptor = GetCompiledPageActionDescriptor(actionDescriptor, viewDescriptor);
-            var entryOptions = new MemoryCacheEntryOptions();
-            for (var i = 0; i < viewDescriptor.ExpirationTokens.Count; i++)
-            {
-                entryOptions.ExpirationTokens.Add(viewDescriptor.ExpirationTokens[i]);
-            }
-
-            return _memoryCache.Set(actionDescriptor, compiledPageActionDescriptor, entryOptions);
+            return cache.GetOrAdd(actionDescriptor, LoadAsyncCore(actionDescriptor));
         }
 
-        private CompiledPageActionDescriptor GetCompiledPageActionDescriptor(PageActionDescriptor actionDescriptor, CompiledViewDescriptor viewDescriptor)
+        private async Task<CompiledPageActionDescriptor> LoadAsyncCore(PageActionDescriptor actionDescriptor)
         {
+            var viewDescriptor = await Compiler.CompileAsync(actionDescriptor.RelativePath);
             var context = new PageApplicationModelProviderContext(actionDescriptor, viewDescriptor.Type.GetTypeInfo());
             for (var i = 0; i < _applicationModelProviders.Length; i++)
             {
@@ -148,6 +161,19 @@ namespace Microsoft.AspNetCore.Mvc.RazorPages.Infrastructure
                     conventions.OfType<TConvention>(),
                     attributes.OfType<TConvention>());
             }
+        }
+
+        private sealed class InnerCache
+        {
+            public InnerCache(int version)
+            {
+                Version = version;
+                Entries = new ConcurrentDictionary<PageActionDescriptor, Task<CompiledPageActionDescriptor>>();
+            }
+
+            public ConcurrentDictionary<PageActionDescriptor, Task<CompiledPageActionDescriptor>> Entries { get; }
+
+            public int Version { get; }
         }
     }
 }
--- a/src/Mvc/Mvc.RazorPages/src/Infrastructure/DefaultPageLoader.cs
+++ b/src/Mvc/Mvc.RazorPages/src/Infrastructure/DefaultPageLoader.cs
@@ -5,23 +5,26 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
+using System.Threading.Tasks;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc.ApplicationModels;
 using Microsoft.AspNetCore.Mvc.Filters;
 using Microsoft.AspNetCore.Mvc.Razor.Compilation;
 using Microsoft.AspNetCore.Mvc.Routing;
+using Microsoft.Extensions.Caching.Memory;
 using Microsoft.Extensions.Options;
 
 namespace Microsoft.AspNetCore.Mvc.RazorPages.Infrastructure
 {
-    internal class DefaultPageLoader : IPageLoader
+    internal class DefaultPageLoader : PageLoaderBase
     {
         private readonly IPageApplicationModelProvider[] _applicationModelProviders;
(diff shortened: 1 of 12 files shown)


src/Mvc/Mvc.RazorPages/src/Infrastructure/PageLoaderMatcherPolicy.cs
```diff
@@ -73,17 +72,44 @@ public Task ApplyAsync(HttpContext httpContext, EndpointSelectorContext context,
             for (var i = 0; i < candidates.Count; i++)
             {
                 ref var candidate = ref candidates[i];
-                var endpoint = (RouteEndpoint)candidate.Endpoint;
+                var endpoint = candidate.Endpoint;
 
                 var page = endpoint.Metadata.GetMetadata<PageActionDescriptor>();
                 if (page != null)
                 {
-                    var compiled = _loader.Load(page);
```
pranavkm: This gets called on every request and we produce a new instance of `CompiledPageActionDescriptor`. In the past the ADP cached this for us, which now has to be pushed to the loader.

src/Mvc/Mvc.RazorPages/src/Infrastructure/PageActionInvokerProvider.cs
```diff
@@ -139,7 +138,19 @@ public void OnProvidersExecuting(ActionInvokerProviderContext context)
...
+                CompiledPageActionDescriptor compiledPageActionDescriptor;
+                var endpointFeature = actionContext.HttpContext.Features.Get<IEndpointFeature>();
+                if (endpointFeature != null)
+                {
+                    // With endpoint routing, PageLoaderMatcherPolicy should have already produced a CompiledPageActionDescriptor.
+                    compiledPageActionDescriptor = (CompiledPageActionDescriptor)actionDescriptor;
+                }
+                else
+                {
+                    compiledPageActionDescriptor = _loader.LoadAsync(actionDescriptor).GetAwaiter().GetResult();
```
pranavkm: Fixing this is a bunch more work. We could leave it less nice without endpoint routing

src/Mvc/Mvc.RazorPages/src/Infrastructure/DefaultPageLoader.cs
```diff
@@ -41,16 +47,31 @@ internal class DefaultPageLoader : IPageLoader
...
             }
 
-            var compileTask = Compiler.CompileAsync(actionDescriptor.RelativePath);
-            var viewDescriptor = compileTask.GetAwaiter().GetResult();
+            if (_memoryCache.TryGetValue(actionDescriptor, out CompiledPageActionDescriptor compiledDescriptor))
+            {
+                return compiledDescriptor;
+            }
 
+            var viewDescriptor = await Compiler.CompileAsync(actionDescriptor.RelativePath);
```
JamesNK: perf nit: Move this async stuff into its own method and only call when page is not in cache. Then LoadAsync can get from cache without an async state machine.

rynowak:
This is D O P E

src/Mvc/Mvc.RazorPages/src/Infrastructure/PageLoaderBase.cs
```diff
@@ -0,0 +1,20 @@
...
+
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Mvc.RazorPages.Infrastructure
+{
+    /// <summary>
+    /// Creates a <see cref="CompiledPageActionDescriptor"/> from a <see cref="PageActionDescriptor"/>.
+    /// </summary>
+#pragma warning disable CS0618 // Type or member is obsolete
+    public abstract class PageLoaderBase : IPageLoader
```
rynowak: How abouts just `PageLoader`.

src/Mvc/Mvc.RazorPages/src/Infrastructure/PageLoaderBase.cs
```diff
@@ -0,0 +1,20 @@
...
+#pragma warning restore CS0618 // Type or member is obsolete
+    {
+        public abstract ValueTask<CompiledPageActionDescriptor> LoadAsync(PageActionDescriptor actionDescriptor);
```
rynowak: docs

src/Mvc/Mvc.RazorPages/src/Infrastructure/PageLoaderBase.cs
```diff
@@ -0,0 +1,20 @@
...
+
+        CompiledPageActionDescriptor IPageLoader.Load(PageActionDescriptor actionDescriptor)
```
rynowak: Does this need to be obsolete too? or doesn't matter.

src/Mvc/Mvc.RazorPages/src/Infrastructure/PageLoaderMatcherPolicy.cs
```diff
@@ -73,17 +72,44 @@ public Task ApplyAsync(HttpContext httpContext, EndpointSelectorContext context,
...
+                    var compiled = _loader.LoadAsync(page);
+                    if (compiled.IsCompletedSuccessfully)
+                    {
+                        candidates.ReplaceEndpoint(i, compiled.Result.Endpoint, candidate.Values);
+                    }
+                    else
+                    {
+                        // In the most common case, LoadAsync will return a synchronous result.
+                        // Avoid going async since this is a fairly hot path.
+                        return ApplyAsyncAwaited(candidates);
```
JamesNK: In ApplyAsyncAwaited you're recalling LoadAsync for event candidate again. Shouldn't you pass in the last task and the index so you don't make unnecessary calls to LoadAsync? Or do you think it is not worth the complexity?

src/Mvc/Mvc.RazorPages/src/Infrastructure/DefaultPageLoader.cs
```diff
@@ -41,16 +47,31 @@ internal class DefaultPageLoader : IPageLoader
...
+        public async override ValueTask<CompiledPageActionDescriptor> LoadAsync(PageActionDescriptor actionDescriptor)
         {
             if (actionDescriptor == null)
             {
                 throw new ArgumentNullException(nameof(actionDescriptor));
             }
 
-            var compileTask = Compiler.CompileAsync(actionDescriptor.RelativePath);
-            var viewDescriptor = compileTask.GetAwaiter().GetResult();
+            if (_memoryCache.TryGetValue(actionDescriptor, out CompiledPageActionDescriptor compiledDescriptor))
```
rynowak: This isn't atomic - does that matter?

JamesNK (commented):
What problem is changing the type to be async solving? In some places it is moving GetResult from one place to another.

src/Mvc/Mvc.RazorPages/src/Infrastructure/PageActionInvokerProvider.cs
rynowak: I think it's it's fine. We can't really do anything about the fact that the old code path is sync.

src/Mvc/Mvc.RazorPages/src/Infrastructure/PageActionInvokerProvider.cs
```diff
@@ -139,7 +138,19 @@ public void OnProvidersExecuting(ActionInvokerProviderContext context)
             IFilterMetadata[] filters;
             if (!cache.Entries.TryGetValue(actionDescriptor, out var cacheEntry))
             {
-                actionContext.ActionDescriptor = _loader.Load(actionDescriptor);
+                CompiledPageActionDescriptor compiledPageActionDescriptor;
+                var endpointFeature = actionContext.HttpContext.Features.Get<IEndpointFeature>();
+                if (endpointFeature != null)
```
rynowak: It might be cleaner to use `MvcOptions` for this.

src/Mvc/Mvc.RazorPages/src/Infrastructure/PageLoaderMatcherPolicy.cs
rynowak: 😍

rynowak: Well. So what this is going to do is call `LoadAsync` twice for the current page instance in quick succession. It would be nice to pass in a few more variables to we could *await* the task we already have and then *resume* the loop on the next trip.

src/Mvc/Mvc.RazorPages/src/Infrastructure/DefaultPageLoader.cs
rynowak: Who was caching this before? This is wierd that we're just adding another random cache, and it's double-wierd that it's using the change tokens as well.

You have cache in the compiler, cache here, and cache in the invoker provider. It seems like this is getting out of hand.

rynowak: This design made sense before because the cache in the compiler caches each type, and the cache in the invoker provide caches each AD. 

Should we just obsolete the loader entirely and make the invoker provider be the cache?

src/Mvc/Mvc.RazorPages/src/Infrastructure/DefaultPageLoader.cs
```diff
@@ -5,23 +5,26 @@
...
 {
-    internal class DefaultPageLoader : IPageLoader
+    internal class DefaultPageLoader : PageLoaderBase
     {
         private readonly IPageApplicationModelProvider[] _applicationModelProviders;
         private readonly IViewCompilerProvider _viewCompilerProvider;
         private readonly ActionEndpointFactory _endpointFactory;
         private readonly PageConventionCollection _conventions;
         private readonly FilterCollection _globalFilters;
+        private readonly MemoryCache _memoryCache;
```
mkArtak: Doesn't look like you ever dispose of this.

mkArtak:
Can you please link the issue this addresses?

src/Mvc/Mvc.RazorPages/src/Infrastructure/IPageLoader.cs
```diff
@@ -1,11 +1,14 @@
...
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
+
 namespace Microsoft.AspNetCore.Mvc.RazorPages.Infrastructure
 {
     /// <summary>
     /// Creates a <see cref="CompiledPageActionDescriptor"/> from a <see cref="PageActionDescriptor"/>.
     /// </summary>
+    [Obsolete("This type is obsolete. Use PageLoaderBase instead.")]
```
mkArtak: Please make sure the referenced issue is marked with `breaking-change` label.

src/Mvc/Mvc.RazorPages/src/Infrastructure/PageLoaderBase.cs
javiercn: It's private isn't it? (as it is an explicit interface implementation)

src/Mvc/Mvc.RazorPages/src/Infrastructure/PageLoaderMatcherPolicy.cs
pranavkm: Ended up doing the continue. I was worried it might have appeared to be more complex than necessary, but it's fairly trivial code .

pranavkm:
@JamesNK filed an issue for this - https://github.com/aspnet/AspNetCore/issues/8016. But in essence we make a blocking call when loading the page, which performs async work. We end up getting reports of application deadlocks due to thread pool starvation. This change attempts to address it fairly cheaply with endpoint routing.

src/Mvc/Mvc.RazorPages/src/CompiledPageActionDescriptor.cs
```diff
@@ -65,5 +65,7 @@ public CompiledPageActionDescriptor(PageActionDescriptor actionDescriptor)
         /// Gets or sets the associated <see cref="Endpoint"/> of this page.
         /// </summary>
         public Endpoint Endpoint { get; set; }
+
+        internal PageActionInvokerCacheEntry PageActionInvokerCacheEntry { get; set; }
```
pranavkm: @rynowak how gross is this? Having this here avoids a second dictionary lookup per page invocation.

rynowak: Action Descriptors are data. They are intentionally constructable and don't have "initialized" state on them. It's important that we can construct valid ADs without needing services.

src/Mvc/Mvc.RazorPages/src/Infrastructure/CompiledPageActionDescriptorCache.cs
```diff
@@ -0,0 +1,299 @@
...
+                {
+                    current = new InnerCache(actionDescriptors.Version);
+                    _currentCache = current;
+                }
+
+                return current;
+            }
+        }
+
+        public virtual ValueTask<CompiledPageActionDescriptor> GetOrAddAsync(PageActionDescriptor actionDescriptor)
```
rynowak: Is this really a good usage of `ValueTask`? I don't think `ValueTask` can be awaited multiple times. Also the usual pattern for this is to cache `Task<T>` instead of caching `T`

src/Mvc/Mvc.RazorPages/src/Infrastructure/CompiledPageActionDescriptorCache.cs
```diff
@@ -0,0 +1,299 @@
...
+
+            return CreateCacheEntryAsync(cache, actionDescriptor);
+        }
+
+        private async ValueTask<CompiledPageActionDescriptor> CreateCacheEntryAsync(InnerCache innerCache, PageActionDescriptor actionDescriptor)
+        {
+            var viewDescriptor = await Compiler.CompileAsync(actionDescriptor.RelativePath);
+            var compiledActionDescriptor = GetCompiledPageActionDescriptor(actionDescriptor, viewDescriptor);
+            var cacheEntry = CreateCacheEntry(compiledActionDescriptor);
+            compiledActionDescriptor.PageActionInvokerCacheEntry = cacheEntry;
```
rynowak: I don't think we should do this.

rynowak:
Does this really need the breaking change label? Is the breaking change that we're not calling an old API anymore?

pranavkm:
> Does this really need the breaking change label? Is the breaking change that we're not calling an old API anymore?

Not any longer. I wasn't registering the `IPageLoader` in one of the iterations which could have been construed as breaking. All we have now is the `ObsoleteAttribute` on `IPageLoader`

pranavkm:
🆙 📅

src/Mvc/Mvc.RazorPages/src/Infrastructure/PageLoader.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace Microsoft.AspNetCore.Mvc.RazorPages.Infrastructure
+{
+    /// <summary>
+    /// Creates a <see cref="CompiledPageActionDescriptor"/> from a <see cref="PageActionDescriptor"/>.
+    /// </summary>
+#pragma warning disable CS0618 // Type or member is obsolete
+    public abstract class PageLoader : IPageLoader
+#pragma warning restore CS0618 // Type or member is obsolete
+    {
+        public abstract Task<CompiledPageActionDescriptor> LoadAsync(PageActionDescriptor actionDescriptor);
```
rynowak: DOCS YO

rynowak (approved):
:shipit:

PR #7943, opened 2019-02-26
[Components] Increase CanTriggerEventHandlerDisposedInEarlierPendingBatchAsync timeout to reduce test flakyness

javiercn:


src/Components/Components/test/RendererTest.cs
```diff
@@ -2566,8 +2566,15 @@ public async Task CanTriggerEventHandlerDisposedInEarlierPendingBatchAsync()
 
             // Act/Assert 3: After we complete the first UI update in which a given
             // event handler ID is disposed, we can no longer reuse that event handler ID
+
+            // From here we can't see when the async disposal is completed. Just give it plenty of time (Task.Yield isn't enough).
+            // There is a small chance in which the continuations from TaskCompletionSource run asynchronously.
+            // In that case we might not be able to see the results from RemoveEventHandlerIds as they might run asynchronously.
+            // For that case, we are going to queue a continuation on render1TCS.Task, include a 1s delay and await the resulting
+            // task to offer the best chance that we get to see the error in all cases.
+            var awaitableTask = render1TCS.Task.ContinueWith(_ => Task.Delay(1000)).Unwrap();
```
javiercn: My expectation is that by doing this we can offer a better chance for the continuation from updatedisplayasync gets executed first.

SteveSandersonMS: I'm fine with expanding the timeout, but I'm curious about why it's implemented like this and not:

```
await render1TCS.Task;
await Task.Delay(1000);
```

Alternatively, if this continues to be a problematic test, we might want to poll. Example: invoking `renderer.DispatchEventAsync` every 500ms for (say) up to 10 seconds, and completing as success when we see it throw, or failing if it still hasn't thrown after 10 seconds.

SteveSandersonMS (approved):
I'd be interested in your comments on my comment, but am happy to approve it now since I'm sure you'll make a reasonable choice about any further updates!

src/Components/Components/test/RendererTest.cs
javiercn: > I'm fine with expanding the timeout, but I'm curious about why it's implemented like this and not:
> 
> ```
> await render1TCS.Task;
> await Task.Delay(1000);
> ```
> Alternatively, if this continues to be a problematic test, we might want to poll. Example: invoking `renderer.DispatchEventAsync` every 500ms for (say) up to 10 seconds, and completing as success when we see it throw, or failing if it still hasn't thrown after 10 seconds.

I'm hoping that by doing the way I do it, it becomes a child task of the TCS and if it decides to execute the continuations asynchronously then it takes the Task.Delay after it has taken previously queued continuations. It's kind of a wild guess that I took by reading the MSDN comments on the scheduler and a bit of the code, but I hope that the scheduler offers some degree of fairness here.

I think the 1000 delay increase is the most important bit. If it continues to be flaky after that we'll try your approach.

PR #7954, opened 2019-02-26
[Components][Mvc] Relayer integration pieces

javiercn:
* Introduce Microsoft.AspNetCore.Mvc.ViewFeatures.Prerendering
* Separate static rendering from prerendering.
* Move Prerendering interfaces to Microsoft.AspNetCore.Components.Server.
* Fix exception handling during prerendering.


eng/ProjectReferences.props
```diff
@@ -10,7 +10,7 @@
     <ProjectReferenceProvider Include="Microsoft.AspNetCore.DeveloperCertificates.XPlat" ProjectPath="$(RepositoryRoot)src\Tools\FirstRunCertGenerator\src\Microsoft.AspNetCore.DeveloperCertificates.XPlat.csproj"  />
     <ProjectReferenceProvider Include="Microsoft.Extensions.ApiDescription.Tasks" ProjectPath="$(RepositoryRoot)src\Mvc\Extensions.ApiDescription.Design\src\Microsoft.Extensions.ApiDescription.Design.csproj"  />
     <ProjectReferenceProvider Include="Microsoft.AspNetCore.SignalR.Specification.Tests" ProjectPath="$(RepositoryRoot)src\SignalR\server\Specification.Tests\src\Microsoft.AspNetCore.SignalR.Specification.Tests.csproj"  />
-    <ProjectReferenceProvider Include="Microsoft.AspNetCore.Blazor.Build" ProjectPath="$(RepositoryRoot)src\Components\Blazor\Build\src\Microsoft.AspNetCore.Blazor.Build.csproj"  />
+    <ProjectReferenceProvider Include="Microsoft.AspNetCore.Blazor.Build" ProjectPath="$(RepositoryRoot)src\Components\blazor\Build\src\Microsoft.AspNetCore.Blazor.Build.csproj"  />
```
rynowak: S P I C Y

eng/ProjectReferences.props
```diff
@@ -107,6 +107,7 @@
     <ProjectReferenceProvider Include="Microsoft.AspNetCore.Mvc.Razor" ProjectPath="$(RepositoryRoot)src\Mvc\Mvc.Razor\src\Microsoft.AspNetCore.Mvc.Razor.csproj" RefProjectPath="$(RepositoryRoot)src\Mvc\Mvc.Razor\ref\Microsoft.AspNetCore.Mvc.Razor.csproj" />
     <ProjectReferenceProvider Include="Microsoft.AspNetCore.Mvc.TagHelpers" ProjectPath="$(RepositoryRoot)src\Mvc\Mvc.TagHelpers\src\Microsoft.AspNetCore.Mvc.TagHelpers.csproj" RefProjectPath="$(RepositoryRoot)src\Mvc\Mvc.TagHelpers\ref\Microsoft.AspNetCore.Mvc.TagHelpers.csproj" />
     <ProjectReferenceProvider Include="Microsoft.AspNetCore.Mvc.Testing" ProjectPath="$(RepositoryRoot)src\Mvc\Mvc.Testing\src\Microsoft.AspNetCore.Mvc.Testing.csproj" RefProjectPath="$(RepositoryRoot)src\Mvc\Mvc.Testing\ref\Microsoft.AspNetCore.Mvc.Testing.csproj" />
+    <ProjectReferenceProvider Include="Microsoft.AspNetCore.Mvc.ViewFeatures.Prerendering" ProjectPath="$(RepositoryRoot)src\Mvc\Mvc.ViewFeatures.Prerendering\src\Microsoft.AspNetCore.Mvc.ViewFeatures.Prerendering.csproj" RefProjectPath="$(RepositoryRoot)src\Mvc\Mvc.ViewFeatures.Prerendering\ref\Microsoft.AspNetCore.Mvc.ViewFeatures.Prerendering.csproj" />
```
rynowak: Can we drop ViewFeatures from the name? ViewFeatures is named like it is because it's a pile of unorganized things. I don't want to spread the terminology contagion.

src/Components/Blazor/Blazor/src/Hosting/WebAssemblyHostBuilder.cs
```diff
@@ -89,8 +89,9 @@ private void CreateServiceProvider()
             services.AddSingleton<IWebAssemblyHost, WebAssemblyHost>();
             services.AddSingleton<IJSRuntime>(WebAssemblyJSRuntime.Instance);
 
-            services.AddSingleton<IUriHelper>(WebAssemblyUriHelper.Instance);
-            services.AddSingleton<HttpClient>(s =>
+            // We need to do this to delay the initialization of WebAssemblyUriHelper
+            services.AddSingleton<IUriHelper>(sp => WebAssemblyUriHelper.Instance);
+            services.AddSingleton(s =>
```
rynowak: This was better before.

eng/ProjectReferences.props
javiercn: Mvc.Components?

src/Components/Components/src/Services/UriHelperBase.cs
```diff
@@ -73,8 +73,26 @@ public void NavigateTo(string uri, bool forceLoad)
         /// Called to initialize BaseURI and current URI before those values the first time.
         /// Override this method to dynamically calculate the those values.
         /// </summary>
-        protected virtual void InitializeState()
+        public virtual void InitializeState(string uriAbsolute, string baseUriAbsolute)
```
rynowak: This seems wrong. to have public.

src/Components/Blazor/Blazor/src/Services/WebAssemblyUriHelper.cs
```diff
@@ -40,8 +40,7 @@ protected override void InitializeState()
             // client-side (Mono) use, so it's OK to rely on synchronicity here.
             var baseUri = WebAssemblyJSRuntime.Instance.Invoke<string>(Interop.GetBaseUri);
             var uri = WebAssemblyJSRuntime.Instance.Invoke<string>(Interop.GetLocationHref);
-            SetAbsoluteBaseUri(baseUri);
-            SetAbsoluteUri(uri);
+            base.InitializeState(uri, baseUri);
```
rynowak: Between you and @pranavkm this seems to change in every PR. Are we converging on a design here? I feel like this might need to have more parts than it currently does.

src/Components/Components/src/Services/UriHelperBase.cs
javiercn: Needs to be public so that we can call it from MVC

src/Components/Blazor/Blazor/src/Services/WebAssemblyUriHelper.cs
javiercn: We are meeting tomorrow (he can't do so today)

src/Components/Server/src/Prerendering/ComponentPrerenderResult.cs
```diff
@@ -0,0 +1,33 @@
...
+
+using System.Collections.Generic;
+using System.IO;
+
+namespace Microsoft.AspNetCore.Components.Server.Prerendering
+{
+    /// <summary>
+    /// Represents the result of a prerendering an <see cref="IComponent"/>.
+    /// </summary>
+    public sealed class ComponentPrerenderResult
```
javiercn: I plan to lift this into Microsoft.AspNetCore.Components so that I can pass in the RenderTreeFrames and perform the serialization into a stream/pipe directly from here instead of relying on the internal list of tokens.

src/Components/Server/src/Circuits/CircuitPrerenderer.cs
```diff
@@ -1,42 +1,58 @@
...
+                var renderResult = await circuitHost.PrerenderComponentAsync(
                     prerenderingContext.ComponentType,
                     prerenderingContext.Parameters);
+
+                return new ComponentPrerenderResult(renderResult);
+        }
+
+        private CircuitHost GetOrCreateCircuitHost(HttpContext context)
+        {
+            if(context.Items.TryGetValue(CircuitHostKey, out var existingHost)) {
```
javiercn: Too much JS here.

src/Components/Server/src/Circuits/CircuitPrerenderer.cs
```diff
@@ -1,42 +1,58 @@
...
+                return (CircuitHost)existingHost;
```
javiercn: We reuse the same circuit host for all prerendered components in a given request. This is equivalent to a non-prerendered scenario with multiple components registered.

src/Mvc/Mvc.ViewFeatures.Prerendering/src/HtmlHelperComponentPrerenderingExtensions.cs
```diff
@@ -0,0 +1,75 @@
...
+        }
+
+        /// <summary>
+        /// Renders the <typeparamref name="TComponent"/> <see cref="IComponent"/>.
+        /// </summary>
+        /// <param name="htmlHelper">The <see cref="IHtmlHelper"/>.</param>
+        /// <param name="parameters">An <see cref="object"/> containing the parameters to pass
+        /// to the component.</param>
+        /// <returns>The HTML produced by the rendered <typeparamref name="TComponent"/>.</returns>
+        public static async Task<IHtmlContent> PrerenderComponentAsync<TComponent>(
```
javiercn: I plan for this to take in a selector parameter so that we don't have to do:
```razor
<app>@(await Html.RenderComponentAsync<TComponent>()))</app>
```
and instead we can simply do
```razor
@(await Html.RenderComponentAsync<TComponent>(selector: "app")))
```

src/Components/Components/src/Services/UriHelperBase.cs
SteveSandersonMS: Isn't this an argument for making something `public` in whatever subclass of `UriHelperBase` is used by MVC, rather than making it public on the base class? Or is that not practical for some reason?

src/Components/Server/src/Circuits/CircuitPrerenderer.cs
SteveSandersonMS: I don't really understand this comment...

src/Components/Server/src/Circuits/CircuitPrerenderer.cs
SteveSandersonMS: Makes sense

src/Components/Server/src/Circuits/CircuitPrerenderer.cs
javiercn: The braces are wrong, lol

src/Components/Components/src/Services/UriHelperBase.cs
javiercn: * We agreed that MVC was going to use a separate codepath to render components statically (not dependent on IComponentPrerenderer).
* We agreed that there were going to be no references from MVC to Components.Server and viceversa.
* This component needs to initialize the UriHelper so that requests can get routed properly. In order to do so, InitializeState needs to be public so that MVC can call into it (see line below).
https://github.com/aspnet/AspNetCore/pull/7954/files#diff-bc676323fcbc8b60ac8dc402fbb55f8aR49

src/Components/Server/src/Prerendering/ComponentPrerenderResult.cs
SteveSandersonMS: Remember that any reads you do from `RenderTreeFrames` must be synchronous, and hence can't be interleaved with network I/O. This is because `RenderTreeFrames` is a mutable structure, and we want to keep it that way.

You could arguably copy its binary data into a separate buffer if you want, though whether that's better than doing the conversion-to-HTML-string up front is unclear.

src/Components/test/E2ETest/Infrastructure/ServerFixtures/ServerFixture.cs
```diff
@@ -38,6 +38,8 @@ protected static string FindSolutionDir()
         {
             var solutionDir = FindSolutionDir();
             return Directory.GetFiles(solutionDir, "*.csproj", SearchOption.AllDirectories)
+                .Where(file => !file.Contains("\\ref\\"))
```
SteveSandersonMS: Is the Windows-specific path separator safe to use here?

src/Mvc/Mvc.ViewFeatures.Prerendering/src/HtmlHelperComponentPrerenderingExtensions.cs
SteveSandersonMS: What's the advantage of that? It just looks harder to understand to me.

src/Mvc/Mvc.ViewFeatures.Prerendering/src/HtmlHelperComponentPrerenderingExtensions.cs
```diff
@@ -0,0 +1,75 @@
...
+            if (htmlHelper == null)
+            {
+                throw new ArgumentNullException(nameof(htmlHelper));
+            }
+
+            return htmlHelper.PrerenderComponentAsync<TComponent>(null);
+        }
+
+        /// <summary>
+        /// Renders the <typeparamref name="TComponent"/> <see cref="IComponent"/>.
```
SteveSandersonMS: XML doc looks weird. Should `<see cref="IComponent"/>` be replaced with the word `component`?

src/Mvc/Mvc.ViewFeatures.Prerendering/test/HtmlHelperComponentPrerenderingExtensionsTests.cs
```diff
@@ -0,0 +1,451 @@
...
+            public void Configure(RenderHandle renderHandle)
+            {
+                _renderHandle = renderHandle;
+            }
+
+            public Task SetParametersAsync(ParameterCollection parameters)
+            {
+                _renderHandle.Render(builder =>
+                {
+                    var s = 0;
```
SteveSandersonMS: Please remove all instances of code that generates sequence numbers programmatically. It's incredibly misleading. Users keep doing this too, perhaps because they copy examples. Sequence numbers should always be literal values.

SteveSandersonMS: It's the sort of thing I'd write an analyzer for if I had enough spare time.

src/Mvc/Mvc.ViewFeatures.Prerendering/test/HtmlHelperComponentPrerenderingExtensionsTests.cs
```diff
@@ -0,0 +1,451 @@
...
+                builder.AddMarkupContent(s++, "\n");
+
+                builder.CloseElement();
+            }
+        }
+
+        private static IHtmlHelper CreateHelper(HttpContext ctx = null, Action<IServiceCollection> configureServices = null)
+        {
+            var services = new ServiceCollection();
+            //services.AddRazorComponents();
```
SteveSandersonMS: Typo?

src/Mvc/Mvc.ViewFeatures/src/Infrastructure/HttpUriHelper.cs
```diff
@@ -3,57 +3,17 @@
 
 using System;
 using Microsoft.AspNetCore.Components.Services;
-using Microsoft.AspNetCore.Http;
-using Microsoft.AspNetCore.Http.Extensions;
 
 namespace Microsoft.AspNetCore.Mvc.ViewFeatures
 {
     internal class HttpUriHelper : UriHelperBase
```
SteveSandersonMS: Out of interest, why is it called `HttpUriHelper`? Is this specific to prerendering, and if so, could it be called `PrerenderingUriHelper` instead?

src/Components/Server/src/Circuits/CircuitPrerenderer.cs
SteveSandersonMS: Oh, I see!

src/Mvc/Mvc.ViewFeatures.Prerendering/src/HtmlHelperComponentPrerenderingExtensions.cs
javiercn: It resembles `AddComponentsHub(selector: "app")` and we want to put the circuit id as an attribute in the selector so that we can resume the circuit from there.

SteveSandersonMS:
No other objections from me! Thanks for proceeding with this, @javiercn!

Since this touches MVC stuff so much, could you get @rynowak's approval before merging?

src/Mvc/Mvc.ViewFeatures/src/Infrastructure/HttpUriHelper.cs
javiercn: It was called that way because it was initialized with the HttpContext. I plan to separate the terminology even more between the basic rendering MVC does and the prerendering. My current thinking is to name the basic MVC Rendering with Static and the full thing Prerendering

src/Mvc/Mvc.ViewFeatures.Prerendering/src/HtmlHelperComponentPrerenderingExtensions.cs
SteveSandersonMS: There are other ways we could do this that wouldn't have to impact the developer-facing API.

src/Components/Server/src/Prerendering/ComponentPrerenderResult.cs
javiercn: Yes, I don't expect this to be much of a problem. The general recommendation would be to consume the results right away, and I have ideas on how to further prevent more renders after we decide we are done with prerendering and until we reconect to the server, so the render tree frames shouldn't change after we are done prerendering for the remainder of the request.

dougbu (commented):
Don't see much for me to comment on. `eng/` changes look fine modulo @rynowak's naming suggestion.

javiercn:
Closing this or as the changes are integrated in a different PR

PR #7956, opened 2019-02-26
Add temporary signcheck exclusion for comhost inside shared runtime

pakrym:
Workaround https://github.com/dotnet/core-setup/issues/5253

Fix signcheck failures:
```

[File] AspNetCoreRuntime.3.0.x64.3.0.0-preview4-19125-09.nupkg, Signed: True

[File] 0833626f316b72011de3acb878c4b3bb.dll, Signed: False, Full Name: content/shared/Microsoft.NETCore.App/3.0.0-preview4-27424-1/comhost.dll [Error] HRESULT: 800b0100 (No signature was present in the subject)

[File] AspNetCoreRuntime.3.0.x86.3.0.0-preview4-19125-09.nupkg, Signed: True

[File] 0833626f316b72011de3acb878c4b3bb.dll, Signed: False, Full Name: content/shared/Microsoft.NETCore.App/3.0.0-preview4-27424-1/comhost.dll [Error] HRESULT: 800b0100 (No signature was present in the subject)

```

### DIFF
--- a/eng/signcheck.exclusions.txt
+++ b/eng/signcheck.exclusions.txt
@@ -1,5 +1,6 @@
 */AppHostTemplate/apphost.exe;AspNetCoreRuntime.*.nupkg; Exclude the apphost because this is expected to be code-signed by customers after the SDK modifies it.
 */runtime.*.microsoft.netcore.dotnetapphost/*/apphost.exe;Microsoft.AspNetCore.AzureAppServices.SiteExtension.*.nupkg; Exclude the apphost because this is expected to be code-signed by customers after the SDK modifies it.
+*/Microsoft.NETCore.App/*/comhost.dll;AspNetCoreRuntime.3.0.*.nupkg; Workaround https://github.com/dotnet/core-setup/issues/5253
 *.js;; Exclude all JavaScript files from codesigning because we don't expect these to run on Windows Script Host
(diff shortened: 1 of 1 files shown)


AaronRobinsonMSFT:
I think I see the question. The `comhost.dll` is similar to `apphost.exe` - which I don't think is normally signed. Whatever the process for `apphost.exe`, `comhost.dll` should follow.

natemcmaster:
That doesn't sound right. Why would comhost.dll not be signed? Is this a file we would expect customers to rename and code-sign themselves? If so, why is it shipping inside the shared runtime? apphost.exe is not inside Microsoft.NETCore.App

pakrym:
@natemcmaster see dotnet/core-setup#5253

natemcmaster:
So, the plan is to remove comhost.dll from the shared framework? If so, I would favor taking an update instead of adding an exclusion.

AaronRobinsonMSFT:
>Is this a file we would expect customers to rename and code-sign themselves?

@natemcmaster Yes.

>If so, I would favor taking an update instead of adding an exclusion.

Agreed. We will address https://github.com/dotnet/core-setup/issues/5253 today.

pakrym:
Ok, if having a broken build for a day is acceptable we can wait for an update

cc @davidfowl

PR #7964, opened 2019-02-26
Add FormPipeReader

jkotalik:
For https://github.com/aspnet/AspNetCore/issues/3967

``` ini

BenchmarkDotNet=v0.10.13, OS=Windows 10.0.17763
Intel Xeon CPU E5-1650 v4 3.60GHz, 1 CPU, 12 logical cores and 6 physical cores
.NET Core SDK=3.0.100-preview-010184
  [Host]     : .NET Core 3.0.0-preview3-27414-8 (CoreCLR 4.6.27414.77, CoreFX 4.7.19.10906), 64bit RyuJIT
  Job-BZPCMW : .NET Core 3.0.0-preview3-27414-8 (CoreCLR 4.6.27414.77, CoreFX 4.7.19.10906), 64bit RyuJIT

Runtime=Core  Server=True  Toolchain=.NET Core 3.0  
RunStrategy=Throughput  

```
|                       Method |      Mean |    Error |   StdDev |    Median |         Op/s |  Gen 0 | Allocated |
|----------------------------- |----------:|---------:|---------:|----------:|-------------:|-------:|----------:|
|                 ReadUtf8Data |  77.75 ns | 1.594 ns | 3.220 ns |  75.92 ns | 12,860,911.2 |      - |       0 B |
|              ReadUnicodeData | 143.71 ns | 2.883 ns | 6.851 ns | 143.97 ns |  6,958,491.6 | 0.0010 |      64 B |
|    ReadUtf8MultipleBlockData |  72.95 ns | 1.490 ns | 3.240 ns |  72.91 ns | 13,708,615.3 |      - |       0 B |
| ReadUnicodeMultipleBlockData | 136.28 ns | 2.751 ns | 5.803 ns | 135.70 ns |  7,338,085.7 | 0.0007 |      64 B |


src/Http/WebUtilities/src/FormPipeReader.cs
```diff
@@ -0,0 +1,345 @@
...
+                        break;
+                    }
+
+                    value = buffer.Slice(sequenceReader.Position);
+
+                    sequenceReader.Advance(value.Length);
+                }
+
+                // Need to call ToArray if the key/value spans multiple segments 
+                var decodedKey = GetDecodedString((key.IsSingleSegment ? key.First : key.ToArray()).Span, encoding);
```
jkotalik: if less that 128, stackalloc, otherwise rent from shared array pool here.

src/Http/WebUtilities/src/FormPipeReader.cs
```diff
@@ -0,0 +1,345 @@
...
+using Microsoft.AspNetCore.Connections.Abstractions.Sources;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNetCore.WebUtilities
+{
+    /// <summary>
+    /// Used to read an 'application/x-www-form-urlencoded' form.
+    /// Internally reads from a PipeReader.
+    /// </summary>
+    public class FormPipeReader
```
davidfowl: Make it static

src/Http/WebUtilities/src/FormPipeReader.cs
```diff
@@ -0,0 +1,345 @@
...
+        /// </summary>
+        /// <param name="cancellationToken">The <see cref="CancellationToken"/>.</param>
+        /// <returns>The collection containing the parsed HTTP form body.</returns>
+        public async Task<Dictionary<string, StringValues>> ReadFormAsync(CancellationToken cancellationToken = default)
+        {
+            KeyValueAccumulator accumulator = default;
+            while (true)
+            {
+                ReadResult readResult;
+                if (!_pipeReader.TryRead(out readResult))
```
davidfowl: Remove this

src/Http/WebUtilities/src/FormPipeReader.cs
```diff
@@ -0,0 +1,345 @@
...
+                var buffer = readResult.Buffer;
+
+                if (!buffer.IsEmpty)
+                {
+                    TryParseFormValues(ref buffer, ref accumulator, readResult.IsCompleted, _encoding, KeyLengthLimit, ValueLengthLimit, ValueCountLimit);
+                }
+
+                if (readResult.IsCompleted)
+                {
+                    break;
```
jkotalik: Throw if there is data here.

src/Http/WebUtilities/src/FormPipeReader.cs
jkotalik: Let's profile it first 😄

src/Http/WebUtilities/perf/Microsoft.AspNetCore.WebUtilities.Performance/FormPipeReaderInternalsBenchmark.cs
```diff
@@ -0,0 +1,56 @@
...
+            _singleSegmentUnicodeReadOnlySequence = new ReadOnlySequence<byte>(Encoding.Unicode.GetBytes("foo=bar&baz=boo"));
+            _multiSegmentUnicodeReadOnlySequence = ReadOnlySequenceFactory.CreateSegments(Encoding.Unicode.GetBytes("foo=bar&baz=boo"), Encoding.Unicode.GetBytes("&haha=hehe&lol=temp"));
+
+            _accumulator = default;
+        }
+
+        [Benchmark]
+        public void ReadUtf8Data()
+        {
+            FormPipeReader.TryParseFormValues(ref _singleSegmentUtf8ReadOnlySequence, ref _accumulator, isFinalBlock: false);
```
Tratcher: Shouldn't isFinalBlock be true for all of these?

src/Http/WebUtilities/src/FormPipeReader.cs
```diff
@@ -0,0 +1,345 @@
...
+                }
+
+                var decodedKey = GetDecodedString(key, encoding);
+                var decodedValue = GetDecodedString(value, encoding);
+
+                AppendAndVerify(ref accumulator, keyLengthLimit, valueLengthLimit, valueCountLimit, decodedKey, decodedValue);
+
+                // Cover case where we don't have an ampersand at the end.
+                consumed += key.Length + value.Length + (ampersand == -1 ? equalsEncoded.Length : equalsEncoded.Length + andEncoded.Length);
+            }
```
Tratcher: There needs to be a check here that if isFinalBlock, make sure we've consumed all of the data. There could be garbage at the end of the form.

src/Http/WebUtilities/src/FormPipeReader.cs
```diff
@@ -0,0 +1,345 @@
...
+                TryParseFormValuesFast(buffer.First.Span,
+                    ref accumulator,
+                    isFinalBlock,
+                    encoding,
+                    keyLengthLimit,
+                    valueLengthLimit,
+                    valueCountLimit,
+                    equalsEncoded,
+                    andEncoded,
+                    out var consumed);
```
Tratcher: Why out consumed only to slice? Can't you pass the buffer in as ref and slice? Or is that slower?

src/Http/WebUtilities/src/FormPipeReader.cs
```diff
@@ -0,0 +1,345 @@
...
+            int valueCountLimit)
+        {
+            // We will eventually be comparing against equals and ampersand, so get the encoded
+            // versions of them. For Utf8/ASCII, these are static, so getting the encoding is zero allocation.
+            var equalsEncoded = GetEqualsForEncoding(encoding);
+            var andEncoded = GetAndForEncoding(encoding);
+
+            if (buffer.IsSingleSegment)
+            {
+                TryParseFormValuesFast(buffer.First.Span,
```
Tratcher: Trying to make this all static is a mess. Nor does it save you anything when you didn't make the type itself static. If it's all got to be static then make a struct with all the invariant parameters including equalsEncoded, etc..

src/Http/WebUtilities/src/FormPipeReader.cs
```diff
@@ -0,0 +1,345 @@
...
+                return "";
+            }
+            else if (encoding == Encoding.UTF8 || encoding == Encoding.ASCII)
+            {
+                // UrlDecoder only works on UTF8 (and implicitly ASCII)
+
+                // We need to create a Span from a ReadOnlySpan. This cast is safe because the memory is still held by the pipe
+                // We will also create a string from it by the end of the function.
+                var span = MemoryMarshal.CreateSpan(ref Unsafe.AsRef(readOnlySpan[0]), readOnlySpan.Length);
+                var bytes = UrlDecoder.DecodeInPlace(span);
```
Tratcher: We'll need to think about how DecodeInPlace affects buffered/seekable data. This span is likely a temp buffer and not the real request body buffer, but keep that in mind.

Tratcher (changes requested):

src/Http/WebUtilities/src/FormPipeReader.cs
```diff
@@ -0,0 +1,345 @@
...
+        {
+            TryParseFormValues(ref buffer, ref accumulator, isFinalBlock, Encoding.UTF8);
+        }
+
+        internal static void TryParseFormValues(ref ReadOnlySequence<byte> buffer, ref KeyValueAccumulator accumulator, bool isFinalBlock, Encoding encoding)
+        {
+            TryParseFormValues(ref buffer, ref accumulator, isFinalBlock, encoding, DefaultKeyLengthLimit, DefaultValueLengthLimit, DefaultValueCountLimit);
+        }
+
+        internal static void TryParseFormValues(
```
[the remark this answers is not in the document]
davidfowl: Yea

src/Http/WebUtilities/src/FormPipeReader.cs
davidfowl: I think that’s fine for now but it’s a waste of an allocation. I agree there are too many arguments here and it’s unclean

src/Http/WebUtilities/src/FormPipeReader.cs
davidfowl: You need to update the sequence so there’s no point slicing the span.

src/Http/WebUtilities/src/FormPipeReader.cs
jkotalik: Yeah I didn't make the final transition to making this all static. There are definitely too many arguments, that's for sure. 

For now, I'll make it a non-static class.

src/Http/WebUtilities/src/FormPipeReader.cs
jkotalik: I don't think I follow. How could there be garbage here? If the encoded bytes had an extra invalid byte at the end?

davidfowl: There’s no check required here if you throw in the async loop

src/Http/WebUtilities/src/FormPipeReader.cs
davidfowl: Without this

src/Http/WebUtilities/src/FormPipeReader.cs
```diff
@@ -31,6 +31,9 @@ public class FormPipeReader
         private PipeReader _pipeReader;
         private Encoding _encoding;
 
+        private static ReadOnlyMemory<byte> _otherEqualEncoding;
```
davidfowl: Unfortunately this defeats the span optimizations. Let’s not do this. We can take the perf hit in the slow path

src/Http/WebUtilities/src/FormPipeReader.cs
```diff
@@ -0,0 +1,332 @@
...
+    /// Used to read an 'application/x-www-form-urlencoded' form.
+    /// Internally reads from a PipeReader.
+    /// </summary>
+    public class FormPipeReader
+    {
+        private static int DefaultValueCountLimit = 1024;
+        private static int DefaultKeyLengthLimit = 1024 * 2;
+        private static int DefaultValueLengthLimit = 1024 * 1024 * 4;
+        private static ReadOnlySpan<byte> UTF8EqualEncoded => new byte[] { (byte)'=' };
+        private static ReadOnlySpan<byte> UTF8AndEncoded => new byte[] { (byte)'&' };
```
davidfowl: We can hardcode utf16 but I dunno how common it is.

Tratcher: Don't bother until you actually find some examples. I haven't seen it used here.

src/Http/WebUtilities/src/FormPipeReader.cs
```diff
@@ -0,0 +1,332 @@
...
+
+                AppendAndVerify(ref accumulator, decodedKey, decodedValue);
+
+                consumed = sequenceReader.Position;
+            }
+
+            buffer = buffer.Slice(consumed);
+        }
+
+        private string GetDecodedStringFromReadOnlySequence(ReadOnlySequence<byte> ros)
```
davidfowl: Should be passed by in and it should be called buffer

src/Http/WebUtilities/src/FormPipeReader.cs
davidfowl: Good

jkotalik: We don't optimize Unicode (still call Uri.Unescape) so I'm not going to bother.

src/Http/WebUtilities/src/FormPipeReader.cs
jkotalik: This is only used for non utf8/ascii.

davidfowl: Yes ignore my comment I didn't; see the other code path below.

src/Http/WebUtilities/src/FormPipeReader.cs
```diff
@@ -0,0 +1,346 @@
...
+        {
+            if (encoding == Encoding.UTF7)
+            {
+                throw new ArgumentException("UTF7 is unsupported and insecure. Please select a different encoding.");
+            }
+
+            _pipeReader = pipeReader;
+            _encoding = encoding;
+
+            if (_encoding != Encoding.UTF8 || _encoding != Encoding.ASCII)
```
campersau:
```suggestion
            if (_encoding != Encoding.UTF8 && _encoding != Encoding.ASCII)
```

src/Http/WebUtilities/src/FormPipeReader.cs
```diff
@@ -0,0 +1,346 @@
...
+        private const int DefaultValueCountLimit = 1024;
+        private const int DefaultKeyLengthLimit = 1024 * 2;
+        private const int DefaultValueLengthLimit = 1024 * 1024 * 4;
+
+        // Used for UTF8/ASCII (precalculated for fast path)
+        private static ReadOnlySpan<byte> UTF8EqualEncoded => new byte[] { (byte)'=' };
+        private static ReadOnlySpan<byte> UTF8AndEncoded => new byte[] { (byte)'&' };
+        // Used for other encodings
+        private static ReadOnlyMemory<byte> _otherEqualEncoding;
+        private static ReadOnlyMemory<byte> _otherAndEncoding;
```
campersau: Why `static`?

Tratcher (dismissed):

src/Shared/UrlDecoder/UrlDecoder.cs
```diff
@@ -271,9 +275,12 @@ private static int UnescapePercentEncoding(ref int scan, Span<byte> buffer)
                 return -1;
             }
 
-            if (SkipUnescape(value1, value2))
+            if (!formDecoding)
             {
-                return -1;
+                if (SkipUnescape(value1, value2))
```
Tratcher: This might be clearer if you passed formDecoding into SkipUnescape.

src/Shared/UrlDecoder/UrlDecoder.cs
```diff
@@ -249,8 +252,9 @@ private static void Copy(int begin, int end, ref int writer, Span<byte> buffer)
         /// </summary>
         /// <param name="scan">The value to read</param>
         /// <param name="buffer">The byte array</param>
+        /// <param name="formDecoding">Whether we are decoding a form or not. Will escape '/' if we are doing form encoding</param>
         /// <returns>The unescaped byte if success. Otherwise return -1.</returns>
-        private static int UnescapePercentEncoding(ref int scan, Span<byte> buffer)
+        private static int UnescapePercentEncoding(ref int scan, Span<byte> buffer, bool formDecoding)
```
Tratcher: Inconsistent use of formDecoding and formEncoding

src/Http/WebUtilities/src/FormPipeReader.cs
```diff
@@ -0,0 +1,346 @@
...
+                return decodedString;
+            }
+        }
+
+        // Check that key/value constraints are met and appends value to accumulator.
+        private KeyValueAccumulator AppendAndVerify(ref KeyValueAccumulator accumulator, string decodedKey, string decodedValue)
+        {
+            if (decodedKey.Length > KeyLengthLimit)
+            {
+                throw new InvalidDataException($"Form key or value length limit {KeyLengthLimit} exceeded.");
```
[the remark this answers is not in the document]
jkotalik: Wouldn't that operate on the byte array length rather than the decoded key/value length though? I don't think there is a better place besides this.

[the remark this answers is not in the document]
Tratcher: No, this is a blocker. What happens if I send a 2gb form key or value? You'd buffer the whole thing.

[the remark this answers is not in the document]
jkotalik: Good point. The fields like ValueCountLimit, KeyLengthLimit, etc are ambiguous to begin with but they were originally counting the number of characters instead of bytes. Should we just start counting bytes instead and change the default based on encoding?

jkotalik:
Updated numbers:

|                   Method |     Mean |     Error |    StdDev |  Op/s |    Gen 0 |   Allocated |
|------------------------- |---------:|----------:|----------:|------:|---------:|------------:|
| ReadSmallFormAsyncStream | 6.176 ms | 0.1285 ms | 0.3339 ms | 161.9 | 328.1250 | 22968.86 KB |
|   ReadSmallFormAsyncPipe | 1.069 ms | 0.0211 ms | 0.0485 ms | 935.8 |   7.8125 |   586.54 KB |

|                    Method |     Mean |     Error |    StdDev |      Op/s |  Gen 0 | Allocated |
|-------------------------- |---------:|----------:|----------:|----------:|-------:|----------:|
|              ReadUtf8Data | 1.130 us | 0.0223 us | 0.0326 us | 884,770.5 | 0.0114 |     792 B |
| ReadUtf8MultipleBlockData | 2.605 us | 0.0529 us | 0.0854 us | 383,814.5 | 0.0153 |    1048 B |

src/Http/WebUtilities/src/FormPipeReader.cs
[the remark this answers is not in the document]
Tratcher: Hmm. For ascii, UTF-16, and UTF32 the ratio of bytes to chars is fixed. For UTF-8 that doesn't work though, and it's twice as bad with the percent encoding. Counting bytes is always going to be the most restrictive option but I don't think UTF-8 gives you a good alternative.

Changing the default based on the encoding would be problematic for the same reason, you don't know how much to expect for UTF-8.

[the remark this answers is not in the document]
jkotalik: How about we use the max bytes per character as a safety net. We check both in both places (after indexOf and the decoded string).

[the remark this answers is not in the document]
davidfowl: Let just make a breaking change to limit the bytes vs chars. That way we can just check as part of the parsing instead of after

src/Shared/UrlDecoder/UrlDecoder.cs
```diff
@@ -52,7 +54,12 @@ public static int DecodeInPlace(Span<byte> buffer)
                     break;
                 }
 
-                if (buffer[sourceIndex] == '%')
+                if (buffer[sourceIndex] == '+' && isFormEncoding)
+                {
+                    // Set it to ' ' when we are doing form encoding.
+                    buffer[sourceIndex] = 0x20;
```
davidfowl: nice!

src/Http/WebUtilities/src/FormPipeReader.cs
```diff
@@ -0,0 +1,342 @@
...
+
+            if (accumulator.ValueCount > ValueCountLimit)
+            {
+                throw new InvalidDataException($"Form value count limit {ValueCountLimit} exceeded.");
+            }
+
+            return accumulator;
+        }
+
+        private string GetDecodedString(in ReadOnlySpan<byte> readOnlySpan)
```
davidfowl: Don't use in for ReadOnlySpan or span its only needed for ReadOnlySequence (since that's a big struct)

src/Http/WebUtilities/src/FormPipeReader.cs
```diff
@@ -0,0 +1,342 @@
...
+
+            if (ros.Length < StackAllocThreshold)
+            {
+                Span<byte> buffer = stackalloc byte[(int)ros.Length];
+                ros.CopyTo(buffer);
+                return GetDecodedString(buffer);
+            }
+            else
+            {
+                var byteArray = ArrayPool<byte>.Shared.Rent((int)ros.Length);
```
khellang: Should this be returned in a finally block or is that not necessary?

src/Http/WebUtilities/src/FormPipeReader.cs
```diff
@@ -0,0 +1,342 @@
...
+
+        // Used for UTF8/ASCII (precalculated for fast path)
+        private static ReadOnlySpan<byte> UTF8EqualEncoded => new byte[] { (byte)'=' };
+        private static ReadOnlySpan<byte> UTF8AndEncoded => new byte[] { (byte)'&' };
+        // Used for other encodings
+        private ReadOnlyMemory<byte> _otherEqualEncoding;
+        private ReadOnlyMemory<byte> _otherAndEncoding;
+
+        private PipeReader _pipeReader;
+        private Encoding _encoding;
```
Yves57: `readonly`?

src/Http/WebUtilities/src/FormPipeReader.cs
```diff
@@ -0,0 +1,342 @@
...
+            {
+                // Slow path for Unicode and other encodings.
+                // Just do raw string replacement.
+                var decodedString = _encoding.GetString(readOnlySpan);
+                decodedString = decodedString.Replace('+', ' ');
+                return Uri.UnescapeDataString(decodedString);
+            }
+        }
+
+        private ReadOnlySpan<byte> GetEqualsForEncoding()
```
Yves57: Just curious, why adding `GetEqualsForEncoding()` and `GetAndForEncoding()` and not initializing `_otherEqualEncoding` and `_otherAndEncoding` directly in the constructor? So it could be possible to remove the lines:
```
var equalsDelimiter = GetEqualsForEncoding();
var andDelimiter = GetAndForEncoding();
```

src/Http/WebUtilities/src/FormPipeReader.cs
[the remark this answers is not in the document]
Tratcher: Max bytes per char for UTF-8 is a 4x multiplier over the number of characters usually used. Counting the actual bytes is better, it will be ascii most of the time. No need to check the string, it's not going to get longer after decoding.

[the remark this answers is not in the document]
davidfowl: Agreed let’s nust check the bytes

src/Http/WebUtilities/src/FormPipeReader.cs
jkotalik: In this can, I don't think we need to, but it's generally a good practice. I'll go ahead and add it.

src/Http/WebUtilities/src/FormPipeReader.cs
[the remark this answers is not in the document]
jkotalik: Throwing for bytes now. It's a bit tacky because I need to check whether to throw in two places. for each key/value 😢

src/Http/WebUtilities/src/FormPipeReader.cs
```diff
@@ -153,11 +154,20 @@ public FormPipeReader(PipeReader pipeReader, Encoding encoding)
...
+                    if (span.Length > KeyLengthLimit)
+                    {
+                        throw new InvalidDataException($"Form key or value length limit {KeyLengthLimit} exceeded.");
+                    }
                     break;
                 }
 
                 key = span.Slice(0, equals);
 
+                if (key.Length > KeyLengthLimit)
```
Tratcher: Why not do this before slicing? equals is an int, no?

src/Http/WebUtilities/src/FormPipeReader.cs
```diff
@@ -153,11 +154,20 @@ public FormPipeReader(PipeReader pipeReader, Encoding encoding)
 
                 if (equals == -1)
                 {
+                    if (span.Length > KeyLengthLimit)
+                    {
+                        throw new InvalidDataException($"Form key or value length limit {KeyLengthLimit} exceeded.");
```
Tratcher: We know it was the key length.

src/Http/WebUtilities/src/FormPipeReader.cs
```diff
@@ -169,6 +179,11 @@ public FormPipeReader(PipeReader pipeReader, Encoding encoding)
                     {
                         // We can't know that what is currently read is the end of the form value, that's only the case if this is the final block
                         // If we're not in the final block, then consume nothing
+                        if (span.Length > ValueLengthLimit)
+                        {
+                            throw new InvalidDataException($"Form key or value length limit {ValueLengthLimit} exceeded.");
```
Tratcher: We know it was the value

src/Http/WebUtilities/src/FormPipeReader.cs
```diff
@@ -182,6 +197,11 @@ public FormPipeReader(PipeReader pipeReader, Encoding encoding)
                     span = span.Slice(ampersand + andDelimiter.Length);
                 }
 
+                if (value.Length > ValueLengthLimit)
```
Tratcher: Can this happen before slicing?

src/Http/WebUtilities/src/FormPipeReader.cs
```diff
@@ -253,29 +293,24 @@ private string GetDecodedStringFromReadOnlySequence(ReadOnlySequence<byte> ros)
...
-                ros.CopyTo(buffer);
-                var decodedString = GetDecodedString(buffer);
-
-                ArrayPool<byte>.Shared.Return(byteArray);
-                return decodedString;
+                try
+                {
+                    Span<byte> buffer = byteArray.AsSpan(0, (int)ros.Length);
+                    ros.CopyTo(buffer);
+                    var decodedString = GetDecodedString(buffer);
```
Tratcher: Why the extra local variable?

src/Http/WebUtilities/test/FormPipeReaderTests.cs
```diff
@@ -284,6 +284,32 @@ public void TryParseFormValues_MultiSegmentSplitAcrossSegmentsThatNeedDecodingWo
             Assert.Equal("wow", dict["\"%-.<>\\^_`{|}"]);
         }
 
+        [Fact]
+        public void TryParseFormValues_MultiSegmentExceedKeyLengthThrows()
+        {
+            var readOnlySequence = ReadOnlySequenceFactory.CreateSegments(Encoding.UTF8.GetBytes("foo=bar&baz=bo"), Encoding.UTF8.GetBytes("o&t="));
```
Tratcher: This fails on the first key in the first sequence, it never gets to the split value (which should be a split key for this test).

src/Http/WebUtilities/test/FormPipeReaderTests.cs
```diff
@@ -284,6 +284,32 @@ public void TryParseFormValues_MultiSegmentSplitAcrossSegmentsThatNeedDecodingWo
...
+            var formReader = new FormPipeReader(null);
+            formReader.KeyLengthLimit = 2;
+
+            Assert.Throws<InvalidDataException>(() => formReader.ParseFormValues(ref readOnlySequence, ref accumulator, isFinalBlock: true));
+        }
+
+        [Fact]
+        public void TryParseFormValues_MultiSegmentExceedValueLengthThrows()
+        {
+            var readOnlySequence = ReadOnlySequenceFactory.CreateSegments(Encoding.UTF8.GetBytes("foo=bar&baz=bo"), Encoding.UTF8.GetBytes("o&t="));
```
Tratcher: `bar` needs to be shorter if you're trying to test `boo` spanning the sequence gap.

src/Http/WebUtilities/test/FormPipeReaderTests.cs
Tratcher: Do you need a separate test for a trailing key that exceeds the limit?

src/Http/WebUtilities/test/FormPipeReaderTests.cs
```diff
@@ -0,0 +1,348 @@
...
+            var dict = accumulator.GetResults();
+            Assert.Equal("bar", dict["foo"]);
+            Assert.Equal("bo" + new string('a', 128), dict["baz"]);
+        }
+
+        [Theory]
+        [MemberData(nameof(Encodings))]
+        public void TryParseFormValues_MultiSegmentSplitAcrossSegmentsWithPlusesWorks(Encoding encoding)
+        {
+            var readOnlySequence = ReadOnlySequenceFactory.CreateSegments(encoding.GetBytes("+++=+++&++++=+++"), encoding.GetBytes("+&+="));
```
pakrym: Use BytePerSegmentFactory where appropriate.

src/Http/WebUtilities/ref/Microsoft.AspNetCore.WebUtilities.netcoreapp3.0.cs
```diff
@@ -1,6 +1,15 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+namespace Microsoft.AspNetCore.Internal
+{
+    public partial class UrlDecoder
```
Tratcher: Why is this public? Isn't it internal shared source?

src/Http/WebUtilities/src/FormPipeReader.cs
```diff
@@ -156,34 +156,36 @@ public FormPipeReader(PipeReader pipeReader, Encoding encoding)
                 {
                     if (span.Length > KeyLengthLimit)
                     {
-                        throw new InvalidDataException($"Form key or value length limit {KeyLengthLimit} exceeded.");
+                        ThrowKeyTooLargeException();
+                        return;
```
Tratcher: return?

src/Http/Http/test/Features/FormFeatureTests.cs
```diff
@@ -82,6 +83,49 @@ public async Task ReadFormAsync_SimpleData_ReturnsParsedFormCollection(bool buff
...
+
+            IFormFeature formFeature = new FormFeature(context.Request, new FormOptions() { BufferBody = bufferRequest });
+            context.Features.Set<IFormFeature>(formFeature);
+
+            var formCollection = await context.Request.ReadFormAsync();
+
+            Assert.Equal("bar", formCollection["foo"]);
+            Assert.Equal("2", formCollection["baz"]);
+
+            // TODO this returns a null stream because we don't wrap stream/pipereader like we do in Kestrel
```
davidfowl: So this is a test issue. Remove the code completely then?

src/Http/WebUtilities/src/FormPipeReader.cs
```diff
@@ -0,0 +1,388 @@
...
+        private const int DefaultValueCountLimit = 1024;
+        private const int DefaultKeyLengthLimit = 1024 * 2;
+        private const int DefaultValueLengthLimit = 1024 * 1024 * 4;
+
+        // Used for UTF8/ASCII (precalculated for fast path)
+        private static ReadOnlySpan<byte> UTF8EqualEncoded => new byte[] { (byte)'=' };
+        private static ReadOnlySpan<byte> UTF8AndEncoded => new byte[] { (byte)'&' };
+
+        // Used for other encodings
+        private ReadOnlyMemory<byte> _otherEqualEncoding;
```
davidfowl: nit; Just store these as byte[], it's much more efficient (even though this is the slow path).

src/Http/WebUtilities/src/FormPipeReader.cs
davidfowl: Remove the TryRead 😄

src/Http/Http/test/Features/FormFeatureTests.cs
jkotalik: Once we fix how wrapping works in HttpAbstractions, I want to revert this code.

src/Http/WebUtilities/src/FormPipeReader.cs
```diff
@@ -0,0 +1,388 @@
...
+                }
+                finally
+                {
+                    ArrayPool<byte>.Shared.Return(byteArray);
+                }
+            }
+        }
+
+        // Check that key/value constraints are met and appends value to accumulator.
+        private KeyValueAccumulator AppendAndVerify(ref KeyValueAccumulator accumulator, string decodedKey, string decodedValue)
```
davidfowl: This should be void

src/Http/WebUtilities/src/FormPipeReader.cs
```diff
@@ -0,0 +1,388 @@
...
+            }
+
+            return accumulator;
+        }
+
+        private string GetDecodedString(ReadOnlySpan<byte> readOnlySpan)
+        {
+            if (readOnlySpan.Length == 0)
+            {
+                return "";
```
davidfowl: nit: string.Empty

jkotalik:
Last test failure: https://github.com/aspnet/AspNetCore-Internal/issues/1383

halter73:
Do we have any end-to-end benchmarks for forum reading @sebastienros?

jkotalik:
MVC may have one, but I doubt it.

sebastienros:
One was done in MVC, and also does auth and database stuff. Here is how the form body is handled with a wrk script: https://github.com/aspnet/AspNetCore/blob/master/src/Mvc/benchmarkapps/BasicViews/post.lua

src/Http/WebUtilities/src/FormPipeReader.cs
```diff
@@ -96,11 +96,12 @@ public FormPipeReader(PipeReader pipeReader, Encoding encoding)
 
                 if (readResult.IsCompleted)
                 {
+                    _pipeReader.AdvanceTo(buffer.End);
```
davidfowl: Hmm I wonder if we should be calling Complete I the finally

halter73: Any reason we shouldn't rely on the server auto-completing the body PipeReader?

davidfowl: I don’t think that’s problematic but this logic did drain the pipe so why hold onto any buffers if we’re done?

src/Http/WebUtilities/src/FormPipeReader.cs
```diff
@@ -0,0 +1,382 @@
...
+        {
+            var sequenceReader = new SequenceReader<byte>(buffer);
+            var consumed = sequenceReader.Position;
+            var equalsDelimiter = GetEqualsForEncoding();
+            var andDelimiter = GetAndForEncoding();
+
+            while (!sequenceReader.End)
+            {
+                // TODO seems there is a bug with TryReadTo (advancePastDelimiter: true). It isn't advancing past the delimiter on second read.
+                if (!sequenceReader.TryReadTo(out ReadOnlySequence<byte> key, equalsDelimiter, advancePastDelimiter: false) ||
```
Tratcher: Any way to make TryReadTo stop if it exceeded the key length limit? Likely not without slicing.

src/Http/WebUtilities/src/FormPipeReader.cs
```diff
@@ -0,0 +1,382 @@
...
+            bool isFinalBlock)
+        {
+            var sequenceReader = new SequenceReader<byte>(buffer);
+            var consumed = sequenceReader.Position;
+            var equalsDelimiter = GetEqualsForEncoding();
+            var andDelimiter = GetAndForEncoding();
+
+            while (!sequenceReader.End)
+            {
+                // TODO seems there is a bug with TryReadTo (advancePastDelimiter: true). It isn't advancing past the delimiter on second read.
```
Tratcher: Is this filed? Link.

src/Http/WebUtilities/src/FormPipeReader.cs
```diff
@@ -0,0 +1,382 @@
...
+                {
+                    ThrowKeyTooLargeException();
+                }
+
+                if (!sequenceReader.TryReadTo(out ReadOnlySequence<byte> value, andDelimiter, false) ||
+                    !sequenceReader.IsNext(andDelimiter, true))
+                {
+                    if (!isFinalBlock)
+                    {
+                        if (sequenceReader.Length > ValueLengthLimit)
```
Tratcher: In the fast version you do this check before checking isFinalBlock. Any reason you don't do that here?

src/Http/WebUtilities/test/FormPipeReaderTests.cs
```diff
@@ -0,0 +1,378 @@
...
+
+            var exception = await Assert.ThrowsAsync<InvalidDataException>(
+                () => ReadFormAsync(new FormPipeReader(bodyPipe) { ValueCountLimit = 3 }));
+            Assert.Equal("Form value count limit 3 exceeded.", exception.Message);
+        }
+
+        [Fact]
+        public async Task ReadFormAsync_KeyLengthLimitMet_Success()
+        {
+            var bodyPipe = await MakePipeReader("foo=1&bar=2&baz=3&baz=4");
```
Tratcher: Make one of these keys 10 long

src/Http/WebUtilities/test/FormPipeReaderTests.cs
```diff
@@ -0,0 +1,378 @@
...
+            Assert.Equal("1", formCollection["foo"].ToString());
+            Assert.Equal("2", formCollection["bar"].ToString());
+            Assert.Equal("3,4", formCollection["baz"].ToString());
+            Assert.Equal(3, formCollection.Count);
+        }
+
+        [Fact]
+        public async Task ReadFormAsync_KeyLengthLimitExceeded_Throw()
+        {
+            var bodyPipe = await MakePipeReader("foo=1&baz1234567890=2");
```
Tratcher: Make it 11 long

src/Http/WebUtilities/test/FormPipeReaderTests.cs
```diff
@@ -0,0 +1,378 @@
...
+            Assert.Equal("1", formCollection["foo"].ToString());
+            Assert.Equal("1234567890", formCollection["bar"].ToString());
+            Assert.Equal("3,4", formCollection["baz"].ToString());
+            Assert.Equal(3, formCollection.Count);
+        }
+
+        [Fact]
+        public async Task ReadFormAsync_ValueLengthLimitExceeded_Throw()
+        {
+            var bodyPipe = await MakePipeReader("foo=1&baz=1234567890123");
```
Tratcher: Make it 11 long

Tratcher (approved):

src/Http/WebUtilities/src/FormPipeReader.cs
davidfowl: @JeremyKuhne That would be a nice feature. We'll have to slice first here.

src/Http/WebUtilities/src/FormPipeReader.cs
jkotalik: https://github.com/dotnet/corefx/commit/b0bc99d6f7943a8081a84bea23c5542831c9359e fixed it, will take a bit to get new dependencies.

Issue #7962, opened 2019-02-26
Allow custom component instantiation


AndreasTruetschel:
Change the access modifier of [ComponentFactory](https://github.com/aspnet/AspNetCore/blob/master/src/Components/Components/src/ComponentFactory.cs) to public to allow instatiation of components.  

An even better solution is to add an interface `IComponentFactory` that is implemented by `ComponentFactory` and that is registered in the `ServiceCollection`.  

[My current use case for this](https://github.com/AI4E/AI4E/blob/3fe96a1816146dbb8d9b12a54d6ae7883fdfce37/src/Blazor/AI4E.Blazor/Components/ViewExtension.cs#L162):
I implemented a system to split a blazor/razor-components project into multiple modules that are loaded lazily. The modules define view-extensions that can be rendered whereever needed. The view-extension rendering engines does not know about the concrete type of the component at compile time but needs to instantiate a component (of a type that is known at runtime), configure it with some values and finally render it. Currently I need to copy the source of `ComponentFactory` to my project to allow this. 

If you like, I can implement this and open a PR for it.

mkArtakMSFT:
We've moved this issue to the Backlog milestone. This means that it is not going to happen for the coming release. We will reassess the backlog following the current release and consider this item at that time. However, keep in mind that there are many other high priority features with which it will be competing for resources.

stsrki:
I would also like this feature to be implemented as it would allow me to easier instantiate custom components. The _PR_ as I can see is fairly straightforward so is there a reason why it's not already merged?

mkArtakMSFT:
Thanks for your PR, @AndreasTruetschel. We still need to think through this. Looking into your workaround, it seems completely reasonable and is what we would recommend. So for now, we're closing your PR till we get a better idea about the direction we would like to take this.

jspuij:
@mkArtakMSFT So the workaround is to compile your own version of Blazor? You cannot be serious about that. It is mindboggling that microsoft forces DI upon everyone through the asp.net core framework, yet do not use the same DI container for Component creation with the argument that the framework needs to be in control of component creation. Yet we DO get parameter injection, which makes even less sense.

Allow us to use the container to instantiate components or provide a good argument why the container should not be allowed to be used to instantiate components.

XomegaNet:
I agree with @jspuij.

[...]

PR #7967, opened 2019-02-26
Make ComponentFactory public and register it in the IServiceCollection

CatoLeanTruetschel:
- Changed access modifier of ComponentFactory to public
- Added an interface IComponentFactory
- Registered the ComponentFactory in the service collection as service IComponentFactory

Addresses #7962


mkArtakMSFT:
Thanks for your PR. We're closing this per https://github.com/aspnet/AspNetCore/issues/7962#issuecomment-528025437

PR #7968, opened 2019-02-26
WebAssemblyRenderer.AddComponentAsync intial parameters overloads

CatoLeanTruetschel:
Added overloads to WebAssemblyRenderer.AddComponentAsync to specify initial parameters

There is currently no specific issue for this.


SteveSandersonMS:
Thanks for the suggestion, but we're trying to avoid extending the renderer API surface area as much as we can until we do a broader refactor on this.

I really do appreciate your interest in contributing, and hopefully we will be able to merge some of your PRs! Just to make that more likely, a couple of thoughts:

 * Normally it's best to file an issue and get agreement that a certain extension or change is desirable and will be accepted before submitting a PR. Often we have specific plans for things we do or don't want to happen in the future, and spontaneous PRs are unlikely to line up with that roadmap just by chance. (It's a bit different with focused bug fixes though - we usually do want to fix bugs!)
 * We would certainly need test coverage for any changes like this one

If you don't mind, I'll close this as it's not something we could take action on in the short term. Hope that's OK, and please do continue filing issues and discussing more contributions! Contributions are important for the health of the project, even if it's not always trivial to line them up with the project plans.

PR #7971, opened 2019-02-26
Cleanup of unused code in Identity

HenrikWM:
Summary of the changes within `UserManager.cs`:
- Removed unused code
- Added `readonly` to a private field
- Renamed local variable from `store` to `claimStore`

### DIFF
--- a/src/Identity/test/Identity.Test/UserManagerTest.cs
+++ b/src/Identity/test/Identity.Test/UserManagerTest.cs
@@ -657,9 +657,9 @@ namespace Microsoft.AspNetCore.Identity.Test
         {
             var manager = MockHelpers.TestUserManager(new NoopUserStore());
             Assert.False(manager.SupportsUserClaim);
-            await Assert.ThrowsAsync<NotSupportedException>(async () => await manager.AddClaimAsync(null, null));
+            await Assert.ThrowsAsync<NotSupportedException>(async () => await manager.AddClaimsAsync(null, null));
             await Assert.ThrowsAsync<NotSupportedException>(async () => await manager.ReplaceClaimAsync(null, null, null));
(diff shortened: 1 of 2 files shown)


HenrikWM:
I’ll fix the tests in a few days.

HenrikWM:
Tests are green ✔️

src/Identity/test/Identity.Test/UserManagerTest.cs
```diff
@@ -657,9 +657,9 @@ public async Task ClaimMethodsFailWhenStoreNotImplemented()
         {
             var manager = MockHelpers.TestUserManager(new NoopUserStore());
             Assert.False(manager.SupportsUserClaim);
-            await Assert.ThrowsAsync<NotSupportedException>(async () => await manager.AddClaimAsync(null, null));
+            await Assert.ThrowsAsync<NotSupportedException>(async () => await manager.AddClaimsAsync(null, null));
```
HaoK: Removing the GetClaimStores is what caused the regression as those are what throw the not supported exceptions

HenrikWM: Yes but the store wasn't really being used in `AddClaimAsync` or `RemoveClaimAsync`. Both are just calling `GetClaimStore` without the claim store being used for anything. It's in `AddClaimsAsync` and `RemoveClaimsAsync` that the store is really used. These two methods are just passing parameters along to the overloads. So we're not changing any behavior, just throwing closing to usage than we were before.

So I've changed the methods to assert for the `NotSupportedException` on the actual methods using the store.

HaoK:
This is a regression, as we rely on the Get checks to throw the exceptions

PR #7980, opened 2019-02-26
SignalR additional xml docs

analogrelay:
Reviewed our xml docs as both ramping back up after leave and docathon. Things look pretty good, added a few things to some public methods.

### DIFF
--- a/src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
+++ b/src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
@@ -63,6 +63,31 @@ namespace Microsoft.AspNetCore.SignalR.Client
         private ConnectionState _connectionState;
         private int _serverProtocolMinorVersion;
 
+        /// <summary>
+        /// Occurs when the connection is closed, either by a call to <see cref="StopAsync(CancellationToken)"/> or a connection error.
+        /// </summary>
+        /// <remarks>
+        /// If this event was triggered from a connection error, the <see cref="Exception"/> that occurred will be passed in as the
+        /// sole argument to this handler. If this event was triggered from a call to <see cref="StopAsync(CancellationToken)"/>, then
+        /// the argument will be <see langword="null"/>.
+        /// </remarks>
+        /// <example>
+        /// The following example attaches a handler to the <see cref="Closed"/> event, and checks the provided argument to determine
+        /// if there was an error:
+        ///
+        /// <code>
+        /// connection.Closed += (exception) => {
+        ///     if (exception == null)
+        ///     {
+        ///         Console.WriteLine("Connection closed without error.");
+        ///     }
+        ///     else
+        ///     {
+        ///         Console.WriteLine($"Connection closed due to an error: {exception.Message}");
+        ///     }
+        /// };
+        /// </code>
+        /// </example>
         public event Func<Exception, Task> Closed;
 
         // internal for testing purposes
@@ -985,11 +1010,17 @@ namespace Microsoft.AspNetCore.SignalR.Client
             }
         }
 
+        /// <summary>
+        /// This method is for internal framework use and should not be called by user code.
+        /// </summary>
         public void ResetSendPing()
         {
             Volatile.Write(ref _nextActivationSendPing, (DateTime.UtcNow + KeepAliveInterval).Ticks);
         }
 
+        /// <summary>
+        /// This method is for internal framework use and should not be called by user code.
+        /// </summary>
         public void ResetTimeout()
         {
             Volatile.Write(ref _nextActivationServerTimeout, (DateTime.UtcNow + ServerTimeout).Ticks);
--- a/src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
+++ b/src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
@@ -64,11 +64,12 @@ namespace Microsoft.AspNetCore.SignalR.Client
         private int _serverProtocolMinorVersion;
 
         /// <summary>
-        /// Occurs when the connection is closed, either by a call to <see cref="StopAsync(CancellationToken)"/> or a connection error.
+        /// Occurs when the connection is closed. The connection could be closed due to an error or due to either the server or client intentionally
+        /// closing the connection without error.
         /// </summary>
         /// <remarks>
         /// If this event was triggered from a connection error, the <see cref="Exception"/> that occurred will be passed in as the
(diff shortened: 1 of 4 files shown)


src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -63,6 +63,31 @@ public partial class HubConnection
         private ConnectionState _connectionState;
         private int _serverProtocolMinorVersion;
 
+        /// <summary>
+        /// Occurs when the connection is closed, either by a call to <see cref="StopAsync(CancellationToken)"/> or a connection error.
```
BrennanConroy: or the server closed the connection gracefully

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -63,6 +63,31 @@ public partial class HubConnection
...
+        /// If this event was triggered from a connection error, the <see cref="Exception"/> that occurred will be passed in as the
+        /// sole argument to this handler. If this event was triggered from a call to <see cref="StopAsync(CancellationToken)"/>, then
+        /// the argument will be <see langword="null"/>.
+        /// </remarks>
+        /// <example>
+        /// The following example attaches a handler to the <see cref="Closed"/> event, and checks the provided argument to determine
+        /// if there was an error:
+        ///
+        /// <code>
+        /// connection.Closed += (exception) => {
```
BrennanConroy: nit: `{` on newline

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -985,11 +1010,17 @@ private async Task ReceiveLoop(ConnectionState connectionState)
             }
         }
 
+        /// <summary>
+        /// This method is for internal framework use and should not be called by user code.
```
BrennanConroy: Or just make it private :)

src/SignalR/common/Http.Connections/src/HttpConnectionContextExtensions.cs
```diff
@@ -9,6 +9,15 @@ namespace Microsoft.AspNetCore.Http.Connections
 {
     public static class HttpConnectionContextExtensions
     {
+        /// <summary>
+        /// Gets the <see cref="HttpContext"/> associated with the connection, if there is one.
+        /// </summary>
+        /// <param name="connection">The <see cref="ConnectionContext"/> representing the connection.</param>
+        /// <returns>The <see cref="HttpContext"/> associated with the connection, or <see langword="null"/> if the connection is not HTTP-based</returns>
```
BrennanConroy: full stop

BrennanConroy (approved):

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -63,6 +63,31 @@ public partial class HubConnection
         private ConnectionState _connectionState;
         private int _serverProtocolMinorVersion;
 
+        /// <summary>
+        /// Occurs when the connection is closed, either by a call to <see cref="StopAsync(CancellationToken)"/> or a connection error.
+        /// </summary>
+        /// <remarks>
+        /// If this event was triggered from a connection error, the <see cref="Exception"/> that occurred will be passed in as the
+        /// sole argument to this handler. If this event was triggered from a call to <see cref="StopAsync(CancellationToken)"/>, then
```
mikaelm12: You might want to add the server closing the connection gracefully case here as well

mikaelm12 (approved):

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
anurse: True, will update.

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
anurse: I didn't want to bog this PR down with that, but I can just remove the doc comment. It was public in 2.2 so our standard policy is to deprecate it over one release first. See https://github.com/aspnet/AspNetCore/issues/7977 for more. I'll remove these changes and we can just discuss things there.

halter73 (approved):

src/SignalR/common/Http.Connections/src/HttpConnectionContextExtensions.cs
```diff
@@ -9,6 +9,15 @@ namespace Microsoft.AspNetCore.Http.Connections
...
     {
+        /// <summary>
+        /// Gets the <see cref="HttpContext"/> associated with the connection, if there is one.
+        /// </summary>
+        /// <param name="connection">The <see cref="ConnectionContext"/> representing the connection.</param>
+        /// <returns>The <see cref="HttpContext"/> associated with the connection, or <see langword="null"/> if the connection is not HTTP-based.</returns>
+        /// <remarks>
+        /// SignalR connections can run on top of HTTP transports like WebSockets or Long Polling, or other non-HTTP transports. As a result,
+        /// this method can sometimes return <see langword="null"/> depending on the configuration of your application.
+        /// </remarks>
```
halter73: Beware that this won't show up in intellisense or the HttpConnectionContextExtensions generated docs. To see this you will have to got to the generated docs page dedicated exclusively to the GetHttpContext() method.

anurse: Yeah, I think I'll put `or <see langword="null" /> if there is no HTTP context` in the summary (and leave this remark that adds more detail).

anurse:
https://github.com/aspnet/AspNetCore/issues/7265 - which should now be skipped. Rerunning

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -63,13 +63,40 @@ public partial class HubConnection
...
+        /// <code>
+        /// connection.Closed += (exception) =>
+        /// {
+        ///     if (exception == null)
+        ///     {
+        ///         Console.WriteLine("Connection closed without error.");
+        ///     }
+        ///     else
+        ///     {
+        ///         Console.WriteLine($"Connection closed due to an error: {exception.Message}");
```
halter73: Nit: I normally prefer printing $"{exception}" over $"{exception.Message}" so you get the Exception type and stack trace too.

halter73 (approved):

PR #7987, opened 2019-02-27
Improve NullHtmlEncoder perf

sebastienros:
It's used anywhere `IHtmlContent` is rendered.

```
|                      Method |      Mean |     Error |    StdDev | Ratio | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|---------------------------- |----------:|----------:|----------:|------:|------------:|------------:|------------:|--------------------:|
|           MvcNullHtmlEncode | 0.9833 ns | 0.0121 ns | 0.0101 ns |  1.00 |           - |           - |           - |                   - |
| InlinedBetterNullHtmlEncode | 0.2979 ns | 0.0147 ns | 0.0138 ns |  0.30 |           - |           - |           - |                   - |
```

### DIFF
--- a/src/Razor/Razor/src/TagHelpers/NullHtmlEncoder.cs
+++ b/src/Razor/Razor/src/TagHelpers/NullHtmlEncoder.cs
@@ -79,8 +79,9 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
                 return;
             }
 
-            var span = value.AsSpan();
-            output.Write(span.Slice(startIndex, characterCount));
+            var span = value.AsSpan(startIndex, characterCount);
+            
+            output.Write(span);
         }
 
         /// <inheritdoc />
--- a/src/Razor/Razor/src/TagHelpers/NullHtmlEncoder.cs
+++ b/src/Razor/Razor/src/TagHelpers/NullHtmlEncoder.cs
@@ -64,6 +64,21 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
 
         public override void Encode(TextWriter output, string value, int startIndex, int characterCount)
         {
+            if (output == null)
+            {
+                throw new ArgumentNullException(nameof(output));
+            }
+
+            if (value == null)
+            {
+                throw new ArgumentNullException(nameof(value));
+            }
+
+            if (characterCount == 0)
+            {
+                return;
+            }
+
             var span = value.AsSpan();
             output.Write(span.Slice(startIndex, characterCount));
         }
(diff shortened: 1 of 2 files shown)


src/Razor/Razor/src/TagHelpers/NullHtmlEncoder.cs
```diff
@@ -11,12 +12,12 @@ namespace Microsoft.AspNetCore.Razor.TagHelpers
...
     /// expected to contain valid HTML.
     /// </summary>
-    public class NullHtmlEncoder : HtmlEncoder
+    public sealed class NullHtmlEncoder : HtmlEncoder
     {
         /// <summary>
         /// Initializes a <see cref="NullHtmlEncoder"/> instance.
         /// </summary>
-        protected NullHtmlEncoder()
+        private NullHtmlEncoder()
```
davidfowl: breaking change

sebastienros: I assume it's not very important as it should not make sense to inherit from something that does nothing, also because we don't expose this type in public surface apis, and because one can easily inherit from the same base class. Is that an issue in this case? Or just a matter of Announcement?

davidfowl: I agree but we need to leave a paper trail so that we can easily query them. I'm not sure this is worth an announcement but maybe it is?

mkArtak:
Addresses https://github.com/aspnet/AspNetCore/issues/8019

mkArtak:
@sebastienros, we track the labels on the issues, so filed one and referenced for that purpose.

src/Razor/Razor/src/TagHelpers/NullHtmlEncoder.cs
mkArtakMSFT: This is resolved now - I've filed an issue and marked as necessary

pranavkm (approved):

src/Razor/Razor/src/TagHelpers/NullHtmlEncoder.cs
```diff
@@ -85,10 +79,12 @@ public override void Encode(TextWriter output, string value, int startIndex, int
                 return;
             }
 
-            output.Write(value.Substring(startIndex, characterCount));
+            var span = value.AsSpan();
+            output.Write(span.Slice(startIndex, characterCount));
```
davidfowl: `value.AsSpan(startIndex, characterCount)`

sebastienros: Done

davidfowl (approved):

anurse:
@sebastienros Helix failures should be resolved by a rebase, please rebase asap.

anurse:
https://github.com/aspnet/AspNetCore-Internal/issues/1898 - Build timeout :(

PR #7996, opened 2019-02-27
Add a RouterBase type

CatoLeanTruetschel:
Added a `RouterBase` type that can be used to implement custom routers without the need to reimplement or copy the complete `Microsoft.AspNetCore.Components.Routing` namespace.   

This adds a method `protected abstract IEnumerable<Type> ResolveRoutableComponents()` that is used to resolve the types of components that can be routed.  

A custom router is free to override this without reimplementing the actual routing mechanism (`RouteTable`, etc)

### DIFF
--- a/src/Components/Components/src/Routing/Router.cs
+++ b/src/Components/Components/src/Routing/Router.cs
@@ -4,10 +4,6 @@
 using System;
 using System.Collections.Generic;
 using System.Reflection;
-using System.Threading.Tasks;
-using Microsoft.AspNetCore.Components.Layouts;
-using Microsoft.AspNetCore.Components.RenderTree;
-using Microsoft.AspNetCore.Components.Services;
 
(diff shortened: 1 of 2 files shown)


SteveSandersonMS:
Thanks for the suggestion, @AndreasTruetschel!

I think we need to do some broader design work on the client-side router before making commitments about how specifically we want to make it more extensible. The existing arrangement is not considered final in any way.

As such, rather than making this change part of core in the short term, would you be open to publishing a separate `ExtensibleRouter` in a NuGet package, so that people who want it can use that instead of the built-in router? That way, the "copying the whole Router namespace" problem only happens once.

AndreasTruetschel:
Thanks for your response @SteveSandersonMS 
As you suggested I forked the Blazor component routing and will develop this separately of this project. This is hosted [here](https://www.nuget.org/packages/AI4E.AspNetCore.Components.Routing/).

PR #8004, opened 2019-02-27
Disable parallel publish for Helix

pakrym:
Let's see if it make Helix publish less flaky.

Fixes: https://github.com/aspnet/AspNetCore/issues/7950


HaoK (approved):

HaoK:
It should get rid of a lot of those file is in use errors that crop up

dougbu (approved):

pakrym:
Unrelated helix failure https://github.com/aspnet/AspNetCore-Internal/issues/1879

PR #8008, opened 2019-02-27
add content type into extension provider for passbook Apple wallet files

Taturevich:
Adding support of `.pkpass` file extension used for Apple wallet passboook files.


Tratcher:
FYI the bar for adding any new items to this list is quite high. Most will be declined.

shirhatti:
What @Tratcher said. We will only add new MIME types for content that ships as part of our template.
You should use a custom [FileExtensionContentTypeProvider](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/static-files?view=aspnetcore-2.2#fileextensioncontenttypeprovider) in your application.

PR #8010, opened 2019-02-27
Fix flaky AbortedStream test in Kestrel

jkotalik:
I think this is causing https://github.com/aspnet/AspNetCore-Internal/issues/1879. This failure is only occurring on linux agents, so will run a few times to verify.

### DIFF
--- a/src/Servers/Kestrel/test/InMemory.FunctionalTests/Http2/Http2TestBase.cs
+++ b/src/Servers/Kestrel/test/InMemory.FunctionalTests/Http2/Http2TestBase.cs
@@ -25,6 +25,7 @@ using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2.HPack;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure;
 using Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.Internal;
 using Microsoft.AspNetCore.Testing;
+using Microsoft.Extensions.Logging;
 using Microsoft.Net.Http.Headers;
 using Moq;
 using Xunit;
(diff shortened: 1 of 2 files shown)


jkotalik:
The reason I think my PR introduced it was because the issue I filed was originally on https://dev.azure.com/dnceng/public/_build/results?buildId=106545, which was the original PR. I thought this test was an artifact of https://github.com/aspnet/AspNetCore/commit/57092e96ac4a45b143bd6edd631923896fe12a56, but it most likely seems it isn't the case.

Tratcher:
Your PR changes the request body pipe, but the stacktrace for the test is in the response connection pipe. They shouldn't be related.

halter73:
We should add the following to the top of the flaky test instead.


```C#
            // Remove callback that completes _pair.Application.Output on abort.
            _mockConnectionContext.Reset()
```

src/Servers/Kestrel/test/InMemory.FunctionalTests/Http2/Http2TestBase.cs
```diff
@@ -177,7 +178,11 @@ public Http2TestBase()
             _mockConnectionContext.Setup(c => c.Abort(It.IsAny<ConnectionAbortedException>())).Callback<ConnectionAbortedException>(ex =>
             {
                 // Emulate transport abort so the _connectionTask completes.
-                Task.Run(() => _pair.Application.Output.Complete(ex));
+                Task.Run(() =>
+                {
+                    TestApplicationErrorLogger.LogInformation($"ConnectionAbortedException thrown, Exception message: {ex.Message}. Completing transport Output.");
```
halter73: This would omit the stack trace which could be useful. I also don't love how severe it looks considering it's pretty common for ConnectionContext.Abort() to be called. What about the following?

```suggestion
                TestApplicationErrorLogger.LogInformation(0, ex, "ConnectionContext.Abort() was called. Completing _pair.Application.Output.");
```

I haven't checked what the output looks like, but I think the test logs should include the full exception. If not, the other alternative is to use `{ex}` instead of `{ex.Message}` since the former just calls Exception.ToString().

halter73 (approved):
pending most recent comment about the exception stack trace

PR #8013, opened 2019-02-28
Send email confirmation after register via external login

HaoK:
Fixes https://github.com/aspnet/AspNetCore/issues/5767

Note this doesn't cause any change in the default experience as the confirmation email is sent silently (and by default its a no-op anyways)  But dropping in a real email sender will just light it up

@ajcvickers @blowdart

### DIFF
--- a/src/Identity/UI/src/Areas/Identity/Pages/V3/Account/ExternalLogin.cshtml.cs
+++ b/src/Identity/UI/src/Areas/Identity/Pages/V3/Account/ExternalLogin.cshtml.cs
@@ -4,9 +4,11 @@
 using System;
 using System.ComponentModel.DataAnnotations;
 using System.Security.Claims;
+using System.Text.Encodings.Web;
 using System.Threading;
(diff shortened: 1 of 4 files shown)


HaoK:
Maybe its worth consolidating the SendConfirmationEmail code block into a helper, since we have 4 copies of this identical block now (boostrap3/4 + External/Register).

mkArtak (approved):

HaoK:
Weird is @mkArtak not considered an approving reviewer?

anurse:
Current failure is https://github.com/aspnet/AspNetCore/issues/8109. Working on getting it skipped.

HaoK:
@ajcvickers can you review?

ajcvickers (approved):

Ponant:
The purpose of an external login is in large part for circumventing email confirmation. So I do not understand the logic behind this besides an logic in the `UserManager`. Cheers

PR #8014, opened 2019-02-28
Update baseline test

mikaelm12:
TeamCity is complaining about permissions for paths. Updating the baseline test to use temp directory.

### DIFF
--- a/test/SharedFx.UnitTests/SharedFxTests.cs
+++ b/test/SharedFx.UnitTests/SharedFxTests.cs
@@ -65,8 +65,8 @@ namespace Microsoft.AspNetCore
                     {
                         var localAssemblyVersion = AssemblyName.GetAssemblyName(file).Version;
                         var dllName = Path.GetFileName(file);
-                        Assert.True(nugetAssemblyVersions.ContainsKey(dllName), $"Expected {dllName} to be in the downloaded dlls");
-                        Assert.True(localAssemblyVersion.CompareTo(nugetAssemblyVersions[dllName]) >= 0, $"Expected the local version of {dllName} to be greater than or equal to the already released version.");
+                        Assert.Contains(dllName, nugetAssemblyVersions.Keys);
+                        Assert.InRange(localAssemblyVersion.CompareTo(nugetAssemblyVersions[dllName]), 0, int.MaxValue);
                     }
                     catch (BadImageFormatException) { }
 
--- a/test/SharedFx.UnitTests/SharedFxTests.cs
+++ b/test/SharedFx.UnitTests/SharedFxTests.cs
@@ -37,12 +37,13 @@ namespace Microsoft.AspNetCore
 
             var zipPath = Path.Combine(AppContext.BaseDirectory, zipName);
 
-            if (!Directory.Exists(AppContext.BaseDirectory + "unzipped"))
+            var tempPath = Path.GetTempPath();
+            if (!Directory.Exists(Path.Combine(tempPath, "unzipped")))
             {
-                ZipFile.ExtractToDirectory(AppContext.BaseDirectory, "unzipped");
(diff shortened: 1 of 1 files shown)


test/SharedFx.UnitTests/SharedFxTests.cs
```diff
@@ -37,12 +37,13 @@ public async Task BaselineTest(SharedFxConfig config)
 
             var zipPath = Path.Combine(AppContext.BaseDirectory, zipName);
 
-            if (!Directory.Exists(AppContext.BaseDirectory + "unzipped"))
+            var tempPath = Path.GetTempPath();
+            if (!Directory.Exists(Path.Combine(tempPath, "unzipped")))
```
dougbu: This test will mess up if the `$env:Temp/unzipped` folder exists and contains something other than what `ZipFile` contains. Probably a good idea to use `Path.GetRandomFileName()` instead of the constant name.

Also, delete the directory if it already exists.

test/SharedFx.UnitTests/SharedFxTests.cs
```diff
@@ -70,6 +71,8 @@ public async Task BaselineTest(SharedFxConfig config)
                 catch (BadImageFormatException) { }
 
             });
+
+            Directory.Delete(Path.Combine(tempPath, "unzipped"), true);
```
dougbu: Do this cleanup in a `finally` block

dougbu (changes requested):

test/SharedFx.UnitTests/SharedFxTests.cs
ryanbrandenburg: Oh, I gave him bad advice to use `GetTempPath` because I was under the impression that the result was random rather than fixed.

mikaelm12: Okay, I'll incorporate this 👍

test/SharedFx.UnitTests/SharedFxTests.cs
```diff
@@ -36,40 +37,45 @@ public async Task BaselineTest(SharedFxConfig config)
...
-                    Assert.True(nugetAssemblyVersions.ContainsKey(dllName), $"Expected {dllName} to be in the downloaded dlls");
-                    Assert.True(localAssemblyVersion.CompareTo(nugetAssemblyVersions[dllName]) >= 0, $"Expected the local version of {dllName} to be greater than or equal to the already released version.");
-                }
-                catch (BadImageFormatException) { }
+                    try
+                    {
+                        var localAssemblyVersion = AssemblyName.GetAssemblyName(file).Version;
+                        var dllName = Path.GetFileName(file);
+                        Assert.True(nugetAssemblyVersions.ContainsKey(dllName), $"Expected {dllName} to be in the downloaded dlls");
+                        Assert.True(localAssemblyVersion.CompareTo(nugetAssemblyVersions[dllName]) >= 0, $"Expected the local version of {dllName} to be greater than or equal to the already released version.");
```
dougbu: In a similar vein, shouldn't this check be `Assert.InRange(localAssemblyVersion.CompareTo(nugetAssemblyVersions[dllName]), 0, int.MaxValue)`?

dougbu (approved):
Small changes only&hellip;

test/SharedFx.UnitTests/SharedFxTests.cs
mikaelm12: I think the `Assert.Contains` change is fine but I'd like to at least keep this Assert.True to keep the error message.

dougbu: I find custom `Assert.X(...)` error messages don't help much. But, I defer to @Eilon here&hellip;

mikaelm12: Okay, I updated it to use `Assert.InRange`. Merging now.

PR #8021, opened 2019-02-28
Handle models containing special characters

CodingSteve01:
Input-Data which contains special characters could break the output html. To avoid this behaviour the input-model is encoded.


pranavkm:
@CodingSteve01 thank you for sending this PR. However it is not entirely clear what the issue you're running in to and how this PR addresses it. Could you please log a work item detailing your scenario and why this is an appropriate fix before sending a PR?

PR #8032, opened 2019-02-28
Update ResolveReferences.targets

natemcmaster:
More recent versions of Microsoft.Bcl.Json.Sources include build targets as well as content files.

cc @ahsonkhan


dougbu (approved):

eng/targets/ResolveReferences.targets
```diff
@@ -137,7 +137,7 @@
 
       <!-- Ensure only content asset are consumed from .Sources packages -->
```
HaoK: nit: comment is now out of date

PR #8035, opened 2019-02-28
Add missing doc comments

ryanbrandenburg:


mikaelm12 (commented):
Do you also need docs for the `Controllerproperties`, `Filters` `Application` and `Attributes` properties in the `ControllerModel` class?

ryanbrandenburg:
Those items probably need comments too, but I didn't feel confident that I could accurately document them given my knowledge. Instead I documented the areas/methods I was comfortable with, figuring some documentation is better than none.

src/Middleware/StaticFiles/src/StaticFileResponseContext.cs
```diff
@@ -12,11 +12,19 @@ namespace Microsoft.AspNetCore.StaticFiles
     /// </summary>
     public class StaticFileResponseContext
     {
+        /// <summary>
+        /// Constructs the <see cref="StaticFileResponseContext"/>.
+        /// </summary>
         [Obsolete("Use the constructor that passes in the HttpContext and IFileInfo parameters: StaticFileResponseContext(HttpContext context, IFileInfo file)", false)]
```
Tratcher: Ah, it's probably time to remove this overload.

Tratcher (approved):

PR #8040, opened 2019-02-28
Update core-setup to version containing signing fix and EF to version…

pakrym:
… containing System.Data fix


dougbu (approved):
If this passes the PR checks before #7894 does, :shipit: 

But, why isn't AspNetCore-Tooling updated in this PR? And, why aren't the versions in #7894 exactly the same (I think the bot got behind but can't see why)?

pakrym:
Because I updated everything manually to minimally viable versions to avoid new bugs.

dougbu:
> Because I updated everything manually to minimally viable versions to avoid new bugs.

Actually, the way I read this, you're ahead of #7894 in at least a few places. The bot hasn't pushed new stuff since yesterday.

I'd like to understand why the bot stopped at 22 commits. (22 may be a limit it has somewhere.)

Issue #4661, opened 2018-02-09
CookiePolicyOptions sets minimumsamesitelevel to lax by default and that breaks oauth providers


javiercn:
This happens when you plug-in the cookie policy middleware on the app and use an oauth provider for authentication.

Eilon:
@blowdart - let's discuss when you're back whether we can lower the default to None.

muratg:
Triage: We can drop this to None only if we check ALL the cookies the ASP.NET stack drops to see if they have appropriate value.

brockallen:
Another idea is to change how the remote authentication handler plumbing works in ASP.NET Core handles responses: https://brockallen.com/2019/01/11/same-site-cookies-asp-net-core-and-external-authentication-providers/

Tratcher:
https://github.com/aspnet/AspNetCore/issues/8212

Issue #2675, opened 2018-01-02
Response cookies Append inconsistently applies SameSite default


aspnet-hello:
_From @Tratcher on Wednesday, December 27, 2017 9:11:02 AM_

The two Append overloads have inconsistent behavior for the samesite property that was added in 2.0.
```
                context.Response.Cookies.Append("Name1", "Value1");
                context.Response.Cookies.Append("Name2", "Value2", new CookieOptions());
```
```
Set-Cookie:Name1=Value1; path=/
Set-Cookie:Name2=Value2; path=/; samesite=lax
```

The two overloads should have the same defaults.

I noticed because we're making changes to CookiePolicy for http://github.com/aspnet/Security/issues/1561 and adding it to the templates. This will have the side-effect of causing the Append API to consistently use the second overload's behavior. 

_Copied from original issue: aspnet/HttpAbstractions#982_

muratg:
This is a bug, but fixing it is a breaking change. We'll fix this in 3.0.0.

muratg:
@Tratcher feel free to close if you feel breaking change is not worth it.

Tratcher:
SameSite has been such a mine field that the best default here would likely be None.

@blowdart

blowdart:
Yea *sigh* :( None should be the default.

Tratcher:
https://github.com/aspnet/AspNetCore/issues/8212

PR #8043, opened 2019-02-28
Change SameSite default to None

Tratcher:
#2675 #4661 Note this changes the basic infrastructure defaults but does not change any given component's behavior as each component already specified their SameSite config.

SameSite defaults:
- Session: Lax
- CookieTempDataProvider: Lax
- Antiforgery: Strict
- CookieAuth: Lax
- Twitter state cookie: Lax
- Remote auth corrilation cookie (OAuth): None
- OIDC nonce cookie: None

[WIP] Running tests to make sure I didn't miss any.

### DIFF
--- a/src/Http/Http.Abstractions/src/CookieBuilder.cs
+++ b/src/Http/Http.Abstractions/src/CookieBuilder.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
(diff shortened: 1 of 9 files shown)


blowdart:
If this doesn't affect any of our components anyway, then why do the change from "most secure" at all? We'd still be broken.

Tratcher:
The point here is that SameSite has been breaking arbitrary components that aren't aware of it so we're changing to an opt-in model. Our components have already opted in as far as they're able.

blowdart:
Opt-in to security is generally not the route we take though

HaoK (approved):
Looks fine code wise, I defer to @blowdart if this is something that's safe or good to do

Tratcher:
If it was a stable security feature I'd agree, but it's not, Apple keeps breaking it.

blowdart:
So now I ask the impossible, if it's affecting other components can I get an idea of how widespread this is?

Tratcher:
We have three datapoints:
1. The CookiePolicy SameSite default broke all remote auth and had to be [disabled](https://github.com/aspnet/AspNetCore/pull/8043/files#diff-2575214c43b0d5764ba24c1978a59c6dL61) in the templates. People still regularly break this if they don't have that line.
1. Anyone calling Response.Cookies.Append also defaulted to Lax. There haven't been as many direct reports of this issue.
1. Anyone using Lax on IOS is broken and they've had to disable it [everywhere](https://github.com/aspnet/AspNetCore/issues/4647). We've gotten a lot of hits on this one and it's not just auth cookies, it affects session, app cookies (See 2), etc..

This change fixes the first two. It doesn't fix the 3rd.

blowdart:
OK fair, I submit :)

blowdart (approved):
Looks meh to me, but what can I do? :D

anurse:
@Tratcher please rebase on `master` ASAP. This is failing due to a flaky test that has been skipped in `master`

PR #8047, opened 2019-02-28
Always load NodeHttpClient to fix SSR

BrennanConroy:
https://github.com/aspnet/AspNetCore/issues/7823


src/SignalR/clients/ts/webpack.config.base.js
```diff
@@ -40,7 +40,10 @@ module.exports = function (modulePath, browserBaseName, options) {
         },
         resolve: {
             extensions: [".ts", ".js"],
-            alias: options.alias,
+            alias: {
+                "./NodeHttpClient": path.resolve(__dirname, "signalr/src/empty.ts"),
```
BrennanConroy: Review: This saves 3kb for the browser .js files
We could just remove this and not save the bytes if we want

src/SignalR/clients/ts/signalr/src/NodeHttpClient.ts
```diff
@@ -1,23 +1,36 @@
...
 export class NodeHttpClient extends HttpClient {
     private readonly logger: ILogger;
-    private readonly request: Request.RequestAPI<Request.Request, Request.CoreOptions, Request.RequiredUriUrl>;
+    private readonly request: typeof requestModule;
     private readonly cookieJar: Request.CookieJar;
 
     public constructor(logger: ILogger) {
         super();
+        if (typeof requestModule === "undefined") {
+            throw new Error("error");
```
muratg: Better error message? :)

muratg (approved):
Looks good. Will you be sending a PR patch as well? (2.2)

src/SignalR/clients/ts/signalr/src/NodeHttpClient.ts
BrennanConroy: Whoops

src/SignalR/clients/ts/webpack.config.base.js
halter73: I think this is the right amount of clever 👍

Of course, I'd like a more first-class way to do this so we wouldn't even have to to waste bytes on a not-implemented NodeHttpClient, but this is certainly better than bundling the real thing in the browser dist.

src/SignalR/clients/ts/signalr/src/empty.ts
```diff
@@ -0,0 +1,13 @@
+import { HttpClient, HttpResponse } from "./HttpClient";
```
halter73: Nit: rename to EmptyNodeHttpClient.ts or something else more descriptive.

src/SignalR/clients/ts/signalr/src/empty.ts
```diff
@@ -0,0 +1,13 @@
...
+import { ILogger } from "./ILogger";
+
+export class NodeHttpClient extends HttpClient {
+    // @ts-ignore
+    public constructor(logger: ILogger) {
+        super();
+    }
+
+    public send(): Promise<HttpResponse> {
+        throw new Error("Method not implemented.");
```
halter73: This error message could be better too. It should probably recommend the correct way to consume the package from node.js.

anurse:
@BrennanConroy reeeeeeee~~mix~~**base**

halter73 (approved):

PR #8049, opened 2019-02-28
Add some additional logging to ErrorPageMiddlewareWebSite

pranavkm:
DeveloperExceptionMiddleware will log an error if rendering the exception page
throws. The test failure in https://github.com/aspnet/AspNetCore-Internal/issues/1730
suggests that we encountered an error like so but do not have anything further to go by.

This change adds logging to the test so we could identify possible issues

### DIFF
--- a/src/Mvc/test/Mvc.FunctionalTests/ErrorPageTests.cs
+++ b/src/Mvc/test/Mvc.FunctionalTests/ErrorPageTests.cs
@@ -33,13 +33,12 @@ namespace Microsoft.AspNetCore.Mvc.FunctionalTests
         {
             _assemblyTestLog = AssemblyTestLog.ForAssembly(GetType().Assembly);
 
-            // var loggerProvider = new XunitLoggerProvider(testOutputHelper);
             var loggerProvider = _assemblyTestLog.CreateLoggerFactory(testOutputHelper, GetType().Name);
 
-
-            var factory = fixture.Factories.FirstOrDefault() ?? fixture.WithWebHostBuilder(b => b.UseStartup<ErrorPageMiddlewareWebSite.Startup>())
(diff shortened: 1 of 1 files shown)


src/Mvc/test/Mvc.FunctionalTests/ErrorPageTests.cs
```diff
@@ -1,27 +1,45 @@
...
+            MvcTestFixture<ErrorPageMiddlewareWebSite.Startup> fixture,
+            ITestOutputHelper testOutputHelper)
         {
-            Client = fixture.CreateDefaultClient();
+            _assemblyTestLog = AssemblyTestLog.ForAssembly(GetType().Assembly);
+
+            // var loggerProvider = new XunitLoggerProvider(testOutputHelper);
+            var loggerProvider = _assemblyTestLog.CreateLoggerFactory(testOutputHelper, GetType().Name);
+
+
```
javiercn: nit: extra line

src/Mvc/test/Mvc.FunctionalTests/ErrorPageTests.cs
```diff
@@ -1,27 +1,45 @@
...
+            var factory = fixture.Factories.FirstOrDefault() ?? fixture.WithWebHostBuilder(b => b.UseStartup<ErrorPageMiddlewareWebSite.Startup>())
```
javiercn: As much as I love terseness, could you break this down a bit? Make it a GetOrCreateLocal function and do
```
if(fixture.Factories.Any()){
return fixture.Factories.First()
}
else{
return fixture.WithWebHostBuilder(builder => {
    builder.UseStartup<...>()
         .ConfigureLogging(s => ....);
}
```

javiercn (approved):

NTaylorMullen (approved):

Issue #7928, opened 2019-02-25
Create endpoint routing alternatives for SpaFallback


rynowak:
We want to add some new functionality for the "SPA fallback" scenario based on endpoint routing.

Semantics:
- A *fallback* should match with the lowest priority by default
- A *fallback* should not match things that look like filenames

Scenarios (priority order):
- Use a static file as a fallback - `routes.MapFallbackToFile("index.html")`
- Use a Razor Page as a fallback - `routes.MapFallbackToController("Index", "Home")`
- Use an MVC action/controller as a fallback - `routes.MapFallbackToPage("/Index")`

We should also make a change to the exising spa fallback middleware so that it short circuits when the request has an endpoint. This will make it work with endpoint routing, even though we will recommend the new options.

mkArtakMSFT:
Thanks Ryan!

PR #8052, opened 2019-03-01
Add fallback routing

rynowak:
Addresses #7928

Adds the idea of a *fallback* route to routing - as an opinionated way to set things up. 

Adds extension methods for:
- *fallback to generic delegate*
- *fallback to static file*
- *fallback to controller/page*

### DIFF
--- a/src/Http/Routing/src/Builder/FallbackEndpointRouteBuilderExtensions.cs
+++ b/src/Http/Routing/src/Builder/FallbackEndpointRouteBuilderExtensions.cs
@@ -4,7 +4,6 @@
 using System;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Routing;
-using Microsoft.AspNetCore.Routing.Patterns;
 
 namespace Microsoft.AspNetCore.Builder
 {
--- a/src/Http/Routing/src/Builder/FallbackEndpointRouteBuilderExtensions.cs
+++ b/src/Http/Routing/src/Builder/FallbackEndpointRouteBuilderExtensions.cs
@@ -1,0 +1,96 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing;
+using Microsoft.AspNetCore.Routing.Patterns;
+
+namespace Microsoft.AspNetCore.Builder
+{
+    /// <summary>
+    /// Contains extension methods for <see cref="IEndpointRouteBuilder"/>.
+    /// </summary>
+    public static class FallbackEndpointRouteBuilderExtensions
+    {
+        /// <summary>
+        /// Adds a specialized <see cref="RouteEndpoint"/> to the <see cref="IEndpointRouteBuilder"/> that will match
+        /// requests for non-file-names with the lowest possible priority.
+        /// </summary>
+        /// <param name="builder">The <see cref="IEndpointRouteBuilder"/> to add the route to.</param>
+        /// <param name="requestDelegate">The delegate executed when the endpoint is matched.</param>
+        /// <returns>A <see cref="IEndpointConventionBuilder"/> that can be used to further customize the endpoint.</returns>
+        /// <remarks>
+        /// <para>
+        /// <see cref="MapFallback(IEndpointRouteBuilder, RequestDelegate)"/> is intended to handle cases where URL path of
+        /// the request does not contain a file name, and no other endpoint has matched. This is convenient for routing
+        /// requests for dynamic content to a SPA framework, while also allowing requests for non-existent files to
+        /// result in an HTTP 404.
+        /// </para>
+        /// <para>
+        /// <see cref="MapFallback(IEndpointRouteBuilder, RequestDelegate)"/> registers an endpoint using the pattern
+        /// <c>{*path:nonfile}</c>. The order of the registered endpoint will be <c>int.MaxValue</c>.
+        /// </para>
+        /// </remarks>
+        public static IEndpointConventionBuilder MapFallback(this IEndpointRouteBuilder builder, RequestDelegate requestDelegate)
+        {
+            if (builder == null)
+            {
+                throw new ArgumentNullException(nameof(builder));
+            }
+
+            if (requestDelegate == null)
+            {
+                throw new ArgumentNullException(nameof(requestDelegate));
+            }
+
+            return builder.MapFallback("{*path:nonfile}", requestDelegate);
+        }
+
+        /// <summary>
+        /// Adds a specialized <see cref="RouteEndpoint"/> to the <see cref="IEndpointRouteBuilder"/> that will match
+        /// the provided pattern with the lowest possible priority.
+        /// </summary>
+        /// <param name="builder">The <see cref="IEndpointRouteBuilder"/> to add the route to.</param>
+        /// <param name="pattern">The route pattern.</param>
+        /// <param name="requestDelegate">The delegate executed when the endpoint is matched.</param>
+        /// <returns>A <see cref="IEndpointConventionBuilder"/> that can be used to further customize the endpoint.</returns>
+        /// <remarks>
+        /// <para>
+        /// <see cref="MapFallback(IEndpointRouteBuilder, string, RequestDelegate)"/> is intended to handle cases where no
+        /// other endpoint has matched. This is convenient for routing requests to a SPA framework.
+        /// </para>
+        /// <para>
+        /// The order of the registered endpoint will be <c>int.MaxValue</c>.
+        /// </para>
+        /// <para>
+        /// This overload will use the provided <paramref name="pattern"/> verbatim. Use the <c>:nonfile</c> route contraint
+        /// to exclude requests for static files.
+        /// </para>
+        /// </remarks>
+        public static IEndpointConventionBuilder MapFallback(
+            this IEndpointRouteBuilder builder,
+            string pattern,
+            RequestDelegate requestDelegate)
+        {
+            if (builder == null)
+            {
+                throw new ArgumentNullException(nameof(builder));
+            }
+
+            if (pattern == null)
+            {
+                throw new ArgumentNullException(nameof(pattern));
+            }
+
+            if (requestDelegate == null)
+            {
+                throw new ArgumentNullException(nameof(requestDelegate));
+            }
+
+            var conventionBuilder = builder.Map(pattern, "Fallback " + pattern, requestDelegate);
+            conventionBuilder.Add(b => ((RouteEndpointBuilder)b).Order = int.MaxValue);
+            return conventionBuilder;
+        }
+    }
+}
--- a/src/Http/Routing/src/Constraints/FileNameRouteConstraint.cs
+++ b/src/Http/Routing/src/Constraints/FileNameRouteConstraint.cs
@@ -1,0 +1,148 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Globalization;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Routing.Constraints
+{
+    /// <summary>
+    /// Constrains a route parameter to represent only file name values. Does not validate that
+    /// the route value contains valid file system characters, or that the value represents
+    /// an actual file on disk.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// This constraint can be used to disambiguate requests for static files versus dynamic
+    /// content served from the application.
+    /// </para>
+    /// <para>
+    /// This constraint determines whether a route value represents a file name by examining
+    /// the last URL Path segment of the value (delimited by <c>/</c>). The last segment
+    /// must contain the dot (<c>.</c>) character followed by one or more non-(<c>.</c>) characters.
+    /// </para>
+    /// <para>
+    /// If the route value does not contain a <c>/</c> then the entire value will be interpreted
+    /// as the last segment.
+    /// </para>
+    /// <para>
+    /// The <see cref="FileNameRouteConstraint"/> does not attempt to validate that the value contains
+    /// a legal file name for the current operating system.
+    /// </para>
+    /// <para>
+    /// The <see cref="FileNameRouteConstraint"/> does not attempt to validate that the value represents
+    /// an actual file on disk.
+    /// </para>
+    /// <para>
+    /// <list type="bullet">  
+    ///     <listheader>  
+    ///         <term>Examples of route values that will be matched as file names</term>  
+    ///         <description>description</description>  
+    ///     </listheader>  
+    ///     <item>  
+    ///         <term><c>/a/b/c.txt</c></term>  
+    ///         <description>Final segment contains a <c>.</c> followed by other characters.</description>  
+    ///     </item>
+    ///     <item>  
+    ///         <term><c>/hello.world.txt</c></term>  
+    ///         <description>Final segment contains a <c>.</c> followed by other characters.</description>  
+    ///     </item>
+    ///     <item>  
+    ///         <term><c>hello.world.txt</c></term>  
+    ///         <description>Final segment contains a <c>.</c> followed by other characters.</description>  
+    ///     </item>
+    ///     <item>  
+    ///         <term><c>.gitignore</c></term>  
+    ///         <description>Final segment contains a <c>.</c> followed by other characters.</description>  
+    ///     </item> 
+    /// </list>
+    /// <list type="bullet">  
+    ///     <listheader>  
+    ///         <term>Examples of route values that will be rejected as non-file-names</term>  
+    ///         <description>description</description>  
+    ///     </listheader>  
+    ///     <item>  
+    ///         <term><c>/a/b/c</c></term>  
+    ///         <description>Final segment does not contain a <c>.</c>.</description>  
+    ///     </item>
+    ///     <item>  
+    ///         <term><c>/a/b.d/c</c></term>  
+    ///         <description>Final segment does not contain a <c>.</c>.</description>  
+    ///     </item>
+    ///     <item>  
+    ///         <term><c>/a/b.d/c/</c></term>  
+    ///         <description>Final segment is empty.</description>  
+    ///     </item>
+    ///     <item>  
+    ///         <term><c></c></term>  
+    ///         <description>Value is empty</description>  
+    ///     </item>
+    /// </list>  
+    /// </para>
+    /// </remarks>
+    public class FileNameRouteConstraint : IRouteConstraint
+    {
+        /// <inheritdoc />
+        public bool Match(
+            HttpContext httpContext,
+            IRouter route,
+            string routeKey,
+            RouteValueDictionary values,
+            RouteDirection routeDirection)
+        {
+            if (routeKey == null)
+            {
+                throw new ArgumentNullException(nameof(routeKey));
+            }
+
+            if (values == null)
+            {
+                throw new ArgumentNullException(nameof(values));
+            }
+
+            if (values.TryGetValue(routeKey, out var obj) && obj != null)
+            {
+                var value = Convert.ToString(obj, CultureInfo.InvariantCulture);
+                return IsFileName(value);
+            }
+
+            // No value or null value.
+            return false;
+        }
+
+        // This is used both here and in NonFileNameRouteConstraint
+        // Any changes to this logic need to update the docs in those places.
+        internal static bool IsFileName(ReadOnlySpan<char> value)
+        {
+            if (value.Length == 0)
+            {
+                // Not a file name because empty.
+                return false;
+            }
+
+            var lastSlashIndex = value.LastIndexOf('/');
+            if (lastSlashIndex >= 0)
+            {
+                value = value.Slice(lastSlashIndex + 1);
+            }
+
+            var dotIndex = value.IndexOf('.');
+            if (dotIndex == -1)
+            {
+                // No dot.
+                return false;
+            }
+
+            for (var i = dotIndex + 1; i < value.Length; i++)
+            {
+                if (value[i] != '.')
+                {
+                    return true;
+                }
+            }
+
+            return false;
+        }
+    }
+}
--- a/src/Http/Routing/src/Constraints/NonFileNameRouteConstraint.cs
+++ b/src/Http/Routing/src/Constraints/NonFileNameRouteConstraint.cs
@@ -1,0 +1,114 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Globalization;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Routing.Constraints
+{
+    /// <summary>
+    /// Constrains a route parameter to represent only non-file-name values. Does not validate that
+    /// the route value contains valid file system characters, or that the value represents
+    /// an actual file on disk.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// This constraint can be used to disambiguate requests for dynamic content versus
+    /// static files served from the application.
+    /// </para>
+    /// <para>
+    /// This constraint determines whether a route value represents a file name by examining
+    /// the last URL Path segment of the value (delimited by <c>/</c>). The last segment
+    /// must contain the dot (<c>.</c>) character followed by one or more non-(<c>.</c>) characters.
+    /// </para>
+    /// <para>
+    /// If the route value does not contain a <c>/</c> then the entire value will be interpreted
+    /// as a the last segment.
+    /// </para>
+    /// <para>
+    /// The <see cref="NonFileNameRouteConstraint"/> does not attempt to validate that the value contains
+    /// a legal file name for the current operating system.
+    /// </para>
+    /// <para>
+    /// <list type="bullet">  
+    ///     <listheader>  
+    ///         <term>Examples of route values that will be matched as non-file-names</term>  
+    ///         <description>description</description>  
+    ///     </listheader>  
+    ///     <item>  
+    ///         <term><c>/a/b/c</c></term>  
+    ///         <description>Final segment does not contain a <c>.</c>.</description>  
+    ///     </item>
+    ///     <item>  
+    ///         <term><c>/a/b.d/c</c></term>  
+    ///         <description>Final segment does not contain a <c>.</c>.</description>  
+    ///     </item>
+    ///     <item>  
+    ///         <term><c>/a/b.d/c/</c></term>  
+    ///         <description>Final segment is empty.</description>  
+    ///     </item>
+    ///     <item>  
+    ///         <term><c></c></term>  
+    ///         <description>Value is empty</description>  
+    ///     </item>
+    /// </list>
+    /// <list type="bullet">  
+    ///     <listheader>  
+    ///         <term>Examples of route values that will be rejected as file names</term>  
+    ///         <description>description</description>  
+    ///     </listheader>  
+    ///     <item>  
+    ///         <term><c>/a/b/c.txt</c></term>  
+    ///         <description>Final segment contains a <c>.</c> followed by other characters.</description>  
+    ///     </item>
+    ///     <item>  
+    ///         <term><c>/hello.world.txt</c></term>  
+    ///         <description>Final segment contains a <c>.</c> followed by other characters.</description>  
+    ///     </item>
+    ///     <item>  
+    ///         <term><c>hello.world.txt</c></term>  
+    ///         <description>Final segment contains a <c>.</c> followed by other characters.</description>  
+    ///     </item>
+    ///     <item>  
+    ///         <term><c>.gitignore</c></term>  
+    ///         <description>Final segment contains a <c>.</c> followed by other characters.</description>  
+    ///     </item> 
+    /// </list>
+    /// </para>
+    /// </remarks>
+    public class NonFileNameRouteConstraint : IRouteConstraint
+    {
+        /// <inheritdoc />
+        public bool Match(
+            HttpContext httpContext,
+            IRouter route,
+            string routeKey,
+            RouteValueDictionary values,
+            RouteDirection routeDirection)
+        {
+            if (routeKey == null)
+            {
+                throw new ArgumentNullException(nameof(routeKey));
+            }
+
+            if (values == null)
+            {
+                throw new ArgumentNullException(nameof(values));
+            }
+
+            if (values.TryGetValue(routeKey, out var obj) && obj != null)
+            {
+                var value = Convert.ToString(obj, CultureInfo.InvariantCulture);
+                return !FileNameRouteConstraint.IsFileName(value);
+            }
+
+            // No value or null value.
+            //
+            // We want to return true here because the core use-case of the constraint is to *exclude*
+            // things that look like file names. There's nothing here that looks like a file name, so
+            // let it through.
+            return true;
+        }
+    }
+}
--- a/src/Http/Routing/src/DataSourceDependentCache.cs
+++ b/src/Http/Routing/src/DataSourceDependentCache.cs
@@ -8,6 +8,8 @@ using Microsoft.AspNetCore.Http;
 
 namespace Microsoft.AspNetCore.Routing
 {
+    // FYI: This class is also linked into MVC. If you make changes to the API you will
+    // also need to change MVC's usage.
     internal sealed class DataSourceDependentCache<T> : IDisposable where T : class
     {
         private readonly EndpointDataSource _dataSource;
--- a/src/Http/Routing/test/UnitTests/Constraints/FIleNameRouteConstraintTest.cs
+++ b/src/Http/Routing/test/UnitTests/Constraints/FIleNameRouteConstraintTest.cs
@@ -1,0 +1,100 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Xunit;
+
+namespace Microsoft.AspNetCore.Routing.Constraints
+{
+    public class FileNameRouteConstraintTest
+    {
+        public static TheoryData<object> FileNameData
+        {
+            get
+            {
+                return new TheoryData<object>()
+                {
+                    "hello.txt",
+                    "hello.txt.jpg",
+                    "/hello.t",
+                    "/////hello.x",
+                    "a/b/c/d.e",
+                    "a/b./.c/d.e",
+                    ".gitnore",
+                    ".a",
+                    "/.......a"
+                };
+            }
+        }
+
+
+        [Theory]
+        [MemberData(nameof(FileNameData))]
+        public void Match_RouteValue_IsFileName(object value)
+        {
+            // Arrange
+            var constraint = new FileNameRouteConstraint();
+
+            var values = new RouteValueDictionary();
+            values.Add("path", value);
+
+            // Act
+            var result = constraint.Match(httpContext: null, route: null, "path", values, RouteDirection.IncomingRequest);
+
+            // Assert
+            Assert.True(result);
+        }
+
+        public static TheoryData<object> NonFileNameData
+        {
+            get
+            {
+                return new TheoryData<object>()
+                {
+                    null,
+                    string.Empty,
+                    "/",
+                    ".",
+                    "..........",
+                    "hello.",
+                    "/hello",
+                    "//",
+                    "//b.c/",
+                    "/////hello.",
+                    "a/b./.c/d.",
+                };
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(NonFileNameData))]
+        public void Match_RouteValue_IsNotFileName(object value)
+        {
+            // Arrange
+            var constraint = new FileNameRouteConstraint();
+
+            var values = new RouteValueDictionary();
+            values.Add("path", value);
+
+            // Act
+            var result = constraint.Match(httpContext: null, route: null, "path", values, RouteDirection.IncomingRequest);
+
+            // Assert
+            Assert.False(result);
+        }
+
+        [Fact]
+        public void Match_MissingValue_IsNotFileName()
+        {
+            // Arrange
+            var constraint = new FileNameRouteConstraint();
+
+            var values = new RouteValueDictionary();
+
+            // Act
+            var result = constraint.Match(httpContext: null, route: null, "path", values, RouteDirection.IncomingRequest);
+
+            // Assert
+            Assert.False(result);
+        }
+    }
+}
--- a/src/Middleware/SpaServices.Extensions/src/SpaDefaultPageMiddleware.cs
+++ b/src/Middleware/SpaServices.Extensions/src/SpaDefaultPageMiddleware.cs
@@ -3,6 +3,7 @@
 
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Http.Endpoints;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 using System;
@@ -24,6 +25,12 @@ namespace Microsoft.AspNetCore.SpaServices
             // Rewrite all requests to the default page
             app.Use((context, next) =>
             {
+                // If we have an Endpoint, then this is a deferred match - just noop.
+                if (context.GetEndpoint() != null)
+                {
+                    return next();
+                }
+
                 context.Request.Path = options.DefaultPage;
                 return next();
             });
@@ -39,6 +46,12 @@ namespace Microsoft.AspNetCore.SpaServices
             // present on disk), the SPA is definitely not going to work.
             app.Use((context, next) =>
             {
+                // If we have an Endpoint, then this is a deferred match - just noop.
+                if (context.GetEndpoint() != null)
+                {
+                    return next();
+                }
+
                 var message = "The SPA default page middleware could not return the default page " +
                     $"'{options.DefaultPage}' because it was not found, and no other middleware " +
                     "handled the request.\n";
--- a/src/Middleware/SpaServices.Extensions/src/StaticFiles/SpaStaticFilesExtensions.cs
+++ b/src/Middleware/SpaServices.Extensions/src/StaticFiles/SpaStaticFilesExtensions.cs
@@ -1,7 +1,8 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Http.Endpoints;
 using Microsoft.AspNetCore.SpaServices.StaticFiles;
 using Microsoft.Extensions.FileProviders;
 using Microsoft.Extensions.Options;
@@ -110,7 +111,6 @@ namespace Microsoft.Extensions.DependencyInjection
                 }
             }
 
-            
             app.UseStaticFiles(staticFileOptions);
         }
 
--- a/src/Middleware/StaticFiles/ref/Microsoft.AspNetCore.StaticFiles.csproj
+++ b/src/Middleware/StaticFiles/ref/Microsoft.AspNetCore.StaticFiles.csproj
@@ -7,6 +7,7 @@
     <Compile Include="Microsoft.AspNetCore.StaticFiles.netcoreapp3.0.cs" />
     <Reference Include="Microsoft.AspNetCore.Hosting.Abstractions"  />
     <Reference Include="Microsoft.AspNetCore.Http.Extensions"  />
+    <Reference Include="Microsoft.AspNetCore.Routing"  />
     <Reference Include="Microsoft.Extensions.FileProviders.Abstractions"  />
     <Reference Include="Microsoft.Extensions.Logging.Abstractions"  />
     <Reference Include="Microsoft.Extensions.WebEncoders"  />
--- a/src/Middleware/StaticFiles/src/LoggerExtensions.cs
+++ b/src/Middleware/StaticFiles/src/LoggerExtensions.cs
@@ -24,6 +24,7 @@ namespace Microsoft.AspNetCore.StaticFiles
         private static Action<ILogger, StringValues, string, Exception> _sendingFileRange;
         private static Action<ILogger, StringValues, string, Exception> _copyingFileRange;
         private static Action<ILogger, Exception> _writeCancelled;
+        private static Action<ILogger, Exception> _endpointMatched;
 
         static LoggerExtensions()
         {
@@ -75,6 +76,10 @@ namespace Microsoft.AspNetCore.StaticFiles
                 logLevel: LogLevel.Debug,
                 eventId: new EventId(14, "WriteCancelled"),
                 formatString: "The file transmission was cancelled");
+            _endpointMatched = LoggerMessage.Define(
+                logLevel: LogLevel.Debug,
+                eventId: new EventId(15, "EndpointMatched"),
+                formatString: "Static files was skipped as the request already matched an endpoint.");
         }
 
         public static void RequestMethodNotSupported(this ILogger logger, string method)
@@ -89,6 +94,11 @@ namespace Microsoft.AspNetCore.StaticFiles
                 physicalPath = "N/A";
             }
             _fileServed(logger, virtualPath, physicalPath, null);
+        }
+
+        public static void EndpointMatched(this ILogger logger)
+        {
+            _endpointMatched(logger, null);
         }
 
         public static void PathMismatch(this ILogger logger, string path)
--- a/src/Middleware/StaticFiles/src/StaticFileMiddleware.cs
+++ b/src/Middleware/StaticFiles/src/StaticFileMiddleware.cs
@@ -72,7 +72,11 @@ namespace Microsoft.AspNetCore.StaticFiles
         {
             var fileContext = new StaticFileContext(context, _options, _matchUrl, _logger, _fileProvider, _contentTypeProvider);
 
-            if (!fileContext.ValidateMethod())
+            if (!fileContext.ValidateNoEndpoint())
+            {
+                _logger.EndpointMatched();
+            }
+            else if (!fileContext.ValidateMethod())
             {
                 _logger.RequestMethodNotSupported(context.Request.Method);
             }
--- a/src/Middleware/StaticFiles/src/StaticFilesEndpointRouteBuilderExtensions.cs
+++ b/src/Middleware/StaticFiles/src/StaticFilesEndpointRouteBuilderExtensions.cs
@@ -1,0 +1,223 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Endpoints;
+using Microsoft.AspNetCore.Routing;
+using Microsoft.AspNetCore.StaticFiles;
+
+namespace Microsoft.AspNetCore.Builder
+{
+    /// <summary>
+    /// Contains extension methods for using static files with endpoint routing.
+    /// </summary>
+    public static class StaticFilesEndpointRouteBuilderExtensions
+    {
+        /// <summary>
+        /// Adds a specialized <see cref="RouteEndpoint"/> to the <see cref="IEndpointRouteBuilder"/> that will match
+        /// requests for non-filenames with the lowest possible priority. The request will be routed to a
+        /// <see cref="StaticFileMiddleware"/> that attempts to serve the file specified by <paramref name="filePath"/>.
+        /// </summary>
+        /// <param name="builder">The <see cref="IEndpointRouteBuilder"/>.</param>
+        /// <param name="filePath">The file path of the file to serve.</param>
+        /// <returns>The <see cref="IEndpointRouteBuilder"/></returns>
+        /// <remarks>
+        /// <para>
+        /// <see cref="MapFallbackToFile(IEndpointRouteBuilder, string)"/> is intended to handle cases where URL path of
+        /// the request does not contain a filename, and no other endpoint has matched. This is convenient for routing
+        /// requests for dynamic content to a SPA framework, while also allowing requests for non-existent files to
+        /// result in an HTTP 404.
+        /// </para>
+        /// <para>
+        /// The default <see cref="StaticFileOptions"/> for the <see cref="StaticFileMiddleware"/> will be used.
+        /// </para>
+        /// <para>
+        /// <see cref="MapFallbackToFile(IEndpointRouteBuilder, string)"/> registers an endpoint using the pattern
+        /// <c>{*path:nonfile}</c>. The order of the registered endpoint will be <c>int.MaxValue</c>.
+        /// </para>
+        /// </remarks>
+        public static IEndpointConventionBuilder MapFallbackToFile(
+            this IEndpointRouteBuilder builder,
+            string filePath)
+        {
+            if (builder == null)
+            {
+                throw new ArgumentNullException(nameof(builder));
+            }
+
+            if (filePath == null)
+            {
+                throw new ArgumentNullException(nameof(filePath));
+            }
+
+            return builder.MapFallback(CreateRequestDelegate(builder, filePath));
+        }
+
+        /// <summary>
+        /// Adds a specialized <see cref="RouteEndpoint"/> to the <see cref="IEndpointRouteBuilder"/> that will match
+        /// requests for non-filenames with the lowest possible priority. The request will be routed to a
+        /// <see cref="StaticFileMiddleware"/> that attempts to serve the file specified by <paramref name="filePath"/>.
+        /// </summary>
+        /// <param name="builder">The <see cref="IEndpointRouteBuilder"/>.</param>
+        /// <param name="filePath">The file path of the file to serve.</param>
+        /// <param name="options"><see cref="StaticFileOptions"/> for the <see cref="StaticFileMiddleware"/>.</param>
+        /// <returns>The <see cref="IEndpointRouteBuilder"/></returns>
+        /// <remarks>
+        /// <para>
+        /// <see cref="MapFallbackToFile(IEndpointRouteBuilder, string, StaticFileOptions)"/> is intended to handle cases
+        /// where URL path of the request does not contain a file name, and no other endpoint has matched. This is convenient
+        /// for routing requests for dynamic content to a SPA framework, while also allowing requests for non-existent files to
+        /// result in an HTTP 404.
+        /// </para>
+        /// <para>
+        /// <see cref="MapFallbackToFile(IEndpointRouteBuilder, string, StaticFileOptions)"/> registers an endpoint using the pattern
+        /// <c>{*path:nonfile}</c>. The order of the registered endpoint will be <c>int.MaxValue</c>.
+        /// </para>
+        /// </remarks>
+        public static IEndpointConventionBuilder MapFallbackToFile(
+            this IEndpointRouteBuilder builder,
+            string filePath,
+            StaticFileOptions options)
+        {
+            if (builder == null)
+            {
+                throw new ArgumentNullException(nameof(builder));
+            }
+
+            if (filePath == null)
+            {
+                throw new ArgumentNullException(nameof(filePath));
+            }
+
+            return builder.MapFallback(CreateRequestDelegate(builder, filePath, options));
+        }
+
+        /// <summary>
+        /// Adds a specialized <see cref="RouteEndpoint"/> to the <see cref="IEndpointRouteBuilder"/> that will match
+        /// requests for non-filenames with the lowest possible priority. The request will be routed to a
+        /// <see cref="StaticFileMiddleware"/> that attempts to serve the file specified by <paramref name="filePath"/>.
+        /// </summary>
+        /// <param name="builder">The <see cref="IEndpointRouteBuilder"/>.</param>
+        /// <param name="pattern">The route pattern.</param>
+        /// <param name="filePath">The file path of the file to serve.</param>
+        /// <returns>The <see cref="IEndpointRouteBuilder"/></returns>
+        /// <remarks>
+        /// <para>
+        /// <see cref="MapFallbackToFile(IEndpointRouteBuilder, string, string)"/> is intended to handle
+        /// cases where URL path of the request does not contain a filename, and no other endpoint has matched. This is
+        /// convenient for routing requests for dynamic content to a SPA framework, while also allowing requests for
+        /// non-existent files to result in an HTTP 404.
+        /// </para>
+        /// <para>
+        /// The default <see cref="StaticFileOptions"/> for the <see cref="StaticFileMiddleware"/> will be used.
+        /// </para>
+        /// <para>
+        /// The order of the registered endpoint will be <c>int.MaxValue</c>.
+        /// </para>
+        /// <para>
+        /// This overload will use the provided <paramref name="pattern"/> verbatim. Use the <c>:nonfile</c> route contraint
+        /// to exclude requests for static files.
+        /// </para>
+        /// </remarks>
+        public static IEndpointConventionBuilder MapFallbackToFile(
+            this IEndpointRouteBuilder builder,
+            string pattern,
+            string filePath)
+        {
+            if (builder == null)
+            {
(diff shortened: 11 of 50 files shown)


rynowak:
It's probably a good idea to look at individual commits because the changes are all over.

src/Http/Routing/src/Builder/FallbackEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,91 @@
...
+        /// <see cref="MapFallback(IEndpointRouteBuilder, RequestDelegate)"/> is intended to handle cases where URL path of
+        /// the request does not contain a filename, and no other endpoint has matched. This is convenient for routing
+        /// requests for dynamic content to a SPA framework, while also allowing requests for non-existent files to
+        /// result in an HTTP 404.
+        /// </para>
+        /// <para>
+        /// <see cref="MapFallback(IEndpointRouteBuilder, RequestDelegate)"/> registers an endpoint using the pattern
+        /// <c>{*path:nonfile}</c>. The order of the registered endpoint will be <c>int.MaxValue</c>.
+        /// </para>
+        /// </remarks>
```
rynowak: The main scenario here is carried over from spa services routing things, and represents the idea that you want the static files middleware to be responsible for things that **look like** files, and you want routing to handle things that **don't**. It's an opinionated way to do a bunch of useful things.

rynowak: The methods here are layered on top of existing primitives (ordering, constraints).

src/Middleware/StaticFiles/ref/Microsoft.AspNetCore.StaticFiles.csproj
```diff
@@ -7,6 +7,7 @@
     <Compile Include="Microsoft.AspNetCore.StaticFiles.netcoreapp3.0.cs" />
     <Reference Include="Microsoft.AspNetCore.Hosting.Abstractions"  />
     <Reference Include="Microsoft.AspNetCore.Http.Extensions"  />
+    <Reference Include="Microsoft.AspNetCore.Routing"  />
```
rynowak: @Tratcher - thoughts? What's the right place for these extension methods to live.

src/Middleware/StaticFiles/src/StaticFilesEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,223 @@
...
+        /// <para>
+        /// This overload will use the provided <paramref name="pattern"/> verbatim. Use the <c>:nonfile</c> route contraint
+        /// to exclude requests for static files.
+        /// </para>
+        /// </remarks>
+        public static IEndpointConventionBuilder MapFallbackToFile(
+            this IEndpointRouteBuilder builder,
+            string pattern,
+            string filePath,
+            StaticFileOptions options)
```
rynowak: @SteveSandersonMS - I carried forward the options from the spa services stuff because it seemed like there was a lot of complexity there that I don't totally get.

I'm interested in whether you think these things can be used to solve the same set of problems or not.

src/Middleware/SpaServices.Extensions/src/SpaDefaultPageMiddleware.cs
```diff
@@ -24,6 +25,12 @@ public static void Attach(ISpaBuilder spaBuilder)
             // Rewrite all requests to the default page
             app.Use((context, next) =>
             {
+                // If we have an Endpoint, then this is a deferred match - just noop.
+                if (context.GetEndpoint() != null)
+                {
+                    return next();
+                }
+
```
rynowak: I coulnd't find any tests. #YOLO

src/Http/Routing/src/DataSourceDependentCache.cs
```diff
@@ -8,6 +8,8 @@
 
 namespace Microsoft.AspNetCore.Routing
 {
+    // FYI: This class is also linked into MVC. If you make changes to the API you will
+    // also need to change MVC's usage.
     internal sealed class DataSourceDependentCache<T> : IDisposable where T : class
```
rynowak: This commit has a bunch of random cleanup because I'm trying to share the core data structure used in action selection in a few places.

src/Mvc/Mvc.Core/src/Infrastructure/ActionSelectionTable.cs
```diff
@@ -0,0 +1,155 @@
...
+                //
+                // `entries` will appear in `OrdinalIgnoreCaseEntries` exactly once and in `OrdinalEntries` once
+                // for each variation of casing that we've seen.
+                if (!ordinalEntries.ContainsKey(routeValues))
+                {
+                    ordinalEntries.Add(routeValues, entries);
+                }
+            }
+
+            return new ActionSelectionTable<T>(version, routeKeys.ToArray(), ordinalEntries, ordinalIgnoreCaseEntries);
```
rynowak: This is all existing code I ripped out of action selector.

src/Mvc/Mvc.Core/src/Infrastructure/StringArrayComparer.cs
```diff
@@ -0,0 +1,79 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.Extensions.Internal;
+
+namespace Microsoft.AspNetCore.Mvc.Infrastructure
+{
+    internal class StringArrayComparer : IEqualityComparer<string[]>
```
rynowak: This is all existing code I ripped out of action selector.

src/Mvc/Mvc.Core/src/Routing/DynamicControllerEndpointMatcherPolicy.cs
```diff
@@ -0,0 +1,113 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing;
+using Microsoft.AspNetCore.Routing.Matching;
+
+namespace Microsoft.AspNetCore.Mvc.Routing
+{
+    internal class DynamicControllerEndpointMatcherPolicy : MatcherPolicy, IEndpointSelectorPolicy
```
rynowak: There are a few different design approaches I could have taken with MVC:

1. Add more nodes to the DFA stuff by doing an application model thing
2. Use the fallback endpoint's delegate to *chain* to an MVC endpoint
3. Use a matcher policy to substitute in another endpoint dynamically.

I rejected 1 because I don't really want this new routing feature to hang off of MVC options/conventions. Another idea that I could do would be to build this into the data sources themselves, which might be fine?

I rejected 2 because it has problems with things like authZ. You want the policies applied to the endpoint to be applied by middleware and that's only possible if endpoint is correct when middleware run.

I did 3 because I think we're **going to need this functionality**. I have another work item planned to implement features that people will use to port their `IRouter` code forward. What's enabled by this change is 100% needed for that.

-----

I'm looking for input on this because there are a few real tradeoffs here in terms of what's possible and how future enhancements will play out.

rynowak: Also @davidfowl I summon you.

src/Mvc/Mvc.RazorPages/src/Infrastructure/DynamicPageEndpointMatcherPolicy.cs
```diff
@@ -0,0 +1,113 @@
...
+            for (var i = 0; i < candidates.Count; i++)
+            {
+                var endpoint = candidates[i].Endpoint;
+
+                var metadata = endpoint.Metadata.GetMetadata<DynamicPageMetadata>();
+                if (metadata == null)
+                {
+                    continue;
+                }
+
```
rynowak: Note to myself, this isn't done. I need to use the pageloader inside the selector.

src/Mvc/Mvc.RazorPages/src/Infrastructure/DynamicPageEndpointMatcherPolicy.cs
```diff
@@ -0,0 +1,113 @@
...
+                foreach (var kvp in matchedValues)
+                {
+                    values.TryAdd(kvp.Key, kvp.Value);
+                }
+
+                candidates.ReplaceEndpoint(i, replacement, values);
+            }
+        }
+    }
+}
```
rynowak: There's a certain amount of de-duplication that could be done between these two pages (controller and pages). 

I'm not incredibly keen because to a certain extent these code paths need to each have their own types for everything because of DI/metadata. Open to suggestions/comments 😆

src/Mvc/samples/MvcSandbox/Startup.cs
```diff
@@ -68,6 +68,8 @@ public void Configure(IApplicationBuilder app)
 
                 builder.MapControllers();
                 builder.MapRazorPages();
+
+                builder.MapFallbackToController("Index", "Home");
```
rynowak: example usage.

src/Mvc/Mvc.RazorPages/src/Builder/RazorPagesEndpointRouteBuilderExtensions.cs
```diff
@@ -31,6 +32,245 @@ public static IEndpointConventionBuilder MapRazorPages(this IEndpointRouteBuilde
...
+            GetOrCreateDataSource(routes);
+
+            // Maps a fallback endpoint with an empty delegate. This is OK because
+            // we don't expect the delegate to run. 
+            routes.MapFallback(context => Task.CompletedTask).Add(b =>
+            {
+                // MVC registers a policy that looks for this metadata.
+                b.Metadata.Add(CreateDynamicPageMetadata(page, area));
+            });
+        }
```
rynowak: I'm really please how much powerful stuff can be built here with the building blocks we have available (ordering, constraints, policies, endpoints). None of these features required new features, it's all existing extensibility. @JamesNK - you're the best around.

src/Http/Routing/test/UnitTests/Constraints/FIleNameRouteConstraintTest.cs
```diff
@@ -0,0 +1,99 @@
...
+            {
+                return new TheoryData<object>()
+                {
+                    "hello.txt",
+                    "hello.txt.jpg",
+                    "/hello.t",
+                    "/////hello.x",
+                    "a/b/c/d.e",
+                    "a/b./.c/d.e",
+                };
```
javiercn: I would also add a leading dot combinations like `.gitignore`

src/Http/Routing/test/UnitTests/Constraints/FIleNameRouteConstraintTest.cs
```diff
@@ -0,0 +1,99 @@
...
+        {
+            get
+            {
+                return new TheoryData<object>()
+                {
+                    null,
+                    string.Empty,
+                    "/",
+                    ".",
+                    ".a",
```
javiercn: Seems like you are not considering this a file, what's the reason?

src/Middleware/StaticFiles/src/StaticFilesEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,223 @@
...
+
+            return builder.MapFallback(pattern, CreateRequestDelegate(builder, filePath, options));
+        }
+
+        private static RequestDelegate CreateRequestDelegate(
+            IEndpointRouteBuilder builder,
+            string filePath,
+            StaticFileOptions options = null)
+        {
+            var app = builder.CreateApplicationBuilder();
```
javiercn: This is neat, I didn't know you could do this.

src/Mvc/Mvc.Core/src/Routing/DynamicControllerEndpointMatcherPolicy.cs
```diff
@@ -0,0 +1,113 @@
...
+        {
+            if (selector == null)
+            {
+                throw new ArgumentNullException(nameof(selector));
+            }
+
+            _selector = selector;
+        }
+
+        public override int Order => int.MinValue + 100;
```
javiercn: Suggestion: Could we make all these offsets consts into an internal shared class that everybody links so that we can see the ordering in one place? It might sounds stupid, but it's really hard to think about the ordering of the system when it's distributed across classes. I also understand that this is kind of unrelated and a nit.

src/Mvc/Mvc.Core/src/Routing/DynamicControllerEndpointMatcherPolicy.cs
javiercn: I believe you are right (as usual) in this.
1 would work but I understand why you don't want it.
2 makes sense.
3. I think I get it.

As someone not that familiar with this part of the codebase it would be great to get a more explicit pros/cons list. As it seems you clearly prefer 3. Is there any drawback to it?

src/Mvc/Mvc.Core/src/Routing/DynamicControllerEndpointSelector.cs
```diff
@@ -0,0 +1,72 @@
...
+        // This is async because the page version will be async. We don't want to put ourselves in a
+        // position where these types become public and then users start using them in sync locations.
+        public Task<IReadOnlyList<RouteEndpoint>> SelectEndpointsAsync(RouteValueDictionary values)
+        {
+            if (values == null)
+            {
+                throw new ArgumentNullException(nameof(values));
+            }
+
+            var state = CurrentState;
```
javiercn: nit: table is much better than state IMHO.

src/Mvc/Mvc.Core/src/Routing/DynamicControllerEndpointSelector.cs
```diff
@@ -0,0 +1,72 @@
...
+
+            // The ActionSelectionTable works based on a string[] of the route values in a pre-calculated order. This code extracts
+            // those values in the correct order.
+            var keys = state.RouteKeys;
+            var extractedValues = new string[keys.Length];
+            for (var i = 0; i < keys.Length; i++)
+            {
+                values.TryGetValue(keys[i], out var obj);
```
javiercn: Is it guaranteed that at this point this data is going to be here? I think that deserves a comment

src/Mvc/Mvc.RazorPages/src/Infrastructure/DynamicPageEndpointMatcherPolicy.cs
javiercn: Make it generic based on the selector and the metadata type :trollface:

src/Mvc/samples/MvcSandbox/Startup.cs
javiercn: This is neat.
Also you might have forgotten to register the page related stuff on DI

src/Mvc/Mvc.Core/src/Routing/DynamicControllerEndpointMatcherPolicy.cs
```diff
@@ -0,0 +1,113 @@
...
+                var endpoint = candidates[i].Endpoint;
+
+                var metadata = endpoint.Metadata.GetMetadata<DynamicControllerMetadata>();
+                if (metadata == null)
+                {
+                    continue;
+                }
+
+                var matchedValues = candidates[i].Values;
+                var endpoints = await _selector.SelectEndpointsAsync(metadata.Values);
```
javiercn: Does select throw when multiple endpoints get picked up?

src/Mvc/Mvc.Core/src/Routing/DynamicControllerEndpointMatcherPolicy.cs
```diff
@@ -0,0 +1,113 @@
...
+                if (endpoints.Count == 0)
+                {
+                    // If there's no match this is a configuration error. We can't really check
+                    // during startup that the action you configured exists.
+                    throw new InvalidOperationException(
+                        "Cannot find the fallback endpoint specified by route values: " + 
+                        "{ " + string.Join(", ", metadata.Values.Select(kvp => $"{kvp.Key}: {kvp.Value}")) + " }.");
+                }
+
+                var replacement = endpoints[0];
```
javiercn: This seems sketchy if the above call doesn't throw and we can't guarantee that this always results in a single endpoint being selected. I might not understand all the details of how this works, but it feels like when you had two matching actions.

javiercn (commented):
These changes look great!
Pretty much inline with what I thought it would look like.

I'm assuming that you use order to prevent legit requests to things like `/api/values` to match this if there is already an action for that thing. I think we could benefit from seeing a more 'realistic' example, as in I want to prevent '/api/*' values from matching to prevent a mistake in calling an API to result in a 200. 

How would that look like in Startup?

javiercn:
I did a first pass, let me know when you are ready for review

src/Http/Routing/src/Builder/FallbackEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,90 @@
...
+namespace Microsoft.AspNetCore.Builder
+{
+    /// <summary>
+    /// Contains extension methods for <see cref="IEndpointRouteBuilder"/>.
+    /// </summary>
+    public static class FallbackEndpointRouteBuilderExtensions
+    {
+        /// <summary>
+        /// Adds a specialized <see cref="RouteEndpoint"/> to the <see cref="IEndpointRouteBuilder"/> that will match
+        /// requests for non-filenames with a very low priority.
```
SteveSandersonMS: `a very low priority` sounds a bit imprecise. Since it's `Order = int.MaxValue`, how about `the lowest possible priority`?

javiercn: Lower number > priority. 0 > 1 if that makes sense. It's been how routing has worked since forever

src/Http/Routing/src/Constraints/FileNameRouteConstraint.cs
```diff
@@ -0,0 +1,145 @@
...
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Routing.Constraints
+{
+    /// <summary>
+    /// Constrains a route parameter to represent only filename values.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// This constraint can be used with a low priority route to disambiguate for requests for
```
SteveSandersonMS: The level of priority in the route doesn't really matter, does it? The constraint only deals with what the URL looks like, not whether it's useful to prioritize it above or below other routes.

src/Http/Routing/src/Constraints/FileNameRouteConstraint.cs
```diff
@@ -0,0 +1,145 @@
...
+    /// <para>
+    /// If the route value does not contain a <c>/</c> then the entire value will be interpreted
+    /// as the last segment.
+    /// </para>
+    /// <para>
+    /// The <see cref="FileNameRouteConstraint"/> does not attempt to validate that the value contains
+    /// a legal file name for the current operating system.
+    /// </para>
+    /// <para>
+    /// <list type="bullet">
```
SteveSandersonMS: I'm interested to know how all this content surfaces to users. It gets injected into the auto-generated API documentation web pages, does it?

src/Http/Routing/test/UnitTests/Constraints/FIleNameRouteConstraintTest.cs
```diff
@@ -0,0 +1,99 @@
+// Copyright (c) .NET Foundation. All rights reserved.
```
SteveSandersonMS: Typo in filename: `FIleNameRouteConstraintTest.cs` should be `FileNameRouteConstraintTest.cs`

src/Http/Routing/test/UnitTests/Constraints/FIleNameRouteConstraintTest.cs
SteveSandersonMS: On balance I agree. Leading-dotted things probably are filenames, and are not likely to be desired as part of the client-side URL space. It's a weird scenario, though.

src/Middleware/SpaServices.Extensions/src/SpaDefaultPageMiddleware.cs
SteveSandersonMS: Historically it was E2E tests, but they got removed. So yeah.

src/Middleware/StaticFiles/src/StaticFilesEndpointRouteBuilderExtensions.cs
SteveSandersonMS: I would strongly expect that `MapFallbackToFile` could replace the bulk of `UseSpaStaticFiles`.

The `StaticFileOptions` are needed if you want to set custom response headers (e.g., for caching policy), which people do.

Other complications in `UseSpaStaticFiles` are to do with things like "at dev time I proxy requests to a SPA dev server, but in production is serve prebuilt static files from my contentroot".

src/Mvc/Mvc.Core/src/Infrastructure/ActionSelectionTable.cs
```diff
@@ -0,0 +1,155 @@
...
+using System.Linq;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc.Abstractions;
+using Microsoft.AspNetCore.Routing;
+
+namespace Microsoft.AspNetCore.Mvc.Infrastructure
+{
+    // Common infrastructure for things that look up actions by route values.
+    //
+    // The ActionSelectionTablee stores a mapping of route-values -> items for each known set of
```
SteveSandersonMS:
```suggestion
    // The ActionSelectionTable stores a mapping of route-values -> items for each known set of
```

src/Mvc/Mvc.RazorPages/src/Builder/RazorPagesEndpointRouteBuilderExtensions.cs
```diff
@@ -31,6 +32,245 @@ public static IEndpointConventionBuilder MapRazorPages(this IEndpointRouteBuilde
...
+        /// </para>
+        /// <para>
+        /// <see cref="MapFallbackToPage(IEndpointRouteBuilder, string, string)"/> does not re-execute routing, and will
+        /// not generate route values based on routes defined elsewhere. When using this overload, the route values provided by matching 
+        /// <paramref name="pattern"/> will be available. 
+        /// </para>
+        /// </remarks>
+        public static void MapFallbackToPage(
+            this IEndpointRouteBuilder routes,
+            string pattern,
```
SteveSandersonMS: Is there a bug here? `pattern` seems to be unused (besides the nullcheck).

src/Mvc/Mvc.RazorPages/src/Builder/RazorPagesEndpointRouteBuilderExtensions.cs
```diff
@@ -31,6 +32,245 @@ public static IEndpointConventionBuilder MapRazorPages(this IEndpointRouteBuilde
             return GetOrCreateDataSource(routes);
         }
 
+        /// <summary>
+        /// Adds a specialized <see cref="RouteEndpoint"/> to the <see cref="IEndpointRouteBuilder"/> that will match
+        /// requests for non-filenames with a very low priority. The request will be routed to a page endpoint that
+        /// matches <paramref name="page"/>.
```
SteveSandersonMS: I'm wondering whether it might be better *not* to have the various `MapFallbackTo*` overloads that don't take a `pattern`. Those ones all "match requests for non-filenames", but TBH that feels like a pretty arbitrary specialization. I know it matches our current use cases, but it feels like a separate concern from "mapping a fallback route".

It's not like developers will be writing lots of calls to `MapFallbackTo*`. Chances are, they will have exactly one such call in their whole application, and in most of those cases it was put there for them by the template anyway.

If the template produced more explicit calls, e.g.,

    routes.MapFallbackToPage("Index.cshtml", "{*path:nonfile}");

... then it would be more obvious that the `nonfile` constraint is being applied, plus how to remove it if you don't want it.

SteveSandersonMS (commented):
The new APIs look super - very clean and to the point.

I can't really comment about the internal implementation details because I have no up-to-date background in MVC action selection, endpoint routing policies, and so on. It looks believable though.

src/Http/Routing/src/Builder/FallbackEndpointRouteBuilderExtensions.cs
SteveSandersonMS: All I'm commenting on here is the phrasing used in the XML docs. `a very low priority` is an imprecise statement (just *how* low is the priority?), whereas `the lowest possible priority` tells you something exact.

anurse:
FYI @rynowak, you should rebase on `master` in your next iteration. The NodeServices tests are failing in Helix right now and they are skipped in `master`. It's not critical since there are other failing tests (since this is a Draft PR).

src/Middleware/StaticFiles/src/StaticFilesEndpointRouteBuilderExtensions.cs
rynowak: OK cool. That was what I was hoping for. I'm confident that we can build more flavours of this if needed.

src/Mvc/Mvc.Core/src/Routing/DynamicControllerEndpointMatcherPolicy.cs
rynowak: @JamesNK @pranavkm @davidfowl

src/Mvc/samples/MvcSandbox/Startup.cs
rynowak: yeah, I haven't written tests yet for the MVC cases, and I've only manually tested the controller case.

src/Http/Routing/test/UnitTests/Constraints/FIleNameRouteConstraintTest.cs
rynowak: yeah on reflection, I agree. I'll update this.

src/Mvc/Mvc.Core/src/Routing/DynamicControllerEndpointSelector.cs
rynowak: Fowler hates this so much, but he's wrong.

src/Mvc/Mvc.Core/src/Routing/DynamicControllerEndpointMatcherPolicy.cs
rynowak: It's kinda a mess no matter what. I have a few ideas.

src/Mvc/Mvc.Core/src/Routing/DynamicControllerEndpointSelector.cs
rynowak: meh.

src/Middleware/StaticFiles/src/StaticFilesEndpointRouteBuilderExtensions.cs
rynowak: @davidfowl hates this, but he's wrong. It's the bees knees.

rynowak: btw the technical term is routerware

src/Mvc/Mvc.Core/src/Routing/DynamicControllerEndpointMatcherPolicy.cs
rynowak: No, the documentation says that it's implementation defined. The reason why I made this implementation defined ordering instead of throwing is that we will have multiple matches in a trivial case - pages that have the filename `Index.cshtml`. This produces two endpoints. 

-----

We could change this depending on scenarios and feedback. Here are the concerns involved, I know @JamesNK and @davidfowl will be interested as well.

1. If you want this to throw - it breaks a common use case
2. If you want this to be implementation defined it's really simple to implement and explain, but the capabilities are pretty limited
3. If you want this to **allow** returning multiple endpoints, and have the rest of the system try to **disambiguate** them for you, then it's a very powerful and confusing system.

----

So let's assume that you want to be able to return multiple endpoints (3) and explore it a bit. Let's consider a basic case, like "I want to return two endpoints, and have HTTP method disambiguation select the best match".

Well, The *dynamic endpoint pattern* allows you to **RUN CODE** to replace the dynamic endpoint with the single endpoint of your choosing - **AFTER** the route pattern has been matched.

So I really need to add multiple endpoints but I can't because ....

This is a intentional limitation of `CandidateSet`. It's a data structure that's highly optimized for a small number of entries, and it doesn't allow you to insert items. We could change this, and it would be relatively inefficient everywhere you use it (relative to something like techempower).

Cool, so now I can insert multiple endpoints. The next problem is that ...

The matcher policy that you write to replace an endpoint runs in a pipeline of matcher policies. ie: some policies might run before, some might run after.  Well, if you want the system to disambiguate for you then you really are concerned about which policies run before you and which after. Policies that ran before you will see the dynamic endpoint, policies that run after you will see the replacement.

The next issue is that we have policies that bake themselves into the DFA. We've generally written our policies to use the DFA instead of running in the pipeline of matcher policies (HTTP method, Consumes, Hostname). This is like the difference between compiled (DFA) and interpreted (matcher policy). The DFA policy implementations become part of the core data structure of routing, but they cannot do this for a dynamic endpoint, because the metadata of that endpoint are unknown. We will need to make it possible for all of our policies to run in DFA model or `IEndpointSelectorPolicy` mode.

Now something like this could be possible.

If we're going to end up here - ie we need these features, then I think I would step back and make this a core feature of DFA matcher rather than a thing built on top.

rynowak: I like option 2 (until we know more or discuss more) because while it might not be convenient, it's not blocking either. Someone who wants to extend routing is taking on some *hard problems* - I don't really assume we can solve them for everyone.

src/Mvc/Mvc.Core/src/Routing/DynamicControllerEndpointMatcherPolicy.cs
rynowak: Resolving this comment since I replied to this concern above.

src/Http/Routing/src/Builder/FallbackEndpointRouteBuilderExtensions.cs
rynowak: sure

src/Http/Routing/src/Constraints/FileNameRouteConstraint.cs
rynowak: I described it here to try and capture the main use case in the docs. I think it's valuable to capture how we expect something to used in a meaningful way. Let me know if you have ideas to improve this.

src/Http/Routing/src/Constraints/FileNameRouteConstraint.cs
rynowak: That's the idea at least. I don't have an example handy.

src/Mvc/Mvc.RazorPages/src/Builder/RazorPagesEndpointRouteBuilderExtensions.cs
rynowak: I mostly agree and it would simplify the docs. I tend to prefer these kinds of things being explicit anyway. What I don't want to is have to put these simpler overloads back as the result of a template review.

src/Http/Routing/src/Builder/FallbackEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,90 @@
...
+            {
+                throw new ArgumentNullException(nameof(builder));
+            }
+
+            if (requestDelegate == null)
+            {
+                throw new ArgumentNullException(nameof(requestDelegate));
+            }
+
+            var coventionBuilder = builder.Map(pattern, "Fallback", requestDelegate);
```
JamesNK: covention -> convention

src/Http/Routing/src/Constraints/FileNameRouteConstraint.cs
```diff
@@ -0,0 +1,145 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Globalization;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Routing.Constraints
+{
+    /// <summary>
+    /// Constrains a route parameter to represent only filename values.
```
JamesNK: IMO it worth commenting that this route constraint only checks whether the parameter is formatted like a file name, and doesn't check whether the file exists on disk. That was my first thought when I saw the class name.

src/Http/Routing/src/Constraints/FileNameRouteConstraint.cs
```diff
@@ -0,0 +1,145 @@
...
+            {
+                var value = Convert.ToString(obj, CultureInfo.InvariantCulture);
+                return IsFileName(value);
+            }
+
+            // No value or null value.
+            return false;
+        }
+
+        internal static bool IsFileName(ReadOnlySpan<char> value)
```
JamesNK: Certain characters are invalid in a file name, e.g. `:`, `&`. What about checking for those? To complicate things, invalid file name characters are probably different between operating systems.

Whether this constraint does or doesn't check for them then it would be worth commenting on in the XML docs.

src/Middleware/SpaServices.Extensions/src/StaticFiles/SpaStaticFilesExtensions.cs
```diff
@@ -110,8 +111,8 @@ public static void UseSpaStaticFiles(this IApplicationBuilder applicationBuilder
                 }
             }
 
-            
-            app.UseStaticFiles(staticFileOptions);
+            // We should only serve files when there is no endpoint.
+            app.MapWhen(context => context.GetEndpoint() == null, a => app.UseStaticFiles(staticFileOptions));
```
JamesNK: This is funky, but should it be added as a setting to staticFileOptions? That would be more user friendly/discoverable.

src/Middleware/StaticFiles/src/StaticFilesEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,223 @@
...
+            if (filePath == null)
+            {
+                throw new ArgumentNullException(nameof(filePath));
+            }
+
+            return builder.MapFallback(CreateRequestDelegate(builder, filePath));
+        }
+
+        /// <summary>
+        /// Registers a fallback <see cref="RouteEndpoint"/> that will serve the provided static file
```
JamesNK: nit: fullstop

src/Middleware/StaticFiles/src/StaticFilesEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,223 @@
...
+
+namespace Microsoft.AspNetCore.Builder
+{
+    /// <summary>
+    /// Contains extension methods for using static files with endpoint routing.
+    /// </summary>
+    public static class StaticFilesEndpointRouteBuilderExtensions
+    {
+        /// <summary>
+        /// Registers a fallback <see cref="RouteEndpoint"/> that will serve the provided static file
```
JamesNK: nit: fullstop

src/Middleware/StaticFiles/src/StaticFilesEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,223 @@
...
+        /// </summary>
+        /// <param name="builder">The <see cref="IEndpointRouteBuilder"/>.</param>
+        /// <param name="filePath">The file path of the file to serve.</param>
+        /// <returns>The <see cref="IEndpointRouteBuilder"/></returns>
+        /// <remarks>
+        /// <para>
+        /// <see cref="MapFallbackToFile(IEndpointRouteBuilder, string)"/> is intended to handle cases where URL path of
+        /// the request does not contain a filename, and no other endpoint has matched. This is convenient for routing
```
JamesNK: "does not contain a filename"

Should this be does?

src/Middleware/StaticFiles/src/StaticFilesEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,223 @@
...
+            if (filePath == null)
+            {
+                throw new ArgumentNullException(nameof(filePath));
+            }
+
+            return builder.MapFallback(CreateRequestDelegate(builder, filePath, options));
+        }
+
+        /// <summary>
+        /// Registers a fallback <see cref="RouteEndpoint"/> that will serve the provided static file
```
JamesNK: nit: fullstop. Repeat as needed.

src/Mvc/Mvc.Core/src/Builder/ControllerEndpointRouteBuilderExtensions.cs
```diff
@@ -152,6 +153,296 @@ public static IEndpointConventionBuilder MapDefaultControllerRoute(this IEndpoin
             routes.MapControllerRoute(name, template, defaultsDictionary, constraintsDictionary, dataTokens);
         }
 
+        /// <summary>
+        /// Adds a specialized <see cref="RouteEndpoint"/> to the <see cref="IEndpointRouteBuilder"/> that will match
+        /// requests for non-filenames with a very low priority. The request will be routed to a controller endpoint that
```
JamesNK: nit: filenames is two words in other comments

src/Mvc/Mvc.Core/src/Builder/ControllerEndpointRouteBuilderExtensions.cs
```diff
@@ -152,6 +153,296 @@ public static IEndpointConventionBuilder MapDefaultControllerRoute(this IEndpoin
...
+        /// matches <paramref name="action"/>, and <paramref name="controller"/>.
+        /// </summary>
+        /// <param name="routes">The <see cref="IEndpointRouteBuilder"/> to add the route to.</param>
+        /// <param name="action">The action name.</param>
+        /// <param name="controller">The controller name.</param>
+        /// <remarks>
+        /// <para>
+        /// <see cref="MapFallbackToController(IEndpointRouteBuilder, string, string)"/> is intended to handle cases where URL path of
+        /// the request does not contain a filename, and no other endpoint has matched. This is convenient for routing
```
JamesNK: filename again. Pick one to be consistent with one.

src/Mvc/Mvc.Core/src/Builder/ControllerEndpointRouteBuilderExtensions.cs
```diff
@@ -152,6 +153,296 @@ public static IEndpointConventionBuilder MapDefaultControllerRoute(this IEndpoin
...
+            }
+
+            EnsureControllerServices(routes);
+
+            // Called for side-effect to make sure that the data source is registered.
+            GetOrCreateDataSource(routes);
+
+            // Maps a fallback endpoint with an empty delegate. This is OK because
+            // we don't expect the delegate to run. 
+            routes.MapFallback(context => Task.CompletedTask).Add(b =>
```
JamesNK: nit: Have a shared EmptyRequestDelegate property for `context => Task.CompletedTask` and reuse

src/Mvc/Mvc.Core/src/Infrastructure/ActionSelectionTable.cs
```diff
@@ -0,0 +1,155 @@
...
+            return CreateCore<ActionDescriptor>(
+
+                // We need to store the version so the cache can be invalidated if the actions change.
+                actions.Version,
+
+                // For action selection, ignore attribute routed actions
+                actions.Items.Where(a => a.AttributeRouteInfo == null),
+
+                a => a.RouteValues.Keys,
+                (a, key) =>
```
JamesNK: nit: use named parameters with these args to make it easier to understand what each is for

src/Mvc/Mvc.Core/src/Microsoft.AspNetCore.Mvc.Core.csproj
```diff
@@ -27,6 +27,10 @@ Microsoft.AspNetCore.Mvc.RouteAttribute</Description>
     <Compile Include="$(SharedSourceRoot)SecurityHelper\**\*.cs" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Compile Include="..\..\..\Http\Routing\src\DataSourceDependentCache.cs" Link="Routing\DataSourceDependentCache.cs" />
```
JamesNK: Move the file to a Shared folder?

JamesNK (commented):
I got up to Microsoft.AspNetCore.Mvc.Core.csproj. Will review the rest later

src/Http/Routing/src/Constraints/FileNameRouteConstraint.cs
rynowak: Sure, I agree. I mentioned it for `MapFallback` docs, but not here.

src/Http/Routing/src/Constraints/FileNameRouteConstraint.cs
rynowak: hmm, I thought I did document that. I'll double check. 

It's definitely intended that we don't validate these. 1 - it's platform specific, 2 - it's not aligned with the purpose of this

src/Middleware/SpaServices.Extensions/src/StaticFiles/SpaStaticFilesExtensions.cs
rynowak: Maybe.... this isn't incredibly common to do. All of our guidance for static files is to put it ahead of routing/mvc. This would only come up in a very specialized use case like this.

src/Mvc/Mvc.Core/src/Builder/ControllerEndpointRouteBuilderExtensions.cs
rynowak: I haven't done that because there really isn't a good place for that to live that wants to be public. It's static anyway so this is already *free*

src/Mvc/Mvc.Core/src/Microsoft.AspNetCore.Mvc.Core.csproj
rynowak: *eh* - I don't really free like this helps.

src/Mvc/Mvc.RazorPages/src/Builder/RazorPagesEndpointRouteBuilderExtensions.cs
rynowak: I mocked this up here: https://gist.github.com/rynowak/ab9305b5f1dbb2d918e568b4129a8829

I can't really see us putting this route in a template because it's a lot to explain. I'm going to leave the simpler overloads in place for now, and we can review and decide later.

src/Middleware/StaticFiles/src/StaticFilesEndpointRouteBuilderExtensions.cs
rynowak: nope - it's correct.

This scenario is:
1. Requests for a file get handled by static file middleware
2. Requests for a non-file get re-routed to the SPA (which happens to a static file in this case)
3. Requests for a file that don't exist on disk result in a 404.

src/Mvc/Mvc.Core/src/Routing/DynamicControllerEndpointSelector.cs
rynowak: I'm not sure what you are concerned about. It's `TryGetValue`?

src/Mvc/Mvc.RazorPages/src/Infrastructure/DynamicPageEndpointMatcherPolicy.cs
rynowak: Resolving this comment. I feel like I solved this for the most part by factoring more code into the action selection table class.

src/Mvc/Mvc.Core/src/Microsoft.AspNetCore.Mvc.Core.csproj
JamesNK: I don't like a project in a different solution referencing a file. It will be easy for someone to change the API and only discover another solution is referencing it when the build fails.

rynowak:
OK this is ready for review.

src/Mvc/Mvc.Core/src/Microsoft.AspNetCore.Mvc.Core.csproj
rynowak: I don't really like it either. I left a comment in `DataSourceDependentCache` so that it's more obvious. I don't see how moving the file on disk helps.

src/Mvc/Mvc.Core/src/Routing/DynamicControllerEndpointSelector.cs
javiercn: Yeah, keep in mind that I'm not 100% familiar with the abstractions, so I don't have a full understanding of things

src/Middleware/SpaServices.Extensions/src/StaticFiles/SpaStaticFilesExtensions.cs
Tratcher: Worlds collide. Shouldn't this check go inside of static files so it doesn't hijack requests that already have routes? Or do we need to start work on static file based endpoints?

src/Middleware/StaticFiles/src/StaticFilesEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,219 @@
...
+            app.Use(next => context =>
+            {
+                context.Request.Path = "/" + filePath;
+
+                return next(context);
+            });
+
+            if (options == null)
+            {
+                app.UseStaticFiles();
```
Tratcher: The main things you get from the static file middleware are an IFileProvider, content-type resolution, and conditionals support (if-modified-since...).

The file provider can be assumed (IWebHostEnvironment.WebRootFileProvider), content-types have defaults or the user can specify the content-type if they're hard-coding the path, and it's not clear conditionals would be useful for fallback scenarios.

I don't think you need the full middleware to handle this. Directly calling one of the HttpResponse.[SendFileAsync](https://github.com/aspnet/AspNetCore/blob/master/src/Http/Http.Extensions/src/SendFileResponseExtensions.cs) overloads seems adequate for this scenario. 

E.g.
- `MapFallbackToFile(... filePath)` content-types is resolved from a default set and the file is resolved from IWebHostEnvironment.WebRootFileProvider
- `MapFallbackToFile(... filePath, contentType)` is resolved from IWebHostEnvironment.WebRootFileProvider
- `MapFallbackToFile(... filePath, contentType fileProvider)` is resolved from the given provider
- `MapFallbackToFile(... iFileInfo, contentType)` sends the given file directly

Yet again, if we rewrote static files as endpoints this wouldn't be an issue, you'd new up a pre-resolved static file endpoint.

src/Middleware/StaticFiles/ref/Microsoft.AspNetCore.StaticFiles.csproj
Tratcher: See my comments elsewhere. It's not clear you need to bring in the middleware in order to serve a single file.

Tratcher (changes requested):

src/Http/Routing/src/Constraints/FileNameRouteConstraint.cs
```diff
@@ -0,0 +1,146 @@
...
+            var dotIndex = value.IndexOf('.');
+            if (dotIndex == -1)
+            {
+                // No dot.
+                return false;
+            }
+
+            for (var i = dotIndex + 1; i < value.Length; i++)
+            {
+                if (value[i] != '.')
```
pranavkm: I don't super follow this. Wouldn't this prevent `foo.bar.baz` which the documentation says is allowed?

src/Http/Routing/src/Constraints/NonFileNameRouteConstraint.cs
```diff
@@ -0,0 +1,114 @@
...
+    ///         <description>Final segment contains a <c>.</c> followed by other characters.</description>  
+    ///     </item>
+    ///     <item>  
+    ///         <term><c>.gitignore</c></term>  
+    ///         <description>Final segment contains a <c>.</c> followed by other characters.</description>  
+    ///     </item> 
+    /// </list>
+    /// </para>
+    /// </remarks>
+    public class NonFileNameRouteConstraint : IRouteConstraint
```
pranavkm: Can you add a `//` comment that says this document needs to be kept in sync with `FileNameRouteConstraint`?

src/Http/Routing/src/Constraints/NonFileNameRouteConstraint.cs
```diff
@@ -0,0 +1,114 @@
...
+
+            if (values == null)
+            {
+                throw new ArgumentNullException(nameof(values));
+            }
+
+            if (values.TryGetValue(routeKey, out var obj) && obj != null)
+            {
+                var value = Convert.ToString(obj, CultureInfo.InvariantCulture);
+                return !FileNameRouteConstraint.IsFileName(value);
```
pranavkm: `FileNameHelper.IsFileName` 🚎

src/Http/Routing/test/UnitTests/Constraints/FIleNameRouteConstraintTest.cs
```diff
@@ -0,0 +1,100 @@
...
+                    ".a",
+                    "/.......a"
+                };
+            }
+        }
+
+
+        [Theory]
+        [MemberData(nameof(FileNameData))]
+        public void Match_RouteValue_IsFileName(object value)
```
pranavkm: Could you also add a unit test that specifically tests `FileNameRouteConstraint.IsFileName`? 

```C#
[Theory]
[MemberData(nameof(FileNameData))]
public void FileNameConstraint_IsFileNameReturnsTrue(string fileName)
{
   // Act
  var result = FileNameConstraint.IsFileName(fileName);

   // Assert
   Assert.True(result);
}
```

The method is internal and we don't necessarily need the ceremony of routing to verify that method.

pranavkm: Same deal with the NonFileNameData

src/Middleware/StaticFiles/src/StaticFilesEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,219 @@
...
+            {
+                throw new ArgumentNullException(nameof(filePath));
+            }
+
+            return builder.MapFallback(CreateRequestDelegate(builder, filePath));
+        }
+
+        /// <summary>
+        /// Adds a specialized <see cref="RouteEndpoint"/> to the <see cref="IEndpointRouteBuilder"/> that will match
+        /// requests for non-filenames with the lowest possible priority. The request will be routed a
```
pranavkm: will be routed ~~a~~ to the StaticFileMiddleware

src/Mvc/Mvc.RazorPages/src/Builder/RazorPagesEndpointRouteBuilderExtensions.cs
```diff
@@ -31,6 +33,253 @@ public static IEndpointConventionBuilder MapRazorPages(this IEndpointRouteBuilde
...
+        /// <param name="page">The action name.</param>
+        /// <remarks>
+        /// <para>
+        /// <see cref="MapFallbackToPage(IEndpointRouteBuilder, string, string)"/> is intended to handle cases where URL path of
+        /// the request does not contain a file name, and no other endpoint has matched. This is convenient for routing
+        /// requests for dynamic content to a SPA framework, while also allowing requests for non-existent files to
+        /// result in an HTTP 404.
+        /// </para>
+        /// <para>
+        /// The order of the registered endpoint will be <c>int.MaxValue</c>.
```
pranavkm: Where does this get set? I didn't find any references to the order being set in the code.

src/Mvc/Mvc.RazorPages/src/Infrastructure/DynamicPageEndpointMatcherPolicy.cs
```diff
@@ -0,0 +1,126 @@
...
+                    continue;
+                }
+
+                var matchedValues = candidates[i].Values;
+                var endpoints = await _selector.SelectEndpointsAsync(metadata.Values);
+                if (endpoints.Count == 0)
+                {
+                    // If there's no match this is a configuration error. We can't really check
+                    // during startup that the action you configured exists.
+                    throw new InvalidOperationException(
```
pranavkm: Use the resx luke

src/Mvc/Mvc.RazorPages/src/Infrastructure/DynamicPageEndpointMatcherPolicy.cs
```diff
@@ -0,0 +1,126 @@
...
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing;
+using Microsoft.AspNetCore.Routing.Matching;
+
+namespace Microsoft.AspNetCore.Mvc.RazorPages.Infrastructure
+{
+    internal class DynamicPageEndpointMatcherPolicy : MatcherPolicy, IEndpointSelectorPolicy
+    {
+        private readonly DynamicPageEndpointSelector _selector;
+        private readonly IPageLoader _loader;
```
pranavkm: This will complain once I get my changes in to make it Obsolete

src/Mvc/Mvc.RazorPages/src/Infrastructure/DynamicPageEndpointSelector.cs
```diff
@@ -0,0 +1,56 @@
...
+                throw new ArgumentNullException(nameof(dataSource));
+            }
+
+            _dataSource = dataSource;
+            _cache = new DataSourceDependentCache<ActionSelectionTable<RouteEndpoint>>(dataSource, Initialize);
+        }
+
+        private ActionSelectionTable<RouteEndpoint> Table => _cache.EnsureInitialized();
+
+        // This is async because the page version will be async. We don't want to put ourselves in a
```
pranavkm: > This is async because the page version will be async.

This is the page version. Neither the controller selector nor the page do async work. Should they be synchronous?

src/Mvc/Mvc.Core/src/Infrastructure/StringArrayComparer.cs
```diff
@@ -0,0 +1,79 @@
...
+        }
+
+        public int GetHashCode(string[] obj)
+        {
+            if (obj == null)
+            {
+                return 0;
+            }
+
+            var hash = new HashCodeCombiner();
```
pranavkm: Can you use `System.HashCode` instead?

src/Mvc/Mvc.Core/src/Routing/DynamicControllerEndpointMatcherPolicy.cs
pranavkm: >  I have another work item planned to implement features that people will use to port their IRouter code forward. 

What sort of routing scenario would this help in?

src/Http/Routing/src/DataSourceDependentCache.cs
```diff
@@ -8,6 +8,8 @@
 
 namespace Microsoft.AspNetCore.Routing
 {
+    // FYI: This class is also linked into MVC. If you make changes to the API you will
```
pranavkm: You kinda don't need this comment per se. The build would fail if you #yolo change this.

src/Http/Routing/src/Constraints/FileNameRouteConstraint.cs
Tratcher: It only needs one non-period character after the first period.

src/Middleware/SpaServices.Extensions/src/StaticFiles/SpaStaticFilesExtensions.cs
rynowak: I'm happy to send another PR that puts this inside of static files if you think that's the right thing to do.

src/Http/Routing/src/Constraints/FileNameRouteConstraint.cs
pranavkm: Hmm, so if I understand correctly, it's trying to disallow files that look like `.....`, while allowing files that look like `..a.b...c.`? This needs some comments.

src/Middleware/StaticFiles/src/StaticFilesEndpointRouteBuilderExtensions.cs
rynowak: @SteveSandersonMS can elaborate more than I can on the scenarios. The existing SPA stuff and the people that use it use quite a lot of those features. 

In particular caching headers, and multiple file providers have come up in our discussions. It would be really nice if as part of this experience users could configure these things in one place and have them work for both static files and this. 

-----

I'm still open to the idea that this would be something different from the static files middleware, but what's the drawback of using the middleware?

src/Middleware/SpaServices.Extensions/src/StaticFiles/SpaStaticFilesExtensions.cs
rynowak: > Or do we need to start work on static file based endpoints so you can remove this middleware entirely?

I thought this was that? What part do you think we should make more first class?

src/Http/Routing/src/Constraints/FileNameRouteConstraint.cs
rynowak: My goal is to allow `foo.bar` and `.foobar` but not nonesense like `..`.

src/Http/Routing/src/Constraints/NonFileNameRouteConstraint.cs
rynowak: nope.

src/Http/Routing/test/UnitTests/Constraints/FIleNameRouteConstraintTest.cs
rynowak: I don't really get why testing the same thing twice will help.

src/Mvc/Mvc.RazorPages/src/Builder/RazorPagesEndpointRouteBuilderExtensions.cs
rynowak: `MapFallback` sets the order

src/Mvc/Mvc.RazorPages/src/Infrastructure/DynamicPageEndpointMatcherPolicy.cs
rynowak: To what end? we have no plans to add localization. It's just homework.

src/Mvc/Mvc.RazorPages/src/Infrastructure/DynamicPageEndpointMatcherPolicy.cs
rynowak: Yes.

src/Mvc/Mvc.RazorPages/src/Infrastructure/DynamicPageEndpointSelector.cs
rynowak: I'll remove the comment - also this will be async once I use your new thing.

 I **insist** that these be asynchronous because I anticipate that they will be public. I do not want them to be made synchronous and then proliferate usage all over the world.

src/Http/Routing/src/DataSourceDependentCache.cs
rynowak: meh.

src/Middleware/SpaServices.Extensions/src/StaticFiles/SpaStaticFilesExtensions.cs
Tratcher: Yes, please update StaticFiles.

src/Mvc/Mvc.RazorPages/src/Builder/RazorPagesEndpointRouteBuilderExtensions.cs
SteveSandersonMS: Alright, I agree.

src/Middleware/StaticFiles/src/StaticFilesEndpointRouteBuilderExtensions.cs
SteveSandersonMS: The strongest reason we had to use `UseStaticFiles` in SPA-land was to get all the ETag stuff for free. You really have to support this, and `HttpResponse.SendFileAsync` won't give it to you.

I agree it would be nice if the relevant parts of `UseStaticFiles` were factored out, but I also feel (like @rynowak) that there's no obvious drawback in just reusing `UseStaticFiles` here as-is.

src/Mvc/Mvc.Core/src/Routing/DynamicControllerEndpointMatcherPolicy.cs
rynowak: This is the typical CMS thing. Write an `IRouter` that treats the URL as a *slug* and then look that up in the database and put in the route values that go to the desired action.

src/Middleware/StaticFiles/ref/Microsoft.AspNetCore.StaticFiles.csproj
rynowak: Discussed offline. We're going to go with this for now and evaluate doing some deeper integration between these components.

src/Middleware/StaticFiles/src/StaticFilesEndpointRouteBuilderExtensions.cs
rynowak: Discussed offline and we're going to keep this as-is for now. We're meeting after preview 4 to design another approach for static files in general.

src/Http/Routing/src/Constraints/NonFileNameRouteConstraint.cs
rynowak: Added to the IsFileName method

src/Middleware/SpaServices.Extensions/src/StaticFiles/SpaStaticFilesExtensions.cs
rynowak: Will be in the next iteration.

src/Mvc/Mvc.RazorPages/src/Infrastructure/DynamicPageEndpointSelector.cs
rynowak: GAH. You are the correct. These are no longer async.

rynowak:
Alright this is updated and I've addressed all of the feedback. @Tratcher - there is a new commit with your name on it.

src/Middleware/StaticFiles/src/LoggerExtensions.cs
```diff
@@ -75,6 +76,10 @@ static LoggerExtensions()
                 logLevel: LogLevel.Debug,
                 eventId: new EventId(14, "WriteCancelled"),
                 formatString: "The file transmission was cancelled");
+            _endpointMatched = LoggerMessage.Define(
+                logLevel: LogLevel.Debug,
+                eventId: new EventId(15, "EndpointMatched"),
+                formatString: "The request matched an endpoint.");
```
Tratcher: "Static files was skipped as the request already matched an endpoint."

src/Middleware/StaticFiles/src/StaticFileMiddleware.cs
```diff
@@ -76,6 +76,10 @@ public async Task Invoke(HttpContext context)
             {
                 _logger.RequestMethodNotSupported(context.Request.Method);
             }
+            else if (!fileContext.ValidateNoEndpoint())
```
Tratcher: Move first.

Tratcher (approved):

rynowak:
@aspnet/brt - Come on Bort! It's time to leave. Sorry, are you talking to me? No, my son is also named Bort.

anurse:
Mop and Bucket man to the exit of the Nauseator!

This is https://github.com/aspnet/AspNetCore/issues/6597 which is **supposedly** to be fixed "soon" (hence I wasn't as aggressive about skipping). Seems @halter73 and @BrennanConroy are debating the right way to fix it though so I'll add a skippy-skip.

rynowak:
> which is supposedly to be fixed "soon" (hence I wasn't as aggressive about skipping)

The only good test is a skipped test!

rynowak:
@aspnet/brt @HaoK - can you take a look at this #helixfail

PR #8053, opened 2019-03-01
Skip react for now

ryanbrandenburg:
We're skipping this test for now to avoid flakiness. Re-instating it once it has been improved is being tracked in https://github.com/aspnet/AspNetCore-Internal/issues/1745#issuecomment-468486109.


ajaybhargavb (approved):

pakrym:
ReactReduxTemplate_Works_NetCore needs to be skipped too.

PR #8093, opened 2019-03-01
Add runtime project back to sharedfx project list

pakrym:
Fixes: https://github.com/aspnet/AspNetCore/issues/8083

### DIFF
--- a/build/SharedFx.targets
+++ b/build/SharedFx.targets
@@ -10,6 +10,7 @@
 
     <FxProjectToBuild Condition=" '$(BuildSiteExtensions)' == 'true' " Include="$(RepositoryRoot)src\SiteExtensions\Runtime\Microsoft.AspNetCore.Runtime.SiteExtension.pkgproj" />
     <FxProjectToBuild Condition=" '$(BuildSiteExtensions)' == 'true' " Include="$(RepositoryRoot)src\SiteExtensions\LoggingBranch\LoggingBranch.csproj" />
+    <FxProjectToBuild Include="$(RepositoryRoot)src\Framework\src\Microsoft.AspNetCore.App.Runtime.csproj" />
   </ItemGroup>
 
   <Target Name="BuildSharedFx" DependsOnTargets="$(BuildSharedFxDependsOn)" />
--- a/build/SharedFx.targets
+++ b/build/SharedFx.targets
@@ -10,7 +10,8 @@
 
     <FxProjectToBuild Condition=" '$(BuildSiteExtensions)' == 'true' " Include="$(RepositoryRoot)src\SiteExtensions\Runtime\Microsoft.AspNetCore.Runtime.SiteExtension.pkgproj" />
     <FxProjectToBuild Condition=" '$(BuildSiteExtensions)' == 'true' " Include="$(RepositoryRoot)src\SiteExtensions\LoggingBranch\LoggingBranch.csproj" />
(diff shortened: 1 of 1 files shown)


build/SharedFx.targets
```diff
@@ -10,6 +10,7 @@
 
     <FxProjectToBuild Condition=" '$(BuildSiteExtensions)' == 'true' " Include="$(RepositoryRoot)src\SiteExtensions\Runtime\Microsoft.AspNetCore.Runtime.SiteExtension.pkgproj" />
     <FxProjectToBuild Condition=" '$(BuildSiteExtensions)' == 'true' " Include="$(RepositoryRoot)src\SiteExtensions\LoggingBranch\LoggingBranch.csproj" />
+    <FxProjectToBuild Include="$(RepositoryRoot)src\Framework\src\Microsoft.AspNetCore.App.Runtime.csproj" />
```
dougbu: Is this reverting part of e7b00a5508 ? Specifically, @natemcmaster's removal of
``` xml
<FxProjectToBuild Condition=" '$(BuildRuntimeArchive)' != 'false' " Include="$(RepositoryRoot)src\Installers\Archive\*.*proj" />
```

Either way, I'd appreciate more context on this.

pakrym: Yes, it reverts that part. There is more context in the original issue. 
Microsoft.AspNetCore.App.Runtime.win-x64 is packaged during the main build but we don't run the full build for x86 and that's why 
Microsoft.AspNetCore.App.Runtime.win-x86 is not being produced. I attempted to add 
Microsoft.AspNetCore.App.Runtime back to shared framework project list so it gets built in win-x86 build.

dougbu (approved):

dougbu:
Except it doesn't look like the change was successful ☹️

dougbu (approved):

build/SharedFx.targets
```diff
@@ -10,6 +10,8 @@
 
     <FxProjectToBuild Condition=" '$(BuildSiteExtensions)' == 'true' " Include="$(RepositoryRoot)src\SiteExtensions\Runtime\Microsoft.AspNetCore.Runtime.SiteExtension.pkgproj" />
     <FxProjectToBuild Condition=" '$(BuildSiteExtensions)' == 'true' " Include="$(RepositoryRoot)src\SiteExtensions\LoggingBranch\LoggingBranch.csproj" />
+
+    <FxProjectToBuild Condition=" '$(TargetRuntimeIdentifier)' == 'win-x86' " Include="$(RepositoryRoot)src\Framework\src\Microsoft.AspNetCore.App.Runtime.csproj" />
```
natemcmaster: This might cause a double-build. This project is already included here:
https://github.com/aspnet/AspNetCore/blob/337370263a7119ee5b8e0a6649d9e2da2fd9f0c3/build/repo.props#L149.  As long as this fixes the problem without double-building the project, this is fine with me as a short-term workaround.

The reason we are missing Microsoft.AspNetCore.App.Runtime.win-x86 is that #7832 moved the project from FxProjectToBuild to ProjectToBuild, which means `/t:BuildSharedFx` is not running the `Pack` target to run on Microsoft.AspNetCore.App.Runtime.csproj.

I would actually like to completely eliminate the `/t:BuildSharedFx` target, but we can leave that for another day.

pakrym: > I would actually like to completely eliminate the `/t:BuildSharedFx` target, but we can leave that for another day.

I completely agree. 

> This might cause a double-build. This project is already included here:

The condition should avoid the double build.

PR #8095, opened 2019-03-01
Avoid race in FileWatcher test

Tratcher:
https://github.com/aspnet/AspNetCore-Internal/issues/1710#issuecomment-468578004

Summary: We don't know why this started but we do see a possible race in the test we can address.

What we know:
- This started failing Feb 11th.
- There's at least one failure in the 2.2 branch, but most are master.
- It's only failing on aspnet-mac-a03
- It's not affecting PRs as it's an internal agent.
- It's not OS or version specific, it has passed on other similar agents
- There was a [change](https://github.com/aspnet/AspNetCore/commit/7d4b6fccffc5ab4bb75a3dfaa20088a022709a0c#diff-06e0fd743053558ea338c479baee65c9R273) merged on Jan 28th  to make these tests async. It seems unrelated as it didn't fail for more than a week, even on the same agent.
- This agent was recreated on Jan 17th which also seems unrelated.
- The only change to this component in the affected time window was https://github.com/aspnet/AspNetCore/pull/7403/files which primarily changes logging.
- The only other possibly related change the day this started failing was an Extensions dependency update: https://github.com/aspnet/AspNetCore/pull/7397/files

### DIFF
--- a/src/Tools/dotnet-watch/test/FileWatcherTests.cs
+++ b/src/Tools/dotnet-watch/test/FileWatcherTests.cs
@@ -246,7 +246,13 @@ namespace Microsoft.DotNet.Watcher.Tools.FunctionalTests
                 File.WriteAllText(Path.Combine(dir, "foo2"), string.Empty);
                 File.WriteAllText(Path.Combine(dir, "foo3"), string.Empty);
                 File.WriteAllText(Path.Combine(dir, "foo4"), string.Empty);
-                File.WriteAllText(Path.Combine(dir, "foo4"), string.Empty);
+
+                // On Unix the native file watcher may surface events from
+                // the recent past. Delay to avoid those.
+                // On Unix the file write time is in 1s increments;
+                // if we don't wait, there's a chance that the polling
+                // watcher will not detect the change
+                await Task.Delay(1250);
 
                 var testFileFullPath = Path.Combine(dir, "foo3");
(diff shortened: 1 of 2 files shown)


BrennanConroy (approved):
Did we want to backport this, or does it not fail on 2.2 often enough?

src/Tools/dotnet-watch/test/FileWatcherTests.cs
```diff
@@ -246,7 +246,13 @@ public async Task MultipleFiles(bool usePolling)
...
                 File.WriteAllText(Path.Combine(dir, "foo3"), string.Empty);
                 File.WriteAllText(Path.Combine(dir, "foo4"), string.Empty);
-                File.WriteAllText(Path.Combine(dir, "foo4"), string.Empty);
+
+                // On Unix the native file watcher may surface events from
+                // the recent past. Delay to avoid those.
+                // On Unix the file write time is in 1s increments;
+                // if we don't wait, there's a chance that the polling
+                // watcher will not detect the change
+                await Task.Delay(1000);
```
Eilon: Do we want to make this a tiny bit larger so that it's not "exactly" 1 second?

Tratcher:
@BrennanConroy lets see if this fixes master first. I only saw one failure in 2.2.

PR #8104, opened 2019-03-01
Publish helix tests in release

pakrym:
Fixes: https://github.com/aspnet/AspNetCore/issues/8103

### DIFF
--- a/eng/helix/helix.proj
+++ b/eng/helix/helix.proj
@@ -21,6 +21,7 @@
     <EnableAzurePipelinesReporter>false</EnableAzurePipelinesReporter>
     <IsExternal>true</IsExternal>
     <Creator>aspnetcore</Creator>
+    <SkipInvalidConfigurations>true</SkipInvalidConfigurations>
   </PropertyGroup>
 
   <Target Name="Gather" BeforeTargets="Build">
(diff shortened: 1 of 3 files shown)


dougbu (commented):
This certainly looks fine from an efficiency perspective -- release bits tend to run faster. But, `Debug.Assert(...)`s should be consistent. What is the root cause of flakiness related to our assertions?

pakrym:
We can investigate the root cause later when we decide that testing debug builds is a good idea. For now, the goal is to make tests less flaky.

dougbu (approved):
:ok: I'm convinced 😺

HaoK (approved):

eng/helix/helix.proj
```diff
@@ -21,6 +21,7 @@
     <EnableAzurePipelinesReporter>false</EnableAzurePipelinesReporter>
     <IsExternal>true</IsExternal>
     <Creator>aspnetcore</Creator>
+    <SkipInvalidConfigurations>true</SkipInvalidConfigurations>
```
HaoK: We should circle back with the helix sdk folks if we have to turn this back on, its this issue, we can reopen it to track if need https://github.com/dotnet/arcade/issues/1605

PR #8106, opened 2019-03-01
Helix fail if exceptions thrown during test discovery

HaoK:
 - Detail 1
 - Detail 2


HaoK:
Oops meant to do draft

Issue #8109, opened 2019-03-01
Kestrel Libuv.FunctionalTests RemoteIPv6Address flaky in Helix


anurse:
This test is flaky in helix on `OSX.1012.Amd64.Open`. See https://mc.dot.net/#/user/aspnetcore/pr~2Faspnet~2Faspnetcore/ci/20190227.45/workItem/Libuv.FunctionalTests-netcoreapp3.0/wilogs

* [ ] Skipped

anurse:
Can we disable a single test on a specific Helix queue? cc @muratg @Eilon @jkotalik

muratg:
cc @HaoK

HaoK:
We don’t have a way to target a specific test for a specific queue yet. I would just skip the whole test on helix for now. I already filed an issue for adding a skip on OS and helix attribute

anurse:
Fixed in #8111

anurse:
Reopening actually, I just skipped it. I'd still like to track fixing the flakiness.

PR #8111, opened 2019-03-01
skip Libuv.FunctionalTests RemoteIPv6Address on Helix

analogrelay:
related #8109

### DIFF
--- a/src/Servers/Kestrel/test/FunctionalTests/RequestTests.cs
+++ b/src/Servers/Kestrel/test/FunctionalTests/RequestTests.cs
@@ -136,6 +136,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests
 
         [ConditionalFact]
         [IPv6SupportedCondition]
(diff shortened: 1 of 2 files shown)


Eilon (approved):

HaoK (approved):
Hopefully this is the main offender, I have a vague recollection of seeing similar tests in here fail before, but I might have skipped them already, cross our fingers I guess

anurse:
If another one shows up, I'll disable the entire project :).

anurse:
I messed something up. I thought it was building on my machine but I was wrong :).

anurse:
Figured it out, the file is also used in Sockets tests. I'm adding some `#ifdef` fun.

HaoK (approved):
Yeah these tests are all shared, so slightly more annoying to disable

HaoK:
Generally I've only seen the libuv bind/functional tests fail (and usually only on non windows, so the flakiness is limited somewhat)

PR #8114, opened 2019-03-01
Fix race in LongPolling causing flaky tests

BrennanConroy:
Fixes https://github.com/aspnet/AspNetCore/issues/6597

### DIFF
--- a/src/SignalR/common/Http.Connections/src/Internal/HttpConnectionContext.cs
+++ b/src/SignalR/common/Http.Connections/src/Internal/HttpConnectionContext.cs
@@ -28,6 +28,7 @@ namespace Microsoft.AspNetCore.Http.Connections.Internal
                                          IHttpTransportFeature,
                                          IConnectionInherentKeepAliveFeature
     {
+        private readonly object _stateLock = new object();
         private readonly object _itemsLock = new object();
         private readonly object _heartbeatLock = new object();
         private List<(Action<object> handler, object state)> _heartbeatHandlers;
@@ -35,7 +36,6 @@ namespace Microsoft.AspNetCore.Http.Connections.Internal
         private PipeWriterStream _applicationStream;
         private IDuplexPipe _application;
         private IDictionary<object, object> _items;
-        private int _status = (int)HttpConnectionStatus.Inactive;
 
         // This tcs exists so that multiple calls to DisposeAsync all wait asynchronously
         // on the same task
@@ -83,7 +83,6 @@ namespace Microsoft.AspNetCore.Http.Connections.Internal
         public HttpTransportType TransportType { get; set; }
 
         public SemaphoreSlim WriteLock { get; } = new SemaphoreSlim(1, 1);
-        public SemaphoreSlim StateLock { get; } = new SemaphoreSlim(1, 1);
 
         // Used for testing only
         internal Task DisposeAndRemoveTask { get; set; }
@@ -96,7 +95,18 @@ namespace Microsoft.AspNetCore.Http.Connections.Internal
 
         public DateTime LastSeenUtc { get; set; }
 
-        public HttpConnectionStatus Status { get => (HttpConnectionStatus)_status; set => Interlocked.Exchange(ref _status, (int)value); }
+        public DateTime? LastSeenUtcIfInactive
+        {
+            get
+            {
+                lock (_stateLock)
+                {
+                    return Status == HttpConnectionStatus.Inactive ? (DateTime?)LastSeenUtc : null;
+                }
+            }
+        }
+
+        public HttpConnectionStatus Status { get; set; } = HttpConnectionStatus.Inactive;
 
         public override string ConnectionId { get; set; }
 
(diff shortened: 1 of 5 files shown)


src/SignalR/common/Http.Connections/src/Internal/HttpConnectionContext.cs
```diff
@@ -95,7 +98,7 @@ public HttpConnectionContext(string id, IDuplexPipe transport, IDuplexPipe appli
 
         public DateTime LastSeenUtc { get; set; }
 
-        public HttpConnectionStatus Status { get; set; } = HttpConnectionStatus.Inactive;
```
halter73: Where else is this setter called? It seems weird to protect state transitions in one place, but not elsewhere.

src/SignalR/common/Http.Connections/src/Internal/HttpConnectionDispatcher.cs
```diff
@@ -331,7 +331,8 @@ private async Task ExecuteAsync(HttpContext context, ConnectionDelegate connecti
                         // Mark the connection as inactive
                         connection.LastSeenUtc = DateTime.UtcNow;
 
-                        connection.Status = HttpConnectionStatus.Inactive;
+                        // This is done outside a lock because the next poll might be waiting in the lock already and waiting for currentRequestTcs to complete
+                        Interlocked.CompareExchange(ref connection._status, (int)HttpConnectionStatus.Inactive, (int)HttpConnectionStatus.Active);
```
halter73: So this only prevents moving from the Disposed to Inactive state, right? Is this by itself enough to fixe the "InvalidOperationException: Reading is not allowed after reader was completed."? Are we sure there's no other way to transition out of the Disposed state?

I think a comment discussing what state transition we're trying to prevent here would be nice.

src/SignalR/common/Http.Connections/src/Internal/HttpConnectionContext.cs
```diff
@@ -68,6 +69,8 @@ public HttpConnectionContext(string id, ILogger logger)
             Features.Set<IHttpContextFeature>(this);
             Features.Set<IHttpTransportFeature>(this);
             Features.Set<IConnectionInherentKeepAliveFeature>(this);
+
+            _status = (int)HttpConnectionStatus.Inactive;
```
halter73: Nit: you can keep the assignment could be inline with the declaration like before

src/SignalR/common/Http.Connections/src/Internal/HttpConnectionContext.cs
halter73: Can we remove the setter now?

src/SignalR/common/Http.Connections/src/Internal/HttpConnectionContext.cs
```diff
@@ -309,6 +310,11 @@ private async Task WaitOnTasks(Task applicationTask, Task transportTask, bool cl
             }
         }
 
+        public bool ChangeState(HttpConnectionStatus from, HttpConnectionStatus to)
```
halter73: TryChangeState

src/SignalR/common/Http.Connections/src/Internal/HttpConnectionDispatcher.cs
```diff
@@ -371,7 +373,7 @@ private async Task ExecuteAsync(HttpContext context, ConnectionDelegate connecti
                 }
 
                 // Mark the connection as active
-                connection.Status = HttpConnectionStatus.Active;
+                connection.ChangeState(HttpConnectionStatus.Inactive, HttpConnectionStatus.Active);
```
halter73: Should we just continue on if the state transition fails? Should we assert the transitions are successful?

src/SignalR/common/Http.Connections/src/Internal/HttpConnectionContext.cs
BrennanConroy: No, it's used in `HttpConnectionContext.DisposeAsync` and it was a public API already, so we can't remove it.

src/SignalR/common/Http.Connections/src/Internal/HttpConnectionDispatcher.cs
BrennanConroy: I think it's fine to let it continue

halter73: What if you're in the HttpConnectionStatus.Disposed state?

BrennanConroy: It can't be because they both use the same lock and there is a check for Disposed state after acquiring the lock in this case.

halter73:
@BrennanConroy I added a commit to this PR to remove the SemaphoreSlim/Interlocked stuff. Tell me if it looks good to you

BrennanConroy:
Your change had a lot of issues including making the bug this PR is fixing happen more frequently.

I'm reverting it, and if you'd like to refactor Long Polling you can do so at a different time after this fix goes in.

halter73:
It's *your* PR. You can just force push my commit away 😉

davidfowl (approved):

PR #8115, opened 2019-03-01
Try updating SDK

pakrym:
cc @sebastienros

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "3.0.100-preview-010184"
+    "version": "3.0.100-preview4-010309"
   },
   "msbuild-sdks": {
     "Yarn.MSBuild": "1.13.0"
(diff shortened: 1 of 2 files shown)


pakrym:
Blocked on https://github.com/dotnet/sdk/pull/3002

sebastienros:
Do you mean you will take a newer sdk instead? But in that case you would just close this PR, so I don't get why it's blocked.

pakrym:
Ok, I'll close it after it fails.

pakrym:
Hm, it looks like it will pass. I guess we are not affected by the mentioned issue.

sebastienros:
Somehow expected, this SDK is much older (2 weeks), the latest is around `3.0.100-preview4-010553` and the issue only appeared yesterday. This is the one I was using on benchmarks.

I think we should still take it, and then the new one when it's ready, because there might be other issues with it.

sebastienros (approved):

PR #8125, opened 2019-03-02
IBufferWriter<T>.Advance(int) invalidates span

AArnott:
Per https://github.com/dotnet/dotnet-api-docs/pull/1949/files#diff-7c9a27ba441e8b9f136c5e7499b6cae3R47, any call to `IBufferWriter<T>.Advance(int)` should be assumed to have invalidated any prior result of `GetSpan(int)`, so we need to reset our Span so it's reacquired on the next write.

### DIFF
--- a/src/Servers/Kestrel/Core/src/Internal/BufferWriter.cs
+++ b/src/Servers/Kestrel/Core/src/Internal/BufferWriter.cs
@@ -69,6 +69,9 @@ namespace System.Buffers
                 _bytesCommitted += buffered;
                 _buffered = 0;
                 _output.Advance(buffered);
+                
+                // Calling IBufferWriter<T>.Advance(int) may have invalidated our cached Span<T>, so reacquire next time.
+                _span = default;
             }
         }
 
--- a/src/Servers/Kestrel/perf/PlatformBenchmarks/BufferWriter.cs
+++ b/src/Servers/Kestrel/perf/PlatformBenchmarks/BufferWriter.cs
@@ -30,6 +30,9 @@ namespace PlatformBenchmarks
             {
                 _buffered = 0;
                 _output.Advance(buffered);
+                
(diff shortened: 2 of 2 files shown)


AArnott:
CC: @davidfowl

benaadams:
If you like that, then you'll love this https://github.com/TechEmpower/FrameworkBenchmarks/blob/14e19d782cc5d7f8164490f3467b3981a2a8e27d/frameworks/CSharp/aspnetcore/PlatformBenchmarks/BenchmarkApplication.Fortunes.cs#L38-L56

src/Servers/Kestrel/Core/src/Internal/BufferWriter.cs
```diff
@@ -69,6 +69,9 @@ public void Commit()
                 _bytesCommitted += buffered;
                 _buffered = 0;
                 _output.Advance(buffered);
+                
+                // Calling IBufferWriter<T>.Advance(int) may have invalidated our cached Span<T>, so reacquire next time.
+                _span = default;
```
benaadams: This will regress `EnsureMore` which immediately sets the span after `Commit`, so should be factored differently

AArnott:
> If you like that, then you'll love this 

Why? I fail to see the relevance.

benaadams:
> > If you like that, then you'll love this

> Why? I fail to see the relevance.

A copy is taken of the `BufferWriter` to `lengthWriter` to later add a length prefix; then lots of Writing occurs which is likely at some point to hit `WriteMultiBuffer` => `EnsureMore` => `Commit` and then the lengthWriter is use to write the final length to a span prior to all the data; after which it does a final `Commit` before moving onto the `FlushAsync` of the `PipeWriter` in the outer caller

anurse:
This got super stale :(. I'm going to close it, since there is some outstanding feedback and it probably needs to be rebased. Feel free to reopen if you'd like to re-submit.

PR #8135, opened 2019-03-03
Add option to restrict the maximum hub message size

davidfowl:
- This change moves the limit checking from the transport layer to the protocol parsing layer. One nice side effect is that it gives us better control over error handling.

Fixes https://github.com/aspnet/AspNetCore/issues/8121 and https://github.com/aspnet/AspNetCore/issues/5307

Notes:
- The transport limits just slow down the client, but it doesn't prevent extra buffering. See the issue for more details.

### DIFF
--- a/src/SignalR/server/Core/src/HubOptions.cs
+++ b/src/SignalR/server/Core/src/HubOptions.cs
@@ -37,9 +37,9 @@ namespace Microsoft.AspNetCore.SignalR
         public IList<string> SupportedProtocols { get; set; } = null;
 
         /// <summary>
-        /// Gets or sets the maximum message size of a single hub message.
+        /// Gets or sets the maximum message size of a single incoming hub message. The default is 32KB.
         /// </summary>
-        public long? MaximumMessageSize { get; set; } = null;
+        public long? MaximumReceiveMessageSize { get; set; } = null;
 
         /// <summary>
         /// Gets or sets a value indicating whether detailed error messages are sent to the client.
--- a/src/SignalR/server/Core/src/HubOptions.cs
+++ b/src/SignalR/server/Core/src/HubOptions.cs
@@ -37,6 +37,11 @@ namespace Microsoft.AspNetCore.SignalR
         public IList<string> SupportedProtocols { get; set; } = null;
 
         /// <summary>
+        /// Gets or sets the maximum message size of a single hub message.
+        /// </summary>
+        public long? MaximumMessageSize { get; set; } = null;
+
+        /// <summary>
         /// Gets or sets a value indicating whether detailed error messages are sent to the client.
         /// Detailed error messages include details from exceptions thrown on the server.
         /// </summary>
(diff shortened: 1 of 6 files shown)


src/SignalR/server/SignalR/test/HubConnectionHandlerTestUtils/Utils.cs
```diff
@@ -72,7 +72,7 @@ public static IServiceProvider CreateServiceProvider(Action<ServiceCollection> a
             return services.BuildServiceProvider();
         }
 
-        public static Connections.ConnectionHandler GetHubConnectionHandler(Type hubType, Action<ServiceCollection> addServices = null, ILoggerFactory loggerFactory = null)
+        public static Connections.ConnectionHandler GetHubConnectionHandler(Type hubType, ILoggerFactory loggerFactory = null, Action<ServiceCollection> addServices = null)
```
davidfowl: Lambdas should be last.

src/SignalR/server/Core/src/HubOptions.cs
```diff
@@ -36,6 +36,11 @@ public class HubOptions
         /// </summary>
         public IList<string> SupportedProtocols { get; set; } = null;
 
+        /// <summary>
+        /// Gets or sets the maximum message size of a single hub message.
+        /// </summary>
+        public long? MaximumMessageSize { get; set; } = null;
```
JamesNK: This is just for receive, right? Should the name be MaximumReceiveMessageSize?

src/SignalR/server/Core/src/Internal/HubOptionsSetup.cs
```diff
@@ -16,6 +16,8 @@ public class HubOptionsSetup : IConfigureOptions<HubOptions>
 
         internal static TimeSpan DefaultClientTimeoutInterval => TimeSpan.FromSeconds(30);
 
+        internal const int DefaultMaximumMessageSize = 32 * 1024 * 1024;
```
JamesNK: Include the default size in the HubOptions.MaximumMessageSize XML doc?

src/SignalR/server/SignalR/test/HubConnectionHandlerTests.cs
```diff
@@ -452,6 +453,191 @@ public async Task HandshakeSuccessSendsResponseWithoutError()
...
+                    await connectionHandlerTask.OrTimeout();
+                }
+            }
+
+            Assert.NotNull(exception);
+            Assert.Equal(exception.Message, $"The maximum message size of {maximumMessageSize}B was exceeded. The message size can be configured in AddHubOptions.");
+        }
+
+        [Fact]
+        public async Task ManyHubMessagesUnderTheMessageSizeButConfiguredWithMax()
```
JamesNK: good test 👍

src/SignalR/server/Core/src/Internal/HubOptionsSetup.cs
davidfowl: Done

JamesNK (approved):

src/SignalR/server/Core/src/HubOptions.cs
```diff
@@ -36,6 +36,11 @@ public class HubOptions
         /// </summary>
         public IList<string> SupportedProtocols { get; set; } = null;
 
+        /// <summary>
+        /// Gets or sets the maximum message size of a single incoming hub message. The default is 32KB.
```
BrennanConroy: "Null means no maximum message size" or is that implicit with nullable types

BrennanConroy (approved):

Issue #7305, opened 2019-02-06
Full screen websites are not full screen!?


rakkarage:
### Describe the bug
All the dotnet new commands that generate a webpage generate it the at wrong size?
A full screen web page does not need scroll bars and it does not need to scroll?

### To Reproduce
https://docs.microsoft.com/en-ca/aspnet/core/getting-started/?view=aspnetcore-2.2&tabs=linux
dotnet new webapi
dotnet new webapp
dotnet new razor
dotnet new mvc
dotnet new etc

### Expected behavior
A full screen web page (for example: google.com) should be the exact size of the page and have no scroll bars and no scroll?
The asp net pages have scroll bars and scroll and are X pixels bugger then the page?
http://jsfiddle.net/robdodson/cZseJ/

### Screenshots

![mkkkm](https://user-images.githubusercontent.com/962340/52352878-6fc19b00-29fb-11e9-874b-5df1f1bf08f2.gif)

<img width="291" alt="home-page" src="https://user-images.githubusercontent.com/962340/52352883-73552200-29fb-11e9-829f-b68562a2b493.png">

I am sure there is an easy solution. Excuse my ignorance.

Thanks.

rakkarage:
http://jsfiddle.net/robdodson/cZseJ/

mkArtakMSFT:
Thanks for contacting us, @rakkarage.
Would you like to submit a PR for this?

mkArtakMSFT:
@ryanbrandenburg, let's see whether @rakkarage would send us a PR for this. If not, we may probably just do this as part of some future templates changes.

rakkarage:
Thanks. Sorry it is currently beyond me. :)

Andrzej-W:
There is a small bug in site.css. We have this in _Layout.cshtml
```html
    <footer class="border-top footer text-muted">
        <div class="container">
            &copy; 2019 - ASPTest2 - <a asp-area="" asp-controller="Home" asp-action="Privacy">Privacy</a>
        </div>
    </footer>

```
and this in site.css
```css
.border-top {
  border-top: 1px solid #e5e5e5;
}

.footer {
  position: absolute;
  bottom: 0;
  width: 100%;
  white-space: nowrap;
  /* Set the fixed height of the footer here */
  height: 60px;
  line-height: 60px; /* Vertically center the text there */
}
```
[...]

PR #8136, opened 2019-03-03
Fixing height of footer

natery2000:
Adjusted the height of the footer css for the template

Addresses #7305


pranavkm:
The issue suggests removing the height entirely:

> In my opinion height: 60px is not needed - you can remove it and scrollbar disappears. I think this the best and the simplest fix.

Should we use that instead?

natery2000:
I can change the PR to that, if we think so.

Issue #7283, opened 2019-02-05
Provide friendly error message when manually constructing UrlHelper with endpoint routing


JamesNK:
re: https://github.com/aspnet/AspNetCore/issues/5135#issuecomment-460777543

`UrlHelper` doesn't work when endpoint routing is enabled. Lets provide friendly error message when someone manually constructs `UrlHelper` when endpoint routing is enabled. Message would explain the cause of the error and how to fix it.

PR #8145, opened 2019-03-04
Add friendly error message to UrlHelper for when there is no IRouter

JamesNK:
Addresses #7283

### DIFF
--- a/src/Mvc/Mvc.Core/src/Routing/UrlHelper.cs
+++ b/src/Mvc/Mvc.Core/src/Routing/UrlHelper.cs
@@ -40,8 +40,8 @@ namespace Microsoft.AspNetCore.Mvc.Routing
                 if (routers.Count == 0)
                 {
                     throw new InvalidOperationException("Could not find an IRouter associated with the ActionContext. "
-                        + "If your application is using endpoint routing then you can get a UrlHelper instance with "
-                        + "dependency injection or use Microsoft.AspNetCore.Routing.LinkGenerator.");
+                        + "If your application is using endpoint routing then you can get a IUrlHelperFactory with "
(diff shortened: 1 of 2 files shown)


src/Mvc/Mvc.Core/src/Routing/UrlHelper.cs
```diff
@@ -32,7 +32,21 @@ public UrlHelper(ActionContext actionContext)
...
         /// </summary>
-        protected IRouter Router => ActionContext.RouteData.Routers[0];
+        protected IRouter Router
+        {
+            get
+            {
+                var routers = ActionContext.RouteData.Routers;
+                if (routers.Count == 0)
+                {
+                    throw new InvalidOperationException("Could not find an IRouter associated with the ActionContext. "
```
JamesNK: I also tried this in the UrlHelper constructor and it broke tests. Placing the check here means that the exception will only be thrown if the IndexOutOfRangeException was already going to be thrown.

rynowak: Sounds good.

src/Mvc/Mvc.Core/src/Routing/UrlHelper.cs
```diff
@@ -32,7 +32,21 @@ public UrlHelper(ActionContext actionContext)
...
+                        + "If your application is using endpoint routing then you can get a UrlHelper instance with "
+                        + "dependency injection or use Microsoft.AspNetCore.Routing.LinkGenerator.");
```
rynowak: This is false. You can't get a `UrlHelper` from DI. That's one of the problems that link generator solves.

users can instead resolve an `IUrlHelperFactory` and put in the action context. Don't have an action context? that's another one of the problems link generator solves.

JamesNK:
:up: :date:

rynowak (approved):

PR #8147, opened 2019-03-04
Don't auto include MVC in all samples

davidfowl:
- Turn of assembly info generation by default in samples
- This stops solutions from building that don't have MVC
- It slows the build of projects that don't need MVC down

PS: It's likely we'll need to turn this on for same of the MVC samples? We'll find out 😄

### DIFF
--- a/eng/Workarounds.props
+++ b/eng/Workarounds.props
@@ -12,6 +12,11 @@
     <_ProjectExtensionsWereImported>true</_ProjectExtensionsWereImported>
     <WixTargetsPath>$(WixInstallPath)\wix2010.targets</WixTargetsPath>
   </PropertyGroup>
+  
+  <PropertyGroup>
+    <!-- Required because the Razor SDK will generate attributes -->
+    <GenerateRazorAssemblyInfo Condition="'$(UsingMicrosoftNETSdkWeb)' == 'true' AND '$(GenerateRazorAssemblyInfo)' == ''">false</GenerateRazorAssemblyInfo>
+  </PropertyGroup>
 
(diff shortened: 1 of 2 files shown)


eng/Workarounds.props
```diff
@@ -12,6 +12,11 @@
     <_ProjectExtensionsWereImported>true</_ProjectExtensionsWereImported>
     <WixTargetsPath>$(WixInstallPath)\wix2010.targets</WixTargetsPath>
   </PropertyGroup>
+  
+  <PropertyGroup>
+    <!-- Required because the Razor SDK will generate attributes -->
+    <GenerateRazorAssemblyInfo Condition="'$(UsingMicrosoftNETSdkWeb)' == 'true' AND '$(GenerateRazorAssemblyInfo)' == ''">false</GenerateRazorAssemblyInfo>
```
davidfowl: I tried adding `AND '$(TargetFrameworkIdentifier)' == '.NETCoreApp'` but it wasn't working. Likely something to do with a variable not being evaluated yet...

davidfowl:
Trying to find a way to set GenerateRazorAssemblyInfo if there's a reference to MVC but my msbuild fu is failing me.

dougbu:
@davidfowl [item functions](https://docs.microsoft.com/en-us/visualstudio/msbuild/item-functions?view=vs-2019) should do the trick. I'm thinking of `@(ReferencePath)` items and a function like `@(MyItem->AnyHaveMetadataValue("MetadataName", "MetadataValue"))`.

Might need to play a bit with bin logs to choose the best metadata to check and the expected value.

rynowak:
Yeah I would deffo expect this to break stuff. You'll have to opt into this in the MVC-using projects. Those failures will be hard to chase down, because they will be things like runtime compilation not working, or views not loading.

PR #8195, opened 2019-03-04
Update to C# 8.0

pranavkm:


BrennanConroy:
...
https://github.com/aspnet/AspNetCore/pull/8128

eng/targets/CSharp.Common.props
```diff
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup>
-    <LangVersion>7.3</LangVersion>
+    <LangVersion>8.0</LangVersion>
```
pranavkm: The SDK we have doesn't support "preview" (https://devblogs.microsoft.com/dotnet/an-update-to-c-versions-and-c-tooling) as yet. I think we may want to eventually just use the LangVersion the SDK suggests. In the meanwhile, this seems like a reasonablestop-gap.

PR #8197, opened 2019-03-05
Isolate message compiler output files

pakrym:
Fixes: https://github.com/aspnet/AspNetCore/issues/8126

Let every project that needs compiled message file compile it on their own. Place compiled output if platform-specific subdirectory so they don't clash.


jkotalik (approved):

pakrym:
https://github.com/aspnet/AspNetCore-Internal/issues/1751

PR #8199, opened 2019-03-05
Make StartAsync not throw if we haven't started the response

jkotalik:
For https://github.com/aspnet/AspNetCore/issues/7779.

One scenario this doesn't support is calling StartAsync after calling GetMemory and using the memory immediately after. @halter73 and I deemed it too difficult to fix well and is an incredibly niche scenario.
```c#
public async Task AppFunc(HttpContext context)
{
    var memory = context.Response.BodyPipe.GetMemory();
    await context.Response.StartAsync();
    // Try using the memory here, you'll get an ODE.
    // Advance will also 
}
```

I added a check that would see if Advance was called immediately after calling StartAsync and throw if so.

### DIFF
--- a/src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
+++ b/src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
@@ -217,7 +217,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                 {
                     if (bytes >= 0)
                     {
-                        if (_currentSegment.Length < _position + bytes)
+                        if (_currentSegment.Length - bytes < _position)
                         {
                             throw new ArgumentOutOfRangeException("Can't advance past buffer size.");
                         }
@@ -227,7 +227,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                 }
                 else if (_autoChunk)
                 {
-                    if (bytes + _advancedBytesForChunk > _currentChunkMemory.Length - BeginChunkLengthMax - EndChunkLength)
+                    if (_advancedBytesForChunk > _currentChunkMemory.Length - BeginChunkLengthMax - EndChunkLength - bytes)
                     {
                         throw new ArgumentOutOfRangeException("Can't advance past buffer size.");
                     }
--- a/src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
+++ b/src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
@@ -43,7 +43,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
         private long _unflushedBytes;
         private bool _autoChunk;
         private readonly PipeWriter _pipeWriter;
-        private const int MemorySizeThreshold = 1024;
         private const int BeginChunkLengthMax = 5;
         private const int EndChunkLength = 2;
 
@@ -606,7 +605,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 
             // If the sizeHint is 0, any capacity will do
             // Otherwise, the buffer must have enough space for the entire size hint, or we need to add a segment.
-            if ((sizeHint == 0 && remainingSize > 0) || (sizeHint > 0 && remainingSize >= Math.Min(MemorySizeThreshold, sizeHint)))
+            if ((sizeHint == 0 && remainingSize > 0) || (sizeHint > 0 && remainingSize >= sizeHint))
             {
                 // We have capacity in the current segment
                 return;
--- a/src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
+++ b/src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
@@ -355,11 +355,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 
                 _position = 0;
 
-                if (_currentSegmentOwner != null)
-                {
-                    _currentSegmentOwner.Dispose();
-                    _currentSegmentOwner = null;
-                }
+                DisposeCurrentSegment();
             }
         }
 
@@ -382,13 +378,17 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                     }
                 }
 
-                if (_currentSegmentOwner != null)
-                {
-                    _currentSegmentOwner.Dispose();
-                }
+                DisposeCurrentSegment();
 
                 CompletePipe();
             }
+        }
+
+        private void DisposeCurrentSegment()
+        {
+            _currentSegmentOwner?.Dispose();
+            _currentSegmentOwner = null;
+            _currentSegment = default;
         }
 
         private void CompletePipe()
@@ -658,7 +658,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             public Memory<byte> Buffer { get; }
             public int Length { get; }
 
-            public ReadOnlySpan<byte> Span => Buffer.Span;
+            public ReadOnlySpan<byte> Span => Buffer.Span.Slice(0, Length);
 
             public CompletedBuffer(IMemoryOwner<byte> owner, Memory<byte> buffer, int length)
             {
--- a/src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
+++ b/src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
@@ -226,14 +226,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                         _position += bytes;
                     }
                 }
-
-                if (_autoChunk)
+                else if (_autoChunk)
                 {
-                    if (bytes < 0)
-                    {
-                        throw new ArgumentOutOfRangeException(nameof(bytes));
-                    }
-
                     if (bytes + _advancedBytesForChunk > _currentChunkMemory.Length - BeginChunkLengthMax - EndChunkLength)
                     {
                         throw new InvalidOperationException("Can't advance past buffer size.");
@@ -318,7 +312,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 
             writer.Commit();
 
-            _unflushedBytes += writer.BytesCommitted;
             _autoChunk = autoChunk;
             // I think I should copy everything to the pipe here....
 
@@ -327,9 +320,16 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             for (var i = 0; i < count; i++)
             {
                 var segment = _completedSegments[i];
-                var memory = _pipeWriter.GetMemory(segment.Length);
-                segment.Buffer.CopyTo(memory);
-                _pipeWriter.Advance(segment.Length);
+
+                if (_autoChunk)
+                {
+                    CommitChunkInternal(ref writer, segment.Span);
+                }
+                else
+                {
+                    writer.Write(segment.Span);
+                    writer.Commit();
+                }
                 segment.Return();
             }
 
@@ -338,10 +338,17 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             if (!_currentSegment.IsEmpty)
             {
                 var segment = _currentSegment.Slice(0, _position);
-                var memory = _pipeWriter.GetMemory(segment.Length);
-                segment.CopyTo(memory);
-                _pipeWriter.Advance(segment.Length);
-                _position = 0;
+
+                if (_autoChunk)
+                {
+                    CommitChunkInternal(ref writer, segment.Span);
+                }
+                else
+                {
+                    writer.Write(segment.Span);
+                    writer.Commit();
+                    _position = 0;
+                }
 
                 if (_currentSegmentOwner != null)
                 {
@@ -350,6 +357,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                 }
             }
 
+            _unflushedBytes += writer.BytesCommitted;
             _startCalled = true;
         }
 
--- a/src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
+++ b/src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Buffers;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO.Pipelines;
 using System.Threading;
@@ -55,6 +56,12 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
         private Memory<byte> _currentChunkMemory;
         private bool _currentChunkMemoryUpdated;
         private IMemoryOwner<byte> _fakeMemoryOwner;
+        private bool _startCalled;
+
+        private List<CompletedBuffer> _completedSegments;
+        private Memory<byte> _currentSegment;
+        private IMemoryOwner<byte> _currentSegmentOwner;
+        private int _position;
 
         public Http1OutputProducer(
             PipeWriter pipeWriter,
@@ -72,6 +79,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             _minResponseDataRateFeature = minResponseDataRateFeature;
             _flusher = new TimingPipeFlusher(pipeWriter, timeoutControl, log);
             _memoryPool = memoryPool;
+            _completedSegments = new List<CompletedBuffer>();
         }
 
         public Task WriteDataAsync(ReadOnlySpan<byte> buffer, CancellationToken cancellationToken = default)
@@ -158,6 +166,11 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                 {
                     return GetFakeMemory(sizeHint);
                 }
+                else if (!_startCalled)
+                {
+                    // TODO how does chunked play with this?
+                    return LeasedMemory(sizeHint);
+                }
                 else if (_autoChunk)
                 {
                     return GetChunkedMemory(sizeHint);
@@ -177,6 +190,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                 {
                     return GetFakeMemory(sizeHint).Span;
                 }
+                else if (!_startCalled)
+                {
+                    return LeasedMemory(sizeHint).Span;
+                }
                 else if (_autoChunk)
                 {
                     return GetChunkedMemory(sizeHint).Span;
@@ -195,6 +212,19 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                 if (_completed)
                 {
                     return;
+                }
+
+                if (!_startCalled)
+                {
+                    if (bytes >= 0)
+                    {
+                        if (_currentSegment.Length < _position + bytes)
+                        {
+                            throw new InvalidOperationException("Can't advance past buffer size.");
+                        }
+
+                        _position += bytes;
+                    }
                 }
 
                 if (_autoChunk)
@@ -290,6 +320,37 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 
             _unflushedBytes += writer.BytesCommitted;
             _autoChunk = autoChunk;
+            // I think I should copy everything to the pipe here....
+
+            // try it for now?
+            var count = _completedSegments.Count;
+            for (var i = 0; i < count; i++)
+            {
+                var segment = _completedSegments[i];
+                var memory = _pipeWriter.GetMemory(segment.Length);
+                segment.Buffer.CopyTo(memory);
+                _pipeWriter.Advance(segment.Length);
+                segment.Return();
+            }
+
+            _completedSegments.Clear();
+
+            if (!_currentSegment.IsEmpty)
+            {
+                var segment = _currentSegment.Slice(0, _position);
+                var memory = _pipeWriter.GetMemory(segment.Length);
+                segment.CopyTo(memory);
+                _pipeWriter.Advance(segment.Length);
+                _position = 0;
+
+                if (_currentSegmentOwner != null)
+                {
+                    _currentSegmentOwner.Dispose();
+                    _currentSegmentOwner = null;
+                }
+            }
+
+            _startCalled = true;
         }
 
         public void Dispose()
@@ -300,6 +361,18 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                 {
                     _fakeMemoryOwner.Dispose();
                     _fakeMemoryOwner = null;
+                }
+
+                if (_completedSegments != null)
+                {
+                    foreach (var segment in _completedSegments)
+                    {
+                        segment.Return();
+                    }
+                }
+                if (_currentSegmentOwner != null)
+                {
+                    _currentSegmentOwner.Dispose();
                 }
 
                 CompletePipe();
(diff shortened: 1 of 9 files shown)


jkotalik:
@anurse test discovery issue?

```
[xUnit.net 00:00:05.78] Microsoft.AspNetCore.SignalR.StackExchangeRedis.Tests: Exception thrown during theory discovery on 'Microsoft.AspNetCore.SignalR.StackExchangeRedis.Tests.RedisEndToEndTests.HubConnectionCanSendAndReceiveMessages'; falling back to single test case.
System.InvalidOperationException: No process is associated with this object.
   at System.Diagnostics.Process.EnsureState(State state)
   at System.Diagnostics.Process.EnsureState(State state)
   at System.Diagnostics.Process.GetWaitState()
   at System.Diagnostics.Process.WaitForExitCore(Int32 milliseconds)
   at System.Diagnostics.Process.WaitForExit()
   at Microsoft.AspNetCore.SignalR.StackExchangeRedis.Tests.Docker.RunProcessAndWait(String fileName, String arguments, String prefix, ILogger logger, TimeSpan timeout, String& output) in /_/src/SignalR/server/StackExchangeRedis/test/Docker.cs:line 163
   at Microsoft.AspNetCore.SignalR.StackExchangeRedis.Tests.Docker.RunCommand(String commandAndArguments, String prefix, ILogger logger, String& output) in /_/src/SignalR/server/StackExchangeRedis/test/Docker.cs:line 135
   at Microsoft.AspNetCore.SignalR.StackExchangeRedis.Tests.Docker.RunCommand(String commandAndArguments, String prefix, String& output) in /_/src/SignalR/server/StackExchangeRedis/test/Docker.cs:line 131
   at Microsoft.AspNetCore.SignalR.StackExchangeRedis.Tests.Docker.Create() in /_/src/SignalR/server/StackExchangeRedis/test/Docker.cs:line 44
   at System.Lazy`1.ViaFactory(LazyThreadSafetyMode mode)
   at System.Lazy`1.ExecutionAndPublication(LazyHelper executionAndPublication, Boolean useDefaultConstructor)
   at System.Lazy`1.CreateValue()
   at System.Lazy`1.get_Value()
   at Microsoft.AspNetCore.SignalR.StackExchangeRedis.Tests.Docker.get_Default() in /_/src/SignalR/server/StackExchangeRedis/test/Docker.cs:line 23
   at Microsoft.AspNetCore.SignalR.StackExchangeRedis.Tests.SkipIfDockerNotPresentAttribute.CheckDocker() in /_/src/SignalR/server/StackExchangeRedis/test/SkipIfDockerNotPresentAttribute.cs:line 17
   at Microsoft.AspNetCore.SignalR.StackExchangeRedis.Tests.SkipIfDockerNotPresentAttribute.get_IsMet() in /_/src/SignalR/server/StackExchangeRedis/test/SkipIfDockerNotPresentAttribute.cs:line 12
   at Microsoft.AspNetCore.Testing.xunit.TestMethodExtensions.EvaluateSkipConditions(ITestMethod testMethod)
   at Microsoft.Extensions.Logging.Testing.LoggedConditionalTheoryDiscoverer.CreateTestCasesForDataRow(ITestFrameworkDiscoveryOptions discoveryOptions, ITestMethod testMethod, IAttributeInfo theoryAttribute, Object[] dataRow)
   at Xunit.Sdk.TheoryDiscoverer.Discover(ITestFrameworkDiscoveryOptions discoveryOptions, ITestMethod testMethod, IAttributeInfo theoryAttribute) in C:\Dev\xunit\xunit\src\xunit.execution\Sdk\Frameworks\TheoryDiscoverer.cs:line 238
```

BrennanConroy:
Didn't realize this was failing builds, should be fixed with https://github.com/aspnet/AspNetCore/pull/8201

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
```diff
@@ -56,6 +57,13 @@ public class Http1OutputProducer : IHttpOutputProducer, IHttpOutputAborter, IDis
         private bool _currentChunkMemoryUpdated;
         private IMemoryOwner<byte> _fakeMemoryOwner;
 
+        // Fields needed to store writes before calling either startAsync or Write/FlushAsync
+        private LinkedList<CompletedBuffer> _completedSegments;
```
davidfowl: Why a linked list?

jkotalik: I'm not sure of the performance of a LinkedList in C#. But we don't need indexing, just iteration, so it seemed logical to use just a LinkedList?

Is it still better to use a List instead?

davidfowl: Yes, just use a list. There's nothing here that requires a linked list.

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
```diff
@@ -197,13 +213,20 @@ public void Advance(int bytes)
...
+                if (!_startCalled)
                 {
-                    if (bytes < 0)
+                    if (bytes >= 0)
                     {
-                        throw new ArgumentOutOfRangeException(nameof(bytes));
-                    }
+                        if (_currentSegment.Length < _position + bytes)
+                        {
+                            throw new InvalidOperationException("Can't advance past buffer size.");
```
davidfowl: ArgumentOutOfRangeException. Also why not slice _currentSegment?

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
```diff
@@ -197,13 +213,20 @@ public void Advance(int bytes)
...
 
-                if (_autoChunk)
+                if (!_startCalled)
                 {
-                    if (bytes < 0)
+                    if (bytes >= 0)
                     {
-                        throw new ArgumentOutOfRangeException(nameof(bytes));
-                    }
+                        if (_currentSegment.Length < _position + bytes)
```
davidfowl: This can overflow

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
```diff
@@ -197,13 +213,20 @@ public void Advance(int bytes)
...
+                        }
 
+                        _position += bytes;
+                    }
+                }
+                else if (_autoChunk)
+                {
                     if (bytes + _advancedBytesForChunk > _currentChunkMemory.Length - BeginChunkLengthMax - EndChunkLength)
                     {
                         throw new InvalidOperationException("Can't advance past buffer size.");
```
davidfowl: This should also be an ArgumentOutOfRangeException

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
jkotalik: _sigh_

jkotalik: What's our recommendation for checking overflows? Do we ever use checked? https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/checked

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
jkotalik: Ah this was very recent!

jkotalik: https://github.com/dotnet/corefx/commit/498a7e169a452d3e53a061c1433aea637506a14a#diff-59ca508ec7e0ac1f99fa1b3f030f25ac

jkotalik:
@JamesNK and @JunTaoLuo are you okay with the restrictions on StartAsync that I listed above? After calling StartAsync, don't use the memory that was obtained before calling StartAsync.

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
Tratcher: `_currentSegment.Length - bytes < _position`

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
```diff
@@ -302,6 +373,20 @@ public void Dispose()
...
+                // Call dispose on any memory that wasn't written.
+                if (_completedSegments != null)
+                {
+                    foreach (var segment in _completedSegments)
+                    {
+                        segment.Return();
+                    }
+                }
+
+                if (_currentSegmentOwner != null)
```
Tratcher: ?.Dispose()

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
```diff
@@ -1066,7 +1072,7 @@ private HttpResponseHeaders CreateResponseHeaders(bool appCompleted)
                 {
                     _keepAlive = false;
                 }
-                else if (appCompleted || !_canWriteResponseBody)
+                else if ((appCompleted || !_canWriteResponseBody) && !_hasAdvanced) // Avoid setting contentLength of 0 if we wrote data before calling CreateResponseHeaders
```
Tratcher: if appCompleted && _hasAdvanced then we can set ContentLength to a specific value and avoid chunking.

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
```diff
@@ -45,7 +45,11 @@ public abstract partial class HttpProtocol : IDefaultHttpContextContainer, IHttp
         // Keep-alive is default for HTTP/1.1 and HTTP/2; parsing and errors will change its value
         // volatile, see: https://msdn.microsoft.com/en-us/library/x13ttww7.aspx
         protected volatile bool _keepAlive = true;
-        private bool _canWriteResponseBody;
+        // _canWriteResponseBody is set in CreateResponseHeaders.
+        // If we are writing with GetMemory/Advance before calling StartAsync, assume we can write and throw away contents if we can't.
+        private bool _canWriteResponseBody = true;
+        private bool _hasAdvanced;
+        private bool _requireGetMemoryNextCall;
```
Tratcher: _isLeasedMemoryValid

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
```diff
@@ -1268,6 +1274,21 @@ public void ReportApplicationError(Exception ex)
...
+            if (bytes < 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(bytes));
+            }
+            else if (bytes > 0)
+            {
+                _hasAdvanced = true;
+            }
+
+            if (_requireGetMemoryNextCall)
```
Tratcher: What would happen if I tried to call Advance without ever calling GetMemory?

jkotalik: No matter what that is invalid, but with this new logic, Advance wouldn't be called on the PipeWriter. Agree that it is a bit confusing. Thoughts on how to improve it?

Tratcher: But what error do you get? If it gets to here this won't trigger because the default value is false. You could change the default.

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
halter73: We should probably do something similar for _advancedBytesForChunk

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
```diff
@@ -197,16 +213,23 @@ public void Advance(int bytes)
                     return;
                 }
 
-                if (_autoChunk)
+                if (!_startCalled)
                 {
-                    if (bytes < 0)
+                    if (bytes >= 0)
                     {
-                        throw new ArgumentOutOfRangeException(nameof(bytes));
```
halter73: Do you still throw if bytes is less than zero?

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
```diff
@@ -506,5 +591,90 @@ private Memory<byte> GetFakeMemory(int sizeHint)
...
+            public int Length { get; }
+
+            public ReadOnlySpan<byte> Span => Buffer.Span;
+
+            public CompletedBuffer(IMemoryOwner<byte> owner, Memory<byte> buffer, int length)
+            {
+                _memoryOwner = owner;
+
+                Buffer = buffer;
+                Length = length;
```
halter73: It doesn't look like CompletedBuffer.Length is referenced anywhere. Shouldn't this be used to slice CompletedBuffer.Span?

JunTaoLuo:
@jkotalik Yes we are fine with that restriction, We only call GetSpan after calling StartAsync.

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
```diff
@@ -288,8 +311,56 @@ private void WriteResponseHeadersInternal(ref BufferWriter<PipeWriter> writer, i
...
+                    writer.Commit();
+                }
+
+                _position = 0;
+
+                if (_currentSegmentOwner != null)
+                {
+                    _currentSegmentOwner.Dispose();
+                    _currentSegmentOwner = null;
+                }
```
halter73: Should we also be setting _currentSegment to default?

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
```diff
@@ -302,6 +373,20 @@ public void Dispose()
...
+                {
+                    foreach (var segment in _completedSegments)
+                    {
+                        segment.Return();
+                    }
+                }
+
+                if (_currentSegmentOwner != null)
+                {
+                    _currentSegmentOwner.Dispose();
```
halter73: Why set _currentSegmentOwner to null in WriteDataWrittenBeforeHeaders but not here?

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
halter73: Should this be initialized to true?

src/Servers/Kestrel/test/InMemory.FunctionalTests/ChunkedResponseTests.cs
```diff
@@ -421,15 +421,21 @@ public async Task ChunksCanBeWrittenManually()
             }
         }
 
-        [Fact]
-        public async Task ChunksWithGetMemoryBeforeFirstFlushStillFlushes()
+        [Theory]
+        [InlineData(false)]
+        [InlineData(true)]
```
halter73: I'd rather have a few tests explicitly for calling/not calling StartAsync rather than adding variations to existing tests.

src/Servers/Kestrel/test/InMemory.FunctionalTests/Http2/Http2StreamTests.cs
```diff
@@ -2634,8 +2634,8 @@ public async Task GetMemoryAdvance_Works()
             };
             await InitializeConnectionAsync(async httpContext =>
             {
+                await Task.CompletedTask;
```
halter73: ?

src/Servers/Kestrel/test/InMemory.FunctionalTests/Http2/Http2StreamTests.cs
```diff
@@ -3174,7 +3214,7 @@ public async Task ResponseBodyAndResponsePipeWorks()
             {
                 var response = httpContext.Response;
                 response.ContentLength = 54;
-                await response.StartAsync();
+                await Task.CompletedTask;
```
halter73: Remove these lines.

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
jkotalik: So I was thinking about doing this in this PR but wasn't a fan. You would need to peek into the number of advanced bytes inside of the Http1OutputProducer which isn't clean. I'd prefer to do it in a separate pr.

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
jkotalik: Yeah I'll make it true by default

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
jkotalik: Yeah, in HttpProtocol now.

JamesNK:
> @JamesNK and @JunTaoLuo are you okay with the restrictions on StartAsync that I listed above? After calling StartAsync, don't use the memory that was obtained before calling StartAsync.

I'm pretty sure that is fine. I'll take a closer look tomorrow.

src/Servers/Kestrel/test/InMemory.FunctionalTests/Http2/Http2StreamTests.cs
jkotalik: Got lazy removing async XD

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
jkotalik: It will never be accessed again, so we don't need to. But I'll go ahead and set it.

src/Servers/Kestrel/test/InMemory.FunctionalTests/ChunkedResponseTests.cs
jkotalik: Are you asking for me to write new tests that try different StartAsync/NonStartAsync scenarios, or are you asking that I split the Theory into two different tests?

I'm assuming you are asking for both when appropriate.

jkotalik:
@JamesNK @JunTaoLuo the only thing I'm concerned about is how you mentioned that you wanted to avoid calling StartAsync to have buffering for free so you can modify headers later. It may cause issues if you call StartAsync later, putting you in a weird state.

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
jkotalik: Yeah.

JamesNK:
We wouldn't call `StartAsync` while in the middle of writing to `Memory<byte>`. The scenario we have is writing to the span/memory and calling Advance, then later calling `StartAsync`.

JunTaoLuo:
> then later calling `StartAsync`.
@JamesNK I think you are thinking of `WriteResponseHeadersAsyncCore` which calls `FlushAsync`: https://github.com/grpc/grpc-dotnet/blob/master/src/Grpc.AspNetCore.Server/Internal/HttpContextServerCallContext.cs#L201.

Or did you mean something else?

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
```diff
@@ -45,7 +45,11 @@ public abstract partial class HttpProtocol : IDefaultHttpContextContainer, IHttp
         // Keep-alive is default for HTTP/1.1 and HTTP/2; parsing and errors will change its value
         // volatile, see: https://msdn.microsoft.com/en-us/library/x13ttww7.aspx
         protected volatile bool _keepAlive = true;
-        private bool _canWriteResponseBody;
+        // _canWriteResponseBody is set in CreateResponseHeaders.
+        // If we are writing with GetMemory/Advance before calling StartAsync, assume we can write and throw away contents if we can't.
+        private bool _canWriteResponseBody = true;
+        private bool _hasAdvanced;
+        private bool _isLeasedMemoryInvalid = true;
```
halter73: Should _hasAdvanced and _isLeasedMemoryInvalid be reset between requests?

jkotalik:
🆙 📅

src/Servers/Kestrel/shared/test/TestApp.cs
```diff
@@ -49,7 +49,6 @@ public static async Task EchoAppPipeWriter(HttpContext httpContext)
             if (buffer.Length > 0)
             {
                 await request.Body.ReadUntilEndAsync(buffer).DefaultTimeout();
-                await response.StartAsync();
```
halter73: I don't think we should have unrelated tests hit the inefficient code path. Tests like ChunksWithGetMemoryBeforeFirstFlushStillFlushes should be sufficient.

src/Servers/Kestrel/test/InMemory.FunctionalTests/ChunkedResponseTests.cs
```diff
@@ -534,6 +579,8 @@ public async Task ChunksWithGetMemoryWithInitialFlushWorks()
             {
                 var response = httpContext.Response;
 
+                await response.StartAsync();
+
```
halter73: The old version of this test would still pass, right?

src/Servers/Kestrel/test/InMemory.FunctionalTests/ChunkedResponseTests.cs
```diff
@@ -581,6 +628,65 @@ public async Task ChunksWithGetMemoryWithInitialFlushWorks()
...
+            var testContext = new TestServiceContext(LoggerFactory);
+
+            using (var server = new TestServer(async httpContext =>
+            {
+                var response = httpContext.Response;
+
+                await response.StartAsync();
+
+                var memory = response.BodyWriter.GetMemory();
+                length.Value = memory.Length - 1;
```
halter73: FYI: if length was a normal int, this test would still work since the closure would box it.

src/Servers/Kestrel/test/InMemory.FunctionalTests/Http2/Http2StreamTests.cs
```diff
@@ -3088,7 +3174,7 @@ public async Task ContentLengthWithGetMemoryWorks()
             {
                 var response = httpContext.Response;
                 response.ContentLength = 12;
-                await response.StartAsync();
+                await Task.CompletedTask;
```
halter73: Nit

src/Servers/Kestrel/test/InMemory.FunctionalTests/Http2/Http2StreamTests.cs
```diff
@@ -2681,7 +2727,6 @@ public async Task WriteAsync_GetMemoryLargeWriteBeforeFirstFlush()
             await InitializeConnectionAsync(async httpContext =>
             {
                 var response = httpContext.Response;
-                await response.StartAsync();
```
halter73: If the test isn't explicitly about omitting StartAsync, and calling StartAsync is more efficient, call StartAsync.

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
halter73: Is there a regression test that will fail if we forgot to reset?

halter73 (approved):
Pending a few nits. I'd like a test verifying _isLeasedMemoryInvalid is reset for example.

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
jkotalik: Ya caught me.

jkotalik:
Talked offline, there is actually a good amount of missed scenarios here, especially with resetting state in the Http1OutputProducer. I'll take a look at how to refactor this to avoid this issue in the future.

jkotalik:
@halter73 I added a reset method to HttpOutputProducer. All other ways weren't clean.

src/Servers/Kestrel/test/InMemory.FunctionalTests/ResponseTests.cs
```diff
@@ -3516,6 +3512,217 @@ public async Task ResponseBodyPipeCompleteWithoutExceptionWritesDoNotThrow()
...
+                        "0",
+                        "",
+                        "");
+                }
+                await server.StopAsync();
+            }
+        }
+
+        [Fact]
+        public async Task ResponseStartCalledStateIsResetWithMultipleReqeusts()
```
jkotalik: @halter73 is this sufficient for the type of test you wanted?

halter73: Yep. I would like to to flip back to chunked one last time to ensure the content length response to chunked response transition also works.

You can shorten the test slightly by pipelining the requests like we do in ZeroContentLengthNotSetAutomaticallyForCertainStatusCodes.

halter73 (approved):

src/Servers/Kestrel/test/InMemory.FunctionalTests/ResponseTests.cs
jkotalik: Its in a for loop so it already does the Chunked to Content Length transition.

PR #8200, opened 2019-03-05
Send window updates based on examined rather than consumed.

jkotalik:
Fixes: https://github.com/aspnet/AspNetCore/issues/8105

Will merge https://github.com/aspnet/AspNetCore/pull/8198 separate from this change (the other half).


src/Servers/Kestrel/Core/src/Internal/Http2/Http2MessageBody.cs
```diff
@@ -14,6 +14,7 @@ public class Http2MessageBody : MessageBody
     {
         private readonly Http2Stream _context;
         private ReadResult _readResult;
+        private SequencePosition? _prevExamined;
```
davidfowl: Looking at the other failure I'm handling in System.IO.Pipelines, I think this is fundamentally busted 😢 . It's impossible to use SequencePosition across read calls today.

jkotalik: Any change we can make to SequencePosition to make this work?

jkotalik:
Need https://github.com/aspnet/AspNetCore/pull/8198 before this can be merged.

src/Servers/Kestrel/Core/src/Internal/Http2/Http2MessageBody.cs
```diff
@@ -63,14 +64,23 @@ public override void AdvanceTo(SequencePosition consumed)
 
         public override void AdvanceTo(SequencePosition consumed, SequencePosition examined)
         {
-            var dataLength = _readResult.Buffer.Slice(_readResult.Buffer.Start, consumed).Length;
+            var dataLength = _readResult.Buffer.Slice(_readResult.Buffer.Start, examined).Length;
+            var newlyExamined = dataLength - _totalExamined;
```
Tratcher: This only make sense if _totalExamined is _totalExamined per _readResult. You would need to reset _totalExamined when you update _readResult.

jkotalik: You are correct. I think I can fix this via using consumed. I'll update it shortly.

src/Servers/Kestrel/Core/src/Internal/Http2/Http2MessageBody.cs
```diff
@@ -63,14 +64,23 @@ public override void AdvanceTo(SequencePosition consumed)
...
             _context.RequestBodyPipe.Reader.AdvanceTo(consumed, examined);
-            OnDataRead(dataLength);
+
+            if (newlyExamined > 0)
+            {
+                OnDataRead(dataLength);
```
halter73: Nit: Rename to OnDataExamined.

Also, when you fix the newlyExamined logic, you'll want to pass in newlyExamined instead of dataLength.

src/Servers/Kestrel/test/InMemory.FunctionalTests/Http2/Http2StreamTests.cs
```diff
@@ -2492,7 +2492,6 @@ public async Task Write_DoesNotBlockWriteLoop()
             await SendWindowUpdateAsync(1, 2);
             await SendWindowUpdateAsync(0, 2);
 
-            // Remaining 1 byte from the first write and then the second write
```
halter73: ?

jkotalik:
@halter73 @Tratcher I updated the logic. I _think_ this works. I'll add tests appropriately once we confirm my logic makes sense.

src/Servers/Kestrel/test/InMemory.FunctionalTests/Http2/Http2StreamTests.cs
jkotalik: Who knows how that happened ¯\\_(ツ)_/¯

src/Servers/Kestrel/Core/src/Internal/Http2/Http2Stream.cs
```diff
@@ -57,7 +57,7 @@ public Http2Stream(Http2StreamContext context)
                 this,
                 context.ServiceContext.Log);
 
-            RequestBodyPipe = CreateRequestBodyPipe(context.ServerPeerSettings.InitialWindowSize);
+            RequestBodyPipe = CreateRequestBodyPipe(context.ServerPeerSettings.InitialWindowSize * 10000); // TODO Once corefx is updated remove this
```
halter73: File an issue, and make the pipe Pause/ResumeWriterThreashold long.MaxValue in the meantime.

src/Servers/Kestrel/Core/src/Internal/Http2/Http2MessageBody.cs
```diff
@@ -63,14 +64,47 @@ public override void AdvanceTo(SequencePosition consumed)
 
         public override void AdvanceTo(SequencePosition consumed, SequencePosition examined)
         {
-            var dataLength = _readResult.Buffer.Slice(_readResult.Buffer.Start, consumed).Length;
+            // This code path is fairly hard to understand so let's break it down with an example
+            // ReadAsync returns a ReadResult of length 50.
+            // Advance(25, 40). The examined length would be 40 and consumed length would be 25.
+            // _totalExaminedInReadResult starts at 0. newlyExamined is 40.
+            // OnDataRead is called with length 40.
+            // _totalExaminedInReadResult is now 40 - 25 = 15.
```
Tratcher: So _totalExaminedInReadResult is the data already examined in the next read result, not this one. That's why you can subtract out consumed.

src/Servers/Kestrel/Core/src/Internal/Http2/Http2Stream.cs
jkotalik: We should get updated dependencies before EOD. Anyways, did we decide what the default will be for the thresholds?

src/Servers/Kestrel/Core/src/Internal/Http2/Http2MessageBody.cs
```diff
@@ -63,14 +64,47 @@ public override void AdvanceTo(SequencePosition consumed)
...
+
+            // The next call to ReadAsync returns 50 again
```
Tratcher: 25 old bytes and 25 new ones?

src/Servers/Kestrel/Core/src/Internal/Http2/Http2MessageBody.cs
```diff
@@ -63,14 +64,47 @@ public override void AdvanceTo(SequencePosition consumed)
...
+            // Advance(5, 5) is called
+            // newlyExamined is 5 - 15, or -10.
+            // Update _totalExaminedInReadResult to 10 as we consumed 5.
+
+            // The next call to ReadAsync returns 50 again
```
Tratcher: 45 old bytes and 5 new ones?

src/Servers/Kestrel/Core/src/Internal/Http2/Http2Stream.cs
Tratcher: The default wouldn't need to change with the new design.

src/Servers/Kestrel/Core/src/Internal/Http2/Http2MessageBody.cs
```diff
@@ -63,14 +64,47 @@ public override void AdvanceTo(SequencePosition consumed)
...
+            // _totalExaminedInReadResult is 10
+            // Advance(50, 50) is called
+            // newlyExamined = 50 - 10 = 40
+            // _totalExaminedInReadResult is now 50
+            // _totalExaminedInReadResult is finally 0 after subtracting consumedLength.
+
+            var examinedLength = _readResult.Buffer.Slice(_readResult.Buffer.Start, examined).Length;
```
halter73: Wouldn't it be way easier to keep around a `_lastExamined` SequencePosition and have the following?

```
    var newlyExamined = _readResult.Buffer.Slice(_lastExamined, consumed).Length;
```

jkotalik: Resolved offline. We can't store the last examined in a field between calls.

src/Servers/Kestrel/Core/src/Internal/Http2/Http2MessageBody.cs
jkotalik: Yep.

src/Servers/Kestrel/Core/src/Internal/Http2/Http2MessageBody.cs
jkotalik: Yep.

src/Servers/Kestrel/Core/src/Internal/Http2/Http2MessageBody.cs
jkotalik: I'll rename the variable.

src/Servers/Kestrel/Core/src/Internal/Http2/Http2MessageBody.cs
```diff
@@ -63,14 +64,47 @@ public override void AdvanceTo(SequencePosition consumed)
...
+            var consumedLength = _readResult.Buffer.Slice(_readResult.Buffer.Start, consumed).Length;
```
halter73: Special case when examined equals consumed, and don't double count the length from start to consumed.

jkotalik:
@aspnet/brt @mikaelm12 latest macOS failure.
```
  com.microsoft.signalr.HubConnectionTest > TransportAllUsesLongPollingWhenServerOnlySupportLongPolling() FAILED
      java.lang.RuntimeException at HubConnectionTest.java:1245
          Caused by: java.lang.Exception at HubConnectionTest.java:1245

```

mikaelm12:
> latest macOS failure. 

Noted. I'll file an issue and fix it. Thanks

src/Servers/Kestrel/Core/src/Internal/Http2/Http2MessageBody.cs
```diff
@@ -63,14 +64,57 @@ public override void AdvanceTo(SequencePosition consumed)
...
+            long examinedLength;
+            long consumedLength;
+            if (consumed.Equals(examined))
+            {
+                examinedLength = _readResult.Buffer.Slice(_readResult.Buffer.Start, examined).Length;
+                consumedLength = examinedLength;
+            }
+            else 
+            {
+                examinedLength = _readResult.Buffer.Slice(_readResult.Buffer.Start, examined).Length;
```
halter73: Calculate length from consumed to examined.

src/Servers/Kestrel/Core/src/Internal/Http2/Http2MessageBody.cs
```diff
@@ -14,6 +14,7 @@ public class Http2MessageBody : MessageBody
     {
         private readonly Http2Stream _context;
         private ReadResult _readResult;
+        private long _totalExaminedInPreviousReadResult;
```
Tratcher: Already examined in next read result?

jkotalik:
Test failure is with InMemory process crash again. :/. 

Besides that 🆙 📅

jkotalik:
@aspnet/brt MacOS timeout near razor component stuff.

src/Servers/Kestrel/Core/src/Internal/Http2/Http2MessageBody.cs
```diff
@@ -63,14 +64,57 @@ public override void AdvanceTo(SequencePosition consumed)
 
         public override void AdvanceTo(SequencePosition consumed, SequencePosition examined)
         {
-            var dataLength = _readResult.Buffer.Slice(_readResult.Buffer.Start, consumed).Length;
+            // This code path is fairly hard to understand so let's break it down with an example
```
davidfowl: ZOMG the performance :sad:

jkotalik: How would you fix it? :)

anurse:
Yeah, we have some thoughts on timeouts, we'll be talking about this on Monday I think.

src/Servers/Kestrel/Core/src/Internal/Http2/Http2MessageBody.cs
davidfowl: Measure first

davidfowl: OK we don’t need to gate on that.

jkotalik:
Port conflict issue with ANCM OutOfProc with music store. https://dev.azure.com/dnceng/public/_build/results?buildId=121641

jkotalik:
@halter73 @davidfowl when you get the chance, can you hook me up with an approval 😄

halter73 (approved):

PR #8220, opened 2019-03-05
EF UserStore FindByEmail will throw on dupes

HaoK:
Fixes https://github.com/aspnet/AspNetCore/issues/7716

@ajcvickers

### DIFF
--- a/src/Identity/EntityFrameworkCore/src/UserOnlyStore.cs
+++ b/src/Identity/EntityFrameworkCore/src/UserOnlyStore.cs
@@ -502,7 +502,7 @@ namespace Microsoft.AspNetCore.Identity.EntityFrameworkCore
         {
             cancellationToken.ThrowIfCancellationRequested();
             ThrowIfDisposed();
-            return Users.FirstOrDefaultAsync(u => u.NormalizedEmail == normalizedEmail, cancellationToken);
+            return Users.SingleOrDefaultAsync(u => u.NormalizedEmail == normalizedEmail, cancellationToken);
         }
 
         /// <summary>
--- a/src/Identity/EntityFrameworkCore/src/UserStore.cs
+++ b/src/Identity/EntityFrameworkCore/src/UserStore.cs
@@ -635,7 +635,7 @@ namespace Microsoft.AspNetCore.Identity.EntityFrameworkCore
         {
             cancellationToken.ThrowIfCancellationRequested();
             ThrowIfDisposed();
-            return Users.FirstOrDefaultAsync(u => u.NormalizedEmail == normalizedEmail, cancellationToken);
+            return Users.SingleOrDefaultAsync(u => u.NormalizedEmail == normalizedEmail, cancellationToken);
         }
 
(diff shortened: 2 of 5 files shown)


src/Identity/EntityFrameworkCore/src/UserOnlyStore.cs
```diff
@@ -502,7 +502,7 @@ public override Task<TUser> FindByEmailAsync(string normalizedEmail, Cancellatio
         {
             cancellationToken.ThrowIfCancellationRequested();
             ThrowIfDisposed();
-            return Users.FirstOrDefaultAsync(u => u.NormalizedEmail == normalizedEmail, cancellationToken);
+            return Users.SingleOrDefaultAsync(u => u.NormalizedEmail == normalizedEmail, cancellationToken);
```
ajcvickers: Seems a bit lazy and not the best experience to just let EF throw. What does the exception message say?

HaoK: The message is not amazing: "Sequence contains more than element"

HaoK: With the context of a callstack of FindByEmail its not too bad maybe

HaoK: Also this has been the general pattern to date for all of the other methods, they all are just FirstOrDefault or SingleOrDefault (when we expect them to be unique)

ajcvickers: Doesn't seem great to me. @blowdart. Thoughts?

HaoK: Generally the user manager does all of the validations, errors at the store level are usually something going really wrong, like you mucked with the schema, or for this bug, you have duplicate emails in your db and try to use it like a key

HaoK: To date don't wrap any exceptions that come from EF, any exceptions at the store level have been uncaught historically as they should be very uncommon (unless its a developer error)

HaoK: Note: the gap here is that typically this type of checking would be done by the UserManager, but we don't have any API to get a list of users with an email, so the manager can't do it.  The first line of defense should be the documentation update (which I'll add now), the fact that this throws now, as opposed to silently giving you the first user when there's a duplicate should be good enough to indicate that they need to take a closer look at what's going on

blowdart: I do prefer it barfing to indicate a store hiccup, rather than giving what could be the wrong user

ajcvickers: Well, if @blowdart is okay with the vague exception message, then okay...

ajcvickers (approved):

PR #8227, opened 2019-03-06
Rename Pipe properties

davidfowl:
- Rename BodyPipe to BodyReader and BodyWriter on the request and response directly

It's a much better name

cc @JamesNK @JunTaoLuo

### DIFF
--- a/src/Http/Http/src/Features/ResponseBodyPipeFeature.cs
+++ b/src/Http/Http/src/Features/ResponseBodyPipeFeature.cs
@@ -21,7 +21,7 @@ namespace Microsoft.AspNetCore.Http.Features
             _context = context;
         }
 
-        public PipeWriter ResponseBodyPipe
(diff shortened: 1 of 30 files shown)


jkotalik (approved):
Yeah, I've been thinking the same thing for a while now. BodyPipe was a mediocre name.

src/Http/Http/src/Features/ResponseBodyPipeFeature.cs
```diff
@@ -21,7 +21,7 @@ public ResponseBodyPipeFeature(HttpContext context)
             _context = context;
         }
 
-        public PipeWriter ResponseBodyPipe
+        public PipeWriter ResponseBodyWriter
```
Tratcher: `Writer` would be sufficient. ResponseBody is redundant with the feature name.

PR #8246, opened 2019-03-06
Update korebuild

ryanbrandenburg:
We need the latest Korebuild which contains a fix that should make trx files discover-able on TeamCity.


Eilon (approved):

dougbu (approved):
Doesn't this also bump up the SDK used in our builds i.e. shoudn't `global.json` change to match?

ryanbrandenburg:
Doug and I talked offline, the SDK is already up to date.

Issue #4679, opened 2016-06-01
Add details to the JwtBearer error messages


Tratcher:
JwtBearer adds error details to the response header for auth failures. Right now those errors are very generic based on exception types. We've asked IdentityModel to give us more details via exception properties.

https://github.com/aspnet/Security/pull/845#issuecomment-222839724
AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet#422
https://github.com/aspnet/Security/issues/776

PR #8259, opened 2019-03-06
Add details to the JwtBearer error messages

Tratcher:
#4679 Bearer allows you to include error details in the WwwAuthenticate response header. We didn't want to expose raw exception messages but we did want to provide some hint at what went wrong. Previously we used generic messages based on the exception type because no details were available. Now that the exceptions have some detailed properties we can include those in a well structured format.

@blowdart


Eilon (approved):
Looks good to me. And thank you for using the correct cultures!

HaoK (approved):

PR #8261, opened 2019-03-06
Consume Microsoft.Internal.Extensions.Refs

pakrym:
Use reference assemblies from Microsoft.Internal.Extensions.Refs to produce AspNetCore targeting pack.

Requires: https://github.com/aspnet/Extensions/pull/1223


anurse:
I think I need more background on how things flow here. I have a relatively good idea on what the packs are but still need some catch-up on build process. Let's chat tomorrow.

analogrelay (commented):
Looks good. Waiting for Draft to be removed before approving, but I follow it :)

eng/Versions.props
```diff
@@ -20,90 +20,91 @@
...
+    <MicrosoftExtensionsParameterDefaultValueSourcesPackageVersion>3.0.0-preview4.19161.2</MicrosoftExtensionsParameterDefaultValueSourcesPackageVersion>
+    <MicrosoftExtensionsPrimitivesPackageVersion>3.0.0-preview4.19161.2</MicrosoftExtensionsPrimitivesPackageVersion>
+    <MicrosoftExtensionsTypeNameHelperSourcesPackageVersion>3.0.0-preview4.19161.2</MicrosoftExtensionsTypeNameHelperSourcesPackageVersion>
+    <MicrosoftExtensionsValueStopwatchSourcesPackageVersion>3.0.0-preview4.19161.2</MicrosoftExtensionsValueStopwatchSourcesPackageVersion>
+    <MicrosoftExtensionsWebEncodersPackageVersion>3.0.0-preview4.19161.2</MicrosoftExtensionsWebEncodersPackageVersion>
+    <MicrosoftInternalExtensionsRefsPackageVersion>3.0.0-preview4.19161.2</MicrosoftInternalExtensionsRefsPackageVersion>
+    <MicrosoftJSInteropPackageVersion>3.0.0-preview4.19161.2</MicrosoftJSInteropPackageVersion>
+    <MonoWebAssemblyInteropPackageVersion>3.0.0-preview4.19161.2</MonoWebAssemblyInteropPackageVersion>
     <!-- Packages from aspnet/EntityFrameworkCore -->
     <dotnetefPackageVersion>3.0.0-preview4.19127.3</dotnetefPackageVersion>
```
dougbu: These packages should also be updated to make the whole set coherent.

src/Framework/ref/Microsoft.AspNetCore.App.Ref.csproj
```diff
@@ -52,12 +52,14 @@ This package is an internal implementation of the .NET Core SDK and is not meant
...
     <ReferenceImplementationAssemblies>true</ReferenceImplementationAssemblies>
+
+    <DisablePackageReferenceRestrictions>true</DisablePackageReferenceRestrictions>
   </PropertyGroup>
 
   <ItemGroup>
     <!-- Note: do not add _TransitiveExternalAspNetCoreAppReference to this list. This is intentionally not listed as a direct package reference. -->
     <Reference Include="@(AspNetCoreAppReference);@(AspNetCoreAppReferenceAndPackage);@(ExternalAspNetCoreAppReference)" />
-
+    <PackageReference Include="Microsoft.Internal.Extensions.Refs" Version="$(MicrosoftInternalExtensionsRefsPackageVersion)" IncludeAssets="Build" />
```
dougbu: Update the other build files to avoid the need for this workaround

dougbu (dismissed):

eng/Versions.props
pakrym: Why? I don't need new versions of those.

dougbu: Because EntityFrameworkCore also references Extensions, core-setup and CoreFx. With a coherent set of these dependencies, restore and test issues are far less likely.

src/Tools/dotnet-user-secrets/src/dotnet-user-secrets.csproj
```diff
@@ -18,6 +18,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <Reference Include="Newtonsoft.Json" />
```
dougbu: Why does this project need to change?

pakrym: It was getting Newtonsoft.Json transitively before.

eng/Versions.props
pakrym: That doesn't matter as long as there is no downgrade. The one that matters is core-setup-corefx coherency.

src/Tools/dotnet-user-secrets/src/dotnet-user-secrets.csproj
pakrym: https://github.com/aspnet/Extensions/pull/1028

eng/Versions.props
dougbu: Not exactly. I agree that -- if EntityFrameworkCore is already using the latest core-setup and CoreFx -- it doesn't need to change here. That's not usually the case however.

src/Tools/dotnet-user-secrets/src/dotnet-user-secrets.csproj
dougbu: 👍

eng/Versions.props
pakrym: > if EntityFrameworkCore is already using the latest core-setup and CoreFx 

Why exactly does this matter?

pakrym:
@dougbu ping.

analogrelay (approved):

dougbu:
> ping

Yup, all is well 😺

PR #8263, opened 2019-03-07
[Helix] Only run a subset of queues for PRs

HaoK:

### DIFF
--- a/eng/targets/Helix.Common.props
+++ b/eng/targets/Helix.Common.props
@@ -16,6 +16,9 @@
         <HelixAvailibleTargetQueue Include="Windows.10.Amd64.EnterpriseRS3.ASPNET.Open" Platform="Windows" EnableByDefault="false" EnableForPr="true" />
         <HelixAvailibleTargetQueue Include="OSX.1012.Amd64.Open" Platform="OSX" EnableForPr="true" />
         <HelixAvailibleTargetQueue Include="Ubuntu.1810.Amd64.Open" Platform="Linux" EnableForPr="true" />
+      
+      <!--
+      
         <HelixAvailibleTargetQueue Include="Ubuntu.1604.Amd64.Open" Platform="Linux" EnableByDefault="false" />
         <HelixAvailibleTargetQueue Include="Centos.7.Amd64.Open" Platform="Linux" EnableByDefault="false" />
         <HelixAvailibleTargetQueue Include="Debian.8.Amd64.Open" Platform="Linux" EnableByDefault="false" />
@@ -23,7 +26,7 @@
         <HelixAvailibleTargetQueue Include="Redhat.7.Amd64.Open" Platform="Linux" EnableByDefault="false" />
         <HelixAvailibleTargetQueue Include="Fedora.27.Amd64.Open" Platform="Linux" EnableByDefault="false" />
         <HelixAvailibleTargetQueue Include="Fedora.28.Amd64.Open" Platform="Linux" EnableByDefault="false" />
-
+     -->
         <!--  TODO: re-enable Debian.9.Arm64.Open and Ubuntu.1804.Arm64.Open    -->
     </ItemGroup>
 </Project>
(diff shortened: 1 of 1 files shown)


HaoK:
@pakrym @anurse do you guys know how I can distinguish between PR runs vs master runs?  I'm thinking we just need to modify the `_HelixProjectTargetQueue` condition to skip the non PR queues by default

https://github.com/aspnet/AspNetCore/blob/master/eng/targets/Helix.targets#L33

pakrym:
What's the reason to do this?

HaoK:
We want to only run a subset of queues for PRs, and leave the rest only for 'real' runs on master to speed up PR validation, so probably WIndows/OSX/Ubuntu for PRs. Everything else only on master.

I guess I can just use check `Build.SourceBranchName` for master for this

pakrym:
Aren't we losing coverage that would result in flakier master?

HaoK:
Yeah its an intentional tradeoff, but if it cuts PR validation in half, that might be compelling

pakrym:
You can use `System.TeamProject` to detect if you are running in public or internal build and set a build property in `helix.yml` (https://github.com/aspnet/AspNetCore/blob/master/.azure/pipelines/jobs/default-build.yml#L94)

> Yeah its an intentional tradeoff, but if it cuts PR validation in half, that might be compelling

Most time is spend publishing and uploading anyway, why would it cut runtime in half?

HaoK:
Publishing/uploading is done per queue, so instead of 10 we would only be doing 3, unless things have drastically changed, previously when targetting 3 queues vs 10, it would take less than 40 minutes to do a helix test run as opposed to 90 minutes now.  I can just test that assumption before adding any other logic I guess

pakrym:
> Publishing/uploading is done per queue

It's incorrect assumption. Publishing is done once per test project and upload results are cached between queues.

pakrym:
You can test all of this locally by doing:

```
.\build.cmd -projects .\src\Hosting\Hosting\test\*.csproj /t:Helix /p:IsHelixJob=true /bl
```

HaoK:
This isn't something good to test locally, it took me 2 hours on wireless just uploading last time.  Measuring real running time of PR runs needs to be done in actual PRs since they have fat pipes

pakrym:
Sure, but this PR won't change anything as both upload and publish are done only once.

HaoK:
So the run failed for some reason with an error about no helix work items to publish but the 3 queues we care about ran in ~hr instead of the usual two hours, so running time is still more or less linear based on the total number of queues which is what I saw in the past.

https://dev.azure.com/dnceng/public/_build/results?buildId=116506 50 minute helix 

vs https://dev.azure.com/dnceng/public/_build/results?buildId=116439 1:48 minutes of helix

HaoK:
I'll do a new run where I just delete the queues so the error goes away in case that was cutting the running time short

pakrym:
The run you linked didn't wait for any queue to finish, if failed and stopped right after publishing

HaoK:
Waiting for the queues only adds 10 minutes or so, so this one waited for all the tests https://dev.azure.com/dnceng/public/_build/results?buildId=116755 and it took 64 minutes, so still saves around 15 minutes, as most runs seem to be in the 1:20 range.  We'll get a few more runs to see what the average is since I need to actually add the logic now

pakrym:
I think we've confirmed that we publish and upload only once and the difference is only in waiting for Helix to find free machine and run tests. 

I severyly doubt that gaining 10 minutes (compare to the build run at the same time when queues are empty: https://dev.azure.com/dnceng/public/_build/results?buildId=116891) by disabling 2/3 of coverage is a worthy tradeoff. 

It would also make life harder for @aspnet/brt because instead of catching failures in PRs they would need to investigate them in master without clear connection to the source of failure.

There still are things we can improve to speedup helix runs:

1. Optimize publishing - right now we have parallel publishing disabled which most defenitely affects publish time
2. Optimize upload time - currently all payloads start archiving and uploading at the same time after all test project publishes finish creating CPU and network contention. It would help if we can start uploading payload immidiately after publish was done for the project (this is only worth it if we get parallel publish). 
3. Take a look at a binlog and see what else takes time

HaoK:
Going to just disable helix on PRs in the medium term until its less flaky and consistently green

PR #8290, opened 2019-03-07
Use ApplicationPartFactory when adding application parts

pranavkm:
Fixes https://github.com/aspnet/AspNetCore/issues/8288

### DIFF
--- a/src/Middleware/CORS/test/FunctionalTests/package.json
+++ b/src/Middleware/CORS/test/FunctionalTests/package.json
@@ -3,7 +3,7 @@
   "devDependencies": {
     "jest": "^23.6.0",
     "merge": "^1.2.1",
-    "puppeteer": "^1.12.2"
+    "puppeteer": "^1.13.0"
   },
   "dependencies": {},
   "scripts": {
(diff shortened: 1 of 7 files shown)


src/Middleware/CORS/test/FunctionalTests/package.json
```diff
@@ -3,7 +3,7 @@
   "devDependencies": {
     "jest": "^23.6.0",
     "merge": "^1.2.1",
-    "puppeteer": "^1.12.2"
+    "puppeteer": "^1.13.0"
```
javiercn: Undo?

javiercn:
@pranavkm When did we break this? 
Do we need to patch this?

NTaylorMullen (approved):

pranavkm:
> Do we need to patch this?

It's not broken, just doesn't do the right thing all the time. It certainly wouldn't qualify for a patch bug given we have had no user reports and there's a trivial workaround to address it.

javiercn (approved):

pranavkm:
Test failure investigation is assigned to me (https://github.com/aspnet/AspNetCore-Internal/issues/1913). I'll skip it in a bit.

src/Middleware/CORS/test/FunctionalTests/package.json
pranavkm: The build keeps updating this file. I've decided to stop fighting it at this point

Issue #6526, opened 2019-01-09
React to SIGTERM exit code change


Tratcher:
https://github.com/dotnet/coreclr/pull/21300#issuecomment-452684375

CoreClr is changing the default behavior for SIGTERM exit codes from zero to non-zero. WebHost and GenericHost both block the ProcessExit event while attempting to shut down.

Proposal:
If the host does shut down gracefully then set the exit code to zero.

tmds:
This change is in coreclr for 3.0. So maybe this should be 3.0 milestone?

Tratcher:
Note for generic host: We're going to have to port some of the Microsoft.AspNetCore.Server.IntegrationTesting infrastructure to Extensions to enable proper functional tests.

PR #8294, opened 2019-03-07
Handle SIGTERM exit code

Tratcher:
#6526 There was a corefx change in 3.0 where on Linux if the shutdown is triggered by SIGTERM then that's signaled with the 143 exit code. Suppress that if we shut down gracefully.

Generic host will need a similar treatment.

@tmds

### DIFF
--- a/src/Hosting/Hosting/src/Internal/WebHostLifetime.cs
+++ b/src/Hosting/Hosting/src/Internal/WebHostLifetime.cs
@@ -13,6 +13,7 @@ namespace Microsoft.AspNetCore.Hosting.Internal
         private readonly string _shutdownMessage;
 
         private bool _disposed = false;
+        private bool _exitedGracefully = false;
 
         public WebHostLifetime(CancellationTokenSource cts, ManualResetEventSlim resetEvent, string shutdownMessage)
         {
@@ -22,6 +23,11 @@ namespace Microsoft.AspNetCore.Hosting.Internal
 
             AppDomain.CurrentDomain.ProcessExit += ProcessExit;
             Console.CancelKeyPress += CancelKeyPress;
+        }
+
+        internal void SetExitedGracefully()
+        {
+            _exitedGracefully = true;
         }
 
         public void Dispose()
@@ -46,6 +52,12 @@ namespace Microsoft.AspNetCore.Hosting.Internal
         private void ProcessExit(object sender, EventArgs eventArgs)
         {
             Shutdown();
+            if (_exitedGracefully)
+            {
+                // On Linux if the shutdown is triggered by SIGTERM then that's signaled with the 143 exit code.
+                // Suppress that since we shut down gracefully. https://github.com/aspnet/AspNetCore/issues/6526
+                Environment.ExitCode = 0;
+            }
         }
 
         private void Shutdown()
--- a/src/Hosting/test/FunctionalTests/ShutdownTests.cs
+++ b/src/Hosting/test/FunctionalTests/ShutdownTests.cs
@@ -28,7 +28,6 @@ namespace Microsoft.AspNetCore.Hosting.FunctionalTests
         [ConditionalFact]
         [OSSkipCondition(OperatingSystems.Windows)]
         [OSSkipCondition(OperatingSystems.MacOSX)]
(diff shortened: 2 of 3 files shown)


tmds:
@Tratcher thank you for addressing this issue.

I hinted about an alternative in related issues.

You could add an `ExitCode` property on the `Host`, and a method that returns this exit code, which is called from Main.

When the Host runs with 'console lifetime', you unconditionally assign `Environment.ExitCode = host.ExitCode` on `ProcessExit` event. When the host doesn't run with console lifetime, you don't touch `Environment.ExitCode`.

This `host.ExitCode` could also be used to set the `ServiceBase.ExitCode` when the Host is used to build a Windows service.

For you to consider. This change on its own LGTM.

CC @davidfowl @halter73

JunTaoLuo (approved):

src/Hosting/Hosting/src/Internal/WebHostLifetime.cs
```diff
@@ -46,6 +52,12 @@ private void CancelKeyPress(object sender, ConsoleCancelEventArgs eventArgs)
         private void ProcessExit(object sender, EventArgs eventArgs)
         {
             Shutdown();
+            if (_exitedGracefully)
+            {
+                // On Linux if the shutdown is triggered by SIGTERM then that's signaled with the 143 exit code.
+                // Suppress that since we shut down gracefully. https://github.com/aspnet/AspNetCore/issues/6526
+                Environment.ExitCode = 0;
```
halter73: How does this interact with a Program.Main() that returns a non-zero exit code?

src/Hosting/test/FunctionalTests/ShutdownTests.cs
```diff
@@ -28,7 +28,6 @@ public class ShutdownTests : LoggedTest
         [ConditionalFact]
         [OSSkipCondition(OperatingSystems.Windows)]
         [OSSkipCondition(OperatingSystems.MacOSX)]
-        [OSSkipCondition(OperatingSystems.Linux)] // https://github.com/aspnet/AspNetCore-Internal/issues/1687
```
halter73: Does this test the exit code?

Tratcher: Yes, we'd skipped it due to the exit code failure.

src/Hosting/Hosting/src/Internal/WebHostLifetime.cs
Tratcher: When the process is being terminated externally then Main isn't guaranteed to finish. There's a race.

Tratcher:
Host is not 1:1 with process, so adding a top level exit code wouldn't make sense. I wonder if there's another way to flow the exit code to console lifetime.

halter73 (approved):

tmds:
> Host is not 1:1 with process, so adding a top level exit code wouldn't make sense. I wonder if there's another way to flow the exit code to console lifetime.

For me, Console lifetime means the host is 1:1 with the process. So setting `Environment.ExitCode` based on the host exit code would be meaningful.

As an example: `System.CommandLine`'s `InvocationContext` has an `int ResultCode`. In `CancelOnProcessTermination` this is assigned to `Environment.ExitCode`: https://github.com/dotnet/command-line-api/blob/4c2c852af940d17e5b07b5291e127bdb56cc02f2/src/System.CommandLine/Invocation/InvocationExtensions.cs#L48.

src/Hosting/Hosting/src/Internal/WebHostLifetime.cs
tmds: When the process isn't terminated externally, this sets the final exit code.
```cs
static int Main(string[] args)
{
    AppDomain.CurrentDomain.ProcessExit += delegate { Environment.ExitCode = 10; };
    return 5;
}
```
has an exit code of `10`.

Tratcher:
@tmds Back up a second to how a host.ExitCode would be determined. Today host startup and shutdown errors result in exceptions being thrown as is typical for .NET classes. Overriding the sigterm case makes sense, but that result would always be 0. When and how would you end up with the host specifying a non-zero exit code?

Issue #8302, opened 2019-03-07
NewtonsoftJsonOutputFormatter fails on sync flushes


Tratcher:
https://github.com/aspnet/AspNetCore/issues/7644#issuecomment-470690541
```
System.InvalidOperationException: Synchronous operations are disallowed. Call WriteAsync or set AllowSynchronousIO to true instead.&#xD;&#xA;   
at Microsoft.AspNetCore.Server.IIS.Core.HttpResponseStream.Write(Byte[] buffer, Int32 offset, Int32 count)&#xD;&#xA;   
at Microsoft.AspNetCore.Server.IIS.Core.WrappingStream.Write(Byte[] buffer, Int32 offset, Int32 count)&#xD;&#xA;   
at Microsoft.AspNetCore.WebUtilities.HttpResponseStreamWriter.FlushInternal(Boolean flushEncoder)&#xD;&#xA;   
at Microsoft.AspNetCore.WebUtilities.HttpResponseStreamWriter.Write(String value)&#xD;&#xA;   
at Newtonsoft.Json.JsonTextWriter.WriteValue(Guid value)&#xD;&#xA;   
at Newtonsoft.Json.JsonWriter.WriteValue(Nullable`1 value)&#xD;&#xA;   
at Newtonsoft.Json.JsonWriter.WriteValue(JsonWriter writer, PrimitiveTypeCode typeCode, Object value)&#xD;&#xA;   
at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.SerializeObject(JsonWriter writer, Object value, JsonObjectContract contract, JsonProperty member, JsonContainerContract collectionContract, JsonProperty containerProperty)&#xD;&#xA;   
at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.SerializeList(JsonWriter writer, IEnumerable values, JsonArrayContract contract, JsonProperty member, JsonContainerContract collectionContract, JsonProperty containerProperty)&#xD;&#xA;   
at Newtonsoft.Json.Serialization.JsonSerializerInternalWriter.Serialize(JsonWriter jsonWriter, Object value, Type objectType)&#xD;&#xA;   
at Newtonsoft.Json.JsonSerializer.SerializeInternal(JsonWriter jsonWriter, Object value, Type objectType)&#xD;&#xA;   
at Newtonsoft.Json.JsonSerializer.Serialize(JsonWriter jsonWriter, Object value)&#xD;&#xA;   
[...]

PR #8303, opened 2019-03-07
Allow sync IO for NewtonsoftJsonOutputFormatter

Tratcher:
#8302 This was missed before because there were no tests that wrote a large enough body to trigger a sync flush.

### DIFF
--- a/src/Mvc/test/Mvc.FunctionalTests/JsonOutputFormatterTests.cs
+++ b/src/Mvc/test/Mvc.FunctionalTests/JsonOutputFormatterTests.cs
@@ -32,7 +32,7 @@ namespace Microsoft.AspNetCore.Mvc.FunctionalTests
             {
                 Id = 1,
                 Alias = "john",
-                description = "Administrator",
+                description = "This is long so we can test large objects " + new string('a', 1024 * 65),
                 Designation = "Administrator",
                 Name = "John Williams"
             };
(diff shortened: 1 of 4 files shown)


Tratcher:
@rynowak I'm going dark soon. If I don't get this in today can you take over?

src/Mvc/test/Mvc.FunctionalTests/JsonOutputFormatterTests.cs
```diff
@@ -32,7 +32,7 @@ public async Task JsonOutputFormatter_ReturnsIndentedJson()
             {
                 Id = 1,
                 Alias = "john",
-                description = "Administrator",
+                description = "This is long so we can test large objects " + new string('a', 1024 * 65),
```
rynowak: <3

rynowak (approved):

rynowak:
Looks like you have some nulls in unit tests. I suspect it's some **ancient** tests that mock the http context instead of newing it up.

Tratcher:
Ack.

rynowak:
> @rynowak I'm going dark soon. If I don't get this in today can you take over?

sure

ikirkpat:
We are using dotnet 3.1 and seem to be having crashes at serialization time. I'm investigating whether or not [8302](https://github.com/dotnet/aspnetcore/issues/8302) is the reason for this. It seems that this PR implies that async IO is not working properly for large responses so this PR just made a workaround to let you serialize syncronously. Is that correct? Has this ever fixed so we can do async IO for JSON? If so, what version of dotnet has the fix?

PR #8309, opened 2019-03-08
Testing broken pipelines on ci no. 2

halter73:
Is [this check](https://github.com/dotnet/corefx/blob/1382754add40c899c141803c1cdb03c335d60095/src/System.IO.Pipelines/src/System/IO/Pipelines/Pipe.cs#L492) enough?

If examined and consumed are canonically the same, but examined points to the BufferSegment before the one pointed to by consumed, the examined BufferSegment would get returned. Considering _bufferSegmentPool is used like a stack, GetLength could quickly be called with a _lastExamined segment that's been reused and been given a new RunningIndex.

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -305,7 +305,7 @@
       <Uri>https://github.com/dotnet/corefx</Uri>
       <Sha>00000</Sha>
     </Dependency>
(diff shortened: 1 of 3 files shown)


jkotalik:
We found it1

PR #8311, opened 2019-03-08
Unpin our System.Net.Http package version

dougbu:
- shipped in the past patch


ryanbrandenburg (approved):
Seems reasonable, what's the history behind why it was pinned?

dougbu:
@ryanbrandenburg we generally pin anything that didn't ship in the previous patch. In this case, the dependency was really old but didn't conflict until 2.1.9.

PR #8313, opened 2019-03-08
Add AddControllers() and AddRazorPages()

rynowak:
These are *new style* for configuring MVC in services. We're adding these to make
things feel a bit more tailored to those particular scenarios.

----

The main reason for this is that we've had repeated community asks for an
*API-optimized* way of configuring MVC. I don't think that using AddMvcCore
is a suitable building block, because it has too many options that you want.

I've think I've identified the reasonable set of features that should be part
of the default experience for APIs. All of these things are already pay-for-play
and are activated by the presence of attributes. The only additional cost is
loading of assemblies and a few additional inspections of the attributes
(cached).

----

Given that we're creating AddControllers, it makes sense that we'd flesh out the
rest of the space that folks are saying *but what about me???*.

**Edit:** (and then there were three)

You can either:
- `AddControllers`
- `AddControllersWithViews`
- `AddRazorPages`

These are all top-level service collection methods, that return the existing `IMvcBuilder`. You can use multiple of these. If you end up using `AddControllers` and `AddRazorPages` you will end up with what `AddMvc` does today.

If you do multiple things to multiple different MVC builder instances, then they run in the order you specified - like what would happen if you use `AddMvc` multiple times in today's landscape. You can't have separate model binders for controllers and pages, or things like that.

*Small note - `AddRazorPages` includes the non-API-focused support for controllers. We'd have to do significant work to take this out. Will do it in the future based on user feedback.*

----

Branding and perception are important, and we've had users ask for more flexibility
in what shows up in their app. The plan is to update the templates to use this experience
in preview4, and see what kind of feedback we get.


src/Mvc/Mvc.Core/src/Builder/ControllerEndpointRouteBuilderExtensions.cs
```diff
@@ -106,7 +106,7 @@ public static IEndpointConventionBuilder MapDefaultControllerRoute(this IEndpoin
 
         /// <summary>
         /// Adds endpoints for controller actions to the <see cref="IEndpointRouteBuilder"/> and specifies a route
-        /// with the given <paramref name="name"/>, <paramref name="areaName"/>, <paramref name="template"/>,
```
rynowak: VS is doing a format.

src/Mvc/Mvc.ViewFeatures/src/ViewComponentResult.cs
```diff
@@ -64,7 +64,15 @@ public override Task ExecuteResultAsync(ActionContext context)
             }
 
             var services = context.HttpContext.RequestServices;
-            var executor = services.GetRequiredService<IActionResultExecutor<ViewComponentResult>>();
+            var executor = services.GetService<IActionResultExecutor<ViewComponentResult>>();
```
rynowak: So one of the things that I added is some guardrails around things like returning `View()` from a controller without the services. I figure this is is the most likely way for someone to experience this particular problem.

I couldn't think of other things that need to be fixed, there aren't many APIs that the user calls that wouldn't work, it's mainly the view result stuff.

Additionally, Identity UI already registers what it needs in services, so we're good there.

src/Mvc/Mvc/src/MvcServiceCollectionExtensions.cs
```diff
@@ -30,69 +31,156 @@ public static IMvcBuilder AddMvc(this IServiceCollection services)
...
 
-            builder.AddCors();
+            // This method excludes all of the view-related services by default.
+            var builder = services
+                .AddMvcCore()
+                .AddApiExplorer()
+                .AddAuthorization()
+                .AddCors()
+                .AddDataAnnotations()
+                .AddFormatterMappings();
```
rynowak: These are all of the things that you'll get by default. 
- API explorer doesn't do anything unless you call its APIs
- Auth is based on attributes
- Cors is based on attributes
- Data annotations is based on attributes
- Format filter is based on attributes

All of them are pay for play. 

I debated leaving out format filter because no one uses it. But I think it's hard to explain why we left it out that it is to just put it in. I still find it super duper vexxxing that format filter requires a gesture.

src/Mvc/Mvc.Core/src/Builder/ControllerEndpointRouteBuilderExtensions.cs
ryanbrandenburg: For anyone who doesn't already know: adding `?w=1` to the end of the url hides whitespace only diff.

ryanbrandenburg (approved):
Makes sense from a templating perspective.

src/Mvc/Mvc/src/MvcServiceCollectionExtensions.cs
```diff
@@ -30,69 +31,156 @@ public static IMvcBuilder AddMvc(this IServiceCollection services)
...
+            var builder = services.AddControllers().AddRazorPages();
+            if (configure != null)
+            {
+                builder.AddRazorPagesOptions(configure);
+            }
 
             return builder;
         }
 
+        internal static void AddDefaultFrameworkParts(ApplicationPartManager partManager)
```
pranavkm: Unrelated, but do you know what this is for?

src/Mvc/Mvc.ViewFeatures/src/ViewComponentResult.cs
pranavkm: We could always add more if we get user reports for `GetRequiredService` failing.

pranavkm (approved):
Do you have a follow up item to fix up the templates?

src/Mvc/Mvc/src/MvcServiceCollectionExtensions.cs
rynowak: I'm not sure. I think it's up to them to decide if this is what they wanted all along.

src/Mvc/Mvc/src/MvcServiceCollectionExtensions.cs
rynowak: Adding default framework parts?

pranavkm: Yeah, but why do we add these? These aren't sources for controllers.

rynowak: I cleaned this up by renaming this method and making sure it's only included by pages/views.

rynowak:
@ryanbrandenburg @pranavkm - can you please take a second look at this? I changed how this works based on feedback from @DamianEdwards  and @davidfowl and updated the PR description

pranavkm:
Code check failed

rynowak:
> Code check failed

That is probably true.

src/ProjectTemplates/Web.ProjectTemplates/content/StarterWeb-FSharp/Startup.fs
```diff
@@ -22,7 +22,8 @@ type Startup private () =
     // This method gets called by the runtime. Use this method to add services to the container.
     member this.ConfigureServices(services: IServiceCollection) =
         // Add framework services.
-        services.AddMvc().AddNewtonsoftJson().AddRazorRuntimeCompilation() |> ignore
+        services.AddControllersWithViews().AddNewtonsoftJson().AddRazorRuntimeCompilation() |> ignore
```
pranavkm: Wouldn't this be `AddRazorPages()`?

rynowak: Starterweb for F# uses both

ryanbrandenburg (approved):
Looks good as far as I can tell.

rynowak:
@pranavkm any more feedback?

pranavkm:
:shipit:

PR #8315, opened 2019-03-08
Returning rejected promises in HttpConnection

mikaelm12:
Fixes: https://github.com/aspnet/AspNetCore/issues/7465
We were throwing in places we should have been returned rejected promises.

### DIFF
--- a/src/SignalR/clients/ts/signalr/src/HttpConnection.ts
+++ b/src/SignalR/clients/ts/signalr/src/HttpConnection.ts
@@ -299,7 +299,7 @@ export class HttpConnection implements IConnection {
     }
 
     private constructTransport(transport: HttpTransportType) {
-    switch (transport) {
+        switch (transport) {
             case HttpTransportType.WebSockets:
                 if (!this.options.WebSocket) {
                     throw new Error("'WebSocket' is not supported in your environment.");
--- a/src/SignalR/clients/ts/signalr/src/HttpConnection.ts
+++ b/src/SignalR/clients/ts/signalr/src/HttpConnection.ts
@@ -115,7 +115,7 @@ export class HttpConnection implements IConnection {
 
     public send(data: string | ArrayBuffer): Promise<void> {
         if (this.connectionState !== ConnectionState.Connected) {
-            throw new Error("Cannot send data if the connection is not in the 'Connected' State.");
+            return Promise.reject(new Error("Cannot send data if the connection is not in the 'Connected' State."));
         }
 
         // Transport will not be null if state is connected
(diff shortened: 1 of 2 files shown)


src/SignalR/clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -293,13 +293,13 @@ export class HttpConnection implements IConnection {
...
-            throw new Error(`Unable to connect to the server with any of the available transports. ${transportExceptions.join(" ")}`);
+            return Promise.reject(new Error(`Unable to connect to the server with any of the available transports. ${transportExceptions.join(" ")}`));
         }
-        throw new Error("None of the transports supported by the client are supported by the server.");
+        return Promise.reject(new Error("None of the transports supported by the client are supported by the server."));
     }
 
     private constructTransport(transport: HttpTransportType) {
-        switch (transport) {
+    switch (transport) {
```
BrennanConroy: revert

src/SignalR/clients/ts/signalr/tests/HttpConnection.test.ts
```diff
@@ -150,6 +150,34 @@ describe("HttpConnection", () => {
...
+                logger,
+            } as IHttpConnectionOptions;
+
+            const connection = new HttpConnection("http://tempuri.org", options);
+            await expect(connection.start(TransferFormat.Text))
+                .rejects
+                .toThrow("Unexpected status code returned from negotiate 999");
+        },
+        "Failed to start the connection: Error: Unexpected status code returned from negotiate 999");
+    });
```
BrennanConroy: We should have a test that uses `.then` and .`catch` to verify that we no longer throw synchronously.

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -293,13 +293,13 @@ export class HttpConnection implements IConnection {
...
             case HttpTransportType.WebSockets:
                 if (!this.options.WebSocket) {
                     throw new Error("'WebSocket' is not supported in your environment.");
```
BrennanConroy: What about this throw?

BrennanConroy: And the ones in `resolveTransport` and `resolveUrl`?

BrennanConroy (approved):
Just a few things you should look at.

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
mikaelm12: Errors in constructTransport and resolveTransport are caught and aggregated into a rejected Promise. `resolveUrl` is only used in the constructor so that should be fine just throwing an error

src/SignalR/clients/ts/signalr/tests/HttpConnection.test.ts
```diff
@@ -150,6 +150,34 @@ describe("HttpConnection", () => {
         });
     });
 
+    it("cannot send with an un-started connection", async () => {
+        await VerifyLogger.run(async (logger) => {
+            const connection = new HttpConnection("http://tempuri.org");
+
+            await expect(connection.send("LeBron James"))
+                .rejects
+                .toThrow("Cannot send data if the connection is not in the 'Connected' State.");
```
halter73: Can you add a test for one of the thrown exceptions that get aggregated into a rejected promise?

halter73 (approved):

PR #8318, opened 2019-03-08
Chunked request parsing does not properly update examined

davidfowl:
Chunked request parsing does not properly update examined
- The chunked parsing logic didn't properly update the examined position when parsing the chunked prefix. This started to throw because Pipe now throws if examined is set to the position before the previous.

### DIFF
--- a/src/Servers/Kestrel/Core/src/Internal/Http/Http1ChunkedEncodingMessageBody.cs
+++ b/src/Servers/Kestrel/Core/src/Internal/Http/Http1ChunkedEncodingMessageBody.cs
@@ -323,6 +323,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                 return;
             }
 
+            // Assigned this before calculating the chunk size since that can throw
+            examined = reader.Position;
+
             var chunkSize = CalculateChunkSize(ch1, 0);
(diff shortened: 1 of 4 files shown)


src/Servers/Kestrel/Core/src/Internal/Http/Http1ChunkedEncodingMessageBody.cs
```diff
@@ -323,6 +323,9 @@ private void ParseChunkedPrefix(ReadOnlySequence<byte> buffer, out SequencePosit
                 return;
             }
 
+            // Assigned this before calculating the chunk size since that can throw
```
pakrym: `try-finally` ?

src/Servers/Kestrel/Core/src/Internal/Http/Http1ChunkedEncodingMessageBody.cs
```diff
@@ -360,6 +363,9 @@ private void ParseChunkedPrefix(ReadOnlySequence<byte> buffer, out SequencePosit
                 ch1 = ch2;
             }
 
+            // Set examined so that we capture the progress that way made
```
Tratcher: We

jkotalik (approved):

PR #8326, opened 2019-03-08
Enable the https dev cert for use with Docker for Windows's default networking setup

DavidSimner:
In Docker for Windows, [Docker recommends that, for development purposes, containers can connect to a service on the host using the DNS name `host.docker.internal`](https://docs.docker.com/docker-for-windows/networking/#i-want-to-connect-from-a-container-to-a-service-on-the-host).

This PR changes how the ASP.NET Core https dev cert is generated so that it is additionally valid for that DNS name.

### DIFF
--- a/src/Shared/CertificateGeneration/CertificateManager.cs
+++ b/src/Shared/CertificateGeneration/CertificateManager.cs
@@ -23,6 +23,7 @@ namespace Microsoft.AspNetCore.Certificates.Generation
         private const string ServerAuthenticationEnhancedKeyUsageOidFriendlyName = "Server Authentication";
 
         private const string LocalhostHttpsDnsName = "localhost";
+        private const string DockerLocalhostHttpsDnsName = "host.docker.internal";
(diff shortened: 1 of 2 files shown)


DavidSimner:
@Eilon: the 14 checks are all successful, can you please assign someone to do a code review?

Eilon:
@DavidSimner sorry for the delay.

@javiercn - can you have a look?

@mkArtakMSFT - any thoughts on this?

javiercn:
@Eilon We have an ongoing discussion about it.

anurse:
@javiercn did this conversation go anywhere? I think this PR is probably pretty stale now unfortunately.

javiercn:
@anurse I'll try to find the conversation, but I don't think we want to add additional DNS names to the cert.

anurse:
Ok, given that I'll close this. If there is still interest in making changes to our dev certs to support scenarios like this, please open a new issue and we can discuss the options.

PR #8331, opened 2019-03-08
Added possibility of specifying port when launching with UseAngularCliServer

jepperaskdk:
Launching the AngularCliServer finds an available port by default. This pull-request allows to pass the port on the Microsoft.AspNetCore.SpaServices.SpaOptions.

Today it is possible to override the port number, using a hack like in this answer: [How to set a default port for ASP.NET Core Angular app](https://stackoverflow.com/a/50980371/3717691), but this PR makes it more simple:

```
app.UseSpa(spa =>
{
    spa.Options.SourcePath = "ClientApp";
    spa.Options.PortNumber = 40000;

    if (env.IsDevelopment())
    {
        spa.UseAngularCliServer(npmScript: "start");
    }
});
```


src/Middleware/SpaServices.Extensions/src/AngularCli/AngularCliMiddleware.cs
```diff
@@ -62,9 +62,17 @@ internal static class AngularCliMiddleware
...
 
         private static async Task<AngularCliServerInfo> StartAngularCliServerAsync(
-            string sourcePath, string npmScriptName, ILogger logger)
+            string sourcePath, string npmScriptName, int portNumber, ILogger logger)
         {
-            var portNumber = TcpPortFinder.FindAvailablePort();
+            if (portNumber == 0) {
+                portNumber = TcpPortFinder.FindAvailablePort();
+            } else {
+                if (!TcpPortFinder.PortAvailable(portNumber)) {
```
SteveSandersonMS: If the developer specified a port explicitly but that port is not available, I think it would be better just to throw and fail rather than trying to pick another port.

This is because, if the developer really wants it to be on a specific port, they probably have a reason for this and something else in their infrastructure will depend on it actually being on that port.

So I recommend removing `PortAvailable` altogether and letting it just fail if the specified port can't be bound.

SteveSandersonMS:
The implementation here is perfectly good (caveat: I posted one change suggestion), and so in that sense, I would approve this PR.

However there's also a product decision to be made about what kinds of changes we're taking in this area, and whether we want to ensure consistency in the capabilities across all related APIs. For example, is it important that this new "specify port" feature should also be added to the React dev server middleware for consistency? cc @danroth27 for PM input on strategy.

danroth27:
I'm fine with merging this and then opening an up-for-grabs issue to track adding the functionality for React.

src/Middleware/SpaServices.Extensions/src/AngularCli/AngularCliMiddleware.cs
jepperaskdk: Yes, I agree. I've removed that part again. Let me know if there's more I can do. :-)

SteveSandersonMS (approved):
Looks good!

jepperaskdk:
> I'm fine with merging this and then opening an up-for-grabs issue to track adding the functionality for React.

It seems that the same code can be applied to [ReactDevelopmentServerMiddleware.StartCreateReactAppServerAsync](https://github.com/aspnet/AspNetCore/blob/master/src/Middleware/SpaServices.Extensions/src/ReactDevelopmentServer/ReactDevelopmentServerMiddleware.cs#L63).

agtokty:
hi, which release will contain this commit ?

pranavkm:
@jepperaskdk could you resolve the merge conflicts and update ref assemblies: https://github.com/aspnet/AspNetCore/blob/master/docs/ReferenceAssemblies.md

jepperaskdk:
> @jepperaskdk could you resolve the merge conflicts and update ref assemblies: https://github.com/aspnet/AspNetCore/blob/master/docs/ReferenceAssemblies.md

Hi, yes I tried, however the command failed with: `TargetFramework: not found` - and the `src/Middleware/SpaServices.Extensions/ref/Microsoft.AspNetCore.SpaServices.Extensions.csproj` does not look right. I should note that I'm a completely new to contributing - I'll have a look at the documentation again.

stephtr:
> ```
> app.UseSpa(spa =>
> {
>     spa.Options.SourcePath = "ClientApp";
>     spa.Options.PortNumber = 40000;
> 
>     if (env.IsDevelopment())
>     {
>         spa.UseAngularCliServer(npmScript: "start");
>     }
> });
> ```

Why should the `PortNumber` be an option and not an argument? To me it feels related more closely to the `npmScript` than to the `SourcePath`.

> If the developer specified a port explicitly but that port is not available, I think it would be better just to throw and fail rather than trying to pick another port.

I opened a similar PR already two years ago (most recent version was #12489). For my use case it would be way more useful not to start the `npmScript` at all if the port isn't available. As `npmScript` I run webpack, which takes quite some time for startup but does then incremental compilation on changes. I didn't want webpack to be restarted when for example `dotnet watch` restarted the web server. I therefore modified the extension such that it starts the `npmScript` if the port is available (webpack hadn't been started) and does nothing if the port is already in use (since that means that I already manually started the script to run independent of the web server).

mkArtakMSFT:
We're closing this PR as we've seen no updates on this for some time now. @jepperaskdk if you would like to move forward with this PR, please reopen and rebase.

mrjamiebowman:
Yeah, this is a real issue for me! I need to specify a port for setting up an OIDC client. I'm surprised the code doesn't allow specifying a port to begin with.

punkouter2018:
So how do I start on a different port?

PR #8334, opened 2019-03-08
Resolve deadlock with `InvokeAsync` in `On` handler

mikaelm12:
Issue: https://github.com/aspnet/AspNetCore/issues/5356

The code 
```csharp
hubConnection.On("Send" (message) => 
{
    await hubConnection.InvokeAsync<object>("Update")
})
```
Would deadlock because we wait on the the completion of each of the `on` handlers.

~~One thing to consider now is that the order in which the handlers complete is not guaranteed since we aren't awaiting them.~~


src/SignalR/clients/csharp/Client/test/FunctionalTests/HubConnectionTests.cs
```diff
@@ -263,7 +303,10 @@ public async Task CanInvokeClientMethodFromServer(string protocolName, HttpTrans
                     await connection.StartAsync().OrTimeout();
 
                     var tcs = new TaskCompletionSource<string>();
-                    connection.On<string>("Echo", tcs.SetResult);
+                    connection.On<string>("Echo", (message) =>
```
BrennanConroy: Why?

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -735,7 +735,7 @@ private async Task DispatchInvocationAsync(InvocationMessage invocation)
             {
                 try
                 {
-                    await handler.InvokeAsync(invocation.Arguments);
+                    handler.InvokeAsync(invocation.Arguments);
```
BrennanConroy: We should be awaiting this always

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -676,7 +676,7 @@ private async Task<(bool close, Exception exception)> ProcessMessagesAsync(HubMe
                     break;
                 case InvocationMessage invocation:
                     Log.ReceivedInvocation(_logger, invocation.InvocationId, invocation.Target, invocation.Arguments);
-                    await DispatchInvocationAsync(invocation);
+                    DispatchInvocationAsync(invocation);
```
BrennanConroy: Keep the method async and do `_ =`

BrennanConroy (commented):
I'm not sure about the out of order possibility now, I'll have to think about it.

src/SignalR/clients/csharp/Client/test/FunctionalTests/HubConnectionTests.cs
mikaelm12: ignore that

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -676,7 +676,7 @@ private async Task<(bool close, Exception exception)> ProcessMessagesAsync(HubMe
                     break;
                 case InvocationMessage invocation:
                     Log.ReceivedInvocation(_logger, invocation.InvocationId, invocation.Target, invocation.Arguments);
-                    await DispatchInvocationAsync(invocation);
+                    _ = DispatchInvocationAsync(invocation);
```
pranavkm: Might help to add a comment to say why it's not being awaited.

halter73:
Do we want to guarantee the On callbacks are at least started in the right order even if their async completion?

mikaelm12:
The comment I made earlier about the ordering guarantee is no longer relevant I believe.  The handlers should execute in the order they were registered in the pr now.

halter73:
I think there's still an issue with ordering if there are multiple callbacks for the same message, since you await each callback before calling the next. This won't cause a deadlock like before because it's not blocking the message processing loop, but it can still lead to callbacks getting invoked in a different order than the messages were received in.

If there are ever multiple callbacks for the same message, I think it would be better to not await the callbacks in the loop, but store the tasks in an array and await them all at the end of the loop.

halter73:
Another question I have is why don't we just dispatch on handler callbacks to the threadpool?

mikaelm12:
The issue with the previous state of this PR was that while then multiple handlers for an invocation would be processed in the order in which they were registered, the order in which incoming messages would be processed was not guaranteed. I've re-introduced our TaskQueue type to fix the ordering.

src/SignalR/clients/csharp/Client.Core/src/TaskQueue.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            // Lock the object for as short amount of time as possible
+            lock (_lockObj)
+            {
+                if (_drained)
+                {
+                    return _lastQueuedTask;
+                }
+
+                var newTask = _lastQueuedTask.ContinueWith((t, s1) => taskFunc(s1), state).Unwrap();
```
davidfowl: We need to re-write this to be async await.

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -365,6 +367,7 @@ private async Task StopAsyncCore(bool disposing)
 
                 CheckDisposed();
                 connectionState = _connectionState;
+                await taskQueue.Drain();
```
halter73: Shouldn't we change states before we start draining? It also doesn't make sense to store away the local `connectionState` before a potentially long await.

Also, we should at the very least log when draining takes a very long time.

halter73 (changes requested):

src/SignalR/clients/csharp/Client.Core/src/TaskQueue.cs
halter73: Just make sure that we continue to dispatch to the threadpool after we change to async await. Without any special parameters, Task.CompletedTask.ContinueWith dispatches. That's a good thing because it stops blocking .On() callbacks from blocking the message loop, not just callbacks that return incomplete tasks but don't block.

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -676,7 +679,8 @@ private async Task<(bool close, Exception exception)> ProcessMessagesAsync(HubMe
                     break;
                 case InvocationMessage invocation:
                     Log.ReceivedInvocation(_logger, invocation.InvocationId, invocation.Target, invocation.Arguments);
-                    await DispatchInvocationAsync(invocation);
+                    // No need to wait on the On handlers logic. Would prevent us from awaiting other invocations.
+                    _ = taskQueue.Enqueue((state) => DispatchInvocationAsync((InvocationMessage)state), invocation);
```
davidfowl: No closures....

mikaelm12: Going with a different approach now.  Going to Use a `Channel<Func<object, task>>` as the underlying queue and instead directly queue the tasks from `handler.InvokeAsync`  rather than the task from `DispatchInvocationAsync`

davidfowl: Just use a Channel<InvocationMessage>

mikaelm12: Yeah, I think that simplifies this a lot. Let me update this

halter73: The remaining question is whether we want On callbacks to wait for the Tasks of all previous On callbacks to complete. Obviously unblocking the ProcessMessagesAsync loop is more important, but it might be a good idea to unblock the client invocation loop as well.

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -342,6 +346,19 @@ private async Task StartAsyncCore(CancellationToken cancellationToken)
             }
         }
 
+        private async Task startProcessingInvocationMessages()
```
halter73: Someone's been writing too much Java 😆

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -735,7 +758,7 @@ private async Task DispatchInvocationAsync(InvocationMessage invocation)
             {
                 try
                 {
-                    await handler.InvokeAsync(invocation.Arguments);
+                    handler.InvokeAsync(invocation.Arguments);
```
halter73: I like not blocking the handlers on the async execution of the previous handler, but I still think the handler should be wrapped in something that awaits it so async exceptions can be logged.

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
mikaelm12: 😄 😆  😅

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
davidfowl: Lets leave it blocking for now. We'll only invoke a single callback at a time and that will be the contract.

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
halter73: Never mind, just go back to awaiting.

davidfowl: 👍

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -389,6 +406,12 @@ private async Task StopAsyncCore(bool disposing)
             {
                 await connectionState.StopAsync();
             }
+
+            if (_invocationMessageChannel != null)
+            {
+                _invocationMessageChannel.Writer.TryComplete();
+                await _invocationMessageReceiveTask;
```
halter73: Move this earlier into the method inside the connection lock.

Seeing the `await connectionState.StopAsync()` outside the connection lock above, I went down the rabbit hole of learning about the private ConnectionState class. I'm now wondering when we stopped using _connectionState to ["// Represent all the transient state about a connection"](https://github.com/aspnet/AspNetCore/blob/ee5d6728879a71eb4b2b16cd6786d28711815563/src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs#L1232)? It was purpose-built for stuff like _invocationMessageChannel and _invocationMessageReceiveTask. I was about to recommend moving the fields, but the are now so many *other* fields defined directly on the HubConnection class that track transient state to the connection, it hardly seems worthwhile.

I [filed an issue](https://github.com/aspnet/AspNetCore/issues/8500) to address that and to make sure that all the references to _connectionState are even safe to begin with. I backlogged it, because I can't really justify spending time right now on a massive refactoring of HubConnection. It's too bad, because using ConnectionState correctly could help us avoid bugs. Perhaps if we find a bunch of HubConnection bugs cropping up, we'll reprioritize.

davidfowl: No thinking about that, I think we should move this to the connection state, or at least into the receive loop itself.

davidfowl: @mikaelm12 lets change this to a local function and local variables. There should be no state inside of the HubConnection, this logic should live and die as part of the ReceiveLoop call.

mikaelm12: Okay, got it 👍

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -970,6 +983,11 @@ private async Task ReceiveLoop(ConnectionState connectionState)
             }
             finally
             {
+                if (invocationMessageChannel != null)
```
halter73: invocationMessageChannel cannot be null

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -1233,8 +1251,7 @@ public Task InvokeAsync(object[] parameters)
             }
         }
 
-        // Represents all the transient state about a connection
-        // This includes binding information because return type binding depends upon _pendingCalls
+        //TODO: Refactor all HubConnection into the ConnectionState class.
```
halter73: Nit:

```suggestion
        //TODO: Refactor all transient state about a connection into the ConnectionState class.
```

halter73 (approved):
pending nits

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
anurse: I don't know if I'd even bother with the TODO. We don't *know* it's worth doing until we see issues that indicate the refactor would solve a problem.

This is getting super nit-picky though so 🤷‍♂️

BrennanConroy (approved):

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
mikaelm12: Oh copy pasta, can't be null as currently written.

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
mikaelm12: I'll just leave it for now with the update so that moving forward it can at least serve as a sort of reminder that new state should be added to the `ConnectionState `object

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -1233,8 +1248,7 @@ public Task InvokeAsync(object[] parameters)
             }
         }
 
-        // Represents all the transient state about a connection
-        // This includes binding information because return type binding depends upon _pendingCalls
+        //TODO: Refactor all transient state about the connection into the ConnectionState class.
```
davidfowl: Is this still needed?

davidfowl (approved):

davidfowl (approved):
Beautiful!

PR #8337, opened 2019-03-08
fixing possible NRE, when viewContext is null

m0sa:
This change addresses a NRE in RazorPageBase.

When `ViewContext` is null this line assigns null to `antiforgery`:

https://github.com/aspnet/AspNetCore/blob/bd6faa5ca1f43f78ea708a92d3ff44608e23381b/src/Mvc/Mvc.Razor/src/RazorPageBase.cs#L647

... and then the next line attempts to call the `SetCookieTokenAndHeader` instance method, on this null value:

https://github.com/aspnet/AspNetCore/blob/bd6faa5ca1f43f78ea708a92d3ff44608e23381b/src/Mvc/Mvc.Razor/src/RazorPageBase.cs#L648

### DIFF
--- a/src/Mvc/Mvc.Razor/src/RazorPageBase.cs
+++ b/src/Mvc/Mvc.Razor/src/RazorPageBase.cs
@@ -644,9 +644,11 @@ namespace Microsoft.AspNetCore.Mvc.Razor
         public virtual HtmlString SetAntiforgeryCookieAndHeader()
         {
             var viewContext = ViewContext;
-            var antiforgery = viewContext?.HttpContext.RequestServices.GetRequiredService<IAntiforgery>();
-            antiforgery.SetCookieTokenAndHeader(viewContext?.HttpContext);
-
+            if (viewContext != null)
+            {
+                var antiforgery = viewContext.HttpContext.RequestServices.GetRequiredService<IAntiforgery>();
+                antiforgery.SetCookieTokenAndHeader(viewContext.HttpContext);
+            }
             return HtmlString.Empty;
         }
 


src/Mvc/Mvc.Razor/src/RazorPageBase.cs
```diff
@@ -644,9 +644,11 @@ public virtual async Task<HtmlString> FlushAsync()
...
         {
             var viewContext = ViewContext;
-            var antiforgery = viewContext?.HttpContext.RequestServices.GetRequiredService<IAntiforgery>();
-            antiforgery.SetCookieTokenAndHeader(viewContext?.HttpContext);
-
+            if (viewContext != null)
+            {
+                var antiforgery = viewContext.HttpContext.RequestServices.GetRequiredService<IAntiforgery>();
+                antiforgery.SetCookieTokenAndHeader(viewContext.HttpContext);
+            }
```
rynowak: @pranavkm - do you know why this method tolerates null? Is this for unit testing?

pranavkm: Fairly certain it's due to unit testing. I think it would be fair to not do null-checks here. (Same with the `FlushAsync` above). We would never expect the ViewContext to be null here when the page is executing.

rynowak: I think we should just take this change then. It doesn't seem overly important which semantic we support

rynowak (approved):

pranavkm:
Thanks for the PR!

PR #8339, opened 2019-03-08
WIP: out-of-HttpContext rendering

m0sa:
Toying around with an idea we discussed yesterday with @rynowak. This pulls out a razor base class that's independent of AspNetCore implementation details, and is meant as a discussion about possible / further steps in the direction to simplify out-of-context rendering of razor views.

Naming is hard, so I did a [McDonalds Theory](https://medium.com/@jonbell/mcdonalds-theory-9216e1c9da7d) thing.

I've encountered following issues:

- TagHelpers are very tied with `Mvc.ViewFeatures.Buffers`
- Most of the Mvc.ViewFeatures.Buffers stuff is `internal` so I didn't attempt to move it
- I tried not to introduce a dependency on the DI package
- cross solution changes are a PITA, making it very hard to contribute from the outside / require multiple PRs to deploy properly, I think? I had to replace `<Reference>` with `<ProjectReference>` to make sure that builds don't break locally...


src/Mvc/Mvc.Razor.RuntimeCompilation/src/Microsoft.AspNetCore.Mvc.Razor.RuntimeCompilation.csproj
```diff
@@ -12,7 +12,7 @@
   <ItemGroup>
     <Reference Include="Microsoft.AspNetCore.Mvc.Razor.Extensions" />
     <Reference Include="Microsoft.AspNetCore.Mvc.RazorPages" />
-    <Reference Include="Microsoft.AspNetCore.Razor.Runtime" />
+    <ProjectReference Include="..\..\..\Razor\Razor.Runtime\src\Microsoft.AspNetCore.Razor.Runtime.csproj" />
```
rynowak: I'd really prefer if we could avoid introducing a dependency between the compilation system and the runtime. I'll keep reading to see if this is really required.

m0sa: I couldn't build locally otherwise, the build seemed to use the nuget packages (the new interfaces I've added in AspNetCore.Razor weren't referencable in AspNetCore.Mvc.Razor)

rynowak:
Sorry for the delay, I've been busy with a few other things..

src/Mvc/Mvc.Razor.RuntimeCompilation/src/Microsoft.AspNetCore.Mvc.Razor.RuntimeCompilation.csproj
rynowak: Nevermind, I see this was already here.

rynowak: I'm not sure why you're having a problem building locally, but I'm sure will figure that out by the end.

src/Mvc/Mvc.Razor/src/RazorPageBase.cs
```diff
@@ -26,74 +25,26 @@ namespace Microsoft.AspNetCore.Mvc.Razor
     /// <summary>
     /// Represents properties and methods that are needed in order to render a view that uses Razor syntax.
     /// </summary>
-    public abstract class RazorPageBase : IRazorPage
+    public abstract class RazorPageBase : RazorRazorPageBase<ViewContext>, IRazorPage
```
rynowak: At a base level I really don't think we should try to make MVC inherit from this new thing. My concern is that if we're going to create a new thing, but have it inherit all of MVC's quirks and requirements. Basically, I think it's a success criteria to try and factor out the portable set of features and expose them through a reasonable design. I don't think it's a success criteria to have code-sharing. 

I'm going to try and focus my feedback on the design of the new thing, and once that's nailed it would be good to talk about what MVC does.

src/Razor/Razor.Runtime/src/RazorRazorPageBase.cs
```diff
@@ -0,0 +1,505 @@
...
+using Microsoft.AspNetCore.Html;
+using Microsoft.AspNetCore.Razor.Runtime.TagHelpers;
+using Microsoft.AspNetCore.Razor.TagHelpers;
+
+namespace Microsoft.AspNetCore.Razor
+{
+    /// <summary>
+    /// Represents properties and methods that are needed in order to render a view that uses Razor syntax.
+    /// </summary>
+    public abstract class RazorRazorPageBase<TOutputContext> : IRazorRazorPage<TOutputContext> where TOutputContext : IOutputContext
```
rynowak: I'm ignoring the name for now 😆

src/Razor/Razor.Runtime/src/RazorRazorPageBase.cs
```diff
@@ -0,0 +1,505 @@
...
+    {
+        private readonly Stack<TextWriter> _textWriterStack = new Stack<TextWriter>();
+        private StringWriter _valueBuffer;
+        private AttributeInfo _attributeInfo;
+        private TagHelperAttributeInfo _tagHelperAttributeInfo;
+
+        public virtual TOutputContext ViewContext { get; set; }
```
rynowak: I feel kinda iffy about putting a context type here, and it would really be preferrable to make this based on inject instead. 

Did you have particular scenarios in mind for this other than the fact that MVC has a context?

src/Razor/Razor.Runtime/src/RazorRazorPageBase.cs
```diff
@@ -0,0 +1,505 @@
...
+        public string Path { get; set; }
+
+        /// <inheritdoc />
+        public IDictionary<string, RenderAsyncDelegate> SectionWriters { get; } =
+            new Dictionary<string, RenderAsyncDelegate>(StringComparer.OrdinalIgnoreCase);
+
+        /// <summary>
+        /// Gets the dynamic view data dictionary.
+        /// </summary>
+        public dynamic ViewBag => ViewContext?.ViewBag;
```
rynowak: I'd really like to avoid having `ViewBag` here. There's an impactful runtime/startup cost to having dynamic types appear.

src/Razor/Razor.Runtime/src/RazorRazorPageBase.cs
```diff
@@ -0,0 +1,505 @@
...
+
+        public string Layout { get; set; }
+
+        /// <summary>
+        /// Gets the <see cref="TextWriter"/> that the page is writing output to.
+        /// </summary>
+        /// <summary>
+        /// Gets the <see cref="TextWriter"/> that the page is writing output to.
+        /// </summary>
+        public virtual TextWriter Output
```
rynowak: Are there good reasons to make any of these things virtual?

src/Razor/Razor.Runtime/src/RazorRazorPageBase.cs
```diff
@@ -0,0 +1,505 @@
...
+
+        /// <summary>
+        /// Creates a named content section in the page that can be invoked in a Layout page using
+        /// <c>RenderSection</c> or <c>RenderSectionAsync</c>
+        /// </summary>
+        /// <param name="name">The name of the section to create.</param>
+        /// <param name="section">The delegate to execute when rendering the section.</param>
+        /// <remarks>This is a temporary placeholder method to support ASP.NET Core 2.0.0 editor code generation.</remarks>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        protected void DefineSection(string name, Func<object, Task> section)
```
rynowak: This is an old back-compat placeholder, code generation doesn't use it so it can be removed.

src/Razor/Razor.Runtime/src/RazorRazorPageBase.cs
```diff
@@ -0,0 +1,505 @@
...
+            }
+            SectionWriters[name] = section;
+        }
+
+
+        /// <summary>
+        /// Writes the specified <paramref name="value"/> with HTML encoding to <see cref="Output"/>.
+        /// </summary>
+        /// <param name="value">The <see cref="object"/> to write.</param>
+        public virtual void Write(object value)
```
rynowak: Similar comment, would be really nice to make these non-virtual.

src/Razor/Razor.Runtime/src/RazorRazorPageBase.cs
```diff
@@ -0,0 +1,505 @@
...
+        }
+
+        private void WritePositionTaggedLiteral(string value, int position)
+        {
+            BeginContext(position, value.Length, isLiteral: true);
+            WriteLiteral(value);
+            EndContext();
+        }
+
+        public abstract void BeginContext(int position, int length, bool isLiteral);
```
rynowak: We don't code generate calls to these APIs anymore, they can be removed.

rynowak: `BeginContext/EndContext

src/Razor/Razor.Runtime/src/RazorRazorPageBase.cs
```diff
@@ -0,0 +1,505 @@
...
+            public string Prefix { get; }
+
+            public int PrefixOffset { get; }
+
+            public string Suffix { get; }
+
+            public int SuffixOffset { get; }
+
+            public bool Suppressed { get; set; }
+        }
```
rynowak: Since all of these ints are unuses, we can shrink this struct.

src/Razor/Razor.Runtime/src/RazorRazorPageBase.cs
```diff
@@ -0,0 +1,505 @@
...
+            }
+        }
+
+        public virtual void BeginWriteAttribute(
+            string name,
+            string prefix,
+            int prefixOffset,
+            string suffix,
+            int suffixOffset,
+            int attributeValuesCount)
```
rynowak: @pranavkm @NTaylorMullen - any objection to simplifying this? We can add a new overload, update the compiler and then reduce this proliferation

@m0sa - FYI these int values are used for some diagnostics things that aren't supported anymore.

src/Razor/Razor/src/IRazorRazorPage.cs
```diff
@@ -0,0 +1,71 @@
...
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Html;
+
+namespace Microsoft.AspNetCore.Razor
+{
+    /// <summary>
+    /// Represents properties and methods that are used by a Razor View for execution.
+    /// </summary>
+    public interface IRazorRazorPage<TOutputContext> where TOutputContext : IOutputContext
```
rynowak: I'm not sure that there's much value in putting this stuff in an interface

src/Razor/Razor/src/IRazorUrlHelper.cs
```diff
@@ -0,0 +1,11 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+using System.IO;
+using Microsoft.AspNetCore.Razor.TagHelpers;
+
+namespace Microsoft.AspNetCore.Razor
+{
+    public interface IRazorUrlHelper
+    {
+    }
+}
```
rynowak: I don't think we'll need to move this down to the Razor layer if we support inject.

src/Razor/Razor.Runtime/src/RazorRazorPageBase.cs
```diff
@@ -0,0 +1,505 @@
...
+        protected internal virtual TextWriter PopWriter()
+        {
+            var viewContext = ViewContext;
+            var writer = _textWriterStack.Pop();
+            viewContext.Writer = writer;
+            return writer;
+        }
+
+        /// <inheritdoc />
+        public abstract string Href(string contentPath);
```
rynowak: I'm not sure that this belong here

src/Razor/Razor.Runtime/src/RazorRazorPageBase.cs
```diff
@@ -0,0 +1,505 @@
...
+using System.ComponentModel;
+using System.Diagnostics;
+using System.IO;
+using System.Text.Encodings.Web;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Html;
+using Microsoft.AspNetCore.Razor.Runtime.TagHelpers;
+using Microsoft.AspNetCore.Razor.TagHelpers;
+
+namespace Microsoft.AspNetCore.Razor
```
rynowak: So to sum up what's here - it looks like a standalone/portable Razor base class would have the following stuff:
- I/O (writing markup and attributes)
- Tag Helpers 
- Layouts (Path/BodyContent/Layout
- Sections 
- Flush

I think we'd also add support for `@inject` - that's pretty essential, and allows you build in a lot of flexibility without putting every feature at the base layer.

Stuff I would cut:
- ViewBag (not something we recommend to users
- URL generation (can't be built-in at this layer)
- "Context" (not clear that this is generalizable)

-----

I'd be pretty happy if this is where we land.

src/Razor/Razor/src/RenderAsyncDelegate.cs
```diff
@@ -3,7 +3,7 @@
 
 using System.Threading.Tasks;
 
-namespace Microsoft.AspNetCore.Mvc.Razor
+namespace Microsoft.AspNetCore.Razor
```
rynowak: I don't think we can make this breaking change, I don't think we can predict the impact of changing this.

src/Razor/Razor.Runtime/src/RazorRazorPageBase.cs
NTaylorMullen: Simplify!!!!! 👍

src/Razor/Razor.Runtime/src/RazorRazorPageBase.cs
m0sa: Only had getting-rid-of-mvc in mind here.. We'd need at least an interface for it to get it in via [Inject], though.

src/Razor/Razor/src/IRazorUrlHelper.cs
m0sa: RazorInjectAttribute is currently in `Microsoft.AspNetCore.Mvc.Razor.Internal`

m0sa:
OK, gave this an overhaul, taking [the review](https://github.com/aspnet/AspNetCore/pull/8339/files#r264813478) into account. This is now a separate class, which nothing calls into (yet), and nothing uses (yet).

The big issues I see here are:

- TagHelpers, they seem to be tightly bound to `ViewContext`, and would have to be refactored in order to work with this, from the looks of it:

https://github.com/aspnet/AspNetCore/blob/master/src/Mvc/Mvc.Razor/src/RazorPageBase.cs#L170

- DI would need to be re-considered. `[RazorInject]` currently lives in the MVC sln.

src/Razor/Razor.Runtime/src/RazorRazorPageBase.cs
m0sa: Not if MVC doesn't inhert from it

src/Razor/Razor/src/IRazorRazorPage.cs
m0sa: It would be nice to have at least some common interface for the code that the code generated by razor calls, instead of relying on convention, and keeping two separate classes in-sync. Especially if customers are allowed to implement and specify their own base classes.

rynowak:
Sorry for the delays on this, just a quick poke to say that I'm pretty busy with summit this week. I hope to get back to this in a few days.

src/Razor/Razor.Runtime/src/Runtime/RazorRazorBasePage.cs
```diff
@@ -0,0 +1,468 @@
...
+        /// <summary>
+        /// Gets or sets the <see cref="TextWriter"/> that the page is writing output to.
+        /// </summary>
+        public TextWriter Output { get; set; }
+
+        /// <inheritdoc />
+        public string Path { get; set; }
+
+        /// <inheritdoc />
+        // REVIEW: using Func<Task> instead of moving RenderAsyncDelegate around, which is a breaking change
```
rynowak: I think putting `Func<Task>` here is fine. Agreed that we can't move that type.

src/Razor/Razor.Runtime/src/Runtime/RazorRazorBasePage.cs
```diff
@@ -0,0 +1,468 @@
...
+        public IHtmlContent BodyContent { get; set; }
+
+        /// <inheritdoc />
+        public IDictionary<string, Func<Task>> PreviousSectionWriters { get; set; }
+
+        /// <summary>
+        /// Gets the <see cref="System.Text.Encodings.Web.HtmlEncoder"/> to use when this <see cref="RazorRazorPageBase"/>
+        /// handles non-<see cref="IHtmlContent"/> C# expressions.
+        /// </summary>
+        public HtmlEncoder HtmlEncoder { get; set; } = HtmlEncoder.Default; // REVIEW: how should we [RazorInject] this in a non-breaking way?
```
rynowak: I was thinking we could do the following for `[RazorInject]`.

Introduce a new type `Microsoft.AspNetCore.Razor.RazorInjectAttribute`.Mark the old one obsolete, but make MVC still honor it. We might want to come up with another name for the new one so they don't clash. As we we can still load old types and MVC still supports the old attribute we are OK.

src/Razor/Razor.Runtime/src/Runtime/RazorRazorBasePage.cs
```diff
@@ -0,0 +1,468 @@
...
+        private TagHelperAttributeInfo _tagHelperAttributeInfo;
+
+        public string Layout { get; set; }
+
+        /// <summary>
+        /// Gets or sets the <see cref="TextWriter"/> that the page is writing output to.
+        /// </summary>
+        public TextWriter Output { get; set; }
+
+        /// <inheritdoc />
```
rynowak: BTW inheritdoc doesn't do anything 😆

src/Razor/Razor.Runtime/src/Runtime/RazorRazorBasePage.cs
```diff
@@ -0,0 +1,468 @@
...
+
+        private struct AttributeInfo
+        {
+            public AttributeInfo(
+                string name,
+                string prefix,
+                int prefixOffset,
+                string suffix,
+                int suffixOffset,
+                int attributeValuesCount)
```
rynowak: All of these integers can be removed and simplified. I'm OK if you want to leave them here for now for us to clean up, it's our mess after all.

rynowak (commented):
It seems like what's here is the right API shape, and it's ready for the next level of polish (tests, docs, etc). At this point it should be possible to use `@inherits` in a `.cshtml` file and have it compile and output text using this as a base class. 

What else you want to try and accomplish is up to your interests and time. I'd be happy with any way you want to break down this work, or any order of operations/stopping point you're prepared to handle.

1. Integration with MVC - make it possible for MVC to use this base-class by creating an adapter for `IRazorPage` in MVC's code.
2. Implementing layouts + the features in [RazorView](https://github.com/aspnet/AspNetCore/blob/master/src/Mvc/Mvc.Razor/src/RazorView.cs#L21)
3. Implementing a loader so you can create instances of these things based on [RazorCompiledItemLoader](https://github.com/aspnet/AspNetCore/blob/master/src/Razor/Razor.Runtime/src/Hosting/RazorCompiledItemLoader.cs)
4. Integrating or polishing the view buffer system that MVC has at this layer

rynowak:
Make sure to rebase if you update this also. We fixed some build issues that were causing Java to get stuck on the CI, and unfortunately that code is in the repo and requires a rebase to pick up.

PR #8340, opened 2019-03-08
Make pubinternal types in Antiforgery internal

pranavkm:
Adresses https://github.com/aspnet/AspNetCore/issues/8308


pranavkm:
Pre-emptively sending out a PR. Most of these types fall in to the category of default implementations of interfaces, so it makes it fairly easy to consider making them true `internal`. `IAntiforgeryFeature` was the one type that stuck out. It's set in `FeatureCollection` but only consumed by `Antiforgery`. Should it be made `public` since it's an actual usable interface?

src/Antiforgery/src/Microsoft.AspNetCore.Antiforgery.csproj
```diff
@@ -1,10 +1,9 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <Description>An antiforgery system for ASP.NET Core designed to generate and validate tokens to prevent Cross-Site Request Forgery attacks.</Description>
     <TargetFramework>netcoreapp3.0</TargetFramework>
     <IsAspNetCoreApp>true</IsAspNetCoreApp>
-    <NoWarn>$(NoWarn);CS1591</NoWarn>
```
Eilon: NICE!

Eilon (approved):

pranavkm:
Spoke to @rynowak offline. `IAntiforgeryFeature` seems like an implementation detail and doesn't need to be public. We can wait for user feedback to see if it needs to be made public.

logic01:
```
using Microsoft.AspNetCore.Antiforgery;
            var serializer = afType
                .GetField("_tokenSerializer", BindingFlags.NonPublic | BindingFlags.Instance)
                .GetValue(_antiForgery) as IAntiforgeryTokenSerializer;
```

Looks like my team is one of the few who used this interface in dotnet core 2.x. Is there any way to get similar functionality?

PR #8344, opened 2019-03-08
[Helix] Reenable music store

HaoK:
Starting point for https://github.com/aspnet/AspNetCore/issues/8343


HaoK:
@aspnet/brt fyi for a MacOS test failure

PR #8358, opened 2019-03-09
Pass the formatter context to the CreateJsonSerializer. It gives ability to override and create serializer specific to the context

mnltejaswini:
 - Pass the formatter context to overridable CreateJsonSerializer. This allows users to override it and create serializer customized to the context.

Addresses:
  In our scenario, we need to select serializer based on the request properties like api version. With context parameter, we can just override CreateJsonSerializer and plug the custom serializer based on the request.


mnltejaswini:
@davidfowl

rynowak:
The way you're supposed to do this is by registering multiple formatters and overriding `CanWrite`/`CanRead` - does that work for you?

mnltejaswini:
@rynowak  Yes that also works, in our case where we want serializer per request's api version, we might need one formatter per version. With this change we have one formatter per service, but hook the appropriate serializer. Which approach do you recommend?

rynowak:
> With this change we have one formatter per service, but hook the appropriate serializer. Which approach do you recommend?

How many versions will you have? How frequently do you expect to bump up version? Basically I want to understand if you're going to have 4-5 or like 10. That would help make a decision about whether new extensibility is needed. 

Regardless, any changes here will go into 3.0 - which I assume you're not using in production yet 😆

mnltejaswini:
@rynowak   We are planning to flight one api version per trimester , not to mention that we need to support old api versions for over an year. So we might have under 10 versions if are actively deprecating the old versions, if not more than 10

pranavkm (approved):

pranavkm:
Thanks @mnltejaswini

Issue #8359, opened 2019-03-09
Update the health check endpoint route builder extensions to make the endpoint display name configurable.


CBaud:
### Is your feature request related to a problem? Please describe.
In ASP.NET Core 3.0 Preview 3, the health check endpoint route builder extensions specify a common display name ("Health checks") for all health check endpoints.

### Describe the solution you'd like
It would be helpful if the extensions allowed the display name to be configurable by the service. For example, a service running on Kubernetes may have two separate health check endpoints for the liveness and readiness probes.

### Describe alternatives you've considered
The alternative would be to manually configure the health check endpoints in my service.

rynowak:
Thanks @CBaud 

FYI I'm not sure you saw my follow up PR - we decided to make this part of the builder returned from `MapHealthChecks`. So this is uniform across all endpoint routing stuff and won't be missed when we add future router-ware. Just mentioning because we ended up not going with your fix. I appreciate that you make the effort and I don't want you to think we don't appreciate it.

PR #8361, opened 2019-03-09
Update the health check endpoint route builder extensions to make the endpoint display name configurable. (#8359)

CBaud:
This updates the health check endpoint route builder extensions to make the display name configurable.

Addresses #8359


CBaud:
FYI @JamesNK @rynowak

rynowak:
Looks good, thanks Chris thanks!

We may end up changing this to use a builder `.AddHealthChecks(...).WithDisplayName(...)` to make it consistent with some other routing stuff. 

I'm going to accept this PR for now since it's an improvement.

rynowak (approved):

PR #8362, opened 2019-03-09
System.Text.Json based formatters

### LINKED ISSUE #7256: Add a System.Text.Json based input \ output formatter

pranavkm:
Fixes #7256


src/Mvc/Mvc.Formatters.Json/src/TranscodingWriteStream.cs
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        public override void Write(byte[] buffer, int offset, int count)
+            => throw new NotSupportedException();
+
+        public override async Task WriteAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        {
+            ThrowArgumentOutOfRangeException(buffer, offset, count);
+
+            var maxChars = Encoding.UTF8.GetMaxCharCount(count);
```
davidfowl: I'm not sure about this. An alternative implementation would be to use the encoder and just pick a buffer size. What if maxChars is huge?

src/Mvc/Mvc.Formatters.Json/src/TranscodingReadStream.cs
```diff
@@ -0,0 +1,93 @@
...
+            => throw new NotSupportedException();
+
+        public override int Read(byte[] buffer, int offset, int count)
+            => throw new NotSupportedException();
+
+        public override async Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        {
+            ThrowArgumentOutOfRangeException(buffer, offset, count);
+
+            if (_reader.EndOfStream)
```
davidfowl: This call synchronous Read on the underlying Stream. Why use the StreamReader at all? Why not just use the decoder?

src/Mvc/Mvc.Formatters.Json/src/TranscodingWriteStream.cs
pranavkm: Yeah, I need to clean this up. Likely used a fixed size char buffer

src/Mvc/Mvc.Formatters.Json/src/TranscodingReadStream.cs
pranavkm: Wasn't aware this did a blocking call. That said, it was a little easier to use the reader compared to the encoder. I can switch it over

src/Mvc/Mvc.Formatters.Json/test/Microsoft.AspNetCore.Mvc.Formatters.Json.Test.csproj
```diff
@@ -0,0 +1,14 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
rynowak: Is there a reason why we need a `Formatters.Json` at all? I don't think that the reasons why we did this in the past are really relevant anymore.

src/Mvc/Mvc.Formatters.Json/src/JsonFormatterOptions.cs
```diff
@@ -0,0 +1,21 @@
...
+    /// Provides options to configure formatters using <c>System.Text.Json</c>.
+    /// </summary>
+    public class JsonFormatterOptions
+    {
+        /// <summary>
+        /// Gets the <see cref="JsonSerializerOptions"/>.
+        /// </summary>
+        public JsonSerializerOptions SerializerOptions { get; } = new JsonSerializerOptions
+        {
+            IgnoreNullPropertyValueOnRead = true,
```
rynowak: It might be worth establishing a pattern now that we add comments to these explaining *why* we have the settings that we do.

src/Mvc/Mvc.Formatters.Json/src/JsonInputFormatter.cs
```diff
@@ -0,0 +1,83 @@
...
+            {
+                throw new ArgumentNullException(nameof(encoding));
+            }
+
+            var httpContext = context.HttpContext;
+            var inputStream = GetInputStream(httpContext, encoding);
+
+            try
+            {
+                var result = await JsonSerializer.ReadAsync(inputStream, context.ModelType, _serializerOptions, httpContext.RequestAborted);
```
rynowak: I have a doubt about using `httpContext.RequestAborted` here. I'm not sure we use cancellation tokens in other places we do input reading because we generally want the request to go through if the client discconnected after sending the data.

src/Mvc/Mvc.Formatters.Json/src/JsonInputFormatter.cs
```diff
@@ -0,0 +1,83 @@
...
+                return InputFormatterResult.Success(result);
+            }
+            catch (JsonReaderException exception)
+            {
+                var inputException = new InputFormatterException(exception.Message, exception);
+                context.ModelState.TryAddModelException(key: string.Empty, inputException);
```
rynowak: Using `string.Empty` as a model state key does *baaaaad* things.

Is there a reason why this is different from what the old one does? Just avoiding throwing?

src/Mvc/Mvc.Formatters.Json/src/JsonFormatterOptions.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    /// <summary>
+    /// Provides options to configure formatters using <c>System.Text.Json</c>.
+    /// </summary>
+    public class JsonFormatterOptions
+    {
+        /// <summary>
+        /// Gets the <see cref="JsonSerializerOptions"/>.
+        /// </summary>
+        public JsonSerializerOptions SerializerOptions { get; } = new JsonSerializerOptions
```
rynowak: implement real property and make this not possible to set, but not to null? I can see someone wanting to replace this object completely.

src/Mvc/Mvc.Formatters.Json/src/JsonInputFormatter.cs
```diff
@@ -0,0 +1,83 @@
...
+            }
+
+            return InputFormatterResult.Failure();
+        }
+
+        private Stream GetInputStream(HttpContext httpContext, Encoding encoding)
+        {
+            if (encoding == UTF8EncodingWithoutBOM || encoding == Encoding.UTF8)
```
rynowak: should this be comparing `encoding.WebName` instead? I don't like that we're comparing instances here.

src/Mvc/Mvc.Formatters.Json/src/JsonOutputFormatter.cs
```diff
@@ -0,0 +1,67 @@
...
+            await JsonSerializer.WriteAsync(context.Object, context.ObjectType, writeStream, _serializerOptions, requestAborted);
+            await writeStream.FlushAsync(requestAborted);
+        }
+
+        private Stream GetWriteStream(HttpContext httpContext, Encoding selectedEncoding)
+        {
+            if (selectedEncoding == Encoding.UTF8)
+            {
+                return httpContext.Response.Body;
+            }
```
rynowak: Does this write a bom? Not needed because the serializer doesn't do it? comment plz

src/Mvc/Mvc.Formatters.Json/src/TranscodingReadStream.cs
```diff
@@ -0,0 +1,93 @@
...
+
+                return Encoding.UTF8.GetBytes(
+                    charBuffer.AsSpan(0, readChars),
+                    buffer.AsSpan(offset, count));
+            }
+            finally
+            {
+                _charPool.Return(charBuffer);
+            }
+        }
```
rynowak: This doesn't seem so terribad to implement.

src/Mvc/Mvc.Formatters.Json/src/TranscodingWriteStream.cs
```diff
@@ -0,0 +1,104 @@
...
+        public TranscodingWriteStream(Stream stream, Encoding targetEncoding)
+        {
+            _stream = stream;
+            _charPool = ArrayPool<char>.Shared;
+            _targetEncoding = targetEncoding;
+        }
+
+        public override bool CanRead => true;
+        public override bool CanSeek => false;
+        public override bool CanWrite => false;
```
rynowak: derp

src/Mvc/Mvc.Formatters.Json/src/TranscodingWriteStream.cs
rynowak: if maxchars is huge it will be pooled?

src/Mvc/Mvc.Formatters.Json/src/JsonInputFormatter.cs
davidfowl: Why not? In case somebody derives?

src/Mvc/Mvc.NewtonsoftJson/src/DependencyInjection/NewtonsoftJsonMvcOptionsSetup.cs
```diff
@@ -59,33 +59,58 @@ internal class NewtonsoftJsonMvcOptionsSetup : IConfigureOptions<MvcOptions>
...
+            var jsonInputFormatter = new NewtonsoftJsonInputFormatter(
                 jsonInputLogger,
                 _jsonOptions.SerializerSettings,
                 _charPool,
                 _objectPoolProvider,
                 options,
-                _jsonOptions));
+                _jsonOptions);
 
+            InsertBefore(options.InputFormatters, JsonInputFormatter, jsonInputFormatter);
```
rynowak: We should remove the other ones too IMO. It's OK if we have to change what the method is called to make that clear. I think users want one thing or the other.

src/Mvc/Mvc.Formatters.Json/src/JsonInputFormatter.cs
rynowak: the encoding collection is mutable, you could create your own instance and add it

this chafes me not because it's likely to be broken, but because it just communicates the wrong thing, you want to handle UTF8 in general, and using something like the webname is the right way to do that.

src/Mvc/Mvc.Formatters.Json/src/JsonInputFormatter.cs
```diff
@@ -0,0 +1,83 @@
...
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            if (encoding == null)
+            {
+                throw new ArgumentNullException(nameof(encoding));
+            }
+
+            var httpContext = context.HttpContext;
+            var inputStream = GetInputStream(httpContext, encoding);
```
rynowak: Do we need to do any work to make this guy check for invalid characters/encoding? I'm not sure if that's something we need or not now that the serializer operates on bytttes

davidfowl: That would invalidate the performance wins no? That would mean everyone that uses the serializer needs to do that as well.

cc @GrabYourPitchforks any guidance here?

rynowak: Security is a thing tho.

davidfowl: GIGO?

src/Mvc/Mvc.Formatters.Json/src/DependencyInjection/JsonFormatterMvcOptionsSetup.cs
```diff
@@ -0,0 +1,32 @@
...
+    internal class JsonFormatterMvcOptionsSetup : IConfigureOptions<MvcOptions>
+    {
+        private readonly JsonFormatterOptions _options;
+
+        public JsonFormatterMvcOptionsSetup(IOptions<JsonFormatterOptions> options)
+        {
+            _options = options.Value;
+        }
+
+
```
jkotalik: nit space

src/Mvc/Mvc.Formatters.Json/src/DependencyInjection/JsonMvcCoreBuilderExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+    /// Extensions methods for configuring MVC via an <see cref="IMvcCoreBuilder"/>.
+    /// </summary>
+    public static class JsonMvcCoreBuilderExtensions
+    {
+        /// <summary>
+        /// Configures Newtonsoft.Json specific features such as input and output formatters.
+        /// </summary>
+        /// <param name="builder">The <see cref="IMvcCoreBuilder"/>.</param>
+        /// <returns>The <see cref="IMvcCoreBuilder"/>.</returns>
+        public static IMvcCoreBuilder AddJsonFormatters(this IMvcCoreBuilder builder)
```
jkotalik:
```suggestion
        public static IMvcCoreBuilder AddJsonFormatter(this IMvcCoreBuilder builder)
```

If you want to keep the method names consistent.

src/Mvc/Mvc.Formatters.Json/src/JsonInputFormatter.cs
jkotalik: Yeah I think having the ct as a parameter to ReadRequestBodyAsync would be better. We don't do this in the FormReader.

src/Mvc/Mvc.Formatters.Json/src/TranscodingReadStream.cs
```diff
@@ -0,0 +1,93 @@
...
+        {
+            ThrowArgumentOutOfRangeException(buffer, offset, count);
+
+            if (_reader.EndOfStream)
+            {
+                return 0;
+            }
+
+            // Given the buffer size, determine the maximum number of characters we could read.
+            var maxCharsToRead = count / _maxCharSizeInBytes;
```
jkotalik: I feel like this will poorly fill up the user passed in buffer. I did something similar with HttpResponseWritingExtensions with encoding: https://github.com/aspnet/AspNetCore/blob/master/src/Http/Http.Abstractions/src/Extensions/HttpResponseWritingExtensions.cs#L123 may be useful.

src/Mvc/Mvc.Formatters.Json/src/TranscodingWriteStream.cs
davidfowl: The array pool will return a heap allocated buffer if the size if too big or the pool is exhausted.

src/Mvc/Mvc.Formatters.Json/src/JsonFormatterOptions.cs
pranavkm: Per https://github.com/aspnet/Mvc/issues/6816#issuecomment-329221624, we made a conscious choice not to allow this with the Newtonsoft.Json SerializerSettings. The same arguments apply here.

src/Mvc/Mvc.Formatters.Json/src/JsonInputFormatter.cs
pranavkm: I can have it throw now. We eventually want to be able to construct a path from the reader and have the response 400 rather than 500.

pranavkm: https://github.com/aspnet/AspNetCore/issues/8474 to track follow up

src/Mvc/Mvc.Formatters.Json/src/JsonInputFormatter.cs
pranavkm: Filed https://github.com/aspnet/AspNetCore/issues/8475 so we don't lose track of this

blowdart: I'd say yes, and you'd reject the json as invalid, What was the behaviour of the previous formatters?

rynowak: The behaviour of the previous formatter was that we passed JSON.NET a streamreader and they did what they do. https://github.com/aspnet/AspNetCore/blob/master/src/Mvc/Mvc.NewtonsoftJson/src/NewtonsoftJsonInputFormatter.cs#L145

This is a new serializer so we need to know that the expectations we have are correct. 

/cc @JamesNK in case he has SPICY thoughts

davidfowl: We can’t have logic to validate the JSON  that’s the deserialzers job, this is about validating UTF8.

davidfowl: It's a good question for @ahsonkhan. I'm thinking the reader should handle this case if that isn't already the case today with an option to turn it off as a fast path if you already did the work yourself.

GrabYourPitchforks: The `System.Text.Json` reader itself should be performing this check on your behalf.

JamesNK: Newtonsoft.Json had a check for bad escape sequences. Otherwise it assumed that all the char values that TextReader gave it were valid.

pranavkm: @ahsonkhan \ @steveharter is this already being done? Do we need a separate issue tracking implementing \ verifying this?

rynowak: cool beans!

ahsonkhan: > is this already being done?

The `Utf8JsonReader.Read()`, and the `JsonDocument.Parse()` on their own validate/reject invalid UTF-8 data as long as it is **not** contained within a JSON string/property name. This means token delimiters, whitespace, etc. all have to be valid UTF-8. Otherwise, as part of JSON validation, these invalid UTF-8 bytes will be detected and rejected (throw `JsonReaderException`). For invalid/ill-formed UTF-8 within quoted strings, if you were to ever materialize the token into a UTF-16 string by calling `Utf8JsonReader.GetString()` or `JsonElement.GetString()`/`GetRawText()`, that will validate and throw. The `Utf8JsonReader` does **not** validate invalid UTF-8 bytes within quoted strings if you never materialize the strings within the reader loop.

```C#
string jsonString = "[[[]], {\"a\":1, \"b\": 2}, 3, {\"a\":{}, \"b\":{\"c\":[]} }, [{\"a\":1, \"b\": 2}, null]]";
byte[] dataUtf8 = Encoding.UTF8.GetBytes(jsonString);

var reader = new Utf8JsonReader(dataUtf8, isFinalBlock: true, default);
var stringIndices = new List<int>();
while (reader.Read())
{
    if (reader.TokenType == JsonTokenType.PropertyName)
    {
        // Move back to after the first quote
        // Before: "<name>":|
        // After: "|<name>":
        stringIndices.Add((int)reader.BytesConsumed - reader.ValueSpan.Length - 2);
    }
}

for (int i = 0; i < dataUtf8.Length; i++)
{
    dataUtf8 = Encoding.UTF8.GetBytes(jsonString);
    dataUtf8[i] = 0xEF;
    if (stringIndices.Contains(i))
    {
        // Invalid UTF-8 bytes contained within quoted strings are not rejected until the string is materialized
        using (JsonDocument doc = JsonDocument.Parse(dataUtf8))
        {
            // navigating to the token/JsonElement with the invalid UTF-8 byte and calling GetString() on it will throw
        }
    }
    else
    {
        Assert.Throws<JsonReaderException>(() => JsonDocument.Parse(dataUtf8));
    }
}
```

The JSON Deserializer APIs (like `JsonSerializer.Parse`) will reject invalid UTF-8 tokens if it occurs anywhere within the JSON payload (including within JSON strings/properties).
```C#
dataUtf8 = Encoding.UTF8.GetBytes("\"hello\"");
Assert.Equal("hello", Serialization.JsonSerializer.Parse<string>(dataUtf8));
dataUtf8[3] = 0xEF;
Assert.Throws<InvalidOperationException>(() => Serialization.JsonSerializer.Parse<string>(dataUtf8));

dataUtf8 = Encoding.UTF8.GetBytes("[ ]");
Assert.Equal(Array.Empty<int>(), Serialization.JsonSerializer.Parse<int[]>(dataUtf8));
dataUtf8[1] = 0xEF;
Assert.Throws<JsonReaderException>(() => Serialization.JsonSerializer.Parse<int[]>(dataUtf8));
```

pranavkm:
🆙 📅

src/Mvc/Mvc.Formatters.Json/src/JsonInputFormatter.cs
ahsonkhan: > I'm thinking the reader should handle this case if that isn't already the case today with an option to turn it off as a fast path if you already did the work yourself.

@davidfowl, based on the above clarification, do you still think the **reader** needs to validate UTF-8 (particularly within quoted strings) even if no one materializes them? Currently, that validation is lazy. The **serializer** does the validation already primarily because it needs to get the string eventually.

src/Mvc/Mvc.Core/src/Formatters/TranscodingReadStream.cs
```diff
@@ -0,0 +1,225 @@
...
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Mvc.Formatters.Json
+{
+    internal sealed class TranscodingReadStream : Stream
+    {
+        internal const int MaxByteBufferSize = 4096;
+        internal const int MaxCharBufferSize = 4 * MaxByteBufferSize;
```
ahsonkhan: Where is this 4x factor coming from?

src/Mvc/Mvc.Core/src/Formatters/TranscodingReadStream.cs
```diff
@@ -0,0 +1,225 @@
...
+            _byteBuffer = _byteBuffer.Slice(bytesUsed);
+
+            _charBuffer = new ArraySegment<char>(_charBuffer.Array, 0, charsUsed);
+
+            return readBytes == 0;
+        }
+
+        private void ThrowArgumentOutOfRangeException(byte[] buffer, int offset, int count)
+        {
+            if (count == 0)
```
ahsonkhan: Should this be `<= 0`?

src/Mvc/Mvc.Core/src/Formatters/TranscodingReadStream.cs
```diff
@@ -0,0 +1,225 @@
...
+                // We need to flush on the last write. This is true when we exhaust the input Stream and any buffered content.
+                var allContentRead = eof && _charBuffer.Count == 0 && _byteBuffer.Count == 0;
+
+                if (_charBuffer.Count > 0 && readBuffer.Count < MaxByteCountForUTF8Char && readBuffer.Count < Encoding.UTF8.GetByteCount(_charBuffer.AsSpan(0, 1)))
+                {
+                    // It's possible that the passed in buffer is smaller than the size required to encode a single
+                    // char. For instance, the JsonSerializer may pass in a buffer of size 1 or 2 which
+                    // is insufficient if the character requires more than 2 bytes to represent. In this case, read
+                    // content in to an overflow buffer and fill up the passed in buffer.
+                    _encoder.Convert(
```
ahsonkhan: I am having difficulty understanding the need for `_overflowBuffer` here. It looks like this buffer will always only be between 1-3 bytes in length. Is that correct? Can you explain why its needed? Why is the JsonSerializer passing in a buffer of size 1 or 2?

src/Mvc/Mvc.Core/src/Formatters/TranscodingWriteStream.cs
```diff
@@ -0,0 +1,156 @@
...
+                        out var bytesUsed,
+                        out encoderCompletd);
+
+                    await _stream.WriteAsync(byteBuffer.AsMemory(0, bytesUsed), cancellationToken);
+                    charsWritten += charsEncoded;
+                }
+            }
+            finally
+            {
+                ArrayPool<byte>.Shared.Return(byteBuffer);
```
ahsonkhan: Clear the array before returning?

src/Mvc/Mvc.Core/src/Formatters/TranscodingWriteStream.cs
```diff
@@ -0,0 +1,156 @@
...
+            }
+
+            // At this point, we've written all the buffered chars to the underlying Stream.
+            _charsDecoded = 0;
+        }
+
+        private void ThrowArgumentException(byte[] buffer, int offset, int count)
+        {
+            if (buffer.Length - offset < count)
```
ahsonkhan: Negative count check?

src/Mvc/Mvc.Core/src/Formatters/TranscodingReadStream.cs
pranavkm: It seemed like a reasonable limit size to for the char buffer. The ctor attempts to use the encoding to determine the max array size, but this is the upper bound for things.

src/Mvc/Mvc.Core/src/Formatters/TranscodingReadStream.cs
pranavkm: In general the buffer size is fairly large. However, the serializer sometimes call with a size 1 byte buffer:

image

The calling code: https://github.com/dotnet/corefx/blob/master/src/System.Text.Json/src/System/Text/Json/Serialization/JsonSerializer.Read.Stream.cs#L64-L70. I considered making an ask that would require the buffer to be at least 6 byte long, but this seemed like the right fix to make regardless. cc @stephentoub in case he's got any thoughts or suggestion here.

src/Mvc/Mvc.Core/src/Formatters/TranscodingWriteStream.cs
pranavkm: We generally don't clear these buffers e.g. https://github.com/aspnet/AspNetCore/blob/master/src/Http/WebUtilities/src/HttpResponseStreamWriter.cs#L78

src/Mvc/Mvc.Core/src/Formatters/TranscodingReadStream.cs
stephentoub: > However, the serializer sometimes call with a size 1 byte buffer

This could be tweaked.  Currently the serializer is trying to fill the buffer it has before passing that along to the Utf8JsonReader, because every call to the reader has relatively expensive setup costs, and we need to read everything from the stream anyway, plus it potentially avoids an extra trip through the reader for an empty buffer when we make the final read on the stream and find that it's eof.

But there's a balance that can be struck.  This is just a matter of finding the right balance / heuristic in https://github.com/dotnet/corefx/blob/b8bc4ff80c5f7baa681e8a569d367356957ba78a/src/System.Text.Json/src/System/Text/Json/Serialization/JsonSerializer.Read.Stream.cs#L58-L62 as to how much to proactively fill the buffer.  As I noted in https://github.com/dotnet/corefx/pull/35823, "We can tweak this further in the future as needed, e.g. only making a second attempt to fill the buffer rather than doing so aggressively."

src/Mvc/Mvc.Core/src/Formatters/SystemTextJsonInputFormatter.cs
```diff
@@ -0,0 +1,95 @@
...
+    {
+        private readonly JsonSerializerOptions _serializerOptions;
+
+        /// <summary>
+        /// Initializes a new instance of <see cref="SystemTextJsonInputFormatter"/>.
+        /// </summary>
+        /// <param name="options">The <see cref="MvcOptions"/>.</param>
+        public SystemTextJsonInputFormatter(MvcOptions options)
+        {
+            _serializerOptions = options.SerializerOptions;
```
rynowak: Should this be `JsonSerializerOptions`?

src/Mvc/Mvc.Core/test/Formatters/TranscodingReadStreamTest.cs
```diff
@@ -0,0 +1,113 @@
...
+
+        [Theory]
+        [MemberData(nameof(ReadAsyncInputLatin))]
+        public Task ReadAsync_Works_WhenInputIs_ASCII(string message)
+        {
+            // Arrange
+            var sourceEncoding = Encoding.ASCII;
+            return ReadAsyncTest(sourceEncoding, message);
+        }
+
```
rynowak: I think we should definitely have a case where we use the read-stream directly. Possibly reading it character by character. I want to make sure we're exercising the special case taht requires overflow buffer.

rynowak (approved):

src/Mvc/Mvc.Core/src/Formatters/TranscodingReadStream.cs
ahsonkhan: If it is primarily for UTF-16 to UTF-8/etc, then 3x is the upper bound.

single UTF-16 character -> 1-3 UTF-8 bytes
surrogate pairs (i.e. 2 UTF-16 characters) -> 4 UTF-8 bytes (i.e. factor is still <= 3x)

src/Mvc/Mvc.Core/src/Formatters/TranscodingWriteStream.cs
ahsonkhan: For some context, see recent corefx discussion on clearing:
https://github.com/dotnet/corefx/pull/35927#discussion_r264083647

tl;dr

We don't have consistent guidance on when to clear atm, but could be helpful for defense in depth.

src/Mvc/Mvc.Core/src/Formatters/TranscodingReadStream.cs
pranavkm: > If we expect arbitrary encodings, then ignore my comment, since I don't know what the worst case factor would be in that case.

I could change it to 3x. Unicode is the 2nd encoding that we register by default with the formatters, so that makes sense.

src/Mvc/Mvc.Core/src/Formatters/TranscodingWriteStream.cs
```diff
@@ -0,0 +1,166 @@
...
+            }
+
+            if (offset <= 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(offset));
+            }
+
+            if (buffer.Length - offset < count)
+            {
+                throw new ArgumentOutOfRangeException(nameof(count));
```
ahsonkhan: nit: Why did you decide on throwing with count? What if `offset > buffer.Length` and `count = 0`?

src/Mvc/Mvc.Core/src/Formatters/TranscodingReadStream.cs
ahsonkhan: cc @grabyourpitchforks

src/Mvc/Mvc.Core/src/Formatters/SystemTextJsonInputFormatter.cs
pranavkm: I'm slightly jumping the gun here, but we're invariably going to start reading other options from `MvcOptions` to configure the serializer in a couple of releases. Plus it also gets us out of the business of setting up the serializer options since `MvcOptions` initializes it correctly.

rynowak: I was just referrring to the property name. Like why `MvcOptions.SerializerSettings` instead of `MvcOptions.JsonSerializerSettings`.

Issue #8372, opened 2019-03-09
Reuse header value strings across requests


benaadams:
### Is your feature request related to a problem? Please describe.

For each request from the browser a lot of repeated headers are sent that are identical to the previous request.

This results in a lot of string allocations, especially for `Accept`, `Cookie` and `User-Agent` e.g.
```
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8
Accept-Encoding: gzip, deflate, br
Accept-Language: en-US,en;q=0.9
Cache-Control: no-cache
Connection: keep-alive
Cookie: (**redacted up to 2KB of data**)
Host: github.com
Pragma: no-cache
Upgrade-Insecure-Requests: 1
User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/72.0.3626.121 Safari/537.36
```

### Describe the solution you'd like
The previous request has already materialized these values as strings, so they could be reused rather than allocating new strings.

### Additional context
Tracking issue

In March 5th, 2019's ASP.NET Community Standup, @DamianEdwards expressed [that he thought this was already happening](https://youtu.be/PC7lnstHBBc?t=4014); then continued to suggest that in a future version [it would be important to continue to reduce memory used for headers](https://youtu.be/PC7lnstHBBc?t=4181).

/cc @davidfowl

valeriob:
Hi, how does it relate to http2 header compression ?

PR #8374, opened 2019-03-09
Reuse previous materialized strings

benaadams:
For each request from the browser a lot of repeated headers are sent that are identical to the previous request. This results in a lot of string allocations, especially for `Accept`, `Cookie` and `User-Agent` when materializing these values.

This change drops the sustained allocations to zero bytes/s for middleware with Json TE headers (e.g. 320MB less per 1M requests); will also drop header string allocation for MVC etc by the same amount.

image

Which fits with the goals of aspnet applications being able to run under lower memory conditions (See: ["Proposal for .NET Core GC Support for Docker Limits"](https://github.com/dotnet/designs/blob/261698cb9c56bab0bc7b54b938fa097d729a144f/proposed/support-for-docker-limits.md))

image

This PR introduces `DisableStringReuse` which went `false` to reduces allocations for repeated request headers.

```csharp
public partial class KestrelServerOptions
{
/// <summary>
/// Gets or sets a value that controls whether the string values materialized
/// will be reused across requests; if they match, or if the strings will always be reallocated.
/// </summary>
/// <remarks>
/// Defaults to false.
/// </remarks>
public bool DisableStringReuse { get; set; } = false;
}
```

Pre
```
               Method |              |       Mean |        Op/s |   Gen 0 | Allocated |
--------------------- |------------- |-----------:|------------:|--------:|----------:|
 PlaintextTechEmpower |              |   650.1 ns | 1,538,199.2 |  8.3008 |    168 KB |
           LiveAspNet |              | 1,487.8 ns |   672,119.6 | 26.3672 |    504 KB |
```
Post
```
               Method | ReuseHeaders |       Mean |        Op/s |   Gen 0 | Allocated |
--------------------- |------------- |-----------:|------------:|--------:|----------:|
 PlaintextTechEmpower |        False |   593.3 ns | 1,685,440.5 |  7.3242 |    144 KB |
           LiveAspNet |        False | 1,092.5 ns |   915,301.9 | 22.4609 |    440 KB |
                      |              |            |             |         |           |
 PlaintextTechEmpower |         True |   526.0 ns | 1,901,193.1 |       - |       0 B |
           LiveAspNet |         True |   871.9 ns | 1,146,875.8 |       - |       0 B |
```
This change drops ~320MB of allocations for ~1M requests of non-pipelined Json TE benchmark
```
wrk -c 128 -t 4 -d 45 -H "Accept: application/json,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7" -H "Connection: keep-alive" http://127.0.0.1:5001
```

Addresses #8372

### DIFF
--- a/src/Servers/Kestrel/Core/ref/Microsoft.AspNetCore.Server.Kestrel.Core.netcoreapp3.0.cs
+++ b/src/Servers/Kestrel/Core/ref/Microsoft.AspNetCore.Server.Kestrel.Core.netcoreapp3.0.cs
@@ -121,6 +121,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core
         public Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.Internal.SchedulingMode ApplicationSchedulingMode { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public System.IServiceProvider ApplicationServices { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public Microsoft.AspNetCore.Server.Kestrel.KestrelConfigurationLoader ConfigurationLoader { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
+        public bool DisableStringReuse { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServerLimits Limits { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
         public Microsoft.AspNetCore.Server.Kestrel.KestrelConfigurationLoader Configure() { throw null; }
         public Microsoft.AspNetCore.Server.Kestrel.KestrelConfigurationLoader Configure(Microsoft.Extensions.Configuration.IConfiguration config) { throw null; }
--- a/src/Servers/Kestrel/Core/ref/Microsoft.AspNetCore.Server.Kestrel.Core.netcoreapp3.0.cs
+++ b/src/Servers/Kestrel/Core/ref/Microsoft.AspNetCore.Server.Kestrel.Core.netcoreapp3.0.cs
@@ -122,6 +122,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core
         public System.IServiceProvider ApplicationServices { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public Microsoft.AspNetCore.Server.Kestrel.KestrelConfigurationLoader ConfigurationLoader { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServerLimits Limits { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
+        public bool ReuseRequestHeaders { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public Microsoft.AspNetCore.Server.Kestrel.KestrelConfigurationLoader Configure() { throw null; }
         public Microsoft.AspNetCore.Server.Kestrel.KestrelConfigurationLoader Configure(Microsoft.Extensions.Configuration.IConfiguration config) { throw null; }
         public void ConfigureEndpointDefaults(System.Action<Microsoft.AspNetCore.Server.Kestrel.Core.ListenOptions> configureOptions) { }
@@ -251,6 +252,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
     public partial interface IHttpHeadersHandler
     {
         void OnHeader(System.Span<byte> name, System.Span<byte> value);
+        void OnHeadersComplete();
     }
     public partial interface IHttpParser<TRequestHandler> where TRequestHandler : Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.IHttpHeadersHandler, Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.IHttpRequestLineHandler
     {
--- a/src/Servers/Kestrel/Core/ref/Microsoft.AspNetCore.Server.Kestrel.Core.netcoreapp3.0.cs
+++ b/src/Servers/Kestrel/Core/ref/Microsoft.AspNetCore.Server.Kestrel.Core.netcoreapp3.0.cs
@@ -122,7 +122,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core
         public System.IServiceProvider ApplicationServices { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public Microsoft.AspNetCore.Server.Kestrel.KestrelConfigurationLoader ConfigurationLoader { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServerLimits Limits { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
-        public bool ReuseRequestHeaders { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public Microsoft.AspNetCore.Server.Kestrel.KestrelConfigurationLoader Configure() { throw null; }
         public Microsoft.AspNetCore.Server.Kestrel.KestrelConfigurationLoader Configure(Microsoft.Extensions.Configuration.IConfiguration config) { throw null; }
         public void ConfigureEndpointDefaults(System.Action<Microsoft.AspNetCore.Server.Kestrel.Core.ListenOptions> configureOptions) { }
--- a/src/Servers/Kestrel/Core/src/Internal/Http/Http1Connection.cs
+++ b/src/Servers/Kestrel/Core/src/Internal/Http/Http1Connection.cs
@@ -307,7 +307,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                     }
                     else
                     {
-                        // Same as pervious
+                        // Same as previous
                         QueryString = _parsedQueryString;
                     }
 
--- a/src/Servers/Kestrel/Core/src/Internal/Http/Http1Connection.cs
+++ b/src/Servers/Kestrel/Core/src/Internal/Http/Http1Connection.cs
@@ -19,6 +19,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
         private const byte ByteAsterisk = (byte)'*';
         private const byte ByteForwardSlash = (byte)'/';
         private const string Asterisk = "*";
+        private const string ForwardSlash = "/";
 
         private readonly HttpConnectionContext _context;
         private readonly IHttpParser<Http1ParsingHandler> _parser;
@@ -268,6 +269,19 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 
             _requestTargetForm = HttpRequestTarget.OriginForm;
 
+            if (target.Length == 1)
+            {
+                // If target.Length == 1 it can only be a forward slash (e.g. home page)
+                // and we know RawTarget and Path are the same and QueryString is Empty
+                RawTarget = ForwardSlash;
+                Path = ForwardSlash;
+                QueryString = string.Empty;
+                // Clear parsedData as we won't check it if we come via this path again,
+                // an setting to null is fast as it doesn't need to use a GC write barrier.
+                _parsedRawTarget = _parsedPath = _parsedQueryString = null;
+                return;
+            }
+
             // URIs are always encoded/escaped to ASCII https://tools.ietf.org/html/rfc3986#page-11
             // Multibyte Internationalized Resource Identifiers (IRIs) are first converted to utf8;
             // then encoded/escaped to ASCII  https://www.ietf.org/rfc/rfc3987.txt "Mapping of IRIs to URIs"
@@ -275,9 +289,45 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             try
             {
                 // Read raw target before mutating memory.
-                RawTarget = target.GetAsciiStringNonNullCharacters();
-                QueryString = query.GetAsciiStringNonNullCharacters();
-                Path = PathNormalizer.DecodePath(path, pathEncoded, RawTarget, query.Length);
+                var previousValue = _parsedRawTarget;
+                if (previousValue == null || previousValue.Length != target.Length ||
+                    !StringUtilities.BytesOrdinalEqualsStringAndAscii(previousValue, target))
+                {
+                    // The previous string does not match what the bytes would convert to,
+                    // so we will need to generate a new string.
+                    RawTarget = _parsedRawTarget = target.GetAsciiStringNonNullCharacters();
+
+                    previousValue = _parsedQueryString;
+                    if (previousValue == null || previousValue.Length != query.Length ||
+                        !StringUtilities.BytesOrdinalEqualsStringAndAscii(previousValue, query))
+                    {
+                        // The previous string does not match what the bytes would convert to,
+                        // so we will need to generate a new string.
+                        QueryString = _parsedQueryString = query.GetAsciiStringNonNullCharacters();
+                    }
+                    else
+                    {
+                        // Same as pervious
+                        QueryString = _parsedQueryString;
+                    }
+
+                    if (path.Length == 1)
+                    {
+                        // If path.Length == 1 it can only be a forward slash (e.g. home page)
+                        Path = _parsedPath = ForwardSlash;
+                    }
+                    else
+                    {
+                        Path = _parsedPath = PathNormalizer.DecodePath(path, pathEncoded, RawTarget, query.Length);
+                    }
+                }
+                else
+                {
+                    // As RawTarget is the same we can reuse the previous parsed values.
+                    RawTarget = _parsedRawTarget;
+                    Path = _parsedPath;
+                    QueryString = _parsedQueryString;
+                }
             }
             catch (InvalidOperationException)
             {
@@ -312,9 +362,25 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             //
             // Allowed characters in the 'host + port' section of authority.
             // See https://tools.ietf.org/html/rfc3986#section-3.2
-            RawTarget = target.GetAsciiStringNonNullCharacters();
+            var previousValue = _parsedRawTarget;
+            if (previousValue == null || previousValue.Length != target.Length ||
+                !StringUtilities.BytesOrdinalEqualsStringAndAscii(previousValue, target))
+            {
+                // The previous string does not match what the bytes would convert to,
+                // so we will need to generate a new string.
+                RawTarget = _parsedRawTarget = target.GetAsciiStringNonNullCharacters();
+            }
+            else
+            {
+                // Reuse previous value
+                RawTarget = _parsedRawTarget;
+            }
+
             Path = string.Empty;
             QueryString = string.Empty;
+            // Clear parsedData for path and queryString as we won't check it if we come via this path again,
+            // an setting to null is fast as it doesn't need to use a GC write barrier.
+            _parsedPath = _parsedQueryString = null;
         }
 
         private void OnAsteriskFormTarget(HttpMethod method)
@@ -331,6 +397,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             RawTarget = Asterisk;
             Path = string.Empty;
             QueryString = string.Empty;
+            // Clear parsedData as we won't check it if we come via this path again,
+            // an setting to null is fast as it doesn't need to use a GC write barrier.
+            _parsedRawTarget = _parsedPath = _parsedQueryString = null;
         }
 
         private void OnAbsoluteFormTarget(Span<byte> target, Span<byte> query)
@@ -346,21 +415,42 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             //    a server MUST accept the absolute-form in requests, even though
             //    HTTP/1.1 clients will only send them in requests to proxies.
 
-            RawTarget = target.GetAsciiStringNonNullCharacters();
-
-            // Validation of absolute URIs is slow, but clients
-            // should not be sending this form anyways, so perf optimization
-            // not high priority
-
-            if (!Uri.TryCreate(RawTarget, UriKind.Absolute, out var uri))
+            var previousValue = _parsedRawTarget;
+            if (previousValue == null || previousValue.Length != target.Length ||
+                !StringUtilities.BytesOrdinalEqualsStringAndAscii(previousValue, target))
             {
-                ThrowRequestTargetRejected(target);
-            }
+                // The previous string does not match what the bytes would convert to,
+                // so we will need to generate a new string.
+                RawTarget = _parsedRawTarget = target.GetAsciiStringNonNullCharacters();
 
-            _absoluteRequestTarget = uri;
-            Path = uri.LocalPath;
-            // don't use uri.Query because we need the unescaped version
-            QueryString = query.GetAsciiStringNonNullCharacters();
+                // Validation of absolute URIs is slow, but clients
+                // should not be sending this form anyways, so perf optimization
+                // not high priority
+
+                if (!Uri.TryCreate(RawTarget, UriKind.Absolute, out var uri))
+                {
+                    ThrowRequestTargetRejected(target);
+                }
+
+                _absoluteRequestTarget = uri;
+                Path = _parsedPath = uri.LocalPath;
+                // don't use uri.Query because we need the unescaped version
+                previousValue = _parsedQueryString;
+                if (previousValue == null || previousValue.Length != query.Length ||
+                    !StringUtilities.BytesOrdinalEqualsStringAndAscii(previousValue, query))
+                {
+                    // The previous string does not match what the bytes would convert to,
+                    // so we will need to generate a new string.
+                    QueryString = _parsedQueryString = query.GetAsciiStringNonNullCharacters();
+                }
+            }
+            else
+            {
+                // As RawTarget is the same we can reuse the previous values.
+                RawTarget = _parsedRawTarget;
+                Path = _parsedPath;
+                QueryString = _parsedQueryString;
+            }
         }
 
         internal void EnsureHostHeaderExists()
--- a/src/Servers/Kestrel/Core/src/Internal/Http/Http1Connection.cs
+++ b/src/Servers/Kestrel/Core/src/Internal/Http/Http1Connection.cs
@@ -288,9 +288,11 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 
             try
             {
+                var disableStringReuse = ServerOptions.DisableStringReuse;
                 // Read raw target before mutating memory.
                 var previousValue = _parsedRawTarget;
-                if (DisableStringReuse || previousValue == null || previousValue.Length != target.Length ||
+                if (disableStringReuse ||
+                    previousValue == null || previousValue.Length != target.Length ||
                     !StringUtilities.BytesOrdinalEqualsStringAndAscii(previousValue, target))
                 {
                     // The previous string does not match what the bytes would convert to,
@@ -298,7 +300,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                     RawTarget = _parsedRawTarget = target.GetAsciiStringNonNullCharacters();
 
                     previousValue = _parsedQueryString;
-                    if (DisableStringReuse || previousValue == null || previousValue.Length != query.Length ||
+                    if (disableStringReuse ||
+                        previousValue == null || previousValue.Length != query.Length ||
                         !StringUtilities.BytesOrdinalEqualsStringAndAscii(previousValue, query))
                     {
                         // The previous string does not match what the bytes would convert to,
@@ -362,8 +365,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             //
             // Allowed characters in the 'host + port' section of authority.
             // See https://tools.ietf.org/html/rfc3986#section-3.2
+
             var previousValue = _parsedRawTarget;
-            if (DisableStringReuse || previousValue == null || previousValue.Length != target.Length ||
+            if (ServerOptions.DisableStringReuse ||
+                previousValue == null || previousValue.Length != target.Length ||
                 !StringUtilities.BytesOrdinalEqualsStringAndAscii(previousValue, target))
             {
                 // The previous string does not match what the bytes would convert to,
@@ -415,8 +420,10 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             //    a server MUST accept the absolute-form in requests, even though
             //    HTTP/1.1 clients will only send them in requests to proxies.
 
+            var disableStringReuse = ServerOptions.DisableStringReuse;
             var previousValue = _parsedRawTarget;
-            if (DisableStringReuse || previousValue == null || previousValue.Length != target.Length ||
+            if (disableStringReuse ||
+                previousValue == null || previousValue.Length != target.Length ||
                 !StringUtilities.BytesOrdinalEqualsStringAndAscii(previousValue, target))
             {
                 // The previous string does not match what the bytes would convert to,
@@ -436,7 +443,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                 Path = _parsedPath = uri.LocalPath;
                 // don't use uri.Query because we need the unescaped version
                 previousValue = _parsedQueryString;
-                if (DisableStringReuse || previousValue == null || previousValue.Length != query.Length ||
+                if (disableStringReuse ||
+                    previousValue == null || previousValue.Length != query.Length ||
                     !StringUtilities.BytesOrdinalEqualsStringAndAscii(previousValue, query))
                 {
                     // The previous string does not match what the bytes would convert to,
--- a/src/Servers/Kestrel/Core/src/Internal/Http/Http1Connection.cs
+++ b/src/Servers/Kestrel/Core/src/Internal/Http/Http1Connection.cs
@@ -290,7 +290,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             {
                 // Read raw target before mutating memory.
                 var previousValue = _parsedRawTarget;
-                if (previousValue == null || previousValue.Length != target.Length ||
+                if (DisableStringReuse || previousValue == null || previousValue.Length != target.Length ||
                     !StringUtilities.BytesOrdinalEqualsStringAndAscii(previousValue, target))
                 {
                     // The previous string does not match what the bytes would convert to,
@@ -298,7 +298,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                     RawTarget = _parsedRawTarget = target.GetAsciiStringNonNullCharacters();
 
                     previousValue = _parsedQueryString;
-                    if (previousValue == null || previousValue.Length != query.Length ||
+                    if (DisableStringReuse || previousValue == null || previousValue.Length != query.Length ||
                         !StringUtilities.BytesOrdinalEqualsStringAndAscii(previousValue, query))
                     {
                         // The previous string does not match what the bytes would convert to,
@@ -363,7 +363,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             // Allowed characters in the 'host + port' section of authority.
             // See https://tools.ietf.org/html/rfc3986#section-3.2
             var previousValue = _parsedRawTarget;
-            if (previousValue == null || previousValue.Length != target.Length ||
+            if (DisableStringReuse || previousValue == null || previousValue.Length != target.Length ||
                 !StringUtilities.BytesOrdinalEqualsStringAndAscii(previousValue, target))
             {
                 // The previous string does not match what the bytes would convert to,
@@ -416,7 +416,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             //    HTTP/1.1 clients will only send them in requests to proxies.
 
             var previousValue = _parsedRawTarget;
-            if (previousValue == null || previousValue.Length != target.Length ||
+            if (DisableStringReuse || previousValue == null || previousValue.Length != target.Length ||
                 !StringUtilities.BytesOrdinalEqualsStringAndAscii(previousValue, target))
             {
                 // The previous string does not match what the bytes would convert to,
@@ -436,12 +436,16 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                 Path = _parsedPath = uri.LocalPath;
                 // don't use uri.Query because we need the unescaped version
                 previousValue = _parsedQueryString;
-                if (previousValue == null || previousValue.Length != query.Length ||
+                if (DisableStringReuse || previousValue == null || previousValue.Length != query.Length ||
                     !StringUtilities.BytesOrdinalEqualsStringAndAscii(previousValue, query))
                 {
                     // The previous string does not match what the bytes would convert to,
                     // so we will need to generate a new string.
                     QueryString = _parsedQueryString = query.GetAsciiStringNonNullCharacters();
+                }
+                else
+                {
+                    QueryString = _parsedQueryString;
                 }
             }
             else
--- a/src/Servers/Kestrel/Core/src/Internal/Http/HttpHeaders.Generated.cs
+++ b/src/Servers/Kestrel/Core/src/Internal/Http/HttpHeaders.Generated.cs
@@ -6,16 +6,16 @@ using System.Collections.Generic;
 using System.Buffers;
 using System.IO.Pipelines;
 using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
 using Microsoft.Extensions.Primitives;
 using Microsoft.Net.Http.Headers;
+using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 {
 
     internal partial class HttpRequestHeaders
     {
-
-        private long _bits = 0;
         private HeaderReferences _headers;
 
         public bool HasConnection => (_bits & 0x2L) != 0;
@@ -703,12 +703,46 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                 _headers._UserAgent = value; 
             }
         }
-        public StringValues HeaderOrigin
+        public StringValues HeaderDNT
         {
             get
             {
                 StringValues value = default;
                 if ((_bits & 0x10000000000L) != 0)
+                {
+                    value = _headers._DNT;
+                }
+                return value;
+            }
+            set
+            {
+                _bits |= 0x10000000000L;
+                _headers._DNT = value; 
+            }
+        }
+        public StringValues HeaderUpgradeInsecureRequests
+        {
+            get
+            {
+                StringValues value = default;
+                if ((_bits & 0x20000000000L) != 0)
+                {
+                    value = _headers._UpgradeInsecureRequests;
+                }
+                return value;
+            }
+            set
+            {
+                _bits |= 0x20000000000L;
+                _headers._UpgradeInsecureRequests = value; 
+            }
+        }
+        public StringValues HeaderOrigin
+        {
+            get
+            {
+                StringValues value = default;
+                if ((_bits & 0x40000000000L) != 0)
                 {
                     value = _headers._Origin;
                 }
@@ -716,7 +750,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             }
             set
             {
-                _bits |= 0x10000000000L;
+                _bits |= 0x40000000000L;
                 _headers._Origin = value; 
             }
         }
@@ -725,7 +759,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             get
             {
                 StringValues value = default;
-                if ((_bits & 0x20000000000L) != 0)
+                if ((_bits & 0x80000000000L) != 0)
                 {
                     value = _headers._AccessControlRequestMethod;
                 }
@@ -733,7 +767,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             }
             set
             {
-                _bits |= 0x20000000000L;
+                _bits |= 0x80000000000L;
                 _headers._AccessControlRequestMethod = value; 
             }
         }
@@ -742,7 +776,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             get
             {
                 StringValues value = default;
-                if ((_bits & 0x40000000000L) != 0)
+                if ((_bits & 0x100000000000L) != 0)
                 {
                     value = _headers._AccessControlRequestHeaders;
                 }
@@ -750,7 +784,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             }
             set
             {
-                _bits |= 0x40000000000L;
+                _bits |= 0x100000000000L;
                 _headers._AccessControlRequestHeaders = value; 
             }
         }
@@ -932,7 +966,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                         }
                         if ("Origin".Equals(key, StringComparison.OrdinalIgnoreCase))
                         {
-                            if ((_bits & 0x10000000000L) != 0)
+                            if ((_bits & 0x40000000000L) != 0)
                             {
                                 value = _headers._Origin;
                                 return true;
@@ -1019,6 +1053,15 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                             if ((_bits & 0x100L) != 0)
                             {
                                 value = _headers._Via;
+                                return true;
+                            }
+                            return false;
+                        }
+                        if ("DNT".Equals(key, StringComparison.OrdinalIgnoreCase))
+                        {
+                            if ((_bits & 0x10000000000L) != 0)
+                            {
+                                value = _headers._DNT;
                                 return true;
                             }
                             return false;
@@ -1227,11 +1270,24 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                         }
                     }
                     break;
+                case 25:
+                    {
+                        if ("Upgrade-Insecure-Requests".Equals(key, StringComparison.OrdinalIgnoreCase))
+                        {
+                            if ((_bits & 0x20000000000L) != 0)
+                            {
+                                value = _headers._UpgradeInsecureRequests;
+                                return true;
+                            }
+                            return false;
+                        }
+                    }
+                    break;
                 case 29:
                     {
                         if ("Access-Control-Request-Method".Equals(key, StringComparison.OrdinalIgnoreCase))
                         {
-                            if ((_bits & 0x20000000000L) != 0)
+                            if ((_bits & 0x80000000000L) != 0)
                             {
                                 value = _headers._AccessControlRequestMethod;
                                 return true;
@@ -1244,7 +1300,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                     {
                         if ("Access-Control-Request-Headers".Equals(key, StringComparison.OrdinalIgnoreCase))
                         {
-                            if ((_bits & 0x40000000000L) != 0)
+                            if ((_bits & 0x100000000000L) != 0)
                             {
                                 value = _headers._AccessControlRequestHeaders;
                                 return true;
@@ -1368,7 +1424,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                         }
                         if ("Origin".Equals(key, StringComparison.OrdinalIgnoreCase))
                         {
-                            _bits |= 0x10000000000L;
+                            _bits |= 0x40000000000L;
                             _headers._Origin = value;
                             return;
                         }
@@ -1430,6 +1486,12 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                         {
                             _bits |= 0x100L;
                             _headers._Via = value;
+                            return;
+                        }
+                        if ("DNT".Equals(key, StringComparison.OrdinalIgnoreCase))
+                        {
+                            _bits |= 0x10000000000L;
+                            _headers._DNT = value;
                             return;
                         }
                     }
@@ -1581,11 +1643,21 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                         }
                     }
                     break;
+                case 25:
+                    {
+                        if ("Upgrade-Insecure-Requests".Equals(key, StringComparison.OrdinalIgnoreCase))
+                        {
+                            _bits |= 0x20000000000L;
+                            _headers._UpgradeInsecureRequests = value;
+                            return;
+                        }
+                    }
+                    break;
                 case 29:
                     {
                         if ("Access-Control-Request-Method".Equals(key, StringComparison.OrdinalIgnoreCase))
                         {
-                            _bits |= 0x20000000000L;
+                            _bits |= 0x80000000000L;
                             _headers._AccessControlRequestMethod = value;
                             return;
                         }
@@ -1595,7 +1667,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                     {
                         if ("Access-Control-Request-Headers".Equals(key, StringComparison.OrdinalIgnoreCase))
                         {
-                            _bits |= 0x40000000000L;
+                            _bits |= 0x100000000000L;
                             _headers._AccessControlRequestHeaders = value;
                             return;
                         }
@@ -1776,9 +1848,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                         }
                         if ("Origin".Equals(key, StringComparison.OrdinalIgnoreCase))
                         {
-                            if ((_bits & 0x10000000000L) == 0)
+                            if ((_bits & 0x40000000000L) == 0)
                             {
-                                _bits |= 0x10000000000L;
+                                _bits |= 0x40000000000L;
                                 _headers._Origin = value;
                                 return true;
                             }
@@ -1872,6 +1944,16 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                             {
                                 _bits |= 0x100L;
                                 _headers._Via = value;
+                                return true;
+                            }
+                            return false;
+                        }
+                        if ("DNT".Equals(key, StringComparison.OrdinalIgnoreCase))
+                        {
+                            if ((_bits & 0x10000000000L) == 0)
+                            {
+                                _bits |= 0x10000000000L;
+                                _headers._DNT = value;
                                 return true;
                             }
                             return false;
@@ -2097,13 +2179,27 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                         }
                     }
                     break;
-                case 29:
+                case 25:
                     {
-                        if ("Access-Control-Request-Method".Equals(key, StringComparison.OrdinalIgnoreCase))
+                        if ("Upgrade-Insecure-Requests".Equals(key, StringComparison.OrdinalIgnoreCase))
                         {
                             if ((_bits & 0x20000000000L) == 0)
                             {
                                 _bits |= 0x20000000000L;
+                                _headers._UpgradeInsecureRequests = value;
+                                return true;
+                            }
+                            return false;
+                        }
+                    }
+                    break;
+                case 29:
+                    {
+                        if ("Access-Control-Request-Method".Equals(key, StringComparison.OrdinalIgnoreCase))
+                        {
+                            if ((_bits & 0x80000000000L) == 0)
+                            {
+                                _bits |= 0x80000000000L;
                                 _headers._AccessControlRequestMethod = value;
                                 return true;
                             }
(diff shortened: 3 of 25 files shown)


src/Servers/Kestrel/Core/src/Internal/Http/HttpRequestHeaders.cs
```diff
@@ -24,33 +65,82 @@ private static long ParseContentLength(string value)
...
-        private void SetValueUnknown(string key, in StringValues value)
+        private void AppendContentLength(Span<byte> value)
         {
-            Unknown[key] = value;
+            if (_contentLength.HasValue)
+            {
+                BadHttpRequestException.Throw(RequestRejectionReason.MultipleContentLengths);
+            }
+
+
```
davidfowl: nit: Extra space

src/Servers/Kestrel/Core/src/Internal/Http/HttpRequestHeaders.cs
```diff
@@ -24,33 +65,82 @@ private static long ParseContentLength(string value)
...
+                BadHttpRequestException.Throw(RequestRejectionReason.InvalidContentLength, value.GetAsciiOrUTF8StringNonNullCharacters());
+            }
+
+            _contentLength = parsed;
         }
 
-        public unsafe void Append(Span<byte> name, string value)
+        private static unsafe bool TryParseNonNegativeInt64(Span<byte> value, out long result)
         {
-            fixed (byte* namePtr = name)
```
davidfowl: Should we use the Utf8Parser here?

src/Servers/Kestrel/Core/src/Internal/Http/HttpRequestHeaders.cs
```diff
@@ -24,33 +65,82 @@ private static long ParseContentLength(string value)
...
+        {
+            Unknown[key] = value;
         }
 
         [MethodImpl(MethodImplOptions.NoInlining)]
-        private unsafe void AppendUnknownHeaders(byte* pKeyBytes, int keyLength, string value)
+        private unsafe void AppendUnknownHeaders(Span<byte> name, Span<byte> value)
         {
-            string key = new string('\0', keyLength);
+            string key = new string('\0', name.Length);
```
davidfowl: string.Create?

src/Servers/Kestrel/Core/ref/Microsoft.AspNetCore.Server.Kestrel.Core.netcoreapp3.0.cs
```diff
@@ -723,7 +727,7 @@ public abstract partial class HttpProtocol : Microsoft.AspNetCore.Http.Features.
     }
     public sealed partial class HttpRequestHeaders : Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpHeaders
     {
-        public HttpRequestHeaders() { }
+        public HttpRequestHeaders(Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServerOptions serverOptions) { }
```
davidfowl: Just pass a bool? I'm not sure we should pass all the options (though I know where could be more in the future...)

src/Servers/Kestrel/Core/src/Internal/Http/HttpRequestHeaders.cs
benaadams: Complicated as can't pass the Span as the `TState` so would still need to fix it to pass as `byte*`; then would need to fix the `Span<char>` in the `SpanAction` pass to `TryGetAsciiString` as it currently is; so would be a bigger change to address all that.

src/Servers/Kestrel/Core/ref/Microsoft.AspNetCore.Server.Kestrel.Core.netcoreapp3.0.cs
benaadams: Changed

src/Servers/Kestrel/Core/src/Internal/Http/HttpRequestHeaders.cs
benaadams: Not sure https://github.com/dotnet/corefx/issues/35814

benaadams:
Using a larger set of headers to represent a browser request
```
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8
Accept-Encoding: gzip, deflate, br
Accept-Language: en-US,en;q=0.9
Cache-Control: no-cache
Connection: keep-alive
Host: github.com
Pragma: no-cache
User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/72.0.3626.121 Safari/537.36
Cookie: prov=20629ccd-8b0f-e8ef-2935-cd26609fc0bc; __qca=P0-1591065732-1479167353442; _ga=GA1.2.1298898376.1479167354; _gat=1; sgt=id=9519gfde_3347_4762_8762_df51458c8ec2; acct=t=why-is-%e0%a5%a7%e0%a5%a8%e0%a5%a9-numeric&s=why-is-%e0%a5%a7%e0%a5%a8%e0%a5%a9-numeric
```
And a 2 minute run
```
wrk -c 128 -t 4 -d 120 -H "Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8" -H "Accept-Encoding: gzip, deflate, br" -H "Accept-Language: en-US,en;q=0.9" -H "Cache-Control: no-cache" -H "Connection: keep-alive" -H "Host: github.com" -H "Pragma: no-cache" -H "User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/72.0.3626.121 Safari/537.36" -H "Cookie: prov=20629ccd-8b0f-e8ef-2935-cd26609fc0bc; __qca=P0-1591065732-1479167353442; _ga=GA1.2.1298898376.1479167354; _gat=1; sgt=id=9519gfde_3347_4762_8762_df51458c8ec2; acct=t=why-is-%e0%a5%a7%e0%a5%a8%e0%a5%a9-numeric&s=why-is-%e0%a5%a7%e0%a5%a8%e0%a5%a9-numeric" -s ./wrk/scripts/pipeline.lua http://127.0.0.1:5001 -- 16
```

Before
```
Pipeline depth: 16
Running 2m test @ http://127.0.0.1:5001
  4 threads and 128 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     5.44ms    3.95ms  56.46ms   76.52%
    Req/Sec    46.44k     9.64k   91.47k    68.45%
  22152336 requests in 2.00m, 2.72GB read
Requests/sec: 184499.18
Transfer/sec:     23.23MB
```
28GB of headers are allocated

image

After
```
Pipeline depth: 16
Running 2m test @ http://127.0.0.1:5001
  4 threads and 128 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     4.77ms    3.48ms  54.37ms   76.96%
    Req/Sec    52.39k    10.31k  113.12k    68.83%
  25000032 requests in 2.00m, 3.07GB read
Requests/sec: 208147.81
Transfer/sec:     26.20MB
```
0.3MB of headers are allocated

image

25M requests in 2 minutes is probably quite a high load though :smile:

src/Servers/Kestrel/Core/src/Internal/Http/HttpRequestHeaders.cs
```diff
@@ -13,6 +13,46 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
...
+        protected override void ClearFast()
+        {
+            if (!_reuseHeaderValues)
+            {
+                // If we aren't reusing headers clear them all
+                Clear(_bits);
+            }
+            else
+            {
+                // If we are reusing headers, store the currently set headers for comparision later
```
jo-ninja: nit: spelling "comparision" => "comparison"
```suggestion
                // If we are reusing headers, store the currently set headers for comparison later
```

src/Servers/Kestrel/Core/src/Internal/Infrastructure/StringUtilities.cs
```diff
@@ -109,6 +112,127 @@ public static unsafe bool TryGetAsciiString(byte* input, char* output, int count
             return isValid;
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveOptimization)]
+        public unsafe static bool BytesOrdinalEqualsStringAndAscii(string previousValue, Span<byte> newValue)
+        {
+            // We just widen the bytes to char for comparision, if either the string or the bytes are not ascii
```
jo-ninja: nit: another "comparision" here (and some other lines below). Might need 'find & replace all' 😉

src/Servers/Kestrel/Core/src/Internal/Infrastructure/StringUtilities.cs
```diff
@@ -109,6 +112,127 @@ public static unsafe bool TryGetAsciiString(byte* input, char* output, int count
...
+            // this will result in non-equality, so we don't need to specifically test for non-ascii.
+            Debug.Assert(previousValue.Length == newValue.Length);
+
+            // Use IntPtr values rather than int, to avoid unnessary 32 -> 64 movs on 64-bit.
```
jo-ninja: nit: "unnessary" => "unnecessary "
```suggestion
            // Use IntPtr values rather than int, to avoid unnecessary 32 -> 64 movs on 64-bit.
```

src/Servers/Kestrel/Core/src/Internal/Http/HttpRequestHeaders.cs
benaadams: Done

src/Servers/Kestrel/Core/src/Internal/Infrastructure/StringUtilities.cs
benaadams: Done

src/Servers/Kestrel/Core/src/Internal/Infrastructure/StringUtilities.cs
benaadams: Went for search and replace :)

benaadams:
Updated perf test results
```
               Method | ReuseHeaders |       Mean |        Op/s |   Gen 0 | Allocated |
--------------------- |------------- |-----------:|------------:|--------:|----------:|
 PlaintextTechEmpower |        False |   676.1 ns | 1,479,039.7 |  8.3008 |    168 KB |
           LiveAspNet |        False | 1,423.3 ns |   702,610.6 | 25.3906 |    504 KB |
                      |              |            |             |         |           |
 PlaintextTechEmpower |         True |   588.4 ns | 1,699,502.5 |  0.4883 |     24 KB |
           LiveAspNet |         True | 1,233.7 ns |   810,542.5 |  2.9297 |     88 KB |
```

benaadams:
Now faster than before when not reusing headers, and faster still when reusing headers
```
               Method | ReuseHeaders |       Mean |        Op/s |   Gen 0 | Allocated |
--------------------- |------------- |-----------:|------------:|--------:|----------:|
 PlaintextTechEmpower |        False |   629.3 ns | 1,589,177.2 |  8.7891 |    168 KB |
           LiveAspNet |        False | 1,357.9 ns |   736,414.0 | 26.3672 |    504 KB |
                      |              |            |             |         |           |
 PlaintextTechEmpower |         True |   552.5 ns | 1,809,801.6 |  0.9766 |     24 KB |
           LiveAspNet |         True | 1,180.8 ns |   846,871.4 |  3.9063 |     88 KB |
```

mattnischan:
Any chance someone with less than altruistic intent could alter something like the "Accept" header value for everyone with some unsafe antics?

benaadams:
> Any chance someone with less than altruistic intent could alter something like the "Accept" header value for everyone with some unsafe antics?

Sort of, it would only change it for the request (the values are only reused on a single connection); and the next request wouldn't match via equality so it would generate a new one; but then you could change it again. 

However; that's a lot of work, since you can happily just replace the value in the headers collection without unsafe.

mattnischan:
Ah OK, I was missing the part where the string is only reused for a connection, not across the application.

benaadams:
>  was missing the part where the string is only reused for a connection, not across the application.

Avoids slow down from contention 😉

benaadams:
```
               Method | ReuseHeaders |       Mean |        Op/s |   Gen 0 | Allocated |
--------------------- |------------- |-----------:|------------:|--------:|----------:|
 PlaintextTechEmpower |        False |   583.3 ns | 1,714,322.7 |  7.3242 |    144 KB |
           LiveAspNet |        False | 1,275.7 ns |   783,867.2 | 24.4141 |    492 KB |
                      |              |            |             |         |           |
 PlaintextTechEmpower |         True |   523.0 ns | 1,912,075.5 |       - |       0 B |
           LiveAspNet |         True | 1,158.9 ns |   862,855.6 |  1.9531 |     76 KB |
```

benaadams:
```
               Method | ReuseHeaders |       Mean |        Op/s |   Gen 0 | Allocated |
--------------------- |------------- |-----------:|------------:|--------:|----------:|
 PlaintextTechEmpower |        False |   593.3 ns | 1,685,440.5 |  7.3242 |  147456 B |
           LiveAspNet |        False | 1,092.5 ns |   915,301.9 | 22.4609 |  450560 B |
                      |              |            |             |         |           |
 PlaintextTechEmpower |         True |   526.0 ns | 1,901,193.1 |       - |       0 B |
           LiveAspNet |         True |   871.9 ns | 1,146,875.8 |       - |       0 B |
```

Tratcher:
Would it be more productive / less dangerous to focus on the HTTP/2 HPACK scenario where this is a first class protocol concept? Or is that already performing well?

sebastienros:
All ran 5 times and excluded the highest and lowest result.

### Json (+2%)

```
| Description |     RPS | CPU (%) | Memory (MB) | Avg. Latency (ms) | Startup (ms) | First Request (ms) | Latency (ms) | Ratio |
| ----------- | ------- | ------- | ----------- | ----------------- | ------------ | ------------------ | ------------ | ----- |
|             | 413,056 |      87 |         157 |              5.72 |          207 |              67.89 |         0.47 |  1.00 |
|         ben | 422,409 |      86 |         156 |              4.77 |          322 |             138.26 |         0.48 |  1.02 |
```

```
Headers["Accept"] = "application/json,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7";
Headers["Connection"] = "keep-alive";
```

### Plaintext (+7%)

```
| Description |       RPS | CPU (%) | Memory (MB) | Avg. Latency (ms) | Startup (ms) | First Request (ms) | Latency (ms) | Ratio |
| ----------- | --------- | ------- | ----------- | ----------------- | ------------ | ------------------ | ------------ | ----- |
|             | 2,327,521 |      93 |         161 |              6.13 |          210 |              37.37 |         0.65 |  1.00 |
|         ben | 2,485,687 |      94 |         161 |              2.98 |          318 |               80.3 |         0.45 |  1.07 |
```

```
Headers["Accept"] = "text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7";
Headers["Connection"] = "keep-alive";
```

jkotalik:
@davidfowl @DamianEdwards do we want this in for preview4?

src/Servers/Kestrel/Core/src/Internal/Infrastructure/StringUtilities.cs
```diff
@@ -109,6 +112,127 @@ public static unsafe bool TryGetAsciiString(byte* input, char* output, int count
             return isValid;
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveOptimization)]
+        public unsafe static bool BytesOrdinalEqualsStringAndAscii(string previousValue, Span<byte> newValue)
```
halter73: Are there plans to add anything to BCL for efficiently comparing a string to an ASCII-encoded span?

benaadams: Not currently when I asked. It would likely need to be a Utf8 to `string` compare to fit in with any current apis, however that's a lot more complicated. 

Not sure where a simpler ascii-to-string Ordinal only would fit to make an api proposal; as it would likely also need IgnoreCase and CultureAware variants to match with the kind of apis corefx would want; and then its suddenly a lot more complicated and a lot of code for probably a niche case.

/cc @GrabYourPitchforks

benaadams:
> Would it be more productive / less dangerous to focus on the HTTP/2 HPACK scenario where this is a first class protocol concept? Or is that already performing well?

It will improve HTTP/2 also as HTTP/2 uses 2 header collections, these collections being one of them.

src/Servers/Kestrel/Core/src/Internal/Http/Http1Connection.cs
```diff
@@ -346,7 +373,14 @@ private void OnAbsoluteFormTarget(Span<byte> target, Span<byte> query)
             //    a server MUST accept the absolute-form in requests, even though
             //    HTTP/1.1 clients will only send them in requests to proxies.
 
-            RawTarget = target.GetAsciiStringNonNullCharacters();
+            var previousValue = RawTarget;
+            if (previousValue == null || previousValue.Length != target.Length ||
+                !StringUtilities.BytesOrdinalEqualsStringAndAscii(previousValue, target))
+            {
+                // The previous string does not matche what the bytes would convert to,
```
benaadams: nit: matche -> match

src/Servers/Kestrel/Core/src/Internal/Http/Http1Connection.cs
```diff
@@ -360,7 +394,14 @@ private void OnAbsoluteFormTarget(Span<byte> target, Span<byte> query)
             _absoluteRequestTarget = uri;
             Path = uri.LocalPath;
             // don't use uri.Query because we need the unescaped version
-            QueryString = query.GetAsciiStringNonNullCharacters();
+            previousValue = QueryString;
+            if (previousValue == null || previousValue.Length != query.Length ||
+                !StringUtilities.BytesOrdinalEqualsStringAndAscii(previousValue, query))
+            {
+                // The previous string does not matche what the bytes would convert to,
```
benaadams: nit: matche -> match

src/Servers/Kestrel/Core/src/Internal/Infrastructure/StringUtilities.cs
GrabYourPitchforks: We did just introduce a type `Utf8` into the `System.Text.Unicode` namespace, and we have plans to add APIs like `ToUpper` to that type. Maybe we'd also put an `Ascii` type in that same namespace? (Yuck.)

benaadams: Only want something equivalent to `string.CompareOrdinal(ReadOnlySpan<byte> asciiBytes)`

Could also go with `Utf8.CompareOrdinal(string, ReadOnlySpan<byte>)` if it was fast in the ascii-only path; Utf8 headers should be rare and is little value in optimizing for them; although I assume corefx would care about the performance of the Utf8 path also since it would be a more generic method.

benaadams:
Ready for review

benaadams:
Tests written, ready for review

benaadams:
Improved some comments and rebased into the "Improve comments" commit

src/Servers/Kestrel/Core/src/Internal/Http/HttpRequestHeaders.cs
benaadams: Also mostly a copy from [aspnet/AspNetCore/src/Http/Headers/src/HeaderUtilities.cs](https://github.com/aspnet/AspNetCore/blob/285110de91f276d0b9302c3d2a84f0dc5af861b1/src/Http/Headers/src/HeaderUtilities.cs#L422-L456) but changed to operate over a `Span<byte>` rather than a `StringSegment`.

benaadams:
1.6M requests over 128 connections
```
wrk -c 128 -t 4 -d 35 -H "Accept: application/json,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7" -H "Connection: keep-alive" http://127.0.0.1:5001
Running 35s test @ http://127.0.0.1:5001
  4 threads and 128 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     2.31ms    3.13ms 102.69ms   94.62%
    Req/Sec    11.61k     3.90k   23.84k    71.18%
  1613343 requests in 35.10s, 203.10MB read
Requests/sec:  45962.35
Transfer/sec:      5.79MB
```
Allocations 
image


Previous for 960k requests over 128 connections

image

src/Servers/Kestrel/Core/src/Internal/Http/HttpRequestHeaders.cs
davidfowl: We should add an overload in the same place:

```c#
public static unsafe bool TryParseNonNegativeInt64(StringSegment value, out long result);
public static unsafe bool TryParseNonNegativeInt64(ReadOnlySpan<byte> span, out long result);
```

benaadams: Moved

src/Servers/Kestrel/Core/src/Internal/Infrastructure/StringUtilities.cs
```diff
@@ -109,6 +113,222 @@ public static unsafe bool TryGetAsciiString(byte* input, char* output, int count
...
+            } while ((byte*)offset < (byte*)count);
+
+            // If we get here (input is exactly a multiple of Vector length) then there are no inequalities via widening;
+            // so the input bytes are both ascii and a match to the string if it was converted via Encoding.ASCII.GetString(...)
+            return true;
+        NotEqual:
+            return false;
+        }
+
+        private unsafe static bool IsValidHeaderString(string value)
```
GrabYourPitchforks: This method is only called within `Debug.Assert` and doesn't need to be optimized. The entire method can be implemented as below:

```cs
try {
  if (value is null) return false;
  new Utf8Encoding(throw: true).GetByteCount(value);
  return true;
} catch (...) {
  return false;
}
```

src/Servers/Kestrel/Core/src/Internal/Infrastructure/StringUtilities.cs
davidfowl: @GrabYourPitchforks can we find a home for this? It would be great if this didn't have to live in Kestrel 😄

GrabYourPitchforks: Open an issue in the corefx repo with what you want the API to be. :)

In general I'm pessimistic about a general-purpose API that compares UTF-8 and UTF-16 directly, as the performance of such a thing would be abysmal except for all-ASCII input. And if you're looking for an _ASCII-specific_ API, definitely call out that specific scenario when you open the issue.

davidfowl: @benaadams can you file an API request?

src/Servers/Kestrel/Core/src/Internal/Infrastructure/StringUtilities.cs
benaadams: Done

src/Servers/Kestrel/Core/src/Internal/Infrastructure/StringUtilities.cs
benaadams: https://github.com/dotnet/corefx/issues/36519

GrabYourPitchforks (changes requested):

src/Http/Headers/ref/Microsoft.Net.Http.Headers.netcoreapp3.0.cs
```diff
@@ -204,6 +204,7 @@ public static partial class HeaderUtilities
         public static bool TryParseDate(Microsoft.Extensions.Primitives.StringSegment input, out System.DateTimeOffset result) { throw null; }
         public static bool TryParseNonNegativeInt32(Microsoft.Extensions.Primitives.StringSegment value, out int result) { throw null; }
         public static bool TryParseNonNegativeInt64(Microsoft.Extensions.Primitives.StringSegment value, out long result) { throw null; }
+        public static bool TryParseNonNegativeInt64(System.ReadOnlySpan<byte> asciiBytes, out long result) { throw null; }
```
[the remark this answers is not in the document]
davidfowl: Yes.

src/Servers/Kestrel/Core/src/Internal/Http/HttpRequestHeaders.cs
```diff
@@ -24,33 +68,42 @@ private static long ParseContentLength(string value)
...
+        {
+            Unknown[key] = value;
         }
 
         [MethodImpl(MethodImplOptions.NoInlining)]
-        private unsafe void AppendUnknownHeaders(byte* pKeyBytes, int keyLength, string value)
+        private unsafe void AppendUnknownHeaders(Span<byte> name, string valueString)
         {
-            string key = new string('\0', keyLength);
+            string key = new string('\0', name.Length);
```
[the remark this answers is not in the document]
GrabYourPitchforks: (I know this is existing code. It doesn't need to be addressed as part of this PR, but it should be addressed at some point.)

src/Http/Headers/src/HeaderUtilities.cs
```diff
@@ -455,6 +455,57 @@ public static unsafe bool TryParseNonNegativeInt64(StringSegment value, out long
...
+        /// <param name="asciiBytes">The Ascii bytes to parse</param>
+        /// <param name="result">
+        /// When this method returns, contains the 64-bit signed integer value equivalent of the number contained
+        /// in the string, if the conversion succeeded, or zero if the conversion failed. The conversion fails if
+        /// <paramref name="asciiBytes"/> is of zero length, is not of the correct format, is negative, or
+        /// represents a number greater than Int64.MaxValue. This parameter is passed uninitialized; any value
+        /// originally supplied in result will be overwritten.
+        /// </param>
+        /// <returns><code>true</code> if parsing succeeded; otherwise, <code>false</code>.</returns>
+        public static unsafe bool TryParseNonNegativeInt64(ReadOnlySpan<byte> asciiBytes, out long result)
```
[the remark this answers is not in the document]
davidfowl: See https://github.com/aspnet/AspNetCore/pull/8374#discussion_r264014520

src/Http/Headers/ref/Microsoft.Net.Http.Headers.netcoreapp3.0.cs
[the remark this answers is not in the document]
benaadams: Was asked to move it to that repo https://github.com/aspnet/AspNetCore/pull/8374#discussion_r270695664 in order to then use it in Kestrel it needs to be public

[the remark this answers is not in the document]
davidfowl: It's like the other parsers on this type. In theory these wouldn't exist really.

src/Servers/Kestrel/Core/src/Internal/Http/HttpRequestHeaders.cs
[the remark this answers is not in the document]
benaadams: Is a messy area https://github.com/aspnet/AspNetCore/pull/8374#discussion_r264014541 and should be picked up in a follow up PR as its just what it does currently. Equally the headers should be passed as `ReadOnlySpan<byte>` rather than `Span<byte>` and certainly before this interface is made public public rather than pubternal
```csharp
public partial interface IHttpHeadersHandler
{
    void OnHeader(System.Span<byte> name, System.Span<byte> value);
    void OnHeadersComplete();
}
```

src/Http/Headers/src/HeaderUtilities.cs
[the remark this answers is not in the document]
benaadams: Would be
```csharp
if (!Utf8Parser.TryParse(ReadOnlySpan<byte> source, out long parsed, out int, char) || parsed < 0)
{
    BadHttpRequestException.Throw(RequestRejectionReason.InvalidContentLength, value.GetAsciiOrUTF8StringNonNullCharacters());
}
```
Which is 
switch(standardFormat) -> TryParseInt64D(source, out value, out bytesConsumed) -> ...

Ends up being quite a heavy weight path when the numbers are commonly 4 digits or less?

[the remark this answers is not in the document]
GrabYourPitchforks: The in-box implementation is already fairly lean, and I'd be shocked if it's the bottleneck in your scenarios. And if it is the bottleneck then the onus is on us (the BCL team) to fix it. We shouldn't have variations of this method propagated throughout first-party shipping code without good reason.

src/Servers/Kestrel/Core/src/Internal/Infrastructure/StringUtilities.cs
```diff
@@ -109,6 +113,179 @@ public static unsafe bool TryGetAsciiString(byte* input, char* output, int count
...
+                    // Read a Vector length from the input as bytes
+                    var vector = Unsafe.ReadUnaligned<Vector<byte>>(ref Unsafe.Add(ref bytes, offset));
+                    // Widen the bytes directly to chars (ushort) as if they were ascii.
+                    // As widening doubles the size we get two vectors back.
+                    Vector.Widen(vector, out var vector0, out var vector1);
+                    // Read two char vectors from the string to perform the match.
+                    var compare0 = Unsafe.ReadUnaligned<Vector<ushort>>(ref Unsafe.As<char, byte>(ref Unsafe.Add(ref str, offset)));
+                    var compare1 = Unsafe.ReadUnaligned<Vector<ushort>>(ref Unsafe.As<char, byte>(ref Unsafe.Add(ref str, offset + Vector<ushort>.Count)));
+
+                    // If the string is not ascii, then the widened bytes cannot match
```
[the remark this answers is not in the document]
benaadams: Shame, will fix.

src/Http/Headers/src/HeaderUtilities.cs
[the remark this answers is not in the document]
benaadams: Will change

src/Servers/Kestrel/Core/src/Internal/Infrastructure/StringUtilities.cs
```diff
@@ -109,6 +113,179 @@ public static unsafe bool TryGetAsciiString(byte* input, char* output, int count
...
+                        }
+                    }
+
+                    // End of input reached, there are no inequalities via widening; so the input bytes are both ascii
+                    // and a match to the string if it was converted via Encoding.ASCII.GetString(...)
+                    return true;
+                }
+
+                // Create a comparision vector for all bits being equal
+                var AllTrue = new Vector<ushort>(ushort.MaxValue);
```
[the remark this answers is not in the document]
benaadams: Does seem to be 1 i.e. returning 1,1,1, ... rather than 0xffff, 0xffff, 0xffff, ...

image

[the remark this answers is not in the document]
benaadams: Whereas MaxValue is
image

src/Servers/Kestrel/Core/src/Internal/Http/HttpHeaders.Generated.cs
```diff
@@ -3462,780 +3593,410 @@ protected override bool CopyToFast(KeyValuePair<string, StringValues>[] array, i
...
+        {
+            ref byte nameStart = ref MemoryMarshal.GetReference(name);
+            ref StringValues values = ref Unsafe.AsRef<StringValues>(null);
+            var flag = 0L;
+
+            // Does the name matched any "known" headers
+            switch (name.Length)
+            {
+                case 2:
+                    if (((Unsafe.Add(ref Unsafe.As<byte, ushort>(ref nameStart), 0) & 0xdfdfu) == 0x4554u))
```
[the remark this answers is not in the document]
benaadams: Not too worried about Endianness [as Kestrel will throw on startup](https://github.com/aspnet/AspNetCore/blob/ff313901de48b1318cf6951181f9a32ee82d8753/src/Servers/Kestrel/Core/src/KestrelServer.cs#L125-L128) if its BigEndian
```csharp
if (!BitConverter.IsLittleEndian)
{
    throw new PlatformNotSupportedException(CoreStrings.BigEndianNotSupported);
}
```
Can't really use `BinaryPrimitives` as its all span based and the amount of slicing that would occur in this method and the tmps it would cause the Jit to choke https://github.com/aspnet/AspNetCore/pull/7724#issuecomment-465375561, without making it a really weird arrangement, extra overheads aside.

Will change to ReadUnaligned though.

src/Http/Headers/ref/Microsoft.Net.Http.Headers.netcoreapp3.0.cs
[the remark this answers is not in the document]
benaadams: Removed

src/Http/Headers/src/HeaderUtilities.cs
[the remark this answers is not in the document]
benaadams: Changed

src/Http/Headers/src/HeaderUtilities.cs
```diff
@@ -455,6 +455,57 @@ public static unsafe bool TryParseNonNegativeInt64(StringSegment value, out long
...
+            {
+                var digit = (long)asciiBytes[i] - (long)0x30;
+                if ((ulong)digit > 9)
+                {
+                    // Not a digit
+                    break;
+                }
+
+                calc = calc * 10 + digit;
+                if (calc < 0)
```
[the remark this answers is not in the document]
benaadams: Removed this method

src/Servers/Kestrel/Core/src/Internal/Infrastructure/StringUtilities.cs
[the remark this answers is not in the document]
benaadams: Left as `new Vector<ushort>(ushort.MaxValue)` as need all the bits set

src/Servers/Kestrel/Core/src/Internal/Http/HttpHeaders.Generated.cs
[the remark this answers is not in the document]
benaadams: Changed to ReadUnaligned; main change in `KnownHeaders.cs` in this commit https://github.com/aspnet/AspNetCore/pull/8374/commits/805505185c5b38db4659d284173b3560ea0f097d which then generates the `HttpHeaders.Generated.cs` file in same commit.

src/Servers/Kestrel/Core/src/Internal/Infrastructure/StringUtilities.cs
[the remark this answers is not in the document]
benaadams: Fixed, but forget to add test for this example. Will Add.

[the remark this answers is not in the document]
benaadams: Added test

src/Servers/Kestrel/Core/src/Internal/Http/HttpRequestHeaders.cs
benaadams: https://github.com/aspnet/AspNetCore/issues/8978

src/Servers/Kestrel/Core/src/Internal/Http/HttpRequestHeaders.cs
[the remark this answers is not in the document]
benaadams: https://github.com/aspnet/AspNetCore/issues/8978 for string

https://github.com/aspnet/AspNetCore/issues/8979 for interfaces

benaadams:
@GrabYourPitchforks have your issues been addressed?

benaadams:
Ubuntu error unrelated
```
Microsoft.DotNet.Watcher.Tools.FunctionalTests.GlobbingAppTests.RenameCompiledFile

System.Threading.Tasks.TaskCanceledException : A task was canceled.

Stack trace
   at Microsoft.DotNet.Watcher.Tools.FunctionalTests.AwaitableProcess.GetOutputLineAsync(String predicateName, Predicate`1 predicate, CancellationToken cancellationToken) in /_/src/Tools/dotnet-watch/test/AwaitableProcess.cs:line 100
   at Microsoft.DotNet.Watcher.Tools.FunctionalTests.AwaitableProcess.GetOutputLineAsync(String message, TimeSpan timeout) in /_/src/Tools/dotnet-watch/test/AwaitableProcess.cs:line 85
   at Microsoft.DotNet.Watcher.Tools.FunctionalTests.GlobbingAppTests.RenameCompiledFile() in /_/src/Tools/dotnet-watch/test/GlobbingAppTests.cs:line 86
```

benaadams:
macOS failure Microsoft.DotNet.Watcher.Tools.FunctionalTests.GlobbingAppTests.RenameCompiledFile https://github.com/aspnet/AspNetCore/issues/8987

src/Servers/Kestrel/Core/src/Internal/Http/Http1Connection.cs
```diff
@@ -275,9 +275,29 @@ private void OnOriginFormTarget(HttpMethod method, HttpVersion version, Span<byt
             try
             {
                 // Read raw target before mutating memory.
-                RawTarget = target.GetAsciiStringNonNullCharacters();
-                QueryString = query.GetAsciiStringNonNullCharacters();
-                Path = PathNormalizer.DecodePath(path, pathEncoded, RawTarget, query.Length);
```
Tratcher: You've got to admit this one is artificial. What scenarios besides benchmarks hit the same URL back-to-back at any scale sufficient for this optimization to be relevant?

benaadams: Lots of users hitting F5 on an overloaded webserver

benaadams: Lots of users hitting F5 on an overloaded webserver; get latest news on a site not doing real-time, to check stock on a sold out item hoping it comes back, trying to book tickets for a popular gig that's just gone on sale, etc

davidfowl: It's fine as long as it doesn't regress performance, that's why the optimizing compare is important.

Tratcher: It has increased complexity and thus reduced maintainability.

Even when you mash F5 it's going to make several requests for different resources, and spread them across different connections. You'll rarely get the same request back-to-back.

benaadams: >  You'll rarely get the same request back-to-back.

Will do if your resources are in CDN; they'll be on a different domain, different "servers"

benaadams: >  You'll rarely get the same request back-to-back.

Will do if your resources are in CDN; they'll be on a different domain, different "servers"

Example: for the url https://www.ageofascent.com/2019/02/04/asp-net-core-saturating-10gbe-at-7-million-requests-per-second/ even though there are 87 requests that make up the page, there is always only a single request to `www.ageofascent.com` which contains the page body, everything else is from other domains.

davidfowl: It’s fine. This doesn’t cross the line for me and doesn’t  make the code that much worse. IMO The part of this change that is concerning is the byte comparsion.

benaadams: > What scenarios besides benchmarks hit the same URL back-to-back at any scale sufficient for this optimization to be relevant?

Long polling uses same path and multiple requests, range requests use the same url and changes the Range header; REST apis based on POST body values, e.g. GraphQL uses a single url and then changes body data; Webhooks will typically POST to a fixed URL, etc.

Although; you raise an important gap, the most common request of a website is for the homepage, that being OriginForm `/`; so I have added an optimization to use an interned string of ForwardSlash even on the first request; where there has been no prior request.

jkotalik:
Note, there is a failing IIS test (which uses kestrel) that is failing from your change consistently. It is a test that sends multiple Hello World requests and verifies the responses are all the same. Test that is failing: https://github.com/aspnet/AspNetCore/blob/834910964737ebbeaed45a965acd124adb59ab46/src/Servers/IIS/IIS/test/Common.FunctionalTests/CommonStartupTests.cs#L30 and helix failure: https://mc.dot.net/#/user/aspnetcore/pr~2Faspnet~2Faspnetcore/ci/20190403.56/workItem/IISExpress.FunctionalTests-netcoreapp3.0/wilogs

benaadams:
Any reason its not showing in the CI checks?

benaadams:
Rebased to investigate

benaadams:
> Note, there is a failing IIS test (which uses kestrel) that is failing from your change consistently. It is a test that sends multiple Hello World requests and verifies the responses are all the same.

Ahh, think I know what this is. 

Currently am assuming if `RawTarget` is the same then `Path` and `Query` are unchanged. However, if something has changed them in the pipeline, like splitting `Path` into `PathBase` then its been modified and that's not a correct assumption.

To take advantage of the cached value and cope with modifications (as the modified value will never match); then will need to store to a parsed value, and test against that.

benaadams:
Still good with the additional header validation and alternate cached vars
```
               Method | ReuseHeaders |       Mean |        Op/s | Scaled | Allocated |
--------------------- |------------- |-----------:|------------:|-------:|----------:|
 PlaintextTechEmpower |        False |   595.3 ns | 1,679,956.8 |   1.00 |    144 KB |
           LiveAspNet |        False | 1,028.1 ns |   972,674.0 |   1.73 |    440 KB |
                      |              |            |             |        |           |
 PlaintextTechEmpower |         True |   516.2 ns | 1,937,167.1 |   1.00 |       0 B |
           LiveAspNet |         True |   849.6 ns | 1,177,080.0 |   1.65 |       0 B |
```

src/Servers/Kestrel/Core/src/KestrelServerOptions.cs
```diff
@@ -54,6 +54,15 @@ public class KestrelServerOptions
...
         public bool AllowSynchronousIO { get; set; } = false;
 
+        /// <summary>
+        /// Gets or sets a value that controls whether the header values materialized as strings will be reused across requests;
+        /// if they match, or if the strings will always be reallocated.
+        /// </summary>
+        /// <remarks>
+        /// Defaults to true.
+        /// </remarks>
+        public bool ReuseRequestHeaders { get; set; } = true;
```
Tratcher: This option is only being provided to disable the feature if something breaks, right? In that case should it be an AppContext switch rather than an API?

benaadams: I was thinking more if your org had some paranoia constraints; since headers contain the cookie and auth headers. However, that should be more unusual than the norm.

How would I set it up as an AppContext switch? Are there examples?

Tratcher: https://github.com/aspnet/AspNetCore/wiki/Quirks-mode-switches-in-patches

benaadams:
This improves HTTP/1.1 and provides some benefit to HTTP/2; however there are more improves for HTTP/2 that can be done (path also for gRPC); extra known header types e.g. `:authority` as well as looking at the HPackDecoder, but I'll look at those in a follow up

src/Servers/Kestrel/Core/src/KestrelServerOptions.cs
benaadams: > Quirks mode switches are currently used only in patches. Quirks mode switches are not used in major/minor versions.

Could just get rid of the option?

Tratcher:
Yeah, for HTTP/2 I'd expect this kind of optimization to be built directly into the HPackDecoder. You don't have anywhere else to cache at the moment.

src/Servers/Kestrel/Core/src/KestrelServerOptions.cs
Tratcher: Depends on our confidence. Might do keep the switch for a few previews.

benaadams:
> Yeah, for HTTP/2 I'd expect this kind of optimization to be built directly into the HPackDecoder. You don't have anywhere else to cache at the moment.

Still adds the byte headers to the regular RequestHeaders dictionary

https://github.com/aspnet/AspNetCore/blob/eaf63a670dde3cbbbaa73fb88653f8e197fbe966/src/Servers/Kestrel/Core/src/Internal/Http2/HPack/HPackDecoder.cs#L338

https://github.com/aspnet/AspNetCore/blob/eaf63a670dde3cbbbaa73fb88653f8e197fbe966/src/Servers/Kestrel/Core/src/Internal/Http2/HPack/HPackDecoder.cs#L349

as these then call though to `HttpProtocol.OnHeader`, so HTTP/2 does still benefit from the saving on string conversion

benaadams:
> Note, there is a failing IIS test (which uses kestrel) that is failing from your change consistently.

Fixed now and passing in Helix.

Will add the AppContext switch

benaadams:
Need to resolve merge conflict

benaadams:
Squashed and rebased

src/Servers/Kestrel/Core/src/Internal/Http/HttpHeaders.Generated.cs
```diff
@@ -3464,778 +3595,407 @@ protected override bool CopyToFast(KeyValuePair<string, StringValues>[] array, i
...
+                    array[arrayIndex] = new KeyValuePair<string, StringValues>("Content-Length", HeaderUtilities.FormatNonNegativeInt64(_contentLength.Value));
+                    ++arrayIndex;
+                }
+            ((ICollection<KeyValuePair<string, StringValues>>)MaybeUnknown)?.CopyTo(array, arrayIndex);
+
+            return true;
+        }
+        
+        [MethodImpl(MethodImplOptions.AggressiveOptimization)]
+        public unsafe void Append(Span<byte> name, Span<byte> value)
```
GrabYourPitchforks: All of this unsafe logic assumes a little-endian machine. I _think_ that's a safe assumption at the moment, and I wouldn't hold up this PR for it, but it should eventually be addressed.

src/Servers/Kestrel/Core/src/Internal/Http/HttpHeaders.Generated.cs
```diff
@@ -3464,778 +3595,407 @@ protected override bool CopyToFast(KeyValuePair<string, StringValues>[] array, i
...
+                        values = ref _headers._Origin;
+                    }
+                    else if ((firstTerm6 == 0x47415250u) && ((Unsafe.ReadUnaligned<ushort>(ref Unsafe.AddByteOffset(ref nameStart, (IntPtr)(2 * sizeof(ushort)))) & 0xdfdfu) == 0x414du))
+                    {
+                        flag = 0x10L;
+                        values = ref _headers._Pragma;
+                    }
+                    break;
+                case 7:
+                    var firstTerm7 = (Unsafe.ReadUnaligned<uint>(ref nameStart) & 0xdfdfdfdfu);
```
GrabYourPitchforks: Potential optimization: Don't be afraid to perform overlapping reads. You can read the first 4 bytes as a DWORD and the last 4 bytes as a DWORD. Sure, there's going to be one overlapping byte, but the fact that fewer overall operations are necessary should still result in a net perf boost.

src/Servers/Kestrel/Core/src/Internal/Infrastructure/StringUtilities.cs
```diff
@@ -109,6 +115,257 @@ public static unsafe bool TryGetAsciiString(byte* input, char* output, int count
...
+                }
+            }
+        }
+
+        /// <summary>
+        /// Given a WORD which represents a buffer of 2 bytes, widens the buffer into 2 WORDs and
+        /// compares them to the WORD buffer with machine endianness.
+        /// </summary>
+        [MethodImpl(MethodImplOptions.AggressiveInlining | MethodImplOptions.AggressiveOptimization)]
+        private static bool WidenTwoAsciiBytesToUtf16AndComapreToChars(ref char charStart, ushort value)
```
GrabYourPitchforks: Typo: _WidenTwoAsciiBytesToUtf16AndComapreToChars_

src/Servers/Kestrel/Core/src/Internal/Infrastructure/StringUtilities.cs
```diff
@@ -109,6 +115,257 @@ public static unsafe bool TryGetAsciiString(byte* input, char* output, int count
...
+        NotEqual:
+            return false;
+        }
+
+        /// <summary>
+        /// Given a DWORD which represents a buffer of 4 bytes, widens the buffer into 4 WORDs and
+        /// compares them to the WORD buffer with machine endianness.
+        /// </summary>
+        [MethodImpl(MethodImplOptions.AggressiveInlining | MethodImplOptions.AggressiveOptimization)]
+        private static bool WidenFourAsciiBytesToUtf16AndComapreToChars(ref char charStart, uint value)
```
GrabYourPitchforks: Typo: _WidenFourAsciiBytesToUtf16AndComapreToChars_

src/Servers/Kestrel/Core/src/Internal/Infrastructure/StringUtilities.cs
```diff
@@ -109,6 +115,257 @@ public static unsafe bool TryGetAsciiString(byte* input, char* output, int count
...
+                    return true;
+                }
+
+                // Create a comparision vector for all bits being equal
+                var AllTrue = new Vector<ushort>(ushort.MaxValue);
+                // do/while as entry condition already checked, remaining length must be Vector<byte>.Count or larger.
+                do
+                {
+                    // Read a Vector length from the input as bytes
+                    var vector = Unsafe.ReadUnaligned<Vector<byte>>(ref Unsafe.Add(ref bytes, offset));
```
GrabYourPitchforks: This is still missing the check for "was the input all-ASCII?" The easiest way to do this would be to interpret the byte input as a `Vector<sbyte>` and to ask "is any element less than 0?" See also https://github.com/dotnet/coreclr/blob/56043b6c896417e792e64a69692fed3251ed3adc/src/System.Private.CoreLib/shared/System/Text/ASCIIUtility.cs#L154 for inspiration.

GrabYourPitchforks (changes requested):

GrabYourPitchforks:
FYI the only thing that's "must address" from my most recent PR comments is the missing "any non-ASCII bytes?" check in the `Vector<T>` code paths. Everything else is just fixing typos / things to address later / whatever.

src/Servers/Kestrel/Core/src/Internal/Http/HttpHeaders.Generated.cs
benaadams: Is very safe (currently) as StartAsync will throw if its not:

https://github.com/aspnet/AspNetCore/blob/b44680762ef34c91f5fe0d411ddd8fe0b4e3d06b/src/Servers/Kestrel/Core/src/KestrelServer.cs#L121-L128

src/Servers/Kestrel/Core/src/Internal/Http/HttpHeaders.Generated.cs
benaadams: Will do in follow up

src/Servers/Kestrel/Core/src/Internal/Infrastructure/StringUtilities.cs
benaadams: Done

src/Servers/Kestrel/Core/src/Internal/Infrastructure/StringUtilities.cs
benaadams: Done

src/Servers/Kestrel/Core/src/Internal/Infrastructure/StringUtilities.cs
benaadams: Done; had the method just didn't plumb it in O_o

benaadams:
@GrabYourPitchforks feedback addressed?

src/Servers/Kestrel/Core/src/Internal/Http/HttpHeaders.Generated.cs
GrabYourPitchforks: Thanks for double-checking. :)

GrabYourPitchforks (approved):

anurse:
Let's either merge this this morning or move it to preview 5 (and merge it like, tomorrow ;)). It looks like a good performance win.

anurse:
Sounds like we need a little more review here and can't squeeze it in to preview 4. Moving to preview 5. Let's try to close this out in the next day or so though (modulo any delays due to branching).

src/Servers/Kestrel/Core/src/Internal/Http/Http1Connection.cs
```diff
@@ -268,16 +269,65 @@ private void OnOriginFormTarget(HttpMethod method, HttpVersion version, Span<byt
...
+            if (target.Length == 1)
+            {
+                // If target.Length == 1 it can only be a forward slash (e.g. home page)
+                // and we know RawTarget and Path are the same and QueryString is Empty
+                RawTarget = ForwardSlash;
+                Path = ForwardSlash;
+                QueryString = string.Empty;
+                // Clear parsedData as we won't check it if we come via this path again,
+                // an setting to null is fast as it doesn't need to use a GC write barrier.
+                _parsedRawTarget = _parsedPath = _parsedQueryString = null;
```
halter73: would it be wrong not to clear these?

src/Servers/Kestrel/Core/src/Internal/Http/Http1Connection.cs
```diff
@@ -268,16 +269,65 @@ private void OnOriginFormTarget(HttpMethod method, HttpVersion version, Span<byt
...
+                    if (previousValue == null || previousValue.Length != query.Length ||
+                        !StringUtilities.BytesOrdinalEqualsStringAndAscii(previousValue, query))
+                    {
+                        // The previous string does not match what the bytes would convert to,
+                        // so we will need to generate a new string.
+                        QueryString = _parsedQueryString = query.GetAsciiStringNonNullCharacters();
+                    }
+                    else
+                    {
+                        // Same as pervious
```
halter73: nit:

```suggestion
                        // Same as previous
```

src/Servers/Kestrel/Core/src/Internal/Http/Http1Connection.cs
```diff
@@ -346,21 +415,42 @@ private void OnAbsoluteFormTarget(Span<byte> target, Span<byte> query)
...
+                Path = _parsedPath = uri.LocalPath;
+                // don't use uri.Query because we need the unescaped version
+                previousValue = _parsedQueryString;
+                if (previousValue == null || previousValue.Length != query.Length ||
+                    !StringUtilities.BytesOrdinalEqualsStringAndAscii(previousValue, query))
+                {
+                    // The previous string does not match what the bytes would convert to,
+                    // so we will need to generate a new string.
+                    QueryString = _parsedQueryString = query.GetAsciiStringNonNullCharacters();
+                }
```
halter73: Shouldn't this have an else with `QueryString = _parsedQueryString;` like in OnOriginFormTarget? If not, should we remove the else from OnOriginFormTarget?

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
```diff
@@ -1490,5 +1504,15 @@ public async ValueTask<FlushResult> WriteAsyncAwaited(Task initializeTask, ReadO
                 return await Output.FlushAsync(cancellationToken);
             }
         }
+
+        private static bool GetDisableHeaderReuseSwitch()
+        {
+            if (AppContext.TryGetSwitch(DisableHeaderReuseSwitch, out var disableHeaderReuse))
```
halter73: Is there a switch to disable the similar changes to the target parsing?

src/Servers/Kestrel/Core/src/Internal/Http/Http1Connection.cs
benaadams: 🤔 I suppose it would just match to the previous request that wasn't a `/` next time; so no.

May be useful for a homepage that is `/` + one request (e.g. service worker).

Happy to take direction here.

src/Servers/Kestrel/Core/test/HttpRequestHeadersTests.cs
```diff
@@ -3,11 +3,13 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Text;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http;
 using Microsoft.Extensions.Primitives;
 using Xunit;
+using static CodeGenerator.KnownHeaders;
```
halter73: Is this being used? Does KnownHeaders still need to be in the shared folder?

src/Servers/Kestrel/shared/KnownHeaders.cs
```diff
@@ -196,6 +93,8 @@ public static string GeneratedFile()
                 "TE",
                 "Translate",
                 "User-Agent",
+                "DNT",
+                "Upgrade-Insecure-Requests"
```
halter73: :+1: Any other headers we should consider adding to the common headers list while we're thinking about it? @Tratcher

src/Servers/Kestrel/shared/KnownHeaders.cs
```diff
@@ -213,8 +112,6 @@ public static string GeneratedFile()
                 PrimaryHeader = requestPrimaryHeaders.Contains("Content-Length")
             }})
             .ToArray();
-            Debug.Assert(requestHeaders.Length <= 64);
-            Debug.Assert(requestHeaders.Max(x => x.Index) <= 62);
```
halter73: Why did these Debug.Asserts get removed? It looks like we're still using `KnownHeader.TestBit()`, so I think it's still a good check to have.

src/Servers/Kestrel/shared/KnownHeaders.cs
benaadams: I'd like to add a few more from https://github.com/aspnet/AspNetCore/blob/master/src/Http/Headers/src/HeaderNames.cs including the http/2 ones in that list of `:authority`,`:method`,`:path`, etc

Which was also part of my motivation for halving the size of `StringValues` https://github.com/aspnet/Extensions/pull/1283; so adding more headers wouldn't be too egregious.

However, was also looking to do that in a follow up, rather than this PR

src/Servers/Kestrel/Core/test/HttpRequestHeadersTests.cs
benaadams: Yes; is used in `[MemberData(nameof(KnownRequestHeaders))]` to check all the `KnownRequestHeaders` work as expected.

src/Servers/Kestrel/Core/src/Internal/Http/Http1Connection.cs
halter73: I'm fine leaving it as is. Better to let the GC collect the strings as soon as we know we'll no longer use them. I just wanted to make sure my understanding was correct.

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
benaadams: No. I saw it more as a paranoia setting https://github.com/aspnet/AspNetCore/pull/8374#discussion_r272666738 as the headers could contain the cookies/auth info (whereas you shouldn't be putting this in the path).

Also the alternate path for startline would be quite messy. Can add it though?

halter73 (approved):
I would like OnOriginFormTarget and OnAbsoluteTarget to be consistent regarding whether it explicitly resets QueryString to _parsedQueryString. I don't think it matters whether they do or don't, but they should be the same.

src/Servers/Kestrel/Core/src/Internal/Http/Http1Connection.cs
benaadams: Yes

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
halter73: Is it really that messy? It is a little bit of paranoia, but I think having a switch that disable all reuse of previous materialize strings would relieve some of that paranoia.

src/Servers/Kestrel/shared/KnownHeaders.cs
halter73: Makes sense. We should probably add the http/2 headers separately.

src/Servers/Kestrel/shared/KnownHeaders.cs
benaadams: Moved it to the start of the `GeneratedFile()` method line 527; also added check for responseHeaders.

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
benaadams: When I was originally thinking about how to do it; I thought it would be two branches with the two forms of parsing.

However, can just add it as first term in the `if` tests; so no it wouldn't be messy.

benaadams:
@halter73 feedback addressed?

benaadams:
Didn't mean to close :(

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
davidfowl: @benaadams we were discussion this and think it should be a regular option with, not an app context switch. It should just be a single option that can also be read from configuration (like we do for a bunch of other settings in kestrel)

benaadams: Added config rather than switch https://github.com/aspnet/AspNetCore/pull/8374/commits/b0dca95ad87dab6adabf7f3665f94cdf5e10bdeb

benaadams:
>  it should be a regular option with, not an app context switch

@davidfowl feedback addressed?  https://github.com/aspnet/AspNetCore/pull/8374/commits/b0dca95ad87dab6adabf7f3665f94cdf5e10bdeb

src/Servers/Kestrel/Core/src/Internal/Http/Http1Connection.cs
benaadams: Fixed

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
Tratcher: @davidfowl why the option? Is there any reason to turn this off other than to mitigate a bug in the implementation?

benaadams: From a runtime point of view the switch is faster since its a `readonly static` so branch eliminates one way or the other. The config gets evaluated on every iteration.

Tratcher: No need for that, it only needs to be evaluated once.

benaadams: Wondering if Options can be "baked" to `readonly static`s currently
e.g.
```csharp
if (buffer.Length >= ServerOptions.Limits.MaxRequestLineSize)
```
is 3 indirections
```asm
G_M62071_IG06:
       488B9780000000       mov      rdx, gword ptr [rdi+128]
       488B5218             mov      rdx, gword ptr [rdx+24]
       8B5220               mov      edx, dword ptr [rdx+32]
       4C63CA               movsxd   r9, edx
       493BC9               cmp      rcx, r9
       7C37                 jl       SHORT G_M62072_IG07
```
if it can be "baked" then it would just be a const compare

```asm
G_M62071_IG06:
       493BC9               cmp      rcx, 8192
       7C37                 jl       SHORT G_M62072_IG07
```

benaadams: Probably not as it would cause a problem if you were running more than once Kestrel instance in a process :-/

benaadams: Squished the api commits; just in case want it reverted https://github.com/aspnet/AspNetCore/pull/8374/commits/b822a0fc1027cda4c39948dbb2b05e73067adfd3

benaadams: A little difference; but not huge

Config
```
               Method |     Mean |        Op/s | Scaled | Allocated |
--------------------- |---------:|------------:|-------:|----------:|
 PlaintextTechEmpower | 541.7 ns | 1,846,001.2 |   1.00 |       0 B |
           LiveAspNet | 891.1 ns | 1,122,221.7 |   1.65 |       0 B |
```         
Switch
```
               Method |     Mean |        Op/s | Scaled | Allocated |
--------------------- |---------:|------------:|-------:|----------:|
 PlaintextTechEmpower | 536.9 ns | 1,862,399.7 |   1.00 |       0 B |
           LiveAspNet | 878.1 ns | 1,138,780.0 |   1.64 |       0 B |
```

halter73:
Thanks @benaadams!

benaadams:
😸 🚀 

image

PR #8415, opened 2019-03-11
Simplify formatting of "WWW-Authenticate" header's content

TechnoBerry:
 - Replace usage of `StringBuilder` to `string#Join`


Tratcher:
This defeats the point of using the string builder to avoid creating excess strings.

PR #8419, opened 2019-03-11
Fix AppOfflineDroppedWhileSiteStarting_SiteShutsDown_InProcess

jkotalik:
Fixes https://github.com/aspnet/AspNetCore/issues/6555
The reason an invalid return code was being thrown was due to an issue with our request draining logic which always caused the server to ungracefully shutdown. The return code was due to an invalid handle with console logging.

### DIFF
--- a/src/Servers/IIS/AspNetCoreModuleV2/AspNetCore/PollingAppOfflineApplication.cpp
+++ b/src/Servers/IIS/AspNetCoreModuleV2/AspNetCore/PollingAppOfflineApplication.cpp
@@ -17,6 +17,8 @@ HRESULT PollingAppOfflineApplication::TryCreateHandler(_In_ IHttpContext* pHttpC
 void
 PollingAppOfflineApplication::CheckAppOffline()
 {
+    SRWSharedLock stopLock(m_stateLock);
+
     if (m_fStopCalled)
     {
         return;
--- a/src/Servers/IIS/AspNetCoreModuleV2/AspNetCore/PollingAppOfflineApplication.cpp
+++ b/src/Servers/IIS/AspNetCoreModuleV2/AspNetCore/PollingAppOfflineApplication.cpp
@@ -17,8 +17,6 @@ HRESULT PollingAppOfflineApplication::TryCreateHandler(_In_ IHttpContext* pHttpC
 void
 PollingAppOfflineApplication::CheckAppOffline()
 {
-    SRWSharedLock stopLock(m_stateLock);
-
     if (m_fStopCalled)
     {
         return;
--- a/src/Servers/IIS/AspNetCoreModuleV2/CommonLib/application.h
+++ b/src/Servers/IIS/AspNetCoreModuleV2/CommonLib/application.h
@@ -23,22 +23,17 @@ public:
         _In_  IHttpContext       *pHttpContext,
         _Outptr_result_maybenull_ IREQUEST_HANDLER  **pRequestHandler) override
     {
-        auto waitForShutdown = false;
         *pRequestHandler = nullptr;
 
+        SRWSharedLock stateLock(m_stateLock);
+
         {
-            SRWSharedLock stopLock(m_stateLock);
+            SRWSharedLock stopLock(m_stopLock);
 
             if (m_fStopCalled)
             {
-                waitForShutdown = true;
+                return S_FALSE;
             }
-        }
-
-        if (waitForShutdown)
-        {
-            THROW_LAST_ERROR_IF(WaitForSingleObject(m_pFinishShutdownEvent, INFINITE) == WAIT_FAILED);
-            return S_FALSE;
         }
 
         TraceContextScope traceScope(pHttpContext->GetTraceContext());
@@ -60,34 +55,23 @@ public:
           m_applicationId(pHttpApplication.GetApplicationId())
     {
         InitializeSRWLock(&m_stateLock);
+        InitializeSRWLock(&m_stopLock);
         m_applicationVirtualPath = ToVirtualPath(m_applicationConfigPath);
-        THROW_LAST_ERROR_IF_NULL(m_pFinishShutdownEvent = CreateEvent(
-            nullptr,  // default security attributes
-            TRUE,     // manual reset event
-            FALSE,    // not set
-            nullptr)); // name
     }
 
     VOID
     Stop(bool fServerInitiated) override
     {
-        auto waitForShutdown = false;
+        SRWExclusiveLock stateLock(m_stateLock);
 
         {
-            SRWExclusiveLock stopLock(m_stateLock);
-
+            SRWSharedLock stopLock(m_stopLock);
             if (m_fStopCalled)
             {
-                waitForShutdown = true;
+                return;
             }
 
             m_fStopCalled = true;
-        }
-
-        if (waitForShutdown)
-        {
-            THROW_LAST_ERROR_IF(WaitForSingleObject(m_pFinishShutdownEvent, INFINITE) == WAIT_FAILED);
-            return;
         }
 
         StopInternal(fServerInitiated);
@@ -149,7 +133,8 @@ public:
     }
 
 protected:
-    SRWLOCK m_stateLock {};
+    SRWLOCK m_stateLock{};
+    SRWLOCK m_stopLock {};
     bool m_fStopCalled;
 
 private:
--- a/src/Servers/IIS/AspNetCoreModuleV2/CommonLib/application.h
+++ b/src/Servers/IIS/AspNetCoreModuleV2/CommonLib/application.h
@@ -75,11 +75,6 @@ public:
         }
 
         StopInternal(fServerInitiated);
-
-        if (m_pFinishShutdownEvent != nullptr)
-        {
-            LOG_IF_FAILED(SetEvent(m_pFinishShutdownEvent));
-        }
     }
 
     virtual
@@ -144,8 +139,6 @@ private:
     std::wstring m_applicationVirtualPath;
     std::wstring m_applicationConfigPath;
     std::wstring m_applicationId;
-
-    HandleWrapper<NullHandleTraits> m_pFinishShutdownEvent;
 
     static std::wstring ToVirtualPath(const std::wstring& configurationPath)
     {
--- a/src/Servers/IIS/AspNetCoreModuleV2/CommonLib/application.h
+++ b/src/Servers/IIS/AspNetCoreModuleV2/CommonLib/application.h
@@ -27,13 +27,11 @@ public:
 
         SRWSharedLock stopLock(m_stopLock);
 
+        // If we have acquired the stopLock, we don't need to acquire the data lock
+        // as m_fStoppedCalled is only set by Stop.
+        if (m_fStopCalled)
         {
-            SRWSharedLock dataLock(m_dataLock);
-
-            if (m_fStopCalled)
-            {
-                return S_FALSE;
-            }
+            return S_FALSE;
         }
 
         TraceContextScope traceScope(pHttpContext->GetTraceContext());
--- a/src/Servers/IIS/AspNetCoreModuleV2/CommonLib/application.h
+++ b/src/Servers/IIS/AspNetCoreModuleV2/CommonLib/application.h
@@ -57,14 +57,16 @@ public:
     VOID
     Stop(bool fServerInitiated) override
     {
-        SRWExclusiveLock stopLock(m_stateLock);
-
-        if (m_fStopCalled)
         {
-            return;
-        }
+            SRWExclusiveLock stopLock(m_stateLock);
 
-        m_fStopCalled = true;
+            if (m_fStopCalled)
+            {
+                return;
+            }
+
+            m_fStopCalled = true;
+        }
 
         StopInternal(fServerInitiated);
     }
--- a/src/Servers/IIS/AspNetCoreModuleV2/CommonLib/application.h
+++ b/src/Servers/IIS/AspNetCoreModuleV2/CommonLib/application.h
@@ -9,6 +9,7 @@
 #include "SRWExclusiveLock.h"
 #include "SRWSharedLock.h"
 #include "exceptions.h"
+#include "HandleWrapper.h"
 
 class APPLICATION : public IAPPLICATION
 {
@@ -22,12 +23,21 @@ public:
         _In_  IHttpContext       *pHttpContext,
         _Outptr_result_maybenull_ IREQUEST_HANDLER  **pRequestHandler) override
     {
+        auto waitForShutdown = false;
         *pRequestHandler = nullptr;
 
-        SRWSharedLock stopLock(m_stateLock);
-
-        if (m_fStopCalled)
         {
+            SRWSharedLock stopLock(m_stateLock);
+
+            if (m_fStopCalled)
+            {
+                waitForShutdown = true;
+            }
+        }
+
+        if (waitForShutdown)
+        {
+            THROW_LAST_ERROR_IF(WaitForSingleObject(m_pFinishShutdownEvent, INFINITE) == WAIT_FAILED);
             return S_FALSE;
         }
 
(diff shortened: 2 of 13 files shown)


src/Servers/IIS/AspNetCoreModuleV2/AspNetCore/PollingAppOfflineApplication.cpp
```diff
@@ -17,6 +17,8 @@ HRESULT PollingAppOfflineApplication::TryCreateHandler(_In_ IHttpContext* pHttpC
 void
 PollingAppOfflineApplication::CheckAppOffline()
 {
+    SRWSharedLock stopLock(m_stateLock);
```
pakrym: This would cause a deadlock. SRW is non-reentrant and non-upgradable Stop call might take a lock.

pakrym: This is also not required, `m_fStopCalled` is optimistic check it doesn't have to be correct every time.

jkotalik: > non-reentrant

My bad, didn't realize that.

src/Servers/IIS/AspNetCoreModuleV2/CommonLib/application.h
```diff
@@ -57,14 +57,16 @@ class APPLICATION : public IAPPLICATION
     VOID
     Stop(bool fServerInitiated) override
     {
-        SRWExclusiveLock stopLock(m_stateLock);
-
-        if (m_fStopCalled)
         {
-            return;
-        }
+            SRWExclusiveLock stopLock(m_stateLock);
```
pakrym: There was a reason we help lock for the entire duration of StopInternal... I wish I remembered why.

src/Servers/IIS/AspNetCoreModuleV2/InProcessRequestHandler/inprocessapplication.cpp
```diff
@@ -534,6 +540,9 @@ IN_PROCESS_APPLICATION::CreateHandler(
     {
         DBG_ASSERT(!m_fStopCalled);
         m_requestCount++;
+
+        LOG_INFOF(L"Adding request. Total Request Count %d", m_requestCount.load());
```
pakrym: LOG_TRACE, this is super spammy

src/Servers/IIS/AspNetCoreModuleV2/InProcessRequestHandler/inprocessapplication.cpp
```diff
@@ -341,9 +343,13 @@ IN_PROCESS_APPLICATION::ExecuteApplication()
 
 void IN_PROCESS_APPLICATION::QueueStop()
 {
-    if (m_fStopCalled)
     {
-        return;
+        SRWSharedLock lock(m_stateLock);
```
pakrym: Not required.

src/Servers/IIS/AspNetCoreModuleV2/InProcessRequestHandler/inprocessapplication.cpp
```diff
@@ -284,6 +285,7 @@ IN_PROCESS_APPLICATION::ExecuteApplication()
         // At this point CLR thread either finished or timed out, abandon it.
         m_clrThread.detach();
 
+        SRWSharedLock lock(m_stateLock);
```
pakrym: Not sure if this one is required.

src/Servers/IIS/AspNetCoreModuleV2/CommonLib/application.h
jkotalik: #YOLOSWAGGINZ

pakrym:
I also don't like that a flaky test is the only thing that caught this pretty common deadlock.

src/Servers/IIS/IIS/test/IIS.Shared.FunctionalTests/RequiresIISAttribute.cs
```diff
@@ -124,12 +124,6 @@ public RequiresIISAttribute(IISCapability capabilities)
                 }
             }
 
-            if (capabilities.HasFlag(IISCapability.ShutdownToken))
```
pakrym: Remove the capability.

src/Servers/IIS/AspNetCoreModuleV2/InProcessRequestHandler/inprocessapplication.cpp
```diff
@@ -545,8 +549,10 @@ void
 IN_PROCESS_APPLICATION::HandleRequestCompletion()
 {
     SRWSharedLock lock(m_stateLock);
-    auto requestCount = m_requestCount--;
-    if (m_fStopCalled && requestCount == 0)
+    m_requestCount--;
```
pakrym: Why did you separate this statement? You can end up calling callback multiple times.

src/Servers/IIS/AspNetCoreModuleV2/CommonLib/application.h
jkotalik: It was used for out of process. I did the minimal and re-added the lock around StopInternal().

pakrym: Now Stop retakes the lock twice for out of process. What was it used for there?

jkotalik: Ahh I now see why we locked the entire call to StopAsync. It makes it so it is guaranteed that the app can be replaced by the time the request returns 503. 

By having the lock in Stop, it means that the next call to TryCreateHandler will not finish until the app is shutdown, meaning the dotnet process has shut down by the time TryCreateHandler returns. 

By removing the lock, we don't affect any behavior, we simply will server AppOffline earlier. However, files may still be locked for a bit.

pakrym: Won't it brake web deploy scenarios?

jkotalik: I don't think so, but we can discuss tomorrow.

jkotalik: We figured out why. It's to prevent a second shutdown request from immediately returning and crashing the process. I'll _try_ to write a test but it's probably going to be racy.

src/Servers/IIS/AspNetCoreModuleV2/InProcessRequestHandler/inprocessapplication.cpp
```diff
@@ -544,9 +543,23 @@ IN_PROCESS_APPLICATION::CreateHandler(
...
+    auto requestCount = --m_requestCount;
+
+    LOG_TRACEF(L"Removing Request %d", requestCount);
+
+    if (m_fStopCalled && requestCount == 0 && !m_blockManagedCallbacks)
+    {
+        CallRequestsDrained();
+    }
+}
+
```
jkotalik: nit

src/Servers/IIS/AspNetCoreModuleV2/InProcessRequestHandler/inprocessapplication.cpp
```diff
@@ -544,9 +543,23 @@ IN_PROCESS_APPLICATION::CreateHandler(
 void
 IN_PROCESS_APPLICATION::HandleRequestCompletion()
 {
-    SRWSharedLock lock(m_stateLock);
-    auto requestCount = m_requestCount--;
-    if (m_fStopCalled && requestCount == 0)
+    //SRWSharedLock stopLock(m_stateLock);
```
jkotalik: nit

src/Servers/IIS/IntegrationTesting.IIS/src/IISDeployer.cs
```diff
@@ -45,6 +46,11 @@ public IISDeployer(IISDeploymentParameters deploymentParameters, ILoggerFactory
 
         public override void Dispose()
         {
```
pakrym: What is this for?

src/Servers/IIS/IIS/test/testassets/InProcessWebSite/Program.cs
```diff
@@ -63,6 +63,18 @@ public static int Main(string[] args)
 
                         Thread.Sleep(Timeout.Infinite);
                     }
+                    break;
+                case "RemoveShutdownLimit":
+                    {
+                        var host = new WebHostBuilder()
+                           .UseIIS()
+                            .UseShutdownTimeout(TimeSpan.FromSeconds(120))
```
pakrym: Says remove but actually increases

src/Servers/IIS/IIS/test/Common.FunctionalTests/AppOfflineTests.cs
```diff
@@ -78,7 +79,11 @@ public async Task AppOfflineDroppedWhileSiteFailedToStartInShim_AppOfflineServed
             AddAppOffline(deploymentResult.ContentRoot);
 
             await AssertAppOffline(deploymentResult);
-            DeletePublishOutput(deploymentResult);
+            RetryHelper.RetryOperation(
```
pakrym: You don't need this retry.

src/Servers/IIS/AspNetCoreModuleV2/InProcessRequestHandler/inprocessapplication.cpp
```diff
@@ -544,9 +543,23 @@ IN_PROCESS_APPLICATION::CreateHandler(
...
+
+    auto requestCount = --m_requestCount;
+
+    LOG_TRACEF(L"Removing Request %d", requestCount);
+
+    if (m_fStopCalled && requestCount == 0 && !m_blockManagedCallbacks)
```
pakrym: You need a lock to read `m_fStopCalled`

src/Servers/IIS/IntegrationTesting.IIS/src/IISDeployer.cs
jkotalik: Making sure we don't dispose twice. I was seeing this when trying some stress tests locally and I think it should be fixed (though it isn't relevant to this pr).

src/Servers/IIS/IntegrationTesting.IIS/src/IISDeployer.cs
```diff
@@ -24,12 +24,13 @@ public class IISDeployer : IISDeployerBase
         private const string DetailedErrorsEnvironmentVariable = "ASPNETCORE_DETAILEDERRORS";
 
         private static readonly TimeSpan _timeout = TimeSpan.FromSeconds(60);
-        private static readonly TimeSpan _retryDelay = TimeSpan.FromMilliseconds(200);
+        private static readonly TimeSpan _retryDelay = TimeSpan.FromMilliseconds(500);
```
pakrym: Why this?

src/Servers/IIS/IIS/test/Common.FunctionalTests/AppOfflineTests.cs
```diff
@@ -170,7 +218,13 @@ public async Task AppOfflineDroppedWhileSiteRunning_SiteShutsDown_OutOfProcess()
 
             AddAppOffline(deploymentResult.ContentRoot);
             await AssertAppOffline(deploymentResult);
-            DeletePublishOutput(deploymentResult);
+
+            // Files should eventually be unlocked, however when AppOffline returns, it doesn't necessarily mean 
+            RetryHelper.RetryOperation(
```
pakrym: Do you need retry here?

src/Servers/IIS/IIS/test/testassets/InProcessWebSite/Startup.cs
```diff
@@ -221,6 +221,12 @@ private async Task HelloWorld(HttpContext ctx)
             await ctx.Response.WriteAsync("Hello World");
         }
 
+        private async Task HelloWorldDelayed(HttpContext ctx)
+        {
+            await ctx.Request.Body.ReadAsync(new byte[1]);
```
pakrym: There are existing endpoint that read request body, reuse one of them.

src/Servers/IIS/IIS/test/Common.FunctionalTests/AppOfflineTests.cs
```diff
@@ -131,17 +132,64 @@ public async Task AppOfflineDroppedWhileSiteStarting_SiteShutsDown_InProcess()
...
+                    "Host: localhost",
+                    "Connection: close",
+                    "",
+                    "");
+
+                connectionList.Add(connection);
+            }
+
+            // Shouldn't stop the server here. Instead should drop app offline 
+            AddAppOffline(deploymentResult.ContentRoot);
```
pakrym: I wonder if the server would even be able to detect and react to app_offline before we get to `connectionList[i].Send/Receive`

src/Servers/IIS/AspNetCoreModuleV2/InProcessRequestHandler/inprocessapplication.cpp
```diff
@@ -544,9 +543,22 @@ IN_PROCESS_APPLICATION::CreateHandler(
 void
 IN_PROCESS_APPLICATION::HandleRequestCompletion()
 {
-    SRWSharedLock lock(m_stateLock);
-    auto requestCount = m_requestCount--;
-    if (m_fStopCalled && requestCount == 0)
+    SRWSharedLock stopLock(m_stateLock);
```
pakrym: I would be wary of calling customer core under the lock.

src/Servers/IIS/AspNetCoreModuleV2/CommonLib/application.h
```diff
@@ -22,12 +23,21 @@ class APPLICATION : public IAPPLICATION
         _In_  IHttpContext       *pHttpContext,
         _Outptr_result_maybenull_ IREQUEST_HANDLER  **pRequestHandler) override
     {
+        auto waitForShutdown = false;
         *pRequestHandler = nullptr;
 
-        SRWSharedLock stopLock(m_stateLock);
+        {
+            SRWSharedLock stopLock(m_stateLock);
```
pakrym: Interlocked would make it so much cleaner.

src/Servers/IIS/IntegrationTesting.IIS/src/IISDeployer.cs
jkotalik: Trying to make logs less spammy. I'll make it quadratic.

src/Servers/IIS/IIS/test/Common.FunctionalTests/AppOfflineTests.cs
jkotalik: Good point.

jkotalik:
Failures were due to a cert issue with Kestrel and failing to download chromium. I'll file an issue.

jkotalik:
For diagnosing https://github.com/aspnet/AspNetCore/issues/7075.

src/Servers/IIS/IIS/test/Common.FunctionalTests/AppOfflineTests.cs
```diff
@@ -131,17 +132,74 @@ public async Task AppOfflineDroppedWhileSiteStarting_SiteShutsDown_InProcess()
...
+                await connection.Receive(
+                  "HTTP/1.1 200 OK", "");
+                await connection.ReceiveHeaders();
+                await connection.Receive("1", $"{i + 1}");
+                connectionList.Add(connection);
+            }
+
+            AddAppOffline(deploymentResult.ContentRoot);
+
+            var task = deploymentResult.HttpClient.GetAsync("/WaitForAppToStartShuttingDown");
```
pakrym: this should be above AddAppOffline

src/Servers/IIS/IIS/test/Common.FunctionalTests/AppOfflineTests.cs
```diff
@@ -131,17 +132,74 @@ public async Task AppOfflineDroppedWhileSiteStarting_SiteShutsDown_InProcess()
...
+         
+            for (var i = 0; i < 2; i++)
+            {
+                await connectionList[i].Send("a", "");
+                await connectionList[i].Receive("", "4", "done");
+                connectionList[i].Dispose();
             }
+
+            await task;
```
pakrym: This should be above the loop

src/Servers/IIS/AspNetCoreModuleV2/InProcessRequestHandler/inprocessapplication.h
```diff
@@ -188,6 +188,8 @@ class IN_PROCESS_APPLICATION : public InProcessApplicationBase
     void
     StopClr();
 
+    void CallRequestsDrained();
```
pakrym: nit formatting.

src/Servers/IIS/AspNetCoreModuleV2/InProcessRequestHandler/inprocessapplication.cpp
```diff
@@ -544,9 +547,22 @@ IN_PROCESS_APPLICATION::CreateHandler(
...
+    }
+}
+
+void IN_PROCESS_APPLICATION::CallRequestsDrained()
+{
+    auto expected = false;
+    if (m_alreadyDrained.compare_exchange_strong(expected, true))
     {
         LOG_INFO(L"Drained all requests, notifying managed.");
         m_RequestsDrainedHandler(m_ShutdownHandlerContext);
```
pakrym: Set `m_RequestsDrainedHandler` to null after running it.

pakrym (approved):

pakrym (approved):

PR #8420, opened 2019-03-11
C++ client low level API

BrennanConroy:
Please look over the API and ignore tests.

http_client: https://github.com/aspnet/AspNetCore/blob/28ccc03fb90499425a9a3deb0c066e8624fff23a/src/SignalR/clients/cpp/src/signalrclient/http_client.h

websocket_client: https://github.com/aspnet/AspNetCore/blob/28ccc03fb90499425a9a3deb0c066e8624fff23a/src/SignalR/clients/cpp/src/signalrclient/default_websocket_client.h

transport: https://github.com/aspnet/AspNetCore/blob/28ccc03fb90499425a9a3deb0c066e8624fff23a/src/SignalR/clients/cpp/src/signalrclient/transport.h

As a reminder, the `http_client` and `websocket_client` are the things we're wanting to be replaceable. By default we will have implementations that use cpprestsdk.


BrennanConroy:
Ping, any feedback?

No feedback is fine too, just want people to take a look :)

One thing I've been thinking about is accepting the `http_client` as a `unique_ptr` to the `hub_connection`, that way the lifetime is well known and there wont be any weird issues with the http_client and connection being in the same scope but destructors running at different times causing dangling pointer usage.

src/SignalR/clients/cpp/src/signalrclient/http_client.h
```diff
@@ -0,0 +1,40 @@
...
+        std::map<std::string, std::string> headers;
+        std::string content;
+        std::chrono::seconds timeout;
+    };
+
+    class http_response
+    {
+    public:
+        int status_code = 0;
+        std::string content;
```
anurse: Are we going to support SSE? If not, and I think we can probably do without it, this should be fine. If we do want to support it, we'll need this to be a stream of some kind.

src/SignalR/clients/cpp/src/signalrclient/http_client.h
```diff
@@ -0,0 +1,40 @@
...
+#pragma once
+
+#include <string>
+#include <functional>
+#include <map>
+#include <chrono>
+
+namespace signalr
+{
+    enum http_method
```
anurse:
```suggestion
    enum class http_method
```

We use C++11 (or higher) right? Seems like we should use a scoped enum if we're building new things. I'd prefer `http_method::GET` over putting `GET` in the global scope.

src/SignalR/clients/cpp/src/signalrclient/websocket_client.h
```diff
@@ -3,21 +3,21 @@
...
+        virtual void start(std::string url, transfer_format format, std::function<void(std::exception_ptr)> callback) = 0;
 
-        virtual pplx::task<std::string> receive() = 0;
+        virtual void stop(std::function<void(std::exception_ptr)> callback) = 0;
 
-        virtual pplx::task<void> close() = 0;
+        virtual void send(std::string payload, std::function<void(std::exception_ptr)> callback) = 0;
 
-        virtual ~websocket_client() {};
+        virtual void receive(std::function<void(std::string, std::exception_ptr)> callback) = 0;
```
anurse: From the naming and structure I presume this calls the callback once for each call to receive, when a single frame is received? Another alternative would be a `on_recieved` API that registers a callback which will be called once for each frame that arrives (and we'd do the looping ourselves).

src/SignalR/clients/cpp/src/signalrclient/transport.h
```diff
@@ -5,35 +5,29 @@
...
+        virtual void start(const std::string& url, transfer_format format, std::function<void(std::exception_ptr)> callback) = 0;
+        virtual void stop(std::function<void(std::exception_ptr)> callback) = 0;
+        virtual void on_close(std::function<void(std::exception_ptr)> callback) = 0;
 
-        void process_response(const std::string &message);
-        void error(const std::exception &e);
+        virtual void send(std::string payload, std::function<void(std::exception_ptr)> callback) = 0;
 
-        logger m_logger;
+        virtual void on_receive(std::function<void(std::string, std::exception_ptr)> callback) = 0;
```
anurse: I'd expect to use the same API in `websocket_client` as we do here, so maybe we should change `websocket_client` to use `on_receive` as well...

analogrelay (commented):
Looks pretty good

anurse:
> One thing I've been thinking about is accepting the `http_client` as a `unique_ptr` to the `hub_connection`, that way the lifetime is well known and there wont be any weird issues with the http_client and connection being in the same scope but destructors running at different times causing dangling pointer usage.

Did you say... ownership?

image

anurse:
I do think that's a good idea though. Always a fan of `unique_ptr` when we feel like it makes sense, and I think it makes sense here.

src/SignalR/clients/cpp/src/signalrclient/http_client.h
davidfowl: We just need to make sure that it can be added.

src/SignalR/clients/cpp/src/signalrclient/default_http_client.cpp
```diff
@@ -0,0 +1,68 @@
...
+    {
+        web::http::method method;
+        if (request.method == http_method::GET)
+        {
+            method = U("GET");
+        }
+        else if (request.method == http_method::POST)
+        {
+            method = U("POST");
+        }
```
halter73: nit: fail properly in else case in case a new method is added, someone does an unsafe cast, etc...

src/SignalR/clients/cpp/src/signalrclient/http_client.h
halter73: We talked about using ostream for this and istream for the request content. The problem is that only offers blocking and polling APIs. The alternative is using [Asio's AsyncRead/WriteStream](http://think-async.com/Asio/asio-1.11.0/doc/asio/reference/AsyncWriteStream.html).

Since we're only aiming for source compatibility initially instead of some stable C ABI, we can probably just add an async stream field to http_request/response later and gracefully skip SSE if the http_client that doesn't use the field. We'd need a new send overload that calls the callback prior to buffering the entire response anyway.

src/SignalR/clients/cpp/src/signalrclient/http_client.h
```diff
@@ -0,0 +1,40 @@
...
+        POST
+    };
+
+    class http_request
+    {
+    public:
+        http_method method;
+        std::map<std::string, std::string> headers;
+        std::string content;
+        std::chrono::seconds timeout;
```
halter73: Maybe timeout should be a separate parameter on send. Or better yet, define a single-method CTS-like class that we return from send instead.

src/SignalR/clients/cpp/src/signalrclient/websocket_transport.cpp
```diff
@@ -41,7 +43,99 @@ namespace signalr
...
+                }
+
+                auto transport = weak_transport.lock();
+                if (transport)
+                {
+                    transport->m_process_response_callback(message, nullptr);
+
+                    if (!cts.get_token().is_canceled())
+                    {
+                        transport->receive_loop(cts);
```
halter73: Could this stack dive if enough messages were buffered in memory that `websocket_client->receive` repeatedly called its callback inline?

src/SignalR/clients/cpp/src/signalrclient/websocket_transport.cpp
```diff
@@ -66,44 +160,37 @@ namespace signalr
...
-            websocket_client->connect(url)
-                .then([transport, connect_tce, receive_loop_cts](pplx::task<void> connect_task)
+            websocket_client->start(url, format, [transport, receive_loop_cts, callback](std::exception_ptr exception)
                 {
                     try
                     {
-                        connect_task.get();
+                        if (exception != nullptr)
+                        {
+                            std::rethrow_exception(exception);
```
halter73: Why not just call the callback immediately with the exception and return?

BrennanConroy: So we can log the exception

src/SignalR/clients/cpp/src/signalrclient/websocket_transport.cpp
BrennanConroy: Definitely could. I believe we talked about this and decided to do it this way for now and re-evaluate later if we need to dispatch.

halter73: If there's a risk, I figure it's easy enough to turn this into a trampoline to avoid the problem. I think it would look something like this.

```c++
auto calling_receive = true;
while (calling_receive) {
    websocket_client->receive([weak_transport, cts, logger, websocket_client, calling_receive](std::string message, std::exception_ptr exception)
        {
           // ...
           if (!cts.get_token().is_canceled())
           {
                if (calling_receive)
                {
                    calling_receive = false
                }
                else
                {
                    transport->receive_loop(cts);
                }
            }
        });

    calling_receive = !calling_receive;
}
```

BrennanConroy: Something like that would work, but we still need to jump off the main thread for the start of the receive loop in order to avoid blocking start. Right now I'm relying on the transport dispatching, like we talked about in the meeting.

halter73: What's the main thread? Whatever thread you call connection.start() on?

Do we jump off the main thread currently before calling websocket_client->receive()? Or does this design prevent us from jumping off the main thread in the future?

I would think the trampoline is strictly better than the recursive version because it avoids potentially stack diving.

src/SignalR/clients/cpp/src/signalrclient/transport.h
BrennanConroy: The question is, do we expect people who provide their own websocket impl to do the receive loop and call our callback, or do we do the receive loop in the transport and just call receive on their impl?

src/SignalR/clients/cpp/src/signalrclient/websocket_transport.cpp
BrennanConroy: >What's the main thread? Whatever thread you call connection.start() on?

Correct.

>Do we jump off the main thread currently before calling websocket_client->receive()? Or does this design prevent us from jumping off the main thread in the future?

We don't do anything threading related yet, except in the http_client and websocket_client as those use pplx. We can add it whenever though.

>I would think the trampoline is strictly better than the recursive version because it avoids potentially stack diving.

I agree.

src/SignalR/clients/cpp/src/signalrclient/connection_impl.cpp
```diff
@@ -224,52 +233,57 @@ namespace signalr
...
+        transport->on_receive([disconnect_cts, connect_request_tce, logger, weak_connection](std::string message, std::exception_ptr exception)
+            {
+                if (exception != nullptr)
                 {
-                    connection->process_response(response);
-                }
-            };
+                    try
+                    {
+                        std::rethrow_exception(exception);
```
jkotalik: Comment on why you need to rethrow at it isn't intuitive.

src/SignalR/clients/cpp/src/signalrclient/connection_impl.cpp
```diff
@@ -224,52 +233,57 @@ namespace signalr
...
+                        // no op after connection started successfully
+                        connect_request_tce.set_exception(exception);
+                    }
+                }
+                else
                 {
-                    logger.log(trace_level::info,
-                        std::string{ "ignoring stray error received after connection was restarted. error: " }
-                        .append(e.what()));
+                    // When a connection is stopped we don't wait for its transport to stop. As a result if the same connection
```
jkotalik: nit: duplication on comments.

src/SignalR/clients/cpp/src/signalrclient/default_websocket_client.cpp
```diff
@@ -21,30 +21,73 @@ namespace signalr
...
-    pplx::task<void> default_websocket_client::connect(const std::string& url)
+    void default_websocket_client::start(std::string url, transfer_format, std::function<void(std::exception_ptr)> callback)
     {
-        return m_underlying_client.connect(utility::conversions::to_string_t(url));
+        m_underlying_client.connect(utility::conversions::to_string_t(url))
+            .then([callback](pplx::task<void> task)
+            {
+                try
+                {
+                    task.get();
```
jkotalik: I think I need a bit more context on what this method is doing. Why do you need to complete the task provided here? Is the callback here only used for exception handling or is it required?

src/SignalR/clients/cpp/src/signalrclient/default_websocket_client.cpp
```diff
@@ -21,30 +21,73 @@ namespace signalr
...
+                    callback(nullptr);
+                }
+                catch (...)
+                {
+                    callback(std::current_exception());
```
jkotalik: I'm not convinced this pattern for Try/Catch is good. Why not catch a std::exception& and pass the ref into the callback? That way you don't need to rethrow to get the exception.

src/SignalR/clients/cpp/src/signalrclient/connection_impl.cpp
```diff
@@ -23,22 +24,30 @@ namespace signalr
 
     std::shared_ptr<connection_impl> connection_impl::create(const std::string& url, trace_level trace_level, const std::shared_ptr<log_writer>& log_writer)
     {
-        return connection_impl::create(url, trace_level, log_writer, std::make_unique<web_request_factory>(), std::make_unique<transport_factory>());
+        return connection_impl::create(url, trace_level, log_writer, nullptr, std::make_unique<transport_factory>());
     }
 
     std::shared_ptr<connection_impl> connection_impl::create(const std::string& url, trace_level trace_level, const std::shared_ptr<log_writer>& log_writer,
-        std::unique_ptr<web_request_factory> web_request_factory, std::unique_ptr<transport_factory> transport_factory)
+        std::unique_ptr<http_client> http_client, std::unique_ptr<transport_factory> transport_factory)
```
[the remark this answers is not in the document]
BrennanConroy: Nah, https://github.com/aspnet/AspNetCore/pull/8420/files#diff-94c5397b9dcb8a27dcc5dd7ca166abe6R42 😛

src/SignalR/clients/cpp/src/signalrclient/default_websocket_client.cpp
BrennanConroy: This is a point in time thing as we're still using `pplx::task` in our code.

What this is doing, is observing any exceptions from the completed task so we don't crash the process, and then forwarding the exception to the new callback based APIs

src/SignalR/clients/cpp/src/signalrclient/connection_impl.cpp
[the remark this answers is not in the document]
jkotalik: Instead of passing nullptr here: https://github.com/aspnet/AspNetCore/pull/8420/files#diff-94c5397b9dcb8a27dcc5dd7ca166abe6R27, pass in https://github.com/aspnet/AspNetCore/pull/8420/files#diff-94c5397b9dcb8a27dcc5dd7ca166abe6R48

src/SignalR/clients/cpp/src/signalrclient/negotiate.cpp
```diff
@@ -0,0 +1,103 @@
...
+                    {
+                        response.url = utility::conversions::to_utf8string(negotiation_response_json[_XPLATSTR("url")].as_string());
+
+                        if (negotiation_response_json.has_field(_XPLATSTR("accessToken")))
+                        {
+                            response.accessToken = utility::conversions::to_utf8string(negotiation_response_json[_XPLATSTR("accessToken")].as_string());
+                        }
+                    }
+
+                    if (negotiation_response_json.has_field(_XPLATSTR("ProtocolVersion")))
```
jkotalik: Is this the typical check you do for SignalR vs SignalR Core.

src/SignalR/clients/cpp/src/signalrclient/websocket_transport.cpp
```diff
@@ -41,7 +43,131 @@ namespace signalr
...
+                    }
+                    catch (...)
+                    {
+                        cts.cancel();
+
+                        logger.log(
+                            trace_level::errors,
+                            std::string("[websocket transport] unknown error occurred when receiving response from websocket"));
+
+                        std::promise<void> prom;
```
jkotalik: nit: prom -> promise.

src/SignalR/clients/cpp/src/signalrclient/websocket_transport.cpp
```diff
@@ -41,7 +43,131 @@ namespace signalr
...
+            {
+                if (exception != nullptr)
+                {
+                    try
+                    {
+                        std::rethrow_exception(exception);
+                    }
+                    catch (const std::exception & e)
+                    {
+                        cts.cancel();
```
jkotalik: nit: simplify these two catches into one?

jkotalik (approved):
LGTM from the perspective of non-signalR dev. IMO let's try to get this in sooner than later.

BrennanConroy:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=128810&view=logs

BrennanConroy:
Test failures: https://github.com/aspnet/AspNetCore-Internal/issues/1756 and https://github.com/aspnet/AspNetCore-Internal/issues/2074

src/SignalR/clients/cpp/src/signalrclient/negotiate.cpp
BrennanConroy: Yes

davidfowl:
:clap:

PR #8424, opened 2019-03-12
Fix SignalR transport test

mikaelm12:
Fixes: https://github.com/aspnet/AspNetCore/issues/8262

There was a race where the transport could stop before the HubConnection finished starting.


BrennanConroy:
Shouldn't `hubConnection.start().timeout(1, TimeUnit.SECONDS).blockingAwait();` be waiting for start to finish?

mikaelm12:
The issue was the hubConnection would continuously poll and before we could receive the 204 response between starting the transport and sending the handshake since we weren't blocking.

BrennanConroy (approved):
Handshake bug :(

halter73 (approved):

PR #8425, opened 2019-03-12
Add HubMetdata

davidfowl:
- Added HubMetadata to detect the hub type on Endpoints
- Added a test

Fixes https://github.com/aspnet/AspNetCore/issues/7783

### DIFF
--- a/src/SignalR/server/Core/src/HubMetadata.cs
+++ b/src/SignalR/server/Core/src/HubMetadata.cs
@@ -1,0 +1,11 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.AspNetCore.SignalR
+{
+    public class HubMetadata
+    {
+        public Type HubType { get; set; }
+    }
+}
--- a/src/SignalR/server/Core/src/HubMetadata.cs
+++ b/src/SignalR/server/Core/src/HubMetadata.cs
@@ -1,11 +1,23 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
 using System;
-using System.Collections.Generic;
-using System.Text;
 
 namespace Microsoft.AspNetCore.SignalR
 {
+    /// <summary>
+    /// Metadata that describes the <see cref="Hub"/> information associated with a specific endpoint.
+    /// </summary>
     public class HubMetadata
     {
-        public Type HubType { get; set; }
+        public HubMetadata(Type hubType)
+        {
+            HubType = hubType;
+        }
+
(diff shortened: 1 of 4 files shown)


src/SignalR/server/Core/src/HubMetadata.cs
```diff
@@ -0,0 +1,11 @@
+using System;
```
BrennanConroy: copyright

src/SignalR/server/SignalR/test/MapSignalRTests.cs
```diff
@@ -185,6 +185,28 @@ void ConfigureRoutes(IEndpointRouteBuilder routes)
...
         }
 
+        [Fact]
+        public void MapHubEndPointRoutingAppliesHubMetadata()
+        {
+            void ConfigureRoutes(IEndpointRouteBuilder routes)
+            {
+                // This "Foo" policy should override the default auth attribute
+                routes.MapHub<AuthHub>("/path")
+                      .RequireAuthorization(new AuthorizeAttribute("Foo"));
```
BrennanConroy: Assert this?

BrennanConroy (approved):

src/SignalR/server/SignalR/test/MapSignalRTests.cs
davidfowl: Let me remove it, it was copy pasta.

src/SignalR/server/Core/src/HubMetadata.cs
```diff
@@ -0,0 +1,11 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.AspNetCore.SignalR
+{
+    public class HubMetadata
+    {
+        public Type HubType { get; set; }
```
rynowak: Do you really want this to be mutable? Everythign we put in metadata is immutable so far.

davidfowl: Ah good catch.

davidfowl: Bleh, making it immutable means constructor hell though.

rynowak: oh no, a single constructor with a single parameter :trollface:

davidfowl: It's fine for this since HubType is *required* anyways. It's the fact that other potentially optional metadata will need to be get only and will require more ctors

analogrelay (commented):
Don't have context to review at this time. Looks fine to me though.

JamesNK (approved):

rynowak (approved):

PR #8457, opened 2019-03-12
C++11 compliant

BrennanConroy:
This is all that was required from a C++ point of view to compile on a mac with `-std=c++11`

### DIFF
--- a/src/SignalR/clients/cpp/src/signalrclient/hub_connection_impl.cpp
+++ b/src/SignalR/clients/cpp/src/signalrclient/hub_connection_impl.cpp
@@ -50,10 +50,8 @@ namespace signalr
 
     void hub_connection_impl::initialize()
     {
-        auto this_hub_connection = shared_from_this();
-
         // weak_ptr prevents a circular dependency leading to memory leak and other problems
-        auto weak_hub_connection = std::weak_ptr<hub_connection_impl>(this_hub_connection);
+        std::weak_ptr<hub_connection_impl> weak_hub_connection = shared_from_this();
 
         m_connection->set_message_received([weak_hub_connection](const std::string& message)
         {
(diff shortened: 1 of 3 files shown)


src/SignalR/clients/cpp/src/signalrclient/hub_connection_impl.cpp
```diff
@@ -50,10 +50,8 @@ namespace signalr
 
     void hub_connection_impl::initialize()
     {
-        auto this_hub_connection = shared_from_this();
-
         // weak_ptr prevents a circular dependency leading to memory leak and other problems
-        auto weak_hub_connection = std::weak_ptr<hub_connection_impl>(this_hub_connection);
```
halter73: Was this calling the copy ctor before?

halter73: Nvm. I see that shared_from_this() returns a shared_ptr. Now it makes sense why you switched from auto to writing out the full type.

halter73 (approved):

halter73:
The AspNetCore-ci (Test: Ubuntu 16.04 x64) build failure seems to be a transient. Retrying:

```
  FAILURE: Build failed with an exception.
  
  * What went wrong:
  A problem occurred configuring root project 'signalr'.
  > Could not resolve all artifacts for configuration ':classpath'.
     > Could not resolve org.slf4j:slf4j-api:1.7.2.
       Required by:
           project : > com.diffplug.spotless:spotless-plugin-gradle:3.14.0 > com.diffplug.spotless:spotless-lib-extra:1.14.0 > org.eclipse.jgit:org.eclipse.jgit:4.9.0.201710071750-r
        > Could not resolve org.slf4j:slf4j-api:1.7.2.
           > Could not get resource 'https://plugins.gradle.org/m2/org/slf4j/slf4j-api/1.7.2/slf4j-api-1.7.2.pom'.
              > Could not GET 'https://plugins.gradle.org/m2/org/slf4j/slf4j-api/1.7.2/slf4j-api-1.7.2.pom'.
                 > Read timed out
```

halter73:
The helix failure isn't yours either: https://github.com/aspnet/AspNetCore/pull/8458

Issue #8164, opened 2019-03-04
[Helix] InMemory.FunctionalTests-netcoreapp3.0 timeout on Ubuntu.1810.Amd64.Open


pakrym:
https://mc.dot.net/#/user/aspnetcore/pr~2Faspnet~2Faspnetcore/ci/20190303.15/workItem/InMemory.FunctionalTests-netcoreapp3.0/wilogs

cc @anurse

anurse:
Hrm, timeout errors are fun... not sure what is equivalent to "skipping" here, since I don't really want to turn the entire test pass off.

anurse:
We need xunit diagnostic logs and our test logs from Helix. Dumps might help too. #7774 is tracking that work.

anurse:
https://mc.dot.net/#/user/aspnetcore/pr~2Faspnet~2Faspnetcore/ci/20190304.3/workItem/InMemory.FunctionalTests-netcoreapp3.0/wilogs as well.

Not sure the best way to mitigate these. We could disable all of InMemory.FunctionalTests in Helix but that's a really big hammer to apply. Both cases have been on Ubuntu (but different versions), we could try just disabling them on those platforms. That's still pretty big.

#7774 is tracking getting more data out of Helix on why these are failing, but I don't expect immediate progress on that.	

(cc @jkotalik @Tratcher @Eilon @muratg for any thoughts).

Tratcher:
Yeah, getting a dump seems like the best route forward here. We've tried the xunit diagnostic logs before but their noise ratio is really bad. If we had a repro we could turn them on just for a given PR.

anurse:
I propose closing this for now. We need to mitigate these better and I think we've decided to just shut the entire test project (InMemory.FunctionalTests) off on the OSes it hangs on. 

To get a clearer picture of that and avoid confusion, we'll look at BRT reports starting from now.

PR #8458, opened 2019-03-13
Skip Kestrel's InMemory.FunctionalTests on macOS

halter73:
- until we figure out what's going on (I hope)

Addresses #8164 

We're still seeing issues in Helix like https://dev.azure.com/dnceng/public/_build/results?buildId=122553

### DIFF
--- a/src/Servers/Kestrel/test/InMemory.FunctionalTests/Properties/AssemblyInfo.cs
+++ b/src/Servers/Kestrel/test/InMemory.FunctionalTests/Properties/AssemblyInfo.cs
@@ -1,8 +1,10 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using Microsoft.AspNetCore.Testing.xunit;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Testing;
 
 [assembly: ShortClassName]
 [assembly: LogLevel(LogLevel.Trace)]
+[assembly: OSSkipCondition(OperatingSystems.MacOSX, SkipReason = "https://github.com/aspnet/AspNetCore/issues/8164")]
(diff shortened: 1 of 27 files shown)


src/Servers/Kestrel/test/InMemory.FunctionalTests/Properties/AssemblyInfo.cs
```diff
@@ -1,8 +1,10 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using Microsoft.AspNetCore.Testing.xunit;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Testing;
 
 [assembly: ShortClassName]
 [assembly: LogLevel(LogLevel.Trace)]
+[assembly: OSSkipCondition(OperatingSystems.MacOSX, SkipReason = "https://github.com/aspnet/AspNetCore/issues/8164")]
```
halter73: @Tratcher @JunTaoLuo I copied this from https://github.com/aspnet/AspNetCore/blob/e6d77a1f6c1c87425e7820e6ded1ce527a66050d/src/Servers/HttpSys/test/FunctionalTests/Properties/AssemblyInfo.cs, but tests still run when I run `dotnet test` on my mac. Am I missing something?

JunTaoLuo: https://github.com/aspnet/Extensions/blob/master/src/TestingUtils/Microsoft.AspNetCore.Testing/src/xunit/OSSkipConditionAttribute.cs#L19

The skip reason is being interpreted as the os version probably. Turn the skip reason as a comment.

halter73: It would be weird for a named parameter to be misinterpreted like that. Either way, I tried it and it doesn't appear to work.

As an aside, can we currently skip iff running on macOS AND as part of a Helix test?

jkotalik (approved):
The right move to do, unfortunately.

halter73:
Is there currently a way to skip tests only if they're both running on macOS and as part of Helix?

src/Servers/Kestrel/test/InMemory.FunctionalTests/Properties/AssemblyInfo.cs
halter73: Turns out I have to use ConditionalFact/Theory 😆

anurse:
This may not be necessary given that we're disabling Helix for PRs entirely for now (cc @HaoK).

Once we have a better infrastructure for flaky tests (`[Flaky]` attribute) we'll be looking at restoring it and the `Flaky` attribute supports targeting specific Helix queues, OSes, etc.

HaoK:
New tweaked plan is to just ignore all helix failures, meaning its always going to be a 'green' check (even if every single test fails on helix) See https://github.com/aspnet/AspNetCore/pull/8464

halter73:
I'm not deleting my branch yet in case we change our mind.

PR #8464, opened 2019-03-13
Ignore helix mission control test failures (test failures won't fail build)

HaoK:
Alternative to https://github.com/aspnet/AspNetCore/pull/8454

This will make it so helix check will pass as long as the jobs are submitted, it will ignore any failures in the helix jobs


dougbu (approved):
(If errors come from anywhere else, may need to use `ContinueOnError: true` in the YAML.)

jkotalik (approved):
Yay!!!

HaoK:
I'll file another PR that turns WaitForWorkItemCompletion to false now as well since there's no need to wait til all the tests finish if we don't care if they are passing now

PR #8478, opened 2019-03-13
Add MemoryCache services where needed

ajcvickers:
Previously the tests were relying on EF adding the services. See https://github.com/aspnet/EntityFrameworkCore/issues/14756 and https://github.com/aspnet/AspNetCore/issues/7728


dougbu (approved):
Thanks!

Eilon (approved):

PR #8482, opened 2019-03-13
Attempt to publish MusicStore to temp directories

pranavkm:
Fixes https://github.com/aspnet/AspNetCore/issues/8439

### DIFF
--- a/src/MusicStore/test/MusicStore.E2ETests/MusicStore.E2ETests.csproj
+++ b/src/MusicStore/test/MusicStore.E2ETests/MusicStore.E2ETests.csproj
@@ -11,8 +11,6 @@
     <NoWarn>$(NoWarn);NU1605</NoWarn>
     <!-- https://github.com/aspnet/AspNetCore/issues/6549 -->
     <BuildHelixPayload>false</BuildHelixPayload>
-    <IsTestProject>false</IsTestProject>
-    <IsReferenceAssemblyProject>false</IsReferenceAssemblyProject>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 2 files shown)


anurse:
Apparently there is infrastructure in IIS's functional tests to publish test projects as a build step which should completely eliminate this issue. I think we should look at adopting that in MusicStore instead of continuing to patch this publish-during-test-run logic.

pranavkm:
Could you point me to where that would be? I think MusicStore intentionally publishes as part of the test primarily because there's variation in the app type (Self-Contained \ portable), bitness, and there's some db stuff that gets reset on a per test basis. I'm not super sure if this would translate well to a set of build time publish.

anurse:
On it, stand by

anurse:
Posted details in the issue: https://github.com/aspnet/AspNetCore/issues/8439#issuecomment-472520196

anurse:
> because there's variation in the app type (Self-Contained \ portable), bitness

The IIS logic handles this by publishing all those variants separately during build.

> and there's some db stuff that gets reset on a per test basis

I believe we can still do that within the test even when publishing during build.

HaoK:
+1 for publishing the test apps during build like IIS, it will need to be done for MusicStore to work on helix anyways (https://github.com/aspnet/AspNetCore/issues/8343)

jkotalik:
Here is a peek at what you need to do: https://github.com/aspnet/AspNetCore/blob/master/src/Servers/IIS/IIS/src/Microsoft.AspNetCore.Server.IIS.csproj#L41

pranavkm:
Superseded by https://github.com/aspnet/AspNetCore/pull/8578

PR #8486, opened 2019-03-13
Add a separate test pass to run flaky tests that doesn't fail the build

analogrelay:
Similar to https://github.com/aspnet/Extensions/pull/1239.

This doesn't actually add the pass yet, just a sample flaky test to use to verify things are working. Sorry Kestrel CODEOWNERS, you're getting pinged on this one because you're the random victim ;). I'll remove reviewers and re-add ones that make sense when this is ready to review.

TODO:
* [x] It worked on Windows, make sure it works on macOS and Ubuntu too
* [ ] Make sure TRX files are still published.

### DIFF
--- a/eng/helix/vstest/runtests.cmd
+++ b/eng/helix/vstest/runtests.cmd
@@ -1,3 +1,6 @@
+REM Disable "!Foo!" expansions because they break the filter syntax
+setlocal disableextensions
+
 set target=%1
 set sdkVersion=%2
 set runtimeVersion=%3
@@ -18,12 +21,35 @@ set HELIX=%helixQueue%
 
 %DOTNET_ROOT%\dotnet vstest %target% -lt >discovered.txt
 find /c "Exception thrown" discovered.txt
-if %errorlevel% equ 0 (
-    echo Exception thrown during test discovery.
-    type discovered.txt
+REM "ERRORLEVEL is not %ERRORLEVEL%" https://blogs.msdn.microsoft.com/oldnewthing/20080926-00/?p=20743/
+if not errorlevel 1 (
+    echo Exception thrown during test discovery. 1>&2
+    type discovered.txt 1>&2
     exit 1
 )
 
-%DOTNET_ROOT%\dotnet vstest %target% --logger:trx --logger:console;verbosity=normal
+set exit_code=0
 
+REM Run non-flaky tests first
+REM We need to specify all possible Flaky filters that apply to this environment, because the flaky attribute
+REM only puts the explicit filter traits the user provided in
+REM Filter syntax: https://github.com/Microsoft/vstest-docs/blob/master/docs/filter.md
+set NONFLAKY_FILTER="Flaky:All!=true&Flaky:Helix:All!=true&Flaky:Helix:Queue:All!=true&Flaky:Helix:Queue:%HELIX%!=true"
+echo Running non-flaky tests.
+%DOTNET_ROOT%\dotnet vstest %target% --logger:trx --logger:console;verbosity=normal --TestCaseFilter:%NONFLAKY_FILTER%
+if errorlevel 1 (
+    echo Failure in non-flaky test 1>&2
+    set exit_code=1
+    REM DO NOT EXIT
+)
+
+set FLAKY_FILTER="Flaky:All=true|Flaky:Helix:All=true|Flaky:Helix:Queue:All=true|Flaky:Helix:Queue:%HELIX%=true"
+echo Running known-flaky tests.
+%DOTNET_ROOT%\dotnet vstest %target% --logger:trx --logger:console;verbosity=normal --TestCaseFilter:%FLAKY_FILTER%
+if errorlevel 1 (
+    echo Failure in flaky test 1>&2
+    REM DO NOT EXIT and DO NOT SET EXIT_CODE to 1
+)
+
+exit %exit_code%
 
--- a/eng/helix/vstest/runtests.sh
+++ b/eng/helix/vstest/runtests.sh
@@ -69,7 +69,7 @@ fi
 
 # Run non-flaky tests first
 # We need to specify all possible Flaky filters that apply to this environment, because the flaky attribute
-# only puts the explicit filter traits the user provided in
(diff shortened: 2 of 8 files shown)


jkotalik (dismissed):
Seems great! 😄

eng/helix/vstest/runtests.cmd
```diff
@@ -1,3 +1,6 @@
+REM Disable "!Foo!" expansions because they break the filter syntax
+setlocal disableextensions
```
anurse: cc @HaoK 	

Even with the planned change to ignore the exit code from helix altogether I'd like to get this change in in some form so that when we do start paying attention to the exit code, we don't start getting known-flaky tests causing failures.

anurse: (Still working on Unix scripts btw)

HaoK: Yup, ignoring the exit code is hopefully just a short/medium term thing that gives us some time to get flaky in place

HaoK: maybe consider passing in the filters to each script (assuming they are the same in the sh version)

anurse: Yeah, they would be the same. I'll look at that. There can be some shell escaping problems with the syntax so it might be tricky, but I'll give it a shot.

HaoK:
Do you think its worth separating flaky and non flaky farther apart for helix?  Basically instead of running both flaky and non flaky on the same helix item (and just ignoring flaky failures), build two helix work items instead of one per test project, only including the additional run helix flaky work item if the test project has any marked (and the script could just take an parameter saying to always report success for that run).

For example:
`20190313.11/workItem/ProjectTemplates.Tests-netcoreapp3.0` excluding flaky tests
`20190313.11/workItem/ProjectTemplates.Tests-netcoreapp3.0-flaky` with the flaky tests filter and alwaysSucceed = true (if we can't detect this at build time, can just add a `FlakyTests=true` msbuild property for those test projects)

anurse:
We could do that, but adding a `FlakyTest` msbuild property seems error-prone. One of the things I like about the current system is that you just slap an attribute on a test and get the behavior, you don't have to remember to update an MSBuild file.

anurse:
🤞 I think I have the YAML and MSBuild goop in place to run the Flaky test pass separately.

anurse:
Helix results look good (https://mc.dot.net/#/user/aspnetcore/pr~2Faspnet~2Faspnetcore/ci/20190314.26/workItem/Libuv.FunctionalTests-netcoreapp3.0/wilogs). Helix is reporting success even though flaky tests failed.

anurse:
I believe this is _basically_ ready for review. I **will remove** the fake-flaky test I added to Kestrel before merging (feel free to withhold final approval until I do that if you'd like) but the rest should be ready to review.

eng/helix/vstest/runtests.sh
```diff
@@ -67,4 +67,23 @@ if grep -q "Exception thrown" discovered.txt; then
     exit 1
 fi
 
-$DOTNET_ROOT/dotnet vstest $1 --logger:trx
+# Run non-flaky tests first
+# We need to specify all possible Flaky filters that apply to this environment, because the flaky attribute
+# only puts the explicit filter traits the user provided in
```
dougbu: "provided in" what?

eng/helix/vstest/runtests.cmd
```diff
@@ -18,12 +21,35 @@ set HELIX=%helixQueue%
...
+    set exit_code=1
+    REM DO NOT EXIT
+)
+
+set FLAKY_FILTER="Flaky:All=true|Flaky:Helix:All=true|Flaky:Helix:Queue:All=true|Flaky:Helix:Queue:%HELIX%=true"
+echo Running known-flaky tests.
+%DOTNET_ROOT%\dotnet vstest %target% --logger:trx --logger:console;verbosity=normal --TestCaseFilter:%FLAKY_FILTER%
+if errorlevel 1 (
+    echo Failure in flaky test 1>&2
+    REM DO NOT EXIT and DO NOT SET EXIT_CODE to 1
```
dougbu: Suggest including similar comments in runtests.sh

eng/helix/vstest/runtests.cmd
```diff
@@ -18,12 +21,35 @@ set HELIX=%helixQueue%
...
+    type discovered.txt 1>&2
     exit 1
 )
 
-%DOTNET_ROOT%\dotnet vstest %target% --logger:trx --logger:console;verbosity=normal
+set exit_code=0
+
+REM Run non-flaky tests first
+REM We need to specify all possible Flaky filters that apply to this environment, because the flaky attribute
+REM only puts the explicit filter traits the user provided in
```
dougbu: "provided in" what?

dougbu (approved):
Minor comments and, yes, remove the temporary test 😺

HaoK (approved):
Looks good, 

one thing that still nags me is running the non flaky + flaky tests together on helix. There's going to be potentially a lot more noise in the helix logs with all the flaky tests at the end so its going to make it even more painful hunting for the non flaky error messages (compared to a clean helix job with its on mission control link + a flaky helix job with a perma red mission control link)

But hopefully once we get trx support for the error reporting maybe it won't be as much of an issue

eng/helix/vstest/runtests.cmd
anurse: shhhhhh

HaoK:
Basically an alternative to consider doing eventually is two helix jobs 

1) helix job with FailOnMissionControlFailure = true (non flaky version), 
2) helix job that passes in the flaky filter to run all tests that runs with FailOnMissionControl = false (so the build ignores any failures from that run)

anurse:
Yeah, I do agree that it won't be the cleanest results format. I could suppress the console output of the flaky run maybe.

We could create separate Helix work items for flaky tests if we thought that was useful but it feels like that will take up a lot more resources than it's worth...

HaoK:
Well its not a big deal right now since we are treating all helix tests as flaky so...there's that...

PR #8490, opened 2019-03-13
Add dependency from AspNetCore targeting pack to dotnet Core targeting pack

pakrym:
Fixes https://github.com/aspnet/AspNetCore/issues/8284

### DIFF
--- a/src/Installers/Rpm/TargetingPack/Rpm.TargetingPack.rpmproj
+++ b/src/Installers/Rpm/TargetingPack/Rpm.TargetingPack.rpmproj
@@ -27,6 +27,7 @@
     </ProjectReference>
 
     <InstallerOwnedDirectory Include="$(RpmPackageInstallRoot)packs/$(TargetingPackName)/" />
+    <RpmDependency Include="dotnet-targeting-pack-$(AspNetCoreMajorVersion).$(AspNetCoreMinorVersion)" Version="$(MicrosoftNETCoreAppPackageVersion.Split('-')[0])" />
   </ItemGroup>
 
   <Import Project="$(MSBuildToolsPath)\Microsoft.Common.targets" />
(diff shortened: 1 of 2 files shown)


src/Installers/Rpm/TargetingPack/Rpm.TargetingPack.rpmproj
```diff
@@ -27,6 +27,7 @@
     </ProjectReference>
 
     <InstallerOwnedDirectory Include="$(RpmPackageInstallRoot)packs/$(TargetingPackName)/" />
+    <RpmDependency Include="dotnet-targeting-pack-$(AspNetCoreMajorVersion).$(AspNetCoreMinorVersion)" Version="$(MicrosoftNETCoreAppPackageVersion.Split('-')[0])" />
```
JunTaoLuo: Why is the package version different between debian and rpm installers? `DotnetRuntimeDependencyVersion` vs `MicrosoftNETCoreAppPackageVersion.Split('-')[0]`?

pakrym: Honestly, I just mirrored what was done in runtime installers but results of both operations would be the same. Also, Debian case uses token replacement and doesn't allow `Split`

JunTaoLuo: Feels strange that they are slightly different but okay.

JunTaoLuo (approved):

natemcmaster:
Sorry for reviewing after you merged, but I think this code won't work for patches. This assumes the runtime version always matches the targeting pack version. The targeting pack is not expected to rev beyond 3.0.0, so this will break in 3.0.1. We should add a separate variable for the targeting pack dependency that accounts for this difference.

JunTaoLuo:
Filed https://github.com/aspnet/AspNetCore/issues/8633.

PR #8517, opened 2019-03-14
Make Kestrels Internals truly internal. 

### LINKED ISSUE #8306: Remove pubternal types from Kestrel
MVC recently made all pubinternal types in MVC internal via https://github.com/aspnet/AspNetCore/issues/4932. As a reminder, `pubinternal` types are types that are declared as public but put in an .Internal namespace. 

In ASP.NET Core 3.0, we should do the same in Kestrel, specifically for the Microsoft.AspNetCore.Server.Kestrel.Core.Internal namespace. Reference assemblies will also show any pubinternal changes.

cc @muratg @halter73 @Tratcher

jkotalik:
Fixes #8306. Big change here so I'll give you the rundown.

**Big Picture:**
- Removes all .Internal namespaces and moves appropriate methods to internal
- Makes all types in Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions public. These types are already exposed on public API
- Makes KestrelServerOptionsSetup public

**Nitty Gritty:**
- Any method that had a pubternal API was either changed to internal or public accordingly
- Tests that took in enums that were changed from public to internal were changed appropriately.

For any public API, we probably need to add xml comments appropriately.


src/Servers/Kestrel/Core/ref/Microsoft.AspNetCore.Server.Kestrel.Core.netcoreapp3.0.cs
```diff
@@ -140,6 +139,11 @@ public partial class KestrelServerOptions
         public void ListenUnixSocket(string socketPath) { }
         public void ListenUnixSocket(string socketPath, System.Action<Microsoft.AspNetCore.Server.Kestrel.Core.ListenOptions> configure) { }
     }
+    public partial class KestrelServerOptionsSetup : Microsoft.Extensions.Options.IConfigureOptions<Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServerOptions>
```
pakrym: Shouldn't be public.

jkotalik:
We should also probably have an API review for these changes if we decide to many anything truly public.

@sebastienros let's also make sure this doesn't break the platformbenchmark scenario.

pranavkm (approved):
Code changes look great. I'd get sign off from someone more familiar with the API being made public

sebastienros:
I don't know exactly what will be broken, but chances are that this will break it. Here is a list of internal namespace usages:

https://github.com/aspnet/AspNetCore/blob/master/src/Servers/Kestrel/perf/PlatformBenchmarks/Program.cs#L7

https://github.com/aspnet/AspNetCore/blob/master/src/Servers/Kestrel/perf/PlatformBenchmarks/IHttpConnection.cs#L6

https://github.com/aspnet/AspNetCore/blob/master/src/Servers/Kestrel/perf/PlatformBenchmarks/BenchmarkApplication.cs#L7

https://github.com/aspnet/AspNetCore/blob/master/src/Servers/Kestrel/perf/PlatformBenchmarks/BenchmarkApplication.HttpConnection.cs#L9

https://github.com/aspnet/AspNetCore/blob/master/src/Servers/Kestrel/perf/PlatformBenchmarks/AsciiString.cs#L6

I would suggest to also change PlatformBenchmarks at the same time such that it will work with this PR, or adapt the bits that are used to make them public. Ideally some parts of PlaintextBenchmarks should be in Kestrel and directly reusable in the app.

sebastienros:
Looking at your changes I assume it's actually a good change for the PlatformBenchmarks as some use types will now become really public. Still good if some of the things in PB were in Kestrel directly.

src/Servers/Kestrel/Core/src/Adapter/ConnectionAdapterContext.cs
```diff
@@ -1,11 +1,11 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.IO;
 using Microsoft.AspNetCore.Connections;
 using Microsoft.AspNetCore.Http.Features;
 
-namespace Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal
```
halter73: I'm not sure we should be changing the namespace unless we're actually making the type internal.

src/Servers/Kestrel/Core/src/Adapter/Internal/LoggingConnectionAdapter.cs
```diff
@@ -6,9 +6,9 @@
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 
-namespace Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal
+namespace Microsoft.AspNetCore.Server.Kestrel.Core.Adapter
 {
-    public class LoggingConnectionAdapter : IConnectionAdapter
+    internal class LoggingConnectionAdapter : IConnectionAdapter
```
halter73: Why are we bothering to change the namespaces? Shouldn't making the classes internal be enough?

src/Servers/Kestrel/Core/ref/Microsoft.AspNetCore.Server.Kestrel.Core.netcoreapp3.0.cs
```diff
@@ -88,7 +87,7 @@ public enum HttpProtocols
     }
     public partial class KestrelServer : Microsoft.AspNetCore.Hosting.Server.IServer, System.IDisposable
     {
-        public KestrelServer(Microsoft.Extensions.Options.IOptions<Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServerOptions> options, Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.Internal.ITransportFactory transportFactory, Microsoft.Extensions.Logging.ILoggerFactory loggerFactory) { }
+        public KestrelServer(Microsoft.Extensions.Options.IOptions<Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServerOptions> options, Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.ITransportFactory transportFactory, Microsoft.Extensions.Logging.ILoggerFactory loggerFactory) { }
```
halter73: I would not change the namespace of anything that remains public as part of this PR. Maybe we can do that as a follow up.

halter73 (changes requested):
YESSSS! Overall, I'm super happy about this change.

src/Servers/Kestrel/Core/src/Adapter/Internal/LoggingConnectionAdapter.cs
jkotalik: That's what @pranavkm did with his PRs. Keeping consistent.

src/Servers/Kestrel/Core/ref/Microsoft.AspNetCore.Server.Kestrel.Core.netcoreapp3.0.cs
jkotalik: Sure.

src/Servers/Kestrel/Core/src/Adapter/ConnectionAdapterContext.cs
jkotalik: These honestly seemed like mistakes, I don't believe they should be pubternal to begin with (they are using in samples and public API). I can revert it for now if you prefer.

src/Servers/Kestrel/Core/src/Adapter/Internal/LoggingConnectionAdapter.cs
halter73: Did @pranavkm still keep the classes moved out of the Internal namespace in 
"Internal" directories? I think it's better to have the namespace structure match the directory structure if possible.

src/Servers/Kestrel/Core/src/Adapter/ConnectionAdapterContext.cs
halter73: These aren't in internal namespaces by accident. It's only a quasi-public, pubternal API currently because we didn't want to commit to a final API before.

Now is definitely not the time to commit to a final API since one of @davidfowl's goals for 3.0 is to migrate connection adapters to project bedrock connection middleware. That will hopefully mean deleting this API entirely.

src/Servers/Kestrel/Core/src/Adapter/Internal/LoggingConnectionAdapter.cs
pranavkm: I emptied .Internal folder along with changing the namespace. In MVC it helped do some code organization since we typically had one .Internal namespace per project which acted as a dumping ground for all kinds of features.

src/Servers/Kestrel/Core/src/Adapter/ConnectionAdapterContext.cs
jkotalik: Got it, I'll do the needful.

src/Servers/Kestrel/Core/src/Adapter/Internal/LoggingConnectionAdapter.cs
jkotalik: I see. @halter73 I agree with you then with keeping the .Internal namespace as it isn't really a dumping ground for us.

davidfowl:
This is going to make it impossible to write a kestrel transport right?

halter73:
> This is going to make it impossible to write a kestrel transport right?

The plan is to leave the transport abstractions and a few other things like connection adapters pubternal in this first pass.

jkotalik:
Updated with feedback. I think this is looking muchhhhh better.

src/Servers/Kestrel/Core/src/BadHttpRequestException.cs
```diff
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
```
halter73: I wish git better tools around normalizing BOMs similar to what it has for line endings gitattributes.

halter73 (approved):

PR #8518, opened 2019-03-14
Display startup errors in ANCM

pakrym:


src/Servers/IIS/IIS/samples/NativeIISSample/Startup.cs
```diff
@@ -117,14 +118,22 @@ public void Configure(IApplicationBuilder app)
...
         {
-            var host = new WebHostBuilder()
-                .UseKestrel()
-                .UseIIS()
-                .UseIISIntegration()
-                .UseStartup<Startup>()
-                .Build();
-
-            host.Run();
+            AppDomain.CurrentDomain.FirstChanceException += (sender, eventArgs) => {
```
pakrym: Had to use `FirstChanceException` instead of `UnhandledException` because we don't have https://github.com/dotnet/coreclr/pull/23054 yet.

src/Servers/IIS/IIS/samples/NativeIISSample/Startup.cs
```diff
@@ -117,14 +118,22 @@ public void Configure(IApplicationBuilder app)
...
+                Console.WriteLine("1");
+                NativeMethods.HttpSetStartupErrorPageContent("<pre>" + HtmlEncoder.Default.Encode(eventArgs.Exception.ToString()) + "</pre>");
```
pakrym: We need to decide the design of this page. Maybe reuse DeveloperExceptionPage code generation.

pakrym: Unfortunately, it has more dependencies then we might want.

Tratcher:
👀

jkotalik:
I'll be taking this over eventually, but we still need some downstream dependencies.

src/Servers/IIS/AspNetCoreModuleV2/InProcessRequestHandler/StartupExceptionApplication.h
```diff
@@ -27,18 +29,20 @@ class StartupExceptionApplication : public InProcessApplicationBase
 
     HRESULT CreateHandler(IHttpContext *pHttpContext, IREQUEST_HANDLER ** pRequestHandler)
     {
-        *pRequestHandler = new ServerErrorHandler(*pHttpContext, 500, 30, "Internal Server Error", m_HR, m_moduleInstance, m_disableLogs, IN_PROCESS_RH_STATIC_HTML);
-        return S_OK;
-    }
+        if (m_errorPageContent.length() > 0)
```
jkotalik: I know security is a concern here, so I propose we make this error page only surface when running behind IISExpress or the development environment variable is set.

Tratcher: IIS Express is an interesting way of checking this. 

ASPNETCORE_DETAILEDERRORS is an existing variable we use for the hosting error page. I'd be leary about using Development.

jkotalik:
I have a prototype of this using replicating the hosting exception page. Currently looks like:

image


As @pakrym  mentioned, this unfortunately brings in some dependencies that will be present in every inproc app. Here are the usings I have for the startup hook.

```c#
using System;
using System.IO;
using System.Linq;
using System.Reflection;
using System.Text;
using System.Text.Encodings.Web;
using Microsoft.AspNetCore.Server.IIS;
using Microsoft.Extensions.FileProviders;
using Microsoft.Extensions.StackTrace.Sources;
```

Tratcher:
Hosting already has all of those dependencies, no? You wouldn't be adding anything new.

src/Servers/IIS/IIS/src/StartupHook.cs
```diff
@@ -0,0 +1,129 @@
...
+            model.OperatingSystemDescription = RuntimeInformation.OSDescription;
+
+            var exceptionDetailProvider = new ExceptionDetailsProvider(
+                new PhysicalFileProvider(contentRoot),
+                sourceCodeLineCount: 6);
+
+            model.ErrorDetails = exceptionDetailProvider.GetDetails(exception);
+
+            var errorPage = new ErrorPage(model);
+            var context = new IntermediateHttpContext();
```
Tratcher: I expect you can get rid of this mock by adding an overload of ExecuteAsync that writes directly to a Stream.

jkotalik: Yeah that would probably be cleaner.

jkotalik: I still may need to mock the response. By default, it sets Response.ContentType to text/html no matter what. I'm going to leave the mock for now.

Tratcher: That line is here (https://github.com/aspnet/AspNetCore/blob/482577287429ca72a1d10fd367fe43556e3621ac/src/Shared/ErrorPage/ErrorPage.cshtml#L18) and would be easy to move out.

jkotalik:
Blocked on https://github.com/dotnet/coreclr/pull/23178 and https://github.com/dotnet/core-setup/issues/5556

jkotalik:
This is now working. Probably a bit of cleanup around the edges (like using host properties rather than environment variables), but overall it looks good.

Here is what the exception page looks like:
image

And showing the .NET Runtime event log:

image

src/Servers/IIS/AspNetCoreModuleV2/CommonLib/ServerErrorHandler.h
```diff
@@ -9,31 +9,40 @@
 class ServerErrorHandler : public REQUEST_HANDLER
 {
 public:
+    ServerErrorHandler(IHttpContext& pContext, USHORT statusCode, USHORT subStatusCode, std::string statusText, HRESULT hr, HINSTANCE module, bool disableStartupPage, BYTE* content, int length) noexcept
+        : ServerErrorHandler(pContext, statusCode, subStatusCode, statusText, hr, module, disableStartupPage, 0, content, length)
+    {
+    }
+
+    ServerErrorHandler(IHttpContext& pContext, USHORT statusCode, USHORT subStatusCode, std::string statusText, HRESULT hr, HINSTANCE module, bool disableStartupPage, int page) noexcept
```
jkotalik: There are two scenarios here. One which loads from a resource file (page number passed in) and the other is passed in via the startup hook. I have a 3rd constructor which takes both.

src/Servers/IIS/IIS/src/StartupHook.cs
```diff
@@ -0,0 +1,77 @@
...
+    public static void Initialize()
+    {
+        if (Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT") != "Development")
+        {
+            return;
+        }
+
+        AppDomain.CurrentDomain.UnhandledException += (sender, eventArgs) =>
+        {
+            var exception = (Exception)eventArgs.ExceptionObject;
```
jkotalik: This is mostly taken from https://github.com/aspnet/AspNetCore/blob/4e44025a52e4b73aa17e09a8041b0e166e0c5ce0/src/Hosting/Hosting/src/Internal/WebHost.cs#L245

src/Servers/IIS/AspNetCoreModuleV2/InProcessRequestHandler/StartupExceptionApplication.h
jkotalik: I did ASPNETCORE_ENVIRONMENT for now because it is the only this that is on by default as an environment variable. There are 3 different things we can key off of: DetailedErrors, CaptureStartupErrors, and Environment. I personally think CaptureStartupErrors is the best thing to use, but I don't see this as an environment variable anywhere.

src/Servers/IIS/AspNetCoreModuleV2/InProcessRequestHandler/inprocessapplication.cpp
```diff
@@ -235,8 +235,17 @@ IN_PROCESS_APPLICATION::ExecuteApplication()
             LOG_INFOF(L"Setting current directory to %s", this->QueryApplicationPhysicalPath().c_str());
         }
 
-        bool clrThreadExited;
+        if (m_pConfig->QueryCallStartupHook())
+        {
+            // Used to display developer exception page when there is an exception in main.
+            // TODO should we preserve whatever is in the environment variable before setting it?
+            SetEnvironmentVariable(L"DOTNET_STARTUP_HOOKS", L"Microsoft.AspNetCore.Server.IIS");
```
anurse: Could you run the following exploratory tests and report back your findings? I want to make sure we understand the behavior of the feature:

1. What happens in the browser, on the first request, if this assembly doesn't exist (i.e. try renaming it to a bogus name)
1. What happens in the browser, on the first request, if this assembly throws from inside it's StartupHook
1. What happens in the browser, on the first request, if you have a startup error **and** you throw from within the UnhandledException handler itself.

I recognize these are super edge cases, but I want to make sure we know what's going on. If there's any way to detect these cases, we can at be prepared to diagnose issues.

src/Servers/IIS/AspNetCoreModuleV2/InProcessRequestHandler/inprocessapplication.cpp
```diff
@@ -235,8 +235,17 @@ IN_PROCESS_APPLICATION::ExecuteApplication()
             LOG_INFOF(L"Setting current directory to %s", this->QueryApplicationPhysicalPath().c_str());
         }
 
-        bool clrThreadExited;
+        if (m_pConfig->QueryCallStartupHook())
+        {
+            // Used to display developer exception page when there is an exception in main.
+            // TODO should we preserve whatever is in the environment variable before setting it?
```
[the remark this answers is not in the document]
davidfowl: It needs to be appeneded.

src/Servers/IIS/AspNetCoreModuleV2/InProcessRequestHandler/managedexports.cpp
```diff
@@ -504,4 +508,17 @@ set_main_handler(_In_ hostfxr_main_fn main)
     IN_PROCESS_APPLICATION::SetMainCallback(main);
 }
 
+EXTERN_C __MIDL_DECLSPEC_DLLEXPORT
+VOID
+http_set_startup_error_page_content(_In_ byte* errorPageContent, int length)
+{
+    // Only set the response if we are running under IISExpress for security reasons.
+    if (g_pHttpServer->IsCommandLineLaunch())
```
davidfowl: Remove this check.

src/Servers/IIS/IIS/src/StartupHook.cs
```diff
@@ -0,0 +1,77 @@
...
+
+internal class StartupHook
+{
+    /// <summary>
+    /// Startup hooks are pieces of code that will run before a users program main executes
+    /// See: https://github.com/dotnet/core-setup/blob/master/Documentation/design-docs/host-startup-hook.md
+    /// </summary>
+    public static void Initialize()
+    {
+        if (Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT") != "Development")
```
davidfowl: https://github.com/aspnet/Extensions/blob/97d7856ea482015209d60cf8955bacf8053f4cb5/src/Hosting/Abstractions/src/HostEnvironmentEnvExtensions.cs#L64

Few things:
- `DOTNET_ENVIRONMENT` and `ASPNETCORE_ENVIRONMENT`
- The comparison should be case insensitive.

src/Servers/IIS/IIS/src/StartupHook.cs
```diff
@@ -0,0 +1,77 @@
...
+        {
+            return;
+        }
+
+        AppDomain.CurrentDomain.UnhandledException += (sender, eventArgs) =>
+        {
+            var exception = (Exception)eventArgs.ExceptionObject;
+
+            // Get the content root from IIS.
+            var iisConfigData = NativeMethods.HttpGetApplicationProperties();
```
davidfowl: Can this code fail in anyway? Should it be in a try/catch?

src/Servers/IIS/IIS/src/StartupHook.cs
```diff
@@ -0,0 +1,77 @@
...
+                new PhysicalFileProvider(contentRoot),
+                sourceCodeLineCount: 6);
+
+            model.ErrorDetails = exceptionDetailProvider.GetDetails(exception);
+
+            var errorPage = new ErrorPage(model);
+
+            var stream = new MemoryStream();
+            // Sync over async here, but you can't have async code in startup hooks.
+            errorPage.ExecuteAsync(stream).GetAwaiter().GetResult();
```
davidfowl: It has nothing to do with the startup hook. It's to do with the callback. It'll never go async because we're writing to a memory stream. Put that in the comment instead.

src/Servers/IIS/IIS/src/StartupHook.cs
```diff
@@ -0,0 +1,77 @@
...
+
+            // Get the raw content and set the error page.
+            stream.Position = 0;
+            var content = stream.ToArray();
+
+            NativeMethods.HttpSetStartupErrorPageContent(content);
```
davidfowl: Is it possible to be running on a version of ANCM without this export? What happens in that case?

src/Servers/IIS/IIS/src/StartupHook.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        AppDomain.CurrentDomain.UnhandledException += (sender, eventArgs) =>
+        {
+            var exception = (Exception)eventArgs.ExceptionObject;
+
+            // Get the content root from IIS.
+            var iisConfigData = NativeMethods.HttpGetApplicationProperties();
+            var contentRoot = iisConfigData.pwzFullApplicationPath.TrimEnd(Path.DirectorySeparatorChar);
+
+            var model = new ErrorPageModel
```
davidfowl: Is this using the hosting page or developer exception page?

src/Servers/IIS/AspNetCoreModuleV2/InProcessRequestHandler/inprocessapplication.cpp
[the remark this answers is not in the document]
jkotalik: It needs to be appended until we have runtime property settings. I don't think a user can configure those at all.

src/Servers/IIS/AspNetCoreModuleV2/InProcessRequestHandler/inprocessapplication.cpp
jkotalik: 1. The app fails to start.
2. The app fails to start.
3. From what I can tell, it just causes the UnhandledException handler to exit early.

src/Servers/IIS/IIS/src/StartupHook.cs
jkotalik: Hosting.

src/Servers/IIS/IIS/src/StartupHook.cs
jkotalik: Possible yes, but it would require doing crazy things. Both the managed and native dlls are in the same package, so you can't reference different versions if you tried. It would require you copy a different inprocess native dll to the published/build directory. In that case, this would throw. Trying to cover this case is pointless. 

The only thing I do want to protect against is making sure we are actually running inprocess in the startup hook.

src/Servers/IIS/IIS/src/StartupHook.cs
jkotalik: If we are running inprocess, no it can't fail.

jkotalik: From what I can tell, throwing in the UnhandledException handler doesn't propagate to the client.

src/Servers/IIS/AspNetCoreModuleV2/InProcessRequestHandler/StartupExceptionApplication.h
```diff
@@ -15,32 +15,32 @@ class StartupExceptionApplication : public InProcessApplicationBase
...
 
     ~StartupExceptionApplication() = default;
 
-    HRESULT CreateHandler(IHttpContext *pHttpContext, IREQUEST_HANDLER ** pRequestHandler)
+    HRESULT CreateHandler(IHttpContext* pHttpContext, IREQUEST_HANDLER** pRequestHandler)
     {
-        *pRequestHandler = new ServerErrorHandler(*pHttpContext, 500, 30, "Internal Server Error", m_HR, m_moduleInstance, m_disableLogs, IN_PROCESS_RH_STATIC_HTML);
-        return S_OK;
-    }
+        *pRequestHandler = new ServerErrorHandler(*pHttpContext, 500, 30, "Internal Server Error", m_HR, m_moduleInstance, m_disableLogs, IN_PROCESS_RH_STATIC_HTML, m_errorPageContent, m_length);
```
Tratcher: Content-Type?

src/Servers/IIS/IIS/samples/NativeIISSample/Startup.cs
```diff
@@ -117,14 +117,15 @@ public void Configure(IApplicationBuilder app)
...
         {
-            var host = new WebHostBuilder()
-                .UseKestrel()
-                .UseIIS()
-                .UseIISIntegration()
-                .UseStartup<Startup>()
-                .Build();
-
-            host.Run();
+            throw new Exception();
```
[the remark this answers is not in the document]
jkotalik: Bad addition... This is a sample

src/Servers/IIS/IIS/src/StartupHook.cs
```diff
@@ -0,0 +1,87 @@
...
+    {
+        if (!NativeMethods.IsAspNetCoreModuleLoaded())
+        {
+            // This means someone set the startup hook for Microsoft.AspNetCore.Server.IIS
+            // but are not running inprocess. Return at this point.
+            return;
+        }
+
+        if (Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT").Equals("Development", StringComparison.OrdinalIgnoreCase) ||
+            Environment.GetEnvironmentVariable("DOTNET_ENVIRONMENT").Equals("Development", StringComparison.OrdinalIgnoreCase))
```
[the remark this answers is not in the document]
jkotalik: @davidfowl and @anurse didn't like running on just IISExpress. Again, ideally we would use CaptureStartupErrors from hosting to decide this.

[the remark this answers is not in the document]
davidfowl: We can also use `ASPNETCORE_DETAILEDERRORS`. This needs to work in development for both IIS and IIS Express.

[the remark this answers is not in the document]
Tratcher: CaptureStartupErrors is too late, but ASPNETCORE_DETAILEDERRORS would work.

src/Servers/IIS/IIS/src/StartupHook.cs
davidfowl: OK. So in practice, this should never throw

src/Servers/IIS/IIS/src/StartupHook.cs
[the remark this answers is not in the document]
jkotalik: I don't want to rely just on ASPNETCORE_DETAILEDERRORS as it isn't default.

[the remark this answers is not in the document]
jkotalik: I'd be okay with checking both Environment and DetailedErrors.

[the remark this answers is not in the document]
davidfowl: Not sure what you mean. @Tratcher is asking how somebody turns it *on* on other environments, not off. By default, if you aren't in *Development*  (Production is our default), then you won't get the error page. You can override that by setting ASPNETCORE_DETAILEDERRORS  to true/1.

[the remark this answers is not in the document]
jkotalik: Sorry I was unclear in what I said above. I agree with what you just said.

src/Servers/IIS/AspNetCoreModuleV2/InProcessRequestHandler/StartupExceptionApplication.h
jkotalik: It's set here: https://github.com/aspnet/AspNetCore/pull/8518/files#diff-1b6e0cba731b19c6281127415e6d3b0fR56. Think we should be setting charset=utf-8?

jkotalik:
🆙 📅

davidfowl (approved):

src/Servers/IIS/AspNetCoreModuleV2/CommonLib/ServerErrorHandler.h
```diff
@@ -9,31 +9,40 @@
 class ServerErrorHandler : public REQUEST_HANDLER
 {
 public:
+    ServerErrorHandler(IHttpContext& pContext, USHORT statusCode, USHORT subStatusCode, std::string statusText, HRESULT hr, HINSTANCE module, bool disableStartupPage, BYTE* content, int length) noexcept
```
BrennanConroy: nit: use `const std::string&` instead of `std::string` to avoid the string copies.

src/Servers/IIS/AspNetCoreModuleV2/InProcessRequestHandler/managedexports.cpp
```diff
@@ -504,4 +507,12 @@ set_main_handler(_In_ hostfxr_main_fn main)
     IN_PROCESS_APPLICATION::SetMainCallback(main);
 }
 
+EXTERN_C __MIDL_DECLSPEC_DLLEXPORT
+VOID
+http_set_startup_error_page_content(_In_ byte* errorPageContent, int length)
+{
+    g_errorPageContent.resize(g_errorPageContent.size() + length);
+    memcpy(&g_errorPageContent[g_errorPageContent.size() - length], errorPageContent, length);
```
BrennanConroy: looks like off-by-one error. Also size will be 0 so lets just use that

anurse:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=160355

src/Servers/IIS/AspNetCoreModuleV2/InProcessRequestHandler/dllmain.cpp
```diff
@@ -128,7 +129,7 @@ CreateApplication(
             std::unique_ptr<InProcessOptions> options;
             THROW_IF_FAILED(InProcessOptions::Create(*pServer, pSite, *pHttpApplication, options));
             // Set the currently running application to a fake application that returns startup exceptions.
-            auto pErrorApplication = std::make_unique<StartupExceptionApplication>(*pServer, *pHttpApplication, g_hServerModule, options->QueryDisableStartUpErrorPage(), hr);
+            auto pErrorApplication = std::make_unique<StartupExceptionApplication>(*pServer, *pHttpApplication, g_hServerModule, options->QueryDisableStartUpErrorPage(), hr, g_errorPageContent);
```
BrennanConroy: If `g_errorPageContent` is only ever used once, and only right here, then we can `std::move` it all the way to where it's assigned and avoid any copies.

src/Servers/IIS/AspNetCoreModuleV2/InProcessRequestHandler/inprocessapplication.h
```diff
@@ -15,6 +15,9 @@ typedef BOOL(WINAPI * PFN_SHUTDOWN_HANDLER) (void* pvShutdownHandlerContext);
 typedef REQUEST_NOTIFICATION_STATUS(WINAPI * PFN_ASYNC_COMPLETION_HANDLER)(void *pvManagedHttpContext, HRESULT hrCompletionStatus, DWORD cbCompletion);
 typedef void(WINAPI * PFN_REQUESTS_DRAINED_HANDLER) (void* pvShutdownHandlerContext);
 
+
```
BrennanConroy: nit: formatting

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/2300

jkotalik:
I'm going to try verifying this change E2E ASAP. I'm confident in the change, but I would like to make sure this works as part of the SDK.

PR #8525, opened 2019-03-14
Make ResetSendPing and ResetTimeout private methods on HubConnection

mikaelm12:
Fixes: https://github.com/aspnet/AspNetCore/issues/7977

Making these methods private as they are only intended for internal use and shouldn't be exposed. Will file an announcement after this is merged since it is a breaking change but it shouldn't be a huge deal.


halter73 (approved):

mikaelm12:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=124543&view=logs

BrennanConroy (approved):

analogrelay (approved):

PR #8529, opened 2019-03-14
Atomic swap the request drained handler when shutting down.

jkotalik:
For https://github.com/aspnet/AspNetCore-Internal/issues/1964 and https://github.com/aspnet/AspNetCore-Internal/issues/1958


pakrym (approved):

jkotalik:
There was a failure with an ANCM test that this fix was originally dealing with. I realized now that if app_offline is dropped too early, the IIS site will never shut down because it creates a polling application.

I'm going to make it so the test only runs with IIS Express and rewrite a test that works with IIS.

src/Servers/IIS/IIS/test/Common.FunctionalTests/Utilities/IISFunctionalTestBase.cs
```diff
@@ -17,7 +17,7 @@ namespace Microsoft.AspNetCore.Server.IIS.FunctionalTests.Utilities
     public class IISFunctionalTestBase : FunctionalTestsBase
     {
         protected static readonly TimeSpan RetryDelay = TimeSpan.FromMilliseconds(100);
-        protected PublishedSitesFixture _fixture;
+        protected PublishedSitesFixture _fixture { get; set; }
```
pakrym: `Fixture`

Issue #8230, opened 2019-03-06
Not Modified (304) returns Content-type


asn19595:
Content-type header is set when the response is 'NotModified' by FileStreamResult. Which strange according to 
https://tools.ietf.org/html/rfc7232#section-4.1

1. Using this version of ASP.NET Core 'v2.2.2'
2. Run this test 'WriteFileAsync_NotModified_RangeRequestedIgnored()' from 'FileStreamResultTest.cs'
https://github.com/aspnet/Mvc/blob/65af12f1f575fdaee893e0232416df29bb83d7fa/test/Microsoft.AspNetCore.Mvc.Core.Test/FileStreamResultTest.cs#L401
3. Add an Assert for Assert.Empty(httpResponse.Headers[HeaderNames.ContentType]);
4. The test fails as it contains the Content-type header with value according to the stream.

Expected behavior:
SHOULD NOT generate representation metadata other than the above listed fields (Content-type)
https://tools.ietf.org/html/rfc7232#section-4.1

mkArtakMSFT:
Thanks for contacting us, @asn19595.
We would happily consider a PR for this.

ChiragRupani:
I would like to work on this issue. I have done setup and also identified the fix (https://github.com/aspnet/AspNetCore/blob/a673be3b9a2464cb20351f381e189a43beb6f426/src/Mvc/Mvc.Core/src/Infrastructure/FileResultExecutorBase.cs#L75) - either not set content type before it or set to Empty in it. I am verifying it.
Does it require to fix for other IActionResultExecutor classes as well?

asn19595:
> I would like to work on this issue. I have done setup and also identified the fix at line
> [AspNetCore/src/Mvc/Mvc.Core/src/Infrastructure/FileResultExecutorBase.cs](https://github.com/aspnet/AspNetCore/blob/a673be3b9a2464cb20351f381e189a43beb6f426/src/Mvc/Mvc.Core/src/Infrastructure/FileResultExecutorBase.cs#L75)
> 
> Line 75 in [a673be3](/aspnet/AspNetCore/commit/a673be3b9a2464cb20351f381e189a43beb6f426)
> 
>  if (preconditionState == PreconditionState.NotModified) 
> 
> Fix is to either not to set content type before it or set to Empty in it. Should I work on it?
> Does it require to fix for other IActionResultExecutor classes as well?

I have already a PR in process for this. Thank anyways.

pranavkm:
[...]

PR #8545, opened 2019-03-15
Set ContentType to null if the response is NotModified

ajinkya-nikam:
Set ContentType to null when we set StatusCode to 304 (not Modified)

Addresses #8230

### DIFF
--- a/src/Mvc/Mvc.Core/src/Infrastructure/FileResultExecutorBase.cs
+++ b/src/Mvc/Mvc.Core/src/Infrastructure/FileResultExecutorBase.cs
@@ -75,6 +75,7 @@ namespace Microsoft.AspNetCore.Mvc.Infrastructure
             if (preconditionState == PreconditionState.NotModified)
             {
                 response.StatusCode = StatusCodes.Status304NotModified;
+                response.ContentType = null;
                 return (range: null, rangeLength: 0, serveBody: false);
             }
             else if (preconditionState == PreconditionState.PreconditionFailed)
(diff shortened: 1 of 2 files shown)


src/Mvc/Mvc.Core/src/Infrastructure/FileResultExecutorBase.cs
```diff
@@ -75,6 +75,7 @@ internal enum PreconditionState
             if (preconditionState == PreconditionState.NotModified)
             {
                 response.StatusCode = StatusCodes.Status304NotModified;
+                response.ContentType = null;
```
asn19595: Set ContentType to null if the response is NotModified.

src/Mvc/Mvc.Core/test/FileStreamResultTest.cs
```diff
@@ -437,6 +437,7 @@ public async Task WriteFileAsync_NotModified_RangeRequestedIgnored()
             Assert.Null(httpResponse.ContentLength);
             Assert.Empty(httpResponse.Headers[HeaderNames.ContentRange]);
             Assert.NotEmpty(httpResponse.Headers[HeaderNames.LastModified]);
+            Assert.Empty(httpResponse.Headers[HeaderNames.ContentType]);
```
asn19595: Check for ContentType not being set.

mkArtak:
Thanks for your effort, @asn19595.
We have a process in place for accepting community PRs, and it starts from filing an issue. We engage into discussion with community members on the issues first to determine what are the problems the person is facing and what are the options for addressing it.
After finding out that the issue is worth fixing, then we move forward with submitting a PR.

Given that you already have a PR, please file an issue with the details of the problem and link it from this PR.

asn19595:
> Thanks for your effort, @asn19595.
> We have a process in place for accepting community PRs, and it starts from filing an issue. We engage into discussion with community members on the issues first to determine what are the problems the person is facing and what are the options for addressing it.
> After finding out that the issue is worth fixing, then we move forward with submitting a PR.
> 
> Given that you already have a PR, please file an issue with the details of the problem and link it from this PR.

Sorry was new to the process. Addressed the issue that is already been filed by me. Thanks.

pranavkm:
There's an ongoing PR that addresses this in a better way: https://github.com/aspnet/AspNetCore/pull/8557. Closing this in favor of the other.

PR #8557, opened 2019-03-15
Removed Content Type header for 304 response since no content is being returned

ChiragRupani:
Summary of the changes
 - Removes Content Type header when preconditional headers process to 304 (NotModified)

Addresses #8230

### DIFF
--- a/src/Mvc/Mvc.Core/src/Infrastructure/FileResultExecutorBase.cs
+++ b/src/Mvc/Mvc.Core/src/Infrastructure/FileResultExecutorBase.cs
@@ -71,7 +71,7 @@ namespace Microsoft.AspNetCore.Mvc.Infrastructure
             // Short circuit if the preconditional headers process to 304 (NotModified) or 412 (PreconditionFailed)
             if (preconditionState == PreconditionState.NotModified)
             {
-                response.StatusCode = StatusCodes.Status304NotModified;                
+                response.StatusCode = StatusCodes.Status304NotModified;
                 return (range: null, rangeLength: 0, serveBody: false);
             }
             else if (preconditionState == PreconditionState.PreconditionFailed)
@@ -87,7 +87,6 @@ namespace Microsoft.AspNetCore.Mvc.Infrastructure
             {
                 // Assuming the request is not a range request, and the response body is not empty, the Content-Length header is set to 
                 // the length of the entire file. 
-
                 // If the request is a valid range request, this header is overwritten with the length of the range as part of the 
                 // range processing (see method SetContentLength).
 
--- a/src/Mvc/Mvc.Core/src/Infrastructure/FileResultExecutorBase.cs
+++ b/src/Mvc/Mvc.Core/src/Infrastructure/FileResultExecutorBase.cs
@@ -52,10 +52,7 @@ namespace Microsoft.AspNetCore.Mvc.Infrastructure
             {
                 throw new ArgumentNullException(nameof(result));
             }
-
-            SetContentType(context, result);
-            SetContentDispositionHeader(context, result);
-
+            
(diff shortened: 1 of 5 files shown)


src/Mvc/Mvc.Core/src/Infrastructure/FileResultExecutorBase.cs
```diff
@@ -75,6 +75,7 @@ internal enum PreconditionState
             if (preconditionState == PreconditionState.NotModified)
             {
                 response.StatusCode = StatusCodes.Status304NotModified;
+                response.Headers.Remove("Content-Type");
```
pranavkm: @Tratcher does a 412 require the `Content-Type` header to be sent? If not, we could move these two lines https://github.com/aspnet/AspNetCore/blob/master/src/Mvc/Mvc.Core/src/Infrastructure/FileResultExecutorBase.cs#L56-L57 to before here - https://github.com/aspnet/AspNetCore/blob/master/src/Mvc/Mvc.Core/src/Infrastructure/FileResultExecutorBase.cs#L86.

Tratcher: No, 412 doesn't need it either. Moving SetContentType and SetContentDispositionHeader would be better than removing the value here.

pranavkm:
@ChiragRupani could you make the changes I suggested here: https://github.com/aspnet/AspNetCore/pull/8557#discussion_r266514066? I'd also recommend rebasing on master. We had some build infrastructure issues that are addressed in more recent commits.

src/Mvc/Mvc.Core/src/Infrastructure/FileResultExecutorBase.cs
ChiragRupani: Done

src/Mvc/Mvc.Core/src/Infrastructure/FileResultExecutorBase.cs
```diff
@@ -74,7 +71,7 @@ internal enum PreconditionState
             // Short circuit if the preconditional headers process to 304 (NotModified) or 412 (PreconditionFailed)
             if (preconditionState == PreconditionState.NotModified)
             {
-                response.StatusCode = StatusCodes.Status304NotModified;
+                response.StatusCode = StatusCodes.Status304NotModified;
```
pranavkm: Could you undo this change?

src/Mvc/Mvc.Core/src/Infrastructure/FileResultExecutorBase.cs
```diff
@@ -83,10 +80,14 @@ internal enum PreconditionState
...
             }
 
+            SetContentType(context, result);
+            SetContentDispositionHeader(context, result);
+
             if (fileLength.HasValue)
             {
                 // Assuming the request is not a range request, and the response body is not empty, the Content-Length header is set to 
                 // the length of the entire file. 
+
```
pranavkm: Undo please

src/Mvc/Mvc.Core/src/Infrastructure/FileResultExecutorBase.cs
ChiragRupani: Removed spaces

src/Mvc/Mvc.Core/src/Infrastructure/FileResultExecutorBase.cs
ChiragRupani: Removed spaces

Tratcher (approved):

pranavkm:
Thanks for the PR!

PR #8558, opened 2019-03-15
Kill java after build

ryanbrandenburg:
To prevent leaving java processes (such as Selenium) around lets add java to the list of processes that we kill when leaving an agent. Only on windows because we don't run Selenium on other OSs.


javiercn (approved):

dougbu (approved):

ryanbrandenburg:
Closing in favor of https://github.com/aspnet/AspNetCore/pull/8568

Issue #5282, opened 2017-10-26
Disconnect / Reconnect


aswanevelder:
Hi guys,

Are there javascript events available yet for Disconnect and Reconnect as was available in SignalR 2?

Thanks

anurse:
Since we're now doing this in 3.0. I've split this issue into separate ones for each client (since the work will be done per-client). I'm going to close this issue in lieu of those separate ones now.

PR #8566, opened 2019-03-15
Implement auto reconnect on SignalR TypeScript client

### LINKED ISSUE #8709: Support automatic reconnect in the TypeScript client
Epic #8797

Split from https://github.com/aspnet/AspNetCore/issues/5282

We will support automatic reconnect in 3.0 **without message replay**. We will also allow customization of the reconnect logic. This will be opt-in via a `HubConnectionBuilder` option. The default behavior will continue to be manual reconnection.

halter73:
Partially addresses #5282

Fixes #8709 (added by @anurse :trollface:)

### DIFF
--- a/src/SignalR/clients/ts/signalr/src/DefaultReconnectPolicy.ts
+++ b/src/SignalR/clients/ts/signalr/src/DefaultReconnectPolicy.ts
@@ -1,0 +1,20 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+import { IReconnectPolicy } from "./IReconnectPolicy";
+
+// 0, 2, 10, 30 second delays before reconnect attempts.
+const DEFAULT_RETRY_DELAYS_IN_MILLISECONDS = [0, 2000, 10000, 30000, null];
+
+/** @private */
+export class DefaultReconnectPolicy implements IReconnectPolicy {
+    private readonly retryDelays: Array<number | null>;
+
+    constructor(retryDelays?: number[]) {
+        this.retryDelays = retryDelays !== undefined ? [...retryDelays, null] : DEFAULT_RETRY_DELAYS_IN_MILLISECONDS;
+    }
+
+    public nextRetryDelayInMilliseconds(previousRetryCount: number): number | null {
+        return this.retryDelays[previousRetryCount];
+    }
+}
--- a/src/SignalR/clients/ts/signalr/src/HttpConnection.ts
+++ b/src/SignalR/clients/ts/signalr/src/HttpConnection.ts
@@ -14,10 +14,10 @@ import { WebSocketTransport } from "./WebSocketTransport";
 
 /** @private */
 const enum ConnectionState {
-    Connecting,
-    Connected,
-    Disconnected,
-    Disconnecting,
+    Connecting = "Connecting ",
+    Connected = "Connected",
+    Disconnected = "Disconnected",
+    Disconnecting = "Disconnecting",
 }
 
 /** @private */
--- a/src/SignalR/clients/ts/signalr/src/HttpConnection.ts
+++ b/src/SignalR/clients/ts/signalr/src/HttpConnection.ts
@@ -198,7 +198,7 @@ export class HttpConnection implements IConnection {
 
             this.transport = undefined;
         } else {
-            this.logger.log(LogLevel.Debug, "HttpConnection.transport became unset because start() failed.");
+            this.logger.log(LogLevel.Debug, "HttpConnection.transport is undefined in HttpConnection.stop() because start() failed.");
             this.stopConnection();
         }
     }
@@ -283,7 +283,6 @@ export class HttpConnection implements IConnection {
         } catch (e) {
             this.logger.log(LogLevel.Error, "Failed to start the connection: " + e);
             this.connectionState = ConnectionState.Disconnected;
-
             this.transport = undefined;
             return Promise.reject(e);
         }
--- a/src/SignalR/clients/ts/signalr/src/HttpConnection.ts
+++ b/src/SignalR/clients/ts/signalr/src/HttpConnection.ts
@@ -53,7 +53,7 @@ export class HttpConnection implements IConnection {
     // connectionStarted is tracked independently from connectionState, so we can check if the
     // connection ever did successfully transition from connecting to connected before disconnecting.
     private connectionStarted: boolean;
-    private baseUrl: string;
+    private readonly baseUrl: string;
     private readonly httpClient: HttpClient;
     private readonly logger: ILogger;
     private readonly options: IHttpConnectionOptions;
--- a/src/SignalR/clients/ts/signalr/src/HttpConnection.ts
+++ b/src/SignalR/clients/ts/signalr/src/HttpConnection.ts
@@ -17,6 +17,7 @@ const enum ConnectionState {
     Connecting,
     Connected,
     Disconnected,
+    Disconnecting,
 }
 
 /** @private */
@@ -49,16 +50,22 @@ if (Platform.isNode && typeof require !== "undefined") {
 /** @private */
 export class HttpConnection implements IConnection {
     private connectionState: ConnectionState;
+    // connectionStarted is tracked independently from connectionState, so we can check if the
+    // connection ever did successfully transition from connecting to connected before disconnecting.
+    private connectionStarted: boolean;
     private baseUrl: string;
     private readonly httpClient: HttpClient;
     private readonly logger: ILogger;
     private readonly options: IHttpConnectionOptions;
     private transport?: ITransport;
-    private startPromise?: Promise<void>;
+    private startInternalPromise?: Promise<void>;
+    private stopPromise?: Promise<void>;
+    private stopPromiseResolver!: (value?: PromiseLike<void>) => void;
     private stopError?: Error;
     private accessTokenFactory?: () => string | Promise<string>;
 
     public readonly features: any = {};
+    public connectionId?: string;
     public onreceive: ((data: string | ArrayBuffer) => void) | null;
     public onclose: ((e?: Error) => void) | null;
 
@@ -89,14 +96,16 @@ export class HttpConnection implements IConnection {
 
         this.httpClient = options.httpClient || new DefaultHttpClient(this.logger);
         this.connectionState = ConnectionState.Disconnected;
+        this.connectionStarted = false;
         this.options = options;
+
         this.onreceive = null;
         this.onclose = null;
     }
 
     public start(): Promise<void>;
     public start(transferFormat: TransferFormat): Promise<void>;
-    public start(transferFormat?: TransferFormat): Promise<void> {
+    public async start(transferFormat?: TransferFormat): Promise<void> {
         transferFormat = transferFormat || TransferFormat.Binary;
 
         Arg.isIn(transferFormat, TransferFormat, "transferFormat");
@@ -104,13 +113,32 @@ export class HttpConnection implements IConnection {
         this.logger.log(LogLevel.Debug, `Starting connection with transfer format '${TransferFormat[transferFormat]}'.`);
 
         if (this.connectionState !== ConnectionState.Disconnected) {
-            return Promise.reject(new Error("Cannot start a connection that is not in the 'Disconnected' state."));
+            return Promise.reject(new Error("Cannot start an HttpConnection that is not in the 'Disconnected' state."));
         }
 
         this.connectionState = ConnectionState.Connecting;
 
-        this.startPromise = this.startInternal(transferFormat);
-        return this.startPromise;
+        this.startInternalPromise = this.startInternal(transferFormat);
+        await this.startInternalPromise;
+
+        // The TypeScript compiler thinks that connectionState must be Connecting here. The TypeScript compiler is wrong.
+        if (this.connectionState as any === ConnectionState.Disconnecting) {
+            // stop() was called and transitioned the client into the Disconnecting state.
+            const message = "Failed to start the HttpConnection before stop() was called.";
+            this.logger.log(LogLevel.Error, message);
+
+            // We cannot await stopPromise inside startInternal since stopInternal awaits the startInternalPromise.
+            await this.stopPromise;
+
+            return Promise.reject(new Error(message));
+        } else if (this.connectionState as any !== ConnectionState.Connected) {
+            // stop() was called and transitioned the client into the Disconnecting state.
+            const message = "HttpConnection.startInternal completed gracefully but didn't enter the connection into the connected state!";
+            this.logger.log(LogLevel.Error, message);
+            return Promise.reject(new Error(message));
+        }
+
+        this.connectionStarted = true;
     }
 
     public send(data: string | ArrayBuffer): Promise<void> {
@@ -123,22 +151,55 @@ export class HttpConnection implements IConnection {
     }
 
     public async stop(error?: Error): Promise<void> {
-        this.connectionState = ConnectionState.Disconnected;
+        if (this.connectionState === ConnectionState.Disconnected) {
+            this.logger.log(LogLevel.Debug, `Call to HttpConnection.stop(${error}) ignored because the connection is already in the disconnected state.`);
+            return Promise.resolve();
+        }
+
+        if (this.connectionState === ConnectionState.Disconnecting) {
+            this.logger.log(LogLevel.Debug, `Call to HttpConnection.stop(${error}) ignored because the connection is already in the disconnecting state.`);
+            return this.stopPromise;
+        }
+
+        this.connectionState = ConnectionState.Disconnecting;
+
+        this.stopPromise = new Promise((resolve) => {
+            // Don't complete stop() until stopConnection() completes.
+            this.stopPromiseResolver = resolve;
+        });
+
+        // stopInternal should never throw so just observe it.
+        await this.stopInternal(error);
+        await this.stopPromise;
+    }
+
+    private async stopInternal(error?: Error): Promise<void> {
         // Set error as soon as possible otherwise there is a race between
         // the transport closing and providing an error and the error from a close message
         // We would prefer the close message error.
         this.stopError = error;
 
         try {
-            await this.startPromise;
+            await this.startInternalPromise;
         } catch (e) {
-            // this exception is returned to the user as a rejected Promise from the start method
+            // This exception is returned to the user as a rejected Promise from the start method.
         }
 
         // The transport's onclose will trigger stopConnection which will run our onclose event.
+        // The transport should always be set if currently connected. If it wasn't set, it's likely because
+        // stop was called during start() and start() failed.
         if (this.transport) {
-            await this.transport.stop();
+            try {
+                await this.transport.stop();
+            } catch (e) {
+                this.logger.log(LogLevel.Error, `HttpConnection.transport.stop() threw error '${e}'.`);
+                this.stopConnection();
+            }
+
             this.transport = undefined;
+        } else {
+            this.logger.log(LogLevel.Debug, "HttpConnection.transport became unset because start() failed.");
+            this.stopConnection();
         }
     }
 
@@ -157,7 +218,7 @@ export class HttpConnection implements IConnection {
                     // No fallback or negotiate in this case.
                     await this.transport!.connect(url, transferFormat);
                 } else {
-                    return Promise.reject(new Error("Negotiation can only be skipped when using the WebSocket transport directly."));
+                    throw new Error("Negotiation can only be skipped when using the WebSocket transport directly.");
                 }
             } else {
                 let negotiateResponse: INegotiateResponse | null = null;
@@ -166,16 +227,16 @@ export class HttpConnection implements IConnection {
                 do {
                     negotiateResponse = await this.getNegotiationResponse(url);
                     // the user tries to stop the connection when it is being started
-                    if (this.connectionState === ConnectionState.Disconnected) {
-                        return;
+                    if (this.connectionState === ConnectionState.Disconnecting || this.connectionState === ConnectionState.Disconnected) {
+                        throw new Error("The connection was stopped during negotiation.");
                     }
 
                     if (negotiateResponse.error) {
-                        return Promise.reject(new Error(negotiateResponse.error));
+                        throw new Error(negotiateResponse.error);
                     }
 
                     if ((negotiateResponse as any).ProtocolVersion) {
-                        return Promise.reject(new Error("Detected a connection attempt to an ASP.NET SignalR Server. This client only supports connecting to an ASP.NET Core SignalR Server. See https://aka.ms/signalr-core-differences for details."));
+                        throw new Error("Detected a connection attempt to an ASP.NET SignalR Server. This client only supports connecting to an ASP.NET Core SignalR Server. See https://aka.ms/signalr-core-differences for details.");
                     }
 
                     if (negotiateResponse.url) {
@@ -194,8 +255,10 @@ export class HttpConnection implements IConnection {
                 while (negotiateResponse.url && redirects < MAX_REDIRECTS);
 
                 if (redirects === MAX_REDIRECTS && negotiateResponse.url) {
-                    return Promise.reject(new Error("Negotiate redirection limit exceeded."));
+                    throw new Error("Negotiate redirection limit exceeded.");
                 }
+
+                this.connectionId = negotiateResponse.connectionId;
 
                 await this.createTransport(url, this.options.transport, negotiateResponse, transferFormat);
             }
@@ -207,12 +270,20 @@ export class HttpConnection implements IConnection {
             this.transport!.onreceive = this.onreceive;
             this.transport!.onclose = (e) => this.stopConnection(e);
 
-            // only change the state if we were connecting to not overwrite
-            // the state if the connection is already marked as Disconnected
-            this.changeState(ConnectionState.Connecting, ConnectionState.Connected);
+            if (this.connectionState === ConnectionState.Connecting) {
+                // Ensure the connection transitions to the connected state prior to completing this.startInternalPromise.
+                // start() will handle the case when stop was called and startInternal exits still in the disconnecting state.
+                this.logger.log(LogLevel.Debug, "The HttpConnection connected successfully.");
+                this.connectionState = ConnectionState.Connected;
+            }
+
+            // stop() is waiting on us via this.startInternalPromise so keep this.transport around so it can clean up.
+            // This is the only case startInternal can exit in neither the connected nor disconnected state because stopConnection()
+            // will transition to the disconnected state. start() will wait for the transition using the stopPromise.
         } catch (e) {
             this.logger.log(LogLevel.Error, "Failed to start the connection: " + e);
             this.connectionState = ConnectionState.Disconnected;
+
             this.transport = undefined;
             return Promise.reject(e);
         }
@@ -262,9 +333,6 @@ export class HttpConnection implements IConnection {
             this.transport = requestedTransport;
             await this.transport.connect(connectUrl, requestedTransferFormat);
 
-            // only change the state if we were connecting to not overwrite
-            // the state if the connection is already marked as Disconnected
-            this.changeState(ConnectionState.Connecting, ConnectionState.Connected);
             return;
         }
 
@@ -272,7 +340,6 @@ export class HttpConnection implements IConnection {
         const transports = negotiateResponse.availableTransports || [];
         for (const endpoint of transports) {
             try {
-                this.connectionState = ConnectionState.Connecting;
                 const transport = this.resolveTransport(endpoint, requestedTransport, requestedTransferFormat);
                 if (typeof transport === "number") {
                     this.transport = this.constructTransport(transport);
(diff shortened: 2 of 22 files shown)


bradygaster (approved):

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -54,13 +56,19 @@ export class HttpConnection implements IConnection {
...
     private startPromise?: Promise<void>;
     private stopError?: Error;
     private accessTokenFactory?: () => string | Promise<string>;
 
     public readonly features: any = {};
     public onreceive: ((data: string | ArrayBuffer) => void) | null;
     public onclose: ((e?: Error) => void) | null;
+    public onreconnecting: ((e?: Error) => void) | null;
+
+    // REVIEW: I would like to have a connectionId param here to highlight it will change,
```
mikaelm12: I like this idea

src/SignalR/clients/ts/signalr/src/DefaultReconnectPolicy.ts
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+import { IReconnectPolicy } from "./IReconnectPolicy";
+
+const DEFAULT_RETRY_DELAYS_IN_MILLISECONDS = [0, 2000, 10000, 30000];
+
+/** @private */
+export class DefaultReconnectPolicy implements IReconnectPolicy {
```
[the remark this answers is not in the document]
halter73: I like this idea. That certainly seems better than adding a bunch of overloads to withAutomaticReconnect.

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
JamesNK: Aren't there are times when the client just won't have the connectionId? e.g. an automatic, non-negotiated (or was it no handshake?) websockets connection

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -376,6 +413,83 @@ export class HttpConnection implements IConnection {
...
+
+                if (this.connectionState !== ConnectionState.Reconnecting) {
+                    return;
+                }
+            }
+
+            nextRetryDelay = reconnectPolicy.nextRetryDelayInMilliseconds(previousRetryCount++, Date.now() - startTime);
+        }
+
+        this.logger.log(LogLevel.Information, "Reconnect retry attempts have been exhausted. Connection disconnecting.");
```
JamesNK: IMO it would be useful info to have how many reconnect attempts before failure in the log message.

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -376,6 +413,83 @@ export class HttpConnection implements IConnection {
...
+            // Exit early if the onreconnecting callback called connection.stop().
+            if (this.connectionState !== ConnectionState.Reconnecting) {
+                return;
+            }
+        }
+
+        const startTime = Date.now();
+
+        while (nextRetryDelay != null) {
+            this.logger.log(LogLevel.Information, `The next reconnect attempt will start in ${nextRetryDelay} ms.`);
```
JamesNK: IMO it would be useful to include the retry count in the log message

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
mikaelm12: Yeah, connecting with WebSockets and skipping negotiation. The client never gets a negotiation response which is where they would get their connection id from the server.  

> or was it no handshake?

Handshake is always required to connect. 😄

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -54,13 +56,20 @@ export class HttpConnection implements IConnection {
     private readonly logger: ILogger;
     private readonly options: IHttpConnectionOptions;
     private transport?: ITransport;
+    private transferFormat?: TransferFormat;
     private startPromise?: Promise<void>;
     private stopError?: Error;
     private accessTokenFactory?: () => string | Promise<string>;
+    private connectionId: string | null;
```
BrennanConroy:
```suggestion
    private connectionId?: string;
```

src/SignalR/clients/ts/signalr/src/HubConnection.ts
```diff
@@ -348,6 +366,26 @@ export class HubConnection {
         }
     }
 
+    /** Registers a handler that will be invoked when the connection starts reconnecting.
+     *
+     * @param {Function} callback The handler that will be invoked when the connection start reconnecting. Optionally receives a single argument containing the error that caused the connection to start reconnecting (if any).
+     */
+    public onreconnecting(callback: (error?: Error) => void) {
+        if (callback) {
+            this.reconnectingCallbacks.push(callback);
```
BrennanConroy: We might just want this to be a single handler that you can overwrite by calling this function again.

If we allow multiple registrations then people will want to be able to unregister.

src/SignalR/clients/ts/signalr/src/HubConnectionBuilder.ts
```diff
@@ -106,6 +107,19 @@ export class HubConnectionBuilder {
         return this;
     }
 
+     /** Configures the {@link @aspnet/signalr.HubConnection} to automatically attempt to reconnect if the connection is lost. */
+    public withAutomaticReconnect(): HubConnectionBuilder {
```
BrennanConroy: How common do you think people would write the following code? And accidentally overwrite their reconnect policy?
```typescript
builder.withUrl("/chat", { reconnectPolicy: new CustomReconnect() })
    .withAutomaticReconnect()
    .build();
```

src/SignalR/clients/ts/signalr/src/IReconnectPolicy.ts
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+/** An abstraction that controls when the client attempts to reconnect and how many times it does so. */
+export interface IReconnectPolicy {
```
BrennanConroy: Need to add this to the index.ts file.

src/SignalR/clients/ts/signalr/src/IReconnectPolicy.ts
```diff
@@ -0,0 +1,16 @@
...
+    /** Called after the transport loses the connection.
+     *
+     * @param {number} previousRetryCount The number of failed reconnect attempts so far.
+     *
+     * @param {number} elapsedMilliseconds The amount of time in milliseconds spent reconnecting so far.
+     *
+     * @returns {number | null} The amount of time to wait before the next reconnect attempt. `null` tells the client to stop retrying and close.
+     */
+    // TODO: add timeElapsedReconnecting parameter
+    nextRetryDelayInMilliseconds(previousRetryCount: number, elapsedMilliseconds: number): number | null;
```
BrennanConroy: We discussed making this `number | Promise<void>`, do we not want to do that anymore?

src/SignalR/samples/SignalRSamples/wwwroot/hubs.html
```diff
@@ -164,6 +175,14 @@ <h4>Group Actions</h4>
             updateButtonState(false);
         });
 
+        connection.onreconnecting(function (e) {
+            addLine('message-list', 'Connection reconnecting: ' + e, 'red');
```
BrennanConroy: nit: yellow?

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -376,6 +413,83 @@ export class HttpConnection implements IConnection {
...
+        }
+
+        if (!this.changeState(ConnectionState.Connected, ConnectionState.Reconnecting)) {
+            return;
+        }
+
+        if (error) {
+            this.logger.log(LogLevel.Information, `Connection reconnecting because of error '${error}'.`);
+        } else {
+            this.logger.log(LogLevel.Information, "Connection reconnecting.");
```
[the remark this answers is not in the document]
halter73: Why do you like that better than "Connection reconnecting."?

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -376,6 +413,83 @@ export class HttpConnection implements IConnection {
...
+            // This exception is returned to the user as a rejected Promise from the start method
+            // or was observed by the last reconnect loop.
+        }
+
+        const reconnectPolicy = this.options.reconnectPolicy as IReconnectPolicy;
+        let previousRetryCount = 0;
+        let nextRetryDelay = reconnectPolicy.nextRetryDelayInMilliseconds(previousRetryCount++, 0);
+
+        if (nextRetryDelay === null) {
+            this.logger.log(LogLevel.Information, "Connection not reconnecting because of the IReconnectPolicy.");
```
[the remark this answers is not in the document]
halter73: There's a weird situation where it's possible to have reconnectPolicy.nextRetryDelayInMilliseconds() return null the first time it's called. That's what this is handling. I don't want to log "Reconnect attempts have stopped" when the HttpConnection never entered the reconnecting state.

That being said, it is possible that the HubConnection never left the reconnecting state from the previous reconnect attempt, and this will be logged. I'll try changing the log based on the previous retry count.

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
halter73: I had it that way, but then the connectionID is really `string | undefined`. Do we prefer that over `string | null`?

src/SignalR/clients/ts/signalr/src/HubConnection.ts
halter73: I'm following the pattern we used for onclose(). I think if we want to change that pattern, we can change it for all onclose/onreconnecting/onreconnected all at once.

src/SignalR/clients/ts/signalr/src/HubConnectionBuilder.ts
halter73: Not very common, but I can see how it could happen. Do you think we should have withAutomaticReconnect no-op if a reconnectPolicy is already set?

src/SignalR/clients/ts/signalr/src/IReconnectPolicy.ts
halter73: https://github.com/aspnet/AspNetCore/issues/9017

src/SignalR/clients/ts/signalr/src/DefaultReconnectPolicy.ts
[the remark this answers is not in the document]
halter73: #9016

halter73:
@JameNK @mikaelm12 @BrennanConroy @anurse This is ready for review.

I only plan to add two more tests similar to AutomaticReconnect.test.ts's "does not reset after hub handshake failure" test that tests the integration between HubConnection and HttpConnection.

1. A test to verify the onreconnected connectionId parameter appropriately provides the new connection id.
1. A test to verify that the client can reconnect again after it successfully and fully reconnects once.

I'm also considering adding a test to verify the reconnect delay timeout gets cleared when the connection stops, but making the delay long enough to be reliably but quick enough not to slow down our tests might be tricky.

src/SignalR/clients/ts/signalr/tests/AutomaticReconnect.test.ts
```diff
@@ -0,0 +1,449 @@
...
+
+            const startPromise = hubConnection.start();
+
+            await TestWebSocket.webSocketSet;
+            await TestWebSocket.webSocket.closeSet;
+
+            TestWebSocket.webSocket.readyState = TestWebSocket.OPEN;
+            TestWebSocket.webSocket.onopen(new TestEvent());
+
+            await new Promise((resolve) => setTimeout(resolve, 100));
```
halter73: This will change to wait for onmessage to be set similar to how we wait for onclose to be set.

halter73:
@BrennanConroy @mikaelm12 ping

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -123,22 +151,55 @@ export class HttpConnection implements IConnection {
...
+            try {
+                await this.transport.stop();
+            } catch (e) {
+                this.logger.log(LogLevel.Error, `HttpConnection.transport.stop() threw error '${e}'.`);
+                this.stopConnection();
+            }
+
             this.transport = undefined;
+        } else {
+            this.logger.log(LogLevel.Debug, "HttpConnection.transport became unset because start() failed.");
```
mikaelm12: nit: Sounds odd. Maybe just say that it's undefined

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -157,7 +218,7 @@ export class HttpConnection implements IConnection {
                     // No fallback or negotiate in this case.
                     await this.transport!.connect(url, transferFormat);
                 } else {
-                    return Promise.reject(new Error("Negotiation can only be skipped when using the WebSocket transport directly."));
```
mikaelm12: Why did you change these? Shouldn't these continue to just reject the promise instead of throwing?

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -207,12 +270,20 @@ export class HttpConnection implements IConnection {
...
+                this.connectionState = ConnectionState.Connected;
+            }
+
+            // stop() is waiting on us via this.startInternalPromise so keep this.transport around so it can clean up.
+            // This is the only case startInternal can exit in neither the connected nor disconnected state because stopConnection()
+            // will transition to the disconnected state. start() will wait for the transition using the stopPromise.
         } catch (e) {
             this.logger.log(LogLevel.Error, "Failed to start the connection: " + e);
             this.connectionState = ConnectionState.Disconnected;
+
```
mikaelm12: nit: Remove new line

src/SignalR/clients/ts/signalr/src/IConnection.ts
```diff
@@ -6,6 +6,7 @@ import { TransferFormat } from "./ITransport";
 /** @private */
 export interface IConnection {
     readonly features: any;
+    readonly connectionId?: string;
```
mikaelm12: Should this should be `string | null` since we are going to expose this?

src/SignalR/clients/ts/signalr/src/HubConnection.ts
```diff
@@ -101,48 +118,110 @@ export class HubConnection {
...
 
+        try {
+            await this.startInternal();
+
+            this.connectionState = HubConnectionState.Connected;
+            this.logger.log(LogLevel.Debug, "HubConnection connected successfully.");
+        } catch (e) {
+            this.connectionState = HubConnectionState.Disconnected;
+            this.logger.log(LogLevel.Debug, `HubConnection failed to start successfully because of error '${e}'.`);
+            return Promise.reject(e);
```
JamesNK: Because this is in an async method, should this be a throw instead of `Promise.reject(e);`

src/SignalR/clients/ts/signalr/src/HubConnection.ts
```diff
@@ -101,48 +118,110 @@ export class HubConnection {
...
+        } catch (e) {
+            this.logger.log(LogLevel.Debug, `Hub handshake failed with error '${e}' during start(). Stopping HubConnection.`);
+
+            this.cleanupTimeout();
+            this.cleanupPingTimer();
+
+            // HttpConnection.stop() should not complete until after the onclose callback is invoked.
+            // This will transition the HubConnection to the disconnected state before HttpConnection.stop() completes.
+            await this.connection.stop(e);
+            return Promise.reject(e);
```
JamesNK: Because this is in an async method, should this be a throw instead of `Promise.reject(e);`

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -349,19 +420,33 @@ export class HttpConnection implements IConnection {
...
         // If we have a stopError, it takes precedence over the error from the transport
         error = this.stopError || error;
+        this.stopError = undefined;
+
+        if (this.connectionState === ConnectionState.Disconnected) {
+            this.logger.log(LogLevel.Debug, `Call to HttpConnection.stopConnection(${error}) was ignored because the connection is already in the disconnected state.`);
+            return;
+        }
+
+        if (this.connectionState === ConnectionState.Connecting) {
```
mikaelm12: Is this the correct behavior? If `stopConnection `gets called while we're in the connecting state should that override connecting?

src/SignalR/clients/ts/signalr/src/HubConnection.ts
halter73: I was wondering this myself. Our TypeScript client switches back-and-forth between throwing and returning rejected promises from async methods.

The only difference I can ascertain between the two is that throwing will trigger the catch of a surrounding try/catch *in the same method*. Does anyone else know of any other differences?

I switch a few returned rejected promises to throws in HttpConnection.startInternal where I need the catch block to run. Everywhere else I try to return a rejected promise since that seems to be the more common patter in our client.

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
halter73: https://github.com/aspnet/AspNetCore/pull/8566#discussion_r272728004

src/SignalR/clients/ts/signalr/src/IConnection.ts
halter73: On HubConnection it should be.

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
halter73: If `stopConnection` was called, we'd be in the Disconnecting state.

src/SignalR/clients/ts/signalr/src/HubConnection.ts
JamesNK: Something to figure out, but not in this PR 😬

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
halter73: This is what the post-startInternal logic in start() is taking care of. By the time start exits, the state will be Disconnected.

BrennanConroy (commented):
I'll review more tonight/tomorrow

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -349,19 +419,33 @@ export class HttpConnection implements IConnection {
...
+            this.logger.log(LogLevel.Warning, `Call to HttpConnection.stopConnection(${error}) was ignored because the connection hasn't yet left the in the connecting state.`);
```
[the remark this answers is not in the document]
halter73: I don't expect this will ever be logged since we just finished awaiting startInternal() here. startInternal() either:

1. Completes after successfully transitioning from the Connecting to Connected state.
2. Fails after moving back into the Disconnected state.
3. Sees that stop/stopInternal() transitioned the connection to the Disconnecting state and exits while still in the Disconnecting state. This is with the realization that startInternal() needs to complete for stopInternal to continue and the expectation that start() will await the stopInternal() promise and only throw after the connection has fully transition into the Disconnected state.

This check and log might be overly defensive, but the logic is complicated enough I'd like to be able to know this would be logged if and when we investigate any connection state issues.

src/SignalR/clients/ts/signalr/src/HubConnection.ts
```diff
@@ -61,11 +75,11 @@ export class HubConnection {
     // create method that can be used by HubConnectionBuilder. An "internal" constructor would just
     // be stripped away and the '.d.ts' file would have no constructor, which is interpreted as a
     // public parameter-less constructor.
-    public static create(connection: IConnection, logger: ILogger, protocol: IHubProtocol): HubConnection {
-        return new HubConnection(connection, logger, protocol);
+    public static create(connection: IConnection, logger: ILogger, protocol: IHubProtocol, reconnectPolicy?: IReconnectPolicy): HubConnection {
```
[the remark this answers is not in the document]
mikaelm12: Should we just add this as an additional overload instead of the breaking change

src/SignalR/clients/ts/signalr/src/HubConnectionBuilder.ts
```diff
@@ -106,6 +114,39 @@ export class HubConnectionBuilder {
...
 
+    /** Configures the {@link @aspnet/signalr.HubConnection} to automatically attempt to reconnect if the connection is lost.
+     * By default, the client will wait 0, 2, 10 and 30 seconds respectively before trying up to 4 reconnect attempts.
+     */
+    public withAutomaticReconnect(): HubConnectionBuilder;
+
+    /** Configures the {@link @aspnet/signalr.HubConnection} to automatically attempt to reconnect if the connection is lost.
+     *
+     * @param {number[]} retryDelays An array containing the delays in milliseconds before trying each reconnect attempt.
+     * The length of the array represents how many failed reconnect attempts it takes before the client will stop attempting to reconnect.
```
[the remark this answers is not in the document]
halter73: The array cannot contain null according to the type definition. If some tried using an array containing null anyway, the client would stop reconnecting after it reached the first null element of the array.

The DefaultReconnectPolicy ctor copies the array and adds the null at the end for you.

https://github.com/aspnet/AspNetCore/blob/ac2006e8910eec11559ecb4269512dad4703d323/src/SignalR/clients/ts/signalr/src/DefaultReconnectPolicy.ts#L14

src/SignalR/clients/ts/signalr/tests/HttpConnection.test.ts
```diff
@@ -95,8 +81,9 @@ describe("HttpConnection", () => {
 
                 await expect(connection.start(TransferFormat.Text))
                     .rejects
-                    .toThrow("Cannot start a connection that is not in the 'Disconnected' state.");
+                    .toThrow("Cannot start an HttpConnection that is not in the 'Disconnected' state.");
             } finally {
+                (options.transport as ITransport).onclose!();
```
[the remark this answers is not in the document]
halter73: Before stop would complete before the transport closed. Now stop waits for the transport to close, so we indicate that the transport already closed before calling connection.stop() to ensure connection.stop() completes.

As for the as cast, I cant just do `options.transport!.onclose!();` because options.transport has the type `HttpTransportType | ITransport | undefined` instead of `ITransport | undefined`.

I could store the transport in a local, but meh.

src/SignalR/clients/ts/signalr/src/HubConnection.ts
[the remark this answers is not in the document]
halter73: It's private to TypeScript. If called from JavaScript or without type checks without the last argument reconnectPolicy is just undefined I think. Same goes for the HubConnection ctor.

That's part of reason I made reconnectPolicy optional instead of creating a NullReconnectPolicy or similar.

src/SignalR/clients/ts/signalr/tests/HubConnection.Reconnect.test.ts
```diff
@@ -0,0 +1,414 @@
...
+import { DefaultReconnectPolicy } from "../src/DefaultReconnectPolicy";
+import { HubConnection, HubConnectionState } from "../src/HubConnection";
+import { JsonHubProtocol } from "../src/JsonHubProtocol";
+
+import { VerifyLogger } from "./Common";
+import { TestConnection } from "./TestConnection";
+import {  PromiseSource } from "./Utils";
+
+describe("auto reconnect", () => {
+    it("is not enabled by default", async () => {
```
mikaelm12: Would we want to do iterate these over the transports and protocols like we do here https://github.com/aspnet/AspNetCore/blob/d1eeb551395fa3013d7a5862bac8cdae01fff3d6/src/SignalR/clients/ts/FunctionalTests/ts/HubConnectionTests.ts#L49

src/SignalR/clients/ts/signalr/tests/HubConnection.Reconnect.test.ts
```diff
@@ -0,0 +1,414 @@
...
+            connection.onclose!(new Error("Connection lost"));
+
+            await closePromise;
+
+            expect(onreconnectingCalled).toBe(false);
+        },
+        "Connection disconnected with error 'Error: WebSocket closed with status code: 0 ().'.");
+    });
+
+    // it("can be opted into", async () => {
```
mikaelm12: Some testing coverage areas - Reconnect with negotiate redirects, reconnect with skip negotiate, passing in your own `IReconnectPolicy `instead of just the `DefaultReconnectPolicy`

mikaelm12: Also being able to successfully reconnect twice.

mikaelm12 (commented):
I think this looks okay for preview 4. We can refine this in the next preview as well. Let's just wrap up the testing

src/SignalR/clients/ts/signalr/src/HubConnectionBuilder.ts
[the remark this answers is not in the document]
BrennanConroy: Fair, but they can provide their own `IReconnectPolicy` and bypass the constructor https://github.com/aspnet/AspNetCore/pull/8566/files/ac2006e8910eec11559ecb4269512dad4703d323#diff-1e166c54389222a5f7d175c9a95b415bR144

src/SignalR/clients/ts/signalr/src/HubConnection.ts
```diff
@@ -480,27 +578,146 @@ export class HubConnection {
...
+            if (this.connectionState !== HubConnectionState.Reconnecting) {
+                this.logger.log(LogLevel.Debug, "Connection left the reconnecting state in onreconnecting callback. Done reconnecting.");
+                return;
+            }
+        }
+
+        while (nextRetryDelay !== null) {
+            this.logger.log(LogLevel.Information, `Reconnect attempt number ${previousReconnectAttempts} will start in ${nextRetryDelay} ms.`);
+
+            await new Promise((resolve) => {
```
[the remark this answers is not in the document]
halter73: Correct. And we've moved into the disconnected state, so that's fine..

src/SignalR/clients/ts/signalr/src/HubConnectionBuilder.ts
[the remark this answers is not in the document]
halter73: In that case, what index goes goes out of bounds? Are you worried previousRetryCount or elapsedMilliseconds?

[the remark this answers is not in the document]
BrennanConroy: `previousRetryCount` would go out of bounds in their array unless they specify null in the last element. Maybe that's ok, as you put a doc comment for the method, maybe we could be more explicit and say you must return a null as the last element?

src/SignalR/clients/ts/signalr/src/HubConnection.ts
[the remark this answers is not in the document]
BrennanConroy: That would be a memory leak then? Granted a very slow one so maybe not important enough to fix.

BrennanConroy (commented):
Looks like there might be a flaky/always failing test?

src/SignalR/clients/ts/signalr/src/HubConnection.ts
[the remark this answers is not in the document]
halter73: No. The timeout is cleared. Everything is gc'd.

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -349,19 +419,33 @@ export class HttpConnection implements IConnection {
...
-    private changeState(from: ConnectionState, to: ConnectionState): boolean {
-        if (this.connectionState === from) {
-            this.connectionState = to;
-            return true;
-        }
-        return false;
-    }
-
     private stopConnection(error?: Error): void {
+        this.logger.log(LogLevel.Debug, `HttpConnection.stopConnection(${error}) called while in sate ${this.connectionState}.`);
```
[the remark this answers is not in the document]
halter73: These are debug logs. If typescript made it easy to turn regular enums to strings, I would consider it as part of this change.

But as it stands, I have to either change the enum to a "string enum" or add a look up table. Not crazy, but not necessary as part of this change.

src/SignalR/clients/ts/signalr/src/HubConnectionBuilder.ts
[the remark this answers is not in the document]
halter73: I doubt most custom implementations will even use arrays. Talking about elements in a data structure people may or may not use could confuse people.

The doc comments for IReconnectPolicy indicates "`null` tells the client to stop retrying and close." I'll make sure this is clear in the main documentation as well.

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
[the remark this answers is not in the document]
halter73: Okay, I fixed it. https://github.com/aspnet/AspNetCore/pull/8566/commits/2112031bef4847b6ff046e7d3a7015b2fd0bbe1b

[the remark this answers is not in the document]
BrennanConroy: Third times a charm
"sate" -> "state"

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -14,9 +14,10 @@ import { WebSocketTransport } from "./WebSocketTransport";
 
 /** @private */
 const enum ConnectionState {
-    Connecting,
-    Connected,
-    Disconnected,
+    Connecting = "Connecting ",
```
BrennanConroy: Nice :+1:

BrennanConroy (approved):
Should the onclosed callback be triggered if the connection stops while in the reconnecting state?
We should file an issue for adding some random delay.
We should file an issue for figuring out how to make reconnect work nicely with handshake failure.
Some of these logs might be better suited as trace.

PR #8568, opened 2019-03-15
Kills java instances for Azure pipelines

javiercn:


eng/scripts/KillProcesses.ps1
```diff
@@ -10,6 +10,12 @@ function _kill($processName) {
     }
 }
 
+function _killSeleniumJavaInstances(){
```
ryanbrandenburg: nit: This name isn't quite accurate, it's killing our java instances, which might be selenium. Don't block on it though.

ryanbrandenburg (approved):

eng/scripts/KillProcesses.ps1
```diff
@@ -22,6 +28,8 @@ _kill vctip.exe
 _kill chrome.exe
 _kill h2spec.exe
 _kill WerFault.exe
+_killSeleniumJavaInstances
```
dougbu: Why isn't this `_kill java.exe`? or a few `_kill` calls for the Java processes we use?

dougbu (changes requested):
Update KillProcesses.sh too

eng/scripts/KillProcesses.ps1
dougbu: To be clear: I won't block on this but it looks more complicated that necessary and will fail the build if a `Stop-Process` fails (unlike in the `_kill` function).

javiercn: I already fixed that locally. I'm trying it out on a different PR. I will update this one and merge it once  I see it working on the other one.

eng/scripts/KillProcesses.ps1
```diff
@@ -10,6 +10,18 @@ function _kill($processName) {
...
 
+function _killJavaInstances() {
+    $_javaProcesses = Get-Process java -ErrorAction SilentlyContinue |
+        Where-Object { $_.Path -like "$env:JAVA_HOME*" };
+    foreach($_javaProcess in $_javaProcesses) {
+        try {
+            Stop-Process $proc
+        } catch {
+            Write-Host "Failed to kill java process: $proc 
+        }
```
dougbu: This looks more robust.

dougbu (approved):
I'm still curious: Why is `_kill java.exe` or a couple of similar commands insufficient? And, why isn't it useful to kill Java processes on non-Windows machines i.e. in KillProcesses.sh (that only kills `dotnet`)?

eng/scripts/KillProcesses.ps1
```diff
@@ -10,6 +10,18 @@ function _kill($processName) {
...
 }
 
+function _killJavaInstances() {
+    $_javaProcesses = Get-Process java -ErrorAction SilentlyContinue |
+        Where-Object { $_.Path -like "$env:JAVA_HOME*" };
+    foreach($_javaProcess in $_javaProcesses) {
+        try {
+            Stop-Process $proc
+        } catch {
+            Write-Host "Failed to kill java process: $proc
```
Eilon: Is there a missing end quote here?

javiercn: Good catch

Eilon:
Force merged.

Eilon:
Thanx @javiercn !

PR #8569, opened 2019-03-15
Update acessTokenProvider selecting logic for restarting connection

mikaelm12:
Fixes: https://github.com/aspnet/AspNetCore/issues/8429

Making sure we don't lose the reference to the user passed in accessTokenProvider. Before if you got redirected by the service, then restarted your connection the accessTokenProvider regiestered with `withAccessTokenProvider` would be lost.

### DIFF
--- a/src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
+++ b/src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
@@ -38,6 +38,7 @@ public class HubConnection {
     private List<OnClosedCallback> onClosedCallbackList;
     private final boolean skipNegotiate;
     private Single<String> accessTokenProvider;
+    private Single<String> redirectAccessTokenProvider;
     private final Map<String, String> headers = new HashMap<>();
     private ConnectionState connectionState = null;
     private HttpClient httpClient;
@@ -255,10 +256,10 @@ public class HubConnection {
             }
 
             if (negotiateResponse.getAccessToken() != null) {
-                this.accessTokenProvider = Single.just(negotiateResponse.getAccessToken());
+                this.redirectAccessTokenProvider = Single.just(negotiateResponse.getAccessToken());
                 // We know the Single is non blocking in this case
                 // It's fine to call blockingGet() on it.
-                String token = this.accessTokenProvider.blockingGet();
+                String token = this.redirectAccessTokenProvider.blockingGet();
                 this.headers.put("Authorization", "Bearer " + token);
             }
 
@@ -308,9 +309,10 @@ public class HubConnection {
         negotiate.flatMapCompletable(url -> {
             logger.debug("Starting HubConnection.");
             if (transport == null) {
+                Single<String> tokenProvider = redirectAccessTokenProvider != null ? redirectAccessTokenProvider : accessTokenProvider;
                 switch (transportEnum) {
                     case LONG_POLLING:
-                        transport = new LongPollingTransport(headers, httpClient, accessTokenProvider);
+                        transport = new LongPollingTransport(headers, httpClient, tokenProvider);
                         break;
                     default:
                         transport = new WebSocketTransport(headers, httpClient);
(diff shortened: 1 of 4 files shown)


mikaelm12:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=125367&view=logs

aspnet-hello:

I've triaged the above build.

src/SignalR/clients/java/signalr/src/test/java/com/microsoft/signalr/HubConnectionTest.java
```diff
@@ -1651,6 +1651,48 @@ public void negotiateRedirectIsFollowed()  {
...
+        HubConnection hubConnection = HubConnectionBuilder
+                .create("http://example.com")
+                .withTransportImplementation(transport)
+                .withHttpClient(client)
+                .withAccessTokenProvider(Single.just("User Registered Token"))
+                .build();
+
+        hubConnection.start().timeout(1, TimeUnit.SECONDS).blockingAwait();
+        assertEquals(HubConnectionState.CONNECTED, hubConnection.getConnectionState());
+        hubConnection.stop();
```
BrennanConroy: await

BrennanConroy (approved):

src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
```diff
@@ -308,9 +309,10 @@ public Completable start() {
         negotiate.flatMapCompletable(url -> {
             logger.debug("Starting HubConnection.");
             if (transport == null) {
+                Single<String> tokenProvider = redirectAccessTokenProvider != null ? redirectAccessTokenProvider : accessTokenProvider;
```
halter73: I noticed that we weren't nulling out the redirect url in stopConnection(), and saw we just flow the redirect url from the return value of startNegotiate() all the way to the call to transport.start(url).

I like this pattern over adding fields that need to be reset. I recommend changing startNegotiate() to return the NegotiateResponse returned from the last call to handleNegotiate().

startNegotiate() could continue to contain the transportEnum selection logic. Since getTransportEnum() exists, we need to store the current transportEnum in a field anyway. We *should* reset transportEnum to its original value in stopConnection().

The finalUrl logic currently in startNegotiate() should move here since this flatMapCompletable callback will have the entire last NegotiateResponse passed in, not just the finalUrl.

Then you can also move the access token logic that's currently in handleNegotiate() right here and use a local tokenProvider instead of the field. Speaking of that logic, I think we should remove the Authorization in stopConnection() if we added one.

src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
```diff
@@ -255,10 +256,10 @@ private void timeoutHandshakeResponse(long timeout, TimeUnit unit) {
             }
 
             if (negotiateResponse.getAccessToken() != null) {
-                this.accessTokenProvider = Single.just(negotiateResponse.getAccessToken());
+                this.redirectAccessTokenProvider = Single.just(negotiateResponse.getAccessToken());
                 // We know the Single is non blocking in this case
                 // It's fine to call blockingGet() on it.
-                String token = this.accessTokenProvider.blockingGet();
+                String token = this.redirectAccessTokenProvider.blockingGet();
                 this.headers.put("Authorization", "Bearer " + token);
```
halter73: Nit: why does this method use `this` everywhere?

src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
mikaelm12: I don't disagree with these changes but are these refactorings something we want to include in the patch since this is something we want to backport? Just a thought.

mikaelm12: I'll just leave the refactor work our of the patch but I'll include it in master.

mikaelm12: >Then you can also move the access token logic that's currently in handleNegotiate() right here and use a local tokenProvider instead of the field.

I think we need to keep  a field for the accessToken because in the redirect scenario we'll get two negotiate responses. The accessToken won't be in the final negotiate response though.

halter73 (approved):

PR #8577, opened 2019-03-15
Add new email/confirmation flows

HaoK:
Part of https://github.com/aspnet/AspNetCore/issues/8356

Still working on fixing/adding tests, for the default confirmation page, we generate a new confirmation link and display it only when in development.

@ajcvickers @blowdart


HaoK:
For this new requireConfirmedAccount setting, are we planning on setting it explicitly in the templates or just turn it on in AddDefaultIdentity?

HaoK:
Tests added/updated and are green again, removing WIP label

src/Identity/UI/src/Areas/Identity/Pages/V3/Account/RegisterConfirmation.cshtml.cs
```diff
@@ -0,0 +1,90 @@
...
+            }
+
+            var user = await _userManager.FindByEmailAsync(email);
+            if (user == null)
+            {
+                return NotFound($"Unable to load user with email '{email}'.");
+            }
+
+            Email = email;
+            IsDevelopment = _hostingEnv.IsDevelopment();
```
javiercn: I don't think we do this in any part of ASP.NET (other than the the old metapackage) and we try to avoid this. (I tried to do this in the past and got smacked for it). The pattern that we follow is to flow this through config or other means. 

If I can provide an alternative suggestion, you can remove the property completely. You can have an if on the view like as follows
```csharp
if(ViewData["EnableConfirmationLink"])
{
    // Display the link
}
```
and you can drop a _ViewStart.cshtml (on the templates) under the Account folder to enable it as follows
```razor
@inject IHostingEnvironment HostingEnvironment;
ViewData["EnableConfirmationLink"] = HostingEnvironment.IsDevelopment
```
This way:
* The user has direct control of when this is enabled.
* We don't depend on hosting environment within the core libraries.

src/Identity/UI/src/Areas/Identity/Pages/V3/Account/RegisterConfirmation.cshtml.cs
```diff
@@ -0,0 +1,90 @@
...
+            if (IsDevelopment)
+            {
+                var userId = await _userManager.GetUserIdAsync(user);
+                var code = await _userManager.GenerateEmailConfirmationTokenAsync(user);
+                EmailConfirmationUrl = Url.Page(
+                    "/Account/ConfirmEmail",
+                    pageHandler: null,
+                    values: new { userId = userId, code = code },
+                    protocol: Request.Scheme);
+
```
javiercn: Extra line, but as mentioned above its better if you follow the `_ViewStart.cshtml` approach

src/Identity/UI/src/Areas/Identity/Pages/V4/Account/ConfirmEmailChange.cshtml.cs
```diff
@@ -0,0 +1,70 @@
...
+            var user = await _userManager.FindByIdAsync(userId);
+            if (user == null)
+            {
+                return NotFound($"Unable to load user with ID '{userId}'.");
+            }
+
+            var result = await _userManager.ChangeEmailAsync(user, email, code);
+            if (!result.Succeeded)
+            {
+                throw new InvalidOperationException($"Error changing email for user with ID '{userId}':");
```
javiercn: Same comments as for the V3 version

src/Identity/UI/src/Areas/Identity/Pages/V4/Account/Manage/Email.cshtml.cs
```diff
@@ -0,0 +1,186 @@
...
+            {
+                return Page();
+            }
+
+            var user = await _userManager.GetUserAsync(User);
+            if (user == null)
+            {
+                return NotFound($"Unable to load user with ID '{_userManager.GetUserId(User)}'.");
+            }
+
```
javiercn: extra line

src/Identity/samples/IdentitySample.DefaultUI/Startup.cs
```diff
@@ -37,9 +46,11 @@ public void ConfigureServices(IServiceCollection services)
 
             services.AddMvc().AddNewtonsoftJson();
 
-            services.AddDefaultIdentity<ApplicationUser>()
+            services.AddDefaultIdentity<ApplicationUser>(o => o.SignIn.RequireConfirmedAccount = true)
                  .AddRoles<IdentityRole>()
                  .AddEntityFrameworkStores<ApplicationDbContext>();
+
+            //services.AddSingleton<IUserConfirmation<ApplicationUser>, BadDude>();
```
javiercn: 😈 Leftovers?

src/Identity/test/Identity.FunctionalTests/Pages/Account/Manage/Email.cs
```diff
@@ -0,0 +1,68 @@
...
+            var manageResponse = await Client.GetAsync(goToManage);
+            var manage = await ResponseAssert.IsHtmlDocumentAsync(manageResponse);
+
+            return new Email(Client, manage, Context);
+        }
+
+        internal object GetEmail() => _emailInput.GetAttribute("value");
+
+        internal object GetNewEmail() => _newEmailInput.GetAttribute("value");
+
```
javiercn: extra line

javiercn (commented):
Overall looks good.

The only real issue I have is with the way we do the confirmation link for development.

src/Identity/UI/src/Areas/Identity/Pages/V3/Account/ConfirmEmailChange.cshtml.cs
```diff
@@ -0,0 +1,70 @@
...
+            {
+                throw new InvalidOperationException($"Error changing email for user with ID '{userId}':");
+            }
+
+            // In our UI email and user name are one and the same, so when we update the email
+            // we need to update the user name.
+            var setUserNameResult = await _userManager.SetUserNameAsync(user, email);
+            if (!setUserNameResult.Succeeded)
+            {
+                throw new InvalidOperationException($"Unexpected error occurred setting name for user with ID '{userId}'.");
```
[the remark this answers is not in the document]
HaoK: I'm fine changing this, but just wanted to point out that this is what the existing code does today https://github.com/aspnet/AspNetCore/blob/master/src/Identity/UI/src/Areas/Identity/Pages/V4/Account/Manage/Index.cshtml.cs#L149

src/Identity/UI/src/Areas/Identity/Pages/V3/Account/RegisterConfirmation.cshtml.cs
HaoK: I can just switch to checking for the no-op default email sender instead to control whether to display the code

javiercn: Hmm, you don't want that to happen in prod, do you?

javiercn: It will poison your data and it will be a mess.

HaoK: We'll go over this with @blowdart at triage this week and hash out this bit then

javiercn: The new way looks definitely better, but I would consider putting something on the template viewstart to make sure this never shows up in prod.

javiercn (approved):
Looks good to me.

ajcvickers (approved):

MarcAir:
There may be a logical error in the confirmation of email.

OnPostChangeEmailAsync does not check to see if the email is used by another user prior to sending confirmation.
or should we consider this something the ConfirmEmail page is to do, since someone could potentially have either errored while entering the new email, or someone trying to exploit the change email function to see if someone is registered with that email?
either way, should we not be checking for email existing in the identity store prior to emailing a request for confirmation?

MarcAir:
something like this? 
basically if email exists, do not send email. But notify user of email sent (for potential nefarious actors) 
```csharp
public async Task<IActionResult> OnPostChangeEmailAsync()
        {
///removed for brevity 
if (Input.NewEmail != email)
{
            var emailExists = await _userManager.FindByEmailAsync(Input.NewEmail);
            if(emailExists == null)//not notifying if email sent or not.
            {
                        var userId = await _userManager.GetUserIdAsync(user);
                        var code = await _userManager.GenerateChangeEmailTokenAsync(user, Input.NewEmail);
                        var callbackUrl = Url.Page(
                                    "/Account/ConfirmEmailChange",
                                    pageHandler: null,
                                    values: new { userId = userId, email = Input.NewEmail, code = code },
                                    protocol: Request.Scheme);

                        await _emailSender.SendEmailAsync(
                                    Input.NewEmail,
                                    "Confirm your email",
                                    $"Please confirm your account by <a href=\"{HtmlEncoder.Default.Encode(callbackUrl)}\">clicking here</a>.");
            }
            StatusMessage = "Confirmation link to change email sent. Please check your email";
            return RedirectToPage();
}
///removed for brevity 
}
```

PR #8578, opened 2019-03-15
Publish MusicStore as part of build

pranavkm:
* Add a target to publish MusicStore as part of E2E tests
* Publishing for x86 causes unit tests in MusicStore to have weird
reference issues. Removing these since they are very pretty low value.


src/MusicStore/test/MusicStore.E2ETests/PublishAndRunTests.cs
```diff
@@ -15,7 +18,7 @@ namespace E2ETests
     public class PublishAndRunTests : LoggedTest
     {
         public static TestMatrix TestVariants
-            => TestMatrix.ForServers(ServerType.IISExpress, ServerType.HttpSys)
+            => TestMatrix.ForServers(ServerType.IISExpress, ServerType.HttpSys, ServerType.Kestrel)
```
jkotalik: As you mentioned, I'm never was a fan of the test matrix as it makes it hard to disable things. Maybe we should just scrap it. Not in this PR but eventually.

src/MusicStore/test/MusicStore.E2ETests/PublishAssetAttribute.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        public ApplicationType ApplicationType { get; }
+
+        public ServerType ServerType { get; }
+
+        public HostingModel HostingModel { get; }
+
+        public string PublishDirectory { get; }
+    }
+
```
jkotalik: nit space.

src/MusicStore/test/MusicStore.Test/CartSummaryComponentTest.cs
```diff
@@ -1,90 +0,0 @@
...
-            services
-                .AddMemoryCache()
-                .AddLogging()
-                .AddDbContext<MusicStoreContext>(b => b.UseInMemoryDatabase("Scratch").UseInternalServiceProvider(efServiceProvider));
-
-            _serviceProvider = services.BuildServiceProvider();
-        }
-
-        [Fact]
-        public async Task CartSummaryComponent_Returns_CartedItems()
```
jkotalik: Just so I know, why were these test dropped?

jkotalik (approved):

src/MusicStore/test/MusicStore.Test/CartSummaryComponentTest.cs
pranavkm: I didn't spend an inordinate time on it, but it looked like incremental builds for the solution were broken because the unit tests would get x86 versions of the binaries and would then fail in exotic ways. We could spend time fixing this, but the tests really offer no value - they're unit testing a sample that we use for product testing.

src/MusicStore/test/MusicStore.E2ETests/MusicStore.E2ETests.csproj
```diff
@@ -39,4 +36,40 @@
...
+    </ItemGroup>
+
+    <MSBuild
+      Projects="$(MSBuildProjectFullPath)"
+      Targets="_PublishAsset"
+      Properties="PublishProject=%(PublishAsset.Identity);Architecture=%(Architecture);ApplicationType=%(ApplicationType);PublishDirectory=%(PublishDirectory)"
+      BuildInParallel="false" />
+  </Target>
+
+  <Target Name="_PublishAsset">
```
pakrym: Did you try reusing targets from testsite.props?

pranavkm: NIH bro. But more seriously, the IIS targets had a bunch of things that look like they are very IIS specific (copying launch settings \ injecting deps file etc). All we need here is to publish a project with different settings. I did re-use the `PublishedApplicationPublisher` which majorly speeds up things.

src/MusicStore/test/MusicStore.E2ETests/PublishAndRunTests.cs
pranavkm: 🤷‍♀️

src/MusicStore/test/MusicStore.E2ETests/MusicStore.E2ETests.csproj
pakrym: I think it be easily extracted, you mostly need https://github.com/aspnet/AspNetCore/blob/master/src/Servers/IIS/build/testsite.props#L75-L113. There are couple workarounds for VS/publish issues there that are universaly useful.

PR #8582, opened 2019-03-16
Attempt to make MaxRequestBufferSizeTests.LargeUpload less flaky

halter73:
This is for the release/2.1 branch, but should probably be ported forward to master so the test can be unskipped.

Fixes https://github.com/aspnet/AspNetCore-Internal/issues/1775 (hopefully)
See https://github.com/aspnet/AspNetCore-Internal/issues/1970 for the master version.

### DIFF
--- a/src/Servers/Kestrel/test/FunctionalTests/MaxRequestBufferSizeTests.cs
+++ b/src/Servers/Kestrel/test/FunctionalTests/MaxRequestBufferSizeTests.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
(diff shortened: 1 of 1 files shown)


dougbu:
@halter73 is this needed in the current patch builds? I believe 2.1 builds have passed AspNetCore for the next release i.e. the builds succeeded (though a few tests may yet fail in the Test: Windows (Server 2012) build). Don't really want to back them up.

Might be fine for 2.2 either way because those AspNetCore builds are blocked by other PRs and their validations.

halter73:
Skipping is fine for 2.1. I would like to get _dataLength to 100MB in all branches we can though.

jkotalik (approved):

anurse:
@Eilon this is a test-only fix, what's the shiproom process here. Do we need to bring it there?

Eilon:
It's "Tell Mode" (send email only) for test-only changes. If we already have the final 2.1 build for this patch, let's get it in now. If not, maybe we wait until that's done so we don't bork anything by mistake.

anurse:
For shiproom: This is a test-only fix to improve build stability. Please let us know when we're good to merge this into the 2.1 branch.

dougbu (approved):

Eilon:
Don't need any shiproom labels, just merge when ready.

PR #8607, opened 2019-03-17
Add skeleton to use components + prerendering

rynowak:
This adds the skeleton needed for components + prerendring development
in the MVC sandbox. We don't currently have a sample app for quick and
dirty testing.

### DIFF
--- a/src/Mvc/samples/MvcSandbox/Startup.cs
+++ b/src/Mvc/samples/MvcSandbox/Startup.cs
@@ -26,6 +26,7 @@ namespace MvcSandbox
             {
                 options.ConstraintMap["slugify"] = typeof(SlugifyParameterTransformer);
             });
+            services.AddRazorComponents();
             services.AddMvc()
                 .AddRazorRuntimeCompilation()
                 .SetCompatibilityVersion(Microsoft.AspNetCore.Mvc.CompatibilityVersion.Latest);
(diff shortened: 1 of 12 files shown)


rynowak:
@aspnet/brt

rynowak:
@javiercn

src/Mvc/samples/MvcSandbox/Startup.cs
```diff
@@ -66,13 +70,10 @@ public void Configure(IApplicationBuilder app)
 
                 builder.MapControllers();
                 builder.MapRazorPages();
-
-                builder.MapFallbackToController("Index", "Home");
+                builder.MapComponentHub<MvcSandbox.Components.App>("app");
```
javiercn: I think you need to pass the path here too.

rynowak: Path to what? This is working

javiercn: Actually, nevermind. This should work just fine.

javiercn (approved):

rynowak:
@aspnet/brt

rynowak:
still fails after rebasing

javiercn:
The agent is busted. Needs to be reboot.

anurse:
Yeah, this is the file locking that's plaguing all our agents.

Issue #8609, opened 2019-03-17
Exceptions thrown during prerendering should be handled by developer exception page


rynowak:
Exceptions in prerendering cause a "white screen". All of the HTML skeleton is transferred to the client, but none of the component output.

The root cause is that `RemoteRenderer.HandleException` just handles and logs all exceptions without firing its `UnhandledException` event. Also nothing subscribes to the `UnhandledException` event.

```
Microsoft.AspNetCore.Components.Server.dll!Microsoft.AspNetCore.Components.Browser.Rendering.RemoteRenderer.HandleException(System.Exception exception) Line 86	C#
 	Microsoft.AspNetCore.Components.dll!Microsoft.AspNetCore.Components.Rendering.Renderer.AddToPendingTasks(System.Threading.Tasks.Task task) Line 329	C#
 	Microsoft.AspNetCore.Components.dll!Microsoft.AspNetCore.Components.Rendering.ComponentState.SetDirectParameters(Microsoft.AspNetCore.Components.ParameterCollection parameters) Line 132	C#
 	Microsoft.AspNetCore.Components.dll!Microsoft.AspNetCore.Components.Rendering.Renderer.RenderRootComponentAsync(int componentId, Microsoft.AspNetCore.Components.ParameterCollection initialParameters) Line 149	C#
 	Microsoft.AspNetCore.Components.dll!Microsoft.AspNetCore.Components.Rendering.HtmlRenderer.CreateInitialRenderAsync(System.Type componentType, Microsoft.AspNetCore.Components.ParameterCollection initialParameters) Line 235	C#
 	Microsoft.AspNetCore.Components.dll!Microsoft.AspNetCore.Components.Rendering.HtmlRenderer.RenderComponentAsync(System.Type componentType, Microsoft.AspNetCore.Components.ParameterCollection initialParameters) Line 52	C#
 	Microsoft.AspNetCore.Components.Server.dll!Microsoft.AspNetCore.Components.Server.Circuits.CircuitHost.PrerenderComponentAsync.AnonymousMethod__0() Line 102	C#
 	Microsoft.AspNetCore.Components.dll!Microsoft.AspNetCore.Components.Rendering.RendererSynchronizationContext.InvokeAsync.AnonymousMethod__0(object _) Line 116	C#
[...]

PR #8616, opened 2019-03-18
Throw unhandled exceptions during prerendering

rynowak:
Fixes: #8609

Currently exceptions thrown during prerendering are simply logged. This
change uses the existing *unhandled exception* mechanism of the
renderer/circuit to throw these. The result is that the developer
exception page just works for prerendering.

### DIFF
--- a/src/Components/Components/src/EventCallbackFactoryBinderExtensions.cs
+++ b/src/Components/Components/src/EventCallbackFactoryBinderExtensions.cs
@@ -11,74 +11,263 @@ namespace Microsoft.AspNetCore.Components
     /// </summary>
     public static class EventCallbackFactoryBinderExtensions
     {
+        private delegate bool BindConverter<T>(object obj, out T value);
+
         // Perf: conversion delegates are written as static funcs so we can prevent
         // allocations for these simple cases.
-        private static Func<object, string> ConvertToString = (obj) => (string)obj;
+        private readonly static BindConverter<string> ConvertToString = ConvertToStringCore;
 
-        private static Func<object, bool> ConvertToBool = (obj) => (bool)obj;
-        private static Func<object, bool?> ConvertToNullableBool = (obj) => (bool?)obj;
-
-        private static Func<object, int> ConvertToInt = (obj) => int.Parse((string)obj);
-        private static Func<object, int?> ConvertToNullableInt = (obj) =>
+        private static bool ConvertToStringCore(object obj, out string value)
         {
-            if (int.TryParse((string)obj, out var value))
+            // We expect the input to already be a string.
+            value = (string)obj;
+            return true;
+        }
+
+        private static BindConverter<bool> ConvertToBool = ConvertToBoolCore;
+        private static BindConverter<bool?> ConvertToNullableBool = ConvertToNullableBoolCore;
+
+        private static bool ConvertToBoolCore(object obj, out bool value)
+        {
+            // We expect the input to already be a bool.
+            value = (bool)obj;
+            return true;
+        }
+
+        private static bool ConvertToNullableBoolCore(object obj, out bool? value)
+        {
+            // We expect the input to already be a bool.
+            value = (bool?)obj;
+            return true;
+        }
+
+        private static BindConverter<int> ConvertToInt = ConvertToIntCore;
+        private static BindConverter<int?> ConvertToNullableInt = ConvertToNullableIntCore;
+
+        private static bool ConvertToIntCore(object obj, out int value)
+        {
+            var text = (string)obj;
+            if (string.IsNullOrEmpty(text))
             {
-                return value;
+                value = default;
+                return false;
             }
 
-            return null;
-        };
-
-        private static Func<object, long> ConvertToLong = (obj) => long.Parse((string)obj);
-        private static Func<object, long?> ConvertToNullableLong = (obj) =>
-        {
-            if (long.TryParse((string)obj, out var value))
+            if (!int.TryParse(text, out var converted))
             {
-                return value;
+                value = default;
+                return false;
             }
 
-            return null;
-        };
+            value = converted;
+            return true;
+        }
 
-        private static Func<object, float> ConvertToFloat = (obj) => float.Parse((string)obj);
-        private static Func<object, float?> ConvertToNullableFloat = (obj) =>
+        private static bool ConvertToNullableIntCore(object obj, out int? value)
         {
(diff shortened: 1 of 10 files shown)


src/Mvc/test/Mvc.FunctionalTests/ComponentRenderingFunctionalTests.cs
```diff
@@ -17,18 +19,17 @@ public class ComponentRenderingFunctionalTests : IClassFixture<MvcTestFixture<Ba
         public ComponentRenderingFunctionalTests(MvcTestFixture<BasicWebSite.StartupWithoutEndpointRouting> fixture)
         {
             Factory = fixture;
-            Client = Client ?? CreateClient(fixture);
```
rynowak: The paradigm used in these tests was broken now that we report errors. The FetchData component is part of all of these tests, and when used with prerendering it was throwing an exception because the service configuration wasn't correct.

I can't help feeling like this is all more complicated than necessary. Using httpclient to call your own API isn't something we'd normally tell users to do.

src/Mvc/test/WebSites/BasicWebSite/RazorComponents/Throws.razor
```diff
@@ -0,0 +1,10 @@
+﻿@page "/components/throws"
+    
+@* This is expected to throw and result in a 500 *@
+@functions { 
+    protected override async Task OnInitAsync()
+    {
+        await base.OnInitAsync();
+        throw new InvalidTimeZoneException("test");
```
rynowak: One of my favourites for testing, I've never seen this in the wild.

src/Mvc/test/WebSites/BasicWebSite/RazorComponents/Throws.razor
```diff
@@ -0,0 +1,10 @@
+﻿@page "/components/throws"
+
```
rynowak: Note that I didn't include a test for the case where a component throws during construction/DI/parameters. That results in https://github.com/aspnet/AspNetCore/issues/8190

javiercn:
I already have tests and a fix for this on the components relayering PR, btw

javiercn (approved):

src/Components/Server/src/Circuits/CircuitPrerenderer.cs
```diff
@@ -40,6 +45,13 @@ public async Task<IEnumerable<string>> PrerenderComponentAsync(ComponentPrerende
             }
         }
 
+        private void CircuitHost_UnhandledException(object sender, UnhandledExceptionEventArgs e)
+        {
+            // Throw all exceptions encountered during pre-rendering so the default developer
+            // error page can respond.
+            throw (Exception)e.ExceptionObject;
```
javiercn: Shouldn't this be ExceptionDispatchInfo.Capture(e).Throw() to preserve the original callstack?

rynowak: 👍

SteveSandersonMS (approved):
Looks solid

rynowak:
@SteveSandersonMS - could you take a second look at this? 

I have more that I want to follow up on but I will do that in a separate PR.

src/Components/Components/src/EventCallbackFactoryBinderExtensions.cs
```diff
@@ -11,74 +11,263 @@ namespace Microsoft.AspNetCore.Components
     /// </summary>
     public static class EventCallbackFactoryBinderExtensions
     {
+        private delegate bool BindConverter<T>(object obj, out T value);
```
rynowak: I reworked all of this logic to avoid throwing exceptions where possible. Now that we're handling the exceptions here it just makes sense to avoid the noise of throwing them in the first place. Now all of this code is tryparse where possible.

src/Components/Components/src/EventCallbackFactoryBinderExtensions.cs
```diff
@@ -11,74 +11,263 @@ namespace Microsoft.AspNetCore.Components
...
+            if (!int.TryParse(text, out var converted))
+            {
+                value = default;
+                return false;
             }
 
-            return null;
-        };
+            value = converted;
+            return true;
```
rynowak: The difference between nullable and non-nullable:
- Nullable treats empty string as `null` and will bind successfully
- non-nullable fails to parse empty string, and will retain the previous value.

It's problematic to convert an empty value to `default` for non-nullable. It will check the value of the text field to `0` instead of `""`.

SteveSandersonMS (approved):
Looks extra-solid now! I'm glad we're finally filling in some of the details of these scenarios.

src/Mvc/test/Mvc.FunctionalTests/ComponentRenderingFunctionalTests.cs
javiercn: you can simply do `Client ?? CreateClient(fixture.WithWebHostBuilder(b => b.ConfigureServices(s => s.AddRazorComponents()))` or extend the MVC fixture and configure it there. Both work. I usually do the WithWebHostBuilder because I'm cheap and I don't bother with inheritance, but this is something you can do a nested class of the test and that's totally fine.

src/Components/Server/src/Circuits/CircuitPrerenderer.cs
```diff
@@ -26,8 +28,12 @@ public async Task<IEnumerable<string>> PrerenderComponentAsync(ComponentPrerende
                 GetFullUri(context.Request),
                 GetFullBaseUri(context.Request));
 
+            // We don't need to unsubscribe because the circuit host object is scoped to this call.
+            circuitHost.UnhandledException += CircuitHost_UnhandledException;
```
javiercn: On second look, I don't think this is correct. The circuit will live longer than the request and when the signalr connection gets stablished you don't want to have this handler throwing.

If you take a look at how i'm doing this on my PR
https://github.com/aspnet/AspNetCore/pull/7954/files#diff-d33c04a2c673816a9a5c6e959a8e5270R91

I'm specifically basing it on whether I'm prerrendering or not.

You can keep this if you want to, but I'm gonna have to change it on my PR so it just adds work to me :(

I'll leave it up to yo.

rynowak: I don't understand this, or how to do extra work to support something that isn't implemented yet, so I'm going to go ahead and merge this 😆

PR #8624, opened 2019-03-18
Start IHostedServices before server

KravaNick:
Fix: #8625

Why when we start server, IHostedService's start after it?
It can cause some runtime problems, when we just started and already receiveing and handling requests.

Isn't it would be a great idea to also with for services to start properly, before start receiveing requests


Tratcher:
In hindsight this was a poor design on our part, but we opted not to change it in WebHost for compatibility reasons. However this has been addressed in the new 3.0 generic Host where the server is an IHostedService and you control this via registration order.

PR #8634, opened 2019-03-18
Fixes #7862

### LINKED ISSUE #7862: Our contributing guidelines don't work when JS is involved
### Describe the bug

Components.sln requires you to have already built the javascript project at src\Components\Browser.JS

It looks like these `npmproj` instance don't participate in the build. The build will fail looking for `components.server.js`

```
Severity	Code	Description	Project	File	Line	Suppression State
Error	CS1566	Error reading resource '_framework\components.server.js' -- 'Could not find a part of the path 'C:\git\aspnet\AspNetCore\src\Components\Browser.JS\src\dist\components.server.js'.'	Microsoft.AspNetCore.Components.Server	C:\git\aspnet\AspNetCore\src\Components\Server\src\CSC	1	Active

```

### To Reproduce

`git clean -xdf` in `src\Components\Browser.JS\src`
Follow the instructions here: https://github.com/aspnet/AspNetCore/blob/master/docs/BuildFromSource.md#building-in-visual-studio
`startvs.cmd src\Components\Components.sln`

Do a build - it will fail with an error message 

### Expected behavior

The build works

## Workaround

The workaround is go to the `Browser.JS` directory and `dotnet build`.

ryanbrandenburg:
Fixes #7862.

### DIFF
--- a/restore.cmd
+++ b/restore.cmd
@@ -1,2 +1,2 @@
 @ECHO OFF
-PowerShell -NoProfile -NoLogo -ExecutionPolicy unrestricted -Command "[System.Threading.Thread]::CurrentThread.CurrentCulture = ''; [System.Threading.Thread]::CurrentThread.CurrentUICulture = '';& '%~dp0build.ps1' -all -nobuild -restore %*; exit $LASTEXITCODE"
+PowerShell -NoProfile -NoLogo -ExecutionPolicy unrestricted -Command "[System.Threading.Thread]::CurrentThread.CurrentCulture = ''; [System.Threading.Thread]::CurrentThread.CurrentUICulture = '';& '%~dp0build.ps1' -all -restore %*; exit $LASTEXITCODE"
--- a/restore.sh
+++ b/restore.sh
@@ -3,4 +3,4 @@
 set -euo pipefail
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-"$DIR/build.sh" --all --restore --no-build "$@"
+"$DIR/build.sh" --all --restore "$@"


natemcmaster (changes requested):
I'm not sure about this solution. This means restore.cmd/sh are basically the same as build.cmd/sh.  The whole point of the restore script is to simply download dependencies, and not to compile projects.

ryanbrandenburg:
The other option I see is to change the guidance to do `.\build.cmd -all` instead of restore. We just have too many projects/scenarios that don't work without a full build first, at which point I feel like there's not much point to the `.\restore` script.

natemcmaster:
I think there are more than two options. The error in #7862 says that the C# compiler fails sometimes when a required .js file is not present on disk. Here are some questions to clarify the scope of the problem. I don’t have the answers, but giving them consideration may lead you to an alternative that you haven’t thought of yet. 

Could we conditionally skip the EmbeddedResource declaration without damaging the product function? 
Could we commit this generated file to source instead of always building on demand? 
Could we make our npmproj files build from Visual Studio so this file will always exist during a build?

rynowak:
My thoughts on these to the best of my knowledge... since I logged the issue.

> Could we conditionally skip the EmbeddedResource declaration without damaging the product function?

I think it would be really problematic to skip this - we like the fact that the build fails without the file present, we don't want this to ever break silently. Skipping the file would break local F5.

> Could we commit this generated file to source instead of always building on demand?

I don't think would be realistic. This is the build output of our .ts code. We'd like to be able to work on the .ts and .cs part seamlessly.

> Could we make our npmproj files build from Visual Studio so this file will always exist during a build?

That would be my preferred option - but I'm hardly familiar with the issues contained therein.

ryanbrandenburg:
The problem this PR was trying to solve is no longer the same behavior.

PR #8638, opened 2019-03-19
React to breaking changes with M.A.Internal

jkotalik:
Reacting to https://github.com/aspnet/AspNetCore/issues/8637.

I think the following will break once we get new deps for aspnetcore.


sebastienros (approved):

jkotalik:
@sebastienros updated.

jkotalik:
I'll file a follow up issue if this passes to decide whether we should keep certain things public.

jkotalik:
@sebastienros can you confirm this change works?

sebastienros:
Currently verifying this

sebastienros:
I can't binary check the changes because the dll replacement happens once the app is published. But this is a compilation issue with the current runtime. However I checked that all the error messages have the corresponding change in this PR.

PR #8656, opened 2019-03-19
Publish MusicStore to separate directories

pranavkm:


src/MusicStore/test/MusicStore.E2ETests/PublishAndRunTests.cs
```diff
@@ -36,6 +36,7 @@ public async Task PublishAndRun_Test(TestVariant variant)
                     ApplicationPath = Helpers.GetApplicationPath(),
                     PublishApplicationBeforeDeployment = true,
                     PreservePublishedApplicationForDebugging = Helpers.PreservePublishedApplicationForDebugging,
+                    PublishedApplicationRootPath = Path.Combine(Path.GetTempPath(), "MusicStore", Path.GetRandomFileName()),
```
jkotalik: I thought if the PublishedApplicationRootPath was not set, it would use a random temp directory. Is that correct?

pranavkm: You're right:

https://github.com/aspnet/AspNetCore/blob/35746adf68798336bbb6d0f709b66a394c2734e9/src/Hosting/Server.IntegrationTesting/src/Deployers/ApplicationDeployer.cs#L86-L88

So why was file locking an issue to begin with? These tests don't run in parallel with each other.

jkotalik: ¯\\_(ツ)_/¯

PR #8668, opened 2019-03-20
Expose ConnectionId  on .NET Client

mikaelm12:
Exposing the ConnectionId on the .NET Client.

Part of: #8681

### DIFF
--- a/src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
+++ b/src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
@@ -123,12 +123,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
         /// <remarks>
         /// This value will change if the connection is stopped and restarted.
         /// </remarks>
-        public string ConnectionId {
-            get
-            {
-                return _connectionId;
-            }
-        }
+        public string ConnectionId => _connectionId;
 
         /// <summary>
         /// Indicates the state of the <see cref="HubConnection"/> to the server.
--- a/src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
+++ b/src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
@@ -54,6 +54,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
         private long _nextActivationSendPing;
         private bool _disposed;
         private bool _hasInherentKeepAlive;
+        private string _connectionId;
 
         private CancellationToken _uploadStreamToken;
 
@@ -115,6 +116,19 @@ namespace Microsoft.AspNetCore.SignalR.Client
         /// Gets or sets the timeout for the initial handshake.
         /// </summary>
         public TimeSpan HandshakeTimeout { get; set; } = DefaultHandshakeTimeout;
+
+        /// <summary>
+        /// Get the connections current connection Id.
+        /// </summary>
+        /// <remarks>
+        /// This value will change if the connection is stopped and restarted.
+        /// </remarks>
+        public string ConnectionId {
+            get
+            {
+                return _connectionId;
+            }
+        }
 
         /// <summary>
         /// Indicates the state of the <see cref="HubConnection"/> to the server.
@@ -338,6 +352,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
 
                 // Start the connection
                 var connection = await _connectionFactory.ConnectAsync(_protocol.TransferFormat);
+                _connectionId = connection.ConnectionId;
                 var startingConnectionState = new ConnectionState(connection, this);
                 _hasInherentKeepAlive = connection.Features.Get<IConnectionInherentKeepAliveFeature>()?.HasInherentKeepAlive ?? false;
 
--- a/src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
+++ b/src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
@@ -118,11 +118,10 @@ namespace Microsoft.AspNetCore.SignalR.Client
         public TimeSpan HandshakeTimeout { get; set; } = DefaultHandshakeTimeout;
 
         /// <summary>
-        /// Get the connections current connection Id.
+        /// Gets the connection's current Id. This value will be cleared when the connection is stopped and will have a new value every time the connection is (re)established.
+        /// This value will be null if the negotiation step is skipped via HttpConnectionOptions or if the WebSockets transport is explicitly specified because the
+        /// client skips negotiation in that case as well.
         /// </summary>
-        /// <remarks>
-        /// This value will change if the connection is stopped and restarted.
-        /// </remarks>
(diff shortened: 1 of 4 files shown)


mikaelm12:
Looks like I need to do the `GenerateReferenceSource` step. Adding public API

src/SignalR/clients/csharp/Client/test/FunctionalTests/HubConnectionTests.cs
```diff
@@ -155,6 +155,42 @@ public async Task CanStopAndStartConnection(string protocolName, HttpTransportTy
...
+                    var originalClientConnectionId = connection.ConnectionId;
+                    var connectionIdFromServer = await connection.InvokeAsync<string>(nameof(TestHub.GetCallerConnectionId)).OrTimeout();
+                    Assert.Equal(connection.ConnectionId, connectionIdFromServer);
+                    await connection.StopAsync().OrTimeout();
+                    Assert.Null(connection.ConnectionId);
+                    await connection.StartAsync().OrTimeout();
+                    connectionIdFromServer = await connection.InvokeAsync<string>(nameof(TestHub.GetCallerConnectionId)).OrTimeout();
+                    Assert.NotEqual(originalClientConnectionId, connectionIdFromServer);
+                    Assert.Equal(connection.ConnectionId, connectionIdFromServer);
+
```
BrennanConroy: nit: newline

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -116,6 +117,19 @@ public partial class HubConnection
         /// </summary>
         public TimeSpan HandshakeTimeout { get; set; } = DefaultHandshakeTimeout;
 
+        /// <summary>
+        /// Get the connections current connection Id.
+        /// </summary>
+        /// <remarks>
+        /// This value will change if the connection is stopped and restarted.
+        /// </remarks>
+        public string ConnectionId {
```
```suggestion
        public string ConnectionId => _connectionId;
```

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -405,6 +420,8 @@ private async Task StopAsyncCore(bool disposing)
                     (_serviceProvider as IDisposable)?.Dispose();
                     _disposed = true;
                 }
+
+                _connectionId = null;
```
BrennanConroy: Not sure if we care about this, lets see what other people think

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -116,6 +117,14 @@ public partial class HubConnection
         /// </summary>
         public TimeSpan HandshakeTimeout { get; set; } = DefaultHandshakeTimeout;
 
+        /// <summary>
+        /// Get the connections current connection Id.
```
JamesNK: connection's

JamesNK: `Get the connection's current Id.` reads better IMO

JamesNK: Will ConnectionId be null in some cases or has that changed?

A null connection ID for an active connection will confuse some people. It should be mentioned in the comments.

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
halter73: I like nulling it out once the connection is stopped.

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
halter73: I agree we should document when this can be null. Is this only when the client is disconnected? Or is it also possible to establish a connection without negotiating using HubConnection?

halter73 (changes requested):
I think this will be good to go once we update the doc comment to discuss when ConnectionId can be null.

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
JamesNK: I believe a HubConnection with the transport explicitly specified as WebSockets will skip negotiation and never have the connection ID sent to the client.

mikaelm12: Yeah, when negotiation is skipped the client connection Id will be null as well as when the client is disconnected. I can file a docs issue so that we follow up once we get this change in across the clients.

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
davidfowl: Stopped or disposed?

halter73: Stopped. The ConnectionId isn't that useful after you've disconnected, and it can take on multiple values prior to being disposed.

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
halter73: Whether or not we decide this is worth mentioning in a doc article, I think we should also mention this in the doc comments. Something similar to what you have for the Java client should be fine. E.g.

```
/// Gets the connections current Id. This value will be cleared when the connection is stopped and will have a new value every time the connection is (re)established.
/// This value will be null if the WebSockets transport is explicitly specified because the client skips negotiation in that case.
```

mikaelm12: I like that wording. We should also note that this will always be null when the `SkipNegotiation` flag is set to true.

halter73 (approved):

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -116,6 +117,13 @@ public partial class HubConnection
         /// </summary>
         public TimeSpan HandshakeTimeout { get; set; } = DefaultHandshakeTimeout;
 
+        /// <summary>
+        /// Gets the connection's current Id. This value will be cleared when the connection is stopped and will have a new value every time the connection is (re)established.
+        /// This value will be null if the negotiation step is skipped via HttpConnectionOptions or if the WebSockets transport is explicitly specified because the
```
BrennanConroy: This is incorrect. Negotiate is only skipped when Websockets **AND** `SkipNegotiate` are set.

JamesNK: I think SkipNegotiate is only possible when Websockets is set. Could simplify the comment to just meantion it.

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
bradygaster: Good coverage w.r.t. docs. thanks!

Issue #8066, opened 2019-03-01
Unable to compile the SignalR C++ client on MacOS


markusstephanides:
### Describe the bug
I'm not able to compile the SignalR C++ client on MacOS.

### To Reproduce
1. Using the most recent version of the repo and compiled the most recent version of the cpprestsdk
2. Running
`CXX=g++ cmake .. -DCMAKE_BUILD_TYPE=Release -DCPPREST_INCLUDE_DIR=<path_to_cpp_rest_include_dir> -DCPPREST_LIB_DIR=<path_to_cpp_rest_lib_dir>`
3. See errors in the screenshot below

### Expected behavior
Successful compilation (library) of the SignalR cpp client.

### Screenshots
![image](https://user-images.githubusercontent.com/6057570/53631683-6353eb00-3c13-11e9-9d2e-1c8c7d22a683.png)

BrennanConroy:
This is expected right now. We haven't put any effort into the cross-plat build with CMAKE yet, but this is definitely on our radar.

anurse:
Closing in lieu of https://github.com/aspnet/AspNetCore/issues/8706 which is tracking our cross-plat build integration work.

PR #8669, opened 2019-03-20
Update cmake for SignalR C++ client on Mac OS X

wegylexy:
Summary of the changes
 - Skip gtest
 - Patch Boost and OpenSSL in `/usr/local/opt`
```sh
brew install cpprestsdk boost openssl
cmake .. -DCMAKE_CXX_FLAGS=-I/usr/local/opt/openssl/include -DCPPREST_INCLUDE_DIR=/usr/local/opt/cpprestsdk/include -DCPPREST_LIB_DIR=/usr/local/opt/cpprestsdk/lib
make
```

Addresses #8066

### DIFF
--- a/src/SignalR/clients/cpp/CMakeLists.txt
+++ b/src/SignalR/clients/cpp/CMakeLists.txt
@@ -10,10 +10,14 @@ include
 "${CPPREST_INCLUDE_DIR}")
 
 find_library(CPPREST_SO NAMES "cpprest" PATHS ${CPPREST_LIB_DIR} REQUIRED)
+if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+	find_package(Boost COMPONENTS chrono thread REQUIRED)
+endif()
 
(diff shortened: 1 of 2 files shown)


BrennanConroy:
Thanks for your interest in the C++ client!

Unfortunately we aren't ready to take CMake changes yet.

We need to setup our CI builds to build and test the client on the different OS's first so we don't accidentally break the build files. And we'll want more than just macOS working.

We have already started looking into making CMake work cross-plat, you can see the WIP changes in the branch https://github.com/aspnet/AspNetCore/tree/brecon/cmake

I've filed https://github.com/aspnet/AspNetCore/issues/8706 so we can track and discuss things there if you'd like.

wegylexy:
It already works on Ubuntu before this change. Mac just needs more help to find Boost and OpenSSL. For Windows, we have Visual Studio. So after this change, all 3 platforms can build.

anurse:
Thanks for the PR, but we have some build process stuff to integrate with (version numbers to align with and MSBuild goop to connect it up to). As @BrennanConroy said, we're working on getting this in soon. I'm going to close this since the changes we've already got will end up conflicting with this.

PR #8685, opened 2019-03-20
MapController in RazorPages

### LINKED ISSUE #8132: Routing doesn't find controllers when using StarterWeb template
The reason is that the template is missing `router.MapControllers()` call in the `Startup.Configure()` method. This results in the Default Identity UI's `AccountController` not routeable hence the error.

ryanbrandenburg:
Fixes #8132.

That `#if` is my attempt at "#if HasAuth". If there's a better way let me know.

### DIFF
--- a/src/ProjectTemplates/Web.ProjectTemplates/content/RazorPagesWeb-CSharp/Startup.cs
+++ b/src/ProjectTemplates/Web.ProjectTemplates/content/RazorPagesWeb-CSharp/Startup.cs
@@ -161,6 +161,9 @@ namespace Company.WebApplication1
             app.UseRouting(routes =>
             {
                 routes.MapRazorPages();
+#if (IndividualAuth || OrganizationalAuth || WindowsAuth)
+                routes.MapControllers();
+#endif
             });
 
             app.UseCookiePolicy();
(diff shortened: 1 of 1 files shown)


javiercn:
I'm not sure you need controllers in razor pages unless it's aad or b2c auth

javiercn:
You should instantiate the different auth options and try that out

ryanbrandenburg:
I already confirmed that it's needed for IndividualAuth in [the repro](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/797529?src=WorkItemMention&src-action=artifact_link). if you still think we need to check all the types do we have existing AD B2C/aad resources to use and/or instructions on how to set them up?

javiercn:
Inside Azure there are samples, grab the settings from there and change the port in the generated app.

> I already confirmed that it's needed for IndividualAuth in [the repro](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/797529?src=WorkItemMention&src-action=artifact_link).

The template uses horrible terminology for the different options. Individual auth == Identity, the repro is for B2C (although AAD is on the same spot)

javiercn:
I don't think it's needed for windows auth either.

rynowak (approved):

JamesNK (approved):

javiercn (approved):
looks good to me

ryanbrandenburg:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=134406

aspnet-hello:

I've triaged the above build.

PR #8688, opened 2019-03-20
Don't explicitly set ValidationProblemDetails status code

khellang:
Re: https://github.com/aspnet/AspNetCore/issues/6145#issuecomment-461640112

Wrapping it in a `BadRequestObjectResult` below will make sure it has the correct status code set:

https://github.com/aspnet/AspNetCore/blob/aa89639d629899985b27b4ad749f707efc8892cd/src/Mvc/Mvc.Core/src/ObjectResult.cs#L61-L64

### DIFF
--- a/src/Mvc/Mvc.Core/src/DependencyInjection/ApiBehaviorOptionsSetup.cs
+++ b/src/Mvc/Mvc.Core/src/DependencyInjection/ApiBehaviorOptionsSetup.cs
@@ -105,10 +105,7 @@ namespace Microsoft.Extensions.DependencyInjection
 
         internal static IActionResult ProblemDetailsInvalidModelStateResponse(ActionContext context)
         {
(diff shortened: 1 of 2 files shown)


rynowak:
@pranavkm

khellang:
This can probably just be closed and rolled into #12298 😊

pranavkm:
@khellang does the factory introduced as part of #12298 solve this problem?

khellang:
This isn't really a problem, it's just some redundant code I saw and thought should be removed 😊

pranavkm:
@khellang while the other PR did not remove this, we do have a way to override the default status. I'm going to err on the side of caution and leave this behavior be :)

Thanks for your PR though.

khellang:
Yeah, I just sent it cause it was explicitly asked for 😂 https://github.com/aspnet/AspNetCore/issues/6145#issuecomment-461640112

Issue #8681, opened 2019-03-20
Expose ConnectionId on SignalR Clients


mikaelm12:
We've had users ask about how to get a client's connectionId Our current guidance to users on how to get it is to create a Hub method to return it to them e.g.

```csharp
public string GetConnectionId() 
{
    return Context.ConnectionId
}

```
Which is less than ideal.

Another thing we planned on doing for the `onReconnected` events is making it have a single parameter which would be the client's new connectionId. 

- [x] .NET
- [x] Java
- [ ] JavaScript

anurse:
Still have JavaScript to do here right? Is this going to land for preview 4 (end of this week)?

PR #8689, opened 2019-03-20
Java Client Expose ConnectionId on HubConnection

mikaelm12:
Part of : #8681


halter73 (approved):
Seems reasonable. I created https://github.com/aspnet/Docs/issues/11687 to track adding a warning to our docs about exposing the ConnectionId publicly.

PR #8690, opened 2019-03-20
[Fixes #1947][Fixes #1978] Fix and identify flaky components E2E Tests

### LINKED ISSUE #1947: Running a Web API using the latest 1.2.0-preview-1* of Microsoft.AspNetCore gives error
I was asked  to create an issue about this here, after first asking about it over at https://github.com/dotnet/cli/issues/5910

## Steps to reproduce
`dotnet new webapi`

Change the `csproj`-file so that it looks like this:

```
<Project Sdk="Microsoft.NET.Sdk.Web">
 <PropertyGroup>
   <TargetFramework>netcoreapp2.0</TargetFramework>
   <RuntimeFrameworkVersion>2.0.0-beta-001509-00</RuntimeFrameworkVersion>
 </PropertyGroup>

 <ItemGroup>
   <PackageReference Include="Microsoft.AspNetCore" Version="1.2.0-preview1-23666" />
   <PackageReference Include="Microsoft.AspNetCore.Mvc" Version="1.2.0-preview1-23666" />
   <PackageReference Include="Microsoft.Extensions.Logging.Debug" Version="1.2.0-preview1-23666" />
 </ItemGroup>
</Project>
```
Followed by
`dotnet run`

## Expected  behavior
```
Hosting environment: Production
Content root path: c:\dev\test
Now listening on: http://localhost:5000
Application started. Press Ctrl+C to shut down.
```

## Actual behavior
```
Unhandled Exception: System.TypeLoadException: Method 'Resolve' in type 'InnerResolver' from assembly 'Microsoft.CodeAnalysis.Razor, Version=1.2.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60' does not have an implementation.
   at Microsoft.CodeAnalysis.Razor.DefaultTagHelperFeature.OnInitialized()
   at Microsoft.AspNetCore.Razor.Evolution.DefaultRazorEngine..ctor(IRazorEngineFeature[] features, IRazorEnginePhase[] phases)
   at Microsoft.AspNetCore.Razor.Evolution.DefaultRazorEngineBuilder.Build()
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitScoped(ScopedCallSite scopedCallSite, ServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitConstructor(ConstructorCallSite constructorCallSite, ServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitTransient(TransientCallSite transientCallSite, ServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitConstructor(ConstructorCallSite constructorCallSite, ServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitScoped(ScopedCallSite scopedCallSite, ServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitConstructor(ConstructorCallSite constructorCallSite, ServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitScoped(ScopedCallSite scopedCallSite, ServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitClosedIEnumerable(ClosedIEnumerableCallSite closedIEnumerableCallSite, ServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitTransient(TransientCallSite transientCallSite, ServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitConstructor(ConstructorCallSite constructorCallSite, ServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitScoped(ScopedCallSite scopedCallSite, ServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitConstructor(ConstructorCallSite constructorCallSite, ServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitScoped(ScopedCallSite scopedCallSite, ServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ServiceProvider.<>c__DisplayClass18_0.<RealizeService>b__0(ServiceProvider provider)
   at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService(IServiceProvider provider, Type serviceType)
   at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService[T](IServiceProvider provider)
   at Microsoft.AspNetCore.Builder.MvcApplicationBuilderExtensions.UseMvc(IApplicationBuilder app, Action`1 configureRoutes)
   at test.Startup.Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerFactory loggerFactory) in c:\dev\test\Startup.cs:line 40
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at Microsoft.AspNetCore.Hosting.ConventionBasedStartup.Configure(IApplicationBuilder app)
   at Microsoft.AspNetCore.Hosting.Internal.WebHost.BuildApplication()
   at Microsoft.AspNetCore.Hosting.WebHostBuilder.Build()
   at test.Program.Main(String[] args) in c:\dev\test\Program.cs:line 15
```

## Environment data
`dotnet --info` output:
.NET Command Line Tools (2.0.0-alpha-004853)

Product Information:
 Version:            2.0.0-alpha-004853
 Commit SHA-1 hash:  e7689a3116

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.14393
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\2.0.0-alpha-004853\

## Notes
If you change the version of the reference packages in the `csproj`-file to `1.2.0-preview1-23656` instead of `1.2.0-preview1-23666` (the version before it) then this error does not occur.

### LINKED ISSUE #1978: Support for "classic" ASP.NET
I was planning to migrate a couple of ASP.NET 4 projects to the new csproj format but [this article](http://www.natemcmaster.com/blog/2017/03/09/vs2015-to-vs2017-upgrade/) says it's not supported.

Is it actually true? If so, it would be a pity since statistically the web project file is where most of our conflicts happen.

Also, current file size: 267 KB...

javiercn:
Marks https://github.com/aspnet/AspNetCore-Internal/issues/1987 as flaky, but there's a real product issue behind it https://github.com/aspnet/AspNetCore/issues/8204

### DIFF
--- a/src/Components/test/E2ETest/ServerExecutionTests/ServerSideAppTest.cs
+++ b/src/Components/test/E2ETest/ServerExecutionTests/ServerSideAppTest.cs
@@ -1,14 +1,13 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
+using System.Linq;
 using Microsoft.AspNetCore.Components.E2ETest.Infrastructure;
 using Microsoft.AspNetCore.Components.E2ETest.Infrastructure.ServerFixtures;
 using Microsoft.AspNetCore.E2ETesting;
 using OpenQA.Selenium;
 using OpenQA.Selenium.Support.UI;
-using System;
-using System.Linq;
-using System.Threading.Tasks;
 using Xunit;
 using Xunit.Abstractions;
 
@@ -41,7 +40,7 @@ namespace Microsoft.AspNetCore.Components.E2ETest.ServerExecutionTests
         [Fact]
         public void HasHeading()
         {
-            Assert.Equal("Hello, world!", Browser.FindElement(By.TagName("h1")).Text);
+            Browser.Equal("Hello, world!", () => Browser.FindElement(By.CssSelector("h1#index")).Text);
         }
 
         [Fact]
@@ -96,7 +95,7 @@ namespace Microsoft.AspNetCore.Components.E2ETest.ServerExecutionTests
         {
             // Navigate to "Fetch Data"
             Browser.FindElement(By.LinkText("Fetch data")).Click();
-            Browser.Equal("Weather forecast", () => Browser.FindElement(By.TagName("h1")).Text);
+            Browser.Equal("Weather forecast", () => Browser.FindElement(By.CssSelector("h1#fetch-data")).Text);
 
             // Wait until loaded
             var tableSelector = By.CssSelector("table.table");
--- a/src/Components/test/E2ETest/Tests/EventTest.cs
+++ b/src/Components/test/E2ETest/Tests/EventTest.cs
@@ -6,6 +6,7 @@ using BasicTestApp;
 using Microsoft.AspNetCore.Components.E2ETest.Infrastructure;
 using Microsoft.AspNetCore.Components.E2ETest.Infrastructure.ServerFixtures;
 using Microsoft.AspNetCore.E2ETesting;
+using Microsoft.AspNetCore.Testing.xunit;
 using OpenQA.Selenium;
 using OpenQA.Selenium.Interactions;
 using Xunit;
@@ -133,6 +134,7 @@ namespace Microsoft.AspNetCore.Components.E2ETest.Tests
         }
 
         [Fact]
+        [Flaky("https://github.com/aspnet/AspNetCore-Internal/issues/1987")]
         public void InputEvent_RespondsOnKeystrokes()
         {
             MountTestComponent<InputEventComponent>();
--- a/src/Components/test/testassets/ComponentsApp.App/Pages/FetchData.cshtml
+++ b/src/Components/test/testassets/ComponentsApp.App/Pages/FetchData.cshtml
@@ -2,7 +2,7 @@
 @page "/fetchdata/{StartDate:datetime}"
 @inject WeatherForecastService ForecastService
 
-<h1>Weather forecast</h1>
+<h1 id="fetch-data">Weather forecast</h1>
 
 <p>This component demonstrates fetching data from the server.</p>
 
--- a/src/Components/test/testassets/ComponentsApp.App/Pages/Index.cshtml
+++ b/src/Components/test/testassets/ComponentsApp.App/Pages/Index.cshtml
@@ -1,5 +1,5 @@
-﻿@page "/"
+@page "/"
 
-<h1>Hello, world!</h1>
+<h1 id="index">Hello, world!</h1>
 
 Welcome to your new app.


src/Components/test/E2ETest/ServerExecutionTests/ServerSideAppTest.cs
```diff
@@ -41,7 +40,7 @@ public void HasTitle()
         [Fact]
         public void HasHeading()
         {
-            Assert.Equal("Hello, world!", Browser.FindElement(By.TagName("h1")).Text);
+            Browser.Equal("Hello, world!", () => Browser.FindElement(By.CssSelector("h1#index")).Text);
```
javiercn: This assert is better because it waits for the element to appear, which is good as the issue could come from the page not being navigated from the previous location. Also uses a more specific way to find the element for that reason.

src/Components/test/E2ETest/ServerExecutionTests/ServerSideAppTest.cs
```diff
@@ -96,7 +95,7 @@ public void HasFetchDataPage()
         {
             // Navigate to "Fetch Data"
             Browser.FindElement(By.LinkText("Fetch data")).Click();
-            Browser.Equal("Weather forecast", () => Browser.FindElement(By.TagName("h1")).Text);
+            Browser.Equal("Weather forecast", () => Browser.FindElement(By.CssSelector("h1#fetch-data")).Text);
```
javiercn: This can happen because we click on the link above and the navigation happens between us finding the current h1 and the one in the page. (Produces a stale reference exception). Use a more specific CSS selector to prevent it.

src/Components/test/E2ETest/Tests/EventTest.cs
```diff
@@ -133,6 +134,7 @@ public void PreventDefault_DotNotApplyByDefault()
         }
 
         [Fact]
+        [Flaky("https://github.com/aspnet/AspNetCore-Internal/issues/1987")]
```
javiercn: The flakiness is due to a product issue. Marking it as such until we fix the underlying bug.

rynowak (approved):

src/Components/test/E2ETest/ServerExecutionTests/ServerSideAppTest.cs
SteveSandersonMS: > Use a more specific CSS selector to prevent it.

I don't think this is correct. The behavior of `WaitAssertCore` is to keep trying the assertion until it passes (until some timeout). Using a more specific selector won't change this, because it would have kept retrying anyway if the assertion failed. Or am I missing something?

I'm not saying your change is wrong; I'm just saying that there must be some other cause for any flakiness here.

src/Components/test/E2ETest/ServerExecutionTests/ServerSideAppTest.cs
SteveSandersonMS: As per other comment, I agree it's right to switch from `Assert` to `Browser` here, but I disagree that we need to change the CSS selector.

src/Components/test/testassets/ComponentsApp.App/Pages/FetchData.cshtml
```diff
@@ -2,7 +2,7 @@
 @page "/fetchdata/{StartDate:datetime}"
 @inject WeatherForecastService ForecastService
 
-<h1>Weather forecast</h1>
+<h1 id="fetch-data">Weather forecast</h1>
```
SteveSandersonMS: Shouldn't be needed

src/Components/test/testassets/ComponentsApp.App/Pages/Index.cshtml
```diff
@@ -1,5 +1,5 @@
-﻿@page "/"
+@page "/"
 
-<h1>Hello, world!</h1>
+<h1 id="index">Hello, world!</h1>
```
SteveSandersonMS: Shouldn't be needed

src/Components/test/E2ETest/ServerExecutionTests/ServerSideAppTest.cs
javiercn: The issue was a stale element exception.
What likely happened was that the page changed from Index to FetchData while Browser.FindElement was executing, so it would find H1 on index and then that node would become detached from the DOM when accessing text, which would trigger the exception.

By using the more specific selector, the first find would fail until we navigated to the new page and find the h1 we were looking for. In general, based on my experience, it pays off to be more specific here.

That's only my theory of what was happening here.

src/Components/test/testassets/ComponentsApp.App/Pages/FetchData.cshtml
javiercn: I've explained my motives above. Is there any specific reason why you think this is an issue? I believe that in principle, being more specific with these tests is better, but I'm all ears.

src/Components/test/E2ETest/ServerExecutionTests/ServerSideAppTest.cs
SteveSandersonMS: The waitassert logic is here: https://github.com/aspnet/AspNetCore/blob/9f1a978230cdd161998815c425bfd2d25e8436b6/src/Shared/E2ETesting/WaitAssert.cs#L55

It does a `try`/`catch` around the assertion code, so if it throws for any reason, we just keep retrying in case it passes before the eventual timeout. As such I don't see how it would make any difference whether the first find fails or not.

src/Components/test/testassets/ComponentsApp.App/Pages/FetchData.cshtml
SteveSandersonMS: Let's be sure we understand the real cause of the issue before we know what the principles are :)

src/Components/test/E2ETest/ServerExecutionTests/ServerSideAppTest.cs
javiercn: Ok. My bad this is actually a timeout that is hidden by this line:
https://github.com/aspnet/AspNetCore/blob/master/src/Shared/E2ETesting/WaitAssert.cs#L69

There is no way to know the real cause I believe as re-running the assertion is not correct.
Given this is the case, we should modify WaitAssertCore to capture the last exception in a local and rethrow that.

Also, how often does this thing poll?

SteveSandersonMS: > we should modify WaitAssertCore to capture the last exception in a local and rethrow that.

Isn't that the same as what it already does? It already keeps running the same callback over and over, and if the `WebDriverWait` times out, then it runs the same callback one more time but this time lets any exception bubble up the call stack.

> Also, how often does this thing poll?

I'm not certain if our code modifies the defaults anywhere, but apparently the default is to poll every 500ms (https://stackoverflow.com/a/44645991).

javiercn: > Isn't that the same as what it already does?

No. It reruns the assertion and that can cause a different exception than the last one.

SteveSandersonMS: It's runs the assertion N times in a loop until there's a timeout, then runs it again for the (N+1)th time. For what reason would we prefer to report the Nth exception instead of the (N+1)th exception?

javiercn: I believe so. If not, I can't think how that test failed.

PR #8691, opened 2019-03-20
Enable AuthSamples.FunctionalTests on helix

HaoK:
Publish additional files needed to be able to run tests from publish directory


HaoK:
@Tratcher looks it just took copying over a few of the samples directories over to the content root (publish dir) is what it takes to get these working on helix

Tratcher (approved):

Issue #5903, opened 2017-08-26
Consider adding a IWebHostBuilder.Configure overload that takes a WebHostBuilderContext


cwe1ss:
Would be nice/consistent to have this additional overload as it also exists for `ConfigureAppConfiguration` and `ConfigureServices`. It would allow us to access the configuration and environment info without having to resolve it ourselves from DI.

danielmarbach:
Sent in a [PR ](https://github.com/aspnet/AspNetCore/pull/8697)that addresses this. I'm still downloading VS2019RC and see if I can even compile my changes locally but the PR is there

PR #8697, opened 2019-03-21
Add overload for Configure that gets WebHostBuilderContext

danielmarbach:
Adds an overload for Configure that gets the WebHostBuilderContext to avoid having to assign `IHostingEnvironment` and `IConfiguration` to static fields like shown in https://docs.microsoft.com/en-us/aspnet/core/fundamentals/startup?view=aspnetcore-2.2#convenience-methods

I stopped expanding this further down to `IStartup` and `IStartupFilter`. I'll leave the breaking change decision to you folks

Addresses #5903

### DIFF
--- a/src/Hosting/Hosting/src/WebHostBuilderExtensions.cs
+++ b/src/Hosting/Hosting/src/WebHostBuilderExtensions.cs
@@ -37,7 +37,7 @@ namespace Microsoft.AspNetCore.Hosting
             return hostBuilder.Configure(configureApp, configureApp.GetMethodInfo().DeclaringType.GetTypeInfo().Assembly.GetName().Name);
         }
 
-        static IWebHostBuilder Configure(this IWebHostBuilder hostBuilder, Action<WebHostBuilderContext, IApplicationBuilder> configureApp, string startupAssemblyName)
+        private static IWebHostBuilder Configure(this IWebHostBuilder hostBuilder, Action<WebHostBuilderContext, IApplicationBuilder> configureApp, string startupAssemblyName)
         {
             if (configureApp == null)
             {
--- a/src/Hosting/Hosting/src/WebHostBuilderExtensions.cs
+++ b/src/Hosting/Hosting/src/WebHostBuilderExtensions.cs
@@ -23,12 +23,26 @@ namespace Microsoft.AspNetCore.Hosting
         /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
         public static IWebHostBuilder Configure(this IWebHostBuilder hostBuilder, Action<IApplicationBuilder> configureApp)
         {
+            return hostBuilder.Configure((_, app) => configureApp(app), configureApp.GetMethodInfo().DeclaringType.GetTypeInfo().Assembly.GetName().Name);
+        }
+
+        /// <summary>
+        /// Specify the startup method to be used to configure the web application.
+        /// </summary>
+        /// <param name="hostBuilder">The <see cref="IWebHostBuilder"/> to configure.</param>
+        /// <param name="configureApp">The delegate that configures the <see cref="IApplicationBuilder"/>.</param>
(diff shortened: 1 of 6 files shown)


Tratcher:
The change looks good overall but there are some suspicious test failures. Did you run the tests locally?

danielmarbach:
When I cloned master github build status indicator already showed some failing builds. So I did rely on build pipelines only. Happy to run the test locally tomorrow

danielmarbach:
I looked into these tests and could not see how they are related to my changes. Did a quick rebase and pushed again

danielmarbach:
Frankly speaking, I'm not sure what I'm supposed to do now

> E:\A\_work\1\s\.dotnet\buildtools\korebuild\3.0.0-build-20190314.2\modules\vstest\module.targets(158,5): error : Test group UnitTests/netcoreapp3.0 failed with exit code '1'. [E:\A\_work\1\s\.dotnet\buildtools\korebuild\3.0.0-build-20190314.2\KoreBuild.proj]

Any changes @Tratcher or @anurse can have a look and see whether this is related to my changes or not?

Tratcher:
See https://dev.azure.com/dnceng/public/_build/results?buildId=130733&view=ms.vss-test-web.build-test-results-tab
~25 failing WebHostBuilder tests so it's almost certainly related to your change. I'll try it locally.

src/Hosting/Hosting/src/WebHostBuilderExtensions.cs
```diff
@@ -22,6 +22,17 @@ public static class WebHostBuilderExtensions
         /// <param name="configureApp">The delegate that configures the <see cref="IApplicationBuilder"/>.</param>
         /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
         public static IWebHostBuilder Configure(this IWebHostBuilder hostBuilder, Action<IApplicationBuilder> configureApp)
+        {
+            return hostBuilder.Configure((_, app) => configureApp(app));
```
Tratcher: I see the problem. This added indirection breaks the following line in Configure:
https://github.com/aspnet/AspNetCore/blob/c3f361fc5395d3ef50b876a7f1850695371e9604/src/Hosting/Hosting/src/WebHostBuilderExtensions.cs#L42
The new declaring type for configureApp is WebHostBuilderExtensions, not Program.

This may be best addressed by adding a third private overload of Configure with most of the functionality but that takes startupAssemblyName and having both of these call into it.

danielmarbach: Thanks for the hint. Didn't spot that. Fixed now

src/Hosting/Hosting/src/WebHostBuilderExtensions.cs
```diff
@@ -22,14 +22,28 @@ public static class WebHostBuilderExtensions
...
+        /// </summary>
+        /// <param name="hostBuilder">The <see cref="IWebHostBuilder"/> to configure.</param>
+        /// <param name="configureApp">The delegate that configures the <see cref="IApplicationBuilder"/>.</param>
+        /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
+        public static IWebHostBuilder Configure(this IWebHostBuilder hostBuilder, Action<WebHostBuilderContext, IApplicationBuilder> configureApp)
+        {
+            return hostBuilder.Configure(configureApp, configureApp.GetMethodInfo().DeclaringType.GetTypeInfo().Assembly.GetName().Name);
+        }
+
+        static IWebHostBuilder Configure(this IWebHostBuilder hostBuilder, Action<WebHostBuilderContext, IApplicationBuilder> configureApp, string startupAssemblyName)
```
Tratcher: private
Always include access modifiers.

Tratcher (approved):

danielmarbach:
Good to go?

Tratcher:
Thanks

PR #8698, opened 2019-03-21
use fields to avoid hard coded state classes

kennywangjin:
- add static read-write fields to EditContext to hold validation state classes
- use these classes to apply field validation state
- user could customize their own classes

### DIFF
--- a/src/Components/Components/src/Forms/EditContext.cs
+++ b/src/Components/Components/src/Forms/EditContext.cs
@@ -13,6 +13,12 @@ namespace Microsoft.AspNetCore.Components.Forms
     /// </summary>
     public sealed class EditContext
     {
+        public static string ModifiedClass = "modified";
+        public static string ValidClass = "valid";
+        public static string InvalidClass = "invalid";
+        public static string ValidationMessageClass = "validation-message";
+        public static string ValidationErrorsClass = "validation-errors";
+
         // Note that EditContext tracks state for any FieldIdentifier you give to it, plus
         // the underlying storage is sparse. As such, none of the APIs have a "field not found"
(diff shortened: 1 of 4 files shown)


src/Components/Components/src/Forms/EditContext.cs
```diff
@@ -13,6 +13,12 @@ namespace Microsoft.AspNetCore.Components.Forms
     /// </summary>
     public sealed class EditContext
     {
+        public static string ModifiedClass = "modified";
```
gfoidl: Why don't you use constants?
And do they have to be `public`? `private` would be enough, as long no use outside this class is planned.

BTW: If you use `static readonly` the JIT can treat them es constants. Anyway, here `const` would be the better way.

SteveSandersonMS: @gfoidl The point of @kennywangjin's PR here is to make these values mutable, hence not `const`.

I'm not sure we'll use this approach, but hopefully it explains why the PR is as it is. It's quite likely that we'd move the responsibility for this into `EditForm` and have it configurable on a per-instance basis instead.

gfoidl: > hopefully it explains why the PR is as it is

Yes, thanks! 
(Unfortunately there's no reference to any issue, etc. in the description of the PR, and the boilerplate "Summary of the changes (Less than 80 chars)" distracted me.)

> have it configurable on a per-instance basis instead

Seems reasonable.

mkArtakMSFT:
Thanks for your effort. We won't be accepting your change as is, as global static properties would allow breaking this from some other place.

PR #8725, opened 2019-03-21
React to sem version 2.0

jkotalik:
Fixes https://github.com/dotnet/core-setup/commit/8ee7c85052193cbf064b7da16c34a8f50d94695a
Reacting to https://github.com/dotnet/core-setup/issues/4795. Hostfxr dlls could start using sem ver 2.0, so let's make sure we can handle it. 

Today, we source copy fx_ver.cpp/h to handle finding the highest version of hostfxr.dll. This updates the source copy and adds a test. It also modifies pal::string_t to std::wstring and a few other small things.

I didn't think of a great way to add a test in managed code; it would require copying another version of hostfxr.dll into the .dotnet directory and then deleting it afterwards. I added a test for global versioning which is used for ANCM out of process that tests

### DIFF
--- a/src/Servers/IIS/AspNetCoreModuleV2/CommonLib/fx_ver.cpp
+++ b/src/Servers/IIS/AspNetCoreModuleV2/CommonLib/fx_ver.cpp
@@ -1,11 +1,32 @@
-// Copyright (c) .NET Foundation and contributors. All rights reserved.
-// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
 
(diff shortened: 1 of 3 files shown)


src/Servers/IIS/AspNetCoreModuleV2/CommonLib/fx_ver.cpp
```diff
@@ -1,11 +1,32 @@
...
-#include <Windows.h>
-#include <sstream>
-#include <cassert>
+static bool validIdentifiers(const std::wstring& ids);
+
+size_t index_of_non_numeric(const std::wstring& str, size_t i)
+{
+    return str.find_first_not_of(TEXT("0123456789"), i);
+}
+
```
BrennanConroy: nit: newline

BrennanConroy (approved):

PR #8726, opened 2019-03-21
backport accessTokenProvider fix

mikaelm12:
Fixes: https://github.com/aspnet/AspNetCore/issues/8640

Backporting a fix for issue: https://github.com/aspnet/AspNetCore/issues/8429 into 2.2.x

# Description:
When connecting to the Azure SignalR Service we take the access token from the negotiate response and wrap that in a single and override the user provided `accessTokenProvider` is there was one. That behavior is correct. What is incorrect is that we don't keep a reference to the original passed in `accessTokenProvider` so we can restore it if we restart the connection. 

# Customer Impact
Without this change customers who have provided an accessToken provider and are connecting to the Azure SignalR Service won't be able to restart their connections properly.

# Regression?
Nope

# Risk:
Low. This just makes sure that we can reset the user provided accessTokenProvider if there was one in the event the client is connecting to the Azure SignalR Service.

### DIFF
--- a/src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
+++ b/src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
@@ -44,6 +44,7 @@ public class HubConnection {
     private List<OnClosedCallback> onClosedCallbackList;
     private final boolean skipNegotiate;
     private Single<String> accessTokenProvider;
+    private Single<String> redirectAccessTokenProvider;
     private final Map<String, String> headers = new HashMap<>();
     private ConnectionState connectionState = null;
     private final HttpClient httpClient;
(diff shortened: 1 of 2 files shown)


BrennanConroy:
cc @anurse

anurse:
Sounds good. I'll bring this to shiproom.

vivmishra:
Approved for 2.2.5
Wait for branch to open.

BrennanConroy (approved):

anurse:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=130379

aspnet-hello:

I've triaged the above build.

dougbu:
@anurse please confirm no further changes are necessary to ensure the updated Java packages are built and flow correctly.

Once that is done, PR looks ready to merge. The branches in this repo are now open.

anurse:
Branch is now open, please merge ASAP.

> please confirm no further changes are necessary to ensure the updated Java packages are built and flow correctly

This is correct, they should be produced entirely during our build. I will need a drop share of the final products of our build (since the Java assets don't flow through the reset of the orchestrated build) when it comes to release time (this is like the NPM packages).

mikaelm12:
I don't have authorization to merge actually. Either @anurse or @dougbu  will have to click the button.

anurse:
Ah yes, I got you covered :).

mikaelm12:
Thanks!

PR #8730, opened 2019-03-22
Copy EndpointFeature when using LongPolling

BrennanConroy:
Fixes https://github.com/aspnet/AspNetCore/issues/8679

@JamesNK Is it ok to reference the feature from the original HttpContext? Or should we do some kind of deep copy?

~~TODO: Write a test~~

### DIFF
--- a/src/SignalR/common/Http.Connections/src/Internal/HttpConnectionDispatcher.cs
+++ b/src/SignalR/common/Http.Connections/src/Internal/HttpConnectionDispatcher.cs
@@ -567,7 +567,7 @@ namespace Microsoft.AspNetCore.Http.Connections.Internal
             requestFeature.PathBase = existingRequestFeature.PathBase;
             requestFeature.QueryString = existingRequestFeature.QueryString;
             requestFeature.RawTarget = existingRequestFeature.RawTarget;
-            var requestHeaders = new Dictionary<string, StringValues>(existingRequestFeature.Headers.Count, StringComparer.Ordinal);
+            var requestHeaders = new Dictionary<string, StringValues>(existingRequestFeature.Headers.Count, StringComparer.OrdinalIgnoreCase);
             foreach (var header in existingRequestFeature.Headers)
             {
                 requestHeaders[header.Key] = header.Value;
@@ -590,6 +590,11 @@ namespace Microsoft.AspNetCore.Http.Connections.Internal
             var responseFeature = new HttpResponseFeature();
 
             var endpointFeature = context.Features.Get<IEndpointFeature>();
+
+            if (endpointFeature != null)
+            {
+                endpointFeature = new CopiedEndpointFeature(endpointFeature.Endpoint);
+            }
 
             var features = new FeatureCollection();
             features.Set<IHttpRequestFeature>(requestFeature);
@@ -674,6 +679,19 @@ namespace Microsoft.AspNetCore.Http.Connections.Internal
         {
             public static EmptyServiceProvider Instance { get; } = new EmptyServiceProvider();
             public object GetService(Type serviceType) => null;
+        }
+
+        private class CopiedEndpointFeature : IEndpointFeature
+        {
+            public CopiedEndpointFeature(Endpoint endpoint)
+            {
+                _endpoint = endpoint;
+            }
+
+            private Endpoint _endpoint;
+
+            // REVIEW: Do we care if the user modifies the Endpoint?
+            public Endpoint Endpoint { get => _endpoint; set => throw new InvalidOperationException("The Endpoint is no longer modifiable."); }
         }
     }
 }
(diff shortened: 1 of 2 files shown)


JamesNK:
> @JamesNK Is it ok to reference the feature from the original HttpContext? Or should we do some kind of deep copy?

An endpoint is immutable so you don't need to copy that.

The implementation of IEndpointFeature is EndpointSelectorContext which also implements some other features - https://github.com/aspnet/AspNetCore/blob/85e2147ff037cc9950249fbb9d095ad47ec4184a/src/Http/Routing/src/EndpointSelectorContext.cs

Nothing on it is cached and reused between requests - today - but you would drag in some extra objects if you didn't create a copy.

BrennanConroy:
Ok, so we should make a custom `IEndpointFeature` class that just copies the `Endpoint`?

src/SignalR/common/Http.Connections/src/Internal/HttpConnectionDispatcher.cs
```diff
@@ -589,10 +589,13 @@ private static HttpContext CloneHttpContext(HttpContext context)
             // The response is a dud, you can't do anything with it anyways
             var responseFeature = new HttpResponseFeature();
 
+            var endpointFeature = context.Features.Get<IEndpointFeature>();
+
             var features = new FeatureCollection();
             features.Set<IHttpRequestFeature>(requestFeature);
             features.Set<IHttpResponseFeature>(responseFeature);
             features.Set<IHttpConnectionFeature>(connectionFeature);
+            features.Set<IEndpointFeature>(endpointFeature);
```
JamesNK: IMO you should create a basic implementation of IEndpointFeature, copy the endpoint across, and set that.

JamesNK (commented):
> @JamesNK Is it ok to reference the feature from the original HttpContext? Or should we do some kind of deep copy?

An endpoint is immutable so you don't need to copy that.

The implementation of IEndpointFeature is EndpointSelectorContext which also implements some other features - https://github.com/aspnet/AspNetCore/blob/85e2147ff037cc9950249fbb9d095ad47ec4184a/src/Http/Routing/src/EndpointSelectorContext.cs

Nothing on it is cached and reused between requests - today - but you would drag in some extra objects if you didn't create a copy.

BrennanConroy:
cc @pranavkm I tested this manually and LongPolling works now. I don't know anything about the Components testing infrastructure, so either you can add a test later, or help me write one.

src/SignalR/common/Http.Connections/src/Internal/HttpConnectionDispatcher.cs
```diff
@@ -672,5 +680,18 @@ private class EmptyServiceProvider : IServiceProvider
...
         }
+
+        private class CopiedEndpointFeature : IEndpointFeature
+        {
+            public CopiedEndpointFeature(Endpoint endpoint)
+            {
+                _endpoint = endpoint;
+            }
+
+            private Endpoint _endpoint;
```
JamesNK: readonly

src/SignalR/common/Http.Connections/src/Internal/HttpConnectionDispatcher.cs
anurse: For the other features we have a basic implementation available in the `Microsoft.AspNetCore.Http` package for exactly this purpose. Should we do something similar here? [`EndpointFeature`](https://github.com/aspnet/AspNetCore/blob/425c196cba530b161b120a57af8f1dd513b96f67/src/Http/Http.Abstractions/src/Routing/EndpointHttpContextExtensions.cs#L65) looks pretty dirt-simple and if we made it public it would basically be the same as whatever we could make here.

JamesNK: I keep forgetting the `SetEndpoint` extension method exists 😕 

No I don't think we need to implement EndpointFeature because people can use `SetEndpoint` instead.

Speaking of which, @BrennanConroy instead of creating a basic endpoint class you could use `newHttpContext.SetEndpoint(endpoint)`

JamesNK (approved):

src/SignalR/common/Http.Connections/src/Internal/HttpConnectionDispatcher.cs
```diff
@@ -600,6 +601,9 @@ private static HttpContext CloneHttpContext(HttpContext context)
             var newHttpContext = new DefaultHttpContext(features);
             newHttpContext.TraceIdentifier = context.TraceIdentifier;
 
+            var endpointFeature = context.Features.Get<IEndpointFeature>();
+            newHttpContext.SetEndpoint(endpointFeature?.Endpoint);
```
Tratcher: @JamesNK is this a deep enough copy? Does the endpoint get disposed or reused anywhere?

Tratcher (approved):

src/SignalR/common/Http.Connections/src/Internal/HttpConnectionDispatcher.cs
JamesNK: Endpoints are 🔥 immutable 🔥

PR #8731, opened 2019-03-22
Experience for Razor SDK

pranavkm:
* Remove Components.Build
* Update MVC projects and templates to use AddRazorSupportForMvc
* Update components templates to use .razor extension

Fixes https://github.com/aspnet/AspNetCore/issues/6392


eng/Version.Details.xml
```diff
@@ -9,21 +9,21 @@
 -->
 <Dependencies>
   <ProductDependencies>
-    <Dependency Name="Microsoft.AspNetCore.Razor.Language" Version="3.0.0-preview4.19168.1">
+    <Dependency Name="Microsoft.AspNetCore.Razor.Language" Version="3.0.0-preview4.19171.3">
```
dougbu: Do not make version changes with the old version of `darc` or manually. In this case, there are newer coherent versions that will be available tomorrow morning.

dougbu (changes requested):

rynowak:
@pranavkm - we don't want to merge the `.razor` change until we have support in tooling for `_Imports.razor`.

SteveSandersonMS:
LGTM besides Ryan's point about not doing all the .cshtml->.razor renames just yet.

I'm unfamiliar with the design intent behind `AddRazorSupportForMvc`. I can see the implementation is [here](https://github.com/aspnet/AspNetCore-Tooling/blob/cfee40a19eb42f0c02365104c0a460e1025ad84c/src/Razor/src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Microsoft.NET.Sdk.Razor.Configuration.targets#L33), but I'd be interested to hear more about the cases where this would and would not be used, if you have a few mins to chat.

ryanbrandenburg (approved):
Looks fine WRT templates.

dougbu (approved):
eng/* changes look good but I didn't review the rest

PR #8748, opened 2019-03-22
Fix jasmine timeout

mikaelm12:
Issue: https://github.com/aspnet/AspNetCore-Internal/issues/1961
We were manually setting the timeout in the HubConnection and Connection tests to 10 seconds. We had logic to set it to 20 seconds but it was getting written over. The logs were explicitly saying that the timeout allowed was only 10 seconds. I did a manual sauce labs build of this and it passed but I want to have it on the daily cadence to see if it brings stability.

### DIFF
--- a/src/SignalR/clients/ts/FunctionalTests/ts/Common.ts
+++ b/src/SignalR/clients/ts/FunctionalTests/ts/Common.ts
@@ -5,7 +5,7 @@ import { HttpTransportType, IHubProtocol, JsonHubProtocol } from "@aspnet/signal
 import { MessagePackHubProtocol } from "@aspnet/signalr-protocol-msgpack";
 
 // On slower CI machines, these tests sometimes take longer than 5s
-jasmine.DEFAULT_TIMEOUT_INTERVAL = 10 * 1000;
+jasmine.DEFAULT_TIMEOUT_INTERVAL = 20 * 1000;
 
 export let ENDPOINT_BASE_URL: string = "";
 export let ENDPOINT_BASE_HTTPS_URL: string = "";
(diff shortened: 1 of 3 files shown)


BrennanConroy:
Wont this be 40 seconds now? https://github.com/aspnet/AspNetCore/blob/93bd018e9e8ef0437cca79fc749d9ac469cc69ff/src/SignalR/clients/ts/FunctionalTests/ts/Common.ts#L36

mikaelm12:
It seems like that logic is getting respected

mikaelm12:
I did  a manual run yesterday with the default set to 10 and removing the overrides in HubConnection and Connection tests and the timeout still read 10s

BrennanConroy:
We should probably figure that out. Maybe we need double quotes here https://github.com/aspnet/AspNetCore/blob/master/src/SignalR/clients/ts/FunctionalTests/scripts/run-tests.ts#L338

halter73:
> I did a manual run yesterday with the default set to 10 and removing the overrides in HubConnection and Connection tests and the timeout still read 10s

I'm not sure I understand this comment. Is this not what you expected?

BrennanConroy:
>Is this not what you expected?

No, we have a check for sauce labs and double the timeout. But it looks like that check isn't working so it stays at 10 seconds.

halter73:
Do we think that's because the "sauce" variable is being appropriately set?

mikaelm12:
Yeah, for what ever reason the doubling logic it's not being respected. At this point I think we should just set the default timeout to 20 seconds so see if this stabilizes the build.

BrennanConroy (approved):

PR #8749, opened 2019-03-22
Mark Flaky Tests

mikaelm12:
Issue: https://github.com/aspnet/AspNetCore-Internal/issues/2040
Marking some tests as flaky


src/Hosting/Hosting/test/Internal/HostingEventSourceTests.cs
```diff
@@ -107,6 +108,7 @@ public void HostStop()
 
         [Theory]
         [MemberData(nameof(RequestStartData))]
+        [Flaky("https://github.com/aspnet/AspNetCore-Internal/issues/2040", "All")]
```
mikaelm12: For whatever reason the `FlakyOn.All` constant wasn't resolving so I just passed in the actual value here

Tratcher (approved):

Issue #5510, opened 2018-04-28
Route segments are not decoded


GoldenCrystal:
### Title
Route segments are not decoded

### Functional impact
Spaces, non-ASCII characters, or other unsafe characters will not work properly inside URLs out of the box.
This impacts routes such as ````@page "/ブラゾル" ````, or any route parameter that would contain unsafe characters.

### Minimal repro steps
1. Create a new Blazor application
2. Create a page like this one:
````
@page "/wiki/{topic}"

<p>@Topic</p>

@functions {
    public string Topic { get; set; }
}
````
3. Start the application and navigate to "/wiki/déjà vu" in your Browser

### Expected result
I expect the page to display "déjà vu"

### Actual result
Instead, I see "d%C3%A9j%C3%A0%20vu"

### Further technical details
I tested this on Chrome 65 and Firefox 59.

It seems that the simplified parsing implementation in [RouteContext](https://github.com/aspnet/Blazor/blob/dev/src/Microsoft.AspNetCore.Blazor/Routing/RouteContext.cs) could easily be improved for handling this scenario.

I'm ok with providing a PR for this, if needed.

rynowak:
We'd be willing accept a PR for this.

mkArtakMSFT:
@javiercn can you please review the proposed PR? Thanks!

javiercn:
As discussed on the PR, we probably want to just use System.Text.Encoding.Web instead

PR #8759, opened 2019-03-23
Blazor: Decode URL segments in RouteContext

hexawyz:
Change `RouteContext`, to URL-decode every route segment. This happens before they are matched in `RouteEntry`, so both static route segments and route parameters are handled.

The current implementation of `TemplateParser` seems pretty relaxed and only splits on the character `/`, so unless we want to support route segments containing escaped `/` characters, the change on `RouteContext` should be enough.

Addresses #5510

### DIFF
--- a/src/Components/Components/src/Routing/RouteContext.cs
+++ b/src/Components/Components/src/Routing/RouteContext.cs
@@ -1,8 +1,9 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
 using System.Collections.Generic;
+using System.Net;
 
 namespace Microsoft.AspNetCore.Components.Routing
 {
@@ -15,6 +16,11 @@ namespace Microsoft.AspNetCore.Components.Routing
             // This is a simplification. We are assuming there are no paths like /a//b/. A proper routing
             // implementation would be more sophisticated.
             Segments = path.Trim('/').Split(Separator, StringSplitOptions.RemoveEmptyEntries);
+            // Individual segments are URL-decoded in order to support arbitrary characters, assuming UTF-8 encoding.
+            for (int i = 0; i < Segments.Length; i++)
+            {
+                Segments[i] = WebUtility.UrlDecode(Segments[i]);
+            }
         }
 
         public string[] Segments { get; }
--- a/src/Components/Components/src/Routing/RouteContext.cs
+++ b/src/Components/Components/src/Routing/RouteContext.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Net;
 
 namespace Microsoft.AspNetCore.Components.Routing
 {
@@ -19,7 +18,7 @@ namespace Microsoft.AspNetCore.Components.Routing
             // Individual segments are URL-decoded in order to support arbitrary characters, assuming UTF-8 encoding.
             for (int i = 0; i < Segments.Length; i++)
             {
-                Segments[i] = WebUtility.UrlDecode(Segments[i]);
+                Segments[i] = Uri.UnescapeDataString(Segments[i]);
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/Components/Components/src/Routing/RouteContext.cs
```diff
@@ -15,6 +16,11 @@ public RouteContext(string path)
             // This is a simplification. We are assuming there are no paths like /a//b/. A proper routing
             // implementation would be more sophisticated.
             Segments = path.Trim('/').Split(Separator, StringSplitOptions.RemoveEmptyEntries);
+            // Individual segments are URL-decoded in order to support arbitrary characters, assuming UTF-8 encoding.
+            for (int i = 0; i < Segments.Length; i++)
+            {
+                Segments[i] = WebUtility.UrlDecode(Segments[i]);
```
javiercn: @Tratcher This is for client side blazor. (Assume no ASP.NET Core)
Is this the best way to decode the segments? (vs `Uri.UnescapeDataString`.  or any other package with minimal dependencies we can bring in).

javiercn: I believe we should use https://www.nuget.org/packages/System.Text.Encodings.Web.

@SteveSandersonMS This will bring an additional package, but I think its important to use this one as it is the blessed one and other methods have nuances that we would need to understand (and would potentially introduce incompatibilities between server and client code)

SteveSandersonMS: @javiercn How much does it increase the default compressed post-linked app size?

javiercn: @SteveSandersonMS I would have to check if we decide to go this route, but the dll is around 54K uncompressed, so I think it'll be fine.

Independent of size, I think it's worth using it as it ensure compatibility with the server and is the blessed implementation.

Tratcher: Where is `path` coming from? What guarantees do we have about it?

javiercn: It’s the url from the browser as returned by window.location.href

SteveSandersonMS: @javiercn I totally agree that it's ideal to share this logic from the `System.Text.Encodings.Web` package. From what you're saying about the 54K uncompressed unlinked size, it sounds like we shouldn't have any problem with that. Nuget.org says it doesn't have any further dependencies.

javiercn: @SteveSandersonMS Let's switch to that then. We probably will see more gains when we link it.

GoldenCrystal: @javiercn Thanks, I wasn't aware of this package. I thought `WebUtility` was still the correct replacement for `HttpUtility`. 😅
I'll switch to this implementation and see how it works.

GoldenCrystal: It took me a bit of time to upgrade my setup and rebuild the whole solution… And now that I've looked at the contents of the package, it seems that it only provides methods for encoding. 😅

See the source here for reference : https://github.com/dotnet/corefx/tree/9e2f3b630a51875fc0455a7cabb8803f6527f81e/src/System.Text.Encodings.Web/src/System/Text/Encodings/Web

In that case, we should pick either `WebUtility.UrlDecode` or `Uri.UnescapeDataString`.
The main difference between `Uri.UnescapeDataString` and `WebUtility.UrlDecode`, would be that the former doesn't convert `+` into space (cf. [documentation](https://docs.microsoft.com/en-us/dotnet/api/system.uri.unescapedatastring)), while the latter does.

In this case, `Uri.UnescapeDataString` might indeed have been the better choice. A quick search in the repository reveals that it is used in quite a few places already: https://github.com/aspnet/AspNetCore/search?q=UnescapeDataString&unscoped_q=UnescapeDataString

If that's ok with you, I'll make that change instead.

Tratcher: UnescapeDataString should be fine for individual path segments.

mkArtakMSFT:
@javiercn have you looked at the latest update of this PR? What is pending here?

javiercn:
I think this can be merged.

javiercn:
Merged! Thanks for your contribution!

PR #8765, opened 2019-03-24
Make UseEndpoints a thing

rynowak:
This change makes the endpoint middleware explicit again, and updates
all of the templates.

The other change here is make UseEndpoints be the place where you
register endpoints. This is vital because it puts your code visually at
the point of the pipeline where it executes.

Lastly, I removed support for UseMvc with endpoint routing. This is
causing issues for some security features, and we're moving in the
direction of trying to make the middleware heavy implementation required
in 3.0. There are some issues we won't be able to fix in MVC if we can't
unambiguously know if UseMvc was used or the middleware.


src/Components/Server/src/Circuits/DefaultCircuitFactory.cs
```diff
@@ -100,7 +100,7 @@ private static IList<ComponentDescriptor> ResolveComponentMetadata(HttpContext h
                 {
                     throw new InvalidOperationException(
                         $"{nameof(ComponentHub)} doesn't have an associated endpoint. " +
-                        "Use 'app.UseRouting(routes => routes.MapComponentHub<App>(\"app\"))' to register your hub.");
+                        "Use 'app.UsedEndpoints(endpoints => endpoints.MapComponentHub<App>(\"app\"))' to register your hub.");
```
JamesNK: Components uses 100% recycled UsedEndpoints.

rynowak: You can clearly see that the new version is 100% green.






(I am proud of that joke)

rynowak:
@JamesNK

src/Components/test/testassets/TestServer/Startup.cs
```diff
@@ -38,24 +38,36 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
...
+                //     endpoints.MapComponentsHub<Index>();
+                //
+                // However it's expressed using routing as a way of checking that
+                // we're not relying on any extra magic inside MapComponentsHub, since it's
                 // important that people can set up these bits of middleware manually (e.g., to
                 // swap in UseAzureSignalR instead of UseSignalR).
-                subdirApp.UseRouting(routes =>
-                    routes.MapHub<ComponentHub>(ComponentHub.DefaultPath).AddComponent<Index>(selector: "root"));
+
+                subdirApp.UseRouting();
```
JamesNK: Is UseRouting required again here? re: UseRouting, line 42

src/Components/test/testassets/TestServer/Startup.cs
```diff
@@ -38,24 +38,36 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
...
+
+                subdirApp.UseEndpoints(endpoints =>
+                {
+                    endpoints.MapHub<ComponentHub>(ComponentHub.DefaultPath).AddComponent<Index>(selector: "root");
```
JamesNK: What are the rules for when there are multiple UseEndpoints? Is the scope that it is used in tracked, e.g. inside a Map, and the UseRouting for that scope gets routes from the correct UseEndpoints.

src/Http/Routing/src/Builder/EndpointRoutingApplicationBuilderExtensions.cs
```diff
@@ -95,5 +113,31 @@ private static void VerifyRoutingIsRegistered(IApplicationBuilder app)
...
+
+        private static void VerifyEndpointRoutingMiddlewareIsRegistered(IApplicationBuilder app, out DefaultEndpointRouteBuilder endpointRouteBuilder)
+        {
+            if (!app.Properties.TryGetValue(EndpointRouteBuilder, out var obj))
+            {
+                var message =
+                    $"{nameof(EndpointRoutingMiddleware)} must be added to the request execution pipeline before {nameof(EndpointMiddleware)}. " +
+                    $"Please add {nameof(EndpointRoutingMiddleware)} by calling '{nameof(IApplicationBuilder)}.{nameof(UseRouting)}' inside the call " +
+                    $"to 'Configure(...)' in the application startup code.";
+                throw new InvalidOperationException(message);
```
JamesNK: Throwing if UseRouting is not present when UseEndpoints is called is nice.

What do you think of expanding this message to explain why routing should be first? I think it is useful to teach the dev about what is going on at each step. On the other hand the message is quite long already.

JamesNK: What about the scenario when some middleware has set an endpoint without routing, and we just want to execute the set endpoint? This is what EndpointMiddleware use to do.

src/Identity/samples/IdentitySample.DefaultUI/Startup.cs
```diff
@@ -58,13 +58,15 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env, ILoggerF
...
             app.UseAuthentication();
+            app.UseAuthorization();
 
-            app.UseMvc(routes =>
+            app.UseEndpoints(endpoints =>
             {
-                routes.MapRoute(
-                    name: "default",
-                    template: "{controller=Home}/{action=Index}/{id?}");
+                endpoints.MapDefaultControllerRoute();
```
JamesNK: Random observation:
It feels *weirder* to mix the setting up of routes on an "endpoints" argument compared to a "routes" argument.

src/Identity/testassets/Identity.DefaultUI.WebSite/StartupBase.cs
```diff
@@ -77,9 +77,23 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
...
             app.UseCookiePolicy();
 
+            app.UseRouting();
+
             app.UseAuthentication();
 
-            app.UseMvc();
+            // This has to be disabled due to https://github.com/aspnet/AspNetCore/issues/8387
+            //
+            // UseAuthorization does not currently work with Razor pages, and it impacts
```
JamesNK: eep

src/Middleware/CORS/samples/SampleDestination/Startup.cs
```diff
@@ -57,7 +57,11 @@ public void ConfigureServices(IServiceCollection services)
 
         public void Configure(IApplicationBuilder app)
         {
-            app.UseRouting(routing =>
+            app.UseRouting();
+
+            app.UseCors();
+
+            app.UseEndpoints(routing =>
```
JamesNK: nit: routing -> endpoints

src/Mvc/Mvc.Core/src/Builder/MvcApplicationBuilderExtensions.cs
```diff
@@ -88,58 +84,23 @@ public static IApplicationBuilder UseMvcWithDefaultRoute(this IApplicationBuilde
...
-                        routerBuilder.DataSources.Add(dataSource);
-                    });
-                }
-
-                return app.UseEndpoint();
+                var message =
+                    "Endpoint Routing does not support 'IApplicationBuilder.UseMvc(...)'. To use " +
+                    "'IApplicationBuilder.UseMvc' set 'MvcOptions.EnableEndpointRouting = false' inside " +
+                    "'ConfigureServices(...).";
+                throw new InvalidOperationException(message);
```
JamesNK: Decisive! I like it

src/Components/test/testassets/TestServer/Startup.cs
JamesNK: Answered as I read through the rest of the PR 😄

src/Components/test/testassets/TestServer/Startup.cs
JamesNK: Answered as I read through the rest of the PR 😄

rynowak: Rule 1: Don't do it, it's kinda nonsense
Rule 2: Why would you do this?
Rule 3: We do our best make it do something reasonable 😁

src/Http/Routing/src/Builder/EndpointRoutingApplicationBuilderExtensions.cs
rynowak: > What do you think of expanding this message to explain why routing should be first? I think it is useful to teach the dev about what is going on at each step. On the other hand the message is quite long already.

I really don't personally think we can use exception messages to teach people things, other than that the current thing is wrong. If you have a suggestion of how to reword this or some specific language you want me to add, I'd be happy to try anyway. 


> What about the scenario when some middleware has set an endpoint without routing, and we just want to execute the set endpoint? This is what EndpointMiddleware use to do.

I don't want this to work. I don't want the endpoint middleware to be optional, or to show conflicting examples.

src/Identity/samples/IdentitySample.DefaultUI/Startup.cs
rynowak: I don't think this can/should change. The metaphor from MVC is *routes* yet routes are not a concept in endpoint routing. It's intentionally names this way so you can easily transition code.

Let the past be the past.

src/Identity/testassets/Identity.DefaultUI.WebSite/StartupBase.cs
rynowak: yep. bugs happen every way.

This was the point where I decided that `UseMvc` needed to stop working with ER.

src/Mvc/Mvc.Core/src/Builder/MvcApplicationBuilderExtensions.cs
rynowak: Thanks, me too!

rynowak:
@JamesNK any other super cool feedback

src/Http/Routing/src/Builder/EndpointRoutingApplicationBuilderExtensions.cs
JamesNK: > If you have a suggestion of how to reword this or some specific language you want me to add, I'd be happy to try anyway.

`$"{nameof(EndpointRoutingMiddleware)} matches routes setup by {nameof(EndpointMiddleware)} and so must be added to the request execution pipeline before {nameof(EndpointMiddleware)}."`

Not super happy with it but it explains a very little why we ask devs to do the things they must do.

JamesNK: > I don't want the endpoint middleware to be optional, or to show conflicting examples.

I get that, but what about if someone has middleware that sets an endpoint. There is no longer a `UseEndpoint` (implicit or explicit) to execute it. This change feels like endpoints are being constrained to always being used with routing.

rynowak: > But what about if someone has middleware that sets an endpoint. There is no longer a UseEndpoint (implicit or explicit) to execute it.

Do we have a scenario for this? I think making the endpoint middleware both implicit and explicit is a decision we will be explaining forever. Making it explicit is a response to feedback that people don't understand how the pipeline works when its implicit.

I'm happy to discuss whatever options you think are improvements. One idea I have would be to add an overload of `UseEndpoints()` that doesn't register any endpoints and thus doesn't require `UseRouter`. 

This all seems so marginal to me since this is trivial to write on your own. I can't think of a reason why we would document or recommend that as a pattern.

JamesNK: I kind of did something like that in this auth sample - https://github.com/aspnet/AspNetCore/pull/4279 - although the reason for setting the endpoints was metadata for the AuthZ middleware. The static file middleware would take care of executing the endpoints.

src/Components/Blazor/Templates/src/content/BlazorHosted-CSharp/BlazorHosted-CSharp.Server/Startup.cs
```diff
@@ -28,9 +28,14 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
                 app.UseBlazorDebugging();
             }
 
-            app.UseMvc(routes =>
+            app.UseRouting();
+
+            app.UseAuthentication();
+            app.UseAuthorization();
```
SteveSandersonMS: Why have `UseAuthentication`/`UseAuthorization` in this template by default? Developers would certainly have to take further steps before these would do anything useful, so they might as well add these themselves at the point where they are setting up auth.

src/Components/test/testassets/ComponentsApp.Server/Startup.cs
```diff
@@ -29,10 +29,13 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
             }
 
             app.UseStaticFiles();
-            app.UseRouting(builder =>
+
+            app.UseRouting();
+
+            app.UseEndpoints(endpoints =>
```
SteveSandersonMS: I'm sure this will be documented, but TBH it's not clear to me from looking at the code how `UseRouting` and `UseEndpoints` do different jobs. I can't guess why both are needed or what kinds of things go into each.

Not saying this is a problem, just that we'll need to have good clear explanations.

src/Http/Routing/src/Builder/EndpointRoutingApplicationBuilderExtensions.cs
rynowak: > The static file middleware would take care of executing the endpoint.

Right, so the point is that someone executes the endpoint and doesn't call next - then you never reach this condition.

src/Components/test/testassets/ComponentsApp.Server/Startup.cs
rynowak: yes, already writing the blog post.

src/Components/Blazor/Templates/src/content/BlazorHosted-CSharp/BlazorHosted-CSharp.Server/Startup.cs
rynowak: This might be a bad copy-paste on my part, I'm happy to take them out.

src/Http/Routing/src/Builder/EndpointRoutingApplicationBuilderExtensions.cs
JamesNK: Yes, in this example.

Basically my scenario is:
1. Middleware decides to send a response to the user and sets an endpoint rather than be terminal (maybe it wants more middleware to run after it)
2. Request gets to the end of the pipeline and the host throws an error saying their is an endpoint so the user must configure UseEndpoints
3. User adds UseEndpoints
4. Now ASP.NET Core throws an error saying the user must configure UseRouting, which user doesn't need/want

rynowak: Resolved this but updating the error message and adding a `UseEndpointExecutor()` method.

rynowak:
@JamesNK updated
@Tratcher - any feedback on this?
@ryanbrandenburg do templates look good to you?

Tratcher:
Nothing James hasn't already covered.

src/ProjectTemplates/Web.ProjectTemplates/content/RazorPagesWeb-CSharp/Startup.cs
```diff
@@ -158,20 +158,22 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
...
 #if (OrganizationalAuth || IndividualAuth)
             app.UseAuthentication();
 #endif
             app.UseAuthorization();
+
+            app.UseEndpoints(endpoints =>
+            {
+                endpoints.MapRazorPages();
+#if (IndividualB2CAuth || OrganizationalAuth)
+                routes.MapControllers();
```
ryanbrandenburg: Is it still routes? I would have thought it would be `endpoints.MapControllers();` now.

ryanbrandenburg (approved):
One major concern, but it should just be a one word fix if I'm right.

src/ProjectTemplates/Web.ProjectTemplates/content/RazorPagesWeb-CSharp/Startup.cs
rynowak: Yes, bad copy-pasto

src/Components/Server/src/Builder/ComponentEndpointRouteBuilderExtensions.cs
```diff
@@ -17,43 +17,43 @@ public static class ComponentEndpointRouteBuilderExtensions
         /// the component <typeparamref name="TComponent"/> to this hub instance as the given DOM <paramref name="selector"/>.
         /// </summary>
         /// <typeparam name="TComponent">The first <see cref="IComponent"/> associated with this <see cref="ComponentHub"/>.</typeparam>
-        /// <param name="routes">The <see cref="RouteBuilder"/>.</param>
+        /// <param name="endpoints">The <see cref="RouteBuilder"/>.</param>
```
JamesNK:
```suggestion
        /// <param name="endpoints">The <see cref="IEndpointRouteBuilder"/>.</param>
```

src/Components/Server/src/Builder/ComponentEndpointRouteBuilderExtensions.cs
```diff
@@ -17,43 +17,43 @@ public static class ComponentEndpointRouteBuilderExtensions
...
+            return endpoints.MapComponentHub(typeof(TComponent), selector, ComponentHub.DefaultPath);
         }
 
         /// <summary>
         /// Maps the SignalR <see cref="ComponentHub"/> to the path <paramref name="path"/> and associates
         /// the component <typeparamref name="TComponent"/> to this hub instance as the given DOM <paramref name="selector"/>.
         /// </summary>
         /// <typeparam name="TComponent">The first <see cref="IComponent"/> associated with this <see cref="ComponentHub"/>.</typeparam>
-        /// <param name="routes">The <see cref="RouteBuilder"/>.</param>
+        /// <param name="endpoints">The <see cref="RouteBuilder"/>.</param>
```
JamesNK:
```suggestion
        /// <param name="endpoints">The <see cref="IEndpointRouteBuilder"/>.</param>
```

src/Components/Server/src/Builder/ComponentEndpointRouteBuilderExtensions.cs
```diff
@@ -66,27 +66,27 @@ public static class ComponentEndpointRouteBuilderExtensions
...
-            return routes.MapComponentHub(typeof(TComponent), selector, path);
+            return endpoints.MapComponentHub(typeof(TComponent), selector, path);
         }
 
         /// <summary>
         /// Maps the SignalR <see cref="ComponentHub"/> to the path <paramref name="path"/> and associates
         /// the component <paramref name="componentType"/> to this hub instance as the given DOM <paramref name="selector"/>.
         /// </summary>
-        /// <param name="routes">The <see cref="RouteBuilder"/>.</param>
+        /// <param name="endpoints">The <see cref="RouteBuilder"/>.</param>
```
JamesNK:
```suggestion
        /// <param name="endpoints">The <see cref="IEndpointRouteBuilder"/>.</param>
```

JamesNK (approved):

PR #8771, opened 2019-03-24
ApiAuthorizationDbContext with types for user, role and other ASP.NET Identity objects

cjaliaga:
Matches all the options that IdentityDbContext offers to define the types of the ASP.NET Identity objects, such user, key, role, user claim, user role, user login, role claim and user token.


blowdart:
This is too big a breaking change. You can always do it manually yourself.

PR #8811, opened 2019-03-26
Fix #8520 - Component parameters not found

### LINKED ISSUE #8520: Default new Razor Components project loads empty page, then get `Failed to reconnect to the server.  [Retry?]` message.
### Describe the bug
Default new Razor Components project loads empty page, then eventually times out with `Failed to reconnect to the server.  [Retry?]` message.

### To Reproduce
1. Dev16 Preview 4.1
1. File/New Project/Razor Components
1. <kbd>F5</kbd>
1. Browser opens with empty page
1. Then eventually get `Failed to reconnect to the server.  [Retry?]` message:
image
1. Clicking Retry doesn't seem to do anything

### Expected behavior
A page should show up

### Additional context
`dotnet --info`:

```
C:\Users\elipton>dotnet --info
 Version:   3.0.100-preview4-010745
 Commit:    7b46750b5c

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.17763
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\3.0.100-preview4-010745\

Host (useful for support):
  Version: 3.0.0-preview4-27513-07
  Commit:  f9965e5a7e

.NET Core SDKs installed:
  2.1.504 [C:\Program Files\dotnet\sdk]
  2.1.600-preview-009472 [C:\Program Files\dotnet\sdk]
  2.1.600-preview-009497 [C:\Program Files\dotnet\sdk]
  2.1.600 [C:\Program Files\dotnet\sdk]
  2.2.200-preview-009748 [C:\Program Files\dotnet\sdk]
  2.2.200-preview-009804 [C:\Program Files\dotnet\sdk]
  2.2.200 [C:\Program Files\dotnet\sdk]
  3.0.100-preview4-010745 [C:\Program Files\dotnet\sdk]

.NET Core runtimes installed:
  <trimmed>
  Microsoft.AspNetCore.App 2.2.2 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.0.0-preview4-19162-05 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 2.1.7 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.8 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.2.1 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.2.2 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 3.0.0-preview4-27513-07 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.WindowsDesktop.App 3.0.0-preview4-27511-44 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
```

And here's what's relevant in the CSPROJ:

```xml
  <ItemGroup>
    <PackageReference Include="Microsoft.AspNetCore.Components.Server" Version="3.0.0-preview4-19162-05" />
    <PackageReference Include="Microsoft.AspNetCore.Mvc.NewtonsoftJson" Version="3.0.0-preview4-19162-05" />
  </ItemGroup>
```

rynowak:
This resolves the issue blocking use of component parameters from our
ref assemblies. Making properties public with private get is our
recommended guidance for wanting documentation to work in the IDE, and
it will be required when you ship ref assemblies - since ref assemblies
only contain public members.

### DIFF
--- a/eng/GenAPI.exclusions.txt
+++ b/eng/GenAPI.exclusions.txt
@@ -3,3 +3,19 @@ T:Microsoft.AspNetCore.Components.RenderTree.RenderTreeFrame
 # Manually implemented - https://github.com/dotnet/arcade/issues/2066
 T:Microsoft.AspNetCore.Mvc.ApplicationModels.PageParameterModel
 T:Microsoft.AspNetCore.Mvc.ApplicationModels.PagePropertyModel
+# Manually implemented - https://github.com/aspnet/AspNetCore/issues/8825
+T:Microsoft.AspNetCore.Components.CascadingValue`1
+T:Microsoft.AspNetCore.Components.Forms.DataAnnotationsValidator
+T:Microsoft.AspNetCore.Components.Forms.EditForm
+T:Microsoft.AspNetCore.Components.Forms.InputBase`1
+T:Microsoft.AspNetCore.Components.Forms.InputCheckbox
+T:Microsoft.AspNetCore.Components.Forms.InputDate`1
+T:Microsoft.AspNetCore.Components.Forms.InputNumber`1
+T:Microsoft.AspNetCore.Components.Forms.InputSelect`1
+T:Microsoft.AspNetCore.Components.Forms.InputText
+T:Microsoft.AspNetCore.Components.Forms.InputTextArea
+T:Microsoft.AspNetCore.Components.Forms.ValidationMessage`1
+T:Microsoft.AspNetCore.Components.Forms.ValidationSummary
+T:Microsoft.AspNetCore.Components.Layouts.LayoutDisplay
+T:Microsoft.AspNetCore.Components.Routing.NavLink
+T:Microsoft.AspNetCore.Components.Routing.Router
--- a/src/Components/Components/ref/Microsoft.AspNetCore.Components.netstandard2.0.Manual.cs
+++ b/src/Components/Components/ref/Microsoft.AspNetCore.Components.netstandard2.0.Manual.cs
@@ -45,3 +45,183 @@ namespace Microsoft.AspNetCore.Components.RenderTree
         public override string ToString() { throw null; }
     }
 }
+
+// Built-in components: https://github.com/aspnet/AspNetCore/issues/8825
+namespace Microsoft.AspNetCore.Components
+{
+    public partial class CascadingValue<T> : Microsoft.AspNetCore.Components.IComponent
+    {
+        public CascadingValue() { }
+        [Microsoft.AspNetCore.Components.ParameterAttribute]
+        public Microsoft.AspNetCore.Components.RenderFragment ChildContent { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } private set { throw null; }}
+        [Microsoft.AspNetCore.Components.ParameterAttribute]
+        public bool IsFixed { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } private set { throw null; }}
+        [Microsoft.AspNetCore.Components.ParameterAttribute]
+        public string Name { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } private set { throw null; }}
+        [Microsoft.AspNetCore.Components.ParameterAttribute]
+        public T Value { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } private set { throw null; }}
+        public void Configure(Microsoft.AspNetCore.Components.RenderHandle renderHandle) { }
+        public System.Threading.Tasks.Task SetParametersAsync(Microsoft.AspNetCore.Components.ParameterCollection parameters) { throw null; }
+    }
+}
+
+namespace Microsoft.AspNetCore.Components.Forms
+{
+    public partial class DataAnnotationsValidator : Microsoft.AspNetCore.Components.ComponentBase
+    {
+        public DataAnnotationsValidator() { }
+        protected override void OnInit() { }
+    }
+
+    public partial class EditForm : Microsoft.AspNetCore.Components.ComponentBase
+    {
+        public EditForm() { }
+        [Microsoft.AspNetCore.Components.ParameterAttribute]
+        public Microsoft.AspNetCore.Components.RenderFragment<Microsoft.AspNetCore.Components.Forms.EditContext> ChildContent { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } private set { throw null; }}
+        [Microsoft.AspNetCore.Components.ParameterAttribute]
+        public Microsoft.AspNetCore.Components.Forms.EditContext EditContext { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } private set { throw null; }}
+        [Microsoft.AspNetCore.Components.ParameterAttribute]
+        public object Model { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } private set { throw null; }}
+        [Microsoft.AspNetCore.Components.ParameterAttribute]
+        public Microsoft.AspNetCore.Components.EventCallback<Microsoft.AspNetCore.Components.Forms.EditContext> OnInvalidSubmit { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } private set { throw null; }}
+        [Microsoft.AspNetCore.Components.ParameterAttribute]
+        public Microsoft.AspNetCore.Components.EventCallback<Microsoft.AspNetCore.Components.Forms.EditContext> OnSubmit { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } private set { throw null; }}
+        [Microsoft.AspNetCore.Components.ParameterAttribute]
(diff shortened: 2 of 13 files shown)


rynowak:
Hmm looking at the ref assembly output, I'm not sure this will fix the problem. The setters aren't generated in ref assemblies.

SteveSandersonMS:
OK, this is inconvenient. It seems that `internal set` is also insufficient.

Maybe the correct fix is for us to change our parameter detection logic so we don't care whether there's a setter or not. As long as something is a property annotated with `[Parameter]`, treat it as a parameter (and let it fail at runtime if there truly isn't a setter). That way it's OK for the reference assemblies not to have setters for those parameters.

The original reason for checking for the existence of `set` was because we didn't originally have `[Parameter]`. Now we do, the developer's intention is pretty clear.

**Potential further improvement:** We have an analyzer that currently errors if you make the setter public. We could change it so that it also errors if you have no setter at all, so nobody would get confused. This wouldn't affect the ref assemblies because we're not running the roslyn analyzer on those.

rynowak:
> Potential further improvement: We have an analyzer that currently errors if you make the setter public. We could change it so that it also errors if you have no setter at all, so nobody would get confused. This wouldn't affect the ref assemblies because we're not running the roslyn analyzer on those.

Or we would make the properties public set/get and add an analyzer to tell you not to set them. 

I've updated this PR to include the setters in the ref assembly. Given that our nightly builds are broken (and have been for weeks) I don't really want to deliberate on the right solution until its working again. I think that this is an awful workaround, and I've logged https://github.com/aspnet/AspNetCore/issues/8825 to follow up on this issue.

src/Components/Components/ref/Microsoft.AspNetCore.Components.netstandard2.0.Manual.cs
```diff
@@ -44,4 +44,146 @@ namespace Microsoft.AspNetCore.Components.RenderTree
...
+}
+
+// Built-in components: https://github.com/aspnet/AspNetCore/issues/8825
+namespace Microsoft.AspNetCore.Components
+{
+    public partial class CascadingValue<T> : Microsoft.AspNetCore.Components.IComponent
+    {
+        public CascadingValue() { }
+        [Microsoft.AspNetCore.Components.ParameterAttribute]
+        public Microsoft.AspNetCore.Components.RenderFragment ChildContent { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } private set { throw null; }}
```
rynowak: adding the `private set { throw null; }` is what requires us to do these manually. Tooling and runtime ignore visibility and requires a setter.

src/Components/Components/ref/Microsoft.AspNetCore.Components.netstandard2.0.cs
```diff
@@ -702,19 +714,25 @@ public partial class InputCheckbox : Microsoft.AspNetCore.Components.Forms.Input
     public partial class InputDate<T> : Microsoft.AspNetCore.Components.Forms.InputBase<T>
     {
         public InputDate() { }
+        [Microsoft.AspNetCore.Components.ParameterAttribute]
+        public string ParsingErrorMessage { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
```
rynowak: Something is still not right here. I expected to see this type removed.

rynowak:
@SteveSandersonMS @pakrym

pakrym:
LGTM

src/Components/Components/ref/Microsoft.AspNetCore.Components.netstandard2.0.Manual.cs
```diff
@@ -44,4 +44,149 @@ namespace Microsoft.AspNetCore.Components.RenderTree
...
+    }
+
+    public partial class ValidationMessage<T> : Microsoft.AspNetCore.Components.ComponentBase, System.IDisposable
+    {
+        public ValidationMessage() { }
+        [Microsoft.AspNetCore.Components.ParameterAttribute]
+        public System.Linq.Expressions.Expression<System.Func<T>> For { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } private set { throw null; }}
+        protected override void BuildRenderTree(Microsoft.AspNetCore.Components.RenderTree.RenderTreeBuilder builder) { }
+        protected override void OnParametersSet() { }
+        void System.IDisposable.Dispose() { }
```
SteveSandersonMS: Could you clarify why only some of the InputBase subclasses are here? Why InputNumber but not InputText?

rynowak: Those types don't have any parameters.

src/Components/Components/ref/Microsoft.AspNetCore.Components.netstandard2.0.cs
```diff
@@ -676,55 +664,19 @@ public sealed partial class FieldChangedEventArgs
...
-        protected override bool TryParseValueFromString(string value, out T result, out string validationErrorMessage) { throw null; }
-    }
-    public partial class InputSelect<T> : Microsoft.AspNetCore.Components.Forms.InputBase<T>
-    {
-        public InputSelect() { }
-        protected override void BuildRenderTree(Microsoft.AspNetCore.Components.RenderTree.RenderTreeBuilder builder) { }
-        protected override bool TryParseValueFromString(string value, out T result, out string validationErrorMessage) { throw null; }
-    }
-    public partial class InputText : Microsoft.AspNetCore.Components.Forms.InputBase<string>
+    public partial class InputText : Microsoft.AspNetCore.Components.ComponentBase
```
SteveSandersonMS: Sorry to still be asking questions, but is it really OK to declare `InputText` as inheriting directly from `ComponentBase`, when it really inherits from `InputBase<string>`?

Is there any risk that this will make the tooling think there are no `Id` or `Class` parameters for `InputText`?

src/Components/Components/ref/Microsoft.AspNetCore.Components.netstandard2.0.Manual.cs
SteveSandersonMS: Makes sense. Thanks.

src/Components/Components/ref/Microsoft.AspNetCore.Components.netstandard2.0.cs
rynowak: whoa, I'm glad you caught this. This looks like a bug in the ref assembly generation, and is not intended.

rynowak: I will move these types to be manual as well and report this issue.

rynowak: https://github.com/dotnet/arcade/issues/2345

eng/GenAPI.exclusions.txt
```diff
@@ -2,4 +2,19 @@
...
+T:Microsoft.AspNetCore.Components.Forms.InputDate`1
+T:Microsoft.AspNetCore.Components.Forms.InputNumber`1
+T:Microsoft.AspNetCore.Components.Forms.InputSelect`1
+T:Microsoft.AspNetCore.Components.Forms.InputText
+T:Microsoft.AspNetCore.Components.Forms.InputTextArea
+T:Microsoft.AspNetCore.Components.Forms.ValidationMessage`1
+T:Microsoft.AspNetCore.Components.Forms.ValidationSummary
+T:Microsoft.AspNetCore.Components.Layouts.LayoutDisplay
+T:Microsoft.AspNetCore.Components.Routing.NavLink
+T:Microsoft.AspNetCore.Components.Routing.Router
```
rynowak: All of our built-in components are now manual. I think this is the best thing to do since it's being spooky.

SteveSandersonMS: Looks like we're missing `InputCheckbox` from here

rynowak: should be fixed now

SteveSandersonMS (approved):
Great!

dougbu (approved):
I'd like to understand why the project's properties are insufficient and when GenAPI.exclusions.txt changes will be needed going forward. But, that shouldn't block this PR&hellip;

rynowak:
@aspnet/BRT

rynowak:
>I'd like to understand why the project's properties are insufficient and when GenAPI.exclusions.txt changes will be needed going forward. But, that shouldn't block this PR…

I'd be happy to chat about this in more detail in the office. Ultimately we don't want to maintain this long term, but we need a solution right now.

rynowak:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=134101

aspnet-hello:

I've triaged the above build.

PR #8818, opened 2019-03-26
Fix rendering inheritance. Fixes #8192

### LINKED ISSUE #8192: Subclassing of Form components doesn't work well
I tried subclassing EditNumber to implement something similar to this

It doesn't work because the genered code from a razor file calls` base.BuildRenderTree` - you get the rendering of the base class and what you wrote.

image

We should be able to fix this by making it unnecessary for a component to call `base.BuildRenderTree`. In general, it's a good idea if we can avoid extensibility points where an inheritor has to call `base`

SteveSandersonMS:
With this PR, it no longer matters whether a `ComponentBase` component's rendering logic calls `base.BuildRenderTree` or not.

We need to merge this change *before* we change the Razor Components compiler so that it no longer generates the `base.BuildRenderTree` calls.

### DIFF
--- a/src/Components/Components/src/ComponentBase.cs
+++ b/src/Components/Components/src/ComponentBase.cs
@@ -25,11 +25,6 @@ namespace Microsoft.AspNetCore.Components
     /// </summary>
     public abstract class ComponentBase : IComponent, IHandleEvent, IHandleAfterRender
     {
-        /// <summary>
-        /// Specifies the name of the <see cref="RenderTree"/>-building method.
-        /// </summary>
-        public const string BuildRenderTreeMethodName = nameof(BuildRenderTree);
-
         private readonly RenderFragment _renderFragment;
         private RenderHandle _renderHandle;
         private bool _initialized;
@@ -41,7 +36,12 @@ namespace Microsoft.AspNetCore.Components
         /// </summary>
         public ComponentBase()
         {
-            _renderFragment = BuildRenderTree;
+            _renderFragment = builder =>
+            {
+                _hasPendingQueuedRender = false;
+                _hasNeverRendered = false;
+                BuildRenderTree(builder);
+            };
         }
 
         /// <summary>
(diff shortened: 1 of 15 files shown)


src/Components/Components/src/ComponentBase.cs
```diff
@@ -25,11 +25,6 @@ namespace Microsoft.AspNetCore.Components
     /// </summary>
     public abstract class ComponentBase : IComponent, IHandleEvent, IHandleAfterRender
     {
-        /// <summary>
-        /// Specifies the name of the <see cref="RenderTree"/>-building method.
-        /// </summary>
-        public const string BuildRenderTreeMethodName = nameof(BuildRenderTree);
```
SteveSandersonMS: This was unused, as far as I can tell.

src/Components/Components/src/ComponentBase.cs
```diff
@@ -52,8 +52,9 @@ protected virtual void BuildRenderTree(RenderTreeBuilder builder)
         {
```
SteveSandersonMS: Technically we could make `BuildRenderTree` into an `abstract` method now if we wanted. I don't personally feel strongly about doing so, since the built-in empty implementation is a legal default that you might even want to use in some scenarios (e.g., for components that don't produce any visible output, but exist only to cause side-effects on inclusion/disposal).

rynowak: I think virtual is fine as-is.

src/Components/Components/src/ComponentBase.cs
rynowak: yep. this was here is a very early iteration to share with the compiler

rynowak (approved):

SteveSandersonMS:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=134858

Hi BRT. I've been trying to get this PR merged all day but the Ubuntu job is in a bad way. One time it timed out, the other it just failed for unrelated reasons.

aspnet-hello:

I've triaged the above build.

ryanbrandenburg:
@SteveSandersonMS doing "@aspnet-hello triage" does not summon the BRT. That's @aspnet/brt. I only noticed this because it violated some naive assumptions RaaS was making about how such comments would function.

ryanbrandenburg:
@aspnet-hello triage https://dnceng.visualstudio.com/public/_build/results?buildId=134970

aspnet-hello:

I've triaged the above build.

SteveSandersonMS:
@ryanbrandenburg Thanks for clarifying.

@aspnet/brt, then :)

ryanbrandenburg:
https://github.com/aspnet/AspNetCore-Internal/issues/2090 was filed about the test failures, @anurse has his minions on the case.

PR #8827, opened 2019-03-26
SignInManager should respect IdentityOptions

cleftheris:
GetExternalLoginInfoAsync should not be using `ClaimTypes.NameIdentifier` directly. There is an option to configure this via `Options.ClaimsIdentity.UserIdClaim`


HaoK:
That setting is referring to what claim will get generated in the output claims identity that identity generates for the user, the external login identity is something different and is not configurable as its a contract with the auth providers

PR #8830, opened 2019-03-26
Make building Components & Mvc solutions work out of the box

pranavkm:
Fixes https://github.com/aspnet/AspNetCore/issues/8413

### DIFF
--- a/src/Components/Browser.JS/Microsoft.AspNetCore.Components.Browser.JS.npmproj
+++ b/src/Components/Browser.JS/Microsoft.AspNetCore.Components.Browser.JS.npmproj
@@ -12,10 +12,18 @@
 
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), Directory.Build.targets))\Directory.Build.targets" />
 
+  <!--
+    Invoking build[.cmd|.sh] does not restore npm projects by default. This is a performance optimization.
+    However, we do no want to ensure yarn's restored before a P2P build happens. We'll conservatively yarn install
+    if the node_modules directory is missing.
+  -->
+  <Target Name="FastRestore" DependsOnTargets="Restore" Condition="!Exists('node_modules')" />
+
   <!-- Override the default 'Build' target from eng/targets/Npm.Common.targets. -->
   <Target Name="Build"
+          DependsOnTargets="FastRestore"
           Inputs="@(WebpackInputs)"
-          Outputs="dist\components.webassembly.js;dist\components.server.js">
+          Outputs="src\dist\components.webassembly.js;src\dist\components.server.js">
     <RemoveDir Directories="dist" />
     <Yarn Command="run build:debug" Condition="'$(Configuration)' == 'Debug'" />
     <Yarn Command="run build:production" Condition="'$(Configuration)' != 'Debug'" />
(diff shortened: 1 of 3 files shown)


src/Components/Browser.JS/Microsoft.AspNetCore.Components.Browser.JS.npmproj
```diff
@@ -12,10 +12,18 @@
...
+    if the node_modules directory is missing.
+  -->
+  <Target Name="FastRestore" DependsOnTargets="Restore" Condition="!Exists('node_modules')" />
+
   <!-- Override the default 'Build' target from eng/targets/Npm.Common.targets. -->
   <Target Name="Build"
+          DependsOnTargets="FastRestore"
           Inputs="@(WebpackInputs)"
-          Outputs="dist\components.webassembly.js;dist\components.server.js">
+          Outputs="src\dist\components.webassembly.js;src\dist\components.server.js">
```
rynowak: Consider moving this to an item group

src/Components/Browser.JS/Microsoft.AspNetCore.Components.Browser.JS.npmproj
```diff
@@ -12,10 +12,18 @@
 
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), Directory.Build.targets))\Directory.Build.targets" />
 
+  <!--
+    Invoking build[.cmd|.sh] does not restore npm projects by default. This is a performance optimization.
+    However, we do no want to ensure yarn's restored before a P2P build happens. We'll conservatively yarn install
```
pranavkm: we do ~~not~~ want to

rynowak (approved):

src/Components/Browser.JS/Microsoft.AspNetCore.Components.Browser.JS.npmproj
```diff
@@ -8,14 +8,24 @@
 
   <ItemGroup>
     <WebpackInputs Include="**\*.ts" Exclude="node_modules\**" />
+    <WebPackOutputs Include="src\dist\components.webassembly.js" />
+    <WebPackOutputs Include="src\dist\components.server.js" />
```
SteveSandersonMS: Nit: for consistency, `WebpackOutputs `

SteveSandersonMS:
I'd love to understand a bit more about this idea that it's legitimate to build Components.Server with `BuildNodeJS==false`. As far as I can tell, that just means we're making a broken build, since Components.Server can't function correctly without the compiled .js files.

If we do need to have this concept of `BuildNodeJS==false`, shouldn't we be excluding things that depend on Node-build resources (such as Components.Server) completely, instead of building a malformed and unfunctional version of them?

SteveSandersonMS:
To be clear, I'm not questing the correctness of your changes here, @pranavkm! This looks totally right within the scope of what we already do. I'm more wondering if, since we're changing this anyway, we have the opportunity to change it more completely so it doesn't do this weird thing any more.

src/Components/Browser.JS/Microsoft.AspNetCore.Components.Browser.JS.npmproj
pranavkm: I'll do it in a followup. I'd rather not wait another day to fix up casing 😄

pranavkm:
> I'd love to understand a bit more about this idea that it's legitimate to build Components.Server with BuildNodeJS==false. As far as I can tell, that just means we're making a broken build, since Components.Server can't function correctly without the compiled .js files.

The flag is set for the linux-musl build. I think the image is missing nodejs and yarn. It's not an immediate problem since builds for the Components.Server package come out from our Windows build, and the musl build runs does not run any tests.

PR #8837, opened 2019-03-26
Limit chunk size when writing chunked responses

### LINKED ISSUE #8815: WEB API Controller method return value causes net::ERR_INCOMPLETE_CHUNKED_ENCODING
### Describe the bug
When we moved to .NET Core 3 controllers stopped working properly. When method from controller returns data (e.g string) of size >= 4099 it causes net::ERR_INCOMPLETE_CHUNKED_ENCODING in chrome.
Tried on preview 3 and preview 4, on .NET Core 2.2 everything was good.
No issues when method returns data <=4098.

### To Reproduce
Method of controller causing this behavior:

```
[HttpGet("TestControllerResponse")]
public IActionResult TestControllerResponse()
{
     string result = "";
     var i = 0;
     //var size = 4098;    // 4098 -> OK
     var size = 4099;    // 4099 -> net::ERR_INCOMPLETE_CHUNKED_ENCODING
     while (++i <= size)
     {
          result = $"{result}+";
     }
     return Ok(result);
}
```

### Expected behavior
As in previous versions, this should be OK, not causing net::ERR_INCOMPLETE_CHUNKED_ENCODING.

### Details from uncaught exceptions:

> fail: Microsoft.AspNetCore.Server.Kestrel[13]
>       Connection id "0HLLHQ2L5SQMD", Request id "0HLLHQ2L5SQMD:0000000E": An unhandled exception was thrown by the application.
> System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values.
>    at System.Memory`1.Slice(Int32 start, Int32 length)
>    at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.Http1OutputProducer.GetChunkedMemory(Int32 sizeHint)
>    at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.Http1OutputProducer.GetSpan(Int32 sizeHint)
>    at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpProtocol.GetSpan(Int32 sizeHint)
>    at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpResponsePipeWriter.GetSpan(Int32 sizeHint)
>    at Microsoft.AspNetCore.Http.HttpResponseWritingExtensions.WriteMultiSegmentEncoded(PipeWriter writer, String text, Encoding encoding, Span`1 destination, Int32 encodedLength)
>    at Microsoft.AspNetCore.Http.HttpResponseWritingExtensions.Write(HttpResponseresponse, String text, Encoding encoding)
>    at Microsoft.AspNetCore.Http.HttpResponseWritingExtensions.WriteAsync(HttpResponse response, String text, Encoding encoding, CancellationToken cancellationToken)
>    at Microsoft.AspNetCore.Mvc.Formatters.StringOutputFormatter.WriteResponseBodyAsync(OutputFormatterWriteContext context, Encoding encoding)
>    at Microsoft.AspNetCore.Mvc.Formatters.TextOutputFormatter.WriteAsync(OutputFormatterWriteContext context)
>    at Microsoft.AspNetCore.Mvc.Infrastructure.ObjectResultExecutor.ExecuteAsync(ActionContext context, ObjectResult result)
>    at Microsoft.AspNetCore.Mvc.ObjectResult.ExecuteResultAsync(ActionContext context)
>    at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.InvokeResultAsync(IActionResult result)
>    at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.InvokeNextResultFilterAsync[TFilter,TFilterAsync]()
>    at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.Rethrow(ResultExecutedContext context)
>    at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.ResultNext[TFilter,TFilterAsync](State& next, Scope& scope, Object& state, Boolean& isCompleted)
>    at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.InvokeResultFilters()
>    at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.InvokeNextResourceFilter()
>    at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.Rethrow(ResourceExecutedContext context)
>    at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.Next(State& next,Scope& scope, Object& state, Boolean& isCompleted)
>    at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.InvokeFilterPipelineAsync()
>    at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.InvokeAsync()
>    at Microsoft.AspNetCore.Routing.EndpointMiddleware.Invoke(HttpContext httpContext)
>    at Microsoft.AspNetCore.Routing.EndpointRoutingMiddleware.Invoke(HttpContexthttpContext)
>    at Microsoft.AspNetCore.Builder.Extensions.MapMiddleware.Invoke(HttpContext context)
>    at Microsoft.AspNetCore.StaticFiles.StaticFileMiddleware.Invoke(HttpContext context)
>    at Microsoft.AspNetCore.Authentication.AuthenticationMiddleware.Invoke(HttpContext context)
>    at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpProtocol.ProcessRequests[TContext](IHttpApplication`1 application)

jkotalik:
Fixes https://github.com/aspnet/AspNetCore/issues/8815.

I'm pretty shocked a bug like this didn't show up anywhere across our stack. To repro this, someone simply needs to call HttpResponse.WriteAsync(string larger than 4096 bytes) when writing a chunked response.

The origin of this issue was this change in corefx: https://github.com/dotnet/corefx/commit/c45d88d5a1326501e31c02e508efb352a492fb59#diff-9a0b343d0ae95ef90f79dcc5a297c613. We made an assumption that if someone calls GetMemory on a PipeWriter, it will never return more memory than what the max buffer size of the MemoryPool is. This was changed by that PR. The issue should have been discovered here: https://github.com/aspnet/AspNetCore/pull/7671/files where we decided to remove the sizeHint check on all of these calls. It was a red flag that GetMemory was suddenly returning more memory than we expected. 

There are two approaches for a fix here:
1. Make it so sizeHint is limited to 4096 before calling GetMemory on the pipe (which is what I did here). 
2. Make it so the Http1OutputProducer works with chunks larger than 4096.

I went with the first approach for now for simplicity. We can consider the second if we see value in it.


src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
```diff
@@ -541,9 +541,13 @@ public void Reset()
         // These methods are for chunked http responses that use GetMemory/Advance
         private Memory<byte> GetChunkedMemory(int sizeHint)
         {
-            // The max size of a chunk will be the size of memory returned from the PipeWriter (today 4096)
-            // minus 5 for the max chunked prefix size and minus 2 for the chunked ending, leaving a total of
-            // 4089.
+            // Calling GetMemory on the PipeWriter can return a size larger than 4096
+            // if the sizeHint is greater.
+            // The max size we allow for chunking is 4096.
+            if (sizeHint > _memoryPool.MaxBufferSize)
```
davidfowl: Hmm given the new behavior of GetMemory, should we use a heap buffer if it’s bigger than the pool size?

jkotalik: That's the thing I'm contemplating. My list above states that either are valid options, just the second one requires more work.

halter73: So the problem with this change is that the HTTP/1 response PipeWriter once again return less than sizeHint bytes from GetMemory even though the new contract is that PipeWriters always return at least sizeHint bytes? 

If so, I can see how proper fix could be a lot more involved than this PR as is. We probably still don't want to send up to 2GB chunks, so it could require creating multiple chunks (and the doing the requisite copying) during a single call to Advance(bytes).

That was the nice thing about sizeHint being a hint instead of a minimum: If the PipeWriter implementation knows it's going to have to chunk the data anyway, it can induce the code using the PipeWriter to do the chunking in a way that will likely be more efficient.

jkotalik: To confirm before I try to prototype it, should we make it so the sizeHint can be greater than 4096?

halter73: I would say so. Let's wait to hear what @davidfowl has to say though.

davidfowl: Yes we have to respect the contract and make it a minimum. We should just change all of the calling code (like IBufferWriter\<byte\>.Write) to call GetMemory with a smaller size buffer (like 4K) to avoid this happening in really exceptional cases. If the caller request 2GB then just throw OOM (or let the array construction do that)

jkotalik: Or just calling GetMemory without a sizeHint.

halter73: >  If the caller request 2GB then just throw OOM (or let the array construction do that)

The bigger concern is when array construction *doesn't* OOM. There are likely at least some clients/proxies that limit received HTTP chunks to a max size of somewhere between 4KB and 2GB. That's why I'm concerned about making a change that allows for greater-than-4KB chunks for the first time.

Example: https://github.com/softwaremill/sttp/issues/36

davidfowl: Then we can throw ourselves. It should be a setting though

halter73: > It should be a setting though

👍

jkotalik: > The bigger concern is when array construction doesn't OOM. There are likely at least some clients/proxies that limit received HTTP chunks to a max size of somewhere between 4KB and 2GB. That's why I'm concerned about making a change that allows for greater-than-4KB chunks for the first time.

I'm fairly certain this 4KB restriction is only there for the PipeWriter scenario. If someone called Response.WriteAsync in the past, I'm fairly sure it would write a single chunk for that write. Ex if someone does `ctx.Response.Body.WriteAsync(2gb string)`, we would write that chunk to the response still.

If we want to throw/limit chunk size, we should do it universally.

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
```diff
@@ -541,58 +543,57 @@ public void Reset()
...
 
-            Debug.Assert(actualMemory.Length <= 4089);
-
             return actualMemory;
         }
 
+        private void UpdateCurrentChunkMemory(int sizeHint)
+        {
+            _currentChunkMemory = _pipeWriter.GetMemory(sizeHint);
+            _currentMemoryPrefixBytes = _currentChunkMemory.Length.ToString("X").Length + BeginChunkFixedLength;
```
davidfowl: This is allocating a string to compute the hex length? Nah

jkotalik:
I wish I can mark a PR as draft retroactively...

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

src/Servers/Kestrel/Core/src/Internal/Http/ChunkWriter.cs
```diff
@@ -44,6 +36,18 @@ public static int BeginChunkBytes(int dataCount, Span<byte> span)
             return count;
         }
 
+        internal static int GetBeginChunkByteCount(int count, out int total, out int shift)
+        {
```
jkotalik: I'm not a fan of how I refactored this, but couldn't find a better way. Thoughts?

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
jkotalik: 🆙 📅 without throwing for OOM.

src/Servers/Kestrel/Core/src/Internal/Http/ChunkWriter.cs
```diff
@@ -16,16 +16,8 @@ internal static class ChunkWriter
...
-            total = (count > 0xffff) ? 0x10 : 0x00;
-            count >>= total;
-            shift = (count > 0x00ff) ? 0x08 : 0x00;
-            count >>= shift;
-            total |= shift;
-            total |= (count > 0x000f) ? 0x04 : 0x00;
-
-            count = (total >> 2) + 3;
+            int total, shift; 
+            var count = GetBeginChunkByteCount(dataCount, out total, out shift);
```
halter73: It doesn't look like the shift out param is actually being used given that it get's reassigned at the beginning of the for loop:

```C#
for (shift = total; shift >= 0; shift -= 4)
```

src/Servers/Kestrel/Core/src/Internal/Http/ChunkWriter.cs
halter73: At a minimum, I would keep out param method private and have the internal version of the method without the out param.

Have you verified GetBeginChunkByteCount works for the all ints between 1 and int.MaxValue? Because it only had to work for ints up to 4096 before, right? It might be worth adding unit tests for both this and BeginChunkBytes.

src/Servers/Kestrel/Core/src/Internal/Http/ChunkWriter.cs
jkotalik: The only reason I kept it was to reuse the int.

src/Servers/Kestrel/Core/src/Internal/Http/ChunkWriter.cs
jkotalik: No, this had to work for all values up to Int.MaxValue. Tests are here: https://github.com/aspnet/AspNetCore/blob/master/src/Servers/Kestrel/Core/test/ChunkWriterTests.cs#L39

jkotalik: I'm more concerned about the logic I added to Http1OutputProducer with regards to all values.

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
```diff
@@ -538,57 +539,67 @@ public void Reset()
...
         private Memory<byte> GetChunkedMemory(int sizeHint)
         {
-            // The max size of a chunk will be the size of memory returned from the PipeWriter (today 4096)
-            // minus 5 for the max chunked prefix size and minus 2 for the chunked ending, leaving a total of
-            // 4089.
+            var originalSizeHint = sizeHint;
+
+            // To guarantee that sizeHint is a minimumSize, we need to call GetMemory with a sizeHint
+            // larger than what was provided. We add the begin and end chunk lengths here.
+            if (sizeHint > 0)
```
halter73: What if the sizeHint is 0, and there is some space remaining in _currentChunkMemory, but not enough space for a chunk prefix/suffix plus real data?

I have the same question for when sizeHint is 0, there's no space remaining in _currentChunkMemory, and _pipeWriter.GetMemory(0) does not return enough space for a chunk?

src/Servers/Kestrel/Core/src/Internal/Http/ChunkWriter.cs
halter73: I wouldn't worry about it unless there's some evidence that we have so many locals that they're spilling out of registers.

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
jkotalik: > What if the sizeHint is 0, and there is some space remaining in _currentChunkMemory, but not enough space for a chunk prefix/suffix plus real data?

The big thing to remember is when someone calls GetMemory and there is space remaining, we don't need to reinsert the chunk prefix/suffix. Here is an example.

Let's say we call GetMemory(0). This will call GetMemory on the PipeWriter, returning a chunk of 4096. We allocate space in the chunk like this:
[2 bytes chunked prefix | 3 bytes chunk length | 4089 bytes chunk data | 2 bytes chunked suffix]

When someone calls Advance(4088) (one less than the length of the chunk), the chunk will now look like:
[2 bytes chunked prefix | 3 bytes chunk length | 4088 bytes used chunk data | 1 byte unused chunk data | 2 bytes chunked suffix]

If someone calls GetMemory(0) it will return the 1 remaining byte in the chunk (without needing to insert the prefix/suffix) because we already wrapped the chunk. 

If someone calls GetMemory(2), we would write the current chunk before getting new memory.

> I have the same question for when sizeHint is 0, there's no space remaining in _currentChunkMemory, and _pipeWriter.GetMemory(0) does not return enough space for a chunk?

 Same applies as above, except the next call to GetMemory(0) would return a new chunk.

src/Http/Http.Abstractions/src/Extensions/HttpResponseWritingExtensions.cs
```diff
@@ -93,9 +93,10 @@ private static async Task StartAndWriteAsyncAwaited(this HttpResponse response,
 
         private static void Write(this HttpResponse response, string text, Encoding encoding)
         {
+            var minimumByteSize = encoding.GetMaxByteCount(1);
```
halter73: The encoding is almost always Encoding.UTF8 right? It might be better to hard code 6 as the minimumByteSize in that case.

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
```diff
@@ -538,57 +539,67 @@ public void Reset()
...
-                // Chunk is completely written, commit it to the pipe so GetMemory will return a new chunk of memory.
                 var writer = new BufferWriter<PipeWriter>(_pipeWriter);
                 WriteCurrentMemoryToPipeWriter(ref writer);
                 writer.Commit();
-
                 _unflushedBytes += writer.BytesCommitted;
-                _currentChunkMemory = _pipeWriter.GetMemory(sizeHint);
-                _currentChunkMemoryUpdated = true;
+
+                UpdateCurrentChunkMemory(sizeHint);
```
halter73: sizeHint could still be 0 here right? What if the underlying pipe only gives you 5 bytes? I think that actualMemory could end up a length 0 bytes. If the underlying pipe gave less than 5 bytes, I think the slice would fail.

src/Servers/Kestrel/Core/src/Internal/Http/Http1OutputProducer.cs
```diff
@@ -538,57 +539,67 @@ public void Reset()
...
 
+        private void UpdateCurrentChunkMemory(int sizeHint)
+        {
+            _currentChunkMemory = _pipeWriter.GetMemory(sizeHint);
+            _currentMemoryPrefixBytes = ChunkWriter.GetBeginChunkByteCount(_currentChunkMemory.Length);
+
+            // Super edge case. If we call GetMemory and it returns 4099, we would originally calculate _currentMemoryPrefixBytes
+            // as 6 bytes. However, after subtracting the bytes for _currentChunkMemory.Length and the endPrefix, the real
+            // body length would be 5 bytes. 
+            _currentMemoryPrefixBytes = ChunkWriter.GetBeginChunkByteCount(_currentChunkMemory.Length - _currentMemoryPrefixBytes - EndChunkLength);
```
halter73: Couldn't this logic be internal to GetBeginChunkByteCount?

halter73: Do you even whether this logic is stable? If the next line was also

```C#
_currentMemoryPrefixBytes = ChunkWriter.GetBeginChunkByteCount(_currentChunkMemory.Length - _currentMemoryPrefixBytes - EndChunkLength);
```

Is it guaranteed that _currentMemoryPrefixBytes won't change again?

halter73:
Is there a maximum chunk size other than int.MaxValue?

src/Servers/Kestrel/Core/src/Internal/Http/ChunkWriter.cs
```diff
@@ -44,6 +36,30 @@ public static int BeginChunkBytes(int dataCount, Span<byte> span)
             return count;
         }
 
+        internal static int GetBeginChunkByteCount(int count)
```
halter73: I'd prefer if this called into the other GetBeginChunkByteCount

jkotalik: I definitely intended to do that, must have missed it.

src/Servers/Kestrel/Core/src/Internal/Http/ChunkWriter.cs
```diff
@@ -44,6 +37,33 @@ public static int BeginChunkBytes(int dataCount, Span<byte> span)
...
         }
 
+        internal static int GetPrefixLength(Memory<byte> memory, int suffixLength)
+        {
+            var prefixLength = GetBeginChunkByteCount(memory.Length);
+
+            // Super edge case. If we call GetMemory and it returns 4099, we would originally calculate _currentMemoryPrefixBytes
+            // as 6 bytes. However, after subtracting the bytes for _currentChunkMemory.Length and the endPrefix, the real
+            // body length would be 5 bytes. 
+            return GetBeginChunkByteCount(memory.Length - prefixLength - suffixLength);
```
halter73: This logic isn't stable on the boundaries. This [gist](https://gist.github.com/halter73/af2b9f78978f83813b19e187c4e5309e) demonstrates how shortening the expected prefix length one byte with the second call to GetBeginChunkByteCount() allows the data length to become one byte longer once again necessitating a longer prefix.

> Given 21 bytes from underlying PipeWriter, slice 1 extra byte off end of actualMemory.

The above output from the gist shows that Memory with a length of 21 bytes is one of the problematic inputs. With this logic, prefixLength is 4 after the first line since it takes 2 hex chars to represent 21 (0x15) plus CRLF. The method the turns around and returns 3 though because 15 (21 - 4 - 2) can be represented with 1 hex char (0xF).

I think ultimately GetPrefixLength should return the lower number in these boundary cases, because returning the larger prefix length would always result in shifting data to the left in Advance() even when the buffer is completely filled. If we return the lower number though, we have to slice one extra byte off the end of the Memory Http1OutputProducer returns from GetMemory() so we don't end up in the situation where we need the longer prefix.

I'm wondering if we should just hard code the mapping from memory lengths to prefix lengths. With 8 or 16 return values depending on whether we special-case the boundaries, it should only take 3 or 4 comparisons. Or we could create a benchmark and let twitter go at it.

and I'm wondering if we should just hard code the ranges on this logic.

jkotalik: I reached the same conclusion as you too. I have a slightly different approach though, I’ll explain tomorrow.

jkotalik:
🆙 📅

src/Servers/Kestrel/Core/src/Internal/Http/ChunkWriter.cs
```diff
@@ -44,6 +44,97 @@ public static int BeginChunkBytes(int dataCount, Span<byte> span)
...
+            var length = memory.Length;
+
+            // If GetMemory returns one of the following values, there is no way to set the prefix/body lengths
+            // such that we either wouldn't have an invalid chunk or would need to copy if the entire memory chunk is used.
+            // For example, if GetMemory returned 21, we would guess that the chunked prefix is 4 bytes initially
+            // and the suffix is 2 bytes, meaning there is 15 bytes remaining to write into. However, 15 bytes only need 3
+            // bytes for the chunked prefix, so we would have to copy once we call advance. Therefore, to avoid this scenario,
+            // we slice the memory by one byte.
+
+            if (length <= 65544)
```
jkotalik: Constants for these or not? Don't know which I prefer here.

halter73: No need for constants IMHO if it's all used in this one method. I wouldn't mind checking in a link to the source that was used to come up with these numbers in case we ever want to tweek the algorithm.

jkotalik:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=137894

jkotalik:
```
YARN : error https: //registry.yarnpkg.com/yauzl/-/yauzl-2.10.0.tgz: Extracting tar content of undefined failed, the file appears to be corrupt: "ENOENT: no such file or directory, stat 'C:\\Users\\vsagent\\AppData\\Local\\Yarn\\Cache\\v4\\npm-yauzl-2.10.0-c7eb17c93e112cb1086fa6d8e51fb0667b79a5f9\\node_modules\\yauzl\\README.md'" [F:\workspace\_work\1\s\src\Components\test\E2ETest\Microsoft.AspNetCore.Components.E2ETests.csproj]
YARN : error An unexpected error occurred: "ENOTEMPTY: directory not empty, rmdir 'C:\\Users\\vsagent\\AppData\\Local\\Yarn\\Cache\\v4\\npm-urijs-1.19.1-5b0ff530c0cbde8386f6342235ba5ca6e995d25a\\node_modules\\urijs'". [F:\workspace\_work\1\s\src\ProjectTemplates\test\ProjectTemplates.Tests.csproj]
```

src/Servers/Kestrel/Core/test/ChunkWriterTests.cs
```diff
@@ -43,5 +43,21 @@ public void ChunkedPrefixMustBeHexCrLfWithoutLeadingZeros(int dataCount, string
 
             Assert.Equal(Encoding.ASCII.GetBytes(expected), span.Slice(0, count).ToArray());
         }
+
+        [Fact]
+        public void ChunkedPrefixReturnsSegmentThatDoesNotNeedToMove()
```
halter73: How long does this test take? It doesn't cover some edge cases like i = 268435467 since it doesn't loop that high.

Instead of testing every i from 5 to 10000000 (5 isn't even valid), it might be better to just test the boundary cases. So basically 6, int.MaxVales, the constants used in GetPrefixBytesForChunk, and the values neighboring those constants.

src/Http/Http.Abstractions/src/Extensions/HttpResponseWritingExtensions.cs
```diff
@@ -126,7 +139,7 @@ private static void WriteMultiSegmentEncoded(PipeWriter writer, string text, Enc
                 writer.Advance(bytesUsed);
                 source = source.Slice(charsUsed);
 
-                destination = writer.GetSpan(encodedLength - totalBytesUsed);
+                destination = writer.GetSpan(minimumByteSize);
```
halter73: Can minimumByteSize ever be greater than `encodedLength - totalBytesUsed`?

halter73: Never mind. I don't know why I though this was a problem. Totally makes sense why you use the smallest value possible. Hopefully it doesn't come back to bite us.

src/Servers/Kestrel/Core/src/Internal/Http/ChunkWriter.cs
jkotalik: I'll include your algorithm just for the shout out.

jkotalik:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=137894

src/Servers/Kestrel/Core/test/ChunkWriterTests.cs
jkotalik: The only reason I didn't test 268435467 was due to OOM. It doesn't take long (less than a second).

jkotalik: I'll rewrite the test to not allocate a large array.

jkotalik: Bleh I don't like how I would need to rewrite it. I would need to either change GetPrefixBytesForChunk to pass in a length or do something hacky.

aspnet-hello:

I've triaged the above build.

aspnet-hello:

I've triaged the above build.

halter73 (approved):

PR #8840, opened 2019-03-26
Update Gradle to version 5.3 - Now with retries

mikaelm12:
Fixes: https://github.com/aspnet/AspNetCore/issues/8804
Related: https://github.com/aspnet/AspNetCore-Internal/issues/2052

Gradle added some resiliency features in 5.0 (we were on 4.9 before :/ ) particularly, retries with exponential back off strategies. This should add some stability to our builds


BrennanConroy (approved):

src/SignalR/clients/java/signalr/src/test/java/com/microsoft/signalr/NegotiateResponseTest.java
```diff
@@ -5,10 +5,11 @@
 
 import static org.junit.jupiter.api.Assertions.*;
 
-import com.google.gson.stream.JsonReader;
+import java.io.StringReader;
+
 import org.junit.jupiter.api.Test;
 
-import java.io.StringReader;
```
halter73: Is there some ordering convention that we're using for the imports?

halter73 (approved):

src/SignalR/clients/java/signalr/src/test/java/com/microsoft/signalr/NegotiateResponseTest.java
mikaelm12: Yeah, java, javax, org, com enforced by this rule https://github.com/aspnet/AspNetCore/blob/1f63f25508dfb8f46715d1842cf3dd49bd21a300/src/SignalR/clients/java/signalr/build.gradle#L47

Eilon (approved):

PR #8841, opened 2019-03-26
Get Binlogs from other linux machines.

jkotalik:
Getting binlogs from other OSs.


dougbu (approved):
Please add the parameter to the macOS build too. If it doesn't break the test build steps, do them as well.

jkotalik:
Yeah I'll make sure this exists for all build configurations. I just wanted parity with what I'm verifying in #8670

PR #8863, opened 2019-03-27
Components: don't block the SignalR loop during init. Fixes #8274

### LINKED ISSUE #8274: Tasks returned from JSRuntime.InvokeAsync during OnInitAsync never finishes in preview3
### Describe the bug
Tasks returned from JSRuntime.InvokeAsync during second call to OnInitAsync (after prerendering) never finishes.

### To Reproduce
Create a new razorcomponents project.

Add a js interop function to Index.cshtml and disable prerendering

```html
    <script>
        window.interopDuringOnInit = function() {
            return "Hello World";
        }
    </script>

    <app></app>

    <script src="_framework/components.server.js"></script>
```

Call js interop during OnAsyncInit in Index.razor

```razor
<h1>@Greeting!</h1>

Welcome to your new app.

@functions {
    string Greeting;

    protected override async Task OnInitAsync()
    {
        try
        {
            Console.WriteLine("Calling browser");
            Greeting = await JSRuntime.InvokeAsync<string>("interopDuringOnInit");
            Console.WriteLine("Returned from browser");
        }
        catch (Exception ex)
        {
            Console.WriteLine("Could not invoke interop during OnInit, " + ex.ToString());
        }
    }
}
```

When OnInitAsync is called as part of an initial load (prerendering disabled) the task never finishes, This leaves the circuit hanging, and no further events are processed.

If this component is loaded from a `IUriHelper.OnLocationChange`, the above code works as expected.

SteveSandersonMS:
For more details, see https://github.com/aspnet/AspNetCore/issues/8274#issuecomment-477203983

### DIFF
--- a/src/Components/Server/src/Circuits/CircuitHost.cs
+++ b/src/Components/Server/src/Circuits/CircuitHost.cs
@@ -104,21 +104,32 @@ namespace Microsoft.AspNetCore.Components.Server.Circuits
             });
         }
 
-        public async Task InitializeAsync(CancellationToken cancellationToken)
+        public void Initialize(CancellationToken cancellationToken)
         {
-            await Renderer.InvokeAsync(async () =>
+            // This Initialize method is fire-and-forget as far as the caller is concerned, because
+            // if it was to await completion, it would be blocking the SignalR message loop. This could
+            // lead to deadlock, e.g., if the init process itself waited for an incoming SignalR message
+            // such as the result of a JSInterop call.
+            Renderer.InvokeAsync(async () =>
             {
-                SetCurrentCircuitHost(this);
-
-                for (var i = 0; i < Descriptors.Count; i++)
+                try
                 {
-                    var (componentType, domElementSelector) = Descriptors[i];
-                    await Renderer.AddComponentAsync(componentType, domElementSelector);
+                    SetCurrentCircuitHost(this);
+
+                    for (var i = 0; i < Descriptors.Count; i++)
+                    {
+                        var (componentType, domElementSelector) = Descriptors[i];
+                        await Renderer.AddComponentAsync(componentType, domElementSelector);
+                    }
+
+                    await OnCircuitOpenedAsync(cancellationToken);
+
+                    await OnConnectionUpAsync(cancellationToken);
                 }
-
-                await OnCircuitOpenedAsync(cancellationToken);
-
-                await OnConnectionUpAsync(cancellationToken);
+                catch (Exception ex)
+                {
+                    Renderer_UnhandledException(this, ex);
+                }
             });
 
             _initialized = true;
--- a/src/Components/Server/src/Circuits/CircuitHost.cs
+++ b/src/Components/Server/src/Circuits/CircuitHost.cs
@@ -104,35 +104,35 @@ namespace Microsoft.AspNetCore.Components.Server.Circuits
             });
         }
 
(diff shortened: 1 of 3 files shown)


src/Components/Server/src/Circuits/CircuitHost.cs
```diff
@@ -104,21 +104,32 @@ public Task<IEnumerable<string>> PrerenderComponentAsync(Type componentType, Par
...
-                    var (componentType, domElementSelector) = Descriptors[i];
-                    await Renderer.AddComponentAsync(componentType, domElementSelector);
-                }
+                    SetCurrentCircuitHost(this);
 
-                await OnCircuitOpenedAsync(cancellationToken);
+                    for (var i = 0; i < Descriptors.Count; i++)
+                    {
+                        var (componentType, domElementSelector) = Descriptors[i];
+                        await Renderer.AddComponentAsync(componentType, domElementSelector);
```
javiercn: Do we still have to work for quiescence here until we declare the circuit opened?

javiercn (approved):
Looks great. Could we perhaps have an E2E test for this?

SteveSandersonMS:
@javiercn I don't think we have any infrastructure for E2E tests specifically about initialization, since all our E2E tests run in a context where the app has already started running. I agree it would be good if we did have more coverage around that, but it's not something I'd like to get deep into at this point.

src/Components/Server/src/Circuits/CircuitHost.cs
SteveSandersonMS: No, if we did that, then JSInterop calls made in the meantime (e.g., during InitAsync) would throw because they are only allowed once the circuit is flagged as opened.

We need the circuit to be regarded as opened synchronously. Since dispatcher calls are serialized, and since (non-quiescent) rendering is synchronous, we already know that we'll only start to process incoming JSInterop calls *after* the first synchronous render, which is the right time to start processing them.

src/Components/Server/src/ComponentHub.cs
```diff
@@ -64,7 +64,7 @@ public override Task OnDisconnectedAsync(Exception exception)
         /// <summary>
         /// Intended for framework use only. Applications should not call this method directly.
         /// </summary>
-        public async Task<string> StartCircuit(string uriAbsolute, string baseUriAbsolute)
+        public string StartCircuit(string uriAbsolute, string baseUriAbsolute)
```
pranavkm: :+1:

pranavkm (approved):

src/Components/Server/src/Circuits/CircuitHost.cs
```diff
@@ -104,21 +104,32 @@ public Task<IEnumerable<string>> PrerenderComponentAsync(Type componentType, Par
             });
         }
 
-        public async Task InitializeAsync(CancellationToken cancellationToken)
```
rynowak: It would be really nice to leave the signature alone and have the hub be the one to fire-and-forget this task. When we have async code we need to be able to await it in a test or errors will creep in. 

Basically have a big, thick, gooey async method that returns a task, and have a small stub (`StartCircuit` in this case) do `_ = InitializeAsync()`

SteveSandersonMS: Sure, done.

rynowak (approved):

PR #8870, opened 2019-03-27
Remove explicit footer height

adrianwright109:
In the default File > New > web project(s) the site.css file has an explicit height set for the .footer class. This causes the page to then have a scrollbar even though the page doesn't need one.

This PR removes the height: 60px; from the .footer css class on the following templates:
RazorPagesWeb-CSharp
StarterWeb-CSharp
StarterWeb-FSharp

Addresses #7305


ryanbrandenburg (approved):
Seems good to me. Thanks for the contribution!

ryanbrandenburg:
I'm re-trying the build, will merge once it passes.

Issue #8873, opened 2019-03-27
TempData has poor round-tripping fidelity


pranavkm:
This issue applies to using `TempDataDictionary` although it is generally much worse with `TempDataAttribute`. By round-tripping, I mean a value being set, serialized to bytes that gets written to cookie or session, and deserialized back to `TempDataDictionary` and read.

Consider, 
```C#
public class IndexModel
{
    [TempData]
    public int Value { get; set; }
    
    public ActionResult OnPost()
    {
        Value = 10;
        return RedirectToPage();
    }
    
    public void OnGet()
    {
        // We should expect Value to be 10 here
    }
}
```

For the above scenario, outside of a few types such as a `int`, `DateTime`, `string` and `Guid`, round-tripping using the Json.NET based serializer will fail due to a mismatched round-trip type. This is because no type information is preserved in the payload, and we get Json.NET's best guess results which for types such as `TimeSpan` or `Uri` is `string`. 

With the `TempData[".."]` lookup, users may have written code anticipating the round trip to be a string value, however, using `TempDataAttribute` would result in a runtime `InvalidCastException`.

rynowak:
Are you suggesting that we take steps to mitigate this or explain it? I'd generally be supportive of something like an analzyer that provides guidance, or better errors messages when it does fail, or blocking things that can't work with an exception.

pranavkm:
> Are you suggesting that we take steps to mitigate this or explain it?

In theory, we could do a better job with `TempData` attributes since we know the `Type` that's being round-tripped. However, in the shorter term, it might be better to limit `TempData` properties to a more limited subset - `int`, `bool`, `string`, and `DateTime`.

mkArtakMSFT:
Let's see whether anybody even asks for it. Parking in the backlog for now.

rynowak:
This is going to be asked for as soon as we ship :) I'd be happy to be wrong.

pranavkm:
> I'd be happy to be wrong.

Closing as we haven't had any specific user feedback here. I guess you have a reason to be happy now 😄

rynowak:
I'm super happy now.

PR #8874, opened 2019-03-27
Add a System.Text.Json based TempDataSerializer

pranavkm:
* Update DefaultTempDataSerializer
* Add common tests for DefaultTempDataSerializer & BsonTempDataSerializer
* Remove uses of NewtonsoftJson in tests solely required for temp-data support

Fixes https://github.com/aspnet/AspNetCore/issues/7255

### DIFF
--- a/src/Mvc/Mvc.NewtonsoftJson/src/BsonTempDataSerializer.cs
+++ b/src/Mvc/Mvc.NewtonsoftJson/src/BsonTempDataSerializer.cs
@@ -149,7 +149,7 @@ namespace Microsoft.AspNetCore.Mvc.NewtonsoftJson
             }
             else
             {
-                return Array.Empty<byte>();;
+                return Array.Empty<byte>();
             }
         }
 
--- a/src/Mvc/Mvc.NewtonsoftJson/src/BsonTempDataSerializer.cs
+++ b/src/Mvc/Mvc.NewtonsoftJson/src/BsonTempDataSerializer.cs
@@ -149,7 +149,7 @@ namespace Microsoft.AspNetCore.Mvc.NewtonsoftJson
             }
             else
             {
-                return new byte[0];
+                return Array.Empty<byte>();;
             }
         }
 
@@ -164,6 +164,8 @@ namespace Microsoft.AspNetCore.Mvc.NewtonsoftJson
                 throw new InvalidOperationException(errorMessage);
             }
         }
+
+        public override bool CanSerializeType(Type type) => CanSerializeType(type, out _);
 
         private static bool CanSerializeType(Type typeToSerialize, out string errorMessage)
         {
@@ -208,6 +210,8 @@ namespace Microsoft.AspNetCore.Mvc.NewtonsoftJson
             }
 
             actualType = actualType ?? typeToSerialize;
+            actualType = Nullable.GetUnderlyingType(actualType) ?? actualType;
+
             if (!IsSimpleType(actualType))
             {
                 errorMessage = Resources.FormatTempData_CannotSerializeType(
--- a/src/Mvc/Mvc.ViewFeatures/src/Infrastructure/ArrayBufferWriter.cs
+++ b/src/Mvc/Mvc.ViewFeatures/src/Infrastructure/ArrayBufferWriter.cs
@@ -1,0 +1,181 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Buffers;
+using System.Diagnostics;
+
+namespace Microsoft.AspNetCore.Mvc.ViewFeatures.Infrastructure
+{
+    // Note: this is currently an internal class that will be replaced with a shared version.
+    internal sealed class ArrayBufferWriter<T> : IBufferWriter<T>, IDisposable
+    {
+        private T[] _rentedBuffer;
+        private int _index;
+
+        private const int MinimumBufferSize = 256;
+
+        public ArrayBufferWriter()
+        {
+            _rentedBuffer = ArrayPool<T>.Shared.Rent(MinimumBufferSize);
+            _index = 0;
+        }
+
+        public ArrayBufferWriter(int initialCapacity)
+        {
+            if (initialCapacity <= 0)
+                throw new ArgumentException(nameof(initialCapacity));
+
+            _rentedBuffer = ArrayPool<T>.Shared.Rent(initialCapacity);
+            _index = 0;
+        }
+
+        public ReadOnlyMemory<T> WrittenMemory
+        {
+            get
+            {
+                CheckIfDisposed();
+
+                return _rentedBuffer.AsMemory(0, _index);
+            }
+        }
+
+        public int WrittenCount
+        {
+            get
+            {
+                CheckIfDisposed();
+
+                return _index;
+            }
+        }
+
+        public int Capacity
+        {
+            get
+            {
+                CheckIfDisposed();
+
+                return _rentedBuffer.Length;
+            }
+        }
+
+        public int FreeCapacity
+        {
+            get
+            {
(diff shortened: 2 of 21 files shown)


src/Mvc/Mvc.ViewFeatures/src/Infrastructure/DefaultTempDataSerializer.cs
```diff
@@ -3,30 +3,142 @@
...
+            using var jsonDocument = JsonDocument.Parse(value);
+            var rootElement = jsonDocument.RootElement;
+            return DeserializeDictionary(rootElement);
+        }
+
+        private IDictionary<string, object> DeserializeDictionary(JsonElement rootElement)
+        {
+            var deserialized = new Dictionary<string, object>(StringComparer.Ordinal);
+
+            foreach (var item in rootElement.EnumerateObject())
```
pranavkm: We're going to have to keep this code, so it's likely the deserializer that would change. This does have fewer features compared to the Json.Net one (it does not support arrays or dictionaries), however round-tripping is really broken with the current implementation (https://github.com/aspnet/AspNetCore/issues/8873). For anything but the most trivial cases, you get InvalidCastException. So doing a more limited set of types that we can correctly support all the time seems like a better solution for now.

src/Mvc/test/WebSites/HtmlGenerationWebSite/Startup.cs
```diff
@@ -19,7 +19,6 @@ public void ConfigureServices(IServiceCollection services)
             // null which is interpreted as true unless element includes an action attribute.
             services.AddMvc(ConfigureMvcOptions)
                 .InitializeTagHelper<FormTagHelper>((helper, _) => helper.Antiforgery = false)
-                .AddNewtonsoftJson()
```
pranavkm: We were using AddNewtonsoftJson to get an implementation of the tempdata serializer. We no longer need it now.

src/Mvc/Mvc.NewtonsoftJson/src/BsonTempDataSerializer.cs
```diff
@@ -149,7 +149,7 @@ public override byte[] Serialize(IDictionary<string, object> values)
             }
             else
             {
-                return new byte[0];
+                return Array.Empty<byte>();;
```
rynowak: 😻

src/Mvc/Mvc.NewtonsoftJson/src/BsonTempDataSerializer.cs
```diff
@@ -208,6 +210,8 @@ private static bool CanSerializeType(Type typeToSerialize, out string errorMessa
             }
 
             actualType = actualType ?? typeToSerialize;
+            actualType = Nullable.GetUnderlyingType(actualType) ?? actualType;
```
rynowak: Is this a new issue found while testing?

src/Mvc/Mvc.ViewFeatures/src/Infrastructure/ArrayBufferWriter.cs
```diff
@@ -0,0 +1,181 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
```
rynowak: License is different because it came from corefx?

src/Mvc/Mvc.ViewFeatures/src/Infrastructure/DefaultTempDataSerializer.cs
```diff
@@ -3,30 +3,142 @@
...
+            {
+                throw new ArgumentNullException(nameof(value));
+            }
+
+            if (value.Length == 0)
+            {
+                return new Dictionary<string, object>();
+            }
+
+            using var jsonDocument = JsonDocument.Parse(value);
```
rynowak: 😻

src/Mvc/Mvc.ViewFeatures/src/Infrastructure/DefaultTempDataSerializer.cs
rynowak: I think as long as the error message is reasonable for preview 4 this is the right thing to do.

I expected us to do something specialized here anyway since we want to serialize each entry specially.

src/Mvc/Mvc.ViewFeatures/src/Infrastructure/DefaultTempDataSerializer.cs
```diff
@@ -3,30 +3,142 @@
...
+                        deserializedValue = item.Value.GetBoolean();
+                        break;
+
+                    case JsonValueType.Number:
+                        deserializedValue = item.Value.GetInt32();
+                        break;
+
+                    case JsonValueType.String:
+                        var stringValue = item.Value.GetString();
+                        if (DateTime.TryParseExact(stringValue, "o", CultureInfo.InvariantCulture, DateTimeStyles.RoundtripKind, out var dateTime))
```
rynowak: Do we keep track of the types we're putting in? 


Seeing logic like this just seems totally flawed? What if a string happened to contain a valid date. I think we have to roundtrip the type names. What did we do with the JSON.NET version?

src/Mvc/Mvc.NewtonsoftJson/src/BsonTempDataSerializer.cs
davidfowl: nit: extra `;`

src/Mvc/Mvc.NewtonsoftJson/src/BsonTempDataSerializer.cs
pranavkm: Yup and it's mostly my fault. We were previously unwrapping as part of the filter - https://github.com/aspnet/AspNetCore/blame/d116ab8550ea67aefd67992072cb3bfe08723f0c/src/Mvc/src/Microsoft.AspNetCore.Mvc.ViewFeatures/Filters/SaveTempDataPropertyFilterBase.cs#L125, but this got missed when I changed it from a static method on the one `TempDataSerializer` type to an instance on `TempDataSerializer`. Added some tests that cover this

src/Mvc/Mvc.ViewFeatures/src/Infrastructure/ArrayBufferWriter.cs
pranavkm: Yeah. I think the plan is for the type to go away. I'll update the license to keep it in sync with the rest of the codes.

src/Mvc/Mvc.ViewFeatures/src/Infrastructure/DefaultTempDataSerializer.cs
pranavkm: With Json.NET, `TypeNameHandling` is null, so we get Json.NET's best effort behavior. From what I can tell,

1) `string`, `bools` are always round-tripped correctly.
2) `int` get round-tripped as `long`. We do special handling for this. Consequently things like `short`, `byte`, or a `long` that fits inside an `int` don't round-trip
3) Json.NET returns a `DateTime` for any string that can be parsed by `DateTime.Parse`. Same goes for `Guid`. However, `DateTimeOffset` round trip as a `DateTime` instance, `TimeSpan` and `Uri` as a `string` value

We could in theory enable `TypeNameHandling` in Json.NET's implementation to get better behavior, part of why I filed #8873, but that flag's controversial. 

For this specific bit, we could remove the special support for `DateTime`. We have one test that fails as a result, but I can change it to do a different data type

rynowak: Ok I didn't realize that's what we did. I think we should try to make this highly compatible with newtonsoft if possible.

pranavkm:
🆙 📅

rynowak (approved):

PR #8877, opened 2019-03-28
Chain coherent dependencies up through Extensions to EF Core packages

dougbu:
- should reduce incoherent PRs we get from `darc`
- update dependencies from 5 repos to the latest coherent set

PR #8887, opened 2019-03-28
Use LangVersion preview rather than 8.0

benaadams:
Reference updates are choking on nullable reference types (e.g. https://github.com/aspnet/AspNetCore/pull/8819, https://github.com/aspnet/AspNetCore/pull/8885); so switch from language "8.0" to "preview" to fix this (although they should be the same at this time, looks to be a roslyn update to put this under the new "preview" version)

/cc @NTaylorMullen,@dougbu

### DIFF
--- a/eng/targets/CSharp.Common.props
+++ b/eng/targets/CSharp.Common.props
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup>
-    <LangVersion>8.0</LangVersion>
+    <LangVersion>preview</LangVersion>
     <AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)..\AspNetCore.snk</AssemblyOriginatorKeyFile>
     <SignAssembly>true</SignAssembly>
 


benaadams:
Might want to cherry pick this into https://github.com/aspnet/AspNetCore/pull/8885 as the current sdk refs don't understand "preview" and the new ones don't allow nullable ref types under "8.0"

dougbu:
Doesn't look like this could pass without updating the SDK. I'm surprised _any_ of the CI checks are succeeding. Probably need to get #8670 in first or to cherry-pick this change in there so we don't forget. (Things may also improve after I get #8877 in but I doubt it.)

dougbu:
> Probably need to get #8670 in first or to cherry-pick this change in there

I take this back. Looks like I need this in #8877. I'll do the cherry-pick and see if that gets the CI checks on #8877 passing.

dougbu:
Oh well, can't get n #8877 working without a new SDK either. We'll need to wait for #8670.

@JunTaoLuo do you have an estimate when #8670 will be in?

NTaylorMullen:
@benaadams we definitely want to move to "preview" as well 😄. Once we're able to consume a new SDK that should be possible. Folks are d ealing with #8670 and once that's in this should be unblocked as well.

benaadams:
Don't think this is needed; looks like its Razor view compilation that's having the issue, not the general source https://github.com/aspnet/AspNetCore/pull/8877#issuecomment-478181211

Issue #8786, opened 2019-03-25
Better ways to manage prerendering+JSInterop interactions


SteveSandersonMS:
### Background

I've tried to collect all the issues related to this. Here's what appears to be the full set:

 * #8422 Inject different services during prerendering
    * This one is about prerendering followed by *client-side* execution. It requests a way to inject different services for the two cases.
    * We already have that capability (since one's on the server, and the other is on the client), so I'm closing the issue.
    * We're not planning to have a separate DI container for *prerendering* as opposed to other server-side execution
 * #8404 JSInterop improvements
    * This is a meta-issue listing several ideas for further JSInterop features. Arguably it serves exactly the same purpose as this issue, so we can probably close both at the same time once something has been implemented.
 * #8592 Navigation is not allowed during prerendering
    * Only tenuously related, because it's specifically about navigation, not JSInterop. Including just to get a broader picture.
 * #8287 Please can we have some way to know we are pre-rendering so our components don't error
    * Duplicate of #8404, so now closed.
 * #8327 . NET core 3 preview 3 - razor component - unable tu use JsRuntime on the component lifecycle
    * Was already closed as a duplicate of #8404, but additionally states the following: *Expected behavior: Having a method in the component where i can call my js after component rendered*
 * #7842 Add new lifecycle method OnAfterFirstRender
    * Not really related to prerendering, but is related to the need to do one-time JSInterop calls after component initialization

### Scenarios

From all the descriptions and comments in the above issues, I extract the following scenarios:

* [A] **During prerendering, I want my component to render different output** (or a blank space). Once we switch into interactive mode, it should render again in regular (non-prerendered) mode.
    * e.g., to render in "ghosted out" mode while non-interactive
    * e.g., to skip certain computations or data access that we don't want to block the initial HTML response on
* [B] I have a component that needs to do JSInterop as part of its initialization, so **I want to defer those JSInterop calls until JSInterop becomes available**.
    * e.g., to initialize a JS widget library that gets attached to my component's rendered DOM elements
    * e.g., to invoke some JS logging library that doesn't make sense to call during prerendering

### Possible solutions

 * [1] We could have a DI service, e.g., `@inject IPrerenderingContext PrerenderingContext` that lets you evaluate `PrerenderingContext.IsPrerendering` anywhere in your lifecycle methods or rendering logic
[...]

PR #8888, opened 2019-03-28
Supply "IsConnected" state info to components

SteveSandersonMS:
This is a partial implementation of what we need for #8786. It's pretty basic and just follows from existing patterns.

Still to do: E2E tests showing the swap from "prerendering/disconnected" into "connected", with a component that renders different output in the two cases.

Question: Do we need some other flag to differentiate the "prerendering" case from the "was interactive but then disconnected" case? I'm not sure we have a use case for that.

### DIFF
--- a/src/Components/Components/ref/Microsoft.AspNetCore.Components.netstandard2.0.cs
+++ b/src/Components/Components/ref/Microsoft.AspNetCore.Components.netstandard2.0.cs
@@ -836,6 +836,10 @@ namespace Microsoft.AspNetCore.Components.Routing
 }
 namespace Microsoft.AspNetCore.Components.Services
 {
+    public partial interface IComponentContext
+    {
+        bool IsConnected { get; }
+    }
     public partial interface IUriHelper
     {
         event System.EventHandler<string> OnLocationChanged;
--- a/src/Components/Components/src/Services/IComponentContext.cs
+++ b/src/Components/Components/src/Services/IComponentContext.cs
@@ -1,0 +1,19 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Components.Services
+{
+    /// <summary>
+    /// Provides information about the environment in which components are executing.
+    /// </summary>
+    public interface IComponentContext
+    {
+        /// <summary>
+        /// Gets a flag to indicate whether there is an active connection to the user's display.
+        /// </summary>
+        /// <example>During prerendering, the value will always be false.</example>
+        /// <example>During server-side execution, the value can be true or false depending on whether there is an active SignalR connection.</example>
+        /// <example>During client-side execution, the value will always be true.</example>
+        bool IsConnected { get; }
+    }
+}
--- a/src/Components/test/testassets/TestServer/Startup.cs
+++ b/src/Components/test/testassets/TestServer/Startup.cs
@@ -1,4 +1,5 @@
 using BasicTestApp;
+using BasicTestApp.RouterTest;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Components.Server;
 using Microsoft.AspNetCore.Hosting;
@@ -67,6 +68,19 @@ namespace TestServer
             app.UseEndpoints(endpoints =>
             {
                 endpoints.MapControllers();
+            });
+
+            // Separately, mount a prerendered server-side Blazor app on /prerendered
+            app.Map("/prerendered", subdirApp =>
+            {
+                subdirApp.UsePathBase("/prerendered");
+                subdirApp.UseStaticFiles();
+                subdirApp.UseRouting();
+                subdirApp.UseEndpoints(endpoints =>
+                {
+                    endpoints.MapFallbackToPage("/PrerenderedHost");
+                    endpoints.MapComponentHub<TestRouter>("app");
+                });
             });
         }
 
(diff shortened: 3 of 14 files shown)


src/Components/Components/src/Services/IComponentContext.cs
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Components.Services
```
rynowak: What would you think about putting this in `Microsoft.AspNetCore.Components`. I really think we'll end up removing `.Services` as a namespace when this gets reviewed in favor of some mix of: 
- moving these types into `.Components`
- moving these types into more specific namespaces like routing/layouts
- moving them into a more generic namespace like `.Infrastructure` for things that most people won't use

src/Components/Components/src/Services/IComponentContext.cs
```diff
@@ -0,0 +1,19 @@
...
+    /// </summary>
+    public interface IComponentContext
+    {
+        /// <summary>
+        /// Gets a flag to indicate whether there is an active connection to the user's display.
+        /// </summary>
+        /// <example>During prerendering, the value will always be false.</example>
+        /// <example>During server-side execution, the value can be true or false depending on whether there is an active SignalR connection.</example>
+        /// <example>During client-side execution, the value will always be true.</example>
+        bool IsConnected { get; }
```
rynowak: Should we document some guarantees about what happens when transition between states? Or is this not the right place for that.

SteveSandersonMS: What kind of guarantees do you have in mind?

rynowak: Like, if you weren't connected when you last rendered - does that mean that you get called again when you are connected?

SteveSandersonMS: Ah, I see. Since we're not proposing to do that, I'm not sure what there would be to say about it here.

I expect you know already, but the intent to deal with the mainstream "transition from prerendered to interactive" case is via lifecycle methods. For a more general "notify me when connection state changes", I'd expect that to be a separate API, which someone could wrap into a cascading value if they wanted.

src/Components/Components/src/Services/IComponentContext.cs
SteveSandersonMS: Totally happy to merge the `Components.Services` namespace into `.Components`. That would also fix one of the other things that's been bothering me lately - that the templates imports don't include `.Services`.

Filed https://github.com/aspnet/AspNetCore/issues/8892

src/Components/Components/src/Services/IComponentContext.cs
rynowak: Ok cool. I guess what I was thinking about is where are we going to document to the pattern for using this property effectively. That's probably not a good fit for the in-code docs, so feel free to ignore.

rynowak (approved):

javiercn (approved):

SteveSandersonMS:
Temporarily rebased onto https://github.com/aspnet/AspNetCore/pull/8910 so that the E2E test can work.

@javiercn, I know you already approved this, but now I've added the E2E test it might be worth taking another look. Maybe the format of test will be useful in your prerendering PR too (which I will read this afternoon).

javiercn:
I already included something similar in my draft PR :(

src/Components/test/testassets/TestServer/Startup.cs
```diff
@@ -68,6 +69,19 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
...
+            {
+                subdirApp.UsePathBase("/prerendered");
+                subdirApp.UseStaticFiles();
+                subdirApp.UseRouting();
+                subdirApp.UseEndpoints(endpoints =>
+                {
+                    endpoints.MapFallbackToPage("/PrerenderedHost");
+                    endpoints.MapComponentHub<TestRouter>("app");
+                });
+            });
```
javiercn: This should not be our experience for having multiple entry points as part of an app. But that requires a separate discussion. See https://github.com/aspnet/AspNetCore/issues/8913

src/Components/Components/ref/Microsoft.AspNetCore.Components.netstandard2.0.cs
```diff
@@ -836,6 +836,10 @@ public enum NavLinkMatch
 }
 namespace Microsoft.AspNetCore.Components.Services
 {
+    public partial interface IComponentContext
```
javiercn: How did you update this? Can you try and get the private members of the components that we ship to show up? That breaks the current E2E test in ProjectTemplates. It would be incredibly dope if we updated and enable that.

javiercn (approved):
Signed off again.

src/Components/Components/ref/Microsoft.AspNetCore.Components.netstandard2.0.cs
SteveSandersonMS: I ran `GenerateReferenceSource` as per https://github.com/aspnet/AspNetCore/blob/master/docs/ReferenceAssemblies.md

What you see here is what it outputs.

src/Components/test/testassets/TestServer/Startup.cs
SteveSandersonMS: Sure, this is just what you have to do today, not a claim that it's the ideal dev experience or that we shouldn't change it.

src/Components/Components/ref/Microsoft.AspNetCore.Components.netstandard2.0.cs
javiercn: I do believe we did this for some other APIs in the assembly. This is an ongoing problem on the templates. Things like `Assembly="typeof(Blah).Assembly"` won't work because it won't be able to find the Assembly property on the component and it will treat it as a string instead when compiling.

SteveSandersonMS: Yes, @rynowak fixed that by telling the tool not to generate anything for those types, and he wrote those ref types manually. I don't think we need to do `IComponentContext` manually though.

rynowak: For now we need to manually generate the reference source of components because non-public methods do not show up in reference assemblies. This should not apply to anything else we need to do. If the contagion starts spreading I would love to learn more about that 😆

PR #8900, opened 2019-03-28
Chain coherent dependencies up through Extensions to EF Core packages

dougbu:
- should reduce incoherent PRs we get from `darc`
- update dependencies from 5 repos to the latest coherent set


dougbu:
/fyi @mmitche we'll see how this fails 😺

dougbu:
@mmitche this is a good sample build containing loads of package downgrade errors when most of the Extensions packages were pulling in a newer Microsoft.NetCore.App package than the projects referenced directly

dougbu:
@mmitche closing this PR but I'll leave the branch around in case you need it. Build https://dev.azure.com/dnceng/9ee6d478-d288-47f7-aacc-f6e6d082ae6d/_build/results?buildId=137146 should also stay around for a while.

Issue #5624, opened 2018-10-31
Make it easier to enable client-side SignalR logging when using server-side Blazor


danroth27:
We should provide a simple way to [enable client-side SignalR logging](https://github.com/aspnet/SignalR/wiki/Diagnostics-Guide#javascript-client-logging) when using server-side Blazor.

sven5:
@danroth27 
Looking forward for this feature. In the meantime, is there any way I can manually enable client-side logging?

Thanks
Sven

danroth27:
@sven5 There are a few [community projects](https://blazor.net/community) out there that you can take a look at:
- https://github.com/BlazorExtensions/Logging
- https://github.com/elmahio/Elmah.Io.Blazor

PR #8901, opened 2019-03-29
Make Blazor initialization configurable

### LINKED ISSUE #6887: Figure out a way to allow configuring SignalR settings on the client.
The current SignalR connection on the client is set up to print info logs, but this isn't required in the general case. However it would be useful to retain the ability to configure the log level among other things, in the event the user's attempting to debug connection issues on the client.

---------------------------------------
## Original issue report

When running a Razor Components app you see the following log output in the browser console.

```
Information: Normalizing '_blazor' to 'http://localhost:58443/_blazor'. - components.server.js:16 
Information: WebSocket connected to ws://localhost:58443/_blazor?id=EQLp2HObPNphy4UacPN7Jw - components.server.js:16 
Information: Using HubProtocol 'messagepack'.
```

We should investigate why we are logging Information level messages by default and how a user would change the log level.

rynowak:
Fixes: #6887 and #6887 and #5624

Sample:
```html
    <script src="_framework/components.server.js" autostart="false"></script>
    <script>
        Blazor.start({
            configureSignalR: function (builder) {
                builder.configureLogging(2); // LogLevel.Information
            }
        });
  </script>
```

Adds support for calling Blazor.start({...}) and passing in a
configuration object.

For now all you can configure is the SignalR HubConnectionBuilder. This
is a priority right now because we want to make configuring SignalR's
logging accessible.

### DIFF
--- a/src/Components/Browser.JS/src/Boot.Server.ts
+++ b/src/Components/Browser.JS/src/Boot.Server.ts
@@ -2,7 +2,7 @@ import '@dotnet/jsinterop';
 import './GlobalExports';
 import * as signalR from '@aspnet/signalr';
 import { MessagePackHubProtocol } from '@aspnet/signalr-protocol-msgpack';
-import { fetchBootConfigAsync, loadEmbeddedResourcesAsync } from './BootCommon';
+import { fetchBootConfigAsync, loadEmbeddedResourcesAsync, shouldAutoStart } from './BootCommon';
 import { CircuitHandler } from './Platform/Circuits/CircuitHandler';
 import { AutoReconnectCircuitHandler } from './Platform/Circuits/AutoReconnectCircuitHandler';
 import RenderQueue from './Platform/Circuits/RenderQueue';
@@ -10,9 +10,21 @@ import { ConsoleLogger } from './Platform/Logging/Loggers';
 import { LogLevel, ILogger } from './Platform/Logging/ILogger';
 import { discoverPrerenderedCircuits, startCircuit } from './Platform/Circuits/CircuitManager';
 
-let renderingFailed = false;
 
-async function boot(): Promise<void> {
+type SignalRBuilder = (builder: signalR.HubConnectionBuilder) => void;
+interface BlazorOptions {
+  configureSignalR?: SignalRBuilder,
+};
+
+let renderingFailed = false;
+let started = false;
+
+async function boot(options?: BlazorOptions): Promise<void> {
+
+  if (started) {
+    throw new Error('Blazor has already started.');
+  }
+  started = true;
 
   // For development.
   // Simply put a break point here and modify the log level during
@@ -30,7 +42,9 @@ async function boot(): Promise<void> {
     return loadEmbeddedResourcesAsync(bootConfig);
   });
 
-  const initialConnection = await initializeConnection(circuitHandlers, logger);
+  // pass options.configureSignalR to configure the signalR.HubConnectionBuilder
+  const configureSignalR = (options && options.configureSignalR) || null;
+  const initialConnection = await initializeConnection(configureSignalR, circuitHandlers, logger);
 
   const circuits = discoverPrerenderedCircuits(document);
   for (let i = 0; i < circuits.length; i++) {
@@ -55,7 +69,7 @@ async function boot(): Promise<void> {
       // We can't reconnect after a failure, so exit early.
       return false;
     }
-    const reconnection = await initializeConnection(circuitHandlers, logger);
+    const reconnection = await initializeConnection(configureSignalR, circuitHandlers, logger);
     const results = await Promise.all(circuits.map(circuit => circuit.reconnect(reconnection)));
 
     if (reconnectionFailed(results)) {
@@ -81,14 +95,20 @@ async function boot(): Promise<void> {
   }
 }
 
-async function initializeConnection(circuitHandlers: CircuitHandler[], logger: ILogger): Promise<signalR.HubConnection> {
+async function initializeConnection(configureSignalR: SignalRBuilder | null, circuitHandlers: CircuitHandler[], logger: ILogger): Promise<signalR.HubConnection> {
+
   const hubProtocol = new MessagePackHubProtocol();
   (hubProtocol as any).name = 'blazorpack';
-  const connection = new signalR.HubConnectionBuilder()
+
+  const connectionBuilder = new signalR.HubConnectionBuilder()
     .withUrl('_blazor')
-    .withHubProtocol(hubProtocol)
-    .configureLogging(signalR.LogLevel.Information)
-    .build();
+    .withHubProtocol(hubProtocol);
+
+  if (configureSignalR) {
+    configureSignalR(connectionBuilder);
+  }
+
+  const connection = connectionBuilder.build();
 
   connection.on('JS.BeginInvokeJS', DotNet.jsCallDispatcher.beginInvokeJSFromDotNet);
   connection.on('JS.RenderBatch', (browserRendererId: number, batchId: number, batchData: Uint8Array) => {
@@ -131,4 +151,7 @@ function unhandledError(connection: signalR.HubConnection, err: Error, logger: I
   }
 }
 
-boot();
+window['Blazor'].start = boot;
+if (shouldAutoStart()) {
+  boot();
+}
--- a/src/Components/Browser.JS/src/BootCommon.ts
+++ b/src/Components/Browser.JS/src/BootCommon.ts
@@ -37,3 +37,10 @@ interface BootJsonData {
   jsReferences: string[];
   linkerEnabled: boolean;
 }
+
+// Tells you if the script was added without <script src="..." autostart="false"></script>
+export function shouldAutoStart() {
+  return document &&
+    document.currentScript &&
+    document.currentScript.getAttribute('autostart') !== 'false';
+}
--- a/src/Components/Browser.JS/src/GlobalExports.ts
+++ b/src/Components/Browser.JS/src/GlobalExports.ts
@@ -1,8 +1,6 @@
-import { platform } from './Environment';
 import { navigateTo, internalFunctions as uriHelperInternalFunctions } from './Services/UriHelper';
 import { internalFunctions as httpInternalFunctions } from './Services/Http';
 import { attachRootComponentToElement } from './Rendering/Renderer';
-import { Pointer } from './Platform/Platform';
 
 // Make the following APIs available in global scope for invocation from JS
 window['Blazor'] = {
--- a/src/Components/test/testassets/ComponentsApp.Server/Pages/Index.cshtml
+++ b/src/Components/test/testassets/ComponentsApp.Server/Pages/Index.cshtml
@@ -15,7 +15,13 @@
 
     <app>@(await Html.RenderComponentAsync<App>(new { Name="Guest" }))</app>
 
-    <script src="_framework/components.server.js"></script>
-
+    <script src="_framework/components.server.js" autostart="false"></script>
+    <script>
+        Blazor.start({
+            configureSignalR: function (builder) {
+                builder.configureLogging(2); // LogLevel.Information
+            }
+        });
+    </script>
 </body>
 </html>
--- a/src/Components/test/testassets/TestServer/Pages/PrerenderedHost.cshtml
+++ b/src/Components/test/testassets/TestServer/Pages/PrerenderedHost.cshtml
@@ -15,15 +15,11 @@
         interactive states, we only load the .js file when told to.
     *@
     <hr />
-    <button id="load-boot-script" onclick="loadBootScript(event)">Load boot script</button>
-    <script>
-    function loadBootScript(event) {
-        event.srcElement.disabled = true;
-        var scriptElem = document.createElement('script');
-        scriptElem.src = '_framework/components.server.js';
-        document.body.appendChild(scriptElem);
-    }
 
+    <button id="load-boot-script" onclick="Blazor.start()">Load boot script</button>
+
+    <script src="_framework/components.server.js" autostart="false"></script>
+    <script>
     // Used by InteropOnInitializationComponent
     function setElementValue(element, newValue) {
       element.value = newValue;
--- a/src/Components/Browser.JS/src/Boot.WebAssembly.ts
+++ b/src/Components/Browser.JS/src/Boot.WebAssembly.ts
@@ -6,9 +6,17 @@ import { getAssemblyNameFromUrl } from './Platform/Url';
 import { renderBatch } from './Rendering/Renderer';
 import { SharedMemoryRenderBatch } from './Rendering/RenderBatch/SharedMemoryRenderBatch';
 import { Pointer } from './Platform/Platform';
-import { fetchBootConfigAsync, loadEmbeddedResourcesAsync } from './BootCommon';
+import { fetchBootConfigAsync, loadEmbeddedResourcesAsync, shouldAutoStart } from './BootCommon';
 
-async function boot() {
+let started = false;
+
+async function boot(options?: any) {
+
+  if (started) {
+    throw new Error('Blazor has already started.');
+  }
+  started = true;
+
   // Configure environment for execution under Mono WebAssembly with shared-memory rendering
   const platform = Environment.setPlatform(monoPlatform);
   window['Blazor'].platform = platform;
@@ -43,4 +51,7 @@ async function boot() {
   platform.callEntryPoint(mainAssemblyName, bootConfig.entryPoint, []);
 }
 
-boot();
+window['Blazor'].start = boot;
+if (shouldAutoStart()) {
+  boot();
+}
--- a/src/Components/Components.sln
+++ b/src/Components/Components.sln
@@ -210,6 +210,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution 
 		..\..\.editorconfig = ..\..\.editorconfig
 	EndProjectSection
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.AspNetCore.Mvc.Components.Prerendering", "..\Mvc\Mvc.Components.Prerendering\src\Microsoft.AspNetCore.Mvc.Components.Prerendering.csproj", "{3A4132B6-60DA-43A0-8E7B-4BF346F3247C}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
@@ -1324,6 +1326,18 @@ Global
 		{9088E4E4-B855-457F-AE9E-D86709A5E1F4}.Release|x64.Build.0 = Debug|Any CPU
 		{9088E4E4-B855-457F-AE9E-D86709A5E1F4}.Release|x86.ActiveCfg = Debug|Any CPU
 		{9088E4E4-B855-457F-AE9E-D86709A5E1F4}.Release|x86.Build.0 = Debug|Any CPU
+		{3A4132B6-60DA-43A0-8E7B-4BF346F3247C}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{3A4132B6-60DA-43A0-8E7B-4BF346F3247C}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{3A4132B6-60DA-43A0-8E7B-4BF346F3247C}.Debug|x64.ActiveCfg = Debug|Any CPU
+		{3A4132B6-60DA-43A0-8E7B-4BF346F3247C}.Debug|x64.Build.0 = Debug|Any CPU
+		{3A4132B6-60DA-43A0-8E7B-4BF346F3247C}.Debug|x86.ActiveCfg = Debug|Any CPU
+		{3A4132B6-60DA-43A0-8E7B-4BF346F3247C}.Debug|x86.Build.0 = Debug|Any CPU
+		{3A4132B6-60DA-43A0-8E7B-4BF346F3247C}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{3A4132B6-60DA-43A0-8E7B-4BF346F3247C}.Release|Any CPU.Build.0 = Release|Any CPU
+		{3A4132B6-60DA-43A0-8E7B-4BF346F3247C}.Release|x64.ActiveCfg = Release|Any CPU
+		{3A4132B6-60DA-43A0-8E7B-4BF346F3247C}.Release|x64.Build.0 = Release|Any CPU
+		{3A4132B6-60DA-43A0-8E7B-4BF346F3247C}.Release|x86.ActiveCfg = Release|Any CPU
+		{3A4132B6-60DA-43A0-8E7B-4BF346F3247C}.Release|x86.Build.0 = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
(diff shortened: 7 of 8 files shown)


src/Components/test/testassets/ComponentsApp.Server/Pages/Index.cshtml
```diff
@@ -14,6 +14,15 @@
...
 
+    <script type="text/javascript">
+        window.Blazor = {
+            callbacks: {
+                configureSignalR: function (builder) {
+                    builder.configureLogging(0);
+                }
+            }
+        }
+    </script>
```
rynowak: A few things to note:

This has to happen before `components.server.js` so that it can be used to configure signalR - we create the connection immediately when `components.server.js` is included.

I made this a callback on `signalR.HubConnectionBuilder` because there might be more things on the signal r connection that are useful. Then we won't have to add more API for dealing with signal r.

I made `window.Blazor.callbacks` its own object so we can treat it as opaque see `GlobalExports.ts`. 

`configureLogging` takes a .ts enum - however we don't have the signal r stuff exported from our script. I think we we end up using this pattern we will want to resolve this somehow, possibly by signal r allowing a string.

src/Components/Browser.JS/src/Boot.Server.ts
```diff
@@ -29,7 +32,8 @@ async function boot() {
   );
 
   window['Blazor'].reconnect = async () => {
-    const reconnection = await initializeConnection(circuitHandlers);
+    // Intentionally re-evaluating configureSignalR so it can change while the app is running.
+    const reconnection = await initializeConnection(window['Blazor'].callbacks && window['Blazor'].callbacks.configureSignalR, circuitHandlers);
```
rynowak: what I had in mind for this is that you could do this in the console if you're dev-ing and see a problem. It would be a little jank mode because you have to type out a function. Does this make sense or should we just evaluate once?

src/Components/Browser.JS/src/GlobalExports.ts
```diff
@@ -1,13 +1,16 @@
-import { platform } from './Environment';
 import { navigateTo, internalFunctions as uriHelperInternalFunctions } from './Services/UriHelper';
 import { internalFunctions as httpInternalFunctions } from './Services/Http';
 import { attachRootComponentToElement } from './Rendering/Renderer';
-import { Pointer } from './Platform/Platform';
+
+// Accept the following callbacks that need to be set before we boot.
+const existing = window['Blazor'];
```
javiercn: The idiomatic way to do this in JS is window['Blazor'] = window['Blazor'] || {};

javiercn:
In general I would hope our approach here would be based on modern javascript (modules) instead of monkeypatching. We should simply allow people to decide when and how to call blazor.boot and have that be the way they can pass parameters and configuration options to it.

```javascript
import { boot } from 'components.server.js'

let options = {
    ...
};

boot(options);
```

This is available already in all major browsers:
https://www.caniuse.com/#feat=es6-module

And it can definitely be transpiled to (ES5/ES3) with pretty much anything.

If there is a contract here between the framework and our customers we should formalize it properly instead of hacking around at adding properties to the window object.

src/Components/Blazor/testassets/StandaloneApp/Pages/Counter.razor
```diff
@@ -1,8 +1,13 @@
...
+@page "/counter"
 
 <h1>Counter</h1>
 
-<p>Current count: @currentCount</p>
+<p>Current count:
+    @currentCount
+    It
+    is
+    Wednesday my dudes
```
SteveSandersonMS: I'm guessing... this is not really going in? Probably you were experimenting with linebreaks for @ajaybhargavb's work.

src/Components/test/testassets/ComponentsApp.App/Pages/Counter.razor
```diff
@@ -1,8 +1,14 @@
...
 
 <h1>Counter</h1>
 
-<p>Current count: @currentCount</p>
+<p>
+    Current count:
+    @currentCount
+    It
+    is
+    Wednesday my dudes
```
SteveSandersonMS: Same

SteveSandersonMS:
I agree with @javiercn that this might be the point at which we should embrace having the developer call some JS function to start the connection. I don't think we need to take a hard dependency on ES modules to do that, though it would be compatible with them for people who want that.

Two simple possibilities:

### 1. Literally a function call

```html
<script src="_framework/components.server.js"></script>
<script>Blazor.start();</script>
```

`Blazor.start` could optionally accept a config param, e.g.,

```js
Blazor.start({ signalR: { ... } });
```

Pros:

 * Totally standard and easy to understand
 * Meshes well with the likely future where we publish our `Microsoft.AspNetCore.Components.Browser.JS` as an NPM package that embeds its `.ts` sources. Developers can reference that package from their own JS/TS modules can call the exported `Blazor.connect` from anywhere in their code
 * Same API can be implemented in both `components.server.js` and `components.webassembly.js`, so developers can control the timing of the WebAssembly startup process if they want. The WebAssembly one would ignore options it doesn't care for (e.g., `signalR: { ... }`).

Cons:

 * An extra line of code in all the template host pages

### 2. Same, but with shorthand for implicit connection

```html
<script src="_framework/components.server.js" start="auto"></script>
```

The logic can check for the presence of the special attribute on its own script tag and, if present, implicitly call its own `Blazor.start()`. It's just a way of not requiring the extra line in the template.

I don't feel strongly about this being super important.

src/Components/Blazor/testassets/StandaloneApp/Pages/Counter.razor
rynowak: Yeah this is leftover from other stuff. Its obviously not right because it's not wednesday anymore

rynowak:
@SteveSandersonMS  for option 2 - what would you think of something like:

```
<script src="_framework/components.server.js" start></script>
```

I think having `start="auto"` is a little unclear because it implies that there are other meaningful options for the attribute value and I don't think we have any mind. Let me know if this isn't a good idea!

SteveSandersonMS:
TBH I’d be totally happy with just doing option 1, but if you do prefer 2, then yes a minimised attribute makes sense.

rynowak:
I did option 1 and 2! Since the example that was discussed here was **opt-in** to auto-start, I went with that, but while was doing it I wondered if the inverse is better because folks won't have to change their code, and it might be less common to want manual-start. Happy to discuss and change this based on feedback.

Current (opt-in):
```
<script src="..." start></script>

OR

<script src="..."></script>
<script src="..." start>Blazor.start(...)</script>
```


With opt-out:
```
<script src="..."></script>

OR

<script src="..." start="false"></script>
<script src="..." start>Blazor.start(...)</script>
```

src/Components/Browser.JS/src/Boot.Server.ts
```diff
@@ -5,11 +5,19 @@ import { MessagePackHubProtocol } from '@aspnet/signalr-protocol-msgpack';
...
 
-async function boot() {
+let started = false;
+
+async function boot(options?: any) {
+
+  if (started) {
+    throw new Error('Blazor has already started.');
+  }
+  started = true;
```
rynowak: this seemed worth doing because you can now manually start things. LMK if you think this is bad somehow.

src/Components/Browser.JS/src/Boot.Server.ts
```diff
@@ -5,11 +5,19 @@ import { MessagePackHubProtocol } from '@aspnet/signalr-protocol-msgpack';
...
+
   const circuitHandlers: CircuitHandler[] = [ new AutoReconnectCircuitHandler() ];
```
rynowak: I took no steps to address the fact that we still have mutable global state for other things like reconnect. LMK if you think that fits in nicely with this PR.

src/Components/Browser.JS/src/Boot.Server.ts
```diff
@@ -5,11 +5,19 @@ import { MessagePackHubProtocol } from '@aspnet/signalr-protocol-msgpack';
...
   window['Blazor'].circuitHandlers = circuitHandlers;
```
rynowak: I took no steps to address the fact that this script relies on mutating global state when `boot()` runs. It seems a little odd to me. LMK if you think this needs to be addressed as part of this change.

src/Components/Browser.JS/src/BootCommon.ts
```diff
@@ -37,3 +37,13 @@ interface BootJsonData {
...
 }
+
+// calls Blazor.start() if the script was added with <script src="..." start></script>
+export function autoStartIfApplicable(start: (options?: any)=> Promise<void>) {
+  if (document &&
+    document.currentScript &&
+    document.currentScript.hasAttribute("start") &&
+    document.currentScript.getAttribute("start") !== "false") {
+      start();
+  }
```
rynowak: LMK if there's a more idiomatic way to do this. i is noob

src/Components/Browser.JS/src/GlobalExports.ts
```diff
@@ -1,8 +1,6 @@
-import { platform } from './Environment';
 import { navigateTo, internalFunctions as uriHelperInternalFunctions } from './Services/UriHelper';
 import { internalFunctions as httpInternalFunctions } from './Services/Http';
 import { attachRootComponentToElement } from './Rendering/Renderer';
-import { Pointer } from './Platform/Platform';
```
rynowak: Removed unused

src/Components/test/testassets/ComponentsApp.Server/Pages/Index.cshtml
```diff
@@ -15,5 +15,12 @@
     <app>@(await Html.RenderComponentAsync<App>())</app>
 
     <script src="_framework/components.server.js"></script>
+    <script>
+        Blazor.start({
+            configureSignalR: function (builder) {
+                builder.configureLogging(2); // LogLevel.Information
```
rynowak: I don't have any cool ideas about making this better inside Blazor. @anurse  and I discussed the possibility of accepting strings in this API but that still feels lame.

Other ideas?

src/Components/test/testassets/TestServer/Pages/PrerenderedHost.cshtml
```diff
@@ -19,6 +19,9 @@
     function loadBootScript(event) {
         event.srcElement.disabled = true;
         var scriptElem = document.createElement('script');
+        scriptElem.onload = function () {
+            Blazor.start();
+        };
```
rynowak: I tried a few things to make dynamic loading of this script work for the test and this was the one that worked. LMK if there is a better way.

src/Components/Browser.JS/src/Boot.Server.ts
SteveSandersonMS: LGTM.

src/Components/Browser.JS/src/Boot.Server.ts
SteveSandersonMS: If you wanted to refactor those other pieces to match your new pattern then I'm sure that would work well, but if you prefer to consider that a separate future work item that would be fine too.

src/Components/Browser.JS/src/Boot.Server.ts
```diff
@@ -18,7 +26,9 @@ async function boot() {
     return loadEmbeddedResourcesAsync(bootConfig);
   });
 
-  const initialConnection = await initializeConnection(circuitHandlers);
+  // pass options.configureSignalR to configure the signalR.HubConnectionBuilder
+  const configureSignalR = (options || {}).configureSignalR
```
SteveSandersonMS: Nit: missing semicolon

SteveSandersonMS: Other nit: `(options && options.configureSignalR) || null` is fractionally more idiomatic, and avoids instantiating an empty object that will then just be discarded.

src/Components/Browser.JS/src/Boot.Server.ts
```diff
@@ -41,12 +51,16 @@ async function boot() {
...
-  const connection = new signalR.HubConnectionBuilder()
+async function initializeConnection(configureSignalR: (builder: signalR.HubConnectionBuilder) => void, circuitHandlers: CircuitHandler[]): Promise<signalR.HubConnection> {
+  const connectionBuilder = new signalR.HubConnectionBuilder()
     .withUrl('_blazor')
-    .withHubProtocol(new MessagePackHubProtocol())
-    .configureLogging(signalR.LogLevel.Information)
-    .build();
+    .withHubProtocol(new MessagePackHubProtocol());
+
+  if (configureSignalR) {
```
SteveSandersonMS: This is suspicious. It implies we're not running TypeScript in strict mode. If we were, it would not be possible for this to be null, as the type isn't nullable. Looking back, I'm surprised it allows L30 - is it implictly interpreting `configureSignalR` as type `any`?

SteveSandersonMS: Ideally I'd expect there to be a type defined at the top of the file like:

```
type SignalRBuilder = (builder: signalR.HubConnectionBuilder) => void;
```

... and then this function's signature would be:

```
async function initializeConnection(configureSignalR: SignalRBuilder | null, circuitHandlers: CircuitHandler[]): Promise<signalR.HubConnection> { ... }
```

... so it's explicit about what's going on.

src/Components/Browser.JS/src/Boot.Server.ts
javiercn: We should file a bug for it and not do anything right now. In the end I will expect for boot to take in some sort of options. Or to have a builder pattern to configure things a la signalr

src/Components/Browser.JS/src/BootCommon.ts
SteveSandersonMS: The chain of `if`s looks fine. However:

 * Why take a callback as a param, given that you're not really making any use of its args or return value? It's declared as accepting `options` and returning a promise, but the usage here supports neither. Instead maybe replace `autoStartIfApplicable` with `shouldAutoStart(): boolean`, and have the upstream code do what it wants with the bool result.
 * Nit: quotes should be single quotes. Javier is adding tslint in his separate PR.

src/Components/test/testassets/ComponentsApp.Server/Pages/Index.cshtml
SteveSandersonMS: I think it's fine

People who consume these libraries via TS imports will be able to pass SignalR enum values. People who consume them as a `<script>` tag won't, but that's just the nature of the SignalR API here (SignalR presumably documents what the accepted values are).

src/Components/test/testassets/TestServer/Pages/PrerenderedHost.cshtml
SteveSandersonMS: Instead of dynamically inserting the script tag, it now seems we can hard-code the script tag into the document as normal, just don't put the `start` on it. Then `loadBootScript` is replaced just by calling `Blazor.start()`.

This is nice because it shows we have test coverage for this supported usage pattern.

src/Components/Browser.JS/src/Boot.Server.ts
rynowak: Lets discuss at standup. I think it's somewhat worthwhile to try and change the pattern we're using once. If this PR is going to introduce the `start()` function that it might make sense to establish the pattern we want to use for everything in this preview.

src/Components/Browser.JS/src/Boot.Server.ts
rynowak: This is that PR though.

src/Components/test/testassets/TestServer/Pages/PrerenderedHost.cshtml
rynowak: solid plan

src/Components/Browser.JS/src/Boot.Server.ts
SteveSandersonMS: I'd be open to doing an all-over cleanup pass over the `.JS` project after preview 4. I'm not sure we'll have time to change all the public-facing JS APIs prior to preview 4 lockdown, but I guess it depends on exact cutoff dates.

rynowak: We're slowing down feature developement **so that** we can polish things that need polishing. Anything that's part of the making the experience consistent and well-thought-through is fair game. 

We could just as well decide to punt this particular item.

SteveSandersonMS:
> So to be clear, we auto-start today, so in preview 4 everyone will have to change their codez to either call start manually, or add the start attribute.

OK, maybe I'm changing my mind on this now. There might be a different way to factor this to avoid the drawback of an opt-out autostart. Let's discuss.

SteveSandersonMS:
We might be better doing opt-out auto-start, e.g., `<script src='...' autostart='false'>` so as to avoid having a strange concept in the 95% of cases where you do just want the app to start up.

Then there would be options for people who want to pass config info:

 * If you're just using `<script>` tags, use `autostart=false` and call `Blazor.boot(myOptions)`
 * If you're importing our library into some larger TS/JS build system, we will expose the `boot` function from a module you can import without implicitly invoking it. Then you can freely call `boot(myOptions)`. For this to work, we need to change `boot.server.ts` to be a trivial entrypoint file that just imports `boot` from some other module and invokes it. This makes sense because its name actually is "**boot**.server.ts".

src/Components/test/testassets/ComponentsApp.Server/Pages/Index.cshtml
anurse: I don't think we document the constant values though. I think it's reasonable to have the SignalR APIs accept strings though. If that's filed somewhere, we can continue with this for now.

src/Components/Browser.JS/src/Boot.Server.ts
```diff
@@ -41,12 +56,16 @@ async function boot() {
   circuitHandlers.forEach(h => h.onConnectionUp && h.onConnectionUp());
 }
 
-async function initializeConnection(circuitHandlers: CircuitHandler[]): Promise<signalR.HubConnection> {
-  const connection = new signalR.HubConnectionBuilder()
+async function initializeConnection(configureSignalR: SignalRBuilder | null, circuitHandlers: CircuitHandler[]): Promise<signalR.HubConnection> {
```
javiercn: null is not cool in the JS world. May I suggest you use SignalRBuilder | undefined? Or better yet. The more idiomatic js way of doing this is through the destructuring syntax `({ configureSignalr: SignalRBuilder?, circuitHandlers: CircuitHandler[]})` (You can do this on JS parameter lists)

src/Components/test/testassets/ComponentsApp.Server/Pages/Index.cshtml
rynowak: logged an issue

rynowak:
> If you're importing our library into some larger TS/JS build system, we will expose the boot function from a module you can import without implicitly invoking it. Then you can freely call boot(myOptions). For this to work, we need to change boot.server.ts to be a trivial entrypoint file that just imports boot from some other module and invokes it. This makes sense because its name actually is "boot.server.ts".

OK. I'm planning to move forward without this part of the change for now. Hope that's alright!

src/Components/Browser.JS/src/Boot.Server.ts
rynowak: Does this matter though? How would an undefined get in here? This isn't something users call directly.

javiercn: I might have been throw off by the typing. You are right it doesn't matter, it's just the typing is not idiomatic.

rynowak:
@SteveSandersonMS @javiercn - any more feedback on this?

src/Components/Browser.JS/src/Boot.Server.ts
```diff
@@ -5,11 +5,24 @@ import { MessagePackHubProtocol } from '@aspnet/signalr-protocol-msgpack';
...
 import { renderBatch } from './Rendering/Renderer';
-import { fetchBootConfigAsync, loadEmbeddedResourcesAsync } from './BootCommon';
+import { fetchBootConfigAsync, loadEmbeddedResourcesAsync, shouldAutoStart } from './BootCommon';
 import { CircuitHandler } from './Platform/Circuits/CircuitHandler';
 import { AutoReconnectCircuitHandler } from './Platform/Circuits/AutoReconnectCircuitHandler';
 
-async function boot() {
+type SignalRBuilder = (builder: signalR.HubConnectionBuilder) => void;
+type BlazorOptions = {
+  configureSignalR: SignalRBuilder | null,
```
javiercn: configureSignalr?: SignalRBuilder

src/Components/Browser.JS/src/Boot.Server.ts
```diff
@@ -5,11 +5,24 @@ import { MessagePackHubProtocol } from '@aspnet/signalr-protocol-msgpack';
...
+};
+
+let started = false;
+
+async function boot(options?: BlazorOptions | null) {
```
javiercn: `| null` should not be needed

src/Components/Browser.JS/src/Boot.Server.ts
```diff
@@ -18,7 +31,9 @@ async function boot() {
     return loadEmbeddedResourcesAsync(bootConfig);
   });
 
-  const initialConnection = await initializeConnection(circuitHandlers);
+  // pass options.configureSignalR to configure the signalR.HubConnectionBuilder
+  const configureSignalR = (options && options.configureSignalR) || null;
```
javiercn: I don't think the parenthesis are required here.

src/Components/Browser.JS/src/Boot.Server.ts
javiercn: Should we mark boot as export?

src/Components/Browser.JS/src/Boot.Server.ts
rynowak: It still helps with clarity. I don't see a reason to take it out.

src/Components/Browser.JS/src/Boot.Server.ts
rynowak: out of scope for this PR. I'm not trying to solve the scenario of using our scripts as a module. Steve created an issue for it.

src/Components/Browser.JS/src/Boot.Server.ts
```diff
@@ -5,11 +5,24 @@ import { MessagePackHubProtocol } from '@aspnet/signalr-protocol-msgpack';
...
 import { internalFunctions as uriHelperFunctions } from './Services/UriHelper';
 import { renderBatch } from './Rendering/Renderer';
-import { fetchBootConfigAsync, loadEmbeddedResourcesAsync } from './BootCommon';
+import { fetchBootConfigAsync, loadEmbeddedResourcesAsync, shouldAutoStart } from './BootCommon';
 import { CircuitHandler } from './Platform/Circuits/CircuitHandler';
 import { AutoReconnectCircuitHandler } from './Platform/Circuits/AutoReconnectCircuitHandler';
 
-async function boot() {
+type SignalRBuilder = (builder: signalR.HubConnectionBuilder) => void;
+type BlazorOptions = {
```
javiercn: Its recommended to use interfaces over type aliases. We should probably mark this as export too.
https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/prefer-interface.md

javiercn (approved):
Looks good to me, I would change the type -> interface for the BlazorOptions. The primary usage of `type` is aliasing and simplifying other type definitions, basic stuff should be done with interface and class.

src/Components/Browser.JS/src/Boot.Server.ts
rynowak: export is out of scope. I'm not solving the scenario of using our startup from a module in this PR.

SteveSandersonMS:
The reason given for not using type alias doesn’t sound applicable to function signature definitions (equivalent of C# delegate type).

SteveSandersonMS:
Oh wait that’s not what you were doing anyway. Please ignore me!

javiercn:
> The reason given for not using type alias doesn’t sound applicable to function signature definitions (equivalent of C# delegate type).

I was pointing to blazor options. For function definitions is totally fine

SteveSandersonMS (approved):
Looks really good!

rynowak:
@guardrex @danroth27 we want to make sure to document this for preview4. See code sample in the header.

Issue #8902, opened 2019-03-29
Support strongly typed builders with IEndpointConventionBuilder


JamesNK:
Map extensions that hang off `IEndpointRouteBuilder` return `IEndpointConventionBuilder`. This is fine and good.

Can we imagine scenarios where someone wants to return a more strongly typed builder? For example, from gRPC's `MapGrpcService` I return an `GrpcEndpointRouteBuilder`. It implements convention builder but there is also some gRPC specific stuff on there, and gRPC extension methods.

```cs
public class GrpcEndpointRouteBuilder : IEndpointConventionBuilder
{
    public GrpcEndpointRouteBuilder ConfigureCrazyGrpcStuff(Action a)
    {
        a();
        return this;
    }
}
```

Today all our `IEndpointConventionBuilder` extension methods return `IEndpointConventionBuilder` which would make using a strongly typed builder difficult.

```cs
routes.MapGrpcService<MyService>()
    .RequireAuthorization("policy")
    .ConfigureCrazyGrpcStuff(() => {}) // error, reference is no longer the gRPC builder
```

Should our extension methods be more like:

```cs
public TBuilder RequireAuthorization<TBuilder>(TBuilder builder, string policyName) where TBuilder : IEndpointConventionBuilder
{
    // stuff
    return builder;
}
```

// @davidfowl @rynowak

rynowak:
It's absolutely intended that you return a strongly-typed builder for a case like this.

We do need to update these methods to have a `T`. That was part of the original design and prototypes but I guess it got lost? 

I agree 100% and this was always the plan.

rynowak:
I would also strongly encourage that you return a CLASS and not an interface so it can grow.

davidfowl:
I dunno, the generics make me sad.

rynowak:
Then I guess you're sad? The builders returned from endpoint routing methods are the application builder replacements. The `IEndpointConventionBuilder` is there to be implemented for all of the *common* stuff.

davidfowl:
I guess I should write a bit more about why it makes me sad. The problem with generics is that they instantly become hard to use once type inference fails. That happens the moment you have a second generic argument that needs to be specified. Maybe we'll get lucky and won't end up with multiple generic arguments but I am suspicious of that.

rynowak:
Yeah agreed. So the example with `MapHub` should change to use typeof instead of non-generic.

mkArtakMSFT:
[...]

PR #8906, opened 2019-03-29
Change IEndpointConventionBuilder extension methods to use generic argument

JamesNK:
Addresses #8902

### DIFF
--- a/build/sources.props
+++ b/build/sources.props
@@ -14,6 +14,7 @@
       https://dotnetfeed.blob.core.windows.net/aspnet-entityframeworkcore/index.json;
       https://dotnetfeed.blob.core.windows.net/aspnet-aspnetcore-tooling/index.json;
       https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json;
+      https://grpc.jfrog.io/grpc/api/nuget/v3/grpc-nuget-dev;
       https://api.nuget.org/v3/index.json;
     </RestoreSources>
     <!-- TODO remove this once we remove usages of Internal.AspNetCore.Sdk. -->
--- a/src/Components/Server/ref/Microsoft.AspNetCore.Components.Server.netcoreapp3.0.cs
+++ b/src/Components/Server/ref/Microsoft.AspNetCore.Components.Server.netcoreapp3.0.cs
@@ -9,10 +9,10 @@ namespace Microsoft.AspNetCore.Builder
     }
     public static partial class ComponentEndpointRouteBuilderExtensions
     {
-        public static Microsoft.AspNetCore.SignalR.ComponentEndpointConventionBuilder MapBlazorHub(this Microsoft.AspNetCore.Routing.IEndpointRouteBuilder endpoints) { throw null; }
-        public static Microsoft.AspNetCore.SignalR.ComponentEndpointConventionBuilder MapBlazorHub(this Microsoft.AspNetCore.Routing.IEndpointRouteBuilder endpoints, System.Type componentType, string selector, string path) { throw null; }
-        public static Microsoft.AspNetCore.SignalR.ComponentEndpointConventionBuilder MapBlazorHub<TComponent>(this Microsoft.AspNetCore.Routing.IEndpointRouteBuilder endpoints, string selector) where TComponent : Microsoft.AspNetCore.Components.IComponent { throw null; }
-        public static Microsoft.AspNetCore.SignalR.ComponentEndpointConventionBuilder MapBlazorHub<TComponent>(this Microsoft.AspNetCore.Routing.IEndpointRouteBuilder endpoints, string selector, string path) where TComponent : Microsoft.AspNetCore.Components.IComponent { throw null; }
+        public static Microsoft.AspNetCore.Components.Server.ComponentEndpointConventionBuilder MapBlazorHub(this Microsoft.AspNetCore.Routing.IEndpointRouteBuilder endpoints) { throw null; }
+        public static Microsoft.AspNetCore.Components.Server.ComponentEndpointConventionBuilder MapBlazorHub(this Microsoft.AspNetCore.Routing.IEndpointRouteBuilder endpoints, System.Type componentType, string selector, string path) { throw null; }
+        public static Microsoft.AspNetCore.Components.Server.ComponentEndpointConventionBuilder MapBlazorHub<TComponent>(this Microsoft.AspNetCore.Routing.IEndpointRouteBuilder endpoints, string selector) where TComponent : Microsoft.AspNetCore.Components.IComponent { throw null; }
+        public static Microsoft.AspNetCore.Components.Server.ComponentEndpointConventionBuilder MapBlazorHub<TComponent>(this Microsoft.AspNetCore.Routing.IEndpointRouteBuilder endpoints, string selector, string path) where TComponent : Microsoft.AspNetCore.Components.IComponent { throw null; }
     }
 }
 namespace Microsoft.AspNetCore.Components.Browser.Rendering
@@ -29,6 +29,11 @@ namespace Microsoft.AspNetCore.Components.Server
         public CircuitOptions() { }
         public System.TimeSpan DisconnectedCircuitRetentionPeriod { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public int MaxRetainedDisconnectedCircuits { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
+    }
+    public sealed partial class ComponentEndpointConventionBuilder : Microsoft.AspNetCore.Builder.IEndpointConventionBuilder, Microsoft.AspNetCore.SignalR.IHubEndpointConventionBuilder
+    {
+        internal ComponentEndpointConventionBuilder() { }
+        public void Add(System.Action<Microsoft.AspNetCore.Builder.EndpointBuilder> convention) { }
     }
     public sealed partial class ComponentHub : Microsoft.AspNetCore.SignalR.Hub
     {
@@ -89,14 +94,6 @@ namespace Microsoft.AspNetCore.Components.Server.Circuits
         protected override void NavigateToCore(string uri, bool forceLoad) { }
         [Microsoft.JSInterop.JSInvokableAttribute("NotifyLocationChanged")]
         public static void NotifyLocationChanged(string uriAbsolute) { }
-    }
-}
-namespace Microsoft.AspNetCore.SignalR
-{
-    public sealed partial class ComponentEndpointConventionBuilder : Microsoft.AspNetCore.Builder.IEndpointConventionBuilder, Microsoft.AspNetCore.SignalR.IHubEndpointConventionBuilder
-    {
-        internal ComponentEndpointConventionBuilder() { }
-        public void Add(System.Action<Microsoft.AspNetCore.Builder.EndpointBuilder> convention) { }
     }
 }
 namespace Microsoft.Extensions.DependencyInjection
--- a/src/Components/Server/ref/Microsoft.AspNetCore.Components.Server.netcoreapp3.0.cs
+++ b/src/Components/Server/ref/Microsoft.AspNetCore.Components.Server.netcoreapp3.0.cs
@@ -5,7 +5,7 @@ namespace Microsoft.AspNetCore.Builder
 {
     public static partial class ComponentEndpointConventionBuilderExtensions
     {
-        public static TBuilder AddComponent<TBuilder>(this TBuilder builder, System.Type componentType, string selector) where TBuilder : Microsoft.AspNetCore.SignalR.HubEndpointConventionBuilder { throw null; }
+        public static TBuilder AddComponent<TBuilder>(this TBuilder builder, System.Type componentType, string selector) where TBuilder : Microsoft.AspNetCore.SignalR.IHubEndpointConventionBuilder { throw null; }
     }
     public static partial class ComponentEndpointRouteBuilderExtensions
     {
@@ -93,9 +93,10 @@ namespace Microsoft.AspNetCore.Components.Server.Circuits
 }
 namespace Microsoft.AspNetCore.SignalR
 {
-    public partial class ComponentEndpointConventionBuilder : Microsoft.AspNetCore.SignalR.HubEndpointConventionBuilder
+    public sealed partial class ComponentEndpointConventionBuilder : Microsoft.AspNetCore.Builder.IEndpointConventionBuilder, Microsoft.AspNetCore.SignalR.IHubEndpointConventionBuilder
     {
-        internal ComponentEndpointConventionBuilder() : base (default(Microsoft.AspNetCore.Builder.IEndpointConventionBuilder)) { }
+        internal ComponentEndpointConventionBuilder() { }
+        public void Add(System.Action<Microsoft.AspNetCore.Builder.EndpointBuilder> convention) { }
     }
 }
 namespace Microsoft.Extensions.DependencyInjection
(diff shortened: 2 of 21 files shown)


src/Components/test/testassets/TestServer/Startup.cs
```diff
@@ -56,7 +56,7 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
 
                 subdirApp.UseEndpoints(endpoints =>
                 {
-                    endpoints.MapHub<ComponentHub>(ComponentHub.DefaultPath).AddComponent<Index>(selector: "root");
+                    endpoints.MapHub<ComponentHub>(ComponentHub.DefaultPath).AddComponent<IEndpointConventionBuilder, Index>(selector: "root");
```
JamesNK: We obviously don't want this.

Does `AddComponent` only make sense in the context of `MapHub`? Should `MapHub` return a custom builder type and `AddComponent` would then be an extension method of the custom builder type.

rynowak: yup.

JamesNK:
```
src\Http\Routing\ref\Microsoft.AspNetCore.Routing.netcoreapp3.0.cs(,): error : Generated code is not up to date in src/Http/Routing/ref/Microsoft.AspNetCore.Routing.netcoreapp3.0.cs.
src\Middleware\CORS\ref\Microsoft.AspNetCore.Cors.netcoreapp3.0.cs(,): error : Generated code is not up to date in src/Middleware/CORS/ref/Microsoft.AspNetCore.Cors.netcoreapp3.0.cs.
src\Security\Authorization\Policy\ref\Microsoft.AspNetCore.Authorization.Policy.netcoreapp3.0.cs(,): error : Generated code is not up to date in src/Security/Authorization/Policy/ref/Microsoft.AspNetCore.Authorization.Policy.netcoreapp3.0.cs.
```

Is there a way to automatically update reference files? (I think I missed the memo about them)

SteveSandersonMS:
@JamesNK https://github.com/aspnet/AspNetCore/blob/master/docs/ReferenceAssemblies.md

src/Components/test/testassets/TestServer/Startup.cs
JamesNK: Which would you prefer:

* `public static IHubEndpointConventionBuilder AddComponent<TComponent>(this TBuilder builder, string selector)` - still has the generic type but would potentially lose the exact builder type
* `public static TBuilder AddComponent<TBuilder>(this TBuilder builder, string selector, Type component) where TBuilder : IHubEndpointConventionBuilder` - requires `typeof(Component)` but keeps the exact builder type

`IHubEndpointConventionBuilder` would be returned by https://github.com/aspnet/AspNetCore/blob/658b37d2bd3b0da2c07e25b53fa5c97ce8d656d3/src/SignalR/server/SignalR/src/HubEndpointRouteBuilderExtensions.cs#L23

rynowak: I don't think either one has strong downsides. 

`IHubConventionBuilder` is already strongly typed so 1 is not a major concern.
People like generics because they make them feel smart so 2 has that drawback. 

I would lean towards 1 because I don't think the drawback will matter in practice. I'd also encourage teh signelRz to return a *sealed class* from `MapHub<>`

JamesNK:
@rynowak @davidfowl Removed generic `AddComponent`, added `HubEndpointConventionBuilder`

JamesNK:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=146435

aspnet-hello:

I've triaged the above build.

src/SignalR/server/SignalR/src/HubEndpointConventionBuilder.cs
```diff
@@ -0,0 +1,30 @@
...
+
+using System;
+using Microsoft.AspNetCore.Builder;
+
+namespace Microsoft.AspNetCore.SignalR
+{
+    /// <summary>
+    /// Builds conventions that will be used for customization of Hub <see cref="EndpointBuilder"/> instances.
+    /// </summary>
+    public class HubEndpointConventionBuilder : IEndpointConventionBuilder
```
rynowak: sealed 😀

rynowak: Well, it already has an internal constructor, so either way

rynowak (approved):

halter73 (approved):
nit: I notice you don't cover all the extension method overloads in the ChainedCall_ReturnedBuilderIsDerivedType tests.

rynowak:
> nit: I notice you don't cover all the extension method overloads in the ChainedCall_ReturnedBuilderIsDerivedType tests.

I don't really think these tests have much value anyway. They test what the compiler does 😆 If you were going to forget to write the method with a generic then you would also forget to add one of these tests.

halter73:
> I don't really think these tests have much value anyway. They test what the compiler does 😆

Fair.

JamesNK:
@davidfowl ping. Like to hear from you on this change before merging 😄

src/Components/Server/src/Builder/ComponentEndpointRouteBuilderExtensions.cs
```diff
@@ -18,7 +19,7 @@ public static class ComponentEndpointRouteBuilderExtensions
         /// </summary>
         /// <param name="endpoints">The <see cref="IEndpointRouteBuilder"/>.</param>
         /// <returns>The <see cref="IEndpointConventionBuilder"/>.</returns>
-        public static IEndpointConventionBuilder MapBlazorHub(this IEndpointRouteBuilder endpoints)
+        public static HubEndpointConventionBuilder MapBlazorHub(this IEndpointRouteBuilder endpoints)
```
davidfowl: Shouldn't this be a different concrete type that derives from `HubEndpointConventionBuilder`?

JamesNK: I didn't change it because the new builder type would be used by nothing.

We could make every framework return a strongly typed builder just in case: MVC, Razor pages, SignalR, Blazor, gRPC. Do we want to?

davidfowl: Isn't that the point of this change? Future proofing? It's the same reason we want to avoid returning IEndpointConventionBuilder right?

halter73: Couldn't you change MapBlazorHub and the like to return a more derived HubEndpointConventionBuilder later? Or would that be breaking? I don't think it would be since it's static.

davidfowl: Pretty sure that’s a breaking change

halter73: What code would break?

JamesNK: Any compiled code would throw a MethodNotFoundException because the signature has changed

halter73: That's what I was wondering. It's interesting the return type is used when looking up the method considering you cannot overload on return type.

halter73: I guess it makes sense as it avoids potentially uglier errors if the return type changed to a non-derived type though.

JamesNK: https://twitter.com/terrajobst/status/1115730579838787584

It is either the biggest co-incidence of all time or I think @davidfowl asked @terrajobst about it 😄

JamesNK: > Isn't that the point of this change? Future proofing? It's the same reason we want to avoid returning IEndpointConventionBuilder right?

Ok. I agree it provides future proofing, I just wanted to double check that we've thought through changing to a concrete type for each framework.

davidfowl: Yes, if we ever want extension methods that apply to components but not hubs we'll need it.

rynowak: We actually have an example of that already.

rynowak: As much as I would like to say it's unimportant because inheritance is bad.

davidfowl:
@JamesNK whats the deal with this PR?

JamesNK:
I'm going to update it today.

src/SignalR/server/SignalR/src/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("Microsoft.AspNetCore.Components.Server, PublicKey=0024000004800000940000000602000000240000525341310004000001000100f33a29044fa9d740c9b3213a93e57c84b472c84e0b8a0e1ae48e67a9f8f6de9d5f7f3d52ac23e48ac51801f1dc950abe901da34d2a9e3baadb141a17c77ef3c565dd5ee5054b91cf63bb3c6ab83f72ab3aafe93d0fc3c2348b764fafb0b1c0733de51459aeab46580384bf9d74c4e28164b7cde247f891ba07891c9d872ad2bb")]
```
JamesNK: ComponentEndpointConventionBuilder inherits from HubEndpointConventionBuilder and HubEndpointConventionBuilder has an internal constructor. Make SignalR server internals visible to Components server so it is callable.

Thoughts on this?

rynowak: ehhhhhhhhhhhhhhhhhhhhhhhhhhhhhh that seems super not good.

BrennanConroy: Why not make the constructor public?

JamesNK: Options here:
1. The current InternalsVisibleTo. It's gross, and introduces internal coupling
2. Make the constructor public. It's not a type we'd want a user to instantiate so a public constructor is not ideal. On the other hand, if Blazor has a reason to extend HubEndpointConventionBuilder then a future framework might have that same reason
3. Remove inheritance and change the AddComponent extension method to only work with ComponentEndpointConventionBuilder. I don't know enough about Blazor design to say whether this is desired.

What do y'all want?

src/Components/Server/ref/Microsoft.AspNetCore.Components.Server.netcoreapp3.0.cs
```diff
@@ -92,6 +91,13 @@ public partial class RemoteUriHelper : Microsoft.AspNetCore.Components.UriHelper
         public static void NotifyLocationChanged(string uriAbsolute) { }
     }
 }
+namespace Microsoft.AspNetCore.SignalR
+{
+    public partial class ComponentEndpointConventionBuilder : Microsoft.AspNetCore.SignalR.HubEndpointConventionBuilder
```
JamesNK: FYI I chose to not seal this because you can't have sealed type as a generic argument - https://stackoverflow.com/questions/1945085/why-cant-we-use-sealed-classes-as-generic-constraints

It makes sense if the type never changes, but not allowing the type as a generic parameter doesn't allow for the type one day changing to unsealed.

JamesNK: @rynowak Change sealed builders at https://github.com/aspnet/AspNetCore/pull/9657 to unsealed? An internal constructor achieves the same result.

rynowak: that would be fine.

src/SignalR/server/SignalR/src/Properties/AssemblyInfo.cs
rynowak: My concern with the constructor isn't that folks would call it, it's that the required parameters will grow over time. I think option 2 is probably best.

src/Components/Server/ref/Microsoft.AspNetCore.Components.Server.netcoreapp3.0.cs
NinoFloris: @JamesNK can't have sealed as a generic argument? you surely meant generic constraint? (upper bound constraint)

halter73: I think you're right @NinoFloris. I don't know why this can't just be sealed.

If someone was planning to add `T : ComponentEndpointConventionBuilder` as a generic constraint and ComponentEndpointConventionBuilder is sealed, you could just use ComponentEndpointConventionBuilder directly instead of introducing T.

JamesNK:
Ran into a fun situation with this PR. gRPC uses WithDisplayName and WithMetadata, and changing them breaks gRPC. There are also gRPC project tests in this solution, and they will fail. tldr; there is a circular dependency between the two repos.

What I'm going to do is update gRPC to not use WithDisplayName and WithMetadata. When a new gRPC version is on NuGet I'll update the template to reference it as part of this PR, then merge it.

src/Components/Server/ref/Microsoft.AspNetCore.Components.Server.netcoreapp3.0.cs
rynowak: OK after some discussion: https://github.com/aspnet/AspNetCore/issues/9316

Could you please seal this and the regular HubConventionBuilder and add an inferface?

src/Components/test/testassets/TestServer/Startup.cs
```diff
@@ -65,7 +65,7 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
 
                 subdirApp.UseEndpoints(endpoints =>
                 {
-                    endpoints.MapHub<ComponentHub>(ComponentHub.DefaultPath).AddComponent<Index>(selector: "root");
+                    endpoints.MapHub<ComponentHub>(ComponentHub.DefaultPath).AddComponent(typeof(Index), selector: "root");
```
dougbu: Huh? This looks like a move _away_ from using a generic argument

build/sources.props
```diff
@@ -14,6 +14,7 @@
       https://dotnetfeed.blob.core.windows.net/aspnet-entityframeworkcore/index.json;
       https://dotnetfeed.blob.core.windows.net/aspnet-aspnetcore-tooling/index.json;
       https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json;
+      https://grpc.jfrog.io/grpc/api/nuget/v3/grpc-nuget-dev;
```
natemcmaster: Is this a permanent feed we will need to reference forever? Or temporary until something makes it up to nuget.org? Assuming it's the later, let's open an issue to track the follow-up item here. Even better, of course, is writing code that documents this intention. For example,
```xml
<!-- Can be removed once Grps.AspNetCore 0.1.21 is released to nuget.org and we update our dependency version -->
<RestoreSources Condition=" '$(GrpcAspNetCoreServerPackageVersion)' == '0.1.21-dev201904300701' ">
  $(RestoreSources);
  https://grpc.jfrog.io/grpc/api/nuget/v3/grpc-nuget-dev;
</RestoreSources>
```

JamesNK: It won't be permanent but it could come and go depending on whether we need to reference a nightly version of gRPC.

src/Components/test/testassets/TestServer/Startup.cs
JamesNK: Covered here - https://github.com/aspnet/AspNetCore/pull/8906#discussion_r270283883

build/sources.props
JamesNK: https://github.com/aspnet/AspNetCore/issues/9878

PR #8907, opened 2019-03-29
Added NegotiateMetadata to negotiate endpoints

davidfowl:
- This makes it a bit cleaner to identify the negotiate when trying to apply policies (like replacing the endpoint)
- Added tests that cover MapConnectionHandler for endpoint routing

See https://github.com/davidfowl/AzureSignalRLightupPrototype/blob/166c69ddc54fc716cb46ab570dc0beb17f46c869/Microsoft.Azure.SignalR.Remix/NegotiateMatcherPolicy.cs#L28 for what it looks like today.

cc @rynowak

### DIFF
--- a/src/SignalR/common/Http.Connections/src/NegotiateMetadata.cs
+++ b/src/SignalR/common/Http.Connections/src/NegotiateMetadata.cs
@@ -1,0 +1,13 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.AspNetCore.Http.Connections
+{
+    /// <summary>
+    /// Metadata to identify the /negotiate endpoint for HTTP dispatcher based
+    /// </summary>
+    public class NegotiateMetadata
+    {
+    }
+}
--- a/src/SignalR/common/Http.Connections/src/NegotiateMetadata.cs
+++ b/src/SignalR/common/Http.Connections/src/NegotiateMetadata.cs
@@ -1,3 +1,6 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
 using System;
 using System.Collections.Generic;
 using System.Text;
@@ -5,7 +8,7 @@ using System.Text;
 namespace Microsoft.AspNetCore.Http.Connections
 {
     /// <summary>
-    /// Metadata to identify the /negotiate endpoint for HTTP dispatcher based
+    /// Metadata to identify the /negotiate endpoint for HTTP connections
     /// </summary>
     public class NegotiateMetadata
     {
--- a/src/SignalR/common/Http.Connections/test/MapConnectionHandlerTests.cs
+++ b/src/SignalR/common/Http.Connections/test/MapConnectionHandlerTests.cs
@@ -11,6 +11,7 @@ using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Connections;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Hosting.Server.Features;
+using Microsoft.AspNetCore.Routing;
 using Microsoft.AspNetCore.SignalR.Tests;
 using Microsoft.AspNetCore.Testing.xunit;
 using Microsoft.Extensions.DependencyInjection;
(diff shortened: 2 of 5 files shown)


src/SignalR/common/Http.Connections/src/NegotiateMetadata.cs
```diff
@@ -0,0 +1,13 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.AspNetCore.Http.Connections
+{
+    /// <summary>
+    /// Metadata to identify the /negotiate endpoint for HTTP dispatcher based
+    /// </summary>
+    public class NegotiateMetadata
+    {
```
JamesNK: @rynowak In routing we had interfaces for some metadata. Is there a rule of thumb for interface vs class?

https://github.com/aspnet/AspNetCore/blob/f10635f6c832641e022a05c7fd97fe512ec844d9/src/Http/Routing/src/ISuppressLinkGenerationMetadata.cs

src/SignalR/common/Http.Connections/src/NegotiateMetadata.cs
```diff
@@ -0,0 +1,13 @@
+using System;
```
BrennanConroy: copyright

src/SignalR/common/Http.Connections/src/NegotiateMetadata.cs
```diff
@@ -0,0 +1,13 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.AspNetCore.Http.Connections
+{
+    /// <summary>
+    /// Metadata to identify the /negotiate endpoint for HTTP dispatcher based
```
BrennanConroy: based...?

src/SignalR/common/Http.Connections/test/MapConnectionHandlerTests.cs
```diff
@@ -68,6 +69,69 @@ public void MapConnectionHandlerFindsAuthAttributesOnDoubleAuthEndPoint()
...
+
+            Assert.Equal(1, authCount);
+        }
+
+        [Fact]
+        public void MapConnectionHandlerEndPointRoutingAppliesAttributesBeforeConventions()
+        {
+            void ConfigureRoutes(IEndpointRouteBuilder endpoints)
+            {
+                // This "Foo" policy should override the default auth attribute
```
BrennanConroy: Should this be additive instead of override?

davidfowl: No.

src/SignalR/common/Http.Connections/src/NegotiateMetadata.cs
davidfowl: I'd like to merge this today sooo, we should totally bikeshed on this as fast as possible. I only made it a class because I was thinking we would add things to it but maybe it could be more metadata. Either way I don't super care.

BrennanConroy (approved):

src/SignalR/common/Http.Connections/src/NegotiateMetadata.cs
halter73: 👍 	for the interface. I would say it's to be more similar to ISuppressLinkGenerationMetadata, but I don't see the need for a bool. The absence or presence of INegotiateMetadata should be sufficient.

src/SignalR/common/Http.Connections/test/MapConnectionHandlerTests.cs
```diff
@@ -68,6 +69,69 @@ public void MapConnectionHandlerFindsAuthAttributesOnDoubleAuthEndPoint()
...
+
+            using (var host = BuildWebHostWithEndPointRouting(ConfigureRoutes))
+            {
+                host.Start();
+
+                var dataSource = host.Services.GetRequiredService<EndpointDataSource>();
+                // We register 2 endpoints (/negotiate and /)
+                Assert.Equal(2, dataSource.Endpoints.Count);
+                Assert.NotNull(dataSource.Endpoints[0].Metadata.GetMetadata<NegotiateMetadata>());
+                Assert.Null(dataSource.Endpoints[1].Metadata.GetMetadata<NegotiateMetadata>());
```
halter73: Nit: Do we want to rely on endpoint order in this test? Maybe it'd be better to replace `Endpoint[0]` with `Single(e => e.DisplayName.Contains("negotiate"))` or similar.

halter73 (approved):

src/SignalR/common/Http.Connections/src/NegotiateMetadata.cs
rynowak: The presence or absence isn't enough as **a general rule** because metadata or ordered and can override.  On a case-by-case basis it can be fine 😆

rynowak: Interfaces are for cases where you might want to compose the metadata with some other type that implements multiple policies.

Also to allow you to use attributes to specify metadata without the metadata being tied to an attribute class. (`IAuthorizeData` vs `AuthorizeAttribute`).

Depending on the user case a class could be fine. Do you plan to have one ever one of these? Does it make sense for someone to customize, extend or replace it?

For instance the MVC `ActionDescriptor` goes in metadata but we have no plans to turn that into an interface.

src/SignalR/common/Http.Connections/test/MapConnectionHandlerTests.cs
rynowak: Auth metadata is combined 😆

davidfowl: Ah the middleware does it. TIL

PR #8910, opened 2019-03-29
Fix base URI during prerendering

SteveSandersonMS:
While creating E2E tests for the new `IComponentContext`, I tried to do prerendering with an app mounted on a non-root pathbase, and found this bug.

### DIFF
--- a/src/Components/Server/test/Circuits/CircuitPrerendererTest.cs
+++ b/src/Components/Server/test/Circuits/CircuitPrerendererTest.cs
@@ -1,0 +1,132 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Components.Server.Circuits;
+using Microsoft.AspNetCore.Components.Services;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.DependencyInjection;
+using Moq;
+using Xunit;
+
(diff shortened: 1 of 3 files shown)


src/Components/Server/test/Circuits/CircuitPrerendererTest.cs
```diff
@@ -0,0 +1,132 @@
...
+        // The only unit tests added here should specifically be about how we're
+        // interacting with the HttpContext for configuring the prerenderer.
+
+        [Fact]
+        public async Task ExtractsUriFromHttpContext_EmptyPathBase()
+        {
+            // Arrange
+            var circuitFactory = new TestCircuitFactory();
+            var circuitPrerenderer = new CircuitPrerenderer(circuitFactory);
+            var httpContext = new Mock<HttpContext>();
```
SteveSandersonMS: I strongly dislike mocking (any set of tests based around mocks tends to degrade into a write-only copy-paste fest), so I'm sad to be using it here. But other than handling this entirely as E2E tests, it's probably the only viable way to cover our interaction with `HttpContext`.

javiercn (approved):

javiercn:
https://github.com/aspnet/AspNetCore/pull/8911/files

src/Components/Server/test/Circuits/CircuitPrerendererTest.cs
javiercn: I forgot to mention. You don't have to mock. You can simply do new DefaultHttpContext() and save yourself some pain. Sorry I missed that.

PR #8932, opened 2019-03-29
System.Text.Json Hub Protocol

### LINKED ISSUE #5387: Add protocol package for System.Text.Json
Epic #8798

### LINKED ISSUE #8969: Current usage of Utf8JsonReader needs to use TextEquals (e.g. NegotiateProtocol/etc.)
Currently, consumers of Utf8JsonReader are comparing the ValueSpan/ValueSequence directly (using SequenceEqual). This could be problematic if the payload contained escaped data which would be incorrectly flagged as not matching (when it should).

For example, the following should be re-written to use the new `TextEquals` APIs:
https://github.com/aspnet/AspNetCore/blob/9de42d516e6ee9c8eb17700e134f3ac8a7a682ff/src/SignalR/common/Http.Connections.Common/src/NegotiateProtocol.cs#L215-L217

```C#
if (reader.TextEquals(TransportPropertyNameBytes))
```

Similarly here/elsewhere:
https://github.com/aspnet/AspNetCore/blob/9de42d516e6ee9c8eb17700e134f3ac8a7a682ff/src/SignalR/common/SignalR.Common/src/Protocol/HandshakeProtocol.cs#L122-L124

See https://github.com/dotnet/corefx/issues/34938 for more context.

cc @BrennanConroy, @JamesNK

BrennanConroy:
Fixes https://github.com/aspnet/AspNetCore/issues/5387
Fixes #8969 

We are blocked on a couple APIs that will be merged soon. So the reading and writing works for only a couple scenarios. Ignore the test failures and the todo comments for using an API that doesn't exist yet.

The meat of this change is in JsonHubProtocol.cs and I've also converted our json unit tests to a base test class so newtonsoft and system.text.json can share a lot of the tests. This part is still a little WIP as I figure out what tests can be shared.


BrennanConroy:
Ping, could you guys pre-review this so that if we magically get new bits this week it'll be easier to get this merged.

src/SignalR/common/Protocols.Json/src/baseline.netcore.json
```diff
@@ -0,0 +1,253 @@
+{
```
davidfowl: We're still using these files even when we have reference assemblies?

BrennanConroy:
cc @ahsonkhan

src/Framework/src/Microsoft.AspNetCore.App.Runtime.csproj
```diff
@@ -119,6 +119,9 @@ This package is an internal implementation of the .NET Core SDK and is not meant
   </PropertyGroup>
 
   <ItemGroup>
+    <!-- Workaround for https://github.com/dotnet/core-setup/issues/5665 -->
```
dougbu: Undo this hack by rebasing on master instead of my dead branch.

dougbu: That is do _not_ check this in.

src/SignalR/clients/ts/FunctionalTests/ComplexObject.cs
```diff
@@ -10,7 +10,7 @@ public class ComplexObject
         public string String { get; set; }
         public int[] IntArray { get; set; }
         public byte[] ByteArray { get; set; }
-        public Guid GUID { get; set; }
+        // public Guid GUID { get; set; }
```
dougbu: Even if this is temporary, delete rather than comment out code

src/SignalR/clients/ts/FunctionalTests/SignalR.Npm.FunctionalTests.npmproj
```diff
@@ -8,6 +8,7 @@
     <_TestSauceArgs>--verbose --no-color --configuration $(Configuration) --sauce-user "$(SauceUser)" --sauce-key "$(SauceKey)"</_TestSauceArgs>
     <_TestSauceArgs Condition="'$(BrowserTestHostName)' != ''">$(_TestSauceArgs) --use-hostname "$(BrowserTestHostName)"</_TestSauceArgs>
     <NpmTestArgs Condition="'$(DailyTests)' != 'true'">run test:local -- --no-color --configuration $(Configuration)</NpmTestArgs>
+    <SkipTests>true</SkipTests>
```
dougbu: This was a temporary hack on my temporary branch. Rebase on master and not my branch.

src/SignalR/clients/ts/FunctionalTests/Startup.cs
```diff
@@ -36,12 +36,12 @@ public void ConfigureServices(IServiceCollection services)
...
-                // we are running the same tests with JSON and MsgPack protocols and having
-                // consistent casing makes it cleaner to verify results
-                options.PayloadSerializerSettings.ContractResolver = new DefaultContractResolver();
-            })
+            // .AddNewtonsoftJsonProtocol(options =>
+            // {
+            //     // we are running the same tests with JSON and MsgPack protocols and having
+            //     // consistent casing makes it cleaner to verify results
+            //     options.PayloadSerializerSettings.ContractResolver = new DefaultContractResolver();
+            // })
```
dougbu: Delete rather than comment out. I'm going to stop saying this though there's more cases below.

src/SignalR/common/Protocols.Json/Directory.Build.props
```diff
@@ -0,0 +1,8 @@
+﻿<Project>
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory)..\, Directory.Build.props))\Directory.Build.props" />
+
+  <PropertyGroup>
+    <NoWarn>$(NoWarn);CS3021</NoWarn>
```
dougbu: What's using `[CLSCompliant(false)]` and why (in a non-compliant assembly)?

src/SignalR/common/Protocols.Json/src/baseline.netcore.json
dougbu: We purposefully removed all of these files from the master branch a while ago.

dougbu (dismissed):
Looks like you need to add a ref assembly for the new project. See https://github.com/aspnet/AspNetCore/blob/master/docs/ReferenceAssemblies.md#when-adding-a-new-project

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
```diff
@@ -0,0 +1,790 @@
...
+        {
+            return HubProtocolExtensions.GetMessageBytes(this, message);
+        }
+
+        private HubMessage ParseMessage(ReadOnlySequence<byte> input, IInvocationBinder binder)
+        {
+            try
+            {
+                // We parse using the Utf8JsonReader directly but this has a problem. Some of our properties are dependent on other properties
+                // and since reading the json might be unordered, we need to store the parsed content as JsonDocument to re-parse when true types are known.
```
ahsonkhan: Have you considered building a document of the whole payload up front instead of using the reader or is that too expensive? I am asking especially since I see list being used to generate the stream ids.

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
```diff
@@ -0,0 +1,790 @@
...
+            foreach (var argument in arguments)
+            {
+                var type = argument?.GetType();
+                if (type == typeof(DateTime))
+                {
+                    writer.WriteStringValue((DateTime)argument);
+                }
+                else if (type == typeof(DateTimeOffset))
+                {
+                    writer.WriteStringValue((DateTimeOffset)argument);
```
ahsonkhan: Was this needed only because `+` is being escaped in the writer by default if written out as string (which JsonElement ends up doing)?

If there was a way to customize the JsonWriter to not escape `+`, for example, would you apply that setting to the writer instance you are using here for all the input data you want to write? I was considering providing a writer option to provide custom `TextEncoder`.

cc @bartonjs, @GrabYourPitchforks

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
```diff
@@ -0,0 +1,790 @@
...
+            }
+
+            return new InvocationMessage(invocationId, target, arguments, streamIds);
+        }
+
+        private bool ReadArgumentAsType(ref Utf8JsonReader reader, IReadOnlyList<Type> paramTypes, int paramIndex)
+        {
+            if (paramIndex < paramTypes.Count)
+            {
+                var paramType = paramTypes[paramIndex];
```
ahsonkhan: This isn't used anywhere.

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
```diff
@@ -0,0 +1,790 @@
...
+            }
+
+            return reader.Read();
+        }
+
+        private object BindType(JsonElement jsonObject, Type type)
+        {
+            if (type == typeof(DateTime))
+            {
+                // TODO: return jsonObject.GetDateTime();
```
ahsonkhan: We have these APIs now.

src/SignalR/common/Shared/SystemTextJsonExtensions.cs
```diff
@@ -99,5 +105,24 @@ public static string ReadAsString(this ref Utf8JsonReader reader, string propert
...
+            {
+                return reader.GetDateTime();
+            }
+            else if (type == typeof(DateTimeOffset) || type == typeof(DateTimeOffset?))
+            {
+                return reader.GetDateTimeOffset();
+            }
+            else
+            {
+                reader.Read();
```
ahsonkhan: `Debug.Assert` this was true?

src/SignalR/common/Shared/SystemTextJsonExtensions.cs
```diff
@@ -99,5 +105,24 @@ public static string ReadAsString(this ref Utf8JsonReader reader, string propert
 
             return reader.GetInt32();
         }
+
+        public static object ReadAsType(this ref Utf8JsonReader reader, Type type)
```
ahsonkhan: Is this called anywhere?

src/SignalR/samples/SignalRSamples/Hubs/Streaming.cs
```diff
@@ -12,16 +12,16 @@ namespace SignalRSamples.Hubs
 {
     public class Streaming : Hub
     {
-        public async IAsyncEnumerable<int> AsyncEnumerableCounter(int count, int delay)
+        public async IAsyncEnumerable<int> AsyncEnumerableCounter(int count, double delay)
```
ahsonkhan: Presumably, this is an unrelated change to the json port, correct?

src/SignalR/server/SignalR/test/HubConnectionHandlerTests.cs
```diff
@@ -2231,7 +2231,7 @@ public async Task HubOptionsCanUseCustomJsonSerializerSettings()
             }
         }
 
-        [Fact]
+        [Fact(Skip = "Camel case is not the default yet")]
```
ahsonkhan: FYI @steveharter, for the backlog of work items :)

src/SignalR/server/SignalR/test/HubConnectionHandlerTestUtils/Hubs.cs
```diff
@@ -23,7 +23,7 @@ public Task ClientSendMethod(string userId, string message)
             return Clients.User(userId).SendAsync("Send", message);
         }
 
-        public Task SendToMultipleUsers(IReadOnlyList<string> userIds, string message)
+        public Task SendToMultipleUsers(List<string> userIds, string message)
```
ahsonkhan: Is this change unrelated to the port?

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
```diff
@@ -0,0 +1,790 @@
...
+                                if (reader.TokenType != JsonTokenType.StartArray)
+                                {
+                                    throw new InvalidDataException($"Expected '{ArgumentsPropertyName}' to be of type {SystemTextJsonExtensions.GetTokenString(JsonTokenType.StartArray)}.");
+                                }
+
+                                hasArguments = true;
+
+                                if (string.IsNullOrEmpty(target))
+                                {
+                                    // We don't know the method name yet so just parse an array of generic JArray
```
ahsonkhan: JArray != JsonDocument

src/SignalR/perf/Microbenchmarks/HubProtocolBenchmark.cs
```diff
@@ -54,15 +54,15 @@ public void GlobalSetup()
...
-        [Benchmark]
-        public void ReadSingleMessage()
-        {
-            var data = new ReadOnlySequence<byte>(_binaryInput);
-            if (!_hubProtocol.TryParseMessage(ref data, _binder, out _))
-            {
-                throw new InvalidOperationException("Failed to read message");
-            }
-        }
+        // [Benchmark]
```
davidfowl: ?

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
davidfowl: What would be the benefit of making a document? For the most part, we don't hit the JsonDocument code path in normal scenarios since our clients send the data in the "right" order. We need to keep the arguments and return type "un parsed" for as long as possible.

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
```diff
@@ -0,0 +1,790 @@
...
+                {
+                    writer.WriteStringValue((DateTime)argument);
+                }
+                else if (type == typeof(DateTimeOffset))
+                {
+                    writer.WriteStringValue((DateTimeOffset)argument);
+                }
+                else
+                {
+                    var bytes = JsonSerializer.ToBytes(argument, type);
```
davidfowl: wut

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
ahsonkhan: The main reason I mentioned this is that you wouldn't have to build up the list and then call ToArray on it. Instead, you could initialize the array based on `element.GetArrayLength()`:
https://github.com/aspnet/AspNetCore/pull/8932/files#diff-81f0f21fef0bc37dc6a70b52a0c3379aR167

But if that is a small part of the payload (or if that list is generally small), it's not worth it.

src/SignalR/samples/SignalRSamples/Hubs/Streaming.cs
BrennanConroy: No, System.Text.Json doesn't support casting double to int automatically.

src/SignalR/server/SignalR/test/HubConnectionHandlerTestUtils/Hubs.cs
BrennanConroy: Nope, `IReadOnlyList<T>` isn't supported yet

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
BrennanConroy: There is no way to serialize into a writer, so we need to get the bytes of the object, then create a JsonElement and write the JsonElement into the writer.

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
BrennanConroy: >Was this needed only because + is being escaped in the writer by default if written out as string (which JsonElement ends up doing)?

Correct. Also it's probably more performant to write it directly with the writer.

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
JamesNK: Raise an issue on corefx. You should be able to do this.

src/SignalR/clients/csharp/Client.Core/src/HubConnectionBuilder.cs
```diff
@@ -25,7 +27,7 @@ public HubConnectionBuilder()
             Services = new ServiceCollection();
             Services.AddSingleton<HubConnection>();
             Services.AddLogging();
-            this.AddNewtonsoftJsonProtocol();
+            this.AddJsonProtocol();
```
anurse: I don't think we're going to be removing Newtonsoft as the default in preview4 in the end. Let's swap this back and make sure we show how to replace Newtonsoft with System.Text.Json in a blog post.

BrennanConroy: I doubt we'll be getting this in for preview4

ahsonkhan: > I don't think we're going to be removing Newtonsoft as the default in preview4 in the end. 

> I doubt we'll be getting this in for preview4

Why?

anurse: > I don't think we're going to be removing Newtonsoft as the default in preview4 in the end.

Discussed with @ahsonkhan offline. There are still things missing that will make the experience pretty rough (https://github.com/aspnet/AspNetCore/issues/8489).

ahsonkhan: > There are still things missing that will make the experience pretty rough

Particularly around dictionary and collection interfaces like IReadonlyList.

Fyi, @steveharter, @layomi, @joshfree

anurse:
If we turn this off by default, can we try to get it in for preview 4 so people can use it?

@BrennanConroy can you update to make this off by default and verify the code necessary to turn Newtonsoft off and System.Text.Json on? That way we can blog about it.

BrennanConroy:
:up: 📅 Newtonsoft is the new old default.

src/SignalR/server/SignalR/test/Internal/DefaultHubProtocolResolverTests.cs
```diff
@@ -70,14 +70,18 @@ public void DefaultHubProtocolResolverReturnsNullForNotSupportedProtocol()
...
+            var jsonProtocol2 = new NewtonsoftJsonHubProtocol();
+            var resolver = new DefaultHubProtocolResolver(new[] {
+                jsonProtocol1,
+                jsonProtocol2
+            }, NullLogger<DefaultHubProtocolResolver>.Instance);
 
-            Assert.Equal($"Multiple Hub Protocols with the name 'json' were registered.", exception.Message);
+            var resolvedProtocol = resolver.GetProtocol(jsonProtocol2.Name, null);
+            Assert.NotEqual(jsonProtocol1, resolvedProtocol);
+            Assert.Equal(jsonProtocol2, resolvedProtocol);
```
halter73: Nit: Assert.Same

src/SignalR/server/SignalR/test/HubConnectionHandlerTestUtils/Hubs.cs
halter73: Is there a skipped test for this?

src/SignalR/server/Core/src/Internal/DefaultHubProtocolResolver.cs
```diff
@@ -23,16 +23,20 @@ public DefaultHubProtocolResolver(IEnumerable<IHubProtocol> availableProtocols,
...
                 if (_availableProtocols.ContainsKey(protocol.Name))
                 {
-                    throw new InvalidOperationException($"Multiple Hub Protocols with the name '{protocol.Name}' were registered.");
+                    Log.OverwroteSignalRProtocol(_logger, protocol.Name, protocol.GetType());
+                    _availableProtocols[protocol.Name] = protocol;
+                }
+                else
+                {
+                    Log.RegisteredSignalRProtocol(_logger, protocol.Name, protocol.GetType());
+                    _availableProtocols.Add(protocol.Name, protocol);
```
halter73: nit: you can still use the indexer setter if the key doesn't exist yet.

src/SignalR/samples/SignalRSamples/Hubs/Streaming.cs
halter73: Is there a skipped test for this?

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
```diff
@@ -0,0 +1,777 @@
...
+                                if (string.IsNullOrEmpty(invocationId))
+                                {
+                                    // If we don't have an invocation id then we need to store it as a JsonDocument so we can parse it later
+                                    resultToken = JsonDocument.ParseValue(ref reader);
+                                }
+                                else
+                                {
+                                    // If we have an invocation id already we can parse the end result
+                                    var returnType = binder.GetReturnType(invocationId);
+                                    if (reader.TokenType == JsonTokenType.Null)
```
mikaelm12: nit: Invert the check to see if `reader.TokenType` isn't null, then you don't need an else clause at all

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
```diff
@@ -0,0 +1,777 @@
...
+                                {
+                                    // If we don't have an id yet then we need to store it as a JsonDocument to parse later
+                                    itemsToken = JsonDocument.ParseValue(ref reader);
+                                    continue;
+                                }
+
+                                try
+                                {
+                                    var itemType = binder.GetStreamItemType(id);
+                                    if (reader.TokenType == JsonTokenType.Null)
```
mikaelm12: Same comment as above.

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
```diff
@@ -0,0 +1,777 @@
...
+                                {
+                                    // We don't know the method name yet so just store the array in JsonDocument
+                                    argumentsToken = JsonDocument.ParseValue(ref reader);
+                                }
+                                else
+                                {
+                                    try
+                                    {
+                                        var paramTypes = binder.GetParameterTypes(target);
+                                        if (reader.TokenType == JsonTokenType.Null)
```
mikaelm12: Here as well.

src/SignalR/common/SignalR.Common/test/Internal/Protocol/JsonHubProtocolTestsBase.cs
```diff
@@ -0,0 +1,334 @@
...
+            //new JsonProtocolTestData("InvocationMessage_HasHeaders", AddHeaders(TestHeaders, new InvocationMessage("123", "Target", new object[] { 1, "Foo", 2.0f })), "{\"type\":1," + SerializedHeaders + ",\"invocationId\":\"123\",\"target\":\"Target\",\"arguments\":[1,\"Foo\",2.0]}"),
+            //new JsonProtocolTestData("StreamInvocationMessage_HasHeaders", AddHeaders(TestHeaders, new StreamInvocationMessage("123", "Target", new object[] { new CustomObject() })), "{\"type\":4," + SerializedHeaders + ",\"invocationId\":\"123\",\"target\":\"Target\",\"arguments\":[{\"StringProp\":\"SignalR!\",\"DoubleProp\":6.2831853071,\"IntProp\":42,\"DateTimeProp\":\"2017-04-11T00:00:00Z\",\"NullProp\":null,\"ByteArrProp\":[1,2,3]}]}"),
+            //new JsonProtocolTestData("CancelInvocationMessage_HasHeaders", AddHeaders(TestHeaders, new CancelInvocationMessage("123")), "{\"type\":5," + SerializedHeaders + ",\"invocationId\":\"123\"}"),
```
halter73: Using MemberData, is it possible to skip specific theory variations?

src/SignalR/common/Shared/SystemTextJsonExtensions.cs
halter73: ping

src/SignalR/common/Shared/SystemTextJsonExtensions.cs
```diff
@@ -24,10 +25,15 @@ public static void EnsureObjectStart(this ref Utf8JsonReader reader)
         {
             if (reader.TokenType != JsonTokenType.StartObject)
             {
-                throw new InvalidDataException($"Unexpected JSON Token Type '{GetTokenString(reader.TokenType)}'. Expected a JSON Object.");
+                throw new InvalidDataException($"Unexpected JSON Token Type '{reader.GetTokenString()}'. Expected a JSON Object.");
```
halter73: How is this called by reader.GetTokenString()?

src/SignalR/common/Shared/SystemTextJsonExtensions.cs
```diff
@@ -24,10 +25,15 @@ public static void EnsureObjectStart(this ref Utf8JsonReader reader)
...
             }
         }
 
+        public static string GetTokenString(this ref Utf8JsonReader reader)
```
halter73: Who calls this? Do we need to keep the JsonTokenType overload public?

src/SignalR/common/Shared/SystemTextJsonExtensions.cs
halter73: Never mind. I was confused and didn't realize this was interpolated.

src/SignalR/common/Protocols.Json/src/JsonHubProtocolOptions.cs
```diff
@@ -0,0 +1,14 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.SignalR.Protocol;
+
+namespace Microsoft.AspNetCore.SignalR
+{
+    /// <summary>
+    /// Options used to configure a <see cref="JsonHubProtocol"/> instance.
+    /// </summary>
+    public class JsonHubProtocolOptions
```
halter73: sealed?

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
```diff
@@ -0,0 +1,777 @@
...
+using Microsoft.AspNetCore.Connections;
+using Microsoft.AspNetCore.Internal;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.AspNetCore.SignalR.Protocol
+{
+    /// <summary>
+    /// Implements the SignalR Hub Protocol using System.Text.Json.
+    /// </summary>
+    public class JsonHubProtocol : IHubProtocol
```
halter73: Should this be sealed?

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
```diff
@@ -0,0 +1,777 @@
...
+                    writer.WriteStringValue((DateTime)argument);
+                }
+                else if (type == typeof(DateTimeOffset))
+                {
+                    writer.WriteStringValue((DateTimeOffset)argument);
+                }
+                else
+                {
+                    var bytes = JsonSerializer.ToBytes(argument, type);
+                    using var token = JsonDocument.Parse(bytes);
```
halter73: +10 pranav points. This reminds me of RAII. We're C++ now 😆

src/SignalR/common/Shared/SystemTextJsonExtensions.cs
BrennanConroy: Yes, it's used at least once

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
BrennanConroy: I don't see a reason to allow or disallow inheriting this class.

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
```diff
@@ -0,0 +1,777 @@
...
+                    token.RootElement.WriteAsValue(ref writer);
```
halter73: We still need to be able to serialize directly into a writer, but I does Utf8JsonWriter have a concept similar Json.NET's WriteRawValue? We'd still allocate an array, but least then we wouldn't need to serialize, parse then serialize again.

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
```diff
@@ -0,0 +1,777 @@
...
+            WriteInvocationId(message, ref writer);
+            if (!string.IsNullOrEmpty(message.Error))
+            {
+                writer.WriteString(ErrorPropertyNameBytes, message.Error);
+            }
+            else if (message.HasResult)
+            {
+                var bytes = JsonSerializer.ToBytes(message.Result, message.Result?.GetType());
+                using var token = JsonDocument.Parse(bytes);
+                token.RootElement.WriteAsProperty(ResultPropertyNameBytes, ref writer);
```
halter73: I'd also create a helper method that's called every time we write an arbitrary type, so when we fix it we don't accidentally leave the inefficient logic laying around somewhere.

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
halter73: Let's use sealed by default for new public classes.

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
halter73: ping

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
bartonjs: > does Utf8JsonWriter have a concept similar Json.NET's WriteRawValue

No.  "really raw" violates indentation settings, can violate the state (opening an array or object it doesn't close, closing the existing array or object), and can violate the character escaping rules for the writer.

So it's something that's hard to add... because it either isn't "raw" or leads to unexpected data.  If it's a token-by-token thing then there's the possibility that it writes stuff before finding out that the payload is invalid...  writing through JsonDocument solves the validity, escaping, and indentation problems; so it's the approach that we are encouraging for now.

src/SignalR/samples/SignalRSamples/Hubs/Streaming.cs
BrennanConroy: No

src/SignalR/server/SignalR/test/HubConnectionHandlerTestUtils/Hubs.cs
BrennanConroy: No

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
JamesNK: Serializing to bytes, then wrapping in a JsonDocument, then writing the JsonDocument is needlessly complex.

JsonSerializer should support serializing directly to a Utf8JsonWriter.

bartonjs: > JsonSerializer should support serializing directly to a Utf8JsonWriter.

Oh, yeah, I agree.  I believe it's on the serializer TODO list (though if you can't find an issue then opening one would have virtue).  I was mainly commenting on WriteRaw for the general case.

JamesNK: ok!

BrennanConroy:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=147444&view=logs

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
BrennanConroy: Filed https://github.com/dotnet/corefx/issues/36714 as the closest thing I could find was the `Utf8JsonWriter` issue talking about `WriteRaw`

ahsonkhan: Do you need equivalent functionality on the reader side? If so, an issue for that would be most welcome :)

I consider this under the umbrella of "interop between serializer and reader/writer" (similar to what we recently added - interop between jsondocument and reader/writer - i.e. https://github.com/dotnet/corefx/issues/36148 / https://github.com/dotnet/corefx/issues/36153).

BrennanConroy: Yes we do, I'm currently looking at this and can hopefully show some perf numbers for the difference between using the JsonDocument approach we currently have vs. using the raw sequence.

BrennanConroy: Filed for that case https://github.com/dotnet/corefx/issues/36717

aspnet-hello:

I've triaged the above build.

anurse:
Ok, after discussions with JSON folks and @BrennanConroy, let's withdraw this from preview 4. It sounds like the big JSON push will be in preview 5

BrennanConroy:
image

dougbu (approved):
Just looked at the eng/ and *.csproj changes. Those look fine but you need another approver for the C#.

src/SignalR/common/Protocols.Json/src/Microsoft.AspNetCore.SignalR.Protocols.Json.csproj
```diff
@@ -0,0 +1,30 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <Description>Implements the SignalR Hub Protocol using System.Text.Json.</Description>
+    <TargetFrameworks>netstandard2.0;netcoreapp3.0</TargetFrameworks>
+    <IsAspNetCoreApp>false</IsAspNetCoreApp>
```
natemcmaster: Is this temporary? Or is it really not part of the plans to JSON protocol support into the shared framework?

BrennanConroy: Temporary

natemcmaster: 👍 Ok, just wanted to make sure.

davidfowl: So it isn't in the shared framework?

BrennanConroy: I guess there is no reason not to add it now.

halter73:
It might be worth combining the three places we call `var bytes = JsonSerializer.ToBytes(...); using var token = JsonDocument.Parse(bytes); token.RootElement.WriteAs...` into a helper method since we'll want to replace that logic later.

src/SignalR/common/Protocols.Json/src/Microsoft.AspNetCore.SignalR.Protocols.Json.csproj
natemcmaster: Fine by me. Any reason we can't also remove the NewtonsoftJson version from the shared framework as a part of this PR?

BrennanConroy: Because we haven't decided to do it yet.

BrennanConroy:
Is that the only thing blocking approval?

If so, could you approve and I'll make the changes soonish.

halter73:
I'll approve it if it really needs to get in soon. I'm hoping someone else like @davidfowl or maybe @JamesNK will also take another look at this and approve it.

I didn't realize we were still rushing to get this in after it missed the preview4 train. I can understand not wanting to deal with merge conflicts piling up.

I haven't been in any discussions about when this needs to be merged or what changes we expect from corefx before we're comfortable pushing. Ideally we wouldn't merge this until we had a not-so-crazily-inefficient way to serialize nested objects of arbitrary types.

halter73 (approved):

src/SignalR/perf/benchmarkapps/BenchmarkServer/Startup.cs
```diff
@@ -23,6 +23,7 @@ public void ConfigureServices(IServiceCollection services)
             {
                 o.EnableDetailedErrors = true;
             })
+            // TODO: Json vs NewtonsoftJson option
```
halter73: Are you going to do this before merging?

src/SignalR/samples/SignalRSamples/Hubs/Streaming.cs
```diff
@@ -12,16 +12,16 @@ namespace SignalRSamples.Hubs
...
         {
             for (var i = 0; i < count; i++)
             {
                 yield return i;
-                await Task.Delay(delay);
+                await Task.Delay((int)delay);
```
halter73: Can we add a skipped test for System.Text.Json casting double to int automatically before merging?

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
```diff
@@ -0,0 +1,758 @@
...
+                    switch (reader.TokenType)
+                    {
+                        case JsonTokenType.PropertyName:
+                            if (reader.TextEquals(TypePropertyNameBytes))
+                            {
+                                type = reader.ReadAsInt32(TypePropertyName);
+
+                                if (type == null)
+                                {
+                                    throw new InvalidDataException($"Missing required property '{TypePropertyName}'.");
```
JamesNK: I'm confused by this. Why would ReadAsInt32 return null, and why does it mean there is no type property? You've just checked you're in the type property so that can't be true.

JamesNK: I looked up ReadAsInt32 https://github.com/aspnet/AspNetCore/blob/9de42d516e6ee9c8eb17700e134f3ac8a7a682ff/src/SignalR/common/Shared/SystemTextJsonExtensions.cs#L86

This message would be more accurate to say that the JSON type is wrong like... https://github.com/aspnet/AspNetCore/blob/9de42d516e6ee9c8eb17700e134f3ac8a7a682ff/src/SignalR/common/Shared/SystemTextJsonExtensions.cs#L97
...instead of saying the property is missing.

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
```diff
@@ -0,0 +1,758 @@
...
+                            {
+                                invocationId = reader.ReadAsString(InvocationIdPropertyName);
+                            }
+                            else if (reader.TextEquals(StreamIdsPropertyNameBytes))
+                            {
+                                reader.CheckRead();
+
+                                if (reader.TokenType != JsonTokenType.StartArray)
+                                {
+                                    throw new InvalidDataException($"Expected '{StreamIdsPropertyName}' to be of type {reader.GetTokenString()}.");
```
JamesNK: Should `reader.GetTokenString()` be `SystemTextJsonExtensions.GetTokenString(JsonTokenType.StartArray)`?

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
```diff
@@ -0,0 +1,758 @@
...
+                                {
+                                    // We don't know the method name yet so just store the array in JsonDocument
+                                    argumentsToken = JsonDocument.ParseValue(ref reader);
+                                }
+                                else
+                                {
+                                    try
+                                    {
+                                        var paramTypes = binder.GetParameterTypes(target);
+                                        if (reader.TokenType != JsonTokenType.Null)
```
JamesNK: Has there been a read between here and the token type check on line 234?

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
```diff
@@ -0,0 +1,758 @@
...
+                                            arguments = BindTypes(token.RootElement, paramTypes);
+                                        }
+                                    }
+                                    catch (Exception ex)
+                                    {
+                                        argumentBindingException = ExceptionDispatchInfo.Capture(ex);
+
+                                        // Could be at any point in argument array JSON when an error is thrown
+                                        // Read until the end of the argument JSON array
+                                        while (reader.CurrentDepth == initialDepth && reader.TokenType == JsonTokenType.StartArray ||
```
JamesNK: Is there good unit tests coverage here? I recall @ahsonkhan discussing changing the reader depth logic. I don't know whether that has happened or not yet.

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
```diff
@@ -0,0 +1,758 @@
...
+                        headers[propertyName] = reader.GetString();
+                        break;
+                    case JsonTokenType.Comment:
+                        break;
+                    case JsonTokenType.EndObject:
+                        return headers;
+                }
+            }
+
+            throw new JsonReaderException("Unexpected end when reading message headers", 0, 0);
```
JamesNK: `0, 0` refers to line and position?

Why not throw InvalidDataException here like CheckRead does?

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
```diff
@@ -0,0 +1,758 @@
...
+            }
+            writer.WriteEndObject();
+            writer.Flush();
+        }
+
+        private void WriteHeaders(ref Utf8JsonWriter writer, HubInvocationMessage message)
+        {
+            if (message.Headers != null && message.Headers.Count > 0)
+            {
+                writer.WriteStartObject(HeadersPropertyNameBytes);
```
JamesNK: When writing a known property name, add `escape: false`. This will tell the writer to skip inspecting the property name and escaping characters. We don't need it because we know this is a valid property name.

It provides a small performance boost.

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
```diff
@@ -0,0 +1,758 @@
...
+                writer.WriteEndObject();
+            }
+        }
+
+        private void WriteCompletionMessage(CompletionMessage message, ref Utf8JsonWriter writer)
+        {
+            WriteInvocationId(message, ref writer);
+            if (!string.IsNullOrEmpty(message.Error))
+            {
+                writer.WriteString(ErrorPropertyNameBytes, message.Error);
```
JamesNK: `escaped: false`, etc

I don't want to comment every one 😄

JamesNK (commented):
Fix these up and then I'll re-review and approve

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
```diff
@@ -0,0 +1,758 @@
...
+        private const string TypePropertyName = "type";
+        private static ReadOnlySpan<byte> TypePropertyNameBytes => new byte[] { (byte)'t', (byte)'y', (byte)'p', (byte)'e' };
+        private const string ErrorPropertyName = "error";
+        private static ReadOnlySpan<byte> ErrorPropertyNameBytes => new byte[] { (byte)'e', (byte)'r', (byte)'r', (byte)'o', (byte)'r' };
+        private const string TargetPropertyName = "target";
+        private static ReadOnlySpan<byte> TargetPropertyNameBytes => new byte[] { (byte)'t', (byte)'a', (byte)'r', (byte)'g', (byte)'e', (byte)'t' };
+        private const string ArgumentsPropertyName = "arguments";
+        private static ReadOnlySpan<byte> ArgumentsPropertyNameBytes => new byte[] { (byte)'a', (byte)'r', (byte)'g', (byte)'u', (byte)'m', (byte)'e', (byte)'n', (byte)'t', (byte)'s' };
+        private const string HeadersPropertyName = "headers";
+        private static ReadOnlySpan<byte> HeadersPropertyNameBytes => new byte[] { (byte)'h', (byte)'e', (byte)'a', (byte)'d', (byte)'e', (byte)'r', (byte)'s' };
```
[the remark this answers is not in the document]
BrennanConroy: I know, I was there ;)

src/SignalR/perf/benchmarkapps/BenchmarkServer/Startup.cs
BrennanConroy: Nope

src/SignalR/samples/SignalRSamples/Hubs/Streaming.cs
BrennanConroy: I'm not sure we want this behavior. But I'll add a commented out test, and we can decide later.

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
BrennanConroy: We have the ArgumentBindingFailure tests, but because we're using JsonDocument.Parse right now, this code path isn't actually used. It will be used once we move off JsonDocument.

JamesNK (approved):

BrennanConroy:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=154173

aspnet-hello:

I've triaged the above build.

src/SignalR/server/Core/src/Internal/DefaultHubProtocolResolver.cs
```diff
@@ -23,16 +23,19 @@ public DefaultHubProtocolResolver(IEnumerable<IHubProtocol> availableProtocols,
...
 
-            // We might get duplicates in _hubProtocols, but we're going to check it and throw in just a sec.
+            // We might get duplicates in _hubProtocols, but we're going to check it and overwrite in just a sec.
             _hubProtocols = availableProtocols.ToList();
             foreach (var protocol in _hubProtocols)
             {
                 if (_availableProtocols.ContainsKey(protocol.Name))
                 {
-                    throw new InvalidOperationException($"Multiple Hub Protocols with the name '{protocol.Name}' were registered.");
+                    Log.OverwroteSignalRProtocol(_logger, protocol.Name, protocol.GetType());
```
davidfowl: Meh. Last one wins, no need to log.

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
```diff
@@ -0,0 +1,760 @@
...
+            {
+                if (arguments == null)
+                {
+                    arguments = new object[paramCount];
+                }
+
+                try
+                {
+                    arguments[paramIndex] = BindType(element, paramTypes[paramIndex]);
+                    ++paramIndex;
```
davidfowl: prefix operator?

BrennanConroy: ++C guy here

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
ahsonkhan: > I recall @ahsonkhan discussing changing the reader depth logic. I don't know whether that has happened or not yet.

It did happen. It matches Json.NET semantics now.

See: https://github.com/dotnet/corefx/pull/36116

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
```diff
@@ -0,0 +1,760 @@
...
+            }
+            writer.WriteEndObject();
+            writer.Flush();
+        }
+
+        private void WriteHeaders(ref Utf8JsonWriter writer, HubInvocationMessage message)
+        {
+            if (message.Headers != null && message.Headers.Count > 0)
+            {
+                writer.WriteStartObject(HeadersPropertyNameBytes, escape: true);
```
ahsonkhan: Was this meant to be `escape: false`?

BrennanConroy: Haha, yeah

PR #8933, opened 2019-03-29
Always replace ANCM in xdt transform.

jkotalik:
Fixes https://github.com/aspnet/AspNetCore/issues/7307

Verified this worked too.


BrennanConroy (approved):

shirhatti:
@jkotalik Late to the party here, but is this already available in Antares?

jkotalik:
Yep!

PR #8935, opened 2019-03-30
Using IAsyncEnumerable in the .NET Client

### LINKED ISSUE #5376: Implement support for IAsyncEnumerable in the .NET client

Split from https://github.com/aspnet/SignalR/issues/3166

Support for IAsyncEnumerable on the client (returning it from `HubConnection.StreamAsync()`). This is just a little thorny because of the open question of .NET Standard support for `IAsyncEnumerable`

mikaelm12:
Fixes: #5376 
We finally get to use the `StreamAsync `method name 😄 

Currently what we have is 
```csharp
public static Task<ChannelReader<TResult>> StreamAsChannelAsync<TResult>(...)
```
This is adding support for 

```csharp
public static IAsyncEnumerable<TResult> StreamAsync<TResult>(...)
```


src/SignalR/clients/csharp/Client.Core/src/HubConnectionExtensions.StreamAsync.cs
```diff
@@ -0,0 +1,280 @@
...
+        /// <summary>
+        /// Invokes a streaming hub method on the server using the specified method name and return type.
+        /// </summary>
+        /// <typeparam name="TResult">The return type of the streaming server method.</typeparam>
+        /// <param name="hubConnection">The hub connection.</param>
+        /// <param name="methodName">The name of the server method to invoke.</param>
+        /// <param name="cancellationToken">The token to monitor for cancellation requests. The default value is <see cref="CancellationToken.None" />.</param>
+        /// <returns>
+        /// A <see cref="Task{TResult}"/> that represents the asynchronous invoke.
+        /// The <see cref="Task{TResult}.Result"/> property returns a <see cref="ChannelReader{T}"/> for the streamed hub method values.
```
mikaelm12: The return values on the doc comments need to be updated

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -436,6 +436,77 @@ private async Task StopAsyncCore(bool disposing)
             }
         }
 
+#if NETCOREAPP3_0
+        public async IAsyncEnumerable<object> StreamAsyncCore(string methodName, Type returnType, object[] args, CancellationToken cancellationToken = default)
```
BrennanConroy: What if we just did
```c#
return await StreamAsChannelCoreAsync(methodName, returnType, args, cancellationToken).ReadAllAsync();
```

Eilon:
Tagging @divega because I know you're passionate about `IAsyncEnumerable`, so FYI.

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
mikaelm12: What we should actually be doing here is using the `MakeCancelableAsyncEnumerableFromChannel<T>` method on the `AsyncEnumerableAdapters` class

anurse: Sounds good. We can just share the source for that file for now.

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -436,6 +437,16 @@ private async Task StopAsyncCore(bool disposing)
...
+        {
+            var stream = (await StreamAsChannelCoreAsyncCore(methodName, returnType, args, cancellationToken)).ReadAllAsync();
+            await foreach(var item in AsyncEnumerableAdapters.MakeCancelableAsyncEnumerable(stream, cancellationToken)) {
```
anurse: You shouldn't need the `await foreach` here anymore right? `MakeCancelableAsyncEnumerable` just returns the think we need. It can just be like this:

```suggestion
            return AsyncEnumerableAdapters.MakeCancelableAsyncEnumerable(stream, cancellationToken);
```

mikaelm12: Is there a better way to do this?

anurse: Ah wait, I see why. Because that would end up being a `Task<IAsyncEnumerable<T>>` when we want an `IAsyncEnumerable<T>`. Unfortunate, but makes sense. Disregard!

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -436,6 +437,16 @@ private async Task StopAsyncCore(bool disposing)
             }
         }
 
+#if NETCOREAPP3_0
+        public async IAsyncEnumerable<object> StreamAsyncCore(string methodName, Type returnType, object[] args, CancellationToken cancellationToken = default)
```
halter73: Make this method generic and have it return `IAsyncEnumerable<TResult>` so you don't need to convert later in HubConnectionExtensions.StreamAsyncCore().

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
halter73: Instead of iterating over MakeCancelableAsyncEnumerable()'s return value, instead just returning it directly.

You might want to create another version of the method which [actually uses the CancellationToken passed to GetAsyncEnumerator()](https://github.com/aspnet/AspNetCore/blob/1cf98a5b5c533d61d4b5c3681680157733dbb244/src/SignalR/server/Core/src/Internal/AsyncEnumerableAdapters.cs#L39-L41), and returns `IAsyncEnumerable<TResult>` instead of `IAsyncEnumerable<object>`. Returning a `IAsyncEnumerable<TResult>` is actually easier since you can always directly return the value from `ChannelReader<TResult>.ReadAllAsync().GetAsyncEnumerator()` from `CancelableAsyncEnumerable<TResult>.GetAsyncEnumerator()` without worrying about wrapping the async enumerators for value types with BoxedAsyncEnumerator.

src/SignalR/clients/csharp/Client/test/FunctionalTests/HubConnectionTests.cs
```diff
@@ -322,6 +322,156 @@ public async Task CanInvokeFromOnHandler(string protocolName, HttpTransportType
...
+                    LoggerFactory.CreateLogger<HubConnectionTests>().LogError(ex, "{ExceptionType} from test", ex.GetType().FullName);
+                    throw;
+                }
+                finally
+                {
+                    await connection.DisposeAsync().OrTimeout();
+                }
+            }
+        }
+
```
halter73: It'd be interested in a test that manually calls GetAsyncEnumerator() with a CancellationToken that gets tripped mid stream.

src/SignalR/clients/csharp/Client.Core/src/Microsoft.AspNetCore.SignalR.Client.Core.csproj
```diff
@@ -13,6 +13,8 @@
     <Compile Include="$(SignalRSharedSourceRoot)PipeWriterStream.cs" Link="PipeWriterStream.cs" />
     <Compile Include="$(SignalRSharedSourceRoot)ReflectionHelper.cs" Link="ReflectionHelper.cs" />
     <Compile Include="$(SignalRSharedSourceRoot)TimerAwaitable.cs" Link="Internal\TimerAwaitable.cs" />
+    <Compile Include="$(SignalRRoot)server\Core\src\Internal\AsyncEnumerableAdapters.cs" Link="Internal\AsyncEnumerableAdapters.cs" />
```
halter73: Move AsyncEnumerableAdapters to SignalRSharedSourceRoot if you can.

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
divega: If you are going to be doing this a lot (creating IAsyncEnumerable\<T> from async methods), you could consider creating a helper like MakeCancelableAsyncEnumerableFromChannel but that takes a Task\<Channel\<T>> and handles it to an IAsyncEnumerable\<T> that defers the await until the first call to MoveNextAsync. Then you can return that directly without creating an iterator method. 

The only observable difference would be that any exception that awaiting StreamAsChannelCoreAsyncCore would cause will be deferred.

This is what we do with ExecuteDataReaderAsync.

halter73 (changes requested):
I would like us to create and use a new `AsyncEnumerableAdapters .MakeCancelableAsyncEnumerable<TResult>()` method that uses the CancellationToken from GetAsyncEnumerator().

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
halter73: That's a good idea. Probably worth making a new issue to track it. We could convert `Task<ChannelReader<T>>` to `ChannelReader<T>` in the same way.

anurse: > Instead of iterating over MakeCancelableAsyncEnumerable()'s return value, just return it directly.

That doesn't really work because of the `await` of `StreamAsChannelCoreAsyncCore` beforehand though, right? That's why this has to be transformed into a generator using the `yield` syntax.

anurse: I think that's something we can look at if we encounter an issue. Let's get this rolling now and see how it lands :).

halter73: If we return the IAsyncEnumerable directly, we'd have to make StreamAsync return `Task<IAsyncEnumerable<TResult>>` which @anures and I decided isn't great. I think using an iterator method is OK then.

We still should use a version of `MakeCancelableAsyncEnumerable<TResult>` that returns `IAsyncEnumerable<TResult>` and uses the CancellationToken passed through to `GetAsyncEnumerator(CancellationToken)` linking it with the CancellationToken from `StreamAsyncCore` if both are non-default.

divega: FWIW, one correction to what I said before: You want to pass a Func<Task<Channel<T>> to allow deferring the call (in this case to StreamAsChannelCoreAsyncCore) completely instead of passing an already initiated Task<T>.

> Probably worth making a new issue to track it.

Sure. If you see value. I am back to lurker mode :smile:

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -436,6 +437,16 @@ private async Task StopAsyncCore(bool disposing)
...
+        {
+            var stream = (await StreamAsChannelCoreAsyncCore(methodName, returnType, args, cancellationToken)).ReadAllAsync();
```
BrennanConroy: Call the public API `StreamAsChannelCoreAsync`, that way you get the `ForceAsync` and the log scope for free

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
mikaelm12: > Task<IAsyncEnumerable<TResult>> which @anurse and I decided isn't great. I think using an iterator method is OK then.

The problem with this is that if we use an iterator and yield return from the result of `MakeCancelableAsyncEnumerable` we actually lose the cancellation token which defeats the purpose.

>We still should use a version of MakeCancelableAsyncEnumerable<TResult> that returns IAsyncEnumerable<TResult>

I worked down this route and realized that the streaming methods like `StreamAsChannelCoreAsync` all return `ChannelReader<object>`. This means that no matter what we would have to iterate and cast on every item.

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -436,6 +437,33 @@ private async Task StopAsyncCore(bool disposing)
...
+        /// Invokes a streaming hub method on the server using the specified method name, return type and arguments.
+        /// </summary>
+        /// <typeparam name="TResult">The return type of the streaming server method.</typeparam>
+        /// <param name="methodName">The name of the server method to invoke.</param>
+        /// <param name="args">The arguments used to invoke the server method.</param>
+        /// <param name="cancellationToken">The token to monitor for cancellation requests. The default value is <see cref="CancellationToken.None" />.</param>
+        /// <returns>
+        /// A <see cref="IAsyncEnumerable{TResult}"/> that represents the stream.
+        /// </returns>
+        public async Task<IAsyncEnumerable<TResult>> StreamAsyncCore<TResult>(string methodName, object[] args, CancellationToken cancellationToken = default)
```
mikaelm12: Changed this to return a `Task<IAsyncEnumerable>` to preserve the cancellation logic from the `CancelableTypedAsyncEnumerable`

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -436,6 +437,33 @@ private async Task StopAsyncCore(bool disposing)
...
+        {
+            var stream = CastIAsyncEnumerable<TResult>((await StreamAsChannelCoreAsync(methodName, typeof(TResult), args, cancellationToken)).ReadAllAsync());
```
mikaelm12: Have to wrap the result of the `ReadAllAsync()` in a cast helper method because all of our streaming methods use `Channel<object>` which comes out to IAsyncEnumerable<object>

src/SignalR/clients/csharp/Client/test/FunctionalTests/HubConnectionTests.cs
```diff
@@ -322,6 +322,241 @@ public async Task CanInvokeFromOnHandler(string protocolName, HttpTransportType
...
+        [Theory]
+        [MemberData(nameof(HubProtocolsAndTransportsAndHubPaths))]
+        [LogLevel(LogLevel.Trace)]
+        public async Task StreamAsyncCanBeCancelledThroughGetEnumerator(string protocolName, HttpTransportType transportType, string path)
```
mikaelm12: This test is still failing though

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
mikaelm12: To be clear, I'd rather just return the `IAsyncEnumerable` but  otherwise we lose the cancel logic

halter73: Can you explain this more?

Having StreamAsync return a `Task<IAsyncEnumerable>` *is* more consistent with StreamAsChannelAsync returning `Task<ChannelReader<TResult>>`, but in an ideal world where we had infinite development time neither would return a Task in 3.0.

Is there some reason cancellation cannot work if StreamAsync doesn't return a Task?

src/SignalR/server/Core/src/Internal/AsyncEnumerableAdapters.cs
```diff
@@ -12,21 +12,55 @@ namespace Microsoft.AspNetCore.SignalR.Internal
...
+
         public static IAsyncEnumerable<object> MakeCancelableAsyncEnumerableFromChannel<T>(ChannelReader<T> channel, CancellationToken cancellationToken = default)
         {
             return MakeCancelableAsyncEnumerable(channel.ReadAllAsync(), cancellationToken);
         }
 
+        private class CancelableTypedAsyncEnumerable<TResult> : IAsyncEnumerable<TResult>
+        {
+            private readonly IAsyncEnumerable<TResult> _asyncEnumerable;
+            private CancellationToken _cancellationToken;
```
halter73: This should really be readonly.

src/SignalR/server/Core/src/Internal/AsyncEnumerableAdapters.cs
```diff
@@ -12,21 +12,55 @@ namespace Microsoft.AspNetCore.SignalR.Internal
...
+                    if (_cancellationToken != default)
+                    {
+                        _cancellationToken = CancellationTokenSource.CreateLinkedTokenSource(_cancellationToken, cancellationToken).Token;
+                    }
+                    else
+                    {
+                        _cancellationToken = cancellationToken;
+                    }
+                }
+                return _asyncEnumerable.GetAsyncEnumerator(_cancellationToken);
```
halter73: Don't reset _cancellationToken, just use a local.

src/SignalR/server/Core/src/Internal/AsyncEnumerableAdapters.cs
```diff
@@ -12,21 +12,55 @@ namespace Microsoft.AspNetCore.SignalR.Internal
...
+                _cancellationToken = cancellationToken;
+            }
+
+            public IAsyncEnumerator<TResult> GetAsyncEnumerator(CancellationToken cancellationToken = default)
+            {
+                if (cancellationToken != default)
+                {
+                    if (_cancellationToken != default)
+                    {
+                        _cancellationToken = CancellationTokenSource.CreateLinkedTokenSource(_cancellationToken, cancellationToken).Token;
```
halter73: In the case we create a linked CTS, we have to store the CTS in a wrapped IAsyncEnumerator, so we can dispose it in DisposeAsync.

src/SignalR/server/Core/src/Internal/AsyncEnumerableAdapters.cs
```diff
@@ -12,21 +12,55 @@ namespace Microsoft.AspNetCore.SignalR.Internal
...
+                }
+                return _asyncEnumerable.GetAsyncEnumerator(_cancellationToken);
+            }
+        }
+
         /// <summary>Converts an IAsyncEnumerable of T to an IAsyncEnumerable of object.</summary>
         private class CancelableAsyncEnumerable<T> : IAsyncEnumerable<object>
         {
             private readonly IAsyncEnumerable<T> _asyncEnumerable;
-            private readonly CancellationToken _cancellationToken;
```
halter73: revert

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
mikaelm12: Before we were doing something similar to 
```csharp
var stream= AsyncEnumerableAdapters.MakeCancelableTypedAsyncEnumerable(enumerable, cancellationToken)
await foreach(var item in stream)
{
    yield return stream
}
```
As I understand it this new AsyncEnumerable that from these returns wont have the  previous cancellationToken that was passed in.

mikaelm12:
> I would like us to create and use a new AsyncEnumerableAdapters .MakeCancelableAsyncEnumerable<TResult>() method that uses the CancellationToken from GetAsyncEnumerator().

We found that the implementation of `IAsyncEnumerrator `doesn't do anything with the cancellation token passed into `GetAsyncEnumerator` which is why I couldn't get that test to pass @halter73

src/SignalR/clients/csharp/Client/test/FunctionalTests/HubConnectionTests.cs
mikaelm12: Passing now

mikaelm12:
This is ready for review again

src/SignalR/common/Shared/AsyncEnumerableAdapters.cs
```diff
@@ -12,16 +12,46 @@ namespace Microsoft.AspNetCore.SignalR.Internal
...
+                _cts = cts;
+            }
+
+            public IAsyncEnumerator<TResult> GetAsyncEnumerator(CancellationToken cancellationToken = default)
+            {
+                if (cancellationToken.CanBeCanceled)
+                {
+                    cancellationToken.Register(() =>
+                    {
+                        _cts.Cancel();
```
BrennanConroy: pass the state in to avoid the allocation

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -436,6 +437,39 @@ private async Task StopAsyncCore(bool disposing)
...
+
+        private async IAsyncEnumerable<T> CastIAsyncEnumerable<T>(ChannelReader<object> reader, CancellationToken token)
+        {
+            while (await reader.WaitToReadAsync(token))
+            {
+                while (reader.TryRead(out var item))
+                {
+                    yield return (T)item;
+                }
+            }
```
halter73: Pass the whole CTS into this method and dispose it at the end in a finally.

src/SignalR/common/Shared/AsyncEnumerableAdapters.cs
```diff
@@ -12,16 +12,46 @@ namespace Microsoft.AspNetCore.SignalR.Internal
...
+            {
+                _asyncEnumerable = asyncEnumerable;
+                _cts = cts;
+            }
+
+            public IAsyncEnumerator<TResult> GetAsyncEnumerator(CancellationToken cancellationToken = default)
+            {
+                if (cancellationToken.CanBeCanceled)
+                {
+                    cancellationToken.Register(() =>
```
halter73: This is the registration I was talking about yesterday. Dispose it in a wrapped IAsyncEnumerator.DisposeAsync() method.

src/SignalR/common/Shared/AsyncEnumerableAdapters.cs
```diff
@@ -12,16 +12,46 @@ namespace Microsoft.AspNetCore.SignalR.Internal
...
+                    {
+                        _cts.Cancel();
+                    });
+                }
+                return _asyncEnumerable.GetAsyncEnumerator(cancellationToken);
```
halter73: No need to pass in the cancellationToken here. If we know it won't be used, it's a little misleading to pass it in.

src/SignalR/common/Shared/AsyncEnumerableAdapters.cs
halter73: I'd also call GetAsyncEnumerator first. Store the result. Then register to the cancellation only if it can be canceled and GetAsyncEnumerator didn't throw. Then return the stored result.

src/SignalR/clients/csharp/Client.Core/ref/Microsoft.AspNetCore.SignalR.Client.Core.csproj
```diff
@@ -11,4 +11,12 @@
     <Reference Include="Microsoft.Extensions.Logging"  />
     <Reference Include="System.Threading.Channels"  />
   </ItemGroup>
+<ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp3.0'">
+    <Compile Include="Microsoft.AspNetCore.SignalR.Client.Core.netcoreapp3.0.cs" />
```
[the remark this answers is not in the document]
mikaelm12: This was generated by the GenerateReferenceSource script

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
halter73: You can just move the call to StreamAsChannelCoreAsync into CastIAsyncEnumerable and await it at the beginning. Then StreamAsyncCore doesn't need to be async or Task-returning.

src/SignalR/common/Shared/AsyncEnumerableAdapters.cs
```diff
@@ -12,16 +12,75 @@ namespace Microsoft.AspNetCore.SignalR.Internal
...
+            {
+                var enumerator = _asyncEnumerable.GetAsyncEnumerator();
+                if (cancellationToken.CanBeCanceled)
+                {
+                    var registration = cancellationToken.Register((ctsState) =>
+                    {
+                        ((CancellationTokenSource)ctsState).Cancel();
+                    }, _cts);
+
+                    return new CancelableEnumerator<TResult>(_asyncEnumerable.GetAsyncEnumerator(), registration);
```
halter73: 👍

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -449,17 +449,17 @@ private async Task StopAsyncCore(bool disposing)
...
             var cts = cancellationToken.CanBeCanceled ? CancellationTokenSource.CreateLinkedTokenSource(cancellationToken) : new CancellationTokenSource();
-            var channel = await StreamAsChannelCoreAsync(methodName, typeof(TResult), args, cts.Token);
-            var stream = CastIAsyncEnumerable<TResult>(channel, cts);
+            var stream = CastIAsyncEnumerable<TResult>(methodName, args, cts);
             var cancelableStream = AsyncEnumerableAdapters.MakeCancelableTypedAsyncEnumerable(stream, cts);
             return cancelableStream;
         }
 
-        private async IAsyncEnumerable<T> CastIAsyncEnumerable<T>(ChannelReader<object> reader, CancellationTokenSource cts)
+        private async IAsyncEnumerable<T> CastIAsyncEnumerable<T>(string methodName, object[] args, CancellationTokenSource cts)
```
halter73: nit: maybe rename StreamAsAsyncEnumerableCore now

halter73 (approved):

mikaelm12:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=145064&view=logs

aspnet-hello:

I've triaged the above build.

mikaelm12:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=145064&view=logs

mikaelm12:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=145064&view=logs&jobId=900d802f-1272-5fca-9818-053d703b72e8

PR #8944, opened 2019-03-30
HostingLogScope now logs RequestPath as Request.BasePath+Request.Path

### LINKED ISSUE #8942: HostingLogScope should include Request.BasePath
When logging the request path the `HostingLogScope` should also include the`HttpContext.Request.BasePath`. 

At the moment only `HttpContext.Request.Path` is logged. This leads to the unexpected behavior that the URL triggering the request is not necessarily equal to the request path logged. This is the case if for example the `UsePathBaseMiddleware` is used.

I expected the request path logged to always be equal to the triggering URL.

In my opinion both `Path` and `BasePath` should be logged. Either as two separate fields or as concatenation.

penenkel:
fixes #8942

### DIFF
--- a/src/Hosting/Hosting/src/Internal/HostingLoggerExtensions.cs
+++ b/src/Hosting/Hosting/src/Internal/HostingLoggerExtensions.cs
@@ -132,7 +132,9 @@ namespace Microsoft.AspNetCore.Hosting.Internal
             public HostingLogScope(HttpContext httpContext, string correlationId)
             {
                 _traceIdentifier = httpContext.TraceIdentifier;
-                _path = (httpContext.Request.PathBase+httpContext.Request.Path).ToString();
+                _path = (httpContext.Request.PathBase.HasValue 
+                         ? httpContext.Request.PathBase + httpContext.Request.Path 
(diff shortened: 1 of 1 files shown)


src/Hosting/Hosting/src/Internal/HostingLoggerExtensions.cs
```diff
@@ -132,7 +132,7 @@ public int Count
             public HostingLogScope(HttpContext httpContext, string correlationId)
             {
                 _traceIdentifier = httpContext.TraceIdentifier;
-                _path = httpContext.Request.Path.ToString();
+                _path = (httpContext.Request.PathBase+httpContext.Request.Path).ToString();
```
davidfowl: This always ends up allocating a string. I suggest checking for an empty `PathBase`:

```
_path = (httpContext.Request.PathBase.HasValue ? httpContext.Request.PathBase + httpContext.Request.Path : httpContext.Request.Path).ToString();
```

The majority of request do not have a PathBase.

Tratcher (approved):

davidfowl:
I just forced a re-run

Tratcher:
Thanks

PR #8955, opened 2019-03-31
Add dynamic controller/page routes

### LINKED ISSUE #8130: MVC doesn't produce endpoints for controllers without a route
If you have controllers without a corresponding route, there will be no endpoints created for these action methods. This is a barrier to some of the more advanced kinds of scenarios we want endpoint routing to support in 3.0.

We can represent these as `Endpoint` (not `RouteEndpoint`) - because they can be executed, but don't have an inherent routing behaviour of their own.

rynowak:
Adds infrastructure for a common IRouter-based pattern. In this pattern,
an extender subclasses Route to post-process the route values before MVC
action selection run. The new infrastructure duplicates this kind of
experience but based on endpoint routing.

The approach in this PR starts at the bottom... meaning that this is the
most-focused and least-invasive way to implement a feature like this.
Similar to fallback routing, this is a pattern built with matcher
policies and metadata rather than a built-in feature of routing.

It's valuable to point out that this approach uses IActionConstraint to
disambiguate between actions. The other way we could go would be to make
the *other* matcher policy implementations able to do this. This would
mean that whenever you have a dynamic endpoint, you will not by using
the DFA for features like HTTP methods. It also means that we need to go
re-implement a bunch of infrastructure.

This PR also adds the concept of an 'inert' endpoint - a non-Routable
endpoint that's created when fallback/dynamic is in use. This seems like
a cleaner design because we don't start *matching* RouteEndpoint
instances for URLs that don't match. This resolves #8130

### DIFF
--- a/src/Mvc/Mvc.Core/ref/Microsoft.AspNetCore.Mvc.Core.netcoreapp3.0.cs
+++ b/src/Mvc/Mvc.Core/ref/Microsoft.AspNetCore.Mvc.Core.netcoreapp3.0.cs
@@ -14,6 +14,7 @@ namespace Microsoft.AspNetCore.Builder
         public static Microsoft.AspNetCore.Builder.ControllerActionEndpointConventionBuilder MapControllerRoute(this Microsoft.AspNetCore.Routing.IEndpointRouteBuilder endpoints, string name, string pattern, object defaults = null, object constraints = null, object dataTokens = null) { throw null; }
         public static Microsoft.AspNetCore.Builder.ControllerActionEndpointConventionBuilder MapControllers(this Microsoft.AspNetCore.Routing.IEndpointRouteBuilder endpoints) { throw null; }
         public static Microsoft.AspNetCore.Builder.ControllerActionEndpointConventionBuilder MapDefaultControllerRoute(this Microsoft.AspNetCore.Routing.IEndpointRouteBuilder endpoints) { throw null; }
+        public static void MapDynamicControllerRoute<TTransformer>(this Microsoft.AspNetCore.Routing.IEndpointRouteBuilder endpoints, string pattern) where TTransformer : Microsoft.AspNetCore.Mvc.Routing.DynamicRouteValueTransformer { }
         public static Microsoft.AspNetCore.Builder.IEndpointConventionBuilder MapFallbackToAreaController(this Microsoft.AspNetCore.Routing.IEndpointRouteBuilder endpoints, string action, string controller, string area) { throw null; }
         public static Microsoft.AspNetCore.Builder.IEndpointConventionBuilder MapFallbackToAreaController(this Microsoft.AspNetCore.Routing.IEndpointRouteBuilder endpoints, string pattern, string action, string controller, string area) { throw null; }
         public static Microsoft.AspNetCore.Builder.IEndpointConventionBuilder MapFallbackToController(this Microsoft.AspNetCore.Routing.IEndpointRouteBuilder endpoints, string action, string controller) { throw null; }
@@ -2909,6 +2910,11 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding.Validation
 }
 namespace Microsoft.AspNetCore.Mvc.Routing
 {
+    public abstract partial class DynamicRouteValueTransformer
+    {
+        protected DynamicRouteValueTransformer() { }
+        public abstract System.Threading.Tasks.Task<Microsoft.AspNetCore.Routing.RouteValueDictionary> TransformAsync(Microsoft.AspNetCore.Http.HttpContext httpContext, Microsoft.AspNetCore.Routing.RouteValueDictionary values);
+    }
     [System.AttributeUsageAttribute(System.AttributeTargets.Method, AllowMultiple=true, Inherited=true)]
     public abstract partial class HttpMethodAttribute : System.Attribute, Microsoft.AspNetCore.Mvc.Routing.IActionHttpMethodProvider, Microsoft.AspNetCore.Mvc.Routing.IRouteTemplateProvider
     {
--- a/src/Mvc/Mvc.Core/src/Builder/ControllerEndpointRouteBuilderExtensions.cs
+++ b/src/Mvc/Mvc.Core/src/Builder/ControllerEndpointRouteBuilderExtensions.cs
@@ -8,6 +8,7 @@ using Microsoft.AspNetCore.Mvc.Core;
 using Microsoft.AspNetCore.Mvc.Routing;
 using Microsoft.AspNetCore.Routing;
 using Microsoft.AspNetCore.Routing.Constraints;
+using Microsoft.AspNetCore.Routing.Patterns;
 using Microsoft.Extensions.DependencyInjection;
 
 namespace Microsoft.AspNetCore.Builder
@@ -212,7 +213,7 @@ namespace Microsoft.AspNetCore.Builder
             EnsureControllerServices(endpoints);
 
             // Called for side-effect to make sure that the data source is registered.
-            GetOrCreateDataSource(endpoints);
+            GetOrCreateDataSource(endpoints).CreateInertEndpoints = true;
 
             // Maps a fallback endpoint with an empty delegate. This is OK because
             // we don't expect the delegate to run.
@@ -288,7 +289,7 @@ namespace Microsoft.AspNetCore.Builder
             EnsureControllerServices(endpoints);
 
             // Called for side-effect to make sure that the data source is registered.
-            GetOrCreateDataSource(endpoints);
+            GetOrCreateDataSource(endpoints).CreateInertEndpoints = true;
 
             // Maps a fallback endpoint with an empty delegate. This is OK because
             // we don't expect the delegate to run.
@@ -356,7 +357,7 @@ namespace Microsoft.AspNetCore.Builder
             EnsureControllerServices(endpoints);
 
             // Called for side-effect to make sure that the data source is registered.
-            GetOrCreateDataSource(endpoints);
+            GetOrCreateDataSource(endpoints).CreateInertEndpoints = true;
 
             // Maps a fallback endpoint with an empty delegate. This is OK because
             // we don't expect the delegate to run.
@@ -434,7 +435,7 @@ namespace Microsoft.AspNetCore.Builder
             EnsureControllerServices(endpoints);
 
             // Called for side-effect to make sure that the data source is registered.
-            GetOrCreateDataSource(endpoints);
+            GetOrCreateDataSource(endpoints).CreateInertEndpoints = true;
 
             // Maps a fallback endpoint with an empty delegate. This is OK because
             // we don't expect the delegate to run.
@@ -445,6 +446,48 @@ namespace Microsoft.AspNetCore.Builder
                 b.Metadata.Add(CreateDynamicControllerMetadata(action, controller, area));
             });
             return builder;
+        }
+
+        /// <summary>
+        /// Adds a specialized <see cref="RouteEndpoint"/> to the <see cref="IEndpointRouteBuilder"/> that will
+        /// attempt to select a controller action using the route values produced by <typeparamref name="TTransformer"/>.
+        /// </summary>
+        /// <param name="endpoints">The <see cref="IEndpointRouteBuilder"/> to add the route to.</param>
+        /// <param name="pattern">The URL pattern of the route.</param>
+        /// <typeparam name="TTransformer">The type of a <see cref="DynamicRouteValueTransformer"/>.</typeparam>
+        /// <remarks>
+        /// <para>
+        /// This method allows the registration of a <see cref="RouteEndpoint"/> and <see cref="DynamicRouteValueTransformer"/>
+        /// that combine to dynamically select a controller action using custom logic.
+        /// </para>
+        /// <para>
+        /// The instance of <typeparamref name="TTransformer"/> will be retrieved from the dependency injection container.
+        /// Register <typeparamref name="TTransformer"/> with the desired service lifetime in <c>ConfigureServices</c>.
+        /// </para>
+        /// </remarks>
+        public static void MapDynamicControllerRoute<TTransformer>(this IEndpointRouteBuilder endpoints, string pattern)
+            where TTransformer : DynamicRouteValueTransformer
+        {
+            if (endpoints == null)
+            {
+                throw new ArgumentNullException(nameof(endpoints));
+            }
+
+            EnsureControllerServices(endpoints);
+
+            // Called for side-effect to make sure that the data source is registered.
+            GetOrCreateDataSource(endpoints).CreateInertEndpoints = true;
+
+            endpoints.Map(
+                pattern, 
+                context =>
+                {
+                    throw new InvalidOperationException("This endpoint is not expected to be executed directly.");
+                })
+                .Add(b =>
+                {
+                    b.Metadata.Add(new DynamicControllerRouteValueTransformerMetadata(typeof(TTransformer)));
+                });
         }
 
         private static DynamicControllerMetadata CreateDynamicControllerMetadata(string action, string controller, string area)
--- a/src/Mvc/Mvc.Core/src/Infrastructure/ActionSelectionTable.cs
+++ b/src/Mvc/Mvc.Core/src/Infrastructure/ActionSelectionTable.cs
@@ -74,23 +74,23 @@ namespace Microsoft.AspNetCore.Mvc.Infrastructure
                 });
         }
 
-        public static ActionSelectionTable<RouteEndpoint> Create(IEnumerable<Endpoint> endpoints)
+        public static ActionSelectionTable<Endpoint> Create(IEnumerable<Endpoint> endpoints)
         {
-            return CreateCore<RouteEndpoint>(
+            return CreateCore<Endpoint>(
                 
                 // we don't use version for endpoints
                 version: 0, 
 
-                // Only include RouteEndpoints and only those that aren't suppressed. 
-                items: endpoints.OfType<RouteEndpoint>().Where(e =>
+                // Exclude RouteEndpoints - we only process inert endpoints here. 
+                items: endpoints.Where(e =>
                 {
-                    return e.Metadata.GetMetadata<ISuppressMatchingMetadata>()?.SuppressMatching != true;
+                    return e.GetType() == typeof(Endpoint);
                 }),
 
-                getRouteKeys: e => e.RoutePattern.RequiredValues.Keys,
+                getRouteKeys: e => e.Metadata.GetMetadata<ActionDescriptor>().RouteValues.Keys,
                 getRouteValue: (e, key) =>
                 {
-                    e.RoutePattern.RequiredValues.TryGetValue(key, out var value);
+                    e.Metadata.GetMetadata<ActionDescriptor>().RouteValues.TryGetValue(key, out var value);
                     return Convert.ToString(value, CultureInfo.InvariantCulture) ?? string.Empty;
                 });
         }
--- a/src/Mvc/Mvc.Core/src/Routing/ActionEndpointFactory.cs
+++ b/src/Mvc/Mvc.Core/src/Routing/ActionEndpointFactory.cs
@@ -20,6 +20,7 @@ namespace Microsoft.AspNetCore.Mvc.Routing
     internal class ActionEndpointFactory
     {
         private readonly RoutePatternTransformer _routePatternTransformer;
+        private readonly RequestDelegate _requestDelegate;
 
         public ActionEndpointFactory(RoutePatternTransformer routePatternTransformer)
         {
@@ -29,6 +30,7 @@ namespace Microsoft.AspNetCore.Mvc.Routing
             }
 
             _routePatternTransformer = routePatternTransformer;
+            _requestDelegate = CreateRequestDelegate();
         }
 
         public void AddEndpoints(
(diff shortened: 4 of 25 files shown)


src/Mvc/Mvc.Core/src/DependencyInjection/MvcCoreServiceCollectionExtensions.cs
```diff
@@ -269,7 +273,6 @@ internal static void AddMvcCoreServices(IServiceCollection services)
             //
             // Endpoint Routing / Endpoints
             //
-            services.TryAddSingleton<ActionEndpointDataSource>();
```
rynowak: This should have been removed in the change to removed support for `UseMvc` + `EnableEndpointRouting == true`.

src/Mvc/Mvc.Core/src/Infrastructure/ActionSelectionTable.cs
```diff
@@ -74,23 +74,23 @@ public static ActionSelectionTable<ActionDescriptor> Create(ActionDescriptorColl
                 });
         }
 
-        public static ActionSelectionTable<RouteEndpoint> Create(IEnumerable<Endpoint> endpoints)
+        public static ActionSelectionTable<Endpoint> Create(IEnumerable<Endpoint> endpoints)
```
rynowak: This now uses inert endpoints instead of `RouteEndpoint`.

src/Mvc/Mvc.Core/src/Routing/ActionEndpointFactory.cs
```diff
@@ -56,6 +57,17 @@ public ActionEndpointFactory(RoutePatternTransformer routePatternTransformer)
...
+            {
+                // For each ActionDescriptor create a single 'inert' Endpoint without routing information.
+                //
+                // This makes those endpoints accessible for dynamic and fallback routing. We don't want to
+                // mix RouteEndpoint that is selected by routing, vs Endpoint that is selected by user code/policies
+                //
+                // This also helps us avoid ambiguities when an endpoint has multiple conventional routes.
+                endpoints.Add(CreateInertEndpoint(action, conventions));
+            }
+
```
rynowak: Aside from this all of the other changes in this file are refactoring to avoid duplication for the route endpoint and inert endpoint.

src/Mvc/Mvc.Core/src/Routing/DynamicControllerEndpointMatcherPolicy.cs
```diff
@@ -42,8 +43,13 @@ public bool AppliesToEndpoints(IReadOnlyList<Endpoint> endpoints)
...
             {
-                var metadata = endpoints[i].Metadata.GetMetadata<DynamicControllerMetadata>();
-                if (metadata != null)
+                if (endpoints[i].Metadata.GetMetadata<DynamicControllerMetadata>() != null)
+                {
+                    // Found a dynamic controller endpoint
+                    return true;
+                }
+
+                if (endpoints[i].Metadata.GetMetadata<DynamicControllerRouteValueTransformerMetadata>() != null)
```
rynowak: The code is like 15% more complicated because we handle two cases in a single policy. The upside is that this class has to maintain a cache of computed data per-endpoint. I don't want to duplicate that cache and use 2x the memory.

src/Mvc/Mvc.Core/src/Routing/DynamicControllerEndpointMatcherPolicy.cs
```diff
@@ -81,40 +87,63 @@ public Task ApplyAsync(HttpContext httpContext, EndpointSelectorContext context,
...
+                var values = new RouteValueDictionary(dynamicValues);
 
                 // Include values that were matched by the fallback route.
-                foreach (var kvp in matchedValues)
+                foreach (var kvp in originalValues)
                 {
                     values.TryAdd(kvp.Key, kvp.Value);
                 }
 
+                var replacement = _selector.SelectBestEndpoint(httpContext, values, endpoints);
```
rynowak: This is where we call into the action constraints. 

The reasons that I went this way instead of using something new to solve the problem:
- This is the simplest thing to do for now
- I'm not sure how common this requirement is outside of MVC.

The thing that this solves is selecting between a GET endpoint and a POST endpoint, etc. It seems obvious someone layering a CMS on top of MVC needs this functionality. Maybe I am just being stupid and this needs to be *built-in*.

src/Mvc/Mvc.Core/src/Routing/DynamicRouteValueTransformer.cs
```diff
@@ -0,0 +1,14 @@
...
+
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing;
+
+namespace Microsoft.AspNetCore.Mvc.Routing
+{
+    public abstract class DynamicRouteValueTransformer
+    {
+        public abstract Task<RouteValueDictionary> TransformAsync(HttpContext httpContext, RouteValueDictionary values);
```
rynowak: This is what I think encapsulates the really common set of requirements for those that subclass `Route`. You want to turn one set of route values into another one, and then let MVC do its thing.

src/Mvc/Mvc.Core/test/Infrastructure/ActionSelectionTableTest.cs
```diff
@@ -300,36 +300,6 @@ public void Select_Endpoint_Match_IncludesAttributeRoutedActions()
             Assert.Single(matches);
         }
 
-        [Fact]
-        public void Select_Endpoint_NoMatch_ExcludesMatchingSuppressedAction()
```
rynowak: No longer a valid test

src/Mvc/test/WebSites/RoutingWebSite/Areas/Admin/DynamicController.cs
```diff
@@ -0,0 +1,22 @@
...
+    [Area("Admin")]
+    public class DynamicController : Controller
+    {
+        public ActionResult Index()
+        {
+            return Content("Hello from dynamic controller: " + Url.Action());
+        }
+
+        [HttpPost]
+        public ActionResult Index(int x = 0)
```
rynowak: Before this change, using a fallback route to map to this action would result in an implementation-defined choice of which method is called.

src/Mvc/Mvc.Core/src/Routing/DynamicControllerEndpointMatcherPolicy.cs
JamesNK: This is the part I don't like. I don't care that it is slow, but I think having an endpoint routing feature (new hotness) reuse action constraints (old and busted) is going to become more and more troublesome. There are already matcher policies in endpoint routing - host matching - that have no equivalent in action constraints.

rynowak:
@JamesNK @pranavkm - can you please take a look? This is for preview 7.

src/Mvc/Mvc.Core/ref/Microsoft.AspNetCore.Mvc.Core.netcoreapp3.0.cs
```diff
@@ -2909,6 +2910,11 @@ public partial class ValidatorCache
 }
 namespace Microsoft.AspNetCore.Mvc.Routing
 {
+    public abstract partial class DynamicRouteValueTransformer
+    {
+        protected DynamicRouteValueTransformer() { }
+        public abstract System.Threading.Tasks.Task<Microsoft.AspNetCore.Routing.RouteValueDictionary> TransformAsync(Microsoft.AspNetCore.Http.HttpContext httpContext, Microsoft.AspNetCore.Routing.RouteValueDictionary values);
```
davidfowl: ValueTask

rynowak: Why? Is the guidance to just use ValueTask **everywhere** now?

rynowak:
@JamesNK @pranavkm - please take a look

src/Mvc/Mvc.Core/src/Builder/ControllerEndpointRouteBuilderExtensions.cs
```diff
@@ -447,6 +448,44 @@ public static ControllerActionEndpointConventionBuilder MapDefaultControllerRout
...
+            }
+
+            EnsureControllerServices(endpoints);
+
+            // Called for side-effect to make sure that the data source is registered.
+            GetOrCreateDataSource(endpoints).CreateInertEndpoints = true;
+
+            // Maps a dynamic controller endpoint with an empty delegate. This is OK because
+            // we don't expect the delegate to run.
+            endpoints.Map(pattern, context => Task.CompletedTask).Add(b =>
```
pranavkm: Should the delegate throw so we when stuff's broken?

src/Mvc/Mvc.Core/ref/Microsoft.AspNetCore.Mvc.Core.netcoreapp3.0.cs
```diff
@@ -14,6 +14,7 @@ public static partial class ControllerEndpointRouteBuilderExtensions
         public static Microsoft.AspNetCore.Builder.ControllerActionEndpointConventionBuilder MapControllerRoute(this Microsoft.AspNetCore.Routing.IEndpointRouteBuilder endpoints, string name, string pattern, object defaults = null, object constraints = null, object dataTokens = null) { throw null; }
         public static Microsoft.AspNetCore.Builder.ControllerActionEndpointConventionBuilder MapControllers(this Microsoft.AspNetCore.Routing.IEndpointRouteBuilder endpoints) { throw null; }
         public static Microsoft.AspNetCore.Builder.ControllerActionEndpointConventionBuilder MapDefaultControllerRoute(this Microsoft.AspNetCore.Routing.IEndpointRouteBuilder endpoints) { throw null; }
+        public static void MapDynamicControllerRoute<TTransformer>(this Microsoft.AspNetCore.Routing.IEndpointRouteBuilder endpoints, string pattern) where TTransformer : Microsoft.AspNetCore.Mvc.Routing.DynamicRouteValueTransformer { }
```
pranavkm: Should this return `IEndpointRouteBuilder` like the rest

src/Mvc/Mvc.Core/src/Routing/ActionEndpointFactory.cs
```diff
@@ -399,8 +397,47 @@ private static (RoutePattern resolvedRoutePattern, IDictionary<string, string> r
             {
                 perRouteConventions[i](builder);
             }
+        }
+
+        private static RequestDelegate CreateRequestDelegate()
+        {
+            // We don't want to close over the retrieve the Invoker Factory in ActionEndpointFactory as
```
pranavkm: > close over the ~~retrieve~~ the Invoker Factory

src/Mvc/Mvc.Core/src/Routing/DynamicControllerEndpointMatcherPolicy.cs
```diff
@@ -83,30 +89,56 @@ public Task ApplyAsync(HttpContext httpContext, CandidateSet candidates)
                 }
 
                 var endpoint = candidates[i].Endpoint;
+                var originalValues = candidates[i].Values;
+
+                RouteValueDictionary dynamicValues = null;
+
+                // We don't expext both of these to be provided, and they are internal so there's
```
pranavkm: ~~expext~~ expect

src/Mvc/Mvc.Core/src/Routing/DynamicControllerEndpointMatcherPolicy.cs
```diff
@@ -83,30 +89,56 @@ public Task ApplyAsync(HttpContext httpContext, CandidateSet candidates)
...
                 }
 
-                var matchedValues = candidates[i].Values;
-                var endpoints = _selector.SelectEndpoints(metadata.Values);
-                if (endpoints.Count == 0)
+                var endpoints = _selector.SelectEndpoints(dynamicValues);
+                if (endpoints.Count == 0 && dynamicControllerMetadata != null)
                 {
-                    // If there's no match this is a configuration error. We can't really check
+                    // If there's no match for a fallback is a configuration error. We can't really check
```
pranavkm: ~~If there's no~~ Having no a match for a fallback is a configuration error.

src/Mvc/Mvc.RazorPages/src/Builder/RazorPagesEndpointRouteBuilderExtensions.cs
```diff
@@ -279,6 +280,54 @@ public static IEndpointConventionBuilder MapFallbackToPage(this IEndpointRouteBu
...
+            {
+                throw new ArgumentNullException(nameof(endpoints));
+            }
+
+            if (pattern == null)
+            {
+                throw new ArgumentNullException(nameof(pattern));
+            }
+
+            if (endpoints == null)
```
JamesNK: One check good, two checks even better 🐴

src/Mvc/Mvc.Core/src/Routing/DynamicControllerEndpointMatcherPolicy.cs
```diff
@@ -83,30 +89,56 @@ public Task ApplyAsync(HttpContext httpContext, CandidateSet candidates)
...
+                {
+                    // Not a dynamic controller.
+                    continue;
+                }
 
-                var metadata = endpoint.Metadata.GetMetadata<DynamicControllerMetadata>();
-                if (metadata == null)
+                if (dynamicValues == null)
                 {
+                    candidates.ReplaceEndpoint(i, null, null);
```
pranavkm: What does this do?

src/Mvc/Mvc.Core/src/Routing/DynamicRouteValueTransformer.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Routing;
+
+namespace Microsoft.AspNetCore.Mvc.Routing
+{
+    public abstract class DynamicRouteValueTransformer
```
pranavkm: Documentation

src/Mvc/Mvc.RazorPages/src/Infrastructure/PageActionEndpointDataSource.cs
```diff
@@ -22,13 +22,17 @@ public PageActionEndpointDataSource(IActionDescriptorCollectionProvider actions,
...
 
-            // IMPORTANT: this needs to be the last thing we do in the constructor. 
+            // IMPORTANT: this needs to be the last thing we do in the constructor.
             // Change notifications can happen immediately!
             Subscribe();
         }
 
         public PageActionEndpointConventionBuilder DefaultBuilder { get; }
 
+        // Used to control whether we create 'inter' (non-routable) endpoints for use in dynamic
```
pranavkm: ~~inter~~ inert

src/Mvc/test/WebSites/RoutingWebSite/StartupForDynamic.cs
```diff
@@ -0,0 +1,66 @@
...
+
+            });
+
+            app.Map("/afterrouting", b => b.Run(c =>
+            {
+                return c.Response.WriteAsync("Hello from middleware after routing");
+            }));
+        }
+
+        private class Transformer : DynamicRouteValueTransformer
```
pranavkm: Nifty

src/Mvc/Mvc.Core/src/Routing/DynamicControllerEndpointMatcherPolicy.cs
JamesNK: My memory is hazy about what this PR use to have to what is has now.

What did you do to remove this?

pranavkm (approved):
👍

src/Mvc/Mvc.Core/src/Routing/DynamicControllerEndpointMatcherPolicy.cs
rynowak: We did a much more complex (and better) thing. We use `IEndpointSelectorPolicy` to disambiguate instead of `IActionConstraint`. This means that all of our built-in policies implement both the DFA mode and selector policy mode (already done).

src/Mvc/Mvc.Core/src/Builder/ControllerEndpointRouteBuilderExtensions.cs
rynowak: Sure.

rynowak: I just want to take a moment, and mention one thing. 

I agree that failing loudly in this case is better than silently. However, usually when this happens, the next piece of feedback is to provide an descriptive and meaningful error message for the code that we don't expect to ever run.

src/Mvc/Mvc.Core/ref/Microsoft.AspNetCore.Mvc.Core.netcoreapp3.0.cs
rynowak: **TLDR:** no because it's not useful

Everywhere else we have *dynamic* behaviour in routing, we don't expose a builder because there isn't anything useful you can do with it.

You might say.... hey, I want to use `RequireHost(...)` - that seems cool and good! It won't work, because by the time we're running policies we're looking at the *real* endpoint, not the one used to enable this functionality. So if you put `HostAttribute` on your action and then used a dynamic controller route to select it, then that would work, and it's **super** by-design that it works that way.

src/Mvc/Mvc.Core/src/Routing/DynamicControllerEndpointMatcherPolicy.cs
rynowak: It "nulls out" the match, in one go it sets the endpoint and rvd to null, and marks it as a non-match.

src/Mvc/Mvc.Core/ref/Microsoft.AspNetCore.Mvc.Core.netcoreapp3.0.cs
HKochniss: I think the guidance (as of "Stephen Toub says so") is that if the method is mostly expected to return synchronously, use a ValueTask, but if the method is expected to do real async work (like let's say > 20% of the time, but I'm not sure where I read it, but that's somehow the number in my head) use Task, as ValueTask comes with it's own struct-based overhead and just wraps a real Task in that case..

davidfowl: `ValueTask<T>`  is still the guidance, there's almost no reason to use `Task<T>` as it forces an allocation. You can also now reuse allocations with `ValueTask<T>`  so it's an extensibility point as well.

> Why? Is the guidance to just use ValueTask everywhere now?

Yes, generic `ValueTask<T>` absolutely, for `ValueTask` is more questionable and has more overhead and is only really valuable as an extensibility point (using `IValueTaskSource`)

HKochniss: hrm didn't sound like it here: https://devblogs.microsoft.com/dotnet/understanding-the-whys-whats-and-whens-of-valuetask/#user-content-should-every-new-asynchronous-api-return-valuetask--valuetasktresult

" There are also some minor costs associated with returning a ValueTask<TResult> instead of a Task<TResult>, e.g. in microbenchmarks it’s a bit faster to await a Task<TResult> than it is to await a ValueTask<TResult>, so if you can use cached tasks (e.g. you’re API returns Task or Task<bool>), you might be better off performance-wise sticking with Task and Task<bool>. ValueTask / ValueTask<TResult> are also multiple words in size, and so when these are awaitd and a field for them is stored in a calling async method’s state machine, they’ll take up a little more space in that state machine object." an yadda.. tbh it doesn#t feel like a complete replacement for Task everywhere.

davidfowl: > However, ValueTask / ValueTask<TResult> are great choices when a) you expect consumers of your API to only await them directly, b) allocation-related overhead is important to avoid for your API, and c) either you expect synchronous completion to be a very common case, or you’re able to effectively pool objects for use with asynchronous completion. When adding abstract, virtual, or interface methods, you also need to consider whether these situations will exist for overrides/implementations of that method.

If you have a virtual member, you can't assume what derived implementations will do. To optimize for common synchronous cases and offer extensibility for derived types, use `ValueTask<T>`.

The framework is the one that is awaiting this so it's mostly up to the implementation to just return synchronously or asynchronously. The other downsides don't apply.

I really want to go back the change a bunch of APIs to return `ValueTask<T>` 😄 . I could squash so many needless allocations in the auth stack..

HKochniss: Appreciate the time you took to answer that. Hopefully someone finds this hidden conversation, reads it and learns, too.

Issue #8536, opened 2019-03-14
Add extensible handlers to DeveloperExceptionPage


DamianEdwards:
Add extensibility point to sub-systems to add handlers via DI that get inspected when an exception occurs. The handler can choose to render a response (e.g. the current EF diagnostic middleware page would change to this) or pass-through (do nothing). The middleware would fallback to rendering an HTML error page if no handler matched and the client sent an Accept header that included text/html, otherwise it would just return the error as plain text. For API responses, MVC would register a handler that formats a `ProblemDetails` with the relevant information according to the configured formatters.

There's 2 work items here:
- [x] - Allowing components to take over the rendering of the error page depending on the exception type (like how EF has the database error page)
- [x] - Rendering a plaintext response if the client isn't a browser (if the Accept header does not contain text/html)

davidfowl:
As a strawman design for the first part of this issue, we could introduce a simple interface like this:

```C#
public interface IDeveloperPageExceptionFilter
{
    Task HandleExceptionAsync(HttpContext context, Exception exception, Func<HttpContext, Exception, Task> next);
}
```

This follows the filter pattern we have elsewhere in the product (middleware, startup filters etc)

- Short circuiting the default logic means you don't call next
- With access to the Exception and next you can change the exception type in a filter and call next with a more specific exception (it would allow for doing things like exception unwrapping)

We could provide an `ExceptionContext` in case we wanted to expand it further in the future with more data (it clashes with the ExceptionContext type in MVC though so maybe a different name).

cc @rynowak

rynowak:
This pattern seems OK, how do we decide on the order? Are these in DI or options? What's the lifetime? Those are some of the rough edges in MVC's filters. If it's based on options then the answers are pretty straightforward IMO.

+1 for the idea of chain of responsibility here. Is the idea that we register the current behaviour as the last filter?

I'm not sure the idea of transforming the exception is valuable. I think it's fine if the contract allows it, but I don't see why its necessary.

I don't think we need a context here. If we get that far, we might want to consider using features or enhancing existing contracts on http context.

note: since nothing clears the endpoint on throw - you could use endpoint metadata in one of these filters.

davidfowl:
> This pattern seems OK, how do we decide on the order?

DI order, like IStartupFilters. I don't think the order will matter that much in practice as I'm hoping the majority use case is to filter on exception type if you're not wholesale changing exception formatting (then you just need to be the first).

> Are these in DI or options?

DI.

[...]

PR #8958, opened 2019-04-01
Initial design for exception page filters

davidfowl:
- This change introduces the concept of an IDeveloperPageException filter that runs whenever the developer exception page has encountered an error. It follows the middleware pattern (chain of resposibility) which allows short circuiting or decorating the default logic.
- Added tests

Contributes to #8536


src/Middleware/Diagnostics/src/DeveloperExceptionPage/IDeveloperPageExceptionFilter.cs
```diff
@@ -0,0 +1,13 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Diagnostics
+{
+    public interface IDeveloperPageExceptionFilter
```
davidfowl: A few questions:
- Should this be in the abstractions package and be renamed to something more general? I don't want to give the impression that these are ASP.NET Core wide exception filters. These only apply to the developer exception page.
- Do we need something similar for the error handler? It already has a way to handle the exception and get at the error (via a feature).

rynowak: I don't think personally think there's much point to abstractions packages anymore. Open to ideas or discussion about the pros and cons, but since we don't ship packages I don't think there are many cons to either choice.

I think the name makes it really clear that it only applies to the developer exception page.

If we add it to the error handler, then it **does**  become ASP.NET wide. However I'm not sure we'd do that. The reason why we're doing this for developer exception page is that we want to plug in more default behaviours and don't want you to end up with 20 lines of code in startup for it. I'm not sure we'd add more default behaviours to error handler. Hearing arguments for that would change my mind 😁

src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
```diff
@@ -40,12 +41,14 @@ public class DeveloperExceptionPageMiddleware
...
         /// <param name="diagnosticSource"></param>
+        /// <param name="filters"></param>
         public DeveloperExceptionPageMiddleware(
             RequestDelegate next,
             IOptions<DeveloperExceptionPageOptions> options,
             ILoggerFactory loggerFactory,
             IWebHostEnvironment hostingEnvironment,
-            DiagnosticSource diagnosticSource)
+            DiagnosticSource diagnosticSource,
+            IEnumerable<IDeveloperPageExceptionFilter> filters)
```
rynowak: breaking change yo

src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
```diff
@@ -40,12 +41,14 @@ public class DeveloperExceptionPageMiddleware
         /// <param name="loggerFactory"></param>
         /// <param name="hostingEnvironment"></param>
         /// <param name="diagnosticSource"></param>
+        /// <param name="filters"></param>
```
rynowak: this are some hq docs

src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
davidfowl: Yes, I know but it's 3.0 and I'm expecting this to be acceptable. I didn't update the refs tho because who ever remembers that the first time 😄

src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
```diff
@@ -63,6 +66,13 @@ public class DeveloperExceptionPageMiddleware
             _fileProvider = _options.FileProvider ?? hostingEnvironment.ContentRootFileProvider;
             _diagnosticSource = diagnosticSource;
             _exceptionDetailsProvider = new ExceptionDetailsProvider(_fileProvider, _options.SourceCodeLineCount);
+            _exceptionHandler = DisplayException;
+
+            foreach (var filter in filters.Reverse())
```
rynowak: needs a null check on filtars

src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
```diff
@@ -63,6 +66,13 @@ public class DeveloperExceptionPageMiddleware
...
+            {
+                var nextFilter = _exceptionHandler;
+                _exceptionHandler = (context, exception) => filter.HandleExceptionAsync(context, exception, nextFilter);
+            }
```
rynowak: the CoR is so elegant and cool

src/Middleware/Diagnostics/test/UnitTests/DeveloperExceptionPageMiddlewareTest.cs
```diff
@@ -3,15 +3,14 @@
...
-using System.Linq;
 using System.Diagnostics;
-using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.TestHost;
 using Microsoft.Extensions.DependencyInjection;
 using Xunit;
+using Microsoft.AspNetCore.Http;
```
rynowak: sort

src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
davidfowl: In case somebody news this up manually?

rynowak: other params has it. I'm not going to make a fuss, I mentioned it in case you just plum forgot. I tend to do this these days because I don't want people writing unit tests that partially pass in stuff.

davidfowl: I added it. But I wasn't sure if we were unit testing `IE<T>` because we think people will pass null manually or something else.

src/Middleware/Diagnostics/ref/Microsoft.AspNetCore.Diagnostics.netcoreapp3.0.cs
```diff
@@ -59,7 +59,7 @@ namespace Microsoft.AspNetCore.Diagnostics
 {
     public partial class DeveloperExceptionPageMiddleware
     {
-        public DeveloperExceptionPageMiddleware(Microsoft.AspNetCore.Http.RequestDelegate next, Microsoft.Extensions.Options.IOptions<Microsoft.AspNetCore.Builder.DeveloperExceptionPageOptions> options, Microsoft.Extensions.Logging.ILoggerFactory loggerFactory, Microsoft.AspNetCore.Hosting.IWebHostEnvironment hostingEnvironment, System.Diagnostics.DiagnosticSource diagnosticSource) { }
+        public DeveloperExceptionPageMiddleware(Microsoft.AspNetCore.Http.RequestDelegate next, Microsoft.Extensions.Options.IOptions<Microsoft.AspNetCore.Builder.DeveloperExceptionPageOptions> options, Microsoft.Extensions.Logging.ILoggerFactory loggerFactory, Microsoft.AspNetCore.Hosting.IWebHostEnvironment hostingEnvironment, System.Diagnostics.DiagnosticSource diagnosticSource, System.Collections.Generic.IEnumerable<Microsoft.AspNetCore.Diagnostics.IDeveloperPageExceptionFilter> filters) { }
```
davidfowl: This is a breaking change. If you want to revolt, tell me and I'll add an overload.

src/Middleware/Diagnostics/src/DeveloperExceptionPage/IDeveloperPageExceptionFilter.cs
```diff
@@ -0,0 +1,13 @@
...
+    {
+        Task HandleExceptionAsync(HttpContext context, Exception exception, Func<HttpContext, Exception, Task> next);
```
Tratcher: An error context object would be preferred. It simplifies the signature and gives you future proofing.
`Task HandleExceptionAsync(ErrorContext context, Func<ErrorContext, Task> next);`

ErrorContext:
- HttpContext
- Exception

src/Middleware/Diagnostics/src/DeveloperExceptionPage/IDeveloperPageExceptionFilter.cs
```diff
@@ -0,0 +1,13 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Diagnostics
```
Tratcher: Add a sample that implements the interface, does content negotiation, and returns a JSON formatted error.

davidfowl: Does this repo have samples that aren't tests? I haven't found them. Where are they?

src/Middleware/Diagnostics/src/DeveloperExceptionPage/IDeveloperPageExceptionFilter.cs
davidfowl: Sure I'll add the an ErrorContext. I need to say it though, these types should really be structs (generally). It's such a waste to allocate an object that wraps a bunch of other objects. This is an error path and is only used in development so I care less.

davidfowl: Also, mutable or immutable? I'd like to make it immutable but then it brings up the question on adding another constructor every time we want to add a field.

src/Middleware/Diagnostics/src/DeveloperExceptionPage/IDeveloperPageExceptionFilter.cs
Tratcher: Oh, I didn't realize those had been dropped. We used to have all of these: https://github.com/aspnet/Diagnostics/tree/master/samples

src/Middleware/Diagnostics/src/DeveloperExceptionPage/IDeveloperPageExceptionFilter.cs
Tratcher: Mutable, it's easier to test and you don't need any constructors.

davidfowl: Made it immutable. Its cleaner as both of those parameters are required to be non null.

src/Middleware/Diagnostics/src/DeveloperExceptionPage/IDeveloperPageExceptionFilter.cs
davidfowl: Good old days. It is unfortunate we have no way to f5 this.

src/Middleware/Diagnostics/src/DeveloperExceptionPage/IDeveloperPageExceptionFilter.cs
davidfowl: Agreed. I was just posing these questions regardless.

src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
```diff
@@ -125,15 +125,15 @@ public async Task Invoke(HttpContext context)
         }
 
         // Assumes the response headers have not been sent.  If they have, still attempt to write to the body.
-        private Task DisplayException(HttpContext context, Exception ex)
+        private Task DisplayException(ErrorContext errorContext)
         {
-            var compilationException = ex as ICompilationException;
+            var compilationException = errorContext.Exception as ICompilationException;
```
Tratcher: Factor out DisplayCompilationException as a filter? Even if you don't register it via DI.

davidfowl: Why?

Tratcher: Proof of concept. This is doing exactly what filters are designed to do.

davidfowl: I don't think it makes any sense. The thing left to resolve is the ordering concern that @rynowak brought up. The slight code make the diff bigger and don't add any value.

src/Middleware/Diagnostics.Abstractions/src/IDeveloperPageExceptionFilter.cs
```diff
@@ -0,0 +1,21 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Diagnostics
+{
+    /// <summary>
+    /// Provides an extensiblity point for changing the behavior of the DeveloperExceptionPageMiddleware />.
```
Eilon: Bad close-angle-bracket.

src/Middleware/Diagnostics.Abstractions/src/IDeveloperPageExceptionFilter.cs
```diff
@@ -0,0 +1,21 @@
...
+    /// </summary>
+    public interface IDeveloperPageExceptionFilter
+    {
+        /// <summary>
+        /// An exception handling method.
```
Eilon: When you do get around to docs, it would be good to explain what I'm supposed to do in this method. Should I always call `next`, or only sometimes?

Eilon (approved):
I dig the design.

src/Middleware/Diagnostics.Abstractions/src/ErrorContext.cs
```diff
@@ -0,0 +1,17 @@
+using System;
```
JamesNK: Add copyright

src/Middleware/Diagnostics.Abstractions/src/ErrorContext.cs
```diff
@@ -0,0 +1,17 @@
...
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Diagnostics
+{
+    public class ErrorContext
```
JamesNK: Add XML docs

src/Middleware/Diagnostics.Abstractions/src/IDeveloperPageExceptionFilter.cs
```diff
@@ -0,0 +1,21 @@
+using System;
```
JamesNK: Add copyright

src/Middleware/Diagnostics.Abstractions/src/IDeveloperPageExceptionFilter.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    /// <summary>
+    /// Provides an extensiblity point for changing the behavior of the DeveloperExceptionPageMiddleware.
+    /// </summary>
+    public interface IDeveloperPageExceptionFilter
+    {
+        /// <summary>
+        /// An exception handling method that is used to either format the exception or delegate to the next handler in the chain.
+        /// </summary>
+        /// <param name="errorContext">The error context</param>
```
JamesNK: nit: missing full stop

(and line below)

src/Middleware/Diagnostics/ref/Microsoft.AspNetCore.Diagnostics.netcoreapp3.0.cs
DamianEdwards: Do we really think there are many (any?) folks newing up this middleware class themselves? That said, generally speaking we should lean to remaining compatible for all changes unless we have good reason not to.

davidfowl: > Do we really think there are many (any?) folks newing up this middleware class themselves?

No.

> That said, generally speaking we should lean to remaining compatible for all changes unless we have good reason not to.

We may have to live with this one. Adding another constructor results in the system picking the "wrong" one. I think the risk is small enough to warrant it. I'm investigating anyways.

JamesNK: Added the breaking change label 👼

Tratcher: It picks the first constructor it can satisfy, put the new one on top.

PR #8960, opened 2019-04-01
Support triggering rendering from OnAfterRenderAsync. Fixes #8435

### LINKED ISSUE #8435: Doing something that requires a re-render in OnAfterRender doesn't seem to work
### Describe the bug
Since you can't (I think) depend on component references to be valid until OnAfterRender, that seems to be the only place you can make calls to your components to do any manual setup. But, if any of those requires a re-render, it doesn't seem to work (I'm on the 0.9.0 blazor stuff in VS Preview 4.1.)

Does the 'need to re-render' flag get cleared upon return from OnAfterRender()? Or is it not legal to do things that will require a re-render in OnAfterRender? If the latter I would think another callback needs to be made of the DoComponentSetup() type or something.

If I make calls to components in OnAfterRender() and they call StateHasChanged(), those changes aren't rendered. Only the stuff (presumably) that where rendered before OnAfterRender() area actually rendered.

Before going through the hassle of trying to create some reproduceable bit, I just wanted to make sure it is supposed to be OK to do this kind of thing. If not, then it's just that we need some sort of component setup callback. If it's supposed to work, maybe someone else can sanity check me, and I can try to create a small scenario that demonstrates it.

The particular scenario is setting the selected row in a list box component. We can't know that until we get initialized and get the data loaded. And then we can't actually set it until the component references are valid. So I'm setting them in OnAfterRender, and calling StateHasChanged(), but those changes are never rendered in the initial display of the screen. The selection visually stays on the 0th row. I can see in the debugger that the call is made and the target index is stored and StateHasChanged() is called. 

But she no redraw.

SteveSandersonMS:
Thanks to @DeanRoddey for reporting this.

This PR fixes the case where you invoke `StateHasChanged` synchronously inside `OnAfterRenderAsync`. Previously such render requests were ignored, because our post-batch cleanup logic was erasing the knowledge that another render had been requested.

Normally this wouldn't have been massively urgent because it's quite an edge case, but pretty soon this will *not* be an edge case. This re-render-from-`OnAfterRenderAsync` technique is central to our plan for the idiomatic way to handle #8786. As such I'm keen to include this fix immediately.

The real value of this PR is mostly its test cases. The actual implementation details of how the renderer was fixed are a minor detail that will probably change anyway with @javiercn's much bigger PR. But the test cases should remain relevant.

### DIFF
--- a/src/Components/Components/perf/RenderTreeDiffBuilderBenchmark.cs
+++ b/src/Components/Components/perf/RenderTreeDiffBuilderBenchmark.cs
@@ -79,7 +79,7 @@ namespace Microsoft.AspNetCore.Components.Performance
         [Benchmark(Description = "RenderTreeDiffBuilder: Input and validation on a single form field.", Baseline = true)]
         public void ComputeDiff_SingleFormField()
         {
-            builder.Clear();
+            builder.ClearStateForCurrentBatch();
             var diff = RenderTreeDiffBuilder.ComputeDiff(renderer, builder, 0, original.GetFrames(), modified.GetFrames());
             GC.KeepAlive(diff);
         }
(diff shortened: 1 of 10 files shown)


javiercn (approved):
LGTM

PR #8967, opened 2019-04-01
Add validation to ensure Cookie.Expiration is not set

HaoK:
Fixes https://github.com/aspnet/AspNetCore/issues/4671


Tratcher (approved):
This fails on the first request, right?

HaoK:
Sorta, it will fail the first time cookie authentication is accessed.  We don't have eager options validation, it will be the first time cookie options is requested

HaoK:
So for the default application cookie it will get hit immediately, if they set it on some random cookie they don't always use, the error won't surface for a while

PR #8988, opened 2019-04-02
Reduce boxing in RedirectRule.cs

cypressious:
`QueryString` is a struct and when concatenating it with a `string` it will be boxed. The `ToString()` method of `QueryString` just calls `ToUriComponent()` so call it directly to prevent boxing.


jkotalik (approved):

jkotalik:
Thanks!

PR #8994, opened 2019-04-02
Test name fixed

### LINKED ISSUE #8922: Preflight CORS request returning NoContent instead of Forbidden
### Describe the bug
I have seen other frameworks return a 403 when preflight cors requests fail. According to the name of this test that may be the intention although the assert is for a no content status.
https://github.com/aspnet/AspNetCore/blob/6dc28d8a4ecb53a1890d0c14c9914b46dc0573b8/src/Middleware/CORS/test/UnitTests/CorsMiddlewareTests.cs#L261

### To Reproduce
1. https://github.com/aspnet/AspNetCore/blob/6dc28d8a4ecb53a1890d0c14c9914b46dc0573b8/src/Middleware/CORS/test/UnitTests/CorsMiddlewareTests.cs#L261
2. Test name doesn't match assert

### Expected behavior
Expected to assert against 403(Forbidden)

flash2048:
Test name fixed #8922

### DIFF
--- a/src/Middleware/CORS/test/UnitTests/CorsMiddlewareTests.cs
+++ b/src/Middleware/CORS/test/UnitTests/CorsMiddlewareTests.cs
@@ -258,7 +258,7 @@ namespace Microsoft.AspNetCore.Cors.Infrastructure
         }
 
         [Fact]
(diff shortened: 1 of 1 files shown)


pranavkm (approved):

pranavkm:
Thanks for the PR!

PR #9000, opened 2019-04-02
Add missing using statement to Blazor hosted template.

danroth27:
Fixes compilation error in Startup.cs:

'IWebHostEnvironment' does not contain a definition for 'IsDevelopment' and the best extension method overload 'HostingEnvironmentExtensions.IsDevelopment(IHostingEnvironment)' requires a receiver of type 'IHostingEnvironment'


rynowak:
@davidfowl @Tratcher - is this right? We expect this using to show up in templates or are we doing a wierd?

Tratcher (approved):

davidfowl:
Yea this is right.

davidfowl (approved):

davidfowl:
Do we have this in all of the templates @danroth27 ?

danroth27:
> Do we have this in all of the templates @danroth27 ?

@davidfowl As appropriate, yes. The Razor Components template already has it. The Blazor client-side template doesn't need it.

davidfowl:
I meant the non Blazor templates

danroth27:
The non-Blazor templates look fine.

PR #9002, opened 2019-04-02
Do not specify `name` of the default (hosted) pools

dougbu:
- aspnet/AspNetCore#8862
- update e2e-test.yml to use modern AzDO job schema
  - https://github.com/dotnet/arcade/blob/master/Documentation/AzureDevOps/PhaseToJobSchemaChange.md
- use BYOC queues on Windows; part of aspnet/AspNetCore-Internal#2033
- pick up latest release/2.1 commit for aspnet/EntityFrameworkCore for related changes

### DIFF
--- a/.azure/pipelines/e2e-tests.yml
+++ b/.azure/pipelines/e2e-tests.yml
@@ -1,9 +1,10 @@
 trigger: none
-phases:
-- phase: Host_Windows
-  queue:
-    name: Hosted VS2017
-    parallel: 8
+
+jobs:
+- job: Host_Windows
+  pool:
+    vmImage: vs2017-win2016
+  strategy:
     matrix:
       Portable_Node8:
         Test.RuntimeIdentifier: none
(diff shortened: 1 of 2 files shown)


.azure/pipelines/e2e-tests.yml
```diff
@@ -44,10 +46,10 @@ phases:
     inputs:
       testRunner: vstest
       testResultsFiles: 'artifacts/logs/**/*.trx'
-- phase: Host_macOS
-  queue:
-    name: Hosted macOS
-    parallel: 4
```
ryanbrandenburg: Looks like we lost the `parallel` parameters, was that intentional?

ryanbrandenburg (approved):
Seems fine so long as the parallel change was intended.

.azure/pipelines/e2e-tests.yml
dougbu: Hmm, may actually be useful here since we're not using the default-build.yml templates. I'll add `maxParallel: 4` to the strategies where I deleted `parallel: 4`.

PR #9003, opened 2019-04-02
Turn on new confirmation flow in templates

HaoK:
Depends on https://github.com/aspnet/AspNetCore/pull/8577


ryanbrandenburg (commented):
The code looks fine but are we sure this is a scenario we want as a default experience for the templates?

HaoK:
@blowdart ?

blowdart:
As discussed, after branching happens.

HaoK:
@blowdart still good to merge and turn this on for preview 7?

blowdart:
Yes, we want the feedback.

blowdart (approved):
Do it, let's get the feedback

Issue #8347, opened 2019-03-08
Remove unused non-default namespaces in template Startup.cs files


Eilon:
Examples:
* https://github.com/aspnet/AspNetCore/blob/f4c80ca99f42cd37dd8128e44eacecc0e584bc51/src/ProjectTemplates/Web.ProjectTemplates/content/Worker-CSharp/Program.cs#L3-L9
   * `System.IO` is not a default namespace, and not used
   * `Microsoft.Extensions.Configuration` and `Microsoft.Extensions.Logging` not used
* https://github.com/aspnet/AspNetCore/blob/f4c80ca99f42cd37dd8128e44eacecc0e584bc51/src/ProjectTemplates/Web.ProjectTemplates/content/StarterWeb-CSharp/Program.cs#L6-L10
   * `Microsoft.AspNetCore`, `Microsoft.Extensions.Configuration`, and `Microsoft.Extensions.Logging` not used
* https://github.com/aspnet/AspNetCore/blob/f4c80ca99f42cd37dd8128e44eacecc0e584bc51/src/ProjectTemplates/Web.ProjectTemplates/content/StarterWeb-CSharp/Startup.cs#L26
   * `Microsoft.AspNetCore.HttpsPolicy` not used
* https://github.com/aspnet/AspNetCore/blob/f4c80ca99f42cd37dd8128e44eacecc0e584bc51/src/ProjectTemplates/Web.ProjectTemplates/content/StarterWeb-CSharp/Startup.cs#L28
   * `Microsoft.AspNetCore.Mvc` not used

And other similar cases in other templates.

PR #9005, opened 2019-04-02
Remove unused `using`-s from Program and Startup classes

mkArtakMSFT:
Addresses #8347

### DIFF
--- a/src/ProjectTemplates/Web.ProjectTemplates/content/EmptyWeb-CSharp/Program.cs
+++ b/src/ProjectTemplates/Web.ProjectTemplates/content/EmptyWeb-CSharp/Program.cs
@@ -3,7 +3,9 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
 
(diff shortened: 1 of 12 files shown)


javiercn (approved):
It's nearly impossible to validate this PR by hand. I'm ok with it so long as the template tests pass. If something breaks we'll get pinged.

ryanbrandenburg (approved):
What @javiercn said. Each of these have tests that compile them, if those pass you're golden.

Eilon (approved):
Looks good! Hopefully it's correct 😄

davidfowl (changes requested):
This isn't a good change. We actually want some of these because of extension methods that exist in various places.

mkArtakMSFT:
@davidfowl, can you be more specific please. Which particular namespaces you'd like to keep?
Here I'm removing only those which aren't used.

davidfowl:
```
using Microsoft.Extensions.Configuration;
```

I can't open up VS now because I'm updating but for example, in Program.cs the logging extensions methods are in Microsoft.Extensions.Logging (AddConsole, AddEventLog etc) and the configuration extension methods are in Microsoft.Extensions.Configuration. Having those unused using these make it easier to find logging providers and configuration providers without having to type it out by hand.

mkArtakMSFT:
Sounds good, @davidfowl. I'll leave these two usings untouched.

davidfowl (approved):

PR #9025, opened 2019-04-03
Server-side Blazor template cleanups for Preview 4

SteveSandersonMS:
Some parts of this should be understood as proposals for discussion. I'm OK with changing this further. The commits in this PR are:

 * Rename `Host.cshtml` to `_Host.cshtml`
    * Is this a legitimate change? My goal is to make the file layout more comprehensible. I find it jarring that we have `Host.cshtml` in the same `Pages` directory as the component pages, and yet we expect people to understand that "host" is in its own special category. By prefixing with `_`, (1) it becomes more visually obvious that it's not a regular file like the others there, and (2) with alphabetical sorting, it moves up to the top alongside `_Imports.razor`, making the groupings easier to parse visually.
    * If I understand correctly, the leading-underscore-on-cshtml convention usually means "not something users can navigate to", which is true in this case, in the sense that visitors *can't* go to `yoursite/Host` in typical cases. This file is an implementation detail of how your components are hosted.
 * **Update: reverted this one for reasons given by @javiercn below** ~~In `Startup.cs`, change `endpoints.MapComponentHub();` to `endpoints.MapComponentHub<App>("app");`~~
    * ~~During preview 4 (and maybe preview 5) I think this is important for discoverability. If people want to disable prerendering, or even more so if they want to use a plain `.html` file as their host page, they have to specify the root component name and selector. But that will be almost impossible to discover unless we put this code here by default.~~
    * ~~I don't think this constitutes any major drawback for the out-of-box case where it's not strictly necessary.~~
    * ~~Longer term, this could change to something like `endpoints.MapComponentHub().AddRootComponent<App>("app");`, though maybe we would decide the discoverability is not a problem once we have some way to emit a non-prerendered placeholder from `_Host.cshtml`, and then just remove it from the default template.~~
 * Remove unneccesary dependency on NewtonsoftJson package
    * Nothing to discuss with that
 * Remove unnecessary MapRazorPages and make blank lines consistent 
    * I think this is non-controversial too
 * Rename `Services` subdirectory to `Data`
    * I know this might sound shallow, but this is something I do care about: it really annoyed me that alphabetically, `Services` was splitting up `Pages` and `Shared` even though the latter two are close friends and `Services` is quite different. In any case, I also found it the assumption that people should have `Services` as an architectural pattern around data was too specific as well. The directory still contains a `WeatherForecastService` so it's still leaning that way, just not as strongly now.
    * I also considered naming the directory `DomainModel` though I'm worried this might lead people to make presumptions about DDD somehow being the one true sanctioned architectural philosophy
    * Overall, I think our goal should be to avoid implying you should organize by file type or role. Almost everyone who builds larger apps recognizes the preference to group by feature area and not as `Controllers`, `Models`, `Partials`, `Pages`, `Services`, `Utils`, `css`, `js`, `images`, etc. At a higher level, splitting by domainmodel-vs-UI is pretty well established though, and often these two are in different projects.
    * There is prior art for `Data` in our templates - see `StarterWeb-CSharp`
 * Remove favicon.ico
    * This was previously a workaround for a limitation in Preview 3 and before. It's just not needed now, and removing it makes this template more consistent with others.


javiercn:
> * Rename `Host.cshtml` to `_Host.cshtml`

I'm ok with this change, the benefits far outweight the drawbacks.

> In `Startup.cs`, change `endpoints.MapComponentHub();` to `endpoints.MapComponentHub<App>("app");`
> * During preview 4 (and maybe preview 5) I think this is important for discoverability. If people want to disable prerendering, or even more so if they want to use a plain `.html` file as their host page, they have to specify the root component name and selector. But that will be almost impossible to discover unless we put this code here by default.
> * I don't think this constitutes any major drawback for the out-of-box case where it's not strictly necessary.
> * Longer term, this could change to something like `endpoints.MapComponentHub().AddRootComponent<App>("app");`, though maybe we would decide the discoverability is not a problem once we have some way to emit a non-prerendered placeholder from `_Host.cshtml`, and then just remove it from the default template.

I don't agree with this reasoning. It's far better for us to ship an experience that looks closer to how we want things to look long term and that exercises our features to get a chance to get feedback on them. I think that for server-side blazor the cshtml entry point is way more important than the `.html`.

May I suggest that instead of changing the method in the template, we keep defaulting to the prerender case and add a comment on top of MapComponentHub that says.

```csharp
// Replace MapComponentHub() for MapComponentHub<App>("app")
// and remove Html.RenderComponentAsync from your _Host.cshtml 
// if you don't want the application to be prerendered.
```

> * WeatherForecastService

call it repository or weatherforecastdata?

> * I'm worried this might lead people to make presumptions about DDD somehow being the one true sanctioned architectural philosophy

But it is 😝 

I would not use the data folder because that is typically reserved for EF stuff.

src/ProjectTemplates/Web.ProjectTemplates/content/RazorComponentsWeb-CSharp/Startup.cs
```diff
@@ -45,16 +42,16 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
...
             app.UseStaticFiles();
 
             app.UseRouting();
 
             app.UseEndpoints(endpoints =>
             {
-                endpoints.MapRazorPages();
-                endpoints.MapComponentHub();
-                endpoints.MapFallbackToPage("/Host");
+                endpoints.MapComponentHub<App>("app");
```
javiercn: If you do this, this will throw away the data from the prerendered circuit and will likely cause wreck havoc on the client as two renderers are trying to render on the same node.

SteveSandersonMS: Ok, I’ll revert this bit then. 

I really don’t want the suggested big comment in the template, so we’ll just have to accept that disabling prerendering is now a nonobvious process. It should get a bit better when there’s a way of emitting the placeholder without prerendering.

SteveSandersonMS:
> I would not use the data folder because that is typically reserved for EF stuff.

That makes it a good thing! Yes, people can and should put their EF stuff here too.

rynowak:
> Remove unnecessary MapRazorPages and make blank lines consistent
> I think this is non-controversial too

FYI the reason why this works is that `MapFallbackToPage` includes it. I think it's fine.

rynowak:
I'm strongly supportive of getting rid of `Services`. The fact that it's there makes it seem like a recommendation from the team and it isn't. My preference would be to have these files in the root, but I'm sure people would object to that also. I think what you're proposing is an improvement.

rynowak:
> If I understand correctly, the leading-underscore-on-cshtml convention usually means "not something users can navigate to", which is true in this case, in the sense that visitors can't go to yoursite/Host in typical cases. This file is an implementation detail of how your components are hosted.

I think this is the right semantic. I'm impressed that it works with the fallback route, and I'm not really sure how it does... but good job me I suppose?

rynowak (approved):
:shipit: from me

SteveSandersonMS:
OK thanks, now I just want some indication of non-objection from @danroth27 then I'll merge.

danroth27:
> OK thanks, now I just want some indication of non-objection from @danroth27 then I'll merge.

Looking now

danroth27 (approved):
LGTM

SteveSandersonMS:
Thanks everyone.

PR #9026, opened 2019-04-03
Fix double connection bug

javiercn:


SteveSandersonMS (approved):
An impressively quick fix!

javiercn:
@SteveSandersonMS There were some issues here that I found. I've fixed them

Issue #8999, opened 2019-04-02
Package Microsoft.AspNetCore.Blazor.Cli 0.9.0-preview4-19201-04 is not compatible with netcoreapp2.2


danroth27:
Repro steps:
- .NET Core 3.0 SDK: 3.0.100-preview4-011022
- Blazor VSIX: 16.0.19201.4
- Create new Blazor app
- Workaround: update csproj 

  ```xml
  <Project Sdk="Microsoft.NET.Sdk.Web">
  
    <PropertyGroup>
      <TargetFramework>netstandard2.0</TargetFramework>
      <RunCommand>dotnet</RunCommand>
      <RunArguments>blazor serve</RunArguments>
      <RestoreAdditionalProjectSources>
        https://dotnet.myget.org/F/aspnetcore-dev/api/v3/index.json;
        https://dotnet.myget.org/F/blazor-dev/api/v3/index.json;
      </RestoreAdditionalProjectSources>
      <LangVersion>7.3</LangVersion>
      <RazorLangVersion>3.0</RazorLangVersion>
      <AddRazorSupportForMvc>false</AddRazorSupportForMvc>
      <GenerateRazorAssemblyInfo>false</GenerateRazorAssemblyInfo>
    </PropertyGroup>
  
    <ItemGroup>
      <PackageReference Include="Microsoft.AspNetCore.Blazor" Version="0.9.0-preview4-19201-04" />
      <PackageReference Include="Microsoft.AspNetCore.Blazor.Build" Version="0.9.0-preview4-19201-04" PrivateAssets="all" />
  
      <DotNetCliToolReference Include="Microsoft.AspNetCore.Blazor.Cli" Version="0.9.0-preview4-19201-04" />
    </ItemGroup>
  
  </Project>
  ```
- Build and run

Actual result:

Package Microsoft.AspNetCore.Blazor.Cli 0.9.0-preview4-19201-04 is not compatible with netcoreapp2.2 (.NETCoreApp,Version=v2.2). Package Microsoft.AspNetCore.Blazor.Cli 0.9.0-preview4-19201-04 supports: netcoreapp3.0 (.NETCoreApp,Version=v3.0)

danroth27:
@SteveSandersonMS

placetobejohan:
I was able to solve this by checking the option **Use previews of the .Net Core SDK** under Tools --> Options --> Projects and Solutions --> .NET Core in VS 2019.

SteveSandersonMS:
[...]

PR #9028, opened 2019-04-03
Replace Blazor CLI tool with new DevServer package

### LINKED ISSUE #8243: Replace Microsoft.AspNetCore.Blazor.Cli with new .DevServer package
To align with future .NET Core SDK changes, we need to replace our `<DotNetCliToolReference>` in the Blazor standalone app with a regular `<PackageReference>` that bundles the dev server.

This was kindly prototyped for us by @wli3 at https://github.com/aspnet/AspNetCore/compare/master...wli3:close-to-blazor-build2?expand=1

SteveSandersonMS:
Fixes #8243 and #8999

We weren't planning to do this so soon, but it seems like the .NET Core SDK has now dropped support for post-2.2 DotNetCliToolReference packages. Even if that's not the case and some other workaround was possible, we did want to make this change for the long term anyway.

The end-user difference is that the Blazor client-side (standalone) template will no longer have a `<DotNetCliToolReference>`, but instead changes as follows: https://github.com/aspnet/AspNetCore/compare/stevesa/replace-cli-tool-with-devserver-package?expand=1#diff-fd13fca5328c4838e54de65fdad455d2

We no longer need the hack that generated the `*.runtimeconfig.json` file, because it's now just a regular published ASP.NET Core application.

The only drawback I can think of is that `DevServer` is a 5MB package, whereas `Cli` was just a few tens of KB. But it doesn't affect people's app sizes, since it's not sent to the browser nor even included in publish output. AFAIK this is the only correct and supported way to do what we want, though.


src/Components/Blazor/DevServer/src/Microsoft.AspNetCore.Blazor.DevServer.csproj
```diff
@@ -0,0 +1,50 @@
...
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <OutputType>Exe</OutputType>
+    <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
+    <AssemblyName>blazor-devserver</AssemblyName>
+    <PackageId>Microsoft.AspNetCore.Blazor.DevServer</PackageId>
+    <IsShippingPackage>true</IsShippingPackage>
+    <HasReferenceAssembly>false</HasReferenceAssembly>
+    <StartupObject>Microsoft.AspNetCore.Blazor.DevServer.Program</StartupObject>
```
pranavkm: Why is this required?

pranavkm (approved):

src/Components/Blazor/DevServer/src/Microsoft.AspNetCore.Blazor.DevServer.csproj
SteveSandersonMS: Because there's more than one `static class Program` with a `static void Main`. We could rename things but specifying this here is equally good.

src/Components/Blazor/DevServer/src/Microsoft.AspNetCore.Blazor.DevServer.csproj
```diff
@@ -0,0 +1,50 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <OutputType>Exe</OutputType>
+    <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
+    <AssemblyName>blazor-devserver</AssemblyName>
+    <PackageId>Microsoft.AspNetCore.Blazor.DevServer</PackageId>
+    <IsShippingPackage>true</IsShippingPackage>
+    <HasReferenceAssembly>false</HasReferenceAssembly>
```
dougbu: Why doesn't this project have a ref assembly?

src/Components/Blazor/DevServer/src/Microsoft.AspNetCore.Blazor.DevServer.nuspec
```diff
@@ -1,32 +1,25 @@
...
     <tags>aspnetcore components</tags>
     <serviceable>true</serviceable>
     <repository type="git" url="https://github.com/aspnet/AspNetCore" commit="$repositorycommit$" />
-    <dependencies>
-      <group targetFramework=".NETCoreApp3.0">
-        <dependency id="Microsoft.AspNetCore.Blazor.Server" version="$blazorversion$" exclude="Build,Analyzers" />
-        <dependency id="Microsoft.AspNetCore.Components.Server" version="$aspnetcoreversion$" exclude="Build,Analyzers" />
-        <dependency id="Microsoft.AspNetCore.SpaServices.Extensions" version="$aspnetcoreversion$" exclude="Build,Analyzers" />
-      </group>
-    </dependencies>
```
dougbu: Really -- no dependencies at all?

dougbu (approved):
Minor questions here and in the code&hellip;

I see the CI verifications all succeeded 😃 But, should Components.sln change too?

SteveSandersonMS:
> But, should Components.sln change too?

The .sln did change: https://github.com/aspnet/AspNetCore/pull/9028/files#diff-7c36ff9630f9661629130352ef599b7d

> Really -- no dependencies at all?

Yes, it's a self-contained pre-published command-line tool.

> Why doesn't this project have a ref assembly?

It has no public API surface. Referencing it only affects how your project starts up, not what code is in it while it is running.

Issue #6392, opened 2019-01-04
Experience for a Component class library


rynowak:
We need to define what the experience is for a class library with components. This probably can't be as simple as a empty project file for the reasons outlined below.

We have to address the following concerns:
- component libraries should be netstandard2.0 for blazor (at least for now)
- our inference of Razor targeting (version, configuration, mvc support) is based on netcoreapp version
- we shouldn't put msbuild files in a package to control the configuration because it will cause conflicts with other scenarios

A simple example of how to solve this would be something like:
```
<Project>
  <PropertyGroup>
    <IsComponentLibrary>true</IsComponentLibrary>
  </PropertyGroup>

....
</Project>
```

rynowak:
For preview 2 we plan to update the Blazor.Build package to hide the necessary boilerplate since this template will still be a separate item in preview 2. No update yet on what the official story will be in RTM

SteveSandersonMS:
> For preview 2 we plan to update the Blazor.Build

This is done, but it's in Components.Build (which is in turn referenced by Blazor.Build for Blazor apps, but not for Razor Components apps).

pranavkm:
I'm trying to lookup where the design notes for this was (I know we spoke about this once), but failing to come up with it. As I recall, the plan was

a) to use RazorLangVersion to light up features. This would allow us to know what feature sets to target when we introduce newer features in the future. This is particularly important for `netstandard2.0` (ns2.0) targeting projects since there isn't a version number we could use to infer this. In `netcoreapp` targeting projects, we can use the netcoreapp version to do this.

b) For 3.0, ns2.0 targeting templates would have `<RazorLangVersion>3.0</RazorLangVersion>` in the projects. We would remove the .Components.Build and Blazor.Build packages since we should be able to infer all the details we need from this alone. `netcoreapp3.0` targeting templates would have no package reference that adds build props or targets.

c) MVC specific features would light up based on the presence of a reference to Microsoft.AspNetCore.App. We could go further and light it up only if we think we have to work with Razor views or Razor Pages (https://github.com/aspnet/AspNetCore/issues/8161).

Does this correctly sum up what was discussed?

rynowak:
> c) MVC specific features would light up based on the presence of a reference to Microsoft.AspNetCore.App. We could go further and light it up only if we think we have to work with Razor views or Razor Pages 

[...]

PR #9033, opened 2019-04-03
Update KoreBuild + SDK

pranavkm:


dougbu (changes requested):
global.json should also change to match the underlying SDK

pranavkm:
@dougbu does this look familiar:

```
F:\workspace\_work\1\s\build\repo.targets(138,5): error MSB4018: The "RepoTasks.RemoveSharedFrameworkDependencies" task failed unexpectedly. [F:\workspace\_work\1\s\.dotnet\buildtools\korebuild\3.0.0-build-20190405.1\KoreBuild.proj]
F:\workspace\_work\1\s\build\repo.targets(138,5): error MSB4018: System.InvalidOperationException: The element 'metadata' in namespace 'http://schemas.microsoft.com/packaging/2013/05/nuspec.xsd' has invalid child element 'frameworkReferences' in namespace 'http://schemas.microsoft.com/packaging/2013/05/nuspec.xsd'. List of possible elements expected: 'packageTypes, releaseNotes, frameworkAssemblies, title, language, developmentDependency, contentFiles, references, summary, license' in namespace 'http://schemas.microsoft.com/packaging/2013/05/nuspec.xsd'. This validation error occurred in a 'frameworkReferences' element. [F:\workspace\_work\1\s\.dotnet\buildtools\korebuild\3.0.0-build-20190405.1\KoreBuild.proj]
F:\workspace\_work\1\s\build\repo.targets(138,5): error MSB4018:    at NuGet.Packaging.Manifest.<>c.<ValidateManifestSchema>b__20_0(Object sender, ValidationEventArgs e) [F:\workspace\_work\1\s\.dotnet\buildtools\korebuild\3.0.0-build-20190405.1\KoreBuild.proj]
F:\workspace\_work\1\s\build\repo.targets(138,5): error MSB4018:    at System.Xml.Schema.XNodeValidator.ValidationCallback(Object sender, ValidationEventArgs e) [F:\workspace\_work\1\s\.dotnet\buildtools\korebuild\3.0.0-build-20190405.1\KoreBuild.proj]
```

dougbu:
@pranavkm I don't recall seeing that error before. Is it already described in an AspNetCore or AspNetCore-Internal issue?

@JunTaoLuo do you have an inkling what's going wrong w/ that SharedFx task?

JunTaoLuo:
Hmm not sure. Though I'm wondering if it's related to https://github.com/NuGet/Home/issues/7342? It's a new feature that just got merged so maybe we need a new version of NuGet?

dougbu:
Hmm, maybe but only if NuGet made a breaking change while adding the new NuGet/Home#7342 feature.

pranavkm:
cc @natemcmaster 

Bumping up NuGet.Build.Tasks to 5.0.0 didn't help. `frameworkReferences` is a fairly new feature, do we need to reference a preview build of NuGet in repo-tasks?

JunTaoLuo:
cc @rrelyea @nkolev92? Can you comment on the NuGet requirements for `frameworkReferences` support?

natemcmaster:
Which package is being processed when the error about  `frameworkReferences` is thrown? We haven't actually started using `<FrameworkReference>` to build our own packages (https://github.com/aspnet/AspNetCore/issues/4257), so I'm surprised this has appeared in generated nuspec.

nkolev92:
As far as I know, the FrameworkReference pack support is not yet merged anywhere. 
It'll be in the 5.1.0 NuGet release, but it will still considered a `preview` feature. 

I think understanding @natemcmaster's point is key.

pranavkm:
@JunTaoLuo would you be able to drive investigating the source of the problem? 

FYI @Eilon since this is blocking some template work (#6392) that requires a fairly new WebSdk.

JunTaoLuo:
Hmm looks like Microsoft.AspNetCore.ApiAuthorization.IdentityServer failed the task. Looking into why.

JunTaoLuo:
The packages we build has the following in the nuspec:
```
    <frameworkReferences>
      <group targetFramework=".NETCoreApp3.0">
        <frameworkReference name="Microsoft.NETCore.App" />
      </group>
    </frameworkReferences>
```
This is probably what's confusing the `Nuget.Packaging.Manifest.ReadFrom` API. Seems like it's related to the SDK update.

natemcmaster:
> As far as I know, the FrameworkReference pack support is not yet merged anywhere.

It must be if the nuspec generated by the SDK has a framework reference in it. 

@pranavkm  try updating to NuGet.Build.Tasks `5.1.0-rtm.5921`. May require adding a new restore source to https://dotnet.myget.org/F/nuget-build/api/v3/index.json (cref https://github.com/dotnet/toolset/blob/release/3.0.1xx/eng/Versions.props#L55)

nkolev92:
5.1.0-rtm-5921 is the first insertion into the CLI/SDK that has the FrameworkReference work as far as NuGet is concerned. 
And the PR to merge that is on hold https://github.com/dotnet/cli/pull/11054

Which version of the SDK is the one that's creating that nuspec? 

I'm not sure **how** the nuget bits got ingested to cause breakages. 
Would this have anything to do with that? https://github.com/dotnet/toolset/pull/593

global.json
```diff
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "3.0.100-preview4-010309"
+    "version": "3.0.100-preview4-011136"
```
natemcmaster: @nkolev92 this is the version of the SDK we're trying to update to.

natemcmaster: image
https://github.com/dotnet/toolset/pull/593#issuecomment-479250753

@nguerrera We're seeing this feature show up in P4 builds of the SDK...did something changed in the last 6 days since this comment, or did this unintentionally get merged into P4?

nguerrera: How does the feature being present break things?

I intentionally did not revert it because I figured we should get coverage of latest bits. There was no expectation of anything breaking. I chatted briefly with @nkolev92 after that comment.

If necessary, we can revert the nuget version in P4.

JunTaoLuo: Note that @pranavkm updated to the latest SDK available as of today.

nguerrera:
> And the PR to merge that is on hold dotnet/cli#11054

That's for 2.1.7xx

For 3.0, it was dotnet/toolset#593

global.json
nguerrera: Understood. The question is why is taking a nuget with the frameworkReferences feature a breaking change?

natemcmaster: It breaks the aspnet build because we have some custom MSBuild tasks reading the nuspec, and NuGet code throws when unrecognized elements appear in the metadata. 

FWIW, though, the reason we have these custom MSBuild tasks was to workaround the absence of this NuGet feature...so having access to this SDK unblocks us to start using `<FrameworkReference>` the *right* way.

@pranavkm was there any particular reason you were trying to update to the latest SDK? If not, maybe we should combine this update with a solution for #4257

nkolev92:
@nguerrera I think regardless of whether it gets reverted or not, the SDK should set "pack=false" to Microsoft.NETCore.App.

nguerrera:
> @nguerrera I think regardless of whether it gets reverted or not, the SDK should set "pack=false" to Microsoft.NETCore.App.

@dsplaisted

global.json
JunTaoLuo: Right, essentially there's a mismatch in the toolset which uses nuget 5.1.0-rtm.5921 to build our packages and the NuGet.Build.Tasks that we use to post process the packages built. To reduce the work that's required to update the SDK which to my understanding is blocking us, why don't we just update the version of NuGet.Build.Tasks that we use to 5.1.0-rtm.5921 as well?

JunTaoLuo: Should be available at https://dotnet.myget.org/feed/nuget-build/package/nuget/NuGet.Build.Tasks/5.1.0-rtm.5921+27af96bdb7ba8d6d7ea9ad53fc76cd1d1aa80703.

pranavkm: > @pranavkm was there any particular reason you were trying to update to the latest SDK? If not, maybe we should combine this update with a solution for #4257

There are changes in the WebSDK that are require to the test the preview4 templates.

JunTaoLuo:
@pranavkm Looks like it's fixed. Feel free to proceed with QB process.

JunTaoLuo (approved):

dougbu (approved):

natemcmaster:
I'm a little concerned about accepting this change so late in the branching process. It sounds like there was a lot for churn in the SDK w.r.t. framework references. Before we accept this, can you try using some of the netcoreapp3.0 packages produced by this build to make sure they still work as expected?

pranavkm:
@natemcmaster I agree, it does seem a bit risky at this point. @danroth27 is considering not taking the template change at this point, so we might be happy taking this update in master instead.

Eilon:
OK if there's too much risk and we're not desperate for this change, then let's just do 'master' and let the bits settle.

pranavkm:
Updated to point this to master.

PR #9037, opened 2019-04-03
Use coventional routes for link generation

rynowak:
This change enables using conventional routes for link generation when
using MVC conventional routes. This change makes MVC link generation
behaviour highly compatible with 2.1.

The way that this works is that we create endpoints for **MATCHING**
using the denormalized conventional route, but we tell those endpoints
to suppress link generation.

For link generation we generate a non-matching endpoints per-route with
the same order value.

I added the concept of *required value any* to link generation. This is
needed because for an endpoint to participate in link generation using
RouteValuesAddress it needs to have some required values. These details
are a little fiddly, but I think it's worth doing this feature
completely.


rynowak:
@JamesNK

rynowak:
@aspnet/hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=142090&view=ms.vss-test-web.build-test-results-tab

aspnet-hello:

I've triaged the above build.

src/Mvc/test/Mvc.FunctionalTests/compiler/resources/HtmlGenerationWebSite.HtmlGeneration_Home.Index.Encoded.html
```diff
@@ -45,26 +45,26 @@
         <a href="/Order/List">Href Order List</a>
     </div>
     <div>
-        <a href="">Non-existent Controller</a>
+        <a href="HtmlEncode[[/NonExistentController]]">Non-existent Controller</a>
```
JamesNK: 🎉

src/Http/Routing/test/UnitTests/LinkGeneratorIntegrationTest.cs
```diff
@@ -0,0 +1,610 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Features;
+using Microsoft.AspNetCore.Routing.Patterns;
+using System.Collections.Generic;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Routing
```
JamesNK: Add tests for using the route name + conventional route? https://github.com/aspnet/AspNetCore/issues/4468

JamesNK: Similar https://github.com/aspnet/AspNetCore/issues/4556#issuecomment-448496662

JamesNK:
Also, there are a lot of routing tests where we have a base class to share them between old routing and endpoint routing. Some tests are "disabled" in endpoint routing. We should audit those and re-enable the tests that include newly supported scenarios.

rynowak:
Ah, that's a good point. I didn't realise we had **disabled tests**. I thought we just updated them to have different expectations.

src/Http/Routing/src/Internal/LinkGenerationDecisionTree.cs
```diff
@@ -21,38 +22,48 @@ public class LinkGenerationDecisionTree
...
+            var attributedEntries = new List<OutboundMatch>();
+            _conventionalEntries = new List<OutboundMatch>();
             for (var i = 0; i < entries.Count; i++)
             {
+                var isAttributeRoute = true;
                 var entry = entries[i];
                 foreach (var kvp in entry.Entry.RequiredLinkValues)
                 {
-                    if (!_knownValues.TryGetValue(kvp.Key, out var values))
+                    if (object.ReferenceEquals(RoutePattern.RequiredValueAny, kvp.Value))
```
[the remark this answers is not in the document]
rynowak: sure.

src/Http/Routing/src/Patterns/DefaultRoutePatternTransformer.cs
```diff
@@ -76,9 +77,29 @@ private RoutePattern SubstituteRequiredValuesCore(RoutePattern original, RouteVa
                     // Ex: {controller=Home}/{action=Index}/{id?} - with required values: { area = "", ... }
                     continue;
                 }
+                else if (object.ReferenceEquals(kvp.Value, RoutePattern.RequiredValueAny))
+                {
+                    // Required value is *any* - this is allowed for a parameter with a default, but not
```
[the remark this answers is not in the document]
rynowak: Sure.

src/Http/Routing/src/Template/TemplateBinder.cs
```diff
@@ -178,7 +178,7 @@ public TemplateValuesResult GetValues(RouteValueDictionary ambientValues, RouteV
                             throw new InvalidOperationException($"Unable to find required value '{key}' on route pattern.");
                         }
 
-                        if (!RoutePartsEqual(ambientValue, _pattern.RequiredValues[key]))
+                        if (!RoutePartsEqual(ambientValue, _pattern.RequiredValues[key]) && !object.ReferenceEquals(RoutePattern.RequiredValueAny, _pattern.RequiredValues[key]))
```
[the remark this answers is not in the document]
rynowak: No, RVA cannot be used as a route value -it will turn into empty string and vanish. I added a comment to RVA's docs.

src/Http/Routing/test/UnitTests/LinkGeneratorIntegrationTest.cs
rynowak: I'll add a test for a conventional route that doesn't match any endpoints.

rynowak:
> Also, there are a lot of routing tests where we have a base class to share them between old routing and endpoint routing. Some tests are "disabled" in endpoint routing. We should audit those and re-enable the tests that include newly supported scenarios.

I found two tests that were relevant to *share*. The tests that are unique to legacy routing are the tests for 404 on HTTP methods, and `IRouter`

rynowak:
@JamesNK updated

src/Http/Routing/src/Patterns/RoutePattern.cs
```diff
@@ -146,5 +151,11 @@ internal string DebuggerToString()
         {
             return RawText ?? string.Join(SeparatorString, PathSegments.Select(s => s.DebuggerToString()));
         }
+
+        [DebuggerDisplay("{DebuggerToString(),nq}")]
```
JamesNK: Fancy. I like it

src/Http/Routing/test/UnitTests/LinkGeneratorIntegrationTest.cs
```diff
@@ -121,10 +121,20 @@ public LinkGeneratorIntegrationTest()
...
                     order: 1000,
                     metadata: new object[] { new SuppressMatchingMetadata(), }),
 
+                // Conventional routed link generation route 3 - this doesn't match any actions.
+                EndpointFactory.CreateRouteEndpoint(
+                    RoutePatternFactory.Parse(
+                        "api/{controller}/{id?}",
+                        defaults: new { },
+                        parameterPolicies: new { },
+                        requiredValues: new { controller = RoutePattern.RequiredValueAny, action = (string)null, area = (string)null, page = (string)null, }),
```
JamesNK: What will happen when someone adds a route like this: `routes.MapRoute("foo", "/foo/foo");`

Are there are no required values, and that means no RequiredValueAny?

rynowak: It would have required value null for everything that MVC thinks is special.

JamesNK: Maybe add an integration test for it? 🙏 

(or create an issue for it if you don't want to do it now)













(or tell me to piss off 😄 )

rynowak:
@JamesNK added a test.

src/Http/Routing/test/UnitTests/LinkGeneratorIntegrationTest.cs
rynowak: Added a test

JamesNK (approved):
:shipit:

rynowak:
@aspnet/aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=143528

aspnet-hello:

I've triaged the above build.

PR #9044, opened 2019-04-03
Added Crankier docs

staff0rd:
Docs describing usage of Crankier

### DIFF
--- a/src/SignalR/perf/benchmarkapps/Crankier/Readme.md
+++ b/src/SignalR/perf/benchmarkapps/Crankier/Readme.md
@@ -36,5 +36,5 @@ dotnet run -- local --target-url https://localhost:5001/echo --workers 10
 Attempt to make 5,000 connections to the `echo` hub using Long Polling
 
 ```
-dotnet run local --target-url https://localhost:5001/echo --connections 1000 --transport LongPolling
+dotnet run -- local --target-url https://localhost:5001/echo --connections 1000 --transport LongPolling
 ```
--- a/src/SignalR/perf/benchmarkapps/Crankier/Readme.md
+++ b/src/SignalR/perf/benchmarkapps/Crankier/Readme.md
@@ -1,0 +1,40 @@
+# Crankier
+
+Load testing for ASP.NET Core SignalR
+
+## Commands
+
+### local
+
+```
+> dotnet run help local
+
+Usage:  local [options]
+
+Options:
+  --target-url <TARGET_URL>                   The URL to run the test against.
+  --workers <WORKER_COUNT>                    The number of workers to use.
+  --connections <CONNECTION_COUNT>            The number of connections per worker to use.
+  --send-duration <SEND_DURATION_IN_SECONDS>  The send duration to use.
+  --transport <TRANSPORT>                     The transport to use (defaults to WebSockets).
+  --worker-debug                              Provide this switch to have the worker wait for the debugger.
+```
+
+Notes:
+
+* `TARGET_URL` needs to be the route to your hub exposed by `UseSignalr` in your application.
+* `CONNECTION_COUNT` defaults to 10,000.
+
+#### Examples
+
+Attempt to make 10,000 connections to the `echo` hub using WebSockets and 10 workers:
+
+```
+dotnet run local --target-url https://localhost:5001/echo --workers 10
+```
+
+Attempt to make 5,000 connections to the `echo` hub using Long Polling
+
+```
+dotnet run local --target-url https://localhost:5001/echo --connections 1000 --transport LongPolling
+```
--- a/src/SignalR/perf/benchmarkapps/Crankier/Readme.md
+++ b/src/SignalR/perf/benchmarkapps/Crankier/Readme.md
@@ -6,6 +6,8 @@ Load testing for ASP.NET Core SignalR
 
 ### local
 
+The `local` command launches a set of local worker clients to establish connections to your SignalR server.
+
 ```
 > dotnet run -- help local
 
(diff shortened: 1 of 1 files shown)


src/SignalR/perf/benchmarkapps/Crankier/Readme.md
```diff
@@ -0,0 +1,40 @@
+# Crankier
+
+Load testing for ASP.NET Core SignalR
+
+## Commands
+
+### local
+
+```
+> dotnet run help local
```
anurse:
```suggestion
> dotnet run -- help local
```

While your example works fine, it's probably best to encourage using `--` here. The `--` option tells `dotnet run` that everything after it is for the app being run and not for the `dotnet run` command itself. This is only really relevant when a command/option in the app conflicts with an option in `dotnet run` but it's good to encourage this separation just in case something changes :)

src/SignalR/perf/benchmarkapps/Crankier/Readme.md
```diff
@@ -0,0 +1,40 @@
...
+
+* `TARGET_URL` needs to be the route to your hub exposed by `UseSignalr` in your application.
+* `CONNECTION_COUNT` defaults to 10,000.
+
+#### Examples
+
+Attempt to make 10,000 connections to the `echo` hub using WebSockets and 10 workers:
+
+```
+dotnet run local --target-url https://localhost:5001/echo --workers 10
```
anurse:
```suggestion
dotnet run -- local --target-url https://localhost:5001/echo --workers 10
```

src/SignalR/perf/benchmarkapps/Crankier/Readme.md
```diff
@@ -0,0 +1,40 @@
...
+```
+
+Attempt to make 5,000 connections to the `echo` hub using Long Polling
+
+```
+dotnet run local --target-url https://localhost:5001/echo --connections 1000 --transport LongPolling
```
anurse:
```suggestion
dotnet run -- local --target-url https://localhost:5001/echo --connections 1000 --transport LongPolling
```

src/SignalR/perf/benchmarkapps/Crankier/Readme.md
```diff
@@ -0,0 +1,40 @@
...
+  --connections <CONNECTION_COUNT>            The number of connections per worker to use.
+  --send-duration <SEND_DURATION_IN_SECONDS>  The send duration to use.
+  --transport <TRANSPORT>                     The transport to use (defaults to WebSockets).
+  --worker-debug                              Provide this switch to have the worker wait for the debugger.
+```
+
+Notes:
+
+* `TARGET_URL` needs to be the route to your hub exposed by `UseSignalr` in your application.
+* `CONNECTION_COUNT` defaults to 10,000.
```
anurse: Might as well add the defaults for `SEND_DURATION_IN_SECONDS` (which is `300`) and `WORKER_COUNT` (which is `1`)

src/SignalR/perf/benchmarkapps/Crankier/Readme.md
```diff
@@ -0,0 +1,40 @@
+# Crankier
+
+Load testing for ASP.NET Core SignalR
+
+## Commands
+
+### local
```
anurse: I'd suggest adding a small overview of the `local` command here. Something like "The `local` command launches a set of local worker clients to establish connections to your SignalR server and send messages for the duration of the test."

We should also call out that the client expects the server hub to have a method like this:

```csharp
public void SendPayload(string payload)
{
	// ... your code ...
}
```

Maybe point to [Client.StartTest](https://github.com/aspnet/AspNetCore/blob/master/src/SignalR/perf/benchmarkapps/Crankier/Client.cs#L73) which defines the test behavior to run.

src/SignalR/perf/benchmarkapps/Crankier/Readme.md
```diff
@@ -0,0 +1,40 @@
...
+
+#### Examples
+
+Attempt to make 10,000 connections to the `echo` hub using WebSockets and 10 workers:
+
+```
+dotnet run local --target-url https://localhost:5001/echo --workers 10
+```
+
+Attempt to make 5,000 connections to the `echo` hub using Long Polling
```
anurse: The command below says `1000` though, right? I may have forgotten since the app was first written but the default worker count is `1` so it should only create `1000` connections.

analogrelay (commented):
Very nice! A couple comments but I don't see any problem taking this change in after some refinement :)

src/SignalR/perf/benchmarkapps/Crankier/Readme.md
staff0rd: Client.StartTest has zero references.  Also I ran the `local` command against my own hub that did not have a `SendPayload` and it worked as expected.  I'll test this further this evening, but I suspect `StartTest` is not actually being called or is being called with `0` message size.

anurse: Ah, you're probably right. This is a bunch of code for the previous iteration of SignalR that was brought forward and hacked up to work so there are definitely some quirks in here :)

src/SignalR/perf/benchmarkapps/Crankier/Readme.md
staff0rd: done

src/SignalR/perf/benchmarkapps/Crankier/Readme.md
staff0rd: I've added the overview, but also confirmed that `SendPayload` is currently not called.

src/SignalR/perf/benchmarkapps/Crankier/Readme.md
staff0rd: my bad, updated.

staff0rd:
@anurse all changes implemented

analogrelay (approved):
Nice, thanks!

Issue #5311, opened 2018-06-27
Test concurrent connection limit on different Azure VM sizes


BrennanConroy:
Collect connection count, CPU %, Memory Usage
Machine Sizes:
- [ ] B1s (4,900 connections, 1GB RAM limiting factor, CPU ~10%) Specs: 1GB RAM, 1 CPU
- [ ] B2s (82,000 connections, CPU and RAM basically 100%) Specs: 4 GB RAM, 2 CPU
- [ ] D2s_v3 (110,000 connections, CPU 100%, RAM 80%) Specs: 8 GB RAM, 2 CPU
- [ ] D4s_v3 (160,000 connections, CPU 100%, RAM 50%, connections started disconnecting) Specs: 16 GB RAM, 4 CPU
- [ ] D8s_v3 Specs: 32 GB RAM, 8 CPU
- [ ] D16s_v3 Specs: 64 GB RAM, 16 CPU
- [ ] D32s_v3 Specs: 128 GB RAM, 32 CPU
- [ ] D64s_v3 Specs: 256 GB RAM, 64 CPU

Note: Test Linux vs. Windows VMs to see if there is a difference

anurse:
Closing this since performance is a constant concern and workstream for us, but the specifics of this issue aren't really relevant any more. The concern is that this data would be based on our specific conditions and would get stale really fast.

PR #9045, opened 2019-04-03
Crankier: Log connection exceptions

staff0rd:
Sends exception messages back to the worker for agent logging so you can see why faulted connections are failing.

### DIFF
--- a/src/SignalR/perf/benchmarkapps/Crankier/Client.cs
+++ b/src/SignalR/perf/benchmarkapps/Crankier/Client.cs
@@ -17,6 +17,7 @@ namespace Microsoft.AspNetCore.SignalR.Crankier
         private bool _sendInProgress;
         private volatile ConnectionState _connectionState = ConnectionState.Connecting;
 
+        public string ErrorMessage { get; private set;}
         public ConnectionState State => _connectionState;
         public async Task CreateAndStartConnectionAsync(string url, HttpTransportType transportType)
         {
@@ -33,7 +34,8 @@ namespace Microsoft.AspNetCore.SignalR.Crankier
                 }
                 else
                 {
-                    Trace.WriteLine($"Connection terminated with error: {ex.GetType()}: {ex.Message}");
+                    ErrorMessage = $"Connection terminated with error: {ex.GetType()}: {ex.Message}";
+                    Trace.WriteLine(ErrorMessage);
                     _connectionState = ConnectionState.Faulted;
                 }
 
@@ -57,7 +59,8 @@ namespace Microsoft.AspNetCore.SignalR.Crankier
                 }
                 catch (Exception ex)
                 {
-                    Trace.WriteLine($"Connection.Start Failed: {ex.GetType()}: {ex.Message}");
+                    ErrorMessage = $"Connection.Start Failed: {ex.GetType()}: {ex.Message}";
+                    Trace.WriteLine(ErrorMessage);
 
                     if (connectCount == 3)
                     {
--- a/src/SignalR/perf/benchmarkapps/Crankier/Client.cs
+++ b/src/SignalR/perf/benchmarkapps/Crankier/Client.cs
@@ -12,13 +12,27 @@ namespace Microsoft.AspNetCore.SignalR.Crankier
 {
     public class Client
     {
+        private readonly int _processId;
+        private readonly IAgent _agent;
         private HubConnection _connection;
         private CancellationTokenSource _sendCts;
         private bool _sendInProgress;
         private volatile ConnectionState _connectionState = ConnectionState.Connecting;
 
-        public string ErrorMessage { get; private set;}
         public ConnectionState State => _connectionState;
+        public Client(int processId, IAgent agent)
+        {
+            _processId = processId;
+            _agent = agent;
+        }
+
+        private void LogFault(string description, Exception exception)
+        {
+            var message = $"{description}: {exception.GetType()}: {exception.Message}";
+            Trace.WriteLine(message);
+            _agent.LogAsync(_processId, message);
+        }
+
         public async Task CreateAndStartConnectionAsync(string url, HttpTransportType transportType)
         {
             _connection = new HubConnectionBuilder()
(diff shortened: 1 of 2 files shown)


src/SignalR/perf/benchmarkapps/Crankier/Client.cs
```diff
@@ -33,7 +34,8 @@ public async Task CreateAndStartConnectionAsync(string url, HttpTransportType tr
                 }
                 else
                 {
-                    Trace.WriteLine($"Connection terminated with error: {ex.GetType()}: {ex.Message}");
+                    ErrorMessage = $"Connection terminated with error: {ex.GetType()}: {ex.Message}";
+                    Trace.WriteLine(ErrorMessage);
```
halter73: What do you think about adding a constructor to Client that takes an IAgent and process id so the clients can log directly?

halter73:
@staff0rd Out of curiosity, what are you using Crankier for? Has it met your needs?

src/SignalR/perf/benchmarkapps/Crankier/Client.cs
staff0rd: Happy to do this, will make the change this evening.

staff0rd:
@halter73 attempting to get an understanding of what limits Azure has on SignalR.  Found some weird stuff like Websockets being limited to 768 for P1V1 & S1 app service plans, have raised tickets with support.  I get 8000+ locally.

Confirmed 350 websocket limit on B1 per the [limits page](https://docs.microsoft.com/en-us/azure/azure-subscription-service-limits#app-service-limits).

I also want to test Azure SignalR Service as the [pricing](https://azure.microsoft.com/en-us/pricing/details/signalr-service/) is not clear to me.  Notably whether connections count as messages (docs state they don't) and what Price/Unit/Day actually means: does it mean max number of units raised in a given 24 period, or, if I raise one unit for 12 hours and bring it back down, do I pay half the price.  Depending on answers to those I'd be interested in verifying @anurse's [tweet](https://twitter.com/anurse/status/983406560880701440?ref_src=twsrc%5Etfw%7Ctwcamp%5Etweetembed%7Ctwterm%5E983406560880701440&ref_url=http%3A%2F%2F0.0.0.0%3A4000%2Fblog%2F2019%2F03%2F20%2Fazure-signalr-service%2F) - providing I can raise the resources on my VS subscription credit.

anurse:
@staff0rd FYI in our experience, Azure App Service WebSocket support can be very limited. Azure SignalR Service is *strongly recommended* when using Azure App Service, since it handles the SignalR connections for you.

anurse:
> Depending on answers to those I'd be interested in verifying @anurse's [tweet](https://twitter.com/anurse/status/983406560880701440?ref_src=twsrc%5Etfw%7Ctwcamp%5Etweetembed%7Ctwterm%5E983406560880701440&ref_url=http%3A%2F%2F0.0.0.0%3A4000%2Fblog%2F2019%2F03%2F20%2Fazure-signalr-service%2F)

FYI that tweet was using a VM, not Azure App Service ;)

staff0rd:
@anurse what spec was the VM?

staff0rd:
@halter73 changes made.

I dunno if conversation responses or resolving conversations trigger notifications?

anurse:
> what spec was the VM?

I honestly don't remember. That tweet was from a very early exploratory test. #5311 has some data that we've gathered based on Azure VM specs.

staff0rd:
@anurse any guidance on how those numbers were achieved? #5311 notes 82,000 connections on B2s - but how are these connections being made?  I ran the benchmark server on a B2s, then ran Crankier but it stops at around 8000 with

> System.Net.Http.HttpRequestException: An operation on a socket could not be performed because the system lacked sufficient buffer space or because a queue was full

BrennanConroy:
You need many client machines.

staff0rd:
Also;

> System.Net.Http.HttpRequestException: Only one usage of each socket address (protocol/network address/port) is normally permitted

staff0rd:
How did you bring the client machines up?  Any automation here?

BrennanConroy:
Manually, no automation, that's why we didn't go too far with it.

staff0rd:
Ok, I [stuck it in a container](https://github.com/staff0rd/docker-crankier), raised 12 of them, and got 80172 connections on a B2s before something blew up and it lost 20k connections.

src/SignalR/perf/benchmarkapps/Crankier/Worker.cs
```diff
@@ -140,9 +140,10 @@ private async Task SendStatusUpdateAsync(CancellationToken cancellationToken)
                         case ConnectionState.Reconnecting:
                             reconnectingCount++;
                             break;
-                        case ConnectionState.Faulted:
+                        case ConnectionState.Faulted: {
                             faultedCount++;
                             break;
+                        }
```
halter73: nit: these brackets are kinda weird

halter73:
> I dunno if conversation responses or resolving conversations trigger notifications?

In my experience, conversation responses trigger notifications, but resolving conversations doesn't.

halter73:
@staff0rd Do you have any details about what blew up in you container-based testing?

src/SignalR/perf/benchmarkapps/Crankier/Worker.cs
staff0rd: agree, removed.

halter73 (approved):

staff0rd:
@halter73 I do not have any info on what happened, and since decommissioned the resources.  Basically it was over 80k and suddenly about 20k connections.  The containers were still running at this time, and further connections were being made afterwards so it slowly started climbing again - however I switched it off as I didn't want to wait for it to get back up to 80k - I think overall it took about 20min to get that high with 12 containers making connections.

Oddly, some of the earlier containers I raised never even made 1,000 connections, when later ones were peaking at over 10,000 connections.  That said, I've got the suspicion that the 20k drop was due to one or two containers failing, but it's difficult for me to watch logs across 12 containers.

I still want to push the connection limit (just for fun with larger VMs), so I may make some changes to the benchmark server to help diagnose what's going on during multi-client tests.

I've also proven that the 768 limit I was getting on S1 App Services is _per client_.  While I can't find any docs to confirm that number, it removes any concern of App Service not being feasible for me in the short to medium term.

halter73:
Thanks!

PR #9046, opened 2019-04-03
[Helix] Try running node services tests on helix again

HaoK:
Node 10.15.3 should be installed on all queues tomorrow, see if anything is blocking this test before then

### DIFF
--- a/eng/helix/content/InstallNode.ps1
+++ b/eng/helix/content/InstallNode.ps1
@@ -9,6 +9,12 @@ $ProgressPreference = 'SilentlyContinue' # Workaround PowerShell/PowerShell#2138
 
 Set-StrictMode -Version 1
 
+if (Test-Path "%HELIX_CORRELATION_PAYLOAD%\node\bin\node.exe")
+{
+    Write-Host "Node.exe found at %HELIX_CORRELATION_PAYLOAD%\node\bin"
+    exit
+}
+
 $nodeFile="node-v$Version-win-x64"
 $url="http://nodejs.org/dist/v$Version/$nodeFile.zip"
 Write-Host "Starting download of NodeJs ${Version} from $url"
@@ -17,4 +23,5 @@ Write-Host "Done downloading NodeJS ${Version}"
 mkdir node -Force
 Expand-Archive "nodejs.zip" -destination node
 Write-Host "Expanded NodeJs"
-copy node/$nodeFile/node.exe .
+mkdir %HELIX_CORRELATION_PAYLOAD%\node\bin -Force
+copy node/$nodeFile/node.exe %HELIX_CORRELATION_PAYLOAD%\node\bin
--- a/eng/helix/content/InstallNode.ps1
+++ b/eng/helix/content/InstallNode.ps1
@@ -18,7 +18,7 @@ if (Test-Path "%HELIX_CORRELATION_PAYLOAD%\node\bin\node.exe")
 $nodeFile="node-v$Version-win-x64"
 $url="http://nodejs.org/dist/v$Version/$nodeFile.zip"
 Write-Host "Starting download of NodeJs ${Version} from $url"
-Invoke-WebRequest -UseBasicParsing -Uri "$url" -Out "nodejs.zip"
+Invoke-WebRequest -UseBasicParsing -Uri "$url" -OutFile "nodejs.zip"
 Write-Host "Done downloading NodeJS ${Version}"
 mkdir node -Force
 Expand-Archive "nodejs.zip" -destination node
--- a/eng/helix/content/InstallNode.ps1
+++ b/eng/helix/content/InstallNode.ps1
@@ -33,13 +33,14 @@ if ((Get-Command "Expand-Archive" -ErrorAction SilentlyContinue))
 }
 else
 {
+    Write-Host "Expand-Archive not found"
     $shell = new-object -com shell.application
-    $zip = $shell.NameSpace("nodejs.zip")
+    $zip = $shell.NameSpace((Get-Location).Path + "\nodejs.zip")
     foreach($item in $zip.items())
     {
         $shell.Namespace($output_dir).copyhere($item)
+        Start-Sleep -milliseconds 500
     }
-    Start-Sleep -s 5
 }
 
 Write-Host "Expanded NodeJs"
--- a/eng/helix/content/InstallNode.ps1
+++ b/eng/helix/content/InstallNode.ps1
@@ -9,6 +9,12 @@ $ProgressPreference = 'SilentlyContinue' # Workaround PowerShell/PowerShell#2138
 
 Set-StrictMode -Version 1
 
+if ((Get-Command "node.exe" -ErrorAction SilentlyContinue)) 
+{ 
+    Write-Host "Found node.exe in PATH"
+    exit
+}
+
 if (Test-Path "%HELIX_CORRELATION_PAYLOAD%\node\bin\node.exe")
 {
     Write-Host "Node.exe found at %HELIX_CORRELATION_PAYLOAD%\node\bin"
@@ -20,8 +26,7 @@ $url="http://nodejs.org/dist/v$Version/$nodeFile.zip"
 Write-Host "Starting download of NodeJs ${Version} from $url"
 Invoke-WebRequest -UseBasicParsing -Uri "$url" -OutFile "nodejs.zip"
 Write-Host "Done downloading NodeJS ${Version}"
-mkdir node -Force
-Expand-Archive "nodejs.zip" -destination node
+Expand-Archive "nodejs.zip"
 Write-Host "Expanded NodeJs"
 mkdir %HELIX_CORRELATION_PAYLOAD%\node\bin -Force
-copy node/$nodeFile/node.exe %HELIX_CORRELATION_PAYLOAD%\node\bin
+copy nodejs/$nodeFile/node.exe %HELIX_CORRELATION_PAYLOAD%\node\bin
--- a/eng/helix/content/InstallNode.ps1
+++ b/eng/helix/content/InstallNode.ps1
@@ -26,7 +26,22 @@ $url="http://nodejs.org/dist/v$Version/$nodeFile.zip"
 Write-Host "Starting download of NodeJs ${Version} from $url"
 Invoke-WebRequest -UseBasicParsing -Uri "$url" -OutFile "nodejs.zip"
 Write-Host "Done downloading NodeJS ${Version}"
-Expand-Archive "nodejs.zip"
+
+if ((Get-Command "Expand-Archive" -ErrorAction SilentlyContinue)) 
+{ 
+    Expand-Archive "nodejs.zip"
+}
+else
+{
+    $shell = new-object -com shell.application
+    $zip = $shell.NameSpace("nodejs.zip")
+    foreach($item in $zip.items())
+    {
+        $shell.Namespace($output_dir).copyhere($item)
+    }
+    Start-Sleep -s 5
+}
+
 Write-Host "Expanded NodeJs"
 mkdir %HELIX_CORRELATION_PAYLOAD%\node\bin -Force
 copy nodejs/$nodeFile/node.exe %HELIX_CORRELATION_PAYLOAD%\node\bin
--- a/eng/helix/content/InstallNode.ps1
+++ b/eng/helix/content/InstallNode.ps1
@@ -30,8 +30,8 @@ Write-Host "Starting download of NodeJs ${Version} from $url"
 Invoke-WebRequest -UseBasicParsing -Uri "$url" -OutFile "nodejs.zip"
 Write-Host "Done downloading NodeJS ${Version}"
 
-mkdir nodejs -Force
-$temp_dir = Join-Path (Get-Location) nodejs
+$temp_dir = Join-Path ([System.IO.Path]::GetTempPath()) [System.Guid]::NewGuid()
+mkdir $temp_dir -Force
 Write-Host "Extracting to $temp_dir"
 
 if (Get-Command -Name 'Microsoft.PowerShell.Archive\Expand-Archive' -ErrorAction Ignore) {
--- a/eng/helix/content/InstallNode.ps1
+++ b/eng/helix/content/InstallNode.ps1
@@ -1,7 +1,10 @@
 
 param(
     [Parameter(Mandatory = $true)]
-    $Version
+    $Version,
+    
+    [Parameter(Mandatory = $true)]
+    $output_dir
 )
 
 $ErrorActionPreference = 'Stop'
@@ -15,9 +18,9 @@ if ((Get-Command "node.exe" -ErrorAction SilentlyContinue))
     exit
 }
 
-if (Test-Path "%HELIX_CORRELATION_PAYLOAD%\node\bin\node.exe")
+if (Test-Path "$output_dir\node.exe")
 {
-    Write-Host "Node.exe found at %HELIX_CORRELATION_PAYLOAD%\node\bin"
+    Write-Host "Node.exe found at $output_dir"
     exit
 }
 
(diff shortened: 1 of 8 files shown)


HaoK:
Fixes https://github.com/aspnet/AspNetCore/issues/8044

HaoK:
node rollout to helix queues hasn't happened yet blocked until that occurs

HaoK:
Almost there 

@MattGal looks like Node was installed correctly on OSX/Windows and these tests are now working on those queues, but the various linux queues are failing due to nodejs not being found in the path still:

(Fedora/Redhat/Debian/Ubunutu/Centos) https://mc.dot.net/#/user/aspnetcore/pr~2Faspnet~2Faspnetcore/ci/20190422.46/workItem/Microsoft.AspNetCore.NodeServices.Tests-netcoreapp3.0/wilogs

HaoK:
@natemcmaster before I go down any crazy rabbit holes, can you take a look at how I'm attempting to install nodejs here in case there's something (much) simpler?  Looks like extract-archive isn't available on Win7/8 so I am attempting to use the built in shell support (from stack overflow shrug)

OSX14 and Win7/8 are the last 3 queues which aren't correctly picking up nodejs now.

eng/helix/content/InstallNode.ps1
```diff
@@ -0,0 +1,48 @@
...
+Write-Host "Done downloading NodeJS ${Version}"
+
+if ((Get-Command "Expand-Archive" -ErrorAction SilentlyContinue)) 
+{ 
+    Expand-Archive "nodejs.zip"
+}
+else
+{
+    Write-Host "Expand-Archive not found"
+    $shell = new-object -com shell.application
```
natemcmaster: Try this instead:

```ps1
            if (Get-Command -Name 'Microsoft.PowerShell.Archive\Expand-Archive' -ErrorAction Ignore) {
                # Use built-in commands where possible as they are cross-plat compatible
                Microsoft.PowerShell.Archive\Expand-Archive -Path $tmpfile -DestinationPath $destPath
            }
            else {
                # Fallback to old approach for old installations of PowerShell
                Add-Type -AssemblyName System.IO.Compression.FileSystem
                [System.IO.Compression.ZipFile]::ExtractToDirectory($tmpfile, $korebuildPath)
            }
```

source: https://github.com/aspnet/BuildTools/blob/ec244eafdc277b29813c81e19e9e9768cad8839a/scripts/bootstrapper/run.ps1#L116-L124

eng/targets/Helix.props
```diff
@@ -35,4 +36,14 @@
     <HelixPreCommand Include="call RunPowershell.cmd mssql\InstallSqlServerLocalDB.ps1 || exit /b 1" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TestDependsOnNode)' == 'true' AND '$(IsWindowsHelixQueue)' == 'false'">
+    <HelixPreCommand Include="./installnode.sh $(NodeVersion) linux-x64" />
```
natemcmaster: You probably don't need to pass in linux-x64 here. You can detect OS instead.

eng/helix/content/installnode.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+
+if type -P "node" &>/dev/null; then
+    echo "node is in \$PATH"
+    exit
+fi
+
+node_version=$1
+platformarch=$2
```
natemcmaster:
```suggestion
if [ "$uname" = "Darwin" ]; then
   platformarch='darwin-x64'
else
   platformarch='linux-x64'
fi
```

eng/targets/Helix.props
```diff
@@ -35,4 +36,14 @@
...
+    <HelixPreCommand Include="export PATH=%24PATH&quot;:&quot;%24HELIX_CORRELATION_PAYLOAD/node/bin" />
```
natemcmaster: Question, mostly out of curiosity. Could we symlink the node executable into some kind of HELIX_BIN directory? This is the 'normal' way to install things in *nix.

HaoK:
Getting pretty close, @natemcmaster 

The new Fedora 28 queue looks like something is different about where the files get dropped from helix content.  Do you know of any differences in how stuff is deployed to the docker queues?

So the content is getting dropped still
```
/home/helixbot/work/a8f6aa6b-18ae-4632-b882-44613fd6c226/Work/401bd1d3-623e-4b6d-b464-116bfd6b4d77/Exec/js/moduleWithDefaultExport.js'"
```

But at runtime, it appears to be working from slightly different directories from the other linux paths:

```
 Microsoft.AspNetCore.NodeServices.HostingModels.NodeInvocationException : Cannot find module '/root/helix/work/Exec/js/testCases'
```

I can change the test to look in a different place, but my guess is that would cause all of the other tests to fail (while fixing the Fedora

natemcmaster:
Sorry, not sure. I haven't looked closely into how the Helix runner launches the docker process. Only thing i can say is that, as a rule of thumb, everything that the test needs should be in the publish output directory, and the test code should be use System.AppContext.BaseDirectory to find those files.

HaoK:
Looks like the issue is the fedora queue is running from "/root/helix/work/workitem/.dotnet20383" which is different from the other ones which look something like: "/home/helixbot/work/9d15a323-9475-41bd-ad6d-5ee1698280af/Work/2b04a8ea-83a2-4dad-9635-a5c7e15760d1/Exec/.dotnet27980"

HaoK:
`DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"` must be different in the dockerized queues which ends up running someplace unexpected for the node tests anyways

HaoK:
Weird, its passing now, the runs seem to hang every so often, but these tests are finally green

https://mc.dot.net/#/user/aspnetcore/pr~2Faspnet~2Faspnetcore/ci/20190528.2/workItem/Microsoft.AspNetCore.NodeServices.Tests~2Fnetcoreapp3.0/wilogs

eng/helix/content/InstallNode.ps1
```diff
@@ -0,0 +1,60 @@
...
+    [Parameter(Mandatory = $true)]
+    $output_dir
+)
+
+$ErrorActionPreference = 'Stop'
+$ProgressPreference = 'SilentlyContinue' # Workaround PowerShell/PowerShell#2138
+
+Set-StrictMode -Version 1
+
+if ((Get-Command "node.exe" -ErrorAction SilentlyContinue))
```
dougbu: Why the extra parens?

eng/helix/content/InstallNode.ps1
```diff
@@ -0,0 +1,60 @@
...
+if (Test-Path "$output_dir\node.exe")
+{
+    Write-Host "Node.exe found at $output_dir"
+    exit
+}
+
+$nodeFile="node-v$Version-win-x64"
+$url="http://nodejs.org/dist/v$Version/$nodeFile.zip"
+Write-Host "Starting download of NodeJs ${Version} from $url"
+Invoke-WebRequest -UseBasicParsing -Uri "$url" -OutFile "nodejs.zip"
```
dougbu: Retries?

eng/helix/content/InstallNode.ps1
```diff
@@ -0,0 +1,60 @@
...
+Write-Host "Copying $temp_dir\$nodeFile\node.exe to $output_dir"
+Copy-Item "$temp_dir\$nodeFile\node.exe" "$output_dir\node.exe"
+
+if (Test-Path "$output_dir\node.exe")
+{
+    Write-Host "Node.exe copied to $output_dir"
+}
+else
+{
+    Write-Host "Node.exe not found at $output_dir"
```
dougbu: How could this be reached?

eng/helix/content/installnode.sh
```diff
@@ -0,0 +1,28 @@
...
+fi
+
+node_version=$1
+echo $uname
+if [ "$uname" = "Darwin" ]; then
+   platformarch='darwin-x64'
+else
+   platformarch='linux-x64'
+fi
+platformarch=$2
```
dougbu: This line should be removed.

eng/targets/Helix.props
```diff
@@ -35,4 +36,13 @@
     <HelixPreCommand Include="call RunPowershell.cmd mssql\InstallSqlServerLocalDB.ps1 || exit /b 1" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TestDependsOnNode)' == 'true' AND '$(IsWindowsHelixQueue)' == 'false'">
+    <HelixPreCommand Include="./installnode.sh $(NodeVersion) linux-x64" Condition="!$(HelixTargetQueue.Contains('OSX'))" />
+    <HelixPreCommand Include="./installnode.sh $(NodeVersion) darwin-x64" Condition="$(HelixTargetQueue.Contains('OSX'))" />
```
dougbu:
```suggestion
    <HelixPreCommand Include="./installnode.sh $(NodeVersion)" />
```
Because `installnode.sh` figures the platform out for itself.

HaoK: I tried that but it doesn't work, OSX would resolve to linux-x64 instead of darwin-x64 and fail

eng/helix/content/InstallNode.ps1
HaoK: I just left it in for debugging purposes, I could remove it but doesn't seem to hurt

eng/helix/content/InstallNode.ps1
HaoK: Helix automatically retries failed work items 3 times, that seems good enough, I don't think it adds much value adding nested retries here

eng/targets/Helix.props
HaoK: That sounds interesting, so are you thinking we create some kind of `/bin` that we put any things that we want to be in the path for runtests, and add that well known `/bin` to the path?  Then we could just add a symlink to that directory instead what i'm currently doing.  Well I guess it'd be the same just using a well known bin instead of a special node dir.

eng/targets/Helix.props
HaoK: Looks like $uname is undefined on the OSX.1014 queue

HaoK:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=203419

eng/helix/content/InstallNode.ps1
```diff
@@ -0,0 +1,60 @@
+
+param(
+    [Parameter(Mandatory = $true)]
+    $Version,
+    
+    [Parameter(Mandatory = $true)]
+    $output_dir
```
natemcmaster:
```suggestion
    $InstallDir
```

Nit: use PowerShell casing conventions. Also, not sure what "output" means in the context of an installer script, so I think "InstallDir" makes a little more sense.

eng/helix/content/InstallNode.ps1
```diff
@@ -0,0 +1,60 @@
+
```
natemcmaster: Can you add a doc-block? Example:
https://github.com/aspnet/AspNetCore/blob/8bb34bb30bd55cb26b928c05a6a12672bbe5b76a/eng/helix/content/mssql/InstallSqlServerLocalDB.ps1#L1-L14

eng/helix/content/InstallNode.ps1
```diff
@@ -0,0 +1,60 @@
...
+}
+
+$nodeFile="node-v$Version-win-x64"
+$url="http://nodejs.org/dist/v$Version/$nodeFile.zip"
+Write-Host "Starting download of NodeJs ${Version} from $url"
+Invoke-WebRequest -UseBasicParsing -Uri "$url" -OutFile "nodejs.zip"
+Write-Host "Done downloading NodeJS ${Version}"
+
+$temppath = [System.IO.Path]::GetTempPath()
+$temp_dir = Join-Path $temppath nodejs
```
natemcmaster:
```suggestion
$tempDir = Join-Path $temppath nodejs
```

Nit: use consistent casing conventions

eng/helix/content/installnode.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
```
natemcmaster:
```suggestion
#!/usr/bin/env bash

# Cause the script to fail if any subcommand fails
set -e
```

 Without this, failures in the pre-command will go unnoticed. https://github.com/dotnet/arcade/issues/2731

eng/helix/content/installnode.sh
```diff
@@ -0,0 +1,28 @@
...
+node_version=$1
+osname=`uname -s`
+echo $osname
+if [ "$osname" = "Darwin" ]; then
+   platformarch='darwin-x64'
+else
+   platformarch='linux-x64'
+fi
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+output_dir=$DIR/node
```
natemcmaster:
```suggestion
output_dir="$DIR/node"
```
Double quotes for safety in case $DIR contains spaces

natemcmaster (commented):
Mostly coding style nits. Looks like this is good! I'm going to leave a 2nd review about changes you could make which would also solve problems with that required overriding the location of `bin/`

src/Middleware/NodeServices/test/Microsoft.AspNetCore.NodeServices.Tests.csproj
```diff
@@ -2,19 +2,16 @@
...
     <OutputPath />
-
+    <TestDependsOnNode>true</TestDependsOnNode>
   </PropertyGroup>
 
   <ItemGroup>
     <Reference Include="Microsoft.AspNetCore.NodeServices" />
     <Reference Include="Microsoft.AspNetCore.TestHost" />
     <Reference Include="Microsoft.Extensions.Logging.Testing" />
+    <HelixContent Include="js\*.*" />
```
natemcmaster:
```suggestion
    <Content Include="js\**\*" />   
```

src/Middleware/NodeServices/test/NodeServicesTest.cs
```diff
@@ -115,7 +115,7 @@ public async Task CanInvokeDefaultModuleExport()
         }
 
         private static string ModulePath(string testModuleName)
-            => $"../../../js/{testModuleName}";
+            => string.IsNullOrEmpty(Environment.GetEnvironmentVariable("helix")) ? $"../../../js/{testModuleName}" : $"js/{testModuleName}";
```
natemcmaster:
```suggestion
            => Path.Combine(AppContext.BaseDirectory, "js", testModuleName);
```

src/Middleware/NodeServices/test/Microsoft.AspNetCore.NodeServices.Tests.csproj
```diff
@@ -2,19 +2,16 @@
 
   <PropertyGroup>
     <TargetFramework>netcoreapp3.0</TargetFramework>
-
-    <!-- Tests do not work on Helix or when bin/ directory is not in project directory due to undeclared dependency on test content. -->
-    <!-- https://github.com/aspnet/AspNetCore/issues/8044 -->
-    <BuildHelixPayload>false</BuildHelixPayload>
     <BaseOutputPath />
```
natemcmaster:
```suggestion
```

src/Middleware/NodeServices/test/Microsoft.AspNetCore.NodeServices.Tests.csproj
```diff
@@ -2,19 +2,16 @@
...
     <OutputPath />
```
natemcmaster:
```suggestion
```

natemcmaster (commented):
I believe if you apply all 4 of these suggested changes, then we don't need to special case running tests in Helix. This is the recommended approach for tests that depend on static content (see https://github.com/aspnet/Extensions/issues/1697)

natemcmaster (approved):
Thanks, Hao!

PR #9059, opened 2019-04-04
Make TestServer IAsyncDisposable

davidfowl:
- This should make it a bit more maintainable than it is today (having to call server.StopAsync before Dispose calls it synchronously)

PS: This is only for the InMemoryTests


JamesNK (approved):

src/Servers/Kestrel/test/InMemory.FunctionalTests/TestTransport/TestServer.cs
```diff
@@ -158,5 +157,12 @@ private static Task CancellationTokenAsTask(CancellationToken token)
             token.Register(() => tcs.SetResult(null));
             return tcs.Task;
         }
+
+        public async ValueTask DisposeAsync()
+        {
+            // The concrete WebHost implements IAsyncDisposable
+            await ((IAsyncDisposable)_host).ConfigureAwait(false).DisposeAsync();
```
JamesNK: If _host is always WebHost then the reference type should just be WebHost

Tratcher (approved):

Issue #6919, opened 2019-01-22
Add a service that allows constructing parts of an ApplicationModel instance


mgubis:
```C#
public abstract class PageApplicationModelPartsFactory
{
    public abstract ParameterModel CreateDefault(....);
} 
```

BTW the compat expectations of an API like this are:

1. This API will respect things like compat switches, so if we make behaviour change it's backed by a compat switch
2. We're allowed to make new overloads and call those instead if requirements change. The existing APIs will work the same and might be obsoleted over time.

--------------------------------------------------

We've been using .NET Core 2.1 for our project so far and tried to upgrade to 2.2. We are using customized IPageApplicationModelProvider. Since we wanted to preserve the functionality of the default razor provider, we are inheriting from DefaultPageApplicationModelProvider (as suggested in official Microsoft paper: https://docs.microsoft.com/en-us/aspnet/core/razor-pages/razor-pages-conventions?view=aspnetcore-2.2 (Replace the default page app model provider). The problem is, that in 2.2, DefaultPageApplicationModelProvider became internal and thus we can't inherit from it.

Is this way of replacing the default page app model provider not supported any more? Is there any other preferred way available?

Thank you.

pranavkm:
> The problem is, that in 2.2, DefaultPageApplicationModelProvider became internal and thus we can't inherit from it.
Unfortunately the type was in the `.Internal` namespace and inheriting from it was never a supported scenario. We have a longer discussion about this here - https://github.com/aspnet/AspNetCore/issues/4932


> Is this way of replacing the default page app model provider not supported any more? Is there any other preferred way available?

In MVC, provider types, including `IPageApplicationModel` provider instances are invoked in their declared order. The expected approach is to register a custom application model provider after the default one (an order of `0` should suffice since the default ones are initialized to -1000.). More commonly, we expect users to author conventions to modify the results of the application model provider.

I'll work on getting the docs updated, it shouldn't have recommended inheriting from the `DefaultApplicationModelProvider`. In the meanwhile, if you could detail your scenario, I'd be happy to provide the code that you would need and possibly use that as the sample in the docs.

mgubis:
Thank you for quick answer - I really appreciate it. 

I'll try to explain our use case:

[...]

PR #9066, opened 2019-04-04
Introduce IPageApplicationModelPartsProvider

mgubis:
DefaultApplicationModelProvider was split into several parts accessible via newly created IPageApplicationModelPartsProvider.

Addresses #6919


pranavkm:
Thanks @mgubis. This looks super nice. I'll work on reviewing and integrating this in the next preview release.

src/Mvc/Mvc.RazorPages/src/ApplicationModels/IPageApplicationModelPartsProvider.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Reflection;
+using System.Text;
+using Microsoft.AspNetCore.Mvc.RazorPages;
+
+namespace Microsoft.AspNetCore.Mvc.ApplicationModels
+{
+    public interface IPageApplicationModelPartsProvider
```
pranavkm: I'm slightly partial to making this an abstract base type, to make it easier to evolve it if we ever need to, but this looks really great as is

yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
pranavkm: Accidentally added?

src/Mvc/Mvc.RazorPages/src/ApplicationModels/DefaultPageApplicationModelPartsProvider.cs
```diff
@@ -0,0 +1,273 @@
+﻿using System;
```
pranavkm: Could you add this license header to the file:

```C#
```

src/Mvc/Mvc.RazorPages/src/ApplicationModels/IPageApplicationModelPartsProvider.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System;
```
pranavkm: Could you add this license header to the file:

```C#
```

pranavkm (commented):
This looks really nice. Let me know if you're able to address the few comments I had

yarn.lock
mgubis: Yes, sorry for this - will fix it asap.

src/Mvc/Mvc.RazorPages/src/ApplicationModels/DefaultPageApplicationModelPartsProvider.cs
mgubis: Sure, I'll add it asap.

src/Mvc/Mvc.RazorPages/src/ApplicationModels/IPageApplicationModelPartsProvider.cs
mgubis: Sure, I'll add it asap.

src/Mvc/Mvc.RazorPages/src/ApplicationModels/IPageApplicationModelPartsProvider.cs
mgubis: I agree it would allow us to add new functionality to the parts provider without breaking existing third-party parts providers. What I don't like about this approach is, that we will be extending silently parts providers, which were created by third-party. They customized it completely and then we will push new code without them knowing.
I would say, that if we want to allow partial update of the parts provider (thus allowing people to accept default functionality and change/extend only parts), we should expose DefaultPageApplicationModelPartsProvider and make methods virtual. Since this is not desired, I would keep the current solution - if somebody creates new parts provider and we add new functionality, he will get compilation error and acknowledges, that there is new stuff he has to implement.
What is your view of this?

pranavkm: >  What I don't like about this approach is, that we will be extending silently parts providers, which were created by third-party.

While that's true, it may hamstrings the ability to add new feature and end up with code that ends up downcasting: 
https://github.com/aspnet/AspNetCore/blob/master/src/Mvc/Mvc.Core/src/ApplicationModels/DefaultApplicationModelProvider.cs#L470-L479.

I'm ok leaving this as is now. That said, could you add a comment to the interface since it's new API:

> Provides parts that are used to construct a `<see cref="PageApplicationModel" /> instance`

mgubis: Okay - comment added; it is definitely appropriate to have the comment there - thanks!

pranavkm (approved):

pranavkm:
Thanks, I'll get this merged once the build passes

mgubis:
Hi @pranavkm, I noticed, that the code has test failures. It doesn't have any problems in my branch, so it is probably connected to something with merge. Is there anything I can help with or will you handle this?
Thanks.

pranavkm:
@mgubis could you update the ref assemblies in the Mvc.RazorPages project. See https://github.com/aspnet/AspNetCore/blob/master/docs/ReferenceAssemblies.md#when-changing-public-api

mgubis:
Thanks @pranavkm for pointing me in right direction. Ref assemblies should be updated now and all checks have passed. Please let me know if I can be of any other help in merging this pull request.

pranavkm:
Thanks for the PR @mgubis

PR #9068, opened 2019-04-04
[Infrastructure] Retry yarn install on failure

javiercn:
Retries yarn install in case of failure

### DIFF
--- a/eng/targets/Npm.Common.targets
+++ b/eng/targets/Npm.Common.targets
@@ -18,7 +18,12 @@
 
   <Target Name="Restore">
     <Message Importance="High" Text="Running yarn install on $(MSBuildProjectFullPath)" />
-    <Yarn Command="install $(InstallArgs)" />
+    <Yarn Command="install $(InstallArgs)" IgnoreExitCode="true">
+      <Output TaskParameter="ExitCode" PropertyName="_YarnExitCode"/>
+    </Yarn>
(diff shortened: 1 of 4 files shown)


analogrelay (approved):

dougbu (approved):

eng/targets/Npm.Common.targets
```diff
@@ -18,7 +18,12 @@
 
   <Target Name="Restore">
     <Message Importance="High" Text="Running yarn install on $(MSBuildProjectFullPath)" />
-    <Yarn Command="install $(InstallArgs)" />
+    <Yarn Command="install $(InstallArgs)" ContinueOnError="true">
+      <Output TaskParameter="ExitCode" PropertyName="_YarnExitCode"/>
+    </Yarn>
+    <Yarn Command="install $(InstallArgs)" Condition="'$(_YarnExitCode)' != '0'">
```
ryanbrandenburg: Are the failures we experience with Yarn usually related to networking problems? If so it would seem we want a small delay between the two tries. If not this is sufficient.

javiercn: It's due to concurrent restores I believe. I'm gonna start here and see if it improves. I have no idea of how to add a delay on MSBuild

anurse: A custom task is always an option, if necessary.

javiercn: Yeah, but lets not go there just yet.

PR #9080, opened 2019-04-04
Standarize E2E test asserts and increase wait time

javiercn:
* Converts the templating asserts to use the shared infrastructure asserts.
* Increases the timeout for templating tests from 10 seconds to 30 to help with reliability.


javiercn:
This was giving me grief today in several of my PRs. Seems that since we've moved to BYOC queues these tests have become unstable again. I've increased the timeout to see if that helps and replaced the asserts for the ones that we use in Blazor which are more robust.

ryanbrandenburg (approved):
Seems fine.

src/ProjectTemplates/test/RazorComponentsTemplateTest.cs
```diff
@@ -75,35 +75,31 @@ private void TestBasicNavigation()
...
-            Assert.Contains(Project.ProjectGuid, Browser.Title);
+            Browser.Equal(Project.ProjectName.Trim(), () => Browser.Title.Trim());
 
             // Initially displays the home page
-            Assert.Equal("Hello, world!", Browser.GetText("h1"));
+            Browser.Equal("Hello, world!", () => Browser.FindElement(By.TagName("h1")).Text);
 
             // Can navigate to the counter page
-            Browser.Click(By.PartialLinkText("Counter"));
-            Browser.WaitForUrl("counter");
```
SteveSandersonMS: @javiercn Just so we can understand what the right APIs are to use in the future, could you clarify why `WaitForUrl` is wrong? Does it not wait in the right way or something?

src/ProjectTemplates/test/RazorComponentsTemplateTest.cs
```diff
@@ -75,35 +75,31 @@ private void TestBasicNavigation()
...
-            Browser.WaitForText("h1", "Counter");
+            Browser.FindElement(By.PartialLinkText("Counter")).Click();
+            Browser.Contains("counter", () => Browser.Url);
+            Browser.Equal("Counter", () => Browser.FindElement(By.TagName("h1")).Text);
 
             // Clicking the counter button works
-            var counterComponent = Browser.FindElement("h1").Parent();
-            var counterDisplay = Browser.FindElement("h1 + p");
```
SteveSandersonMS: I see you didn't want to hang onto the same element references, but to save me from making mistakes in the future, could you explain why this is wrong to do?

I would have reasoned that, by the time we are on the "Counter" page, the `counterDisplay` element instance is going to remain fixed, so it's safe to hold it. Arguably it's even preferable to do so, as it shows we're not doing completely broken things like tearing down the entire UI and recreating it after each click.

src/Shared/E2ETesting/WaitAssert.cs
```diff
@@ -57,16 +61,24 @@ private static void WaitAssertCore(IWebDriver driver, Action assertion, TimeSpan
...
-                assertion();
+                if (lastException != null)
+                {
+                    ExceptionDispatchInfo.Capture(lastException).Throw();
+                }
+                else
+                {
+                    // Instead of reporting it as a timeout, report the Xunit exception
+                    assertion();
+                }
```
SteveSandersonMS: Can you clarify why it's advantageous to do this? I recall we discussed this before. What reason is there to prefer `lastException` over invoking `assertion` again? This logic seems really arbitrary and misleading about its purpose.

javiercn: Rerunning the assertion might make it fail for other reasons.

src/ProjectTemplates/test/RazorComponentsTemplateTest.cs
javiercn: It's mostly that it was in a separate class with a different way to wait and with different settings to configure it. I want us to share the asserts we use for E2E testing so that when we make an improvement it translates to the rest of the codebase.

src/Shared/E2ETesting/WaitAssert.cs
SteveSandersonMS: We're repeating the same discussion from last time :)

If we run the assertion 15 times and it fails each time, then we run it for a 16th time and it fails again, I don't see why we should regard the 15th exception as any more "correct" than the 16th or the 1st or any other in the sequence.

src/ProjectTemplates/test/RazorComponentsTemplateTest.cs
javiercn: I didn't think much of that. I was trying to make the tests terser

src/ProjectTemplates/test/RazorComponentsTemplateTest.cs
SteveSandersonMS: Ah, so `WaitForUrl` is our own custom thing and not something built-in? If so then I totally understand. Yes it definitely makes sense to have only one set of custom wait-assertions!

src/ProjectTemplates/test/RazorComponentsTemplateTest.cs
SteveSandersonMS: OK, sounds fine.

src/Shared/E2ETesting/WaitAssert.cs
javiercn: I believe 15th is equally incorrect to 1st, but more correct than 16th. If you feel strongly about it, I'll revert it (i'm doing some changes in spa templates, so I can do it while I'm there).

src/ProjectTemplates/test/RazorComponentsTemplateTest.cs
javiercn: Yeah, all these methods were helpers on the templating test project. The ones in blazor are better IMO. If we need more, they should be added there.

src/Shared/E2ETesting/WaitAssert.cs
SteveSandersonMS: Nah, on further consideration, it doesn't really matter. There'll be no observable difference, so either way is fine. Sorry for raising it.

PR #9088, opened 2019-04-04
Mark test as condition so it is actually skipped on helix.

jkotalik:


ryanbrandenburg (approved):

SteveSandersonMS:
There's nothing different about these two components E2E tests compared with all the other ones. I'm not sure how it can be right to make these two specifically as flaky.

There's a reasonable chance that @javiercn's other PR #9080 might help reduce flakiness across the board with the E2E tests. Can we see how that goes before picking these two to mark as flaky?

SteveSandersonMS:
BTW what do you think, @javiercn? I know you've been battling test flakiness lately.

jkotalik:
>  Can we see how that goes before picking these two to mark as flaky?

I don't think we should hesitate on marking a test as flaky. They are still run as part of every build and are still triaged by RAAS. CanRenderTextOnlyComponent has failed twice already too. If one of these two tests fail again, it will block someone from potentially merging a PR that shouldn't be blocked.

jkotalik:
Going to close this as I haven't seen these be flaky now and based on @SteveSandersonMS description, these tests aren't special (it means any test could fail).

PR #9094, opened 2019-04-04
Fix template baseline

ryanbrandenburg:


jkotalik (approved):
If it works!

mkArtakMSFT (approved):

javiercn (approved):

javiercn:
How did this test even pass the checks in the old PR?

ryanbrandenburg:
The jist of it is that our PR's went in without either being rebased onto the other, but the failure only happens when they're combined (I added a test for "razorcomponents" but you were changing the name). That kind of timing problem is seems unavoidable to me without requiring a latest rebase check pass before you merge, which would make merging a PR glacially slow.

Issue #8387, opened 2019-03-10
Authorization broken in Razor Pages .Net Core 3.0 preview 3


Ponant:
Asp.Net Core Identity 3.0 Preview 3 template with RP
Authorize attribute does nothing on the Privacy page
AuthorizePage nothing
AuthorizeFolder nothing

You can also navigate to /Identity/Account/Manage, which should trigger back to login whereas it shows Unable to load user with ID '' gotten from the get request.
Is this planned for preview 4?
Related to https://github.com/aspnet/AspNetCore/issues/7011 which is flagged as done and closed but does not work.

Eilon:
@Ponant can you show exactly what modifications you made to the app after the initial project was created?

Ponant:
@Eilon , go to the Identity/Account/Manage page of the template; there is no need to change anything to the template to get this bug.
Otherwise, you can do this
```
    [Authorize]
    public class PrivacyModel : PageModel
```
but you will be able to access the Privacy page

Also
 services.AddMvc().AddRazorPagesOptions(o=>o.Conventions.AuthorizePage("/Privacy"))
                .AddNewtonsoftJson();

Eilon:
@javiercn / @pranavkm - any idea?

javiercn:
Not sure what's happening here, but I'll take a look.

Ponant:
Just create the identity template and navigate to identity/account/manage and you should see it traversing the get request whereas it requires authorization. You can also use the AuthorizePage and the like and it should just ignore them. Let me know if you can reproduce. Cheers

mkArtak:
Thanks for the details, @Ponant.
I could repro this using Preview3 bits.
This is where the failure is coming from: https://github.com/aspnet/AspNetCore/blob/c95ee2b051814b787b07f55ff224d03d550aafeb/src/Identity/samples/IdentitySample.DefaultUI/Areas/Identity/Pages/Account/Manage/Index.cshtml.cs#L68
Obviously, the expectation here is that the user should be known. So indeed, authorization was completely bypassed.

Ponant:
@mkArtak , out of curiosity, do you know from where the bug comes from?

mkArtak:
Not really, @Ponant. Let's wait for @pranavkm to handle this - he'll do a great job here.

rynowak:
[...]

PR #9099, opened 2019-04-05
Make AuthorizeFilter work in endpoint routing

pranavkm:
Fixes https://github.com/aspnet/AspNetCore/issues/8387

### DIFF
--- a/src/Identity/test/Identity.FunctionalTests/IdentityUserTests/IdentityUserAuthorizationWithoutEndpointRoutingTests.cs
+++ b/src/Identity/test/Identity.FunctionalTests/IdentityUserTests/IdentityUserAuthorizationWithoutEndpointRoutingTests.cs
@@ -1,0 +1,16 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Identity.DefaultUI.WebSite;
+using Microsoft.AspNetCore.Identity.EntityFrameworkCore;
+
+namespace Microsoft.AspNetCore.Identity.FunctionalTests.IdentityUserTests
+{
+    public class IdentityUserAuthorizationWithoutEndpointRoutingTests : AuthorizationTests<StartupWithoutEndpointRouting, IdentityDbContext>
+    {
+        public IdentityUserAuthorizationWithoutEndpointRoutingTests(ServerFactory<StartupWithoutEndpointRouting, IdentityDbContext> serverFactory)
+            : base(serverFactory)
+        {
+        }
+    }
+}
--- a/src/Middleware/CORS/src/Infrastructure/CorsMiddleware.cs
+++ b/src/Middleware/CORS/src/Infrastructure/CorsMiddleware.cs
@@ -145,8 +145,23 @@ namespace Microsoft.AspNetCore.Cors.Infrastructure
             // Get the most significant CORS metadata for the endpoint
             // For backwards compatibility reasons this is then downcast to Enable/Disable metadata
             var corsMetadata = endpoint?.Metadata.GetMetadata<ICorsMetadata>();
+
             if (corsMetadata is IDisableCorsAttribute)
             {
+                var isOptionsRequest = string.Equals(
+                    context.Request.Method,
+                    CorsConstants.PreflightHttpMethod,
+                    StringComparison.OrdinalIgnoreCase);
+
+                var isCorsPreflightRequest = isOptionsRequest && context.Request.Headers.ContainsKey(CorsConstants.AccessControlRequestMethod);
+
+                if (isCorsPreflightRequest)
+                {
+                    // If this is a preflight request, and we disallow CORS, complete the request
+                    context.Response.StatusCode = StatusCodes.Status204NoContent;
+                    return;
+                }
+
                 await _next(context);
                 return;
             }
--- a/src/Mvc/Mvc.Core/src/ApplicationModels/ActionAttributeRouteModel.cs
+++ b/src/Mvc/Mvc.Core/src/ApplicationModels/ActionAttributeRouteModel.cs
@@ -132,11 +132,12 @@ namespace Microsoft.AspNetCore.Mvc.ApplicationModels
             if (controllerMetadata != null)
             {
                 // It is criticial to get the order in which metadata appears in endpoint metadata correct. More significant metadata
-                // must appear later in the sequence.
-                // In this case, the parameter named 'metadata' is the controller endpoint metadata and should appear earlier than the existing metadata
-                // that represents the action metadata.
-                var combinedMetadata = Enumerable.Concat(controllerMetadata, selector.EndpointMetadata).ToList();
-                selector.EndpointMetadata = combinedMetadata;
+                // must appear later in the sequence. In this case, the values in `controllerMetadata` should have their order
+                // preserved, but appear earlier than the entries in `selector.EndpointMetadata`.
+                for (var i = 0; i < controllerMetadata.Count; i++)
+                {
+                    selector.EndpointMetadata.Insert(i, controllerMetadata[i]);
+                }
             }
         }
 
--- a/src/Mvc/Mvc.Core/src/ApplicationModels/ActionAttributeRouteModel.cs
+++ b/src/Mvc/Mvc.Core/src/ApplicationModels/ActionAttributeRouteModel.cs
@@ -131,9 +131,13 @@ namespace Microsoft.AspNetCore.Mvc.ApplicationModels
         {
             if (metadata != null)
             {
-                for (var i = 0; i < metadata.Count; i++)
+                for (var i = metadata.Count - 1; i >= 0; i--)
                 {
-                    selector.EndpointMetadata.Add(metadata[i]);
+                    // It is criticial to get the order in which metadata appears in endpoint metadata correct. More significant metadata
+                    // must appear later in the sequence.
+                    // In this case, the parameter named 'metadata' is the controller endpoint metadata and should appear earlier than the existing metadata
+                    // that represents the action metadata.
+                    selector.EndpointMetadata.Insert(0, metadata[i]);
                 }
             }
         }
--- a/src/Mvc/Mvc.Core/src/ApplicationModels/ActionAttributeRouteModel.cs
+++ b/src/Mvc/Mvc.Core/src/ApplicationModels/ActionAttributeRouteModel.cs
@@ -127,18 +127,16 @@ namespace Microsoft.AspNetCore.Mvc.ApplicationModels
             }
         }
 
-        private static void AddEndpointMetadata(SelectorModel selector, IList<object> metadata)
+        private static void AddEndpointMetadata(SelectorModel selector, IList<object> controllerMetadata)
         {
-            if (metadata != null)
+            if (controllerMetadata != null)
             {
-                for (var i = metadata.Count - 1; i >= 0; i--)
-                {
-                    // It is criticial to get the order in which metadata appears in endpoint metadata correct. More significant metadata
-                    // must appear later in the sequence.
-                    // In this case, the parameter named 'metadata' is the controller endpoint metadata and should appear earlier than the existing metadata
-                    // that represents the action metadata.
-                    selector.EndpointMetadata.Insert(0, metadata[i]);
-                }
+                // It is criticial to get the order in which metadata appears in endpoint metadata correct. More significant metadata
+                // must appear later in the sequence.
+                // In this case, the parameter named 'metadata' is the controller endpoint metadata and should appear earlier than the existing metadata
+                // that represents the action metadata.
+                var combinedMetadata = Enumerable.Concat(controllerMetadata, selector.EndpointMetadata).ToList();
+                selector.EndpointMetadata = combinedMetadata;
             }
         }
 
--- a/src/Mvc/Mvc.Core/src/ApplicationModels/AuthorizationApplicationModelProvider.cs
+++ b/src/Mvc/Mvc.Core/src/ApplicationModels/AuthorizationApplicationModelProvider.cs
@@ -6,16 +6,21 @@ using System.Collections.Generic;
 using System.Linq;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc.Authorization;
+using Microsoft.Extensions.Options;
 
 namespace Microsoft.AspNetCore.Mvc.ApplicationModels
 {
     internal class AuthorizationApplicationModelProvider : IApplicationModelProvider
     {
+        private readonly MvcOptions _mvcOptions;
         private readonly IAuthorizationPolicyProvider _policyProvider;
 
-        public AuthorizationApplicationModelProvider(IAuthorizationPolicyProvider policyProvider)
+        public AuthorizationApplicationModelProvider(
+            IAuthorizationPolicyProvider policyProvider,
+            IOptions<MvcOptions> mvcOptions)
         {
             _policyProvider = policyProvider;
+            _mvcOptions = mvcOptions.Value;
         }
 
         public int Order => -1000 + 10;
@@ -30,6 +35,13 @@ namespace Microsoft.AspNetCore.Mvc.ApplicationModels
             if (context == null)
             {
                 throw new ArgumentNullException(nameof(context));
+            }
+
+            if (_mvcOptions.EnableEndpointRouting)
+            {
+                // When using endpoint routing, the AuthorizationMiddleware does the work that Auth filters would otherwise perform.
+                // Consequently we do not need to convert authorization attributes to filters.
+                return;
             }
 
             foreach (var controllerModel in context.Result.Controllers)
--- a/src/Mvc/Mvc.Core/src/Authorization/AuthorizeFilter.cs
+++ b/src/Mvc/Mvc.Core/src/Authorization/AuthorizeFilter.cs
@@ -8,6 +8,7 @@ using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Authorization.Policy;
+using Microsoft.AspNetCore.Http.Endpoints;
 using Microsoft.AspNetCore.Mvc.ApplicationModels;
 using Microsoft.AspNetCore.Mvc.Core;
 using Microsoft.AspNetCore.Mvc.Filters;
@@ -22,9 +23,6 @@ namespace Microsoft.AspNetCore.Mvc.Authorization
     /// </summary>
     public class AuthorizeFilter : IAsyncAuthorizationFilter, IFilterFactory
     {
-        // Property key set by authorization middleware when it is run
-        private const string AuthorizationMiddlewareInvokedKey = "__AuthorizationMiddlewareInvoked";
-
         private AuthorizationPolicy _effectivePolicy;
 
         /// <summary>
@@ -116,6 +114,7 @@ namespace Microsoft.AspNetCore.Mvc.Authorization
             {
                 return Task.FromResult(Policy);
             }
+
             if (PolicyProvider == null)
             {
                 throw new InvalidOperationException(
@@ -127,7 +126,7 @@ namespace Microsoft.AspNetCore.Mvc.Authorization
             return AuthorizationPolicy.CombineAsync(PolicyProvider, AuthorizeData);
         }
 
-        private async Task<AuthorizationPolicy> GetEffectivePolicyAsync(AuthorizationFilterContext context)
+        internal async Task<AuthorizationPolicy> GetEffectivePolicyAsync(AuthorizationFilterContext context)
         {
             if (_effectivePolicy != null)
             {
@@ -154,6 +153,27 @@ namespace Microsoft.AspNetCore.Mvc.Authorization
                 }
             }
 
+            var endpoint = context.HttpContext.GetEndpoint();
+            if (endpoint != null)
+            {
+                // When doing endpoint routing, MVC does not create filters for any authorization specific metadata i.e [Authorize] does not
+                // get translated into AuthorizeFilter. Consequently, there are some rough edges when an application uses a mix of AuthorizeFilter
+                // explicilty configured by the user (e.g. global auth filter), and uses endpoint metadata.
+                // To keep the behavior of AuthFilter identical to pre-endpoint routing, we will gather auth data from endpoint metadata
+                // and produce a policy using this. This would mean we would have effectively run some auth twice, but it maintains compat.
+                var policyProvider = PolicyProvider ?? context.HttpContext.RequestServices.GetRequiredService<IAuthorizationPolicyProvider>();
+                var endpointAuthorizeData = endpoint.Metadata.GetOrderedMetadata<IAuthorizeData>() ?? Array.Empty<IAuthorizeData>();
+
+                var endpointPolicy = await AuthorizationPolicy.CombineAsync(policyProvider, endpointAuthorizeData);
+                if (endpointPolicy != null)
+                {
+                    builder.Combine(endpointPolicy);
+                }
+
+                // We cannot cache the policy since it varies by endpoint metadata.
+                canCache = false;
+            }
+
             effectivePolicy = builder?.Build() ?? effectivePolicy;
 
             // We can cache the effective policy when there is no custom policy provider
@@ -173,12 +193,6 @@ namespace Microsoft.AspNetCore.Mvc.Authorization
                 throw new ArgumentNullException(nameof(context));
             }
 
-            if (context.HttpContext.Items.ContainsKey(AuthorizationMiddlewareInvokedKey))
-            {
-                // Authorization has already run in middleware. Don't re-run for performance
-                return;
-            }
-
             if (!context.IsEffectivePolicy(this))
             {
                 return;
@@ -196,7 +210,7 @@ namespace Microsoft.AspNetCore.Mvc.Authorization
             var authenticateResult = await policyEvaluator.AuthenticateAsync(effectivePolicy, context.HttpContext);
 
             // Allow Anonymous skips all authorization
-            if (HasAllowAnonymous(context.Filters))
+            if (HasAllowAnonymous(context))
             {
                 return;
             }
@@ -226,14 +240,24 @@ namespace Microsoft.AspNetCore.Mvc.Authorization
             return AuthorizationApplicationModelProvider.GetFilter(policyProvider, AuthorizeData);
         }
 
-        private static bool HasAllowAnonymous(IList<IFilterMetadata> filters)
+        private static bool HasAllowAnonymous(AuthorizationFilterContext context)
         {
+            var filters = context.Filters;
             for (var i = 0; i < filters.Count; i++)
             {
                 if (filters[i] is IAllowAnonymousFilter)
                 {
                     return true;
                 }
+            }
+
+            // When doing endpoint routing, MVC does not add AllowAnonymousFilters for AllowAnonymousAttributes that
+            // were discovered on controllers and actions. To maintain compat with 2.x, 
+            // we'll check for the presence of IAllowAnonymous in endpoint metadata.
+            var endpoint = context.HttpContext.GetEndpoint();
+            if (endpoint?.Metadata?.GetMetadata<IAllowAnonymous>() != null)
+            {
+                return true;
             }
 
             return false;
--- a/src/Mvc/Mvc.Core/src/Infrastructure/ControllerActionInvokerProvider.cs
+++ b/src/Mvc/Mvc.Core/src/Infrastructure/ControllerActionInvokerProvider.cs
@@ -16,7 +16,7 @@ namespace Microsoft.AspNetCore.Mvc.Infrastructure
     internal class ControllerActionInvokerProvider : IActionInvokerProvider
     {
         private readonly ControllerActionInvokerCache _controllerActionInvokerCache;
-        private readonly IReadOnlyList<IValueProviderFactory> _valueProviderFactories;
+        private readonly IList<IValueProviderFactory> _valueProviderFactories;
         private readonly int _maxModelValidationErrors;
         private readonly ILogger _logger;
         private readonly DiagnosticListener _diagnosticListener;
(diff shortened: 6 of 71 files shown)


pranavkm:
I'll work on adding more tests \ but this would be a start

src/Identity/test/Identity.FunctionalTests/IdentityUserTests/IdentityUserAuthorizationWithoutEndpointRoutingTests.cs
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Identity.DefaultUI.WebSite;
+using Microsoft.AspNetCore.Identity.EntityFrameworkCore;
+
+namespace Microsoft.AspNetCore.Identity.FunctionalTests.IdentityUserTests
+{
+    public class IdentityUserAuthorizationWithoutEndpointRoutingTests : AuthorizationTests<StartupWithoutEndpointRouting, IdentityDbContext>
```
pranavkm: Adding a few IdentityUI tests to work with non-endpoint routing for sanity

src/Middleware/CORS/src/Infrastructure/CorsMiddleware.cs
```diff
@@ -145,8 +145,23 @@ private async Task InvokeCore(HttpContext context, ICorsPolicyProvider corsPolic
             // Get the most significant CORS metadata for the endpoint
             // For backwards compatibility reasons this is then downcast to Enable/Disable metadata
             var corsMetadata = endpoint?.Metadata.GetMetadata<ICorsMetadata>();
+
             if (corsMetadata is IDisableCorsAttribute)
             {
+                var isOptionsRequest = string.Equals(
```
pranavkm: Matches the current behavior of the CORS filter in MVC. The filter short-circuits if a `IDisableCorsAttribute` is present and you're doing a preflight request

src/Mvc/Mvc.Core/src/ApplicationModels/ActionAttributeRouteModel.cs
```diff
@@ -131,9 +131,13 @@ private static void AddEndpointMetadata(SelectorModel selector, IList<object> me
         {
             if (metadata != null)
             {
-                for (var i = 0; i < metadata.Count; i++)
+                for (var i = metadata.Count - 1; i >= 0; i--)
```
pranavkm: This was messing up the `GetMetadata` call in Auth. If you had `AuthAttribute` on a controller and action, we were currently returning the value from the controller.

src/Mvc/Mvc.Core/src/Authorization/AuthorizeFilter.cs
```diff
@@ -154,6 +153,27 @@ private async Task<AuthorizationPolicy> GetEffectivePolicyAsync(AuthorizationFil
...
+                var endpointAuthorizeData = endpoint.Metadata.GetOrderedMetadata<IAuthorizeData>() ?? Array.Empty<IAuthorizeData>();
+
+                var endpointPolicy = await AuthorizationPolicy.CombineAsync(policyProvider, endpointAuthorizeData);
+                if (endpointPolicy != null)
+                {
+                    builder.Combine(endpointPolicy);
+                }
+
+                // We cannot cache the policy since it varies by endpoint metadata.
+                canCache = false;
```
pranavkm: 😢

src/Mvc/Mvc.Core/test/ApplicationModels/ControllerActionDescriptorProviderTests.cs
```diff
@@ -266,16 +266,16 @@ public void GetDescriptors_EndpointMetadata_ContainsAttributesFromActionAndContr
...
-                metadata => Assert.IsType<AllowAnonymousAttribute>(metadata),
-                metadata => Assert.IsType<AuthorizeAttribute>(metadata));
+                metadata => Assert.IsType<AuthorizeAttribute>(metadata),
+                metadata => Assert.IsType<AllowAnonymousAttribute>(metadata));
 
             var authorizeAction = Assert.Single(descriptors, a => a.RouteValues["action"] == "AuthorizeAction");
 
             Assert.NotNull(authorizeAction.EndpointMetadata);
 
             Assert.Collection(authorizeAction.EndpointMetadata,
```
pranavkm: Results of the ordering change

src/Mvc/Mvc.Core/test/Authorization/AuthorizeFilterTest.cs
```diff
@@ -527,6 +512,89 @@ public void CreateInstance_ReturnsNewFilterIfPolicyAndPolicyProviderAreNotSet(Au
             Assert.Same(policyProvider, actual.PolicyProvider);
         }
 
+        [Fact]
+        public async Task GetEffectivePolicyAsync_ReturnsCurrentPolicy_WhenNoEndpointMetadataIsAvailable()
```
pranavkm: A couple of unit tests to go through the motion of verifying the combining code works

src/Mvc/Mvc.Cors/src/CorsApplicationModelProvider.cs
```diff
@@ -30,8 +39,21 @@ public void OnProvidersExecuting(ApplicationModelProviderContext context)
                 throw new ArgumentNullException(nameof(context));
             }
 
+            if (_mvcOptions.EnableEndpointRouting)
+            {
+                // When doing endpoint routing, translate IEnableCorsAttribute to an HttpMethodMetadata with CORS enabled.
+                ConfigureCorsEndpointMetadata(context.Result);
```
pranavkm: Should this go in endpoint routing? Right now, you could configure `ICorsMetadata` for an endpoint, but that would not cause the `HttpMethodMetadata` to change to allow CORS

src/Mvc/Mvc.Cors/src/CorsAuthorizationFilter.cs
```diff
@@ -110,8 +110,7 @@ public async Task OnAuthorizationAsync(Filters.AuthorizationFilterContext contex
                     !StringValues.IsNullOrEmpty(accessControlRequestMethod))
                 {
                     // If this was a preflight, there is no need to run anything else.
-                    // Also the response is always 200 so that anyone after mvc can handle the pre flight request.
-                    context.Result = new StatusCodeResult(StatusCodes.Status200OK);
+                    context.Result = new StatusCodeResult(StatusCodes.Status204NoContent);
```
pranavkm: This matches what the CORS middleware does.

src/Mvc/Mvc.Cors/src/CorsHttpMethodActionConstraint.cs
```diff
@@ -8,7 +8,6 @@
 
 namespace Microsoft.AspNetCore.Mvc.Cors
 {
-    // Don't casually change the name of this. We reference the full type name in ActionConstraintCache.
```
pranavkm: We no longer produce instances of this when doing endpoint routing, so the comment is no longer relevant (I think)

src/Mvc/Mvc.RazorPages/src/ApplicationModels/CompiledPageActionDescriptorBuilder.cs
```diff
@@ -61,6 +61,18 @@ internal static class CompiledPageActionDescriptorBuilder
...
         }
 
+        private static IList<object> CreateEndPointMetadata(PageApplicationModel applicationModel)
+        {
+            var handlerMetatdata = applicationModel.HandlerTypeAttributes;
+            var endPointMetadata = applicationModel.EndpointMetadata;
+
+            // It is criticial to get the order in which metadata appears in endpoint metadata correct. More significant metadata
+            // must appear later in the sequence.
+            // In this case, handlerMetadata is attributes on the Page \ PageModel, and endPointMetadata is configured via conventions. and
```
pranavkm: `AuthorizePage(..)` would have added at the end of the filter collection making it the effective filter. This matches the behavior

src/Mvc/Mvc.RazorPages/src/ApplicationModels/PageConventionCollection.cs
```diff
@@ -27,6 +28,13 @@ public PageConventionCollection(IList<IPageConvention> conventions)
         {
         }
 
+        internal PageConventionCollection(MvcOptions mvcOptions)
```
pranavkm: There isn't a way to set the collection on the options today. Seemed like it was OK to let this be internal

src/Mvc/Mvc.RazorPages/src/DependencyInjection/PageConventionCollectionExtensions.cs
```diff
@@ -60,7 +61,31 @@ public static PageConventionCollection ConfigureFilter(this PageConventionCollec
...
 
         /// <summary>
-        /// Adds a <see cref="AllowAnonymousFilter"/> to the page with the specified name.
+        /// Adds the specified <paramref name="convention"/> to <paramref name="conventions"/>.
+        /// The added convention will apply to all handler properties and parameters on handler methods.
+        /// </summary>
+        /// <param name="conventions">The <see cref="PageConventionCollection"/> to configure.</param>
+        /// <param name="convention">The <see cref="IParameterModelBaseConvention"/> to apply.</param>
+        /// <returns>The <see cref="PageConventionCollection"/>.</returns>
+        public static PageConventionCollection Add(this PageConventionCollection conventions, IParameterModelBaseConvention convention)
```
pranavkm: This got moved so all the auth stuff is grouped together.

src/Middleware/CORS/src/Infrastructure/CorsMiddleware.cs
```diff
@@ -145,8 +145,23 @@ private async Task InvokeCore(HttpContext context, ICorsPolicyProvider corsPolic
...
+
+                var isCorsPreflightRequest = isOptionsRequest && context.Request.Headers.ContainsKey(CorsConstants.AccessControlRequestMethod);
+
+                if (isCorsPreflightRequest)
+                {
+                    // If this is a preflight request, and we disallow CORS, complete the request
+                    context.Response.StatusCode = StatusCodes.Status204NoContent;
+                    return;
+                }
+
```
rynowak: That's kinda interesting that we missed this in the middleware. so were preflight requests going through to the the thing that handles the request? I'm not 100% certain of the ramifications of this change.

src/Mvc/Mvc.Core/src/ApplicationModels/ActionAttributeRouteModel.cs
rynowak: ugh.

rynowak: I don't actually get this. So we have a list of metadata, and you're reversing it... this seems like the wrong fix the problem.

rynowak: Metadata is always `most significant = last`. so whatever is putting them in the wrong order is what needs the fix.

src/Mvc/Mvc.Core/src/ApplicationModels/AuthorizationApplicationModelProvider.cs
```diff
@@ -32,6 +37,13 @@ public void OnProvidersExecuting(ApplicationModelProviderContext context)
                 throw new ArgumentNullException(nameof(context));
             }
 
+            if (_mvcOptions.EnableEndpointRouting)
+            {
+                // When using endpoint routing, the AuthorizationMiddleware does the work that Auth filters would otherwise perform.
```
rynowak: 👍

src/Mvc/Mvc.Core/src/Authorization/AuthorizeFilter.cs
rynowak: meh.

src/Mvc/Mvc.Core/src/Authorization/AuthorizeFilter.cs
```diff
@@ -154,6 +153,27 @@ private async Task<AuthorizationPolicy> GetEffectivePolicyAsync(AuthorizationFil
...
+                // and produce a policy using this. This would mean we would have effectively run some auth twice, but it maintains compat.
+                var policyProvider = PolicyProvider ?? context.HttpContext.RequestServices.GetRequiredService<IAuthorizationPolicyProvider>();
+                var endpointAuthorizeData = endpoint.Metadata.GetOrderedMetadata<IAuthorizeData>() ?? Array.Empty<IAuthorizeData>();
+
+                var endpointPolicy = await AuthorizationPolicy.CombineAsync(policyProvider, endpointAuthorizeData);
+                if (endpointPolicy != null)
+                {
+                    builder.Combine(endpointPolicy);
+                }
+
```
rynowak: Do we need to handle `IAllowAnonymous` here?

rynowak: Ah, I see it's below.

src/Mvc/Mvc.Core/src/Infrastructure/ControllerActionInvokerProvider.cs
```diff
@@ -16,7 +16,7 @@ namespace Microsoft.AspNetCore.Mvc.Infrastructure
     internal class ControllerActionInvokerProvider : IActionInvokerProvider
     {
         private readonly ControllerActionInvokerCache _controllerActionInvokerCache;
-        private readonly IReadOnlyList<IValueProviderFactory> _valueProviderFactories;
+        private readonly IList<IValueProviderFactory> _valueProviderFactories;
```
rynowak: uhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhhh

src/Mvc/Mvc.Core/src/Infrastructure/CopyOnWriteList.cs
```diff
@@ -8,15 +8,15 @@ namespace Microsoft.AspNetCore.Mvc.Infrastructure
...
         private List<T> _copy;
 
-        public CopyOnWriteList(IReadOnlyList<T> source)
+        public CopyOnWriteList(IList<T> source)
         {
             _source = source;
         }
 
-        private IReadOnlyList<T> Readable => _copy ?? _source;
+        private IList<T> Readable => _copy ?? _source;
```
rynowak: I'm not super a fan of this.

src/Mvc/Mvc.Cors/src/CorsApplicationModelProvider.cs
rynowak: I'm not sure what you mean *go in endpoint routing*. what would that look like?

src/Mvc/Mvc.Cors/src/CorsAuthorizationFilter.cs
rynowak: I'm not too concerned about changing 200 to 204, consistency is good.

rynowak: I looked and the spec says any 2XX is legal.

src/Mvc/Mvc.Cors/src/CorsHttpMethodActionConstraint.cs
rynowak: Cool 👍 Always good to remove teh uglies

src/Mvc/Mvc.RazorPages/src/ApplicationModels/CompiledPageActionDescriptorBuilder.cs
rynowak: Did we do this before? I really don't want to make attributes on handler methods *just work* since they don't work on legacy routing and legacy routing is still supported.

src/Mvc/Mvc.RazorPages/src/ApplicationModels/PageApplicationModel.cs
```diff
@@ -154,5 +156,10 @@ public PageApplicationModel(PageApplicationModel other)
         /// Gets the sequence of <see cref="PagePropertyModel"/> instances on <see cref="PageHandlerModel"/>.
         /// </summary>
         public IList<PagePropertyModel> HandlerProperties { get; }
+
+        /// <summary>
+        /// Gets the endpoint metadata for this action.
+        /// </summary>
+        public IList<object> EndpointMetadata { get; }
```
rynowak: Adding this because there wasn't a way to do this before? I thought we had this on pageroutemodel?

src/Mvc/Mvc.RazorPages/src/ApplicationModels/PageApplicationModel.cs
```diff
@@ -47,6 +48,7 @@ public class PageApplicationModel
             HandlerMethods = new List<PageHandlerModel>();
             HandlerProperties = new List<PagePropertyModel>();
             HandlerTypeAttributes = handlerAttributes;
+            EndpointMetadata = new CopyOnWriteList<object>(ActionDescriptor.EndpointMetadata ?? Array.Empty<object>());
```
rynowak: I don't think this is a good enough reason to change the contract of `CopyOnWriteList`. We use that in a few perf-sensitive places that run every request. This happens once per page.

src/Mvc/Mvc.RazorPages/src/ApplicationModels/PageConventionCollection.cs
rynowak: This is really ugly, but it might be the best thing we can do.

src/Mvc/Mvc.RazorPages/src/DependencyInjection/PageConventionCollectionExtensions.cs
```diff
@@ -378,8 +446,18 @@ public static PageConventionCollection AuthorizeAreaFolder(this PageConventionCo
...
+            {
+                if (conventions.MvcOptions.EnableEndpointRouting)
+                {
+                    model.EndpointMetadata.Add(new AuthorizeAttribute(policy));
+                }
+                else
+                {
+                    model.Filters.Add(new AuthorizeFilter(policy));
+                }
+            });
```
rynowak: The constructor thing on the collection is gross, but this works pretty much like what I expected.

src/Mvc/Mvc.RazorPages/src/ApplicationModels/PageConventionCollection.cs
```diff
@@ -14,6 +14,7 @@ public class PageConventionCollection : Collection<IPageConvention>
         /// Initializes a new instance of the <see cref="PageConventionCollection"/> class that is empty.
         /// </summary>
         public PageConventionCollection()
+            : this((MvcOptions)null)
```
rynowak: Would it make things better/easier if this was `IServiceProvider` instead? I'm worried about cycles in options, not sure if `IServiceProvider` makes that better or now.

src/Mvc/Mvc.RazorPages/src/Infrastructure/PageActionInvokerProvider.cs
```diff
@@ -27,7 +26,7 @@ internal class PageActionInvokerProvider : IActionInvokerProvider
         private readonly IRazorPageFactoryProvider _razorPageFactoryProvider;
         private readonly IActionDescriptorCollectionProvider _collectionProvider;
         private readonly IFilterProvider[] _filterProviders;
-        private readonly IReadOnlyList<IValueProviderFactory> _valueProviderFactories;
+        private readonly IList<IValueProviderFactory> _valueProviderFactories;
```
rynowak: 👎

src/Mvc/Mvc.RazorPages/src/ApplicationModels/CompiledPageActionDescriptorBuilder.cs
rynowak: Ignore me. This says handler type.

src/Mvc/test/Mvc.FunctionalTests/GlobalAuthorizationFilterEndpointRoutingTest.cs
```diff
@@ -21,11 +20,5 @@ public GlobalAuthorizationFilterEndpointRoutingTest(MvcTestFixture<SecurityWebSi
             builder.UseStartup<SecurityWebSite.StartupWithGlobalDenyAnonymousFilter>();
 
         public WebApplicationFactory<SecurityWebSite.StartupWithGlobalDenyAnonymousFilter> Factory { get; }
-
-        [Fact(Skip = "https://github.com/aspnet/AspNetCore/issues/8387")]
```
rynowak: Worth searching for #8387 - there may be more of these *L U R K I N G*

rynowak (commented):
This looks a lot like what I expected!

There are a few really ugly things here, but that's not really something we can avoid.

src/Mvc/Mvc.Core/src/Authorization/AuthorizeFilter.cs
HaoK: On the bright side, there might have been some subtle bugs with policy caching + combining of policies that were in 2.2 (but appear to have been accidentally fixed in 3.0) i.e. https://github.com/aspnet/AspNetCore/issues/7687

So its slower, but safer

src/Mvc/Mvc.Core/test/Authorization/AuthorizeFilterTest.cs
```diff
@@ -527,6 +512,89 @@ public void CreateInstance_ReturnsNewFilterIfPolicyAndPolicyProviderAreNotSet(Au
...
+                .AddSingleton<IAuthorizationPolicyProvider>(policyProvider)
+                .BuildServiceProvider();
+
+            ActionContext.HttpContext.SetEndpoint(new Endpoint(
+                _ => null,
+                new EndpointMetadataCollection(new AuthorizeAttribute("policy2")),
+                "test"));
+            var context = new AuthorizationFilterContext(ActionContext, new[] { filter, });
+
+
```
HaoK: nit: extra line

src/Mvc/Mvc.Core/src/ApplicationModels/AuthorizationApplicationModelProvider.cs
Tratcher: If ever there were a place you wanted to confirm that the auth middlewere was added...

src/Mvc/Mvc.RazorPages/src/DependencyInjection/PageConventionCollectionExtensions.cs
```diff
@@ -313,13 +371,23 @@ public static PageConventionCollection AuthorizeFolder(this PageConventionCollec
                 throw new ArgumentException(Resources.ArgumentCannotBeNullOrEmpty, nameof(folderPath));
             }
 
-            var authorizeFilter = new AuthorizeFilter(policy);
-            conventions.AddFolderApplicationModelConvention(folderPath, model => model.Filters.Add(authorizeFilter));
+            var authorizeAttribute = new AuthorizeAttribute(policy);
```
HaoK: is this attribute instance used anywhere?  the endpoint metadata is newing another one up below

src/Mvc/Mvc.RazorPages/src/DependencyInjection/PageConventionCollectionExtensions.cs
```diff
@@ -289,13 +337,23 @@ public static PageConventionCollection AuthorizeAreaPage(this PageConventionColl
                 throw new ArgumentException(Resources.ArgumentCannotBeNullOrEmpty, nameof(pageName));
             }
 
-            var authorizeFilter = new AuthorizeFilter(policy);
-            conventions.AddAreaPageApplicationModelConvention(areaName, pageName, model => model.Filters.Add(authorizeFilter));
+            var authorizeAttribute = new AuthorizeAttribute(policy);
```
HaoK: is this attribute instance used anywhere? the endpoint metadata is newing another one up below

src/Mvc/Mvc.RazorPages/src/DependencyInjection/PageConventionCollectionExtensions.cs
```diff
@@ -378,8 +446,18 @@ public static PageConventionCollection AuthorizeAreaFolder(this PageConventionCo
                 throw new ArgumentException(Resources.ArgumentCannotBeNullOrEmpty, nameof(folderPath));
             }
 
-            var authorizeFilter = new AuthorizeFilter(policy);
-            conventions.AddAreaFolderApplicationModelConvention(areaName, folderPath, model => model.Filters.Add(authorizeFilter));
+            var authorizeAttribute = new AuthorizeAttribute(policy);
```
HaoK: is this attribute instance used anywhere? the endpoint metadata is newing another one up below

src/Mvc/test/Mvc.FunctionalTests/AuthMiddlewareAndFilterTestBase.cs
```diff
@@ -0,0 +1,174 @@
...
+
+        [Fact]
+        public Task GlobalAuthFilter_CombinesWithAuthAttributeSpecifiedViaConvention()
+        {
+            // Arrange
+            var page = "AuthorizePageViaConvention";
+
+            return LoginAB(page);
+        }
+
```
HaoK: Historically the place we see the most bugs is when there's a action that is trying to mix cookies and bearer, if its not too hard, do you think you can add one or two tests for that scenario here?

Basically something like:

```C#
[Authorize] // requires any authenticated user (aka the application cookie typically)
public class Controller {
   [Authorize(AuthenticationScheme = "Bearer")]
   public void Api();

   public void Cookie();

   [AllowAnonymous]
   public void PartyOn();
```

And ideally there would be a few tests that calls these actions at least twice in a few different orders (i.e. anon, bearer, cookie vs cookie, anon, bearer, etc) to make sure that we catch something like https://github.com/aspnet/AspNetCore/issues/7687

src/Mvc/Mvc.Core/src/ApplicationModels/ActionAttributeRouteModel.cs
pranavkm: `SelectorModel selector` is the one from `ActionModel` `IList<object> metadata` is the controller metadata. We're combining the two. We used to append `metadata` to the `selector.EndpointMetadata`. This would mean you got `SelectorModel.EndpointMetadata = { ActionMetadata, ControllerMetadata }` prior to this change at the end of this call. This change makes it so that you get `SelectorModel.EndpointMetadata = {  ControllerMetadata, ActionMetadata }`

see https://github.com/aspnet/AspNetCore/pull/9099/files#diff-e47afec1f608e7785f1bd3b4e4c6427eL277 for an example of this

src/Mvc/Mvc.Cors/src/CorsApplicationModelProvider.cs
pranavkm: One of

1) Have `EndpointMetadata` complain if you have cors metadata but do not have `HttpMethodData.AllowCors` set when constructing the `Endpoint`
2) Have the builder update the `HttpMethodData` when it sees cors attributes
3) Remove `HttpMethodData.AllowCors` and use the presence of cors metadata to do CORS. Particularly feasible if `ICorsMetadata` is in abstractions.

The problem I have with this behavior now is that it's not enough to say here's CORS stuff for you endpoint, you have to explicitly update another metadata for it to work.

src/Middleware/CORS/src/Infrastructure/CorsMiddleware.cs
pranavkm: We don't have any tests outside of MVC that use `IDisableCorsAttribute`. It's only meaningful (not really) when you `AllowCors` on controllers and turn it off for an action. Prior to this change, the filter would have done the needful, at least with MVC.

src/Mvc/Mvc.Core/src/Infrastructure/ControllerActionInvokerProvider.cs
pranavkm: Super perf! Fine, I'll change these.

src/Mvc/Mvc.RazorPages/src/ApplicationModels/PageApplicationModel.cs
pranavkm: Yes, we have `EndpointMetadata` on `PageRouteModel`. However, here's how things work with Razor Pages:

`PageRouteModel` -> `PageActionDescriptor` -> `PageApplicationModel` -> `Apply conventions` -> `CompiledPageActionDescriptor`

When going from `PageRouteModel` to `PageActionDescriptor`: we've already turned individual selector models in to individual `PageActionDescriptor` instances. Secondly, we cannot expose selector model on `PageApplicationModel` since it's too late to influence the route, but you do get to influence route metadata.

src/Mvc/Mvc.RazorPages/src/ApplicationModels/PageConventionCollection.cs
pranavkm: We could lazy load the options when you do a convention. Conventions get applied much later during execution, so things are definitely wired up by then

src/Mvc/Mvc.Core/src/ApplicationModels/ActionAttributeRouteModel.cs
rynowak: Reversing a list is not the correct solution. It means that some part of the system is using the wrong representation and possibly exposing it to extensibility.

src/Mvc/Mvc.Cors/src/CorsApplicationModelProvider.cs
rynowak: I'll log and issue for this.

rynowak: https://github.com/aspnet/AspNetCore/issues/9127

src/Mvc/Mvc.Core/src/ApplicationModels/AuthorizationApplicationModelProvider.cs
rynowak: Right, this could work for MVC controllers, not for pages. It doesn't help with the case also where you put auth middleware somewhere crazy.

src/Mvc/Mvc.RazorPages/src/ApplicationModels/PageApplicationModel.cs
rynowak: okie dokie.

pranavkm (commented):
:shipit:

rynowak (approved):

Issue #8898, opened 2019-03-28
Create an analyzer for Endpoint Routing + UseMVC


rynowak:
In 3.0 we won't support this combo anymore, since we expect this to impact lots of folks updating to 3.0 this definitely deserves an analyzers so users can know.

The conditions I would expect to check for is:
- Is `MvcOptions.EndableEndpointRouting == false` not found?
- Is `UseMvc()` present

rynowak:
https://github.com/aspnet/AspNetCore/tree/rynowak/startup-analyzer

JamesNK:
Questions:

1. Does CompatibilityVersion.Version_2_1 still disable endpoint routing? From what I can tell it does not.
2. How funky do we want to get with detecting the correct Startup.cs? For example, someone may have renamed it. Would require parsing `Program.Main` (which, if we want to be really thorough would require us to look in the csproj file to see if the entry point had changed).
3. I'll remove the other analysis logic that is in your branch. Can look at in the future. Cool?

rynowak:
1. We don't need to handle that. These analyzers are 3.0+ and 3.0 doesn't support that setting.

2. Not funky at all. We should do something really simple.

3. Yes please, we should do something reasonable to try and make the migration experience have better feedback.

JamesNK:
Help file article - https://github.com/aspnet/AspNetCore/issues/8898

PR #9100, opened 2019-04-05
Add StartupAnalzyer

JamesNK:
Addresses #8898


JamesNK:
I have:

* Modified the registration callbacks so validation executes after analysis
* Added a filter for the Startup.cs file
* Added more tests
* Removed order and service analysis/validation

TODO: Actually test in VS 😬

src/Mvc/Mvc.Analyzers/src/Startup/MvcOptionsAnalysis.cs
```diff
@@ -0,0 +1,44 @@
...
+            var analysis = new MvcOptionsAnalysis((IMethodSymbol)context.OwningSymbol);
+
+            context.RegisterOperationAction(context =>
+            {
+                if (context.Operation is ISimpleAssignmentOperation operation &&
+                    operation.Value.ConstantValue.HasValue &&
+                    operation.Target is IPropertyReferenceOperation property &&
+                    property.Member.Name == "EnableEndpointRouting")
+                {
+                    analysis.EndpointRoutingEnabled = operation.Value.ConstantValue.Value as bool?;
```
pranavkm: Should this be a `bool` cast? You couldn't assign it anything else that would compile

src/Mvc/Mvc.Analyzers/src/Startup/ServicesAnalysis.cs
```diff
@@ -0,0 +1,52 @@
...
+
+            var services = ImmutableArray.CreateBuilder<ServicesItem>();
+            context.RegisterOperationAction(context =>
+            {
+                // We're looking for usage of extension methods, so we need to look at the 'this' parameter
+                // rather than invocation.Instance.
+                if (context.Operation is IInvocationOperation invocation &&
+                    invocation.Instance == null &&
+                    invocation.Arguments.Length >= 1 &&
+                    invocation.Arguments[0].Parameter.Type.Name == "IServiceCollection")
```
pranavkm: In terms of error checking, do we need to verify `Parameter`, `Type` are defined? For instance if the analyzer is invoked when the code is incomplete or when symbols are missing, would this null ref?

src/Mvc/Mvc.Analyzers/src/Startup/ServicesAnalysis.cs
```diff
@@ -0,0 +1,52 @@
...
+
+            return analysis;
+        }
+
+        public ServicesAnalysis(IMethodSymbol configureServicesMethod) 
+            : base(configureServicesMethod)
+        {
+        }
+
+        public ImmutableArray<ServicesItem> Services { get; private set; } = ImmutableArray<ServicesItem>.Empty;
```
pranavkm: Any reason to prefer this to a regular list?

src/Mvc/Mvc.Analyzers/src/Startup/StartupAnalyzer.Diagnostics.cs
```diff
@@ -0,0 +1,20 @@
...
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Microsoft.AspNetCore.Analyzers
+{
+    public partial class StartupAnalzyer : DiagnosticAnalyzer
+    {
+        internal readonly static DiagnosticDescriptor UnsupportedUseMvcWithEndpointRouting = new DiagnosticDescriptor(
+            "MVC1005", 
+            "Cannot use UseMvc with Endpoint Routing.", 
+            "Using '{0}' to configure MVC is not supported while using Endpoint Routing. To continue using '{0}', please set 'MvcOptions.EnableEndpointRounting = false' inside '{1}'.",
```
pranavkm: Should this guide the user to use `UseEndpointRouting` instead? All migrating users would see this error and we want to steer them towards endpoint routing. 
Additionally there's a way to do help link with a diagnostic. We should do a placeholder aka.ms link and point it to a migration guide

src/Mvc/Mvc.Analyzers/src/Startup/StartupAnalyzer.Events.cs
```diff
@@ -0,0 +1,33 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace Microsoft.AspNetCore.Analyzers
+{
+    public partial class StartupAnalzyer : DiagnosticAnalyzer
```
pranavkm: `internal` to keep with the theme?

src/Mvc/Mvc.Analyzers/src/Startup/StartupAnalyzer.Events.cs
```diff
@@ -0,0 +1,33 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
```
pranavkm: Super unrelated, we should consider renaming the project. The name's pretty immaterial to users now that they never see it, but we'd likely use it for entirely non-MVC scenarios now

src/Mvc/Mvc.Analyzers/src/Startup/StartupAnalzyer.cs
```diff
@@ -0,0 +1,134 @@
...
+        {
+            SupportedDiagnostics = ImmutableArray.Create<DiagnosticDescriptor>(new[]
+            {
+                UnsupportedUseMvcWithEndpointRouting,
+            });
+
+            // By default just analyze file named Startup.cs in the root directory
+            // Analzyer only runs for C# so limiting to *.cs file is fine
+            // Can be overriden for unit testing other file names
+            StartupFilePredicate = path => string.Equals(path, "Startup.cs", StringComparison.OrdinalIgnoreCase);
```
pranavkm: Interesting

pranavkm (commented):
Looks pretty neat

src/Mvc/Mvc.Analyzers/src/Startup/MvcOptionsAnalysis.cs
rynowak: It's an analzyer so I wrote this to be extra super duper safe. For instance what happens if the user put something in that doesn't actually compile? I think the analyzer would still see it.

src/Mvc/Mvc.Analyzers/src/Startup/ServicesAnalysis.cs
rynowak: That's a great question.

src/Mvc/Mvc.Analyzers/src/Startup/ServicesAnalysis.cs
rynowak: I want it to be immutable in general. Most of the roslyn code uses the immutable types so I'm following the style.

src/Mvc/Mvc.Analyzers/src/Startup/StartupAnalyzer.Diagnostics.cs
rynowak: Yes, we should do that.

I think we should possibly include both?

src/Mvc/Mvc.Analyzers/src/Startup/StartupAnalyzer.Diagnostics.cs
```diff
@@ -0,0 +1,20 @@
...
+            "Usage", 
+            DiagnosticSeverity.Warning,
```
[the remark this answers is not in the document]
rynowak: I could go either way. I lean towards warning because we're going to throw anyway.

src/Mvc/Mvc.Analyzers/src/Startup/StartupAnalyzer.Diagnostics.cs
```diff
@@ -0,0 +1,20 @@
...
+    public partial class StartupAnalzyer : DiagnosticAnalyzer
+    {
+        internal readonly static DiagnosticDescriptor UnsupportedUseMvcWithEndpointRouting = new DiagnosticDescriptor(
+            "MVC1005", 
+            "Cannot use UseMvc with Endpoint Routing.", 
+            "Using '{0}' to configure MVC is not supported while using Endpoint Routing. To continue using '{0}', please set 'MvcOptions.EnableEndpointRounting = false' inside '{1}'.",
+            "Usage", 
+            DiagnosticSeverity.Warning, 
+            isEnabledByDefault: true,
+            helpLinkUri: null); // TODO docs link.
```
rynowak: See, I was all over that aka.ms

src/Mvc/Mvc.Analyzers/src/Startup/StartupAnalyzer.Events.cs
rynowak: I'm in favour of us shipping one analyzer for ASP.NET Core and one for Microsoft.Extensions. 

I don't want to make this part of james work.

src/Mvc/Mvc.Analyzers/src/Startup/StartupAnalyzer.Events.cs
rynowak: If that's allowed sure.

src/Mvc/Mvc.Analyzers/src/Startup/StartupAnalyzer.Diagnostics.cs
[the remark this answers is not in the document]
JamesNK: I think it should be a warning at most. This will never be 100% perfect. A dev might have set the flag in a way that we don't detect.

A warning generally won't block someone, and if they take some time to clean up any warnings they have they can suppress it then.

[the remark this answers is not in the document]
rynowak: The answer won't ever be incredibly cut and dry btw. You can configure analyzers so, you can actually change the level of any error, in addition to suppressing the warning.

src/Mvc/Mvc.Analyzers/src/Startup/StartupAnalyzer.Events.cs
JamesNK: Other analyzers in this project are internal. Leaving public for consistency.

src/Mvc/Mvc.Analyzers/src/Startup/StartupAnalyzer.Diagnostics.cs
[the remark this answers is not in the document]
pranavkm: I'm not super sure it this applies to user created analyzers, but errors produced by Roslyn's stylistic code fixes (such as use throws expression instead of if checks) show up as errors in the error window, but don't actually fail the build.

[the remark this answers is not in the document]
rynowak: The stylistic ones are **definitely** not errors, they are `Info`. https://github.com/dotnet/roslyn/blob/master/src/Compilers/Core/Portable/Diagnostic/DiagnosticSeverity.cs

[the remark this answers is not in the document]
rynowak: My opinon btw is the warning is the best value for what we intend.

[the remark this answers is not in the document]
JamesNK: +1

rynowak (approved):

JamesNK:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=145222

aspnet-hello:

I've triaged the above build.

PR #9101, opened 2019-04-05
Marking a good number of tests as flaky.

jkotalik:
Trying to figure out what kestrel test is hanging. Using binary elimination as we have no access to logs easily.


dougbu (commented):
What's the plan here?
- Will the deleted tests come back once the retries have helped figure the issue out?
- Or, are you not going to check this in i.e. just run the PR validation pipelines until you're done?

Tratcher:
We came up with an idea to put a long running test warning timer directly into our custom test runner.

Or failing that you could turn the xunit diagnostic logs on at least temporarily. They're very noisy but good for finding hangs.

jkotalik:
> Or, are you not going to check this in i.e. just run the PR validation pipelines until you're done?

This.

jkotalik:
> Or failing that you could turn the xunit diagnostic logs on at least temporarily. They're very noisy but good for finding hangs.

Sure. I think I can enable it here: https://github.com/aspnet/AspNetCore/blob/3cd5054eb5d8851b25881cd05cccb931aaa022e6/src/Servers/Kestrel/xunit.runner.json

src/Servers/Kestrel/xunit.runner.json
```diff
@@ -3,5 +3,6 @@
   "appDomain": "denied",
   "methodDisplay": "method",
   "longRunningTestSeconds": 60,
-  "maxParallelThreads": -1
+  "maxParallelThreads": -1,
+  "diagnosticMessages": true
```
anurse: I totally forgot about this! Yeah if you don't turn this on it doesn't actually *report* the long running tests... so annoying.

jkotalik: I don't think this is actually being picked up by xunit 😢

jkotalik:
Found it! ContentLengthReadAsyncSingleBytesAtATime is our offender.

jkotalik:
This has been updated. I'd like to get this in soon to start unblock more PR builds.

src/Servers/Kestrel/test/InMemory.FunctionalTests/Http2/Http2TimeoutTests.cs
```diff
@@ -492,7 +492,7 @@ public async Task DATA_Sent_TooSlowlyDueToOutputFlowControlOnLargeWrite_AbortsCo
             _mockConnectionContext.VerifyNoOtherCalls();
         }
 
-        [Fact]
+        [Fact(Skip = "https://github.com/aspnet/AspNetCore-Internal/issues/2197")]
```
jkotalik: This test needs to be skipped entirely because it causes a test hang.

ryanbrandenburg (approved):
Good finds.

src/Servers/Kestrel/xunit.runner.json
jkotalik: Also going to leave on diagnostic messages for now (to help us find more hangs if they occur). Will disable eventually 😄

PR #9120, opened 2019-04-05
Disable IISExpress tests

jkotalik:
Trying to unblock PR checks.

### DIFF
--- a/src/Servers/test/FunctionalTests/ResponseCompressionTests.cs
+++ b/src/Servers/test/FunctionalTests/ResponseCompressionTests.cs
@@ -32,7 +32,7 @@ namespace ServerComparison.FunctionalTests
         }
 
         public static TestMatrix NoCompressionTestVariants
-            => TestMatrix.ForServers(ServerType.IISExpress, ServerType.Kestrel, ServerType.Nginx,  ServerType.HttpSys)
+            => TestMatrix.ForServers(ServerType.Kestrel, ServerType.Nginx,  ServerType.HttpSys)
                 .WithTfms(Tfm.NetCoreApp30)
                 .WithAllHostingModels();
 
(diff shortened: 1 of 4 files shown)


src/Servers/test/FunctionalTests/ResponseCompressionTests.cs
```diff
@@ -32,7 +32,7 @@ public ResponseCompressionTests(ITestOutputHelper output) : base(output)
         }
 
         public static TestMatrix NoCompressionTestVariants
-            => TestMatrix.ForServers(ServerType.IISExpress, ServerType.Kestrel, ServerType.Nginx,  ServerType.HttpSys)
+            => TestMatrix.ForServers(ServerType.Kestrel, ServerType.Nginx,  ServerType.HttpSys)
```
Tratcher: Only use comments so it's more obvious it needs to be reverted.
```suggestion
            => TestMatrix.ForServers(/*ServerType.IISExpress, */ ServerType.Kestrel, ServerType.Nginx,  ServerType.HttpSys)
```

dougbu (approved):
It sounds like the AzDO change is going to be rolled back, perhaps quicker than we can be sure this resolves the issue. Suggest we might not need this change.

src/Servers/test/FunctionalTests/ResponseCompressionTests.cs
jkotalik: I don't think it matters. We are just going to do a straight revert on this no matter what.

dougbu:
@jkotalik suggest closing this in favour of #9122

jkotalik:
I'm going to let both of these run to completion.

jkotalik:
Yeah there are a few more tests that failed too.

PR #9122, opened 2019-04-05
Disable IISExpress tests

jkotalik:


dougbu (approved):
Wow, it worked even though https://docs.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-build-tools?view=vs-2019 says the component doesn't exist!!

Or, have the agents been rolled back already?

src/DefaultBuilder/test/Microsoft.AspNetCore.FunctionalTests/WebHostFunctionalTests.cs
```diff
@@ -151,7 +151,7 @@ public void LoggingConfigurationSectionPassedToLoggerByDefault()
             }
         }
 
-        [ConditionalFact]
+        [ConditionalFact(Skip = "https://github.com/aspnet/AspNetCore-Internal/issues/2170")]
         [OSSkipCondition(OperatingSystems.Linux | OperatingSystems.MacOSX)]
```
dougbu: Minor but this is really `[Fact(Skip = "https://github.com/aspnet/AspNetCore-Internal/issues/2170")]` and there's no need for the `[OSSkipCondition(OperatingSystems.Linux | OperatingSystems.MacOSX)]`.

eng/scripts/vs.buildtools.json
```diff
@@ -25,6 +25,7 @@
         "Microsoft.VisualStudio.Workload.NetCoreBuildTools",
         "Microsoft.VisualStudio.Workload.VCTools",
         "Microsoft.VisualStudio.Workload.VisualStudioExtensionBuildTools",
-        "Microsoft.VisualStudio.Workload.WebBuildTools"
+        "Microsoft.VisualStudio.Workload.WebBuildTools",
+        "Microsoft.VisualStudio.Component.IISExpress"
```
dougbu: Why is this still changing? I would have thought #9120 was the way to go since this change didn't work.

dougbu (approved):
If this works, great. But, why is vs.buildtools.json still changing?

PR #9131, opened 2019-04-05
Safe utf8 string conversion

benaadams:
Saw this, thought @blowdart might like it


GrabYourPitchforks:
Separately: We should also consider whether we want to block ill-formed UTF-8 from coming into the system at all. The default behavior of `Encoding.UTF8` is that it replaces invalid sequences with `\uFFFD`. (And this behavior is changing ever so slightly in .NET Core 3.0, so 2.x and 3.0 will potentially produce different numbers of `\uFFFD` chars given invalid input.)

benaadams:
Could use a throwing fallback?

GrabYourPitchforks:
I don't know if it would be appropriate to use a throwing fallback here as part of this PR. The web server framework probably isn't expecting these particular calls to throw right now. It's just something I wanted to throw out for future consideration.

blowdart:
Someone has hacked @benaadams's github account. We need to rollback everything from at least the last week.

JamesNK (approved):

davidfowl:
LGTM. This is hold over from .Net fx support

davidfowl (approved):

PR #9149, opened 2019-04-06
Dispose configuration in WebHost

poke:
This is a companion-PR to aspnet/Extensions#1361 which addresses aspnet/Extensions#938: When disposing the WebHost, the underlying configurations will also be disposed by the service provider.

~Since the `IConfiguration` that is passed to the `WebHost` constructor is the *host* configuration, and does not access the configuration directly (the `Startup` may though), the `WebHost` has to resolve the app configuration from the service provider. This is done early within `Initialize` as soon as the service provider is available to avoid having to retrieve it later only for disposal.~

The host configuration is also disposed ~by the `WebHost`~ transitively. While the host configuration only ever contains a single environment variables configuration provider (which is not disposable itself), disposing the configuration will also make sure to dispose the change registration.

The drawback for this is that using the `WebHostBuilder.GetSetting()` *after* the web host is disposed means that the operation is made on an already disposed configuration root. The way the configuration root is currently implemented, this should not actually prevent this from working though.

### DIFF
--- a/src/Hosting/Hosting/src/WebHostBuilder.cs
+++ b/src/Hosting/Hosting/src/WebHostBuilder.cs
@@ -175,6 +175,10 @@ namespace Microsoft.AspNetCore.Hosting
             {
                 host.Initialize();
 
+                // resolve configuration explicitly once to mark it as resolved within the
+                // service provider, ensuring it will be properly disposed with the provider
+                _ = host.Services.GetService<IConfiguration>();
+
                 var logger = host.Services.GetRequiredService<ILogger<WebHost>>();
 
                 // Warn about duplicate HostingStartupAssemblies
@@ -264,12 +268,13 @@ namespace Microsoft.AspNetCore.Hosting
 
             var builder = new ConfigurationBuilder()
                 .SetBasePath(_hostingEnvironment.ContentRootPath)
-                .AddConfiguration(_config);
+                .AddConfiguration(_config, shouldDisposeConfiguration: true);
 
             _configureAppConfigurationBuilder?.Invoke(_context, builder);
 
             var configuration = builder.Build();
-            services.AddSingleton<IConfiguration>(configuration);
+            // register configuration as factory to make it dispose with the service provider
+            services.AddSingleton<IConfiguration>(_ => configuration);
             _context.Configuration = configuration;
 
             var listener = new DiagnosticListener("Microsoft.AspNetCore");
--- a/src/Hosting/Hosting/test/WebHostTests.cs
+++ b/src/Hosting/Hosting/test/WebHostTests.cs
@@ -975,6 +975,54 @@ namespace Microsoft.AspNetCore.Hosting
             Assert.Contains("ConfigureServices", ex.Message);
         }
 
+        [Fact]
+        public void Dispose_DisposesAppConfiguration()
+        {
+            var providerMock = new Mock<ConfigurationProvider>().As<IDisposable>();
+            providerMock.Setup(d => d.Dispose());
+
+            var sourceMock = new Mock<IConfigurationSource>();
+            sourceMock.Setup(s => s.Build(It.IsAny<IConfigurationBuilder>()))
+                .Returns((ConfigurationProvider)providerMock.Object);
+
+            var host = CreateBuilder()
+                .ConfigureAppConfiguration(configuration =>
+                {
+                    configuration.Add(sourceMock.Object);
+                })
+                .Build();
+
+            providerMock.Verify(c => c.Dispose(), Times.Never);
+
+            host.Dispose();
+
+            providerMock.Verify(c => c.Dispose(), Times.AtLeastOnce());
+        }
+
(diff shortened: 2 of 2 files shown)


poke:
The disposal of the host configuration depends on the changes to `ChainedConfigurationProvider` in aspnet/Extensions#1361, and is not covered by tests (because there’s no proper access to it).

src/Hosting/Hosting/src/WebHostBuilder.cs
```diff
@@ -269,7 +269,8 @@ private IServiceCollection BuildCommonServices(out AggregateException hostingSta
             _configureAppConfigurationBuilder?.Invoke(_context, builder);
 
             var configuration = builder.Build();
-            services.AddSingleton<IConfiguration>(configuration);
+            // register configuration as factory to make it dispose with the service provider
+            services.AddSingleton<IConfiguration>(_ => configuration);
```
Tratcher: Also needs to set shouldDisposeConfiguration https://github.com/aspnet/AspNetCore/blob/b8317c9f768460b509e95925a65d3ab78feba779/src/Hosting/Hosting/src/WebHostBuilder.cs#L267

src/Hosting/Hosting/src/Internal/WebHost.cs
```diff
@@ -129,6 +129,10 @@ public void Initialize()
 
                 _applicationServicesException = ExceptionDispatchInfo.Capture(ex);
             }
+
+            // resolve configuration explicitly to mark it as resolved within the service
+            // provider, ensuring it will be properly disposed with the provider
+            _ = _applicationServices.GetService<IConfiguration>();
```
Tratcher: This could still happen in WebHostBuilder just after calling Initialize.
https://github.com/aspnet/AspNetCore/blob/b8317c9f768460b509e95925a65d3ab78feba779/src/Hosting/Hosting/src/WebHostBuilder.cs#L178

src/Hosting/Hosting/src/WebHostBuilder.cs
Tratcher: Oh, right, cross repo change..

poke: Good that you mention it though. I probably would have forgotten it after all.

For the repo changes, I have to wait until the maestro bot detects the changes and makes a PR, right?

poke: I honestly don’t know how that would work.. I’ve looked at the file that contains all the package version numbers but it says that it should only be updated automatically using “Darc”. But the Darc readme basically tells me that I won’t be able to use it… 😅 

Guess I’ll just wait for the dependencies to be updated 🤷‍♂️

anurse: Sounds like you're in the **darc** about dependency updates.
<br />
<br />
<br />
<br />
<br />
<br />
<br />
<br />
... I'll show myself out.

poke:
Since there wasn’t any PR from the bot over the course of the day that increased the version on the M.E.Configuration package, I just went totally rogue and edited the section that _“should ONLY be updated by automation”_… manually.

That way, I can at least compile the final version of this change and hopefully it does work on the CI end too.

If we need to wait for a _proper_ update of the versions, then that’s of course fine with me too.

Tratcher:
Yeah, we're going to need to wait for the rest of Extensions deps to get updated. Pulling just one isn't a good idea. @dougbu can you notify us here after the next update?

poke:
I don’t have the tools so that’s the best I can do right now 😁 And yeah, the CI really doesn’t like this apparently 😅 

But at least I could push the latest code out, so you can look over it even if it’s not ready to merge yet.

dougbu:
>  @dougbu can you notify us here after the next update?

We should get new packages from EntityFrameworkCore and AspNetCore-Tooling every day. Those pull in the latest coherent Extensions packages. Coherency usually means Extensions build from day 1 is pulled into EF and Tooling on day 2 and those are pulled into this repo on day 3. Of course, the packages are all reving in parallel.

Which specific Extensions build are you looking for?

Tratcher:
@dougbu 3.0.0-preview5.19208.3 or later

dougbu:
:ok: 3.0.0-preview5.19208.3 Extensions packages should be in AspNetCore tomorrow morning. AspNetCore-Tooling and EntityFrameworkCore have them now. And, both have had successful official builds since taking those dependencies.

poke:
The dependencies arrived and the PR is updated! Should be good to go now 😊

Tratcher (approved):

Tratcher:
[Ongoing build outage, I'll followup]

poke:
Yeah, I know, I’m already keeping an eye on #9263 and am ready to rebase to make the CI green.

poke:
It’s green! :tada: :tada:

Tratcher:
Thanks :-)

PR #9169, opened 2019-04-08
Remove reference to Microsoft.AspNetCore.Components.Server

danroth27:
Microsoft.AspNetCore.Components.Server no longer exists and is subsumed by Microsoft.AspNetCore.Blazor.Server


pranavkm (approved):
cc @javiercn

javiercn (approved):

HaoK:
The failure happens before it gets to helix, https://dev.azure.com/dnceng/public/_build/results?buildId=146976 shows that its dying trying to Install build tools for VS 2019 so maybe a temporary/flaky thing (retry?)

PR #9174, opened 2019-04-08
Stabilize StreamAsyncCanBeCanceledThroughGetAsyncEnumerator test

mikaelm12:
Fixes: https://github.com/aspnet/AspNetCore-Internal/issues/2196
We were seeing some flakiness on some of the PR builds for this test. 

I updated this test to use a strategy that we have in a similar test for our channel streaming API
https://github.com/aspnet/AspNetCore/blob/8bfbc4a2ac8f32e61a9d6eee8c55901771478428/src/SignalR/clients/csharp/Client/test/FunctionalTests/HubConnectionTests.cs#L728-L742

Instead of only expected 1 item to have been written to the stream just enforce that we at least gotten one item and that less than the total amount, in this case 1000 items, were written because of the token getting triggered.

### DIFF
--- a/src/SignalR/clients/csharp/Client/test/FunctionalTests/HubConnectionTests.cs
+++ b/src/SignalR/clients/csharp/Client/test/FunctionalTests/HubConnectionTests.cs
@@ -449,7 +449,7 @@ namespace Microsoft.AspNetCore.SignalR.Client.FunctionalTests
 
                     var cts = new CancellationTokenSource();
 
-                    var stream = connection.StreamAsync<int>("Stream", 5, cts.Token);
+                    var stream = connection.StreamAsync<int>("Stream", 1000, cts.Token);
                     var results = new List<int>();
 
                     var enumerator = stream.GetAsyncEnumerator();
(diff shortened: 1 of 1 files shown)


src/SignalR/clients/csharp/Client/test/FunctionalTests/HubConnectionTests.cs
```diff
@@ -689,8 +689,7 @@ public async Task StreamAsyncCanBeCanceledThroughGetEnumerator(string protocolNa
                         }
                     });
 
-                    Assert.Single(results);
-                    Assert.Equal(0, results[0]);
+                    Assert.True(results.Count > 0 && results.Count < 1000);
```
halter73: Do you know why this is needed here an not in StreamAsyncCanBeCanceled?

halter73 (approved):
It looks like you just missed the other test it with this change.

I see the CanCloseStreamMethodEarly method now. Just make sure you get StreamAsyncCanBeCanceled too, because that's gotta be at least as flaky.

src/SignalR/clients/csharp/Client/test/FunctionalTests/HubConnectionTests.cs
mikaelm12: It should be there as well. I'll add it

BrennanConroy (approved):

PR #9175, opened 2019-04-08
Fix folder version for global module to only have 3 numbers

jkotalik:
### Description

Changes the default folder path for the AspNetCoreModule out of process request handler from a folder path of MAJOR.MINOR.PATCH.BUILD to MAJOR.MINOR.PATCH.

### Reason for change:

Without this change, ANCM+OutOfProcess will not work, giving a failure saying the out of process request handler cannot be loaded.

### What commit caused this failure

We recently took in a new dependency from core-setup to find the folder hostfxr is located in. Hostfxr recently was updated to support sem ver 2.0 (https://semver.org/), which we needed to react to if/when they start to support it (https://github.com/aspnet/AspNetCore/commit/7a1a53d76d62783ae33f116309c5d3745f00dd94 for the commit we made). We use this type (fx_ver) to also look for the request handler for ANCM out of process. However, the folder ANCM out of process is located in follows windows folder versions (MAJOR.MINOR.PATCH.BUILD) rather than sem var. Before updating our core-setup dependency, this folder version would still be parsed and valid. However, after this change, it says it cannot find the folder. 

### Why wasn't this caught earlier?

We added many automated tests checking for folder versions that contained MAJOR.MINOR.PATCH, but during installer cleanup, we decided to simplify the use of the MSI version and Folder version into the same value. We didn't add regression tests for this scenario at all. 

### Any concern of breaking changes?

No. The reason dlls are in folders are for rapid update scenarios, which have never been used yet. MSIs today remove the folder on uninstall, so there will never be more than one folder in the directory.

### DIFF
--- a/src/Installers/Windows/AspNetCoreModule-Setup/ANCMIISExpressV2/ancm_iis_expressv2.wxs
+++ b/src/Installers/Windows/AspNetCoreModule-Setup/ANCMIISExpressV2/ancm_iis_expressv2.wxs
@@ -173,7 +173,7 @@
                 <RegistryValue Name="TypesSupported" Type="integer" Value="7"/>
               </RegistryKey>
             </Component>
-            <Directory Id="HandlerVersionDir" Name="$(var.ANCMMsiVersion)" >
+            <Directory Id="HandlerVersionDir" Name="$(var.ANCMFolderVersion)" >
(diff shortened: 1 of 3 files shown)


jkotalik:
cc @Eilon @shirhatti @DamianEdwards

anurse:
QB approved. Continue to **hold** until I confirm the branch status.

Eilon:

And please also resolve https://devdiv.visualstudio.com/DevDiv/_workitems/edit/843075 when done.

analogrelay (approved):

PR #9177, opened 2019-04-08
Rename StreamPipeReaderOptions to StreamPipeReaderAdapterOptions 

jkotalik:
### Description

Changes the name of StreamPipeReaderOptions to StreamPipeReaderAdapterOptions to avoid a name collision with System.IO.Pipelines.

### Reason for change

Builds will fail without this change, saying that there is a nameconflict on StreamPipeReaderOptions.

### What commit caused this failure

https://github.com/dotnet/corefx/commit/dc965e16c4f71163df9de013951c95916992c908 caused this name conflict and upgrading dependencies caught it.

### Why wasn't it caught earlier?

Upgrading dependencies takes too long and no one was on top of this earlier.

### Any concerns of breaking changes?

This is a breaking change, but the type was unused and will be removed within the timeframe of preview4 to preview5.

cc @Eilon @DamianEdwards


dougbu (changes requested):
Do not attempt to merge this into the preview 4 branch. We don't have coherent builds of EntityFrameworkCore or AspNetCore-Tooling that have pushed into the `.NET Core 3 Release` channel available yet. (The versions will actually go _backwards_.)

Suggest opening a PR that only changes the names of our methods to avoid conflicts. That'll work now and when we get new dependencies.

anurse:
Don't worry, we also need QB approval ;). I'll sync up with you before we attempt to merge anything to preview 4.

jkotalik:
🆙 📅

dougbu (approved):
Works for me but must wait for QB approval

anurse:
QB approved. Continue to **hold** until I confirm the branch status.

Eilon:

And please also resolve https://devdiv.visualstudio.com/DevDiv/_workitems/edit/843068 when done.

PR #9181, opened 2019-04-08
Complain if auth hasn't been set up correctly

pranavkm:
Fixes https://github.com/aspnet/AspNetCore/issues/9041


pranavkm:
Plain text benchmarks:

| Description |     RPS | CPU (%) | Memory (MB) | Avg. Latency (ms) | Startup (ms) | First Request (ms) | Latency (ms) |
| ----------- | ------- | ------- | ----------- | ----------------- | ------------ | ------------------ | ------------ |
| Baseline            | 1,042,838 |      95 |         161 |              2.73 |          318 |             114.95 |         0.47 |
| Check HttpContext.Items first| 939,324 |      94 |         158 |              2.69 |          329 |             109.38 |         0.34 |
| Check endpoint first | 1,022,092 |      95 |         158 |              3.57 |          331 |             108.98 |          0.5 |
| Also check Cors | 1,008,837 |      95 |         158 |              2.75 |          321 |             115.12 |         0.72 |
|     Do auth | 114,312 |     100 |         173 |             20.01 |          418 |             183.73 |         0.93 |

Auth one is a likely worse than normal since I was manually signing in every request.

pranavkm:
cc @davidfowl

src/Http/Routing/src/EndpointMiddleware.cs
```diff
@@ -52,6 +60,34 @@ public async Task Invoke(HttpContext httpContext)
             await _next(httpContext);
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
davidfowl: Why?

pranavkm: Absolutely no good reason.

src/Security/Authorization/Core/src/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Runtime.CompilerServices;
+using Microsoft.AspNetCore.Authorization;
+
+[assembly: TypeForwardedTo(typeof(IAuthorizeData))]
```
Tratcher: Where these interfaces used by anyone but us prior to 3.0? What for?

src/Http/Http.Abstractions/src/IAllowAnonymous.cs
```diff
@@ -4,7 +4,7 @@
 namespace Microsoft.AspNetCore.Authorization
 {
     /// <summary>
-    /// Marker interface to enable the <see cref="AllowAnonymousAttribute"/>.
+    /// Marker interface to allow anonymous.
```
rynowak: This doc comment makes me sad.

src/Http/Routing/src/EndpointMiddleware.cs
rynowak: IMO better to just inline the code.

src/Http/Routing/src/EndpointMiddleware.cs
```diff
@@ -52,6 +60,34 @@ public async Task Invoke(HttpContext httpContext)
...
+            if (endpoint.Metadata.GetMetadata<ICorsMetadata>() != null &&
+                !httpContext.Items.ContainsKey(CorsMiddlewareInvokedKey))
+            {
+                ThrowMissingCorsMiddlewareException(endpoint);
+            }
+        }
+
+        private static void ThrowMissingAuthMiddlewareException(Endpoint endpoint)
+        {
+            throw new InvalidOperationException($"Endpoint {endpoint.DisplayName} contains authorization metadata, " +
```
rynowak: This needs to include instructions about how to fix the problem. Imagine you see this error for the first time in preview 5.

src/Security/Authorization/Core/src/Properties/AssemblyInfo.cs
rynowak: By anyone who wanted to? They are public.

src/SignalR/server/SignalR/test/Startup.cs
```diff
@@ -25,11 +22,16 @@ public void ConfigureServices(IServiceCollection services)
                 options.DefaultAuthenticateScheme = CookieAuthenticationDefaults.AuthenticationScheme;
                 options.DefaultChallengeScheme = CookieAuthenticationDefaults.AuthenticationScheme;
             }).AddCookie();
+
+            services.AddAuthorizationPolicyEvaluator();
```
rynowak: It's super wierd to me that this is separate from just authorization/authentication. It's a super undiscoverable name.

src/Http/Routing/src/EndpointMiddleware.cs
```diff
@@ -35,6 +41,8 @@ public async Task Invoke(HttpContext httpContext)
             var endpoint = httpContext.Features.Get<IEndpointFeature>()?.Endpoint;
             if (endpoint?.RequestDelegate != null)
             {
+                EnsureRequisiteMiddlewares(httpContext, endpoint);
```
rynowak: This should have an opt-out on `RouteOptions`

src/SignalR/server/SignalR/test/Startup.cs
davidfowl: See https://github.com/aspnet/AspNetCore/issues/6899

src/Http/Routing/src/EndpointMiddleware.cs
```diff
@@ -11,6 +14,9 @@ namespace Microsoft.AspNetCore.Routing
 {
     internal sealed class EndpointMiddleware
     {
+        internal const string AuthorizationMiddlewareInvokedKey = "__AuthorizationMiddlewareInvoked";
```
davidfowl: I wonder if these should be features... Then we could back them by fields in Kestrel and avoid the dictionary look up altogether

pranavkm: > Is it possible to use mvc without ever initilizing Items?

MVC uses it in a couple of places (https://github.com/aspnet/AspNetCore/blob/c565386a3ed135560bc2e9017aa54a950b4e35dd/src/Mvc/Mvc.Core/src/Routing/UrlHelperFactory.cs, https://github.com/aspnet/AspNetCore/blob/c565386a3ed135560bc2e9017aa54a950b4e35dd/src/Mvc/Mvc.ViewFeatures/src/Filters/SaveTempDataFilter.cs#L34) so generally you'd see it initialized if MVC is involved.

src/SignalR/server/SignalR/test/Startup.cs
pranavkm: https://github.com/aspnet/AspNetCore/issues/9208 to track this

src/Http/Http.Abstractions/src/IAllowAnonymous.cs
pranavkm: Do you have suggestions on what we could do to make this better?

src/Http/Routing/src/EndpointMiddleware.cs
```diff
@@ -52,6 +60,34 @@ public async Task Invoke(HttpContext httpContext)
...
+                !httpContext.Items.ContainsKey(CorsMiddlewareInvokedKey))
+            {
+                ThrowMissingCorsMiddlewareException(endpoint);
+            }
+        }
+
+        private static void ThrowMissingAuthMiddlewareException(Endpoint endpoint)
+        {
+            throw new InvalidOperationException($"Endpoint {endpoint.DisplayName} contains authorization metadata, " +
+                "but a middleware was not found that supports authorization.");
```
JamesNK: We say "a middleware", but we really require middleware that sets `AuthorizationMiddlewareInvokedKey`.

Is this a problem for people who think that they can implement their own middleware to fix this problem? I think this exception could explicitly say `.UseAuthorization()` is required. It is the only middleware that knows to set `AuthorizationMiddlewareInvokedKey`

The alternative is providing a more first class way of saying that auth has run for the request.

src/Http/Routing/src/EndpointMiddleware.cs
```diff
@@ -52,6 +60,34 @@ public async Task Invoke(HttpContext httpContext)
...
+        }
+
+        private static void ThrowMissingCorsMiddlewareException(Endpoint endpoint)
+        {
+            throw new InvalidOperationException($"Endpoint {endpoint.DisplayName} contains CORS metadata, " +
+                "but a middleware was not found that supports CORS.");
```
JamesNK: Same issue as above. Only `.UseCors()` knows to set the right items key.

src/Http/Routing/src/EndpointMiddleware.cs
pranavkm: The error message is a bit more explicit about what to do resolve it and includes a line saying you could add auth. 

> The alternative is providing a more first class way of saying that auth has run for the request.

We could add a gesture for this or turn it in to a feature with settable properties. Either of these seem viable to me

src/SignalR/clients/ts/FunctionalTests/Startup.cs
```diff
@@ -73,17 +77,33 @@ public void ConfigureServices(IServiceCollection services)
                     {
                         OnMessageReceived = context =>
                         {
-                            var signalRTokenHeader = context.Request.Query["access_token"];
+                            var endpoint = context.HttpContext.Features.Get<IEndpointFeature>()?.Endpoint;
```
pranavkm: Auth currently runs twice in SignalR (#9216). Perhaps it would be safer to leave that be in preview4. This particular change was the result of the WebSocketMiddleware not having run prior to endpoint routing - consequently the check `context.HttpContext.WebSockets.IsWebSocketRequest` failed.

pranavkm:
🆙 📅

src/SignalR/clients/ts/FunctionalTests/Startup.cs
```diff
@@ -73,17 +77,33 @@ public void ConfigureServices(IServiceCollection services)
...
+                            if (endpoint != null && endpoint.Metadata.GetMetadata<HubMetadata>() != null)
```
pranavkm: @davidfowl this particular change was suggested by @BrennanConroy. In lieu of checking if a particular request is a WebSocket request, this always checks if the query string is specified. However it needs to read the Auth header if available.

src/Http/Http.Abstractions/src/IAllowAnonymous.cs
halter73:
```suggestion
    /// Marker interface to allow access to anonymous users.
```

maybe?

src/SignalR/clients/ts/FunctionalTests/Startup.cs
halter73: I think this is reasonable.

The slight concern I have is that this could encourage people to put an access_token in the query string when they could have set an Authorization header. We'd like to encourage using the header over the query string since the query string is more likely to be logged.

I don't think that concern is big enough to warrant changing this logic though.

Eilon:

Please also resolve https://devdiv.visualstudio.com/DevDiv/_workitems/edit/843164when done.

pranavkm:
@halter73 \ @BrennanConroy how do I proceed with the SignalR functional test failure? The Windows build doesn't say what test failed and I'm not sure how to get to the logs to diagnose the Ubuntu failure.

natemcmaster:
👀 @aspnet/build - this will probably be the last Preview 4 change.

BrennanConroy:
The windows test is the same failure as the ubuntu one.

You can look in the `<os>_Test_Logs` artifacts folder for "browserlogs.console.log" and "ts.functionaltests.dotnet.log" for the server and client side logs. I looked at them last night and couldn't see any obvious issue. I would just skip the test for now to get this PR in.

pranavkm:
I'm giving it one more go putting back the custom CORS code that was in there. In the absence of it, I'm going to consider skipping the test.

pranavkm:
CORS seems to have done the trick.

BrennanConroy:
That is odd... any idea why?

pranavkm:
Not entirely sure. Perhaps the way the CI runs tests requires CORS to have run prior to `redirect`?

src/Http/Routing/src/RouteOptions.cs
```diff
@@ -28,6 +28,22 @@ public class RouteOptions
...
+        /// metadata will go unevaluated.
+        /// </para>
+        /// <para>
+        /// When <see langword="false"/>, prior to the execution of the endpoint, routing will verify that
+        /// all unevaluated metadata has been accounted for. Setting this property to <see langword="true"/>
+        /// suppresses this check.
+        /// </para>
+        /// </summary>
+        /// <value>Defaults to <see langword="true"/>.</value>
+        public bool SuppressCheckForUnevaluatedEndpointMetadata { get; set; }
```
rynowak: Can we call this `SuppressCheckForUnhandledSecurityMetadata` - I want to make this really clear that we intend it for security related features.

rynowak (approved):

natemcmaster:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=150893&view=logs

natemcmaster:
Previous test run failed on one flaky test. This has QB and review approvals, so I'm bringing it in with admin permissions.

PR #9186, opened 2019-04-08
Revert "Use new Fedora queue (#9178)"

jkotalik:
This reverts commit 5fb1ac22e9dd0ed0efd97dfa72d3cfca4ec19aed.

Apparently the fedora 28 queue may be working again and the script we run doesn't work on docker images. Still investigating if we can update our script.


dougbu (approved):

jkotalik:
I'm going to take a stab at doing this the right way in a separate branch.

PR #9188, opened 2019-04-09
Expose Connection Id in SignalR TS Client

### LINKED ISSUE #8681: Expose ConnectionId on SignalR Clients

mikaelm12:
Fixes #8681 
Last part of the exposing connection id issue.

### DIFF
--- a/src/SignalR/clients/ts/FunctionalTests/ts/HubConnectionTests.ts
+++ b/src/SignalR/clients/ts/FunctionalTests/ts/HubConnectionTests.ts
@@ -879,6 +879,33 @@ describe("hubConnection", () => {
             expect(newConnectionId).toBe(serverConnectionId);
 
             await hubConnection.stop();
+            expect(hubConnection.connectionId).toBeNull();
+
+            done();
+        } catch (err) {
+            fail(err);
+            done();
+        }
+    });
+
+    it("connection id is alwys null is negotiation is skipped", async (done) => {
+        try {
+            const hubConnection = getConnectionBuilder(
+                    HttpTransportType.WebSockets,
+                    undefined,
+                    { skipNegotiation: true },
+                )
+                .build();
+
+            expect(hubConnection.connectionId).toBeNull();
+
+            await hubConnection.start();
+
+            expect(hubConnection.connectionId).toBeNull();
+
+            await hubConnection.stop();
+
+            expect(hubConnection.connectionId).toBeNull();
 
             done();
         } catch (err) {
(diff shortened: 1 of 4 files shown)


src/SignalR/clients/ts/signalr/src/HubConnection.ts
```diff
@@ -120,6 +120,17 @@ export class HubConnection {
         return this.connectionState;
     }
 
+    /** Represents the connection id of the {@link HubConnection} on the server. The connection id will be null when the connection is either
+     *  in the disconnected state or if the negotiation step was skipped.
+     */
+    get connectionId(): string | null {
+        if (this.connection.connectionId) {
```
davidfowl: Can connection be null?

JamesNK: Does `return this.connection.connectionId || null` work here? (I'm guessing the if is to go from string/undefined to string/null)

JamesNK: The property comment explains when connectionId could be null.

davidfowl: connection not connection property. If connection is null it'll null ref.

mikaelm12: The underlying connection should always be defined

mikaelm12: But we can add a check here if you want

davidfowl: We don't need it if it's never null

src/SignalR/clients/ts/FunctionalTests/ts/HubConnectionTests.ts
```diff
@@ -838,6 +841,52 @@ describe("hubConnection", () => {
         }
     });
 
+    it("connection id matches server side connection id", async (done) => {
```
davidfowl: Do we have a test that verifies when it's null?

mikaelm12: Theres' an assert before start is called, there could be another for after stop. I'll add another test for when we skip negotiate and the id is always null.

davidfowl (approved):

PR #9206, opened 2019-04-09
Make NodeJS and Java required to build

pranavkm:


poke:
Is there a way to opt out of that somewhow? Like, I probably don’t ever want to contribute to the Java SignalR client, so I really don’t want to have to install the Java SDK just to be able to compile non-Java stuff in the repo.

natemcmaster:
Maybe I missed something while I was out, but I thought NodeJS and Java were still optional, and only needed if you add additional paramaters to opt-in to nodejs/java projects, e.g. `build.sh --build-nodejs`

pranavkm:
@natemcmaster, building Blazor.Server requires `yarn` to be present. The generated binary is incorrect without the embedded js files. Additionally, building template tests now fail if you do not have java installed: https://github.com/aspnet/AspNetCore/blob/master/src/Shared/E2ETesting/E2ETesting.targets#L37-L38. The error appears in the default code path when you do `build`. I'm simply updating the docs to reflect these new requirements

pranavkm:
@poke generally building individual "sub-folders" should not require java or NodeJS to be installed. However building the repo does require these tools, the build fails otherwise.

natemcmaster (approved):
Thanks for clarifying @pranavkm !

ryanbrandenburg (approved):

PR #9222, opened 2019-04-09
Fix stale references

ryanbrandenburg:
Attempted fix for https://github.com/aspnet/AspNetCore-Internal/issues/2136. I wasn't able to get this to re-produce, but seeing as the `FindElement` is on the same line as the `.Text` here it seems likely that the problem here is that Browsers are getting re-used.

### DIFF
--- a/src/Shared/E2ETesting/BrowserTestBase.cs
+++ b/src/Shared/E2ETesting/BrowserTestBase.cs
@@ -10,7 +10,7 @@ using Xunit.Abstractions;
 namespace Microsoft.AspNetCore.E2ETesting
 {
     [CaptureSeleniumLogs]
-    public class BrowserTestBase : IClassFixture<BrowserFixture>, IAsyncLifetime
+    public abstract class BrowserTestBase : IClassFixture<BrowserFixture>, IAsyncLifetime
     {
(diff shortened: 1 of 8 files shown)


src/Shared/E2ETesting/BrowserTestBase.cs
```diff
@@ -32,14 +32,16 @@ public BrowserTestBase(BrowserFixture browserFixture, ITestOutputHelper output)
...
+
         public Task DisposeAsync()
         {
             return Task.CompletedTask;
         }
 
         public virtual async Task InitializeAsync()
         {
-            var (browser, logs) = await BrowserFixture.GetOrCreateBrowserAsync(Output);
+            var (browser, logs) = await BrowserFixture.GetOrCreateBrowserAsync(Output, BrowserIsolationContext);
```
javiercn: This is only needed in a couple of tests where we want to ensure a clean slate within the tests. Please undo this change

javiercn (changes requested):
I don't see how this would fix the problem. There is likely a way more targeted fix that can be done in the specific test and I feel this will only destabilize other tests. Reusing the browser is not a problem unless you create cookies and things like that.

PR #9224, opened 2019-04-09
Make DataProtection certificate tests assertions stricter about which  cert is used in the test

natemcmaster:
Addresses aspnet/AspNetCore-Internal#2177

This test only fails on some CI agents. As these tests rely on a certificate store shared by other builds, it's probably that the failures are the result of some other dotnet team adding a cert called 'TestCert' into the X509Store.

This changes the tests to pull from the X509Store the _exact_ cert we want the test to use.


dougbu (approved):
🥇

natemcmaster:
Ok, nope, that wasn't the right solution for the problem. Will continue to investigate.

Issue #9121, opened 2019-04-05
MvcCoreLoggerExtensions - Information Leakage Security Vulnerability


aaronlcope:
### Describe the bug
Information leakage found in MVCCoreLoggerExtensions.cs facility.

On line 195 of [MvcCoreLoggerExtensions.cs](https://github.com/aspnet/AspNetCore/blob/c565386a3ed135560bc2e9017aa54a950b4e35dd/src/Mvc/Mvc.Core/src/MvcCoreLoggerExtensions.cs)

The action method arguments are logged. MVC applications may contain authentication screens where user sensitive credentials can be obtained from these log files.

### To Reproduce
Set up logging in an MVC app, develop a login screen, configure logging at INFO-level and watch the logs spit out user credentials.

### Expected behavior
User credentials should never be logged. In fact, any unpredictable data (that is user input) should be left to the implementor to decide if it should be logged. I would think the correct approach is to not log the arguments at all and leave that up to the MVC developer to choose carefully and control that logging on their own.

blowdart:
Interesting. As Information isn't the default loglevel you wouldn't get this logged by default, but when you do tweak the logging level yes, this is a potential problem if you haven't secured your logs like best practice would dictate.

So I'm not classifying this as a security patch Tuesday type bug, but rather something that will be fixed in the next scheduled releases for 2.x and above. The change will be that rather than the argument name/value pair being logged you'll only see the argument names. Tracing will instead log the full name/value pair.

blowdart:
@mkArtakMSFT Can you assign this to someone please?

aaronlcope:
@blowdart  - thanks for the eyes on this. much appreciated.

blowdart:
For future reference, security issues should first go to secure@microsoft.com, where they can qualify for our [bug bounty](https://www.microsoft.com/en-us/msrc/bounty-dot-net-core). Github issues can't qualify.

mkArtakMSFT:
Thanks @blowdart.
@pranavkm lets get this fixed for 2.1, 2.2 as well as the upcoming releases.

poke:
Why is the fix for this to completely remove the logging of the arguments? Wouldn’t it make more sense to just reduce the log level on the arguments as the [log level documentation](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/?view=aspnetcore-2.2#log-level) would suggest? On `Debug` or even `Trace`, this shouldn’t be a problem at all.

[...]

PR #9227, opened 2019-04-09
Don't log action and page handler arguments above Trace level

pranavkm:
Fixes https://github.com/aspnet/AspNetCore/issues/9121

### Impact
The contract for logging says any sensitive information would be logged at `LogLevel.Trace`. However, MVC logs some user inputs which could contain potentially sensitive information at `LogLevel.Info`

Workaround
Workaround is to disable Info logging for Mvc.

Risk
This should be fairly low risk. Logging MVC is disabled in production applications.

### DIFF
--- a/eng/PatchConfig.props
+++ b/eng/PatchConfig.props
@@ -28,6 +28,7 @@ Later on, this will be checked using this condition:
   </PropertyGroup>
   <PropertyGroup Condition=" '$(VersionPrefix)' == '2.1.11' ">
     <PackagesInPatch>
+      Microsoft.AspNetCore.Mvc.Core;
     </PackagesInPatch>
   </PropertyGroup>
 
--- a/eng/PatchConfig.props
+++ b/eng/PatchConfig.props
@@ -29,6 +29,7 @@ Later on, this will be checked using this condition:
   <PropertyGroup Condition=" '$(VersionPrefix)' == '2.1.11' ">
     <PackagesInPatch>
       Microsoft.AspNetCore.Mvc.Core;
+      Microsoft.AspNetCore.Mvc.RazorPages;
     </PackagesInPatch>
   </PropertyGroup>
 
--- a/src/Mvc/Mvc.Core/src/Internal/MvcCoreLoggerExtensions.cs
+++ b/src/Mvc/Mvc.Core/src/Internal/MvcCoreLoggerExtensions.cs
@@ -174,7 +174,7 @@ namespace Microsoft.AspNetCore.Mvc.Internal
                 "Executing action method {ActionName} - Validation state: {ValidationState}");
 
             _actionMethodExecutingWithArguments = LoggerMessage.Define<string, string[]>(
-                LogLevel.Debug,
+                LogLevel.Trace,
                 3,
                 "Executing action method {ActionName} with arguments ({Arguments})");
 
@@ -816,7 +816,7 @@ namespace Microsoft.AspNetCore.Mvc.Internal
                 var validationState = context.ModelState.ValidationState;
                 _actionMethodExecuting(logger, actionName, validationState, null);
 
-                if (arguments != null && logger.IsEnabled(LogLevel.Debug))
+                if (arguments != null && logger.IsEnabled(LogLevel.Trace))
                 {
                     var convertedArguments = new string[arguments.Length];
                     for (var i = 0; i < arguments.Length; i++)
--- a/src/Mvc/Mvc.Core/src/Internal/MvcCoreLoggerExtensions.cs
+++ b/src/Mvc/Mvc.Core/src/Internal/MvcCoreLoggerExtensions.cs
@@ -39,6 +39,7 @@ namespace Microsoft.AspNetCore.Mvc.Internal
         private static readonly Action<ILogger, string, Exception> _contentResultExecuting;
 
         private static readonly Action<ILogger, string, ModelValidationState, Exception> _actionMethodExecuting;
+        private static readonly Action<ILogger, string, string[], Exception> _actionMethodExecutingWithArguments;
         private static readonly Action<ILogger, string, string, double, Exception> _actionMethodExecuted;
 
         private static readonly Action<ILogger, string, string[], Exception> _logFilterExecutionPlan;
@@ -171,6 +172,11 @@ namespace Microsoft.AspNetCore.Mvc.Internal
                 LogLevel.Information,
                 1,
                 "Executing action method {ActionName} - Validation state: {ValidationState}");
+
+            _actionMethodExecutingWithArguments = LoggerMessage.Define<string, string[]>(
+                LogLevel.Debug,
+                3,
+                "Executing action method {ActionName} with arguments ({Arguments})");
 
             _actionMethodExecuted = LoggerMessage.Define<string, string, double>(
                 LogLevel.Information,
@@ -809,6 +815,17 @@ namespace Microsoft.AspNetCore.Mvc.Internal
 
                 var validationState = context.ModelState.ValidationState;
                 _actionMethodExecuting(logger, actionName, validationState, null);
+
+                if (arguments != null && logger.IsEnabled(LogLevel.Debug))
+                {
+                    var convertedArguments = new string[arguments.Length];
+                    for (var i = 0; i < arguments.Length; i++)
+                    {
+                        convertedArguments[i] = Convert.ToString(arguments[i]);
+                    }
+
+                    _actionMethodExecutingWithArguments(logger, actionName, convertedArguments, null);
+                }
             }
         }
 
--- a/src/Mvc/Mvc.Core/src/Internal/MvcCoreLoggerExtensions.cs
+++ b/src/Mvc/Mvc.Core/src/Internal/MvcCoreLoggerExtensions.cs
@@ -39,7 +39,6 @@ namespace Microsoft.AspNetCore.Mvc.Internal
         private static readonly Action<ILogger, string, Exception> _contentResultExecuting;
 
         private static readonly Action<ILogger, string, ModelValidationState, Exception> _actionMethodExecuting;
-        private static readonly Action<ILogger, string, string[], ModelValidationState, Exception> _actionMethodExecutingWithArguments;
         private static readonly Action<ILogger, string, string, double, Exception> _actionMethodExecuted;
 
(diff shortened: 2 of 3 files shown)


dougbu (commented):
Which milestone is this PR intended for? Doesn't look like an approval process has begun.

poke:
As per #9121, can we please keep logging the action arguments, but just use a different log level that allows us to include possibly sensitive information? So `Debug` or `Trace`? Having this information available by default is very valuable.

pranavkm:
@poke sure. I can log it as a separate log message at debug level. I do want to keep the log level for the current action executing \ page executing unchanged at Info.

poke:
That’s perfect! Thanks a lot!

src/Mvc/Mvc.Core/src/Internal/MvcCoreLoggerExtensions.cs
```diff
@@ -173,10 +173,10 @@ static MvcCoreLoggerExtensions()
                 1,
                 "Executing action method {ActionName} - Validation state: {ValidationState}");
 
-            _actionMethodExecutingWithArguments = LoggerMessage.Define<string, string[], ModelValidationState>(
-                LogLevel.Information,
-                1,
-                "Executing action method {ActionName} with arguments ({Arguments}) - Validation state: {ValidationState}");
+            _actionMethodExecutingWithArguments = LoggerMessage.Define<string, string[]>(
+                LogLevel.Debug,
```
rynowak: This has to be `Trace` not debug. Debug is not supposed to include sensitive information.

src/Mvc/Mvc.Core/src/Internal/MvcCoreLoggerExtensions.cs
```diff
@@ -173,10 +173,10 @@ static MvcCoreLoggerExtensions()
...
+                3,
+                "Executing action method {ActionName} with arguments ({Arguments})");
```
rynowak: What happened to the original log message?

IMO it makes much more sense to log what we did before at info (without the arguments) and then make a separate log statement that shows the arguments.

src/Mvc/Mvc.Core/src/Internal/MvcCoreLoggerExtensions.cs
pranavkm: I did consider setting this to trace. While not identical, we do log some other things including the RVD at Debug level: https://github.com/aspnet/AspNetCore/blob/ee0fb6d0e268afcb4848c64ef42a0cb0f48f8e3f/src/Mvc/Mvc.Core/src/Internal/MvcCoreLoggerExtensions.cs#L779-L782

Should I do a pass over these?

src/Mvc/Mvc.Core/src/Internal/MvcCoreLoggerExtensions.cs
pranavkm: That's still there. See `_actionMethodExecuting`

src/Mvc/Mvc.Core/src/Internal/MvcCoreLoggerExtensions.cs
rynowak: We don't consider route values to be sensitive.

dougbu:
@pranavkm the next shiproom meeting is tomorrow at 11:00. Please add the requisite label and approval template before that. We aren't meeting again before the merge deadline and discussions in the room tend to go quicker.

vivmishra:
Approved for 2.1.11, 2.2.5 and 3.0.

pranavkm:
Logs from [controllers](https://gist.github.com/pranavkm/8c1e5c1c8d7798dc635451e86e9702be) and [Razor Pages](https://gist.github.com/pranavkm/7a490df90c715596eff8e1ab0ea525d2) with this change

poke:
With this change, is there a way now to say which overload of a method is being executed?

Say there's one `Login()` and one `Login(LoginViewModel)`. Before, you could tell (from the info logs) which overload was being executed. Now that information is gone and you would actually need the debug logs.

dougbu (approved):
@pranavkm please let me know if this is ready to merge. That is, do you have additional changes you want to make?

pranavkm:
@poke unfortunately that seems to be the case. I had a look at if there's a straightforward solution to including this information in these logs, but that not seem to be the case. We can certainly make this nicer in 3.0. For 2.1 and 2.2, a possible workaround add a trace level filter for the category `Microsoft.AspNetCore.Mvc.Internal.ControllerActionInvoker`:

```
  "Logging": {
    "LogLevel": {
      "Default": "Warning",
      "Microsoft.AspNetCore.Mvc.Internal.ControllerActionInvoker": "Trace"
    }
```

dougbu:
@pranavkm please let me know if this is ready to merge.

dougbu:
:up::date: to do the needful in PatchConfig.props

poke:
@pranavkm What do you think about including the number of arguments in the log message? That way it should give a good idea which overload is called (not perfect but better than nothing). And it would be straight forward to do so to.

(I don't want to further delay this PR with this though, so feel free to merge this for now and maybe we can look at it again for 3.0)

rynowak:
We talked about including the types of the arguments on Thursday. We have some existing code that does this elsewhere.

natemcmaster:
Once merged, will this need to be part of 2.2.x as well? If so, make sure the bot-generated PR is completed, which should be opened shortly after this PR is completed.

pranavkm:
🆙 📅

pranavkm:
Here's what it looks like now:

```
info: Microsoft.AspNetCore.Hosting.Internal.WebHost[1]
      Request starting HTTP/1.1 GET http://localhost:51001/
info: Microsoft.AspNetCore.Mvc.Internal.ControllerActionInvoker[3]
      Route matched with {action = "Index", controller = "Home", page = ""}. Executing controller action with signature Microsoft.AspNetCore.Mvc.IActionResult Index() on controller MvcSandbox.Controllers.HomeController (MvcSandbox).
info: Microsoft.AspNetCore.Mvc.Internal.ControllerActionInvoker[1]
      Executing action method MvcSandbox.Controllers.HomeController.Index (MvcSandbox) - Validation state: Valid
info: Microsoft.AspNetCore.Mvc.Internal.ControllerActionInvoker[2]
      Executed action method MvcSandbox.Controllers.HomeController.Index (MvcSandbox), returned result Microsoft.AspNetCore.Mvc.ViewResult in 5.5025ms.
info: Microsoft.AspNetCore.Mvc.ViewFeatures.ViewResultExecutor[1]
      Executing ViewResult, running view Index.
info: Microsoft.AspNetCore.Mvc.ViewFeatures.ViewResultExecutor[4]
      Executed ViewResult - view Index executed in 283.7445ms.
info: Microsoft.AspNetCore.Mvc.Internal.ControllerActionInvoker[2]
      Executed action MvcSandbox.Controllers.HomeController.Index (MvcSandbox) in 402.5448ms
info: Microsoft.AspNetCore.Hosting.Internal.WebHost[2]
      Request finished in 944.9806ms 200 text/html; charset=utf-8
```

rynowak (approved):

eng/PatchConfig.props
```diff
@@ -28,6 +28,7 @@ Later on, this will be checked using this condition:
   </PropertyGroup>
   <PropertyGroup Condition=" '$(VersionPrefix)' == '2.1.11' ">
     <PackagesInPatch>
+      Microsoft.AspNetCore.Mvc.Core;
```
pranavkm: @natemcmaster \ @dougbu does this look correct?

natemcmaster: Yes.

natemcmaster: Based on source code changes, looks like you need to update Mvc.RazorPages as well.

pranavkm:
@natemcmaster \ @dougbu could you merge?

pranavkm:
Verified this change is also present in 2.2.5.

fiksen99:
Which version did this make it into? We are using 2.1.7 and I see no 2.1 updates since

vivmishra:
You can view the various downloads for 2.X releases at https://dotnet.microsoft.com/download/dotnet-core 
We are at v2.1.12 for 2.1.

fiksen99:
We're targetting .NET Framework with ASP.NET Core, I see the
Microsoft.AspNetCore package has not been updated since v 2.1.7

I've reached out to the feature team also, but may as well ask here - is
this package going to be updated for those of us on .NET Framework and
ASP.NET Core 2.1 LTS?

pranavkm:
@fiksen99 for .NET Framework, the change would be in the corresponding `Microsoft.AspNetCore.Mvc` package, not in the metapackage. I'd recommend using the NuGet UI to update to the latest patch releases of all of the Microsoft..* package.

fiksen99:
Thanks, I initially thought that would be the case, but in the `Microsoft.AspNetCore.Mvc` nuget, we see the lastest 2.1.X update is 2.1.3 in October 2018
image
https://www.nuget.org/packages/Microsoft.AspNetCore.Mvc/

poke:
The changes are in [AspNetCore.Mvc.Core](https://www.nuget.org/packages/Microsoft.AspNetCore.Mvc.Core) and [AspNetCore.Mvc.RazorPages](https://www.nuget.org/packages/Microsoft.AspNetCore.Mvc.RazorPages), version 2.1.11 each.

I am not sure why the parent package AspNetCore.Mvc wasn't updated to transitively deliver these versions. I would assume that this is actually a publishing bug, since neither Mvc.Core nor Mvc.RazorPages are packages that you usually have to depend on explicitly on netfx.

fiksen99:
Thanks. Do you know who owns updating the AspNetCore.Mvc package, and are they aware of this issue?

Are there any potential implications of referenceing AspNetCore.Mvc 2.1.3 and then directly referencing AspNetCore.Mvc.Core 2.1.11+? Or are they intended to be backwards compatible?

poke:
> Are there any potential implications of referenceing AspNetCore.Mvc 2.1.3 and then directly referencing AspNetCore.Mvc.Core 2.1.11+?

That should work. Referencing the newer version explicitly will make AspNetCore.Mvc use that version (instead of the older one) too.

fiksen99:
makes sense. Would be great to get the entire main Mvc package released too. I assume given 2.1 is LTS, the intention is that these packages continue to receive updates

Issue #9170, opened 2019-04-08
Update daily build info with correct feeds/blobs


Eilon:
The information here seems out of date:

https://github.com/aspnet/AspNetCore/blob/master/docs/DailyBuilds.md

It lists the .NET Core blob feed but not the ASP.NET Core blob feed.

cc @dougbu @danroth27

dougbu:
If we need an another blob feed or aspnetcore-dev from GitHub, something is busted. The instructions should be (and were when I changed that Markdown recently) sufficient and correct. @danroth27 what in particular was missing when attempting to consume our packages?

danroth27:
The Blazor packages (ex Microsoft.AspNetCore.Blazor) were missing.

danroth27:
It looks like the Blazor packages are on the dotnet blob feed, but they are old:

![image](https://user-images.githubusercontent.com/1874516/55764253-a29a0680-5a1f-11e9-8887-067542d44061.png)

The correct versions should match the .NET Core 3.0 verisons (3.0.0-previewN-xxxx). The correct packages appear to be on the aspnetcore blog feed:

![image](https://user-images.githubusercontent.com/1874516/55764285-c8bfa680-5a1f-11e9-8d78-8f652565d54f.png)

dougbu:
Oh, I remember now. @natemcmaster broke our stuff across multiple blob feeds and what's currently in `DailyBuilds.md` only works for projects that use packages from the shared framework. I'll send a PR to add the additional blob stores.

dougbu:
95cda2c627fb

PR #9234, opened 2019-04-10
Add more feeds to DailyBuilds.md

dougbu:
- #9170

### DIFF
--- a/docs/DailyBuilds.md
+++ b/docs/DailyBuilds.md
@@ -14,6 +14,10 @@ If you want to download the latest daily build and use it in a project, then you
       <packageSources>
           <clear />
           <add key="dotnet-core" value="https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json" />
+          <add key="extensions" value="https://dotnetfeed.blob.core.windows.net/aspnet-extensions/index.json" />
+          <add key="entityframeworkcore" value="https://dotnetfeed.blob.core.windows.net/aspnet-entityframeworkcore/index.json" />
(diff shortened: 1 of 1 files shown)


dougbu:
@danroth27 please confirm your previously-failing tests work with this new configuration

davidfowl:
Why do we need this many feeds now?

dougbu:
@davidfowl root cause is the lock contention we and our partners ran into when everything was pushing to a single blob store. (Our partners are considering breaking up the core-setup store as well.)

Good news is this is temporary. Will be moving to entirely new feeds&hellip;

danroth27 (approved):

docs/DailyBuilds.md
```diff
@@ -14,6 +14,10 @@ If you want to download the latest daily build and use it in a project, then you
       <packageSources>
           <clear />
           <add key="dotnet-core" value="https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json" />
+          <add key="extensions" value="https://dotnetfeed.blob.core.windows.net/aspnet-extensions/index.json" />
+          <add key="entityframeworkcore" value="https://dotnetfeed.blob.core.windows.net/aspnet-entityframeworkcore/index.json" />
+          <add key="aspnetcore-tooling" value="https://dotnetfeed.blob.core.windows.net/aspnet-aspnetcore-tooling/index.json;" />
```
natemcmaster: Does the semicolon at the end cause issue for the NuGet client? (applies to the next line, too)

dougbu: Good catch. I'm not sure about what problems it might cause but I'll clean these up in a moment&hellip;

PR #9252, opened 2019-04-10
Fix #8974 by supporting strings in configureLogging for SignalR JS client

### LINKED ISSUE #8974: Make it possible to set SignalR log level in javascript with strings
see: https://github.com/aspnet/AspNetCore/blob/master/src/SignalR/clients/ts/signalr/src/HubConnectionBuilder.ts#L30

It would be really useful we we could pass in a string with one of the named values of `LogLevel`. We have a use case for this in Blazor where the user won't be writing typescript and probably won't have SignalR imported.

analogrelay:
Fixes #8974 

This allows users to pass in a string for the log level when calling `configureLogging`. This makes it much easier to configure logging in environments where you don't have access to the enum definition (such as in Blazor when you're just doing this in a script block in your page).

Note: I also grouped all the `configureLogging` tests into their own `describe` because there were now quite a few :).

Test output for HubConnectionBuilder.test.ts:

```
 PASS  signalr/tests/HubConnectionBuilder.test.ts (6.62s)
  HubConnectionBuilder
    √ withUrl throws if url is null (3ms)
    √ withHubProtocol throws if protocol is null
    √ withUrl throws if url is undefined
    √ withHubProtocol throws if protocol is undefined (1ms)
    √ builds HubConnection with HttpConnection using provided URL (12ms)
    √ can configure transport type
    √ can configure hub protocol (2ms)
    √ reconnectPolicy undefined by default (1ms)
    √ withAutomaticReconnect throws if reconnectPolicy is already set
    √ withAutomaticReconnect uses default retryDelays when called with no arguments (1ms)
    √ withAutomaticReconnect uses custom retryDelays when provided
    √ withAutomaticReconnect uses a custom IReconnectPolicy when provided
    configureLogging
      √ throws if logger is null (1ms)
      √ throws if logger is undefined
      √ accepts LogLevel.None
      √ accepts LogLevel.Critical
      √ accepts LogLevel.Error
      √ accepts LogLevel.Warning (1ms)
      √ accepts LogLevel.Information
      √ accepts LogLevel.Debug
      √ accepts LogLevel.Trace (1ms)
      √ accepts "critical" as an alias for LogLevel.Critical
      √ accepts "debug" as an alias for LogLevel.Debug (1ms)
      √ accepts "error" as an alias for LogLevel.Error
      √ accepts "info" as an alias for LogLevel.Information
      √ accepts "information" as an alias for LogLevel.Information (1ms)
      √ accepts "none" as an alias for LogLevel.None
      √ accepts "trace" as an alias for LogLevel.Trace
      √ accepts "warn" as an alias for LogLevel.Warning
      √ accepts "warning" as an alias for LogLevel.Warning (1ms)
      √ allows logger to be replaced (1ms)
      √ configures logger for both HttpConnection and HubConnection (2ms)
      √ does not replace HttpConnectionOptions logger if provided (1ms)
```

cc @rynowak

### DIFF
--- a/src/SignalR/clients/ts/signalr/src/HubConnectionBuilder.ts
+++ b/src/SignalR/clients/ts/signalr/src/HubConnectionBuilder.ts
@@ -13,18 +13,17 @@ import { JsonHubProtocol } from "./JsonHubProtocol";
 import { NullLogger } from "./Loggers";
 import { Arg, ConsoleLogger } from "./Utils";
 
-// This is exported from this module but NOT from index, so it's "internal" for testing purposes
-export const LogLevelNameMapping = {
-    // TS Lint prefers alphabetical sorting...
-    critical: LogLevel.Critical,
+// tslint:disable:object-literal-sort-keys
+const LogLevelNameMapping = {
+    trace: LogLevel.Trace,
     debug: LogLevel.Debug,
-    error: LogLevel.Error,
     info: LogLevel.Information,
     information: LogLevel.Information,
-    none: LogLevel.None,
-    trace: LogLevel.Trace,
     warn: LogLevel.Warning,
     warning: LogLevel.Warning,
+    error: LogLevel.Error,
+    critical: LogLevel.Critical,
+    none: LogLevel.None,
 };
 
 function parseLogLevel(name: keyof typeof LogLevelNameMapping): LogLevel {
--- a/src/SignalR/clients/ts/signalr/src/HubConnectionBuilder.ts
+++ b/src/SignalR/clients/ts/signalr/src/HubConnectionBuilder.ts
@@ -13,6 +13,32 @@ import { JsonHubProtocol } from "./JsonHubProtocol";
 import { NullLogger } from "./Loggers";
 import { Arg, ConsoleLogger } from "./Utils";
 
+// This is exported from this module but NOT from index, so it's "internal" for testing purposes
+export const LogLevelNameMapping = {
+    // TS Lint prefers alphabetical sorting...
+    critical: LogLevel.Critical,
+    debug: LogLevel.Debug,
+    error: LogLevel.Error,
+    info: LogLevel.Information,
+    information: LogLevel.Information,
+    none: LogLevel.None,
+    trace: LogLevel.Trace,
+    warn: LogLevel.Warning,
+    warning: LogLevel.Warning,
+};
+
+function parseLogLevel(name: keyof typeof LogLevelNameMapping): LogLevel {
+    // Case-insensitive matching via lower-casing
+    // Yes, I know case-folding is a complicated problem in Unicode, but we only support
+    // the ASCII strings defined in LogLevelNameMapping anyway, so it's fine -anurse.
+    const mapping = LogLevelNameMapping[name.toLowerCase()];
+    if (typeof mapping !== "undefined") {
+        return mapping;
+    } else {
+        throw new Error(`Unknown log level: ${name}`);
+    }
+}
+
 /** A builder for configuring {@link @aspnet/signalr.HubConnection} instances. */
 export class HubConnectionBuilder {
     /** @internal */
@@ -47,12 +73,15 @@ export class HubConnectionBuilder {
      * @param {LogLevel | ILogger} logging An object implementing the {@link @aspnet/signalr.ILogger} interface or {@link @aspnet/signalr.LogLevel}.
      * @returns The {@link @aspnet/signalr.HubConnectionBuilder} instance, for chaining.
      */
-    public configureLogging(logging: LogLevel | ILogger): HubConnectionBuilder;
-    public configureLogging(logging: LogLevel | ILogger): HubConnectionBuilder {
+    public configureLogging(logging: LogLevel | keyof typeof LogLevelNameMapping | ILogger): HubConnectionBuilder;
+    public configureLogging(logging: LogLevel | keyof typeof LogLevelNameMapping | ILogger): HubConnectionBuilder {
         Arg.isRequired(logging, "logging");
 
         if (isLogger(logging)) {
             this.logger = logging;
+        } else if (typeof logging === "string") {
+            const logLevel = parseLogLevel(logging);
+            this.logger = new ConsoleLogger(logLevel);
         } else {
             this.logger = new ConsoleLogger(logging);
         }
@@ -92,7 +121,7 @@ export class HubConnectionBuilder {
         // Flow-typing knows where it's at. Since HttpTransportType is a number and IHttpConnectionOptions is guaranteed
         // to be an object, we know (as does TypeScript) this comparison is all we need to figure out which overload was called.
         if (typeof transportTypeOrOptions === "object") {
-            this.httpConnectionOptions = {...this.httpConnectionOptions, ...transportTypeOrOptions};
+            this.httpConnectionOptions = { ...this.httpConnectionOptions, ...transportTypeOrOptions };
         } else {
             this.httpConnectionOptions = {
                 ...this.httpConnectionOptions,
--- a/src/SignalR/clients/ts/signalr/src/HubConnectionBuilder.ts
+++ b/src/SignalR/clients/ts/signalr/src/HubConnectionBuilder.ts
@@ -69,11 +69,11 @@ export class HubConnectionBuilder {
 
     /** Configures custom logging for the {@link @aspnet/signalr.HubConnection}.
      *
-     * @param {LogLevel | ILogger} logging An object implementing the {@link @aspnet/signalr.ILogger} interface or {@link @aspnet/signalr.LogLevel}.
+     * @param {LogLevel | string | ILogger} logging A {@link @aspnet/signalr.LogLevel}, a string representing a LogLevel, or an object implementing the {@link @aspnet/signalr.ILogger} interface. See {@link https://docs.microsoft.com/en-us/aspnet/core/signalr/configuration#configure-logging|the documentation for client logging configuration} for more details.
      * @returns The {@link @aspnet/signalr.HubConnectionBuilder} instance, for chaining.
      */
     public configureLogging(logging: LogLevel | keyof typeof LogLevelNameMapping | ILogger): HubConnectionBuilder;
-    public configureLogging(logging: LogLevel | keyof typeof LogLevelNameMapping | ILogger): HubConnectionBuilder {
+    public configureLogging(logging: LogLevel | string | ILogger): HubConnectionBuilder {
         Arg.isRequired(logging, "logging");
 
         if (isLogger(logging)) {
--- a/src/SignalR/clients/ts/signalr/src/HubConnectionBuilder.ts
+++ b/src/SignalR/clients/ts/signalr/src/HubConnectionBuilder.ts
@@ -26,7 +26,7 @@ const LogLevelNameMapping = {
     none: LogLevel.None,
 };
 
-function parseLogLevel(name: keyof typeof LogLevelNameMapping): LogLevel {
+function parseLogLevel(name: string): LogLevel {
     // Case-insensitive matching via lower-casing
     // Yes, I know case-folding is a complicated problem in Unicode, but we only support
     // the ASCII strings defined in LogLevelNameMapping anyway, so it's fine -anurse.
@@ -69,7 +69,8 @@ export class HubConnectionBuilder {
 
     /** Configures custom logging for the {@link @aspnet/signalr.HubConnection}.
      *
-     * @param {LogLevel | string | ILogger} logging A {@link @aspnet/signalr.LogLevel}, a string representing a LogLevel, or an object implementing the {@link @aspnet/signalr.ILogger} interface. See {@link https://docs.microsoft.com/en-us/aspnet/core/signalr/configuration#configure-logging|the documentation for client logging configuration} for more details.
+     * @param {LogLevel | string | ILogger} logging A {@link @aspnet/signalr.LogLevel}, a string representing a LogLevel, or an object implementing the {@link @aspnet/signalr.ILogger} interface.
+     *    See {@link https://docs.microsoft.com/en-us/aspnet/core/signalr/configuration#configure-logging|the documentation for client logging configuration} for more details.
      * @returns The {@link @aspnet/signalr.HubConnectionBuilder} instance, for chaining.
      */
     public configureLogging(logging: LogLevel | string | ILogger): HubConnectionBuilder {
--- a/src/SignalR/clients/ts/signalr/src/HubConnectionBuilder.ts
+++ b/src/SignalR/clients/ts/signalr/src/HubConnectionBuilder.ts
@@ -69,10 +69,18 @@ export class HubConnectionBuilder {
 
     /** Configures custom logging for the {@link @aspnet/signalr.HubConnection}.
      *
+     * @param {string} logLevel A string representing a LogLevel setting a minimum level of messages to log.
+     *    See {@link https://docs.microsoft.com/en-us/aspnet/core/signalr/configuration#configure-logging|the documentation for client logging configuration} for more details.
+     */
+    public configureLogging(logLevel: string): HubConnectionBuilder;
+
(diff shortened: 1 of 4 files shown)


src/SignalR/clients/ts/signalr/src/HubConnectionBuilder.ts
```diff
@@ -13,6 +13,32 @@ import { JsonHubProtocol } from "./JsonHubProtocol";
...
+    // TS Lint prefers alphabetical sorting...
+    critical: LogLevel.Critical,
+    debug: LogLevel.Debug,
+    error: LogLevel.Error,
+    info: LogLevel.Information,
+    information: LogLevel.Information,
+    none: LogLevel.None,
+    trace: LogLevel.Trace,
+    warn: LogLevel.Warning,
+    warning: LogLevel.Warning,
```
halter73: Should we capitalize the first letter like we do for the LogLevel enum?

src/SignalR/clients/ts/signalr/src/HubConnectionBuilder.ts
```diff
@@ -13,6 +13,32 @@ import { JsonHubProtocol } from "./JsonHubProtocol";
 import { NullLogger } from "./Loggers";
 import { Arg, ConsoleLogger } from "./Utils";
 
+// This is exported from this module but NOT from index, so it's "internal" for testing purposes
+export const LogLevelNameMapping = {
+    // TS Lint prefers alphabetical sorting...
```
halter73: Put `/* tslint:disable:object-literal-sort-keys */` above the declaration and sort based log severity.

src/SignalR/clients/ts/signalr/tests/HubConnectionBuilder.test.ts
```diff
@@ -111,98 +134,160 @@ describe("HubConnectionBuilder", () => {
...
-        try {
-            await connection.start();
-        } catch {
-            // Ignore failures
-        }
+        });
 
-        // A HubConnection message
-        expect(logger.messages).toContain("Starting HubConnection.");
+        const levelNames = Object.keys(LogLevelNameMapping) as Array<keyof typeof LogLevelNameMapping>;
```
halter73: Maybe use a different mapping object for the tests, so test will fail if we make a breaking change without realizing.

src/SignalR/clients/ts/signalr/src/HubConnectionBuilder.ts
halter73: If not, we might want to change our HubConnection state enum to use all lower-case characters too:

https://github.com/aspnet/AspNetCore/blob/4964b338abe02bdbb21ff69928d38b11e85c90c9/src/SignalR/clients/ts/signalr/src/HubConnection.ts#L17-L28

anurse: It needs to be lowercase for the parsing to work since we do `name.toLowerCase()`. This type, and it's members, don't appear to users, it's just an internal type used for the parser and for the tests.

src/SignalR/clients/ts/signalr/tests/HubConnectionBuilder.test.ts
anurse: That's reasonable. I was thinking exactly the opposite (to save having to update the test) but I think you're right, it's better to know.

anurse:
:up: :date:

src/SignalR/clients/ts/signalr/src/Utils.ts
```diff
@@ -146,27 +146,34 @@ export class SubjectSubscription<T> implements ISubscription<T> {
 /** @private */
 export class ConsoleLogger implements ILogger {
     private readonly minimumLogLevel: LogLevel;
+    public outputConsole: {
```
BrennanConroy: nit: public for testing

BrennanConroy (approved):

anurse:
Update to fix build.

src/SignalR/clients/ts/signalr/src/HubConnectionBuilder.ts
halter73: OK. I see it's only used as input to configureLogging.

So in JavaScript, you could capitalize the string however you want, but in TypeScript you're still limited to lowercase strings because configureLogging expects a `keyof typeof LogLevelNameMapping`.

src/SignalR/clients/ts/signalr/src/HubConnectionBuilder.ts
```diff
@@ -47,12 +72,15 @@ export class HubConnectionBuilder {
      * @param {LogLevel | ILogger} logging An object implementing the {@link @aspnet/signalr.ILogger} interface or {@link @aspnet/signalr.LogLevel}.
      * @returns The {@link @aspnet/signalr.HubConnectionBuilder} instance, for chaining.
      */
-    public configureLogging(logging: LogLevel | ILogger): HubConnectionBuilder;
-    public configureLogging(logging: LogLevel | ILogger): HubConnectionBuilder {
+    public configureLogging(logging: LogLevel | keyof typeof LogLevelNameMapping | ILogger): HubConnectionBuilder;
+    public configureLogging(logging: LogLevel | keyof typeof LogLevelNameMapping | ILogger): HubConnectionBuilder {
```
halter73: Update doc comments to indicate that a string, or more specifically a `keyof typeof LogLevelNameMapping`.

src/SignalR/clients/ts/signalr/src/HubConnectionBuilder.ts
```diff
@@ -13,6 +13,31 @@ import { JsonHubProtocol } from "./JsonHubProtocol";
 import { NullLogger } from "./Loggers";
 import { Arg, ConsoleLogger } from "./Utils";
 
+// tslint:disable:object-literal-sort-keys
+const LogLevelNameMapping = {
```
halter73: Does this need to be exported now that it's included in the public configureLogging type signature?

It might be better to change configureLogging to take a `keyof typeof LogLevel` `keyof typeof LogLevelNameMapping`. You can still use the LogLevelNameMapping and call toLowerCase on the argument.

As an upside, hat would be make things more consistent with the capitalization of our HubConnectionStates.

halter73: `keyof typeof LogLevel` doesn't work since we want to allow "warn" and "info". I say don't export anything new, and change the configureLogging type signature to take a string with a link to our docs from the doc comment.

src/SignalR/clients/ts/signalr/src/HubConnectionBuilder.ts
```diff
@@ -47,12 +72,15 @@ export class HubConnectionBuilder {
      * @param {LogLevel | ILogger} logging An object implementing the {@link @aspnet/signalr.ILogger} interface or {@link @aspnet/signalr.LogLevel}.
```
halter73:
```suggestion
     * @param {LogLevel | ILogger | string} logging An object implementing the {@link @aspnet/signalr.ILogger} interface or {@link @aspnet/signalr.LogLevel}.
```

src/SignalR/clients/ts/signalr/src/HubConnectionBuilder.ts
halter73:
```suggestion
    public configureLogging(logging: LogLevel | string | ILogger): HubConnectionBuilder {
```

src/SignalR/clients/ts/signalr/src/Utils.ts
anurse:
```suggestion
	// Public for testing purposes.
    public outputConsole: {
```

anurse: Oh man, that worked. Cool!

anurse:
Feedback applied!

src/SignalR/clients/ts/signalr/src/HubConnectionBuilder.ts
```diff
@@ -44,15 +69,18 @@ export class HubConnectionBuilder {
 
     /** Configures custom logging for the {@link @aspnet/signalr.HubConnection}.
      *
-     * @param {LogLevel | ILogger} logging An object implementing the {@link @aspnet/signalr.ILogger} interface or {@link @aspnet/signalr.LogLevel}.
+     * @param {LogLevel | string | ILogger} logging A {@link @aspnet/signalr.LogLevel}, a string representing a LogLevel, or an object implementing the {@link @aspnet/signalr.ILogger} interface. See {@link https://docs.microsoft.com/en-us/aspnet/core/signalr/configuration#configure-logging|the documentation for client logging configuration} for more details.
      * @returns The {@link @aspnet/signalr.HubConnectionBuilder} instance, for chaining.
      */
-    public configureLogging(logging: LogLevel | ILogger): HubConnectionBuilder;
-    public configureLogging(logging: LogLevel | ILogger): HubConnectionBuilder {
+    public configureLogging(logging: LogLevel | keyof typeof LogLevelNameMapping | ILogger): HubConnectionBuilder;
```
halter73: I still wonder if we need to export LogLevelNameMapping if we keep the `keyof typeof LogLevelNameMapping` in one of the public signatures. I'd just remove it.

anurse: This is just a derp. The fact that this signature even exists is an artifact of a previous change.

```suggestion
```

src/SignalR/clients/ts/signalr/src/Utils.ts
```diff
@@ -146,27 +146,35 @@ export class SubjectSubscription<T> implements ISubscription<T> {
 /** @private */
 export class ConsoleLogger implements ILogger {
     private readonly minimumLogLevel: LogLevel;
+	// Public for testing purposes.
```
halter73: This comment was indented with a tab.

```YARN : error : F:/workspace/_work/1/s/src/SignalR/clients/ts/signalr/src/Utils.ts[149, 1]: space indentation expected [F:\workspace\_work\1\s\src\SignalR\clients\ts\signalr\signalr.npmproj]``

double nit: Add newline above.

File: src/SignalR/clients/ts/signalr/src/HubConnectionBuilder.ts [hunk 10]
```diff
@@ -44,15 +69,17 @@ export class HubConnectionBuilder {
 
     /** Configures custom logging for the {@link @aspnet/signalr.HubConnection}.
      *
-     * @param {LogLevel | ILogger} logging An object implementing the {@link @aspnet/signalr.ILogger} interface or {@link @aspnet/signalr.LogLevel}.
+     * @param {LogLevel | string | ILogger} logging A {@link @aspnet/signalr.LogLevel}, a string representing a LogLevel, or an object implementing the {@link @aspnet/signalr.ILogger} interface. See {@link https://docs.microsoft.com/en-us/aspnet/core/signalr/configuration#configure-logging|the documentation for client logging configuration} for more details.
```
halter73:
```
YARN : error : F:/workspace/_work/1/s/src/SignalR/clients/ts/signalr/src/HubConnectionBuilder.ts[72, 1]: Exceeds maximum line length of 300 [F:\workspace\_work\1\s\src\SignalR\clients\ts\signalr\signalr.npmproj]
```

halter73 (approved):
LGTM. Just some tslint cleanup left.

anurse:
:'(

src/SignalR/clients/ts/signalr/src/HubConnectionBuilder.ts
```diff
@@ -44,15 +69,17 @@ export class HubConnectionBuilder {
...
      * @returns The {@link @aspnet/signalr.HubConnectionBuilder} instance, for chaining.
      */
-    public configureLogging(logging: LogLevel | ILogger): HubConnectionBuilder;
-    public configureLogging(logging: LogLevel | ILogger): HubConnectionBuilder {
+    public configureLogging(logging: LogLevel | string | ILogger): HubConnectionBuilder {
```
mikaelm12: Should the argument here also be `keyof typeof  LogLevelNameMapping` ?

mikaelm12: Ah no. realized as soon I I wrote it. It's kind of the point of the pr

src/SignalR/clients/ts/signalr/src/HubConnectionBuilder.ts
```diff
@@ -13,6 +13,31 @@ import { JsonHubProtocol } from "./JsonHubProtocol";
...
+
+function parseLogLevel(name: keyof typeof LogLevelNameMapping): LogLevel {
+    // Case-insensitive matching via lower-casing
+    // Yes, I know case-folding is a complicated problem in Unicode, but we only support
+    // the ASCII strings defined in LogLevelNameMapping anyway, so it's fine -anurse.
+    const mapping = LogLevelNameMapping[name.toLowerCase()];
+    if (typeof mapping !== "undefined") {
+        return mapping;
+    } else {
+        throw new Error(`Unknown log level: ${name}`);
```
mikaelm12: Nit: Do we want to preserve the casing on the user input here?

src/SignalR/clients/ts/signalr/src/HubConnectionBuilder.ts
anurse: That variable isn't exported from `index` so consumers won't be able to see it.

src/SignalR/clients/ts/signalr/src/HubConnectionBuilder.ts
anurse: This does preserve that casing.

src/SignalR/clients/ts/signalr/tests/HubConnectionBuilder.test.ts
```diff
@@ -111,98 +148,160 @@ describe("HubConnectionBuilder", () => {
...
 
-        // A HubConnection message
-        expect(logger.messages).toContain("Starting HubConnection.");
+        const levelNames = Object.keys(ExpectedLogLevelMappings) as Array<keyof typeof ExpectedLogLevelMappings>;
+        for (const str of levelNames) {
+            const mapped = ExpectedLogLevelMappings[str];
+            const mappedName = LogLevel[mapped];
+            it(`accepts "${str}" as an alias for LogLevel.${mappedName}`, async () => {
+                const builder = new HubConnectionBuilder()
+                    .configureLogging(str);
```
mikaelm12: It's trying to assign str to an `ILogger` instead of a string here

mikaelm12 (dismissed):
Test Failure. It looks like `configureLogging` is trying to define one of the string log level args as an ILogger which causes the test to fail.
```
  FAIL signalr/tests/HubConnectionBuilder.test.ts
    ΓùÅ Test suite failed to run
  
      signalr\tests\HubConnectionBuilder.test.ts (203,39): Argument of type '"trace" | "debug" | "info" | "information" | "warn" | "warning" | "error" | "critical" | "none"' is not assignable to parameter of type 'ILogger'.
        Type '"trace"' is not assignable to type 'ILogger'.
  
        at Object.runTsDiagnostics (common/node_modules/ts-jest/src/utils.ts:227:11)
        at process (common/node_modules/ts-jest/src/preprocessor.ts:47:5)
        at Object.process (common/node_modules/ts-jest/index.js:8:51)
        at ScriptTransformer.transformSource (common/node_modules/@jest/transform/build/ScriptTransformer.js:387:35)
        at ScriptTransformer._transformAndBuildScript (common/node_modules/@jest/transform/build/ScriptTransformer.js:457:40)
        at ScriptTransformer.transform (common/node_modules/@jest/transform/build/ScriptTransformer.js:513:25)
```

anurse:
Blocked on @mikaelm12 's change request ;).

BrennanConroy:
Unblocked now :D

anurse:
hat tip

mikaelm12:
This is why I don't like using the request changes feature lol

anurse:
I only use it when I specifically feel the need to see the result of my feedback before signing off.

Issue #8941, opened 2019-03-30
Add retry to VS installation script


dougbu:
When used in a CI build step (Install Build Tools for Visual Studio 2019), the installation part of InstallVisualStudio.ps1 sometimes silently fails. Add a retry and, if possible, display more information when problems arise.

Example build failures:
- https://dev.azure.com/dnceng/internal/_build/results?buildId=137826
- https://dev.azure.com/dnceng/internal/_build/results?buildId=137832
- https://dev.azure.com/dnceng/internal/_build/results?buildId=138515

dougbu:
f46d5b3af2f2

PR #9260, opened 2019-04-10
Ensure InstallVisualStudio.ps1 doesn't mix RTM and preview versions if both are installed

natemcmaster:
Attempting to resolve https://github.com/aspnet/AspNetCore-Internal/issues/2210


JunTaoLuo (approved):

dougbu (approved):
Hmm, guessing the agents are being upgraded again. Once again we weren't notified.

Should we address #8941 and (hopefully) find a way to get more information into the logs about any VS installation failure at the same time?

natemcmaster:
This issue is blocking P4 builds, so I don't want to address #8941 just yet. Seems this fix is sufficient to unblock P4 for now.

It appears the VS Build Tools preview version was actually 16 P4, not 16.1 previews. The agents must be stale.

jkotalik:
Hey I think this is causing my build to fail: https://dev.azure.com/dnceng/public/_build/results?buildId=150728

poke:
@jkotalik Once #9263 is merged, you should be able to rebase onto master and the builds should go through.

PR #9262, opened 2019-04-10
Produce Corresponding symbols in installer output for ANCM

jkotalik:
We don't export corresponding symbols for msis with ANCM. This is a simple approach for now.

### DIFF
--- a/src/Installers/Windows/AspNetCoreModule-Setup/ANCMIISExpressV2/AncmIISExpressV2.wixproj
+++ b/src/Installers/Windows/AspNetCoreModule-Setup/ANCMIISExpressV2/AncmIISExpressV2.wixproj
@@ -11,6 +11,8 @@
     <SuppressIces>ICE03</SuppressIces>
     <DisableGuidGeneration>true</DisableGuidGeneration>
     <SchemaVersion>2.0</SchemaVersion>
+    <TempPlatform Condition="'$(Platform)' == 'x64'">x64</TempPlatform>
+    <TempPlatform Condition="'$(Platform)' == 'x86'">Win32</TempPlatform>
   </PropertyGroup>
 
   <ItemGroup>
@@ -50,6 +52,16 @@
       <DoNotHarvest>True</DoNotHarvest>
     </ProjectReference>
   </ItemGroup>
+
+  <Target Name="CopyBuildOutputToArtifactDirectory"
+          Condition=" '$(IsProductInstaller)' == 'true' "
+          AfterTargets="Build">
+    <ItemGroup>
+      <BuildContentForAncm Include="$(RepositoryRoot)src\Servers\IIS\AspNetCoreModuleV2\AspNetCore\bin\$(Configuration)\$(TempPlatform)\*.*" />
+    </ItemGroup>
+
+    <Copy SourceFiles="@(BuildContentForAncm)" DestinationFolder="$(InstallersOutputPath)\IISExpressSymbols\$(TempPlatform)" />
+  </Target>
 
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), Directory.Build.targets))\Directory.Build.targets" />
   <Import Project="..\build\settings.props" />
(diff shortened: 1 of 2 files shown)


joeloff:
Eventually what you'd like is a release pipeline that can run off a specific signed build. We use that for some of our webtools builds today and it takes care of publishing our installers to VSDrop, archiving symbols and creating an insertion PR.

jkotalik:
@aspnet/build my build is failing for installing VS reasons I think.

jkotalik:
Nvrm, I see that https://github.com/aspnet/AspNetCore/pull/9266 will correct it.

jkotalik:
@aspnet/build this is a high priority issue for inserting ANCM into visual studio. Feedback is appreciated so I can get this merged soon.

src/Installers/Windows/AspNetCoreModule-Setup/ANCMIISExpressV2/AncmIISExpressV2.wixproj
```diff
@@ -51,6 +53,15 @@
     </ProjectReference>
   </ItemGroup>
 
+  <Target Name="CopyBuildOutputToArtifactDirectory"
```
jkotalik: Not fond of this approach where I just copy the output like this, but didn't think of a better way.

src/Installers/Windows/AspNetCoreModule-Setup/ANCMIISExpressV2/AncmIISExpressV2.wixproj
```diff
@@ -11,6 +11,8 @@
     <SuppressIces>ICE03</SuppressIces>
     <DisableGuidGeneration>true</DisableGuidGeneration>
     <SchemaVersion>2.0</SchemaVersion>
+    <TempPlatform Condition="'$(Platform)' == 'x64'">x64</TempPlatform>
+    <TempPlatform Condition="'$(Platform)' == 'x86'">Win32</TempPlatform>
```
dougbu: Does this never run on ARM?

src/Installers/Windows/AspNetCoreModule-Setup/ANCMIISExpressV2/AncmIISExpressV2.wixproj
dougbu: Suggest instead including the relevant files in the build artifacts and doing everything necessary in our build pipeline. Work w/ @aspnet/build to make the pipeline changes once the artifacts are available.

That is, change YAML rather than MSBuild.

src/Installers/Windows/AspNetCoreModule-Setup/ANCMIISExpressV2/AncmIISExpressV2.wixproj
jkotalik: Nope.

dougbu:
Should mention #9263 brings in the change you duplicated in InstallVisualStudio.ps1 That looks like it'll be in master in less than an hour. 2nd attempt already further along.

jkotalik:
Sorry was impatient and wanted to verify my change. I'll revert appropriately.

src/Installers/Windows/AspNetCoreModule-Setup/ANCMIISExpressV2/AncmIISExpressV2.wixproj
jkotalik: I disagree. If I change the YAML, I would need to point to a folder path that brings in src\servers\IIS... and do that for multiple folders. It doesn't seem like a great solution.

I also have concerns that the pdbs/dlls that we would get from the end of build won't match the build from the installers. I already tried to get pdbs from a different component than the installers and they didn't match, meaning during a complete build of aspnetcore, native dlls get rebuilt multiple times and those rebuilds actually change the dll. This is a separate concern that I would like to resolve later, but for the mean time, the safest option is to make so immediately after building the installers, we grab the dlls/pdbs and put them in the artifacts to guarantee they match.

This is a very time sensitive change. If this is a large enough concern to block merging this PR, then I can figure out a better way to fix it. But I don't see much benefit.

dougbu (approved):

dougbu:
@jkotalik https://dnceng.visualstudio.com/internal/_build/results?buildId=151176 went well and I've retried https://dev.azure.com/dnceng/public/_build/results?buildId=151165.

Plase confirm the ANCM artifacts from https://dnceng.visualstudio.com/internal/_build/results?buildId=151176 look correct.

jkotalik:
Verified that the symbols are correct and are matching. 😄

joeloff:
Let me know if you want me to archive a new build for the insertion

PR #9264, opened 2019-04-10
SignalR: Benchmark server logging

staff0rd:
Gives a better idea of what the benchmark server is doing.  Helpful for getting an overview when using multiple crankier clients to load test.

* Timestamp is time since first connection made
* Displays current connection count
* Displays peak connection count
* When connection counts are > 100, limits logging to `(connectionCount % 100 == 0)` to avoid console spam

Example from a B2s instance;

```
[00:18:23] Connected: 76200 current, 76200 peak.
[00:18:34] Connected: 76300 current, 76300 peak.
[00:18:43] Connected: 76400 current, 76400 peak.
[00:19:32] Connected: 76500 current, 76500 peak.
```

I also have [this](https://github.com/staff0rd/docker-crankier/blob/master/Start-Containers.ps1) if you wanna raise multiple crankier clients in containers.  Remember to pull them down when you're finished as they'll live (bill) forever if you don't turn them off.


src/SignalR/perf/benchmarkapps/BenchmarkServer/Hubs/EchoHub.cs
```diff
@@ -30,6 +35,29 @@ public async Task Broadcast(int duration)
             Console.WriteLine("Broadcast exited: Sent {0} messages", sent);
         }
 
+        public static string Status => $"{_connectionCount} current, {_peakConnectionCount} peak.";
+
+        public void LogConnections(string label) {
+            if (_connectionCount < 1000 || _connectionCount % 100 == 0)
```
halter73: Do you find this a bit noisy for the first 1000 connections?

staff0rd: I did this so I can get an earlier understanding of how quickly multiple workers (and clients) are connecting, including if any are yet to connect.

That said, after 100-200 that early understanding has been received and the rest up until 1000 is just noise.  I'll drop it down to 100.

src/SignalR/perf/benchmarkapps/BenchmarkServer/Hubs/EchoHub.cs
```diff
@@ -30,6 +35,29 @@ public async Task Broadcast(int duration)
...
+            {
+                var timeSinceServerStart = DateTime.Now.Subtract(_serverStart).ToString(@"hh\:mm\:ss");
+                Console.WriteLine($"[{timeSinceServerStart}] {label}: {Status}");
+            }
+        }
+
+        public override Task OnConnectedAsync() {
+            _connectionCount++;
```
halter73:
```suggestion
            Interlocked.Increment(ref _connectionCount);
```

https://docs.microsoft.com/en-us/dotnet/api/system.threading.interlocked.increment?view=netframework-4.7.2

src/SignalR/perf/benchmarkapps/BenchmarkServer/Hubs/EchoHub.cs
```diff
@@ -30,6 +35,29 @@ public async Task Broadcast(int duration)
...
+            _peakConnectionCount = Math.Max(_connectionCount, _peakConnectionCount);
+            LogConnections("Connected");
+            return Task.CompletedTask;
+        }
+
+        public override Task OnDisconnectedAsync(Exception exception) {
+            _connectionCount--;
```
halter73:
```suggestion
            Interlocked.Decrement(ref _connectionCount);
```

https://docs.microsoft.com/en-us/dotnet/api/system.threading.interlocked.decrement?view=netframework-4.7.2

src/SignalR/perf/benchmarkapps/BenchmarkServer/Hubs/EchoHub.cs
halter73: Before the if you should use Interlocked.Read to get connectionCount locally without tearing:

```
var connectionCount = Interlocked.Read(ref _connectionCount);
```

https://docs.microsoft.com/en-us/dotnet/api/system.threading.interlocked.read?view=netcore-3.0

src/SignalR/perf/benchmarkapps/BenchmarkServer/Hubs/EchoHub.cs
```diff
@@ -30,6 +35,29 @@ public async Task Broadcast(int duration)
...
+            if (_connectionCount < 1000 || _connectionCount % 100 == 0)
+            {
+                var timeSinceServerStart = DateTime.Now.Subtract(_serverStart).ToString(@"hh\:mm\:ss");
+                Console.WriteLine($"[{timeSinceServerStart}] {label}: {Status}");
+            }
+        }
+
+        public override Task OnConnectedAsync() {
+            _connectionCount++;
+            _peakConnectionCount = Math.Max(_connectionCount, _peakConnectionCount);
```
halter73:
```suggestion
            var savedPeakConnectionCount = Interlocked.Read(ref _peakConnectionCount);

            while (savedConnectionCount > savedPeakConnectionCount) {
                if (Interlocked.CompareExchange(ref _peakConnectionCount, connectionCount, savedPeakConnectionCount) == savedPeakConnectionCount) {
                    break;
                }
                else
                {
                	savedPeakConnectionCount = Interlocked.Read(ref _peakConnectionCount);
                }
            }
```

To do this safely without locking you would need to do something like the above. At this point, it's probably just easier just to track _connectionCount and _peakConnectionCount inside the lock. The lock would be acquired for such short intervals, that I doubt it would cause performance issues.

https://docs.microsoft.com/en-us/dotnet/api/system.threading.interlocked.compareexchange?view=netcore-3.0

src/SignalR/perf/benchmarkapps/BenchmarkServer/Hubs/EchoHub.cs
```diff
@@ -10,6 +10,11 @@ namespace BenchmarkServer.Hubs
 {
     public class EchoHub : Hub
     {
+        static int _connectionCount = 0;
+        static int _peakConnectionCount = 0;
```
anurse:
```suggestion
        private static int _peakConnectionCount = 0;
```

src/SignalR/perf/benchmarkapps/BenchmarkServer/Hubs/EchoHub.cs
```diff
@@ -10,6 +10,11 @@ namespace BenchmarkServer.Hubs
 {
     public class EchoHub : Hub
     {
+        static int _connectionCount = 0;
```
anurse:
```suggestion
        private static int _connectionCount = 0;
```

src/SignalR/perf/benchmarkapps/BenchmarkServer/Hubs/EchoHub.cs
staff0rd: Paused on this for a while as I needed to read up on `Interlocked`.  Even so, after making all changes (except this one), I agree that using `lock` is probably going to be neater, certainly because in the future I can foresee a `totalDisconnectedCount` or similar.

To my understanding I'll need a single lock object, and I'll use it above and in all other places that I'm currently using `Interlocked` to manipulate those `int` values.

halter73: > To my understanding I'll need a single lock object, and I'll use it above and in all other places that I'm currently using Interlocked to manipulate those int values.

Just lock all reads and writes to all the static fields you added.

If you want to go for bonus points, you could make a EchoHubConnectionCounter [singleton service](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection?view=aspnetcore-2.2) instead of using the statics and register it in the [sample's Startup.ConfigureServices](https://github.com/aspnet/AspNetCore/blob/feff89639c0ddc0b59b90d440db530f936610e8f/src/SignalR/perf/benchmarkapps/BenchmarkServer/Startup.cs#L20). You could then inject the EchoHubConnectionCounter via the EchoHub constructor.

src/SignalR/perf/benchmarkapps/BenchmarkServer/Hubs/EchoHub.cs
staff0rd: switched for `EchoHubConnectionCounter`

src/SignalR/perf/benchmarkapps/BenchmarkServer/Hubs/EchoHub.cs
staff0rd: switched for `EchoHubConnectionCounter`

staff0rd:
Updates completed, this is now pending your review @anurse, @halter73

BrennanConroy (changes requested):
I would prefer not touching the BenchmarkServer project if possible.

I'm fine with us creating a new example server project in the Crankier folder that is specific to Crankier usage.

anurse:
Yeah, it's not super clear but the BenchmarkServer app is actually used other benchmarking logic we do elsewhere. Introducing some locks and logging may adversely affect that benchmark.

Rather than an entirely separate app, a separate Hub with it's own endpoint URL would be OK with me.

staff0rd:
@anurse that will not build because of #9976.

How about `/crankier/client` and `/crankier/server`?

halter73:
@BrennanConroy @anurse  What if instead of creating yet another benchmark server app, we added an opt-in flag like `--enable-stdout-connection-counting`. Without the flag, we could register a `NoOpEchoHubConnectionCounter` which would have no measurable performance impact.

anurse:
I'm very in favor of any option that doesn't involve a whole new test app :). Command line arg sounds _great_.

staff0rd:
@BrennanConroy thoughts on options above?

BrennanConroy:
I'm still not a fan of making a precedent that we can modify the `BenchmarkServer` app as long as it "doesn't affect performance".

In your other PR you were saying you are not able to build this app because of the Directory.Build.Props file which we are planning on keeping, so I'm not sure how this will help.

staff0rd:
@anurse @halter73 What if I were to alter Crankier such that if you passed a `serve` argument it would host and expose its own Hub specific to the purposes it needs?

anurse:
Sure, that sounds like a good approach.

staff0rd:
I'll open a new PR with a different approach.

PR #9269, opened 2019-04-11
Flow ANCM version number in the build output

### LINKED ISSUE #4016: Emit ANCM version information as part of the build
In 2.2, we can modify the version of the produced internal ANCM package- https://github.com/dotnet/versions/blob/master/build-info/dotnet/product/cli/release/2.2/build.xml#L20
We'll need a new approach for 3.0 which @natemcmaster will help define.

### LINKED ISSUE #7548: Upgrade IIS C++ projects to build with platformtoolset v142 (VS2019)

```
```

analogrelay:
Fixes #4016 
Also fixes #7548 because I was getting build errors that told me I should update them to v142 anyway :)

This drops a file: `artifacts/installers/aspnetcoremodule.version` with contents like:

```
13.0.19101.0
[commit hash]
```

Still outstanding (need some assistance from @aspnet/build): working out how we could flow this forward to the final build output. This needs a little more understanding of Maestro/BAR than I have off-hand. I can add it to the manifest in [Maestro.csproj](https://github.com/aspnet/AspNetCore/blob/4964b338abe02bdbb21ff69928d38b11e85c90c9/build/Maestro/Maestro.csproj#L17) but I'm not sure of how that all flows to the final output drop.

### DIFF
--- a/eng/scripts/vs.buildtools.json
+++ b/eng/scripts/vs.buildtools.json
@@ -17,8 +17,6 @@
         "Microsoft.VisualStudio.Component.NuGet.BuildTools",
         "Microsoft.VisualStudio.Component.VC.ATL",
         "Microsoft.VisualStudio.Component.VC.Tools.x86.x64",
-        "Microsoft.VisualStudio.Component.VC.v141.ATL",
-        "Microsoft.VisualStudio.Component.VC.v141.x86.x64",
         "Microsoft.VisualStudio.Component.Windows10SDK.17134",
         "Microsoft.VisualStudio.Workload.ManagedDesktopBuildTools",
         "Microsoft.VisualStudio.Workload.MSBuildTools",
--- a/eng/scripts/vs.buildtools.json
+++ b/eng/scripts/vs.buildtools.json
@@ -17,6 +17,8 @@
         "Microsoft.VisualStudio.Component.NuGet.BuildTools",
         "Microsoft.VisualStudio.Component.VC.ATL",
         "Microsoft.VisualStudio.Component.VC.Tools.x86.x64",
+        "Microsoft.VisualStudio.Component.VC.v141.ATL",
+        "Microsoft.VisualStudio.Component.VC.v141.x86.x64",
         "Microsoft.VisualStudio.Component.Windows10SDK.17134",
         "Microsoft.VisualStudio.Workload.ManagedDesktopBuildTools",
         "Microsoft.VisualStudio.Workload.MSBuildTools",
--- a/src/Servers/IIS/AspNetCoreModuleV2/AspNetCore/AspNetCore.vcxproj
+++ b/src/Servers/IIS/AspNetCoreModuleV2/AspNetCore/AspNetCore.vcxproj
@@ -305,4 +305,12 @@
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
   <ImportGroup Label="ExtensionTargets">
   </ImportGroup>
+  <Target Name="AfterBuild">
+    <!-- Drop a file in the artifacts directory containing the ANCM version number -->
+    <ItemGroup>
+      <VersionFileContents Include="$(AspNetCoreModuleVersionMajor).$(AspNetCoreMinorVersion).$(AssemblyBuild).$(AspNetCorePatchVersion)" />
+      <VersionFileContents Include="$(CommitHash)" />
+    </ItemGroup>
+    <WriteLinesToFile File="$(InstallersOutputPath)aspnetcoremodule.version" Lines="@(VersionFileContents)" OverWrite="true" WriteOnlyWhenDifferent="True" />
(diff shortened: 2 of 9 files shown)


eng/scripts/vs.buildtools.json
```diff
@@ -17,8 +17,6 @@
         "Microsoft.VisualStudio.Component.NuGet.BuildTools",
         "Microsoft.VisualStudio.Component.VC.ATL",
         "Microsoft.VisualStudio.Component.VC.Tools.x86.x64",
-        "Microsoft.VisualStudio.Component.VC.v141.ATL",
```
jkotalik: IIRC there wasn't a v142 equivalent for this package. As long as things build correct, I think we are fine to remove this.

src/Servers/IIS/build/Build.Settings
```diff
@@ -4,7 +4,7 @@
      <SolutionDir Condition="$(SolutionDir) == '' Or $(SolutionDir) == '*Undefined*'">$(MSBuildThisFileDirectory)..\</SolutionDir>
      <Configuration Condition="'$(Configuration)'==''">Debug</Configuration>
      <Platform Condition="'$(Platform)' == ''">Win32</Platform>
-     <PlatformToolset>v141</PlatformToolset>
+     <PlatformToolset>v142</PlatformToolset>
```
jkotalik: Ah, I think I missed this one before. When you open the IIS sln now, do you get a warning about needing to upgrade to v142 still?

src/Servers/IIS/AspNetCoreModuleV2/AspNetCore/AspNetCore.vcxproj
```diff
@@ -305,4 +305,12 @@
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
   <ImportGroup Label="ExtensionTargets">
   </ImportGroup>
+  <Target Name="AfterBuild">
+    <!-- Drop a file in the artifacts directory containing the ANCM version number -->
+    <ItemGroup>
+      <VersionFileContents Include="$(AspNetCoreModuleVersionMajor).$(AspNetCoreMinorVersion).$(AssemblyBuild).$(AspNetCorePatchVersion)" />
```
jkotalik: I'm going to want to verify these numbers tomorrow. I don't believe these are correct, but want to confirm before I correct you.

eng/scripts/vs.buildtools.json
natemcmaster: "Microsoft.VisualStudio.Component.VC.ATL" is the v142 equivalent, and it's listed already. :)

src/Servers/IIS/AspNetCoreModuleV2/AspNetCore/AspNetCore.vcxproj
```diff
@@ -305,4 +305,12 @@
...
+      <VersionFileContents Include="$(CommitHash)" />
```
natemcmaster:
```suggestion
      <VersionFileContents Include="$(SourceRevisionId)" />
```
CommitHash is going to be deprecated. SourceRevisionId is a well-known, built-in property in Microsoft.Common.targets available after the  InitializeSourceControlInformation target . Example usage:

https://github.com/aspnet/AspNetCore/blob/78bc2a10b9ea617269527f4091903471a299e197/eng/targets/CSharp.Common.targets#L9-L17

natemcmaster (commented):
BAR doesn't support anything except NuGet packages right now, so there isn't really anything we can do with this file except upload it to our blob storage.

One minor tweak to make but otherwise looks fine

anurse:
> except upload it to our blob storage.

Looking at the release pipeline, it appears that will happen automatically.

anurse:
Looks like there is something broken about the C++ resolution with this fix:

```
TRACKER : error TRK0005: Failed to locate: "CL.exe". The system cannot find the file specified.
```

src/Servers/IIS/build/Build.Settings
anurse: > When you open the IIS sln now, do you get a warning about needing to upgrade to v142 still

You do not.

eng/scripts/vs.buildtools.json
anurse: Yep, I came to the same conclusion looking at the component directory: https://docs.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-community?view=vs-2019

src/Servers/IIS/AspNetCoreModuleV2/AspNetCore/AspNetCore.vcxproj
anurse: Awesome, I'm glad this is in `Microsoft.Common.targets` now!

natemcmaster (approved):

PR #9283, opened 2019-04-11
Cache OnStarting callback delegate

benaadams:
Static cache the OnStarting callback delegate

Drops this allocation:

image

/cc @rynowak


src/Mvc/Mvc.ViewFeatures/src/Filters/SaveTempDataFilter.cs
```diff
@@ -43,36 +44,37 @@ public void OnResourceExecuting(ResourceExecutingContext context)
 
             if (!context.HttpContext.Response.HasStarted)
             {
-                context.HttpContext.Response.OnStarting((state) =>
-                {
-                    var httpContext = (HttpContext)state;
-
-                    var saveTempDataContext = GetTempDataContext(context.HttpContext);
```
benaadams: Was capturing `context` even though `httpContext` was passed in as state.

davidfowl (approved):

PR #9284, opened 2019-04-11
Lazy allocate dictionary in ItemsDictionary

benaadams:
Lazy allocate dictionary in ItemsDictionary (doesn't need to allocate a dict to find out an item is not there)

### DIFF
--- a/src/Http/Http/src/Internal/ItemsDictionary.cs
+++ b/src/Http/Http/src/Internal/ItemsDictionary.cs
@@ -8,111 +8,147 @@ namespace Microsoft.AspNetCore.Http.Internal
 {
     public class ItemsDictionary : IDictionary<object, object>
     {
+        private IDictionary<object, object> _items;
+
         public ItemsDictionary()
-            : this(new Dictionary<object, object>())
-        {
-        }
+        {}
 
         public ItemsDictionary(IDictionary<object, object> items)
         {
-            Items = items;
+            _items = items;
         }
 
-        public IDictionary<object, object> Items { get; }
+        public IDictionary<object, object> Items => this;
 
         // Replace the indexer with one that returns null for missing values
         object IDictionary<object, object>.this[object key]
         {
             get
             {
-                object value;
-                if (Items.TryGetValue(key, out value))
+                if (_items != null && _items.TryGetValue(key, out var value))
                 {
                     return value;
                 }
                 return null;
             }
-            set { Items[key] = value; }
+            set
+            {
+                EnsureDictionary();
+                _items[key] = value;
+            }
         }
 
         void IDictionary<object, object>.Add(object key, object value)
         {
-            Items.Add(key, value);
+            EnsureDictionary();
+            _items.Add(key, value);
         }
 
         bool IDictionary<object, object>.ContainsKey(object key)
-        {
-            return Items.ContainsKey(key);
-        }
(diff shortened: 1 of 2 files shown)


src/Http/Http/src/Internal/ItemsDictionary.cs
```diff
@@ -8,111 +8,160 @@ namespace Microsoft.AspNetCore.Http.Internal
...
+            value = null;
+            return _items != null && _items.TryGetValue(key, out value);
         }
 
         ICollection<object> IDictionary<object, object>.Values
         {
-            get { return Items.Values; }
+            get
+            {
+                if (_items == null)
```
Tratcher:
```
EnsureDictionary()
{
 if (_items == null)...
}
```

benaadams: Done

src/Http/Http/src/Internal/ItemsDictionary.cs
```diff
@@ -8,111 +8,147 @@ namespace Microsoft.AspNetCore.Http.Internal
...
+            => _items != null && _items.Contains(item);
 
         void ICollection<KeyValuePair<object, object>>.CopyTo(KeyValuePair<object, object>[] array, int arrayIndex)
         {
-            Items.CopyTo(array, arrayIndex);
-        }
+            if (_items == null)
+            {
+                //Delegate to Empty Dictionary to do the argument checking.
+                EmptyDictionary.Collection.CopyTo(array, arrayIndex);
```
Tratcher: If the collection is empty then isn't this a no-op?

src/Http/Http/src/Internal/ItemsDictionary.cs
```diff
@@ -8,111 +8,147 @@ namespace Microsoft.AspNetCore.Http.Internal
...
+            value = null;
+            return _items != null && _items.TryGetValue(key, out value);
         }
 
         ICollection<object> IDictionary<object, object>.Values
         {
-            get { return Items.Values; }
+            get
+            {
+                EnsureDictionary();
```
davidfowl: Why allocate here if the inner dictionary is null?

src/Http/Http/src/Internal/ItemsDictionary.cs
```diff
@@ -8,111 +8,147 @@ namespace Microsoft.AspNetCore.Http.Internal
...
-            return Items.ContainsKey(key);
-        }
+            => _items != null && _items.ContainsKey(key);
 
         ICollection<object> IDictionary<object, object>.Keys
         {
-            get { return Items.Keys; }
+            get
+            {
+                EnsureDictionary();
```
davidfowl: Why allocate here if the inner dictionary is null?

src/Http/Http/src/Internal/ItemsDictionary.cs
benaadams: Mistakenly thought the collections were mutable

benaadams: Changed

src/Http/Http/src/Internal/ItemsDictionary.cs
benaadams: Changed

davidfowl (approved):

PR #9288, opened 2019-04-11
Make Identity UI sample work

javiercn:
@HaoK


javiercn:
Regenerated the migrations and added EF tool locally.

HaoK (approved):

PR #9293, opened 2019-04-11
[Https][Tooling] Add digital signature to the KeyUsage of the HTTPS dev-cert

javiercn:


analogrelay (approved):
Looks OK to me. If the user had the previous cert installed, what process would they go through to update it? Will the first-run experience attempt to auto-upgrade the cert?

Tratcher (approved):

javiercn:
No, they will have to remove the very from the store and then recreate it.

We don’t want this to affect people with existing certs that don’t need it.

Specially if they trusted the cert already (it would require them to trust it again)

If we see that this becomes a problem then we can think of auto-upgrading

anurse:
My concern is that if you're running Windows 10, you'll reboot one day and your ASP.NET Core app will be broken, even if you're up-to-date on the latest SDK. That's a bad experience.

javiercn:
@anurse The counter point is that if we upgrade it automatically, then you install a new SDK in your box and it breaks the trust relationship with your cert. (And we can't remove things from the trust store without prompting, which is also annoying).

The certs lasts for 1 year and they get automatically renewed after that, so when its time you'll get an updated cert.

Lets see if this becomes a problem and then we can decide what to do.

anurse:
Fair points, I'm fine getting customer feedback on this :)

PR #9296, opened 2019-04-11
Fix resolving custom services from startup

sergeyshaykhullin:
Issue here https://github.com/aspnet/AspNetCore/issues/9294
 - Pass to `HostServiceProvider` built `IServiceProvider` from already registered services
- `HostServiceProvider.GetService` tries to resolve that services

Should i write some tests for that?


Tratcher:
This behavior change was intentional. Closing this per https://github.com/aspnet/AspNetCore/issues/9294#issuecomment-482277461. We'll add more details there.

PR #9305, opened 2019-04-11
Remove Routing pubternal

rynowak:
This change makes a bunch of random routing classes internal instead of
pubternal. Most of these have no use case and aren't valuable at all
outside of routing code.

The one exception here is that the only way to construct a
`TemplateBinder` in 1.0-2.2 is using a constructor that accepts some
pubternal types. I don't think it's a good idea to just yank this since
the usage is pretty wide.

Instead I added a factory service so you can create this, and marked the
constructor `[Obsolete]` where these types are touched. We can make the
actual breaking change in the next major release.


src/Http/Routing/src/Template/TemplateBinderFactory.cs
```diff
@@ -0,0 +1,24 @@
...
+        /// <summary>
+        /// Creates a new <see cref="TemplateBinder"/> from the provided <paramref name="template"/> and
+        /// <paramref name="defaults"/>.
+        /// </summary>
+        /// <param name="template">The route template.</param>
+        /// <param name="defaults">A collection of extra default values that do not appear in the route template.</param>
+        /// <returns>A <see cref="TemplateBinder"/>.</returns>
+        public abstract TemplateBinder Create(RouteTemplate template, RouteValueDictionary defaults);
+
+        public abstract TemplateBinder Create(RoutePattern pattern);
```
JamesNK: Add XML comments

JamesNK (approved):

src/Http/Routing/ref/Microsoft.AspNetCore.Routing.netcoreapp3.0.cs
```diff
@@ -490,71 +490,19 @@ public partial class StringRouteConstraint : Microsoft.AspNetCore.Routing.IParam
...
-    [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
-    public readonly partial struct BufferValue
-    {
-        private readonly object _dummy;
-        private readonly int _dummyPrimitive;
-        public BufferValue(string value, bool requiresEncoding) { throw null; }
-        public bool RequiresEncoding { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
-        public string Value { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
-    }
     public partial class DfaGraphWriter
```
pranavkm: Should this be `internal`? Is there a good reason to have this be part of reference assemblies?

jkotalik:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=152425

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/1791

src/Http/Routing/ref/Microsoft.AspNetCore.Routing.netcoreapp3.0.cs
rynowak: Read the comments on the class. This is intentionally pubternal.

PR #9310, opened 2019-04-12
Client to Server Streaming with IAsyncEnumerable

mikaelm12:
~Not ready for review. Just hacking something that works together.~ Time for some eyes on this. Got rid of a lot of similar code.
We're essentially adding checks for an `IAsyncEnumerable` type in the invocation params, associated themwith a stream id similarly to how we do with channels, then consume items from the stream and send them as `StreamItem`  messages.

### DIFF
--- a/src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
+++ b/src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
@@ -513,8 +513,12 @@ namespace Microsoft.AspNetCore.SignalR.Client
             var moreReaders = PackageIAsyncEnumerableParams(ref args, out var secondStreamIds);
             if (streamIds != null)
             {
-                streamIds.AddRange(secondStreamIds);
-            } else
+                if (secondStreamIds != null)
+                {
+                    streamIds.AddRange(secondStreamIds);
+                }
+            }
+            else
             {
                 streamIds = secondStreamIds;
             }
@@ -652,7 +656,18 @@ namespace Microsoft.AspNetCore.SignalR.Client
             for (var i = 0; i < args.Length; i++)
             {
                 var type = args[i].GetType();
-                if (type.GetInterfaces().Any(t => t.GetGenericTypeDefinition() == typeof(IAsyncEnumerable<>)))
+                var isStreamType = type.GetInterfaces().Any(t =>
+                {
+                    if (t.IsGenericType)
+                    {
+                        return t.GetGenericTypeDefinition() == typeof(IAsyncEnumerable<>);
+                    }
+                    else
+                    {
+                        return false;
+                    }
+                });
+                if (isStreamType)
                 {
                     if (readers == null)
                     {
--- a/src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
+++ b/src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
@@ -42,6 +42,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
         private readonly SemaphoreSlim _connectionLock = new SemaphoreSlim(1, 1);
 
         private static readonly MethodInfo _sendStreamItemsMethod = typeof(HubConnection).GetMethods(BindingFlags.NonPublic | BindingFlags.Instance).Single(m => m.Name.Equals("SendStreamItems"));
+        private static readonly MethodInfo _sendIAsyncStreamItemsMethod = typeof(HubConnection).GetMethods(BindingFlags.NonPublic | BindingFlags.Instance).Single(m => m.Name.Equals("SendIAsyncEnumerableStreamItems"));
 
         // Persistent across all connections
         private readonly ILoggerFactory _loggerFactory;
@@ -508,6 +509,18 @@ namespace Microsoft.AspNetCore.SignalR.Client
 
             var readers = PackageStreamingParams(ref args, out var streamIds);
 
+#if NETCOREAPP3_0
+            var moreReaders = PackageIAsyncEnumerableParams(ref args, out var secondStreamIds);
+            if (streamIds != null)
+            {
+                streamIds.AddRange(secondStreamIds);
+            } else
+            {
+                streamIds = secondStreamIds;
+            }
+#endif
+
+
             CheckDisposed();
             await WaitConnectionLockAsync();
 
@@ -534,6 +547,9 @@ namespace Microsoft.AspNetCore.SignalR.Client
 
             LaunchStreams(readers, cancellationToken);
 
+#if NETCOREAPP3_0
+            LaunchStreamsUsingIAsyncEnumerable(moreReaders, cancellationToken);
+#endif
             return channel;
         }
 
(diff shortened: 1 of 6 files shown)


src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -506,7 +508,7 @@ async Task OnStreamCanceled(InvocationRequest irq)
                 irq.Dispose();
             }
 
-            var readers = PackageStreamingParams(ref args, out var streamIds);
```
halter73: Why did you turn the readers local into a _readers field? It only appears to be used in PackageStreamingParams and LaunchStreams, and LaunchStreams always seems to be called directly after PackageStreamingParams, so I don't see the need.

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -625,9 +631,37 @@ private async Task SendStreamItems<T>(string streamId, ChannelReader<T> reader,
             await SendWithLock(CompletionMessage.WithError(streamId, responseError));
         }
 
+#if NETCOREAPP3_0
+        // this is called via reflection using the `_sendStreamItems` field
+        private async Task SendIAsyncEnumerableStreamItems<T>(string streamId, IAsyncEnumerable<T> stream, CancellationToken token)
+        {
+            Log.StartingStream(_logger, streamId);
+
+            var combinedToken = CancellationTokenSource.CreateLinkedTokenSource(_uploadStreamToken, token).Token;
```
halter73: It doesn't look like this combinedToken is being used yet. I'll review more once this is further along.

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -572,24 +571,31 @@ async Task OnStreamCanceled(InvocationRequest irq)
...
             {
                 var reader = kvp.Value;
 
                 // For each stream that needs to be sent, run a "send items" task in the background.
                 // This reads from the channel, attaches streamId, and sends to server.
                 // A single background thread here quickly gets messy.
+#if NETCOREAPP3_0
+                if (ReflectionHelper.IsIAsyncEnumerable(reader.GetType()))
+                {
+                    _ = _sendIAsyncStreamItemsMethod
```
halter73: Long term, we should store the Takss returned by _sendIAsyncStreamItemsMethod and _sendStreamItemsMethod and track them during connection teardown and log warnings if they don't complete within a timeout.

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
mikaelm12: This was from an intermediate state of the pr that I missed during clean up.

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -625,6 +632,36 @@ private async Task SendStreamItems<T>(string streamId, ChannelReader<T> reader,
             await SendWithLock(CompletionMessage.WithError(streamId, responseError));
         }
 
+#if NETCOREAPP3_0
+        // this is called via reflection using the `_sendStreamItems` field
```
```suggestion
        // this is called via reflection using the `_sendIAsyncStreamItemsMethod` field
```

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -625,6 +632,36 @@ private async Task SendStreamItems<T>(string streamId, ChannelReader<T> reader,
...
+        private async Task SendIAsyncEnumerableStreamItems<T>(string streamId, IAsyncEnumerable<T> stream, CancellationToken token)
+        {
+            Log.StartingStream(_logger, streamId);
+
+            var combinedCts = CancellationTokenSource.CreateLinkedTokenSource(_uploadStreamToken, token);
+
+            stream = AsyncEnumerableAdapters.MakeCancelableTypedAsyncEnumerable(stream, combinedCts);
```
BrennanConroy: Consider making a common stream method so we don't duplicate the streaming code, for example
```c#
async Task CommonStreaming(Func<CancellationTokenSource, Task> createAndConsumeStream)
{
    Log.StartingStream(_logger, streamId);
    var combinedCts = CancellationTokenSource.CreateLinkedTokenSource(_uploadStreamToken, token);

    try
    {
        await createAndConsumeStream(combinedCts);
    }
    catch (OperationCanceledException)
    {
        Log(...)
        responseError...
    }

    Log(...);
    await SendWithLock(...);
}
```

src/SignalR/clients/csharp/Client/test/FunctionalTests/HubConnectionTests.cs
```diff
@@ -661,6 +661,106 @@ public async Task CanStreamToAndFromClientInSameInvocation(string protocolName,
...
+            using (StartServer<Startup>(out var server))
+            {
+                var connection = CreateHubConnection(server.Url, path, transportType, protocol, LoggerFactory);
+                try
+                {
+                    async IAsyncEnumerable<int> clientStreamData()
+                    {
+                        for (var i = 0; i < 1000; i++)
+                        {
+                            await Task.Delay(10);
```
BrennanConroy: super nit: put the delay after the yield, on a fast machine the test will be slightly faster that way 😆

src/SignalR/clients/csharp/Client/test/FunctionalTests/Hubs.cs
```diff
@@ -43,6 +43,9 @@ public string GetCallerConnectionId()
 
         public ChannelReader<string> StreamEcho(ChannelReader<string> source) => TestHubMethodsImpl.StreamEcho(source);
 
+        public ChannelReader<int> StreamEchoInt(ChannelReader<int> source) => TestHubMethodsImpl.StreamEchoInt(source);
+
+
```
BrennanConroy: nit: remove newline

src/SignalR/clients/csharp/Client/test/FunctionalTests/Hubs.cs
```diff
@@ -212,6 +219,29 @@ public static ChannelReader<string> StreamEcho(ChannelReader<string> source)
 
             return output.Reader;
         }
+
+        public static ChannelReader<int> StreamEchoInt(ChannelReader<int> source)
+        {
+            var output = Channel.CreateUnbounded<int>();
+            _ = Task.Run(async () => {
```
BrennanConroy: nit: I think we want the `{` on a newline

src/SignalR/common/Shared/ReflectionHelper.cs
```diff
@@ -13,6 +15,13 @@ internal static class ReflectionHelper
         public static bool IsStreamingType(Type type, bool mustBeDirectType = false)
         {
             // TODO #2594 - add Streams here, to make sending files easy
+
+#if NETCOREAPP3_0
+            if (IsIAsyncEnumerable(type))
```
BrennanConroy: Remind me, what is this for?

halter73: For distinguishing streaming params vs normal params.

mikaelm12: Exactly . This is just the helper specifically for detecting IAsyncEnumerable

mikaelm12:
⬆️ 📅

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -590,31 +589,69 @@ private void LaunchStreams(Dictionary<string, object> readers, CancellationToken
...
-            Log.StartingStream(_logger, streamId);
-
-            var combinedToken = CancellationTokenSource.CreateLinkedTokenSource(_uploadStreamToken, token).Token;
 
-            string responseError = null;
-            try
+            async Task ReadChannelStream(CancellationTokenSource tokenSource)
             {
-                while (await reader.WaitToReadAsync(combinedToken))
+                while (await reader.WaitToReadAsync(token))
```
BrennanConroy: These should be using `tokenSource`

BrennanConroy (approved):
One thing needs to be fixed.

I wont make you do it in this PR, but we should consider refactoring this a little to avoid the captures in the local functions.

mikaelm12:
Hm there was a failure in one of the tests on the Mac Agent
```
Failed   Microsoft.AspNetCore.SignalR.Client.FunctionalTests.HubConnectionTests.CanCancelIAsyncEnumerableClientToServerUpload(protocolName: "messagepack", transportType: WebSockets, path: "/default")
  Error Message:
   System.Exception : 1 error(s) logged.
  Microsoft.AspNetCore.SignalR.HubConnectionHandler - ErrorProcessingRequest - TZXBaDo_sm2QkkjEa-5guQ - Error when processing requests.
  ===================
  System.Collections.Generic.KeyNotFoundException: No stream with id '1' could be found.
```
This means that the stream type wasn't detected on the client side and there the invocation didn't register the streaming id with the server.

mikaelm12:
The error seemed to be an issue with how the MessagePackHubProtocol was (or was not actually) handling StreamBindingFailureMessages. It was just throwing instead of sending the binding failure messages

mikaelm12:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=160515&view=logs

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/2306

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -590,31 +589,68 @@ private void LaunchStreams(Dictionary<string, object> readers, CancellationToken
                 // For each stream that needs to be sent, run a "send items" task in the background.
                 // This reads from the channel, attaches streamId, and sends to server.
                 // A single background thread here quickly gets messy.
+#if NETCOREAPP3_0
+                if (ReflectionHelper.IsIAsyncEnumerable(reader.GetType()))
+                {
+                    _ = _sendIAsyncStreamItemsMethod
+                        .MakeGenericMethod(reader.GetType().GetInterface("IAsyncEnumerable`1").GetGenericArguments())
```
davidfowl: @mikaelm12 is this right?

src/SignalR/clients/csharp/Client/test/FunctionalTests/HubConnectionTests.cs
```diff
@@ -661,6 +661,106 @@ public async Task CanStreamToAndFromClientInSameInvocation(string protocolName,
...
+        [LogLevel(LogLevel.Trace)]
+        public async Task CanStreamToServerWithIAsyncEnumerable(string protocolName, HttpTransportType transportType, string path)
+        {
+            var protocol = HubProtocols[protocolName];
+            using (StartServer<Startup>(out var server))
+            {
+                var connection = CreateHubConnection(server.Url, path, transportType, protocol, LoggerFactory);
+                try
+                {
+                    async IAsyncEnumerable<string> clientStreamData()
```
davidfowl: dude, casing, 😄

halter73: Is the case settled on local functions? I could see the argument for using camelCase like you would for anything else with a local scope.

davidfowl: Yes it settled it’s always pascal case because this is c# not java or JavaScript

mikaelm12: 😅

PR #9311, opened 2019-04-12
 Lazy create MVC ResourceInvoker statemachines

benaadams:
Creates quite a few for a simple sync controller method ([plaintext](https://github.com/aspnet/Benchmarks/blob/master/src/Benchmarks/Controllers/HomeController.cs#L15-L19))

+3% gain on mvc plaintext 

* ResourceInvoker faster task completion check

* Lazy create ResourceInvoker.InvokeAsync statemachine

* Lazy create ResourceInvoker.InvokeFilterPipelineAsync statemachine

* Lazy create ResourceInvoker.InvokeResultAsync statemachine

* Lazy create ResourceInvoker.InvokeNextResourceFilter statemachine

* Lazy create ResourceInvoker.InvokeNextExceptionFilterAsync statemachine

* Lazy create ResourceInvoker.InvokeAlwaysRunResultFilters statemachine

* Lazy create ResourceInvoker.InvokeResultFilters statemachine

* Lazy create ResourceInvoker.InvokeNextResultFilterAsync statemachine

* Lazy create ResourceInvoker.InvokeNextResultFilterAwaitedAsync statemachine

* Lazy create ResourceInvoker.InvokeNextResourceFilterAwaitedAsync statemachine

Before
```
 wrk -c 128 -t 4 -d 35 -H "Accept: application/json,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7" -H "Connection: keep-alive" http://127.0.0.1:5000/mvc/plaintext -s ./wrk/scripts/pipeline.lua -- 16
Pipeline depth: 16

Running 35s test @ http://127.0.0.1:5000/mvc/plaintext
  4 threads and 128 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     6.79ms    4.34ms  79.39ms   71.56%
    Req/Sec    39.97k     5.09k   61.52k    70.19%
  5571990 requests in 35.07s, 701.43MB read
Requests/sec: 158875.01
Transfer/sec:     20.00MB
```
After
```
wrk -c 128 -t 4 -d 35 -H "Accept: application/json,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7" -H "Connection: keep-alive" http://127.0.0.1:5000/mvc/plaintext -s ./wrk/scripts/pipeline.lua -- 16
Pipeline depth: 16

Running 35s test @ http://127.0.0.1:5000/mvc/plaintext
  4 threads and 128 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     6.54ms    4.27ms  59.78ms   72.49%
    Req/Sec    41.23k     5.89k   63.74k    69.81%
  5747056 requests in 35.09s, 723.47MB read
Requests/sec: 163757.40
Transfer/sec:     20.61MB
```

Before calls this elides (highlighted)

image

After calls

image

That section of stack trace is also improved

Before

image

After

image

/cc @rynowak @pranavkm

### DIFF
--- a/src/Mvc/Mvc.Core/src/Infrastructure/ResourceInvoker.cs
+++ b/src/Mvc/Mvc.Core/src/Infrastructure/ResourceInvoker.cs
@@ -1011,16 +1011,41 @@ namespace Microsoft.AspNetCore.Mvc.Infrastructure
             }
         }
 
-        private async Task InvokeAlwaysRunResultFilters()
+        private Task InvokeAlwaysRunResultFilters()
         {
-            var next = State.ResultBegin;
-            var scope = Scope.Invoker;
-            var state = (object)null;
-            var isCompleted = false;
-
-            while (!isCompleted)
+            try
             {
-                await ResultNext<IAlwaysRunResultFilter, IAsyncAlwaysRunResultFilter>(ref next, ref scope, ref state, ref isCompleted);
+                var next = State.ResultBegin;
+                var scope = Scope.Invoker;
+                var state = (object)null;
+                var isCompleted = false;
+
+                while (!isCompleted)
+                {
+                    var lastTask = ResultNext<IAlwaysRunResultFilter, IAsyncAlwaysRunResultFilter>(ref next, ref scope, ref state, ref isCompleted);
+                    if (!lastTask.IsCompletedSuccessfully)
+                    {
+                        return Awaited(this, lastTask, next, scope, state, isCompleted);
+                    }
+                }
+
+                return Task.CompletedTask;
+            }
+            catch (Exception ex)
+            {
+                // Wrap non task-wrapped exceptions in a Task, 
+                // as this isn't done automatically since the method is not async.
+                return Task.FromException(ex);
+            }
+
+            static async Task Awaited(ResourceInvoker invoker, Task lastTask, State next, Scope scope, object state, bool isCompleted)
+            {
+                await lastTask;
+
+                while (!isCompleted)
+                {
+                    await invoker.ResultNext<IAlwaysRunResultFilter, IAsyncAlwaysRunResultFilter>(ref next, ref scope, ref state, ref isCompleted);
+                }
             }
         }
 
--- a/src/Mvc/Mvc.Core/src/Infrastructure/ResourceInvoker.cs
+++ b/src/Mvc/Mvc.Core/src/Infrastructure/ResourceInvoker.cs
@@ -57,46 +57,139 @@ namespace Microsoft.AspNetCore.Mvc.Infrastructure
             _cursor = new FilterCursor(filters);
         }
 
-        public virtual async Task InvokeAsync()
+        public virtual Task InvokeAsync()
         {
+            if (_diagnosticListener.IsEnabled() || _logger.IsEnabled(LogLevel.Information))
+            {
+                return Logged(this);
+            }
+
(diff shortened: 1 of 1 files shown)


rynowak:
> +3% gain on mvc plaintext

This is super dope!

src/Mvc/Mvc.Core/src/Infrastructure/ResourceInvoker.cs
```diff
@@ -57,46 +57,139 @@ internal abstract class ResourceInvoker
...
+            }
 
-                _diagnosticListener.BeforeAction(
-                    _actionContext.ActionDescriptor,
-                    _actionContext.HttpContext,
-                    _actionContext.RouteData);
+            if (invokeException != null)
+            {
+                return Awaited(this, Task.FromException(invokeException), scope);
+            }
```
rynowak: Is there a reason not to fold this into the `catch` block?

rynowak: Is this pattern going to negatively effect the quality of callstacks?

benaadams: Good point. Its a vestige of the multiple iterations I went through to deal with the double finally block

benaadams: > Is this pattern going to negatively effect the quality of callstacks?

Yes; because local function names become garbage and there is no contract with the C# compiler that the runtime can rely on to turn them back into anything sensible https://github.com/dotnet/roslyn/issues/22857. This came up when turning async callstacks into something nice https://github.com/dotnet/coreclr/pull/14652 /cc @tmat @danmosemsft 

I can move them out into regular static methods, and they will show up sensibility if preferred?

rynowak: I'm more concerned with whether callstacks get truncated. Do you have an example of what thowing an exception from a controller looks like (developer exception page) from your current PR?

rynowak:
I also certify this change as H E L L A S P I C Y

src/Mvc/Mvc.Core/src/Infrastructure/ResourceInvoker.cs
benaadams: Interestingly its the same for
```csharp
throw new Exception();
```
and
```csharp
await Task.Delay(100);
throw new Exception();
```

image

Whereas I thought the second one would look terribad.

benaadams: In a page, again the delayed and immediate produce the same stack traces which is
image

benaadams: the other version

image

benaadams:
> I also certify this change as H E L L A S P I C Y

Can do same for ControllerActionInvoker etc if you are happy with this kinda change

rynowak:
Exceptions look good!. Yeah it would be awesome if you want to do this for the other invokers.

rynowak (approved):

benaadams:
>  Yeah it would be awesome if you want to do this for the other invokers.

ControllerInvoker https://github.com/aspnet/AspNetCore/pull/9533

Issue #8519, opened 2019-03-14
Allow configuration of ValidationVisitor.ValidateComplexTypesIfChildValidationFails via MvcOptions


bordecal:
### Is your feature request related to a problem? Please describe.
Configuration of ValidationVisitor.ValidateComplexTypesIfChildValidationFails is not currently possible without implementing a custom `IObjectModelValidator`.  Since `DefaultObjectValidator` has internal visibility, it cannot be inherited in user code in order to apply customization of this property, which means that a separate subclass of `ObjectModelValidator` must be created.  While doing so is relatively trivial (including reproducing the current behaviour of `DefaultObjectValidator`), a custom subclass would not inherit any changes made to `DefaultObjectValidator` in the future, making this a somewhat brittle approach.

### Describe the solution you'd like
The `MvcOptions` class already contains multiple properties for customizing validation behaviour.  Would it be possible to add support for toggling `ValidateComplexTypesIfChildValidationFails` as well?

### Describe alternatives you've considered
Making it possible to inherit from `DefaultObjectValidator` in user code might also be an option here, but it doesn't seem to be particularly consistent with the existing design around customizing validation behaviour.

### Additional context
N/A

mkArtakMSFT:
Thanks for contacting us, @bordecal.
Can you please provide more details regarding what you're trying to achive. It's not clear what's the scenario you need this configuration for.

bordecal:
The goal is simply to allow complete validation results to be returned to an endpoint caller under all circumstances, including when a validation failure is in part due to an invalid child object.  Setting `ValidateComplexTypesIfChildValidationFails` to `true` on any `ValidationVisitor` used in controller validation is necessary to accomplish this.

mkArtakMSFT:
That seems reasonable. Would you be willing to send us a PR for this?

bordecal:
Sure.  Is there anything I should do to indicate that I'm working on it (to avoid having someone else pick it up in parallel)?

bordecal:
@mkArtakMSFT : Is there anything I need to do to request review of the PR, or will that happen automatically?

mkArtakMSFT:
Thanks for your contribution, @bordecal!

PR #9312, opened 2019-04-12
Allow ValidationVisitor.ValidateComplexTypesIfChildValidationFails to be configured via MvcOptions

bordecal:
Add `MvcOptions.ValidateComplexTypesIfChildValidationFails` property to allow configuration of `ValidationVisitor` generated by `DefaultObjectValidator`

Addresses #8519


bordecal:
_Some notes to reviewers:_

If you have concerns about the test modifications and additions, here are a few details regarding why I chose the current approaches:

- For the integration tests, I chose to replace the inheritance approach since the main goal of this changeset is to remove the need to subclass in order to control the value of `ValidationVisitor.ValidateComplexTypesIfChildValidationFails`.  If you would like the inheritance approach to remain in the tests, I can certainly restore it and add additional tests for the non-inheritance scenario.
- For the functional tests, I originally put the non-override scenarios in `InputObjectValidationTests` and created a separate test class for the override scenarios.  However, I ended up feeling that the relationship between the tests in the two classes was less than obvious, and the extent of duplication was pretty painful as well.  I've moved the tests for both the override and non-override scenarios into a separate test class (`InputParentValidationTests`), with a nested class for each scenario.  I did a bit of duplication reduction by having each of these inherit from a common base class, but not as much as I usually would since the existing functional tests do seem to avoid aggressive duplication reduction (presumably in favour of test readability).  If you would prefer an alternate approach (e.g.: no nesting and/or no inheritance, etc.), just let me know.
- For the functional tests, I also built the request content via JSON serialization rather than a string literal for the sake of readability and ease of traceability.  On the other hand, I'm not sure how deliberate the decision to use string literals extensively in the other validation tests may have been, and I can certainly change the approach if you would prefer that string literals be used exclusively for such scenarios.

pranavkm (approved):

pranavkm:
Thanks foe the PR!

bordecal:
Thanks!

PR #9318, opened 2019-04-12
Remove target invocation exceptions from Startup invocation

davidfowl:
- This change takes advatage of the new DoNotWrapExceptions binding flag to avoid throwing a TargetInvocationException when invoking Configure/ConfigureServices and ConfigureContainer

PS: This is a better version of https://github.com/aspnet/AspNetCore/pull/7230

cc @pranavkm @rynowak (awareness of this feature, we may want to use it in more places)

### DIFF
--- a/src/Hosting/Hosting/src/Internal/MethodInfoExtensions.cs
+++ b/src/Hosting/Hosting/src/Internal/MethodInfoExtensions.cs
@@ -1,3 +1,6 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
 using System.Reflection;
 
 namespace Microsoft.AspNetCore.Hosting.Internal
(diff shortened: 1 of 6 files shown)


Tratcher (approved):

pranavkm:
MVC uses [`ObjectMethodExecutor`](https://github.com/aspnet/AspNetCore/blob/master/src/Shared/ObjectMethodExecutor/ObjectMethodExecutor.cs#L151-L189) which creates a delegate. Presumably that would not benefit from this.

src/Hosting/Hosting/src/Internal/MethodInfoExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+    internal static class MethodInfoExtensions
+    {
+        // This version of MethodInfo.Invoke removes TargetInvocationExceptions
+        public static object InvokeWithoutWrappingExceptions(this MethodInfo methodInfo, object obj, object[] parameters)
+        {
+            // These are the default arguments passed when methodInfo.Invoke(obj, parameters) are called. We do the same
+            // here but specify BindingFlags.DoNotWrapExceptions to avoid getting TAE (TargetInvocationException)
+            // methodInfo.Invoke(obj, BindingFlags.Default, binder: null, parameters: parameters, culture: null)
+
+            return methodInfo.Invoke(obj, BindingFlags.DoNotWrapExceptions, binder: null, parameters: parameters, culture: null);
```
anurse: TIL

PR #9323, opened 2019-04-12
Enable custom restore

javiercn:
This makes it so that yarn install runs at restore time on the "csproj" projects that require it.


javiercn:
The related build pull request is here https://github.com/aspnet/BuildTools/pull/956

natemcmaster (approved):

PR #9332, opened 2019-04-12
Use default VC toolset

dougbu:
- aspnet/AspNetCore-Internal#2246

### DIFF
--- a/src/Servers/IIS/AspNetCoreModuleV2/AspNetCore/AspNetCore.vcxproj
+++ b/src/Servers/IIS/AspNetCoreModuleV2/AspNetCore/AspNetCore.vcxproj
@@ -33,12 +33,14 @@
     <UseDebugLibraries>true</UseDebugLibraries>
     <PlatformToolset>v142</PlatformToolset>
     <CharacterSet>Unicode</CharacterSet>
+    <VCToolsVersion />
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'" Label="Configuration">
     <ConfigurationType>DynamicLibrary</ConfigurationType>
     <UseDebugLibraries>true</UseDebugLibraries>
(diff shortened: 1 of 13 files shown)


javiercn (approved):
Based on what the VS error tells me and what @jkotalik suggested, this should be ok.

jkotalik:
@javiercn can you try that locally to make sure it works? We won't get validation on CI with this change.

jkotalik:
From @javiercn 
> It's like 6:30pm in Spain

Aka I'll test it shortly.

javiercn:
@jkotalik You're such a :trollface:

I'll test it once its merged
😄

jkotalik:
This the only thing I got working locally. I also think it's the only way this will work on CI.

jkotalik (approved):

src/Servers/IIS/AspNetCoreModuleV2/AspNetCore/AspNetCore.vcxproj
```diff
@@ -33,26 +33,30 @@
     <UseDebugLibraries>true</UseDebugLibraries>
     <PlatformToolset>v142</PlatformToolset>
     <CharacterSet>Unicode</CharacterSet>
+    <VCToolsVersion />
```
dougbu: Lost of repetition here. Can these be in the "Globals" property group or an unconditional "Configuration" property group?

dougbu (commented):
If it works, :ok: but what about Installers\Windows\AspNetCoreModule-Setup/*/*.vcxproj ?

jkotalik:
@dougbu this is what was generated when using the Property Pages in VC++ tooling. I agree it is horrible and I can simplify it, but it is auto generated.

jkotalik:
🆙 📅

jkotalik:
This is blocking people so if this passes, I'm merging btw.

jkotalik:
gah communication errors <_<

PR #9333, opened 2019-04-12
Wait to dispose RequestAborted CTS

jkotalik:
A redo of https://github.com/aspnet/AspNetCore/pull/4447. 

When we tried this PR before, it caused a pretty nasty deadlock between Reset and Cancel in HttpProcotol. This should avoid it by capturing the CTS in a local inside of the lock and then cancelling outside of the lock.

This change will eventually be applied to IIS (and then I'd probably add a helper because the pattern will be the same). However, I want to make sure this change stabilizes on CI and doesn't have flakiness.

If this change works well, we should consider porting it back to 2.2.x.

### DIFF
--- a/src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
+++ b/src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
@@ -377,7 +377,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             _manuallySetRequestAbortToken = null;
             _preventRequestAbortedCancellation = false;
 
-            // Lock to prevent CancelRequestAbortedToken from attempting to cancel an disposed CTS.
+            // Lock to prevent CancelRequestAbortedToken from attempting to cancel a disposed CTS.
             CancellationTokenSource localAbortCts = null;
             lock (_abortLock)
             {
--- a/src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
+++ b/src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
@@ -35,7 +35,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
         private Stack<KeyValuePair<Func<object, Task>, object>> _onCompleted;
 
         private object _abortLock = new object();
-        private volatile bool _requestAborted;
+        private volatile bool _connectionAborted;
         private bool _preventRequestAbortedCancellation;
         private CancellationTokenSource _abortedCts;
         private CancellationToken? _manuallySetRequestAbortToken;
@@ -257,7 +257,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                         return new CancellationToken(false);
                     }
 
-                    if (_requestAborted)
+                    if (_connectionAborted)
                     {
                         return new CancellationToken(true);
                     }
@@ -378,14 +378,17 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             _preventRequestAbortedCancellation = false;
 
             // Lock to prevent CancelRequestAbortedToken from attempting to cancel an disposed CTS.
+            CancellationTokenSource cts = null;
             lock (_abortLock)
             {
-                if (!_requestAborted)
+                if (_abortedCts != null)
                 {
-                    _abortedCts?.Dispose();
+                    cts = _abortedCts;
                     _abortedCts = null;
                 }
             }
+
+            cts?.Dispose();
 
             if (_wrapperObjectsToDispose != null)
             {
@@ -442,9 +445,17 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
         {
             try
             {
-                _abortedCts.Cancel();
-                _abortedCts.Dispose();
-                _abortedCts = null;
+                CancellationTokenSource cts = null;
+                lock (_abortLock)
+                {
+                    if (_abortedCts != null)
+                    {
+                        cts = _abortedCts;
+                        _abortedCts = null;
+                    }
+                }
+
+                cts?.Cancel();
             }
             catch (Exception ex)
             {
@@ -456,15 +467,17 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
         {
             lock (_abortLock)
             {
-                if (_requestAborted)
+                if (_connectionAborted)
                 {
                     return;
                 }
 
-                _requestAborted = true;
+                _connectionAborted = true;
             }
 
-            if (_abortedCts != null)
+            // The check to _hasAborted is an optimization to not schedule a cancellation if the context is already disposed.
+            // In CancelRequestAbortedToken, we will relock to check again.
+            if (_abortedCts != null && !_preventRequestAbortedCancellation)
             {
                 // Potentially calling user code. CancelRequestAbortedToken logs any exceptions.
                 ServiceContext.Scheduler.Schedule(state => ((HttpProtocol)state).CancelRequestAbortedToken(), this);
@@ -481,14 +494,12 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
         {
             lock (_abortLock)
             {
-                if (_requestAborted)
+                if (_connectionAborted)
                 {
                     return;
                 }
 
                 _preventRequestAbortedCancellation = true;
-                _abortedCts?.Dispose();
-                _abortedCts = null;
             }
         }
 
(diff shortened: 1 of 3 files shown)


jkotalik:
Fixes https://github.com/aspnet/AspNetCore/issues/4422

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
```diff
@@ -378,15 +378,18 @@ public void Reset()
             _preventRequestAbortedCancellation = false;
 
             // Lock to prevent CancelRequestAbortedToken from attempting to cancel an disposed CTS.
```
Tratcher:
```suggestion
            // Lock to prevent CancelRequestAbortedToken from attempting to cancel a disposed CTS.
```

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
```diff
@@ -456,15 +476,15 @@ protected void AbortRequest()
...
                 {
                     return;
                 }
 
-                _requestAborted = true;
+                _connectionAborted = true;
             }
 
-            if (_abortedCts != null)
+            if (_abortedCts != null && !_preventRequestAbortedCancellation)
```
Tratcher: _preventRequestAbortedCancellation and _abortedCts need to be checked under the lock?

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
```diff
@@ -442,9 +445,26 @@ private void CancelRequestAbortedToken()
...
             {
-                _abortedCts.Cancel();
-                _abortedCts.Dispose();
-                _abortedCts = null;
+                CancellationTokenSource localAbortCts = null;
+                lock (_abortLock)
+                {
+                    if (_abortedCts != null)
+                    {
+                        if (!_preventRequestAbortedCancellation)
```
halter73: In the case when _preventRequestAbortedCancellation is true here, we should dispose _abortedCts (outside the lock ofc) in addition to nulling it out.

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
halter73: I agree. I was about to say the real _preventRequestAbortedCancellation and _abortedCts checks happen inside the scheduled CancelRequestAbortedToken method, but _preventRequestAbortedCancellation is resettable, so I'd feel better if we determined a `shouldScheduleCancellation` condition inside the above _abortLock.

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
```diff
@@ -442,9 +445,26 @@ private void CancelRequestAbortedToken()
...
+                        _abortedCts = null;
+                    }
+                }
+
+                // There is technically a race here where _preventRequestAbortCancellation
+                // changes to true right here and then we call cts.Cancel.
+                // However, this race already existed, where the cts was canceled just before
+                // the _preventRequestAbortedCancellation is set to true.
+                // _preventRequestAbortedCancellation is mostly a heuristic rather than a guarantee
+                // we need to follow.
```
halter73: This comment is fine for providing an explanation for your design in this PR. Generally speaking though, it's pretty obvious that RequestAborted can fire right before/as _preventRequestAbortedCancellation is set to true.

halter73: I might add a comment saying that we intentionally don't dispose canceled CTSs though.

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
halter73: I found [the comment](https://github.com/aspnet/KestrelHttpServer/pull/2977#discussion_r222387365) I made to @Tratcher the last time he brought this up. I'm not sure I was wrong, but I've learned since then to be extremely careful with this code.

jkotalik: I'll go ahead and do the local.

halter73: While you're at it move the `_preventRequestAbortedCancellation = false` in Reset() down a few lines into the _abortLock.

Like the local, I don't think it's truly necessary, but I think Reset() is the last place _preventRequestAbortedCancellation is accessed without a lock, and I'd feel feel warm and fuzzy if all access was done with a lock.

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
```diff
@@ -442,9 +445,37 @@ private void CancelRequestAbortedToken()
...
+                var shouldDispose = false;
+                var shouldCancel = false;
+                lock (_abortLock)
+                {
+                    if (_abortedCts != null)
+                    {
+                        localAbortCts = _abortedCts;
+                        _abortedCts = null;
+
+                        if (_preventRequestAbortedCancellation)
```
jkotalik: I think I need to this about this change more... one second.

jkotalik:
🆙 📅 . IMO this change is ready to go :shipit:

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
```diff
@@ -375,18 +375,21 @@ public void Reset()
             Scheme = _scheme;
 
             _manuallySetRequestAbortToken = null;
-            _preventRequestAbortedCancellation = false;
 
-            // Lock to prevent CancelRequestAbortedToken from attempting to cancel an disposed CTS.
+            // Lock to prevent CancelRequestAbortedToken from attempting to cancel a disposed CTS.
+            CancellationTokenSource localAbortCts = null;
```
halter73: nit: newline

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
```diff
@@ -442,9 +445,24 @@ private void CancelRequestAbortedToken()
         {
             try
             {
-                _abortedCts.Cancel();
-                _abortedCts.Dispose();
-                _abortedCts = null;
+                CancellationTokenSource localAbortCts = null;
+                var shouldCancel = false;
+                lock (_abortLock)
```
halter73: nit: newline

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
```diff
@@ -454,17 +472,19 @@ private void CancelRequestAbortedToken()
 
         protected void AbortRequest()
         {
+            var shouldScheduleCancellation = false;
```
halter73: nit: newline

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
```diff
@@ -442,9 +445,24 @@ private void CancelRequestAbortedToken()
...
+                    {
+                        localAbortCts = _abortedCts;
+                        _abortedCts = null;
+                        shouldCancel = true;
+                    }
+                }
+
+                // If we cancel the cts, we don't dispose as people may still be using
+                // the cts. It also isn't necessary to dispose a canceled cts.
+                if (shouldCancel)
```
halter73: nit:

```suggestion
                if (localAbortCts != null)
```

jkotalik: smart.

halter73: Or just go back to `localAbortedCts?.Cancel()`.

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
```diff
@@ -442,9 +445,24 @@ private void CancelRequestAbortedToken()
         {
             try
             {
-                _abortedCts.Cancel();
-                _abortedCts.Dispose();
-                _abortedCts = null;
+                CancellationTokenSource localAbortCts = null;
+                var shouldCancel = false;
```
halter73: shouldCancel and localAbortCts are redundant

halter73 (approved):

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
```diff
@@ -375,18 +375,22 @@ public void Reset()
...
 
-            // Lock to prevent CancelRequestAbortedToken from attempting to cancel an disposed CTS.
+            // Lock to prevent CancelRequestAbortedToken from attempting to cancel a disposed CTS.
+            CancellationTokenSource localAbortCts = null;
+
             lock (_abortLock)
             {
-                if (!_requestAborted)
+                _preventRequestAbortedCancellation = false;
+                if (_abortedCts != null)
```
Tratcher: You don't need this if check, you could always perform the inner operations.

PR #9334, opened 2019-04-12
Default IdentityUI to Bootstrap4

### LINKED ISSUE #8353: Make Bootstrap4 the default UIFramework for the Identity UI
Change this to `Bootstrap4`:

https://github.com/aspnet/AspNetCore/blob/bfec2c14be1e65f7dd361a43950d4c848ad0cd35/src/Identity/UI/src/IdentityBuilderUIExtensions.cs#L33

This will help avoid all the places in the templates where the value must be set to Bootstrap4 explicitly, such as:

https://github.com/aspnet/AspNetCore/blob/bc009fc0f180532e003b7d3e466c5a1a2456a630/src/ProjectTemplates/Web.ProjectTemplates/content/RazorPagesWeb-CSharp/Startup.cs#L73

Note: This is super Announcement-worthy, so please make sure an announcement issue gets logged for this.

So, the work here is to:
* [ ] Make the default Bootstrap4
* [ ] Remove the setting from all the templates
* [ ] Post an announcement

ryanbrandenburg:
Fixes #8353.

### DIFF
--- a/src/Identity/UI/src/IdentityBuilderUIExtensions.cs
+++ b/src/Identity/UI/src/IdentityBuilderUIExtensions.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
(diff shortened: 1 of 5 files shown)


javiercn:
Doesn't this break any test? I think there are tests that need to change as part of this fix, as we test both bootstrap3 and bootstrap4 versions, I have the feeling that changing the default will make it so that bootrstrap3 tests don't test bootrstrap3 but 4 as I imagine they rely on the defaults.

javiercn:
Other than that, looks good.

Eilon (approved):
The code change looks good. I defer to @javiercn on other matters.

ryanbrandenburg:
Which tests are you referring to? If you [mean these](https://github.com/aspnet/AspNetCore/blob/master/src/Identity/test/Identity.FunctionalTests/Bootstrap3Tests/Bootstrap3AuthorizationTests.cs) then it would seem that it's already explicit.

javiercn (approved):

PR #9336, opened 2019-04-12
Add Site Extension to patch

BrennanConroy:
Microsoft.AspNetCore.AzureAppServicesIntegration and Microsoft.AspNetCore.AzureAppServices.HostingStartup are where the Logging reference comes from so we need to patch those too.

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -76,6 +76,7 @@
     <MicrosoftExtensionsLocalizationAbstractionsPackageVersion>2.2.0</MicrosoftExtensionsLocalizationAbstractionsPackageVersion>
     <MicrosoftExtensionsLocalizationPackageVersion>2.2.0</MicrosoftExtensionsLocalizationPackageVersion>
     <MicrosoftExtensionsLoggingAbstractionsPackageVersion>2.2.0</MicrosoftExtensionsLoggingAbstractionsPackageVersion>
+    <MicrosoftExtensionsLoggingAzureAppServicesPackageVersion>2.2.5-servicing-19212-04</MicrosoftExtensionsLoggingAzureAppServicesPackageVersion>
     <MicrosoftExtensionsLoggingConfigurationPackageVersion>2.2.0</MicrosoftExtensionsLoggingConfigurationPackageVersion>
     <MicrosoftExtensionsLoggingConsolePackageVersion>2.2.0</MicrosoftExtensionsLoggingConsolePackageVersion>
     <MicrosoftExtensionsLoggingDebugPackageVersion>2.2.0</MicrosoftExtensionsLoggingDebugPackageVersion>
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -76,7 +76,6 @@
     <MicrosoftExtensionsLocalizationAbstractionsPackageVersion>2.2.0</MicrosoftExtensionsLocalizationAbstractionsPackageVersion>
     <MicrosoftExtensionsLocalizationPackageVersion>2.2.0</MicrosoftExtensionsLocalizationPackageVersion>
     <MicrosoftExtensionsLoggingAbstractionsPackageVersion>2.2.0</MicrosoftExtensionsLoggingAbstractionsPackageVersion>
-    <MicrosoftExtensionsLoggingAzureAppServicesPackageVersion>2.2.5-servicing-19212-04</MicrosoftExtensionsLoggingAzureAppServicesPackageVersion>
     <MicrosoftExtensionsLoggingConfigurationPackageVersion>2.2.0</MicrosoftExtensionsLoggingConfigurationPackageVersion>
     <MicrosoftExtensionsLoggingConsolePackageVersion>2.2.0</MicrosoftExtensionsLoggingConsolePackageVersion>
     <MicrosoftExtensionsLoggingDebugPackageVersion>2.2.0</MicrosoftExtensionsLoggingDebugPackageVersion>
--- a/build/repo.props
+++ b/build/repo.props
@@ -18,7 +18,7 @@
     <SignCheckExclusionsFile>$(RepositoryRoot)eng\signcheck.exclusions.txt</SignCheckExclusionsFile>
     <SharedSourcesFolder>$(RepositoryRoot)src\Shared\</SharedSourcesFolder>
     <SharedFxArchitecture Condition="'$(SharedFxArchitecture)' == ''">$(SharedFxRid.Substring($([MSBuild]::Add($(SharedFxRid.LastIndexOf('-')), 1))))</SharedFxArchitecture>
-    <BuildSiteExtensions>true</BuildSiteExtensions>
+    <BuildSiteExtensions Condition=" '$(VersionPrefix)' == '2.2.5' " >true</BuildSiteExtensions>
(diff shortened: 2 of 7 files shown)


pakrym:
LGTM if it works

build/dependencies.props
```diff
@@ -75,7 +75,7 @@
     <MicrosoftExtensionsLocalizationAbstractionsPackageVersion>2.2.0</MicrosoftExtensionsLocalizationAbstractionsPackageVersion>
     <MicrosoftExtensionsLocalizationPackageVersion>2.2.0</MicrosoftExtensionsLocalizationPackageVersion>
     <MicrosoftExtensionsLoggingAbstractionsPackageVersion>2.2.0</MicrosoftExtensionsLoggingAbstractionsPackageVersion>
-    <MicrosoftExtensionsLoggingAzureAppServicesPackageVersion>2.2.0</MicrosoftExtensionsLoggingAzureAppServicesPackageVersion>
+    <MicrosoftExtensionsLoggingAzureAppServicesPackageVersion>2.2.5-servicing-19212-04</MicrosoftExtensionsLoggingAzureAppServicesPackageVersion>
```
natemcmaster: To make this work with ProdCon, you'll need to also move this up into the first property group in this file https://github.com/aspnet/AspNetCore/blob/fc7ef32c73089c7db338fa0acd88c86fda7712a3/build/dependencies.props#L2-L3

build/repo.props
```diff
@@ -18,7 +18,7 @@
     <SignCheckExclusionsFile>$(RepositoryRoot)eng\signcheck.exclusions.txt</SignCheckExclusionsFile>
     <SharedSourcesFolder>$(RepositoryRoot)src\Shared\</SharedSourcesFolder>
     <SharedFxArchitecture Condition="'$(SharedFxArchitecture)' == ''">$(SharedFxRid.Substring($([MSBuild]::Add($(SharedFxRid.LastIndexOf('-')), 1))))</SharedFxArchitecture>
-    <BuildSiteExtensions>false</BuildSiteExtensions>
+    <BuildSiteExtensions>true</BuildSiteExtensions>
```
natemcmaster:
```suggestion
    <BuildSiteExtensions Condition=" '$(VersionPrefix)' == '2.2.5' " >true</BuildSiteExtensions>
```

Instead of us having to remember to revert this later, can you make this conditional?

src/SiteExtensions/LoggingAggregate/build.cmd
```diff
@@ -1,3 +1,3 @@
 @ECHO OFF
-SET RepoRoot="%~dp0..\..\.."
-%RepoRoot%\build.cmd -LockFile %RepoRoot%\korebuild-lock.txt -Path %~dp0 %*
+SET RepoRoot=%~dp0..\..\..
+%RepoRoot%\build.cmd -LockFile %RepoRoot%\korebuild-lock.txt -projects %~dp0**\*.*proj %*
```
natemcmaster: IIRC, this code still requires using `-path` not `-projects`. We didn't completely backport all the build.cmd changes to 2.x branches.

natemcmaster (changes requested):

src/SiteExtensions/LoggingAggregate/build.cmd
BrennanConroy: Wasn't working correctly for me until I made these changes :/

natemcmaster: I suspect you need the first change (remove the quotes), but not sure about the second. src/SiteExtensions/LoggingAggregate/ still has its own 'subfolder' of KoreBuild stuff, like 'src/SiteExtensions/LoggingAggregate/build/' which would only be imported when using `-path`.

BrennanConroy: `C:\Users\Brennan\.dotnet\buildtools\korebuild\2.2.1-build-20190318.1\KoreBuild.Common.targets(81,5): error MSB3191: Unable to create directory "E:\ASPNETCORE\aspnetcore\src\SiteExtensions\LoggingAggregate\"\artifacts\".`

natemcmaster: Looks a like a problem with an extra trailing backslash. How did this work before, i wonder?

build/dependencies.props
natemcmaster: Can you remove this line here now that it's been copied into the section above?

analogrelay (approved):
Ditto what @pakrym said :)

src/SiteExtensions/LoggingAggregate/build.cmd
BrennanConroy: No idea, it's all new to me

build/dependencies.props
natemcmaster: Incidentally, this change will fix https://github.com/aspnet/AspNetCore-Internal/issues/2252

anurse:
We're treating this "catch-up" changes as tell-mode (cc @Eilon) so please notify `netcoreship` about the change and that this is in regard to "re-shipping" the previous patches that were not properly shipped.

BrennanConroy:
@aspnet/build The siteextension build artifact only has microsoft.aspnetcore.azureappservices.hostingstartup as version 2.25-servicing-* and the other bits are 2.2.0. Also, it only contains 2.2.x64 and not 2.1 and 2.2.x86.

src/SiteExtensions/LoggingAggregate/build.cmd
natemcmaster: I still suspect this will fail, but if you validated it, :shipit:

natemcmaster (approved):

BrennanConroy:
@aspnet/build 
Could I get assistance for
>The siteextension build artifact only has microsoft.aspnetcore.azureappservices.hostingstartup as version 2.2.5-servicing-* and the other bits are 2.2.0.

The other two packages that are bundled as 2.2.0 are `microsoft.aspnetcore.azureappservicesintegration` and `microsoft.extensions.logging.azureappservices`. But they should also be 2.2.5-*

natemcmaster:
> The other two packages that are bundled as 2.2.0

Bundled in what? The offline package cache?

natemcmaster:
Or does the site extension re-distribute these assemblies somehow? If so, it's likely a problem with way the site extensions is getting versions during the build. @pakrym would know that dependency flow works.

dougbu:
@BrennanConroy @natemcmaster @anurse are there still open questions on this PR? Or, are we just waiting for the validation build to complete?

BrennanConroy:
I'm waiting for the build now. I think everything has been figured out.

dougbu:
@BrennanConroy something else went in. You need to resolve the conflicts in this PR.

BrennanConroy:
Yeah, but I'd like the build to go through first so I can verify the artifacts.

BrennanConroy:
Confirmed that servicing versions appear in 2.2 site extension.

Merging in ~1 hr 40 minutes when build passes.

BrennanConroy:
... another conflict...

BrennanConroy:
@aspnet/build Please merge

dougbu:
Done 😺

Issue #5810, opened 2018-07-25
Why isn't MemoryStream in DataProtectionTokenProvider disposed?


Ztarbox:
This question is a duplicate from https://github.com/aspnet/AspNetIdentity/issues/7

But since core has more contributors, I reissue it over here:
The `MemoryStream `in `GenerateAsync `and `ValidateAsync `are both not disposed.
https://github.com/aspnet/Identity/blob/c7276ce2f76312ddd7fccad6e399da96b9f6fae1/src/Identity/DataProtectionTokenProvider.cs#L72
Also the disposing by the `BinaryReader `and `BinaryWriter `is "disabled" through setting the `leaveOpen `parameter to `true`.
https://github.com/aspnet/Identity/blob/c7276ce2f76312ddd7fccad6e399da96b9f6fae1/src/Identity/DataProtectionTokenProvider.cs#L174

Shouldn't the `MemoryStream `be disposed, cause we don't need him anymore?

blowdart:
@HaoK can you review the PR please? This got lost :(

CBaud:
I'm not trying to steal @zachesposito's thunder, but jumping in since this hasn't been touched for a while.

davidfowl:
Disposing a MemoryStream is pretty much a noop.

CBaud:
> Disposing a MemoryStream is pretty much a noop.

True, but that is an implementation detail. As a class that implements IDisposable, the intended contract is for the object to get disposed when it is no longer needed. Ignoring that abstraction creates tighter coupling. Enforcing the contract makes this class more resilient against changes to the design of MemoryStream. But that's just my 2 cents.

blowdart:
Closing as we didn't take the PR because it doesn't really give us anything

PR #9339, opened 2019-04-12
Dispose MemoryStream in DataProtectorTokenProvider

CBaud:
Addresses #5810


davidfowl:
Dispose is a noop. Not sure this is required

gfoidl:
I wouldn't dispose the MS here. It makes the code* bigger, without any real win, and smaller code is preferable.

The [updated internal state](https://github.com/dotnet/coreclr/blob/ce952e77ddd2ee9c4e017770573f44dc316db6a8/src/System.Private.CoreLib/shared/System/IO/MemoryStream.cs#L131-L135) is not relevant, as the MS is used only locally and is not shared to other consumers.

\* not only the C# code, but also the IL, and hence it has more work for the JIT.

HaoK:
Agree to close @ajcvickers ?

PR #9340, opened 2019-04-13
Call DiagnosticListener.OnActivityImport before starting Activity to support sampling better

lmolkova:
 - Allows a tracing system to make sampling decision on Activity or modify it before Activity is started

See https://github.com/dotnet/corefx/issues/36349 for more info

### DIFF
--- a/src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs
+++ b/src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs
@@ -273,7 +273,7 @@ namespace Microsoft.AspNetCore.Hosting.Internal
                 }
             }
 
-            _diagnosticListener.OnActivityExport(activity, httpContext.Request);
+            _diagnosticListener.OnActivityImport(activity, httpContext);
 
             if (_diagnosticListener.IsEnabled(ActivityStartKey))
             {
--- a/src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs
+++ b/src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs
@@ -273,6 +273,8 @@ namespace Microsoft.AspNetCore.Hosting.Internal
                 }
             }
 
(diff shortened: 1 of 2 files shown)


src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs
```diff
@@ -273,6 +273,8 @@ private Activity StartActivity(HttpContext httpContext)
                 }
             }
 
+            _diagnosticListener.OnActivityExport(activity, httpContext.Request);
```
davidfowl: Should this be the HttpRequest or some anonymous objects?

lmolkova: This is intended to be request. 

Request is passed into OnActivityExport, OnActivityImport and IsEnabled callbacks. They should be used to quickly filter activities. Write events get anonymous objects.

lmolkova:
I just realized it should be OnActivityImport. Will fix by Monday

src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs
davidfowl: If you ever need anything else, it’ll be a breaking change (or I guess we can shuttle it through the HttpContext)

lmolkova: The main scenario for the callback is setting trace flags (sampling bit). All info to make the decision should be already set on Activity (trace context). 

Even request is optional. There is not much info available at this point except the Request and it's hard to imagine a legit scenario where we'll need more.

lmolkova: Ok, I changed the extra argument in OnActivityImport to be HttpContext (not request) just in case we'll ever need to extend it

davidfowl (approved):

PR #9341, opened 2019-04-13
Allow headers to match on ReferenceEquals before OrdinalIgnoreCase

benaadams:
Provide a performance advantage to using the `Microsoft.Net.Http.Headers.HeaderNames` constants.

Change `HeaderNames` to static properties rather than string constants; so reference equality works; prior to falling back to an OrdinalIgnoreCase characters comparison. (string constants embed directly in the calling assembly so don't end up being reference equal)

Use these properties in Kestrel header methods rather than its own string constants so the reference equality works there.

ResponseHeaderCollectionBenchmark (Common is the main change +18%)
```diff
      Method |                 Type |        Mean |         Op/s | Allocated |
 ----------- |--------------------- |------------:|-------------:|----------:|
- SetHeaders |               Common |   889.92 ns |  1,123,698.8 |       0 B |
+ SetHeaders |               Common |   751.84 ns |  1,330,077.7 |       0 B |
- SetHeaders |  ContentLengthString |    51.64 ns | 19,363,867.1 |       0 B |
+ SetHeaders |  ContentLengthString |    51.90 ns | 19,268,524.6 |       0 B |
- SetHeaders |            Plaintext |    55.45 ns | 18,034,860.0 |       0 B |
+ SetHeaders |            Plaintext |    51.40 ns | 19,455,753.2 |       0 B |
- SetHeaders |              Unknown | 1,539.01 ns |    649,770.4 |       0 B |
+ SetHeaders |              Unknown | 1,589.23 ns |    629,237.4 |       0 B |
```

The inlined string property becomes a constant mov at Jit time e.g.
```csharp
if (ReferenceEquals(HeaderNames.Host, key))
```
becomes
```asm
G_M13073_IG06:
       mov      rdx, 0x15B8E12EBA8        ; HeaderNames:get_Host():ref
       cmp      gword ptr [rdx], rdi      ; ReferenceEquals
       jne      SHORT G_M13073_IG07
```

For plaintext with the default builder (including host filtering)

Before
image

After
image

----
Background

Noticed this while looking at something else https://github.com/aspnet/AspNetCore/pull/9359

image

If you use the `HeaderNames` class they don't match with reference equality e.g.
```csharp
context.Request.Headers[HeaderNames.Host]
```
as they are defined in a different assembly to the string constants used in Kestrel; so it moves to an OrdinalIgnoreCase compare. Using the same `HeaderNames` header names in Kestrel means it will first match using [ReferenceEquals in coreclr](https://github.com/dotnet/coreclr/blob/0fed62e68d1f8f8a756ec8ba5348bfde61057666/src/System.Private.CoreLib/shared/System/String.Comparison.cs#L638-L644) 
```
public bool Equals(string? value, StringComparison comparisonType)
{
    if (object.ReferenceEquals(this, value))
    {
        CheckStringComparison(comparisonType);
        return true;
    }
```
and won't need to fallback to the `OrdinalIgnoreCase` when `HeaderNames` is used e.g when setting
```csharp
response.ContentType = "text/plain";
```
as [DefaultHttpResponse](https://github.com/aspnet/AspNetCore/blob/e04c79bd3dfc0634af1ac6b578124630a5be247d/src/Http/Http/src/Internal/DefaultHttpResponse.cs#L89-L99) also uses these constants
```csharp
set
{
    if (string.IsNullOrEmpty(value))
    {
        HttpResponseFeature.Headers.Remove(HeaderNames.ContentType);
    }
    else
    {
        HttpResponseFeature.Headers[HeaderNames.ContentType] = value;
    }
}
```


benaadams:
Need to investigate more as they are consts in the ref assembly so it just might embed them and not work.

benaadams:
Works now, though the string properties don't inline https://github.com/dotnet/coreclr/issues/23969; however reference equality works without needing to fall back to the OrdinalIgnoreCase compare.

benaadams:
ResponseHeaderCollectionBenchmark
```diff
      Method |                 Type |        Mean |         Op/s | Allocated |
 ----------- |--------------------- |------------:|-------------:|----------:|
- SetHeaders |               Common |   889.92 ns |  1,123,698.8 |       0 B |
+ SetHeaders |               Common |   751.84 ns |  1,330,077.7 |       0 B |
- SetHeaders |  ContentLengthString |    51.64 ns | 19,363,867.1 |       0 B |
+ SetHeaders |  ContentLengthString |    51.90 ns | 19,268,524.6 |       0 B |
- SetHeaders |            Plaintext |    55.45 ns | 18,034,860.0 |       0 B |
+ SetHeaders |            Plaintext |    51.40 ns | 19,455,753.2 |       0 B |
- SetHeaders |              Unknown | 1,539.01 ns |    649,770.4 |       0 B |
+ SetHeaders |              Unknown | 1,589.23 ns |    629,237.4 |       0 B |
```

src/Mvc/test/WebSites/BasicWebSite/RequestIdMiddleware.cs
```diff
@@ -26,7 +27,7 @@ public Task Invoke(HttpContext context)
                 throw new InvalidOperationException("RequestId should be null here");
             }
 
-            var requestId = context.Request.Headers["RequestId"];
```
benaadams: Note: this was `"RequestId"` but `HostingApplicationDiagnostics.cs` uses `"Request-Id"`; now they are both using the same value via `HeaderNames.RequestId` (of `"Request-Id"`)

src/Http/Headers/ref/Microsoft.Net.Http.Headers.netcoreapp3.0.cs
```diff
@@ -118,74 +118,87 @@ public partial class EntityTagHeaderValue
     }
     public static partial class HeaderNames
     {
-        public const string Accept = "Accept";
```
davidfowl: I think this is a breaking change? These can't be used in attributes anymore?

Tratcher: Indeed. Do we have any examples where it's relevant?

benaadams: Was one in http2 tests; though that was testing the headers

image

Wouldn't think it was common?

benaadams: Suppose its a choice:

* Fast header string matching (extremely common; used in every request, 1258 uses in ASP.NET Core itself)

* Use in attributes (extremely uncommon, 2 test uses in ASP.NET Core itself and those testing http/2 protocol psudeo headers)

davidfowl: It’s not common no. The only other thing is that I’d like to eventually burn the headers assembly to the ground

halter73: Furthermore, I don't think this would be a binary breaking change since anything that referenced the constant would still have the constant burned into the assembly without ever actually referencing the HeaderNames class in the IL.

src/Http/Headers/src/HeaderNames.cs
```diff
@@ -5,73 +5,170 @@ namespace Microsoft.Net.Http.Headers
...
+        private readonly static string _userAgent = "User-Agent";
+        private readonly static string _vary = "Vary";
+        private readonly static string _via = "Via";
+        private readonly static string _warning = "Warning";
+        private readonly static string _webSocketSubProtocols = "Sec-WebSocket-Protocol";
+        private readonly static string _wWWAuthenticate = "WWW-Authenticate";
+        private readonly static string _xFrameOptions = "X-Frame-Options";
+
+        // Use properties rather than constants so ReferenceEquals works
+        public static string Accept => _accept;
```
halter73: Any particular reason not to just use public static fields directly?

src/Servers/Kestrel/shared/KnownHeaders.cs
```diff
@@ -94,7 +94,11 @@ static KnownHeaders()
                 "Translate",
                 "User-Agent",
                 "DNT",
-                "Upgrade-Insecure-Requests"
+                "Upgrade-Insecure-Requests",
+                "Request-Id",
+                "Correlation-Context",
+                "TraceParent",
+                "TraceState"
```
halter73: So I see you *did* come up with a few more non-HTTP/2 headers. Do these make sense to you @Tratcher?

src/Servers/Kestrel/shared/KnownHeaders.cs
```diff
@@ -647,27 +651,43 @@ protected override bool TryGetValueFast(string key, out StringValues value)
...
-                            {{
-                                value = _headers._{header.Identifier};
-                                return true;
-                            }}
-                            return false;")}
-                        }}")}
-                    }}
-                    break;")}
+                {{{Each(byLength.OrderBy(h => !h.PrimaryHeader), header => $@"
+                    if (ReferenceEquals(HeaderNames.{header.Identifier}, key))
```
halter73: Is manually calling object.ReferenceEquals and then following up with string.Equals StringComparison.OrdinalIgnoreCase really faster than doing the string comparison up front? If so, shouldn't we just add an object.ReferenceEquals fast path to the start of string.Equals? Or at the very least add a helper method to Kestrel that does this?

Also, AFIACT the body of this if block is identical to the body of the string.Equals if block below. Wouldn't it be better to just add `|| HeaderNames.{header.Identifier}.Equals(key, StringComparison.OrdinalIgnoreCase)` to this if condition even if it isn't part of some helper method?

src/Servers/Kestrel/shared/KnownHeaders.cs
```diff
@@ -698,62 +727,95 @@ protected override bool AddValueFast(string key, StringValues value)
...
+                        {{
+                            {header.ClearBit()};
+                            _headers._{header.Identifier} = default(StringValues);{(header.EnhancedSetter == false ? "" : $@"
+                            _headers._raw{header.Identifier} = null;")}
+                            return true;
+                        }}
+                        return false;")}
+                    }}")}
+    {Each(byLength.OrderBy(h => !h.PrimaryHeader), header => $@"
+                    if (HeaderNames.{header.Identifier}.Equals(key, StringComparison.OrdinalIgnoreCase))
```
halter73: Everywhere there are two if blocks in a row with identical bodies and one checking object.ReferenceEquals and the other checking string.Equals, I have the same questions I did for TryGetValueFast.

src/Servers/Kestrel/shared/KnownHeaders.cs
benaadams: They are from hosting as it was checking each header that was the same length with IgnoreOrdinalCase for
```csharp
if (!headers.TryGetValue(TraceParentHeaderName, out var correlationId))
{
    headers.TryGetValue(RequestIdHeaderName, out correlationId);
```

src/Middleware/Session/ref/Microsoft.AspNetCore.Session.csproj
```diff
@@ -10,5 +10,6 @@
     <Reference Include="Microsoft.Extensions.Caching.Abstractions"  />
     <Reference Include="Microsoft.Extensions.Logging.Abstractions"  />
     <Reference Include="Microsoft.Extensions.Options"  />
+    <Reference Include="Microsoft.Net.Http.Headers"  />
```
Tratcher: A direct reference shouldn't be needed here, you already added it to Http.Abstractions

Tratcher (commented):
@anurse @Eilon for breaking change awareness.

src/Servers/Kestrel/shared/KnownHeaders.cs
benaadams: string.Equals does have a ReferenceEquals at the start (though its uninlinable since its kinda chunky)

The two issues this resolves are
1. Quick inlined reference check
2. Check all the references first before checking IgnoreOrdinalCase

So if there were 3 headers of that length; and it was the 3rd header it would do

ReferenceEquals, IgnoreOrdinalCase, ReferenceEquals, IgnoreOrdinalCase, ReferenceEquals => match

With this approach it instead does

ReferenceEquals, ReferenceEquals, ReferenceEquals => match

Only if none of the ReferenceEquals match does it then start doing IgnoreOrdinalCase. 

Noticed this issue after adding `TraceParent` and trying to workout why `ContentMD5` was being tested IgnoreOrdinalCase.

src/Servers/Kestrel/shared/KnownHeaders.cs
benaadams: As above. Checks all the ReferenceEquals first; then does OrdinalIgnoreCase; rather than interleaving the IgnoreCase checks

src/Http/Headers/src/HeaderNames.cs
benaadams: 🤔 Guess not

benaadams: Or initialise the static properties directly

benaadams: Changed

src/Middleware/Session/ref/Microsoft.AspNetCore.Session.csproj
benaadams: Changed

benaadams:
I may have sent these changes from the future?
```
F:\workspace\_work\1\s\build\tasks\RepoTasks.csproj :
 error MSB4236: The SDK 'Microsoft.NET.Sdk' specified could not be found.
```

benaadams:
retriggered CI

src/Http/Headers/ref/Microsoft.Net.Http.Headers.netcoreapp3.0.cs
anurse: @halter73 is correct. We should announce it as breaking but I don't have a problem taking it.

Eilon: I did a quick search and it looks like this API is used at least a little bit by "the Internet". Should we keep it around to avoid the breaking change?

(I agree it's only a source code breaking change, but that's still a breaking change.)

Tratcher: Keeping it would involve duplication, which would get pretty confusing.

benaadams: To caveat the scope, its source code breaking if used in attributes or other string const initalization; otherwise isn't source code breaking.

Eilon: Oh I didn't see the whole change. I think this change is fine then. The few cases I found on the internet were not in attribute cases where a const is required. They were just regular imperative code. Carry on.

halter73 (approved):

src/Middleware/WebSockets/ref/Microsoft.AspNetCore.WebSockets.netcoreapp3.0.cs
```diff
@@ -41,14 +41,8 @@ public static partial class Constants
     {
         public static partial class Headers
         {
-            public const string Connection = "Connection";
             public const string ConnectionUpgrade = "Upgrade";
-            public const string SecWebSocketAccept = "Sec-WebSocket-Accept";
```
jkotalik: This seems like a slightly worse breaking change because it straight up removes the constants. However, it is in the internal namespace. Can you try making this class internal too?

jkotalik (approved):

src/Middleware/WebSockets/ref/Microsoft.AspNetCore.WebSockets.netcoreapp3.0.cs
benaadams: Internalized

benaadams:
Build: Windows x64/x86 Job failed on a YARN error (transient)

Eilon:
I think @javiercn was looking at YARN reliability.

benaadams:
> I think @javiercn was looking at YARN reliability.

Can you restart that job? 😉 I only have the power to restart everything, not an individual item.

halter73:
> Can you restart that job? 😉 I only have the power to restart everything, not an individual item.

Generally I wait until all the builds are complete before triaging and restarting the failed build. I'm not sure if you can restart just the failed build before then. I'm setting a reminder to requeue once this is finished.

javiercn:
> I think @javiercn was looking at YARN reliability.

Yeah, but this PR needs to be rebased on top of master. I just moved the yarn install for all projects to the beginning of the build and made them happen sequentially, that seems to have fixed the issue you are seeing.

halter73:
> Yeah, but this PR needs to be rebased on top of master.

Doesn't our PR validation automatically rebase on the target branch?

javiercn:
> Doesn't our PR validation automatically rebase on the target branch?

It doesn't AFAIK. It simply checks out the repo/branch and builds it, you still have to deal with merges/rebases.

halter73:
Looks good to go. I'll merge this tomorrow so I remember to make the breaking change announcement at the same time.

anurse:
> It doesn't AFAIK. It simply checks out the repo/branch and builds it, you still have to deal with merges/rebases.

It doesn't rebase, but it **does** use the "merge ref" from GitHub (`refs/pull/9341/merge` for example) which contains the **merge** of this PR and `master`. Though, I don't know when GitHub updates this ref (do they do it for every `master` commit or just for every commit to the PR).

anurse:
At a minimum, it's safe to assume the build is incorporating all changes from the most recent commit to the PR branch **and** all changes that were in `master` **at the time the most recent PR commit was made**

Eilon:
Can probably close & re-open the PR to get the master ref to update and re-try?

benaadams:
@davidfowl merge conflicted me

benaadams:
merge conflict resolved

halter73:
aspnet/Announcements#356

src/Http/Headers/ref/Microsoft.Net.Http.Headers.netcoreapp3.0.cs
NickCraver: We use these quite a bit in `case` statements for what it's worth...though I guess not anymore. Bleh.

benaadams: Sorry :-/

PR #9342, opened 2019-04-13
Support plain text exception formatting

davidfowl:
- If there's no Accept: text/html then print the exception.ToString as plaintext

#8536

Here's an example of the same error page after the change, hitting it from postman and from the browser.

image

image

### DIFF
--- a/src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
+++ b/src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using System.Text;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Diagnostics.Internal;
@@ -135,7 +136,18 @@ namespace Microsoft.AspNetCore.Diagnostics
             if (acceptHeader != null && !acceptHeader.Contains("text/html", StringComparison.OrdinalIgnoreCase))
             {
                 httpContext.Response.ContentType = "text/plain";
-                return httpContext.Response.WriteAsync(errorContext.Exception.ToString());
+
+                var sb = new StringBuilder();
+                sb.AppendLine(errorContext.Exception.ToString());
+                sb.AppendLine();
+                sb.AppendLine("HEADERS");
+                sb.AppendLine("=======");
+                foreach (var pair in httpContext.Request.Headers)
+                {
+                    sb.AppendLine($"{pair.Key} = {pair.Value}");
+                }
+
+                return httpContext.Response.WriteAsync(sb.ToString());
             }
 
             if (errorContext.Exception is ICompilationException compilationException)
--- a/src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
+++ b/src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
@@ -23,7 +23,7 @@ using Microsoft.Net.Http.Headers;
 namespace Microsoft.AspNetCore.Diagnostics
 {
     /// <summary>
-    /// Captures synchronous and asynchronous exceptions from the pipeline and generates HTML error responses.
+    /// Captures synchronous and asynchronous exceptions from the pipeline and generates error responses.
     /// </summary>
     public class DeveloperExceptionPageMiddleware
     {
--- a/src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
+++ b/src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
@@ -146,7 +146,7 @@ namespace Microsoft.AspNetCore.Diagnostics
                 sb.AppendLine("=======");
                 foreach (var pair in httpContext.Request.Headers)
                 {
-                    sb.AppendLine($"{pair.Key} = {pair.Value}");
+                    sb.AppendLine($"{pair.Key}: {pair.Value}");
                 }
 
                 return httpContext.Response.WriteAsync(sb.ToString());
--- a/src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
+++ b/src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
@@ -17,6 +17,7 @@ using Microsoft.Extensions.FileProviders;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Microsoft.Extensions.StackTrace.Sources;
+using Microsoft.Net.Http.Headers;
 
 namespace Microsoft.AspNetCore.Diagnostics
 {
@@ -127,13 +128,22 @@ namespace Microsoft.AspNetCore.Diagnostics
         // Assumes the response headers have not been sent.  If they have, still attempt to write to the body.
         private Task DisplayException(ErrorContext errorContext)
         {
-            var compilationException = errorContext.Exception as ICompilationException;
-            if (compilationException != null)
+            var httpContext = errorContext.HttpContext;
+            var acceptHeader = httpContext.Request.Headers[HeaderNames.Accept].ToString();
+
+            // If the client does not ask for HTML just format the exception as plain text
+            if (acceptHeader != null && !acceptHeader.Contains("text/html", StringComparison.OrdinalIgnoreCase))
             {
-                return DisplayCompilationException(errorContext.HttpContext, compilationException);
+                httpContext.Response.ContentType = "text/plain";
+                return httpContext.Response.WriteAsync(errorContext.Exception.ToString());
             }
 
-            return DisplayRuntimeException(errorContext.HttpContext, errorContext.Exception);
+            if (errorContext.Exception is ICompilationException compilationException)
+            {
+                return DisplayCompilationException(httpContext, compilationException);
+            }
+
+            return DisplayRuntimeException(httpContext, errorContext.Exception);
         }
 
         private Task DisplayCompilationException(
--- a/src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
+++ b/src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
@@ -135,7 +135,7 @@ namespace Microsoft.AspNetCore.Diagnostics
             var acceptHeader = headers.Accept;
 
             // If the client does not ask for HTML just format the exception as plain text
-            if (acceptHeader != null && !acceptHeader.Any(h => h.IsSubsetOf(_textPlainMediaType)))
+            if (acceptHeader == null || !acceptHeader.Any(h => h.IsSubsetOf(_textPlainMediaType)))
             {
                 httpContext.Response.ContentType = "text/plain";
 
--- a/src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
+++ b/src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
@@ -34,6 +34,7 @@ namespace Microsoft.AspNetCore.Diagnostics
         private readonly DiagnosticSource _diagnosticSource;
         private readonly ExceptionDetailsProvider _exceptionDetailsProvider;
         private readonly Func<ErrorContext, Task> _exceptionHandler;
+        private static readonly MediaTypeHeaderValue _textPlainMediaType = new MediaTypeHeaderValue("text/html");
 
         /// <summary>
         /// Initializes a new instance of the <see cref="DeveloperExceptionPageMiddleware"/> class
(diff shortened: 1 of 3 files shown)


src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
```diff
@@ -127,13 +128,22 @@ public async Task Invoke(HttpContext context)
         // Assumes the response headers have not been sent.  If they have, still attempt to write to the body.
         private Task DisplayException(ErrorContext errorContext)
         {
-            var compilationException = errorContext.Exception as ICompilationException;
-            if (compilationException != null)
+            var httpContext = errorContext.HttpContext;
+            var acceptHeader = httpContext.Request.Headers[HeaderNames.Accept].ToString();
+
+            // If the client does not ask for HTML just format the exception as plain text
+            if (acceptHeader != null && !acceptHeader.Contains("text/html", StringComparison.OrdinalIgnoreCase))
```
rynowak: yeah but what if the accept header contains `some-cool-text/html` :trollface:  

It's wired to not parse this properly if you're going to look at it.

src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
```diff
@@ -127,13 +128,22 @@ public async Task Invoke(HttpContext context)
...
+            {
+                httpContext.Response.ContentType = "text/plain";
+                return httpContext.Response.WriteAsync(errorContext.Exception.ToString());
+            }
+
+            if (errorContext.Exception is ICompilationException compilationException)
             {
-                return DisplayCompilationException(errorContext.HttpContext, compilationException);
+                return DisplayCompilationException(httpContext, compilationException);
```
rynowak: Is the reason not to go down this code path that it's pretty hardcoded to HTML?

OR that the scenario doesn't match? Right now compilation exception => views => browser...

src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
davidfowl: Don’t care. We can complicate it when the bugs come in (which they won’t 😬)

src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
davidfowl: I didn't think it warranted special formatting (especially since runtime compilation is off by default). Do you think we should format compilation errors in plaintext differently? Bare in mind you’d only hit this if you were using test server or postman making a request for a view that wasn’t already compiled *and* there happens to be an error.

src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
```diff
@@ -127,13 +128,22 @@ public async Task Invoke(HttpContext context)
...
-            var compilationException = errorContext.Exception as ICompilationException;
-            if (compilationException != null)
+            var httpContext = errorContext.HttpContext;
+            var acceptHeader = httpContext.Request.Headers[HeaderNames.Accept].ToString();
+
+            // If the client does not ask for HTML just format the exception as plain text
+            if (acceptHeader != null && !acceptHeader.Contains("text/html", StringComparison.OrdinalIgnoreCase))
+            {
+                httpContext.Response.ContentType = "text/plain";
+                return httpContext.Response.WriteAsync(errorContext.Exception.ToString());
```
Tratcher: What happened to the idea of using the same model and rendering most of the same content but as text?

davidfowl: I didn't see the benefit.

Tratcher: Of including the headers, cookies, etc? They're just as relevant to the API scenario.

davidfowl: So you want to dump all of that information verbatim split by new lines or something? I don't have a problem with that but maybe instead of dumping all of that information, we can decide what is relevant (and doesn't come back in the HTTP response itself).

Tratcher: Right. We do something similar to debug reverse proxies, IIS, etc..
https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/proxy-load-balancer?view=aspnetcore-2.2#troubleshoot
https://github.com/aspnet/AspNetCore/blob/c6df1b8a14858ede60ccb530022f9201511a1c40/src/Servers/IIS/IIS/samples/NativeIISSample/Startup.cs#L32-L101

Note the request headers are relevant because they can be transformed by intermediaries and middleware. You want to know what the server saw.

davidfowl: Updated it to add request headers

davidfowl:
🆙 📅

src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
pranavkm: I missed the previous conversation on this. Fine, I can see why this isn't absolutely necessary.

src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
Tratcher: Instant debt. Just do it right the first time.

src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
```diff
@@ -127,13 +129,33 @@ public async Task Invoke(HttpContext context)
...
+                httpContext.Response.ContentType = "text/plain";
+
+                var sb = new StringBuilder();
+                sb.AppendLine(errorContext.Exception.ToString());
+                sb.AppendLine();
+                sb.AppendLine("HEADERS");
+                sb.AppendLine("=======");
+                foreach (var pair in httpContext.Request.Headers)
+                {
+                    sb.AppendLine($"{pair.Key} = {pair.Value}");
```
Tratcher:
```suggestion
                    sb.AppendLine($"{pair.Key}: {pair.Value}");
```

src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
Eilon: Yeah come on please fix this.

src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
davidfowl: Yes we discussed it offline. If it becomes useful in the future then we can do so.

src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
davidfowl: Just to be pedantic since this change is taking longer than it needs to. What's wrong with `=`?

Tratcher: It's just not the expected format for HTTP headers.

src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
davidfowl: DONE! Fixed for future browsers we can't predict will ask for html.

src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
davidfowl: It's ok this isn't an HTTP response, it's a plaintext rendering so I'm going to leave it as `=`

pranavkm (approved):

rynowak (approved):

src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
Tratcher: For more examples the chrome and IE dev tools also render it as `header: value`. That's the layout people are used to seeing headers in. `header = value` looks like query parameters.

src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
```diff
@@ -127,13 +129,34 @@ public async Task Invoke(HttpContext context)
...
         private Task DisplayException(ErrorContext errorContext)
         {
-            var compilationException = errorContext.Exception as ICompilationException;
-            if (compilationException != null)
+            var httpContext = errorContext.HttpContext;
+            var headers = httpContext.Request.GetTypedHeaders();
+            var acceptHeader = headers.Accept;
+
+            // If the client does not ask for HTML just format the exception as plain text
+            if (acceptHeader != null && !acceptHeader.Any(h => h.IsSubsetOf(new MediaTypeHeaderValue("text/html"))))
```
Tratcher: Thank you 😁 

Doesn't this create a new MediaTypeHeaderValue inside a loop?

Tratcher (approved):

src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
davidfowl: Yea, the performance of this is probably really bad, though, we're on an exceptional path 😄

davidfowl: I just fixed this performance bug!

src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
davidfowl: > For more examples the chrome and IE dev tools also render it as header: value. That's the layout people are used to seeing headers in.

That's a fair point.

> header = value looks like query parameters.

Good thing we have a giant HEADERS title in the respones 😄

src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
```diff
@@ -17,6 +18,7 @@
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Microsoft.Extensions.StackTrace.Sources;
+using Microsoft.Net.Http.Headers;
 
 namespace Microsoft.AspNetCore.Diagnostics
 {
```
Eilon: I can't put a comment in the right place, but please update the doc comment right below this. It says it `generates HTML error responses.`

Eilon (approved):
Nice. Added one comment above.

src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
davidfowl: Done.

src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
```diff
@@ -32,6 +34,7 @@ public class DeveloperExceptionPageMiddleware
         private readonly DiagnosticSource _diagnosticSource;
         private readonly ExceptionDetailsProvider _exceptionDetailsProvider;
         private readonly Func<ErrorContext, Task> _exceptionHandler;
+        private static readonly MediaTypeHeaderValue _textPlainMediaType = new MediaTypeHeaderValue("text/html");
```
nil4: Shouldn't the value be `text/plain`?

campersau: Then the variable name is a bit missleading.

davidfowl: oops!

nil4: @davidfowl seems that the lines below seem to be wanting `text/plain` and not `text/html`:

https://github.com/aspnet/AspNetCore/blob/9bcb63c65495e6ad074258fd959183d49d5f703a/src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs#L138-L140

davidfowl: ?

nil4: Never-mind me -- I mixed up the initial PR content with the latest code, missing the changes you committed in the meantime. All 👍 now...

PR #9359, opened 2019-04-14
Improve HostFiltering

benaadams:
Is now included in DefaultBuilder and was showing up in traces

image

TryGetValueFast is addressed separately by https://github.com/aspnet/AspNetCore/pull/9341

Blocked by https://github.com/aspnet/Extensions/pull/1696

### DIFF
--- a/src/Middleware/HostFiltering/src/HostFilteringMiddleware.cs
+++ b/src/Middleware/HostFiltering/src/HostFilteringMiddleware.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.CompilerServices;
 using System.Text;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
@@ -68,19 +69,13 @@ namespace Microsoft.AspNetCore.HostFiltering
 
             if (!CheckHost(context, allowedHosts))
             {
-                context.Response.StatusCode = 400;
-                if (_options.IncludeFailureMessage)
-                {
-                    context.Response.ContentLength = DefaultResponse.Length;
-                    context.Response.ContentType = "text/html";
-                    return context.Response.Body.WriteAsync(DefaultResponse, 0, DefaultResponse.Length);
-                }
-                return Task.CompletedTask;
+                return HostValidationFailed(context);
             }
 
             return _next(context);
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         private IList<StringSegment> EnsureConfigured()
         {
             if (_allowAnyNonEmptyHost == true || _allowedHosts?.Count > 0)
@@ -88,10 +83,28 @@ namespace Microsoft.AspNetCore.HostFiltering
                 return _allowedHosts;
             }
 
+            return Configure();
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private Task HostValidationFailed(HttpContext context)
+        {
+            context.Response.StatusCode = 400;
+            if (_options.IncludeFailureMessage)
+            {
+                context.Response.ContentLength = DefaultResponse.Length;
+                context.Response.ContentType = "text/html";
+                return context.Response.Body.WriteAsync(DefaultResponse, 0, DefaultResponse.Length);
+            }
+            return Task.CompletedTask;
+        }
+
+        private IList<StringSegment> Configure()
+        {
             var allowedHosts = new List<StringSegment>();
             if (_options.AllowedHosts?.Count > 0 && !TryProcessHosts(_options.AllowedHosts, allowedHosts))
             {
-                _logger.LogDebug("Wildcard detected, all requests with hosts will be allowed.");
+                _logger.WildcardDetected();
                 _allowedHosts = allowedHosts;
                 _allowAnyNonEmptyHost = true;
                 return _allowedHosts;
@@ -104,7 +117,7 @@ namespace Microsoft.AspNetCore.HostFiltering
 
             if (_logger.IsEnabled(LogLevel.Debug))
             {
-                _logger.LogDebug("Allowed hosts: {Hosts}", string.Join("; ", allowedHosts));
+                _logger.AllowedHosts(string.Join("; ", allowedHosts));
             }
 
             _allowedHosts = allowedHosts;
@@ -142,37 +155,61 @@ namespace Microsoft.AspNetCore.HostFiltering
         }
 
         // This does not duplicate format validations that are expected to be performed by the host.
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         private bool CheckHost(HttpContext context, IList<StringSegment> allowedHosts)
         {
-            var host = new StringSegment(context.Request.Headers[HeaderNames.Host].ToString()).Trim();
+            var host = context.Request.Headers[HeaderNames.Host].ToString();
 
-            if (StringSegment.IsNullOrEmpty(host))
+            if (host.Length == 0)
             {
-                // Http/1.0 does not require the host header.
-                // Http/1.1 requires the header but the value may be empty.
-                if (!_options.AllowEmptyHosts)
-                {
-                    _logger.LogInformation("{Protocol} request rejected due to missing or empty host header.", context.Request.Protocol);
-                    return false;
-                }
-                _logger.LogDebug("{Protocol} request allowed with missing or empty host header.", context.Request.Protocol);
-                return true;
+                return IsEmptyHostAllowed(context);
+            }
+
+            // Host header should not require trimming
+            // rfc7230  3.2.4.  Field Parsing https://tools.ietf.org/html/rfc7230#page-25
+            // Optional whitespace occurring before the first non-whitespace octet of the field value or after the last
+            // non-whitespace octet of the field value ought to be excluded by
+            // parsers when extracting the field value from a header field
+            if (host[0] == ' ' || host[host.Length - 1] == ' ')
+            {
+                _logger.InvalidTrailingOrLeadingWhitespaceInHost(host);
+                return false;
             }
 
             if (_allowAnyNonEmptyHost == true)
             {
-                _logger.LogTrace("All hosts are allowed.");
+                _logger.AllHostsAllowed();
                 return true;
             }
 
-            if (HostString.MatchesAny(host, allowedHosts))
+            return CheckHostInAllowList(allowedHosts, host);
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private bool CheckHostInAllowList(IList<StringSegment> allowedHosts, string host)
+        {
+            if (HostString.MatchesAny(new StringSegment(host), allowedHosts))
             {
-                _logger.LogTrace("The host '{Host}' matches an allowed host.", host);
+                _logger.AllowedHostMatched(host);
                 return true;
             }
-            
-            _logger.LogInformation("The host '{Host}' does not match an allowed host.", host);
+
+            _logger.NoAllowedHostMatched(host);
             return false;
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private bool IsEmptyHostAllowed(HttpContext context)
+        {
+            // Http/1.0 does not require the host header.
+            // Http/1.1 requires the header but the value may be empty.
+            if (!_options.AllowEmptyHosts)
+            {
+                _logger.RequestRejectedMissingHost(context.Request.Protocol);
+                return false;
+            }
+            _logger.RequestAllowedMissingHost(context.Request.Protocol);
+            return true;
         }
     }
 }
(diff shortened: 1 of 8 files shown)


src/Middleware/HostFiltering/src/HostFilteringMiddleware.cs
```diff
@@ -142,37 +155,61 @@ private bool IsTopLevelWildcard(string host)
         }
 
         // This does not duplicate format validations that are expected to be performed by the host.
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         private bool CheckHost(HttpContext context, IList<StringSegment> allowedHosts)
         {
-            var host = new StringSegment(context.Request.Headers[HeaderNames.Host].ToString()).Trim();
+            var host = context.Request.Headers[HeaderNames.Host].ToString();
```
benaadams: @Tratcher is there a reason to allow whitespace before and after the host value? Kestrel will trim it and I assume IIS will also?

rfc7230  3.2.4.  Field Parsing says https://tools.ietf.org/html/rfc7230#page-25
```
Optional whitespace occurring before the first non-whitespace octet of the field value or after the last
non-whitespace octet of the field value ought to be excluded by
parsers when extracting the field value from a header field
```
I've changed it to reject rather than trim

Tratcher: It was not making any assumptions about the underlying server. It was valid input so we tried to allow for it. Does this part really show up in your trace?

benaadams: StringSegment is a 16 byte struct so passed via stack copy
```
Type layout for 'StringSegment'
Size: 16 bytes. Paddings: 0 bytes (%0 of empty space)
|=================================================|
|   0-7: String <Buffer>k__BackingField (8 bytes) |
|-------------------------------------------------|
|  8-11: Int32 <Offset>k__BackingField (4 bytes)  |
|-------------------------------------------------|
| 12-15: Int32 <Length>k__BackingField (4 bytes)  |
|=================================================|
```

`StringSegment.Trim()` => `TrimStart().TrimEnd()`

TrimStart and TrimEnd both fix the string as well as having a loop both of which block inlining; so it gets stack copies for each of them, then stack copied back out for the return; in addition to the fixing cost.

Could change Trim to first check is the start and end are whitespace etc and optimize there; however its either a parser error, or a user error (host set in pipeline with leading/trailing whitespace) and `TryProcessHosts` doesn't trim the hosts in the allow lists so its not a "round-trippable" error as it won't match if used in both places; so seems easiest to reject?

benaadams: > Does this part really show up in your trace?
```
dotnet new web
dotnet publish -c Release
```
```
 wrk -c 128 -t 4 -d 120 -H "Accept: application/json,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7" -H "Connection: keep-alive" http://127.0.0.1:5000/ -s ./wrk/scripts/pipeline.lua -- 16
```
On sampling it CheckHost+EnsureConfigured shows up as >10% of the rest of the pipeline 

CheckHost+EnsureConfigured: 20,569+1198 = 21,767ms
EndpointRoutingMiddleware onwards: 213,871ms

21,767/213,871 * 100 = 10.2%

image

benaadams: TrimStart and TrimEnd are players here; though the effect of logging is more significant

image

src/Middleware/HostFiltering/test/HostFilteringMiddlewareTests.cs
```diff
@@ -80,14 +80,14 @@ public async Task AllowsEmptyHost(bool allowed, int status)
...
+                        ctx.Request.Headers[HeaderNames.Host] = "";
                         return next();
                     });
                     app.UseHostFiltering();
                     app.Run(c =>
                     {
-                        Assert.True(c.Request.Headers.TryGetValue(HeaderNames.Host, out var host));
-                        Assert.True(StringValues.Equals(" ", host));
+                        var host = c.Request.Headers[HeaderNames.Host];
+                        Assert.True(StringValues.Equals(new StringValues(""), host.ToString()));
```
Tratcher: Does this maintain/test the distinction between there being no host header and there being an empty host header? The spec makes a distinction.

benaadams: Hmm... looking into it. Checking what the parser does.

benaadams: All looks ok (in Kestrel)

I think there is an issue with `Microsoft.AspNetCore.Http.HeaderDictionary`

https://github.com/aspnet/AspNetCore/blob/4649449206edae4992e09419334e736726fd2c8e/src/Http/Http/src/HeaderDictionary.cs#L77-L80

It should be testing an `IsNull` property (which doesn't exist) rather than the `IsNullOrEmpty`?

Tratcher: StringValues doesn't even have an IsNull :-)
The issue you highlight would only be on the setter, it would be fine if the Store already contained the empty value. I think all of the servers implement their own request headers structure, and this isn't an issue for response headers. That just leaves TestServer?

benaadams: >  That just leaves TestServer?

Yes, but I assume its why the test was inserting `" "` rather than `""` as a value, as empty gets thrown away

benaadams: Ah, Kestrel does similar thing; but uses `.Count == 0` which I guess would be the equivalent of `.IsNull`

benaadams: Resolved

Tratcher:
Hmm, that had some upstream side-effects. https://dev.azure.com/dnceng/public/_build/results?buildId=157194&view=ms.vss-test-web.build-test-results-tab

anurse:
Moving to preview 6. Please do not merge until master is open for preview 6 changes. We're locking down preview 5 early due to a tight bit of release scheduling ;)

benaadams:
@Tratcher upstream fixes https://github.com/aspnet/AspNetCore/pull/9359/commits/20dd7e480e94977d3733e744120a90192e33b6cb

benaadams:
Rebased

src/Middleware/HostFiltering/src/HostFilteringMiddleware.cs
```diff
@@ -142,37 +155,61 @@ private bool IsTopLevelWildcard(string host)
...
-                return true;
+                return IsEmptyHostAllowed(context);
+            }
+
+            // Host header should not require trimming
+            // rfc7230  3.2.4.  Field Parsing https://tools.ietf.org/html/rfc7230#page-25
+            // Optional whitespace occurring before the first non-whitespace octet of the field value or after the last
+            // non-whitespace octet of the field value ought to be excluded by
+            // parsers when extracting the field value from a header field
+            if (char.IsWhiteSpace(host[0]) || char.IsWhiteSpace(host[host.Length - 1]))
```
Tratcher: At this point your enforcing a header parsing rule. If there's an issue here it's with the underlying request parsing, not with the client's request. You can remove this whitespace check.

src/Middleware/ResponseCompression/test/BodyWrapperStreamTests.cs
```diff
@@ -17,7 +17,7 @@ public class BodyWrapperStreamTests
     {
         [Theory]
         [InlineData(null, "Accept-Encoding")]
-        [InlineData("", "Accept-Encoding")]
+        [InlineData("", ",Accept-Encoding")]
```
Tratcher: This is not desirable output. An empty element should be ignored/replaced.

benaadams: Change `StringValues` to not output empty strings?

Tratcher: Or Headers.Append to ignore an empty value?

benaadams: Empty string is different from not present, as you highlighted https://github.com/aspnet/AspNetCore/pull/9359#discussion_r275417254

> Does this maintain/test the distinction between there being no host header and there being an empty host header? The spec makes a distinction.

Tratcher: A present but empty header has a meaning in the case of Host. Empty values in a header list do not have a meaning and should be suppressed. (e.g. ", , ,,, , , value" is the same as "value"). We do that when writing response headers to the wire, no?

benaadams: > We do that when writing response headers to the wire, no?

Yes, though empty values can be happily added to the header; which is likely why they are skipped on output.

Will make change to `StringValues` to skip them also for `.ToString()`; if the header has a real value

benaadams: https://github.com/aspnet/Extensions/pull/1696

src/Middleware/HostFiltering/src/HostFilteringMiddleware.cs
benaadams: Dropped

src/Middleware/ResponseCompression/test/BodyWrapperStreamTests.cs
benaadams: Changed

benaadams:
Won't pass until aspnet/Extensions#1696 is merged, due to the current behaviour of `StringValues`

benaadams:
Pre
image

Post
image

Tratcher:
Ready to rebase on aspnet/Extensions#1696?

benaadams:
Rebased

src/Middleware/HostFiltering/src/HostFilteringMiddleware.cs
```diff
@@ -142,37 +155,50 @@ private bool IsTopLevelWildcard(string host)
...
 
         // This does not duplicate format validations that are expected to be performed by the host.
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         private bool CheckHost(HttpContext context, IList<StringSegment> allowedHosts)
         {
-            var host = new StringSegment(context.Request.Headers[HeaderNames.Host].ToString()).Trim();
+            var host = context.Request.Headers[HeaderNames.Host].ToString();
 
-            if (StringSegment.IsNullOrEmpty(host))
+            if (0u >= (uint)host.Length)
```
jkotalik: Why not `0 == (uint)host.Length`?

src/Middleware/HostFiltering/test/HostFilteringMiddlewareTests.cs
```diff
@@ -97,6 +97,37 @@ public async Task AllowsEmptyHost(bool allowed, int status)
             Assert.Equal(status, (int)response.StatusCode);
         }
 
+        [Fact]
+        public async Task WhitespaceHostRejected()
```
jkotalik: @Tratcher I okay if we take this change (to reject whitespace hosts). Thoughts?

jkotalik (approved):

src/Middleware/HostFiltering/src/HostFilteringMiddleware.cs
benaadams: Weird...

benaadams: Let me recheck here

src/Middleware/HostFiltering/test/HostFilteringMiddlewareTests.cs
benaadams: Dropped this rejection https://github.com/aspnet/AspNetCore/pull/9359#discussion_r282624054 as it should never happen; so should drop the test too

src/Middleware/HostFiltering/src/HostFilteringMiddleware.cs
benaadams: Oh, ok. Its because the Jit likes `>=` tests on `uint`s to do bounds check elimination and the next test *was* checking if the first char was space; so it eliminated the first char bounds check.

However, that check is now gone, so can just do a regular zero test.

src/Middleware/HostFiltering/test/HostFilteringMiddlewareTests.cs
benaadams: As the whitespace reject has gone; this test is just checking if whitespace is in the allow list; so it isn't adding anything.

benaadams: Addressed

src/Middleware/HostFiltering/src/HostFilteringMiddleware.cs
benaadams: Addressed

Tratcher (approved):

benaadams:
Rebased

benaadams:
Raised issue for Helix issue https://github.com/aspnet/AspNetCore/issues/11038

Issue #5926, opened 2018-06-12
Ambient distributed tracing context


SergeyKanzhelev:
## Motivation

Propagate W3C headers to dependant components and make distributed tracing context available even when monitoring is not enabled.

Propagation of headers will ensure that distributed trace will correlate services even when unmonitored ASP.NET app is deployed in-the-middle. Let's take service `A` that injects distributed tracing context and calls ASP.NET app `B`. Service `B` calls service `C` that expects distributed tracing context in incoming request.

If `B` is not being monitored for distributed traces - it should propagate distributed tracing context unmodified. When analysing telemetry from services `A` and `C` it will look like `A` calls `C` directly. Once distributed tracing monitoring will be enabled for `B` - `B` have to start modifying context according to W3C specification. So telemetry analysis will show `A` calls `B` which calls `C`. 

In both scenarios any other telemetry like `ILogger` traces or stack dumps can be attributed with the distributed tracing context. This will allow to correlate failures of service `B` with the distributed trace those failures affected.

CC: @seth-capistron, @vancem, @jacpull, @lmolkova, @glennc, @davidfowl

muratg:
cc @shirhatti @davidfowl @Tratcher

anurse:
Epic https://github.com/aspnet/AspNetCore/issues/8924

anurse:
@davidfowl to investigate the design here to enable `Activity` tracking by default.

davidfowl:
Closing this

Issue #5918, opened 2018-04-16
Parent Request Id is mislabeled as Correlation Id in log scope


ngbrown:
For the upcoming for ASP.Net Core 2.1; pull #1138 added to the log scope a `CorrelationId`, but the contents are only the parent request identifier and does not appear to be set if there was no parent request (no `Request-Id` header in the incoming HTTP request).

There is documentation/specifications (mostly by @lmolkova) around these terms at:

* [HttpCorrelationProtocol](https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/HttpCorrelationProtocol.md)
* [FlatRequestId](https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/FlatRequestId.md)
* [HierarchicalRequestId](https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/HierarchicalRequestId.md)

Specifically in the [Activity User Guide](https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/ActivityUserGuide.md), it says:

> When application calls external dependency to complete an operation, it may need to pass some of the context (e.g. ___correlation id___) [emphasis added] along with dependency call to be able to correlate logs from multiple service.

I see that the expectation is that in the logs, there will be a "correlation id" of the context, or `CorrelationId` if you will, that can be used to link the logs entries across multiple services handling a single user request.

In the current preview code, the Log scope outputs a `CorrelationId` which is the value of an incoming header `Request-Id`, and only if that header is set.  This is not useful for seeing the entire request across multiple services in the logs.  What is needed is adding to the log scope the root request id to the log scope of all services involved in handling the request, including within the root service that generated the first request id.

muratg:
@pakrym thoughts on the scope of this one? You agree that removing it from logging scope is the right approach?

pakrym:
I, unfortunately, don't have enough context on this. Would need to discuss with @davidfowl as it seems he understands what's wrong here.

lkts:
I want to agree here that `CorrelationId` is confusing in current implementation because it is always `null` for application which serves "initial" requests. I would expect it to be `Activity.Current.Id` as mentioned here.

anurse:
Epic https://github.com/aspnet/AspNetCore/issues/8924

anurse:
[...]

Issue #6895, opened 2019-01-20
Caching per-request objects on reused HttpContexts


JamesNK:
ASP.NET Core 3 adds HttpContext reuse - [Reuse HttpContext object per HTTP/1 connection and HTTP/2 stream](https://github.com/aspnet/AspNetCore/pull/6424)

App frameworks and middleware often have their own objects that are allocated every request. Instead of app frameworks/middleware caching with a shared object pool (creating thread contention), what opportunities are there for to cache lightweight per-request objects on `HttpContext`?

Because cached objects would eventually be on every HttpContext this functionality should only be used for lightweight objects, and not something large like buffers.

e.g. generated gRPC methods have objects passed to them like a stream reader, stream writer and context - `Task SayHellos(HelloRequest request, IServerStreamWriter<HelloReply> responseStream, ServerCallContext context)`. These objects are mostly wrappers over the underlying HttpContext.

* Before: IServerStreamWriter and ServerCallContext are allocated every request
* After: IServerStreamWriter and ServerCallContext are fetched from the `HttpContext`. App framework will create and cache them if they don't exist, or reset the existing instances for the current `HttpContext`.

JamesNK:
@davidfowl

davidfowl:
I think resettable features is the feature here. A feature can opt into controlling creation and reset.

JamesNK:
> Could have a cache that drops any object from the pool if it wasn't requested during that request; so it would only live on for 1 request (or a parameterised number e.g. if not used for x Requests then drop)

That's a clever idea. It could be something a feature could track, e.g. each time it is reset it could increment an "unused" counter by 1, when it is "used" then it goes back to zero, and if it is reset and the counter exceeds a number then heavy objects are culled (or could a feature remove itself from its HttpContext?).

> I think resettable features is the feature here. A feature can opt into controlling creation and reset.

What happens to features currently when the HttpContext is reused?

I don't know if this is a thing to worry about but imagine a web app that uses lots of middleware and app frameworks. Eventually there would be a lot of reset-able features in the mix on the `HttpContext`. Could the time spent each request resetting those features have an impact?

JamesNK:
Profiling gRPC and the largest allocation is the gRPC context. Caching this would almost halve allocated memory.

![image](https://user-images.githubusercontent.com/303201/54496790-85ef3080-48b0-11e9-92e5-55553ef69983.png)

jkotalik:
[...]

PR #9361, opened 2019-04-14
Create an activity more often than before

davidfowl:
- Create an Activity if there's a diagnostic listener attached with the activity name or if logging is enabled.
- Add the activity id to the logging scope (and call the field ActivityId)

cc @lmolkova @SergeyKanzhelev 

#5926 
#5918

### DIFF
--- a/src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs
+++ b/src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs
@@ -43,7 +43,7 @@ namespace Microsoft.AspNetCore.Hosting.Internal
 
             if (HostingEventSource.Log.IsEnabled())
             {
-                context.EventLogEnabled = true;
+                context.EventSourceEnabled = true;
                 // To keep the hot path short we defer logging in this function to non-inlines
                 RecordRequestStartEventLog(httpContext);
             }
@@ -51,13 +51,13 @@ namespace Microsoft.AspNetCore.Hosting.Internal
             var diagnosticListenerEnabled = _diagnosticListener.IsEnabled();
             var loggingEnabled = _logger.IsEnabled(LogLevel.Critical);
 
+            if (loggingEnabled || (diagnosticListenerEnabled && _diagnosticListener.IsEnabled(ActivityName, httpContext)))
+            {
+                context.Activity = StartActivity(httpContext);
+            }
 
             if (diagnosticListenerEnabled)
             {
-                if (_diagnosticListener.IsEnabled(ActivityName, httpContext))
-                {
-                    context.Activity = StartActivity(httpContext);
-                }
                 if (_diagnosticListener.IsEnabled(DeprecatedDiagnosticsBeginRequestKey))
                 {
                     startTimestamp = Stopwatch.GetTimestamp();
@@ -68,16 +68,10 @@ namespace Microsoft.AspNetCore.Hosting.Internal
             // To avoid allocation, return a null scope if the logger is not on at least to some degree.
             if (loggingEnabled)
             {
-                // Get the request ID (first try TraceParent header otherwise Request-ID header
-                if (!httpContext.Request.Headers.TryGetValue(TraceParentHeaderName, out var correlationId))
-                {
-                    httpContext.Request.Headers.TryGetValue(RequestIdHeaderName, out correlationId);
-                }
-
                 // Scope may be relevant for a different level of logging, so we always create it
                 // see: https://github.com/aspnet/Hosting/pull/944
                 // Scope can be null if logging is not on.
-                context.Scope = _logger.RequestScope(httpContext, correlationId);
+                context.Scope = _logger.RequestScope(httpContext, context.Activity.Id);
 
                 if (_logger.IsEnabled(LogLevel.Information))
                 {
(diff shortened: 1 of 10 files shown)


src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs
```diff
@@ -51,13 +51,13 @@ public void BeginRequest(HttpContext httpContext, ref HostingApplication.Context
             var diagnosticListenerEnabled = _diagnosticListener.IsEnabled();
             var loggingEnabled = _logger.IsEnabled(LogLevel.Critical);
```
benaadams: Is `_logger.IsEnabled(LogLevel.Critical)` the right level to create an activity? (As generally it will always be on at critical or above)

davidfowl: In theory, there should *always* be an activity (I need to run some perf tests to see how this affects the default scenarios).  This basically replaces the current request id. The current logic assumes if you're going to log a a scope, then the activity should be created.

We added [this](https://github.com/aspnet/Extensions/blob/e26bf4b2baf715c8f83765dd3e91910c705cc32c/src/Logging/Logging/src/LoggerFilterOptions.cs#L13) but there's kinda a chicken and egg problem as we need to make the activity to be a part of the scope before knowing if the scope if there.

davidfowl: There may also need to be another explicit way to turn it on or off outside logging but that's sorta why this is a draft.

benaadams: Resetable Activity on the connection using something like: https://github.com/aspnet/AspNetCore/issues/6895

Currently problematic as Activity doesn't support being cleared.

davidfowl: The server would need to own activity creation in that case. I'm not sure we'll come up with anything reasonable in the 3.0 time frame for #6895. I think coupling it to logging is the best bet mostly because the data is for the logging system and that scenario has a bunch of allocations already (the scope, the messages being logged etc). 

Maybe the log level needs to be adjusted to something like information but the category is this internal category (which might be a bit weird).

benaadams: https://github.com/aspnet/Extensions/pull/1422

Then something like?
```csharp
var scope = logger.BeginScope(
    (state) => new HostingLogScope(state.httpContext, state.activityId), 
    (httpContext, activityId));

if (!scope.IsNullScope() || (diagnosticListenerEnabled && _diagnosticListener.IsEnabled(ActivityName, httpContext)))
{
    context.Activity = StartActivity(httpContext);
}
```

davidfowl: This only helps when scopes are off. We could just add a `DisableActivityCreation` to the `WebHostOptions` if we wanted to remove the allocation in benchmarks.

benaadams: Is there anyway to do an inversion? Say for example AppInisights is set to sample 5% of activities rather than 100% is there a way to ask it "will you consume this activity" and create based on that; rather than create it 100% of the time and throw it away unobserved 95% of the time?

davidfowl: Yes there's a plan to support sampling but this feature needs to work with logging when app insights isn't on so that it's possible to correlate logs without too much setup. 

See https://github.com/aspnet/AspNetCore/pull/9340 for the sampling PR. The activity still needs to be created though (so there's a chicken and egg problem)

SergeyKanzhelev: @benaadams we will need an `Activity` not only to track something, but also to propagate the decision that we decided NOT to track something. So head-based sampling would work across the components.

pranavkm (approved):

src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs
```diff
@@ -68,16 +68,10 @@ public void BeginRequest(HttpContext httpContext, ref HostingApplication.Context
...
-                if (!httpContext.Request.Headers.TryGetValue(TraceParentHeaderName, out var correlationId))
-                {
-                    httpContext.Request.Headers.TryGetValue(RequestIdHeaderName, out correlationId);
-                }
-
                 // Scope may be relevant for a different level of logging, so we always create it
                 // see: https://github.com/aspnet/Hosting/pull/944
                 // Scope can be null if logging is not on.
-                context.Scope = _logger.RequestScope(httpContext, correlationId);
+                context.Scope = _logger.RequestScope(httpContext, context.Activity.Id);
```
davidfowl: @SergeyKanzhelev @lmolkova should we split this out into span id and trace id in the logs?

SergeyKanzhelev: This is what loggers are typically doing. Like in OpenCensus: https://github.com/census-instrumentation/opencensus-specs/blob/c3fd41b28fed96ceffc8c35af280d74cacea3ab2/trace/LogCorrelation.md#key-names-for-tracing-data or Elastic: https://github.com/elastic/ecs/issues/99

PR #9370, opened 2019-04-14
HeaderPropagation Middleware: handle content headers

alefranz:
This PR addresses feedback @benaadams added to https://github.com/aspnet/AspNetCore/pull/7921 after it was merged:
- Handle content headers.
- Avoid array allocation on the common scenario of a single value for the header.

The bug with content headers was even worst then anticipated, as `Contains` called on `HttpRequestHeaders` throws for content headers. For this reason I had to switch to use `TryGetValues`.

/cc @rynowak

### DIFF
--- a/src/Middleware/HeaderPropagation/src/HeaderPropagationMessageHandler.cs
+++ b/src/Middleware/HeaderPropagation/src/HeaderPropagationMessageHandler.cs
@@ -53,11 +53,31 @@ namespace Microsoft.AspNetCore.HeaderPropagation
             {
                 var outputName = string.IsNullOrEmpty(entry?.OutboundHeaderName) ? headerName : entry.OutboundHeaderName;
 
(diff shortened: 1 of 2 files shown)


benaadams (approved):

alefranz:
Thanks Ben for the prompt review!

Tratcher (approved):

rynowak (approved):

rynowak:
Thanks!

clairepat:
Look, I’m an old flight nurse so I’m not stupid, I could probably start your hearts if they stopped in front of me, but this coding is confounding. If you could please just tell me what you would do given the extraordinary situation. Not what to do mind you, what you personally would do. I need a little guidance, I’m way out of my depth here, at least in this situation. I can fly with trauma patients all over the southeast at one time, but nothing this intense. I was fine on five figure a year income. So please what would you do? Not, mind you, what should I do.

PR #9372, opened 2019-04-14
Enhancements for Dynamic Endpoints

rynowak:
A series of incremental improvements for scenarios around dynamic endpoints:
- Make our built-in policies better behaved for dynamic endpoints
- Allow changing the size of the candidate set
- Use these new features in MVC

### DIFF
--- a/src/Http/Routing/src/Matching/CandidateSet.cs
+++ b/src/Http/Routing/src/Matching/CandidateSet.cs
@@ -2,8 +2,12 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Buffers;
 using System.Collections;
+using System.Collections.Generic;
 using System.Collections.Specialized;
+using System.Diagnostics;
+using System.Linq;
 using System.Runtime.CompilerServices;
 using Microsoft.AspNetCore.Http;
 
@@ -18,7 +22,7 @@ namespace Microsoft.AspNetCore.Routing.Matching
     {
         private const int BitVectorSize = 32;
 
-        private readonly CandidateState[] _candidates;
+        private CandidateState[] _candidates;
 
         /// <summary>
         /// <para>
@@ -55,8 +59,6 @@ namespace Microsoft.AspNetCore.Routing.Matching
                 throw new ArgumentException($"The provided {nameof(endpoints)}, {nameof(values)}, and {nameof(scores)} must have the same length.");
             }
 
-            Count = endpoints.Length;
-
             _candidates = new CandidateState[endpoints.Length];
             for (var i = 0; i < endpoints.Length; i++)
             {
@@ -66,8 +68,6 @@ namespace Microsoft.AspNetCore.Routing.Matching
 
         internal CandidateSet(Candidate[] candidates)
         {
-            Count = candidates.Length;
-
             _candidates = new CandidateState[candidates.Length];
             for (var i = 0; i < candidates.Length; i++)
             {
@@ -78,7 +78,7 @@ namespace Microsoft.AspNetCore.Routing.Matching
         /// <summary>
         /// Gets the count of candidates in the set.
         /// </summary>
-        public int Count { get; }
+        public int Count => _candidates.Length;
 
         /// <summary>
         /// Gets the <see cref="CandidateState"/> associated with the candidate <see cref="Endpoint"/>
@@ -153,7 +153,7 @@ namespace Microsoft.AspNetCore.Routing.Matching
         /// <param name="index">The candidate index.</param>
         /// <param name="endpoint">
         /// The <see cref="Endpoint"/> to replace the original <see cref="Endpoint"/> at
-        /// the <paramref name="index"/>. If <paramref name="endpoint"/> the candidate will be marked
+        /// the <paramref name="index"/>. If <paramref name="endpoint"/> is <c>null</c>. the candidate will be marked
         /// as invalid.
         /// </param>
         /// <param name="values">
@@ -167,7 +167,7 @@ namespace Microsoft.AspNetCore.Routing.Matching
             {
                 ThrowIndexArgumentOutOfRangeException();
             }
-
+            
             _candidates[index] = new CandidateState(endpoint, values, _candidates[index].Score);
 
             if (endpoint == null)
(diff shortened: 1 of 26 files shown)


src/Http/Routing/src/Matching/HostMatcherPolicy.cs
```diff
@@ -48,6 +62,93 @@ public bool AppliesToEndpoints(IReadOnlyList<Endpoint> endpoints)
             });
         }
 
+        public Task ApplyAsync(HttpContext httpContext, EndpointSelectorContext context, CandidateSet candidates)
```
rynowak: I didn't seem much reusable from the jump table implementation because the order of operations is different. It seemed more maintainable to just create a second code path that's about 85% separate.

src/Http/Routing/src/Matching/HttpMethodMatcherPolicy.cs
```diff
@@ -62,6 +83,104 @@ public bool AppliesToEndpoints(IReadOnlyList<Endpoint> endpoints)
...
+            // Returning a 405 here requires us to return keep track of all 'seen' HTTP methods. We allocate to
+            // keep track of this beause we either need to keep track of the HTTP methods or keep track of the
+            // endpoints - both allocate.
+            //
+            // Those code only runs in the presence of dynamic endpoints anyway.
+            //
+            // We want to return a 405 iff we eliminated ALL of the currently valid endpoints due to HTTP method
+            // mismatch.
+            bool? needs405Endpoint = null;
+            HashSet<string> methods = null;
```
rynowak: This implemetatnion is pretty trivial without the support for 405 :(

src/Http/Routing/src/Matching/HostMatcherPolicy.cs
```diff
@@ -48,6 +62,93 @@ public bool AppliesToEndpoints(IReadOnlyList<Endpoint> endpoints)
...
+        {
+            if (httpContext == null)
```
davidfowl: @benaadams Should this error handling be moved to another method?

rynowak: This is interface dispatch. it's not going to get inlined

src/Http/Routing/src/Matching/CandidateSet.cs
```diff
@@ -18,19 +18,7 @@ public sealed class CandidateSet
...
-
-        // We inline storage for 4 candidates here to avoid allocations in common
-        // cases. There's no real reason why 4 is important, it just seemed like 
-        // a plausible number.
-        private CandidateState _state0;
-        private CandidateState _state1;
-        private CandidateState _state2;
-        private CandidateState _state3;
-
-        private CandidateState[] _additionalCandidates;
```
rynowak: @davidfowl - this boy is two allocations now but much slimmer. Some hot code paths got simpler too.

src/Http/Routing/src/Matching/CandidateSet.cs
```diff
@@ -244,15 +124,7 @@ public bool IsValidCandidate(int index)
...
-            if (Count < BitVectorSize)
-            {
-                // Get the n-th bit
-                return _validity[0x00000001 << index];
-            }
-            else
-            {
-                return _largeCapactityValidity[index];
-            }
+            return _candidates[index].Score >= 0;
```
rynowak: We're able to simplify this dramatically by using the score to store whether or not the candidate is valid

src/Http/Routing/src/Matching/HttpMethodMatcherPolicy.cs
```diff
@@ -62,6 +83,104 @@ public bool AppliesToEndpoints(IReadOnlyList<Endpoint> endpoints)
...
+                if (!candidates.IsValidCandidate(i))
+                {
+                    continue;
+                }
+
+                var httpMethod = httpContext.Request.Method;
+                if (metadata.AcceptCorsPreflight &&
+                    string.Equals(httpMethod, PreflightHttpMethod, StringComparison.OrdinalIgnoreCase) &&
+                    httpContext.Request.Headers.ContainsKey(OriginHeader) &&
+                    httpContext.Request.Headers.TryGetValue(AccessControlRequestMethod, out var accessControlRequestMethod) &&
```
benaadams: `Request.Headers.TryGetValue(AccessControlRequestMethod` 
=> `Request.Headers.TryGetValue(HeaderNames.AccessControlRequestMethod`

src/Http/Routing/src/Matching/HttpMethodMatcherPolicy.cs
```diff
@@ -62,6 +83,104 @@ public bool AppliesToEndpoints(IReadOnlyList<Endpoint> endpoints)
...
+
+                if (!candidates.IsValidCandidate(i))
+                {
+                    continue;
+                }
+
+                var httpMethod = httpContext.Request.Method;
+                if (metadata.AcceptCorsPreflight &&
+                    string.Equals(httpMethod, PreflightHttpMethod, StringComparison.OrdinalIgnoreCase) &&
+                    httpContext.Request.Headers.ContainsKey(OriginHeader) &&
```
benaadams: `Request.Headers.ContainsKey(OriginHeader`
=> `Request.Headers.ContainsKey(HeaderNames.OriginHeader`

Lookup may get faster https://github.com/aspnet/AspNetCore/pull/9341

src/Http/Routing/src/Matching/HttpMethodMatcherPolicy.cs
benaadams: Also lift Request.Headers to a local rather than looking it up via 2 interfaces twice
```csharp
var headers = httpContext.Request.Headers;
if (metadata.AcceptCorsPreflight &&
    string.Equals(httpMethod, PreflightHttpMethod, StringComparison.OrdinalIgnoreCase) &&
    headers.ContainsKey(HeaderNames.OriginHeader) &&
    headers.TryGetValue(HeaderNames.AccessControlRequestMethod, out var accessControlRequestMethod) &&
    !StringValues.IsNullOrEmpty(accessControlRequestMethod))
```

e.g. non-faded are calls made getting item from RequestHeaders via httpContext:

image


The long time spent in string.Equals, string.EqualsOridinalIgnoreCase for a 4 letter string "Host" is what lead to #9341 and its acceleration for `HeaderNames.` strings

rynowak: Much <3 thanks!

src/Http/Routing/src/Matching/HostMatcherPolicy.cs
davidfowl: Sure but there's code bloat here and nobody is going to pass null except for a unit test 😄

rynowak: Don't blame me. C# should have a feature for this.

rynowak:
@JamesNK - added the thing we talked about today.

src/Http/Routing/src/Matching/HostMatcherPolicy.cs
```diff
@@ -48,6 +62,93 @@ public bool AppliesToEndpoints(IReadOnlyList<Endpoint> endpoints)
...
+                    // Can match any host.
+                    continue;
+                }
+
+                var matched = false;
+                var (requestHost, requestPort) = GetHostAndPort(httpContext);
+                for (var j = 0; j < hosts.Count; j++)
+                {
+                    var host = hosts[j].AsSpan();
+                    var port = "".AsSpan();
```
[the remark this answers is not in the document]
rynowak: Made this into `ReadOnlySpan<char>.Empty`

rynowak:
@JamesNK - ready for your cool feedback

src/Http/Routing/src/Matching/CandidateSet.cs
```diff
@@ -302,39 +167,197 @@ public void ReplaceEndpoint(int index, Endpoint endpoint, RouteValueDictionary v
...
+                    // score 3: `C2
+                    // score 4: D
+
+                    // Candidates before index can be copied unchanged.
+                    for (var i = 0; i < index; i++)
+                    {
+                        candidates[i] = original[i];
+                    }
+
+                    var buffer = endpoints.ToArray();
```
JamesNK: This would happen per-request right? I worry about CMSes out there that have tens of thousands of endpoints. Large blogs, news sites, etc. Is it worth having an API that allows a user to pass in a pre-sorted list of endpoints that they're cached?

src/Http/Routing/src/Matching/HttpMethodMatcherPolicy.cs
```diff
@@ -62,6 +80,105 @@ public bool AppliesToEndpoints(IReadOnlyList<Endpoint> endpoints)
...
+                throw new ArgumentNullException(nameof(candidates));
+            }
+
+            // Returning a 405 here requires us to return keep track of all 'seen' HTTP methods. We allocate to
+            // keep track of this beause we either need to keep track of the HTTP methods or keep track of the
+            // endpoints - both allocate.
+            //
+            // Those code only runs in the presence of dynamic endpoints anyway.
+            //
+            // We want to return a 405 iff we eliminated ALL of the currently valid endpoints due to HTTP method
```
JamesNK:
```suggestion
            // We want to return a 405 if we eliminated ALL of the currently valid endpoints due to HTTP method
```

src/Mvc/Mvc.Core/src/Routing/ConsumesMatcherPolicy.cs
```diff
@@ -24,16 +22,140 @@ internal class ConsumesMatcherPolicy : MatcherPolicy, IEndpointComparerPolicy, I
...
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            if (candidates == null)
+            {
+                throw new ArgumentNullException(nameof(candidates));
+            }
+
+            // We want to return a 415 iff we eliminated ALL of the currently valid endpoints due to content type
```
JamesNK:
```suggestion
            // We want to return a 415 if we eliminated ALL of the currently valid endpoints due to content type
```

src/Http/Routing/src/Matching/CandidateSet.cs
rynowak: I don't think passing in 1000 endpoints here is viable whether we pre-sort them or not. You're asking us to iterate 1000 endpoints multiple times and process policies on all of them. At that point you're holding it wrong. 

I don't think this is a implementation strategy we should try to enable. If you're writing a CMS with 1000s of endpoints you need to apply some kind of lookup yourself and not *just* use `IEndpointSelectorPolicy`. I haven't heard anyone ask for this.

src/Mvc/Mvc.Core/src/Routing/ConsumesMatcherPolicy.cs
rynowak: No, this is correct. If and only if.

src/Http/Routing/src/Matching/HttpMethodMatcherPolicy.cs
rynowak: This is intended

rynowak:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=157692&view=logs

rynowak:
@jamesnk any other feedback?

aspnet-hello:

I've triaged the above build.

JamesNK (approved):

PR #9390, opened 2019-04-15
Update SQLite versions

ajcvickers:


natemcmaster (changes requested):
A few more steps are required as described in https://github.com/aspnet/AspNetCore/blob/release/2.1/docs/PreparingPatchUpdates.md

* Update submodules.props https://github.com/aspnet/AspNetCore/blob/release/2.1/build/submodules.props#L40
* Update versions.props in the EFCore code to ship a new patch version
* Update the EF Core submodule in this repo

ajcvickers:
@natemcmaster I was taking this open from the two you presented because I thought you meant that was the only thing we needed to do. If we have to do this other stuff anyway, then I'll take the other option.

PR #9399, opened 2019-04-15
Remove RequiredPolicy

### LINKED ISSUE #9398: Remove AuthorizationOptions.RequiredPolicy
Discussed this offline and support for this newly added option can be removed

HaoK:
Fixes https://github.com/aspnet/AspNetCore/issues/9398

### DIFF
--- a/src/Security/Authorization/Core/src/AuthorizationPolicy.cs
+++ b/src/Security/Authorization/Core/src/AuthorizationPolicy.cs
@@ -176,17 +176,6 @@ namespace Microsoft.AspNetCore.Authorization
                 }
             }
 
-            var requiredPolicy = await policyProvider.GetRequiredPolicyAsync();
-            if (requiredPolicy != null)
-            {
-                if (policyBuilder == null)
-                {
-                    policyBuilder = new AuthorizationPolicyBuilder();
-                }
-
-                policyBuilder.Combine(requiredPolicy);
-            }
-
             return policyBuilder?.Build();
         }
     }
--- a/src/Security/Authorization/Policy/src/AuthorizationMiddleware.cs
+++ b/src/Security/Authorization/Policy/src/AuthorizationMiddleware.cs
@@ -51,7 +51,7 @@ namespace Microsoft.AspNetCore.Authorization
 
             // IMPORTANT: Changes to authorization logic should be mirrored in MVC's AuthorizeFilter
             var authorizeData = endpoint?.Metadata.GetOrderedMetadata<IAuthorizeData>() ?? Array.Empty<IAuthorizeData>();
-            if (!authorizeData.Any())
+            if (authorizeData.Count() == 0)
             {
                 await _next(context);
                 return;
--- a/src/Security/Authorization/Policy/src/AuthorizationMiddleware.cs
+++ b/src/Security/Authorization/Policy/src/AuthorizationMiddleware.cs
@@ -51,7 +51,7 @@ namespace Microsoft.AspNetCore.Authorization
 
             // IMPORTANT: Changes to authorization logic should be mirrored in MVC's AuthorizeFilter
             var authorizeData = endpoint?.Metadata.GetOrderedMetadata<IAuthorizeData>() ?? Array.Empty<IAuthorizeData>();
-            if (authorizeData.Length == 0)
+            if (!authorizeData.Any())
             {
                 await _next(context);
                 return;
--- a/src/Security/Authorization/Policy/src/AuthorizationMiddleware.cs
+++ b/src/Security/Authorization/Policy/src/AuthorizationMiddleware.cs
@@ -51,6 +51,11 @@ namespace Microsoft.AspNetCore.Authorization
 
             // IMPORTANT: Changes to authorization logic should be mirrored in MVC's AuthorizeFilter
             var authorizeData = endpoint?.Metadata.GetOrderedMetadata<IAuthorizeData>() ?? Array.Empty<IAuthorizeData>();
+            if (authorizeData.Length == 0) {
+                await _next(context);
+                return;
+            }
+            
             var policy = await AuthorizationPolicy.CombineAsync(_policyProvider, authorizeData);
             if (policy == null)
             {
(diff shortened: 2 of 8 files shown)


src/Security/Authorization/Core/src/AuthorizationPolicy.cs
```diff
@@ -176,17 +176,6 @@ public static async Task<AuthorizationPolicy> CombineAsync(IAuthorizationPolicyP
                 }
             }
 
-            var requiredPolicy = await policyProvider.GetRequiredPolicyAsync();
```
pranavkm: If we remove this, we should modify `AuthorizationMiddleware` to no-op if no auth metadata is available for the endpoint. 

See https://github.com/aspnet/AspNetCore/blob/master/src/Security/Authorization/Policy/src/AuthorizationMiddleware.cs#L53:

```C#
var authorizeData = endpoint?.Metadata.GetOrderedMetadata<IAuthorizeData>();
if (authorizeData == null)
{
   return _next(context);
}
```

cc @JamesNK \ @rynowak  to confirm this behavior.

JamesNK: Could someone have a policy provider - https://github.com/aspnet/AspNetCore/blob/1cd544042d6fba73f844b5e6d5f10d9e7dfa0367/src/Security/Authorization/Policy/src/AuthorizationMiddleware.cs#L54 - that still returns a policy even if there is no authz metadata?

JamesNK: https://github.com/aspnet/AspNetCore/blob/1cd544042d6fba73f844b5e6d5f10d9e7dfa0367/src/Security/Authorization/Core/src/AuthorizationPolicy.cs

It looks like that after this change the policy provider would never be call if there is policy metadata. No-oping would be good. We could even skip the async state machine in the AuthZ middleware if there is no metadata.

pranavkm: > - that still returns a policy even if there is no authz metadata

As you noted, CombineAsync does not call the policy provider in the event there's no auth metadata.

HaoK: Added a no-op check if there's no authorize data to the middleware
```C#
            if (authorizeData.Length == 0) {
                await _next(context);
                return;
            }
```

src/Security/Authorization/Policy/src/AuthorizationMiddleware.cs
```diff
@@ -51,6 +51,11 @@ public async Task Invoke(HttpContext context)
 
             // IMPORTANT: Changes to authorization logic should be mirrored in MVC's AuthorizeFilter
             var authorizeData = endpoint?.Metadata.GetOrderedMetadata<IAuthorizeData>() ?? Array.Empty<IAuthorizeData>();
+            if (authorizeData.Length == 0) {
```
pranavkm: Formatting.

pranavkm (approved):

JamesNK (approved):

src/Security/Authorization/Policy/src/AuthorizationMiddleware.cs
```diff
@@ -51,6 +51,12 @@ public async Task Invoke(HttpContext context)
 
             // IMPORTANT: Changes to authorization logic should be mirrored in MVC's AuthorizeFilter
             var authorizeData = endpoint?.Metadata.GetOrderedMetadata<IAuthorizeData>() ?? Array.Empty<IAuthorizeData>();
```
pranavkm: @rynowak \ @JamesNK can we have `GetOrderedMetadata` return an `IReadOnlyList<T>`? It internally caches an `T[]` so there isn't any code change involved with it.

pranavkm: @Eilon, @rynowak pointed out that this would be a breaking change since the API exists in 2.2. However, this seems like a worthwhile change to make - other APIs and users are likely to benefit from this change. How do you feel about this change?

pranavkm: We could just do `Count()` here. It'll effectively be the same: https://github.com/dotnet/corefx/blob/master/src/System.Linq/src/System/Linq/Count.cs#L12-L22

HaoK:
@pranavkm test failure which doesn't look related to this change, is this a known issue?  Should I just retry?

```
   at System.Net.Http.HttpContent.LoadIntoBufferAsyncCore(Task serializeToStreamTask, MemoryStream tempBuffer)
   at System.Net.Http.HttpClient.FinishSendAsyncBuffered(Task`1 sendTask, HttpRequestMessage request, CancellationTokenSource cts, Boolean disposeCts)
   at Microsoft.AspNetCore.Mvc.FunctionalTests.HttpClientExtensions.GetHtmlDocumentAsync(HttpClient client, String requestUri) in /_/src/Mvc/test/Mvc.FunctionalTests/Infrastructure/HttpClientExtensions.cs:line 18
   at Microsoft.AspNetCore.Mvc.FunctionalTests.FlushPointTest.FlushFollowedByLargeContent() in /_/src/Mvc/test/Mvc.FunctionalTests/FlushPointTest.cs:line 45
----- Inner Stack Trace -----
   at System.IO.Pipelines.PipeCompletion.ThrowLatchedException()
   at System.IO.Pipelines.Pipe.GetReadResult(ReadResult& result)
   at System.IO.Pipelines.Pipe.GetReadAsyncResult()
   at Microsoft.AspNetCore.TestHost.ResponseStream.ReadAsync(Byte[] buffer, Int32 offset, Int32 count, CancellationToken cancellationToken) in /_/src/Hosting/TestHost/src/ResponseStream.cs:line 114
   at System.IO.Stream.CopyToAsyncInternal(Stream destination, Int32 bufferSize, CancellationToken cancellationToken)
   at System.Net.Http.StreamToStreamCopy.<>c.<DisposeSourceWhenCompleteAsync>b__1_0(Task completed, Object innerSource)
   at System.Threading.Tasks.ContinuationTaskFromTask.InnerInvoke()
   at System.Threading.Tasks.Task.<>c.<.cctor>b__274_0(Object obj)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot, Thread threadPoolThread)
   at System.Net.Http.HttpContent.LoadIntoBufferAsyncCore(Task serializeToStreamTask, MemoryStream tempBuffer)
----- Inner Stack Trace -----
```

pranavkm:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=159145

pranavkm:
Yeah, I kicked off a retry. It's a fairly new test, so perhaps there's a bug :(

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/2300

PR #9405, opened 2019-04-16
Suppress warnings about breaking changes to Newtonsoft.Json dependency in SignalR

natemcmaster:
Resolves https://github.com/aspnet/AspNetCore-Internal/issues/1940

This adds support for listing a reference removal as an allowable breaking change.

### DIFF
--- a/docs/BuildErrors.md
+++ b/docs/BuildErrors.md
@@ -14,4 +14,4 @@ in a previous release of this assembly. See <./ReferenceResolution.md> for how t
 
 > error BUILD002: Package references changed since the last release...
 
-Similar to BUILD001, but this error not suppressable. This error only appears in servicing builds which should not change references between assemblies or packages.
+Similar to BUILD001, but this error is not suppressable. This error only appears in servicing builds, which should not change references between assemblies or packages.
--- a/docs/BuildErrors.md
+++ b/docs/BuildErrors.md
@@ -1,0 +1,17 @@
+Build Errors
+------------
+
+This document is for common build errors and how to resolve them.
+
+### Warning BUILD001
+
+> warning BUILD001: Package references changed since the last release...
+
+This warning indicates a breaking change might have been made to a package or assembly due to the removal of a reference which was used
+in a previous release of this assembly. See <./ReferenceResolution.md> for how to suppress.
+
+### Error BUILD002
+
+> error BUILD002: Package references changed since the last release...
+
+Similar to BUILD001, but this error not suppressable. This error only appears in servicing builds which should not change references between assemblies or packages.
--- a/docs/ReferenceResolution.md
+++ b/docs/ReferenceResolution.md
@@ -25,6 +25,7 @@ The requirements that led to this system are:
 * Name the .csproj file to match the assembly name.
 * Run `build.cmd /t:GenerateProjectList` when adding new projects
 * Use [eng/tools/BaseLineGenerator/](/eng/tools/BaselineGenerator/README.md) if you need to update baselines.
+* If you need to make a breaking change to dependencies, you may need to add `<SuppressBaselineReference>`.
 
 ## Important files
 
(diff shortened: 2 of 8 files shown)


src/SignalR/common/Protocols.Json/src/Microsoft.AspNetCore.SignalR.Protocols.Json.csproj
```diff
@@ -27,4 +27,9 @@
     <Reference Include="System.Runtime.CompilerServices.Unsafe" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(AspNetCoreMajorMinorVersion)' == '3.0'">
```
BrennanConroy: This is technically a new package in 3.0.

We replaced the old Microsoft.AspNetCore.SignalR.Protocols.Json package with the Microsoft.AspNetCore.SignalR.Protocols.NewtonsoftJson package and then removed the old one.

So this is now a new package and probably shouldn't have a baseline associated with it.

natemcmaster: Even so, it's the same package ID and the build system (and users) see this as a breaking change. It's okay since this was planned. This documents that intention in the code.

docs/BuildErrors.md
```diff
@@ -0,0 +1,17 @@
+Build Errors
```
anurse: 👏 👏 👏 👏 👏 👏

src/SignalR/clients/csharp/Client.Core/src/Microsoft.AspNetCore.SignalR.Client.Core.csproj
```diff
@@ -28,4 +28,9 @@
     <Reference Include="System.Threading.Channels" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(AspNetCoreMajorMinorVersion)' == '3.0'">
+    <!-- This dependency was replaced by Protocols.NewtonsoftJson between 3.0 and 2.2. This suppression can be removed after 3.0 is complete. -->
```
halter73: Should we file an issue to do this after branching for 3.0? I guess it's not a big deal if we don't remove it right away.

halter73 (approved):

docs/BuildErrors.md
```diff
@@ -0,0 +1,17 @@
...
+> warning BUILD001: Package references changed since the last release...
+
+This warning indicates a breaking change might have been made to a package or assembly due to the removal of a reference which was used
+in a previous release of this assembly. See <./ReferenceResolution.md> for how to suppress.
+
+### Error BUILD002
+
+> error BUILD002: Package references changed since the last release...
+
+Similar to BUILD001, but this error not suppressable. This error only appears in servicing builds which should not change references between assemblies or packages.
```
dougbu: Minor but this will slow readers: "error **is** not&hellip;"

docs/ReferenceResolution.md
```diff
@@ -67,3 +68,16 @@ Steps for adding a new package dependency to an existing project. Let's say I'm
 
        If you don't know the commit hash of the source code used to produce "0.0.1-beta-1", you can use `000000` as a placeholder for `Sha`
        as its value is unimportant and will be updated the next time the bot runs.
+
+## Example: make a breaking change to references
+
+If Microsoft.AspNetCore.Banana in 2.1 had a reference to `Microsoft.AspNetCore.Orange`, but in 3.0 this reference is changing to `Microsoft.AspNetCore.BetterThanOrange`, you would need to make these changes to the .csproj file
```
dougbu: 🍌 is already better than orange (🔶) ❕

src/SignalR/clients/csharp/Client.Core/src/Microsoft.AspNetCore.SignalR.Client.Core.csproj
dougbu: Well, such an issue is likely to get lost between now and RTM if it has to wait in the backlog. But, if you create a 4.0 milestone for it, that'll work.

dougbu (approved):

dougbu:
Hard to tell but the build failures in attempt 1 are likely flakiness. I started attempt 2.

BrennanConroy:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=156855&view=logs

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/2267

src/SignalR/clients/csharp/Client.Core/src/Microsoft.AspNetCore.SignalR.Client.Core.csproj
BrennanConroy: We don't need to file an issue, this will go away when we make System.Text.Json the default.

natemcmaster: It doesn't hurt if it stays around, but feel free to file an issue to clean up later. I'm writing these suppressions with the condition that checks version to makes sure the code doesn't become a problem later when we branch for 3.1 and beyond.

PR #9430, opened 2019-04-16
Make IHubClients and IHubCallerClients Abstract Classes

### LINKED ISSUE #5382: Make IHubClients<T> an abstract class
So that we can add new operators without breaking things. We'd keep the interface, but change the properties to use the abstract class (we could reargue keeping the interface later ;))

mikaelm12:
Fixes: https://github.com/aspnet/AspNetCore/issues/5382

I obsoleted the `IHubClients` and `IHubCallerClients` interfaces and created abstract base classes to replace them.

### DIFF
--- a/src/SignalR/server/Core/src/HubClientBase.cs
+++ b/src/SignalR/server/Core/src/HubClientBase.cs
@@ -1,51 +1,28 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
 using System;
 using System.Collections.Generic;
-using System.Text;
 
 namespace Microsoft.AspNetCore.SignalR
 {
-    public class HubClientBase : HubClientsBase<IClientProxy>
+    public abstract class HubClientsBase : HubClientsBase<IClientProxy>
     {
-        public override IClientProxy All => throw new NotImplementedException();
+        public override abstract IClientProxy All { get; }
+        public override abstract IClientProxy AllExcept(IReadOnlyList<string> excludedConnectionIds);
 
-        public override IClientProxy AllExcept(IReadOnlyList<string> excludedConnectionIds)
-        {
-            throw new NotImplementedException();
-        }
+        public override abstract IClientProxy Client(string connectionId);
 
-        public override IClientProxy Client(string connectionId)
-        {
-            throw new NotImplementedException();
-        }
+        public override abstract IClientProxy Clients(IReadOnlyList<string> connectionIds);
 
-        public override IClientProxy Clients(IReadOnlyList<string> connectionIds)
-        {
-            throw new NotImplementedException();
-        }
+        public override abstract IClientProxy Group(string groupName);
 
-        public override IClientProxy Group(string groupName)
-        {
-            throw new NotImplementedException();
-        }
+        public override abstract IClientProxy GroupExcept(string groupName, IReadOnlyList<string> excludedConnectionIds);
 
-        public override IClientProxy GroupExcept(string groupName, IReadOnlyList<string> excludedConnectionIds)
-        {
-            throw new NotImplementedException();
-        }
+        public override abstract IClientProxy Groups(IReadOnlyList<string> groupNames);
 
-        public override IClientProxy Groups(IReadOnlyList<string> groupNames)
-        {
-            throw new NotImplementedException();
-        }
+        public override abstract IClientProxy User(string userId);
 
-        public override IClientProxy User(string userId)
-        {
-            throw new NotImplementedException();
-        }
-
-        public override IClientProxy Users(IReadOnlyList<string> userIds)
-        {
-            throw new NotImplementedException();
-        }
+        public override abstract IClientProxy Users(IReadOnlyList<string> userIds);
     }
 }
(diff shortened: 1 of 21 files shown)


src/SignalR/server/Core/src/HubClientsBase`T.cs
```diff
@@ -0,0 +1,77 @@
...
+
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.SignalR
+{
+    /// <summary>
+    /// An abstraction that provides access to client connections.
+    /// </summary>
+    /// <typeparam name="T">The client invoker type.</typeparam>
+    public abstract class HubClientsBase<T>
```
halter73: What do we think about removing the Base suffix from all these abstract classes? So this would just be `HubClients<T>`, `HubCallerClientsBase<T>` would be `HubCallerClients<T>` etc...

halter73: Also, even though it's obsolete, this should implement `IHubClients<T>`. All the new abstract base classes should implement the interfaces they're replacing for max compatibility until the interfaces are fully removed.

davidfowl:
> I obsoleted the IHubClients and IHubCallerClients interfaces and created abstract base classes to replace them.

Talk to me about this. Lets *not* do that, just have the abstract classes implement the interfaces. This should be as non-breaking as possible. I've discussed the end-to-end design with @halter73 but I dunno if we wrote it down anywhere.

halter73:
> I've discussed the end-to-end design with @halter73 but I dunno if we wrote it down anywhere.

Here's the original misguided PR: https://github.com/aspnet/AspNetCore/pull/7762

Here's what we talked about on the whiteboard:

HubClients-whiteboard

src/SignalR/server/Core/src/HubClientsBase`T.cs
mikaelm12: > So this would just be HubClients<T>, HubCallerClientsBase<T> ...

We already have the concrete class `HubCallerClients`. That's why I added the base suffix.

mikaelm12:
Not gonna lie, I can't read this.

halter73:
Part of the reason I stopped working on this is that I learned @davidfowl wanted to support [adding a second generic parameter for single clients (e.g. Caller, not All) which allows for interfaces with value-returning methods](https://github.com/aspnet/AspNetCore/issues/5382#issuecomment-466608984).

It's basically what you did plus thinking about how we can add support for invoking value-returning methods on a single client via `Hub(Caller)Clients<TAll, TSingle>` in a non-breaking way. TAll would be an interface type with all void-returning methods, and TSingle would be allowed to have methods with return types.

The non-generic Hub(Caller)Clients would implement `Hub(Caller)Clients<IClientProxy, IClientProxy>`. To invoke value-returning methods on a single client, you'd have to derive from `Hub<TAll, TSingle>`. You won't be able to invoke value-returning methods on a single client with a non-gerneric hub.

Before merging, we should see what it looks like to add the types and methods necessary to support invoking value-returning single-client methods. Since we'd be creating a new `Hub<TAll, TSingle>` type that would likely derive directly from `Hub` instead of `Hub<T>`, I think we'll be OK.

It's funny because we'll see little direct benefit in terms of back-compat from switching Hub(Caller)Clients to abstract base classes from interfaces, since supporting value-returning methods on the caller will require a second generic parameter which will break either the abstract base class or the interface. If we ever want to add a new default method to HubClients, I guess it helps there, but with default interface methods I'm still not completely sure how critical it is.

I'm wondering if we should just punt this abstract base class refactoring until we start actually working on client return values. Although, I guess I don't see a huge harm in adding the abstract base classes aside from a tiny bit of extra complexity.

src/SignalR/server/Core/src/HubClientBase.cs
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) .NET Foundation. All rights reserved.
```
halter73: This filename is wrong. It's HubClientBase instead of HubClientsBase.

src/SignalR/server/Core/src/Internal/HubClients`T.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace Microsoft.AspNetCore.SignalR.Internal
 {
-    internal class HubClients<THub, T> : IHubClients<T> where THub : Hub
+    internal class HubClients<THub, T> : HubClientsBase<T> where THub : Hub
```
halter73: This class the reason you didn't want to name HubClientsBase HubClients? Just rename this one to DefaultHubClients. It's internal and in another namespace so the HubClients name isn't burned. We should use it for the public type.

halter73:
>  If we ever want to add a new default method to HubClients, I guess it helps there, but with default interface methods, I'm still not completely sure how critical it is.

Never mind about the default interface methods. It seems that's still in the [prototype phase](https://github.com/dotnet/csharplang/blob/master/proposals/default-interface-methods.md). Given that, I agree with @davidfowl about switching to use abstract base classes without deprecating the interface for 3.0.

halter73:
@DamianEdwards Says default interface methods are slated to arrive in preview5. In that case, I'm back to thinking we should close this PR. Keep the branch around in case default interface methods don't make it though.

anurse:
Let's close this then and defer action until Preview 6.

Issue #9092, opened 2019-04-04
OAuth authentication has no way to intercept the Token request


Menighin:
### Is your feature request related to a problem? Please describe.
I'm trying to implement OAuth authentication with PKCE. 
The PKCE part requires that I send a `code_verifier` with the token request but this doesn't seem to be possible. I couldn't find anyway of intercepting the Token Request in order to put the `code_verifier`. As far as I understood, AspNetCore automatically get the authorization code and request for a token with it, all under the hood.

### Describe the solution you'd like
To have an event like `OnAuthorizationCodeReceived` like we do on the `OpenId` middleware. Or maybe an even more specific, something like `OnRequestForToken`.

### Describe alternatives you've considered
Implement all from scratch or use a 3rd party lib (any recomendations?)

### Additional context
I'm trying to implement following [this tutorial](https://www.scottbrady91.com/OpenID-Connect/ASPNET-Core-using-Proof-Key-for-Code-Exchange-PKCE) that uses `OpenId`. I couldn't find anything like the `OnAuthorizationCodeReceived` event on the OAuth so I'm stuck in a scenario where I request for authorization with `code_challenge` but I can't get the token because I can´t find a way to put the `code_ verifier` in the request.

Thanks!

Tratcher:
The recommendation for customizing the code exchange is to derive from OAuthHandler and override ExchangeCodeAsync. E.g. https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/blob/af1e3373cf00e82b186ea2a4bba0696d476b0dff/src/AspNet.Security.OAuth.Fitbit/FitbitAuthenticationHandler.cs#L64

Menighin:
Thanks Tratcher! That indeed seems like a more concise and smart way of doing it :)
Still think it would be nice to have and event on this step though!

Menighin:
@Tratcher I was giving this a try and I'm kinda stuck with one thing:
I must create a new `code_verifier` for each request. Meaning that this is dynamic and not an option of the `AuthenticationHandler`. This `code_verifier` is created before the challenge request (right now I have override the `HandleChallengeAsync` and put the logic there). Later on, when I get the authorization code on the `ExchangeCodeAsync` and will request for the token, how can I get the same `code_verfier`  I used before?

[...]

PR #9448, opened 2019-04-17
OAuth - Creating event for ExchangeCode

Menighin:
Summary of the changes
 - Created new event `OnExchangeCode` with default implementation it had before
 - Created `OAuthExchangeCodeContext` to hold properties necessary for developer to customize their own `ExchangeCode` instead of overriding the whole `OAuthHandler`.

Addresses #9092 (Use case described on the issue)


src/Security/Authentication/OAuth/src/Events/OAuthEvents.cs
```diff
@@ -25,6 +29,41 @@ public class OAuthEvents : RemoteAuthenticationEvents
             return Task.CompletedTask;
         };
 
+        /// <summary>
+        /// Gets or sets the delegate that is invoked when the ExchangeCode method is invoked.
+        /// </summary>
+        public Func<OAuthExchangeCodeContext, Task<OAuthTokenResponse>> OnExchangeCode { get; set; } = async context =>
```
Tratcher: This is more code than is appropriate for events. Deriving from OAuthHandler is more appropriate once you need to replace this much.

Is there a more specific event that would still be useful? Like adding an event on line 45 that passed the parameters dictionary so it could be augmented? The default implementation would no-op. Would that be enough for your PKCE example?

src/Security/Authentication/OAuth/src/OAuthHandler.cs
```diff
@@ -159,42 +159,10 @@ protected override async Task<HandleRequestResult> HandleRemoteAuthenticateAsync
             }
         }
 
-        protected virtual async Task<OAuthTokenResponse> ExchangeCodeAsync(string code, string redirectUri)
+        protected virtual async Task<OAuthTokenResponse> ExchangeCodeAsync(string code, string redirectUri, AuthenticationProperties properties)
```
Tratcher: Is there a non-breaking way to add this new parameter? E.g. with a new overload?

Tratcher (changes requested):

src/Security/Authentication/OAuth/src/OAuthHandler.cs
Menighin: @Tratcher hi, I was trying to make this non-breaking but I really don't see how... An overload  won't work because now it is called from `HandleRemoteAuthenticateAsync` with the `properties` parameter. So to make, it non breakable I would need the new method to call the old one. And that is not possible because I would need to pass the `tokenRequestParameters` setted in the new method to the old one... I couldn't think of anything else, do you see a way out of this? How bad is this breaking change? n_n'

Tratcher: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/search?q=ExchangeCodeAsync&unscoped_q=ExchangeCodeAsync
It would break at least 9 providers, which is enough to try to mitigate. Let's think about it while we deal with the other issue.

Menighin: Ok, the first one about simply augmenting the Dictionary, I totally agree with you and I already changed it.
Will be giving some thought about how to not break it... But I don't see now how to make it without some smelly artifact...

Tratcher (commented):
Better, I added some cleanup comments. I'll follow up on the API break.

Tratcher:
@PinpointTownes 

So I looked at the 9 known providers that override ExchangeCode to see why.
https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/search?q=ExchangeCodeAsync&unscoped_q=ExchangeCodeAsync

Summary:
reddit: Move the clientid & secret to the auth header, optional user agent
fitbit: same as reddit
yandex: same as reddit
yahoo: same as reddit
visual studio: weird request parameter names, values
weixin: get+query
stackexchange: x-www-form-urlencoded response
yammer: weird response format
qq: get+query & form response

I'd be more willing to take the break API if I thought they could take advantage of the change and the new event. The Visual Studio provider looks like the only one that could remove it's overload and use the event instead to remap parameters. That tells me the event might not be general enough. We also might need to add first class support for some of these other scenarios. 

A full half of the providers made the same customization (reddit-like), moving the client id and secret to the auth header. If we made that a first class option then they could remove their overrides.

The other providers are unique enough and mess with both the request and response so neither first classing their behavior nor the new event seem helpful.

src/Security/Authentication/OAuth/src/Events/OAuthEvents.cs
```diff
@@ -37,5 +42,13 @@ public class OAuthEvents : RemoteAuthenticationEvents
         /// </summary>
         /// <param name="context">Contains redirect URI and <see cref="AuthenticationProperties"/> of the challenge.</param>
         public virtual Task RedirectToAuthorizationEndpoint(RedirectContext<OAuthOptions> context) => OnRedirectToAuthorizationEndpoint(context);
+
+        /// <summary>
+        /// Invoked before the request to exchange the code for the access token.
+        /// </summary>
+        /// <param name="context">Contains the code returned, the redirect URI and the <see cref="AuthenticationProperties"/>.</param>
+        /// <returns></returns>
+        public virtual Task ExchangeCode(OAuthExchangeCodeContext context) => OnExchangeCode(context);
```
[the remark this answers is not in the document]
Menighin: I think BeforeExchangeCode is better, what you think?

PinpointTownes:
@Tratcher supporting basic authentication in both the OAuth2 and OIDC handlers would be an excellent idea, since it's a common source of pain and could help cut some code in the aspnet-contrib providers.

I wonder if this event should also allow handling the code redeeming dance manually, pretty much like the OIDC handler with its `AuthorizationCodeReceived` event. Is there a particular reason not to align the two handlers?

Tratcher:
FYI we're considering first class PKCE support that would make this unnecessary. https://github.com/aspnet/AspNetCore/issues/7734

src/Security/Authentication/OAuth/src/OAuthHandler.cs
brentschmaltz: @Tratcher is it possible to add a new API ExchangeCodeAsnyc(ExchangeCodeContext context)?

This would future proof us a bit. We have used this in other places.

Tratcher: @brentschmaltz yes, that's plausible. Either way it's a breaking change for many providers. If we added PKCE support directly we might be able to avoid this break. https://github.com/aspnet/AspNetCore/pull/9448#issuecomment-496999910

Tratcher:
I don't know how neccissary this will be after https://github.com/aspnet/AspNetCore/pull/10928.

Tratcher:
Please rebase on https://github.com/aspnet/AspNetCore/pull/10928 and we can see what's left.

Tratcher:
Closing this for now due to the unresolved merge conflict and design questions. We can revisit in the future.

Issue #6882, opened 2019-01-20
Update Blazor (ASP.NET Core Hosted) project template so that it is easy to toggle between client-side and server-side Blazor


danroth27:
The work that remains here is to break up `UseBlazor` into more parts so that it looks similar to the patterns for server-side.

StevenTCramer:
@rynowak  looks like you are going this direction. (One configuration for both client and sever side)

https://github.com/TimeWarpEngineering/blazor-state/blob/master/test/TestApp/TestApp.Client/wwwroot/BlazorDualMode.js

Gives the ability to select client-side or server-side Blazor at run-time without recompiling.

https://github.com/TimeWarpEngineering/blazor-state/blob/master/test/TestApp/TestApp.Server/Startup.cs

This allows the same server to handle either type of client.  The default usage is first visit the client operates server-side second-visit when all is cached it switches to client side.   But all kinds of logic could be put in this to determine dynamically where is the best user experience.

https://thefreezeteam.com/razor-components-dynamic-dual-mode/

This is a very nice use case that I just hope we can maintain :)

Andrzej-W:
Will this enable server side prerendering for Blazor Client Side app?

MaherJendoubi:
I am facing the following regression after using this updated template https://github.com/ElectronNET/Electron.NET/issues/277

PR #9449, opened 2019-04-17
Fix: #6882 - Explode UseBlazor

rynowak:
This design concept splits up UseBlazor into a usage of static files and
a usage of MapFallback.

---

I ignored auto-rebuild support because it's currently dead code until we
have that in VS. If we add auto-rebuild to ASP.NET Core - we'd probably want
to make that a separate gesture inside `IsDevelopement()` like other
dev-time features anyway.

---

The static-files wireup is pretty thick, and comes with some caveats.
This requires you to use the StaticFiles middleware without passing in
an explicit options instance. This gives us parity with what we're doing
for server-side, but it still gives me a bit of pause. Note that this
supports multiple calls to include multiple projects.

Another approach would be to make a separate `UseClientAppFiles()`
middleware. This wouldn't be quite so fragile.

Basically the approach in the PR optimizes the Startup code being
simpler but requiring a little more explanation. We could optimize the
other way by making an explicit middleware.

---

I feel pretty good about the wireup with routing to use the
`index.html` from the client app. I think it's pretty to-the-point.

---

Canonical sample:

```C#
app.UseClientSideBlazorFiles(FindClientAssemblyPath(environment));

app.UseRouting();

app.UseEndpoints(endpoints =>
{
    endpoints.MapFallbackToClientSideBlazor(FindClientAssemblyPath(environment), "index.html");
});
```

I'm not officially proposing these method names. I want to pick something to get the discussion going about this approach and whether it feels good. it certainly is pretty *consistent* with server-side.


javiercn:
What do you need TProgram for in Map Blazor Client-Side?

I hope/expect that to go away in the future, we shouldn’t require loading/referencing an assembly we don’t need.

I expect razor client-side projects to just be static files on the hosted app by the time we publish. Does that make sense?

rynowak:
> What do you need TProgram for in Map Blazor Client-Side?

Client side blazor outputs a config file that's includes file paths for this scenario. Using the type/assembly to find that file is the sanest way to do what it does today.

Personally, I've love it if your work solved all of these problems for us, but I don't consider removing `BlazorConfig` and all of the smarts that it has part of this change.

Also aren't you on vacation?

SteveSandersonMS:
This looks like a really good set of improvements to me. However, your other suggestion about UseClientAppFiles sounds even better I think. 

The fragility of requiring UseStaticFiles without options is my main concern. 

Do you think it would work to merge the index.html fallback into UseClientAppFiles? That feels intuitively good to me, since it’s all about static content. Obviously it would have to be optional in case you wanted to fall back to a cshtml instead. Example:

```
        {
            services.AddBlazorClient<Client.Program>();
        }

        // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
        public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
        {
            if (env.IsDevelopment())
            {
                app.UseBlazorClientDebugging();
            }

            app.UseStaticFiles();

            app.UseRouting();

            app.UseEndpoints(endpoints =>
            {
                endpoints.MapBlazorClientFiles<Client.Program>(fallbackToIndexHtml: true);
            });
        }
```

SteveSandersonMS:
Actually does that not work because UseEndpoints is not a legit place to attach a further UseStaticFiles middleware?

If so then maybe outside the UseEndpoints block:

    app.UseBlazorClientFiles<Program>(fallbackToIndexHtml: true);

Just wondering if this really is reducible to a one-liner while honouring the design of endpoint routing. If not, I’m ok with multiple calls!

rynowak:
> Actually does that not work because UseEndpoints is not a legit place to attach a further UseStaticFiles middleware?

We've come up with various plans to use static files + routing in the past, but we're not planning to add anything in this release.

For now I don't think it makes sense to make this a one-liner. We'd either be doing something odd with static file service, or losing the "fallback" behaviour.

SteveSandersonMS:
Ok, so we’d have both UseBlazorClientFiles and MapFallbackToBlazorClient? If so, fine with me.

rynowak:
Updated sample in the main post.

I'm looking for feedback on a few points:
- method names
- does it feel good to see `index.html` listed explicitly? I'm trying to slim down the number of overloads
- If we don't want `index.html` do we need to make it configurable? It was not in the past

I added overloads to `MapFallbackToClientSideBlazor` to accept a route pattern. This allows you to use it with multiple different routes if you have multiple entry points. You'd currently need to use `Map()` with the middleware as well.

rynowak:
And yes, I know codecheck is going to fail :) I want to get feedback on the APIs before I bother updating ref.

src/Components/Blazor/DevServer/src/Server/Startup.cs
```diff
@@ -30,15 +31,23 @@ public void ConfigureServices(IServiceCollection services)
...
         {
             app.UseDeveloperExceptionPage();
             app.UseResponseCompression();
             EnableConfiguredPathbase(app, configuration);
 
-            var clientAssemblyPath = FindClientAssemblyPath(app);
-            app.UseBlazor(new BlazorOptions { ClientAssemblyPath = clientAssemblyPath });
             app.UseBlazorDebugging();
+
+            app.UseStaticFiles();
```
SteveSandersonMS: Is this line needed? I didn't think the dev server had any static files of its own to serve.

src/Components/Blazor/Server/src/Builder/BlazorHostingApplicationBuilderExtensions.cs
```diff
@@ -0,0 +1,98 @@
...
+            }
+
+            var options = new StaticFileOptions()
+            {
+                ContentTypeProvider = contentTypeProvider,
+                FileProvider = new CompositeFileProvider(fileProviders),
+                OnPrepareResponse = CacheHeaderSettings.SetCacheHeaders,
+            };
+
+            app.UseStaticFiles(options);
```
SteveSandersonMS: It's good that we're registering a separate `StaticFilesMiddleware` rather than trying to hook into the developer's own one, because it allows the `FileExtensionContentTypeProvider` rules and caching headers to be different. Not sure that would be possible without having a separate `StaticFilesMiddleware`.

SteveSandersonMS:
Just to check - did you verify this all works post-publishing? We don't have E2E tests for client-side Blazor's post-publishing state, though we do for server-side Blazor.

src/Components/Blazor/Server/src/Builder/BlazorHostingEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,179 @@
...
+    /// </summary>
+    public static class BlazorHostingEndpointRouteBuilderExtensions
+    {
+        /// <summary>
+        /// Adds a low-priority endpoint that will serve the the file specified by <paramref name="filePath"/> from the client-side
+        /// Blazor application specified by <typeparamref name="TClientApp"/>.
+        /// </summary>
+        /// <typeparam name="TClientApp">A type in the client-side application.</typeparam>
+        /// <param name="endpoints">The <see cref="IEndpointRouteBuilder"/>.</param>
+        /// <param name="filePath">The relative path to the entry point of the client-side application.</param>
```
SteveSandersonMS: This description might need tweaking. What is the path relative to? I think it's to the `wwwroot` directory within the client-side application - is that right?

src/Components/Blazor/Server/src/Builder/BlazorHostingEndpointRouteBuilderExtensions.cs
```diff
@@ -0,0 +1,179 @@
...
+        /// <returns>The <see cref="IApplicationBuilder"/>.</returns>
+        /// <remarks>
+        /// <para>
+        /// This method is intended to handle cases where URL path of the request does not contain a filename, and no other
+        /// endpoint has matched. This is convenient for routing requests for dynamic content to the client-side blazor
+        /// application, while also allowing requests for non-existent files to result in an HTTP 404.
+        /// </para>
+        /// </remarks>
+        public static IEndpointConventionBuilder MapFallbackToClientSideBlazor<TClientApp>(this IEndpointRouteBuilder endpoints, string pattern, string filePath)
```
SteveSandersonMS: `pattern` isn't used anywhere in the method, besides the null check.

It would be possible to add E2E tests for these methods by having a bunch of different `Startup` classes in `BasicTestApp`, and then `TestServer` could map different subdirectories to different startup classes. Currently it maps `/subdir` to the main `Startup` class.

SteveSandersonMS:
Looks even better to me now :) I really like this design.

> does it feel good to see index.html listed explicitly? I'm trying to slim down the number of overloads

Yes, I think that is good. Seeing `index.html` in the call makes it seem slightly less magic and provides a strong hint about what it's really doing.

The number of overloads of `MapFallbackToClientSideBlazor` is already large, so I support not doubling that. If there are any overloads that we think are not core scenarios, I'd be in favour of dropping them and requiring people in non-core scenarios to use the uber-overload that requires you to specify everything. Though TBH the only one I see that looks like a candidate for dropping is `MapFallbackToClientSideBlazor(this IEndpointRouteBuilder endpoints, string clientAssemblyFilePath, string filePath)`.

> I added overloads to MapFallbackToClientSideBlazor to accept a route pattern

I'm neutral about that. Using `Map` or `MapWhen` to achieve this doesn't seem problematic to me.

> method names

The ones you have look good.

Possibly consider renaming `MapFallbackToClientSideBlazor` to `MapFallbackToClientSideBlazorFile`, since it literally does only map fallback to a single file, not to the whole contents of the project as it seems to imply.

However that is a really long method name. I do like the following two shortenings:

 * `UseBlazorClientFiles`
 * `MapFallbackToBlazorClientFile`

... but I accept that would be less consistent with the "client-side Blazor" branding and sacrifices symmetry with `AddServerSideBlazor`.

src/Components/Blazor/Server/src/Builder/BlazorHostingApplicationBuilderExtensions.cs
rynowak: yeah right now that's not possible.

rynowak:
> Just to check - did you verify this all works post-publishing? We don't have E2E tests for client-side Blazor's post-publishing state, though we do for server-side Blazor.

No, this is still a draft PR. I was waiting for feedback before doing things like adding new tests and final verification.

src/Components/Blazor/Server/src/Builder/BlazorHostingEndpointRouteBuilderExtensions.cs
rynowak: yes. I'll clarify

src/Components/Blazor/Server/src/Builder/BlazorHostingEndpointRouteBuilderExtensions.cs
rynowak: fixed the bug. 

Will add tests for these all.

rynowak:
> I'm neutral about that. Using Map or MapWhen to achieve this doesn't seem problematic to me.

I like having the methods that take the route pattern because it's much harder to reason about mixing Map and routing. Your code ends up getting complex for something we could do easily by taking the route pattern.

rynowak:
@SteveSandersonMS - updated. I kept the naming as-is for consistency with the server side story.

src/Components/test/E2ETest/Tests/ClientSideHostingTest.cs
```diff
@@ -0,0 +1,75 @@
...
+        {
+            return TestServer.Program.BuildWebHost<TestServer.StartupWithMapFallbackToClientSideBlazor>(args);
+        }
+
+        [Fact]
+        public void MapFallbackToClientSideBlazor_FilePath()
+        {
+            Navigate("/filepath", noReload: true);
+            WaitUntilLoaded();
+            Assert.NotNull(Browser.FindElement(By.Id("test-selector")).Text);
```
SteveSandersonMS: This assertion is a bit puzzling to me. Why check the nullness of the element's text property? Why are we checking something about the test selector anyway?

I would have expected something like checking that the "Hello, world!" heading has been rendered instead.

Same with other tests below.

src/Components/test/E2ETest/Tests/ClientSideHostingTest.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        private IWebHost BuildWebHost(string[] args)
+        {
+            return TestServer.Program.BuildWebHost<TestServer.StartupWithMapFallbackToClientSideBlazor>(args);
+        }
+
+        [Fact]
+        public void MapFallbackToClientSideBlazor_FilePath()
+        {
+            Navigate("/filepath", noReload: true);
```
SteveSandersonMS: I think we should have `noReload: false` in all these cases.

I know that in practice, it probably is going to reload in all cases because the URL is changing, but we don't need to couple the tests to that inference. We do want it to reload in all cases so that the startup logic is tested.

Also I appreciate it's a bit annoying that the condition `noReload` is phrased negatively and it's unclear what the default is.

SteveSandersonMS (approved):
Looks super! Glad to see how simple you were able to make the E2E tests.

src/Components/test/E2ETest/Tests/ClientSideHostingTest.cs
rynowak: Sure, that's fine. I just wanted to make it something that won't change.

rynowak: This page doesn't have a heading, it just has that select. I'm updating this to just drop `.Text` to make it clear that it's about seeing that element come up. I chose this because it's used in so many tests - which makes it very unlikely these tests would break without something going wrong.

PR #9475, opened 2019-04-17
Implement MaxRequestBodySize feature for IIS inprocess

### LINKED ISSUE #8293: This server does not support the IHttpRequestBodySizeFeature
I have an MVC action decorated with `RequestSizeLimitAttribute`. However, I'm seeing the following message in my logs every time I hit that route: `A request body size limit could not be applied. This server does not support the IHttpRequestBodySizeFeature`. 

I'm running ASP.NET Core 2.2 with IIS on Windows Server 2016.

I couldn't manage to find not even a single mention of this in Google. I'm guessing it's because I'm running the app behind IIS. If that's the case, does that mean that _only_ web.config limits are supported?

jkotalik:
Fixes https://github.com/aspnet/AspNetCore/issues/8293.

I'm okay if this slips to preview6. I'll leave comments in the diff.


src/Servers/IIS/IIS/src/BadHttpRequestException.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.IO;
+using System.Runtime.CompilerServices;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNetCore.Server.IIS
+{
+    public sealed class BadHttpRequestException : IOException
+    {
+        private BadHttpRequestException(string message, int statusCode, RequestRejectionReason reason)
```
jkotalik: I didn't like having IIS use the Kestrel type for BadHttpRequestException, so I copied it here. I'm only using one exception type right now.

src/Servers/IIS/IIS/src/RequestRejectionReason.cs
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Server.IIS
+{
+    internal enum RequestRejectionReason
```
jkotalik: This was also copied (again I only use one value, so I can probably reduce it).

src/Servers/IIS/IIS/test/Common.FunctionalTests/Inprocess/MaxRequestBodySizeTests.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    [Collection(PublishedSitesCollection.Name)]
+    public class MaxRequestBodySizeTests : IISFunctionalTestBase
+    {
+        public MaxRequestBodySizeTests(PublishedSitesFixture fixture) : base(fixture)
+        {
+        }
+
+        [ConditionalFact]
+        public async Task MaxRequestBodySizeE2EWorks()
```
jkotalik: This is an E2E test that uses IIS and IISExpress. Other tests use HostableWebCore. This is more of a sanity check that the feature works E2E

anurse:
Moving this out of preview 5. We are starting to limit preview 5 work to essential stuff only.

src/Servers/IIS/IIS/src/BadHttpRequestException.cs
```diff
@@ -0,0 +1,44 @@
...
+            : base(message)
+        {
+            StatusCode = statusCode;
+            Reason = reason;
+        }
```
halter73: nit: at least copy the newline 😉

src/Servers/IIS/IIS/src/Core/IISHttpContextOfT.cs
```diff
@@ -47,7 +53,7 @@ public override async Task<bool> ProcessRequestAsync()
             {
                 _streams.Stop();
 
-                if (!HasResponseStarted && _applicationException == null && _onStarting != null)
+                if (!HasResponseStarted && _requestRejectedException == null && _applicationException == null && _onStarting != null)
```
halter73: Kestrel doesn't check _requestRejectedException before calling FireOnStarting post _application.ProcessRequestAsync(). Should it?

src/Servers/IIS/IIS/src/RequestRejectionReason.cs
halter73: Reduce it. It's internal.

src/Servers/IIS/IIS/test/IIS.Tests/MaxRequestBodySizeTests.cs
```diff
@@ -0,0 +1,258 @@
...
+            using (var testServer = await TestServer.Create(
+                async ctx =>
+                {
+                    var buffer = new byte[1];
+                    requestRejectedEx1 = await Assert.ThrowsAsync<BadHttpRequestException>(
+                        async () => await ctx.Request.Body.ReadAsync(buffer, 0, 1));
+                    requestRejectedEx2 = await Assert.ThrowsAsync<BadHttpRequestException>(
+                        async () => await ctx.Request.Body.ReadAsync(buffer, 0, 1));
+                    throw requestRejectedEx2;
+                }, LoggerFactory, new IISServerOptions { MaxRequestBodySize = 0 }))
```
halter73: Do we verify that "Content-Length: 0" and empty chunked request bodies are still allowed with MaxRequestBodySize = 0?

halter73 (approved):

jkotalik:
I'm going to block myself on merging this until I figure out why a set of IIS tests aren't running on PRs.

src/Servers/IIS/IIS/src/Core/IISHttpContextOfT.cs
jkotalik: I think it should, but it would be a "breaking" behavioral change. I'm going to remove the check for IIS for now and we can discuss adding it to both later.

src/Servers/IIS/IIS/test/IIS.Tests/MaxRequestBodySizeTests.cs
jkotalik: Added tests.

jkotalik:
Re-enabled some IIS tests, this should now report success/failures correctly now.

src/Servers/IIS/IIS/src/Core/IISHttpContext.IO.cs
```diff
@@ -96,12 +98,19 @@ private async Task ReadBody()
 
                     var read = await AsyncIO.ReadAsync(memory);
 
+                    _consumedBytes += read;
```
Tratcher: This needs to go inside the `if (read != -1)` check, and then MaxRequestBodySize needs to move down.

src/Servers/IIS/IIS/test/IIS.Tests/MaxRequestBodySizeTests.cs
```diff
@@ -0,0 +1,309 @@
...
+                        "");
+                    await connection.Receive("HTTP/1.1 413 Payload Too Large");
+                }
+            }
+
+            Assert.Equal(CoreStrings.BadRequest_RequestBodyTooLarge, exception.Message);
+        }
+
+        [ConditionalFact]
+        public async Task DoesNotRejectRequestWithContentLengthHeaderExceedingGlobalLimitIfLimitDisabledPerRequest()
```
Tratcher: And chunked?

Tratcher (changes requested):
I also wonder if we can integrate closer with maxAllowedContentLength, or even override it? Let's talk offline.

src/Servers/IIS/IIS/src/Core/IISHttpContext.IO.cs
halter73: Nice catch. Is there anyway we could create a regression test for this?

jkotalik: IMO too hard, it would require me figuring out a way to consistently make IIS return an exception on every read. I don't know a way to make that happen consistently.

jkotalik:
This now adds support for reading the IIS content length limit and logs a warning per request/in the constructor for the Server.

src/Servers/IIS/AspNetCoreModuleV2/InProcessRequestHandler/managedexports.cpp
```diff
@@ -216,6 +217,7 @@ http_get_application_properties(
 
     auto const serverAddresses = BindingInformation::Format(pConfiguration.QueryBindings(), pInProcessApplication->QueryApplicationVirtualPath());
     pIISCofigurationData->pwzBindings = SysAllocString(serverAddresses.c_str());
+    pIISCofigurationData->maxRequestBodySize = pInProcessApplication->QueryConfig().QueryMaxRequestBodySizeLimit();
```
anurse: `pIISCofigurationData`

A man pauses for a second and then his eye twitches

anurse: I won't make you change it and cycle another build if people approve this as-is, but if you go back in to this code, a rename to fix the typo would be nice ;P

jkotalik: `if(`

anurse: I have a problem. But I also have a [solution?](https://github.com/aspnet/Extensions/pull/1654)

src/Servers/IIS/IIS/src/CoreStrings.resx
```diff
@@ -147,4 +147,22 @@
...
+    <value>The maximum request body size cannot be modified after the request has been upgraded.</value>
+  </data>
+  <data name="NonNegativeNumberOrNullRequired" xml:space="preserve">
+    <value>Value must be null or a non-negative number.</value>
+  </data>
+  <data name="BadRequest" xml:space="preserve">
+    <value>Bad request.</value>
+  </data>
+  <data name="MaxRequestLimitWarning" xml:space="preserve">
+    <value>Increasing the MaxRequestBodySize conflicts with the max value for IIS limit maxAllowedContentLength. HTTP requests that have a content length greater than maxAllowedContentLength will still be rejected by IIS. You can disable the limit by either removing or setting the maxAllowedContentLength value to a higher limit.</value>
```
Tratcher: Needs to say where to set maxAllowedContentLength (web.config, what section?). Linking to a doc would help too.

src/Servers/IIS/AspNetCoreModuleV2/InProcessRequestHandler/InProcessOptions.cpp
```diff
@@ -71,6 +72,16 @@ InProcessOptions::InProcessOptions(const ConfigurationSource &configurationSourc
     const auto anonAuthSection = configurationSource.GetSection(CS_ANONYMOUS_AUTHENTICATION_SECTION);
     m_fAnonymousAuthEnabled = anonAuthSection && anonAuthSection->GetBool(CS_ENABLED).value_or(false);
 
+    const auto maxRequestBodySize = configurationSource.GetSection(CS_MAX_REQUEST_BODY_SIZE_SECTION);
```
Tratcher: I'll ask you offline about when this is present, absent, etc..

Tratcher (approved):

src/Servers/IIS/IIS/src/CoreStrings.resx
jkotalik: Yeah, I think I'll make a retroactive PR to link in a doc here.

src/Servers/IIS/AspNetCoreModuleV2/InProcessRequestHandler/InProcessOptions.cpp
jkotalik: I actually think this may be trickier than I thought. Default values won't be present here.

jkotalik: Fixed this now. If the section is present, it means the feature is installed at a minimum.

jkotalik: So we can default it 30MB if the feature is enabled but doesn't have a value aka default.

jkotalik:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=180325

aspnet-hello:

discostu105:
We just upgraded to preview-6 and suddenly our web.config setting () seems to not be working anymore. We run on IIS in-process. Here is the config: https://github.com/Dynatrace/superdump/blob/master/src/SuperDumpService/web.config

Anything wrong with the config at first glance? Should we file an issue for this? @jkotalik

Tratcher:
>  seems to not be working anymore

What is the error?

discostu105:
the setting from web.config is not respected anymore, so it uses a much lower default.

anurse:
If you're seeing the setting be ignored, definitely file a new bug and provide a runnable sample that reproduces the problem. We don't track merged PRs like this so it's likely this will get lost if we stop checking our GitHub notifications ;). Also, include any logs from the app since it looks like we should be logging a warning if there's a conflict between these options.

@jkotalik From a very cursory look at the code, I wonder if the fact that we set a default size limit in the managed code is conflicting here. Perhaps the default (if the setting isn't set) should be to just use whatever the web.config value is (if one is present)?

PR #9476, opened 2019-04-17
Make System.Text.Json the default for SignalR and remove Newtonsoft from shared framework

### LINKED ISSUE #4260: Remove Newtonsoft.Json from the shared framework

### LINKED ISSUE #9439: Make System.Text.Json SignalR Protocol package the default
Also, ensure Newtonsoft.Json is removed from the shared framework.

This is blocked on some corefx Json work (cc @BrennanConroy).

BrennanConroy:
Fixes #9439 
Fixes #4260

Review: How do we want to expose json options? I'm currently wrapping the serializer options and passing the values through.

### DIFF
--- a/eng/SharedFramework.External.props
+++ b/eng/SharedFramework.External.props
@@ -68,7 +68,6 @@
 
       If these are needed as direct dependencies, it is okay to change them to ExternalAspNetCoreAppReference and move up into sections above.
     -->
-    <_TransitiveExternalAspNetCoreAppReference Include="Newtonsoft.Json"                                    Version="$(NewtonsoftJsonPackageVersion)" />
     <_TransitiveExternalAspNetCoreAppReference Include="System.Security.Cryptography.Pkcs"                  Version="$(SystemSecurityCryptographyPkcsPackageVersion)" />
     <_TransitiveExternalAspNetCoreAppReference Include="System.Security.Permissions"                        Version="$(SystemSecurityPermissionsPackageVersion)" />
   </ItemGroup>
--- a/src/SignalR/clients/csharp/Client.Core/src/HubConnectionBuilder.cs
+++ b/src/SignalR/clients/csharp/Client.Core/src/HubConnectionBuilder.cs
@@ -25,7 +25,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
             Services = new ServiceCollection();
             Services.AddSingleton<HubConnection>();
             Services.AddLogging();
-            this.AddNewtonsoftJsonProtocol();
+            this.AddJsonProtocol();
         }
 
         /// <inheritdoc />
--- a/src/SignalR/clients/ts/FunctionalTests/ts/HubConnectionTests.ts
+++ b/src/SignalR/clients/ts/FunctionalTests/ts/HubConnectionTests.ts
@@ -457,12 +457,11 @@ describe("hubConnection", () => {
 
                 const complexObject = {
                     ByteArray: protocol.name === "json"
-                        ? "aGVsbG8="
+                        ? new Array(0x68, 0x65, 0x6c, 0x6c, 0x6f)
                         : new Uint8Array([0x68, 0x65, 0x6c, 0x6c, 0x6f]),
                     DateTime: protocol.name === "json"
                         ? "2002-04-01T10:20:15Z"
                         : new Date(Date.UTC(2002, 3, 1, 10, 20, 15)), // Apr 1, 2002, 10:20:15am UTC
-                    GUID: "00010203-0405-0607-0706-050403020100",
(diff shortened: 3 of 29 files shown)


natemcmaster (approved):
Approved w.r.t. shared framework changes.

src/SignalR/clients/ts/FunctionalTests/ts/HubConnectionTests.ts
```diff
@@ -457,12 +457,11 @@ describe("hubConnection", () => {
 
                 const complexObject = {
                     ByteArray: protocol.name === "json"
-                        ? "aGVsbG8="
+                        ? new Array(0x68, 0x65, 0x6c, 0x6c, 0x6f)
```
BrennanConroy: Filed https://github.com/dotnet/corefx/issues/36974 for this

eng/SharedFramework.External.props
```diff
@@ -68,7 +68,6 @@
 
       If these are needed as direct dependencies, it is okay to change them to ExternalAspNetCoreAppReference and move up into sections above.
     -->
-    <_TransitiveExternalAspNetCoreAppReference Include="Newtonsoft.Json"                                    Version="$(NewtonsoftJsonPackageVersion)" />
```
anurse: @pranavkm @rynowak Is MVC good with this change? We are making the switch to System.Text.Json in SignalR by default now.

@ryanbrandenburg Do our templates reference Newtonsoft anywhere or did we already remove all that?

pranavkm: What does this do?

> Do our templates reference Newtonsoft anywhere or did we already remove all that?

MVC's templates still do. We were waiting for https://github.com/dotnet/corefx/issues/36024 to land since it affects quite a few of our tests

pranavkm (approved):

eng/SharedFramework.External.props
anurse: This removes Newtonsoft.Json from the shared framework. We can't do that if MVC isn't ready to do it though.

My concern is that this is going to land super hot for preview 5 and I want everybody on the same page. Maybe we should leave Newtonsoft in the shared framework, even though we're changing the default.

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
```diff
@@ -756,5 +761,18 @@ private HubMessage ApplyHeaders(HubMessage message, Dictionary<string, string> h
...
+
+        internal static JsonSerializerOptions CreateDefaultSerializerSettings()
+        {
+            var options = new JsonSerializerOptions();
+            options.WriteIndented = false;
+            options.ReadCommentHandling = JsonCommentHandling.Disallow;
+            options.AllowTrailingCommas = false;
+            options.IgnoreNullValues = false;
+            options.IgnoreReadOnlyProperties = false;
+            // TODO: camelCase
```
mikaelm12: Tracking this somewhere?

BrennanConroy: Yes, in my head. I'll be doing this today so no need to track it.

eng/SharedFramework.External.props
anurse: Ok, if we can land this soon that would be best. We need to try our templates after this change and make sure we haven't broken the immediate experience.

src/SignalR/clients/csharp/Client.Core/src/HubConnectionBuilder.cs
```diff
@@ -25,7 +25,7 @@ public HubConnectionBuilder()
             Services = new ServiceCollection();
             Services.AddSingleton<HubConnection>();
             Services.AddLogging();
-            this.AddNewtonsoftJsonProtocol();
+            this.AddJsonProtocol();
```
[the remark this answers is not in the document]
joshfree: 😀

mikaelm12 (approved):
File an issue for reviewing the options.

BrennanConroy:
Filed https://github.com/aspnet/AspNetCore/issues/9519 for options review

PR #9477, opened 2019-04-17
partial revert "Flow ANCM version number in the build output (#9269)"

natemcmaster:
This partially reverts commit 01b0c888d2e42674f9dabbbce95eaca149c29917. Sometime between the [successful PR check](https://dev.azure.com/dnceng/public/_build/results?buildId=151625) for #9269 (Apr. 11) and today when the PR was merged, something started causing consistent build failures for ANCM Windows installers.

FYI @aspnet/build - going to merge as soon as this PR check is green.

### DIFF
--- a/korebuild.json
+++ b/korebuild.json
@@ -16,6 +16,8 @@
       "requiredWorkloads": [
         "Microsoft.VisualStudio.Component.VC.ATL",
         "Microsoft.VisualStudio.Component.VC.Tools.x86.x64",
+        "Microsoft.VisualStudio.Component.VC.v141.ATL",
+        "Microsoft.VisualStudio.Component.VC.v141.x86.x64",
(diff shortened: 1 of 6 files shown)


jkotalik (approved):

JunTaoLuo (approved):

natemcmaster:
Updated to be a partial revert. I don't think the part of #9269 that produced a text file with versions in it is the problem. Seems like the real issue was updating the C++ build to v142 toolset. Maybe the conflicting change was merged with #9332 ?

dougbu (approved):

jkotalik:
I don't think it conflicts at all with that change. Those were for local reasons only.

I don't have a full understanding of what version of VS we are using on CI. I know we tried to roll forward to RTM, and are maybe back to a preview now, but the behavior we see on CI is different than what we see locally. I don't see this repro locally.

PR #9482, opened 2019-04-17
Require ClaimsPrincipal.Identity.IsAuthenticated for SignIn by default

### LINKED ISSUE #9255: Consider making SignInAsync throw for an unauthenticated identity
Yeah, I'll admit it. 😢 I burned 1.5 hours trying to use this API and not knowing why it didn't work.

The difference here is between:

```C#
var identity = new ClaimsIdentity(new[] { new Claim("ClaimA", "Value") });
await HttpContext.SignInAsync(scheme: null, new ClaimsPrincipal(identity));
```

```C#
var identity = new ClaimsIdentity(new[] { new Claim("ClaimA", "Value") }, CookieAuthenticationDefaults.AuthenticationScheme);
await HttpContext.SignInAsync(scheme: null, new ClaimsPrincipal(identity));
```

The shorter overload of `ClaimsIdentity` does the wrong thing for most cases. I'm also not sure if we should be validating whether scheme and the authentication scheme passed to `SignInAsync` should match.

----

Since there are use cases for `SignInAsync` with an unauthenticated identity, consider adding more parameters, or using a different method name for that case. Both of these can communicate that you're doing something non-default. 

/cc @HaoK @blowdart

HaoK:
Fixes https://github.com/aspnet/AspNetCore/issues/9255

### DIFF
--- a/src/Http/Authentication.Core/src/AuthenticationService.cs
+++ b/src/Http/Authentication.Core/src/AuthenticationService.cs
@@ -6,6 +6,7 @@ using System.Linq;
 using System.Security.Claims;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Options;
 
 namespace Microsoft.AspNetCore.Authentication
 {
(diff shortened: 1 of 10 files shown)


HaoK:
@rynowak would this exception have helped?

src/Http/Authentication.Core/src/AuthenticationService.cs
```diff
@@ -146,6 +154,14 @@ public virtual async Task SignInAsync(HttpContext context, string scheme, Claims
                 throw new ArgumentNullException(nameof(principal));
             }
 
+            if (Options.RequireAuthenticatedSignIn)
+            {
+                if (principal.Identity == null || !principal.Identity.IsAuthenticated)
```
Tratcher: These error conditions are quite different and warrant different messages.

Tratcher (approved):

src/Http/Authentication.Core/src/AuthenticationService.cs
HaoK: Sure split them into their own

HaoK:
Adding @rynowak and @pranavkm for review since this exposed a small issue with 3 of the existing mvc auth tests, they were returning unauthenticated before, when they really should have been access denied

PR #9487, opened 2019-04-18
Fix AbortedStream_ResetsAndDrainsRequest_RefusesFramesAfterCooldownExpires 

jkotalik:
Fixes https://github.com/aspnet/AspNetCore-Internal/issues/1879.

The reason this test was flaky was due to a race between `WaitForConnectionErrorAsync` calling `Output.Complete()` and the cancellation token check before calling `await SendDataAsync(1, new byte[100], endStream: false);`. The output would be complete and we would try to send data, which would throw because the pipe is complete.

The fix I have is to delay calling Complete until we know the loop exited.

### DIFF
--- a/src/Servers/Kestrel/test/InMemory.FunctionalTests/Http2/Http2TimeoutTests.cs
+++ b/src/Servers/Kestrel/test/InMemory.FunctionalTests/Http2/Http2TimeoutTests.cs
@@ -11,6 +11,7 @@ using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure;
 using Microsoft.AspNetCore.Testing;
 using Microsoft.AspNetCore.Testing.xunit;
+using Microsoft.Extensions.Logging.Testing;
 using Microsoft.Net.Http.Headers;
 using Moq;
 using Xunit;
@@ -193,11 +194,13 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Tests
         }
 
         [Theory]
-        [Flaky("https://github.com/aspnet/AspNetCore-Internal/issues/1879", FlakyOn.All)]
+        [Repeat(20)]
         [InlineData((int)Http2FrameType.DATA)]
         [InlineData((int)Http2FrameType.CONTINUATION)]
         public async Task AbortedStream_ResetsAndDrainsRequest_RefusesFramesAfterCooldownExpires(int intFinalFrameType)
         {
+            var closeLock = new object();
+            var closed = false;
             var finalFrameType = (Http2FrameType)intFinalFrameType;
             // Remove callback that completes _pair.Application.Output on abort.
             _mockConnectionContext.Reset();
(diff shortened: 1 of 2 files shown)


src/Servers/Kestrel/test/InMemory.FunctionalTests/Http2/Http2TimeoutTests.cs
```diff
@@ -193,11 +194,13 @@ public async Task ResponseDrain_SlowerThanMinimumDataRate_AbortsConnection()
         }
 
         [Theory]
-        [Flaky("https://github.com/aspnet/AspNetCore-Internal/issues/1879", FlakyOn.All)]
+        [Repeat(20)]
```
anurse: Is the intent to remove this before checking in?

analogrelay (commented):
Cheer

src/Servers/Kestrel/test/InMemory.FunctionalTests/Http2/Http2TimeoutTests.cs
jkotalik: Ehh I'd say we keep it. The test is fast so it doesn't increase test time significantly. If it is flaky, I'd rather know earlier than later. Eventually we can remove repeats once we deem a test not flaky.

anurse: I'm not opposed to that idea either. I'd like to discuss that a bit though (cc @aspnet/build).

For context, the `Repeat` attribute runs the test multiple times and **requires that it pass EVERY time** (unlike the `Retry` attribute which only requires that it pass **once**).

If the test is fast, I don't mind leaving some `Repeat` attributes in so that we can keep flakiness (if any) visible. However, I also think we need some tracking of test run time to make sure we know this low-hanging fruit is available if these tests start causing PR validation to get slow.

> The test is fast so it doesn't increase test time significantly

Can you get some quick numbers by running just this test from the command line with and without the attribute (use something like PowerShell's `Measure-Command { dotnet test --filter ... }`)? Make sure to run each scenario twice and take the second time to let it warm up and exclude common time-sinks like building. I'd like to know what "doesn't increase test time significantly" is before we let these go in :).

jkotalik: dotnet reports test time 😄 

Total time to run 2 tests 20 times is 3.4910 Seconds. So if we get to the point where we are needing an extra 3 seconds on test runs, we can remove the Repeat attribute.

halter73 (approved):

src/Servers/Kestrel/test/InMemory.FunctionalTests/Http2/Http2TimeoutTests.cs
halter73: I'd remove this. The previous observed flakiness shows this test is already called enough that we should see issues without repeating.

Furthermore this test calls `Task.Delay` in a loop. I get that it's only a millisecond each time, but we don't know how many times this will be repeated. 3.5 seconds is quite a lot for a test that isn't even fully end-to-end. There are *a lot* of HTTP/2 tests. Slow CI machines will likely take longer than the 3.5 seconds you see on your machine.

PR #9488, opened 2019-04-18
Refactor of Startup analyzer code

rynowak:
Independently of this PR, I'm working on a high level plan (in document form) for this area in general. SIT TIGHT.

What's in this PR - cleans up some of the ideas surrounding how this will work overall. The analysis is three phase:

- Find startup classes
- Collect data for `Configure` and `ConfigureServices` methods
- Add diagnostics based on this data

The reasons for this factoring:

- To have good performance we run on what Roslyn wants us to run on. So we centralize the logic to find "interesting" code.
- Also for perf/factoring, we want to separate the collecting of data (what middleware?) from the things that consume that data, so that it's only done once. I think we'll end up with up to 10 different kinds of *analysis* by the end.
- Lastly we expect to have *many* things that run over the computed data to produce diagnostics, this is cheap given that we've already computed all of the state at that point.

This also leads to the pattern of the main analzyer code which is structured around this divison of work. Doing analysis a-type-at-a-time is going to have the best UX and perf.

I think sending a PR is meaningful at this point because it reflects the right structure. The remainder of the work involves writing new analysis kinds, improving existing analysis kinds, and writing new validations.

------

There are lots more investments that are *not* in this PR. We need to produce much more structure data in a bunch of areas, and we need to think about how this will be data-driven by the runtime.


eng/Versions.props
```diff
@@ -148,9 +148,9 @@
     <MicrosoftAzureKeyVaultPackageVersion>2.3.2</MicrosoftAzureKeyVaultPackageVersion>
     <MicrosoftBuildFrameworkPackageVersion>15.8.166</MicrosoftBuildFrameworkPackageVersion>
     <MicrosoftBuildUtilitiesCorePackageVersion>15.8.166</MicrosoftBuildUtilitiesCorePackageVersion>
-    <MicrosoftCodeAnalysisCommonPackageVersion>2.8.0</MicrosoftCodeAnalysisCommonPackageVersion>
-    <MicrosoftCodeAnalysisCSharpPackageVersion>2.8.0</MicrosoftCodeAnalysisCSharpPackageVersion>
-    <MicrosoftCodeAnalysisCSharpWorkspacesPackageVersion>2.8.0</MicrosoftCodeAnalysisCSharpWorkspacesPackageVersion>
+    <MicrosoftCodeAnalysisCommonPackageVersion>3.0.0</MicrosoftCodeAnalysisCommonPackageVersion>
+    <MicrosoftCodeAnalysisCSharpPackageVersion>3.0.0</MicrosoftCodeAnalysisCSharpPackageVersion>
+    <MicrosoftCodeAnalysisCSharpWorkspacesPackageVersion>3.0.0</MicrosoftCodeAnalysisCSharpWorkspacesPackageVersion>
```
rynowak: We need 3.0 for some of the improved APIs like `RegisterSymbolStartAction`/`RegisterSymbolEndAction`. It's pretty strongly recommended that we use these new APIs, and there's no where I can think of that we need to run and doesn't support 3.0

src/Components/Analyzers/src/Microsoft.AspNetCore.Components.Analyzers.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard1.3</TargetFramework>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
rynowak: Required to target 3.0. Everywhere we need to run supports it.

src/Mvc/Mvc.Analyzers/src/Startup/MiddlewareAnalyzer.cs
```diff
@@ -0,0 +1,48 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Operations;
+
+namespace Microsoft.AspNetCore.Analyzers
+{
+    internal class MiddlewareAnalyzer
```
rynowak: It's a common pattern writing stateful analyzers to call things `FooAnalyzer` even though they aren't registered as top-level analyzers.

src/Mvc/Mvc.Analyzers/src/Startup/MiddlewareAnalyzer.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        public void AnalyzeConfigureMethod(OperationBlockStartAnalysisContext context)
+        {
+            var configureMethod = (IMethodSymbol)context.OwningSymbol;
+            var middleware = ImmutableArray.CreateBuilder<MiddlewareItem>();
+
+            // Note: this is a simple source-order implementation. We don't attempt perform data flow
+            // analysis in order to determine the actual order in which middleware are ordered.
+            //
+            // This can currently be confused by things like Map(...)
```
rynowak: THIS IS SUPER HACK MODE 😆

rynowak: There's an existing package from Roslyn that we can use to do more sophicsticated analysis, but it's not needed in this PR.

src/Mvc/Mvc.Analyzers/src/Startup/MiddlewareOrderingAnalzyer.cs
```diff
@@ -0,0 +1,61 @@
...
+
+namespace Microsoft.AspNetCore.Analyzers
+{
+    internal class MiddlewareOrderingAnalyzer
+    {
+        // This should probably be a multi-map, but oh-well.
+        private readonly static ImmutableDictionary<string, string> MiddlewareHappensAfterMap = ImmutableDictionary.CreateRange<string, string>(new[]
+        {
+            new KeyValuePair<string, string>("UseAuthorization", "UseAuthentication"),
+        });
```
rynowak: This is the kind of info we want to get from the runtime so the analyzer can support all versions. For now I'm just pointing that out.

src/Mvc/Mvc.Analyzers/src/Startup/MiddlewareRequiredServiceAnalzyer.cs
```diff
@@ -0,0 +1,89 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Collections.Immutable;
```
rynowak: > When in Rome...

src/Mvc/Mvc.Analyzers/src/Startup/StartupAnalzyer.cs
```diff
@@ -12,44 +11,16 @@ namespace Microsoft.AspNetCore.Analyzers
...
 
-        private readonly static Func<CompilationAnalysisContext, ConcurrentBag<StartupComputedAnalysis>, StartupDiagnosticValidator>[] DiagnosticValidatorFactories = new Func<CompilationAnalysisContext, ConcurrentBag<StartupComputedAnalysis>, StartupDiagnosticValidator>[]
-        {
-            UseMvcDiagnosticValidator.CreateAndInitialize,
-        };
-
-#pragma warning restore RS1008 // Avoid storing per-compilation data into the fields of a diagnostic analyzer.
-
-        public StartupAnalzyer()
+        public override void Initialize(AnalysisContext context)
```
rynowak: The diff is garbo here. I recommend looking at this method in detail because it's elegant and cool..

src/Mvc/Mvc.Analyzers/src/Startup/StartupFacts.cs
```diff
@@ -21,7 +49,7 @@ public static bool IsConfigureServices(StartupSymbols symbols, IMethodSymbol sym
                 return false;
             }
 
-            if (!string.Equals(symbol.Name, SymbolNames.ConfigureServicesMethod, StringComparison.Ordinal))
+            if (!string.Equals(symbol.Name, "ConfigureServices", StringComparison.Ordinal))
```
rynowak: LOL YOLO

rynowak:
Oh yeah, and this needs to move outside of the 'MVC' analyzers at some point too.

anurse:
@rynowak I'm doing a little PR housekeeping. Is this still a thing?

rynowak:
Yeah please don't hurt me. I'll get back to this pretty soon.

anurse:
Nah it's cool, just checking in :).

rynowak:
Ok good. I'll make sure to #YOLOMerge it in a few weeks and break the build.... (just kidding)

anurse:
Genie from Disney's Aladdin saying "Well I feel sheepish"

rynowak:
@pranavkm - can you look at this please 👍

src/Mvc/Mvc.Analyzers/src/Startup/StartupAnalysisContext.cs
```diff
@@ -1,22 +1,98 @@
...
+        }
 
         public StartupSymbols StartupSymbols { get; }
-#pragma warning restore RS1012 // Start action has no registered actions.
+
+        public T? GetRelatedSingletonAnalysis<T>(INamedTypeSymbol type) where T : class
+        {
+            lock (_lock)
+            {
+                if (_analysesByType.TryGetValue(type, out var list))
```
pranavkm: Would it benefit if the value was an `ImmutableList` \ `ImmutableArray` and then do the iteration outside the `lock` or just use a `ConcurrentDictionary` at that point?

src/Mvc/Mvc.Analyzers/src/Startup/StartupAnalyzer.Diagnostics.cs
```diff
@@ -1,30 +1,47 @@
...
+                    // ASP
+                    BuildServiceProviderShouldNotCalledInConfigureServicesMethod,
+
+                    // MVC
+                    UnsupportedUseMvcWithEndpointRouting,
+                });
+            }
+
+            internal readonly static DiagnosticDescriptor BuildServiceProviderShouldNotCalledInConfigureServicesMethod = new DiagnosticDescriptor(
+                "ASP0000",
```
pranavkm: `ASP1000`? 0000 feels like somebody forgot to initialize the string correctly

src/Mvc/Mvc.Analyzers/src/Startup/StartupAnalyzer.Diagnostics.cs
```diff
@@ -1,30 +1,47 @@
...
+                "Do not call 'IServiceCollection.BuildServiceProvider' in 'ConfigureServices'",
+                "Calling 'BuildServiceProvider' from application code results in an additional copy of singleton services being created. Consider alternatives such as dependency injecting services as parameters to 'Configure'.",
+                "Design",
+                DiagnosticSeverity.Warning,
+                isEnabledByDefault: true,
+                helpLinkUri: "https://aka.ms/AA5k895");
```
pranavkm: Can you file an issue to point this to the correct doc article?

src/Mvc/Mvc.Analyzers/test/Mvc.Analyzers.Test.csproj
```diff
@@ -20,6 +20,10 @@
     <Content Include="TestFiles\**\*.*" CopyToPublishDirectory="PreserveNewest" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Content Remove="TestFiles\StartupAnalyzerTest\MvcOptions_UseMvcAndEndpointRoutingEnabled.cs" />
```
pranavkm: Remove?

pranavkm (approved):

src/Mvc/Mvc.Analyzers/src/Startup/StartupAnalysisContext.cs
rynowak: I have no reason to believe the lock is causing a perf problem. Ultimately it makes sense to split this context class into a builder + results for design clarity though. I'll try it out.

src/Mvc/Mvc.Analyzers/src/Startup/StartupAnalyzer.Diagnostics.cs
rynowak: Seems perfect to me.

src/Mvc/Mvc.Analyzers/src/Startup/StartupAnalyzer.Diagnostics.cs
rynowak: Sure.

PR #8330, opened 2019-03-08
Skip IIS Express on Windows 7

pakrym:
https://github.com/aspnet/AspNetCore/issues/8329

PR #9490, opened 2019-04-18
Revert "Skip IIS Express on Windows 7 (#8330)"

jkotalik:
This reverts commit b2a58ab8ded6bc2ab55edb835b00d8b601921c4f.

We disabled a large number of OutOfProc tests due to a change in coreclr. This reverts it as it has been fixed.

### DIFF
--- a/src/Servers/IIS/IIS/test/Common.FunctionalTests/OutOfProcess/AspNetCorePortTests.cs
+++ b/src/Servers/IIS/IIS/test/Common.FunctionalTests/OutOfProcess/AspNetCorePortTests.cs
@@ -37,7 +37,7 @@ namespace Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests
                from s in new string[] { (_minPort - 1).ToString(), (_maxPort + 1).ToString(), "noninteger" }
                select new object[] { v, s };
 
-        [ConditionalTheory(Skip = "https://github.com/aspnet/AspNetCore/issues/8329")]
+        [ConditionalTheory]
         [MemberData(nameof(TestVariants))]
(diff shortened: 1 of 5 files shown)


src/Servers/IIS/IIS/test/Common.FunctionalTests/OutOfProcess/AspNetCorePortTests.cs
```diff
@@ -37,7 +37,7 @@ public static IEnumerable<object[]> InvalidTestVariants
                from s in new string[] { (_minPort - 1).ToString(), (_maxPort + 1).ToString(), "noninteger" }
                select new object[] { v, s };
 
-        [ConditionalTheory(Skip = "https://github.com/aspnet/AspNetCore/issues/8329")]
+        [ConditionalTheory]
```
dougbu: Suggest switching the tests w/ no remaining conditions (e.g. no `[RequiresIIS]`) to use normal `[Fact]` and `[Theory]`

dougbu (approved):

src/Servers/IIS/IIS/test/Common.FunctionalTests/OutOfProcess/AspNetCorePortTests.cs
jkotalik: There are assembly wide skip conditions for requiring iis: https://github.com/aspnet/AspNetCore/blob/3477daf3c4f530dff80f197e0642cb39a26fb07b/src/Servers/IIS/IIS/test/IIS.Shared.FunctionalTests/Properties/AssemblyInfo.cs#L9. So these still need to be conditional facts.

Issue #9208, opened 2019-04-09
Reconsider AddAuthorizationPolicyEvaluator required for the authZ middleware


pranavkm:
Setting up the Auth middleware requires calling `services.AddAuthorizationPolicyEvaluator`. The name isn't particularly discoverable. We should instead see if we could repurpose `services.AddAuthorization`

Eilon:
We should look at merging the AuthZ.Core and AuthZ.Policy into one assembly and have `.AddAuthorization()` do all the work.

Eilon:
@JamesNK do you have cycles for this in preview5?

JamesNK:
If my gRPC tasks go well, I think so.

Double checking we want this:
Core depends on logging and options. Policy depends on authentication, HTTP and routing.

Are we ok with authz assembly having hard dependencies on authentication, HTTP and routing?

Eilon:
@rynowak / @davidfowl - any thoughts on the dependency changes?

rynowak:
hmm. I thought I had commented on this. I don't think there's a reason not to do this - it's better for users. 

There's little meaning to our assembly factoring anymore as long as it doesn't make building the product insane.

davidfowl:
> Are we ok with authz assembly having hard dependencies on authentication, HTTP and routing?

Yes this makes sense to me. I'm not aware of anyone that was trying to use auth outside of http scenarios.

> There's little meaning to our assembly factoring anymore as long as it doesn't make building the product insane.

Not true but I don't want to discuss that on github 😄. Remember there's still a thing called self contained applications. If no layers exist, trimming is *extremely* hard.

PR #9492, opened 2019-04-18
Move Microsoft.AspNetCore.Authorization.Policy types

JamesNK:
Addresses #9208


HaoK (approved):
Should we also try to get rid of the .Policy namespace as part of this move?  I can't imagine there are situations where you can only use any of the Policy classes without also using anything in the base Authorization namespace

JamesNK:
We could, but changing the namespace would be a breaking change.

Eilon (approved):
Change looks good. I agree we should avoid unnecessary breaking changes here.

PR #9506, opened 2019-04-18
Lazy create EndpointRouting statemachines

benaadams:
* Lazy create EndpointRouting statemachines
* Lazy create EndpointMiddleware statemachine

image

Normally complete sync; so skip the async machines in the common path

/cc @rynowak @pranavkm

### DIFF
--- a/src/Http/Routing/src/EndpointMiddleware.cs
+++ b/src/Http/Routing/src/EndpointMiddleware.cs
@@ -31,7 +31,7 @@ namespace Microsoft.AspNetCore.Routing
             _routeOptions = routeOptions?.Value ?? throw new ArgumentNullException(nameof(routeOptions));
         }
 
-        public async Task Invoke(HttpContext httpContext)
+        public Task Invoke(HttpContext httpContext)
         {
             var endpoint = httpContext.Features.Get<IEndpointFeature>()?.Endpoint;
             if (endpoint?.RequestDelegate != null)
@@ -39,7 +39,7 @@ namespace Microsoft.AspNetCore.Routing
                 if (_routeOptions.SuppressCheckForUnhandledSecurityMetadata)
                 {
                     // User opted out of this check.
-                    return;
+                    return Task.CompletedTask;
                 }
 
                 if (endpoint.Metadata.GetMetadata<IAuthorizeData>() != null &&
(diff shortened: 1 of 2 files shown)


src/Http/Routing/src/EndpointMiddleware.cs
```diff
@@ -58,17 +58,34 @@ public async Task Invoke(HttpContext httpContext)
...
+                    var requestTask = endpoint.RequestDelegate(httpContext);
+                    if (!requestTask.IsCompletedSuccessfully)
+                    {
+                        return AwaitRequestTask(endpoint, requestTask, _logger);
+                    }
                 }
-                finally
+                catch (Exception exception)
                 {
                     Log.ExecutedEndpoint(_logger, endpoint);
```
campersau: What if the `requestTask` was completed synchronously, then this wouldn't get logged right?

pranavkm:
For my own benefit, how much of an impact does this have? I was planning on looking at the AuthZ middleware at some point and it would be nice to know this makes a reasonable difference before I dive in

src/Http/Routing/src/EndpointMiddleware.cs
benaadams: Good catch

benaadams:
> For my own benefit, how much of an impact does this have on throughput?

Depends how many; ResourceInvoker was quite good (+3%) https://github.com/aspnet/AspNetCore/pull/9311 but that removed 6; so maybe +0.5% per item on a `dotnet new mvc` + the [`PlainTextActionResult`](https://github.com/aspnet/Benchmarks/blob/6a91ebd2dc574e0ed1f13a4a76930b51944825b3/src/Benchmarks/Controllers/HomeController.cs#L42-L55)

Would be less if the controller was doing more; so it had less impact on Json for example

src/Http/Routing/src/EndpointMiddleware.cs
benaadams: Fixed

src/Http/Routing/src/EndpointRoutingMiddleware.cs
```diff
@@ -53,15 +53,24 @@ internal sealed class EndpointRoutingMiddleware
...
+            }
+
+            var matchTask = matcherTask.Result.MatchAsync(httpContext, feature);
+            if (!matchTask.IsCompletedSuccessfully)
+            {
+                return AwaitMatch(this, httpContext, feature, matchTask);
+            }
 
-            await matcher.MatchAsync(httpContext, feature);
             if (feature.Endpoint != null)
```
Tratcher: Can you call CompleteAwaited here to avoid the code duplication? (And give ti a better name?)

benaadams: Sure

rynowak (approved):

PR #9507, opened 2019-04-18
Lazy create StaticFileMiddleware statemachine

benaadams:
* Lazy create StaticFileMiddleware statemachine

@rynowak @JamesNK @Tratcher question...

`dotnet new mvc` etc adds `app.UseStaticFiles();` prior to `app.UseRouting()` and `app.UseEndpoints()`  (also before auth) is this desirable?

https://github.com/aspnet/AspNetCore/blob/8b536b08ca76126e58f662b7f500bb191116f2bd/src/ProjectTemplates/Web.ProjectTemplates/content/StarterWeb-CSharp/Startup.cs#L157-L174

e.g. it outputs
```csharp
app.UseStaticFiles();

app.UseCookiePolicy();

app.UseRouting();

app.UseAuthorization();

app.UseEndpoints(endpoints =>
{
    endpoints.MapControllerRoute(
        name: "default",
        pattern: "{controller=Home}/{action=Index}/{id?}");
    endpoints.MapRazorPages();
});
```
For the call chain

image

### DIFF
--- a/src/Middleware/StaticFiles/src/StaticFileMiddleware.cs
+++ b/src/Middleware/StaticFiles/src/StaticFileMiddleware.cs
@@ -130,7 +130,8 @@ namespace Microsoft.AspNetCore.StaticFiles
                     {
                         context.Response.Clear();
                     }
-                    break;
+                    await _next(context);
+                    return;
                 case StaticFileContext.PreconditionState.NotModified:
                     _logger.FileNotModified(fileContext.SubPath);
                     await fileContext.SendStatusAsync(Constants.Status304NotModified);
--- a/src/Middleware/StaticFiles/src/StaticFileMiddleware.cs
+++ b/src/Middleware/StaticFiles/src/StaticFileMiddleware.cs
@@ -68,7 +68,7 @@ namespace Microsoft.AspNetCore.StaticFiles
         /// </summary>
         /// <param name="context"></param>
         /// <returns></returns>
-        public async Task Invoke(HttpContext context)
+        public Task Invoke(HttpContext context)
         {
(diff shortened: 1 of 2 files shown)


rynowak:
> dotnet new mvc etc adds app.UseStaticFiles(); prior to app.UseRouting() and app.UseEndpoints() (also before auth) is this desirable?

Yes. Static files short circuits when it matches and doesn't interact with any of these things. So it makes sense to put before.

rynowak (approved):
From me, but @Tratcher should look as well.

benaadams:
@Tratcher the ignore whitespace diff is clearest to look at

Tratcher (approved):

src/Middleware/StaticFiles/src/StaticFileMiddleware.cs
```diff
@@ -95,52 +95,57 @@ public async Task Invoke(HttpContext context)
...
-                        await fileContext.SendStatusAsync(Constants.Status412PreconditionFailed);
+                        await fileContext.SendAsync();
+                        _logger.FileServed(fileContext.SubPath, fileContext.PhysicalPath);
                         return;
+                    }
+                    catch (FileNotFoundException)
+                    {
+                        context.Response.Clear();
+                    }
+                    break;
```
Tratcher: This case used to fall though to next but now returns OK. It's causing test failures.

Tratcher (changes requested):

src/Middleware/StaticFiles/src/StaticFileMiddleware.cs
benaadams: Ah, good spot, added

Tratcher (approved):

PR #9515, opened 2019-04-18
Stop running InstallVisualStudio.ps1 during CI builds

natemcmaster:
Resolves https://github.com/aspnet/AspNetCore-Internal/issues/2194

VS2019.BT queues should have VS 2019 RTM now, so this step _should_ be unnecessary.

### DIFF
--- a/.azure/pipelines/jobs/default-build.yml
+++ b/.azure/pipelines/jobs/default-build.yml
@@ -120,9 +120,6 @@ jobs:
   steps:
   - checkout: self
     clean: true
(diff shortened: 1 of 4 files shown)


dougbu (approved):
It worked ❕

Can the LocalDB installation step also be removed from /.azure/pipelines/ci.yml ?

natemcmaster:
Not yet. We would need to have SQL server preinstalled or switch agent queues. I'm working with the dnceng folks on that...hitting some pushback, so we'll keep installing on-demand.

PR #9517, opened 2019-04-18
[Identity] Make store Contexts virtual

### LINKED ISSUE #5806: Make Context virtual in UserStore
Could the [Context property](https://github.com/aspnet/Identity/blob/9ecbefcf21d12d90ac187e1eefe02ed0219edc23/src/EF/UserStore.cs#L126) be marked as virtual in UserStore? I would like to implement my own getter so I can better manage the lifecycle of the DbContext (to enforce better control over the transaction boundaries).
I have implemented an unit of work using the [DbContextScope library](https://github.com/apawsey/DbContextScope) and it's working great throughout my app, so I would like to use it for Identity as well.

HaoK:
Fixes https://github.com/aspnet/AspNetCore/issues/5806


src/Identity/Core/src/CoreIdentityServiceCollectionExtensions.cs
```diff
@@ -12,7 +12,7 @@ namespace Microsoft.Extensions.DependencyInjection
     /// <summary>
     /// Contains extension methods to <see cref="IServiceCollection"/> for configuring identity services.
     /// </summary>
-    public static class IdentityServiceCollectionExtensions
+    public static class CoreIdentityServiceCollectionExtensions
```
ajcvickers: Why rename? (This is a breaking change...)

ajcvickers: Looks okay otherwise.

ajcvickers: Never mind. Looked at the issue and realized it is an intentional break.

ajcvickers (approved):

src/Identity/Core/src/CoreIdentityServiceCollectionExtensions.cs
HaoK: Turns out this causes issues with templating tests which must not be rebuilding this, gonna back out this change for now

PR #9521, opened 2019-04-18
Avoid async in AuthorizationMiddleware when no metadata

JamesNK:
Minor performance optimization. Avoid the overhead of an async state machine when it is not needed.

### DIFF
--- a/src/Security/Authorization/Policy/src/AuthorizationMiddleware.cs
+++ b/src/Security/Authorization/Policy/src/AuthorizationMiddleware.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Authentication;
@@ -37,7 +38,7 @@ namespace Microsoft.AspNetCore.Authorization
             _policyProvider = policyProvider;
         }
 
-        public async Task Invoke(HttpContext context)
+        public Task Invoke(HttpContext context)
         {
             if (context == null)
             {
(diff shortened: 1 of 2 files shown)


Tratcher (approved):

src/Security/Authorization/Policy/src/AuthorizationMiddleware.cs
```diff
@@ -53,10 +54,14 @@ public async Task Invoke(HttpContext context)
             var authorizeData = endpoint?.Metadata.GetOrderedMetadata<IAuthorizeData>() ?? Array.Empty<IAuthorizeData>();
             if (authorizeData.Count() == 0)
```
benaadams: Drop the `?? Array.Empty<IAuthorizeData>()`

And test for `if (authorizeData == null || authorizeData.Count() == 0)`

Rather than invoking Linq `.Count()` on an empty array?

benaadams:
You'll also have to remove `[System.Diagnostics.DebuggerStepThroughAttribute]` from the ref assembly for this method for code check to pass

src/Security/Authorization/Policy/src/AuthorizationMiddleware.cs
```diff
@@ -49,14 +50,18 @@ public async Task Invoke(HttpContext context)
...
+                return _next(context);
             }
-            
+
+            return EvaluatePolicy(context, endpoint, authorizeData);
+        }
+
+        private async Task EvaluatePolicy(HttpContext context, Endpoint endpoint, IEnumerable<IAuthorizeData> authorizeData)
+        {
+            // IMPORTANT: Changes to authorization logic should be mirrored in MVC's AuthorizeFilter
```
pranavkm: This is pretty much no longer true.

pranavkm (approved):
👍

benaadams (approved):

PR #9531, opened 2019-04-18
Set camelCase as default in JsonHubProtocol

BrennanConroy:
The testing is a little wonky because of some things still not supported by System.Text.Json yet.


src/SignalR/common/SignalR.Common/test/Internal/Protocol/JsonHubProtocolTests.cs
```diff
@@ -102,8 +102,66 @@ public void MagicCast()
             new JsonProtocolTestData("CompletionMessage_HasFloatResult", CompletionMessage.WithResult("123", 2.0f), true, true, "{\"type\":3,\"invocationId\":\"123\",\"result\":2}"),
             new JsonProtocolTestData("StreamInvocationMessage_HasFloatArgument", new StreamInvocationMessage("123", "Target", new object[] { 1, "Foo", 2.0f }), true, true, "{\"type\":4,\"invocationId\":\"123\",\"target\":\"Target\",\"arguments\":[1,\"Foo\",2]}"),
             new JsonProtocolTestData("InvocationMessage_StringIsoDateArgument", new InvocationMessage("Method", new object[] { "2016-05-10T13:51:20+12:34" }), true, true, "{\"type\":1,\"target\":\"Method\",\"arguments\":[\"2016-05-10T13:51:20\\u002b12:34\"]}"),
+
```
mikaelm12: nit: remove new line

src/SignalR/common/Protocols.Json/src/JsonHubProtocolOptions.cs
```diff
@@ -21,5 +21,20 @@ public JsonHubProtocolOptions()
         public bool IgnoreNullValues { get => _serializerOptions.IgnoreNullValues; set => _serializerOptions.IgnoreNullValues = value; }
```
mikaelm12: nit: Probably not necessary now but needs xml docs and newlines.

src/SignalR/common/Protocols.Json/src/JsonHubProtocolOptions.cs
```diff
@@ -21,5 +21,20 @@ public JsonHubProtocolOptions()
...
         public bool WriteIndented { get => _serializerOptions.WriteIndented; set => _serializerOptions.WriteIndented = value; }
         public bool AllowTrailingCommas { get => _serializerOptions.AllowTrailingCommas; set => _serializerOptions.AllowTrailingCommas = value; }
+        public bool UseCamelCase
```
mikaelm12: Do we not want to expose our own enum rather than mapping a bool to an enum? I guess this can just be part of the api review.

mikaelm12: New public API - Run the tool

src/SignalR/common/Protocols.Json/src/JsonHubProtocolOptions.cs
```diff
@@ -21,5 +21,20 @@ public JsonHubProtocolOptions()
...
+        {
+            get => _serializerOptions.PropertyNamingPolicy == JsonNamingPolicy.CamelCase;
+            set
+            {
+                if (value == true)
```
mikaelm12:
```suggestion
                if (value)
```

mikaelm12 (approved):
Be sure to track the testing change you want to make

src/SignalR/common/Protocols.Json/src/JsonHubProtocolOptions.cs
```diff
@@ -21,5 +21,20 @@ public JsonHubProtocolOptions()
...
         public bool WriteIndented { get => _serializerOptions.WriteIndented; set => _serializerOptions.WriteIndented = value; }
         public bool AllowTrailingCommas { get => _serializerOptions.AllowTrailingCommas; set => _serializerOptions.AllowTrailingCommas = value; }
+        public bool UseCamelCase
+        {
+            get => _serializerOptions.PropertyNamingPolicy == JsonNamingPolicy.CamelCase;
+            set
+            {
+                if (value)
+                {
+                    _serializerOptions.PropertyNamingPolicy = JsonNamingPolicy.CamelCase;
```
mikaelm12: If you want to be cool `_serializerOptions.PropertyNamingPolicy = value ? JsonNamingPolicy.CamelCase : null`

src/SignalR/common/Protocols.Json/src/JsonHubProtocolOptions.cs
JamesNK: `JsonNamingPolicy` is an abstract base class. `JsonNamingPolicy.CamelCase` is a static instance of `CamelCaseNamingPolicy`.

The question is whether it makes sense for people to configure their own naming policy. If so then expose PropertyNamingPolicy. If someone wants camelcase then they can set `JsonNamingPolicy.CamelCase`. If they want something else then they can assign their own implementation of JsonNamingPolicy.

JamesNK: Get consensus here before merging please :)

BrennanConroy: We're changing this in preview 6 and have filed https://github.com/aspnet/AspNetCore/issues/9519 to review the API

anurse: There are preview 6 changes planned for JSON that will likely make this entirely moot. We may be able to just expose their options directly in that. We can't use their types right now because of the sources package.

anurse: Followed up with @JamesNK offline. I'm OK with this as the preview 5 version of the API given the review in preview 6.

Issue #8710, opened 2019-03-21
Support automatic reconnect in the .NET client


anurse:
Epic https://github.com/aspnet/AspNetCore/issues/8797

Split from https://github.com/aspnet/AspNetCore/issues/5282

We will support automatic reconnect in 3.0 **without message replay**. We will also allow customization of the reconnect logic. This will be opt-in via a `HubConnectionBuilder` option. The default behavior will continue to be manual reconnection.

anurse:
Pre-triage notes:
* Seems reasonable for preview 5 given that TS already drove the "design".
* Need to determine if we can close on this quickly enough

anurse:
Triage Decision: We expect to have this done (in an initial form at least) for preview 5.

mikaelm12:
*Acceptance checklist** (check one item)
* [ ] We decided not to take this fix.
* [ ] The fix is tests-only.
* [X] The fix contains product changes (**check all items below**).
	* [X] Relevant XML documentation comments for new public APIs are present.
	* [ ] Narrative docs (docs.microsoft.com) are updated. (**check one item below**)
    	* [X] The change requires a new article/updated article. An issue with an outline has been filed here: [aspnet/AspNetCore.Docs/issues/12586]
		* [ ] The change requires a change to an existing article. A **closed** docs PR with these changes is linked here: [PR LINK]
		* [ ] The change requires no docs changes.
	* [X] Verification has been completed. (**check one item below**)
    	* [X] The change is in the shared framework and was verified against the following versions 
			* SDK installer: [3.0.100-preview6-012105]
			* ASP.NET Core Runtime: [3.0.0-preview6-19278-05]
			* .NET Core Runtime: [3.0.0-preview6-27728-04]
	    * [ ] The change is in an OOB NuGet/NPM/JAR package and was verified against the following version of that package: [PACKAGE ID] [VERSION]

PR #9535, opened 2019-04-19
Auto-reconnect support for .NET SignalR client

halter73:
Addresses #8710


src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -1204,19 +1240,214 @@ private async Task PingServer()
...
+
+                await Task.Delay(nextRetryDelay.Value);
+
+                if (State != HubConnectionState.Reconnecting) {
+                    //_logger.log(LogLevel.Debug, "Connection left the reconnecting state during reconnect delay. Done reconnecting.");
+                    return;
+                }
+
+                try {
+                    // TODO: Use a token that is canceled in StopAsync.
```
halter73: I still need to do work to ensure stop always works while the HubConnection is reconnecting. And I also need to add tests of course.

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -1204,19 +1240,214 @@ private async Task PingServer()
...
+                {
+                    await reconnecting.Invoke(closeException);
+                }
+                catch (Exception ex)
+                {
+                    Log.ErrorDuringReconnectingEvent(_logger, ex);
+                }
+            }
+
+            // There is no need to start a new task if there is no Closed event registered
```
steji113:
```suggestion
            // There is no need to start a new task if there is no Reconnecting event registered
```

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -1204,19 +1240,214 @@ private async Task PingServer()
...
+                {
+                    await reconnected.Invoke(ConnectionId);
+                }
+                catch (Exception ex)
+                {
+                    Log.ErrorDuringReconnectedEvent(_logger, ex);
+                }
+            }
+
+            // There is no need to start a new task if there is no Closed event registered
```
steji113:
```suggestion
            // There is no need to start a new task if there is no Reconnected event registered
```

src/SignalR/clients/csharp/Client.Core/src/Internal/DefaultRetryPolicy.cs
```diff
@@ -0,0 +1,41 @@
...
+
+            for (int i = 0; i < retryDelays.Length; i++)
+            {
+                _retryDelays[i] = retryDelays[i];
+            }
+        }
+
+        public TimeSpan? NextRetryDelay(RetryContext retryContext)
+        {
+            return DEFAULT_RETRY_DELAYS_IN_MILLISECONDS[retryContext.PreviousRetryCount];
```
steji113: Should this be indexing into `_retryDelays`?

src/SignalR/clients/csharp/Client.Core/src/HubConnectionState.cs
```diff
@@ -13,8 +13,16 @@ public enum HubConnectionState
...
         Disconnected,
         /// <summary>
+        /// The hub connection is connecting.
+        /// </summary>
+        Connecting,
+        /// <summary>
         /// The hub connection is connected.
         /// </summary>
-        Connected
+        Connected,
```
steji113: Is this considered a [breaking change](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/breaking-change-rules.md#property-field-parameter-and-return-values) since it is changing the value of the `Connected` enum?

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -1204,19 +1240,214 @@ private async Task PingServer()
...
+                ReleaseConnectionLock();
+            }
 
+            // Dispose the connection
+            await CloseAsync(connectionState.Connection);
+
+            // Cancel any outstanding invocations within the connection lock
+            connectionState.CancelOutstandingInvocations(connectionState.CloseException);
+
+            if (connectionState.Stopping || _reconnectPolicy != null)
```
steji113: Should this be `_reconnectPolicy == null`?

steji113: Should this be `_reconnectPolicy == null`?

src/SignalR/clients/csharp/Client.Core/src/HubConnectionState.cs
mikaelm12: This isn't actually changing the value of this enum. It only looks like a diff because we don't use [dangling commas](https://medium.com/@nikgraf/why-you-should-enforce-dangling-commas-for-multiline-statements-d034c98e36f8)

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -1348,15 +1579,17 @@ public Task InvokeAsync(object[] parameters)
...
+        // TODO: Refactor all transient state about the connection into the ConnectionState class.
         private class ConnectionState : IInvocationBinder
         {
-            private volatile bool _stopping;
             private readonly HubConnection _hubConnection;
 
-            private TaskCompletionSource<object> _stopTcs;
             private readonly object _lock = new object();
             private readonly Dictionary<string, InvocationRequest> _pendingCalls = new Dictionary<string, InvocationRequest>(StringComparer.Ordinal);
+            private TaskCompletionSource<object> _stopTcs;
```
mikaelm12: nit: The spacing on all these members is weird

src/SignalR/clients/csharp/Client.Core/src/HubConnectionBuilderExtensions.cs
```diff
@@ -23,5 +24,49 @@ public static IHubConnectionBuilder ConfigureLogging(this IHubConnectionBuilder
...
+
+        /// <summary>
+        /// Configures the <see cref="HubConnection"/> to automatically attempt to reconnect if the connection is lost.
+        /// The client will wait the default 0, 2, 10 and 30 seconds respectively before trying up to four reconnect attempts.
+        /// </summary>
+        /// <param name="hubConnectionBuilder">The <see cref="IHubConnectionBuilder" /> to configure.</param>
+        /// <returns>The same instance of the <see cref="IHubConnectionBuilder"/> for chaining.</returns>
+        public static IHubConnectionBuilder WithAutomaticReconnect(this IHubConnectionBuilder hubConnectionBuilder)
+        {
+            // REVIEW: Should we throw if this would override another IRetryPolicy?
```
mikaelm12: > Should we throw if this would override another IRetryPolicy?

I think this sounds reasonable

src/SignalR/clients/csharp/Client.Core/src/IRetryPolicy.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.SignalR.Client
+{
+    public interface IRetryPolicy
```
mikaelm12: You actually call this `IReconnectPolicy` in the TypeScript

https://github.com/aspnet/AspNetCore/blob/a6fd8da9e87d29515db3196cae7b70ea40f90381/src/SignalR/clients/ts/signalr/src/IReconnectPolicy.ts#L5

mikaelm12: I still think a build would be nice for configuration here but it can wait 
https://github.com/aspnet/AspNetCore/issues/9016

src/SignalR/clients/csharp/Client.Core/src/HubConnectionState.cs
steji113: Sorry the comment is on this line, but the change is inserting the `Connecting` enum above `Connected` which makes `Connecting` = 1 and `Connected` = 2 when it was 1 previously.

src/SignalR/clients/csharp/Client.Core/src/Internal/DefaultRetryPolicy.cs
```diff
@@ -0,0 +1,41 @@
...
+        private TimeSpan?[] _retryDelays;
+
+        public DefaultRetryPolicy()
+        {
+            _retryDelays = DEFAULT_RETRY_DELAYS_IN_MILLISECONDS;
+        }
+
+        public DefaultRetryPolicy(TimeSpan[] retryDelays)
+        {
+            _retryDelays = new TimeSpan?[retryDelays.Length + 1];
```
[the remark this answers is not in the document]
steji113: I think this was done because a null needs to be added at the end hence the `retryDelays.Length + 1`. The null is what triggers the end of the reconnection attempts.

src/SignalR/clients/csharp/Client.Core/src/HubConnectionState.cs
mikaelm12: Good point! I think you're right

halter73: It's definitely a breaking change. The question is, do we care more about breaking someone (if anyone) who might be treating HubConnection.State as an int? Or do we care more about having the enum values listed in a logical order?

I could be convinced either way.

src/SignalR/clients/csharp/Client.Core/src/Internal/DefaultRetryPolicy.cs
halter73: Yes.

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
halter73: Yep. As you can tell, this PR is still in the very early stages.

steji113: Sorry, thanks for humoring my review, I have been a big SignalR fan and have been eagerly watching for this feature throughout the clients. It has also been a great learning experience seeing how it is implemented under the covers. Didn't mean to rush you! Thank you guys for all the hard work.

src/SignalR/clients/csharp/Client.Core/src/IRetryPolicy.cs
halter73: I was going to bring up the idea of renaming this in the TS client too for Preview6.

The reason being is we might want to use the same interface to configure retry behavior after a failed start. Currently, we never retry on initial start failures no matter the reconnect policy, and we don't have any immediate plans to start doing so. But if the future we want to add a retry-start policy, it would be nice if the interface it used was called IRetryPolicy instead of IReconnectPolicy.

Thoughts?

halter73: If we start adding more options like retrying on initial start failures, this starts to become even more appealing.

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
halter73: I was trying to make the spacing more sane by keeping the readonly fields together and keeping the fields protected by the lock near the lock while also keeping the shutdown-related state fields near each other. I didn't want to add a comment because _stopTcs isn't completely protect by the _lock, but I do think the spacing is more logical than before.

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
halter73: Thanks for the review. You've raised a lot of good points. I just wanted to make sure you knew there's a lot left for me to do.

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -96,13 +98,49 @@ public partial class HubConnection
...
+        /// <remarks>
+        /// The <see cref="string"/> parameter will be the <see cref="HubConnection"/>'s new ConnectionId or null if negotiation was skipped.
+        /// </remarks>
+        /// <example>
+        /// The following example attaches a handler to the <see cref="Reconnected"/> event, and checks the provided argument to log the ConnectionId.
+        ///
+        /// <code>
+        /// connection.Reconnected += (connectionId) =>
+        /// {
+        ///     Console.WriteLine($"Connection completed reconnecting with ConnectionId: {connectionId}");
```
```suggestion
        ///     Console.WriteLine($"Connection successfully reconnected. The ConnectionId is now: {connectionId}");
```

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -193,10 +239,21 @@ public HubConnection(IConnectionFactory connectionFactory, IHubProtocol protocol
         public async Task StartAsync(CancellationToken cancellationToken = default)
         {
             CheckDisposed();
+
+            if (!TryChangeState(HubConnectionState.Disconnected, HubConnectionState.Connecting))
+            {
+                throw new InvalidOperationException($"The ${nameof(HubConnection)} cannot be started if it is not in the disconnected state.");
```
BrennanConroy: Whats with the $ in the string?

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -193,10 +239,21 @@ public HubConnection(IConnectionFactory connectionFactory, IHubProtocol protocol
...
+            }
+
             using (_logger.BeginScope(_logScope))
             {
                 await StartAsyncCore(cancellationToken).ForceAsync();
             }
+
+            if (!TryChangeState(HubConnectionState.Connecting, HubConnectionState.Connected))
+            {
+                throw new IOException($"The ${nameof(HubConnection)} left the connecting state during start.");
```
BrennanConroy: Is this the right exception type?

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -137,19 +175,27 @@ public partial class HubConnection
...
-        {
-            get
-            {
-                // Copy reference for thread-safety
-                var connectionState = _connectionState;
-                if (connectionState == null || connectionState.Stopped)
-                {
-                    return HubConnectionState.Disconnected;
-                }
+        public HubConnectionState State { get; private set; }
```
BrennanConroy: Small issue with this change is that the following code could now fail. We could decide it's not important enough, but I want to bring it up at least.
```csharp
if (connection.State == HubConnectionState.Connected || connection.State == HubConnectionState.Reconnecting)
{
    return something;
}
Assert.False(connection.State == HubConnectionState.Connected); // This could be hit now
```

src/SignalR/clients/csharp/Client.Core/src/HubConnectionBuilderExtensions.cs
halter73: Considering this is effectively adding a service, I feel it would be more consistent with the rest of the framework to have the last reconnect configuration win.

halter73:
@mikaelm12 @BrennanConroy @davidfowl Feel free to go ahead and review this.

I plan to add three more tests ("can reconnect", "can reconnect after negotiate redirect", "can reconnect after skipping negotiate") which will be end-to-end similar to [these TS tests](https://github.com/aspnet/AspNetCore/pull/8566/files#diff-1c30bdd79e21ac4ce12ab9663aa438e0), and that's it.

src/SignalR/clients/csharp/Client/test/UnitTests/HubConnectionTests.Reconnect.cs
```diff
@@ -0,0 +1,892 @@
...
+                var firstException = new Exception();
+                testConnectionFactory.CurrentTestConnection.CompleteFromTransport(firstException);
+
+                Assert.Same(firstException, await reconnectingErrorTcs.Task.OrTimeout());
+                Assert.Single(retryContexts);
+                Assert.Same(firstException, retryContexts[0].RetryReason);
+                Assert.Equal(0, retryContexts[0].PreviousRetryCount);
+                Assert.Equal(TimeSpan.Zero, retryContexts[0].ElapsedTime);
+
+                await hubConnection.StopAsync();
```
BrennanConroy:
```suggestion
                await hubConnection.StopAsync().OrTimeout();
```

BrennanConroy (commented):
Still have to review the meat of the PR, but here's some feedback in the meantime.

src/SignalR/samples/ClientSample/HubSample.cs
```diff
@@ -46,6 +46,8 @@ public static async Task<int> ExecuteAsync(string baseUrl)
                 connectionBuilder.WithUrl(uri);
             }
 
+            connectionBuilder.WithAutomaticReconnect();
```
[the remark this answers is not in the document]
halter73: Good catch. Teaches me to be careful when reordering assignments at the last minute without rerunning tests. https://github.com/aspnet/AspNetCore/pull/9535/commits/cb8d28bfef50100724c8a4d74197552431aa7d67

src/SignalR/clients/csharp/Client.Core/src/IRetryPolicy.cs
mikaelm12: I could be convinced either way as far as the naming goes. As long as we are consistent. 

> Currently, we never retry on initial start failures

Don't we try to fallback? Would you configure retry policies that would take precedence over transport fall back? The next question becomes would we go through the retry strategy over all the transports? Questions for another time though.

src/SignalR/samples/ClientSample/HubSample.cs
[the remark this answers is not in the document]
BrennanConroy: I still don't see how this reconnects, you removed the loop that waits for the connection to be started again.

BrennanConroy (commented):
Last I checked there were some nullrefs in a couple tests.

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -1084,144 +1185,290 @@ async Task StartProcessingInvocationMessages(ChannelReader<InvocationMessage> in
...
         }
 
-        private void OnServerTimeout()
+        private OperationCanceledException GetOperationCanceledException(string message, Exception innerException, CancellationToken cancellationToken)
         {
-            _connectionState.CloseException = new TimeoutException(
-                $"Server timeout ({ServerTimeout.TotalMilliseconds:0.00}ms) elapsed without receiving a message from the server.");
-            _connectionState.Connection.Transport.Input.CancelPendingRead();
+#if NETCOREAPP3_0
+            return new OperationCanceledException(message, innerException, _state.StopCts.Token);
```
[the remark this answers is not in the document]
halter73: I'm pretty sure it just sets the OperationCanceledException.CancellationToken property.

https://docs.microsoft.com/en-us/dotnet/api/system.operationcanceledexception.-ctor?view=netcore-3.0#System_OperationCanceledException__ctor_System_String_System_Exception_System_Threading_CancellationToken_

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -1084,144 +1185,290 @@ async Task StartProcessingInvocationMessages(ChannelReader<InvocationMessage> in
...
+        }
 
-            // Cancel any outstanding invocations within the connection lock
-            connectionState.CancelOutstandingInvocations(connectionState.CloseException);
+        private void RunCloseEvent(Exception closeException)
+        {
+            var closed = Closed;
 
-            if (connectionState.CloseException != null)
+            async Task RunClosedEventAsync()
```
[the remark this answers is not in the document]
halter73: I'm intentionally allocating a closure. This happens once per connection close. We're already allocating a state machine so we can await AwaitableThreadPool.Yield().

src/SignalR/clients/csharp/Client.Core/src/Internal/DefaultRetryPolicy.cs
```diff
@@ -0,0 +1,41 @@
...
+
+            for (int i = 0; i < retryDelays.Length; i++)
```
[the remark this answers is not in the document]
halter73: I don't think looping to copy is that bad in this case. I'm not sure I trust Array.Copy to do the write thing when copying an array of TimeSpan structs to an array of `Nullable<TimeSpan>` structs.

src/SignalR/clients/csharp/Client/test/UnitTests/HubConnectionTests.Reconnect.cs
```diff
@@ -0,0 +1,892 @@
...
+                    }
+
+                    return Task.CompletedTask;
+                }
+
+                testConnectionFactory = new ReconnectingConnectionFactory(() => new TestConnection(OnTestConnectionStart));
+                builder.Services.AddSingleton<IConnectionFactory>(testConnectionFactory);
+
+                var retryContexts = new List<RetryContext>();
+                var mockReconnectPolicy = new Mock<IRetryPolicy>();
```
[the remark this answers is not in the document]
halter73: If we want to clean that up later, we can.

src/SignalR/clients/csharp/Client/test/UnitTests/HubConnectionTests.Reconnect.cs
```diff
@@ -0,0 +1,892 @@
...
+
+                await closedErrorTcs.Task.OrTimeout();
+
+                Assert.Equal(0, reconnectingCount);
+                Assert.Equal(0, reconnectedCount);
+            }
+        }
+
+        [Fact]
+        public async Task ReconnectDoesntStartIfConnectionIsLostDuringInitialHandshake()
```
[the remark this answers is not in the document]
halter73: This is the same as it is in TS.

src/SignalR/clients/csharp/Client/test/UnitTests/HubConnectionTests.Reconnect.cs
```diff
@@ -0,0 +1,892 @@
...
+
+                Assert.Same(firstException, await reconnectingErrorTcs.Task.OrTimeout());
+                Assert.Single(retryContexts);
+                Assert.Same(firstException, retryContexts[0].RetryReason);
+                Assert.Equal(0, retryContexts[0].PreviousRetryCount);
+                Assert.Equal(TimeSpan.Zero, retryContexts[0].ElapsedTime);
+
+                var secondException = new Exception();
+                var stopTask = hubConnection.StopAsync();
+                connectionStartTcs.SetResult(null);
```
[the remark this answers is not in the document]
halter73: The stopTask we verify RanToCompletion at the end of the test awaits the ReconnectTask.

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -194,7 +233,43 @@ public async Task StartAsync(CancellationToken cancellationToken = default)
...
+                }
+
+                // The StopCts is canceled at the start of StopAsync should be reset every time the connection finishes stopping.
+                // If this token is currently canceled, it means that StartAsync was called while StopAsync was still running.
+                if (_state.StopCts.Token.IsCancellationRequested)
+                {
+                    throw new InvalidOperationException($"The {nameof(HubConnection)} cannot be started while {nameof(StopAsync)} is running.");
+                }
+
                 await StartAsyncCore(cancellationToken).ForceAsync();
```
```suggestion
                 await StartAsyncCore(cancellationToken);
```

src/SignalR/clients/csharp/Client/test/UnitTests/HubConnectionTests.ConnectionLifecycle.cs
```diff
@@ -149,14 +149,17 @@ Task<ConnectionContext> ConnectionFactory(TransferFormat format)
                     // Wait to hit DisposeAsync on TestConnection (which should be after StopAsync has cleared the connection state)
                     await syncPoint.WaitForSyncPoint();
 
-                    // We should be able to start now, and StopAsync hasn't completed, nor will it complete while Starting
+                    // We should not yet bet able to start now because StopAsync hasn't completed
```
BrennanConroy:
```suggestion
                    // We should not yet be able to start now because StopAsync hasn't completed
```

BrennanConroy (approved):
I'd like you to take a look at the `stopCts` race I mentioned and see if it's actually an issue

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -1084,144 +1185,290 @@ async Task StartProcessingInvocationMessages(ChannelReader<InvocationMessage> in
...
+            {
+                SafeAssert(ReferenceEquals(_state.CurrentConnectionStateUnsynchronized, null),
+                    "Someone other than Reconnect set the connection state!");
+
+                var elapsedTime = DateTime.UtcNow - reconnectStartTime;
+                Log.ReconnectAttempsExhausted(_logger, previousReconnectAttempts, elapsedTime);
+
+                _state.ChangeState(HubConnectionState.Reconnecting, HubConnectionState.Disconnected);
+
+                var message = $"Reconnect retries have been exhausted after {previousReconnectAttempts} failed attempts and {elapsedTime} elapsed. Disconnecting.";
```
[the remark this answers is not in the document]
halter73: I think "00:01:22 elapsed" is pretty understandable. I trust the default settings for TimeSpan.ToString() to produce something reasonable. Since I don't see an alternative format suggested, I'll leave this open to future PRs.

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -397,9 +456,31 @@ private Task CloseAsync(ConnectionContext connection)
         // if we're disposing.
         private async Task StopAsyncCore(bool disposing)
         {
-            // Block a Start from happening until we've finished capturing the connection state.
+            // StartAsync acquires the connection lock for the duration of the handshake.
+            // ReconnectAsync also acquires the connection lock for reconnect attempts and handshakes.
+            // Cancel the StopCts without acquiring the lock so we can short-circuit it.
+            _state.StopCts.Cancel();
```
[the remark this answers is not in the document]
halter73: It does seem scary to cancel the StopCts outside the lock, but we don't need to worry about the connection ending up in a no-longer-startable state because [this logic](https://github.com/aspnet/AspNetCore/pull/9535/files#diff-f10b84c92f9edf5f451fd3213c4a6a3bR265) in StartAsync() will reset the StopCts if it fails in this way.

src/SignalR/clients/csharp/Client.Core/src/HubConnectionBuilderExtensions.cs
```diff
@@ -23,5 +24,49 @@ public static IHubConnectionBuilder ConfigureLogging(this IHubConnectionBuilder
...
+            // REVIEW: If we have a separate retry policy for retrying on start failures vs retrying only after
+            // a connection was fully established, would we create an IStartRetryPolicyHolder with a single IRetryPolicy
+            // property so they could independently retrieved from the IoC container?
+
+            hubConnectionBuilder.Services.AddSingleton<IRetryPolicy>(new DefaultRetryPolicy());
```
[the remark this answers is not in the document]
halter73: I'd rather have the last configuration win. The TryAdd extension method on ServiceCollection would be tempting if it returned a bool saying whether the add succeeded. Then we could throw for double configuration, like we do in the TS client. We could obviously just enumerate the ServiceCollection ourselves to see if theres already an IRetryPolicy policy there, but that seems like a lot of effort. If someone really cares they can send a PR.

halter73:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=184100

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/2488

src/SignalR/clients/csharp/Client.Core/src/HubConnectionState.cs
mikaelm12: Use the natural ordering and just file an announcement

anurse: I'd rather just preserve the ordering. Yes, we do breaking changes in major versions, but only ones that provide value. Adding a new enum variant should be done in a non-breaking way IMO. Just tack this on to the end instead of in the middle.

PR #9537, opened 2019-04-19
UriHelper use Ordinal string.IndexOf

benaadams:
Was wondering why the Proxy scenario was hitting Globablization https://github.com/aspnet/AspNetCore/issues/9404#issuecomment-484752406

`string.IndexOf(string)` is current culture by default; unlike most of the other string apis which are Ordinal; so switch to using Ordinal IndexOf to find things.

/cc @GrabYourPitchforks as I know you love the inconsistency of these apis 😉

### DIFF
--- a/src/Http/Http.Extensions/src/UriHelper.cs
+++ b/src/Http/Http.Extensions/src/UriHelper.cs
@@ -12,7 +12,7 @@ namespace Microsoft.AspNetCore.Http.Extensions
     public static class UriHelper
     {
         private const char ForwardSlash = '/';
-        private const char Pound = '#';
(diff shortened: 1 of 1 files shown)


stephentoub (approved):

JamesNK (approved):

Tratcher (approved):

filipnavara (approved):

GrabYourPitchforks (approved):

GrabYourPitchforks:
Also, enable Roslyn Code Analyzers over the source. It would have caught this. :)
Reference: https://github.com/dotnet/roslyn-analyzers/pull/915

PR #9539, opened 2019-04-19
Use .IndexOf(char/byte) where possible

benaadams:
As its faster. Also string.IndexOf(string) is by default CurrentCulture so its much faster than that.

Follow up to https://github.com/aspnet/AspNetCore/pull/9537

FormPipeReader goes a bit ugly using the ternaries

/cc @GrabYourPitchforks for more spice


rynowak:
:shipit: from  me on the routing change. Someone else should take a look at the form reader part :)

benaadams:
>  Someone else should take a look at the form reader part :)

I think I'm going to back that out as its too ugly; and the current version shouldn't be too bad, even if its doing extra work.

benaadams:
K, dropped the Formreader change

Tratcher (approved):

anurse:
Looks nice and safe to me. I'll merge now for preview 5.

PR #9549, opened 2019-04-19
Lazy create ExceptionHandlerMiddleware statemachine

benaadams:
One of the last remaining holdouts:

image

### DIFF
--- a/src/Middleware/Diagnostics/src/ExceptionHandler/ExceptionHandlerMiddleware.cs
+++ b/src/Middleware/Diagnostics/src/ExceptionHandler/ExceptionHandlerMiddleware.cs
@@ -26,13 +26,13 @@ namespace Microsoft.AspNetCore.Diagnostics
             RequestDelegate next,
             ILoggerFactory loggerFactory,
             IOptions<ExceptionHandlerOptions> options,
-            DiagnosticListener diagnosticSource)
+            DiagnosticListener diagnosticListener)
         {
             _next = next;
             _options = options.Value;
             _logger = loggerFactory.CreateLogger<ExceptionHandlerMiddleware>();
             _clearCacheHeadersDelegate = ClearCacheHeaders;
-            _diagnosticListener = diagnosticSource;
+            _diagnosticListener = diagnosticListener;
             if (_options.ExceptionHandler == null)
             {
                 if (_options.ExceptionHandlingPath == null)
--- a/src/Middleware/Diagnostics/src/ExceptionHandler/ExceptionHandlerMiddleware.cs
+++ b/src/Middleware/Diagnostics/src/ExceptionHandler/ExceptionHandlerMiddleware.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Diagnostics;
+using System.Runtime.ExceptionServices;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Diagnostics.Internal;
@@ -19,19 +20,19 @@ namespace Microsoft.AspNetCore.Diagnostics
         private readonly ExceptionHandlerOptions _options;
         private readonly ILogger _logger;
         private readonly Func<object, Task> _clearCacheHeadersDelegate;
-        private readonly DiagnosticSource _diagnosticSource;
+        private readonly DiagnosticListener _diagnosticListener;
 
         public ExceptionHandlerMiddleware(
             RequestDelegate next,
             ILoggerFactory loggerFactory,
             IOptions<ExceptionHandlerOptions> options,
-            DiagnosticSource diagnosticSource)
+            DiagnosticListener diagnosticSource)
         {
             _next = next;
             _options = options.Value;
             _logger = loggerFactory.CreateLogger<ExceptionHandlerMiddleware>();
             _clearCacheHeadersDelegate = ClearCacheHeaders;
-            _diagnosticSource = diagnosticSource;
+            _diagnosticListener = diagnosticSource;
             if (_options.ExceptionHandler == null)
             {
                 if (_options.ExceptionHandlingPath == null)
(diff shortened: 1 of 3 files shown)


src/Middleware/Diagnostics/src/ExceptionHandler/ExceptionHandlerMiddleware.cs
```diff
@@ -19,19 +20,19 @@ public class ExceptionHandlerMiddleware
         private readonly ExceptionHandlerOptions _options;
         private readonly ILogger _logger;
         private readonly Func<object, Task> _clearCacheHeadersDelegate;
-        private readonly DiagnosticSource _diagnosticSource;
+        private readonly DiagnosticListener _diagnosticListener;
```
benaadams: Change from `DiagnosticSource` to `DiagnosticListener` as second has fast enabled check (and used elsewhere for it)

How breaking is a change to a standard DI type?

Tratcher: Officially, changing the constructor is breaking and we wouldn't do it in a minor or patch. However, since this is a major and the constructor is only expected to be called via DI it's likely fine.

src/Middleware/Diagnostics/src/ExceptionHandler/ExceptionHandlerMiddleware.cs
```diff
@@ -45,71 +46,106 @@ public class ExceptionHandlerMiddleware
...
-                        _diagnosticSource.Write("Microsoft.AspNetCore.Diagnostics.HandledException", new { httpContext = context, exception = ex });
-                    }
+        private async Task HandleException(HttpContext context, ExceptionDispatchInfo edi)
+        {
+            _logger.UnhandledException(edi.SourceException);
+            // We can't do anything if the response has already started, just abort.
+            if (context.Response.HasStarted)
+            {
+                _logger.ResponseStartedErrorHandler();
+                edi.Throw();
```
Tratcher: Would the stack be any cleaner if we included these first few lines directly in the catch block and re-threw directly?

Tratcher (approved):

src/Middleware/Diagnostics/src/ExceptionHandler/ExceptionHandlerMiddleware.cs
```diff
@@ -19,19 +20,19 @@ public class ExceptionHandlerMiddleware
...
         private readonly Func<object, Task> _clearCacheHeadersDelegate;
-        private readonly DiagnosticSource _diagnosticSource;
+        private readonly DiagnosticListener _diagnosticListener;
 
         public ExceptionHandlerMiddleware(
             RequestDelegate next,
             ILoggerFactory loggerFactory,
             IOptions<ExceptionHandlerOptions> options,
-            DiagnosticSource diagnosticSource)
+            DiagnosticListener diagnosticSource)
```
NinoFloris: Can the name be changed with it?

benaadams: Sure :)

src/Middleware/Diagnostics/src/ExceptionHandler/ExceptionHandlerMiddleware.cs
benaadams: The first catch (sync) *should* be uncommon; as most exceptions *should* be wrapped in a task; so its more about reusing the same code.

For the async am trying to avoid recreating this issue https://github.com/dotnet/roslyn/issues/26567 by leaving the catch prior to doing all the extra work

Tratcher:
Woops, ref assembly...

Tratcher (approved):

Tratcher:
Note to self: merge after branching.

src/Middleware/Diagnostics/src/ExceptionHandler/ExceptionHandlerMiddleware.cs
davidfowl: Does the exception handling path need to be faster?

src/Middleware/Diagnostics/src/ExceptionHandler/ExceptionHandlerMiddleware.cs
```diff
@@ -45,71 +46,106 @@ public class ExceptionHandlerMiddleware
...
+            }
+            finally
+            {
+                context.Request.Path = originalPath;
             }
+
+            edi.Throw(); // Re-throw the original if we couldn't handle it
         }
 
         private Task ClearCacheHeaders(object state)
```
davidfowl: static?

benaadams: Changed

src/Middleware/Diagnostics/src/ExceptionHandler/ExceptionHandlerMiddleware.cs
Tratcher: A lot of this is about making the success case faster.

nixa333:
I'd like to understand why you did it this way and apparently I don't currently have enough knowledge to figure it out myself. 
Why do we choose to do the following:
- Using the local static function Awaited, instead of private non-static function
- Not awaiting inside Invoke method but returning the uncompleted Task

If someone would be so kind to explain this, or point me to the right keywords I could google, I'd be grateful.

Cheers!

davidfowl:
> Using the local static function Awaited, instead of private non-static function

Purely a style thing. Local functions are great when the function is used in a single place.

> Not awaiting inside Invoke method but returning the uncompleted Task

We're trying to avoid the async state machine because of the CPU overhead. This code optimizes for the non-exceptional path.

PR #9555, opened 2019-04-19
Lock during finalizer to not race with dispose in MemoryPoolBlock

jkotalik:
Fixes https://github.com/aspnet/AspNetCore-Internal/issues/2324

From what I can tell, the reason these tests were failing were due to a race between Dispose and the Finalizer of the MemoryPoolBlock. The test causing it was https://github.com/aspnet/AspNetCore/blob/77d66682ab29240f3e1b030e24b38baa7b6788cd/src/Servers/Kestrel/Core/test/PipelineExtensionTests.cs#L87, which has a Dispose method on the test itself.

My only thought was the _memoryPool that is created (DiagnosticMemoryPool) isn't calling dispose on the underlying MemoryPool (SlabMemoryPool).

### DIFF
--- a/src/Servers/Kestrel/Core/test/SlabMemoryPoolTests.cs
+++ b/src/Servers/Kestrel/Core/test/SlabMemoryPoolTests.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
+// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System.Buffers;
--- a/src/Servers/Kestrel/Core/test/SlabMemoryPoolTests.cs
+++ b/src/Servers/Kestrel/Core/test/SlabMemoryPoolTests.cs
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft. All rights reserved.
+﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System.Buffers;
--- a/src/Shared/Buffers.MemoryPool/SlabMemoryPool.cs
+++ b/src/Shared/Buffers.MemoryPool/SlabMemoryPool.cs
@@ -160,6 +160,18 @@ namespace System.Buffers
             }
         }
 
+        internal void RefreshBlock(MemoryPoolSlab slab, int offset, int length)
+        {
+            lock (_disposeSync)
+            {
+                if (!_isDisposed && slab != null && slab.IsActive)
+                {
+                    // Need to make a new object because this one is being finalized
+                    Return(new MemoryPoolBlock(this, slab, offset, length));
+                }
+            }
+        }
+
         protected override void Dispose(bool disposing)
         {
             if (_isDisposed)
--- a/src/Shared/Buffers.MemoryPool/SlabMemoryPool.cs
+++ b/src/Shared/Buffers.MemoryPool/SlabMemoryPool.cs
@@ -160,6 +160,7 @@ namespace System.Buffers
             }
         }
 
+        // This method can ONLY be called from the finalizer of MemoryPoolBlock
         internal void RefreshBlock(MemoryPoolSlab slab, int offset, int length)
         {
             lock (_disposeSync)
@@ -167,6 +168,8 @@ namespace System.Buffers
                 if (!_isDisposed && slab != null && slab.IsActive)
                 {
                     // Need to make a new object because this one is being finalized
+                    // Note, this must be called within the _disposeSync lock because the block
+                    // could be disposed at the same time as the finalizer.
                     Return(new MemoryPoolBlock(this, slab, offset, length));
                 }
             }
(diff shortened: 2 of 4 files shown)


anurse:
If you feel super confident that the risk is very low, I'm OK with merging this for preview 5. But I'd lean towards deferrring to preview 6 :).

jkotalik:
Going to wait until preview6 for all of my test failure fixes.

src/Shared/Buffers.MemoryPool/SlabMemoryPool.cs
```diff
@@ -160,6 +160,18 @@ internal void Return(MemoryPoolBlock block)
             }
         }
 
+        internal void RefreshBlock(MemoryPoolSlab slab, int offset, int length)
+        {
+            lock (_disposeSync)
+            {
+                if (!_isDisposed && slab != null && slab.IsActive)
+                {
+                    // Need to make a new object because this one is being finalized
```
halter73: Maybe add a comment indicating that the MemoryPoolBlock ctor itself must be called with the _disposeSync lock.

src/Shared/Buffers.MemoryPool/SlabMemoryPool.cs
```diff
@@ -160,6 +160,18 @@ internal void Return(MemoryPoolBlock block)
             }
         }
 
+        internal void RefreshBlock(MemoryPoolSlab slab, int offset, int length)
```
halter73: I think this should have a name or comment that makes this abundantly clear that this is only called from the MemoryPoolBlock finalizer.

src/Servers/Kestrel/Core/test/SlabMemoryPoolTests.cs
```diff
@@ -26,4 +26,4 @@ public void DisposeWithActiveBlocksWorks()
             memoryPool.Dispose();
         }
     }
-}
\ No newline at end of file
+}
```
halter73: Nit: I normally don't complain about BOM changes and newlines at the end of files, but if the file is otherwise untouched, leave it.

halter73 (approved):

src/Shared/Buffers.MemoryPool/SlabMemoryPool.cs
```diff
@@ -160,6 +160,21 @@ internal void Return(MemoryPoolBlock block)
             }
         }
 
+        // This method can ONLY be called from the finalizer of MemoryPoolBlock
+        internal void RefreshBlock(MemoryPoolSlab slab, int offset, int length)
+        {
+            lock (_disposeSync)
```
halter73: What happens if the GC suspended all other threads while the MemoryPoolBlock finalizer runs and one of those suspended threads has the _disposeSync lock? Normally the _disposeSync lock isn't taken for long, but if the thread was suspended, I could see it potentially deadlocking.

@Maoni0 Can you safely acquire a lock in a finalizer?

halter73: I found https://blogs.msdn.microsoft.com/yunjin/2005/07/05/special-threads-in-clr/

>Because there is only one thread to run all finalizers, if one finalizer is blocked, no other finalizers could run. So it is discouraged to take any lock in finalizer. Also see [Maoni Stephens's blog](http://blogs.msdn.com/maoni/archive/2004/11/04/252697.aspx) for details about finalizer thread.

It recommends not taking any lock, but only because it can stop other finalizers from running. It doesn't say whether the GC suspends normal threads while the finalizer is running.

[This doc](https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/fundamentals#concurrent-garbage-collection) seems to indicate the GC will not suspend other threads .NET Core unless concurrent garbage collection is disabled. That said, as a server, we need to make sure we don't deadlock when concurrent garbage connection is disabled considering we recommend disabling it "[t]o improve performance when several processes are running, disable concurrent garbage collection."

@Maoni0 Please feel free to correct any inaccuracies in my understanding.

halter73: Further reading now leads me to believe that finalizers don't run "during" a GC where the runtime might suspend threads, but that it's merely triggered by the collection adding objects-waiting-to-be-finalized to the finalizer queue. So the object being finalized is still alive as far as GC is concerned until the object is dequeued from the finalizer queue by the finalizer thread and the object is no longer referenced (that is *if* the object is no longer referenced). When collections are actually run, even the finalizer thread is suspended at least if concurrent garbage collection is disabled.

This is making me think that the effort we go through to reconstruct the MemoryPoolBlock might not be worth it, and that we should just add `this` back to the pool in the finalizer and "resurrect" the MemoryPoolBlock.

I'm not writing this long explanation, because I'm confident I'm correct. I just want to explain my current understanding so the experts can clear up what I'm wrong about.

Maoni0: sorry I had not gotten around to responding to this - to GC the finalizer thread is no different from any other managed threads - it will be suspended when GC calls SuspendEE to suspend managed threads.

halter73: Thanks!

PR #9558, opened 2019-04-19
Add support for System.Text.Json based JsonResult executor

### LINKED ISSUE #9554: Add System.Text.Json based JsonResultExecutor
This was missed as part of https://github.com/aspnet/AspNetCore/issues/7256.

pranavkm:
Fixes https://github.com/aspnet/AspNetCore/issues/9554


src/Mvc/Mvc.Core/src/Formatters/TranscodingReadStream.cs
```diff
@@ -224,11 +225,13 @@ public override void Write(byte[] buffer, int offset, int count)
 
         protected override void Dispose(bool disposing)
         {
-            ArrayPool<char>.Shared.Return(_charBuffer.Array);
```
pranavkm: We were prone to returning a buffer twice if the buffer was double-disposed. Guarding against it. In addition `base.Dispose` calls a synchronous `Flush`.  Callers already explicitly call `FlushAsync` inside of a `try` block, so there isn't any benefit to doing a sync flush here. Discovered this when I was adding tests for new JsonResult executor that verifies no sync operation is performed on the Response stream.

rynowak (approved):

src/Mvc/Mvc.Core/src/Formatters/TranscodingReadStream.cs
davidfowl: Just FYI returning the buffer twice taints the pool for every other component in the system.

pranavkm: Yup. We don't have code that double-disposes (and the type's internal), but it's better to be defensive here.

Issue #7492, opened 2019-02-12
Rename items (elements) used in service reference scenarios


dougbu:
This is a pivot on what users see in their projects when using service reference features.

Current items are
``` xml
<ServiceFileReference Include="SomeFile" ... />
<ServiceUriReference Include="https://SomeUri" ... />
<ServiceProjectReference Include="../SomeProject/SomeProject.csproj" ... />
```
New elements will be
~~`<ProtobufReference Include="SomeFile" ... />`~~
``` xml
<OpenApiReference Include="SomeFile" ... />
<OpenApiProjectReference Include="../SomeProject/SomeProject.csproj" ... />
```
~~`<OpenApiReference Include="SomeFile" SourceUri="https://SomeUri" ... />`~~
~~`<ProtobufReference Include="../SomeProject/SomeProject.csproj" ... />`~~
``` xml
<OpenApiProjectReference Include="../SomeProject/SomeProject.csproj" ... />
```

dougbu:
Open questions:
1. Do we use the new item names when calling out to code generation partners? I lean toward "yes" for consistency but that adds more parts that need to change. @glennc do you want to make the call here?
2. Is the `SourceUri` case interesting for gRPC too? @shirhatti?
3. Is the `<ProtobufReference Include="../SomeProject/SomeProject.csproj" ... />` case a bit too confusing? Naming (of the extracted document -- just in the Open API case, the C# files, and the the generated class) might be simpler and more consistent if this were something like `<ProtobufReference Include="SomeFile" SourceProject="../SomeProject/SomeProject.csproj" ... />`. @glennc?
4. We're leaning toward supporting the (Open API-only) case where the service project does _not_ generate a document on disk when it builds. Will there be an explicit gesture to "refresh" the document a client project extracts? Or, will this be done automatically e.g. on every non-design-time build? @glennc and @mlorbetske I think both of you have opinions here&hellip;

glennc:
We talked about how to handle the upgrade from preview3 to 4 with the new names. What we want is for it to fail hard and fast so that customers know straight away to go rename the element.

glennc:
Answering open questions:

1. Yeah, consistency seems good and we have to update them anyway
2. This doesn't matter anymore due to the naming, we also cut URI support
3. The project case is just a wildacard for * references
4. Doesn't matter given the new split.

dougbu:
From discussions today:
[...]

Issue #4945, opened 2018-11-19
Change default document directory to obj


glennc:
Today the directory that we store the API document is empty, meaning that we put it in the root of the project. For this type of asset, one that is purely used to make sure we don't redo work we don't need to, I think obj is a better default.

glennc:
1. Is it fine to leave the code generation default to the root of the client project? Yes.
2. Will users by default check the documents in on the server project? Because if yes, we should put it in root. I think we should start with no until people tell us otherwise. So the server doc would go to into obj or somewhere not visible by default.

dougbu:
ce8f053af7e9 addresses code generation location. Will add a note to remind me to include documents location in #8242

Issue #4944, opened 2018-11-19
Generated files re-appearing without explicit build


glennc:
If I delete a generated client file then it will reappear immediately. My theory for this was that changing the files in the project was causing a design-time build to happen.

glennc:
From discussion today: Implement ability to either generate files during design time build or not. Once we have that configuration (A property, we think global to the project is OK) we will play with the feature both ways and see which experience we like better for the default.

dougbu:
ce8f053af7e9

Issue #4916, opened 2018-09-19
Remove need for users to special-case second and later clients


dougbu:
#### Is this a Bug or Feature request?
Enhancement that's part of #4896

#### Steps to reproduce (preferably a link to a GitHub repo with a repro project)
See https://github.com/dougbu/CodeGeneration repo and its [`ClientConsole` project](https://github.com/dougbu/CodeGeneration/blob/fbc0e82aa186f4d8de2ba7ece21f87d550be1817/ClientConsole/ClientConsole.csproj). Lines of particular interest are https://github.com/dougbu/CodeGeneration/blob/fbc0e82aa186f4d8de2ba7ece21f87d550be1817/ClientConsole/ClientConsole.csproj#L23-L26

#### Description of the problem
NSwag code generation creates `Exception` classes by default. Unfortunately, these classes are not nested and do not have unique names. So, when multiple generated clients are included in a project, compilation errors ensue.

This is not mandatory for the Preview3 milestone because multiple generated clients per project is a grow-up scenario and users _can_ provide the necessary `%(NSwagCSharpCodeGeneratorOptions)` metadata to avoid problems.

Most likely fix is to update the [`GetProjectReferenceMetadata`](https://github.com/aspnet/Mvc/blob/792d5b49a27f5d4d99b3238019deaed5ba9eca0f/src/Microsoft.Extensions.ApiDescription.Client/GetFileReferenceMetadata.cs) task to add `FirstForGenerator` metadata to the relevant items.

#### Version of `Microsoft.AspNetCore.Mvc` or `Microsoft.AspNetCore.App` or `Microsoft.AspNetCore.All`
See the features/client.code.generation branch branch. That work is intended for the 2.2 Preview 3 milestone.

dougbu:
Another oops, I'm working on this now but not #6582 (that one will come just after #8242).

dougbu:
ce8f053af7e9

PR #9559, opened 2019-04-19
Rename all the service reference things

dougbu:
- #7492
- remove document generation from client project; will be included in Web API project infrastructure (coming soon)

Also
- provide `%(FirstForGenerator)` metadata, #4916
- add `$(OpenApiGenerateAtDesignTime)` property (default `true` for now), #4944
- generate code in `obj` directory by default, #4945


dougbu:
Pity GitHub doesn't recognize the .props and .targets files were renamed before things in them were renamed. If you're primarily interested in the delta from existing code, I suggest
``` PowerShell
git checkout dougbu/code.generation.renames.7492
git show --find-renames=20 -- *.props
git show --find-renames=20 -- *.targets
```

dougbu:
@natemcmaster I know you found one typo. Anything else ?

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.targets
```diff
@@ -0,0 +1,151 @@
...
+    -->
+    <ProjectReference Include="@(OpenApiProjectReference)" Exclude="@(ProjectReference)" ReferenceOutputAssembly="false" />
+    <ProjectReference Update="@(OpenApiProjectReference)" OpenApiReference="true" />
+  </ItemGroup>
+
+  <Target Name="_CreateOpenApiReferenceItemsForOpenApiProjectReferences"
+      AfterTargets="ResolveProjectReferences"
+      Condition="'%(ProjectReferenceWithConfiguration.OpenApiReference)' == 'true' AND '$(DesignTimeBuild)' != 'true' AND '$(BuildingProject)' == 'true'"
+      Inputs="%(ProjectReferenceWithConfiguration.Identity)"
+      Outputs="&lt;not-a-file !&gt;">
```
natemcmaster: This is a unsupported usage for inputs and outputs. It looks to me like you're trying to batch over ProjectReferenceWithConfiguration. You don't have to do this with target batching.

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.targets
```diff
@@ -0,0 +1,151 @@
...
+    <PropertyGroup>
+      <_Identity>%(ProjectReferenceWithConfiguration.Identity)</_Identity>
+      <_Temporary />
+    </PropertyGroup>
+
+    <MSBuild Target="OpenApiGetDocuments"
+        Project="$(_Identity)"
```
natemcmaster:
```suggestion
        Project="@(ProjectReferenceWithConfiguration)"
```

No need to create an intermediate property.

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.targets
```diff
@@ -0,0 +1,151 @@
...
+        Properties="%(ProjectReferenceWithConfiguration.SetConfiguration); %(ProjectReferenceWithConfiguration.SetPlatform); %(ProjectReferenceWithConfiguration.SetTargetFramework)"
+        RebaseOutputs="true"
+        RemoveProperties="%(ProjectReferenceWithConfiguration.GlobalPropertiesToRemove);TargetFrameworks;RuntimeIdentifier">
+      <Output TaskParameter="TargetOutputs" PropertyName="_Temporary" />
```
natemcmaster: There are some MSBuild way of doing this which don't require manually matching up some returned metadata from  a target. I would start by changing this to return an item instead of a property.

```suggestion
      <Output TaskParameter="TargetOutputs" ItemName="_ResolvedOpenApiProjectWithDocuments" />
```

Next, make the target you're invoking, in this case "OpenApiGetDocuments", look like this:

```xml
<Target Name="OpenApiGetDocuments"
              Returns="@(OpenApiProjectWithDocuments)">
   <ItemGroup>
       <OpenApiProjectWithDocuments Include="$(MSBuildProjectFullPath)">
          <Documents>$(stuffHere)</Documents>
      </OpenApiProjectWithDocuments>
  </ItemGroup>
</Target>
```

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.targets
```diff
@@ -0,0 +1,151 @@
...
+    </MSBuild>
+
+    <ItemGroup>
+      <OpenApiProjectReference Update="$(_Identity)">
```
natemcmaster: This doesn't work during target execution. It's one of those quirks. `Update` is only allowed during static evaluation. I believe newer versions of MSBuild will properly throw an error here. Older versions allowed this, but behave incorrectly.

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.targets
```diff
@@ -0,0 +1,151 @@
...
+  <Target Name="_GetMetadataForOpenApiReferences" Condition="'@(OpenApiReference)' != ''">
+    <ItemGroup>
+      <_Temporary Remove="@(_Temporary)" />
+    </ItemGroup>
+
+    <GetFileReferenceMetadata Inputs="@(OpenApiReference)"
+        Extension="$(DefaultLanguageSourceExtension)"
+        Namespace="$(RootNamespace)"
+        OutputDirectory="$(OpenApiDefaultOutputDirectory)">
+      <Output TaskParameter="Outputs" ItemName="_Temporary" />
```
natemcmaster:
```suggestion
      <Output TaskParameter="Outputs" ItemName="OpenApiReferenceWithMetadata" />
```

A good practice in MSBuild is to use new item groups as you transform the data, instead of updating the items in place. This helps with extensibility because then people extending your targets don't need to worry so much about the order in which targets execute.

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.targets
```diff
@@ -0,0 +1,151 @@
...
+      <_Files Include="@(OpenApiReference -> '%(OutputPath)')"
+          Condition="$([System.IO.File]::Exists('%(OpenApiReference.OutputPath)'))">
+        <OutputPathExtension>$([System.IO.Path]::GetExtension('%(OpenApiReference.OutputPath)'))</OutputPathExtension>
+      </_Files>
+      <_Directories Remove="@(_Directories)" />
+      <_Directories Include="@(OpenApiReference -> '%(OutputPath)')"
+          Condition="Exists('%(OpenApiReference.OutputPath)') AND ! $([System.IO.File]::Exists('%(OpenApiReference.OutputPath)'))" />
+
+      <!-- If OutputPath is a file, add it directly to relevant items. -->
+      <TypeScriptCompile Remove="@(_Files)" Condition="'%(_Files.OutputPathExtension)' == '.ts'" />
```
natemcmaster: Don't forget about `.tsx` :)

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.targets
```diff
@@ -0,0 +1,151 @@
...
+
+      <Compile Remove="@(_Files)"
+          Condition="'$(DefaultLanguageSourceExtension)' != '.ts' AND '%(_Files.OutputPathExtension)' == '$(DefaultLanguageSourceExtension)'" />
+      <Compile Include="@(_Files)"
+          Condition="'$(DefaultLanguageSourceExtension)' != '.ts' AND '%(_Files.OutputPathExtension)' == '$(DefaultLanguageSourceExtension)'">
+        <SourceDocument>%(OpenApiReference.FullPath)</SourceDocument>
+      </Compile>
+
+      <!-- Otherwise, add all descendant files with the expected extension. -->
+      <TypeScriptCompile Remove="@(_Directories -> '%(Identity)/**/*.ts')" />
```
natemcmaster: I'm assuming you're doing the `Remove` to avoid duplicate entries, right? A more efficient way to do this is in MSBuild is
```xml
<TypeScriptCompile Include=" new pattern here " Exclude="@(TypeScriptCompile)" />
```

natemcmaster (commented):
Only had time to look at Microsoft.Extensions.ApiDescription.Client.targets.

dougbu:
@glennc this PR partially addresses #7491 because it provides a default `%(CodeGenerator)` value. Does not add proposed `%(Language)` metadata.

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.targets
```diff
@@ -0,0 +1,149 @@
...
+      <!-- If OutputPath is a file, add it directly to relevant items. -->
+      <TypeScriptCompile Include="@(_Files)"
+          Exclude="@(TypeScriptCompile)"
+          Condition="'%(_Files.OutputPathExtension)' == '.ts' OR '%(_Files.OutputPathExtension)' == '.tsx'">
+        <SourceDocument>%(_Files.FullPath)</SourceDocument>
+      </TypeScriptCompile>
+
+      <Compile Include="@(_Files)"
+          Exclude="@(Compile)"
+          Condition="'$(DefaultLanguageSourceExtension)' != '.ts' AND '%(_Files.OutputPathExtension)' == '$(DefaultLanguageSourceExtension)'">
```
pranavkm: Have we tested this with F# projects? If not, should we limit this to 'Condition="'$(Language)'=='C#'"`?

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.targets
```diff
@@ -0,0 +1,149 @@
...
+  </PropertyGroup>
+
+  <!-- OpenApiProjectReference support. -->
+
+  <ItemGroup>
+    <!--
+      Do not change %(ReferenceOutputAssembly) if project contains duplicate @(ProjectReference) and
+      @(OpenApiProjectReference) items.
+    -->
+    <ProjectReference Include="@(OpenApiProjectReference)" Exclude="@(ProjectReference)">
```
pranavkm: Does this work if the project have different TFMs? Like does it work if you reference a `netcoreapp` project from a `netstandard` project? I know we struggled with this when we were authoring test projects in Razor.

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.targets
```diff
@@ -0,0 +1,149 @@
...
+      <NoWarn>NU1702</NoWarn>
+      <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
+    </ProjectReference>
+    <ProjectReference Update="@(OpenApiProjectReference)">
```
pranavkm: Any reason not to inline this with the previous include?

src/Mvc/Extensions.ApiDescription.Client/src/Microsoft.Extensions.ApiDescription.Client.nuspec
```diff
@@ -21,9 +21,5 @@
     <file src="buildMultiTargeting\*" target="buildMultiTargeting" />
     <file src="bin\$configuration$\net461\Microsoft.Extensions.ApiDescription.Tasks.*" target="tasks\net461" />
     <file src="bin\$configuration$\netstandard2.0\Microsoft.Extensions.ApiDescription.Tasks.*" target="tasks\netstandard2.0" />
-    <file src="..\..\dotnet-getdocument\src\bin\$configuration$\netcoreapp2.1\publish\*.*" target="tools" />
```
pranavkm: What happened to these?

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.targets
```diff
@@ -0,0 +1,149 @@
...
+
+    <GetFileReferenceMetadata Inputs="@(OpenApiReference)"
+        Extension="$(DefaultLanguageSourceExtension)"
+        Namespace="$(RootNamespace)"
+        OutputDirectory="$(OpenApiDefaultOutputDirectory)">
+      <Output TaskParameter="Outputs" ItemName="_Temporary" />
+    </GetFileReferenceMetadata>
+
+    <ItemGroup>
+      <OpenApiReference Remove="@(OpenApiReference)" />
```
pranavkm: Should this be a different itemgroup rather than re-using `OpenApiReference`? It would be easier to debug this and wouldn't modify a user specified item.

src/Mvc/Extensions.ApiDescription.Client/src/Microsoft.Extensions.ApiDescription.Client.nuspec
dougbu: They aren't necessary with the new client / server split. The tool will come back in Microsoft.Extensions.ApiDescription.Server

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.targets
dougbu: It doesn't work perfectly yet and will be refined. As-is, it brings a large reliability improvement over explicit `<MSBuild />` tasks to restore and build the server project.

For later consideration, could you point me to the Razor solution to this issue?

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.targets
dougbu: Because when `@(ProjectReference)` and `@(OpenApiProjectReference)` overlap we need to get the documents. The previous element is about the non-overlapping case and this covers everything.

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.targets
dougbu: No, this is intended to flow `@(OpenApiProjectReference)` items into `@(OpenApiReference)`. Besides, modifying user-specified items e.g. `@(Compile)` is just normal.

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.targets
dougbu: This condition is correct because compiled languages other than TypeScript are not mixed into other projects. That is, this is about the _exception_ to normal mono-language projects.

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.props
```diff
@@ -0,0 +1,67 @@
...
+      for output files that do not yet exist during design-time builds.
+    -->
+    <OpenApiGenerateAtDesignTime Condition="'$(OpenApiGenerateAtDesignTime)' == ''">true</OpenApiGenerateAtDesignTime>
+
+    <!--
+      $(OpenApiDefaultOutputDirectory) value is interpreted relative to the project folder, unless already an
+      absolute path.
+    -->
+    <OpenApiDefaultOutputDirectory
+        Condition="'$(OpenApiDefaultOutputDirectory)' == ''">$(BaseIntermediateOutputPath)</OpenApiDefaultOutputDirectory>
```
natemcmaster: FYI - BaseIntermediateOutputPath might not be fully set when this file is evaluated. Would recommend moving this to a .targets file instead. BaseIntermediateOutputPath isn't defined until Microsoft.NET.DefaultOutputPaths.targets is evaluated, which happens at the end of project evaluation. Props file evaluation happens near the beginning of project eval.

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.targets
natemcmaster: MSBuild and the SDK don't handle duplicate `<ProjectReference>` items well. This can cause some strange behaviors such as https://github.com/Microsoft/msbuild/issues/2688. Consider making this overlap an error.

natemcmaster (commented):
A few suggestions:

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.props
dougbu: Hmm, it's working at the moment. I'll move it if we see issues or if I otherwise _must_ update this PR.

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.targets
dougbu: The two blocks are separate specifically to avoid creating duplicate `<ProjectReference>` items. Why make overlap between `@(ProjectReference)` and `@(OpenApiProjectReference)` an error with that in place?

pranavkm (approved):

PR #9572, opened 2019-04-19
React to new System.Text.Json features

pranavkm:


src/Mvc/Mvc.Core/src/MvcOptions.cs
```diff
@@ -329,7 +329,20 @@ public int MaxModelBindingRecursionDepth
...
+            // from deserialization errors that might occur from deeply nested objects.
+            // This value is the same for model binding and Json.Net's serialization.
+            MaxDepth = DefaultMaxModelBindingRecursionDepth,
+
+            // Use camel casing for properties
+            PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
+
+            // Ignore casing when serializing values. Setting this to true keeps parity with Json.NET's
+            // behavior.
+            PropertyNameCaseInsensitive = true,
```
pranavkm: How do you feel about this? In the absence of this property, it is required to send camel-case values to successfully deserialize. Most of our tests use pascal-case when sending data. 

We can update tests, but I'm not sure how sad users would be when random casing doesn't work.

JamesNK: I think true is the right default for MVC. Also it keeps closer to Json.NET so less friction when upgrading. If someone wants to be strict then they can always change the setting.

pranavkm: I just spoke to @rynowak offline and he's of the opinion that having it be case sensitive by default would be the right way to go. Users are always free to change it.

rynowak: We should use the best default regardless of migration. You can always turn this *ON* for migrating.

I'm interested to hear arguments about case-insensitive is the right default for a REST API.

pranavkm: Made this case sensitive now and updated tests to be 🐫 cased

JamesNK: > I'm interested to hear arguments about case-insensitive is the right default for a REST API.

Nothing other than it slightly reduces issues when using an API and I don't see much disadvantage. I don't feel strongly about it. As long as it is configurable

pranavkm: We could always revisit this once we hear what users have to say with the preview5 bits.

rynowak: >  Nothing other than it slightly reduces issues when using an API and I don't see much disadvantage. I don't feel strongly about it. As long as it is configurable

APIs document their contract and are precise about what they support. JSON is a case sensitive format, and it's reasonable to follow along with that. I expect if we asked API users they would overwhelmingly support us following the spec.

rynowak (approved):

rynowak:
@ahsonkhan @steveharter  FYI

src/Mvc/Mvc.Core/src/MvcOptions.cs
JamesNK: JSON being case sensitive in the browser is prior art, not a spec. There is no spec about how JSON should map to a .NET object :yum:

When it comes to JSON my philosophy has been be flexible in what you accept and exact in what you produce.

src/Mvc/test/Mvc.FunctionalTests/SystemTextJsonOutputFormatterTest.cs
```diff
@@ -29,67 +29,13 @@ public override async Task Formatting_StringValueWithUnicodeContent()
...
-            // Act
-            var response = await Client.GetAsync($"/JsonOutputFormatter/{nameof(JsonOutputFormatterController.SimpleModelResult)}");
-
-            // Assert
-            await response.AssertStatusCodeAsync(HttpStatusCode.OK);
-            Assert.Equal(expected, await response.Content.ReadAsStringAsync());
-        }
-
-        [Fact]
-        public override async Task Formatting_CollectionType()
```
ahsonkhan: Why remove these tests?

src/Mvc/Mvc.ViewFeatures/test/Infrastructure/TempDataSerializerTestBase.cs
```diff
@@ -173,6 +173,72 @@ public void RoundTripTest_GuidValue()
...
+            var bytes = testProvider.Serialize(input);
+            var values = testProvider.Deserialize(bytes);
+
+            // Assert
+            var roundTripValue = Assert.IsType<string>(values[key]);
+            Assert.Equal(value.ToString(), roundTripValue);
+        }
+
+        [Fact]
+        public virtual void RoundTripTest_StringThatIsNotCompliantGuid()
```
ahsonkhan: Just to verify, we are ok with this, correct?

src/Mvc/Mvc.ViewFeatures/src/Infrastructure/DefaultTempDataSerializer.cs
```diff
@@ -46,19 +45,17 @@ internal class DefaultTempDataSerializer : TempDataSerializer
...
-                        else if (Guid.TryParseExact(stringValue, "B", out var guid))
+                        else if (item.Value.TryGetDateTime(out var dateTime))
                         {
-                            deserializedValue = guid;
+                            deserializedValue = dateTime;
                         }
                         else
                         {
-                            deserializedValue = stringValue;
+                            deserializedValue = item.Value.GetString();
```
ahsonkhan: Will the JSON strings more often than not be GUID/DateTimes?

src/Mvc/Mvc.Core/test/Formatters/JsonInputFormatterTestBase.cs
```diff
@@ -194,8 +195,8 @@ protected async Task ReadAsync_ReadsValidArray_AsList(Type requestedType)
 
             // Assert
             Assert.False(result.HasError);
-            var integers = Assert.IsType<List<int>>(result.Model);
-            Assert.Equal(new int[] { 0, 23, 300 }, integers);
```
ahsonkhan: What motivated this change?

src/Mvc/Mvc.ViewFeatures/src/Infrastructure/DefaultTempDataSerializer.cs
pranavkm: I'd expect the latter. It's not uncommon for users to JSON serialize on complex models so that they could put it in this dictionary. Is there a way to improve this?

src/Mvc/Mvc.ViewFeatures/test/Infrastructure/TempDataSerializerTestBase.cs
pranavkm: Yup, this is fine. We just need to able to document "a" behavior so we don't find it confusing \ surprising.

src/Mvc/test/Mvc.FunctionalTests/SystemTextJsonOutputFormatterTest.cs
pranavkm: They were overriding the baseline tests https://github.com/aspnet/AspNetCore/blob/master/src/Mvc/test/Mvc.FunctionalTests/JsonOutputFormatterTestBase.cs#L117-L129 because they were different from what we would want (i.e did not do camel-case). Now that it matches the baseline, we can have it call the one on the base.

src/Mvc/Mvc.Core/test/Formatters/JsonInputFormatterTestBase.cs
pranavkm: Good one. I thought I'd bring it up - @layomia was there a reason System.Text.Json returns a type that [wraps `List<T>` ](https://github.com/dotnet/corefx/blob/master/src/System.Text.Json/src/System/Text/Json/Serialization/Converters/DefaultEnumerableConverter.cs#L7) rather than just returning a `List<T>`? Not that it ultimately matters, but if I remember correctly we agreed on just doing a `List<T>` like Json.NET does.

ahsonkhan: Also cc @steveharter

PR #9582, opened 2019-04-19
Update solutions

rynowak:
We have a new SignalR package and it needs to be the added.


pranavkm (approved):

jkotalik (approved):

jkotalik:
_groovy_

BrennanConroy (approved):
Is there a way to automate this or is it just seen by opening .sln's?

rynowak:
@BrennanConroy there's a script you can run, but it requires you to knwo what solution needs to be updated. It also doesn't currently work 😆 

What I saw this broken I actually want to the git history to find out who added the project so I could *inform* them of their deeds, and then I realized that we don't really have a foolproof workflow 😆

PR #9585, opened 2019-04-19
Run selenium tests on mac and linux

ryanbrandenburg:
At this point this is just exploration to see how much work this is.


.azure/pipelines/jobs/default-build.yml
```diff
@@ -137,7 +137,14 @@ jobs:
     - powershell: |
         Write-Host "##vso[task.setvariable variable=SeleniumProcessTrackingFolder]$(BuildDirectory)\obj\selenium\"
         ./eng/scripts/InstallGoogleChrome.ps1
-
+      displayName: Install Chrome
+  - ${{ if eq(parameters.agentOs, 'Linux') }}:
```
natemcmaster: This should be unnecessary. Chrome is pre-installed in our Linux agents

https://github.com/Microsoft/azure-pipelines-image-generation/blob/master/images/linux/Ubuntu1604-README.md

natemcmaster (commented):
Can we de-duplicate Chrome installs? It's my understanding that this would be the 2nd download of Chrome as puppeteer already installs it.

https://github.com/aspnet/AspNetCore/blob/f0ef784783a6716e25287196d66fbb9f6222abf3/src/Middleware/CORS/test/FunctionalTests/CORS.FunctionalTests.npmproj#L15-L18

cc @pranavkm

PR #9586, opened 2019-04-19
Don't override jasmine.DEFAULT_TIMEOUT_INTERVAL from Common.ts

halter73:
I noticed in https://github.com/aspnet/AspNetCore-Internal/issues/2334 that a functional test defined in HubConnectionTests.ts was timing out after only 10 seconds instead of the 20 seconds configured in Common.ts.

This happens despite the fact that HubConnectionTests.ts imports Common.ts. I think this is caused by one or both of WebSocketTests.ts and WebWorkerTests.ts reconfiguring jasmine.DEFAULT_TIMEOUT_INTERVAL back to 10 seconds by the time the tests defined in HubConnectionTests.ts are run.

### DIFF
--- a/src/SignalR/clients/ts/FunctionalTests/ts/WebSocketTests.ts
+++ b/src/SignalR/clients/ts/FunctionalTests/ts/WebSocketTests.ts
@@ -4,9 +4,6 @@
 import { ECHOENDPOINT_URL } from "./Common";
 import "./LogBannerReporter";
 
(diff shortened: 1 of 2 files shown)


JamesNK (approved):

mikaelm12 (approved):

BrennanConroy (approved):

anurse:
I'm OK taking this in preview 5, merging now to make sure it lands before branching.

PR #9593, opened 2019-04-20
Skip a flaky test

davidfowl:
It keeps failing, I dunno if it's a product issue but skipping it for now.

### DIFF
--- a/src/DataProtection/Extensions/test/DataProtectionProviderTests.cs
+++ b/src/DataProtection/Extensions/test/DataProtectionProviderTests.cs
@@ -114,8 +114,7 @@ namespace Microsoft.AspNetCore.DataProtection
             });
         }
 
-        [ConditionalFact]
-        [Flaky("https://github.com/aspnet/AspNetCore-Internal/issues/2177", FlakyOn.AzP.Windows)]
+        [ConditionalFact(Skip = "https://github.com/aspnet/AspNetCore-Internal/issues/2177")]
         [X509StoreIsAvailable(StoreName.My, StoreLocation.CurrentUser)]
         [SkipOnHelix] // https://github.com/aspnet/AspNetCore/issues/6720
         public void System_UsesProvidedDirectoryAndCertificate()


natemcmaster:
Failing where? CI or locally?

src/DataProtection/Extensions/test/DataProtectionProviderTests.cs
```diff
@@ -114,8 +114,7 @@ public void System_UsesProvidedDirectory_WithConfigurationCallback()
             });
         }
 
-        [ConditionalFact]
-        [Flaky("https://github.com/aspnet/AspNetCore-Internal/issues/2177", FlakyOn.AzP.Windows)]
```
natemcmaster: This attribute is meant to quarantine the test while still giving us runs of the test so we can identify just exactly _how_ flaky something is. It shouldn't be causing build failures. I recommend we leave as is since this change would completely disable the test.

davidfowl: It’s failing PR checks. Does that mean the attribute isn’t working?

jkotalik: https://dev.azure.com/dnceng/public/_build/results?buildId=164123 seems like the correct behavior. The test is failing on the run, so it reports that the test run succeeded with issues. The build isn't failing.

jkotalik: Do you have an example where there is a red X?

davidfowl: Let me see if I can dig up one of the changes. It saw it fail at least twice this weekend.

anurse: > It saw it fail at least twice this weekend.

The test may fail, but shouldn't produce a failing build (unless **other** non-quarantined tests _also_ failed). If you find a case (**in `master`**) where this test was the one that failed the build, that would be a bug in `Flaky`.

davidfowl: What I saw was a red PR check not green but 2 tests were marked as failed. This one and some other test. If this test fails when marked as flaky and another test fails and it isn't marked flaky, will it show up in the checks output?

anurse: I don't know if it distinguishes clearly which test comes from which run. The Tests pane in AzDO will show **both** quarantined and unquarantined tests.

anurse: You can always fall back to looking at the build log, checking if the main run failed and grepping for `[FAIL]`.

davidfowl: OK so maybe it was the other test. I'll wait for the next one.

Issue #9594, opened 2019-04-20
Improve performance of ASP.NET Core with logging on 


davidfowl:
With the default settings and a hello world middleware, we go from 0 allocations to the following with logging turned on at the "Warning" level:

![image](https://user-images.githubusercontent.com/95136/56454076-9572f680-6300-11e9-985c-5dbbac1cd6cf.png)

PR #9595, opened 2019-04-20
Don't fire DiagnosticSource.StopActivity if there's no DiagnosticListener

davidfowl:
- Don't fire the event if DiagnosticSource.StartActivity was never called. This avoids allocating the anonymous object and extra strings.

Contributes to #9594

### DIFF
--- a/src/Hosting/Hosting/test/HostingApplicationTests.cs
+++ b/src/Hosting/Hosting/test/HostingApplicationTests.cs
@@ -57,6 +57,51 @@ namespace Microsoft.AspNetCore.Hosting.Tests
         }
 
         [Fact]
+        public void ActivityStopDoesNotFireIfNoListenerAttachedForStart()
+        {
+            // Arrange
+            var diagnosticSource = new DiagnosticListener("DummySource");
+            var logger = new LoggerWithScopes(isEnabled: true);
+            var hostingApplication = CreateApplication(out var features, diagnosticSource: diagnosticSource, logger: logger);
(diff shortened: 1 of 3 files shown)


src/Hosting/Hosting/test/HostingApplicationTests.cs
```diff
@@ -56,6 +56,51 @@ public void CreateContextWithEnabledLoggerCreatesActivityAndSetsActivityIdInScop
...
+            var diagnosticSource = new DiagnosticListener("DummySource");
+            var logger = new LoggerWithScopes(isEnabled: true);
+            var hostingApplication = CreateApplication(out var features, diagnosticSource: diagnosticSource, logger: logger);
+            var startFired = false;
+            var stopFired = false;
+
+            diagnosticSource.Subscribe(new CallbackDiagnosticListener(pair =>
+            {
+                // This should not fire
+                if (pair.Key == "Microsoft.AspNetCore.Hosting.HttpRequestIn.Start")
```
BrennanConroy: Not a fan of non-shared strings. Since we return false from the predicate below, no events will fire so we could just set a single bool in the subscribe func and check that?

BrennanConroy (approved):

src/Hosting/Hosting/test/HostingApplicationTests.cs
davidfowl: I initially had that but I wanted it to be specific.

PR #9599, opened 2019-04-20
Native build fails if repo is in a path containing spaces (e.g. C:\Users\Bob

zippy1981:
I checked out the source code to `C:\Users\Justin Dearing\source\repos\aspnet\AspNetCore` on my laptop and ran `build.cmd` and I got errors on the vcxproj builds. I traced it down to this.

### DIFF
--- a/src/Servers/IIS/build/native.targets
+++ b/src/Servers/IIS/build/native.targets
@@ -7,7 +7,7 @@
   <ItemGroup Condition="@(MessageFile->Count()) != 0">
     <CustomBuild Include="@(MessageFile)">
       <FileType>Document</FileType>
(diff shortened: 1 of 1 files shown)


zippy1981:
There is something else broken later in the build process related to spaces regarding dotnet.exe calling the DLL in `src\Servers\IIS\IIS\test\testassets\TestTasks`. I can fix that next, but should that be a separate PR. I'm actually trying to hack on managed code now so I'm skipping the full build and moving on to the PR I actually want to do.

jkotalik (approved):

jkotalik:
I'll merge this after we branch for preview5. Thanks for finding this!

Issue #9274, opened 2019-04-11
Problem with RedirectWwwRule


Bartmax:
### Is your feature request related to a problem? Please describe.
When using [`RedirectWwwRule`](https://github.com/aspnet/AspNetCore/blob/master/src/Middleware/Rewrite/src/Internal/RedirectToWwwRule.cs) it redirects everything that doesn't start with `www.`
When I create a site on azure for, for example: `example.com` (using the `RedirectWwwRule`) and trying to setup some staging so I can test my site on : `example-staging.azurewebsites.net` i get redirected to `www.example-staging.azurewebsites.net` which makes it impossible to use on this particular scenario. Also testing in `example.azurewebsites.net` turns into `www.example.azurewebsites.net`
*This is not limited to azure.*

### Describe the solution you'd like
Since most of the usages for `RedirectWwwRule` would be for naked domains, it would be great if we have it to work only on naked domains (no subdomain urls). Since this is a breaking change, we may add an option to the rule like to opt-in on naked only redirects.

### Describe alternatives you've considered
Currently an alternative is to make my own Rule which renders the existing Rule completely obsolete.

jkotalik:
Seems like an interesting idea. Have you tried to write your own custom rule to implement this?

Bartmax:
Detecting a naked domain is not a trivial thing (and maybe that's why the rule does not try to do it)

I found 2 workarounds that I'm using based on my use-case scenario.

1) Wrap the code inside a check on production environment.
```c#
if (env.IsProduction()) {
// do de rewrite here
}
```

2) I created a custom IRule that accepts a nake domain string value and I check if the host is naked on this fashion and if it is, I'm applying the redirect. Nothing too fancy
```c#
public class RedirectNakedToWwwRule : IRule
    {
        public readonly string _domain;
        public readonly int _statusCode;

        public RedirectNakedToWwwRule(string domain, int statusCode)
        {
            _domain = domain;
            _statusCode = statusCode;
        }

        public virtual void ApplyRule(RewriteContext context)
        {
            var req = context.HttpContext.Request;

            if (req.Host.Value.Equals(_domain, StringComparison.OrdinalIgnoreCase))
            {
                var wwwHost = new HostString($"www.{_domain}");
                var newUrl = UriHelper.BuildAbsolute(req.Scheme, wwwHost, req.PathBase, req.Path, req.QueryString);
                var response = context.HttpContext.Response;
                response.StatusCode = _statusCode;
[...]

PR #9601, opened 2019-04-20
add option to limit domains on AddRedirectToWww

delebru:
@jkotalik: This is the idea I had in mind, let me know if you have any thoughts.

Addresses #9274

### DIFF
--- a/src/Middleware/Rewrite/src/Internal/RedirectToWwwRule.cs
+++ b/src/Middleware/Rewrite/src/Internal/RedirectToWwwRule.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Linq;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Extensions;
 using Microsoft.AspNetCore.Rewrite.Logging;
@@ -12,9 +13,20 @@ namespace Microsoft.AspNetCore.Rewrite.Internal
     public class RedirectToWwwRule : IRule
     {
         public readonly int _statusCode;
+        public string[] _domains = new string[0];
 
         public RedirectToWwwRule(int statusCode)
         {
+            _statusCode = statusCode;
+        }
+        
+        public RedirectToWwwRule(int statusCode, params string[] domains)
+        {
+            if (domains == null)
+            {
+                throw new ArgumentNullException(nameof(domains));
+            }
+            _domains = domains;
             _statusCode = statusCode;
         }
 
@@ -28,6 +40,12 @@ namespace Microsoft.AspNetCore.Rewrite.Internal
             }
 
             if (req.Host.Value.StartsWith("www.", StringComparison.OrdinalIgnoreCase))
+            {
+                context.Result = RuleResult.ContinueRules;
+                return;
+            }
+            
+            if (_domains.Length > 0 && !_domains.Any(d => req.Host.Host.Equals(d, StringComparison.OrdinalIgnoreCase)))
             {
                 context.Result = RuleResult.ContinueRules;
                 return;
--- a/src/Middleware/Rewrite/src/Internal/RedirectToWwwRule.cs
+++ b/src/Middleware/Rewrite/src/Internal/RedirectToWwwRule.cs
@@ -2,7 +2,6 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Linq;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Extensions;
 using Microsoft.AspNetCore.Rewrite.Logging;
@@ -26,6 +25,10 @@ namespace Microsoft.AspNetCore.Rewrite.Internal
             {
                 throw new ArgumentNullException(nameof(domains));
             }
+            if (domains.Length < 1)
+            {
+                throw new ArgumentException(nameof(domains));
+            }
             _domains = domains;
             _statusCode = statusCode;
         }
@@ -45,7 +48,16 @@ namespace Microsoft.AspNetCore.Rewrite.Internal
                 return;
             }
             
-            if (_domains.Length > 0 && !_domains.Any(d => req.Host.Host.Equals(d, StringComparison.OrdinalIgnoreCase)))
+            var domainRequiresRedirection = false;
+            foreach (var domain in _domains)
+            {
+                if (domain.Equals(req.Host.Host, StringComparison.OrdinalIgnoreCase))
+                {
+                    domainRequiresRedirection = true;
+                    break;
+                }
+            }
+            if (!domainRequiresRedirection)
             {
                 context.Result = RuleResult.ContinueRules;
                 return;
--- a/src/Middleware/Rewrite/src/Internal/RedirectToWwwRule.cs
+++ b/src/Middleware/Rewrite/src/Internal/RedirectToWwwRule.cs
@@ -48,16 +48,16 @@ namespace Microsoft.AspNetCore.Rewrite.Internal
                 return;
             }
             
-            var domainRequiresRedirection = false;
+            var requestRequiresRedirection = false;
             foreach (var domain in _domains)
             {
                 if (domain.Equals(req.Host.Host, StringComparison.OrdinalIgnoreCase))
                 {
-                    domainRequiresRedirection = true;
+                    requestRequiresRedirection = true;
                     break;
                 }
             }
-            if (!domainRequiresRedirection)
+            if (!requestRequiresRedirection)
             {
                 context.Result = RuleResult.ContinueRules;
                 return;
--- a/src/Middleware/Rewrite/src/RewriteOptionsExtensions.cs
+++ b/src/Middleware/Rewrite/src/RewriteOptionsExtensions.cs
@@ -129,6 +129,17 @@ namespace Microsoft.AspNetCore.Rewrite
         }
 
         /// <summary>
+        /// Permanently redirects the request to the www subdomain if the request is non-www.
+        /// </summary>
+        /// <param name="options">The <see cref="RewriteOptions"/>.</param>
+        /// <param name="domains">List of domains on which the www. rule will be applied.</param>
+        /// <returns></returns>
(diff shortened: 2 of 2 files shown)


jkotalik:
This should target the master branch. I don't think we would consider adding a feature to a patch like this.

src/Middleware/Rewrite/src/Internal/RedirectToWwwRule.cs
```diff
@@ -32,6 +44,12 @@ public virtual void ApplyRule(RewriteContext context)
                 context.Result = RuleResult.ContinueRules;
                 return;
             }
+            
+            if (_domains.Length > 0 && !_domains.Any(d => req.Host.Host.Equals(d, StringComparison.OrdinalIgnoreCase)))
```
jkotalik: Few performance ideas:
- I'd throw in the constructor if domains.Length == 0. That way, you will always have at least one domain.
- Using LINQ here seems inefficient, it's probably more performant to just have a foreach loop through each domain.

src/Middleware/Rewrite/src/Internal/RedirectToWwwRule.cs
```diff
@@ -12,11 +13,22 @@ namespace Microsoft.AspNetCore.Rewrite.Internal
     public class RedirectToWwwRule : IRule
     {
         public readonly int _statusCode;
+        public string[] _domains = new string[0];
```
jkotalik: You don't need to set this to a default value if you take my suggestion on throwing if the domains is an empty array.

src/Middleware/Rewrite/src/Internal/RedirectToWwwRule.cs
delebru: Thanks for the review. I just pushed a commit adding the throw and replacing linq with a foreach. For further improvements we could replace:
```foreach (var domain in _domains)
{
    if (domain.Equals(req.Host.Host, StringComparison.OrdinalIgnoreCase))
    {
        domainRequiresRedirection = true;
        break;
    }
}
```
with:
```for (var i = 0; i < _domains.Length; i++)
{
    if (_domains[i].ReferenceEquals(req.Host.Host) || _domains[i].Equals(req.Host.Host, StringComparison.OrdinalIgnoreCase))
    {
        domainRequiresRedirection = true;
        break;
    }
}
```
Let me know what you'd like to see that change commited.

src/Middleware/Rewrite/src/Internal/RedirectToWwwRule.cs
delebru: I initialized the domains array with a default value in case the original constructor is called where we can't throw.

If that's a "deal breaker" we could try a different approach by adding another boolean that defines if the rule should apply to all domains. We can set that boolean to true on the original constructor and use it to skip the check on the domains array when applying the rule: 

```
public class RedirectToWwwRule : IRule
{
    public readonly int _statusCode;
    public readonly string[] _domains;
    public readonly bool _applyToAllDomains;

    public RedirectToWwwRule(int statusCode)
    {
        _statusCode = statusCode;
        _applyToAllDomains = true;
    }
        
    public RedirectToWwwRule(int statusCode, params string[] domains)
    {
        if (domains == null || domains.Length == 0)
        {
            throw new ArgumentNullException(nameof(domains));
        }
        if (domains.Length < 1)
        {
            throw new ArgumentException(nameof(domains));
        }
        _domains = domains;
        _applyToAllDomains = false;
        _statusCode = statusCode;
    }

    public virtual void ApplyRule(RewriteContext context)
    {
        var req = context.HttpContext.Request;
        if (req.Host.Host.Equals("localhost", StringComparison.OrdinalIgnoreCase))
        {
            context.Result = RuleResult.ContinueRules;
            return;
        }

        if (req.Host.Value.StartsWith("www.", StringComparison.OrdinalIgnoreCase))
        {
            context.Result = RuleResult.ContinueRules;
            return;
        }

        if (!_applyToAllDomains)
        {
            var domainRequiresRedirection = false;
            foreach (var domain in _domains)
            {
                if (domain.Equals(req.Host.Host, StringComparison.OrdinalIgnoreCase))
                {
                    domainRequiresRedirection = true;
                    break;
                }
            }
            if (!domainRequiresRedirection)
            {
                context.Result = RuleResult.ContinueRules;
                return;
            }
        }

        var wwwHost = new HostString($"www.{req.Host.Value}");
        var newUrl = UriHelper.BuildAbsolute(req.Scheme, wwwHost, req.PathBase, req.Path, req.QueryString);
        var response = context.HttpContext.Response;
        response.StatusCode = _statusCode;
        response.Headers[HeaderNames.Location] = newUrl;
        context.Result = RuleResult.EndResponse;
        context.Logger?.RedirectedToWww();
    }
}
```

delebru:
> This should target the master branch. I don't think we would consider adding a feature to a patch like this.

That's my bad, didn't intend for this to be used as a patch, just a mistake on the PR target branch. I'll open a new PR to master once we have a preferred solution from the previous comments.

src/Middleware/Rewrite/src/Internal/RedirectToWwwRule.cs
```diff
@@ -32,6 +47,21 @@ public virtual void ApplyRule(RewriteContext context)
...
+            var requestRequiresRedirection = false;
+            foreach (var domain in _domains)
+            {
+                if (domain.Equals(req.Host.Host, StringComparison.OrdinalIgnoreCase))
+                {
+                    requestRequiresRedirection = true;
+                    break;
+                }
+            }
+            if (!requestRequiresRedirection)
```
jkotalik: nit: newline.

src/Middleware/Rewrite/src/Internal/RedirectToWwwRule.cs
jkotalik: Do that except with a null check instead:

```c#
public class RedirectToWwwRule : IRule
{
    public readonly int _statusCode;
    public readonly string[] _domains;

    public RedirectToWwwRule(int statusCode)
    {
        _statusCode = statusCode;
    }

    public RedirectToWwwRule(int statusCode, params string[] domains)
    {
        if (domains == null)
        {
            throw new ArgumentNullException(nameof(domains));
        }
        _domains = domains;
        _statusCode = statusCode;
    }

    public virtual void ApplyRule(RewriteContext context)
    {
        var req = context.HttpContext.Request;
        if (req.Host.Host.Equals("localhost", StringComparison.OrdinalIgnoreCase))
        {
            context.Result = RuleResult.ContinueRules;
            return;
        }

        if (req.Host.Value.StartsWith("www.", StringComparison.OrdinalIgnoreCase))
        {
            context.Result = RuleResult.ContinueRules;
            return;
        }

//      if (_domains != null)
//      {
            var requestRequiresRedirection = false;
            foreach (var domain in _domains)
            {
                if (domain.Equals(req.Host.Host, StringComparison.OrdinalIgnoreCase))
                {
                    requestRequiresRedirection = true;
                    break;
                }
            }
            if (!requestRequiresRedirection)
            {
                context.Result = RuleResult.ContinueRules;
                return;
            }
//      }

        var wwwHost = new HostString($"www.{req.Host.Value}");
        var newUrl = UriHelper.BuildAbsolute(req.Scheme, wwwHost, req.PathBase, req.Path, req.QueryString);
        var response = context.HttpContext.Response;
        response.StatusCode = _statusCode;
        response.Headers[HeaderNames.Location] = newUrl;
        context.Result = RuleResult.EndResponse;
        context.Logger?.RedirectedToWww();
    }
}
```

src/Middleware/Rewrite/src/Internal/RedirectToWwwRule.cs
```diff
@@ -12,11 +12,26 @@ namespace Microsoft.AspNetCore.Rewrite.Internal
...
+        public RedirectToWwwRule(int statusCode, params string[] domains)
+        {
+            if (domains == null)
+            {
+                throw new ArgumentNullException(nameof(domains));
+            }
+            if (domains.Length < 1)
+            {
+                throw new ArgumentException(nameof(domains));
+            }
```
jkotalik: nit: newline

src/Middleware/Rewrite/src/Internal/RedirectToWwwRule.cs
jkotalik: I don't think _domains[i].ReferenceEquals(req.Host.Host) would ever be true, as the domain you are passing into the rule wouldn't be the same string passed into the WwwRule.

src/Middleware/Rewrite/src/RewriteOptionsExtensions.cs
```diff
@@ -147,5 +168,17 @@ public static RewriteOptions AddRedirectToWww(this RewriteOptions options, int s
...
+        /// <summary>
+        /// Redirect the request to the www subdomain if the incoming request is non-www.
+        /// </summary>
+        /// <param name="options">The <see cref="RewriteOptions"/>.</param>
+        /// <param name="statusCode">The status code to add to the response.</param>
+        /// <param name="domains">List of domains on which the www. rule will be applied.</param>
+        public static RewriteOptions AddRedirectToWww(this RewriteOptions options, int statusCode, params string[] domains)
+        {
+            options.Rules.Add(new RedirectToWwwRule(statusCode, domains));
+            return options;
```
jkotalik: Make sure to add some tests!

Bartmax:
> I don't think _domains[i].ReferenceEquals(req.Host.Host) would ever be true, as the domain you are passing into the rule wouldn't be the same string passed into the WwwRule.

Not sure I follow,
`ReferenceEquals("string", "string")  // returns true`
I see huge improvement (10x) using ReferenceEquals

src/Middleware/Rewrite/src/Internal/RedirectToWwwRule.cs
delebru: Oh you're right! The strings will on most cases come from different instances.

src/Middleware/Rewrite/src/Internal/RedirectToWwwRule.cs
delebru: Yes! How didn't I see that option... Thanks, I love this solution :)

delebru:
I'm closing this PR now. I'll add the tests and all the suggestions from the latest reviews and open a new PR to master tonight. Thanks!

PR #9618, opened 2019-04-21
Cache some Headers lookups

benaadams:
`context.Request.Headers` goes via feature reference lookups; so once its been resolved, use that for further use rather than resolving again.


jkotalik (approved):

jkotalik:
@anurse this is low risk so going to merge for preview5.

PR #9622, opened 2019-04-21
Avoid allocating the CandidateSet when there's a single match

davidfowl:
- Add fast path for 0 route values, and policies

Scenario, hello world plaintext (the /plaintext sample in the RoutingSandbox) hit with 10K requests.

Allocations before:

image

Allocations after:

image

Microbenchmarks:

MatcherSingleEntryBenchmark

Before:

```
           Method |     Mean |     Error |    StdDev |        Op/s | Scaled | ScaledSD |  Gen 0 | Allocated |
----------------- |---------:|----------:|----------:|------------:|-------:|---------:|-------:|----------:|
         Baseline | 106.5 ns |  1.762 ns |  1.472 ns | 9,387,641.5 |   1.00 |     0.00 | 0.0074 |      40 B |
              Dfa | 185.6 ns |  3.724 ns |  8.556 ns | 5,388,922.3 |   1.74 |     0.08 | 0.0212 |     112 B |
 LegacyTreeRouter | 954.0 ns | 20.355 ns | 38.231 ns | 1,048,254.6 |   8.96 |     0.37 | 0.0458 |     408 B |
     LegacyRouter | 711.2 ns | 13.520 ns | 11.985 ns | 1,406,107.4 |   6.68 |     0.14 | 0.0525 |     280 B |
```

After:

```
           Method |      Mean |      Error |     StdDev |         Op/s | Scaled | ScaledSD |  Gen 0 | Allocated |
----------------- |----------:|-----------:|-----------:|-------------:|-------:|---------:|-------:|----------:|
         Baseline |  96.00 ns |  0.8115 ns |  0.7193 ns | 10,416,201.9 |   1.00 |     0.00 | 0.0076 |      40 B |
              Dfa |  97.07 ns |  1.0586 ns |  0.9902 ns | 10,301,416.6 |   1.01 |     0.01 |      - |       0 B |
 LegacyTreeRouter | 913.16 ns | 17.9336 ns | 17.6132 ns |  1,095,100.5 |   9.51 |     0.19 | 0.0772 |     408 B |
     LegacyRouter | 679.49 ns | 10.7855 ns | 10.0888 ns |  1,471,691.7 |   7.08 |     0.11 | 0.0534 |     280 B |
```

### DIFF
--- a/src/Http/Routing/src/EndpointSelectorContext.cs
+++ b/src/Http/Routing/src/EndpointSelectorContext.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
@@ -24,7 +24,7 @@ namespace Microsoft.AspNetCore.Routing
         /// </summary>
         public RouteValueDictionary RouteValues
         {
-            get => _routeValues;
+            get => _routeValues ?? (_routeValues = new RouteValueDictionary());
             set
             {
                 _routeValues = value;
--- a/src/Http/Routing/src/Matching/DfaMatcher.cs
+++ b/src/Http/Routing/src/Matching/DfaMatcher.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
@@ -53,7 +53,8 @@ namespace Microsoft.AspNetCore.Routing.Matching
             // FindCandidateSet will process the DFA and return a candidate set. This does
             // some preliminary matching of the URL (mostly the literal segments).
             var (candidates, policies) = FindCandidateSet(httpContext, path, segments);
-            if (candidates.Length == 0)
+            var candidateCount = candidates.Length;
+            if (candidateCount == 0)
             {
                 if (log)
                 {
@@ -66,6 +67,26 @@ namespace Microsoft.AspNetCore.Routing.Matching
             if (log)
             {
                 Logger.CandidatesFound(_logger, path, candidates);
+            }
+
+            var policyCount = policies.Length;
+
+            // This is a fast path for single candidate, 0 policies and default selector
+            if (policyCount == 0 && candidateCount == 1 && _selector is DefaultEndpointSelector)
+            {
+                ref var candidate = ref candidates[0];
+                var flags = candidate.Flags;
+
+                // Just strict path matching
+                if (flags == Candidate.CandidateFlags.None)
+                {
+                    // TODO: Remove this allocation
+                    context.RouteValues = new RouteValueDictionary();
+                    context.Endpoint = candidate.Endpoint;
+
+                    // We're done
+                    return Task.CompletedTask;
+                }
             }
 
             // At this point we have a candidate set, defined as a list of endpoints in
@@ -83,7 +104,7 @@ namespace Microsoft.AspNetCore.Routing.Matching
             // `candidateSet` is the mutable state that we pass to the EndpointSelector.
             var candidateSet = new CandidateSet(candidates);
 
-            for (var i = 0; i < candidates.Length; i++)
+            for (var i = 0; i < candidateCount; i++)
             {
                 // PERF: using ref here to avoid copying around big structs.
                 //
@@ -165,7 +186,7 @@ namespace Microsoft.AspNetCore.Routing.Matching
                 }
             }
 
-            if (policies.Length == 0)
+            if (policyCount == 0)
             {
                 // Perf: avoid a state machine if there are no polices
                 return _selector.SelectAsync(httpContext, context, candidateSet);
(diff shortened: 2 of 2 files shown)


src/Http/Routing/src/Matching/DfaMatcher.cs
```diff
@@ -68,6 +69,26 @@ public sealed override Task MatchAsync(HttpContext httpContext, EndpointSelector
                 Logger.CandidatesFound(_logger, path, candidates);
             }
 
+            var policyCount = policies.Length;
+
+            // This is a fast path for single candidate, 0 policies and default selector
+            if (policyCount == 0 && candidateCount == 1 && _selector is DefaultEndpointSelector)
+            {
+                ref var candidate = ref candidates[0];
+                var flags = candidate.Flags;
```
JamesNK: The flags are only used once. Is there any benefit from a local variable?

src/Http/Routing/src/Matching/DfaMatcher.cs
```diff
@@ -68,6 +69,26 @@ public sealed override Task MatchAsync(HttpContext httpContext, EndpointSelector
                 Logger.CandidatesFound(_logger, path, candidates);
             }
 
+            var policyCount = policies.Length;
+
+            // This is a fast path for single candidate, 0 policies and default selector
+            if (policyCount == 0 && candidateCount == 1 && _selector is DefaultEndpointSelector)
```
JamesNK: How fast is `is`? Is it worth doing this check once in the constructor and setting a bool on DfaMatcher to check here?

src/Http/Routing/src/Matching/DfaMatcher.cs
davidfowl: Prolly not.

src/Http/Routing/src/Matching/DfaMatcher.cs
davidfowl: Good idea.

rynowak: yeah it's probably more consistent to make this a flag on candidates if possible.

src/Http/Routing/src/EndpointSelectorContext.cs
```diff
@@ -24,7 +24,7 @@ public sealed class EndpointSelectorContext : IEndpointFeature, IRouteValuesFeat
         /// </summary>
         public RouteValueDictionary RouteValues
         {
-            get => _routeValues;
+            get => _routeValues ?? (_routeValues = new RouteValueDictionary());
```
rynowak: If we're going to do this, we can take advantage of it in more cases, including the ones that involve cnadidate set.

src/Http/Routing/src/Matching/DfaMatcher.cs
rynowak: eh, might not actually be a good fit for flags.

rynowak:
####?s

src/Http/Routing/src/EndpointSelectorContext.cs
JamesNK: This is beyond the scope of this PR, but what about reusing RVD instances? It would be similar to how we're reusing HttpContext et al. It's basically a per-request allocations that is tied to a feature.

Reusing RVD, and the routing feature, would save allocations on every request, not just those with no route values.

[Caching per-request objects on reused HttpContexts](https://github.com/aspnet/AspNetCore/issues/6895)

davidfowl: We can just straight up do this in Kestrel I fear we won't be able to do the more generic caching for 3.0

davidfowl: > If we're going to do this, we can take advantage of it in more cases, including the ones that involve cnadidate set.

Yes, we can, I just wanted to minimize the code change for now. I have to look at see what things are assuming route values are never null when looking at CandidateState.RouteValues (maybe I can do it statiaclly with #enable nullable 😄 )

davidfowl:
> ####?s

Will get some. I went awaty but now I'm back

src/Http/Routing/src/EndpointSelectorContext.cs
rynowak: CandidateSet has some special semantics around RVD that should just be removed. If treats null RVD as a shorthand for not-a-match in some APIs.

src/Http/Routing/src/Matching/DfaMatcher.cs
```diff
@@ -16,13 +16,15 @@ internal sealed class DfaMatcher : Matcher
...
         private readonly int _maxSegmentCount;
+        private readonly bool _isDefaultEndpointSelector;
 
         public DfaMatcher(ILogger<DfaMatcher> logger, EndpointSelector selector, DfaState[] states, int maxSegmentCount)
         {
             _logger = logger;
             _selector = selector;
             _states = states;
             _maxSegmentCount = maxSegmentCount;
+            _isDefaultEndpointSelector = selector is DefaultEndpointSelector;
```
rynowak: I'd feel slightly better if this was an exact type equality check. This isn't public, but that would capture the intent better.

rynowak (approved):

src/Http/Routing/src/Matching/DfaMatcher.cs
davidfowl: Isn’t it sealed? Future proofing?

JamesNK (approved):

PR #9628, opened 2019-04-22
Spanify path -> string logic

davidfowl:
- Use spans instead of byte[] and array segments when turning the URL into a string


jkotalik (approved):

Tratcher:
Hmm, how did you modify a bunch of shared Http.Sys code without needing to react to it in HttpSys?

davidfowl:
byte[] has an implicit conversion to `Span<byte>`

davidfowl:
@Tratcher looked at the callsite in HttpSys, it's because of var. Either that or I'm missing something. Both get reduced allocations for free 😄

PR #9646, opened 2019-04-22
Use correct path to x86 installers to ANCM

jkotalik:
Right now, the x86 msis for ANCM contain the wrong bitness for dlls. This addresses it.

### DIFF
--- a/src/Installers/Windows/AspNetCoreModule-Setup/ANCMIISExpressV2/ancm_iis_expressv2.wxs
+++ b/src/Installers/Windows/AspNetCoreModule-Setup/ANCMIISExpressV2/ancm_iis_expressv2.wxs
@@ -162,12 +162,21 @@
         <Directory Id="INSTALLLOCATION" ShortName="ANCM" Name="$(var.ProductName)">
           <Directory Id="VersionDir" Name="$(var.ProductVersionString)">
             <Component Id="AspNetCoreModule" Guid="84ed6ce6-c8a3-4fa8-a872-c98a1d15dd4f" Win64="$(var.IsWin64)">
-              <File Id="AspNetCoreModuleDll"
-                    Name="aspnetcorev2.dll"
-                    Source="$(var.AspNetCoreV2.TargetPath)"
-                    DiskId="1"
-                    Vital="yes">
-              </File>
+                <?if $(var.Platform) = "x64" ?>
+                    <File Id="AspNetCoreModuleDll"
+                        Name="aspnetcorev2.dll"
+                        Source="$(var.AspNetCoreV2.TargetPath)"
+                        DiskId="1"
+                        Vital="yes">
+                    </File>
+                <?else ?>
+                    <File Id="AspNetCoreModuleDll"
+                        Name="aspnetcorev2.dll"
+                        Source="$(var.AspNetCoreV2WoW64.TargetPath)"
+                        DiskId="1"
+                        Vital="yes">
+                    </File>
+                <?endif?>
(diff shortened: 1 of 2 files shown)


src/Installers/Windows/AspNetCoreModule-Setup/ANCMIISExpressV2/ancm_iis_expressv2.wxs
```diff
@@ -162,25 +162,43 @@
...
+                    <File Id="AspNetCoreModuleDll"
+                        Name="aspnetcorev2.dll"
+                        Source="$(var.AspNetCoreV2.TargetPath)"
+                        DiskId="1"
+                        Vital="yes">
+                    </File>
+                <?else ?>
+                    <File Id="AspNetCoreModuleDll"
+                        Name="aspnetcorev2.dll"
+                        Source="$(var.AspNetCoreV2WoW64.TargetPath)"
```
joeloff: Why don't you just define a single variable in the .wixproj or pass it through using -d, then you don't need to duplicate the File elements.

src/Installers/Windows/AspNetCoreModule-Setup/ANCMIISExpressV2/ancm_iis_expressv2.wxs
```diff
@@ -162,25 +162,43 @@
         <Directory Id="INSTALLLOCATION" ShortName="ANCM" Name="$(var.ProductName)">
           <Directory Id="VersionDir" Name="$(var.ProductVersionString)">
             <Component Id="AspNetCoreModule" Guid="84ed6ce6-c8a3-4fa8-a872-c98a1d15dd4f" Win64="$(var.IsWin64)">
```
joeloff: Is there a reason why the component GUIDs aren't generated and why you have the same GUIDs for both 64-bit and 32-bit components?

jkotalik: You can't install the 64 and 32 bit msis on the same machine, so we should be okay in that scenario.

Why GUIDs aren't generated is a different issue that should probably be resolved.

src/Installers/Windows/AspNetCoreModule-Setup/ANCMIISExpressV2/ancm_iis_expressv2.wxs
jkotalik: Right now these are defined via project references here: https://github.com/aspnet/AspNetCore/blob/master/src/Installers/Windows/AspNetCoreModule-Setup/Directory.Build.props#L45. In the wixproj file, can I just use $(AspNetCoreV2.TargetPath)?

jkotalik: I tried a few things and nothing worked, ideas?

joeloff: if you have a .wixproj, you can do something like <DefineConstants>$(DefineConstants);Foo=$(SomePath)</DefineConstants>

Then in your .wixproj, you'll be able to access $(var.Foo). If you conditionally set $(SomePath) on the architecture you'll have one constant defined for both bitness

jkotalik: Went with an approach where I define the constants in the wxs file instead (as the variables I wanted were defined in DefineConstants).

joeloff (approved):

jkotalik:
Verified this too.

PR #9649, opened 2019-04-22
Fix race in dotnet-watch test

BrennanConroy:
Possibly makes the `RunsWithIterationEnvVariable` test not flaky anymore.

The race is that we wait for the "Waiting for a file to change" log and then set the file time, but if we set the time before the `fileSetWatcher.OnFileChange` handler is registered then we could miss the file changed event.

### DIFF
--- a/src/Tools/dotnet-watch/src/DotNetWatcher.cs
+++ b/src/Tools/dotnet-watch/src/DotNetWatcher.cs
@@ -93,8 +93,6 @@ namespace Microsoft.DotNet.Watcher
 
                     if (finishedTask == processTask)
                     {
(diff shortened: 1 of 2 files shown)


natemcmaster (dismissed):
Can we find an alternative solution? Moving this output line results in the app always showing this message at the beginning of calling `dotnet watch`. Doesn't seem like the right behavior

image

BrennanConroy:
Sure, would you be ok with adding another overload like `public async Task<string> GetChangedFileAsync(CancellationToken cancellationToken, Action startedWatching)`?

natemcmaster:
Yeah, that would be fine.

BrennanConroy:
Ping

natemcmaster (approved):

PR #9652, opened 2019-04-22
Remove redundant IEnumerable interface from ArrayRange<T>

SteveSandersonMS:
No urgency on this since it won't affect shipping code.

`ArrayRange<T>` declared itself as `IEnumerable`/`IEnumerable<T>`, but this was only used by test code. It's a good job it wasn't used in product code, since as @davidfowl noticed, it allocated twice on every call.

To ensure we don't accidentally start using the allocatey methods in production code, this PR removes those interfaces. There quite a lot of noise making the tests continue to work, but it's all simple mechanical changes.

**Reviewers**: The easiest way to read this PR is each commit separately.

### DIFF
--- a/src/Components/Blazor/Build/test/ComponentRenderingRazorIntegrationTest.cs
+++ b/src/Components/Blazor/Build/test/ComponentRenderingRazorIntegrationTest.cs
@@ -350,7 +350,7 @@ namespace Test
             // Assert: Captured ChildContent frames are correct
             var childFrames = GetFrames((RenderFragment)frames[2].AttributeValue);
             Assert.Collection(
-                childFrames,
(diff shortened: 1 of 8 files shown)


rynowak (approved):

davidfowl:
Ref assemblies

javiercn (approved):
Looks good, but as fowler points out, update ref assemblies.

I wish updating ref assemblies was something we could automate with a bot or similar.

SteveSandersonMS:
I did update the ref assemblies. The error was because the version of the tool I ran locally (from the repo) disagreed with the version in CI about the correct ordering for enum entries. 

To work around, I’ve manually avoided committing the changes to enum entry ordering, and now the build passed.

PR #9662, opened 2019-04-22
Use correct path to x86 installers to ANCM (#9646)

jkotalik:
#### Description 
Installing the 32-bit version of the Windows Hosting Bundle on a 32-bit OS installs the 64-bit version of the ASP.NET Core Module instead of the 32-bit version. This is due to some MSI authoring problems. This issue was introduced with some refactoring to the MSI authoring done in January. We don't have sufficient verification for installing the 32-bit Hosting Bundle and deploying to that server. We'll be adding verification scenarios to explicitly ensure 32-bit installation is tested moving forward.
#### Customer Impact 
 Installing the 32-bit version of the 3.0 Windows Hosting Bundle will cause existing ASP.NET Core apps (including 2.x apps) to break.
#### Regression? 
 Yes, introduced in January (in earlier 3.0 previews)
#### Risk 
 Low, the updated MSI has been validated in a 32-bit OS and drops the correct DLLs. 2.x and 3.x apps have been tested after installing this fixed MSI and function correctly.


anurse:
Note @aspnet/build we are in the process of getting QB approval.

anurse:
This is Approved for check-in by the QB. Merge when ready.

analogrelay (approved):

PR #9688, opened 2019-04-23
Update reference assemblies WIP

JunTaoLuo:
Trying to address: https://github.com/aspnet/AspNetCore-Internal/issues/2372.

Updated reference assemblies. EDIT: seems like running `dotnet msbuild /t:GenerateReferenceSource` on macOS doesn't work.


JunTaoLuo:
WIP, triggering for tests.

src/Antiforgery/ref/Microsoft.AspNetCore.Antiforgery.csproj
```diff
@@ -1,9 +1,10 @@
+
 <!-- This file is automatically generated. -->
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFrameworks>netcoreapp3.0</TargetFrameworks>
   </PropertyGroup>
-  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp3.0'">
+  <ItemGroup Condition=" == netcoreapp3.0">
```
JunTaoLuo: This looks super wrong. Not sure what the script is doing here.

natemcmaster (approved):

natemcmaster:
Huh, this is strange. CodeCheck.ps1 failed on the PR check.

image

JunTaoLuo:
Yea, and it's the exact same error as the original build: https://dnceng.visualstudio.com/public/_build/results?buildId=164916. So code check doesn't like either order?!

natemcmaster:
I think there is a legit bug in GenAPI. Investigating...

natemcmaster:
See https://github.com/dotnet/arcade/issues/2607#issuecomment-485949756. GenAPI produces different results when running on `dotnet msbuild` vs `msbuild.exe`. We'll need a new version of GenAPI to fix this.

JunTaoLuo:
Going to wait for the fix in arcade.

PR #9690, opened 2019-04-23
Cleanup unnecessary files added during 2.2 => master merge

natemcmaster:
When we merge 2.x to master, we don't need to bring these files forward. They are not used in the master branch.

Also, added more info to the error message about regenerating ref code. This message was already in the aspnet/Extensions version of CodeCheck.ps1


dougbu (approved):
Whatever 😺

PR #9706, opened 2019-04-24
Make ClientHandler constructor internal

JunTaoLuo:
Addresses: https://github.com/aspnet/AspNetCore/issues/9532. I'm assuming we're okay with this breaking change?

### DIFF
--- a/src/Hosting/TestHost/ref/Microsoft.AspNetCore.TestHost.netcoreapp3.0.cs
+++ b/src/Hosting/TestHost/ref/Microsoft.AspNetCore.TestHost.netcoreapp3.0.cs
@@ -5,7 +5,7 @@ namespace Microsoft.AspNetCore.TestHost
 {
     public partial class ClientHandler : System.Net.Http.HttpMessageHandler
     {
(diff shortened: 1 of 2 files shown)


src/Hosting/TestHost/ref/Microsoft.AspNetCore.TestHost.netcoreapp3.0.cs
```diff
@@ -5,7 +5,7 @@ namespace Microsoft.AspNetCore.TestHost
 {
     public partial class ClientHandler : System.Net.Http.HttpMessageHandler
     {
-        public ClientHandler(Microsoft.AspNetCore.Http.PathString pathBase, Microsoft.AspNetCore.Hosting.Server.IHttpApplication<Microsoft.AspNetCore.Hosting.Internal.HostingApplication.Context> application) { }
+        internal ClientHandler() { }
```
Tratcher: Why is this still listed in the ref assembly?

JunTaoLuo: I was curious as well... especially since the constructor parameters doesn't even match. But that's what the tool did.

Tratcher: It seems to do it for any non-static type without a public constructor. That makes sense, the default constructor is public unless otherwise stated.

Tratcher (approved):

PR #9708, opened 2019-04-24
Make RouteOptions.EndpointDataSources internal

JamesNK:
I noticed RouteOptions.EndpointDataSources was only set in DI and the setter is internal.

Return a collection if not set for use outside of DI, e.g. unit testing.

### DIFF
--- a/src/Http/Routing/src/RouteOptions.cs
+++ b/src/Http/Routing/src/RouteOptions.cs
@@ -3,13 +3,25 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using Microsoft.AspNetCore.Routing.Constraints;
 
 namespace Microsoft.AspNetCore.Routing
 {
     public class RouteOptions
     {
-        public ICollection<EndpointDataSource> EndpointDataSources { get; internal set; }
+        /// <summary>
+        /// Gets a collection of <see cref="EndpointDataSource"/> instances configured with routing.
+        /// </summary>
+        internal ICollection<EndpointDataSource> EndpointDataSources
+        {
+            get
+            {
+                Debug.Assert(_endpointDataSources != null, "Endpoint data sources should have been set in DI.");
+                return _endpointDataSources;
+            }
+            set => _endpointDataSources = value;
+        }
 
         /// <summary>
         /// Gets or sets a value indicating whether all generated paths URLs are lower-case. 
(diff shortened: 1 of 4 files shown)


src/Http/Routing/src/RouteOptions.cs
```diff
@@ -9,7 +9,22 @@ namespace Microsoft.AspNetCore.Routing
 {
     public class RouteOptions
     {
-        public ICollection<EndpointDataSource> EndpointDataSources { get; internal set; }
+        /// <summary>
+        /// Gets a collection of <see cref="EndpointDataSource"/> instances configured with routing.
+        /// </summary>
+        public ICollection<EndpointDataSource> EndpointDataSources
```
rynowak: Does it make sense to think about different fixes for this? (or even making this internal/separate).

I'm not sure that *fixing* this behaviour will fix a real problem. I think if you try to use the `UseEndpoints` and this you're going to have a bad time regardless.

I'm not totally certain that we have a use case for users to interact with this property.

JamesNK: UseRouting won't use these, but anyone else integrating with endpoints will, e.g. link generation.

If we don't make this modifiable then UseEndpoints would be the only way to specify a data source.

rynowak: > If you want to write to this property, that will be useful because ??????

We don't have any scenarios or documentation that use this. It literally has not come up once. I'm pretty lukewarm about public API for things that we wouldn't tell people to do.

JamesNK: Fair enough. It was added to replace registering endpoint data sources via DI which I loathed, but we have UseEndpoints now.

I'll make it internal and see whether any problems come up.

JamesNK:
🆙 📅

src/Http/Routing/src/RouteOptions.cs
```diff
@@ -49,6 +61,7 @@ public class RouteOptions
         public bool SuppressCheckForUnhandledSecurityMetadata { get; set; }
 
         private IDictionary<string, Type> _constraintTypeMap = GetDefaultConstraintMap();
+        private ICollection<EndpointDataSource> _endpointDataSources;
```
rynowak: Can you just move these fields to the top? This isn't the way our guidelines be.

rynowak (approved):

PR #9709, opened 2019-04-24
Speed up PathStringHelper

benaadams:
Called by `PathString.ToString()` in `HostingLogScope` as top contributor

image

As they are per `char` methods they multiply up significantly

image

With change moves to lowest contributor

image

Contributes to https://github.com/aspnet/AspNetCore/issues/9594

/cc @davidfowl

### DIFF
--- a/src/Http/Http.Abstractions/src/Internal/PathStringHelper.cs
+++ b/src/Http/Http.Abstractions/src/Internal/PathStringHelper.cs
@@ -53,7 +53,8 @@ namespace Microsoft.AspNetCore.Http.Internal
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static bool IsPercentEncodedChar(string str, int index)
         {
-            if (str[index] == '%' && index < str.Length - 2)
+            var len = (uint)str.Length;
+            if (str[index] == '%' && index < len - 2)
             {
                 return AreFollowingTwoCharsHex(str, index);
             }
--- a/src/Http/Http.Abstractions/src/Internal/PathStringHelper.cs
+++ b/src/Http/Http.Abstractions/src/Internal/PathStringHelper.cs
@@ -1,11 +1,14 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
+
 namespace Microsoft.AspNetCore.Http.Internal
 {
     internal class PathStringHelper
     {
-        private static bool[] ValidPathChars = {
+        private static readonly bool[] ValidPathChars = {
             false, false, false, false, false, false, false, false,     // 0x00 - 0x07
             false, false, false, false, false, false, false, false,     // 0x08 - 0x0F
             false, false, false, false, false, false, false, false,     // 0x10 - 0x17
@@ -24,24 +27,57 @@ namespace Microsoft.AspNetCore.Http.Internal
             true,  true,  true,  false, false, false, true,  false,     // 0x78 - 0x7F
         };
 
+        // Map from an ASCII char to its hex value, e.g. arr['b'] == 11. 0xFF means it's not a hex digit.
+        // From coreclr/src/System.Private.CoreLib/shared/System/Number.Parsing.cs
+        // byte[] rather than ReadOnlySpan<byte> to avoid double .Length check https://github.com/dotnet/coreclr/issues/24209
+        private static readonly byte[] CharToHexLookup = new byte[]
+        {
+            0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, // 15
+            0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, // 31
+            0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, // 47
+            0x0,  0x1,  0x2,  0x3,  0x4,  0x5,  0x6,  0x7,  0x8,  0x9,  0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, // 63
+            0xFF, 0xA,  0xB,  0xC,  0xD,  0xE,  0xF,  0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, // 79
+            0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, // 95
+            0xFF, 0xa,  0xb,  0xc,  0xd,  0xe,  0xf // 102
+        };
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static bool IsValidPathChar(char c)
         {
-            return c < ValidPathChars.Length && ValidPathChars[c];
+            // Use local array and uint .Length compare to elide the bounds check on array access
+            var validChars = ValidPathChars;
+            var i = (int)c;
+            return (uint)i < (uint)validChars.Length && validChars[i];
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static bool IsPercentEncodedChar(string str, int index)
         {
-            return index < str.Length - 2
-                && str[index] == '%'
-                && IsHexadecimalChar(str[index + 1])
-                && IsHexadecimalChar(str[index + 2]);
+            if (str[index] == '%' && index < str.Length - 2)
+            {
+                return AreFollowingTwoCharsHex(str, index);
+            }
+
+            return false;
         }
 
-        public static bool IsHexadecimalChar(char c)
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private static bool AreFollowingTwoCharsHex(string str, int index)
         {
-            return ('0' <= c && c <= '9')
-                || ('A' <= c && c <= 'F')
-                || ('a' <= c && c <= 'f');
+            Debug.Assert(index < str.Length - 2);
+
+            var charToHex = CharToHexLookup;
+            var i1 = (int)str[index + 1];
+            var i2 = (int)str[index + 2];
+            if ((uint)i1 < (uint)charToHex.Length &&
+                charToHex[i1] != 0xff &&
+               (uint)i2 < (uint)charToHex.Length &&
+                charToHex[i2] != 0xff)
+            {
+                return true;
+            }
+
+            return false;
         }
     }
 }
--- a/src/Http/Http.Abstractions/src/Internal/PathStringHelper.cs
+++ b/src/Http/Http.Abstractions/src/Internal/PathStringHelper.cs
@@ -6,7 +6,7 @@ using System.Runtime.CompilerServices;
 
 namespace Microsoft.AspNetCore.Http.Internal
 {
-    internal class PathStringHelper
+    internal static class PathStringHelper
     {
         private static readonly bool[] ValidPathChars = {
             false, false, false, false, false, false, false, false,     // 0x00 - 0x07
@@ -70,12 +70,11 @@ namespace Microsoft.AspNetCore.Http.Internal
             var charToHex = CharToHexLookup;
             var i1 = (int)str[index + 1];
             var i2 = (int)str[index + 2];
-            if ((uint)i1 < (uint)charToHex.Length &&
-                charToHex[i1] != 0xff &&
-               (uint)i2 < (uint)charToHex.Length &&
-                charToHex[i2] != 0xff)
+            if ((uint)i1 < (uint)charToHex.Length
+                && (uint)i2 < (uint)charToHex.Length)
             {
-                return true;
+                var tmp = charToHex[i1] | charToHex[i2];
+                return tmp != 0xFF;
             }
 
             return false;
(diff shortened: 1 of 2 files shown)


src/Http/Http.Abstractions/src/Internal/PathStringHelper.cs
```diff
@@ -24,24 +27,57 @@ internal class PathStringHelper
...
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static bool IsPercentEncodedChar(string str, int index)
         {
-            return index < str.Length - 2
-                && str[index] == '%'
-                && IsHexadecimalChar(str[index + 1])
-                && IsHexadecimalChar(str[index + 2]);
+            if (str[index] == '%' && index < str.Length - 2)
```
gfoidl:
```suggestion
            var len = (uint)str.Length;
            if (str[index] == '%' && index < len - 2)
```
Reuses the str-length, instead of reloading.
```diff
G_M56184_IG02:
+	   8B4708               mov      eax, dword ptr [rdi+8]
+      8BD0                 mov      edx, eax
+      3BF0                 cmp      esi, eax
-      3B7708               cmp      esi, dword ptr [rdi+8]
       7320                 jae      SHORT G_M56184_IG05
       4863C6               movsxd   rax, esi
       66837C470C25         cmp      word  ptr [rdi+2*rax+12], 37
       7511                 jne      SHORT G_M56184_IG03
+      4863C6               movsxd   rax, esi	   
-      8B4708               mov      eax, dword ptr [rdi+8]
       83C0FE               add      eax, -2
+      483BC2               cmp      rax, rdx	   
-      3BC6                 cmp      eax, esi
       7E07                 jle      SHORT G_M56184_IG03
       E824F4FFFF           call     PathStringHelper:AreFollowingTwoCharsHex(ref,int):bool
       EB02                 jmp      SHORT G_M56184_IG04
```

src/Http/Http.Abstractions/src/Internal/PathStringHelper.cs
```diff
@@ -24,24 +27,57 @@ internal class PathStringHelper
             true,  true,  true,  false, false, false, true,  false,     // 0x78 - 0x7F
         };
 
+        // Map from an ASCII char to its hex value, e.g. arr['b'] == 11. 0xFF means it's not a hex digit.
+        // From coreclr/src/System.Private.CoreLib/shared/System/Number.Parsing.cs
+        // byte[] rather than ReadOnlySpan<byte> to avoid double .Length check https://github.com/dotnet/coreclr/issues/24209
+        private static readonly byte[] CharToHexLookup = new byte[]
```
gfoidl: You can reduce the size of the lut if only lowercase is handled.
And for making the char lowercase just set the 5th bit `c |= (char)(1 << 5)` (or use 0x20 directly).
No check for `A-Z` should be needed, as all other value will result in `0xFF`.

So the lut may fit into a cache line.

Same can be applied for `ValidPathChars`.

src/Http/Http.Abstractions/src/Internal/PathStringHelper.cs
```diff
@@ -24,24 +27,57 @@ internal class PathStringHelper
...
+        private static bool AreFollowingTwoCharsHex(string str, int index)
         {
-            return ('0' <= c && c <= '9')
-                || ('A' <= c && c <= 'F')
-                || ('a' <= c && c <= 'f');
+            Debug.Assert(index < str.Length - 2);
+
+            var charToHex = CharToHexLookup;
+            var i1 = (int)str[index + 1];
+            var i2 = (int)str[index + 2];
```
gfoidl: Unfortunately one bound check isn't elided when flipped, so that `[index + 2]` comes first.
Is string total different than a regular (sz) array?

src/Http/Http.Abstractions/src/Internal/PathStringHelper.cs
```diff
@@ -24,24 +27,57 @@ internal class PathStringHelper
...
+            if ((uint)i1 < (uint)charToHex.Length &&
+                charToHex[i1] != 0xff &&
+               (uint)i2 < (uint)charToHex.Length &&
+                charToHex[i2] != 0xff)
+            {
+                return true;
```
gfoidl: Maybe the following is better?
```c#
if ((uint)i1 < (uint)charToHex.Length
 && (uint)i2 < (uint)charToHex.Length)
{
    var tmp = charToHex[i1] | charToHex[i2];
    return tmp != 0xFF;
}

return false;
```

```diff
       3BF2                 cmp      esi, edx
       7624                 jbe      SHORT G_M42979_IG04
-      4863D2               movsxd   rdx, edx
-      807C1010FF           cmp      byte  ptr [rax+rdx+16], 255
-      7415                 je       SHORT G_M42979_IG04
-      3BF7                 cmp      esi, edi
-      7611                 jbe      SHORT G_M42979_IG04
-      4863FF               movsxd   rdi, edi
-      807C3810FF           cmp      byte  ptr [rax+rdi+16], 255
-      7407                 je       SHORT G_M42979_IG04
-      B801000000           mov      eax, 1
+      3BF7                 cmp      esi, edi
+      7620                 jbe      SHORT G_M42979_IG04
+      4863F2               movsxd   rsi, edx
+      400FB6743010         movzx    rsi, byte  ptr [rax+rsi+16]
+      4863FF               movsxd   rdi, edi
+      0FB6443810           movzx    rax, byte  ptr [rax+rdi+16]
+      0BC6                 or       eax, esi
+      3DFF000000           cmp      eax, 255
+      0F95C0               setne    al
+      0FB6C0               movzx    rax, al

G_M42979_IG03:
       5D                   pop      rbp
       C3                   ret 
```

Safes two branches.

src/Http/Http.Abstractions/src/Internal/PathStringHelper.cs
benaadams: Its a tricky one; you have to guarantee that the indexer is both positive and less than length.

If `index` was `-2` then `[index + 2]` would work but `[index + 1]` would fail so accessing using a variable plus offset doesn't give that guarantee; unlike accessing with consts.

benaadams: Also you have to do it without adding additional checks, as that's just adding the bounds checks back manually.

src/Http/Http.Abstractions/src/Internal/PathStringHelper.cs
benaadams: Done

src/Http/Http.Abstractions/src/Internal/PathStringHelper.cs
gfoidl: > unlike accessing with consts.

This was the point I missed (mentally). JIT does this only on const indexes. Thanks.

src/Http/Http.Abstractions/src/Internal/PathStringHelper.cs
gfoidl: An idea...

The lut could be ommited, and the following code used
```c#
private static bool AreFollowingTwoCharsHex(string str, int index)
{
    Debug.Assert(index < str.Length - 2);

    int i1 = str[index + 1];
    int i2 = str[index + 2];

    return IsHexadecimalChar(i1) && IsHexadecimalChar(i2);
}

private static bool IsHexadecimalChar(int c)
{
    return (byte)(c - 48) <= 9 || (byte)((c & ~0x20) - 65) <= 5;
}
```
`IsHexadecimalChar` was there before, but this version is better optimized.

<details>
   <summary>codegen</summary>

```asm
; Assembly listing for method PathStringHelper:AreFollowingTwoCharsHex(ref,int):bool
; Emitting BLENDED_CODE for X64 CPU with AVX - Unix
; optimized code
; rbp based frame
; partially interruptible
; Final local variable assignments
;
;  V00 arg0         [V00,T00] (  5,  5   )     ref  ->  rdi         class-hnd
;  V01 arg1         [V01,T01] (  4,  4   )     int  ->  rsi
;  V02 loc0         [V02,T06] (  3,  2   )     int  ->  rdi
;# V03 OutArgs      [V03    ] (  1,  1   )  lclBlk ( 0) [rsp+0x00]   "OutgoingArgSpace"
;  V04 tmp1         [V04,T02] (  3,  5   )     int  ->  rax         "impAppendStmt"
;  V05 tmp2         [V05,T07] (  2,  1   )    bool  ->  rax         "Inline return value spill temp"
;  V06 tmp3         [V06,T08] (  3,  1.50)    bool  ->  rax         "Inline return value spill temp"
;  V07 cse0         [V07,T03] (  3,  3   )     int  ->  rax         "ValNumCSE"
;  V08 cse1         [V08,T04] (  3,  3   )     int  ->  rdx         "ValNumCSE"
;  V09 cse2         [V09,T05] (  3,  3   )     int  ->  rsi         "ValNumCSE"
;
; Lcl frame size = 0

G_M2516_IG01:
       55                   push     rbp
       488BEC               mov      rbp, rsp
       90                   nop

G_M2516_IG02:
       8D4601               lea      eax, [rsi+1]
       8B5708               mov      edx, dword ptr [rdi+8]
       3BC2                 cmp      eax, edx
       7364                 jae      SHORT G_M2516_IG08
       4863C0               movsxd   rax, eax
       0FB744470C           movzx    rax, word  ptr [rdi+2*rax+12]
       83C602               add      esi, 2
       3BF2                 cmp      esi, edx
       7355                 jae      SHORT G_M2516_IG08
       4863F6               movsxd   rsi, esi
       0FB77C770C           movzx    rdi, word  ptr [rdi+2*rsi+12]
       8D70D0               lea      esi, [rax-48]
       400FB6F6             movzx    rsi, sil
       83FE09               cmp      esi, 9
       7E16                 jle      SHORT G_M2516_IG03
       83E0DF               and      eax, -33
       83C0BF               add      eax, -65
       0FB6C0               movzx    rax, al
       83F805               cmp      eax, 5
       0F9EC0               setle    al
       0FB6C0               movzx    rax, al
       85C0                 test     eax, eax
       7427                 je       SHORT G_M2516_IG06

G_M2516_IG03:
       8D47D0               lea      eax, [rdi-48]
       0FB6C0               movzx    rax, al
       83F809               cmp      eax, 9
       7E15                 jle      SHORT G_M2516_IG04
       83E7DF               and      edi, -33
       83C7BF               add      edi, -65
       400FB6C7             movzx    rax, dil
       83F805               cmp      eax, 5
       0F9EC0               setle    al
       0FB6C0               movzx    rax, al
       EB05                 jmp      SHORT G_M2516_IG05

G_M2516_IG04:
       B801000000           mov      eax, 1

G_M2516_IG05:
       5D                   pop      rbp
       C3                   ret

G_M2516_IG06:
       33C0                 xor      eax, eax

G_M2516_IG07:
       5D                   pop      rbp
       C3                   ret

G_M2516_IG08:
       E81899B878           call     CORINFO_HELP_RNGCHKFAIL
       CC                   int3

; Total bytes of code 121, prolog size 5 for method PathStringHelper:AreFollowingTwoCharsHex(ref,int):bool
; ============================================================
```
</details>

But I don't know if this is better than the lut-variant. It depends on the input-string and if the lut is already in the cache, etc. Benchmark-results may vary.

So far I couldn't come up with a way where `i1` and `i2` are checked as "packed chars" (i.e. `(i1 << 16) | i2`) at once.

benaadams: Thank you, resolved.

src/Http/Http.Abstractions/src/Internal/PathStringHelper.cs
```diff
@@ -1,47 +1,64 @@
...
 
-        public static bool IsHexadecimalChar(char c)
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static bool IsHexadecimalChar(char c)
         {
-            return ('0' <= c && c <= '9')
-                || ('A' <= c && c <= 'F')
-                || ('a' <= c && c <= 'f');
+            // Between 0 - 9 or uppercased between A - F
+            return (c - '0') <= 9 || ((c & ~0x20) - 'A') <= ('F' - 'A');
```
gfoidl: The casts to an unsigned is needed to correctly handle values <= '0' (now this will result in `true`, e.g. for `\0 - '0'` the first condition is met). Analogous for the seconds condition.

In the proposal I had `byte`, but this isn't enough to cover the whole char-range.
```suggestion
            return (uint)(c - '0') <= 9 || (uint)((c & ~0x20) - 'A') <= ('F' - 'A');
```

So the following test will pass:
```c#
[Fact]
public void IsHexadecimalChar_all_chars()
{
    for (char c = char.MinValue; c < (char)char.MaxValue; ++c)
    {
        bool expected = ('0' <= c && c <= '9') || ('A' <= c && c <= 'F') || ('a' <= c && c <= 'f');
        bool actual = IsHexadecimalChar(c);

        Assert.Equal(expected, actual);
    }
}
```

src/Http/Http.Abstractions/src/Internal/PathStringHelper.cs
```diff
@@ -1,47 +1,64 @@
...
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static bool IsValidPathChar(char c)
         {
-            return c < ValidPathChars.Length && ValidPathChars[c];
+            // Use local array and uint .Length compare to elide the bounds check on array access
+            var validChars = ValidPathChars;
+            var i = (int)c;
+
+            // Array is in chunks of 32 bits, so get offset by dividing by 32
+            var offset = i / 32;
```
gfoidl:
```suggestion
            var offset = i >> 5;		// i / 32;
```
AFAIK JIT does this optimization only for [unsigned](https://sharplab.io/#v2:EYLgHgbALANALiAhgZwLYB8ACAGABJgRgG4BYAKEwGZ8AmXAYVwG9zc39qBLAOzlwFkAFDz6IAlLgC8APlyJcAelyUapMuw64AriIGCdvORJlzFy1eQC+QA=)

gfoidl (approved):

davidfowl:
We don't have any microbenchmarks here do we.

src/Http/Http.Abstractions/src/Internal/PathStringHelper.cs
```diff
@@ -1,47 +1,67 @@
...
+            var i = (int)c;
+
+            // Array is in chunks of 32 bits, so get offset by dividing by 32
+            var offset = i >> 5;		// i / 32;
+            // Significant bit position is the remainder of the above calc; i % 32 => i & 31
+            var significantBit = 1u << (i & 31);
+
+            // Check offset in bounds and check if significant bit set
+            return (uint)offset < (uint)validChars.Length &&
+                ((validChars[offset] & significantBit) == significantBit);
```
benaadams: `!= 0` prob better here

davidfowl (approved):
Nice

benaadams:
Moves to lowest contributor

image

PR #9718, opened 2019-04-24
[Project templates] Sync logout between SPA and main app. Disable react service worker

javiercn:
Sync the logout between the SPA an Identity.
Disable react service worker when used with auth as it causes issues.

Addresses https://github.com/aspnet/AspNetCore/issues/9403 (the critical aspects)

### DIFF
--- a/src/ProjectTemplates/Web.Spa.ProjectTemplates/content/React-CSharp/ClientApp/src/components/api-authorization/AuthorizeService.js
+++ b/src/ProjectTemplates/Web.Spa.ProjectTemplates/content/React-CSharp/ClientApp/src/components/api-authorization/AuthorizeService.js
@@ -310,6 +310,10 @@ export class AuthorizeService {
         });
 
         this.userManager = new UserManager(settings);
+        this.userManager.events.addUserSignedOut(async () => {
+            await this.userManager.removeUser();
+            this.updateState(undefined);
+        });
     }
 
     static get instance() { return authService }
--- a/src/ProjectTemplates/Web.Spa.ProjectTemplates/content/React-CSharp/ClientApp/src/index.js
+++ b/src/ProjectTemplates/Web.Spa.ProjectTemplates/content/React-CSharp/ClientApp/src/index.js
@@ -29,7 +29,7 @@ ReactDOM.render(
 //// Uncomment the lines below to register the generated service worker.
 //// By default create-react-app includes a service worker to improve the
 //// performance of the application by caching static assets. This service
-//// worker, can however try and handle paths for the identity UI, so it is
+//// worker can interfere with the Identity UI, so it is
 //// disabled by default when Identity is being used.
 ////
 ////registerServiceWorker();
--- a/src/ProjectTemplates/Web.Spa.ProjectTemplates/content/React-CSharp/ClientApp/src/index.js
+++ b/src/ProjectTemplates/Web.Spa.ProjectTemplates/content/React-CSharp/ClientApp/src/index.js
@@ -4,13 +4,13 @@ import ReactDOM from 'react-dom';
 import { BrowserRouter } from 'react-router-dom';
 import App from './App';
 ////#if (IndividualLocalAuth)
-// Uncomment the lines below to register the generated service worker.
-// By default create-react-app includes a service worker to improve the
-// performance of the application by caching static assets. This service
-// worker, can however try and handle paths for the identity UI, so it is
-// disabled by default when Identity is being used.
-//
-//import registerServiceWorker from './registerServiceWorker';
+//// Uncomment the lines below to register the generated service worker.
+//// By default create-react-app includes a service worker to improve the
+//// performance of the application by caching static assets. This service
+//// worker, can however try and handle paths for the identity UI, so it is
+//// disabled by default when Identity is being used.
+////
+////import registerServiceWorker from './registerServiceWorker';
 ////#else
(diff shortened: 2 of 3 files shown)


src/ProjectTemplates/Web.Spa.ProjectTemplates/content/React-CSharp/ClientApp/src/index.js
```diff
@@ -14,4 +25,14 @@ ReactDOM.render(
   </BrowserRouter>,
   rootElement);
 
+////#if (IndividualLocalAuth)
+//// Uncomment the lines below to register the generated service worker.
+//// By default create-react-app includes a service worker to improve the
+//// performance of the application by caching static assets. This service
+//// worker, can however try and handle paths for the identity UI, so it is
```
SteveSandersonMS:
```suggestion
//// worker can interfere with the Identity UI, so it is
```

src/ProjectTemplates/Web.Spa.ProjectTemplates/content/React-CSharp/ClientApp/src/index.js
```diff
@@ -3,7 +3,18 @@ import React from 'react';
...
 import { BrowserRouter } from 'react-router-dom';
 import App from './App';
+////#if (IndividualLocalAuth)
+//// Uncomment the lines below to register the generated service worker.
+//// By default create-react-app includes a service worker to improve the
+//// performance of the application by caching static assets. This service
+//// worker, can however try and handle paths for the identity UI, so it is
+//// disabled by default when Identity is being used.
+////
+////import registerServiceWorker from './registerServiceWorker';
```
SteveSandersonMS: Do we definitely need to include a repeat of this comment block, and the conditional `import` here?

If it's possible just to *always* have the `import registerServiceWorker` line, and then only toggle whether it gets used, I think that would be better.

If you're forced to do this as-is due to ESLint rules or something, I'd consider suppressing that rule here.

src/ProjectTemplates/Web.Spa.ProjectTemplates/content/React-CSharp/ClientApp/src/components/api-authorization/AuthorizeService.js
```diff
@@ -310,6 +310,10 @@ export class AuthorizeService {
         });
 
         this.userManager = new UserManager(settings);
+        this.userManager.events.addUserSignedOut(async () => {
+            await this.userManager.removeUser();
+            this.updateState(undefined);
```
SteveSandersonMS: I don't really know oidc-client well enough to know what this really does (e.g., `this.userManager.removeUser`), so I can only assume it's correct and has the desired behaviour. Same with the Angular version.

src/ProjectTemplates/Web.Spa.ProjectTemplates/content/React-CSharp/ClientApp/src/index.js
javiercn: It's because of linting. We could not include the commented out import and just include the comment in the lower line and tell people to uncomment the import too in the comment?

SteveSandersonMS: Yes, that sounds better than duplicating the comment.

src/ProjectTemplates/Web.Spa.ProjectTemplates/content/React-CSharp/ClientApp/src/components/api-authorization/AuthorizeService.js
javiercn: Yeah, that just clears up the state from the local user cache. And then we trigger an update to the user. That makes it so that when you log out from Identity, the SPA also gets notified and updates its state.

SteveSandersonMS (approved):
The oidc-client APIs are outside my knowledge, but 👍 to the rest.

src/ProjectTemplates/Web.Spa.ProjectTemplates/content/React-CSharp/ClientApp/src/index.js
javiercn: Done

ryanbrandenburg (approved):
Some questions:
- The effects of not registering a ServiceWorker are only 1. reduced performace due to no/less cachine and 2. Worse/No offline behavior.
- Is there a plan to come back to this later and make it work with IdentityUI? It seems like there ought to be some way to invalidate the cache (which I assume was the problem) on logout/login.

That said, :shipit: for now to make the baseline scenario better.

javiercn:
> * 2. Worse/No offline behavior.

Users can enable it and configure the service worker themselves, but we can't do it for you.

The problem is the worker intercepts the request to /Identity/Account/Register, so the path needs to be excluded, but CRA doesn't allow you to do that (yet) out of the box.

mkArtakMSFT:
@javiercn please update this PR to target the `master` branch (preview6). This has not been approved for preview5 release.

javiercn:
I'll do this tomorrow. I have a separate PR out for master in case this didn't meet the bar

javiercn:
Closing this PR as it was not approved for preview5.

javiercn:
PR for preview6 is here https://github.com/aspnet/AspNetCore/pull/9691

Issue #4597, opened 2018-12-12
.net Core 2.2 -  Generate RouteData from Url path


mgolois:
Hello,

I'm looking for documentation on the new endpoint routing system. I'm upgrading from .net core 2.1 to 2.2. I have a scenario where I need to generate the RouteData object from a url path.
I could not get it to work with IUrlHelper.

Thanks!

MarxITSolutions:
I am having the same issue after scaffolding the Identity. The callbackUrl always returns null. Page: Register.cshtml.cs                     
var callbackUrl = Url.Page(
                        "/Account/ConfirmEmail",
                        pageHandler: null,
                        values: new { userId = user.Id, code = code },
                        protocol: Request.Scheme);

rynowak:
@mgolois  - hi, sorry for the delayed response. It's not 100% clear to me what you're trying to do, so I'm guessing.

If what you want is to parse a URL into route values, you can use the `TemplateMatcher` for that https://github.com/aspnet/AspNetCore/blob/master/src/Http/Routing/src/Template/TemplateMatcher.cs#L10

I'm closing this issue since I believe I answered the question. Feel free to open another issue is this is not what you need.

-------

@MarxITSolutions  - if you're still having this problem could you please open a new issue and include a minimal repro of the problem so that we can investigate? Thanks!

mgolois:
@rynowak , in order to use TemplateMatcher I need to get the teamples, so all route collections.
With the new route system this no longer works:
`httpcontext.GetRouteData().Routers.OfType<IRouteCollection>().List();`

Basically, after upgrading to 2.2 I would like to use the new routing system, the application has a method that contains a list of strings (urlPaths) and needs to generate RouteData of each urlPaths.
eg:
    /Students/GradeBook/Index: area=Students, controller=GradeBook, action=Index
   /Home/Index: controller=Home, action=Index

oferns:
I also would like to do this. How can we get a collection of routers in the new route system?

oferns:
Ok. So i dont know about @rynowak but I use conventional route patterns. I know everyone loves attribute routing but I don't see the appeal but I believe this approach could work with that as well. What I am doing now is after registering the routes in the UseMvc extension method like this 
   
```
 app.UseMvc(routes => {
                routes.MapRoute(
                    name: "Default",
                    template: "{action}",
                    defaults: new { controller = "Home", action = "Index" }
                );

```
[...]

PR #9728, opened 2019-04-24
Fixes: #4597 Parse URI path with an endpoint

rynowak:
Adds functionality to LinkParser to parse a URI path given a way to
find an endpoint. This is the replacement for various machinications
using the global route collection and `RouteData.Routers` in earlier
versions.

For now I'm just adding a way to do this using Endpoint Name since it's
a pretty low level feature. Endpoint Name is also very direct, so it
feels good for something like this.

Example usage:

```C#
[HttpPut]
[Route("/Products/{id}", Name = "ProductsRoute")]
public async Task<IActionResult> AddRelatedProduct(int projectId, string path)
{
   var values = parser.ParsePathByEndpointName("ProductsRoute", path);
   var relatedId = (string)values["id"];

   // Do something with this information
}
```

### DIFF
--- a/src/Http/Routing/src/DefaultLinkParser.cs
+++ b/src/Http/Routing/src/DefaultLinkParser.cs
@@ -1,0 +1,236 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.Routing
+{
+    internal class DefaultLinkParser : LinkParser, IDisposable
+    {
+        private readonly ParameterPolicyFactory _parameterPolicyFactory;
+        private readonly ILogger<DefaultLinkParser> _logger;
+        private readonly IServiceProvider _serviceProvider;
+
+        // Caches RoutePatternMatcher instances
+        private readonly DataSourceDependentCache<ConcurrentDictionary<RouteEndpoint, MatcherState>> _matcherCache;
+
+        // Used to initialize RoutePatternMatcher and constraint instances
+        private readonly Func<RouteEndpoint, MatcherState> _createMatcher;
+
+        public DefaultLinkParser(
+            ParameterPolicyFactory parameterPolicyFactory,
+            EndpointDataSource dataSource,
+            ILogger<DefaultLinkParser> logger,
+            IServiceProvider serviceProvider)
+        {
+            _parameterPolicyFactory = parameterPolicyFactory;
+            _logger = logger;
+            _serviceProvider = serviceProvider;
+
+            // We cache RoutePatternMatcher instances per-Endpoint for performance, but we want to wipe out
+            // that cache is the endpoints change so that we don't allow unbounded memory growth.
+            _matcherCache = new DataSourceDependentCache<ConcurrentDictionary<RouteEndpoint, MatcherState>>(dataSource, (_) =>
+            {
+                // We don't eagerly fill this cache because there's no real reason to. Unlike URL matching, we don't
+                // need to build a big data structure up front to be correct.
+                return new ConcurrentDictionary<RouteEndpoint, MatcherState>();
+            });
+
+            // Cached to avoid per-call allocation of a delegate on lookup.
+            _createMatcher = CreateRoutePatternMatcher;
+        }
+
+        public override RouteValueDictionary ParsePathByAddress<TAddress>(TAddress address, PathString path)
+        {
+            var endpoints = GetEndpoints(address);
+            if (endpoints.Count == 0)
+            {
+                return null;
+            }
+
+            for (var i = 0; i < endpoints.Count; i++)
+            {
+                var endpoint = endpoints[i];
+                if (TryParse(endpoint, path, out var values))
+                {
+                    Log.PathParsingSucceeded(_logger, path, endpoints);
+                    return values;
+                }
+            }
+
+            Log.PathParsingFailed(_logger, path, endpoints);
+            return null;
+        }
+
+        private List<RouteEndpoint> GetEndpoints<TAddress>(TAddress address)
+        {
+            var addressingScheme = _serviceProvider.GetRequiredService<IEndpointAddressScheme<TAddress>>();
+            var endpoints = addressingScheme.FindEndpoints(address).OfType<RouteEndpoint>().ToList();
+
+            if (endpoints.Count == 0)
+            {
+                Log.EndpointsNotFound(_logger, address);
+            }
+            else
+            {
+                Log.EndpointsFound(_logger, address, endpoints);
+            }
+
+            return endpoints;
+        }
+
+
+        private MatcherState CreateRoutePatternMatcher(RouteEndpoint endpoint)
+        {
+            var constraints = new Dictionary<string, List<IRouteConstraint>>(StringComparer.OrdinalIgnoreCase);
+
+            var policies = endpoint.RoutePattern.ParameterPolicies;
+            foreach (var kvp in policies)
+            {
+                var constraintsForParameter = new List<IRouteConstraint>();
+                var parameter = endpoint.RoutePattern.GetParameter(kvp.Key);
+                for (var i = 0; i < kvp.Value.Count; i++)
+                {
+                    var policy = _parameterPolicyFactory.Create(parameter, kvp.Value[i]);
+                    if (policy is IRouteConstraint constraint)
+                    {
+                        constraintsForParameter.Add(constraint);
+                    }
+                }
+
+                if (constraintsForParameter.Count > 0)
+                {
+                    constraints.Add(kvp.Key, constraintsForParameter);
+                }
+            }
+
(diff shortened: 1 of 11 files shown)


rynowak:
@JamesNK  looking for your COOL high level feedback on this before adding tests.

src/Http/Routing.Abstractions/src/LinkGenerator.cs
```diff
@@ -147,5 +148,32 @@ public abstract class LinkGenerator
             PathString pathBase = default,
             FragmentString fragment = default,
             LinkOptions options = default);
+
+        /// <summary>
+        /// Attempts to pass the provided <paramref name="path"/> using the route pattern
```
rynowak: should say parse :)

src/Http/Routing/src/DefaultLinkGenerator.cs
```diff
@@ -333,7 +388,7 @@ public static RouteValueDictionary GetAmbientValues(HttpContext httpContext)
 
         public void Dispose()
         {
-            _cache.Dispose();
+            _binderCache.Dispose();
```
rynowak: need to add my new cache here as well

rynowak:
This is also pretty simple and straightforward compared to what we have today 👍

rynowak:
There's a few more decisions here to be made like - does this apply route values? Does this run constraints? I'm planning to reach out to some of the folks that asked for this feature and poll them.

rynowak:
@oferns @mgolois - I'm looking at adding an API to address your feedback here: https://github.com/aspnet/AspNetCore/issues/4597

Could you take a look at this and let me know what you think?

src/Http/Routing/src/DefaultLinkGenerator.cs
```diff
@@ -280,13 +297,43 @@ private List<RouteEndpoint> GetEndpoints<TAddress>(TAddress address)
...
+            for (var i = 0; i < endpoints.Count; i++)
+            {
+                var endpoint = endpoints[i];
+                if (TryParse(endpoint, path, out var values))
+                {
+                    return values;
+                }
+            }
+
+            Log.LinkGenerationFailed(_logger, endpoints);
```
JamesNK: Wrong logging method

JamesNK (commented):
Seems alright.

It does feel a little odd to have matching on link generator. Consider whether a new type makes sense. `LinkParser`?

src/Http/Routing/src/DefaultLinkGenerator.cs
```diff
@@ -280,13 +297,43 @@ private List<RouteEndpoint> GetEndpoints<TAddress>(TAddress address)
             return null;
         }
 
+        public override RouteValueDictionary ParsePathByAddress<TAddress>(TAddress address, PathString path)
+        {
+            var endpoints = GetEndpoints(address);
+            if (endpoints.Count == 0)
```
oferns: How/where is GetEndpoints implemented? This was my issue, getting access to the routes/endpoints that used to be in RouteData.Routes

rynowak: Get endpoints is implemented on the base class. 

If you want a list of all of the endpoints in the app, you do do that by grabbing `EndpointDataSource` from DI. If you want the endpoint that matched the current request, use this.  https://github.com/aspnet/AspNetCore/blob/425c196cba530b161b120a57af8f1dd513b96f67/src/Http/Http.Abstractions/src/Routing/EndpointHttpContextExtensions.cs

rynowak:
@JamesNK - ready for review. I decided to go with a separate class.

src/Http/Routing/src/DefaultLinkParser.cs
```diff
@@ -0,0 +1,236 @@
...
+            }
+            else
+            {
+                Log.EndpointsFound(_logger, address, endpoints);
+            }
+
+            return endpoints;
+        }
+
+
```
JamesNK: too liney

src/Http/Routing/src/DefaultLinkParser.cs
```diff
@@ -0,0 +1,236 @@
...
+            public static class EventIds
+            {
+                public static readonly EventId EndpointsFound = new EventId(100, "EndpointsFound");
+                public static readonly EventId EndpointsNotFound = new EventId(101, "EndpointsNotFound");
+
+                public static readonly EventId PathParsingSucceeded = new EventId(102, "PathParsingSucceeded");
+                public static readonly EventId PathParsingFailed = new EventId(103, "PathParsingFailed");
+            }
+
+            private static readonly Action<ILogger, IEnumerable<string>, object, Exception> _endpointsFound = LoggerMessage.Define<IEnumerable<string>, object>(
```
JamesNK: I didn't know logging supported enumerable. Could you add a unit test that checks that the log message looks like what you expected when there are multiple matches.

Alternatively I think it is fine to just have a count in the log message.

src/Http/Routing/test/UnitTests/DefaultLinkParserTest.cs
```diff
@@ -0,0 +1,170 @@
...
+
+            // Act
+            var values = parser.ParsePathByAddress(0, "/");
+
+            // Assert
+            MatcherAssert.AssertRouteValuesEqual(new { controller = "Home", action = "Index", }, values);
+        }
+
+        [Fact]
+        public void ParsePathByAddresss_HasMatches_RunsConstraints()
```
JamesNK: 👍

src/Http/Routing/src/LinkParserEndpointNameAddressExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// <summary>
+    /// Extension methods for using <see cref="LinkParser"/> with an endpoint name.
+    /// </summary>
+    public static class LinkParserEndpointNameAddressExtensions
+    {
+        /// <summary>
+        /// Attempts to parse the provided <paramref name="path"/> using the route pattern
+        /// specified by the <see cref="Endpoint"/> matching <paramref name="endpointName"/>.
+        /// </summary>
+        /// <param name="parser">The <see cref="LinkGenerator"/>.</param>
```
JamesNK:
```suggestion
        /// <param name="parser">The <see cref="LinkParser"/>.</param>
```

src/Http/Routing/src/DefaultLinkParser.cs
```diff
@@ -0,0 +1,236 @@
...
+            {
+                return null;
+            }
+
+            for (var i = 0; i < endpoints.Count; i++)
+            {
+                var endpoint = endpoints[i];
+                if (TryParse(endpoint, path, out var values))
+                {
+                    Log.PathParsingSucceeded(_logger, path, endpoints);
```
JamesNK: Should this log message just mention the endpoint that was successful?

src/Http/Routing/src/DefaultLinkParser.cs
rynowak: I did know that  😆

src/Http/Routing/src/LinkParserEndpointNameAddressExtensions.cs
rynowak: Ooops!

rynowak:
@JamesNK updated

src/Http/Routing/src/DefaultLinkParser.cs
```diff
@@ -0,0 +1,235 @@
...
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.Routing
+{
+    internal class DefaultLinkParser : LinkParser, IDisposable
```
JamesNK: tis never registered with DI

Add an integration test that uses it

rynowak:
@JamesNK updated

src/Mvc/test/Mvc.FunctionalTests/LinkParserTest.cs
```diff
@@ -0,0 +1,96 @@
...
+                p =>
+                {
+                    Assert.Equal("action", p.Name);
+                    Assert.Equal("AnotherRoute", p.Value.Value<string>());
+                },
+                p =>
+                {
+                    Assert.Equal("controller", p.Name);
+                    Assert.Equal("LinkParser", p.Value.Value<string>());
+                },
```
rynowak: The fact that controller/action are included here (non-parameter defaults) might seem surprising at first glance but I think the alternative is WAYYYY more surprising. 

You really want this to round-trip the values from link generation and omitting non-parameter defaults would be strange and arbitrary

JamesNK (approved):

marono:
Will this make it into .NET Core 3.0? Thanks!

rynowak:
This is already in 3.0 preview 6 which will be the next preview release.

holytshirt:
Having been looking how to do this in 2.2 and just found this :/. 

Hurry up 3.0 and thank you!
This is a big missing piece when doing a proper ReST apis, as you don't post Ids, you post uris.

rynowak:
@holytshirt - thanks.

I'm surprised we took this long to figure this out! I've been doing work on routing now for a few years, and I saw a big outcry from users when I broke the existing way of doing this 😆 - now this is something proper that is in the docs.

gbrunton:
@rynowak  Where in the docs is this because I'm having a hard time find it?

rynowak:
It's mentioned here: https://docs.microsoft.com/en-us/aspnet/core/migration/22-to-30?view=aspnetcore-3.1&tabs=visual-studio#routing-startup-code

I've opened #16913 to track adding some more information.

PR #9740, opened 2019-04-25
Fix documentation typo

BrunoMVPCosta:
Looking at the Endpoint class [documentation](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.endpoint?view=aspnetcore-2.2), there is `Respresents` instead of `Represents`.

This PR fixes this documentation typo.


analogrelay (approved):

anurse:
Thanks for the typo fix!

Issue #9213, opened 2019-04-09
Consider exposing Bedrock's "Connection Features" on HttpContext


anurse:
I'm filing this to track some ideas that came up while @Tratcher was looking at Kerberos/NTLM auth. The exact requirements are still coming but I wanted to file this to start some parallel discussion on things that may help make the implementation smooth.

NTLM requires **mandatory** caching of security context information between requests occurring on the same connection. So, given a connection `C` and NTLM-authenticated requests `R1` and `R2` (with `R2` following after `R1`), it is **not possible** to authenticate `R2` without using cached data from the authentication process in `R1`.

Using current features, it is relatively simple to implement this by caching this data keyed off the [Connection ID](https://github.com/aspnet/AspNetCore/blob/4134d02dab44717d049d1e55abb473ec04ddf47d/src/Http/Http.Abstractions/src/ConnectionInfo.cs#L16). However this has a few problems:

1. It's a little clunky to have to maintain a separate cache when there is generally a connection state object in the server
1. It is difficult to reliably expire this cache unless the server exposes a "Connection Ended" event of some kind.

```csharp
public interface INtlmConnectionStateFeature
{
    // ... ntlmy data ...
}
```

The auth middleware can implement this feature entirely, however, it needs to be able to store it somewhere that is guaranteed to live across the entire connection. Bedrock's "Connection Features" is a perfect place for this, however it is not exposed up through the stack.

If we had a way to Get/Set connection-level features, we could implement this in the auth handler with pseudo-code like this:

```csharp
var currentState = context.Connection.Features.Get<INtlmConnectionStateFeature>();
if (currentState == null)
{
    currentState = new NtlmConnectionState();
    context.Connections.Features.Set(currentState);
}
PerformAuthentication(currentState);
```

My proposal is this:
1. Add a new feature to the **request features**: `IHttpConnectionFeaturesFeature` (name can be bikeshed later)

```csharp
public interface IHttpConnectionFeaturesFeature
{
    IFeatureCollection ConnectionFeatures { get; }
}
```

2. Add **read-only** `ConnectionFeatures` property to [`ConnectionInfo`](https://github.com/aspnet/AspNetCore/blob/4134d02dab44717d049d1e55abb473ec04ddf47d/src/Http/Http.Abstractions/src/ConnectionInfo.cs#L11) which returns `null` if there is no `IHttpConnectionFeaturesFeature` present.

[...]

PR #9754, opened 2019-04-25
Add IConnectionCompleteFeature.OnCompleted

Tratcher:
#9213 This is a copy of the HttpResponse.OnCompleted feature down to the connection level. It's intended for resource cleanup at the end of the connection lifecycle and will be used by KerberosAuth to eagerly Dispose state (https://github.com/aspnet/AspNetCore/issues/8897).

### DIFF
--- a/src/Servers/Kestrel/Core/test/ConnectionDispatcherTests.cs
+++ b/src/Servers/Kestrel/Core/test/ConnectionDispatcherTests.cs
@@ -7,6 +7,7 @@ using System.IO.Pipelines;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Connections.Features;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Internal;
 using Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.Internal;
 using Microsoft.AspNetCore.Testing;
@@ -68,6 +69,46 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Tests
 
             mockPipeWriter.Verify(m => m.Complete(It.IsAny<Exception>()), Times.Once());
             mockPipeReader.Verify(m => m.Complete(It.IsAny<Exception>()), Times.Once());
+        }
+
+        [Fact]
+        public async Task OnConnectionFiresOnCompleted()
+        {
+            var serviceContext = new TestServiceContext();
+            var dispatcher = new ConnectionDispatcher(serviceContext, _ => Task.CompletedTask);
+
+            var connection = new Mock<TransportConnection> { CallBase = true }.Object;
+            connection.ConnectionClosed = new CancellationToken(canceled: true);
+            var completeFeature = connection.Features.Get<IConnectionCompleteFeature>();
+
+            Assert.NotNull(completeFeature);
+            object stateObject = new object();
+            object callbackState = null;
+            completeFeature.OnCompleted(state => { callbackState = state; return Task.CompletedTask; }, stateObject);
+
+            await dispatcher.OnConnection(connection);
+
+            Assert.Equal(stateObject, callbackState);
+        }
+
+        [Fact]
+        public async Task OnConnectionOnCompletedExceptionCaught()
+        {
+            var serviceContext = new TestServiceContext();
+            var dispatcher = new ConnectionDispatcher(serviceContext, _ => Task.CompletedTask);
+
+            var connection = new Mock<TransportConnection> { CallBase = true }.Object;
+            connection.ConnectionClosed = new CancellationToken(canceled: true);
+            var completeFeature = connection.Features.Get<IConnectionCompleteFeature>();
+
+            Assert.NotNull(completeFeature);
+            object stateObject = new object();
+            object callbackState = null;
+            completeFeature.OnCompleted(state => { callbackState = state; throw new InvalidTimeZoneException(); }, stateObject);
+
+            await dispatcher.OnConnection(connection);
+
+            Assert.Equal(stateObject, callbackState);
         }
     }
 }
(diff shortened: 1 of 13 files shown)


src/Servers/Kestrel/Transport.Abstractions/src/Internal/TransportConnection.FeatureCollection.cs
```diff
@@ -100,5 +105,68 @@ void IConnectionHeartbeatFeature.OnHeartbeat(System.Action<object> action, objec
...
+            {
+                try
+                {
+                    var task = entry.Key.Invoke(entry.Value);
+                    if (!ReferenceEquals(task, Task.CompletedTask))
+                    {
+                        return FireOnCompletedAwaited(task, onCompleted);
+                    }
+                }
+                catch (Exception)
```
Tratcher: Is there a way you'd recommend logging these exceptions? TransportConnection does not have a logger.

anurse:
cc @BrennanConroy @mikaelm12 @halter73 We may want to implement this feature in SignalR. It could be useful.

davidfowl:
How is this different from IConnectionLifetime

Tratcher:
IConnectionLifetime ConnectionClosed will fire immediately, possibly in the middle of requests. Trying to use it for cleanup would cause lots of race conditions and ObjectDisposedExceptions up stack. OnCompleted acts as a finally block for the connection, allowing you to clean up resources once you're sure everybody is done using them.

src/Servers/Kestrel/Transport.Abstractions/src/Internal/TransportConnection.FeatureCollection.cs
jkotalik: Can we make TransportConnection start taking a logger?

davidfowl: It could be a property.

halter73: I think I prefer the property since that's how everything else is set on the TransportConnection. I would just make sure it doesn't nullref if the logger's not set.

src/Servers/Kestrel/Core/test/ConnectionDispatcherTests.cs
```diff
@@ -69,5 +70,45 @@ public async Task OnConnectionCompletesTransportPipesAfterReturning()
...
+            var dispatcher = new ConnectionDispatcher(serviceContext, _ => Task.CompletedTask);
+
+            var connection = new Mock<TransportConnection> { CallBase = true }.Object;
+            connection.ConnectionClosed = new CancellationToken(canceled: true);
+            var completeFeature = connection.Features.Get<IConnectionCompleteFeature>();
+
+            Assert.NotNull(completeFeature);
+            object stateObject = new object();
+            object callbackState = null;
+            completeFeature.OnCompleted(state => { callbackState = state; throw new InvalidTimeZoneException(); }, stateObject);
```
halter73: `InvalidTimeZoneException` interesting choice.

src/Servers/Kestrel/Transport.Abstractions/src/Internal/TransportConnection.FeatureCollection.cs
```diff
@@ -100,5 +105,68 @@ void IConnectionHeartbeatFeature.OnHeartbeat(System.Action<object> action, objec
         {
             OnHeartbeat(action, state);
         }
+
+        public void OnCompleted(Func<object, Task> callback, object state)
```
halter73:
```suggestion
        void IConnectionCompleteFeature.OnCompleted(Func<object, Task> callback, object state)
```

src/Servers/Kestrel/Core/test/ConnectionDispatcherTests.cs
Tratcher: It's always fun to find a weird exception that would get your attention if it was ever thrown. My personal favorite is java has a HeadlessException.

src/Servers/Kestrel/Transport.Abstractions/src/Internal/TransportConnection.FeatureCollection.cs
```diff
@@ -100,5 +105,68 @@ void IConnectionHeartbeatFeature.OnHeartbeat(System.Action<object> action, objec
...
+        {
+            if (_onCompleted == null)
+            {
+                _onCompleted = new Stack<KeyValuePair<Func<object, Task>, object>>();
+            }
+            _onCompleted.Push(new KeyValuePair<Func<object, Task>, object>(callback, state));
+        }
+
+        public Task FireOnCompleted()
```
halter73: I don't love FireOnCompleted() being public since it's not callled/needed by the transports that might reference Transport.Abstractions. I guess the same could be said for various property setters.

Out of caution, I think we should throw if this gets called multiple times. Or if someone calls OnCompleted after FireOnCompleted was already called.

halter73: Another alternative is exposing the callbacks themselves via the TransportConnection and let the ConnectionDispatcher invoke them directly. This neatly solves the logging problem, but everything else about it is worse IMHO.

anurse: `TrySetCompleted`? :)

anurse: Or just `Complete`/`TryComplete`, since generally that's the naming pattern we've used in other places.

halter73 (approved):

davidfowl (approved):
LGTM

PR #9755, opened 2019-04-25
Be more defensive in HtmlDirectoryFormatter

analogrelay:
Fixes aspnet/AspNetCore-Internal#2245

Our tests were flaking out due to this possible race where the directory browser middleware enumerates the directory but a file is deleted while we're formatting the response. This change just catches the FNF/DNF exceptions and suppresses the file from the directory listing

### DIFF
--- a/src/Middleware/StaticFiles/src/HtmlDirectoryFormatter.cs
+++ b/src/Middleware/StaticFiles/src/HtmlDirectoryFormatter.cs
@@ -33,7 +33,7 @@ namespace Microsoft.AspNetCore.StaticFiles
                 throw new ArgumentNullException(nameof(encoder));
             }
             _htmlEncoder = encoder;
-        } 
+        }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


src/Middleware/StaticFiles/src/HtmlDirectoryFormatter.cs
```diff
@@ -33,7 +33,7 @@ public HtmlDirectoryFormatter(HtmlEncoder encoder)
                 throw new ArgumentNullException(nameof(encoder));
             }
             _htmlEncoder = encoder;
-        } 
+        }
```
anurse: Just some auto trailing whitespace cleanup

Tratcher (approved):

Tratcher:
Needs more usings.

anurse:
Whoops, should've used VS :). Or just... `dotnet build`

PR #9756, opened 2019-04-25
Add CertificateAuthentication

HaoK:
Initial port of the code with all unit tests passing and some stuff commented out so we can track diffs in git.

Take a look at the commented out changes i needed to make for tests @blowdart 

Next steps will be to remove all the commented out code and move the headerforwarding into the existing middleware


src/Security/Authentication/Certificate/src/README.md
```diff
@@ -0,0 +1,241 @@
+﻿# Microsoft.AspNetCore.Authentication.Certificate
+ 
+*This handler is available for ASP.NET Core 2.0 only. Not due to technical reasons, I'm just lazy.*
```
blowdart: This line should probably go away ;)

src/Security/Authentication/test/CertificateTests.cs
```diff
@@ -0,0 +1,739 @@
...
+        //            ValidateCertificateUse = false,
+        //            Events = sucessfulValidationEvents
+        //        },
+        //        Certificates.RootedServerEku);
+
+        //    var response = await server.CreateClient().GetAsync("https://example.com/");
+        //    Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+        //}
+
+        [Fact]
```
blowdart: Delete test, without rooted certs, revocation can't happen

blowdart:
Couple of things to be deleted, but yea, fair enough

HaoK:
Ok I'll delete all the commented out stuff next so its easy for you to see what got removed in the next commit

src/Security/Authentication/Certificate/src/X509CertificateExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+namespace Microsoft.AspNetCore.Authentication.Certificate
+{
+    public static class X509Certificate2Extensions
+    {
+        public static bool IsSelfSigned(this X509Certificate2 certificate)
+        {
+            return certificate.SubjectName.RawData.SequenceEqual(certificate.IssuerName.RawData);
+        }
+
+        public static string SHA256Thumprint(this X509Certificate2 certificate)
```
bartonjs: If you're netcoreapp30 this is just `cert.GetCertHashString(HashAlgorithmName.SHA256)`.  (Seems weird to me that you'd be putting public extensions here)

Eilon:
When this is non-draft, please tag `@aspnet/build` to review the package addition and other infra changes.

src/Security/Authentication/Certificate/src/X509CertificateExtensions.cs
HaoK: Nice, switched to that

src/Security/Authentication/Certificate/src/CertificateForwarderMiddleware.cs
```diff
@@ -0,0 +1,70 @@
...
+{
+
+    // TODO: move to header forwarder
+    public class CertificateForwarderMiddleware
+    {
+        private readonly RequestDelegate _next;
+        private readonly CertificateForwarderOptions _options;
+        private readonly ILogger _logger;
+
+        public CertificateForwarderMiddleware(
```
Tratcher: This should not be a new middleware, it should go into the existing FowardedHeaders middleware as another optional header.

src/Security/Authentication/Certificate/src/CertificateValidator.cs
```diff
@@ -0,0 +1,23 @@
...
+    public static class CertificateValidator
+    {
+        /// <summary>
+        /// Disables connection based client certificate validation so the middleware can handle it instead.
+        /// </summary>
+        /// <param name="certificate">The certificate used to authenticate the remote party.</param>
+        /// <param name="chain">The chain of certificate authorities associated with the remote certificate.</param>
+        /// <param name="errors">One or more errors associated with the remote certificate.</param>
+        /// <returns>A Boolean value that determines whether the specified certificate is accepted for authentication.</returns>
+        public static bool DisableChannelValidation(X509Certificate2 certificate, X509Chain chain, SslPolicyErrors errors)
```
Tratcher: This isn't adding anything meaningful, and it's Kestrel specific. We should instead look at a new bool option for kestrel that disables client cert validation.

src/Security/Authentication/Certificate/src/CertificateForwarderMiddleware.cs
HaoK: You mean the HeaderPropagation middleware?  That isn't in the shared framework currently when I tried to reference it from security it said nope...

Tratcher: No, this one:
https://github.com/aspnet/AspNetCore/blob/master/src/Middleware/HttpOverrides/src/ForwardedHeadersMiddleware.cs

src/Security/Authentication/Certificate/src/CertificateAuthenticationHandler.cs
```diff
@@ -0,0 +1,271 @@
...
+
+                var validateCertificateContext = new ValidateCertificateContext(Context, Scheme, Options)
+                {
+                    ClientCertificate = clientCertificate
+                };
+
+                await Events.ValidateCertificate(validateCertificateContext);
+
+                if (validateCertificateContext.Result != null &&
+                    validateCertificateContext.Result.Succeeded)
```
DAllanCarr: Can we use `?` here to collapse initial null check.

src/Security/Authentication/Certificate/src/CertificateAuthenticationHandler.cs
```diff
@@ -0,0 +1,271 @@
...
+                {
+                    return Success(validateCertificateContext.Principal, clientCertificate);
+                }
+
+                if (validateCertificateContext.Result != null &&
```
DAllanCarr: ^^

src/Security/Authentication/Certificate/src/CertificateForwarderMiddleware.cs
HaoK: Alright as we discussed, moved the CertForwarderMiddleware stuff in to HttpOverrides.

What do you guys think about getting rid of the AddCertificateForwader method on configure services and switching to just having that only as part of UseCertificateForwarder()?  Do we really need to allow them to configure the options in DI?  Can still take the Lambda, but then we wouldn't need two methods.

Would just be a single call in Configure
```
app.UseCertificateForwarder(options => options.CertificateHeader = headerName);
```

src/Security/Authentication/Certificate/src/CertificateValidator.cs
HaoK: Should I do this in a separate PR?

src/Security/Authentication/Certificate/src/CertificateForwarderMiddleware.cs
Tratcher: We've avoided adding any new middleware extensions that passes options to UseXx. It's a lot cleaner to keep all of the config in ConfigureServices. AddCertificateForwader is primarily for discoverability.

HaoK: Yeah I only asked because the other two had UseXx that took options :)

src/Security/Authentication/Certificate/src/CertificateValidator.cs
Tratcher: Delete this one now and file a new issue for Kestrel.

HaoK: Cool, filed https://github.com/aspnet/AspNetCore/issues/10351

src/Security/Authentication/Certificate/src/CertificateAuthenticationHandler.cs
```diff
@@ -0,0 +1,269 @@
...
+        }
+
+        private AuthenticateResult Success(ClaimsPrincipal principal, X509Certificate2 certificate)
+        {
+            var props = new AuthenticationProperties
+            {
+                Items =
+                {
+                    {
+                        CertificateAuthenticationDefaults.CertificateItemsKey, certificate.GetRawCertDataString()
```
Tratcher: Why would this get put this in the properties collection? Use the ClaimsIdentity.BootstrapContext if you must. Whatever you do, do it before the ValidateCertificate event.

They could also just read the cert off the Connection if they wanted, not sure why we need to bundle it.

src/Security/Authentication/Certificate/src/CertificateAuthenticationExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+using Microsoft.AspNetCore.Authentication.Certificate;
+
+namespace Microsoft.Extensions.DependencyInjection
+{
+    /// <summary>
+    /// Extension methods to add Certificate authentication capabilities to an HTTP application pipeline.
+    /// </summary>
+    public static class CertificateAuthenticationAppBuilderExtensions
+    {
+        public static AuthenticationBuilder AddCertificate(this AuthenticationBuilder builder)
```
Tratcher: Doc comment all the things...

src/Security/Authentication/Certificate/src/CertificateAuthenticationHandler.cs
```diff
@@ -0,0 +1,269 @@
...
+        {
+            // You only get client certificates over HTTPS
+            if (!Context.Request.IsHttps)
+            {
+                return AuthenticateResult.NoResult();
+            }
+
+            var clientCertificate = await Context.Connection.GetClientCertificateAsync();
+
+            // This should never be the case, as cert authentication happens long before ASP.NET kicks in.
```
Tratcher: What does this mean?

src/Security/Authentication/Certificate/src/CertificateAuthenticationHandler.cs
```diff
@@ -0,0 +1,269 @@
...
+
+        protected override async Task<AuthenticateResult> HandleAuthenticateAsync()
+        {
+            // You only get client certificates over HTTPS
+            if (!Context.Request.IsHttps)
+            {
+                return AuthenticateResult.NoResult();
+            }
+
+            var clientCertificate = await Context.Connection.GetClientCertificateAsync();
```
Tratcher: Can throw. Expand the try block?

src/Security/Authentication/Certificate/src/CertificateAuthenticationHandler.cs
```diff
@@ -0,0 +1,269 @@
...
+
+            // This should never be the case, as cert authentication happens long before ASP.NET kicks in.
+            if (clientCertificate == null)
+            {
+                Logger.LogDebug("No client certificate found.");
```
Tratcher: Extract logger messages. https://github.com/aspnet/AspNetCore/blob/master/src/Middleware/StaticFiles/src/LoggerExtensions.cs

src/Security/Authentication/Certificate/src/CertificateAuthenticationHandler.cs
```diff
@@ -0,0 +1,269 @@
...
+                var chain = new X509Chain
+                {
+                    ChainPolicy = chainPolicy
+                };
+
+                var certificateIsValid = chain.Build(clientCertificate);
+
+                if (!certificateIsValid)
+                {
+                    using (Logger.BeginScope(clientCertificate.GetCertHashString(HashAlgorithmName.SHA256)))
```
Tratcher: Creating a scope seems excessive.

src/Security/Authentication/Certificate/src/CertificateAuthenticationHandler.cs
```diff
@@ -0,0 +1,269 @@
...
+                }
+
+                var validateCertificateContext = new ValidateCertificateContext(Context, Scheme, Options)
+                {
+                    ClientCertificate = clientCertificate
+                };
+
+                await Events.ValidateCertificate(validateCertificateContext);
+
+                if (validateCertificateContext?.Result?.Succeeded == true)
```
Tratcher:
```
if (validateCertificateContext.Result != null) return validateCertificateContext.Result;
```

src/Security/Authentication/Certificate/src/CertificateAuthenticationExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+            => builder.AddCertificate(CertificateAuthenticationDefaults.AuthenticationScheme);
+
+        public static AuthenticationBuilder AddCertificate(this AuthenticationBuilder builder, string authenticationScheme)
+            => builder.AddCertificate(authenticationScheme, configureOptions: null);
+
+        public static AuthenticationBuilder AddCertificate(this AuthenticationBuilder builder, Action<CertificateAuthenticationOptions> configureOptions)
+            => builder.AddCertificate(CertificateAuthenticationDefaults.AuthenticationScheme, configureOptions);
```
Tratcher: Call the overload that does not set the display name. This auth handler won't work in the normal Identity select-your-auth-type flow, it doesn't support Challenge.

There is one flow where you could allow the user to prompt for a cert:
1) Use HttpSys with lazy cert negotiation enabled (default)
1) The user selects Certificate from the list.
1) The controller calls AuthenticateAsync("Certificate") rather than challenge.
1) The browser prompts for a certificate (on the same request).
1) The controller does whatever signin action it wants.

src/Security/Authentication/Certificate/src/CertificateAuthenticationHandler.cs
```diff
@@ -0,0 +1,269 @@
...
+                    return authenticationFailedContext.Result;
+                }
+
+                throw;
+            }
+        }
+
+        protected override Task HandleChallengeAsync(AuthenticationProperties properties)
+        {
+            // Certificate authentication takes place at the connection level. We can't prompt once we're in
```
Tratcher: Let's put an event here for the Identity auth selection case. For HttpSys a dev could call AuthenticateAsync (in the selection case it would not have been called) and it would trigger a browser prompt on the same request. If that failed then a 403 is warranted. If it succeeded they could redirect to the next stage of sign-in and the cert would be preserved in the server for the next request.

src/Security/Authentication/Certificate/src/CertificateAuthenticationHandler.cs
```diff
@@ -0,0 +1,269 @@
...
+                    }
+                    return AuthenticateResult.Fail("Client certificate failed validation.");
+                }
+
+                var validateCertificateContext = new ValidateCertificateContext(Context, Scheme, Options)
+                {
+                    ClientCertificate = clientCertificate
+                };
+
+                await Events.ValidateCertificate(validateCertificateContext);
```
Tratcher: CertificateValidated

In the other providers we build the Principal before calling this success event and pass it in. Then the dev can customize the identity without creating it from scratch.
https://github.com/aspnet/AspNetCore/blob/4e44025a52e4b73aa17e09a8041b0e166e0c5ce0/src/Security/Authentication/OpenIdConnect/src/OpenIdConnectHandler.cs#L598-L604

src/Security/Authentication/Certificate/src/README.md
```diff
@@ -0,0 +1,239 @@
+# Microsoft.AspNetCore.Authentication.Certificate
```
Tratcher: Move to a doc PR/bug?

src/Security/Authentication/test/CertificateTests.cs
```diff
@@ -0,0 +1,631 @@
...
+                wireUpHeaderMiddleware: true);
+
+            var client = server.CreateClient();
+            client.DefaultRequestHeaders.Add("X-ARR-ClientCert", "OOPS" + Convert.ToBase64String(Certificates.SelfSignedValidWithNoEku.RawData));
+            var response = await client.GetAsync("https://example.com/");
+            Assert.Equal(HttpStatusCode.Forbidden, response.StatusCode);
+        }
+
+        [Fact]
+        public async Task VerifySettingTheHeaderOnTheForwarderOptionsWorks()
```
Tratcher: Move forwarder tests to HttpOverride.Tests.

Tratcher (changes requested):
PM code 😢 

Also get rid of the MessagePack submodule change.

src/Security/Authentication/Certificate/src/Microsoft.AspNetCore.Authentication.Certificate.csproj
```diff
@@ -0,0 +1,17 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <Description>ASP.NET Core middleware that enables an application to support certificate authentication.</Description>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <IsAspNetCoreApp>true</IsAspNetCoreApp>
+    <DefineConstants>$(DefineConstants);SECURITY</DefineConstants>
+    <NoWarn>$(NoWarn);CS1591</NoWarn>
```
natemcmaster:
```suggestion
```

Let's start this off right and document the API. Would recommend not suppressing CS1591 for new code.

eng/SharedFramework.Local.props
```diff
@@ -47,6 +47,7 @@
     <AspNetCoreAppReference Include="Microsoft.AspNetCore.Server.Kestrel" />
     <AspNetCoreAppReference Include="Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions" />
     <AspNetCoreAppReference Include="Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets" />
+    <AspNetCoreAppReference Include="Microsoft.AspNetCore.Authentication.Certificate" />
```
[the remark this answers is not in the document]
HaoK: I just ran `GenerateProjectList.ps1` these changes came from that auto'magically'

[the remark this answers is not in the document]
Tratcher: I assume that's because of IsAspNetCoreApp in the csproj.
https://github.com/aspnet/AspNetCore/pull/9756/files#diff-f9b22fb425e6bb8b26f9464265f51ea2R6

Going into the shared framework needs to be an explicit decision. @anurse @blowdart

src/Middleware/HttpOverrides/src/CertificateForwarderExtensions.cs
```diff
@@ -2,13 +2,14 @@
...
 using System;
-
 using Microsoft.AspNetCore.HttpOverrides;
 using Microsoft.Extensions.DependencyInjection;
-using System.Security.Cryptography.X509Certificates;
 
 namespace Microsoft.AspNetCore.Builder
 {
+    /// <summary>
+    /// Extension methods for using the certificate fowarder
```
Eilon: Need full proper grammar (i.e. complete sentences), so need to end with a period.

src/Security/Authentication/test/CertificateTests.cs
HaoK: I created a modified copy of these tests in the http override but I kept the end to end tests here as well, since we do want to test cert auth + the forwarder actually authenticates / forbids correctly

src/Security/Authentication/Certificate/src/CertificateAuthenticationHandler.cs
HaoK: I don't think that's the intent here, @blowdart can confirm, but it looks like this event is meant to be used to do more complicated custom validation, as opposed to a 'Success' event

src/Security/Authentication/Certificate/src/CertificateAuthenticationHandler.cs
HaoK: Can we file a separate issue/ask to add support for this later?  This doesn't sound like something that needs to be in the initial chcekpoint, @blowdart thoughts?

eng/SharedFramework.Local.props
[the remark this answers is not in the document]
natemcmaster: Thanks for catching this, Chris. Please review these requirements as a part of this discussion: https://github.com/aspnet/AspNetCore/blob/master/docs/SharedFramework.md#what-goes-into-the-shared-framework

src/Middleware/HttpOverrides/src/CertificateForwarderMiddleware.cs
```diff
@@ -0,0 +1,68 @@
...
+                throw new ArgumentNullException(nameof(options));
+            }
+
+            _options = options.Value;
+            _logger = loggerFactory.CreateLogger<CertificateForwarderMiddleware>();
+        }
+
+        public async Task Invoke(HttpContext httpContext)
+        {
+            if (!string.IsNullOrWhiteSpace(_options.CertificateHeader))
```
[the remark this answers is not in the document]
HaoK: Fair enough, will switch to options validation to guard against this ever being null or empty

src/Security/Authentication/Certificate/src/CertificateAuthenticationExtensions.cs
HaoK: Its only the AddRemoteScheme method that takes a display name, I don't believe this currently is setting a display name so it won't show up in identity flows by default.

eng/SharedFramework.Local.props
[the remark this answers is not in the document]
HaoK: Doesn't look like any of the other authentication schemes are in the shared framework including cookies, so seems like this would be a no, removing this for now unless @blowdart really does want this included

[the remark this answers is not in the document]
HaoK: Never mind, cookies is there, I'll leave this for now until we figure out whether it belongs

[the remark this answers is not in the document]
blowdart: No, it doesn't.

src/Security/Authentication/Certificate/src/CertificateAuthenticationHandler.cs
blowdart: But only http.sys. While that'd be nice, we can't make kestrel do it, so yes, nice to have only

src/Security/Authentication/Certificate/src/CertificateAuthenticationHandler.cs
blowdart: It's a bit of both, but yea, mostly custom validation. Weird we do both. Could move the call to CreatePrincipal from line 124 up, then you get the default principal that you can fuck around with in the event

eng/SharedFramework.Local.props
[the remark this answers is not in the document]
HaoK: Ok removed from shared framework

src/Middleware/HttpOverrides/src/CertificateForwarderMiddleware.cs
```diff
@@ -0,0 +1,68 @@
...
+using System.Security.Cryptography.X509Certificates;
+using System.Threading.Tasks;
+
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.AspNetCore.HttpOverrides
+{
+    public class CertificateForwarderMiddleware
```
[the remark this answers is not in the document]
HaoK: Doc comments added

src/Middleware/HttpOverrides/src/CertificateForwarderMiddleware.cs
```diff
@@ -0,0 +1,78 @@
...
+using Microsoft.AspNetCore.Http;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.AspNetCore.HttpOverrides
+{
+    /// <summary>
+    /// Middleware that converts a forward header into a client certificate if found.
+    /// </summary>
+    public class CertificateForwarderMiddleware
```
davidfowl: Why isn;t this part of fowarded headers middleware?

src/Middleware/HttpOverrides/src/CertificateForwarderOptions.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// </summary>
+    public class CertificateForwarderOptions
+    {
+        /// <summary>
+        /// The name of the header containing the client certificate.
+        /// </summary>
+        /// <remarks>
+        /// This defaults to X-ARR-ClientCert, which is the header Azure Web Apps uses.
+        /// </remarks>
+        public string CertificateHeader { get; set; } = "X-ARR-ClientCert";
```
davidfowl: Can we avoid this coupling in our stack? Default it to null and we can do the Azure specific cert options elsewhere (much like we do forwarded headers).

src/Security/Authentication/Certificate/samples/Certificate.Sample/Startup.cs
```diff
@@ -0,0 +1,69 @@
...
+                config.Filters.Add(new AuthorizeFilter(policy));
+            });
+        }
+
+        // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
+        public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
+        {
+            app.UseRouting();
+
+            if (((IHostEnvironment)env).IsDevelopment())
```
davidfowl: That's not how you do this, you just import a different namespace

src/Security/Authentication/Certificate/src/CertificateAuthenticationHandler.cs
```diff
@@ -0,0 +1,269 @@
...
+                Logger.LogDebug("No client certificate found.");
+                return AuthenticateResult.NoResult();
+            }
+
+            // If we have a self signed cert, and they're not allowed, exit early and not bother with
+            // any other validations.
+            if (clientCertificate.IsSelfSigned() &&
+                !Options.AllowedCertificateTypes.HasFlag(CertificateTypes.SelfSigned))
+            {
+                Logger.LogWarning("Self signed certificate rejected, subject was {0}", clientCertificate.Subject);
```
davidfowl: Needs proper structured logging instead of ordinal parameters.

src/Security/Authentication/Certificate/src/CertificateAuthenticationHandler.cs
davidfowl: Agreed, this doesn't add much value.

src/Security/Authentication/Certificate/src/CertificateTypes.cs
```diff
@@ -0,0 +1,29 @@
...
+
+        /// <summary>
+        /// SelfSigned certificates.
+        /// </summary>
+        SelfSigned = 2,
+
+        /// <summary>
+        /// All certificates.
+        /// </summary>
+        All = 3
```
davidfowl: nit: All = Chained | SelfSigned

src/Security/Authentication/Certificate/src/X509CertificateExtensions.cs
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Linq;
```
davidfowl: LINQ? If it's comparing bytes you can use the Span SequenceEqual (https://docs.microsoft.com/en-us/dotnet/api/system.memoryextensions.sequenceequal?view=netcore-3.0#System_MemoryExtensions_SequenceEqual__1_System_ReadOnlySpan___0__System_ReadOnlySpan___0__)

src/Middleware/HttpOverrides/src/CertificateForwarderMiddleware.cs
Tratcher: We looked at it but the ForwardedHeaders middleware is much more complex. That one handles lists of forwarders and steps through the list. This one only handles a single value.

There's another middleware in this package already that similarly handles a single method override header.

src/Middleware/HttpOverrides/src/CertificateForwarderOptions.cs
Tratcher: Note defaulting to null would mean the config is invalid by default.

I don't see a defacto standard name for this one, "X-Client-Cert" is the closest I can find.

src/Security/Authentication/Certificate/src/CertificateAuthenticationHandler.cs
Tratcher: It's run after the built in validation has already passed, we think it's valid. If you wanted to replace the validation you'd need an event at line 62.

src/Middleware/HttpOverrides/src/CertificateForwarderMiddleware.cs
davidfowl: Why is this different? I'm not convinced this should be a separate middleware. Don't all of the spame spoofing concerns apply? Why doesn't this have those features? Why are certs different from *all* of the other headers forwarded by a reverse proxy?

src/Security/Authentication/Certificate/src/CertificateAuthenticationHandler.cs
```diff
@@ -0,0 +1,269 @@
...
+
+        protected override Task HandleChallengeAsync(AuthenticationProperties properties)
+        {
+            // Certificate authentication takes place at the connection level. We can't prompt once we're in
+            // user code, so the best thing to do is Forbid, not Challenge.
+            Response.StatusCode = 403;
+            return Task.CompletedTask;
+        }
+
+        protected override Task HandleForbiddenAsync(AuthenticationProperties properties)
```
Tratcher: This is the same as the base implementation, why did you override it?

src/Middleware/HttpOverrides/src/CertificateForwarderOptions.cs
davidfowl: Defaulting to something azure specific is a no go, so invalid by default means it should throw or 400 like we spoke about.

Tratcher: For null it would throw on startup when Options validation ran.

HaoK: We wouldn't need the validation then, it could just no-op by default if the header is null/empty like it did before

src/Security/Authentication/Certificate/src/CertificateAuthenticationHandler.cs
HaoK: @blowdart any issue with just removing this from the items?

src/Middleware/HttpOverrides/src/CertificateForwarderOptions.cs
Tratcher: No, that's a pit of failure. If you're adding the middleware then you want it to work or tell you why not.

src/Security/Authentication/Certificate/src/CertificateAuthenticationHandler.cs
blowdart: @leastprivilege - this was from you, so, does it need to be a claim?

src/Middleware/HttpOverrides/src/CertificateForwarderMiddleware.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        /// <summary>
+        /// Looks for the presence of a <see cref="CertificateForwarderOptions.CertificateHeader"/> header in the request,
+        /// if found, converts this header to a ClientCertificate set on the connection.
+        /// </summary>
+        /// <param name="httpContext">The <see cref="HttpContext"/>.</param>
+        /// <returns>A <see cref="Task"/>.</returns>
+        public async Task Invoke(HttpContext httpContext)
+        {
+            var clientCertificate = await httpContext.Connection.GetClientCertificateAsync();
```
Tratcher: https://github.com/aspnet/AspNetCore/blob/49c01eefecf1dbd75e5536aa803d689390c6770a/src/Servers/IIS/IISIntegration/src/ForwardedTlsConnectionFeature.cs

HaoK: Ok took a stab at this, take a look and see if its better

src/Security/Authentication/Certificate/src/X509CertificateExtensions.cs
HaoK: Updated to use span sequence equals

src/Security/Authentication/Certificate/src/CertificateAuthenticationHandler.cs
HaoK: Structured logging added

src/Middleware/HttpOverrides/src/CertificateForwarderOptions.cs
HaoK: So how do we want this to behave by default?  `UseCertificateHeaderForwarding` without calling Configure and specifying a header will throw and say please call `AddCertificateHeaderForwarding` and specify a header name?

src/Security/Authentication/Certificate/src/CertificateAuthenticationHandler.cs
HaoK: @blowdart do we need to elaborate this comment? (or remove?)

src/Security/Authentication/Certificate/src/CertificateAuthenticationHandler.cs
HaoK: Ok renamed

src/Security/Authentication/Certificate/src/README.md
HaoK: Yeah docs with this will come later, readme seems ok to leave in the meantime

src/Security/Authentication/Certificate/src/CertificateAuthenticationHandler.cs
blowdart: There's a null check that comes after. It should never be null, hence the comment.

src/Security/Authentication/Certificate/src/README.md
```diff
@@ -0,0 +1,239 @@
...
+For real functionality you will probably want to call a service registered in DI which talks to a database or other type of 
+user store. You can grab your service by using the context passed into your delegates, like so
+
+```c#
+services.AddAuthentication(CertificateAuthenticationDefaults.AuthenticationScheme)
+        .AddCertificate(options =>
+        {
+            options.Events = new CertificateAuthenticationEvents
+            {
+                OnValidateCertificate = context =>
```
Tratcher: Outdated

src/Security/Authentication/Certificate/src/README.md
```diff
@@ -0,0 +1,239 @@
...
+
+### Kestrel
+
+In program.cs configure `UseKestrel()` as follows.
+
+```c#
+public static IWebHost BuildWebHost(string[] args) =>
+    WebHost.CreateDefaultBuilder(args)
+           .UseStartup<Startup>()
+           .UseKestrel(options =>
```
Tratcher: Update this section to match the changes in the sample.

src/Security/Authentication/Certificate/src/README.md
```diff
@@ -0,0 +1,239 @@
...
+called after certificate has been validated, passed validation, abut before the default principal has been created. This allows you to perform your own validation, for example 
+checking if the certificate is one your services knows about, and to construct your own principal. For example,
+
+```c#
+services.AddAuthentication(CertificateAuthenticationDefaults.AuthenticationScheme)
+        .AddCertificate(options =>
+        {
+            options.Events = new CertificateAuthenticationEvents
+            {
+                OnValidateCertificate = context =>
```
Tratcher: Outdated

src/Security/Authentication/Certificate/src/Microsoft.AspNetCore.Authentication.Certificate.csproj
```diff
@@ -0,0 +1,15 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <Description>ASP.NET Core middleware that enables an application to support certificate authentication.</Description>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <DefineConstants>$(DefineConstants);SECURITY</DefineConstants>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <PackageTags>aspnetcore;authentication;security;x509;certificate</PackageTags>
```
natemcmaster:
```suggestion
    <PackageTags>aspnetcore;authentication;security;x509;certificate</PackageTags>
    <IsShippingPackage>true</IsShippingPackage>
```

If you want this to ship to nuget.org, add this property. Otherwise, this package will be considered internal-only (which is okay if you're still in the early stages of building out this API)

src/Security/Authentication/Certificate/src/CertificateAuthenticationHandler.cs
```diff
@@ -0,0 +1,254 @@
...
+                    {
+                        chainErrors.Add($"{validationFailure.Status} {validationFailure.StatusInformation}");
+                    }
+                    Logger.CertificateFailedValidation(clientCertificate.Subject, chainErrors);
+                    return AuthenticateResult.Fail("Client certificate failed validation.");
+                }
+
+                var certificateValidatedContext = new CertificateValidatedContext(Context, Scheme, Options)
+                {
+                    ClientCertificate = clientCertificate
```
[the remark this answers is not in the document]
HaoK: Ah right, updated

src/Middleware/HttpOverrides/ref/Microsoft.AspNetCore.HttpOverrides.netcoreapp3.0.cs
```diff
@@ -3,6 +3,11 @@
 
 namespace Microsoft.AspNetCore.Builder
 {
+    public static partial class CertificateForwarderExtensions
+    {
+        public static Microsoft.Extensions.DependencyInjection.IServiceCollection AddCertificateHeaderForwarding(this Microsoft.Extensions.DependencyInjection.IServiceCollection services, System.Action<Microsoft.AspNetCore.HttpOverrides.CertificateForwarderOptions> configure) { throw null; }
+        public static Microsoft.AspNetCore.Builder.IApplicationBuilder UseCertificateHeaderForwarding(this Microsoft.AspNetCore.Builder.IApplicationBuilder app) { throw null; }
```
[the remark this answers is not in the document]
HaoK: I also renamed the service method to AddCertificateForwarding to match

HaoK:
@blowdart @Tratcher thinks the behavior for the certificate forwarder should pickup the header and use that as an override for any existing cert, which is the opposite of the behavior you had before (which would fallback to the header).  For now I went with the override behavior, but you guys can discuss what's the ideal behavior

src/Security/Authentication/Certificate/src/CertificateAuthenticationDefaults.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        /// <summary>
+        /// The default value used for CertificateAuthenticationOptions.AuthenticationScheme
+        /// </summary>
+        public const string AuthenticationScheme = "Certificate";
+
+        /// <summary>
+        /// The name used for the items dictionary on AuthenticateResult
+        /// </summary>
+        public const string CertificateItemsKey = AuthenticationScheme;
```
Tratcher: Remove

src/Security/Authentication/Certificate/src/CertificateAuthenticationHandler.cs
```diff
@@ -0,0 +1,251 @@
...
+
+                var certificateValidatedContext = new CertificateValidatedContext(Context, Scheme, Options)
+                {
+                    ClientCertificate = clientCertificate,
+                    Principal = CreatePrincipal(clientCertificate)
+                };
+
+                await Events.CertificateValidated(certificateValidatedContext);
+
+                if (certificateValidatedContext?.Result != null)
```
Tratcher:
```suggestion
                if (certificateValidatedContext.Result != null)
```

src/Security/Authentication/Certificate/src/CertificateAuthenticationHandler.cs
Tratcher: Removed?

Tratcher (approved):

src/Middleware/HttpOverrides/src/CertificateForwarderExtensions.cs
```diff
@@ -0,0 +1,55 @@
...
+    /// </summary>
+    public static class CertificateForwarderExtensions
+    {
+        /// <summary>
+        /// Adds a middleware to the pipeline that will look for a certificate in a request header
+        /// decode it, and updates HttpContext.Connection.ClientCertificate.
+        /// </summary>
+        /// <param name="app"></param>
+        /// <returns></returns>
+        public static IApplicationBuilder UseCertificateForwarding(this IApplicationBuilder app)
```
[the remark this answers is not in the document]
HaoK: Filed https://github.com/aspnet/AspNetCore/issues/10679

HaoK:
@aspnet/build any ideas why I'm getting this error on ci builds only for this PR?

```
/.azure/pipelines/helix-test.yml: Could not find /.azure/pipelines/jobs/default-build.yml in repository self hosted on https://api.github.com using commit 5ff3b2aeadbc1775f718f2938171f96f5a9b3f84. GitHub reported the error, "Not Found"
```

dougbu:
> any ideas why I'm getting this error on ci builds only for this PR?

It's not only your PR and it's not only this repo e.g. https://github.com/aspnet/EntityFrameworkCore/pull/15878/checks?check_run_id=139437341. It appears to be GitHub flakiness but I'm not sure.

natemcmaster:
@HaoK just sent mail about this problem, which is affecting everyone, not just this PR. AzDO + GitHub are investigating.

PR #9759, opened 2019-04-25
RequiredPolicy reborn and less demanding as FallbackPolicy

HaoK:
Revert changes to required policy and auth z middleware, switch required policy behavior to be a fallback policy instead which is only used when Combine is called with no `IAuthorizeData`.  

Note this has no impact on `IAuthorizationService`, callers would be able to take advantage of this feature by calling the appropriate Combine (which is aware of Fallback policy), but there's nothing in the service which will use this policy automatically.

### DIFF
--- a/src/Security/Authorization/Core/ref/Microsoft.AspNetCore.Authorization.netcoreapp3.0.cs
+++ b/src/Security/Authorization/Core/ref/Microsoft.AspNetCore.Authorization.netcoreapp3.0.cs
@@ -45,14 +45,15 @@ namespace Microsoft.AspNetCore.Authorization
     public partial class AuthorizationMiddleware
     {
         public AuthorizationMiddleware(Microsoft.AspNetCore.Http.RequestDelegate next, Microsoft.AspNetCore.Authorization.IAuthorizationPolicyProvider policyProvider) { }
+        [System.Diagnostics.DebuggerStepThroughAttribute]
         public System.Threading.Tasks.Task Invoke(Microsoft.AspNetCore.Http.HttpContext context) { throw null; }
     }
     public partial class AuthorizationOptions
     {
         public AuthorizationOptions() { }
         public Microsoft.AspNetCore.Authorization.AuthorizationPolicy DefaultPolicy { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
-        public bool InvokeHandlersAfterFailure { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public Microsoft.AspNetCore.Authorization.AuthorizationPolicy FallbackPolicy { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
+        public bool InvokeHandlersAfterFailure { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public void AddPolicy(string name, Microsoft.AspNetCore.Authorization.AuthorizationPolicy policy) { }
         public void AddPolicy(string name, System.Action<Microsoft.AspNetCore.Authorization.AuthorizationPolicyBuilder> configurePolicy) { }
         public Microsoft.AspNetCore.Authorization.AuthorizationPolicy GetPolicy(string name) { throw null; }
@@ -131,8 +132,8 @@ namespace Microsoft.AspNetCore.Authorization
     {
         public DefaultAuthorizationPolicyProvider(Microsoft.Extensions.Options.IOptions<Microsoft.AspNetCore.Authorization.AuthorizationOptions> options) { }
         public System.Threading.Tasks.Task<Microsoft.AspNetCore.Authorization.AuthorizationPolicy> GetDefaultPolicyAsync() { throw null; }
-        public virtual System.Threading.Tasks.Task<Microsoft.AspNetCore.Authorization.AuthorizationPolicy> GetPolicyAsync(string policyName) { throw null; }
         public System.Threading.Tasks.Task<Microsoft.AspNetCore.Authorization.AuthorizationPolicy> GetFallbackPolicyAsync() { throw null; }
+        public virtual System.Threading.Tasks.Task<Microsoft.AspNetCore.Authorization.AuthorizationPolicy> GetPolicyAsync(string policyName) { throw null; }
     }
     public partial class DefaultAuthorizationService : Microsoft.AspNetCore.Authorization.IAuthorizationService
     {
@@ -161,8 +162,8 @@ namespace Microsoft.AspNetCore.Authorization
     public partial interface IAuthorizationPolicyProvider
     {
         System.Threading.Tasks.Task<Microsoft.AspNetCore.Authorization.AuthorizationPolicy> GetDefaultPolicyAsync();
-        System.Threading.Tasks.Task<Microsoft.AspNetCore.Authorization.AuthorizationPolicy> GetPolicyAsync(string policyName);
         System.Threading.Tasks.Task<Microsoft.AspNetCore.Authorization.AuthorizationPolicy> GetFallbackPolicyAsync();
+        System.Threading.Tasks.Task<Microsoft.AspNetCore.Authorization.AuthorizationPolicy> GetPolicyAsync(string policyName);
     }
     public partial interface IAuthorizationRequirement
     {
--- a/src/Security/Authorization/Core/ref/Microsoft.AspNetCore.Authorization.netcoreapp3.0.cs
+++ b/src/Security/Authorization/Core/ref/Microsoft.AspNetCore.Authorization.netcoreapp3.0.cs
@@ -52,7 +52,7 @@ namespace Microsoft.AspNetCore.Authorization
         public AuthorizationOptions() { }
         public Microsoft.AspNetCore.Authorization.AuthorizationPolicy DefaultPolicy { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public bool InvokeHandlersAfterFailure { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
-        public Microsoft.AspNetCore.Authorization.AuthorizationPolicy RequiredPolicy { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
+        public Microsoft.AspNetCore.Authorization.AuthorizationPolicy FallbackPolicy { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public void AddPolicy(string name, Microsoft.AspNetCore.Authorization.AuthorizationPolicy policy) { }
         public void AddPolicy(string name, System.Action<Microsoft.AspNetCore.Authorization.AuthorizationPolicyBuilder> configurePolicy) { }
         public Microsoft.AspNetCore.Authorization.AuthorizationPolicy GetPolicy(string name) { throw null; }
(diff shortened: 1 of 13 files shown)


HaoK:
Initial stab at what we discussed today with @rynowak @Tratcher @blowdart 

cc @pranavkm @JamesNK

src/Security/Authorization/Core/ref/Microsoft.AspNetCore.Authorization.netcoreapp3.0.cs
```diff
@@ -160,6 +162,7 @@ public partial interface IAuthorizationPolicyProvider
     {
         System.Threading.Tasks.Task<Microsoft.AspNetCore.Authorization.AuthorizationPolicy> GetDefaultPolicyAsync();
         System.Threading.Tasks.Task<Microsoft.AspNetCore.Authorization.AuthorizationPolicy> GetPolicyAsync(string policyName);
+        System.Threading.Tasks.Task<Microsoft.AspNetCore.Authorization.AuthorizationPolicy> GetFallbackPolicyAsync();
```
rynowak: Are we really just making breaking changes to this interface? Don't people implement it?

src/Security/Authorization/Core/src/AuthorizationPolicy.cs
```diff
@@ -176,7 +176,17 @@ public static async Task<AuthorizationPolicy> CombineAsync(IAuthorizationPolicyP
                 }
             }
 
+            // If we have no policy by now, use the fallback policy if we have one
```
rynowak: Do we know the set of places where this method is called?
1. MVC (AuthorizeFilter)
2. AuthorizationMiddleware
3. ????

This change won't impact MVC because we never call this method with empty authorize data.

src/Security/Authorization/Core/ref/Microsoft.AspNetCore.Authorization.netcoreapp3.0.cs
Tratcher: C#8 default interface methods?

src/Security/Authorization/Core/src/AuthorizationPolicy.cs
Tratcher: Oh, I thought the middleware would call GetFallbackPolicyAsync directly?
What does SignalR do?

src/Security/Authorization/Core/src/AuthorizationPolicy.cs
```diff
@@ -176,7 +176,17 @@ public static async Task<AuthorizationPolicy> CombineAsync(IAuthorizationPolicyP
...
+            if (policyBuilder == null)
+            {
+                var requiredPolicy = await policyProvider.GetFallbackPolicyAsync();
```
Tratcher: fallbackPolicy

src/Security/Authorization/Core/ref/Microsoft.AspNetCore.Authorization.netcoreapp3.0.cs
HaoK: Yeah but we already broke it by adding RequiredPolicy and removing it was going to break it too, is renaming it worse/better?

HaoK: There certainly are some people implementing this (I know identity server folks were, but it should be fairly uncommon)  When we regressed a major scenario in 2.x with policy providers we only got a handful of reports

src/Security/Authorization/Core/src/AuthorizationPolicy.cs
pranavkm: SignalR calls this here - https://github.com/aspnet/AspNetCore/blob/master/src/SignalR/common/Http.Connections/src/Internal/AuthorizeHelper.cs#L24, but it's the same behavior where it no-ops when no `AuthorizeData` is present. 

If the intent for this is to work specifically for AuthZMiddleware, 
1) Have we considered an option where this is an option passed to the `AuthZ` middleware? `app.AddAuthorizationPolicyEvaluator(option => option.FallbackPolicy = ...)`. It does make this a non-breaking change 

2) Alternatively, have the `AuthZMiddleware` call `GetFallbackPolicyAsync` when there's no data. This makes it so that no existing consumers of this API are inadvertently affected this change.

Both options mean that the feature applies specifically to `AuthZMiddleware` as documented in the remarks section of this feature.

HaoK: Perhaps I should just reword the comment to also explain that this policy is applied as part of AuthorizationPolicy.Combine.  Neither default nor fallback policy really makes sense when talking about the explicit `IAuthorizationService` since that takes either a policy name, or an explicit set of requirements.

 I don't think the intent for this feature is to 'only' apply to the middleware, if we wanted, I think we could if we wanted, also provide a parameterless overload that didn't specify a policy name/requirements `AuthorizeAsync(user, resource)` that would use the Fallback policy.

src/Security/Authorization/Core/src/DefaultAuthorizationPolicyProvider.cs
```diff
@@ -39,6 +40,15 @@ public Task<AuthorizationPolicy> GetDefaultPolicyAsync()
             return GetCachedPolicy(ref _cachedDefaultPolicy, _options.DefaultPolicy);
         }
 
+        /// <summary>
+        /// Gets the fallback authorization policy.
+        /// </summary>
+        /// <returns>The fallback authorization policy.</returns>
+        public Task<AuthorizationPolicy> GetFallbackPolicyAsync()
```
davidfowl: I kinda want to make new methods ValueTask

HaoK:
So if we really don't want to do a breaking change (and/or don't want to introduce a full on new concept), what about a slightly less flexbile option called `UseDefaultPolicyAsFallback = true` bool, which wouldn't require a new Policy, and the middleware would just use the default policy as the fallback which satisfies most of the scenarios we care about still right?

src/Security/Authorization/Core/ref/Microsoft.AspNetCore.Authorization.netcoreapp3.0.cs
JamesNK: > Yeah but we already broke it by adding RequiredPolicy and removing it was going to break it too, is renaming it worse/better?

RequiredPolicy was added after 2.2 - https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.authorization.iauthorizationpolicyprovider?view=aspnetcore-2.2

Adding it and removing it during 3.0 dev had no effect outside of preview users.

pranavkm: @JamesNK I think the concern is that adding stuff to the interface is a breaking change and we could consider using something like downcasting to a type that apps support for this new feature instead of introducing breaking changes.

src/Security/Authorization/Core/src/AuthorizationPolicy.cs
rynowak: I'm not proposing we change anything, but I want to make the following clear:

> I don't think the intent for this feature is to 'only' apply to the middleware, if we wanted

My understanding is that this feature **CANNOT** work in MVC using `IRouter`. The solution for that user is to apply the `AuthorizeFilter` globally. 

I think that this is a OK resolution.

rynowak:
> So if we really don't want to do a breaking change (and/or don't want to introduce a full on new concept), what about a slightly less flexbile option 

So if the breaking change is approved, then I supposed it doesn't matter what the new thing is called. If the intent is for this type to grow horizontally (adding new methods and properties) then it should be a class so you can do that, or we need to use default interface methods.

Even if we're allowed to make a breaking change, it's still painful for the community and should be avoided where possible.

HaoK:
Ok added the default interface method, so this is no longer a breaking change as a result right?

src/Security/Authorization/Core/src/IAuthorizationPolicyProvider.cs
```diff
@@ -22,5 +22,11 @@ public interface IAuthorizationPolicyProvider
         /// </summary>
         /// <returns>The default authorization policy.</returns>
         Task<AuthorizationPolicy> GetDefaultPolicyAsync();
+
+        /// <summary>
+        /// Gets the fallback authorization policy.
+        /// </summary>
+        /// <returns>The fallback authorization policy.</returns>
+        Task<AuthorizationPolicy> GetFallbackPolicyAsync() => GetDefaultPolicyAsync();
```
Tratcher: Returning null would be a better default. FallbackPolicy defaults to null where DefaultPolicy requires auth.

HaoK: Yeah good catch, I'll add a test for the default behavior too

src/Security/Authorization/Core/src/AuthorizationOptions.cs
```diff
@@ -27,6 +27,14 @@ public class AuthorizationOptions
         /// </remarks>
         public AuthorizationPolicy DefaultPolicy { get; set; } = new AuthorizationPolicyBuilder().RequireAuthenticatedUser().Build();
 
+        /// <summary>
+        /// Gets or sets the fallback authorization policy used by <see cref="AuthorizationPolicy.CombineAsync(IAuthorizationPolicyProvider, IEnumerable{IAuthorizeData})"/>
+        /// when no IAuthorizeData have been provided. As a result, the <see cref="AuthorizationMiddleware"/> uses the fallback policy
+        /// if there are no <see cref="IAuthorizeData"/> instances for a resource. If a resource has any <see cref="IAuthorizeData"/>
+        /// then they are evaluated instead of the fallback policy. By default the fallback policy is null, and is not used in any way by the default <see cref="IAuthorizationService"/>.
+        /// </summary>
+        public AuthorizationPolicy FallbackPolicy { get; set; }
```
pranavkm: Should we add a note that says this is unsupported by MVC and SignalR when not using endpoint routing?

pranavkm (approved):

src/Security/Authorization/Core/src/AuthorizationOptions.cs
blowdart: yes you should, otherwise I'll forget what this is again, and ask for it to be removed ;)

HaoK: Is that true?  I thought the middleware would always apply fallback policy even if there's no endpoint metadata, this logic comes from Combine...

When there's no endpoint routing, that just means metadata = null, which should still use fallback policy (just there would be no resource)?

pranavkm: We do not enforce that the authZ middleware is configured when you do not have endpoint metadata. Your app will function with no indication that the fallback policy is being ignored. While you could add the authz middleware and have it work, I think it's much better to say that for SignalR and MVC, the recommended way is to use endpoint routing for this feature to work reliably.

HaoK: So is your concern, that they might not have UseAuthorization() at all?  In that case, they don't have the middleware so nothing would be using the fallback policy, and that seems ok since the comments do specifically say this is something the middleware does.

My question was more whether there was any situation where the fallback policy would be ignored WHEN they do have UseAuthorization() in their pipeline?

HaoK: Would it be clearer if I just add an explicit sentence at saying something like "fallbackPolicy usually will have no effect unless you have the AuthZ middleware in your pipeline."

pranavkm: > fallbackPolicy usually will have no effect unless you have the AuthZ middleware in your pipeline."

That sounds great.

HaoK:
So I as able to build locally with the new interface default method but the CI builds are not happy.  Do I need to flip a flag somewhere for the CI?

Tratcher:
@aspnet/build Are default interfaces available yet?

dougbu:
@Tratcher as far as I can tell, the default interface feature didn't make it into C# 8 yet. No idea how your local tests are working @HaoK.

HaoK:
I'm just running dotnet test on the AuthZ tests, maybe that builds against ref assemblies which aren't affected?  dotnet build in the AuthZ project works too, no idea

natemcmaster:
https://github.com/dotnet/csharplang/issues/406 and https://github.com/dotnet/csharplang/blob/master/proposals/default-interface-methods.md say the feature is in-progress. We're using whichever version of the C# compiler is bundled in the .NET Core SDK (controlled by global.json). From the repo root, you can run this to find the compiler version: `dotnet .dotnet/x64/sdk/3.0.100-preview4-011136/Roslyn/bincore/csc.dll /version`. At the moment, that means we have Roslyn 3.100.19.20602 (0bafe122).

HaoK:
So what should do I about this PR in the short term without default interface implementations, should we just take a breaking change now and add the default implementation when available?  (File a tracking issue)

rynowak:
> 
> 
> So what should do I about this PR in the short term without default interface implementations, should we just take a breaking change now and add the default implementation when available? (File a tracking issue)

That seems like a good way to make progress

HaoK:
Filed https://github.com/aspnet/AspNetCore/issues/10222 to track the default interface issue

Tratcher (approved):

PR #9763, opened 2019-04-25
IAsyncEnumerable Hub Parameters

### LINKED ISSUE #7930: Implement support for IAsyncEnumerable parameters on the Server

See #5357 for the same issue pertaining to IAsyncEnumerable return values instead of parameters.

mikaelm12:
Fixes: https://github.com/aspnet/AspNetCore/issues/7930
Pr for IAsyncEnumerable params on Hub methods. This wraps up the streaming epic.


src/SignalR/clients/csharp/Client/test/FunctionalTests/HubConnectionTests.cs
```diff
@@ -357,6 +357,51 @@ await foreach (var streamValue in asyncEnumerable)
...
+        [LogLevel(LogLevel.Trace)]
+        public async Task IAsyncEnumerableServerParam(string protocolName, HttpTransportType transportType, string path)
+        {
+            var protocol = HubProtocols[protocolName];
+            using (StartServer<Startup>(out var server))
+            {
+                var connection = CreateHubConnection(server.Url, path, transportType, protocol, LoggerFactory);
+                try
+                {
+                    async IAsyncEnumerable<int> clientStreamData()
```
davidfowl: Casing!

src/SignalR/clients/csharp/Client/test/FunctionalTests/HubConnectionTests.cs
```diff
@@ -357,6 +357,51 @@ await foreach (var streamValue in asyncEnumerable)
...
+                            yield return i;
+                            await Task.Delay(10);
+                        }
+                    }
+                    var stream = clientStreamData();
+
+                    await connection.StartAsync().OrTimeout();
+                    var expectedValue = 0;
+                    var streamTo = 5;
+                    var asyncEnumerable = connection.StreamAsyncCore<int>("StreamIAsyncConsumer", new object[] { stream });
```
davidfowl: Why isn’t this calling StreamAsync?

src/SignalR/clients/csharp/Client/test/FunctionalTests/Hubs.cs
```diff
@@ -242,6 +248,14 @@ public static ChannelReader<int> StreamEchoInt(ChannelReader<int> source)
 
             return output.Reader;
         }
+
+        public static async IAsyncEnumerable<int> StreamIAsyncConsumer(IAsyncEnumerable<int> source)
+        {
+            await foreach(var item in source)
```
davidfowl: Why the loop?

src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs
```diff
@@ -278,6 +280,13 @@ private Task ProcessStreamItem(HubConnectionContext connection, StreamItemMessag
                                     Log.StartingParameterStream(_logger, hubMethodInvocationMessage.StreamIds[streamPointer]);
                                     var itemType = descriptor.StreamingParameters[streamPointer];
                                     arguments[parameterPointer] = connection.StreamTracker.AddStream(hubMethodInvocationMessage.StreamIds[streamPointer], itemType);
+
+                                    // Check if we have an IAsyncEnumerable here and wrap the channel provided by AddStream
+                                    if (ReflectionHelper.IsIAsyncEnumerable(descriptor.OriginalParameterTypes[parameterPointer]))
+                                    {
```
davidfowl: I’m not a fan of this. Everything on the sever should be in terms of IAsyncEnumerable not channel

src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs
```diff
@@ -381,6 +390,17 @@ async Task ExecuteInvocation()
             }
         }
 
+        private static async IAsyncEnumerable<T> ConvertStream<T>(ChannelReader<T> reader)
```
davidfowl: This is built into ChannelReader. It’s ReadAllAsync

src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs
```diff
@@ -278,6 +280,13 @@ private Task ProcessStreamItem(HubConnectionContext connection, StreamItemMessag
                                     Log.StartingParameterStream(_logger, hubMethodInvocationMessage.StreamIds[streamPointer]);
                                     var itemType = descriptor.StreamingParameters[streamPointer];
                                     arguments[parameterPointer] = connection.StreamTracker.AddStream(hubMethodInvocationMessage.StreamIds[streamPointer], itemType);
+
+                                    // Check if we have an IAsyncEnumerable here and wrap the channel provided by AddStream
+                                    if (ReflectionHelper.IsIAsyncEnumerable(descriptor.OriginalParameterTypes[parameterPointer]))
```
davidfowl: This logic should be moved into AddStream

src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs
mikaelm12: Forgot about that API

src/SignalR/clients/csharp/Client/test/FunctionalTests/HubConnectionTests.cs
mikaelm12: It can. Was using this API in another test and copied it

src/SignalR/clients/csharp/Client/test/FunctionalTests/HubConnectionTests.cs
mikaelm12: Too much Java. This doesn't look wrong anymore 😅

src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs
```diff
@@ -27,6 +27,8 @@ public partial class DefaultHubDispatcher<THub> : HubDispatcher<THub> where THub
         private readonly ILogger<HubDispatcher<THub>> _logger;
         private readonly bool _enableDetailedErrors;
 
+        private static readonly MethodInfo _convertToStream = typeof(DefaultHubDispatcher<THub>).GetMethods(BindingFlags.NonPublic | BindingFlags.Static).Single(m => m.Name.Equals("ConvertStream"));
```
davidfowl: Remove this.

src/SignalR/server/Core/src/StreamTracker.cs
```diff
@@ -16,16 +16,37 @@ namespace Microsoft.AspNetCore.SignalR
...
+            // Check if we have an IAsyncEnumerable here and wrap the channel provided by AddStream
+            if (ReflectionHelper.IsIAsyncEnumerable(type))
+            {
+                return _convertToStream.MakeGenericMethod(itemType).Invoke(null, new object[] { reader });
+            }
+
+            return reader;
+        }
+
+        private static async IAsyncEnumerable<T> ConvertStream<T>(ChannelReader<T> reader)
```
davidfowl: Why? Didn’t we decide you can call ReadAllAsync?

src/SignalR/clients/csharp/Client/test/FunctionalTests/Hubs.cs
mikaelm12: You can't just return the IAsyncEnumerable here

src/SignalR/server/Core/src/StreamTracker.cs
mikaelm12: So we can't actually. We need the type that we pass into the MakeGenericMethod call

src/SignalR/clients/csharp/Client/test/FunctionalTests/Hubs.cs
davidfowl: What am I missing?

mikaelm12: Nvm, I forgot to take out the async keyword when I tried this and the return errors that you need to use a yield return so that was a little misleading. Can definitely just return the source directly.

src/SignalR/clients/csharp/Client/test/FunctionalTests/HubConnectionTests.cs
anurse: We need enforced editorconfig to save us!

src/SignalR/clients/csharp/Client/test/FunctionalTests/Hubs.cs
```diff
@@ -45,6 +45,8 @@ public string GetCallerConnectionId()
 
         public ChannelReader<int> StreamEchoInt(ChannelReader<int> source) => TestHubMethodsImpl.StreamEchoInt(source);
 
+        public IAsyncEnumerable<int> StreamIAsyncConsumer(IAsyncEnumerable<int> source) => TestHubMethodsImpl.StreamIAsyncConsumer(source);
```
halter73: nit:

```suggestion
        public IAsyncEnumerable<int> StreamIAsyncConsumer(IAsyncEnumerable<int> source) => source;
```

And delete TestHubMethodsImpl. StreamIAsyncConsumer.

src/SignalR/clients/csharp/Client/test/FunctionalTests/HubConnectionTests.cs
```diff
@@ -357,6 +357,51 @@ await foreach (var streamValue in asyncEnumerable)
             }
         }
 
+        [Theory]
+        [MemberData(nameof(HubProtocolsAndTransportsAndHubPaths))]
+        [LogLevel(LogLevel.Trace)]
```
halter73: Why do we need trace logs for this test but not the whole FunctionalTests project?

src/SignalR/clients/csharp/Client/test/FunctionalTests/HubConnectionTests.cs
```diff
@@ -357,6 +357,51 @@ await foreach (var streamValue in asyncEnumerable)
             }
         }
 
+        [Theory]
+        [MemberData(nameof(HubProtocolsAndTransportsAndHubPaths))]
```
halter73: Why would this test be affected by hub paths? This is 15 variations! 2 hub protocols * 3 transport types * 3 paths - 3 variations since MessagePack and SSE are incompatible no matter the path.

And then to compound the problem, each variation waits at least 50ms in Task.Delays on top of the time needed to start and stop the server, start and stop the connection, etc...

Generally speaking, avoid variations in tests unless the variations really are interesting. If you're using the same MemberData for a bunch of different tests, that's usually a sign of way too many variations.

src/SignalR/clients/csharp/Client/test/FunctionalTests/HubConnectionTests.cs
```diff
@@ -357,6 +357,51 @@ await foreach (var streamValue in asyncEnumerable)
...
+                            yield return i;
+                            await Task.Delay(10);
+                        }
+                    }
+                    var stream = ClientStreamData();
+
+                    await connection.StartAsync().OrTimeout();
+                    var expectedValue = 0;
+                    var streamTo = 5;
+                    var asyncEnumerable = connection.StreamAsync<int>("StreamIAsyncConsumer", stream );
```
BrennanConroy:
```suggestion
                    var asyncEnumerable = connection.StreamAsync<int>("StreamIAsyncConsumer", stream);
```

src/SignalR/clients/csharp/Client/test/FunctionalTests/HubConnectionTests.cs
```diff
@@ -357,6 +357,51 @@ await foreach (var streamValue in asyncEnumerable)
...
+                        {
+                            yield return i;
+                            await Task.Delay(10);
+                        }
+                    }
+                    var stream = ClientStreamData();
+
+                    await connection.StartAsync().OrTimeout();
+                    var expectedValue = 0;
+                    var streamTo = 5;
```
BrennanConroy: Use this in the local function

src/SignalR/clients/csharp/Client/test/FunctionalTests/HubConnectionTests.cs
mikaelm12: This particular test doesn't need trace logs. We just following the surrounding tests.

halter73: @JunTaoLuo Do you know how to change the default min log level for the entire test project?

src/SignalR/clients/csharp/Client/test/FunctionalTests/Hubs.cs
mikaelm12: Good catch. This method on all the hubs can be simplified now.

src/SignalR/common/Shared/ReflectionHelper.cs
```diff
@@ -38,6 +38,11 @@ public static bool IsStreamingType(Type type, bool mustBeDirectType = false)
 #if NETCOREAPP3_0
         public static bool IsIAsyncEnumerable(Type type)
         {
+            if (type.IsGenericType)
```
BrennanConroy: Does the current way not detect `IAsncEnumerable` correctly?

halter73: Correct. Previously, `ReflectionHelper.IsIAsyncEnumerable(typeof(IAsyncEnumerable<Foo>))` would return false because `typeof(IAsyncEnumerable<Foo>).GetInterfaces()` returns an empty array.

We mostly used to call IsIAsyncEnumerable with the result of object.GetType() prior to this PR to find streaming parameters in server-to-client invocations. In that case, the type would have to be concrete, and the old version of IsIAsyncEnumerable did the job.

This was and is still [called indirectly from HubMethodDescriptor](https://github.com/aspnet/AspNetCore/blob/4ddf91c09261436bcca1c0de2bb56269b6c229d4/src/SignalR/server/Core/src/Internal/HubMethodDescriptor.cs), and there the type could be an interface, so this had to be fixed to make IAsyncEnumerable hub parameters to work.

src/SignalR/clients/csharp/Client/test/FunctionalTests/HubConnectionTests.cs
mikaelm12: I will agree this is a slippery slope particularly because of the delays in this test. I changed it to just vary over protocols, fixing the hub path at default and transport to websockets (was considering varying over all transports as well though.)

src/SignalR/server/Core/src/StreamTracker.cs
```diff
@@ -21,11 +21,19 @@ internal class StreamTracker
         /// <summary>
         /// Creates a new stream and returns the ChannelReader for it as an object.
         /// </summary>
-        public object AddStream(string streamId, Type itemType)
+        public object AddStream(string streamId, Type itemType, Type type)
```
davidfowl: Change type to targetType

src/SignalR/server/Core/src/StreamTracker.cs
```diff
@@ -21,11 +21,19 @@ internal class StreamTracker
...
+
             var newConverter = (IStreamConverter)_buildConverterMethod.MakeGenericMethod(itemType).Invoke(null, Array.Empty<object>());
             _lookup[streamId] = newConverter;
-            return newConverter.GetReaderAsObject();
+            var reader = newConverter.GetReaderAsObject(ReflectionHelper.IsIAsyncEnumerable(type));
+
+            return reader;
+        }
+
+        private static IAsyncEnumerable<T> ConvertStream<T>(ChannelReader<T> reader)
```
davidfowl: Can be deleted

src/SignalR/server/Core/src/StreamTracker.cs
```diff
@@ -79,7 +87,7 @@ private static IStreamConverter BuildStream<T>()
         private interface IStreamConverter
         {
             Type GetItemType();
-            object GetReaderAsObject();
+            object GetReaderAsObject(bool isIAsyncEnumerable = false);
```
davidfowl: Might be cleaner to pass the target type here and do the IsAsyncEnumerable check on the inside

src/SignalR/clients/csharp/Client/test/FunctionalTests/Hubs.cs
```diff
@@ -45,6 +45,9 @@ public string GetCallerConnectionId()
 
         public ChannelReader<int> StreamEchoInt(ChannelReader<int> source) => TestHubMethodsImpl.StreamEchoInt(source);
 
+
```
davidfowl: nit: remove extra space

src/SignalR/common/Shared/ReflectionHelper.cs
davidfowl: Might be worth added some tests for this

src/SignalR/clients/csharp/Client/test/FunctionalTests/HubConnectionTests.cs
BrennanConroy: Didn't VS2019 get something like this that we can check in to github?

src/SignalR/clients/csharp/Client/test/FunctionalTests/HubConnectionTests.cs
BrennanConroy: We only really need Trace when the test is flaky and we're gathering more data. By default the level is Debug, I think it's fine to remove this and keep the default.

src/SignalR/clients/csharp/Client/test/FunctionalTests/HubConnectionTests.cs
```diff
@@ -357,6 +357,53 @@ await foreach (var streamValue in asyncEnumerable)
             }
         }
 
+        [Theory]
+        [InlineData("json")]
+        [InlineData("messagepack")]
+        [LogLevel(LogLevel.Trace)]
+        public async Task IAsyncEnumerableServerParam(string protocolName)
```
BrennanConroy: nit: Change the test name to something more test-like

src/SignalR/server/SignalR/test/Internal/ReflectionHelperTests.cs
```diff
@@ -0,0 +1,55 @@
...
+            async IAsyncEnumerable<int> Stream()
+            {
+                await Task.Delay(10);
+                yield return 1;
+            }
+
+            object streamAsObject = Stream();
+            yield return new object[]
+            {
+                streamAsObject.GetType(),
```
BrennanConroy: Not sure if this is any different than the `typeof(IAsyncEnumerable<object>)` test. You should make a class that implements `IAsyncEnumerable<T>` and test against that.

src/SignalR/server/SignalR/test/Internal/ReflectionHelperTests.cs
```diff
@@ -0,0 +1,55 @@
...
+using System.Threading.Channels;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Microsoft.AspNetCore.SignalR.Tests.Internal
+{
+    public class ReflectionHelperTests
+    {
+        [Theory]
+        [MemberData(nameof(GetPersonFromDataGenerator))]
```
BrennanConroy: Where did "Person" come from?

src/SignalR/server/SignalR/test/Internal/ReflectionHelperTests.cs
mikaelm12: This covers the second part of the `IsAsyncEnumerable` check. `IsGenericType` returns false for this.
But I do agree that we should add another case for a class that implements `IAsyncEnumerable<T>`

src/SignalR/server/SignalR/test/Internal/ReflectionHelperTests.cs
mikaelm12: ignore. Copy pasta

davidfowl (approved):

halter73 (approved):

BrennanConroy (approved):

PR #9774, opened 2019-04-26
Make endpoint routing allocation free in common scenarios

### LINKED ISSUE #6596: First class IEndpointFeature and IRoutingFeature in Kestrel
Today Kestrel has a fast path to avoid dictionary lookups by exposing features as fields on the it's IFeatureCollection implementation. We need to add the very commonly used IEndpointFeature and IRoutingFeature to this set this set:

https://github.com/aspnet/AspNetCore/blob/14e606997f8ef409f0ead89c0eda972bd09f96ab/src/Servers/Kestrel/tools/CodeGenerator/HttpProtocolFeatureCollection.cs#L22

cc @rynowak @JamesNK

davidfowl:
- This change tries to remove the EndpointSelectorContext allocation by making it a wrapper struct over the HttpContext. Unlike before, the HttpContext gets mutated once any component in the routnig pipeline sets a non null endpoint. This used to happen after the processing was complete.
- This change also implements the IRouteValuesFeature and IEndpointFeature in HttpProtocol to avoid the feature allocation and feature collection version churn.

Open issues:
- The log that checks if the endpoit is non null is wrong. If code sets an end point manually before routing runs it'll say match success. We'd need to compare the end point before and after routing to see if it changed to be accurate.
- Currently I removed the IRoutingFeature (partially why this is a draft). We need to push it into the HttpRequest feature in order to make it lazy. We should only be allocating this thing if it's requested. I know @rynowak wants it to go away but maybe it can't right now?
- I didn't do thourough testing yet, it's late and I'll just wait for the CI to explode.

Plaintext without logging now indistinguishable from middleware from an allocation profile:

image

Fixes https://github.com/aspnet/AspNetCore/issues/6596

### DIFF
--- a/src/Http/Routing.Abstractions/src/RoutingHttpContextExtensions.cs
+++ b/src/Http/Routing.Abstractions/src/RoutingHttpContextExtensions.cs
@@ -25,14 +25,7 @@ namespace Microsoft.AspNetCore.Routing
             }
 
             var routingFeature = httpContext.Features.Get<IRoutingFeature>();
-
-            if (routingFeature == null)
-            {
-                // REVIEW: Handle data tokens somehow?
-                return new RouteData(httpContext.Request.RouteValues);
-            }
-
-            return routingFeature.RouteData;
+            return routingFeature?.RouteData ?? new RouteData(httpContext.Request.RouteValues);
         }
 
         /// <summary>
@@ -54,8 +47,7 @@ namespace Microsoft.AspNetCore.Routing
                 throw new ArgumentNullException(nameof(key));
             }
 
-            return httpContext.Features.Get<IRoutingFeature>()?.RouteData.Values[key] ??
-                   httpContext.Features.Get<IRouteValuesFeature>()?.RouteValues[key];
+            return httpContext.Features.Get<IRouteValuesFeature>()?.RouteValues[key];
         }
     }
 }
--- a/src/Http/Routing.Abstractions/src/RoutingHttpContextExtensions.cs
+++ b/src/Http/Routing.Abstractions/src/RoutingHttpContextExtensions.cs
@@ -24,8 +24,15 @@ namespace Microsoft.AspNetCore.Routing
                 throw new ArgumentNullException(nameof(httpContext));
             }
 
-            var routingFeature = httpContext.Features[typeof(IRoutingFeature)] as IRoutingFeature;
-            return routingFeature?.RouteData;
+            var routingFeature = httpContext.Features.Get<IRoutingFeature>();
+
+            if (routingFeature == null)
+            {
+                // REVIEW: Handle data tokens somehow?
+                return new RouteData(httpContext.Request.RouteValues);
+            }
+
+            return routingFeature.RouteData;
         }
 
         /// <summary>
--- a/src/Http/Routing.Abstractions/src/RoutingHttpContextExtensions.cs
+++ b/src/Http/Routing.Abstractions/src/RoutingHttpContextExtensions.cs
@@ -3,6 +3,7 @@
 
 using System;
 using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Features;
 
 namespace Microsoft.AspNetCore.Routing
 {
@@ -46,8 +47,8 @@ namespace Microsoft.AspNetCore.Routing
                 throw new ArgumentNullException(nameof(key));
             }
 
-            var routingFeature = httpContext.Features[typeof(IRoutingFeature)] as IRoutingFeature;
-            return routingFeature?.RouteData.Values[key];
+            return httpContext.Features.Get<IRoutingFeature>()?.RouteData.Values[key] ??
+                   httpContext.Features.Get<IRouteValuesFeature>()?.RouteValues[key];
         }
     }
 }
--- a/src/Http/Routing/perf/EndpointRoutingBenchmarkBase.cs
+++ b/src/Http/Routing/perf/EndpointRoutingBenchmarkBase.cs
@@ -131,10 +131,9 @@ namespace Microsoft.AspNetCore.Routing
         protected (HttpContext httpContext, RouteValueDictionary ambientValues) CreateCurrentRequestContext(
             object ambientValues = null)
         {
-            var feature = new EndpointSelectorContext { RouteValues = new RouteValueDictionary(ambientValues) };
+            
             var context = new DefaultHttpContext();
-            context.Features.Set<IEndpointFeature>(feature);
-            context.Features.Set<IRouteValuesFeature>(feature);
+            var feature = new EndpointSelectorContext(context) { RouteValues = new RouteValueDictionary(ambientValues) };
 
             return (context, feature.RouteValues);
         }
--- a/src/Http/Routing/perf/EndpointRoutingBenchmarkBase.cs
+++ b/src/Http/Routing/perf/EndpointRoutingBenchmarkBase.cs
@@ -45,7 +45,7 @@ namespace Microsoft.AspNetCore.Routing
             return CreateServices().GetRequiredService<DfaMatcherBuilder>();
         }
 
-        private protected static  int[] SampleRequests(int endpointCount, int count)
+        private protected static int[] SampleRequests(int endpointCount, int count)
         {
             // This isn't very high tech, but it's at least regular distribution.
             // We sort the route templates by precedence, so this should result in
@@ -129,7 +129,6 @@ namespace Microsoft.AspNetCore.Routing
         protected (HttpContext httpContext, RouteValueDictionary ambientValues) CreateCurrentRequestContext(
             object ambientValues = null)
         {
-            
             var context = new DefaultHttpContext();
             context.Request.RouteValues = new RouteValueDictionary(ambientValues);
 
--- a/src/Http/Routing/perf/EndpointRoutingBenchmarkBase.cs
+++ b/src/Http/Routing/perf/EndpointRoutingBenchmarkBase.cs
@@ -7,14 +7,12 @@ using System.Runtime.CompilerServices;
 using System.Text;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
-using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Routing.Matching;
 using Microsoft.AspNetCore.Routing.Patterns;
 using Microsoft.AspNetCore.Routing.Template;
 using Microsoft.AspNetCore.Routing.Tree;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
-using Microsoft.Extensions.ObjectPool;
 
 namespace Microsoft.AspNetCore.Routing
 {
@@ -133,9 +131,9 @@ namespace Microsoft.AspNetCore.Routing
         {
             
             var context = new DefaultHttpContext();
-            var feature = new EndpointSelectorContext(context) { RouteValues = new RouteValueDictionary(ambientValues) };
+            context.Request.RouteValues = new RouteValueDictionary(ambientValues);
 
-            return (context, feature.RouteValues);
+            return (context, context.Request.RouteValues);
         }
 
         protected void CreateOutboundRouteEntry(TreeRouteBuilder treeRouteBuilder, RouteEndpoint endpoint)
--- a/src/Http/Routing/perf/Matching/MatcherSingleEntryBenchmark.cs
+++ b/src/Http/Routing/perf/Matching/MatcherSingleEntryBenchmark.cs
@@ -18,8 +18,6 @@ namespace Microsoft.AspNetCore.Routing.Matching
         private Matcher _route;
         private Matcher _tree;
 
-        private EndpointSelectorContext _feature;
-
         [GlobalSetup]
         public void Setup()
         {
@@ -35,8 +33,6 @@ namespace Microsoft.AspNetCore.Routing.Matching
             _dfa = SetupMatcher(CreateDfaMatcherBuilder());
             _route = SetupMatcher(new RouteMatcherBuilder());
             _tree = SetupMatcher(new TreeRouterMatcherBuilder());
-
-            _feature = new EndpointSelectorContext();
         }
 
         private Matcher SetupMatcher(MatcherBuilder builder)
@@ -48,8 +44,8 @@ namespace Microsoft.AspNetCore.Routing.Matching
         [Benchmark(Baseline = true)]
         public async Task Baseline()
         {
-            var feature = _feature;
             var httpContext = Requests[0];
+            var feature = new EndpointSelectorContext(httpContext);
 
             await _baseline.MatchAsync(httpContext, feature);
             Validate(httpContext, Endpoints[0], feature.Endpoint);
@@ -58,8 +54,8 @@ namespace Microsoft.AspNetCore.Routing.Matching
         [Benchmark]
         public async Task Dfa()
         {
-            var feature = _feature;
             var httpContext = Requests[0];
+            var feature = new EndpointSelectorContext(httpContext);
 
             await _dfa.MatchAsync(httpContext, feature);
             Validate(httpContext, Endpoints[0], feature.Endpoint);
@@ -68,12 +64,11 @@ namespace Microsoft.AspNetCore.Routing.Matching
         [Benchmark]
         public async Task LegacyTreeRouter()
         {
-            var feature = _feature;
-
             var httpContext = Requests[0];
+            var feature = new EndpointSelectorContext(httpContext);
 
-            // This is required to make the legacy router implementation work with global routing.
-            httpContext.Features.Set<IEndpointFeature>(feature);
+            //// This is required to make the legacy router implementation work with global routing.
+            //httpContext.Features.Set<IEndpointFeature>(feature);
 
             await _tree.MatchAsync(httpContext, feature);
             Validate(httpContext, Endpoints[0], feature.Endpoint);
@@ -82,11 +77,11 @@ namespace Microsoft.AspNetCore.Routing.Matching
         [Benchmark]
         public async Task LegacyRouter()
         {
-            var feature = _feature;
             var httpContext = Requests[0];
+            var feature = new EndpointSelectorContext(httpContext);
 
             // This is required to make the legacy router implementation work with global routing.
-            httpContext.Features.Set<IEndpointFeature>(feature);
+            //httpContext.Features.Set<IEndpointFeature>(feature);
 
             await _route.MatchAsync(httpContext, feature);
             Validate(httpContext, Endpoints[0], feature.Endpoint);
--- a/src/Http/Routing/perf/Matching/MatcherSingleEntryBenchmark.cs
+++ b/src/Http/Routing/perf/Matching/MatcherSingleEntryBenchmark.cs
@@ -11,8 +11,6 @@ namespace Microsoft.AspNetCore.Routing.Matching
     // Just like TechEmpower Plaintext
     public partial class MatcherSingleEntryBenchmark : EndpointRoutingBenchmarkBase
     {
-        private const int SampleCount = 100;
-
         private BarebonesMatcher _baseline;
         private Matcher _dfa;
         private Matcher _route;
@@ -67,9 +65,6 @@ namespace Microsoft.AspNetCore.Routing.Matching
             var httpContext = Requests[0];
             var feature = new EndpointSelectorContext(httpContext);
 
-            //// This is required to make the legacy router implementation work with global routing.
-            //httpContext.Features.Set<IEndpointFeature>(feature);
-
             await _tree.MatchAsync(httpContext, feature);
             Validate(httpContext, Endpoints[0], feature.Endpoint);
         }
@@ -79,9 +74,6 @@ namespace Microsoft.AspNetCore.Routing.Matching
         {
             var httpContext = Requests[0];
             var feature = new EndpointSelectorContext(httpContext);
-
-            // This is required to make the legacy router implementation work with global routing.
-            //httpContext.Features.Set<IEndpointFeature>(feature);
 
             await _route.MatchAsync(httpContext, feature);
             Validate(httpContext, Endpoints[0], feature.Endpoint);
--- a/src/Http/Routing/src/EndpointRoutingMiddleware.cs
+++ b/src/Http/Routing/src/EndpointRoutingMiddleware.cs
@@ -56,7 +56,7 @@ namespace Microsoft.AspNetCore.Routing
 
         public Task Invoke(HttpContext httpContext)
         {
-            var feature = new EndpointSelectorContext();
+            var feature = new EndpointSelectorContext(httpContext);
 
             // There's an inherent race condition between waiting for init and accessing the matcher
             // this is OK because once `_matcher` is initialized, it will not be set to null again.
@@ -97,8 +97,6 @@ namespace Microsoft.AspNetCore.Routing
             {
                 // Set the endpoint feature only on success. This means we won't overwrite any
                 // existing state for related features unless we did something.
-                SetFeatures(httpContext, feature);
-
                 Log.MatchSuccess(_logger, feature);
             }
             else
@@ -107,15 +105,6 @@ namespace Microsoft.AspNetCore.Routing
             }
 
             return _next(httpContext);
-        }
-
-        private static void SetFeatures(HttpContext httpContext, EndpointSelectorContext context)
-        {
-            // For back-compat EndpointSelectorContext implements IEndpointFeature,
-            // IRouteValuesFeature and IRoutingFeature
-            httpContext.Features.Set<IRoutingFeature>(context);
-            httpContext.Features.Set<IRouteValuesFeature>(context);
-            httpContext.Features.Set<IEndpointFeature>(context);
         }
 
         // Initialization is async to avoid blocking threads while reflection and things
--- a/src/Http/Routing/src/EndpointRoutingMiddleware.cs
+++ b/src/Http/Routing/src/EndpointRoutingMiddleware.cs
@@ -58,50 +58,68 @@ namespace Microsoft.AspNetCore.Routing
         {
             var feature = new EndpointSelectorContext(httpContext);
 
+            // Store the previous endpoint before matching starts, this lets us restore whatever endpoint was
+            // set at the end of matching if nothing matched
+            var previousEndpoint = feature.Endpoint;
+
+            // We need to set the endpoint to null here so that we can observe any mutations
+            // made by the matcher
+            feature.Endpoint = null;
+
             // There's an inherent race condition between waiting for init and accessing the matcher
             // this is OK because once `_matcher` is initialized, it will not be set to null again.
             var matcherTask = InitializeAsync();
             if (!matcherTask.IsCompletedSuccessfully)
             {
-                return AwaitMatcher(this, httpContext, feature, matcherTask);
+                return AwaitMatcher(this, httpContext, feature, previousEndpoint, matcherTask);
             }
 
             var matchTask = matcherTask.Result.MatchAsync(httpContext, feature);
             if (!matchTask.IsCompletedSuccessfully)
             {
-                return AwaitMatch(this, httpContext, feature, matchTask);
+                return AwaitMatch(this, httpContext, feature, previousEndpoint, matchTask);
             }
 
-            return SetRoutingAndContinue(httpContext, feature);
+            return SetRoutingAndContinue(httpContext, feature, previousEndpoint);
 
             // Awaited fallbacks for when the Tasks do not synchronously complete
-            static async Task AwaitMatcher(EndpointRoutingMiddleware middleware, HttpContext httpContext, EndpointSelectorContext feature, Task<Matcher> matcherTask)
+            static async Task AwaitMatcher(EndpointRoutingMiddleware middleware,
+                                           HttpContext httpContext,
+                                           EndpointSelectorContext feature,
+                                           Endpoint previousEndpoint,
+                                           Task<Matcher> matcherTask)
             {
                 var matcher = await matcherTask;
                 await matcher.MatchAsync(httpContext, feature);
-                await middleware.SetRoutingAndContinue(httpContext, feature);
+                await middleware.SetRoutingAndContinue(httpContext, feature, previousEndpoint);
             }
 
-            static async Task AwaitMatch(EndpointRoutingMiddleware middleware, HttpContext httpContext, EndpointSelectorContext feature, Task matchTask)
+            static async Task AwaitMatch(EndpointRoutingMiddleware middleware,
+                                         HttpContext httpContext,
+                                         EndpointSelectorContext feature,
+                                         Endpoint previousEndpoint,
+                                         Task matchTask)
             {
                 await matchTask;
-                await middleware.SetRoutingAndContinue(httpContext, feature);
+                await middleware.SetRoutingAndContinue(httpContext, feature, previousEndpoint);
             }
 
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        private Task SetRoutingAndContinue(HttpContext httpContext, EndpointSelectorContext feature)
+        private Task SetRoutingAndContinue(HttpContext httpContext, EndpointSelectorContext feature, Endpoint previousEndpoint)
         {
-            if (feature.Endpoint != null)
+            // Restore the previous endpoint (which might be null) if it's not set
+            feature.Endpoint ??= previousEndpoint;
+
+            // If there was no mutation of the endpoint then log failure
+            if (ReferenceEquals(feature.Endpoint, previousEndpoint))
             {
-                // Set the endpoint feature only on success. This means we won't overwrite any
-                // existing state for related features unless we did something.
-                Log.MatchSuccess(_logger, feature);
+                Log.MatchFailure(_logger);
             }
             else
             {
-                Log.MatchFailure(_logger);
+                Log.MatchSuccess(_logger, feature);
             }
 
             return _next(httpContext);
--- a/src/Http/Routing/src/EndpointRoutingMiddleware.cs
+++ b/src/Http/Routing/src/EndpointRoutingMiddleware.cs
@@ -6,7 +6,6 @@ using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
-using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Routing.Matching;
 using Microsoft.Extensions.Logging;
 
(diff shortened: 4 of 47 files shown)


src/Http/Routing/test/UnitTests/EndpointSelectorContextTest.cs
```diff
@@ -9,51 +9,52 @@
...
-                    "test"),
-            };
-
-            // Act
-            var routeData = ((IRoutingFeature)context).RouteData;
-
-            // Assert
-            Assert.Empty(routeData.DataTokens);
-        }
+        //[Fact]
```
davidfowl: TBD

src/Http/Routing/src/EndpointRoutingMiddleware.cs
```diff
@@ -97,8 +97,6 @@ private Task SetRoutingAndContinue(HttpContext httpContext, EndpointSelectorCont
             {
                 // Set the endpoint feature only on success. This means we won't overwrite any
                 // existing state for related features unless we did something.
-                SetFeatures(httpContext, feature);
-
                 Log.MatchSuccess(_logger, feature);
```
davidfowl: This isn't right

src/Http/Routing/src/EndpointSelectorContext.cs
```diff
@@ -3,68 +3,52 @@
...
+            }
             set
             {
-                _routeValues = value;
-
-                // RouteData will be created next get with new Values
-                _routeData = null;
+                if (value != null)
+                {
+                    _httpContext.SetEndpoint(value);
```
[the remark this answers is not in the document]
JamesNK: Ah never mind, you've added it to HttpProtocol.

src/Http/Routing/src/EndpointSelectorContext.cs
```diff
@@ -3,68 +3,52 @@
 
 using System;
 using Microsoft.AspNetCore.Http;
-using Microsoft.AspNetCore.Http.Features;
+using Microsoft.AspNetCore.Http.Endpoints;
 
 namespace Microsoft.AspNetCore.Routing
 {
-    public sealed class EndpointSelectorContext : IEndpointFeature, IRouteValuesFeature, IRoutingFeature
+    public struct EndpointSelectorContext
```
JamesNK: This will break anyone who has written and distributed an endpoint selector. Can you think of any other side effects @rynowak?

davidfowl: I’m like 87% sure we’re ok breaking that.

rynowak: I think we should break it. There are a few other things we should think about breaking.

rynowak:
> Currently I removed the IRoutingFeature (partially why this is a draft). We need to push it into the HttpRequest feature in order to make it lazy. We should only be allocating this thing if it's requested. I know @rynowak wants it to go away but maybe it can't right now?

I think we should lean into this more. MVC could produce a `RouteData` eagerly or on-demand.

src/Http/Routing/perf/Matching/MatcherSingleEntryBenchmark.cs
```diff
@@ -68,12 +64,11 @@ public async Task Dfa()
...
         {
-            var feature = _feature;
-
             var httpContext = Requests[0];
+            var feature = new EndpointSelectorContext(httpContext);
 
-            // This is required to make the legacy router implementation work with global routing.
-            httpContext.Features.Set<IEndpointFeature>(feature);
+            //// This is required to make the legacy router implementation work with global routing.
+            //httpContext.Features.Set<IEndpointFeature>(feature);
```
rynowak: I think this can just be removed right? Now this feature is always present.

davidfowl: Yea I wasn’t sure at first

rynowak: It's ok. we all feel that way sometimes.

src/Http/Routing/src/EndpointRoutingMiddleware.cs
rynowak: Well, it used to be lel

davidfowl:
> I think we should lean into this more. MVC could produce a RouteData eagerly or on-demand. I want to understand what would happen if we did something other than implementing it on the context directly.

What do you want to do?

src/Http/Routing/src/EndpointRoutingMiddleware.cs
```diff
@@ -109,15 +107,6 @@ private Task SetRoutingAndContinue(HttpContext httpContext, EndpointSelectorCont
             return _next(httpContext);
         }
 
-        private static void SetFeatures(HttpContext httpContext, EndpointSelectorContext context)
-        {
-            // For back-compat EndpointSelectorContext implements IEndpointFeature,
-            // IRouteValuesFeature and IRoutingFeature
-            httpContext.Features.Set<IRoutingFeature>(context);
-            httpContext.Features.Set<IRouteValuesFeature>(context);
-            httpContext.Features.Set<IEndpointFeature>(context);
```
rynowak: I'm really excited to see this disappear. I expect this to be a huge perf win.

src/Http/Routing/src/EndpointSelectorContext.cs
rynowak: I think that set of people is probably just @commonsensesoftware - either way we'll announcement it.

src/Http/Routing/src/EndpointSelectorContext.cs
```diff
@@ -3,68 +3,52 @@
...
+        private HttpContext _httpContext;
 
-        /// <summary>
-        /// Gets or sets the selected <see cref="Http.Endpoint"/> for the current
-        /// request.
-        /// </summary>
-        public Endpoint Endpoint { get; set; }
+        public EndpointSelectorContext(HttpContext httpContext)
+        {
+            _httpContext = httpContext;
```
rynowak: nullified check

src/Http/Routing/src/EndpointSelectorContext.cs
```diff
@@ -3,68 +3,52 @@
...
+            {
+                return _httpContext.GetEndpoint();
+            }
             set
             {
-                _routeValues = value;
-
-                // RouteData will be created next get with new Values
-                _routeData = null;
+                if (value != null)
```
rynowak: This should still mutate the context. It needs to be possible to set it to null.

src/Http/Routing/src/EndpointRoutingMiddleware.cs
rynowak: Yah, to keep the same behaviour you'd have to save and restore endpoint and rvd. 

OR 

we could dicuss this more and see what the right behaviour is when you ahve multiple routes. I know Jimmy is particularly passionate about this topic, but I made us just do something simple.

src/Http/Routing/src/EndpointSelectorContext.cs
```diff
@@ -3,68 +3,52 @@
...
-                    {
-                        var dataTokens = _routeData.DataTokens;
-                        foreach (var kvp in dataTokensMetadata.DataTokens)
-                        {
-                            _routeData.DataTokens.Add(kvp.Key, kvp.Value);
-                        }
-                    }
-                }
-
-                return _routeData;
```
rynowak: yeah, looking at this code, I think I'd feel really good about making `RouteData` lazy-init in MVC.

src/Http/Routing/test/UnitTests/EndpointSelectorContextTest.cs
rynowak: hue

rynowak: I'd imagine the best thing to do would be to make this part of `ActionContext` - and basically say that MVC polyfills this for ER and nothing else.

rynowak: I can send a commit if you want to see what that looks like.

src/Http/Routing/test/UnitTests/Matching/TreeRouterMatcherBuilder.cs
```diff
@@ -99,11 +99,7 @@ public VirtualPathData GetVirtualPath(VirtualPathContext context)
 
             public async Task RouteAsync(RouteContext routeContext)
             {
-                var context = (EndpointSelectorContext)routeContext.HttpContext.Features.Get<IEndpointFeature>();
-
-                // This is needed due to a quirk of our tests - they reuse the endpoint feature.
-                context.Endpoint = null;
```
rynowak: Still needed eh?

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.FeatureCollection.cs
```diff
@@ -18,16 +19,18 @@
...
+                                          IRequestBodyPipeFeature,
+                                          IHttpUpgradeFeature,
+                                          IHttpConnectionFeature,
+                                          IHttpRequestLifetimeFeature,
+                                          IHttpRequestIdentifierFeature,
+                                          IHttpBodyControlFeature,
+                                          IHttpMaxRequestBodySizeFeature,
+                                          IHttpResponseStartFeature,
+                                          IEndpointFeature,
+                                          IRouteValuesFeature
```
rynowak: <3

rynowak:
I'm really surprised and pleased how streamlined this is.

rynowak:
> 
> 
> > I think we should lean into this more. MVC could produce a RouteData eagerly or on-demand. I want to understand what would happen if we did something other than implementing it on the context directly.
> 
> What do you want to do?

I put a suggestion elsewhere while I was reading through the PR - but I think it's clearer to me now.

I would have MVC create the `RouteData` and then pass it in here: https://github.com/aspnet/AspNetCore/blob/master/src/Mvc/Mvc.Abstractions/src/ActionContext.cs#L64

This will result in an allocation for now. But we have two paths forward:
- Try to make RouteData lazy-init inside MVC (might be hard)
- Wait until MVC removes support for `IRouter` to remove it. 

We would announce/document that Endpoint Routing doesn't use `IRouteDataFeature`. If we have extension methods that interact with that, we should obsolete them now or in 4.0.

rynowak:
Here's a really stupid question. Given that this is just a wrapper for HttpContext, why does it need to exist at all. Since we're mutating HttpContext in place I think there's just no purpose to it.

JamesNK:
I does provide clean API for people to set routing stuff onto the context. For example, SetEndpoint is an extension method that lives in its own namespace. It isn't easily discoverable.

rynowak:
> I does provide clean API for people to set routing stuff onto the context. For example, SetEndpoint is an extension method that lives in its own namespace. It isn't easily discoverable. 

We could change that tho?

rynowak:
@davidfowl 4e4538407a05de27d6c6a9c58a41b5ec0747ac1d

src/Http/Routing/test/UnitTests/Matching/TreeRouterMatcherBuilder.cs
davidfowl: Is it? I couldn't figure that out.

davidfowl:
> I does provide clean API for people to set routing stuff onto the context. For example, SetEndpoint is an extension method that lives in its own namespace. It isn't easily discoverable.

I was considering lifting Endpoint to `HttpContext`, I wasn't sure though. Feels like it might be crossing the generic HttpContext line (maybe I'm just paranoid).

src/Http/Routing/test/UnitTests/EndpointSelectorContextTest.cs
davidfowl: Make a change and I can cherry pick it (like I did with your other change).

davidfowl:
> Here's a really stupid question. Given that this is just a wrapper for HttpContext, why does it need to exist at all. Since we're mutating HttpContext in place I think there's just no purpose to it.

The more I look at this change the more I feel like we should just pass the HttpContext instead of having the EndpointSelectorContext.

rynowak:
> I does provide clean API for people to set routing stuff onto the context. For example, SetEndpoint is an extension method that lives in its own namespace. It isn't easily discoverable.

I should have made this clear, but I guess I'm not optimizing for it being easy. It's somewhat rare to set these properties in a policy, it's for the short-circuit case.

src/Http/Routing/src/Matching/DfaMatcher.cs
```diff
@@ -34,11 +34,6 @@ public sealed override Task MatchAsync(HttpContext httpContext, EndpointSelector
                 throw new ArgumentNullException(nameof(httpContext));
             }
 
-            if (context == null)
-            {
-                throw new ArgumentNullException(nameof(context));
-            }
-
```
rynowak: Take a look here: https://github.com/aspnet/AspNetCore/pull/9774/files#diff-1473cfa07c9b9a3090e60b371891a1f4L310 This has to do something different now to process short-circuits.

src/Http/Routing/test/UnitTests/Matching/TreeRouterMatcherBuilder.cs
davidfowl: Yep

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
```diff
@@ -326,6 +329,7 @@ public void Reset()
         {
             _onStarting?.Clear();
             _onCompleted?.Clear();
+            _routeValues?.Clear();
```
davidfowl: @rynowak and @JamesNK this opens the door to re-using RouteValues but we'd need to make changes to routing to avoid reassiging this field and instead just adding values. It doesn't need to happen in this PR though but give it some thought.

src/Http/Routing/test/testassets/RoutingWebSite/UseEndpointRoutingStartup.cs
```diff
@@ -173,7 +173,7 @@ private void SetupBranch(IApplicationBuilder app, string name)
             app.UseRouting();
             app.UseEndpoints(endpoints =>
             {
-                endpoints.MapGet("api/get/{id}", (context) => context.Response.WriteAsync($"{name} - API Get {context.GetRouteData().Values["id"]}"));
+                endpoints.MapGet("api/get/{id}", (context) => context.Response.WriteAsync($"{name} - API Get {context.Request.RouteValues["id"]}"));
```
davidfowl: @rynowak RouteData is null now.

src/Http/Routing/src/EndpointSelectorContext.cs
davidfowl: Marked the PR as breaking change

src/Http/Routing.Abstractions/src/RoutingHttpContextExtensions.cs
```diff
@@ -46,8 +47,8 @@ public static object GetRouteValue(this HttpContext httpContext, string key)
                 throw new ArgumentNullException(nameof(key));
             }
 
-            var routingFeature = httpContext.Features[typeof(IRoutingFeature)] as IRoutingFeature;
-            return routingFeature?.RouteData.Values[key];
+            return httpContext.Features.Get<IRoutingFeature>()?.RouteData.Values[key] ??
```
davidfowl: @rynowak do we want this to work? Is this a legacy method?

src/Http/Routing/test/UnitTests/EndpointRoutingMiddlewareTest.cs
```diff
@@ -61,52 +61,6 @@ public async Task Invoke_OnCall_WritesToConfiguredLogger()
             Assert.Equal(expectedMessage, write.State?.ToString());
         }
 
-        [Fact]
-        public async Task Invoke_BackCompatGetRouteValue_ValueUsedFromEndpointFeature()
```
davidfowl: These were testing interop between RouteData on the IRoutingFeature and IRouteValuesFeature. These might still be valid but we need to make a decision about RouteData first.

src/Http/Routing.Abstractions/src/RoutingHttpContextExtensions.cs
JamesNK: GetRouteValue/GetRouteData are legacy

src/Http/Routing/src/EndpointSelectorContext.cs
```diff
@@ -3,68 +3,49 @@
 
 using System;
 using Microsoft.AspNetCore.Http;
-using Microsoft.AspNetCore.Http.Features;
+using Microsoft.AspNetCore.Http.Endpoints;
 
 namespace Microsoft.AspNetCore.Routing
 {
-    public sealed class EndpointSelectorContext : IEndpointFeature, IRouteValuesFeature, IRoutingFeature
+    public struct EndpointSelectorContext
```
JamesNK: readonly 😄

src/Http/Routing/src/EndpointSelectorContext.cs
```diff
@@ -3,68 +3,49 @@
...
     {
-        private RouteData _routeData;
-        private RouteValueDictionary _routeValues;
+        private HttpContext _httpContext;
```
JamesNK: readonly 😄

src/Http/Routing/test/UnitTests/EndpointSelectorContextTest.cs
```diff
@@ -1,9 +1,8 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System.Linq;
 using Microsoft.AspNetCore.Http;
-using Microsoft.AspNetCore.Routing.Matching;
+using Microsoft.AspNetCore.Http.Endpoints;
```
JamesNK: This Get/SetEndpoint + this namespace are new in 3.0. Not a breaking change if we want to move them

src/Http/Routing/test/testassets/RoutingWebSite/UseEndpointRoutingStartup.cs
JamesNK: It should be created lazily if GetRouteData is called for backwards compat

src/Http/Routing/src/EndpointSelectorContext.cs
davidfowl: The properties are settable.

src/Http/Routing/test/UnitTests/EndpointSelectorContextTest.cs
davidfowl: Not as part of this PR, that'll affect alot more code than I want to touch

src/Http/Routing.Abstractions/src/RoutingHttpContextExtensions.cs
davidfowl: Does that mean GetRouteValue (which is visible by default) doesn't work when endpoint routing runs?

src/Http/Routing/src/EndpointSelectorContext.cs
davidfowl: Hmm, seems to not complain.

src/Http/Routing/test/testassets/RoutingWebSite/UseEndpointRoutingStartup.cs
davidfowl: Maybe, I added a change that news it up over the current route values. We'll need to decide what to do here all up.

src/Http/Routing/src/EndpointSelectorContext.cs
JamesNK: It's the struct fields that matter. The properties are modifying HttpContext and friends.

src/Http/Routing.Abstractions/src/RoutingHttpContextExtensions.cs
JamesNK: It worked because the RouteData was created when requested and wrapped the route values - https://github.com/aspnet/AspNetCore/blob/master/src/Http/Routing/src/EndpointSelectorContext.cs#L43-L68

GetRouteValue should be changed to get the value from httpContext.Request.RouteValues to avoid allocating a RouteData.

src/Http/Routing/src/EndpointSelectorContext.cs
davidfowl: Yep

src/Http/Routing.Abstractions/src/RoutingHttpContextExtensions.cs
davidfowl: OK I made some changes.  The last thing to figure out is data tokens. I have some thoughts.

src/Http/Routing/test/testassets/RoutingWebSite/UseEndpointRoutingStartup.cs
davidfowl: This is fixed but I'm going to leave the 3.0 way in here

davidfowl:
@rynowak so GetRouteValue will continue to work forever. It's now based on the new `IRouteValuesFeature`. I updated the old routing middleware to set the RouteValues on the request in the middleware (tiny change). `GetRouteData` will either get the `IRoutingFeature` or it will new up one on the fly over the `HttpRequest.RouteValues`.

DataTokens is the last feature we need to figure out. After looking at the code a bit, help me understand why we need to preserve that feature in the endpoint routing MVC case? Is it just to preserve some back compat?

src/Http/Routing/perf/EndpointRoutingBenchmarkBase.cs
```diff
@@ -131,12 +129,11 @@ protected RouteEndpoint CreateEndpoint(string template, string httpMethod)
         protected (HttpContext httpContext, RouteValueDictionary ambientValues) CreateCurrentRequestContext(
             object ambientValues = null)
         {
-            var feature = new EndpointSelectorContext { RouteValues = new RouteValueDictionary(ambientValues) };
+
```
jo-ninja: nit: Remove blank line.
```suggestion
```

src/Http/Routing/src/EndpointRoutingMiddleware.cs
```diff
@@ -56,68 +56,75 @@ internal sealed class EndpointRoutingMiddleware
...
-            {
-                // Set the endpoint feature only on success. This means we won't overwrite any
-                // existing state for related features unless we did something.
-                SetFeatures(httpContext, feature);
+            // Restore the previous endpoint (which might be null) if it's not set
+            feature.Endpoint ??= previousEndpoint;
 
-                Log.MatchSuccess(_logger, feature);
+            // If there was no mutation of the endpoint then log failure
+            if (ReferenceEquals(feature.Endpoint, previousEndpoint))
```
JamesNK: I think this is wrong. The previous endpoint could be the same as the match endpoint, e.g. someone could:
1. Create a RoutingEndpoint in `UseEndpoints`
2. In middleware prior to routing they could fetch it from the endpoint datasource and set it on the feature
3. Routing middleware matches the endpoint already on feature

feature.Endpoint is set to null at the start of routing. If it is null at the end then there was no match. Check for null here and then restore the previousEndpoint inside the if.

src/Http/Routing/src/EndpointRoutingMiddleware.cs
```diff
@@ -56,68 +56,75 @@ internal sealed class EndpointRoutingMiddleware
...
-            var feature = new EndpointSelectorContext();
+            var feature = new EndpointSelectorContext(httpContext);
+
+            // Store the previous endpoint before matching starts, this lets us restore whatever endpoint was
+            // set at the end of matching if nothing matched
+            var previousEndpoint = feature.Endpoint;
+
+            // We need to set the endpoint to null here so that we can observe any mutations
+            // made by the matcher
+            feature.Endpoint = null;
```
JamesNK: Is there much perf impact with EndpointSelectorContext.Endpoint fetching the IEndpointFeature from HttpContext each time it gets and sets the endpoint?

What do you think of EndpointSelectorContext not wrapping HttpContext, and instead getting and setting the endpoint on the struct? The endpoint would then be set on HttpContext at the end of routing. Although would that require the struct be passed around by ref?

src/Http/Routing/src/EndpointRoutingMiddleware.cs
JamesNK: Add a unit test for this

src/Http/Routing.Abstractions/src/RoutingHttpContextExtensions.cs
```diff
@@ -46,8 +47,7 @@ public static object GetRouteValue(this HttpContext httpContext, string key)
                 throw new ArgumentNullException(nameof(key));
             }
 
-            var routingFeature = httpContext.Features[typeof(IRoutingFeature)] as IRoutingFeature;
-            return routingFeature?.RouteData.Values[key];
+            return httpContext.Features.Get<IRouteValuesFeature>()?.RouteValues[key];
```
rynowak: This doesn't really work because you expect `RouteData.Values` to be the source of truth. This implementation allows them to be different collections.  

`RouteData` is actually pretty complex - it is used to push and pop state (by MVC) which won't trigger updates to `httpContext.Request.RouteValues`. Trying to do a faithful job here would require some changes inside `RouteData`. 

I'd suggest just leaving this API alone and even marking it obsolete. If you're using old stuff, do `RouteData.Values`. If you're doing new stuff use `httpContext.Request.RouteValues`. 

I think if we're going to do more here than just leave this alone it will need some more design work.

src/Http/Routing/src/EndpointRoutingMiddleware.cs
davidfowl: The struct cannot be passed by ref as all of the methods are async, I don’t think it’ll have any real performance impact and if we wanted to solve this, I’d do it by hoisting the Endpoint to an HttpContext property

rynowak:
> DataTokens is the last feature we need to figure out. After looking at the code a bit, help me understand why we need to preserve that feature in the endpoint routing MVC case? Is it just to preserve some back compat?

It's not really back compat. It's a feature that people use that we haven't tried to replace. Users expect to be able to attach a property bag to a *route* and then read those values later. I think the solution that I created in that one commit is good enough. 

Outside of MVC's scenarios I'd prefer to just leave `IRoutingFeature` and `DataTokens` alone so they can keep working the way they do today for that user.

rynowak:
> The more I look at this change the more I feel like we should just pass the HttpContext instead of having the EndpointSelectorContext.

Did you decide not to do this?

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
rynowak: Reusing route values will involves some surgery to replace/reuse the underlying array. The 'slots' optimization that's use to build the route values was a huge perf win, and that likely is more impactful is saving an alloc of the wrapper object. 

Put another way, building the dictionary using `Add` and other methods that set values in the dictionary is slow. Precomputing which indexes the values belong in is fast. So if we want to look at reusing RVD, we also have to look at manipulating its guts directly.

src/Http/Routing.Abstractions/src/RoutingHttpContextExtensions.cs
```diff
@@ -23,8 +24,8 @@ public static RouteData GetRouteData(this HttpContext httpContext)
                 throw new ArgumentNullException(nameof(httpContext));
             }
 
-            var routingFeature = httpContext.Features[typeof(IRoutingFeature)] as IRoutingFeature;
-            return routingFeature?.RouteData;
+            var routingFeature = httpContext.Features.Get<IRoutingFeature>();
+            return routingFeature?.RouteData ?? new RouteData(httpContext.Request.RouteValues);
```
rynowak: Should this assign `routingFeature.RouteData`?

src/Http/Routing.Abstractions/src/RoutingHttpContextExtensions.cs
rynowak: Now that I think about this more I'm objecting to it less. This only really causes problems when you use both sets of APIs (new and old) and use the old ones to set state, and the new ones to read it.

src/Http/Routing/src/EndpointRoutingMiddleware.cs
davidfowl: Sounds like an extreme edge case but so that I’m clear, pick the exact same instance that would have matched the endpoint and observe that the logs are bad yes?

src/Http/Routing/src/EndpointRoutingMiddleware.cs
```diff
@@ -56,68 +56,75 @@ internal sealed class EndpointRoutingMiddleware
...
-            return SetRoutingAndContinue(httpContext, feature);
+            return SetRoutingAndContinue(httpContext, feature, previousEndpoint);
 
             // Awaited fallbacks for when the Tasks do not synchronously complete
-            static async Task AwaitMatcher(EndpointRoutingMiddleware middleware, HttpContext httpContext, EndpointSelectorContext feature, Task<Matcher> matcherTask)
+            static async Task AwaitMatcher(EndpointRoutingMiddleware middleware,
+                                           HttpContext httpContext,
+                                           EndpointSelectorContext feature,
+                                           Endpoint previousEndpoint,
+                                           Task<Matcher> matcherTask)
```
rynowak: aligning parameters is awful, just let vs format it the default way

src/Http/Routing/src/EndpointRoutingMiddleware.cs
rynowak: This needs to preserve route values as well.

src/Http/Routing.Abstractions/src/RoutingHttpContextExtensions.cs
davidfowl: I dunno feels wrong and side effecty. I don’t want to create a feature here especially since it really isn’t faithful part of the future. Can you think of a downside to not setting the feature?

src/Http/Routing/src/EndpointRoutingMiddleware.cs
```diff
@@ -56,68 +56,75 @@ internal sealed class EndpointRoutingMiddleware
 
         public Task Invoke(HttpContext httpContext)
         {
-            var feature = new EndpointSelectorContext();
+            var feature = new EndpointSelectorContext(httpContext);
+
+            // Store the previous endpoint before matching starts, this lets us restore whatever endpoint was
+            // set at the end of matching if nothing matched
+            var previousEndpoint = feature.Endpoint;
```
rynowak: Is it worth a discussion about whether the current behaviour of the middleware is even desirable? For instance if the routing middleware no-op'ed when there was an endpoint does taht make more sense? 

I don't feel like we ever really answered these questions because we didn't have good scenarios in mind either way.

src/Http/Routing/src/EndpointRoutingMiddleware.cs
davidfowl: lol sure I prefer alignment but whatevs

src/Http/Routing.Abstractions/src/RoutingHttpContextExtensions.cs
davidfowl: Right, I think this API should just “work”. I’d the 3 collections get out of sync it’s be there was mutation involved and maybe we should update those places to also implement this feature to respect the mutation. I know this is all outside of the scope of this PR but I guess that’s why get for touching IRoutingFeature

davidfowl:
> It's not really back compat. It's a feature that people use that we haven't tried to replace. Users expect to be able to attach a property bag to a route and then read those values later. I think the solution that I created in that one commit is good enough.

> Outside of MVC's scenarios I'd prefer to just leave IRoutingFeature and DataTokens alone so they can keep working the way they do today for that user.

Either we think it’s useful and it should be hoisted to the IRouteValues feature (per route, per request, items bag) or its not and it’s only applicable when using IRouterBased routing or in Mvc.

Btw does MVC (I haven’t looked) have a heavy reliance on RouteData or is it mostly round tripping user state from route definition to action invocation?

src/Http/Routing/src/EndpointRoutingMiddleware.cs
davidfowl: I was actually thinking that was a better behavior.

davidfowl:
> Did you decide not to do this?

Yes I want to get this in with the current shape and iterate. Once we solve these big rocks and then come back to this later.

src/Http/Routing/src/EndpointRoutingMiddleware.cs
davidfowl: I'm going to implement this as it simplifies alot

src/Http/Routing.Abstractions/src/RoutingHttpContextExtensions.cs
rynowak: I guess I'm wondering about the case where you have a routing feature but route data is null. This method would return (but not store) a new RouteData object.

src/Http/Routing/src/EndpointRoutingMiddleware.cs
```diff
@@ -58,62 +57,49 @@ public Task Invoke(HttpContext httpContext)
...
 
-            // Store the previous endpoint before matching starts, this lets us restore whatever endpoint was
-            // set at the end of matching if nothing matched
-            var previousEndpoint = feature.Endpoint;
-
-            // We need to set the endpoint to null here so that we can observe any mutations
-            // made by the matcher
-            feature.Endpoint = null;
+            // There's already an endpoint, skip maching completely
+            if (feature.Endpoint != null)
```
JamesNK: Should be an option?

src/Http/Routing/src/EndpointRoutingMiddleware.cs
JamesNK: I'll be annoying and ask whether it should be an option

davidfowl: No.

src/Http/Routing.Abstractions/src/RoutingHttpContextExtensions.cs
davidfowl: > I guess I'm wondering about the case where you have a routing feature but route data is null. This method would return (but not store) a new RouteData object.

When would that happen?

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
davidfowl: It also means that CandidateState needs to expose the slots rather than the RVD https://github.com/aspnet/AspNetCore/blob/2b2aeee9f869ec43f65c188a66d9285858823628/src/Http/Routing/src/Matching/CandidateState.cs#L53. That complicates the API a bit.

JamesNK: RVD has a constructor to wrap the underlying array. RVD could be extended to provide a way to get the underlying array instance.

davidfowl: It would only need to allocate if there were complex segments or constraints:
 https://github.com/aspnet/AspNetCore/blob/2b2aeee9f869ec43f65c188a66d9285858823628/src/Http/Routing/src/Matching/DfaMatcher.cs#L163

https://github.com/aspnet/AspNetCore/blob/2b2aeee9f869ec43f65c188a66d9285858823628/src/Http/Routing/src/Matching/DfaMatcher.cs#L153

It seems like we need to pass RouteValues to legacy APIs that have them in the signature. Complex segments could be fixed it we cared but it's likely they are uncommon enough to not worry about (this is an internal API).

`IRouteConstraint` takes `RouteValues` as well (and has an `IRouter` in the signature....) so we need to allocate there.

In the other cases, we can make a new API like `FromArray(slots)` but an instance version (it would need to deal with edge cases though). Anyhow, thats future work.

src/Http/Routing.Abstractions/src/RoutingHttpContextExtensions.cs
davidfowl: I’ll make the change but I’m not sure when this would happen...

src/Http/Routing/src/EndpointRoutingMiddleware.cs
davidfowl: Does it need to preserve route values if we short circuit here?

src/Http/Routing/src/EndpointRoutingMiddleware.cs
rynowak: You're not being annoying - but it shouldn't be an option. This should work (and be documented) behaving in a single way.

src/Http/Routing/src/EndpointRoutingMiddleware.cs
```diff
@@ -93,31 +98,19 @@ static async Task AwaitMatch(EndpointRoutingMiddleware middleware, HttpContext h
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         private Task SetRoutingAndContinue(HttpContext httpContext, EndpointSelectorContext feature)
         {
-            if (feature.Endpoint != null)
+            // If there was no mutation of the endpoint then log failure
+            if (feature.Endpoint is null)
```
rynowak: ugh. `is null` is so objectionable.

src/Http/Routing/src/EndpointRoutingMiddleware.cs
```diff
@@ -56,7 +55,13 @@ internal sealed class EndpointRoutingMiddleware
 
         public Task Invoke(HttpContext httpContext)
         {
-            var feature = new EndpointSelectorContext();
+            var feature = new EndpointSelectorContext(httpContext);
+
+            // There's already an endpoint, skip maching completely
+            if (feature.Endpoint != null)
+            {
+                return _next(httpContext);
```
rynowak: This seems like this case should be logged.

src/Http/Routing/src/RouterMiddleware.cs
```diff
@@ -40,11 +40,16 @@ public async Task Invoke(HttpContext httpContext)
...
             else
             {
-                httpContext.Features[typeof(IRoutingFeature)] = new RoutingFeature()
+                var routingFeature = new RoutingFeature()
                 {
-                    RouteData = context.RouteData,
+                    RouteData = context.RouteData
                 };
 
+                // Set the RouteValues on the current request
```
rynowak: If you want extra #RyanPoints update this comment to say that this is being down to keep `IRoutingFeature` in sync with `IRouteValuesFeature`. If you want even more extra points include ascii at of N'sync

src/Http/Routing/test/UnitTests/EndpointRoutingMiddlewareTest.cs
```diff
@@ -36,6 +35,24 @@ public async Task Invoke_OnCall_SetsEndpointFeature()
             Assert.NotNull(endpointFeature);
         }
 
+        [Fact]
+        public async Task Invoke_SkipsRouting_IfEndpointSet()
```
rynowak: 👍

rynowak (approved):

src/Http/Routing/src/EndpointRoutingMiddleware.cs
davidfowl: We can make it an option when the masses complain (I mean @sebastienros )

src/Http/Routing/src/EndpointRoutingMiddleware.cs
davidfowl: is so good

src/Http/Routing/src/EndpointRoutingMiddleware.cs
JamesNK: Sure. I'm fine with it not being an option straight away, if ever.

src/Http/Routing/src/EndpointRoutingMiddleware.cs
davidfowl: Done

JamesNK (approved):

src/Http/Routing/src/EndpointSelectorContext.cs
commonsensesoftware: Thanks for the heads up. I looked at this change and it looks to be a non-issue for me. Best as I can tell, this will be 100% source compatible with what I already have. The only write scenario I have is explicitly setting the endpoint as a short-circuit for invalid client requests (ex: duplicate API versions, no API version, etc). It appears this change doesn't alter that behavior.

What will this mean for route constraints - anything? I believe today they are still called in _Endpoint Routing_, but supply `null` for the **IRouter** parameter. That doesn't affect me, but I was wondering if I should expect a possible change there as well. I only use a route constraint for URL segment versioning so I don't need to perform _magic string_ parsing on the URL to find where the version value is in the path.

rynowak: There's no change for `IRouteContraint`, they are used by both routing systems.

PR #9785, opened 2019-04-26
Placeholder RenderTreeBuilder.SetKey() to enable aspnetcore-tooling work

SteveSandersonMS:
I've done most of the compiler work for `key` in aspnetcore-tooling, but the tests can't pass yet because the output now tries to call `builder.SetKey(...)`, but it doesn't yet exist.

This small placeholder will make it possible to complete the PR in aspnetcore-tooling, after which I'll be able to add the real runtime logic in this repo.


javiercn (approved):
I trust what you are doing here.

rynowak (commented):
:shipit:

PR #9788, opened 2019-04-26
Avoid allocations in more cases

rynowak:
Updates to DFA Matcher to avoid allocations in more cases. This makes
the matcher more pay-for-play.

- Avoid allocating an RVD while matching if possible
- Avoid allocating the candidate set unless necessary

First, avoid creating the RVD unless there are parameters or
constraints. This means that the candidate set can contain null route
value dictionaries. This seems fine because selectors are already
low-level. The route values feature will allocate an RVD when accessed,
so code in MVC or middleware won't even see a null RVD.

Secondly, avoid creating the candidate set unless there are selectors.
This will save an allocation most of the time since we won't need to run
selectors is really common cases. The candidate set is needed because
selectors mix mutability and async, so it needs to be a reference type.
However the default case is that we *don't* need to run selectors. The
impact of this is that we make a bunch of methods have an instance
variant and a static variant that operates on the array.

------

I ran some perf tests on this, and it helps the case where you have multiple endpoints that match but no policies. We don't have any tests that directly benchmark this scenario, but it's somewhat common when you use parameters and literals in the same position, or use constraints.

Numbers from `GithubMatcherBenchmark` - this benchmark has some cases like this, and there's a marginal improvement across the aggregate.

**Before**

``` ini

BenchmarkDotNet=v0.10.13, OS=Windows 10.0.17763
Intel Core i7-8650U CPU 1.90GHz (Kaby Lake R), 1 CPU, 8 logical cores and 4 physical cores
.NET Core SDK=3.0.100-preview6-011546
  [Host]     : .NET Core 3.0.0-preview6-27623-04 (CoreCLR 4.6.27622.74, CoreFX 4.700.19.22212), 64bit RyuJIT
  Job-VZXIJQ : .NET Core 3.0.0-preview6-27623-04 (CoreCLR 4.6.27622.74, CoreFX 4.700.19.22212), 64bit RyuJIT

Runtime=Core  Server=True  Toolchain=.NET Core 3.0  
RunStrategy=Throughput  

```
|   Method |     Mean |    Error |   StdDev |        Op/s | Scaled | ScaledSD |   Gen 0 | Allocated |
|--------- |---------:|---------:|---------:|------------:|-------:|---------:|--------:|----------:|
| Baseline | 147.4 ns | 2.075 ns | 1.839 ns | 6,783,660.9 |   1.00 |     0.00 |  1.7090 |   9.49 KB |
|      Dfa | 354.7 ns | 4.589 ns | 3.583 ns | 2,819,073.1 |   2.41 |     0.04 | 10.8643 |  57.07 KB |

**After**

``` ini

BenchmarkDotNet=v0.10.13, OS=Windows 10.0.17763
Intel Core i7-8650U CPU 1.90GHz (Kaby Lake R), 1 CPU, 8 logical cores and 4 physical cores
.NET Core SDK=3.0.100-preview6-011546
  [Host]     : .NET Core 3.0.0-preview6-27623-04 (CoreCLR 4.6.27622.74, CoreFX 4.700.19.22212), 64bit RyuJIT
  Job-IRODAN : .NET Core 3.0.0-preview6-27623-04 (CoreCLR 4.6.27622.74, CoreFX 4.700.19.22212), 64bit RyuJIT

Runtime=Core  Server=True  Toolchain=.NET Core 3.0  
RunStrategy=Throughput  

```
|   Method |     Mean |    Error |   StdDev |        Op/s | Scaled | ScaledSD |  Gen 0 | Allocated |
|--------- |---------:|---------:|---------:|------------:|-------:|---------:|-------:|----------:|
| Baseline | 144.1 ns | 1.231 ns | 1.152 ns | 6,940,674.8 |   1.00 |     0.00 | 1.8311 |   9.49 KB |
|      Dfa | 338.6 ns | 2.982 ns | 2.490 ns | 2,953,168.6 |   2.35 |     0.02 | 9.5215 |  52.12 KB |


src/Http/HttpAbstractions.sln
```diff
@@ -113,7 +113,7 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.AspNetCore.HttpOv
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.AspNetCore.Server.IISIntegration", "..\Servers\IIS\IISIntegration\src\Microsoft.AspNetCore.Server.IISIntegration.csproj", "{1062FCDE-E145-40EC-B175-FDBCAA0C59A0}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.AspNetCore.WebUtilities.Performance", "WebUtilities\perf\Microsoft.AspNetCore.WebUtilities.Performance\Microsoft.AspNetCore.WebUtilities.Performance.csproj", "{21AC56E7-4E77-4B0E-B63E-C8E836E4D14E}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.AspNetCore.WebUtilities.Performance", "WebUtilities\perf\Microsoft.AspNetCore.WebUtilities.Performance\Microsoft.AspNetCore.WebUtilities.Performance.csproj", "{21AC56E7-4E77-4B0E-B63E-C8E836E4D14E}"
```
rynowak: VS did this. YOLO

src/Http/HttpAbstractions.sln
```diff
@@ -642,7 +642,7 @@ Global
 		{1A866315-5FD5-4F96-BFAC-1447E3CB4514} = {80A090C8-ED02-4DE3-875A-30DCCDBD84BA}
 		{068A1DA0-C7DF-4E3C-9933-4E79A141EFF8} = {80A090C8-ED02-4DE3-875A-30DCCDBD84BA}
 		{8C635944-51F0-4BB0-A89E-CA49A7D9BE7F} = {FB2DCA0F-EB9E-425B-ABBC-D543DBEC090F}
-		{1A74D674-5D19-4575-B443-8B7ED433EF2B} = {793FFE24-138A-4C3D-81AB-18D625E36230}
+		{1A74D674-5D19-4575-B443-8B7ED433EF2B} = {14A7B3DE-46C8-4245-B0BD-9AFF3795C163}
```
rynowak: I moved the routing benchmarks project into the right place in the solution structure.

src/Http/Routing/src/Matching/CandidateSet.cs
```diff
@@ -20,9 +20,7 @@ namespace Microsoft.AspNetCore.Routing.Matching
     /// </summary>
     public sealed class CandidateSet
     {
-        private const int BitVectorSize = 32;
```
rynowak: dead code

src/Http/Routing/src/Matching/CandidateSet.cs
```diff
@@ -20,9 +20,7 @@ namespace Microsoft.AspNetCore.Routing.Matching
...
-
-        private CandidateState[] _candidates;
+        internal CandidateState[] Candidates;
```
rynowak: DefaultEndpointSelector needs this

src/Http/Routing/src/Matching/CandidateSet.cs
```diff
@@ -142,8 +150,15 @@ public void SetValidity(int index, bool value)
                 ThrowIndexArgumentOutOfRangeException();
             }
 
-            ref var original = ref _candidates[index];
-            _candidates[index] = new CandidateState(original.Endpoint, original.Values, original.Score >= 0 ^ value ? ~original.Score : original.Score);
+            ref var original = ref Candidates[index];
```
rynowak: The pattern of all of these methods now is that there's an instance method that turns an index into a `ref CandidateState`, and a static one with all of the logic.

src/Http/Routing/src/Matching/DfaMatcher.cs
```diff
@@ -111,17 +111,13 @@ public sealed override Task MatchAsync(HttpContext httpContext, EndpointSelector
                 // candidate: readonly data about the endpoint and how to match
                 // state: mutable storarge for our processing
                 ref var candidate = ref candidates[i];
-                ref var state = ref candidateSet[i];
+                ref var state = ref candidateState[i];
+                state = new CandidateState(candidate.Endpoint, candidate.Score);
```
rynowak: This initialization used to take place in the CandidateSet constructor.

src/Http/Routing/src/Matching/CandidateSet.cs
```diff
@@ -142,8 +151,15 @@ public void SetValidity(int index, bool value)
...
+            SetValidatity(ref original, value);
+        }
+
+        internal static void SetValidatity(ref CandidateState candidate, bool value)
```
Eilon: Funny typo.

davidfowl (approved):
I approve this message

davidfowl:
What allocations are left? Just the selector context?

JamesNK (approved):

rynowak:
> What allocations are left? Just the selector context?

Selector context - AND the `CandiateState[]` that we allocate.

src/Http/Routing/src/Matching/CandidateSet.cs
rynowak: YEAH BRO

rynowak:
@aspnet-hello  triage https://dev.azure.com/dnceng/public/_build/results?buildId=171062

PR #9793, opened 2019-04-26
Header propagation sample and UX improvements

rynowak:
Creating a sample for the header propagation package and some various
misc UX improvements based on building the sample.

A small list:
- allow duplicate inbound names
- de-dupe based on outbound names
- add sugar for configuration
- simplify pattern for transforming values
- add error message for missing middleware

Also a few small perf things.

I started this out by wanting to remove the following from the
configuration pattern:
```C#
options.Headers.Add("X-TraceId", null);
```

This pattern with null is undiscoverable, but we didn't provide
something simpler. The most common case was to add a custom collection
type so we can define sugar methods.

The next realization is that in practical case (dist tracing sample) you
either way to *key* off of the same inbound header twice, or you don't
have an inbound header at all, and you will synthesize the value every
time. This means that the way we're treating inbound header names is a
bit wrong. We don't want inbound header names to be unique, we want
*outbound header names to be unique*.

Next, I want to consolidate DefaultValue and ValueFactory. The problems
I saw with this:
- DefaultValue is a trap. It's rare to use a static value.
- ValueFactory really wants the header name *and* value

I think what's there now is much more terse to work with.

/cc @Tratcher @davidfowl @alefranz


src/Middleware/HeaderPropagation/samples/HeaderPropagationSample/HeaderPropagationSample/Startup.cs
```diff
@@ -0,0 +1,102 @@
...
+            // 1. Forward a header as-is
+            // 2. Forward a header using a different header name
+            // 3. Generate a new header value, or conditionally generate a header value
+            services.AddHeaderPropagation(options =>
+            {
+                // Propagate the X-TraceId as a representation of the overall transaction.
+                options.Headers.Add("X-TraceId");
+
+                // Propagate the X-SpanId as X-ParentSpanId to represent the 'parent' of this HTTP call.
+                options.Headers.Add(new HeaderPropagationEntry("X-SpanId", "X-ParentSpanId"));
```
rynowak: I considered also adding *sugar* methods for these, but I think we'd want to get creative. The meaning of `Add(string)` is obvious, it's less so with this.  I'd be happy to consider suggestions 😀

src/Middleware/HeaderPropagation/src/HeaderPropagationEntry.cs
```diff
@@ -13,44 +12,72 @@ namespace Microsoft.AspNetCore.HeaderPropagation
...
+            }
+
+            if (outboundHeaderName == null)
+            {
+                throw new ArgumentNullException(nameof(outboundHeaderName));
+            }
+
+            InboundHeaderName = inboundHeaderName;
+            OutboundHeaderName = outboundHeaderName;
+        }
```
rynowak: Life got much simpler once these things are specified in the constructor.

src/Middleware/WebSockets/samples/EchoApp/Startup.cs
```diff
@@ -9,6 +9,7 @@
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
```
rynowak: This was failing to build.

src/Middleware/HeaderPropagation/samples/HeaderPropagationSample/HeaderPropagationSample/Program.cs
```diff
@@ -0,0 +1,21 @@
...
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.Hosting;
+
+namespace HeaderPropagationSample
+{
+    public class Program
+    {
+        public static void Main(string[] args)
+        {
+            var host = new WebHostBuilder()
```
Tratcher: We need to get into the habit of having samples use generic host, CreateDefaultBuilder, etc.. We no longer have cross-repo layering restrictions writing them, so they should all look like the templates.

src/Middleware/HeaderPropagation/samples/HeaderPropagationSample/HeaderPropagationSample/Startup.cs
Tratcher: If these are your three main use cases then I think they do warrant sugar. 
`options.Headers.Add(new HeaderPropagationEntry("X-SpanId", "X-ParentSpanId"));`
is no more specific about what the parameters mean than
`options.Headers.Add("X-SpanId", "X-ParentSpanId");`

It's not like you have a wide variety of overloads that would confuse the meaning of the second string parameter.

src/Middleware/HeaderPropagation/samples/HeaderPropagationSample/HeaderPropagationSample/Startup.cs
```diff
@@ -0,0 +1,102 @@
...
+
+                // Generate a new X-SpanId to represent this call, if it's being traced.
+                options.Headers.Add(new HeaderPropagationEntry("X-SpanId")
```
Tratcher: Worth a little sugar: `Add(string header, Func<Context, string> factory)`
```
options.Headers.Add("X-SpanId", context => 
  context.HttpContext.Request.Headers.ContainsKey("X-TraceId") ? Guid.NewGuid().ToString() : null);
```

src/Middleware/HeaderPropagation/samples/HeaderPropagationSample/HeaderPropagationSample/Startup.cs
davidfowl: I fear using this for distributed tracing works against the right way to do it (what we’re baking in). Are there any other reasons to use this besides tracing?

rynowak: > I fear using this for distributed tracing works against the right way to do it (what we’re baking in). Are there any other reasons to use this besides tracing?

I feel like we talked about this before, but I'm sure it's going to come up again every time we look at this code 😆. 

We're building-in one particular implementation of tracing - this is a general building block for *your stuff*. Real things you can use this for:
- Doing a different tracing thing than we're doing.
- Various routing/versioning strategies: https://istio.io/docs/tasks/traffic-management/request-routing/
- A/B testing with service -> service calls
- Fowarding things like a language for resources/messages

I'm happy to show another sample here if you feel like this makes life too confusing. I agree that our guidance should be to use the built-in tracing features once they are available.

src/Middleware/HeaderPropagation/samples/HeaderPropagationSample/HeaderPropagationSample/Program.cs
rynowak: Cool beans.

src/Middleware/HeaderPropagation/samples/HeaderPropagationSample/HeaderPropagationSample/Startup.cs
davidfowl: Yes please add any other sample,one of those sounds good. The problem with having a distributed tracing scenario is that it only 1/2 works. You need to also store things in logs and have other parts of the system use it (outside of ASP.NET
Core) which is why the Activity based imply works well.

src/Middleware/HeaderPropagation/src/HeaderPropagationMiddleware.cs
```diff
@@ -35,33 +35,43 @@ public HeaderPropagationMiddleware(RequestDelegate next, IOptions<HeaderPropagat
...
+            // Perf: avoid foreach since we don't define a struct enumerator.
+            var entries = _options.Headers;
+            for (var i = 0; i < entries.Count; i++)
             {
-                var values = GetValues(headerName, entry, context);
+                var entry = entries[i];
 
-                if (!StringValues.IsNullOrEmpty(values))
+                // We intentionally process entries in order, and allow earlier entries to
+                // take precedence over later entries when they have the same output name.
```
alefranz: Not sure what the convention is, but I would add a test to make sure this behavior stick as the comment can be missed.

src/Middleware/HeaderPropagation/samples/HeaderPropagationSample/HeaderPropagationSample/Startup.cs
alefranz: Hi Ryan,
glad we are on the same page on the possible use cases for this.
I guess adding all those scenarios in the example will help to not get this comment lost. let me know if I can help.


------
David - you are on point mentioning logs, as it is relevant for all the scenarios, not only for tracing. As I started to replace the current logic on applications with this package (as I published a backport package to 2.1-2.2), it has been the main pain point (on top of the clumsy setup which now Ryan got much better :) ).

I was originally tempted to include the ability to also add the header to the logging scope, but it seemed out of scope. However right now to add a value to the logging scope you either have to capture it from the `HttpContext` again on a separate middleware or, expecially when you rely on the factory, you have to access the `HeaderPropagationValues` (which hold these values) from a separate middleware to add those values to the logging scope, which feels wrong (I guess ideally it should be internal but then you will close this door).

Sorry for going off topic, but It would be really nice to have your opinion on this. Thank you.

src/Middleware/HeaderPropagation/src/HeaderPropagationMiddleware.cs
```diff
@@ -35,33 +35,43 @@ public HeaderPropagationMiddleware(RequestDelegate next, IOptions<HeaderPropagat
...
+                if (!headers.ContainsKey(entry.OutboundHeaderName))
                 {
-                    _values.Headers.TryAdd(headerName, values);
+                    var value = GetValue(context, entry);
+                    if (!StringValues.IsNullOrEmpty(value))
+                    {
+                        headers.Add(entry.OutboundHeaderName, value);
```
alefranz: I see you have now switched to store the header with the outbound header name instead of the inbound one.
If you remember I was torn on this in the original PR and I probably switch a couple times.

The reason why I settled on using the inbound name was to play better with the ability to specify the headers per client which we discussed to add later (but unfortunately I haven't had time to work on it yet but I would like to discuss it later so I can work on it).

So the idea is that given an header I receive (e.g. `Accept-Language`, I'm intentionally not going to use tracing as example :) ) I want to propagate it to my dependencies which unfortunately have different conventions (one wants `X-Language`, the other `Lang`) so the output name will be overridable and being able to then configure the clients in terms of the original name accepted by my API/website (as well as being able to find it in the `HeaderPropagationValues` for logging as mentioned in the other comment) seems the best option.

What do you think? what was your thought process that led to use the outbound name as key?

Thank you

alefranz (commented):
Nice Ryan! Much easier to use it now!

src/Middleware/HeaderPropagation/samples/HeaderPropagationSample/HeaderPropagationSample/Program.cs
rynowak: So I'm trying to follow your advice on this. I have currently ended up in a pit of failure that I am already familiar with

1. Adding `ConfigureWebHostDefaults` requires `Microsoft.AspNetCore`
2. Adding `Microsoft.AspNetCore` requires `Microsoft.AspNetCore.Server.IIS`
3. Adding `Microsoft.AspNetCore.Server.IIS` results in:

```
2>C:\Program Files (x86)\Microsoft Visual Studio\d16.1stg\MSBuild\Current\Bin\Microsoft.Common.CurrentVersion.targets(4552,5): error MSB3030: Could not copy the file "C:\git\aspnet\AspNetCore\src\Servers\IIS\AspNetCoreModuleV2\InProcessRequestHandler\bin\Debug\Win32\aspnetcorev2_inprocess.dll" because it was not found.
2>C:\Program Files (x86)\Microsoft Visual Studio\d16.1stg\MSBuild\Current\Bin\Microsoft.Common.CurrentVersion.targets(4552,5): error MSB3030: Could not copy the file "C:\git\aspnet\AspNetCore\src\Servers\IIS\AspNetCoreModuleV2\InProcessRequestHandler\bin\Debug\x64\aspnetcorev2_inprocess.dll" because it was not found.
2>Done building project "Microsoft.AspNetCore.Server.IIS.csproj" -- FAILED.
```

I thought we fixed this?

src/Middleware/HeaderPropagation/src/HeaderPropagationMiddleware.cs
rynowak: Sure 👍 Adding `MultipleEntries_AddsFirstToProduceValue`

src/Middleware/HeaderPropagation/samples/HeaderPropagationSample/HeaderPropagationSample/Program.cs
Tratcher: That's odd, it's working in my sample
https://github.com/aspnet/AspNetCore/pull/9831/files#diff-55f73ec5609ad0e27d098fa76a7e32a4R17

Have you run build on the repo recently?
@jkotalik

src/Middleware/HeaderPropagation/src/HeaderPropagationMiddleware.cs
rynowak: > So the idea is that given an header I receive (e.g. Accept-Language, I'm intentionally not going to use tracing as example :) ) I want to propagate it to my dependencies which unfortunately have different conventions (one wants X-Language, the other Lang) so the output name will be overridable and being able to then configure the clients in terms of the original name accepted by my API/website (as well as being able to find it in the HeaderPropagationValues for logging as mentioned in the other comment) seems the best option.

I'm glad that you're bringing this up because I wanted to get your opinion as well. I'll come back to your example because I think my changes open up an easy path to what you're looking for.

I think storing the header values by their *inbound names* raises a few questions that are hard to answer, or have unsatisfying answers:
- What 'inbound name' do I use for a header that is always generated? 🤷‍♂ 
- Can I map the same inbound header to two outbound names? No 😢 
- Can I map two different headers to the same outbound name? Yes, but the behaviour is undefined 😢

I started contemplating this example while writing the sample [here](https://github.com/aspnet/AspNetCore/pull/9793/files#diff-3e011b5a88b7a99dc13256c30c9859a7R18) and these problems bit me. I have both a generated header `X-SpanId` and a header that's mapped to a different outbound name `X-SpanId -> X-ParentSpanId`. 

Since I'm mapping `X-SpanId` to `X-ParentSpanId` then that mapping needs to use `X-SpanId` as its inbound name. 

So then what *inbound name* do I associate with my *generated* `X-SpanId`. Well it could be anything but one of the other header names. I originally started with `Add("__ignoreme", "X-SpanId", ....);`

This just made me feel sad and like the concepts aren't working for us. Since we're responsible for documentation and guidance, it never feels good to tell a user "make up a name, it doesn't matter" - believe it or not that really confuses and upsets people.

So I went back to requirements. What are the actual constraints of the system? The actual constraint is that outbound headers must not contain duplicates when used with a single HttpClient. 

So what if we lean into that?? From that I derived what's in the PR, and added the (quite elegant IMO) ability to add multiple mappings that run in order. This is a nice little convenience so that you can avoid writing complicated code in a lambda. 

Another option that I considered is to associate a non-semantic "logical name" with each mapping. Then you have no restrictions other than *no-duplicated-logical-names*. I don't want to do this because it really increases the complexity, and I don't think there's a need to go that far.

----

Let's go back to your example.

Does it make sense to forward `Accept-Language` as `Accept-Language` sometimes and `X-Lang` sometimes? Sure it might for your scenario - and it's not blocked by this change.

A different question? Does it make sense to map two different headers to the same *outbound name* **inside a single request**? I don't think you would do that - because then you're not really *propagating headers* - you're doing something else. This is the use case that is blocked by keying on the outbound header name.

I think there's a simple solution to what you proposed here, which is to pass a filter predicate to each message handler with an allow-list of filters to include.  So your code would look like:

```
services.AddHeaderPropagation(options =>
{
    options.Headers.Add("Accept-Language");
    options.Headers.Add("Accept-Language", "X-Lang");
});

services.AddHttpClient("a").AddHeaderPropagation(h => h == "Accept-Language");
services.AddHttpClient("b").AddHeaderPropagation(h => h == "X-Lang");
```

Hopefully this all makes sense. I'm fairly convinced this is the right thing to do, but still interested in your thoughts.

src/Middleware/HeaderPropagation/samples/HeaderPropagationSample/HeaderPropagationSample/Startup.cs
rynowak: @davidfowl - do you have a doc on what's going on with Activity and logging that we can link here for @alefranz

src/Middleware/HeaderPropagation/samples/HeaderPropagationSample/HeaderPropagationSample/Program.cs
rynowak: I have other solutions that use `Microsoft.AspNetCore.Servers.IIS` that are working fine. I'll try building and see if that fixes it.

rynowak: So I did a build, and did a bunch of things and I can't get this to work. The `AuthSamples.sln` doesn't have IIS in it, so I tried removing it. No love.

I'll ask around tomorrow and see if anyone knows what else I should try but at this point I've spent an hour and a half just trying to get this to build.

rynowak: Something seems really wrong with this solution. There are a bunch of build errors that show up every other time I build the solution.

Tratcher: Building in VS can be like that sometimes. Can you build it from command line?

Tratcher: If you get stuck you can still use generic host with its ConfigureHostDefaults, without ConfigureWebHostDefaults, Microsoft.AspNetCore, or IIS. You'd call ConfigureWebHost instead.

rynowak:
Updated this

Tratcher:
Ref assemblies

Tratcher (approved):

src/Middleware/HeaderPropagation/src/HeaderPropagationMiddleware.cs
alefranz: *tl;dr - scroll down to the conclusion*

Hi Ryan,
I appreciate your interest in my opinion.

Let me clarify that I am not arguing that this approach is incorrect, as usual there are compromises to make and probably t is the best compromise. I just want to make sure that some scenarios, which in my experience are pretty common, are not overlooked.
These are not ideal scenario to have to deal with but many times APIs/websites are aggregating mutliple APIs that are out of your control (or they are but are expensive to change) and all use different conventions to gather the same input.

> * What 'inbound name' do I use for a header that is always generated? 🤷‍♂
agree it will be not ideal to have to make up a name, but I believe if's ok-ish if we consider it a logical name. The problem is that even if using the output header name, if this header need to be passed to multiple APIs with different names, you would still need the ability to rename it and using one of these names as key as pros and cons comparing to use a  logical name.
e.g. (in made up new syntax, just to be self explanatory)
```csharp
services.AddHeaderPropagation(o =>o.Headers.Capture("X-SpanId").OrUseIfMissing( x=>Guid.NewGuid()));

services.AddHttpClient("a").AddHeaderPropagation(o => o.Propagate("X-SpanId"));
services.AddHttpClient("b").AddHeaderPropagation(o => o.Propagate("X-SpanId").Named("TransactionId"));
```
> * Can I map the same inbound header to two outbound names? No 😢
I don't get this one. To me it feels similar to the above: you capture inbound header, optionally specify a default outbound header, and later when you use it you can override it.

```csharp
services.AddHeaderPropagation(o =>o.Headers.Capture("Accept-Language").Named("X-Lang"));

services.AddHttpClient("a").AddHeaderPropagation(o => o.Propagate("Lang").Named("Language")); // Here I override the name
services.AddHttpClient("b").AddHeaderPropagation(o => o.Propagate("X-Lang"));
services.AddHttpClient("c").AddHeaderPropagation(o => o.Propagate("X-Lang")));
```

Probably having the ability to specify a default outbound header name can add complexity as you don't really see it when you setup the client, so your approach would be more obvious (as the identifier of the header is what will be used as name if I don't override it). It has the drawback that when all that when all the dependencies have a different names you either have to pick one or make up a logical name and override in all. I think I'm happy with this

```csharp
services.AddHeaderPropagation(o =>o.Headers.Capture("Accept-Language").Named("Lang"));
// outbound name is Lang but I always override it so it is used only as logical name

services.AddHttpClient("a").AddHeaderPropagation(o => o.Propagate("Lang").Named("Language"));
services.AddHttpClient("b").AddHeaderPropagation(o => o.Propagate("Lang").Named("X-Lang"));
```

> * Can I map two different headers to the same outbound name? Yes, but the behaviour is undefined 😢

Do you mean the value can be passed in either one or the other header and you pick one?
e.g. your API can receive the language either in Accept-Language or on a deprecated X-Lang header?  
Yes this is awkward if you use the inbound header name. 
I think this scenario is quite edge case as you later mention:
> Does it make sense to map two different headers to the same _outbound name_ **inside a single request**? I don't think you would do that - because then you're not really _propagating headers_ - you're doing something else. This is the use case that is blocked by keying on the outbound header name.

So I guess you meant something different in one place or the other?

Also because to me if feels if using the outbound header name it actually works
```csharp
services.AddHeaderPropagation(o =>o.Headers.CaptureFromFactory(/* Take from context Accept-Language or, if missing, take X-Lang*/).Named("Lang"));  // it will be stored as Lang and passed to clients as Lang
```

-----
### Conclusion
In conclusion what really matter, in my option, is the ability to specify the headers you want to pass per client, not only as a filter but with the ability to specify a different name.
I don't think using the inbound name block any functionality, but this updated approach looks more obvious when you set it up per client (as the name you see it in your filter is the name that by default will be used ad outbound name in the client).
I think we should actually see it as a logical name used to store the header, which by default it is used as outbound name unless it is overridden in the specific client.

-----
If it's ok with you, after this get merged, I can submit a proposal (or PR) to configure the clients. I haven't figured out a way to specify the configuration of each `HttpClient` in config, but I guess we can start with configuration in code only.

*Someone will then argue: what if one of my dependencies needs that value propagated as parameter in query string? I guess this would be a different feature proposal that risk going too far out of scope to a more generic "parameters propagation thingy™️"*

Sorry for the long reply, here's a 🥔 .

Thank you,
 Alessio

rynowak: > If it's ok with you, after this get merged, I can submit a proposal (or PR) to configure the clients. I haven't figured out a way to specify the configuration of each HttpClient in config, but I guess we can start with configuration in code only.

Sure, I'd love to see more from you on this. I think what's in the PR is a positive change, and I'm open to more design changes up to the 3.0 release if we're adding more features.

alefranz: Thanks. I'll try to open a PR later today.

PR #9796, opened 2019-04-27
Low hanging allocation fruit

davidfowl:
- We were allocating a Task<int> for every read, return a ValueTask<int> for the cases where reads complete synchronously.


davidfowl:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=171200

aspnet-hello:

I've triaged the above build.

jkotalik (approved):

PR #9806, opened 2019-04-28
Avoid buffering during input formatting for longer than necessary

pranavkm:
EnableRewind uses FileBufferingReadStream which is not disposed until the response is completed.
This results in holding on to internal buffers for significantly longer than necessary. Changing it
to return the buffers immediately improved the allocations and throughput.

|                Description |    RPS | CPU (%) | Memory (MB) | Avg. Latency (ms) | Startup (ms) | First Request (ms) | Latency (ms) | Ratio |
| -------------------------- | ------ | ------- | ----------- | ----------------- | ------------ | ------------------ | ------------ | ----- |
| Newtonsoft.Json - old - 500 bytes |  49,661 |      98 |       2,349 |              7.32 |          514 |              87.97 |         0.49 |  1.00 |
| Newtonsoft.Json - new - 500 bytes | 120,273 |      94 |         174 |              2.32 |          509 |              80.54 |         0.36 |  2.42 |
| Newtonsoft.Json - old - 2k | 29,759 |      97 |         382 |              9.11 |          556 |              80.14 |          0.5 |  1.00 |
| Newtonsoft.Json - new - 2k | 36,079 |      97 |         186 |              7.53 |          507 |              84.96 |         0.41 |  1.21 |
| Newtonsoft.Json - old - 8 kbytes | 29,784 |      98 |         343 |              9.41 |          506 |              81.49 |         0.36 |  1.00 |
| Newtonsoft.Json - new - 8 kbytes | 35,408 |      98 |         187 |              8.79 |          503 |              79.64 |         0.33 |  1.19 |
| Newtonsoft.Json - old - 40 kbytes | 2,937 |      76 |       2,004 |             89.87 |          512 |              89.18 |         0.49 |  1.00 |
| Newtonsoft.Json - new - 40 kbytes | 8,416 |      99 |         202 |              30.8 |          534 |              78.79 |         0.47 |  2.87 |

### DIFF
--- a/src/Http/WebUtilities/src/FileBufferingReadStream.cs
+++ b/src/Http/WebUtilities/src/FileBufferingReadStream.cs
@@ -7,6 +7,7 @@ using System.Diagnostics;
 using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Internal;
 
 namespace Microsoft.AspNetCore.WebUtilities
 {
@@ -32,6 +33,16 @@ namespace Microsoft.AspNetCore.WebUtilities
         private bool _completelyBuffered;
 
         private bool _disposed;
+
+        /// <summary>
+        /// Initializes a new instance of <see cref="FileBufferingReadStream" />.
+        /// </summary>
+        /// <param name="inner">The wrapping <see cref="Stream" />.</param>
+        /// <param name="memoryThreshold">The maximum size to buffer in memory.</param>
+        public FileBufferingReadStream(Stream inner, int memoryThreshold)
+            : this(inner, memoryThreshold, bufferLimit: null, tempFileDirectoryAccessor: AspNetCoreTempDirectory.TempDirectoryFactory)
+        {
+        }
 
         public FileBufferingReadStream(
             Stream inner,
@@ -223,13 +234,19 @@ namespace Microsoft.AspNetCore.WebUtilities
                 {
                     oldBuffer.Position = 0;
                     var rentedBuffer = _bytePool.Rent(Math.Min((int)oldBuffer.Length, _maxRentedBufferSize));
-                    var copyRead = oldBuffer.Read(rentedBuffer, 0, rentedBuffer.Length);
-                    while (copyRead > 0)
+                    try
                     {
-                        _buffer.Write(rentedBuffer, 0, copyRead);
-                        copyRead = oldBuffer.Read(rentedBuffer, 0, rentedBuffer.Length);
+                        var copyRead = oldBuffer.Read(rentedBuffer, 0, rentedBuffer.Length);
+                        while (copyRead > 0)
+                        {
+                            _buffer.Write(rentedBuffer, 0, copyRead);
+                            copyRead = oldBuffer.Read(rentedBuffer, 0, rentedBuffer.Length);
+                        }
                     }
-                    _bytePool.Return(rentedBuffer);
+                    finally
+                    {
+                        _bytePool.Return(rentedBuffer);
+                    }
                 }
                 else
                 {
@@ -277,14 +294,19 @@ namespace Microsoft.AspNetCore.WebUtilities
                 {
                     oldBuffer.Position = 0;
                     var rentedBuffer = _bytePool.Rent(Math.Min((int)oldBuffer.Length, _maxRentedBufferSize));
-                    // oldBuffer is a MemoryStream, no need to do async reads.
-                    var copyRead = oldBuffer.Read(rentedBuffer, 0, rentedBuffer.Length);
-                    while (copyRead > 0)
+                    try
                     {
-                        await _buffer.WriteAsync(rentedBuffer, 0, copyRead, cancellationToken);
-                        copyRead = oldBuffer.Read(rentedBuffer, 0, rentedBuffer.Length);
+                        // oldBuffer is a MemoryStream, no need to do async reads.
+                        var copyRead = oldBuffer.Read(rentedBuffer, 0, rentedBuffer.Length);
+                        while (copyRead > 0)
+                        {
+                            await _buffer.WriteAsync(rentedBuffer, 0, copyRead, cancellationToken);
+                            copyRead = oldBuffer.Read(rentedBuffer, 0, rentedBuffer.Length);
+                        }
+                    }finally
+                    {
+                        _bytePool.Return(rentedBuffer);
                     }
-                    _bytePool.Return(rentedBuffer);
                 }
                 else
                 {
--- a/src/Http/WebUtilities/src/FileBufferingReadStream.cs
+++ b/src/Http/WebUtilities/src/FileBufferingReadStream.cs
@@ -303,7 +303,8 @@ namespace Microsoft.AspNetCore.WebUtilities
                             await _buffer.WriteAsync(rentedBuffer, 0, copyRead, cancellationToken);
                             copyRead = oldBuffer.Read(rentedBuffer, 0, rentedBuffer.Length);
                         }
-                    }finally
+                    }
+                    finally
                     {
                         _bytePool.Return(rentedBuffer);
                     }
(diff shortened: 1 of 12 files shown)


src/Http/WebUtilities/src/FileBufferingReadStream.cs
```diff
@@ -223,13 +234,19 @@ public override int Read(byte[] buffer, int offset, int count)
                 {
                     oldBuffer.Position = 0;
                     var rentedBuffer = _bytePool.Rent(Math.Min((int)oldBuffer.Length, _maxRentedBufferSize));
-                    var copyRead = oldBuffer.Read(rentedBuffer, 0, rentedBuffer.Length);
```
pranavkm: `_buffer.Write` may throw. Putting the renting inside of a try-finally block.

src/Mvc/Mvc.Formatters.Xml/test/XmlDataContractSerializerInputFormatterTest.cs
```diff
@@ -242,19 +231,6 @@ public async Task BuffersRequestBody_ByDefaultUsingMvcOptions()
 
             Assert.Equal(expectedInt, model.SampleInt);
             Assert.Equal(expectedString, model.sampleString);
-
-            Assert.True(httpContext.Request.Body.CanSeek);
-            httpContext.Request.Body.Seek(0L, SeekOrigin.Begin);
-
-            result = await formatter.ReadAsync(context);
```
pranavkm: The guidance for users should be to explicitly call `EnableRewind` in a middleware prior to getting to MVC if they want to read the body again after input formatting rather than rely on the formatter.

src/Mvc/Mvc.NewtonsoftJson/src/NewtonsoftJsonInputFormatter.cs
```diff
@@ -129,117 +128,128 @@ public virtual InputFormatterExceptionPolicy ExceptionPolicy
...
+                await readStream.DrainAsync(CancellationToken.None);
+                readStream.Seek(0L, SeekOrigin.Begin);
             }
 
-            using (var streamReader = context.ReaderFactory(request.Body, encoding))
+            var successful = true;
+            Exception exception = null;
+            object model;
+
+            using (var streamReader = context.ReaderFactory(readStream, encoding))
```
pranavkm: Just moved the `ErrorHandler` out of the try-catch block since it doesn't need to be there. Besides disposing the `FileBufferingReadStream`, the rest of this is unchanged.

src/Http/WebUtilities/src/FileBufferingReadStream.cs
davidfowl: Add a test?

src/Mvc/Mvc.Formatters.Xml/src/XmlDataContractSerializerInputFormatter.cs
```diff
@@ -118,43 +117,55 @@ public override async Task<InputFormatterResult> ReadRequestBodyAsync(InputForma
             }
 
             var request = context.HttpContext.Request;
+            Stream readStream = new NonDisposableStream(request.Body);
```
davidfowl: Why is this needed?

Tratcher (commented):
What about the XML perf?

src/Mvc/Mvc.Formatters.Xml/src/XmlSerializerInputFormatter.cs
```diff
@@ -99,39 +98,43 @@ public virtual InputFormatterExceptionPolicy ExceptionPolicy
             }
 
             var request = context.HttpContext.Request;
-
+            Stream readStream = new NonDisposableStream(request.Body);
```
dougbu: Why not `var`?

src/Mvc/Mvc.Formatters.Xml/src/XmlDataContractSerializerInputFormatter.cs
dougbu: And, why not `var`?

dougbu (commented):
Any particular thing you wanted me to look at @pranavkm?

src/Mvc/Mvc.Core/src/MvcOptions.cs
```diff
@@ -103,7 +103,8 @@ public MvcOptions()
         public FormatterCollection<IInputFormatter> InputFormatters { get; }
 
         /// <summary>
-        /// Gets or sets the flag to buffer the request body in input formatters. Default is <c>false</c>.
+        /// Gets or sets the flag that determines if buffering is disabled for input formatters that
```
rynowak:
```suggestion
        /// Gets or sets a value that determines if buffering is disabled for input formatters that
```

rynowak:
So what the impact of this if you wanted to rewind and reread the stream? Would you see a non-seekable stream that's at the end?

What's required if a user wants to enforce buffering (and allow rewinding)?

I'm asking because I'm sure someone has done this, and we're changing the default.

src/Mvc/Mvc.Core/src/MvcOptions.cs
clairepat: I can't tell exactly what the final outcome shows. The Win Zip Files won't work

clairepat: The blue highlights?

clairepat:
I only want to know where to look, unless you have a total in front of you, well, that and the info to deposit

clairepat:
I'm really not, I promise asking anybody to change anything. I just want, what? The amount of the repo and the means to deposit it when you say it's okay. The carpenter ants are early this year and I'm ready to move. That's all. I'm not trying to be pushy.

clairepat:
Look y'all, I'm really not trying to make anybody make it bigger, whatever. When you have no money and cannot work because you're a disabled 62 year old RN, SSDI is nothing, an eighth of my salary, we're living in my father in law's guesthouse in the country. It's really just an overly rustic cabin, it's wood and inside five days the carpenter ants will be everywhere. I promise, I just need a gauge on where to start looking for better housing, that's all.

clairepat:
I have lots of watch only addresses from the 42,000 addresses published to Reddit by Blockchain. It really doesn't get more public domain than that, right? I'll upgrade to the most expensive plan you have if you will make repos of them all. I'm just having a reaction formation from being forced into poverty. I was the primary breadwinner. Poverty is horrible, to say the least, I would like to help others, that was my life. I'm not a greedy person at all. I promise.

clairepat:
OMG, are you serious? Is that actually 105,050 BTC worth of code? It can't be, I mean, I didn't ask for that much but if that's what it is, I'm okay with that. I'm weeping because I think that's what it means. If that's what it is, then good grief, I will live out my days in comfort in a position to help more people.

clairepat:
Just show me where to sign, or whatever, please. If it's half that it's beautiful, This is nothing like I hoped, and had no idea I'd get more than a couple of hundred grand. I keep thinking I fell asleep and will be terribly disappointed. Do I need to click the link to avoid buffering during input formatting? I mean like it reads at the top of the page. Just tell me what to do, please. Anything to hel you and pranavkm.

src/Mvc/Mvc.Formatters.Xml/src/XmlDataContractSerializerInputFormatter.cs
pranavkm: This is existing code. Users may opt out of buffering, we still need to read it in that case. However we do not want to close the stream once the reader is done.

pranavkm: We assign this to a `FileBufferingReadStream` in the `if`, so cannot be an instance of `NonDisposableStream`

src/Mvc/Mvc.Formatters.Xml/src/XmlSerializerInputFormatter.cs
pranavkm: We assign this to a FileBufferingReadStream in the if, so cannot be an instance of NonDisposableStream

pranavkm:
> What's required if a user wants to enforce buffering (and allow rewinding)?

You can explicitly call `EnableRewind` in a middleware prior to getting to MVC and things should work as before. In fact, we avoid double buffering if we think the request stream is buffered.

pranavkm:
🆙 📅

rynowak (approved):

PR #9811, opened 2019-04-28
Remove useless ToString

meziantou:
`_options.Cookie.Path` is already a string, so ToString is useless


analogrelay (approved):
Looks good to me! Thanks!

anurse:
@Tratcher any concerns? I verified that all the above cases are indeed `string` values (not `PathString` or anything funky like that).

Tratcher (approved):

PR #9815, opened 2019-04-29
 Reduce StaticFileContext allocation/struct copy

benaadams:
Lazy alloc the TypedHeaders `_requestHeaders` and `_responseHeaders`; waiting until StaticFiles is actually producing a file rather than just the middleware executing and checking if it should be serving a file.

Move `ServeStaticFile()` to `StaticFileContext` rather than passing the large struct into the method in the middleware and causing a large struct copy.

Combine the three bools `_isGet`, `_isHead` and `_isRangeRequest` in to a single `_requestType` `byte` flags enum.

Change `PreconditionState` from `int` enum to `byte` enum for the 4 states.

### DIFF
--- a/src/Middleware/StaticFiles/src/StaticFileContext.cs
+++ b/src/Middleware/StaticFiles/src/StaticFileContext.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Threading;
@@ -76,23 +77,6 @@ namespace Microsoft.AspNetCore.StaticFiles
             _ifUnmodifiedSinceState = PreconditionState.Unspecified;
             _range = null;
             _requestType = RequestType.Unspecified;
-        }
-
-        internal enum PreconditionState : byte
-        {
-            Unspecified,
-            NotModified,
-            ShouldProcess,
-            PreconditionFailed
-        }
-
-        [Flags]
-        internal enum RequestType : byte
-        {
-            Unspecified = 0b_000,
-            IsHead = 0b_001,
-            IsGet = 0b_010,
-            IsRange = 0b_100,
         }
 
         private RequestHeaders RequestHeaders => (_requestHeaders ??= _request.GetTypedHeaders());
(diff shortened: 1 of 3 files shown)


src/Middleware/StaticFiles/src/StaticFileContext.cs
```diff
@@ -320,6 +324,54 @@ public Task SendStatusAsync(int statusCode)
...
+                    {
+                        context.Response.Clear();
+                    }
+                    await next(context);
+                    return;
+                case PreconditionState.NotModified:
+                    _logger.FileNotModified(SubPath);
+                    await SendStatusAsync(Constants.Status304NotModified);
+                    return;
+
```
DAllanCarr: Nit: remove newline

src/Middleware/StaticFiles/src/StaticFileContext.cs
```diff
@@ -320,6 +324,54 @@ public Task SendStatusAsync(int statusCode)
...
+                case PreconditionState.PreconditionFailed:
+                    _logger.PreconditionFailed(SubPath);
+                    await SendStatusAsync(Constants.Status412PreconditionFailed);
+                    return;
+
```
DAllanCarr: ^^

Tratcher:
@aspnet-hello triage

src/Middleware/StaticFiles/src/StaticFileContext.cs
```diff
@@ -78,56 +76,54 @@ public StaticFileContext(HttpContext context, StaticFileOptions options, PathStr
...
-        {
-            get { return _subPath.Value; }
-        }
+        public bool IsGetMethod => (_requestType & RequestType.IsGet) != 0;
 
-        public string PhysicalPath
-        {
-            get { return _fileInfo?.PhysicalPath; }
-        }
+        public bool IsRangeRequest => (_requestType & RequestType.IsRange) != 0;
```
Tratcher: Add a private setter? Would prefer to keep the bitwise code isolated.

src/Middleware/StaticFiles/src/StaticFileContext.cs
```diff
@@ -78,56 +76,54 @@ public StaticFileContext(HttpContext context, StaticFileOptions options, PathStr
...
-            return _isGet || _isHead;
+            var isValid = false;
+            if (HttpMethods.IsGet(_method))
+            {
+                _requestType |= RequestType.IsGet;
+                isValid = true;
+            }
+            else
+            {
+                _requestType &= ~RequestType.IsGet;
```
Tratcher: Why would you need to clear this? It should have started clear.

Tratcher (approved):

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/2423

src/Middleware/StaticFiles/src/StaticFileContext.cs
benaadams: Was just following the original logic. Will change; also if GET it can't be HEAD

src/Middleware/StaticFiles/src/StaticFileContext.cs
benaadams: Can use `HasFlag` instead; keep forgetting it should be fast now.

benaadams:
Feedback addressed

src/Middleware/StaticFiles/src/StaticFileContext.cs
Tratcher: What about the private setter?

benaadams: Done

Tratcher:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=172654

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/2394
https://github.com/aspnet/AspNetCore-Internal/issues/2407

benaadams:
@Tratcher; delayed creating and using stack for StaticFileContext until validation hits the point of checking file exists; just before trying to serve it, so cost is not paid for other endpoints

benaadams:
Also reordered the declarations of the fields and the initialization of the fields in StaticFileContext to match orders

PR #9829, opened 2019-04-29
RunTemplateTests

ryanbrandenburg:
Clarify intentions of `RunTemplateTests` setting.

### DIFF
--- a/src/ProjectTemplates/test/ProjectTemplates.Tests.csproj
+++ b/src/ProjectTemplates/test/ProjectTemplates.Tests.csproj
@@ -8,7 +8,7 @@
     <TestGroupName>ProjectTemplates.E2ETests</TestGroupName>
     <DefineConstants>$(DefineConstants);XPLAT</DefineConstants>
 
-    <RunTemplateTests>true</RunTemplateTests>
+    <RunTemplateTests Condition="'$(RunTemplateTests)' == ''" >true</RunTemplateTests>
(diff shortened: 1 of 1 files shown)


src/ProjectTemplates/test/ProjectTemplates.Tests.csproj
```diff
@@ -8,7 +8,7 @@
     <TestGroupName>ProjectTemplates.E2ETests</TestGroupName>
     <DefineConstants>$(DefineConstants);XPLAT</DefineConstants>
 
-    <RunTemplateTests>true</RunTemplateTests>
+    <RunTemplateTests Condition="'$(RunTemplateTests)' == ''" >true</RunTemplateTests>
```
javiercn: `?` These tests already ran in all the platforms. Are you trying to run the Selenium bits? (which only run on windows)

javiercn: If that's the case `SeleniumE2ETestsSupported` that's what you are looking to set

ryanbrandenburg: No, I'm just trying to clarify that `RunTemplateTests` can be (and is) set elsewhere, and that should be the value if it is.

ryanbrandenburg:
⌚️

javiercn:
This is a draft PR, can you mark it as non-draft? Is this the full scope of the change?

ryanbrandenburg:
Yes and yes.

javiercn (approved):

Issue #4662, opened 2018-01-26
xplat Integrated Windows Authentication (ntlm/Kerberos)


Eilon:
_From @evil-shrike on January 25, 2018 12:41_

I haven't found any ticket for tracking the xplat Windows-auth support in aspnet/dotnet core. 
I understand that currently we have only out-of-the-box Windows auth support in IIS/IISExpress and http.sys and only on Windows.

There's a decent article in docs https://docs.microsoft.com/en-us/aspnet/core/security/authentication/windowsauth?branch=pr-en-us-3622&tabs=aspnetcore2x on how to configure Windows auth. But it says nothing about non-Windows use-cases. 

There're many issues here on Github and questions around on sites like SO on "how to implement Windows authentication in aspnet core on Linux". But all issues are closed with reasoning like "Kestrel doesn't implement Windows auth currenly, use IIS or https.sys". This is understandable but it's unclear what to do if we need to host an app on Linux and support integrated Windows auth (in intranet networks).

So there are several points which need to be addressed:
* what options do exist currently (at the time of 2.0/2.1) for getting integrated Windows authentication on Linux and other *nix OSes?
* what are plans (if any) for improving/adding such functionality?

I understand that strictly speaking this issue is related to Kestrel project probably, but it's more easily to find it here though. Moreover it's implementation details where Windows-auth support should live. 

Anyway it should be some source of information describing what to do with aspnet core apps on Linux for adding Windows auth.

Some additional info:
* middleware for ntlm support - https://github.com/OneBitSoftware/Microsoft.AspNetCore.Authentication.ActiveDirectory (ntlm only, Windows only - "Achieving server-side NTLM handshaking depends the Windows platform due to interop, dependency on domain-joined machines and secur32.dll")
* it's possible to authenticate via Kerberos on Linux - https://technet.microsoft.com/en-us/library/2008.12.linux.aspx
* there's Application Proxy with Kerberos Constrained Delegation in Windows Server and Azure AD which seems to solve such problems - https://blogs.technet.microsoft.com/applicationproxyblog/2015/09/21/all-you-want-to-know-about-kerberos-constrained-delegation-kcd/ but it's complex
* there're modules for Nginx (http://nginx.org/en/docs/http/ngx_http_upstream_module.html#ntlm) (requires commercial license) and Apache httpd (http://adldap.sourceforge.net/wiki/doku.php?id=mod_auth_ntlm_winbind) (it's unclear how stable it is) to add ntlm support
[...]

PR #9831, opened 2019-04-29
Add new Negotiate Auth handler

Tratcher:
#4662 This is a new package for Negotiate (NTLM/Kerberos) auth. It will not be part of the shared framework.

Starting as a draft while we work out some of the error conditions and add the new connection cleanup feature: https://github.com/aspnet/AspNetCore/pull/9754.

@blowdart


blowdart:
I am here for "Misadventures in Dockerland"

dougbu (commented):
Projects and infrastructure changes look good. No comment on the C# files.

Tratcher:
@blowdart I tried to run a hello world sample in docker before I even started auth. It failed miserably and I wasted a day trying to figure it out before deleting it.

src/Security/Authentication/Negotiate/src/Internal/ReflectedNegotiateState.cs
```diff
@@ -0,0 +1,96 @@
...
+        private static readonly MethodInfo _closeContext;
+
+        private readonly object _instance;
+
+        static ReflectedNegotiateState()
+        {
+            var ntAuthType = typeof(AuthenticationException).Assembly.GetType("System.Net.NTAuthentication");
+            _constructor = ntAuthType.GetConstructors(BindingFlags.NonPublic | BindingFlags.Instance).First();
+            _getOutgoingBlob = ntAuthType.GetMethods(BindingFlags.NonPublic | BindingFlags.Instance).Where(info =>
+                info.Name.Equals("GetOutgoingBlob") && info.GetParameters().Count() == 2).Single();
```
mconnew: Do you want to narrow down to NonPublic? If the names stay the same when the API becomes public, this will now break.

src/Security/Authentication/Negotiate/src/NegotiateHandler.cs
```diff
@@ -0,0 +1,334 @@
...
+                var authorizationHeader = Request.Headers[HeaderNames.Authorization];
+
+                if (StringValues.IsNullOrEmpty(authorizationHeader))
+                {
+                    if (_negotiateState?.IsCompleted == false)
+                    {
+                        throw new InvalidOperationException("An anonymous request was received in between authentication handshake requests.");
+                    }
+                    Logger.LogDebug($"No Authorization header");
+                    return false;
```
mconnew: Does prefixing with a $ add any extra runtime overhead when you don't actually have any dynamic content in the string?

src/Security/Authentication/Negotiate/src/NegotiateHandler.cs
```diff
@@ -0,0 +1,334 @@
...
+                }
+
+                var authorization = authorizationHeader.ToString();
+                Logger.LogTrace($"Authorization: " + authorization);
+                string token = null;
```
mconnew: Using string interpolation but then concatenating the changing part seems conflicting.

src/Security/Authentication/Negotiate/src/NegotiateHandler.cs
```diff
@@ -0,0 +1,334 @@
...
+                _negotiateState = persistence?.State;
+
+                var authorizationHeader = Request.Headers[HeaderNames.Authorization];
+
+                if (StringValues.IsNullOrEmpty(authorizationHeader))
+                {
+                    if (_negotiateState?.IsCompleted == false)
+                    {
+                        throw new InvalidOperationException("An anonymous request was received in between authentication handshake requests.");
+                    }
```
mconnew: Is there no localization on exception message strings?

src/Security/Authentication/Negotiate/src/NegotiateHandler.cs
```diff
@@ -0,0 +1,334 @@
...
+                }
+
+                // Deal with connection credential persistence.
+
+                if (_negotiateState.Protocol == "NTLM" && !Options.PersistNtlmCredentials)
+                {
+                    // NTLM was already put in the persitence cache on the prior request so we could complete the handshake.
+                    // Take it out if we don't want it to persist.
+                    Debug.Assert(object.ReferenceEquals(persistence?.State, _negotiateState),
+                        "NTLM is a two stage process, it must have already been in the cache for the handshake to succeed.");
```
mconnew: You are using an IOptionsMonitor which means the persistance can change. Debug.Assert is meant for when there's a problem with unexpected state indicating a bug but you can get here simply by having the option value change.

src/Security/Authentication/Negotiate/src/NegotiateHandler.cs
```diff
@@ -0,0 +1,334 @@
...
+            {
+                throw new InvalidOperationException("AuthenticateAsync must not be called before the UseAuthentication middleware.");
+            }
+
+            if (IsHttp2)
+            {
+                // Not supported. We don't throw because Negotiate may be set as the default auth
+                // handler on a server that's running HTTP/1 and HTTP/2. We'll challenge HTTP/2 requests
+                // that require auth and they'll downgrade to HTTP/1.1.
+                return AuthenticateResult.NoResult();
```
mconnew: Is this behavior of downgrading to 1.1 when challenged by spec or convention? It sounds like it's the client which is responsible for the downgrade to 1.1? If so, then what happens if a client decides it knows how to do Negotiate auth over 2.0?

src/Security/Authentication/Negotiate/src/NegotiateHandler.cs
```diff
@@ -0,0 +1,334 @@
...
+            if (!_negotiateState.IsCompleted)
+            {
+                // This case should have been rejected by HandleRequestAsync
+                throw new InvalidOperationException("Attempting to use an incomplete authentication context.");
+            }
+
+            // Make a new copy of the user for each request, they are mutable objects and
+            // things like ClaimsTransformation run per request.
+            var identity = _negotiateState.GetIdentity();
+            ClaimsPrincipal user;
```
mconnew: Only use var when it's obvious what type of variable is being returned, e.g. when using new. I have no idea what type identity is by reading this code.

src/Security/Authentication/Negotiate/src/NegotiateHandler.cs
```diff
@@ -0,0 +1,334 @@
...
+            if (identity is WindowsIdentity winIdentity)
+            {
+                user = new WindowsPrincipal(winIdentity);
+                Response.RegisterForDispose(winIdentity);
+            }
+            else
+            {
+                user = new ClaimsPrincipal(new ClaimsIdentity(identity));
+            }
```
mconnew: I think NegotiateStream uses GenericIdentity.

src/Security/Authentication/Negotiate/src/NegotiateHandler.cs
```diff
@@ -0,0 +1,334 @@
...
+            var persistence = new AuthPersistence();
+            RegisterForConnectionDispose(persistence);
+            items[AuthPersistenceKey] = persistence;
+            return persistence;
+        }
+
+        private IDictionary<object, object> GetConnectionItems()
+        {
+            var connectionItems = Context.Features.Get<IConnectionItemsFeature>()?.Items;
+            if (connectionItems == null)
```
mconnew: You can simplify this as:
```csharp
var connectionItems = Context.Features.Get<IConnectionItemsFeature>()?.Items ?? throw new NotSupportedException($"Negotiate authentication requires a server that supports {nameof(IConnectionItemsFeature)} like Kestrel.");
```

src/Security/Authentication/Negotiate/src/NegotiateHandler.cs
```diff
@@ -0,0 +1,334 @@
...
+            {
+                throw new NotSupportedException($"Negotiate authentication requires a server that supports {nameof(IConnectionCompleteFeature)} like Kestrel.");
+            }
+            connectionCompleteFeature.OnCompleted(DisposeState, authState);
+        }
+
+        private static Task DisposeState(object state)
+        {
+            ((IDisposable)state).Dispose();
+            return Task.CompletedTask;
```
mconnew: Is it guaranteed that state will implement IDisposable?

src/Security/Authentication/Negotiate/src/NegotiateOptions.cs
```diff
@@ -0,0 +1,39 @@
...
+        /// <summary>
+        /// Indicates if NTLM credentials should be persisted and re-used for subsquent anonymous requests.
+        /// This option must not be used if connections may be shared by requests from different users.
+        /// The default is true.
+        /// </summary>
+        public bool PersistNtlmCredentials { get; set; } = true;
+
+        // For testing
+        internal INegotiateStateFactory StateFactory { get; set; } = new ReflectedNegotiateStateFactory();
+    }
```
mconnew: This is internal and the only place setting this property is test code. So basically you have a setter only for the purpose of testing. Isn't this what DI is designed to avoid?

src/Security/Authentication/Negotiate/src/Internal/INegotiateState.cs
```diff
@@ -13,6 +13,8 @@ internal interface INegotiateState : IDisposable
 
         bool IsCompleted { get; }
 
+        string Protocal { get; }
```
nil4: Typo?

src/Security/Authentication/Negotiate/src/Internal/ReflectedNegotiateState.cs
```diff
@@ -12,23 +12,25 @@ namespace Microsoft.AspNetCore.Authentication.Negotiate
...
-            var constructor = ntAuthType.GetConstructors(BindingFlags.NonPublic | BindingFlags.Instance).First();
-            // internal NTAuthentication(bool isServer, string package, NetworkCredential credential, string spn, ContextFlagsPal requestedContextFlags, ChannelBinding channelBinding)
-            var credential = CredentialCache.DefaultCredentials;
-            _instance = constructor.Invoke(new object[] { true, "Negotiate", credential, null, 0, null });
+            _constructor = ntAuthType.GetConstructors(BindingFlags.NonPublic | BindingFlags.Instance).First();
             _getOutgoingBlob = ntAuthType.GetMethods(BindingFlags.NonPublic | BindingFlags.Instance).Where(info =>
                 info.Name.Equals("GetOutgoingBlob") && info.GetParameters().Count() == 2).Single();
             _isCompleted = ntAuthType.GetMethods(BindingFlags.NonPublic | BindingFlags.Instance).Where(info =>
                 info.Name.Equals("get_IsCompleted")).Single();
+            _protocal = ntAuthType.GetMethods(BindingFlags.NonPublic | BindingFlags.Instance).Where(info =>
```
nil4: Typo (`protocal`)?

src/Security/Authentication/Negotiate/src/Internal/ReflectedNegotiateState.cs
```diff
@@ -69,6 +78,11 @@ public bool IsCompleted
             get => (bool)_isCompleted.Invoke(_instance, Array.Empty<object>());
         }
 
+        public string Protocal
```
nil4: 😕

src/Security/Authentication/Negotiate/src/NegotiateHandler.cs
Tratcher: I haven't found a spec for it, but all of the clients I tested did do the downgrade except SocketsHttpHandler (it just returned the 401 to the caller).

If a client tries to do Negotiate over HTTP/2 we won't honor it and they'll get a second 401 challenge when they reach the restricted resource.

src/Security/Authentication/Negotiate/src/NegotiateHandler.cs
Tratcher: Yes, but AspNetCore uses ClaimsPrincipal/Identity for everything. GenericIdentity converts easily to ClaimsIdentity.

src/Security/Authentication/Negotiate/src/NegotiateHandler.cs
Tratcher: Yes, in this scenario state is always AuthPersistence.

mconnew: Is it worth adding a Debug.Assert?

src/Security/Authentication/Negotiate/src/Internal/ReflectedNegotiateState.cs
Tratcher: The API proposal is for a new type with reduced surface area. See my branch here:
https://github.com/dotnet/corefx/blob/3e1bbb6cc6c201464f8371ba51c4ab9933e4d86c/src/System.Net.Security/src/System/Net/Security/NegotiateAuthState.cs

When that gets merged I will either have to be careful not to break myself though any private API  changes or react quickly to the breaks (which shouldn't be hard if the changes are constrained to this file).

src/Security/Authentication/Negotiate/src/NegotiateHandler.cs
Tratcher: Leftover from when I had more parameters. Removed.

src/Security/Authentication/Negotiate/src/NegotiateHandler.cs
Tratcher: No, AspNetCore is not currently localized and is inconsistent about the use of resource files. Will revisit when localization happens.

src/Security/Authentication/Negotiate/src/NegotiateHandler.cs
Tratcher: Options changing under us is an interesting topic, but it doesn't affect this assert.

This assert pairs with line 140 above saying all NTLM handshakes require multiple requests and the incomplete state was temporarily persisted regardless of the value of PersistNtlmCredentials.

If the persistence option was changed between requests then you could have two things happen:
- Persistence enabled: No problem, nothing was persisted before, the next request will persist.
- Persistence disabled: Existing connections will maintain their persistence, new connections will not. I don't see anything here that will break.

src/Security/Authentication/Negotiate/src/NegotiateHandler.cs
Tratcher: If this were public or internal maybe, but since it's only used privately from one place I'm not concerned about it. Worst case we'll get a type case exception that should show up in tests.

src/Security/Authentication/Negotiate/src/NegotiateOptions.cs
Tratcher: You're right, but I'd have to make INegotiateStateFactory and INegotiateState public as well. We don't want to do that when this contract exists only for testing.

src/Security/Authentication/Negotiate/src/NegotiateHandler.cs
anurse: To answer your original question (since it seemed interesting): No, the compiler optimizes it out:

https://sharplab.io/#v2:C4LglgNgNAJiDUAfAAgBgATIIwG4CwAUMgMyYBM6AwugN6HoOanIAs6AcgPYAKAhgE68AtgGcAFAEpa9RrOwBOMQBIARADFOnFRPwFZAXxkMjTTGwDqYYAAs+g0ZOl7ZjBcvWbaLMvu26DhPpAA=

Tratcher:
@aspnet/build how do I tell it that the Negotiate.Client and Server projects are test assets and should not be included in eng/ProjectReferences.props?

Tratcher:
`F:\workspace\_work\1\s\eng\targets\Helix.targets(45,81): error MSB4184: The expression """.Contains()" cannot be evaluated. Method 'System.String.Contains' not found. [F:\workspace\_work\1\s\src\Security\Authentication\Negotiate\test\Negotiate.FunctionalTest\Microsoft.AspNetCore.Authentication.Negotiate.FunctionalTest.csproj]`

natemcmaster:
> how do I tell it that the Negotiate.Client and Server projects are test assets and should not be included in eng/ProjectReferences.props?

Put the projects in a folder named 'testassets'.

Tratcher:
Ah, magic.

Tratcher:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=189801

aspnet-hello:

Tratcher:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=189801

Tratcher:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=192518

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/2534

src/Security/Authentication/Negotiate/test/Negotiate.FunctionalTest/CrossMachineReadMe.md
```diff
@@ -0,0 +1,48 @@
+Cross Machine Tests
```
HaoK: I bet this will be fun to get working on helix...

src/Security/Authentication/Negotiate/samples/NegotiateAuthSample/Startup.cs
```diff
@@ -0,0 +1,40 @@
...
+        {
+            app.UseDeveloperExceptionPage();
+            app.UseAuthentication();
+            app.UseAuthorization();
+            app.Run(HandleRequest);
+        }
+
+        public async Task HandleRequest(HttpContext context)
+        {
+            var user = context.User.Identity;
```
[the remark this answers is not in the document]
Tratcher: 😁 there was a vision behind those UseAuthorization asks.

HaoK (approved):

PR #9832, opened 2019-04-29
Make sure read is async first before cancelling.

jkotalik:
Fixes https://github.com/aspnet/AspNetCore-Internal/issues/1798.

We currently have two theories for why this test is flaky:
1. ReadAsync isn't going async, which doesn't seem to be the case as we have a test that confirms ReadAsync goes async everytime.
2. CTS.Cancel isn't being picked up in the PipeReader, which causes ReadAsync to not be canceled.

I'm thinking it's the second now. If this test isn't flaky now, I'm going to start investigating Pipelines.

### DIFF
--- a/src/Servers/IIS/IIS/test/IIS.Tests/ClientDisconnectTests.cs
+++ b/src/Servers/IIS/IIS/test/IIS.Tests/ClientDisconnectTests.cs
@@ -182,7 +182,7 @@ namespace Microsoft.AspNetCore.Server.IISIntegration.FunctionalTests
         [Flaky("https://github.com/aspnet/AspNetCore-Internal/issues/1798", FlakyOn.All)]
         public async Task ReaderThrowsCancelledException()
         {
-            var requestStartedCompletionSource = CreateTaskCompletionSource();
+            var readIsAsyncCompletionSource = CreateTaskCompletionSource();
             var requestCompletedCompletionSource = CreateTaskCompletionSource();
 
             Exception exception = null;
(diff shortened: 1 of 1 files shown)


src/Servers/IIS/IIS/test/IIS.Tests/ClientDisconnectTests.cs
```diff
@@ -179,10 +179,10 @@ public async Task WriterThrowsCancelledException()
         }
 
         [ConditionalFact]
-        [Flaky("https://github.com/aspnet/AspNetCore-Internal/issues/1798", FlakyOn.All)]
+        [Repeat(500)]
         public async Task ReaderThrowsCancelledException()
```
BrennanConroy: nit: `ReaderThrowsCanceledException`

BrennanConroy (approved):
Change looks good.

I'd be interested to see if you find a race between `cts.Cancel()` and some internals of the `ReadAsync` method.

How are we tracking removing the `Repeat` attribute?

src/Servers/IIS/IIS/test/IIS.Tests/ClientDisconnectTests.cs
```diff
@@ -179,10 +179,10 @@ public async Task WriterThrowsCancelledException()
         }
 
         [ConditionalFact]
-        [Flaky("https://github.com/aspnet/AspNetCore-Internal/issues/1798", FlakyOn.All)]
+        [Repeat(500)]
```
Tratcher: 500? How long does this test take?

jkotalik: I’ll remove this after I verified it doesn’t repro.

src/Servers/IIS/IIS/test/IIS.Tests/ClientDisconnectTests.cs
```diff
@@ -179,10 +179,10 @@ public async Task WriterThrowsCancelledException()
         }
 
         [ConditionalFact]
-        [Flaky("https://github.com/aspnet/AspNetCore-Internal/issues/1798", FlakyOn.All)]
-        public async Task ReaderThrowsCancelledException()
+        [Repeat]
```
jkotalik: Leaving this at 10 runs for now. This test takes a second, so 10 seconds total.

jkotalik:
> How are we tracking removing the Repeat attribute?

I think the process I prefer is to occasionally sweep different areas for the Repeat attribute every month or so and remove them if they haven't failed.

PR #9836, opened 2019-04-29
Move from Microsoft.Bcl.Json.Sources to System.Text.Json

dougbu:
- fix downgrade of CoreFx packages in core-sdk
- remove System.Runtime.CompilerServices.Unsafe references from affected projects in hopes they're no longer needed

details:
- CoreFx no longer produces the Microsoft.Bcl.Json.Sources package
- System.Text.Json is now .NET Standard 2.0-compatible
- see also dotnet/corefx#37129 and aspnet/Extensions#1610


BrennanConroy:
You missed the other projects that use System.Text.Json, like MVC and Security

BrennanConroy (approved):

natemcmaster:
You'll need to add this assembly to this section to resolve the error in ResolveReferences.targets.

https://github.com/aspnet/AspNetCore/blob/ba346b45a22295955787477f6869f310aa62e224/eng/SharedFramework.External.props#L84-L92

src/SignalR/common/SignalR.Common/ref/Microsoft.AspNetCore.SignalR.Common.csproj
```diff
@@ -5,12 +5,11 @@
   </PropertyGroup>
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
     <Compile Include="Microsoft.AspNetCore.SignalR.Common.netstandard2.0.cs" />
-    <Reference Include="Microsoft.AspNetCore.Connections.Abstractions" />
-    <Reference Include="Microsoft.Extensions.Options" />
-    <Reference Include="System.Buffers" />
-    <Reference Include="System.Text.Json" />
+    <Reference Include="Microsoft.AspNetCore.Connections.Abstractions"  />
```
BrennanConroy: Whats with all the extra spaces in the references?

dougbu: That's just the way the generator works and it's not our generator. In any case GenApi isn't about pretty files 😺

BrennanConroy: Oh it's the ref file, carry on 😄

src/SignalR/common/SignalR.Common/src/Microsoft.AspNetCore.SignalR.Common.csproj
```diff
@@ -25,9 +25,10 @@
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
-    <Reference Include="Microsoft.Bcl.Json.Sources" />
-    <Reference Include="System.Buffers" />
-    <Reference Include="System.Runtime.CompilerServices.Unsafe" />
+    <Reference Include="System.Text.Json" />
+
+    <!-- System.Text.Json brings System.Buffers in transitively. No need for both here. -->
+    <SuppressBaselineReference Include="System.Buffers" />
```
natemcmaster: Can you move this into the block below? This will prevent suppressions from bleeding over into the next major or minor version.

src/SignalR/common/Http.Connections.Common/src/Microsoft.AspNetCore.Http.Connections.Common.csproj
```diff
@@ -20,9 +20,10 @@
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'" >
-    <Reference Include="Microsoft.Bcl.Json.Sources" />
-    <Reference Include="System.Buffers" />
-    <Reference Include="System.Runtime.CompilerServices.Unsafe" />
+    <Reference Include="System.Text.Json" />
+
+    <!-- System.Text.Json brings System.Buffers in transitively. No need for both here. -->
+    <SuppressBaselineReference Include="System.Buffers" />
```
natemcmaster: Same here.

natemcmaster (changes requested):
One nit about code organization. Otherwise looks good.

src/SignalR/common/SignalR.Common/src/Microsoft.AspNetCore.SignalR.Common.csproj
dougbu: I'll try it.

dougbu:
:up::date: for @natemcmaster

natemcmaster (approved):

Issue #8232, opened 2019-03-06
Support 'key' to guarantee preservation of elements/components


SteveSandersonMS:
This is inspired by, and would basically be equivalent to, the `key` feature in React. It provides a way to tell the system how you want it to preserve child element/component instances when diffing a list.

### Use case

Consider the existing situation where you *don't* have this feature:

```html
<div>
    @foreach (var flight in Flights)
    {
        <DetailsCard Flight="@flight" />
    }
</div>
```

If you add a new flight into the middle of the `Flights` list, what you'd want is that all the existing `DetailsCard` instances are unaffected, and one new `DetailsCard` gets created and put into the rendered output.

To visualize this, if `Flights` previously contained `[F0, F1, F2]`, then this is the **before** state:

 * DetailsCard0, with Flight=F0
 * DetailsCard1, with Flight=F1
 * DetailsCard2, with Flight=F2

... and this is the **desired after** state, given we insert a new item `FNew` at index 1:

 * DetailsCard0, with Flight=F0
 * DetailsCardNew, with Flight=FNew
 * DetailsCard1, with Flight=F1
 * DetailsCard2, with Flight=F2

However, the **actual after** state with the present diff algorithm is this:

 * DetailsCard0, with Flight=F0
 * DetailsCard1, with Flight=FNew
 * DetailsCard2, with Flight=F1
 * DetailsCardNew, with Flight=F2

The system has no way to know that DetailsCard2 or DetailsCard3 should preserve their associations with their older Flight instances, so it just re-associates them with whatever Flight matches their position in the list. As a result, DetailsCard1 and DetailsCard2 rebuild themselves completely using new data, which is wasteful and sometimes even leads to user-visible problems (e.g., input focus is unexpectedly lost).

#### Fixing this with 'key'

The developer should be able to make their app more efficient (and sometimes, better behaving) using `key` like this:

```html
<div>
    @foreach (var flight in Flights)
    {
        <DetailsCard key="@flight" Flight="@flight" />
    }
</div>
```

Here, `key` is a new built-in intrinsic that you can use on **any component**, or **any element**. The result of this will be that you do get the **desired after** state shown above.

To achieve this, the diff algorithm:

  * Notices if you've supplied a `key`, then from that point onwards within the current parent frame (in the above example, the `<div>`), it changes the rules for retaining child elements/components
  * In this mode, previous child elements/components are reused if and only if their old frames have the same `key` value as the new frames.
[...]

PR #9850, opened 2019-04-30
Key support. Implements #8232

SteveSandersonMS:
This is the runtime part of allowing syntax like `<something key="@somekey" />` to guarantee affinity of elements/components with model objects.

### Common use case

```cshtml
@foreach (var person in People)
{
    <div key="@person">... some complex DOM structure goes here ...</div>
}
```

... or:

```cshtml
@foreach (var person in People)
{
    <PersonDisplay key="@person" PublicInfo="@person.PublicInfo" />
}
```

The use of `key` changes how the diff algorithm handles changes to the `People` list:

 * Without `key`, it matches model objects with elements/components in the order they both appear. So if, say, `People[0]` was removed, then it would reassociate each element/component with the `n+1`-th item in the list, and finally delete the last element/component. This can cause unnecessary extra re-rendering of child components or even problems such as making the focus appear to move to an unexpected textbox (if the focus was in a textbox for `Person[2]`, that textbox would now be reassociated with what was previously `Person[3]`, which looks as if the focus moved (although really, it's the rest of the world that moved and the focus stayed still)).
 * With `key`, it matches model objects with elements/components based on equality of the supplied `key` object. So if, say, `People[0]` was removed, then it would know simply to delete the first element/component, and leave everything else alone.

Similarly, if the entries in `People` were shuffled, then:

 * Without `key`, it would reassociate the `i`-th element/component with the new entry `People[i]`, and re-render. This performs a fine-grained update on all descendants (e.g., updating all the text nodes individually).
 * With `key`, it would simply re-order the elements/components to match the new ordering. It would not necessarily re-render descendants (though in practice, in the components case, it does ask them if they want to re-render because it doesn't know whether the `People` entries themselves have mutated).

Using `key` remains optional, but it's nearly always a good idea to use it any time you render a list whose data is subject to insertions/deletions/reordering.

The other possible reason to use `key` is if you want to guarantee **non**-preservation of a given subtree when some data changes. Example:

```
    <div key="@myModel.Id">... some complex DOM structure ...</div>
```

Without `key`, the renderer will respond to changes by recursing into the descendants and updating each affected descendant individually. With `key`, if `myModel.Id` changes, it will throw out the entire subtree and insert a new one - this guarantees that things like focus state are *not* retained.

### Code review notes

While implementing this, I hit what I felt were complexity limits in a couple of areas, forcing me to refactor before making substantial additions. You may find it easier to read the individual commits in order. In particular, I refactored:

 * `RenderTreeFrame`, because the old system of having *many* ad-hoc constructors corresponding to different scenarios became intractible. There would soon be clashes of constructor signatures, and it was hard to be sure that all the `With...` methods (that clone-with-mutatation the immutable struct) would correctly preserve all the fields that were not being modified. Basically, this is crying out for C# 8.x *records*. I changed it to have one constructor per frame type, where that constructor forces you to specify all the fields applicable to that frame type. It's therefore unambiguous where to add any new fields and ensures you can't forget to copy other ones.
 * `RenderTreeBuilder`, mainly to reorganize the `AppendDiffEntriesForRange` logic. Previously it mixed "deciding what to do" with "doing that thing". Those two concepts are now separated, so it's easier to have multiple distinct code paths that result in the same outcome (e.g., "treat as insert" or "match and recurse into descendants"). The code diff for this refactoring is hard to read, unfortunately, because it looks as it basically everything changed even though mostly it was moving blocks of code around.
    * I also tried breaking up `AppendDiffEntriesForRange` into multiple methods, because it's huge. However, benchmarking showed about a 10% slowdown on Mono WebAssembly after doing that, which I would think is because it has to execute a *lot* more instructions to do all the parameter passing. So I have left it as one big method, but tried to clarify the different areas of responsibility.

The logic around detecting keyed inserts/deletions is really simple and doesn't need much of an explanation. Handling moves, however, is much more complicated. My goal was to find a way of distributing the workload between the .NET side and the JS side such that it was no worse than O(N) on either side, and that the .NET side didn't involve any per-render allocations in typical cases. The solution was not generate a list of "moves" (I can't think of a way to do that without either allocating for each key or doing O(N^2) lookups, but if anyone has a solution, I'd be interested!). Instead it generates a description of the final state as a permutation of the normal post-edit state. On the JS side, we can apply the permutation in O(N) time with O(N) allocations (inserting marker nodes into the DOM to track where things will get inserted) - this small number of allocations on the JS side doesn't concern me at all.


src/Components/Blazor/Build/src/ReferenceFromSource.props
```diff
@@ -11,7 +11,7 @@
 
   <PropertyGroup>
     <BlazorBuildReferenceFromSource>true</BlazorBuildReferenceFromSource>
-    <BlazorJsPath>$(RepositoryRoot)src\Components\Browser.JS\dist\$(Configuration)\blazor.*.js</BlazorJsPath>
+    <BlazorJsPath>$(RepositoryRoot)src\Components\Browser.JS\dist\$(Configuration)\blazor.*.js.*</BlazorJsPath>
```
javiercn: Source maps?

SteveSandersonMS: Yes: https://github.com/aspnet/AspNetCore/pull/9850/commits/75df885226cab766dca8a22cb3d0595122a70c48

src/Components/Components/src/RenderTree/RenderTreeDiffBuilder.cs
```diff
@@ -35,107 +36,256 @@ public static void DisposeFrames(RenderBatchBuilder batchBuilder, ArrayRange<Ren
...
             int newStartIndex, int newEndIndexExcl)
         {
+            // This is deliberately a very large method. Parts of it could be factored out
+            // into other private methods, but doing so comes at a consequential perf cost,
+            // because it involves so much parameter passing. You can think of the code here
+            // as being several private methods (delimited by #region) pre-inlined.
+            //
+            // A naive "extract methods"-type refactoring will worsen perf by about 10%. So,
+            // if you plan to refactor this, be sure to benchmark the old and new versions
+            // on Mono WebAssembly.
```
javiercn: Is this true even with `[MethodImpl(MethodImplOptions.AggressiveInlining)]` applied to the child methods?

src/Components/Components/src/RenderTree/RenderTreeEdit.cs
```diff
@@ -1,37 +1,47 @@
...
 
-using System;
+using System.Runtime.InteropServices;
 
 namespace Microsoft.AspNetCore.Components.RenderTree
 {
     /// <summary>
     /// Represents a single edit operation on a component's render tree.
     /// </summary>
+    [StructLayout(LayoutKind.Explicit)]
```
SteveSandersonMS: Client-side Blazor has always relied on knowing the struct layout of `RenderTreeEdit`. This now makes it explicit that we're doing that and what offsets we expect.

src/Components/Components/src/RenderTree/StackObjectPool.cs
```diff
@@ -0,0 +1,73 @@
...
+
+using System;
+
+namespace Microsoft.AspNetCore.Components.RenderTree
+{
+    // This is a very simple object pool that requires Get and Return calls to be
+    // balanced as in a stack. It retains up to 'maxPreservedItems' instances in
+    // memory, then for any further requests it supplies untracked instances.
+
+    internal class StackObjectPool<T> where T : class, new()
```
SteveSandersonMS: I did consider bringing in `ObjectPool` from the extensions repo. However that involves referencing a whole extra assembly and its functionality is much more general than we need. And not that it will matter a lot, but `StackObjectPool` will perform slightly better.

If at some later date we end up using `ObjectPool` in client-side Blazor for other reasons, we can consider replacing `StackObjectPool` with it.

src/Components/Components/src/RenderTree/RenderTreeBuilder.cs
```diff
@@ -454,13 +454,30 @@ public void OpenComponent(int sequence, Type componentType)
         /// <param name="value">The value for the key.</param>
         public void SetKey(object value)
         {
-            // This is just a placeholder to enable work in parallel in the
-            // aspnetcore-tooling repo.
-            //
-            // The real implementation will involve multiple overloads, likely:
-            //   SetKey(int value) -- underlying logic
-            //   SetKey<T>(T value) where T: struct -- avoids boxing 'value' before calling .GetHashCode()
-            //   SetKey(object value) -- performs null check before calling .GetHashCode()
```
SteveSandersonMS: In the end, there's no need for the anticipated overloads, because we're now storing the key as an `object` instead of an `int`-valued hashcode.

src/Components/Components/test/RenderTreeDiffBuilderTest.cs
```diff
@@ -103,6 +103,216 @@ public void RecognizesOldItemsBeingRemoved()
                 entry => AssertEdit(entry, RenderTreeEditType.RemoveFrame, 1));
         }
```
SteveSandersonMS: TODO: Also add tests here showing how we handle clashing keys.

src/Components/Components/src/RenderTree/RenderTreeDiffBuilder.cs
SteveSandersonMS: Yes, because the Mono WebAssemby interpreter isn't JITing.

src/Components/Components/src/RenderTree/RenderTreeDiffBuilder.cs
```diff
@@ -35,107 +36,256 @@ public static void DisposeFrames(RenderBatchBuilder batchBuilder, ArrayRange<Ren
...
+                            // back in sync)
+                            action = newSeq < oldSeq ? DiffAction.Insert : DiffAction.Delete;
+
+                            if (oldLoopedBack)
                             {
-                                oldLoopsBackLater = true;
-                                break;
+                                // If both old and new have now looped back, we must reset their 'looped back'
+                                // tracker so we can treat them as proceeding through the same loop block
+                                prevOldSeq = prevNewSeq = -1;
```
javiercn: nit: We don't do this through the codebase. It's easy to miss when modifying code.

src/Components/Components/src/RenderTree/RenderTreeDiffBuilder.cs
```diff
@@ -35,107 +36,256 @@ public static void DisposeFrames(RenderBatchBuilder batchBuilder, ArrayRange<Ren
...
-                        hasMoreOld = oldEndIndexExcl > oldStartIndex;
-                        prevOldSeq = oldSeq;
+                        // It's much easier for the recipient to handle if we're explicit about
+                        // when the list is finished
+                        diffContext.Edits.Append(RenderTreeEdit.PermutationListEnd());
                     }
+                    #endregion
+
+                    keyedItemInfos.Clear();
+                    diffContext.KeyedItemInfoDictionaryPool.Return(keyedItemInfos);
```
javiercn: I understand why the two lines above are there, but I'm not sure why anything else is part of the finally. I imagine that if there is an exception here we don't care about the batch so why do any work?

The system might also be in an unknown state and those lines above have a chance of causing an exception before we get to these two lines, so I think that the least we do here the better. and those lines can be moved to the try block.

src/Components/Components/src/RenderTree/RenderTreeFrame.cs
```diff
@@ -119,6 +133,12 @@ namespace Microsoft.AspNetCore.Components.RenderTree
         /// </summary>
         [FieldOffset(24)] internal readonly ComponentState ComponentState;
 
+        /// <summary>
+        /// If the <see cref="FrameType"/> property equals <see cref="RenderTreeFrameType.Component"/>,
+        /// gets the component's diffing key, or null if none was specified.
+        /// </summary>
+        [FieldOffset(32)] public readonly object ComponentKey;
```
javiercn: I imagine that this value is always set to something "serializable" even though its marked as object given the assumption that we send this to the client.

javiercn (commented):
I've taken an initial look through the source code. Everything looks good, although TBH I'm not an expert yet on the diffing code, so I will have to take a deeper look probably tomorrow. I plan to clone the branch and play a bit with it, but overall looks great!

If I understood correctly there are two cases for using key:
* Components: The entire component gets rerendered (this happens today without key, so no issue here?)
* Elements: Instead of moving the world around the first item, things are moved around. Do subtrees get diffed between matching key elements or do we always render the new entire subtree?

Regarding the algorithm. Are we using a specific diffing algorithm from literature or did we just "craft" our own. I imagine you've done this, but we could potentially do some research and see if we can find anything already invented that might perform better in terms of time/space.

I've taken a quick look at the tests and I couldn't find any E2E test. Shouldn't we have some?

src/Components/Components/src/RenderTree/RenderTreeDiffBuilder.cs
SteveSandersonMS: OK, will change it.

src/Components/Components/src/RenderTree/RenderTreeDiffBuilder.cs
SteveSandersonMS: OK that's fair - I'll move the rest out of the `finally`.

src/Components/Components/src/RenderTree/RenderTreeFrame.cs
SteveSandersonMS: We never serialize or send this object to the client. It doesn't have to be serializable.

SteveSandersonMS:
> If I understood correctly there are two cases for using key:

The common use cases are described as best as I can in the PR description. It's not just about the first item in a list; it's about general insertions/deletions/reorderings and preserving the association between model objects and elements/components.

> Regarding the algorithm. Are we using a specific diffing algorithm from literature or did we just "craft" our own.

At a high level, our algorithm for elements/components/regions is essentially "merge join". This is much simpler than a general-purpose diff, as well as faster, and is possible only because we have the compile step that lets us inject sequence numbers. We only need a single linear pass, which is much cheaper than most general-purpose diffs that don't require sequence numbers. When we detect reordered attributes or keys, our algorithm is essentially "hash join", which is again O(N) time.

> I've taken a quick look at the tests and I couldn't find any E2E test. Shouldn't we have some?

This is noted in the PR description as "still to do".

javiercn:
> This is noted in the PR description as "still to do".

Fair enough, didn't see it.

src/Components/Components/test/RenderTreeDiffBuilderTest.cs
SteveSandersonMS: Done

src/Components/Components/src/RenderTree/StackObjectPool.cs
rynowak: Yeah, I really wish that I had made object pool a source package instead of a public api 🥀

src/Components/Components/src/RenderTree/RenderTreeBuilder.cs
```diff
@@ -454,13 +454,30 @@ public void OpenComponent(int sequence, Type componentType)
...
+            {
+                throw new InvalidOperationException("Cannot set a key outside the scope of a component or element.");
+            }
+
+            var parentFrameIndexValue = parentFrameIndex.Value;
+            ref var parentFrame = ref _entries.Buffer[parentFrameIndexValue];
+            switch (parentFrame.FrameType)
+            {
+                case RenderTreeFrameType.Element:
+                    _entries.Buffer[parentFrameIndexValue] = parentFrame.WithElementKey(value);
```
rynowak: out of curiousity would `parentFrame = parentFrame.WithElementKey(value);` work here?

src/Components/Components/src/RenderTree/RenderTreeFrame.cs
```diff
@@ -17,13 +16,22 @@ namespace Microsoft.AspNetCore.Components.RenderTree
...
-        // This is not optimal for the Mono-WebAssembly case because that's always 32-bit so the
-        // reference-type fields could be reduced to 4 bytes each. We could use ifdefs to have
-        // different fields offsets for the 32 and 64 bit compile targets, but then we'd have the
-        // complexity of needing different binaries when loaded into Mono-WASM vs desktop.
-        // Eventually we might stop using this shared memory interop altogether (and would have to
-        // if running as a web worker) so for now to keep things simple, treat reference types as
-        // 8 bytes here.
+
+        // Although each frame type uses the slots for different purposes, the runtime does not
+        // allow reference type slots to overlap with each other or with value-type slots.
```
rynowak: > the runtime does not allow reference type slots to overlap with each other or with value-type slots

CoreCLR does allow this, but it produces unverifiable (and thus unsafe) code. I don't recommend we do it, but I have built and thrown away things that do it before 😈

src/Components/Components/src/RenderTree/RenderTreeFrame.cs
```diff
@@ -184,66 +204,72 @@ namespace Microsoft.AspNetCore.Components.RenderTree
         /// </summary>
         [FieldOffset(16)] public readonly string MarkupContent;
 
-        private RenderTreeFrame(int sequence, string elementName, int elementSubtreeLength)
+        // Element constructor
+        private RenderTreeFrame(int sequence, int elementSubtreeLength, string elementName, object elementKey)
             : this()
```
rynowak: Is there a particular purpose here to calling `this()`?

src/Components/Components/src/RenderTree/RenderTreeFrame.cs
```diff
@@ -303,25 +317,31 @@ internal static RenderTreeFrame ComponentReferenceCapture(int sequence, Action<o
...
 
         internal RenderTreeFrame WithElementReferenceCaptureId(string elementReferenceCaptureId)
-            => new RenderTreeFrame(Sequence, ElementReferenceCaptureAction, elementReferenceCaptureId);
+            => new RenderTreeFrame(Sequence, elementReferenceCaptureAction: ElementReferenceCaptureAction, elementReferenceCaptureId);
+
+        internal RenderTreeFrame WithElementKey(object elementKey)
+            => new RenderTreeFrame(Sequence, elementSubtreeLength: ElementSubtreeLength, ElementName, elementKey);
+
+        internal RenderTreeFrame WithComponentKey(object componentKey)
+            => new RenderTreeFrame(Sequence, componentSubtreeLength: ComponentSubtreeLength, ComponentType, ComponentState, componentKey);
```
rynowak: Should we update the `ToString()` implementation?

src/Components/Components/src/RenderTree/StackObjectPool.cs
```diff
@@ -0,0 +1,73 @@
...
+        public T Get()
+        {
+            _numSuppliedItems++;
+
+            if (_numSuppliedItems <= _maxPreservedItems)
+            {
+                if (_numTrackedItems < _numSuppliedItems)
+                {
+                    // Need to allocate a new one
+                    var newItem = new T();
```
rynowak: Be aware that this won't be super fast. `new T()` is actually `Activator.CreateInstance()`. If this class doesn't need to be generic, then consider doing a coupled thing.

src/Components/Components/src/RenderTree/StackObjectPool.cs
```diff
@@ -0,0 +1,73 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Components.RenderTree
+{
+    // This is a very simple object pool that requires Get and Return calls to be
+    // balanced as in a stack. It retains up to 'maxPreservedItems' instances in
+    // memory, then for any further requests it supplies untracked instances.
```
rynowak: I assume no need to be thread-safe then?

src/Components/Components/src/Rendering/KeyedItemInfo.cs
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Components.Rendering
+{
+    // Used internally during diffing to track what we know about
+    // keyed items and their positions
+    internal struct KeyedItemInfo
```
rynowak: Does this need to be mutable? Immutable is nice where possible since it's less error prone

src/Components/Components/src/RenderTree/RenderTreeDiffBuilder.cs
rynowak: I'm all for big methods. Bring it on

rynowak:
I looked over the first few commits, and got lost when we got to the diffing algo changes. This is an area where I haven't really gone deep in the past, and I think it would be more valuable for someone besides me to become a second expect on this (👀 in the direction of Team Blazor EU). 

I don't want to create noise by adding low value comments :)

src/Components/Components/src/RenderTree/RenderTreeBuilder.cs
SteveSandersonMS: Yes it would, and I suppose theoretically it could be faster because it doesn't have to recompute the pointer to the array entry. Not sure it would really make any difference in practice.

Anyway I've changed it to work as you suggest, with a clarifying comment.

src/Components/Components/src/RenderTree/RenderTreeFrame.cs
SteveSandersonMS: OK, well if one day we get desperate to shrink `RenderTreeFrame` by a small amount, we can consider doing that! For now let's keep it safe.

src/Components/Components/src/RenderTree/RenderTreeFrame.cs
SteveSandersonMS: If you don't, the compiler gets mad because there are unassigned readonly fields.

This whole approach of having multiple fields sharing memory slots, and having the whole struct be readonly, and the fact constructors can only be differentiated by signature and not by any other kind of name or identifier, hits compiler limitations all over the place. Something like C's `union` would be a more natural way to expressive this.

I think the low-levelness of this struct within the rendering system justifies going to all these lengths but I wouldn't do it in less perf-critical places.

src/Components/Components/src/RenderTree/RenderTreeFrame.cs
SteveSandersonMS: OK, have done. It won't always be useful because if the key doesn't have a good `ToString` of its own, it will just be represented as its type name with nothing to differentiate it. But it is useful if the key is a string/int/etc.

src/Components/Components/src/RenderTree/StackObjectPool.cs
SteveSandersonMS: Good tip - thanks. I had previously assumed otherwise, but having read up on it, I see why now.

I've changed `StackObjectPool` to require you to pass an instance factory (i.e., `Func<T>`) to its constructor.

SteveSandersonMS: TBH it wouldn't have mattered in our use cases. The whole point of these instances being pooled is that the cost of creating/collecting them amortises to zero if your app keeps running.

src/Components/Components/src/RenderTree/StackObjectPool.cs
SteveSandersonMS: No, it's not making any claims of thread-safety (like most other collection structures). We only use it within the context of a single renderer, where we've already gone to great lengths to ensure calls and hence renders are serialized.

src/Components/Components/src/Rendering/KeyedItemInfo.cs
SteveSandersonMS: Previously I was OK with the mutability because this type is only used in a very localised way within `RenderTreeDiffBuilder` and doesn't really get passed around from one place to another.

However on further consideration, it's not really a big chore to make it immutable, and the perf cost of a bit more copying isn't going to hurt us (it only happens in the specific case where we detect a moved frame).

So, done.

SteveSandersonMS:
OK, hopefully this is now ready to go.

SteveSandersonMS:
Ping to reviewers on this. 

@javiercn If it would be helpful to have a call about this so I can walk you through the algorithm, please let me know!

src/Components/Components/test/RenderTreeDiffBuilderTest.cs
```diff
@@ -1492,6 +1903,207 @@ public void DoesNotInvokeAssignerAgainForRetainedComponents()
             Assert.Empty(referenceFrames);
         }
 
+        [Fact]
```
javiercn: Does it make sense to test something like the frame type changing with the given same key?
For example:
```
element[k1]
element[k2]

component[k2]
component[k1]
```

Also, can multiple key sequences appear on the same component/element at different ranges?
For example a component like:
```
<section>
foreach(var item in FirstList){
  <div @key=@item.Id>
  </div>
}
</section>
<p>Some extra content</p>
<section>
foreach(var item in SecondList){
  <div @key=@item.Id>
  </div>
}
</section>
```

src/Components/test/testassets/BasicTestApp/KeyCasesTreeNode.razor
```diff
@@ -0,0 +1,53 @@
+<div class="node">
```
javiercn: Should we have some E2E test that involves inputs to set expectations?
* For example, when we move an input the focus doesn't get lost.
* We could probably check that the cursor/selection stays in place.
* Could we have a test for inputs moving elsewhere in the tree? (like the checkboxes issue that was reported the other day).

javiercn (approved):
I think the change looks great!

@SteveSandersonMS You've done a remarkable job here. Other than a couple of comments about test cases that I'm not sure if we should add, I think it's pretty polished and good to go.

I'll let up to you whether you think the cases I've mentioned make sense or if we can live without them.

src/Components/Components/test/RenderTreeDiffBuilderTest.cs
SteveSandersonMS: > Does it make sense to test something like the frame type changing with the given same key?

Great suggestion. Previously we didn't handle that because it wasn't possible for a `.razor` component. However it is technically now possible when matching on keys. Added support and test.

> Also, can multiple key sequences appear on the same component/element at different ranges?

This isn't an issue, because we only match within the closest container (element or region). In your example the keys won't interfere with each other because each `<section>` is diffed separately.

javiercn: @SteveSandersonMS You're right. My sample is tainted by the `<section>`. What if those were just `<h1>` (and no `</section>`)

SteveSandersonMS: If they are all in the same container, then by design we will match across all those keys. There's nothing to differentiate different ranges in that case.

javiercn: @SteveSandersonMS Ok. I think its fair to tell people in that case to nest them into a common element. The scenario that I was thinking of was some sort of reporting UI with different views over the same data, but now that I think more of it, you would likely put each of those collections within some container, like a list or a table, so I think its fair to tell people to do it that way.

SteveSandersonMS:
@javiercn 

> Should we have some E2E test that involves inputs to set expectations?
> For example, when we move an input the focus doesn't get lost.
> We could probably check that the cursor/selection stays in place.
> Could we have a test for inputs moving elsewhere in the tree? (like the checkboxes issue that was reported the other day).

I'm very glad you suggested this. Adding these E2E tests uncovered two problems:

 * We were not preserving focus state when moving elements. I just totally forgot this was needed. I should have known because [we do the exact same thing in knockout.js when reordering elements](https://github.com/knockout/knockout/blob/b81762131f96b1042fc70eb4ea6bdc898475b942/src/binding/editDetection/arrayToDomNodeChildren.js#L212) :) Fixed.
 * There's an utterly bizarre behavior in Chrome that I can only guess is a browser bug. It doesn't happen in Firefox. I found that when moving nodes, the act of calling `node.insertBefore` from inside an `input` event can trigger a `change` event at that point in the call stack. So this leads to nested event handlers, which is not something we support. In turn it causes nested rendering, which leads to corruption and a renderer exception. Firefox behaves better by not firing the `change` event until after the `input` handler has returned.
    * This only affects client-side Blazor, because that's the only case where events are truly synchronous. On server-side Blazor, the asyncness of running the event handler across the network means the handlers aren't nested on the call stack. Plus the server-side sync context would run the event handlers in series anyway.
    * To work around this, and to protect us against other more general cases of nested event handling (e.g., maybe DOM mutation observers could cause this), I've made some changes in client-side Blazor's `RendererRegistryEventDispatcher`. It now explicitly checks for nested event dispatch, and if that happens, it queues items so they run in series (though they still start synchronously). This makes things consistent both across server/client hosting and across browsers.

There are now E2E test cases for all things you suggested (typing while boxes move around, and checking checkboxes in lists that move as you check them).

If you want to review the last few new commits please go ahead! I'm still planning to merge this once the CI checks complete because this PR has been waiting for so long and I think the final changes (which are mostly E2E tests) won't really need much further review.

javiercn:
> I'm very glad you suggested this. Adding these E2E tests uncovered two problems

Yay! I added value :)

Taking a look at the new commits.

javiercn (approved):
New changes look good too!

SteveSandersonMS:
Windows, Mac, and Linux builds all failed for entirely different reasons, none of which appear related to this PR :(

PR #9865, opened 2019-04-30
Added Microsoft.Extensions.Logging.EventLog to the shared framework

davidfowl:
Pre-empting https://github.com/aspnet/Extensions/pull/1623


dougbu (approved):

natemcmaster (changes requested):
This adds at least one new transitive dependency to our sharedfx, System.Diagnostics.EventLog.dll. You'll need to add to this section to get tests to pass

https://github.com/aspnet/AspNetCore/blob/9d5a8b3ead8cca083d182daae6b0b78c2a642ea6/eng/SharedFramework.External.props#L63-L73

natemcmaster (approved):

PR #9877, opened 2019-05-01
Make stream buffer capacity configurable

### LINKED ISSUE #4399: Allow the user to configure the bounds used for the client->server streaming channel

Currently there is an unbounded channel on the server used for the client->server streaming.
https://github.com/aspnet/AspNetCore/blob/e310ccac7b30e901f657f81c86c4048fecfedf7b/src/SignalR/src/Microsoft.AspNetCore.SignalR.Core/StreamTracker.cs#L81

We should add some sort of bounds so if the client sends too fast it will get back-pressured.

mikaelm12:
Fixes: https://github.com/aspnet/AspNetCore/issues/4399

Adds `StreamBufferCapacity` hub option.


src/SignalR/server/SignalR/test/AddSignalRTests.cs
```diff
@@ -80,6 +80,20 @@ public void HubSpecificOptionsDoNotAffectGlobalHubOptions()
...
+        {
+            var serviceCollection = new ServiceCollection();
+
+            serviceCollection.AddSignalR().AddHubOptions<CustomHub>(options =>
+            {
+                options.StreamBufferCapacity = 42;
+            });
+
+            var serviceProvider = serviceCollection.BuildServiceProvider();
+            Assert.Equal(42, serviceProvider.GetRequiredService<IOptions<HubOptions<CustomHub>>>().Value.StreamBufferCapacity);
```
halter73: Can we add a more end-to-end test that verifies the client observes the back pressure once the specified capacity is reached?

mikaelm12: Yeah, this still needs real tests. Should have included that in the description.

src/SignalR/common/testassets/Tests.Utils/HubConnectionContextUtils.cs
```diff
@@ -29,13 +29,13 @@ public static HubConnectionContext Create(ConnectionContext connection, IHubProt
 
         public static MockHubConnectionContext CreateMock(ConnectionContext connection)
         {
-            return new MockHubConnectionContext(connection, TimeSpan.FromSeconds(15), NullLoggerFactory.Instance, TimeSpan.FromSeconds(15));
+            return new MockHubConnectionContext(connection, TimeSpan.FromSeconds(15), NullLoggerFactory.Instance, TimeSpan.FromSeconds(15), 10);
```
jkotalik:
```suggestion
            return new MockHubConnectionContext(connection, TimeSpan.FromSeconds(15), NullLoggerFactory.Instance, TimeSpan.FromSeconds(15), streamBufferCapacity: 10);
```

src/SignalR/server/Core/src/HubConnectionContext.cs
```diff
@@ -50,13 +51,15 @@ public class HubConnectionContext
         /// <param name="keepAliveInterval">The keep alive interval. If no messages are sent by the server in this interval, a Ping message will be sent.</param>
         /// <param name="loggerFactory">The logger factory.</param>
         /// <param name="clientTimeoutInterval">Clients we haven't heard from in this interval are assumed to have disconnected.</param>
-        public HubConnectionContext(ConnectionContext connectionContext, TimeSpan keepAliveInterval, ILoggerFactory loggerFactory, TimeSpan clientTimeoutInterval)
+        /// <param name="streamBufferCapacity">The buffer size for client upload streams</param>
+        public HubConnectionContext(ConnectionContext connectionContext, TimeSpan keepAliveInterval, ILoggerFactory loggerFactory, TimeSpan clientTimeoutInterval, int streamBufferCapacity)
```
jkotalik: I think you need to keep the original constructor here to not introduce a breaking change.

src/SignalR/server/Core/src/HubConnectionHandler.cs
```diff
@@ -73,6 +73,7 @@ public override async Task OnConnectedAsync(ConnectionContext connection)
             var keepAlive = _hubOptions.KeepAliveInterval ?? _globalHubOptions.KeepAliveInterval ?? HubOptionsSetup.DefaultKeepAliveInterval;
             var clientTimeout = _hubOptions.ClientTimeoutInterval ?? _globalHubOptions.ClientTimeoutInterval ?? HubOptionsSetup.DefaultClientTimeoutInterval;
             var handshakeTimeout = _hubOptions.HandshakeTimeout ?? _globalHubOptions.HandshakeTimeout ?? HubOptionsSetup.DefaultHandshakeTimeout;
+            var streamBufferCapacity = _hubOptions.StreamBufferCapacity ?? _globalHubOptions.StreamBufferCapacity?? HubOptionsSetup.DefaultStreamBufferCapacity;
```
jkotalik:
```suggestion
            var streamBufferCapacity = _hubOptions.StreamBufferCapacity ?? _globalHubOptions.StreamBufferCapacity ?? HubOptionsSetup.DefaultStreamBufferCapacity;
```

src/SignalR/server/Core/src/StreamTracker.cs
```diff
@@ -16,14 +16,20 @@ namespace Microsoft.AspNetCore.SignalR
...
+            _streamBufferCapacity = streamBufferCapacity;
+        }
+
         /// <summary>
         /// Creates a new stream and returns the ChannelReader for it as an object.
         /// </summary>
         public object AddStream(string streamId, Type itemType)
         {
-            var newConverter = (IStreamConverter)_buildConverterMethod.MakeGenericMethod(itemType).Invoke(null, Array.Empty<object>());
+            var newConverter = (IStreamConverter)_buildConverterMethod.MakeGenericMethod(itemType).Invoke(null, new object[] { _streamBufferCapacity });
```
jkotalik: Nit: can we reuse the array here rather than newing up one every time? Maybe create it in the constructor?

src/SignalR/server/SignalR/test/AddSignalRTests.cs
davidfowl: @halter73 How would you expect to observe that backpressure?

BrennanConroy:
Are you doing
>2. Add similar option to the client config (also global to the connection) which is used to configure the channel bounds for StreamAsChannelAsync

In this PR?

src/SignalR/server/Core/src/HubConnectionContext.cs
davidfowl: That’s right the breaks the SignalR service

src/SignalR/server/SignalR/test/AddSignalRTests.cs
halter73: Once the bounded channel fills up, the DispatchMessagesAsync loop will wait until ChannelWriter.WriteAsync completes.

Eventually, if no one's reading HubMessages off of the transport's PipeReader, the transport induces backpressure on the client that prevents the client from writing new stream items, invocations or any other message until the backpressure is relieved.

Of course, the transport only sees the messages as bytes over a pipe, and if Kestrel tests have proven one thing: it's proven that it's hard to reliably test that backpressure is induced over an actual socket because there can be arbitrary amounts of buffering happening at different layers.

The real way to reliably test that the StreamBufferCapacity has an effect is to verify the [head-of-line blocking](https://en.wikipedia.org/wiki/Head-of-line_blocking) induced by the bounded channel. I would do this by:

1. Call a Hub method that accepts a stream but doesn't read it.
2. Send enough items from the client to completely fill the bounded Channel.
3. Try to invoke a second hub method.
4. Verify the second hub method doesn't get called.
5. Now read an item from the stream passed to the first hub method.
6. Verify the second hub method now does get called.

Long term, will probably want to use credit based flow control similar to HTTP/2 so clients know how many stream items they can send before inducing head-of-line blocking.

src/SignalR/server/Core/src/StreamTracker.cs
```diff
@@ -88,11 +94,11 @@ private class ChannelConverter<T> : IStreamConverter
         {
             private Channel<T> _channel;
 
-            public ChannelConverter()
+            public ChannelConverter(int streamBufferCapacity)
             {
                 // TODO: Make this configurable or figure out a good limit
```
anurse: Remove this! We did it :)

mikaelm12: True

mikaelm12:
> Are you doing <...> In this PR?

Wasn't planning on it. I'll file a follow up for it.

halter73:
> Are you doing
> > 2. Add similar option to the client config (also global to the connection) which is used to configure the channel bounds for StreamAsChannelAsync

This would require protocol changes to support some sort of flow control like I mentioned in https://github.com/aspnet/AspNetCore/pull/9877#discussion_r280188387 to be truly useful. Otherwise, the client will always empty the channel immediately unless the entire TCP connection is facing backpressure.

src/SignalR/server/Core/src/StreamTracker.cs
```diff
@@ -16,14 +16,20 @@ namespace Microsoft.AspNetCore.SignalR
     internal class StreamTracker
     {
         private static readonly MethodInfo _buildConverterMethod = typeof(StreamTracker).GetMethods(BindingFlags.NonPublic | BindingFlags.Static).Single(m => m.Name.Equals("BuildStream"));
+        private readonly object[] _streamBufferCapacityArg;
         private ConcurrentDictionary<string, IStreamConverter> _lookup = new ConcurrentDictionary<string, IStreamConverter>();
 
+        public StreamTracker(int streamBufferCapacity = 10)
```
BrennanConroy: remove default param, this is an internal class and we should always pass a value.

src/SignalR/server/Core/src/HubOptions.cs
```diff
@@ -46,5 +46,10 @@ public class HubOptions
         /// Detailed error messages include details from exceptions thrown on the server.
         /// </summary>
         public bool? EnableDetailedErrors { get; set; } = null;
+
+        /// <summary>
+        /// Gets or sets the max buffer size for client upload streams.
```
BrennanConroy: Should we include the default value in our comments? I see we only did it for `MaximumReceiveMessageSize`, but I think we should be doing it for all of the options.

src/SignalR/server/SignalR/test/HubConnectionHandlerTests.cs
```diff
@@ -2911,6 +2911,37 @@ bool ExpectedErrors(WriteContext writeContext)
...
+                    options.StreamBufferCapacity = 1;
+                });
+            });
+
+            var connectionHandler = serviceProvider.GetService<HubConnectionHandler<MethodHub>>();
+
+            using (var client = new TestClient())
+            {
+                var connectionHandlerTask = await client.ConnectAsync(connectionHandler).OrTimeout();
+                await client.BeginUploadStreamAsync("invocation", nameof(MethodHub.StreamDontRead), new[] { "id" }, Array.Empty<object>());
```
BrennanConroy: nit: `OrTimeout`

BrennanConroy: nit: "invocationId"

src/SignalR/server/SignalR/test/HubConnectionHandlerTests.cs
```diff
@@ -2911,6 +2911,37 @@ bool ExpectedErrors(WriteContext writeContext)
...
+
+                foreach (var letter in new[] { "A", "B", "C", "D", "E" })
+                {
+                    await client.SendHubMessageAsync(new StreamItemMessage("id", letter)).OrTimeout();
+                }
+
+                var ex = await Assert.ThrowsAsync<TimeoutException>(async () =>
+                {
+                    await client.SendInvocationAsync("Echo", "test");
+                    var result = (CompletionMessage)await client.ReadAsync().OrTimeout(5000);
```
BrennanConroy: :( not a fan

mikaelm12: Not a huge fan either but it does demonstrate the behavior we want.

mikaelm12:
😄

src/SignalR/server/SignalR/test/HubConnectionHandlerTests.cs
```diff
@@ -2911,6 +2911,37 @@ bool ExpectedErrors(WriteContext writeContext)
...
         }
 
+        [Fact]
+        public async Task StreamUploadBufferCapacityBlocksOtherInvocations()
+        {
+            var serviceProvider = HubConnectionHandlerTestUtils.CreateServiceProvider(services =>
+            {
+                services.Configure<HubOptions>(options =>
+                {
+                    options.StreamBufferCapacity = 1;
```
halter73: Wouldn't this test still pass with a StreamBufferCapacity of 5 since you send 5 items? I think the capacity and items sent should match in this test to better demonstrate the relationship.

src/SignalR/server/SignalR/test/HubConnectionHandlerTests.cs
halter73: Can we verify that the CompletionMessage eventually arrives after the hub starts reading the stream items?

halter73 (approved):

src/SignalR/server/SignalR/test/HubConnectionHandlerTests.cs
mikaelm12: 4, but yeah. I agree

src/SignalR/server/SignalR/test/HubConnectionHandlerTests.cs
mikaelm12: Talked about this irl with Stephen. This gets messy because you have to synchronize blocking reading the incoming stream items to verify processing invocations is blocked, then you'd have to unblock (the actual tricky part) to verify that invocations can continue to be processed.

PR #9880, opened 2019-05-01
Seal some internal derived classes

benaadams:
Where they inherit from something and nothing derives from them, or can derive from them as they are internal. Since the Jit finally pays attention to it now.


Eilon (approved):
I am personally a fan of this. I always used to do this and people made fun of me. I feel vindicated 😄

PR #9881, opened 2019-05-01
Expose JsonSerializerOptions directly in SignalR

BrennanConroy:
We now use a binary instead of sources so we can expose the json options directly!

### DIFF
--- a/src/SignalR/common/Protocols.Json/ref/Microsoft.AspNetCore.SignalR.Protocols.Json.netstandard2.0.cs
+++ b/src/SignalR/common/Protocols.Json/ref/Microsoft.AspNetCore.SignalR.Protocols.Json.netstandard2.0.cs
@@ -6,10 +6,7 @@ namespace Microsoft.AspNetCore.SignalR
     public partial class JsonHubProtocolOptions
     {
         public JsonHubProtocolOptions() { }
-        public bool AllowTrailingCommas { get { throw null; } set { } }
-        public bool IgnoreNullValues { get { throw null; } set { } }
-        public bool UseCamelCase { get { throw null; } set { } }
-        public bool WriteIndented { get { throw null; } set { } }
+        public System.Text.Json.Serialization.JsonSerializerOptions PayloadSerializerSettings { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
     }
 }
 namespace Microsoft.AspNetCore.SignalR.Protocol
--- a/src/SignalR/common/Protocols.Json/ref/Microsoft.AspNetCore.SignalR.Protocols.Json.netstandard2.0.cs
+++ b/src/SignalR/common/Protocols.Json/ref/Microsoft.AspNetCore.SignalR.Protocols.Json.netstandard2.0.cs
@@ -6,7 +6,7 @@ namespace Microsoft.AspNetCore.SignalR
     public partial class JsonHubProtocolOptions
     {
         public JsonHubProtocolOptions() { }
-        public System.Text.Json.Serialization.JsonSerializerOptions PayloadSerializerSettings { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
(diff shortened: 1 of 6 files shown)


src/SignalR/common/Protocols.Json/src/JsonHubProtocolOptions.cs
```diff
@@ -11,30 +11,9 @@ namespace Microsoft.AspNetCore.SignalR
...
-                else
-                {
-                    _serializerOptions.PropertyNamingPolicy = null;
-                }
-            }
-        }
+        /// <summary>
+        /// Gets or sets the settings used to serialize invocation arguments and return values.
+        /// </summary>
+        public JsonSerializerOptions PayloadSerializerSettings { get; set; } = JsonHubProtocol.CreateDefaultSerializerSettings();
```
rynowak: Payload is kinda a wierd word. Not wierd like *walking dan back to building 18 the long-way* wierd, but still a little wierd. You do you if SignalR already uses this term for stuff 😆

rynowak: Also a little strange to see `public FooOptions  FooBar*Settings*` - even though we usually use options for `IOptions<>`-related stuff

BrennanConroy: We do :)
https://github.com/aspnet/AspNetCore/blob/master/src/SignalR/common/Protocols.NewtonsoftJson/src/NewtonsoftJsonHubProtocolOptions.cs#L17

src/SignalR/common/SignalR.Common/test/Internal/Protocol/JsonHubProtocolTests.cs
```diff
@@ -24,8 +25,11 @@ protected override IHubProtocol GetProtocolWithOptions(bool useCamelCase, bool i
         {
             var protocolOptions = new JsonHubProtocolOptions()
             {
-                IgnoreNullValues = ignoreNullValues,
-                UseCamelCase = useCamelCase,
+                PayloadSerializerSettings = new JsonSerializerOptions()
```
halter73: This line does highlight the Settings vs Options weirdness.

src/SignalR/common/Protocols.Json/src/JsonHubProtocolOptions.cs
anurse: I'm generally all for consistency but also happy to re-open this bikeshed a little given the breaking-ness of the protocol change in 3.0.

"Payload" was chosen original because the serializer options **only apply to arguments and return-values** as opposed to applying to the entire protocol packet. I still think that's the most reasonable name but am happy to entertain alternatives :).

As for "Settings", we used that when the type we configured was `JsonSerializerSettings` but now it's `JsonSerializerOptions`. I'm OK with changing that to `Options`.

`PayloadSerializerOptions` makes sense to me if we're keeping Payload around.

halter73: 👍 for `PayloadSerializerOptions` but no biggie.

halter73 (approved):

src/SignalR/common/Protocols.Json/ref/Microsoft.AspNetCore.SignalR.Protocols.Json.netstandard2.0.cs
```diff
@@ -6,10 +6,7 @@ namespace Microsoft.AspNetCore.SignalR
     public partial class JsonHubProtocolOptions
     {
         public JsonHubProtocolOptions() { }
-        public bool AllowTrailingCommas { get { throw null; } set { } }
-        public bool IgnoreNullValues { get { throw null; } set { } }
-        public bool UseCamelCase { get { throw null; } set { } }
-        public bool WriteIndented { get { throw null; } set { } }
+        public System.Text.Json.Serialization.JsonSerializerOptions PayloadSerializerSettings { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
```
mikaelm12: Re run the tool

mikaelm12 (approved):
Re run the tool and you should be fine

PR #9898, opened 2019-05-01
Remove HubProtocol minor version

### LINKED ISSUE #5320: Remove minorVersion from handshake response
~~The master branch needs to be updated with the new information regarding the `minorVersion` field in the Handshake Response.~~

Let's remove the minorVersion from the handshake response, since we aren't using it yet (see comments below).

mikaelm12:
Fixes: #5320 
We introduced the HubProtocol minor version in 3.0. We don't plan on using it (we can add it back later if we change our minds) and it's a breaking change so let's remove it,

### DIFF
--- a/src/SignalR/common/Protocols.Json/ref/Microsoft.AspNetCore.SignalR.Protocols.Json.netcoreapp3.0.cs
+++ b/src/SignalR/common/Protocols.Json/ref/Microsoft.AspNetCore.SignalR.Protocols.Json.netcoreapp3.0.cs
@@ -18,7 +18,6 @@ namespace Microsoft.AspNetCore.SignalR.Protocol
     {
         public JsonHubProtocol() { }
         public JsonHubProtocol(Microsoft.Extensions.Options.IOptions<Microsoft.AspNetCore.SignalR.JsonHubProtocolOptions> options) { }
-        public int MinorVersion { get { throw null; } }
         public string Name { get { throw null; } }
         public Microsoft.AspNetCore.Connections.TransferFormat TransferFormat { get { throw null; } }
         public int Version { get { throw null; } }
--- a/src/SignalR/common/SignalR.Common/src/Protocol/HandshakeResponseMessage.cs
+++ b/src/SignalR/common/SignalR.Common/src/Protocol/HandshakeResponseMessage.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 namespace Microsoft.AspNetCore.SignalR.Protocol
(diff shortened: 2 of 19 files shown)


src/SignalR/common/SignalR.Common/src/Protocol/HandshakeResponseMessage.cs
```diff
@@ -18,34 +18,13 @@ public class HandshakeResponseMessage : HubMessage
         /// </summary>
         public string Error { get; }
 
-        /// <summary>
-        /// Highest minor protocol version that the server supports.
-        /// </summary>
-        public int MinorVersion { get; }
```
halter73: We should include the ProtocolVersion in the handshake response now IMHO. We could add it later, but it might be quicker and easier to to just replace MinorVersion with ProtocolVersion instead of making that two separate changes.

mikaelm12:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=175569&view=logs

aspnet-hello:

src/SignalR/common/SignalR.Common/src/Protocol/HandshakeResponseMessage.cs
mikaelm12: I get your point but along that line of logic we should just leave the Minor protocol version though, right? If we don't plan to use it in the handshake response now then I think we might as well leave it out.

halter73: I don't feel too strongly about it. I'm fine just removing it for now.

halter73 (approved):

src/SignalR/common/Protocols.Json/ref/Microsoft.AspNetCore.SignalR.Protocols.Json.netcoreapp3.0.cs
```diff
@@ -18,7 +18,6 @@ public sealed partial class JsonHubProtocol : Microsoft.AspNetCore.SignalR.Proto
     {
         public JsonHubProtocol() { }
         public JsonHubProtocol(Microsoft.Extensions.Options.IOptions<Microsoft.AspNetCore.SignalR.JsonHubProtocolOptions> options) { }
-        public int MinorVersion { get { throw null; } }
```
halter73: This was new in 3.0, right?

mikaelm12: Yup. Here's the 2.2 branch of the JsonHubProtocol https://github.com/aspnet/AspNetCore/blob/release/2.2/src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
No reference to MinorVersion

mikaelm12:
Some benchmarking data for this change. Some minor improvements on Handshake.

Before

Method |     Mean |     Error |   StdDev |      Op/s |  Gen 0 | Allocated |
---------------------- |---------:|----------:|---------:|----------:|-------:|----------:|
 SuccessHandshakeAsync | 6.385 us | 0.6681 us | 1.970 us | 156,606.9 | 0.1297 |     680 B |

After

Method |     Mean |     Error |   StdDev |      Op/s |  Gen 0 | Allocated |
---------------------- |---------:|----------:|---------:|----------:|-------:|----------:|
 SuccessHandshakeAsync | 6.340 us | 0.5179 us | 1.519 us | 157,723.7 | 0.1526 |     648 B |

PR #9899, opened 2019-05-01
Localization prep - move strings from theme.xml to theme.wxl

joeloff:
Moving hardcoded strings from the bundle theme into wxl as part of localization prep work.


Eilon (approved):
The change looks alright. Where will actual loc work happen, though?

joeloff:
For now the .wxl files will be picked up from src and we'll get localized copies back. Those would go into the repo and we'll add them to setup as additional payloads.

natemcmaster (approved):

PR #9903, opened 2019-05-01
fix #7916 by using DOTNET_RUNNING_IN_CONTAINERS env var to detect container

### LINKED ISSUE #7916: Data Protection key generation warnings should also happen on Kubernetes
### Background
The default behaviour of [ASP.NET Core Data Protection](https://docs.microsoft.com/en-us/aspnet/core/security/data-protection/introduction) is to create a data protection key ring when the web application starts.  It is documented [here](https://docs.microsoft.com/en-us/aspnet/core/security/data-protection/configuration/default-settings) that `When hosting in a Docker container, keys should be persisted in a folder that's a Docker volume (a shared volume or a host-mounted volume that persists beyond the container's lifetime) or in an external provider ...`

For users who don't read the extensive documentation, there is also a runtime check in `FileSystemXmlRepository` that logs a warning if the documentation hasn't been followed.

This feature request is about extending that existing check.

Currently, the check only checks for Docker, but there are many other container platforms, for example Kubernetes, which it doesn't currently detect, and therefore it doesn't generate the log entry on those platforms.

The impact this might have on users is that their keys are lost, or that each replica of the web application creates its own key, leading to a confusing situation where if the load balancer ever sends traffic to a different replica an error would occur.  I believe that the development work to implement this feature request is justified because of:
1. the severity of these impacts
2. plus because we are talking about the default behaviour here the likelihood is very high

The log entry is generated here: https://github.com/aspnet/AspNetCore/blob/def36fab1e45ef7f169dfe7b59604d0002df3b7c/src/DataProtection/DataProtection/src/Repositories/FileSystemXmlRepository.cs#L37

And the test for Docker is here: https://github.com/aspnet/AspNetCore/blob/def36fab1e45ef7f169dfe7b59604d0002df3b7c/src/DataProtection/DataProtection/src/Internal/DockerUtils.cs#L96

The problem is that it tests for an internal implementation detail of Docker, whereas it would be better if that test was a more general test for containers.  An example might be to see if the environment variable `DOTNET_RUNNING_IN_CONTAINER` is set to `true`

### Describe alternatives you've considered
Re-designing the Data Protection system or changing its defaults would also be possible, but I think that the least intrusive solution is changing the detection code to show the existing log message when it is relevant.

analogrelay:
Fixes #7916

I was assigned this back before @natemcmaster got back :). Finally just decided to do it rather than punting it around. Testing this looks tricky, I'll do a quick manual pass once it lands in the SDK. It's a pretty simple change.

All the official .NET Core images set this environment variable (Windows and Linux) so it should serve as a reasonable first thing to look for and make us a little more resiliant to running in other container runtimes.


src/DataProtection/DataProtection/src/Internal/DockerUtils.cs
```diff
@@ -11,13 +11,13 @@ namespace Microsoft.AspNetCore.DataProtection.Internal
 {
     internal static class DockerUtils
```
natemcmaster: As long as we're at it, rename to `ContainerUtils`?

natemcmaster (approved):

src/DataProtection/DataProtection/src/Internal/DockerUtils.cs
anurse: Yeah, I thought about that. I'll do it :)

blowdart (approved):
Looks fine to me

anurse:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=175726

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/2450

anurse:
> The remote name could not be resolved: 'xsp2tfsprodcus2file0.queue.core.windows.net'

:( Will retrigger when Azure is feeling less under-the-weather

anurse:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=180794

aspnet-hello:

src/DataProtection/DataProtection/src/Internal/ContainerUtils.cs
```diff
@@ -77,14 +77,21 @@ internal static bool IsDirectoryMounted(DirectoryInfo directory, IEnumerable<str
             return false;
         }
 
-        private static bool IsProcessRunningInDocker()
+        private static bool IsProcessRunningInContainer()
         {
+            // Official .NET Core images (Windows and Linux) set this. So trust it if it's there.
+            if(string.Equals(Environment.GetEnvironmentVariable("DOTNET_RUNNING_IN_CONTAINERS"), "true", StringComparison.OrdinalIgnoreCase))
```
DavidSimner: There's a typo here, instead of `DOTNET_RUNNING_IN_CONTAINERS` it should be `DOTNET_RUNNING_IN_CONTAINER`.

See, e.g. https://github.com/dotnet/dotnet-docker/blob/master/2.1/runtime-deps/stretch-slim/amd64/Dockerfile#L21 for where it is set, or e.g. https://www.hanselman.com/blog/DetectingThatANETCoreAppIsRunningInADockerContainerAndSkippableFactsInXUnit.aspx for another source.

anurse: Looks like the environment variable isn't consistent... See https://github.com/dotnet/dotnet-docker/blob/4fd3b74fc5b7444f39f3fdcffa45eef467bccb24/2.1/aspnet/nanoserver-1803/amd64/Dockerfile

anurse: Perhaps we should just check both then :)

anurse: Never mind, I see that the plural is deprecated. Will update!

PR #9906, opened 2019-05-02
Add a top-level node for analyzers

rynowak:
This is the start of a reorg of the analyzers we *ship*. Currently we have some analyzers sprinkled around for various features, but we're also writing analyzers that cover ASP.NET Core as a whole.


src/Analyzers/Analzyers/test/Microsoft.AspNetCore.Analyzers.Test.csproj
```diff
@@ -0,0 +1,30 @@
...
+    <PreserveCompilationContext>true</PreserveCompilationContext>
+    <RootNamespace>Microsoft.AspNetCore.Analyzers</RootNamespace>
+    <!-- https://github.com/aspnet/AspNetCore/issues/6549 -->	
+    <BuildHelixPayload>false</BuildHelixPayload>
+   </PropertyGroup>
+
+  <ItemGroup>
+    <None Include="xunit.runner.json" CopyToOutputDirectory="PreserveNewest" />
+    <Compile Include="$(SharedSourceRoot)test\SkipOnHelixAttribute.cs" />
+    <Content Include="TestFiles\**\*.*" CopyToPublishDirectory="PreserveNewest" />
```
rynowak: The pattern here is kinda interesting. These are both source files in the tests and content that are used by the tests.

src/Analyzers/Analzyers/test/Microsoft.AspNetCore.Analyzers.Test.csproj
```diff
@@ -0,0 +1,30 @@
...
+    <ProjectReference Include="..\..\Mvc.Analyzers\src\Microsoft.AspNetCore.Mvc.Analyzers.csproj" />
+    <Reference Include="Microsoft.AspNetCore.Analyzer.Testing" />
+    <Reference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" />
+
+    <!-- 
+        These tests include startup code for most parts of our stack. This list will grow as we add more.
+    -->
+    <Reference Include="Microsoft.AspNetCore.Components.Server" />
+    <Reference Include="Microsoft.AspNetCore.Mvc" />
+    <Reference Include="Microsoft.AspNetCore.SignalR" />
```
rynowak: Currently blocked on adding these to the solution.

natemcmaster:
Seems fine so far. The CI isn't actually building this code yet though. To get CI and builds, we have two options:

1. Include the src/Analyzers/ folder in the main repo build (add it to this list https://github.com/aspnet/AspNetCore/blob/master/build/repo.props#L145). The packages will build as a part of the runtime build and version with it.
2. Introduce a new, unique pipeline that only builds code from src/Analyzers/. We haven't done this before, but I believe it should be possible. It would give us the ability to produce builds for just the analyzers, without needing to rebuild the entire runtime. Since analyzers would version and ship on their own schedule that can be aligned to VS and the SDK instead of the runtime.

Thoughts?

src/Analyzers/Analzyers/test/Microsoft.AspNetCore.Analyzers.Test.csproj
natemcmaster: This is the issue we talked about yesterday, right? Did you get the 'AddAllProjectsToSolution' script to work?

rynowak:
I think option 1 actually makes the most sense. There will be cases where product code changes will break analyzer tests. I like the idea that we'd do 2 also, but I still think 1 is needed.

src/Analyzers/Analzyers/test/Microsoft.AspNetCore.Analyzers.Test.csproj
rynowak: Yeah I got this to work thanks!

rynowak:
@natemcmaster @pranavkm - ready for review. 

My immediate need is to get the code that's in the shared package into a feed for the publish folks to use. There will be a few more steps/PRs to reorganise the existing code.

build/repo.props
```diff
@@ -166,6 +166,7 @@
                           $(RepositoryRoot)src\MusicStore\**\*.*proj;
                           $(RepositoryRoot)src\SignalR\**\*.csproj;
                           $(RepositoryRoot)src\Components\**\*.csproj;
+                          $(RepositoryRoot)src\Analyzers\**\*.csproj;
```
rynowak: @natemcmaster does order matter here? Right before templates seemed like the right place if it does.

eng/ProjectReferences.props
```diff
@@ -131,5 +131,6 @@
...
     <ProjectReferenceProvider Include="Microsoft.AspNetCore.Components" ProjectPath="$(RepositoryRoot)src\Components\Components\src\Microsoft.AspNetCore.Components.csproj" RefProjectPath="$(RepositoryRoot)src\Components\Components\ref\Microsoft.AspNetCore.Components.csproj" />
     <ProjectReferenceProvider Include="Microsoft.AspNetCore.Components.Server" ProjectPath="$(RepositoryRoot)src\Components\Server\src\Microsoft.AspNetCore.Components.Server.csproj" RefProjectPath="$(RepositoryRoot)src\Components\Server\ref\Microsoft.AspNetCore.Components.Server.csproj" />
+    <ProjectReferenceProvider Include="Microsoft.AspNetCore.Analyzers.FeatureDetection.Sources" ProjectPath="$(RepositoryRoot)src\Analyzers\shared\FeatureDetection\Microsoft.AspNetCore.Analyzers.FeatureDetection.Sources.csproj" RefProjectPath="$(RepositoryRoot)src\Analyzers\shared\ref\Microsoft.AspNetCore.Analyzers.FeatureDetection.Sources.csproj" />
```
rynowak: Should I expect to see both projects in here?

rynowak: Also, did I do the source package right? I don't expect this to product a ref....

eng/Versions.props
```diff
@@ -150,9 +150,9 @@
     <MicrosoftAzureStorageBlobPackageVersion>10.0.1</MicrosoftAzureStorageBlobPackageVersion>
     <MicrosoftBuildFrameworkPackageVersion>15.8.166</MicrosoftBuildFrameworkPackageVersion>
     <MicrosoftBuildUtilitiesCorePackageVersion>15.8.166</MicrosoftBuildUtilitiesCorePackageVersion>
-    <MicrosoftCodeAnalysisCommonPackageVersion>2.8.0</MicrosoftCodeAnalysisCommonPackageVersion>
-    <MicrosoftCodeAnalysisCSharpPackageVersion>2.8.0</MicrosoftCodeAnalysisCSharpPackageVersion>
-    <MicrosoftCodeAnalysisCSharpWorkspacesPackageVersion>2.8.0</MicrosoftCodeAnalysisCSharpWorkspacesPackageVersion>
+    <MicrosoftCodeAnalysisCommonPackageVersion>3.0.0</MicrosoftCodeAnalysisCommonPackageVersion>
+    <MicrosoftCodeAnalysisCSharpPackageVersion>3.0.0</MicrosoftCodeAnalysisCSharpPackageVersion>
+    <MicrosoftCodeAnalysisCSharpWorkspacesPackageVersion>3.0.0</MicrosoftCodeAnalysisCSharpWorkspacesPackageVersion>
```
rynowak: This version should be supported everywhere we are supported. We will need some new features from this version in the very near future.

src/Analyzers/Analyzers/src/Microsoft.AspNetCore.Analyzers.csproj
```diff
@@ -0,0 +1,40 @@
...
+        iconUrl=$(PackageIconUrl);
+        repositoryUrl=$(RepositoryUrl);
+        repositoryCommit=$(SourceRevisionId);
+        copyright=$(Copyright);
+
+        OutputBinary=$(OutputPath)$(AssemblyName).dll;
+        OutputSymbol=$(OutputPath)$(AssemblyName).pdb;
+      </NuspecProperties>
+    </PropertyGroup>
+  </Target>
```
rynowak: I'm following the pattern of the analyzers we currently have in the repo.

src/Analyzers/Analyzers/src/Microsoft.AspNetCore.Analyzers.csproj
```diff
@@ -0,0 +1,40 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <Description>CSharp Analyzers for ASP.NET Core.</Description>
+    <PackageTags>aspnetcore</PackageTags>
+     <!-- This package is for internal use only. It contains analyzers that are bundled in the .NET Core WebSDK. -->
+    <IsShippingPackage>false</IsShippingPackage>
+
+    <TargetFramework>netstandard2.0</TargetFramework>
```
rynowak: As of 3.0 Roslyn is netstandard2.0. This should work everywhere we work.

src/Analyzers/Analyzers/src/Microsoft.AspNetCore.Analyzers.csproj
```diff
@@ -0,0 +1,40 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <Description>CSharp Analyzers for ASP.NET Core.</Description>
+    <PackageTags>aspnetcore</PackageTags>
+     <!-- This package is for internal use only. It contains analyzers that are bundled in the .NET Core WebSDK. -->
+    <IsShippingPackage>false</IsShippingPackage>
```
rynowak: These will go in the SDK, so there's no need to produce a package.

src/Analyzers/Analyzers/src/StartupFacts.cs
```diff
@@ -0,0 +1,126 @@
...
+        //
+        // The ConfigureServices method follows the naming pattern `Configure{Environment?}Services` (ignoring case).
+        // The ConfigureServices method must be public.
+        // The ConfigureServices method can be instance or static.
+        // The ConfigureServices method cannot have other parameters besides IServiceCollection.
+        //
+        // The ConfigureServices method does not actually need to accept IServiceCollection
+        // but we exclude that case because a ConfigureServices method that doesn't accept an
+        // IServiceCollection can't do anything interesting to analysis.
+        public static bool IsConfigureServices(StartupSymbols symbols, IMethodSymbol symbol)
```
rynowak: @Tratcher - any cool thoughts?

rynowak: I'm making the simplifying assumption that this analyzer won't work with the pattern where you return an `IServiceProvider`. The analyzers are for 3.0+

src/Analyzers/Analyzers/src/SymbolNames.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    internal static class SymbolNames
+    {
+        public static readonly string ConfigureServicesMethodPrefix = "Configure";
+
+        public static readonly string ConfigureServicesMethodSuffix = "Services";
+
+        public static readonly string ConfigureMethodPrefix = "Configure";
+
+        public static class IApplicationBuilder
```
rynowak: I use this pattern in a few other places like Razor where we need to keep track of type names and other constants related to types/methods. Defining a hierarchy like this makes it easy to complete in VS.

src/Mvc/Mvc.Analyzers/src/Startup/StartupAnalysisContext.cs
```diff
@@ -7,6 +7,7 @@ namespace Microsoft.AspNetCore.Analyzers
 {
     internal class StartupAnalysisContext
     {
+#pragma warning disable RS1012 // Start action has no registered actions.
```
rynowak: The 3.0 Roslyn package comes with some new rules for analyzers/codefixes. 

I have another change planned that avoids this warning so suppressing it for now.

src/Mvc/Mvc.Api.Analyzers/src/AddResponseTypeAttributeCodeFixAction.cs
```diff
@@ -37,6 +37,8 @@ public AddResponseTypeAttributeCodeFixAction(Document document, Diagnostic diagn
             _diagnostic = diagnostic;
         }
 
+        public override string EquivalenceKey => _diagnostic.Location.ToString();
```
rynowak: @pranavkm

src/Analyzers/Analyzers/src/Microsoft.AspNetCore.Analyzers.csproj
```diff
@@ -0,0 +1,40 @@
...
+
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <IncludeBuildOutput>false</IncludeBuildOutput>
+    <GenerateDocumentationFile>false</GenerateDocumentationFile>
+    <NuspecFile>$(MSBuildProjectName).nuspec</NuspecFile>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Reference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" PrivateAssets="All" />
```
rynowak: @pranavkm - so it turns out that this is required for code fixes. I think the idea is that you're safe to depend on this assembly in your analyzers but expected not to use any of the code from unless something calls your codefix. Seems scary to me but I think this is the recommended pattern.

src/Analyzers/Analyzers/test/StartupFactsTest.cs
```diff
@@ -0,0 +1,142 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Analyzers.TestFiles.StartupFactsTest;
+using Microsoft.CodeAnalysis;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Analyzers
+{
+    public class StartupFactsTest : AnalyzerTestBase
+    {
+        [Theory]
```
pranavkm: 😞

build/repo.props
natemcmaster: Order does not matter.

src/Analyzers/Analyzers/src/Microsoft.AspNetCore.Analyzers.csproj
natemcmaster: If there is no need to make a .nupkg file, set `<IsPackable>false</IsPackable>`.

eng/ProjectReferences.props
natemcmaster: Looks like the .Sources project is still producing a binary, so no..this wasn't done right. This repo doesn't produce any .Sources packages right now, so this will be the first. Recommendations given below.

src/Analyzers/shared/FeatureDetection/Microsoft.AspNetCore.Analyzers.FeatureDetection.Sources.csproj
```diff
@@ -0,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
natemcmaster:
```suggestion
    <!-- This does not represent the TFM for the code. It's only here because /t:Pack requires it. -->
    <TargetFramework>netstandard1.0</TargetFramework>
    <IsPackable>true</IsPackable>
    <IsShipping>false</IsShipping>
    <NoBuild>true</NoBuild>
    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
    <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
    <EnableDefaultItems>false</EnableDefaultItems>
    <IncludeBuildOutput>false</IncludeBuildOutput>
    <IncludeSymbols>false</IncludeSymbols>
    <ContentTargetFolders>contentFiles</ContentTargetFolders>
    <DisableImplicitFrameworkReferences>true</DisableImplicitFrameworkReferences>
    <DefaultExcludeItems>$(DefaultExcludeItems);$(BaseOutputPath);$(BaseIntermediateOutputPath);</DefaultExcludeItems>
    <NoWarn>$(NoWarn);NU5105;CS8021</NoWarn>
    <IsProjectReferenceProvider>false</IsProjectReferenceProvider>
  </PropertyGroup>

  <ItemGroup>
    <Compile Include="$(MSBuildProjectDirectory)\**\*.cs" Exclude="$(DefaultExcludeItems)">
      <Pack>true</Pack>
      <PackagePath>$(ContentTargetFolders)\cs\netstandard1.0\shared\</PackagePath>
    </Compile>
  </ItemGroup>
```

Here is the goop to make this a .Sources packages, contain only content, not a binary.

src/Analyzers/shared/FeatureDetection/Microsoft.AspNetCore.Analyzers.FeatureDetection.Sources.csproj
```diff
@@ -0,0 +1,16 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Reference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Compile Include="../../Analyzers/src/StartupFacts.cs" Link="%(FileName)%(Extension)" />
```
natemcmaster:
```suggestion
    <Compile Include="../../Analyzers/src/StartupFacts.cs">
      <Pack>true</Pack>
      <PackagePath>$(ContentTargetFolders)\cs\netstandard1.0\shared\</PackagePath>
    </Compile>
```

Do this for all .cs files

natemcmaster (changes requested):
A few changes required to make .Sources packages actually work. 

Once you have the updates, can you ping me again? I want to inspect the actual .nupkg file to make sure it has the right content and layout.

eng/ProjectReferences.props
rynowak: Should I undo changes to this file after fixing that?

src/Analyzers/Analyzers/src/StartupFacts.cs
```diff
@@ -0,0 +1,126 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.CodeAnalysis;
+
+namespace Microsoft.AspNetCore.Analyzers
+{
+    internal static class StartupFacts
+    {
+        public static bool IsStartupClass(StartupSymbols symbols, INamedTypeSymbol type)
```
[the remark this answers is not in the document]
rynowak: I left this out because you can get around this by using `UseStartup<>` I tried to avoid things that would give us false negatives.

src/Analyzers/Analyzers/src/StartupFacts.cs
```diff
@@ -0,0 +1,126 @@
...
+                throw new ArgumentNullException(nameof(symbol));
+            }
+
+            if (symbol.DeclaredAccessibility != Accessibility.Public)
+            {
+                return false;
+            }
+
+            if (symbol.Name == null ||
+                !symbol.Name.StartsWith(SymbolNames.ConfigureServicesMethodPrefix, StringComparison.OrdinalIgnoreCase) ||
```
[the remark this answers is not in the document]
rynowak: VB though. This is a little more friendly for VB people.

src/Analyzers/Analyzers/src/StartupFacts.cs
```diff
@@ -0,0 +1,126 @@
...
+                !symbol.Name.EndsWith(SymbolNames.ConfigureServicesMethodSuffix, StringComparison.OrdinalIgnoreCase))
+            {
+                return false;
+            }
+
+            if (symbol.Parameters.Length != 1)
```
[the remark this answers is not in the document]
rynowak: I avoided the patterns where you return something, because those are legacy and this targets 3.0+

src/Analyzers/Analyzers/test/StartupFactsTest.cs
rynowak: Yeah, it's really the most reasonable thing to do for these cases.

src/Analyzers/Analyzers/src/SymbolNames.cs
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.CodeAnalysis;
+
+namespace Microsoft.AspNetCore.Analyzers
+{
+    internal static class SymbolNames
+    {
+        public static readonly string ConfigureServicesMethodPrefix = "Configure";
```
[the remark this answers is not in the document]
rynowak: Habit...

src/Analyzers/shared/FeatureDetection/Microsoft.AspNetCore.Analyzers.FeatureDetection.Sources.csproj
rynowak: My goal was to have this project still compile in VS. It looks like if this is going to be ns10 then that is busto. Should I remove the package references and remove from the solution?

rynowak:
@natemcmaster updated

src/Analyzers/Analyzers/src/Microsoft.AspNetCore.Analyzers.csproj
```diff
@@ -0,0 +1,42 @@
...
+  </ItemGroup>
+
+  <Target Name="PopulateNuspec" BeforeTargets="GenerateNuspec">
```
natemcmaster: I see you set IsPackable = false. Is there any point to keeping this target and the .nuspec file around?

src/Analyzers/shared/FeatureDetection/Microsoft.AspNetCore.Analyzers.FeatureDetection.Sources.csproj
```diff
@@ -0,0 +1,45 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
natemcmaster: I see there is one more thing you have to do. You'll need to change to explicit top/bottom imports like this:

```xml
<Project>
  <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk" />

  rest of project stuff goes here

  <Import Project="Sdk.targets" Sdk="Microsoft.NET.Sdk" />

  <!-- This project does not produce a binary. Disable the compiler. -->
  <Target Name="Compile" />
  <Target Name="CopyFilesToOutputDirectory" />
</Project>
```

src/Analyzers/shared/FeatureDetection/Microsoft.AspNetCore.Analyzers.FeatureDetection.Sources.csproj
natemcmaster: We haven't tried to do that with other .Sources packages because there wasn't any demand. Typically code in a .Sources package are compiled by at least one other project in the repo (and if they code isn't...then I have to ask, why is it here?)

But if you want to try and make it work, that's okay with me.

src/Analyzers/Analyzers/src/Microsoft.AspNetCore.Analyzers.csproj
rynowak: Actually I'm wrong, we do need to produce a package for this. We need this so it can be picked up as an asset for the web sdk.

rynowak:
@natemcmaster  updated

natemcmaster (approved):

natemcmaster:
LGTM. Thanks for updating.

rynowak:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=182265&view=logs

aspnet-hello:

PR #9908, opened 2019-05-02
Remove AsyncLocal from HeaderPropagation middleware

benaadams:
Request scope `HeaderPropagationValues` rather than storing it in an `AsyncLocal`

Contributes to: https://github.com/aspnet/AspNetCore/issues/9813


rynowak:
This is the opposite of how we want this to work 😆

benaadams:
> This is the opposite of how we want this to work 😆

Hmm?

Tratcher:
> This is the opposite of how we want this to work 😆

Oh? It could work. 

I recall injecting into Invoke is slow because it uses ActivatorUtilities.

benaadams:
> I recall injecting into Invoke is slow because it uses ActivatorUtilities.

O RLY? Challenge accepted...

Tratcher:
> > I recall injecting into Invoke is slow because it uses ActivatorUtilities.
> 
> O RLY? Challenge accepted...

Certainly slower than constructor injection. DI does a lot of optimizations and caching for constructor injection, but Invoke injection is just GetRequiredService in a loop. Invoke injection is also per request which makes the effect far more significant.

rynowak:
> Oh? It could work.
> 
> I recall injecting into Invoke is slow because it uses ActivatorUtilities.

We decided to use AsyncLocal because of the thread safety and fire-and-forget issues. A few people tried implementing this using `IHttpContextAccessor` and ended up injuring themselves.

Tratcher:
I understand, but they were trying to access HttpContext inside SendAsync. Doing it in the handler constructor does look safer, that should only be created once per request, correct?

rynowak:
No, the handler is cached for an arbitrary amount of time.

benaadams:
>  A few people tried implementing this using `IHttpContextAccessor` and ended up injuring themselves.

`IHttpContextAccessor` isn't used to get headers from HttpContext; its just used to get the `RequestServices` container, because `DefaultHttpClientFactory` [sets up a subscope](https://github.com/aspnet/Extensions/blob/557995ec322f1175d6d8a72a41713eec2d194871/src/HttpClientFactory/Http/src/DefaultHttpClientFactory.cs#L159-L160) which hides it

rynowak:
> `IHttpContextAccessor` isn't used to get headers from HttpContext; its just used to get the `RequestServices` container, because `DefaultHttpClientFactory` [sets up a subscope](https://github.com/aspnet/Extensions/blob/557995ec322f1175d6d8a72a41713eec2d194871/src/HttpClientFactory/Http/src/DefaultHttpClientFactory.cs#L159-L160) which hides it

The message handler will outlive the request scope, and will be reused across many requests.

benaadams:
> No, the handler is cached for an arbitrary amount of time.

Ah, should be able to tweak for that.

rynowak:
> 
> 
> > No, the handler is cached for an arbitrary amount of time.
> 
> Ah, should be able to tweak for that.

The way I see this we have two places to run code:
1. In the message handler constructor (bad because cached)
2. In the SendAsync method (bad because threading and fire-and-forget)

benaadams:
Can solve this except in one scenario; if SendAsync is called **after** the request has finished.

i.e. is it 

1. fire and forget of SendAsync or 
2. a whole independent Task.Run that's fire and forgotten, and at some point in the future calls SendAsync?

rynowak:
Yes. These are the reasons why we decided on this solution.

src/Middleware/HeaderPropagation/src/HeaderPropagationMessageHandler.cs
```diff
@@ -49,6 +64,12 @@ public HeaderPropagationMessageHandler(IOptions<HeaderPropagationOptions> option
         /// <returns>The task object representing the asynchronous operation.</returns>
         protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
         {
+            var inputValues = _accessor?.HttpContext?.RequestServices.GetService<HeaderPropagationValues>() ??
```
Tratcher: Nope, HttpContext is completely off limits here due to mutithreading.

benaadams: Will have to go bigger then....

rynowak:
@benaadams - we intentionally went with async-local as a solution so that this can support fire-and-forget work and multithreaded work that's spawned from a request. Users that tried to build header propagation on their own was a common source of people trying to use `HttpContext` in an unsupported way, which is what led us here 😆 

If you (or someone like you) plans to use this and wants to make it work without the async local, I'd be happy to accept a PR like this with an option as long as the current "works all of the time" behaviour is the default. 

I'm still not totally convinced by this PR because it relies on `IHttpContextAccessor`. So if you're not already using that, you just traded one dependency on async-local for another.

benaadams:
There is a cost per async local; both in allocations from in and out of a setting async method for the changed immuatable async local dictionaries and the new `ExecutionContext`s to support them; and in the increased cost of restoring the contexts; though as long as they don't have change notifications its not too bad.

However; I think it needs a more holistic approach for all the ASP.NET Core `AsyncLocal`s, rather than looking at one in isolation. So closing this 😄

rynowak:
> However; I think it needs a more holistic approach for all the ASP.NET Core `AsyncLocal`s, rather than looking at one in isolation. So closing this 😄

OK, looking forward to what you come up with!

benaadams:
Related allocations 132MB and 3.4M objects for 312k requests with just 2 AyncLocals (Activity and logging scope) and the growth as you add them isn't linear.

image

PR #9919, opened 2019-05-02
Add --mutex option to yarn commands

natemcmaster:
Attempt to improve reliability of running yarn commands. Our project structure might cause yarn to be launched multiple times. According to yarn docs, this option should avoid conflicts between multiple instances of yarn.
https://yarnpkg.com/en/docs/cli/#toc-concurrency-and-mutex

Also, testing to see if removing the `yarn add puppeteer ` command helps with reliability and perf. The puppeteer package is already part of the package.json file for the CORS functional tests.

cref https://github.com/aspnet/AspNetCore-Internal/issues/2383

### DIFF
--- a/eng/targets/Npm.Common.targets
+++ b/eng/targets/Npm.Common.targets
@@ -18,7 +18,7 @@
 
   <Target Name="Restore">
     <Message Importance="High" Text="Running yarn install on $(MSBuildProjectFullPath)" />
-    <Yarn Command="install --mutex file $(InstallArgs)" StandardOutputImportance="High" StandardErrorImportance="High" />
+    <Yarn Command="install --mutex network $(InstallArgs)" StandardOutputImportance="High" StandardErrorImportance="High" />
   </Target>
 
   <Target Name="PrepareForBuild">
--- a/eng/targets/Npm.Common.targets
+++ b/eng/targets/Npm.Common.targets
@@ -18,7 +18,7 @@
 
   <Target Name="Restore">
     <Message Importance="High" Text="Running yarn install on $(MSBuildProjectFullPath)" />
-    <Yarn Command="install $(InstallArgs)" StandardOutputImportance="High" StandardErrorImportance="High" />
+    <Yarn Command="install --mutex file $(InstallArgs)" StandardOutputImportance="High" StandardErrorImportance="High" />
   </Target>
 
   <Target Name="PrepareForBuild">
--- a/src/Middleware/CORS/test/FunctionalTests/CORS.FunctionalTests.npmproj
+++ b/src/Middleware/CORS/test/FunctionalTests/CORS.FunctionalTests.npmproj
@@ -12,10 +12,5 @@
     <IsTestProject>false</IsTestProject>
   </PropertyGroup>
 
-  <Target Name="InstallPuppeteer" BeforeTargets="Restore">
-    <!-- Explicitly install puppeteer. This will install the bundled Chromium as part of restore instead of as part of execution -->
-    <Yarn Command="add puppeteer" />
-  </Target>
-
   <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), Directory.Build.targets))\Directory.Build.targets" />
 </Project>
--- a/src/Middleware/NodeServices/src/Microsoft.AspNetCore.NodeServices.csproj
+++ b/src/Middleware/NodeServices/src/Microsoft.AspNetCore.NodeServices.csproj
@@ -21,7 +21,7 @@
 
   <Target Name="YarnInstall">
     <Message Text="Running yarn install on $(MSBuildProjectFile)" Importance="High" />
-    <Yarn Command="install" />
+    <Yarn Command="install --mutex file" />
   </Target>
 
   <Target Name="_IsCustomRestoreTargetSupported" Returns="@(CustomRestoreTargets)" Condition="'$(BuildNodeJs)' == 'true'">
--- a/src/Middleware/NodeServices/src/Microsoft.AspNetCore.NodeServices.csproj
+++ b/src/Middleware/NodeServices/src/Microsoft.AspNetCore.NodeServices.csproj
@@ -21,7 +21,7 @@
 
   <Target Name="YarnInstall">
     <Message Text="Running yarn install on $(MSBuildProjectFile)" Importance="High" />
-    <Yarn Command="install --mutex file" />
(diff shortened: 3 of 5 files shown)


eng/targets/Npm.Common.targets
```diff
@@ -18,7 +18,7 @@
 
   <Target Name="Restore">
     <Message Importance="High" Text="Running yarn install on $(MSBuildProjectFullPath)" />
-    <Yarn Command="install $(InstallArgs)" StandardOutputImportance="High" StandardErrorImportance="High" />
+    <Yarn Command="install --mutex network $(InstallArgs)" StandardOutputImportance="High" StandardErrorImportance="High" />
```
anurse: Is there a reason we're using `network` instead of `file`? It seems like `network` is more likely to get us in to port conflict issues...

natemcmaster: I tried file first, thinking the same thing, but that failed right away with
```
YARN : error : Unable to update lock within the stale threshold [F:\workspace\_work\1\s\src\SignalR\clients\ts\common\common.npmproj]
```

Google didn't turn up a useful answer for what causes this.

ryanbrandenburg: Yeah, I was thinking that using the file mutex with an absolute location (like `/tmp/.yarn-mutex`) would be the most reliable.

anurse: Ok. We can see how `network` goes :).

natemcmaster: Let's use the scientific method. Hypothesis - file mutex is more reliable.

Data so far:
* file mutex - 13 successful PR checks, 2 failures
* network mutex - 15 successful PR checks, 0 failures

I can trigger more runs before merging if we don't want to experiment with this on the master branch.

ryanbrandenburg: No need for more tests, I just didn't refresh the page before you said you'd tested it. If you've got data that trumps my hunch :)

src/Middleware/CORS/test/FunctionalTests/CORS.FunctionalTests.npmproj
```diff
@@ -12,10 +12,5 @@
     <IsTestProject>false</IsTestProject>
   </PropertyGroup>
 
-  <Target Name="InstallPuppeteer" BeforeTargets="Restore">
-    <!-- Explicitly install puppeteer. This will install the bundled Chromium as part of restore instead of as part of execution -->
-    <Yarn Command="add puppeteer" />
```
pranavkm: > The puppeteer package is already part of the package.json file for the CORS functional tests.

Here's the reason it was previously added: https://github.com/GoogleChrome/puppeteer/issues/1602#issuecomment-355545915. I'm happy to consider this if it works without it.

eng/targets/Npm.Common.targets
anurse: SCIENCE!

javiercn:
AFAIK this won't do anything will it? We don't run yarn in parallel no more as we do restore at the beginning of the project. Other runs should noop. If that's not the case, we could disable running yarn when there's already a node_modules present and be done with it.

src/Middleware/CORS/test/FunctionalTests/CORS.FunctionalTests.npmproj
natemcmaster: Tests pass on CI and I don't see the "Chromium revision is not downloaded" error. Is that sufficient validation?

By the way, we are already installing Chrome globally for some Kestrel tests. Could we switch the lighter version of this package, puppeteer-core, which doesn't bundle Chrome? This saves ~170MB Mac, ~282MB Linux, ~280MB Win.

natemcmaster:
> We don't run yarn in parallel

Actually, I suspect we _are_ running yarn in parallel and this is why we are hitting so many issues. For instance, see this build: https://dnceng.visualstudio.com/public/_build/results?buildId=175189&view=logs&jobId=a89a2a8c-213d-51e6-fe71-933481fa439c&taskId=701cb6c8-a755-541b-8b0e-d4afeec5a1e8&lineStart=2821&lineEnd=2836&colStart=1&colEnd=1. The console output appears to indicate parallel runs of yarn because the log messages appear in this order
```
Microsoft.AspNetCore.Components.E2ETests.csproj:    Running yarn install on Microsoft.AspNetCore.Components.E2ETests.csproj
ProjectTemplates.Tests.csproj:                      Running yarn install on ProjectTemplates.Tests.csproj
ProjectTemplates.Tests.csproj:                      YARN : error https: //registry.yarnpkg.com/selenium-standalone/-/selenium-standalone-6.15.6.tgz: Extracting tar content of undefined failed, the file appears to be corrupt: "ENOENT: no such file or directory, stat 'C:\\Users\\vsagent\\AppData\\Local\\Yarn\\Cache\\v4\\npm-selenium-standalone-6.15.6-55cc610e405dd74e312c77f3ac7336795ca2cd0c\\node_modules\\selenium-standalone\\test\\default-downloads-test.js'" [F:\workspace\_work\1\s\src\Components\test\E2ETest\Microsoft.AspNetCore.Components.E2ETests.csproj]
Microsoft.AspNetCore.Components.E2ETests.csproj:    YARN : error An unexpected error occurred: "https://registry.yarnpkg.com/selenium-standalone/-/selenium-standalone-6.15.6.tgz: ENOENT: no such file or directory, open 'C:\\Users\\vsagent\\AppData\\Local\\Yarn\\Cache\\v4\\npm-selenium-standalone-6.15.6-55cc610e405dd74e312c77f3ac7336795ca2cd0c\\node_modules\\selenium-standalone\\.yarn-tarball.tgz'". [F:\workspace\_work\1\s\src\ProjectTemplates\test\ProjectTemplates.Tests.csproj]
````

src/Middleware/CORS/test/FunctionalTests/CORS.FunctionalTests.npmproj
pranavkm: Their documentation says it's not guaranteed to work with anything outside the bundled version, but if it works, I'm happy with that.

dougbu (approved):

src/Middleware/CORS/test/FunctionalTests/CORS.FunctionalTests.npmproj
natemcmaster: Let's give it a try. If things start failing, I'll revert this piece.

PR #9935, opened 2019-05-02
Make some tests Flaky

ryanbrandenburg:


jkotalik (approved):

src/Identity/test/Identity.Test/IdentityUIScriptsTest.cs
```diff
@@ -79,6 +79,7 @@ public static IEnumerable<object[]> ScriptWithFallbackSrcData
 
         [Theory]
         [MemberData(nameof(ScriptWithFallbackSrcData))]
+        [Flaky("https://github.com/aspnet/AspNetCore-Internal/issues/2267", FlakyOn.All)]
```
pranavkm: The test's failed only on MacOS. Can we limit it to that platform?

pranavkm (changes requested):

pranavkm (approved):

PR #9957, opened 2019-05-03
Stop caching policies in AuthorizeFilter

### LINKED ISSUE #9779: Authorized call to action with explicit `AuthenticationSchemes` uses wrong authentication scheme, but only after a different action was called
### Describe the bug
In my original application, I'm configuring two authentication schemes like so:

```cs
services.AddAuthentication("Bearer")
		.AddBasic() // uses scheme "Basic"
		.AddIdentityServerAuthentication(); // uses scheme "Bearer"
```

My controller looks like this

```cs
[Authorize]
[Route("")]
public class ValuesController : ControllerBase
{
    [HttpGet("bar")]
    public void Bar()
    {
    }

    [Authorize(AuthenticationSchemes = "Basic")]
    [HttpGet("foo")]
    public void Foo()
    {
    }
}
```

At first, when I call `/foo` using `Basic` authentication, the request will succeed. Only after I've called `/bar` using `Bearer` authentication, calling `/foo` with `Basic` authentication will fail with a 401.

I've set breakpoints in the respective authentication handlers and I can confirm that the `Basic` handler is called on the first call but not on the last one.

Now, the error can be reproduced without using `AddIdentityServerAuthentication()` by adding `[AllowAnonymous]` to the `Bar()` method. In that case, the following exception is thrown on the third call

```
System.InvalidOperationException: No authenticationScheme was specified, and there was no DefaultChallengeScheme found.
   at Microsoft.AspNetCore.Authentication.AuthenticationService.ChallengeAsync(HttpContext context, String scheme, AuthenticationProperties properties)
   at Microsoft.AspNetCore.Mvc.ChallengeResult.ExecuteResultAsync(ActionContext context)
   at Microsoft.AspNetCore.Mvc.Internal.ResourceInvoker.InvokeResultAsync(IActionResult result)
   at Microsoft.AspNetCore.Mvc.Internal.ResourceInvoker.InvokeAlwaysRunResultFilters()
   at Microsoft.AspNetCore.Mvc.Internal.ResourceInvoker.InvokeFilterPipelineAsync()
   at Microsoft.AspNetCore.Mvc.Internal.ResourceInvoker.InvokeAsync()
   at Microsoft.AspNetCore.Routing.EndpointMiddleware.Invoke(HttpContext httpContext)
   at Microsoft.AspNetCore.Routing.EndpointRoutingMiddleware.Invoke(HttpContext httpContext)
   at Microsoft.AspNetCore.Authentication.AuthenticationMiddleware.Invoke(HttpContext context)
   at Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddleware.Invoke(HttpContext context)

```

### To Reproduce

- Check out https://github.com/cypressious/AspNetcoreAuthIssue and run it
- Call
```
localhost:5010/foo
Authorization: Basic YWRtaW46YWRtaW4=
```
- Call `localhost:5010/bar` 
- Call again
```
localhost:5010/foo
Authorization: Basic YWRtaW46YWRtaW4=
```

### Expected behavior
The last call to `localhost:5010/foo` succeeds

### global.json

```json
{
  "sdk": {
    "version": "2.2.104"
  }
}
```

HaoK:
Fixes https://github.com/aspnet/AspNetCore/issues/9779

### DIFF
--- a/src/Mvc/Mvc.Core/src/Authorization/AuthorizeFilter.cs
+++ b/src/Mvc/Mvc.Core/src/Authorization/AuthorizeFilter.cs
@@ -126,10 +126,8 @@ namespace Microsoft.AspNetCore.Mvc.Authorization
 
         internal async Task<AuthorizationPolicy> GetEffectivePolicyAsync(AuthorizationFilterContext context)
         {
-            var effectivePolicy = await ComputePolicyAsync();
-
             // Combine all authorize filters into single effective policy that's only run on the closest filter
-            var builder = new AuthorizationPolicyBuilder(effectivePolicy);
+            var builder = new AuthorizationPolicyBuilder(await ComputePolicyAsync());
             for (var i = 0; i < context.Filters.Count; i++)
             {
                 if (ReferenceEquals(this, context.Filters[i]))
@@ -162,7 +160,7 @@ namespace Microsoft.AspNetCore.Mvc.Authorization
                 }
             }
 
-            return builder?.Build() ?? effectivePolicy;
+            return builder.Build();
         }
 
         /// <inheritdoc />
--- a/src/Mvc/Mvc.Core/src/Authorization/AuthorizeFilter.cs
+++ b/src/Mvc/Mvc.Core/src/Authorization/AuthorizeFilter.cs
@@ -23,8 +23,6 @@ namespace Microsoft.AspNetCore.Mvc.Authorization
     /// </summary>
     public class AuthorizeFilter : IAsyncAuthorizationFilter, IFilterFactory
     {
-        private AuthorizationPolicy _effectivePolicy;
-
         /// <summary>
         /// Initializes a new <see cref="AuthorizeFilter"/> instance.
         /// </summary>
@@ -128,13 +126,7 @@ namespace Microsoft.AspNetCore.Mvc.Authorization
 
         internal async Task<AuthorizationPolicy> GetEffectivePolicyAsync(AuthorizationFilterContext context)
         {
-            if (_effectivePolicy != null)
-            {
-                return _effectivePolicy;
-            }
-
             var effectivePolicy = await ComputePolicyAsync();
-            var canCache = PolicyProvider == null;
 
             // Combine all authorize filters into single effective policy that's only run on the closest filter
             var builder = new AuthorizationPolicyBuilder(effectivePolicy);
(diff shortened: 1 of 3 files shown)


src/Mvc/Mvc.Core/src/Authorization/AuthorizeFilter.cs
```diff
@@ -169,20 +160,9 @@ internal async Task<AuthorizationPolicy> GetEffectivePolicyAsync(AuthorizationFi
...
-            effectivePolicy = builder?.Build() ?? effectivePolicy;
-
-            // We can cache the effective policy when there is no custom policy provider
-            if (canCache)
-            {
-                _effectivePolicy = effectivePolicy;
             }
 
-            return effectivePolicy;
+            return builder?.Build() ?? effectivePolicy;
```
rynowak: Am I crazy? It doesn't look like it's actually possible for `builder` to be null here.

src/Mvc/Mvc.Core/test/Authorization/AuthorizeFilterTest.cs
```diff
@@ -513,26 +513,6 @@ public void CreateInstance_ReturnsNewFilterIfPolicyAndPolicyProviderAreNotSet(Au
             Assert.Same(policyProvider, actual.PolicyProvider);
         }
 
-        [Fact]
-        public async Task GetEffectivePolicyAsync_ReturnsCurrentPolicy_WhenNoEndpointMetadataIsAvailable()
```
rynowak: Is it worth a test that we don't cache? IE, we see the policy provider called twice. I'm thinking less about correctness and more about preventing an *optimization* in the future. Another idea would be a big scary comment.

src/Mvc/Mvc.Core/src/Authorization/AuthorizeFilter.cs
HaoK: Yeah looks like I can simplify this method a bit, I'll clean it up

src/Mvc/Mvc.Core/test/Authorization/AuthorizeFilterTest.cs
HaoK: Yeah I saw a test that was verifying this already in the case of a policy provider in the past , I'll find it and copy and verify it for the normal code path as well

rynowak (approved):

HaoK:
Will open separate PRs for 2.1/2.2

PR #9958, opened 2019-05-03
Use NuGet.Config for SignalR benchmarks

BrennanConroy:


src/SignalR/perf/benchmarkapps/Directory.Build.props
```diff
@@ -1,5 +1,2 @@
 <Project>
-
-  <Import Project="$(MSBuildThisFileDirectory)..\..\..\..\build\sources.props" />
-
-</Project>
\ No newline at end of file
+</Project>
```
sebastienros: You are correct here to keep the file, so we don't inherit from the parent one.

sebastienros (approved):

natemcmaster (approved):

PR #9962, opened 2019-05-04
Read interface IList.Count once rather than per iteration

benaadams:
Cache `IList.Count` when used as an iteration condition as it is an uninlinable interface call so needs to be evaluated each time.

30M calls for 44k requests to `dotnet new mvc` homepage; or ~680 calls per request.

image

/cc @rynowak

### DIFF
--- a/src/Razor/Razor.Runtime/src/Runtime/TagHelpers/TagHelperRunner.cs
+++ b/src/Razor/Razor.Runtime/src/Runtime/TagHelpers/TagHelperRunner.cs
@@ -45,7 +45,7 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
                 var task = tagHelpers[i].ProcessAsync(tagHelperContext, tagHelperOutput);
                 if (!task.IsCompletedSuccessfully)
                 {
-                    return Awaited(task, executionContext, i, count);
+                    return Awaited(task, executionContext, i + 1, count);
                 }
             }
 
--- a/src/Razor/Razor.Runtime/src/Runtime/TagHelpers/TagHelperRunner.cs
+++ b/src/Razor/Razor.Runtime/src/Runtime/TagHelpers/TagHelperRunner.cs
@@ -20,7 +20,7 @@ namespace Microsoft.AspNetCore.Razor.Runtime.TagHelpers
         /// </param>
         /// <returns>Resulting <see cref="TagHelperOutput"/> from processing all of the
         /// <paramref name="executionContext"/>'s <see cref="ITagHelper"/>s.</returns>
-        public async Task RunAsync(TagHelperExecutionContext executionContext)
+        public Task RunAsync(TagHelperExecutionContext executionContext)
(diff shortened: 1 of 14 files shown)


rynowak:
😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻 😻

src/Razor/Razor.Runtime/src/Runtime/TagHelpers/TagHelperRunner.cs
```diff
@@ -20,42 +20,65 @@ public class TagHelperRunner
...
             var tagHelperOutput = executionContext.Output;
 
-            for (var i = 0; i < executionContext.TagHelpers.Count; i++)
+            for (var i = 0; i < count; i++)
             {
-                await executionContext.TagHelpers[i].ProcessAsync(tagHelperContext, tagHelperOutput);
+                var task = tagHelpers[i].ProcessAsync(tagHelperContext, tagHelperOutput);
+                if (!task.IsCompletedSuccessfully)
+                {
+                    return Awaited(task, executionContext, i, count);
```
rynowak: Is this right? don't we need to increment `i`. it looks like this will initialize the `i`th entry twice if there are > `i` entries.

rynowak:
What's the number of calls after the changes?

src/Razor/Razor.Runtime/src/Runtime/TagHelpers/TagHelperRunner.cs
benaadams: Good catch

rynowak (approved):

benaadams:
14M for 53.5k requests so ~680 => ~267 per request

image

rynowak:
awesome!

rynowak:
It's a shame that the simplest way to implement a collection type leads you down this path - thinking of https://github.com/aspnet/AspNetCore/blob/master/src/Razor/Razor/src/TagHelpers/ReadOnlyTagHelperAttributeList.cs and other types we have that subclass `Collection<>` or `ReadOnlyCollection<>`.

I've been thinking about how I want to implement https://github.com/aspnet/AspNetCore/blob/master/src/Http/Http.Abstractions/src/Routing/EndpointMetadataCollection.cs#L100 - because we need to change this in 3.0 to return something better than `IEnumerabl<>`.

rynowak:
Thanks @benaadams

Issue #3667, opened 2018-10-18
ContentDispositionHeaderValue does not sanitize newlines


CornedBee:
### Describe the bug

On some systems (not Windows, but Linux), filenames may contain newline characters. Thus, it should be theoretically valid to have a Content-disposition header filename contain such a newline character, or else at least clear them out when sanitizing.

However, [ContentDispositionHeaderValue.Sanitize](https://github.com/aspnet/HttpAbstractions/blob/8e4f7365adf304e9704b1db43e3095a267583513/src/Microsoft.Net.Http.Headers/ContentDispositionHeaderValue.cs#L437) only replaces characters outside the ASCII range, not characters that are otherwise invalid in headers.

### To Reproduce
1. Using this version of ASP.NET Core: tested on 2.0, though I expect 2.1 still exhibits the problem.
2. Run this code '....'

```
public class MyController : Controller
{
  [HttpGet("/test")]
  public IActionResult Test()
  {
    var content = Encoding.UTF8.GetBytes("Hello");
    return File(content, "text/plain", "File\nName.txt");
  }
}
```

3. With these arguments '....'

This eventually throws an exception, complaining that "\n" is not valid in HTTP headers.

### Expected behavior

I expected the newline in the filename to either be correctly encoded so that it is preserved (and possibly let the browser deal with it in some system-specific way), or else be removed or substituted in the header.

PR #9971, opened 2019-05-04
ContentDispositionHeaderValue must encode/sanitize new-line char

huysentruitw:
ContentDispositionHeaderValue must encode/sanitize new-line char
in filename.

Addresses #3667

### DIFF
--- a/src/Http/Headers/src/ContentDispositionHeaderValue.cs
+++ b/src/Http/Headers/src/ContentDispositionHeaderValue.cs
@@ -444,7 +444,7 @@ namespace Microsoft.Net.Http.Headers
                 for (int i = 0; i < result.Length; i++)
                 {
                     var c = result[i];
-                    if ((int)c > 0x7f || c == '\n')
+                    if ((int)c > 0x7f || (int)c < 0x20)
                     {
                         c = '_'; // Replace out-of-range characters
                     }
@@ -465,14 +465,14 @@ namespace Microsoft.Net.Http.Headers
                 && value.EndsWith("\"", StringComparison.Ordinal);
         }
 
-        // tspecials are required to be in a quoted string. Only non-ascii and new-line needs to be encoded.
+        // tspecials are required to be in a quoted string. Only non-ascii needs to be encoded.
         private bool RequiresEncoding(StringSegment input)
         {
             Contract.Assert(input != null);
 
             for (int i = 0; i < input.Length; i++)
             {
-                if ((int)input[i] > 0x7f || input[i] == '\n')
+                if ((int)input[i] > 0x7f || input[i] < 0x20)
                 {
                     return true;
                 }
--- a/src/Http/Headers/src/ContentDispositionHeaderValue.cs
+++ b/src/Http/Headers/src/ContentDispositionHeaderValue.cs
@@ -444,7 +444,7 @@ namespace Microsoft.Net.Http.Headers
                 for (int i = 0; i < result.Length; i++)
                 {
                     var c = result[i];
-                    if ((int)c > 0x7f)
+                    if ((int)c > 0x7f || c == '\n')
                     {
                         c = '_'; // Replace out-of-range characters
                     }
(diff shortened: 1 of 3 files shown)


src/Http/Headers/src/ContentDispositionHeaderValue.cs
```diff
@@ -444,7 +444,7 @@ private StringSegment Sanitize(StringSegment input)
                 for (int i = 0; i < result.Length; i++)
                 {
                     var c = result[i];
-                    if ((int)c > 0x7f)
+                    if ((int)c > 0x7f || c == '\n')
```
samsosa: Maybe
```suggestion
                    if ((int)c > 0x7f || (int)c < 0x20)
```
?

src/Http/Headers/src/ContentDispositionHeaderValue.cs
```diff
@@ -465,14 +465,14 @@ private bool IsQuoted(StringSegment value)
...
-        // tspecials are required to be in a quoted string.  Only non-ascii needs to be encoded.
+        // tspecials are required to be in a quoted string. Only non-ascii and new-line needs to be encoded.
         private bool RequiresEncoding(StringSegment input)
         {
             Contract.Assert(input != null);
 
             for (int i = 0; i < input.Length; i++)
             {
-                if ((int)input[i] > 0x7f)
+                if ((int)input[i] > 0x7f || input[i] == '\n')
```
samsosa: Maybe
```suggestion
                if ((int)input[i] > 0x7f || (int)input[i] < 0x20)
```
?

src/Http/Headers/src/ContentDispositionHeaderValue.cs
huysentruitw: I like the idea, couldn't find this in the header spec though.

Tratcher:
These look like real test failures in MVC caused by your change. At first glance it looks like the expected results need to be updated.
```
  Failed   Microsoft.AspNetCore.Mvc.FileResultTest.GetHeaderValue_Produces_Correct_ContentDisposition(input: "\x15", expectedOutput: "attachment; filename=\"\x15\"; filename*=UTF-8''%1"...)
  Error Message:
   Assert.Equal() Failure
                                   ↓ (pos 21)
  Expected: ···ttachment; filename=""; filename*=UTF-8''%15
  Actual:   ···ttachment; filename=_; filename*=UTF-8''%15
                                   ↑ (pos 21)
```

huysentruitw:
Oops, forgot to run tests in the other solutions. There are 34 tests failing, all related to non ASCII characters below 0x20. According to the test data [here](https://github.com/aspnet/AspNetCore/blob/c565386a3ed135560bc2e9017aa54a950b4e35dd/src/Mvc/Mvc.Core/test/FileResultTest.cs#L196) I think we should only handle the `\n` case. 

@Tratcher what do you think?

anurse:
Test failure looks unrelated. I'll have our bot file an issue for this and then retrigger the Helix run.

anurse:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=180492

aspnet-hello:

huysentruitw:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=180492

samsosa:
@huysentruitw 
The Linux filesystem ext2, ext3, ext4 allows all characters except NUL or / within filenames.

The [RFC7230](https://tools.ietf.org/html/rfc7230#section-3.2.6) is not very clear overall but specifies that (only) any visible character can be used. Every octet in the upper half (0x80-0xFF) should be opaque, but some of the lower ASCII symbols are indeed not visible but unfortunately not further mentioned.

Ultimately, the problem is almost a matter of taste because only a few people stick to the RFCs anyway.

huysentruitw:
@samsosa thanks for this clarification!

src/Http/Headers/src/ContentDispositionHeaderValue.cs
Tratcher: Filtering control characters makes sense to me. We have filename* for preserving the full name.

Tratcher:
Filtering control characters is a reasonable behavior here. filename* can be used for higher fidelity.

huysentruitw:
@Tratcher does this mean we go with the `< 0x20` instead of just `\n` and alter the MVC tests?

Tratcher:
Yes please.

src/Mvc/Mvc.Core/test/FileResultTest.cs
```diff
@@ -155,6 +155,7 @@ public async Task ExecuteResultAsync_ThrowsException_IfCannotResolveLoggerFactor
                     { "^", "attachment; filename=^; filename*=UTF-8''^" },
                     { "`", "attachment; filename=`; filename*=UTF-8''`" },
                     { "|", "attachment; filename=|; filename*=UTF-8''|" },
+                    { "a\tb", "attachment; filename=a_b; filename*=UTF-8''a%09b" },
```
huysentruitw: Needs to be moved to the 'values need to escape' block. Will fix tomorrow.

Tratcher (approved):

Tratcher:
Thanks

PR #9972, opened 2019-05-04
Add some caching to CORS

pranavkm:


src/Middleware/CORS/src/Infrastructure/CorsMiddleware.cs
```diff
@@ -129,7 +129,7 @@ public Task Invoke(HttpContext context, ICorsPolicyProvider corsPolicyProvider)
             return InvokeCore(context, corsPolicyProvider);
         }
 
-        private async Task InvokeCore(HttpContext context, ICorsPolicyProvider corsPolicyProvider)
```
rynowak: This can be merged into the `Invoke` method. They were previously separated to avoid the state machine, but now there's no real need to.

src/Middleware/CORS/src/Infrastructure/DefaultCorsPolicyProvider.cs
```diff
@@ -30,7 +30,13 @@ public Task<CorsPolicy> GetPolicyAsync(HttpContext context, string policyName)
...
             }
 
-            return Task.FromResult(_options.GetPolicy(policyName ?? _options.DefaultPolicyName));
+            policyName ??= _options.DefaultPolicyName;
+            if (_options.PolicyMap.TryGetValue(policyName, out var result))
+            {
+                return result.policyTask;
+            }
+
+            return Task.FromResult<CorsPolicy>(null);
```
rynowak: Does this benefit from caching? Or does .NET do this for you?

rynowak (approved):

src/Middleware/CORS/src/Infrastructure/DefaultCorsPolicyProvider.cs
JamesNK: As far as I know there is no caching.

You should return a static readonly instance.

src/Middleware/CORS/src/Infrastructure/CorsOptions.cs
```diff
@@ -12,22 +13,16 @@ namespace Microsoft.AspNetCore.Cors.Infrastructure
     public class CorsOptions
     {
         private string _defaultPolicyName = "__DefaultCorsPolicy";
-        private IDictionary<string, CorsPolicy> PolicyMap { get; } = new Dictionary<string, CorsPolicy>();
+
+        internal IDictionary<string, (CorsPolicy policy, Task<CorsPolicy> policyTask)> PolicyMap { get; }
```
JamesNK: Is the tuple to avoid accessing Task.Result? The task will always be complete so I don't think it is a big deal.

A comment explaining why a task is cached would be useful

src/Middleware/CORS/src/Infrastructure/CorsOptions.cs
```diff
@@ -12,22 +13,16 @@ namespace Microsoft.AspNetCore.Cors.Infrastructure
...
+            = new Dictionary<string, (CorsPolicy, Task<CorsPolicy>)>();
```
JamesNK: Should this dictionary have a string comparer for the key?

src/Middleware/CORS/src/Infrastructure/CorsMiddleware.cs
```diff
@@ -183,14 +182,30 @@ private async Task InvokeCore(HttpContext context, ICorsPolicyProvider corsPolic
...
+                    return InvokeCoreAwaited(context, policyTask);
+                }
+
+                corsPolicy = policyTask.GetAwaiter().GetResult();
             }
 
+            return EvaluateAndApplyPolicy(context, corsPolicy);
+        }
+
+        private async Task InvokeCoreAwaited(HttpContext context, Task<CorsPolicy> policyTask)
```
JamesNK: nit: static local method?

pranavkm: Can't be static, because it references a non-static method, but we can make it a local function

pranavkm:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=186758&view=ms.vss-test-web.build-test-results-tab&runId=4500858&paneView=debug

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/2465

Issue #5942, opened 2018-10-17
Request.Host not set by Microsoft.AspNetCore.TestHost.WebSocketClient


kenlyon:
I am using a `Microsoft.AspNetCore.TestHost.TestServer` to test my application.

Whenever I use `TestServer.CreateClient()` then `HttpClient.PostAsync()`, I can see that my controller correctly has the `Request.Host` property matching `TestServer.BaseAddress`.

However, when I use `TestServer.CreateWebSocketClient()` then `WebSocketClient.ConnectAsync()`, I find that the controller's `Request.Host` is always `null`. 

This is a problem for me as I try to generate a new url relative to the requested url in the controller using `new Uri(Request.GetEncodedUrl())`. The call to `Request.GetEncodedUrl()` returns a url with no host (e,g, "http:///path/here") and `new Uri()` throws a `UriFormatException` because there's no _hostname_.

KristianJakubik:
Hi, I would like to work on this issue. Is it still relevant?

kenlyon:
I think it is.

anurse:
https://github.com/aspnet/AspNetCore/commit/f2f1d7751710342d22c55edf83a01615409f2a49

PR #9973, opened 2019-05-04
TestHost.WebSocketClient should set Request.Host

huysentruitw:
TestHost.WebSocketClient should set Request.Host

Addresses #5942

### DIFF
--- a/src/Hosting/TestHost/src/WebSocketClient.cs
+++ b/src/Hosting/TestHost/src/WebSocketClient.cs
@@ -61,11 +61,9 @@ namespace Microsoft.AspNetCore.TestHost
                 request.Scheme = scheme;
                 if (!request.Host.HasValue)
                 {
-                    request.Host = HostString.FromUriComponent(uri);
-                    if (uri.IsDefaultPort)
-                    {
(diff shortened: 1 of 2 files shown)


src/Hosting/TestHost/src/WebSocketClient.cs
```diff
@@ -59,6 +59,14 @@ public async Task<WebSocket> ConnectAsync(Uri uri, CancellationToken cancellatio
                 scheme = (scheme == "ws") ? "http" : scheme;
                 scheme = (scheme == "wss") ? "https" : scheme;
                 request.Scheme = scheme;
+                if (!request.Host.HasValue)
+                {
+                    request.Host = HostString.FromUriComponent(uri);
```
davidfowl: Check if the uri.IsDefaultPort right here so you can avoid the double assignment?

huysentruitw: I've got my inspiration from [here](https://github.com/aspnet/AspNetCore/blob/master/src/Hosting/TestHost/src/ClientHandler.cs#L83)

analogrelay (approved):
LGTM. Thanks!

Issue #7449, opened 2019-02-11
TestServer HttpContext.Abort does nothing


JamesNK:
Hosting a website with TestHost and calling `HttpContext.Abort()` inside a request does not set `HttpContext.RequestAborted`.

huysentruitw:
Does this ever work (even outside `TestHost`) ? I'm trying to see what needs to be done here, but all I find is that [`HttpContext.Abort()`](https://github.com/aspnet/AspNetCore/blob/4646ed56189a8ee36ced6397dd4f86fce9e6a8e8/src/Http/Http/src/DefaultHttpContext.cs#L163) calls `LifetimeFeature.Abort();` which is an [empty method](https://github.com/aspnet/AspNetCore/blob/425c196cba530b161b120a57af8f1dd513b96f67/src/Http/Http/src/Features/HttpRequestLifetimeFeature.cs#L12).

huysentruitw:
Never mind, it's supposed to be empty as the purpose is to let users implement their own feature implementing `IHttpRequestLifetimeFeature`. 😎

davidfowl:
Yes it works outside of test host. Nothing uses that default feature implementation, servers implement their own.

Tratcher:
Accepted via tests, no out of band SDK testing required.

PR #9975, opened 2019-05-04
Calling abort inside TestServer handler should set RequestAborted

huysentruitw:
Calling abort inside TestServer handler should set RequestAborted

Addresses #7449

### DIFF
--- a/src/Hosting/TestHost/src/RequestLifetimeFeature.cs
+++ b/src/Hosting/TestHost/src/RequestLifetimeFeature.cs
@@ -1,24 +1,22 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
 using System.Threading;
+using Microsoft.AspNetCore.Http.Features;
 
-namespace Microsoft.AspNetCore.Http.Features
+namespace Microsoft.AspNetCore.TestHost
 {
     internal class RequestLifetimeFeature : IHttpRequestLifetimeFeature
     {
-        private readonly CancellationTokenSource _abortableCancellationTokenSource = new CancellationTokenSource();
-        private CancellationTokenSource _linkedCancellationTokenSource;
+        private readonly CancellationTokenSource _cancellationTokenSource = new CancellationTokenSource();
 
-        public CancellationToken RequestAborted
+        public RequestLifetimeFeature()
         {
-            get => _linkedCancellationTokenSource?.Token ?? _abortableCancellationTokenSource.Token;
-            set => _linkedCancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(
-                    token1: _abortableCancellationTokenSource.Token,
-                    token2: value);
+            RequestAborted = _cancellationTokenSource.Token;
         }
 
-        public void Abort() => _abortableCancellationTokenSource.Cancel();
+        public CancellationToken RequestAborted { get; set; }
+
+        public void Abort() => _cancellationTokenSource.Cancel();
     }
 }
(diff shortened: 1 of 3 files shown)


src/Hosting/TestHost/src/RequestLifetimeFeature.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace Microsoft.AspNetCore.Http.Features
+{
+    internal class RequestLifetimeFeature : IHttpRequestLifetimeFeature
+    {
+        private readonly CancellationTokenSource _cancellationTokenSource = new CancellationTokenSource();
+
+        public CancellationToken RequestAborted
+        {
+            get => _cancellationTokenSource.Token;
+            set => throw new NotSupportedException();
```
JamesNK: You can support with set like this: https://github.com/grpc/grpc-dotnet/blob/0be935fca79674be56c9fe7e69f6708306889ae2/testassets/FunctionalTestsWebsite/Infrastructure/TestHttpRequestLifetimeFeature.cs

JamesNK (commented):
Looks ok to me. Should be reviewed by a TestHost Expert(tm)

src/Hosting/TestHost/src/RequestLifetimeFeature.cs
huysentruitw: That's lying to the API user though. That setter should not exist IMO as setting it breaks the Abort functionality.

src/Hosting/TestHost/src/RequestLifetimeFeature.cs
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading;
+
+namespace Microsoft.AspNetCore.Http.Features
```
Tratcher:
```suggestion
namespace Microsoft.AspNetCore.TestHost
```

src/Hosting/TestHost/src/RequestLifetimeFeature.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    internal class RequestLifetimeFeature : IHttpRequestLifetimeFeature
+    {
+        private readonly CancellationTokenSource _abortableCancellationTokenSource = new CancellationTokenSource();
+        private CancellationTokenSource _linkedCancellationTokenSource;
+
+        public CancellationToken RequestAborted
```
Tratcher: It's the caller's responsibility to link tokens, and they may choose to link them in other ways. You can replace _linkedCancellationTokenSource with a field for the current CancellationToken initialized to _abortableCancellationTokenSource.Token.

```suggestion
        public CancellationToken RequestAborted { get; set; } = _abortableCancellationTokenSource.Token;
```

Tratcher (changes requested):

src/Hosting/TestHost/src/RequestLifetimeFeature.cs
huysentruitw: So, when the caller passes a `CancellationToken`, the `Abort()` method stops working. That's very weird to me 🤔

Tratcher: Yes, that's the behavior of all of our servers (though Abort also does other things in most servers like close the socket). The caller setting CancellationToken is intentionally customizing the behavior, they can choose how the linking happens.

JamesNK (approved):

JamesNK:
Waiting on @Tratcher approval before merging.

Tratcher:
Thanks

PR #8492, opened 2019-03-13
Fix SignalR build from VS

mikaelm12:
Fixing build from VS for SignalR.

PR #9976, opened 2019-05-05
SignalR: Fix Crankier build

staff0rd:
Fixes Crankier & BenchMarkServer builds.

A couple of questions;

1. Why were the projects dropped from SignalR.sln?  (#8492)
2. In #9958 I see a note from @sebastienros that says we don't want to inherit Directory.Build.props, but both .csproj are using `Reference`.

### DIFF
--- a/src/SignalR/perf/benchmarkapps/Directory.Build.props
+++ b/src/SignalR/perf/benchmarkapps/Directory.Build.props
@@ -1,2 +1,0 @@
-<Project>
-</Project>
--- a/src/SignalR/perf/benchmarkapps/BenchmarkServer/Directory.Build.props
+++ b/src/SignalR/perf/benchmarkapps/BenchmarkServer/Directory.Build.props
@@ -1,0 +1,2 @@
+<Project>
+</Project>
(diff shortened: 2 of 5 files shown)


src/SignalR/perf/benchmarkapps/Directory.Build.props
```diff
@@ -1,2 +0,0 @@
-<Project>
```
BrennanConroy: Move these to the `BenchmarkServer` folder. We want to keep them for that project.

staff0rd: That project does not build without them, as it uses `Reference`...

BrennanConroy: That project isn't really meant to be used by itself. It's part of our benchmarks https://github.com/aspnet/Benchmarks

staff0rd: Odd - I'm using it to point Crankier at.  Also see #9264.

BrennanConroy: Oh... don't touch that project please

staff0rd: done

staff0rd:
@BrennanConroy can you please review this again?  I've closed #9264 and will instead move that functionality to Crankier as suggested.  This PR now represents moving the Directory* stuff so it only affects the benchmark server and the `IHostingEnvironment` deprecation update.

BrennanConroy:
Thanks for the PR!

Issue #9194, opened 2019-04-09
Investigate C#8/Nullable with MVC and Components


rynowak:

rynowak:
Closing this since I did investigation. Will follow up with more issues later

PR #9978, opened 2019-05-05
Adds inferred [Required] for non-null ref types

rynowak:
Follow up from #9194

This change adds the automatic inference of [Required] for non-nullable
properties and parameters. This means that if you opt into nullable
context in C#8, we'll start treating those types as-if you put
[Required] on them.

This provides a nice invariant to rely on, namely that MVC will honor
your declared nullability contract OR report a validation error. This
reinforces the guidance already published by the C# team for using
POCOs/DTOs with nullability. See
https://github.com/aspnet/specs/blob/master/notes/3_0/nullable.md for my
analysis on the topic.

### DIFF
--- a/src/Mvc/Mvc.Core/src/ModelBinding/Metadata/ValidationMetadata.cs
+++ b/src/Mvc/Mvc.Core/src/ModelBinding/Metadata/ValidationMetadata.cs
@@ -1,4 +1,3 @@
-
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
--- a/src/Mvc/Mvc.Core/src/ModelBinding/Metadata/ValidationMetadata.cs
+++ b/src/Mvc/Mvc.Core/src/ModelBinding/Metadata/ValidationMetadata.cs
@@ -1,3 +1,4 @@
+
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
--- a/src/Mvc/Mvc.Core/src/MvcOptions.cs
+++ b/src/Mvc/Mvc.Core/src/MvcOptions.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
 using System.Text.Json;
 using System.Text.Json.Serialization;
 using Microsoft.AspNetCore.Mvc.ApplicationModels;
@@ -101,6 +102,25 @@ namespace Microsoft.AspNetCore.Mvc
         /// Gets a list of <see cref="IInputFormatter"/>s that are used by this application.
         /// </summary>
         public FormatterCollection<IInputFormatter> InputFormatters { get; }
+
+        /// <summary>
+        /// Gets or sets a value that detemines if the inference of <see cref="RequiredAttribute"/> for
+        /// for properties and parameters of non-nullable reference types is suppressed. If <c>false</c> 
+        /// (the default), then all non-nullable reference types will behave as-if <c>[Required]</c> has 
+        /// been applied. If <c>true</c>, this behavior will be suppressed; nullable reference types and 
+        /// non-nullable reference types will behave the same for the purposes of validation.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// This option controls whether MVC model binding and validation treats nullable and non-nullable 
+        /// reference types differently. 
+        /// </para>
+        /// <para>
+        /// By default, MVC will treat a non-nullable reference type parameters and properties as-if 
+        /// <c>[Required]</c> has been applied, resulting in validation errors when no value was bound.
+        /// </para>
+        /// </remarks>
+        public bool SuppressImplicitRequiredAttributeForNonNullableReferenceTypes { get; set; }
 
         /// <summary>
         /// Gets or sets a value that determines if buffering is disabled for input formatters that
--- a/src/Mvc/Mvc.Core/src/MvcOptions.cs
+++ b/src/Mvc/Mvc.Core/src/MvcOptions.cs
@@ -119,6 +119,11 @@ namespace Microsoft.AspNetCore.Mvc
         /// By default, MVC will treat a non-nullable reference type parameters and properties as-if 
         /// <c>[Required]</c> has been applied, resulting in validation errors when no value was bound.
         /// </para>
+        /// <para>
+        /// MVC does not support non-nullable reference type annotations on type arguments and type parameter
+        /// contraints. The framework will not infer any validation attributes for generic-typed properties
+        /// or collection elements.
+        /// </para>
         /// </remarks>
         public bool SuppressImplicitRequiredAttributeForNonNullableReferenceTypes { get; set; }
 
--- a/src/Mvc/Mvc.DataAnnotations/src/DataAnnotationsMetadataProvider.cs
+++ b/src/Mvc/Mvc.DataAnnotations/src/DataAnnotationsMetadataProvider.cs
@@ -410,16 +410,25 @@ namespace Microsoft.AspNetCore.Mvc.DataAnnotations
         internal static bool IsNonNullable(IEnumerable<object> attributes)
         {
             // [Nullable] is compiler synthesized, comparing by name. 
-            var nullableAtttribute = attributes
-                .Where(a => string.Equals(a.GetType().FullName, NullableAttributeFullTypeName))
+            var nullableAttribute = attributes
+                .Where(a => string.Equals(a.GetType().FullName, NullableAttributeFullTypeName, StringComparison.Ordinal))
                 .FirstOrDefault();
+            if (nullableAttribute == null)
+            {
+                return false;
+            }
 
+            // We don't handle cases where generics and NNRT are used. This runs into a
+            // fundamental limitation of ModelMetadata - we a single the Type and Property/Parameter
+            // to look up the metadata. However when generics are involved and NNRT is in use
+            // the distance between the [Nullable] and member we're looking at is potentially
+            // unbounded. 
+            //
             // See: https://github.com/dotnet/roslyn/blob/master/docs/features/nullable-reference-types.md#annotations
-            if (nullableAtttribute != null &&
-                nullableAtttribute.GetType().GetField(NullableFlagsFieldName) is FieldInfo field &&
-                field.GetValue(nullableAtttribute) is byte[] flags
-                && flags.Length >= 0
-                && flags[0] == 1)
+            if (nullableAttribute.GetType().GetField(NullableFlagsFieldName) is FieldInfo field &&
+                field.GetValue(nullableAttribute) is byte[] flags && 
+                flags.Length >= 0 && 
+                flags[0] == 1) // First element is the property/parameter type.
             {
                 return true;
             }
--- a/src/Mvc/Mvc.DataAnnotations/src/DataAnnotationsMetadataProvider.cs
+++ b/src/Mvc/Mvc.DataAnnotations/src/DataAnnotationsMetadataProvider.cs
@@ -345,7 +345,7 @@ namespace Microsoft.AspNetCore.Mvc.DataAnnotations
             // This allows the developer to specify [Required] to customize the error message, so
             // if they already have [Required] then there's no need for us to do this check.
             if (!_options.SuppressImplicitRequiredAttributeForNonNullableReferenceTypes &&
-                requiredAttribute == null && 
+                requiredAttribute == null &&
                 !context.Key.ModelType.IsValueType &&
 
                 // Look specifically at attributes on the property/parameter. [Nullable] on
@@ -419,7 +419,7 @@ namespace Microsoft.AspNetCore.Mvc.DataAnnotations
         // Internal for testing
         internal static bool IsNonNullable(IEnumerable<object> attributes)
         {
-            // [Nullable] is compiler synthesized, comparing by name. 
+            // [Nullable] is compiler synthesized, comparing by name.
             var nullableAttribute = attributes
                 .Where(a => string.Equals(a.GetType().FullName, NullableAttributeFullTypeName, StringComparison.Ordinal))
                 .FirstOrDefault();
@@ -429,15 +429,15 @@ namespace Microsoft.AspNetCore.Mvc.DataAnnotations
             }
 
             // We don't handle cases where generics and NNRT are used. This runs into a
-            // fundamental limitation of ModelMetadata - we a single the Type and Property/Parameter
+            // fundamental limitation of ModelMetadata - we use a single Type and Property/Parameter
             // to look up the metadata. However when generics are involved and NNRT is in use
             // the distance between the [Nullable] and member we're looking at is potentially
-            // unbounded. 
+            // unbounded.
             //
             // See: https://github.com/dotnet/roslyn/blob/master/docs/features/nullable-reference-types.md#annotations
             if (nullableAttribute.GetType().GetField(NullableFlagsFieldName) is FieldInfo field &&
-                field.GetValue(nullableAttribute) is byte[] flags && 
-                flags.Length >= 0 && 
+                field.GetValue(nullableAttribute) is byte[] flags &&
+                flags.Length >= 0 &&
                 flags[0] == 1) // First element is the property/parameter type.
             {
                 return true;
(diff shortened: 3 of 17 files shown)


rynowak:
One other idea (that I didn't do, and think we shouldn't do) would be to use the nullability information to create related objects. So for the following, we could auto-create `Address` even if the data was missing:

```C#
public class Person
{
    public string FirstName { get; set; } = default!;
    public Address Address { get; set; } = default!;
}

public class Address { }
```

I didn't do this, because IMO you have a better option if you want this behaviour.

```C#
public class Person
{
    public string FirstName { get; set; } = default!;
    public Address Address { get; } = new Address();
}

public class Address { }
```

U

rynowak:
@roji FYI

src/Mvc/Mvc.Core/src/ModelBinding/Metadata/ValidationMetadata.cs
```diff
@@ -1,3 +1,4 @@
+
 // Copyright (c) .NET Foundation. All rights reserved.
```
rynowak: @JamesNK

src/Mvc/Mvc.Core/src/MvcOptions.cs
```diff
@@ -102,6 +103,25 @@ public MvcOptions()
...
+        /// <para>
+        /// This option controls whether MVC model binding and validation treats nullable and non-nullable 
+        /// reference types differently. 
+        /// </para>
+        /// <para>
+        /// By default, MVC will treat a non-nullable reference type parameters and properties as-if 
+        /// <c>[Required]</c> has been applied, resulting in validation errors when no value was bound.
+        /// </para>
+        /// </remarks>
+        public bool SuppressImplicitRequiredAttributeForNonNullableReferenceTypes { get; set; }
```
rynowak: I challenge anyone to come up with a better property name. This is a masterpiece.

src/Mvc/Mvc.DataAnnotations/src/DataAnnotationsMetadataProvider.cs
```diff
@@ -23,19 +23,31 @@ internal class DataAnnotationsMetadataProvider :
...
+        private const string NullableAttributeFullTypeName = "System.Runtime.CompilerServices.NullableAttribute";
+        private const string NullableFlagsFieldName = "NullableFlags";
+
         private readonly IStringLocalizerFactory _stringLocalizerFactory;
+        private readonly MvcOptions _options;
         private readonly MvcDataAnnotationsLocalizationOptions _localizationOptions;
 
         public DataAnnotationsMetadataProvider(
-            IOptions<MvcDataAnnotationsLocalizationOptions> options,
+            MvcOptions options,
```
rynowak: Adding the constructor parameter here was the cause of most of the churn. I did a fair bit of mechanical cleanup based on this.

src/Mvc/Mvc.Core/src/ModelBinding/Metadata/ValidationMetadata.cs
JamesNK: image

src/Mvc/Mvc.Core/src/MvcOptions.cs
benaadams: > I challenge anyone to come up with a better property name. This is a masterpiece.

Invert the true/false and call it `RequireValueForNonNullableReferences  { get; set; } = true`?

rynowak: nah, I want it to be false by default. I'm actually happy with the name.

rynowak:
Will add some functional tests tomorrow to capture the E2E scenario. I'm confident that this is working, but we need tests in place that capture the whole workflow to make sure there aren't other blockers.

src/Mvc/Mvc.DataAnnotations/src/DataAnnotationsMetadataProvider.cs
```diff
@@ -380,5 +405,26 @@ private static string GetDisplayGroup(FieldInfo field)
 
             return string.Empty;
         }
+
+        // Internal for testing
+        internal static bool IsNonNullable(IEnumerable<object> attributes)
+        {
+            // [Nullable] is compiler synthesized, comparing by name. 
+            var nullableAtttribute = attributes
+                .Where(a => string.Equals(a.GetType().FullName, NullableAttributeFullTypeName))
```
pranavkm: `attributes.FirstOrDefault(a => string.Equals(a.GetType().FullName, NullableAttributeFullTypeName, StringComparison.Ordinal))`

src/Mvc/Mvc.DataAnnotations/src/DataAnnotationsMetadataProvider.cs
```diff
@@ -380,5 +405,26 @@ private static string GetDisplayGroup(FieldInfo field)
 
             return string.Empty;
         }
+
+        // Internal for testing
+        internal static bool IsNonNullable(IEnumerable<object> attributes)
+        {
+            // [Nullable] is compiler synthesized, comparing by name.
```
pranavkm: Let's keep an eye on https://github.com/dotnet/corefx/issues/36222 in case they start exposing it

src/Mvc/Mvc.DataAnnotations/src/DataAnnotationsMetadataProvider.cs
```diff
@@ -380,5 +405,26 @@ private static string GetDisplayGroup(FieldInfo field)
...
+            var nullableAtttribute = attributes
+                .Where(a => string.Equals(a.GetType().FullName, NullableAttributeFullTypeName))
+                .FirstOrDefault();
+
+            // See: https://github.com/dotnet/roslyn/blob/master/docs/features/nullable-reference-types.md#annotations
+            if (nullableAtttribute != null &&
+                nullableAtttribute.GetType().GetField(NullableFlagsFieldName) is FieldInfo field &&
+                field.GetValue(nullableAtttribute) is byte[] flags
+                && flags.Length >= 0
+                && flags[0] == 1)
```
pranavkm: Does this have trouble with collection types? 

> [Nullable(new[] { 1, 2 })] string[] NotNullMaybeNull; // string?[]!

if not, we should add some comments explaining how this works when more than one value is present in the `NullableAttribute` byte array

src/Mvc/Mvc.DataAnnotations/src/DataAnnotationsMetadataProvider.cs
```diff
@@ -380,5 +405,26 @@ private static string GetDisplayGroup(FieldInfo field)
 
             return string.Empty;
         }
+
+        // Internal for testing
+        internal static bool IsNonNullable(IEnumerable<object> attributes)
```
roji: Note the almost identical logic about to go into EF Core: https://github.com/aspnet/EntityFrameworkCore/pull/15499/files#diff-0ebd43bbbecf10e14f32e65b224af763R28

I actually cache some type information about NullableAttribute for better perf, while supporting multiple NullableAttribute (since it may be synthesized several times into different participating assemblies...).

roji:
Some very similar nullability handling code going into ASP.NET and EF at the very same time (https://github.com/aspnet/EntityFrameworkCore/pull/15499)...

> One other idea (that I didn't do, and think we shouldn't do) would be to use the nullability information to create related objects. So for the following, we could auto-create Address even if the data was missing:

I agree, the less magic going on, the better...

src/Mvc/Mvc.DataAnnotations/src/DataAnnotationsMetadataProvider.cs
rynowak: Yeah, I hadn't thought about this case, but it's worth discussion. 

Note that we don't have anything like this today so this would be a one-off thing or something built into the model binders for collections. We don't support you placing `[Required]` on a type parameter.

rynowak:
> I agree, the less magic going on, the better...

The reason I thought about this at all.... I want to be able to make statements like: "if we violate your nulllability contract then we give you a validation error". I could see someone interpreting that rule in favor of this, but I think it would have serious unintended consequences. Model binding creates objects based on the presence of data in the request that matches the model. Doing this based on nullabitity contracts could lead to infinite recursion instead 😆

rynowak:
@jeremyskinner

src/Mvc/Mvc.DataAnnotations/src/DataAnnotationsMetadataProvider.cs
rynowak: I would certainly prefer that, but this code won't really need to change, it will still be correct.

src/Mvc/Mvc.DataAnnotations/src/DataAnnotationsMetadataProvider.cs
```diff
@@ -380,5 +405,26 @@ private static string GetDisplayGroup(FieldInfo field)
...
+        {
+            // [Nullable] is compiler synthesized, comparing by name. 
+            var nullableAtttribute = attributes
+                .Where(a => string.Equals(a.GetType().FullName, NullableAttributeFullTypeName))
+                .FirstOrDefault();
+
+            // See: https://github.com/dotnet/roslyn/blob/master/docs/features/nullable-reference-types.md#annotations
+            if (nullableAtttribute != null &&
```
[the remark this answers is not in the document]
rynowak: So is your feedback that should I add more t?

At @terrajobst 's suggestion, I got $wife one of these: https://www.amazon.com/Breville-BTM800XL-One-Touch-Tea-Maker/dp/B003LNOPSG It's really nice. We call him T3PO

src/Mvc/Mvc.DataAnnotations/src/DataAnnotationsMetadataProvider.cs
rynowak: @roji does you code need to be thread-safe? That appears to not be...

src/Mvc/Mvc.DataAnnotations/src/DataAnnotationsMetadataProvider.cs
rynowak: There's actually another problem here.

Consider a case like:

```C#
    class Program
    {
        public string Foo { get; set; } = default!;

        public Container<string?> Bar { get; set; } = default!;
    }

    class Container<T>
    {
        public T Item { get; set; } = default!;
    }
```

ModelMetadata is an API you can use to get the metadata for a property, parameter or type.  We're primarily interested in the property case, where you pass in the container type and property info. However, with nnrt, this isn't enough information.  In this example for property `Item` the only place where the metadata reflect the nnrt information is on property `Bar`. 

You need to know the property, the containing type and now the *containing property* in order to create a correct model metadata. Basically with generics in the picture, our current architecture can't support this usage. Put another way, nnrt + generics is *just* type erasure :(

rynowak:
Upated

JamesNK:
Thinking about this some more and it feels a little odd that non-nullable reference types are now required, but non-nullable value types are not. Consider:

```cs
    public class Person
    {
        public string Name { get; set; } = default!;
        public int Age { get; set; }
    }
```

Name is required, Age is not.

rynowak:
`[Required]` means not-null. 

When we look at the *types* of properties and parameters, we're only talking about the allowed state of the model. MVC has the ability to validate the state of the model using `[Required]`. This makes sense because the contract implied by nnrt is about the legal state of a property/parameter/variable. I think that this is the most natural way to think about nnrt by far.

Also note that we really have two meanings for required. Value types like `int` generate client validation messages that require you to specify the values in the form. When validation runs we don't validate the non-null-ness of your `int` property because it's always in a legal state already.

JamesNK:
I see. I'm use to looking at required from a deserialization viewpoint. Often it requires the presence of the data, rather than caring about the value.

I thought Newtonsoft.Json use `[Required]` but it doesn't. There is a `[JsonRequired]` attribute that requires the data exist AND it is not null.

So would required on a non-nullable value type never do anything?

```cs
    public class Person
    {
        [Required]
        public int Age { get; set; }
    }
```

JamesNK:
https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.requiredattribute?view=netframework-4.8#remarks

> The RequiredAttribute attribute specifies that when a field on a form is validated, the field must contain a value. A validation exception is raised if the property is null, contains an empty string (""), or contains only white-space characters.

The added required attribute should set AllowEmptyStrings to true. And add unit tests for empty strings.

src/Mvc/Mvc.DataAnnotations/src/DataAnnotationsMetadataProvider.cs
roji: Nope, the "convention" you are seeing is a composable piece of code that runs once at startup, during EF Core's model building, and instances aren't supposed to be thread-safe. However, we've seen cases of EF Core models which were so big that the model building process was quite slow, which is why I'm doing this caching.

rynowak:
> The added required attribute should set AllowEmptyStrings to true. And add unit tests for empty and white-space only strings

I'll do it for the sake of #Pedantry. In practice it only will matter if you do something wierd. MVC already handles this case for you be treating whitespace/empty strings as null (ConvertEmptyStringToNull 
 defaults to true) https://github.com/aspnet/AspNetCore/blob/master/src/Mvc/Mvc.Core/src/ModelBinding/Binders/SimpleTypeModelBinder.cs#L71

JamesNK:
Ah ok, I wasn't sure if that flag was being used by the validation logic. No point setting it then 😬

A comment and test would be nice tho for clarity

rynowak:
I'm making the change, it's a little more correct so we might as well.

JeremySkinner:
@rynowak One question on this...will it behave in the same way as the current handling for non-nullable value types? (ie if no required rule is explicitly run, or the DataAnnotations provider is disabled, then the default "A value is required" message will be added instead?)

rynowak:
> @rynowak One question on this...will it behave in the same way as the current handling for non-nullable value types? (ie if no required rule is explicitly run, or the DataAnnotations provider is disabled, then the default "A value is required" message will be added instead?)

This is a super set. 

`ModelMetadata.IsRequired` is the thing that adds client side validation messages using jQuery validation. This does `ModelMetadata.IsRequired` and adds an implicit `[Required]`. Is there a specific case or scenario you are worried about?

src/Mvc/Mvc.DataAnnotations/src/DataAnnotationsMetadataProvider.cs
```diff
@@ -380,5 +415,35 @@ private static string GetDisplayGroup(FieldInfo field)
...
+            var nullableAttribute = attributes
+                .Where(a => string.Equals(a.GetType().FullName, NullableAttributeFullTypeName, StringComparison.Ordinal))
+                .FirstOrDefault();
+            if (nullableAttribute == null)
+            {
+                return false;
+            }
+
+            // We don't handle cases where generics and NNRT are used. This runs into a
+            // fundamental limitation of ModelMetadata - we a single the Type and Property/Parameter
```
pranavkm: > we a single the Type and Property/Parameter

Typo

src/Mvc/test/Mvc.FunctionalTests/NonNullableReferenceTypesTest.cs
```diff
@@ -0,0 +1,116 @@
...
+
+namespace Microsoft.AspNetCore.Mvc.FunctionalTests
+{
+    public class NonNullableReferenceTypesTest : IClassFixture<MvcTestFixture<BasicWebSite.StartupWithoutEndpointRouting>>
+    {
+        // Some tests require comparing the actual response body against an expected response baseline
+        // so they require a reference to the assembly on which the resources are located, in order to
+        // make the tests less verbose, we get a reference to the assembly with the resources and we
+        // use it on all the rest of the tests.
+        private static readonly Assembly _resourcesAssembly = typeof(BasicTests).GetTypeInfo().Assembly;
```
pranavkm: Remove

src/Mvc/test/Mvc.FunctionalTests/NonNullableReferenceTypesTest.cs
```diff
@@ -0,0 +1,116 @@
...
+        public async Task CanUseNonNullableReferenceType_WithController_OmitData_ValidationErrors()
+        {
+            // Arrange
+            var parser = new HtmlParser();
+
+            // Act 1
+            var response = await Client.GetAsync("http://localhost/NonNullable");
+
+            // Assert 1
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
```
pranavkm: If you do `response.AssertStatusCodeAsync(HttpStatusCode.OK)`, it'll include the body if this is not OK. Way more useful than a `Expected: OK, Actual: InternalServerError`

pranavkm (approved):

JeremySkinner:
>  Is there a specific case or scenario you are worried about? 

No, was just curious - should all work fine for me. Thanks!

src/Mvc/Mvc.DataAnnotations/src/DataAnnotationsMetadataProvider.cs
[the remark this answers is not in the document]
terrajobst: image


Did you know that [T-3PO actually](https://starwars.fandom.com/wiki/T-3PO_(blue)) existed in the Star Wars universe?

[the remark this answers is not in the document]
rynowak: I was wondering when you were going to show up.

PR #9979, opened 2019-05-05
SignalR: Fix linux build

staff0rd:
Needed when building the BenchmarkServer on linux.

Not needed when building on Windows.


BrennanConroy:
This doesn't seem right. We're running BenchmarkServer on linux many times a day without issues. And our CI should be failing on linux if this is broken.

What version of the sdk and runtime are you using?

staff0rd:
This was failing on the weekend but as of lastest master does not seem to be failing.

FYI I was using `dotnet-install.sh --channel master`.

PR #9997, opened 2019-05-06
Strawman for new authorization packages that can be shared between server and client

SteveSandersonMS:
Our goal is that the same authorization attributes, DI services, and classes should be usable both in server-side ASP.NET Core (without any changes to customer code) *and* in client-side Blazor.

To make that possible, we have to introduce at least one new package where we can put things that should be common to both server and client. Restrictions include:

 * Package must be `netstandard2.0`
 * Package must be shipped both in the shared framework and as a NuGet package
 * Cannot reference APIs that only make sense on the server such as `Http.Abstractions` (or `HttpContext` in particular)

This PR moves the majority of classes out of `Microsoft.AspNetCore.Authorization` and puts them in a new package, `Microsoft.AspNetCore.Authorization.Common`. I'm *not* in any way in love with that name. I'm happy to call it anything else if someone has a preference. It just has to mean "authorization things that work both on the server and on the client".

This shouldn't be a breaking change, since `Microsoft.AspNetCore.Authorization` now references the `.Common` package, so all existing types continue to work (if you rebuild your app). I haven't set up actual type forwarders for people who reference things using reflection. Not sure if we would do that. I also haven't moved any tests since this is just a draft.

Similarly I created `Microsoft.AspNetCore.Authorization.Common.Abstractions` to hold the two interfaces we need that used to be in `Http.Abstractions`.

I haven't built anything end-to-end with this yet so don't even know if it meets our needs. Currently this is just seeking feedback on (a) the whole idea of moving all these things around, and (b) the package naming.


src/Security/Authorization/Common/src/Policy/CommonPolicyEvaluator.cs
```diff
@@ -0,0 +1,60 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Security.Claims;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Authorization.Policy
+{
+    public class CommonPolicyEvaluator : ICommonPolicyEvaluator
```
HaoK: Do we still need a PolicyEvaluator?  It was only introduced to tie authN + authZ together, I thought on the client side there's no authentication needed, so wouldn't the base `IAuthorizationService` be sufficient?

HaoK:
We seem to use Core everywhere, maybe Authorization.Core instead?

SteveSandersonMS:
Since this is probably superseded by #10021, I'll leave it alone for now. Once #10021 is merged, this can probably just be closed.

src/Security/Authorization/Common/src/Policy/CommonPolicyEvaluator.cs
SteveSandersonMS: Yes, we do want to evaluate policies in client-side code.

> I thought on the client side there's no authentication needed

That's not correct. Although authentication isn't *enforced* there (all true enforcement is server-side), we still want policies to exist as a way of informing the client about what the user is going to be allowed to do.

SteveSandersonMS:
Superseded by #10021.

PR #10007, opened 2019-05-06
Add Windows 7 and 8.1 testing on Helix

natemcmaster:
Resolves https://github.com/aspnet/AspNetCore-Internal/issues/1942

* Add new test queues for Win 7 and Win 8.1
* Install SQL Server Local DB in Helix VMs when needed


natemcmaster:
On Windows 7 and 8.1, these test projects completely fail due to lack of SQL Server, as predicted.

image

@HaoK @ajcvickers I know you were having conversations with the dnceng team about SQL Server in Helix. What was the outcome of that thread? It appears we also need SQL Server to run these unit tests for AspNetCore on Helix.

```
Error Message:
 Assert.Contains() Failure
Not found: In Visual Studio, use the Package Manager Console to scaffold a new migration and apply it to the database:
In value:  System.Data.SqlClient.SqlException (0x80131904): A network-related or instance-specific error occurred while establishing a connection to SQL Server. The
```

HaoK:
LocalDb was what was missing for these (and why we switched to the VS queue), maybe in the short term we just skip these tests on these new queues?

natemcmaster:
> switched to the VS queue

Is that the dnceng team recommendation? "Just use VS images"? If so, we'll have to ask them to provision VS images for Win 7 and Win 8.

HaoK:
I think that was just the easiest option since they already had a a pool setup with VS, I found the thread from last time, https://github.com/dotnet/core-eng/issues/5032

jkotalik:
@natemcmaster can you try adding these windows queues for IIS tests too? See https://github.com/aspnet/AspNetCore/blob/783131a7501a1ba5865838e9c4a8d2d5b2f4238d/src/Servers/IIS/IIS/test/FunctionalTest.props. 

Ideally, controlling the machines IIS tests run on should be next to the helix props files.

src/Servers/IIS/IIS/test/FunctionalTest.props
```diff
@@ -16,7 +16,9 @@
     <HelixProjectPlatform Condition="'$(IsIISTest)' == 'true'" Remove="@(HelixProjectPlatform)" />
 
     <HelixProjectTargetQueue Condition="'$(IsIISTest)' == 'true'" Include="Windows.10.Amd64.EnterpriseRS3.ASPNET.Open" />
-
+    <HelixProjectTargetQueue Condition="'$(IsIISTest)' == 'true'" Include="Windows.81.Amd64.Open" />
```
natemcmaster: @jkotalik like this?

We'll see how it goes.

natemcmaster:
Re: SQL Server

@HaoK we talked about this requirement in build team sync up yesterday. Sounds like the dnceng team really doesn't want to pre-install LocalDB SQL Server into Helix machines for us. We came up with some alternatives:
* Only run SQL Server tests on Linux and use Docker to start a mssql DB server
* Switch Identity tests to use the SQLite provider
* Install SQL Server LocalDB on-demand in the helix VM using a script

Thoughts? I'm tempted to start with option 3, install sql server on-demand.

HaoK:
Yeah installing it on demand seems reasonable, option 2 wouldn't likely be too bad either, as it would just be tweaking the connection string in one place I think. @ajcvickers any thoughts on this, is there a long term benefit between testing localdb vs sqlite for our identity efprovider tests?

natemcmaster:
It appears option 3 - install mssql on-demand - works well enough for the tests in this repo.

eng/helix/content/README.md
```diff
@@ -0,0 +1,3 @@
+# Helix content
+
+The content of this folder is included in the test payload for each Helix work item. The code here is mean to be used along side test binaries.
```
dougbu: Suggest "alongside" (one word) 😈

eng/targets/Helix.Common.props
```diff
@@ -12,6 +12,8 @@
     <HelixAvailablePlatform Include="Linux" />
 
     <HelixAvailableTargetQueue Include="Windows.10.Amd64.ClientRS4.VS2017.Open" Platform="Windows" />
+    <HelixAvailableTargetQueue Include="Windows.81.Amd64.Open" Platform="Windows" />
+    <HelixAvailableTargetQueue Include="Windows.7.Amd64.Open" Platform="Windows" />
```
dougbu: Are the agents in these queues actually running the named OS?

dougbu (approved):

eng/targets/Helix.Common.props
natemcmaster: They run the Windows Server equivalent, but per discussion with Eilon, we think this is sufficient for downlevel Windows testing.

jkotalik:
Bleh all IIS tests were skipped... I should have kept better track of these machines as it look likes IIS isn't installed on them.

natemcmaster:
I think something in Helix test reporting is still busted. I'm adding two new queues, but the # of test results didn't increase by much.

Should I be seeing something in the logs about collection of test results from a .trx file? See https://mc.dot.net/#/user/aspnetcore/pr~2Faspnet~2Faspnetcore/ci/20190508.9/workItem/AuthSamples.FunctionalTests-netcoreapp3.0/wilogs for instance

cc @JunTaoLuo @alexperovich

natemcmaster:
> Bleh all IIS tests were skipped...

How did you figure that out? I'm assuming you saw this in logs because I don't see any tests reported as 'skipped'.

image

jkotalik:
Yeah went into the logs: https://mc.dot.net/#/user/aspnetcore/pr~2Faspnet~2Faspnetcore/ci/20190508.9/workItem/IISExpress.FunctionalTests-netcoreapp3.0/wilogs

alexperovich:
I think the `exit 0` at the end of your runtests.cmd is causing an issue.
From this log: https://helix.dot.net/api/2018-03-14/jobs/c3451e6a-3bf5-4337-974e-6c854d9e706d/workitems/AuthSamples.FunctionalTests-netcoreapp3.0/console

The last line executed is the `exit 0`. I think it needs to be `exit /b 0`, the test reporter is never called because IIRC exit without /b exits out of every `call`ed thing.

natemcmaster:
Thanks! I'll give that a try.

natemcmaster:
I think the max retries being set to '4' is causing timeouts on Helix testing. I'd like to set this back to 2.

@jkotalik @Tratcher - I'm seeing a few hundred HttpSys and IIS test failures on all versions of Windows (both the new ones I'm adding and queues that have been enabled for a while) https://dev.azure.com/dnceng/public/_build/results?buildId=183898&view=ms.vss-test-web.build-test-results-tab. Are these test failures a known issue?

Tratcher:
The HttpSys test failures are all HTTPS tests. It relies on the IIS Express certificate and bindings to be set up, it re-uses them. Is IIS Express not installed?

natemcmaster:
Apparently not. Is there a way we can script the installation of IIS Express? The dnceng team is pushing us to install test dependencies on our own and on-demand if possible.

natemcmaster:
Digging through this, we found a few things that require followup

1. Cert generation is busted on test runs. Likely due to old versions of PowerShell which don't support the `-notafter` parameter - @jkotalik 
2. Running the InstallIISFeatures.ps1 script can probably be removed - @jkotalik 
3. We need to add 'SkipOnHelix' attribute to IIS tests that consistently fail on Helix @jkotalik 
3. Windows7 does not have IIS features ( https://github.com/dotnet/core-eng/issues/5506) - @natemcmaster  to follow-up with dnceng team
5. Apparently `MaxRetryCount = 2` means only run tests once? Seems like a bug in Helix - @natemcmaster to followup

jkotalik:
I'll file a follow up issue for some HttpSys/IIS tests that will be disabled due to cert issues on helix. Things look good on helix besides that!

jkotalik:
@natemcmaster we should either remove running UpdateIISExpressCertificate on Windows 7 or ignore the exit code. Right now the IISExpress tests are failing.

natemcmaster:
:up: :date:

* Added a pre-generated TestCert.pfx file and updated the UpdateIISExpressCertificate.ps1 script to install that instead of generating a cert on the fly.
* Disabled Windows 7 tests for IIS. (Once https://github.com/dotnet/core-eng/issues/5506 is complete, we can re-enable)

natemcmaster:
Ok, this is mostly working now. There were a few flaky tests in Kestrel, but I think we should merge.

Biggest improvement here is that we're actually getting test results from Windows now. About 400k test results per build :)

jkotalik (approved):
Merge merge merge

PR #10009, opened 2019-05-06
add support for both env vars in #7916 fix

analogrelay:
Thanks @DavidSimner for saving me from having to wait for end-to-end verification to find this ;).

I decided to support both the (now deprecated) `DOTNET_RUNNING_IN_CONTAINERS` variable (used in https://github.com/dotnet/dotnet-docker/blob/4fd3b74fc5b7444f39f3fdcffa45eef467bccb24/2.1/aspnet/nanoserver-1803/amd64/Dockerfile and https://github.com/dotnet/dotnet-docker/blob/4fd3b74fc5b7444f39f3fdcffa45eef467bccb24/2.1/aspnet/nanoserver-1809/amd64/Dockerfile) as well as `DOTNET_RUNNING_IN_CONTAINER` (the intended variable name). The check is done once per process at start-up.

Since we know the value we expect is `true`, I didn't do what we've done for other "boolean" environment variables and add logic to support both `1` and `true`. I could do that if we felt it was useful though.

### DIFF
--- a/src/DataProtection/DataProtection/src/Internal/ContainerUtils.cs
+++ b/src/DataProtection/DataProtection/src/Internal/ContainerUtils.cs
@@ -80,7 +80,9 @@ namespace Microsoft.AspNetCore.DataProtection.Internal
         private static bool IsProcessRunningInContainer()
         {
             // Official .NET Core images (Windows and Linux) set this. So trust it if it's there.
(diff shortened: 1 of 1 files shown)


natemcmaster:
> add logic to support both 1 and true. I could do that if we felt it was useful though.

Might as well. It's a common-enough expectation for env vars.

natemcmaster (approved):

anurse:
Done.

natemcmaster:
:shipit:

PR #10015, opened 2019-05-06
OpenAPI ServiceReference tool

ryanbrandenburg:
The OpenAPI part of https://github.com/aspnet/specs/pull/276.

#### Description
Create a CLI tool which can be used to add OpenAPIReferences to a project, to be used by the Service Reference feature to generate a client for the given service.

#### Customer impact
Without this tool customers will have to either add service references by hand or through Visual Studio.

#### Regression?
No, this is an entirely new package.

#### Risk
Very low chance of breaking existing workflows since the tool is not currently available. Moderate-to-high risk of unintended behavior in the tool since it's entirely new code which hasn't yet been exercised by users.

### DIFF
--- a/src/Tools/README.md
+++ b/src/Tools/README.md
@@ -1,18 +1,23 @@
-DotNetTools
-===========
+# DotNetTools
 
-## Projects
+## Bundled tools
 
-The folder contains command-line tools for ASP.NET Core that are bundled* in the .NET Core CLI. Follow the links below for more details on each tool.
+The folder contains command-line tools for ASP.NET Core. The following tools are bundled* in the .NET Core CLI. Follow the links below for more details on each tool.
 
- - [dotnet-watch](dotnet-watch/README.md)
- - [dotnet-user-secrets](dotnet-user-secrets/README.md)
- - [dotnet-sql-cache](dotnet-sql-cache/README.md)
- - [dotnet-dev-certs](dotnet-dev-certs/README.md)
+- [dotnet-watch](dotnet-watch/README.md)
+- [dotnet-user-secrets](dotnet-user-secrets/README.md)
+- [dotnet-sql-cache](dotnet-sql-cache/README.md)
+- [dotnet-dev-certs](dotnet-dev-certs/README.md)
 
 *\*This applies to .NET Core CLI 2.1.300-preview2 and up. For earlier versions of the CLI, these tools must be installed separately.*
 
 *For 2.0 CLI and earlier, see <https://github.com/aspnet/DotNetTools/tree/rel/2.0.0/README.md> for details.*
+
+## Non-bundled tools
+
+The following tools are produced by us but not bundled in the .NET Core CLI. They must be aquired independently.
+
+- [dotnet-openapi](dotnet-openapi/README.md)
 
 ## Usage
 
@@ -23,10 +28,11 @@ dotnet watch
 dotnet user-secrets
 dotnet sql-cache
 dotnet dev-certs
+dotnet openapi
 ```
 
 Add `--help` to see more details. For example,
 
-```
+```sh
 dotnet watch --help
 ```
--- a/src/Tools/Shared/TestHelpers/TemporaryCSharpProject.cs
+++ b/src/Tools/Shared/TestHelpers/TemporaryCSharpProject.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
@@ -6,12 +6,12 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Text;
 
-namespace Microsoft.DotNet.Watcher.Tools.Tests
+namespace Microsoft.Extensions.Tools.Internal
 {
     public class TemporaryCSharpProject
     {
         private const string Template =
- @"<Project Sdk=""Microsoft.NET.Sdk"">
+ @"<Project Sdk=""{2}"">
   <PropertyGroup>
     {0}
     <OutputType>Exe</OutputType>
@@ -23,18 +23,21 @@ namespace Microsoft.DotNet.Watcher.Tools.Tests
 
         private readonly string _filename;
         private readonly TemporaryDirectory _directory;
-        private List<string> _items = new List<string>();
-        private List<string> _properties = new List<string>();
+        private readonly List<string> _items = new List<string>();
+        private readonly List<string> _properties = new List<string>();
 
-        public TemporaryCSharpProject(string name, TemporaryDirectory directory)
+        public TemporaryCSharpProject(string name, TemporaryDirectory directory, string sdk)
         {
             Name = name;
             _filename = name + ".csproj";
             _directory = directory;
+            Sdk = sdk;
         }
 
         public string Name { get; }
         public string Path => System.IO.Path.Combine(_directory.Root, _filename);
+
+        public string Sdk { get; }
 
         public TemporaryCSharpProject WithTargetFrameworks(params string[] tfms)
         {
@@ -95,7 +98,7 @@ namespace Microsoft.DotNet.Watcher.Tools.Tests
 
         public void Create()
         {
-            _directory.CreateFile(_filename, string.Format(Template, string.Join("\r\n", _properties), string.Join("\r\n", _items)));
+            _directory.CreateFile(_filename, string.Format(Template, string.Join("\r\n", _properties), string.Join("\r\n", _items), Sdk));
         }
 
         public class ItemSpec
--- a/src/Tools/dotnet-openapi/src/Application.cs
+++ b/src/Tools/dotnet-openapi/src/Application.cs
@@ -1,0 +1,108 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using System.Reflection;
+using System.Threading.Tasks;
+using Microsoft.Build.Locator;
+using Microsoft.DotNet.OpenApi.Commands;
+using Microsoft.Extensions.CommandLineUtils;
+
+namespace Microsoft.DotNet.OpenApi
+{
+    internal class Application : CommandLineApplication
+    {
+        static Application()
+        {
+            MSBuildLocator.RegisterDefaults();
+        }
+
+        public Application(
+            string workingDirectory,
+            Func<string, Task<Stream>> downloadProvider,
+            TextWriter output = null,
+            TextWriter error = null)
+        {
+            DownloadProvider = downloadProvider;
+            Out = output ?? Out;
+            Error = error ?? Error;
+
+            WorkingDirectory = workingDirectory;
+
+            Name = "openapi";
+            FullName = "OpenApi reference management tool";
+            Description = "OpenApi reference management operations.";
+            ShortVersionGetter = GetInformationalVersion;
+
+            HelpOption("-?|-h|--help");
+
+            Invoke = () => {
+                ShowHelp();
+                return 0;
+            };
+
+            Commands.Add(new AddCommand(this));
+            Commands.Add(new RemoveCommand(this));
+            Commands.Add(new RefreshCommand(this));
+        }
+
+        public Func<string, Task<Stream>> DownloadProvider { get; }
+
+        public string WorkingDirectory { get; }
+
+        public new int Execute(params string[] args)
+        {
+            try
+            {
+                return base.Execute(args);
+            }
+            catch (AggregateException ex) when (ex.InnerException != null)
+            {
+                foreach (var innerException in ex.Flatten().InnerExceptions)
+                {
+                    Error.WriteLine(innerException.Message);
+                    if (!(innerException is ArgumentException))
+                    {
+                        Error.WriteLine(innerException.StackTrace);
+                    }
+                }
+                return 1;
+            }
+
+            catch (ArgumentException ex)
+            {
+                // Don't show a call stack when we have unneeded arguments, just print the error message.
+                // The code that throws this exception will print help, so no need to do it here.
+                Error.WriteLine(ex.Message);
+                return 1;
+            }
+            catch (CommandParsingException ex)
+            {
+                // Don't show a call stack when we have unneeded arguments, just print the error message.
+                // The code that throws this exception will print help, so no need to do it here.
+                Error.WriteLine(ex.Message);
+                return 1;
+            }
+            catch (OperationCanceledException)
+            {
+                // This is a cancellation, not a failure.
+                Error.WriteLine("Cancelled");
+                return 1;
+            }
+            catch (Exception ex)
+            {
+                Error.WriteLine(ex.Message);
+                Error.WriteLine(ex.StackTrace);
+                return 1;
+            }
+        }
+
+        private string GetInformationalVersion()
+        {
+            var assembly = typeof(Application).GetTypeInfo().Assembly;
+            var attribute = assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>();
+            return attribute.InformationalVersion;
+        }
+    }
+}
--- a/src/Tools/dotnet-openapi/src/Commands/AddFileCommand.cs
+++ b/src/Tools/dotnet-openapi/src/Commands/AddFileCommand.cs
@@ -1,0 +1,66 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.Extensions.CommandLineUtils;
+using Microsoft.Extensions.Tools.Internal;
+
+namespace Microsoft.DotNet.OpenApi.Commands
+{
+    internal class AddFileCommand : BaseCommand
+    {
+        private const string CommandName = "file";
+
+        public AddFileCommand(AddCommand parent)
+            : base(parent, CommandName)
+        {
+            _sourceFileArg = Argument(SourceProjectArgName, $"The openapi file to add. This must be a path to local openapi file(s)", multipleValues: true);
+        }
+
+        internal readonly CommandArgument _sourceFileArg;
+
+        private readonly string[] ApprovedExtensions = new[] { ".json", ".yaml", ".yml" };
+
+        protected override async Task<int> ExecuteCoreAsync()
+        {
+            var projectFilePath = ResolveProjectFile(ProjectFileOption);
+
+            Ensure.NotNullOrEmpty(_sourceFileArg.Value, SourceProjectArgName);
+
+            foreach (var sourceFile in _sourceFileArg.Values)
+            {
+                var codeGenerator = CodeGenerator.NSwagCSharp;
+                EnsurePackagesInProject(projectFilePath, codeGenerator);
+                if (IsLocalFile(sourceFile))
+                {
+                    if (ApprovedExtensions.Any(e => sourceFile.EndsWith(e)))
+                    {
+                        await Warning.WriteLineAsync($"The extension for the given file '{sourceFile}' should have been one of: {string.Join(",", ApprovedExtensions)}.");
+                        await Warning.WriteLineAsync($"The reference has been added, but may fail at build-time if the format is not correct.");
+                    }
+                    AddServiceReference(OpenApiReference, projectFilePath, sourceFile);
+                }
+                else
+                {
+                    throw new ArgumentException($"{SourceProjectArgName} of '{sourceFile}' was not valid. Valid values are a JSON file or a YAML file");
+                }
+            }
+
+            return 0;
+        }
+
+        private bool IsLocalFile(string file)
+        {
+            return File.Exists(GetFullPath(file));
+        }
+
+        protected override bool ValidateArguments()
+        {
+            Ensure.NotNullOrEmpty(_sourceFileArg.Value, SourceProjectArgName);
+            return true;
+        }
+    }
+}
--- a/src/Tools/dotnet-openapi/src/Commands/AddURLCommand.cs
+++ b/src/Tools/dotnet-openapi/src/Commands/AddURLCommand.cs
@@ -1,0 +1,72 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.IO;
+using System.Threading.Tasks;
+using Microsoft.Extensions.CommandLineUtils;
+using Microsoft.Extensions.Tools.Internal;
+
+namespace Microsoft.DotNet.OpenApi.Commands
+{
+    internal class AddURLCommand : BaseCommand
+    {
+        private const string CommandName = "url";
+        private const string DefaultOpenAPIDir = "openapi";
+        private const string DefaultOpenAPIFile = "openapi.json";
+
+        private const string OutputFileName = "--output-file";
+
+        public AddURLCommand(AddCommand parent)
+            : base(parent, CommandName)
+        {
+            _outputFileOption = Option(OutputFileName, "The destination to download the remote Open API file to.", CommandOptionType.SingleValue);
+            _sourceFileArg = Argument(SourceProjectArgName, $"The openapi file to add. This must be a URL to a remote openapi file.", multipleValues: true);
+        }
+
+        internal readonly CommandOption _outputFileOption;
+
+        internal readonly CommandArgument _sourceFileArg;
+
+        protected override async Task<int> ExecuteCoreAsync()
+        {
+            var projectFilePath = ResolveProjectFile(ProjectFileOption);
+
+            var sourceFile = Ensure.NotNullOrEmpty(_sourceFileArg.Value, SourceProjectArgName);
+
+            string outputFile;
+            if (_outputFileOption.HasValue())
+            {
+                outputFile = _outputFileOption.Value();
+            }
+            else
+            {
+                outputFile = Path.Combine(DefaultOpenAPIDir, DefaultOpenAPIFile);
+            }
+            var codeGenerator = CodeGenerator.NSwagCSharp;
+            EnsurePackagesInProject(projectFilePath, codeGenerator);
+
+            if (IsUrl(sourceFile))
+            {
+                var destination = GetFullPath(outputFile);
+                // We have to download the file from that URL, save it to a local file, then create a AddServiceLocalReference
+                // Use this task https://github.com/aspnet/AspNetCore/commit/91dcbd44c10af893374cfb36dc7a009caa4818d0#diff-ea7515a116529b85ad5aa8e06e4acc8e
+                await DownloadToFileAsync(sourceFile, destination, overwrite: false);
+
+                AddServiceReference(OpenApiReference, projectFilePath, outputFile, sourceFile);
+            }
+            else
+            {
+                Error.Write($"{SourceProjectArgName} was not valid. Valid values are URLs");
+                return 1;
+            }
+
+            return 0;
+        }
+
+        protected override bool ValidateArguments()
+        {
+            Ensure.NotNullOrEmpty(_sourceFileArg.Value, SourceProjectArgName);
+            return true;
+        }
+    }
+}
--- a/src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
+++ b/src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
@@ -1,0 +1,340 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Security.Cryptography;
+using System.Threading.Tasks;
+using Microsoft.Build.Evaluation;
+using Microsoft.Extensions.CommandLineUtils;
+
+namespace Microsoft.DotNet.OpenApi.Commands
+{
+    internal abstract class BaseCommand : CommandLineApplication
+    {
+        protected string WorkingDirectory;
+
+        protected const string SourceProjectArgName = "source-file";
+        public const string OpenApiReference = "OpenApiReference";
+        public const string OpenApiProjectReference = "OpenApiProjectReference";
+        protected const string SourceUrlAttrName = "SourceUrl";
+
+        public BaseCommand(CommandLineApplication parent, string name)
+        {
+            Parent = parent;
+            Name = name;
+            Out = parent.Out ?? Out;
+            Error = parent.Error ?? Error;
+
+            ProjectFileOption = Option("-p|--project", "The project to add a reference to", CommandOptionType.SingleValue);
+
+            Help = HelpOption("-?|-h|--help");
+            if (Parent is Application)
+            {
+                WorkingDirectory = ((Application)Parent).WorkingDirectory;
+            }
+            else
+            {
+                WorkingDirectory = ((Application)Parent.Parent).WorkingDirectory;
+            }
+
+            OnExecute(ExecuteAsync);
+        }
+
+        public CommandOption ProjectFileOption { get; }
+
+        internal CommandOption Help { get; }
+
+        public TextWriter Warning
+        {
+            get { return Out; }
+        }
+
+        protected abstract Task<int> ExecuteCoreAsync();
+
+        protected abstract bool ValidateArguments();
+
+        private async Task<int> ExecuteAsync()
+        {
+            if (!ValidateArguments() || Help.HasValue())
+            {
+                ShowHelp();
+                return 1;
+            }
+
+            return await ExecuteCoreAsync();
+        }
+
+        internal FileInfo ResolveProjectFile(CommandOption projectOption)
+        {
+            string project;
+            if (projectOption.HasValue())
+            {
+                project = projectOption.Value();
+                project = GetFullPath(project);
+                if (!File.Exists(project))
+                {
+                    throw new ArgumentException($"The project '{project}' does not exist.");
+                }
+            }
+            else
+            {
+                var projects = Directory.GetFiles(WorkingDirectory, "*.csproj", SearchOption.TopDirectoryOnly);
+                if (projects.Length == 0)
+                {
+                    throw new ArgumentException("No project files were found in the current directory. Either move to a new directory or provide the project explicitly");
+                }
+                if (projects.Length > 1)
+                {
+                    throw new ArgumentException("More than one project was found in this directory, either remove a duplicate or explicitly provide the project.");
+                }
+
+                project = projects[0];
+            }
+
+            return new FileInfo(project);
+        }
+
+        protected Project LoadProject(FileInfo projectFile)
+        {
+            var project = ProjectCollection.GlobalProjectCollection.LoadProject(
+                projectFile.FullName,
+                globalProperties: null,
+                toolsVersion: null);
+            project.ReevaluateIfNecessary();
+            return project;
+        }
+
+        internal bool IsProjectFile(string file)
+        {
+            return File.Exists(Path.GetFullPath(file)) && file.EndsWith(".csproj");
+        }
+
+        internal bool IsUrl(string file)
+        {
+            return Uri.TryCreate(file, UriKind.Absolute, out var _) && file.StartsWith("http");
+        }
+
+        internal void AddServiceReference(
+            string tagName,
+            FileInfo projectFile,
+            string sourceFile,
+            string sourceUrl = null)
+        {
+            var project = LoadProject(projectFile);
+            var items = project.GetItems(tagName);
+            var fileItems = items.Where((i) => {
+                return string.Equals(GetFullPath(i.EvaluatedInclude), GetFullPath(sourceFile), StringComparison.Ordinal);
+            });
+
+            if (fileItems.Count() > 1)
+            {
+                Warning.Write($"More than one reference to {sourceFile} already exists. Duplicate references could lead to unexpected behavior.");
+                return;
+            }
+
+            if (sourceUrl != null)
+            {
+                if(items.Any(i => string.Equals(i.GetMetadataValue(SourceUrlAttrName), sourceUrl)))
+                {
+                    Warning.Write($"A reference to '{sourceUrl}' already exists in '{project.FullPath}'.");
+                    return;
+                }
+            }
+
+            if (fileItems.Count() == 0)
+            {
+                var metadata = new Dictionary<string, string>();
+
+                if (!string.IsNullOrEmpty(sourceUrl))
+                {
+                    metadata[SourceUrlAttrName] = sourceUrl;
+                }
+
+                project.AddElementWithAttributes(tagName, sourceFile, metadata);
+            }
+            else
+            {
+                Warning.Write($"A reference to '{sourceFile}' already exists in '{project.FullPath}'.");
+                return;
+            }
+        }
+
+        internal async Task DownloadToFileAsync(string url, string destinationPath, bool overwrite)
+        {
+            Application application;
+            if (Parent is Application)
+            {
+                application = (Application)Parent;
+            }
+            else
+            {
+                application = (Application)Parent.Parent;
+            }
+
+            var content = await application.DownloadProvider(url);
+            await WriteToFileAsync(content, destinationPath, overwrite);
+        }
+
+        internal void EnsurePackagesInProject(FileInfo projectFile, CodeGenerator codeGenerator)
+        {
+            var packages = GetServicePackages(codeGenerator);
+            foreach (var (packageId, version) in packages)
+            {
+                var args = new string[] {
+                    "add",
+                    "package",
+                    packageId,
+                    "--version",
+                    version,
+                    "--no-restore"
+                };
+
+                var muxer = DotNetMuxer.MuxerPathOrDefault();
+                if(string.IsNullOrEmpty(muxer))
+                {
+                    throw new ArgumentException($"dotnet was not found on the path.");
+                }
+
+                var startInfo = new ProcessStartInfo
+                {
+                    FileName = muxer,
+                    Arguments = string.Join(" ", args),
+                    WorkingDirectory = projectFile.Directory.FullName,
+                    RedirectStandardError = true,
+                    RedirectStandardOutput = true,
+                };
+
+                var process = Process.Start(startInfo);
+
+                if (!process.WaitForExit(20*1000))
+                {
+                    throw new ArgumentException($"Adding package `{packageId}` to `{projectFile.Directory}` took too long.");
+                }
+
+                if (process.ExitCode != 0)
+                {
+                    Error.Write(process.StandardError.ReadToEnd());
+                    Error.Write(process.StandardOutput.ReadToEnd());
+                    throw new ArgumentException($"Could not add package `{packageId}` to `{projectFile.Directory}`");
+                }
+            }
+        }
+
+        internal string GetFullPath(string path)
+        {
+            return Path.IsPathFullyQualified(path)
+                ? Path.GetFullPath(path)
+                : Path.GetFullPath(path, WorkingDirectory);
+        }
+
+        private static IEnumerable<Tuple<string, string>> GetServicePackages(CodeGenerator type)
+        {
+            var name = Enum.GetName(typeof(CodeGenerator), type);
+            var attributes = typeof(Program).Assembly.GetCustomAttributes<AssemblyMetadataAttribute>();
+            var attribute = attributes.Single(a => string.Equals(a.Key, name, StringComparison.OrdinalIgnoreCase));
+
+            var packages = attribute.Value.Split(';', StringSplitOptions.RemoveEmptyEntries);
+            var result = new List<Tuple<string, string>>();
+            foreach (var package in packages)
+            {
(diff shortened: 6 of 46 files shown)


src/Tools/README.md
```diff
@@ -23,6 +24,7 @@ dotnet watch
 dotnet user-secrets
 dotnet sql-cache
 dotnet dev-certs
+dotnet openapi
```
natemcmaster: Is this really going to be bundled in the .NET Core SDK? If not, this command deserves to be documented differently.

src/Tools/dotnet-openapi/test/OpenApiAddTests.cs
```diff
@@ -0,0 +1,245 @@
...
+namespace Microsoft.DotNet.OpenApi.Add.Tests
+{
+    public class OpenApiAddTests : OpenApiTestBase
+    {
+        public OpenApiAddTests(ITestOutputHelper output) : base(output){}
+
+        [Fact]
+        public void OpenApi_Add_GlobbingOpenApi()
+        {
+            throw new NotImplementedException();
```
natemcmaster: Eh?

src/Tools/dotnet-openapi/test/dotnet-openapi.Tests.csproj
```diff
@@ -0,0 +1,40 @@
...
+      <_Parameter1>TestSettings:RuntimeFrameworkVersion</_Parameter1>
+      <_Parameter2>$(RuntimeFrameworkVersion)</_Parameter2>
+    </AssemblyAttribute>
+  </ItemGroup>
+
+  <Target Name="CleanTestProjects" BeforeTargets="CoreCompile">
+    <RemoveDir Directories="$(TargetDir)TestProjects" Condition="Exists('$(TargetDir)TestProjects')" />
+  </Target>
+
+  <Target Name="PublishDotNetWatch" BeforeTargets="Build">
```
natemcmaster: Copy pasta. Needs a new name 🙂

build/repo.props
```diff
@@ -51,7 +51,7 @@
 
     <!-- These projects are meant to be executed by tests. -->
     <ProjectToExclude Include="
-                      $(RepositoryRoot)src\Tools\dotnet-watch\test\TestProjects\**\*.csproj;
+                      $(RepositoryRoot)src\Tools\Tests.Common\TestProjects\**\*.csproj;
```
natemcmaster: Nit: put in a folder named testassets. This is a shared test library, not a new package, so needs to be categorized as a test asset. This happens automatically if the parent folder is named testassets.

src/Tools/Tests.Common/Tests.Common.csproj
```diff
@@ -0,0 +1,24 @@
...
+    <Content Include="TestProjects\**\*" CopyToOutputDirectory="PreserveNewest" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Folder Include="Properties\" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <EmbeddedResource Update="TestContent\*.txt">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
```
natemcmaster: This is an unusual choice for embedded content. Any reason why it’s both embedded and copied to build output?

src/Tools/dotnet-openapi/test/OpenApiTestBase.cs
```diff
@@ -0,0 +1,71 @@
...
+    public class OpenApiTestBase : IDisposable
+    {
+        protected readonly TemporaryDirectory _tempDir;
+        protected readonly TestConsole _console;
+        private readonly StringBuilder _output = new StringBuilder();
+        protected readonly StringBuilder _error = new StringBuilder();
+        protected readonly ITestOutputHelper _outputHelper;
+
+        // TODO: Use a more permanent URL
+        protected const string SwaggerJsonUrl = "https://raw.githubusercontent.com/glennc/clientgen/master/ConsoleClient/Server.v1.json";
```
natemcmaster: This is not okay. Can you copy this content into our source code instead?

src/Tools/dotnet-openapi/src/Program.cs
```diff
@@ -0,0 +1,441 @@
...
+                    _reporter.Error(process.StandardError.ReadToEnd());
+                    _reporter.Error($"Could not add package `{packageId}` to `{projectFile.Directory}`");
+                    throw new ArgumentException();
+                }
+            }
+        }
+
+        private XmlDocument LoadProject(FileInfo projectFile, out XmlElement projNode)
+        {
+            var projXml = new XmlDocument();
```
natemcmaster: MSBuild project evaluation is more nuanced than parsing XML. I don’t know the requirements of this tool, but if projects are allowed to define conditional items, vary things by target framework, or use Imports, the simple XML evaluation is going to be insufficient. We can chat about the right way to evaluate and edit an MSBuild file if that’s the case.

src/Tools/dotnet-watch/test/dotnet-watch.Tests.csproj
```diff
@@ -10,11 +10,11 @@
   <ItemGroup>
     <Compile Include="$(ToolSharedSourceRoot)TestHelpers\**\*.cs" />
     <Compile Include="$(SharedSourceRoot)test\SkipOnHelixAttribute.cs" />
-    <Content Include="TestProjects\**\*" CopyToOutputDirectory="PreserveNewest" />
```
natemcmaster: I believe this will cause failures in dotnet-watch tests.

natemcmaster (commented):
A few questions:

* Why has all of the dotnet-watch test content been moved into a shared test library? I don’t see it being used by dotnet-openapi tests nor would I expect it to be.
* The code indicates this will be a new first-class, built-in dotnet CLI command. Is that the actual intention for this tool? Or was that an just the result of copying code from dotnet-watch?

src/Tools/dotnet-openapi/test/OpenApiAddTests.cs
ryanbrandenburg: I realized I hadn't accounted for this as I was sending the PR so I created a test that would fail every time to remind me to deal with it before making this a non-Design PR.

src/Tools/README.md
ryanbrandenburg: @shirhatti now that I think about it I seem to remember you saying these wouldn't be bundled, but I'm not sure. Can we get clarification?

src/Tools/dotnet-openapi/test/OpenApiTestBase.cs
```diff
@@ -19,7 +19,7 @@ public class OpenApiTestBase : IDisposable
         protected readonly ITestOutputHelper _outputHelper;
 
         // TODO: Use a more permanent URL
-        protected const string SwaggerJsonUrl = "https://raw.githubusercontent.com/glennc/clientgen/master/ConsoleClient/Server.v1.json";
+        protected const string SwaggerJsonUrl = "https://raw.githubusercontent.com/aspnet/AspNetCore/rybrande/ServiceReference/src/Tools/testassets/TestContent/swagger.json.txt";
```
natemcmaster: Still not okay to rely on an ephemeral URL, even if it's from our repo. What I really meant was that we shouldn't be relying on a remote file to exist on GitHub at all. I would recommend mocking the downloading instead of actually downloading remote code during the test.

ryanbrandenburg:
Failures seem to be related to signing, I'll check with @natemcmaster about what went wrong.

ryanbrandenburg:
:up::date:

dougbu:
@ryanbrandenburg what parts of this PR should we ignore? Or, is it no longer in "[Design]"?

dougbu (changes requested):
Only about a third of the way through this so far ☹️

src/Tools/dotnet-openapi/src/Commands/AddFileCommand.cs
```diff
@@ -0,0 +1,70 @@
...
+{
+    internal class AddFileCommand : BaseCommand
+    {
+        private const string CommandName = "file";
+
+        public AddFileCommand(AddCommand parent)
+            : base(parent, CommandName)
+        {
+            _classNameOpt = Option("-c|--class-name", "The name of the class to be generated", CommandOptionType.SingleValue);
+            _outputFileOpt = Option("-o|--output-file", "The name of the file to output the swagger file to", CommandOptionType.SingleValue);
```
dougbu: I am not sure it's necessary to handle all metadata options `@(OpenApiReference)` supports. I suggest supporting only `--download-dir` (as speced) and perhaps `--namespace` i.e. neither of these.

src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
```diff
@@ -0,0 +1,208 @@
...
+using Microsoft.Extensions.Tools.Internal;
+
+namespace Microsoft.DotNet.OpenApi.Commands
+{
+    internal abstract class BaseCommand : CommandLineApplication
+    {
+        protected string WorkingDir;
+
+        protected const string SourceFileArgName = "source-file";
+        protected const string DefaultSwaggerFile = "swagger.v1.json";
```
dougbu: Swagger is an older term. Stick with Open API everywhere.

src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
```diff
@@ -0,0 +1,208 @@
...
+        protected const string SourceUrlAttrName = "SourceUrl";
+        
+        public BaseCommand(CommandLineApplication parent, string name)
+        {
+            Parent = parent;
+            Name = name;
+            Out = parent.Out ?? Out;
+            Error = parent.Error ?? Error;
+
+            SourceFileArg = Argument(SourceFileArgName, $"The openapi file to {name}. This can be a path to a local openapi file, " +
```
dougbu: Move this up into `AddFileCommand` and similar so it isn't part of `AddCommand` and can have a more specific description

src/Tools/dotnet-openapi/src/Application.cs
```diff
@@ -0,0 +1,92 @@
...
+{
+    internal class Application : CommandLineApplication
+    {
+        static Application()
+        {
+            MSBuildLocator.RegisterDefaults();
+        }
+
+        public Application(
+            string workingDir,
```
dougbu: Be consistent throughout this PR. I prefer `workingDirectory`.

src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
```diff
@@ -0,0 +1,208 @@
...
+               "a URI to a remote openapi file or a path to a *.csproj file containing openapi endpoints", multipleValues: true);
```
dougbu: The correct term is `URL` except when using the `Uri` class

src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
```diff
@@ -0,0 +1,208 @@
...
+                if (csprojs.Length == 0)
+                {
+                    Error.Write("No csproj files were found in the current directory. Either move to a new directory or provide the project explicitly");
+                }
+                if (csprojs.Length > 1)
+                {
+                    Error.Write("More than one csproj was found in this directory, either remove a duplicate or explicitly provide the project.");
+                }
+
+                csproj = csprojs.Single();
```
dougbu:
```suggestion
                csproj = csprojs[0];
```
No need for Linq here

src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
```diff
@@ -0,0 +1,208 @@
...
+                csproj = projectOption.Value();
+                csproj = Path.Combine(WorkingDir, csproj);
+                if (!File.Exists(csproj))
+                {
+                    Error.Write("The given csproj does not exist.");
+                }
+            }
+            else
+            {
+                var csprojs = Directory.GetFiles(WorkingDir, "*.csproj", SearchOption.TopDirectoryOnly);
```
dougbu: s/csproj/project/ throughout

src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
```diff
@@ -0,0 +1,208 @@
...
+        }
+
+        internal bool IsUrl(string file)
+        {
+            return Uri.TryCreate(file, UriKind.Absolute, out var _) && file.StartsWith("http");
+        }
+
+        public async Task DownloadAndOverwriteAsync(string sourceFile, string destinationPath, bool overwrite)
+        {
+            Application parent;
```
dougbu: s/parent/application/

src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
```diff
@@ -0,0 +1,208 @@
...
+        {
+            return File.Exists(file) && file.EndsWith(".csproj");
+        }
+
+        internal bool IsUrl(string file)
+        {
+            return Uri.TryCreate(file, UriKind.Absolute, out var _) && file.StartsWith("http");
+        }
+
+        public async Task DownloadAndOverwriteAsync(string sourceFile, string destinationPath, bool overwrite)
```
dougbu: - odd method name given this doesn't always overwrite
- odd first parameter name given it must be a URL

src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
```diff
@@ -0,0 +1,208 @@
...
+            else
+            {
+                parent = (Application)Parent.Parent;
+            }
+
+            var content = await parent.DownloadProvider(sourceFile);
+            await WriteToFile(content, destinationPath, overwrite);
+        }
+
+        private async Task WriteToFile(string content, string destinationPath, bool overwrite)
```
dougbu: Not a fan of serializing the downloaded content to a `string`. Suggest copying the `Stream` to the output file directly

src/Tools/dotnet-openapi/src/Commands/RefreshCommand.cs
```diff
@@ -0,0 +1,62 @@
...
+            {
+                var destination = FindReferenceFromUrl(projectFile, sourceFile);
+                using (var client = new HttpClient())
+                {
+                    await DownloadAndOverwriteAsync(sourceFile, destination, overwrite: true);
+                }
+            }
+            else
+            {
+                Error.Write($"'dotnet openapi refresh' must be given a url");
```
dougbu: s/url/URL/

dougbu (dismissed):
I'm making slow progress here and am going to wait until initial comments are addressed. Suggest we go through that (non-design I hope) iteration together to save time

src/Tools/Shared/TestHelpers/TemporaryCSharpProject.cs
```diff
@@ -1,17 +1,17 @@
...
 using System.Text;
 
-namespace Microsoft.DotNet.Watcher.Tools.Tests
+namespace Microsoft.DotNet.Tools
 {
     public class TemporaryCSharpProject
     {
         private const string Template =
- @"<Project Sdk=""Microsoft.NET.Sdk"">
+ @"<Project Sdk=""Microsoft.NET.Sdk.Web"">
```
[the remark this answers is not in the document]
ryanbrandenburg: We actually run the Temporary project in one case as a smoke test, without making this a web SDK it fails to compile because of missing namespaces. It shouldn't really affect any of the other tests because none of them build their projects.

dougbu:
@ryanbrandenburg [NSwag.ApiDescription.Client v12.3.0](https://www.nuget.org/packages/NSwag.ApiDescription.Client/) is now available

RicoSuter:
@dougbu yep, but there we will nit to issue a patch soon.. but shouldnt be a problem for you.. i will unlist the old package. Ping me if there is a problem..

ryanbrandenburg:
#FunFact: `dotnet-openapi` is already taken. We're going to have to come up with a different name.

RicoSuter:
Taken for what functionality?

dougbu:
> Taken for what functionality?

Looks like a `dotnet` tool that does code generation based on an Open API document. The [README](https://dev.azure.com/janbaarssen/_git/Open%20API%20Generator?path=%2FREADME.md&version=GBmaster) in their repository mentions it's a .NET Core port of [OpenAPITools/openapi-generator](https://github.com/OpenAPITools/openapi-generator)

src/Tools/README.md
ryanbrandenburg: I've updated this README to reflect that not everything will be shipping in the CLI

shirhatti:
> #FunFact: dotnet-openapi is already taken. We're going to have to come up with a different name.

We messed up. Should have reserved the package name 😢 

@JunTaoLuo Let's start a thread with NuGet folks to ensure we at least reserve the dotnet-grpc package name

src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
```diff
@@ -0,0 +1,290 @@
...
+                    Arguments = string.Join(" ", args),
+                    WorkingDirectory = projectFile.Directory.FullName,
+                    RedirectStandardError = true,
+                    RedirectStandardOutput = true,
+                };
+
+                var process = Process.Start(startInfo);
+                process.WaitForExit(20 * 1000);
+
+                if (process.ExitCode != 0)
```
JunTaoLuo: Hmm I think you should check the bool returned for WaitForExit instead of just the ExitCode in case the process didn't run to completion. In that case ExitCode will throw InvalidOperationException I believe. Also 20 seconds is a long time, since this isn't doing a restore. I would time it out earlier.

src/Tools/dotnet-openapi/src/Commands/RefreshCommand.cs
```diff
@@ -0,0 +1,73 @@
...
+        protected override async Task<int> ExecuteCoreAsync()
+        {
+            var projectFile = ResolveProjectFile(ProjectFileOption);
+
+            var sourceFile = Ensure.NotNullOrEmpty(_sourceFileArg.Value, SourceProjectArgName);
+
+            if (IsUrl(sourceFile))
+            {
+                var destination = FindReferenceFromUrl(projectFile, sourceFile);
+                using (var client = new HttpClient())
```
JunTaoLuo: FYI HttpClient is designed to be reused instead of being instantiated and disposed. However, since the commandline tool isn't long lived, it's not a problem here.

src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
RicoSuter: If this actually waiting until the openapi spec has been generated then it might make sense to even increase it as it might take quite some time if there is a big API surface (ie lots of types have to be reflected, xml docs loaded etc)

src/Tools/dotnet-openapi/src/Commands/RefreshCommand.cs
ryanbrandenburg: Good to bring it up though, I've definitely been guilty of throwing them away too much before, so reminders never hurt.

ryanbrandenburg: Also, turns out we didn't actually use that client.

src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
ryanbrandenburg: To my knowledge the spec is not generated when you add the packages we add, so I think @JunTaoLuo is right that we should reduce the timeout. I made it 5 seconds, which is still probably more than is needed, but not by an insane amount if you account for slow machines.

src/Tools/dotnet-openapi/src/ProjectExtensions.cs
```diff
@@ -0,0 +1,104 @@
...
+                ret.Condition = condStr;
+            }
+
+            root.InsertAfterChild(ret, root.LastItemGroup());
+            return ret;
+        }
+
+        public static void AddElementWithAttributes(this Project project, string tagName, string include, IDictionary<string, string> metadata)
+        {
+            var root = ProjectRootElement.Open(project.FullPath);
```
JunTaoLuo: I found a better API: call Project.AddItem(tag, file).Single() which gives a ProjectItem. You can then call ProjectItem.Xml.AddMetadata(attributename, attributevalue, expressAsAttribute).

Not sure why AddItem gives an ienumerable of items though. Maybe it's evaluating globs?

ryanbrandenburg: Nice! I knew about the `AddItem` function but couldn't figure out how to make it let me add metadata as attributes.

ryanbrandenburg:
There seems to be a Linux-only hang in either the tool or the tests. I'm trying to learn more from a VM now.

ryanbrandenburg:
These Helix failures are an actual problem. NSwag.ApiDescrption.Client depends on NSwag.MSBuild, which contains the tool `dotnet-nswag`, which relies on netcoreapp2.1, so we get the error:

```
dotnet "C:\dotnetbuild\work\6d36a60c-3569-4678-a190-76a12eee5099\Payload\home\.nuget\packages\nswag.msbuild\12.0.0\build\../tools/NetCore21/dotnet-nswag.dll" swagger2csclient /className:openapiClient /namespace:testproj /input:C:\Users\runner\AppData\Local\Temp\dotnet-tool-tests\eb3f933870184eecb405591d011d1a40\openapi.json /output:obj\openapiClient.cs 
It was not possible to find any compatible framework version
The specified framework 'Microsoft.NETCore.App', version '2.1.0' was not found.
- The following frameworks were found:
3.0.0-preview5-27626-15 at [C:\dotnetbuild\work\6d36a60c-3569-4678-a190-76a12eee5099\Payload\sdk\x64\shared\Microsoft.NETCore.App]
3.0.0-preview6-27720-09 at [C:\dotnetbuild\work\6d36a60c-3569-4678-a190-76a12eee5099\Payload\sdk\x64\shared\Microsoft.NETCore.App]
```

This leads to two questions:
1. Is this an acceptable UX? @danroth27.
2. If we go forward with the requirement that NSwag scenarios need netcoreapp2.1 how do we tell Helix to install it for this scenario (preferably only this one, we don't want to pollute).

dougbu:
Adding @glennc 

@ryanbrandenburg .NET Core 2.1 is the LTS version and the NSwag targets use that version when generating code for projects targeting any .NET Core App version. (Will use a .NET Framework executable when targeting that framework.) NSwag.MSBuild doesn't contain an assembly for .NET Core 3.0. So, you're stuck needing .NET Core 2.1.

ryanbrandenburg:
(I don't know why I can't get it into my head that @glennc is the PM for this feature). @dougbu I know why and how netcoreapp2.1 is being used now, but the question still remains if that's an acceptable UX. I don't know of any other tool of component we ship which requires a framework other than the "current" one for that train.

To be clear, I fully expect the answer here to be "yes, this is fine", but I still think we ought to have the discussion with @glencc so we know exactly what we're getting into.

shirhatti:
It's worth looking into whether netcoreapp2.1 is actually required for dotnet-nswag. In other tools that we ship, we've enabled `rollForwardOnNoCandidateFx`. If possible, we should get the nswag tool to enable this as well.

dougbu:
> If possible, we should get the nswag tool to enable this as well.

I wrote the target that invokes the NSwag tool and could send them a PR to add `--roll-forward-on-no-candidate-fx 2` if that's what we want.

shirhatti:
I'd rather see it added to the runtimeconfig.json as opposed to passed to the tool invocation in the target so that everyone benefits.

src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
```diff
@@ -0,0 +1,294 @@
...
+            }
+
+            await Out.WriteAsync($"Downloading to '{destinationPath}'.");
+            var reachedCopy = false;
+            try
+            {
+                if (destinationExists)
+                {
+                    // Check hashes before using the downloaded information.
+                    var downloadHash = DownloadFileExtensions.GetHash(content);
```
JunTaoLuo: Curious, since all this is doing is comparing hashes, if the local copy was modified after download and is newer, it would be overwritten by the remote copy? Is this the behaviour we want?

src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
```diff
@@ -0,0 +1,294 @@
...
+                    await outStream.FlushAsync();
+                }
+            }
+            catch (Exception ex)
+            {
+                await Error.WriteAsync($"Downloading failed.");
+                await Error.WriteAsync(ex.ToString());
+                if (reachedCopy)
+                {
+                    File.Delete(destinationPath);
```
JunTaoLuo: What about the directory that was created?

src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
dougbu: This makes sense from my perspective: The user added or refreshed an `<OpenApiReference>` to a URL and the chosen location on disk _happened_ to match an existing file. **Definitely** expect the file to be overwritten in the `dotnet openapi refresh` case when the download is different.

Guess we could make downloading over an existing file an error in the `dotnet openapi add` case (but I'm not sure it's necessary). @glennc I defer to you on best behaviour here.

src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
ryanbrandenburg: What if multiple levels of directories were created? What if other files were put into that directory, or some other directory somewhere along the chain we created? Doing "the right thing" in this case is actually kind of complicated, and I don't think it's really worth it since this failure mode is much less likely than getting non-200 from the result. CC @dougbu (who wrote this code for a different thing) for opinions.

ryanbrandenburg:
> I'd rather see it added to the runtimeconfig.json as opposed to passed to the tool invocation in the target so that everyone benefits.

I'll get with @glennc and see what we can do.

ryanbrandenburg:
Other than changing the name and getting an updated version of NSwag this should be pretty close to the final form, can I get re-reviews?

src/Tools/dotnet-openapi/src/Commands/AddFileCommand.cs
```diff
@@ -0,0 +1,60 @@
...
+                }
+            }
+
+            return Task.FromResult(0);
+        }
+
+        private bool IsLocalFile(string file)
+        {
+            var fullPath = GetFullPath(file);
+            return File.Exists(fullPath) && (file.EndsWith(".json") || file.EndsWith(".yaml") || file.EndsWith(".yml"));
```
JunTaoLuo: Nit: A bit of a weird scenario but what if the file is in the correct format but doesn't have the correct extension? Maybe output a warning and try adding it anyway, and let it fail during build?

src/Tools/dotnet-openapi/src/Commands/AddURLCommand.cs
```diff
@@ -0,0 +1,72 @@
...
+            else
+            {
+                outputFile = Path.Combine(DefaultOpenAPIDir, DefaultOpenAPIFile);
+            }
+            var codeGenerator = CodeGenerator.NSwagCSharp;
+            EnsurePackagesInProject(projectFilePath, codeGenerator);
+
+            if (IsUrl(sourceFile))
+            {
+                var destination = Path.Combine(WorkingDirectory, outputFile);
```
JunTaoLuo: What if the outputFile is an absolute path?

src/Tools/dotnet-openapi/src/Commands/AddURLCommand.cs
```diff
@@ -0,0 +1,72 @@
...
+                // We have to download the file from that URL, save it to a local file, then create a AddServiceLocalReference
+                // Use this task https://github.com/aspnet/AspNetCore/commit/91dcbd44c10af893374cfb36dc7a009caa4818d0#diff-ea7515a116529b85ad5aa8e06e4acc8e
+                await DownloadToFileAsync(sourceFile, destination, overwrite: false);
```
JunTaoLuo: Why not overwrite?

src/Tools/dotnet-openapi/src/DownloadFile.cs
```diff
@@ -0,0 +1,76 @@
...
+    {
+        public static async Task<string> DownloadAsync(this HttpClient httpClient,
+            string uri)
+        {
+            if (string.IsNullOrEmpty(uri))
+            {
+                throw new ArgumentException("Uri parameter must not be null or empty.");
+            }
+
+            if (string.IsNullOrEmpty(uri))
```
JunTaoLuo: Huh? There is no DestinationPath on this method...

JunTaoLuo: In fact is this being used anywhere at all?

src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
```diff
@@ -0,0 +1,308 @@
...
+            return await ExecuteCoreAsync();
+        }
+
+        internal FileInfo ResolveProjectFile(CommandOption projectOption)
+        {
+            string project;
+            if (projectOption.HasValue())
+            {
+                project = projectOption.Value();
+                project = Path.Combine(WorkingDirectory, project);
```
JunTaoLuo: Again, what about absolute paths?

src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
```diff
@@ -0,0 +1,308 @@
...
+                if (projects.Length == 0)
+                {
+                    Error.Write("No project files were found in the current directory. Either move to a new directory or provide the project explicitly");
+                }
+                if (projects.Length > 1)
+                {
+                    Error.Write("More than one project was found in this directory, either remove a duplicate or explicitly provide the project.");
+                }
+
+                project = projects[0];
```
JunTaoLuo: In the error cases, this would either be null or the first project, would recommend throwing instead and abort the whole operation and specify the option instead.

src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
```diff
@@ -0,0 +1,308 @@
...
+            }
+
+            return new FileInfo(project);
+        }
+
+        protected Project LoadProject(FileInfo projectFile)
+        {
+            var project = ProjectCollection.GlobalProjectCollection.LoadProject(
```
JunTaoLuo: Why not the Project(string path) constructor? Or Project.FromFile(string path)?

src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
```diff
@@ -0,0 +1,308 @@
...
+                projectFile.FullName,
+                globalProperties: null,
+                toolsVersion: null);
+            project.ReevaluateIfNecessary();
```
JunTaoLuo: Might not need this if the constructor is called. At least, I didn't have to.

src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
```diff
@@ -0,0 +1,308 @@
...
+        internal void AddServiceReference(
+            string tagName,
+            FileInfo projectFile,
+            string sourceFile,
+            string sourceUrl = null)
+        {
+            var project = LoadProject(projectFile);
+            var items = project.GetItems(tagName);
+            var item = items.FirstOrDefault((i) => {
+                return string.Equals(GetFullPath(i.EvaluatedInclude), GetFullPath(sourceFile), StringComparison.Ordinal);
```
JunTaoLuo: Would there ever be duplicates? Why FirstOrDefault instead of SingleOrDefault? In fact, I think duplicates warrant a warning.

src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
```diff
@@ -0,0 +1,308 @@
...
+            });
+
+            if (sourceUrl != null)
+            {
+                var urlMatch = items.SingleOrDefault(i => string.Equals(i.GetMetadataValue(SourceUrlAttrName), sourceUrl));
```
JunTaoLuo: so items.Any(...)?

src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
```diff
@@ -0,0 +1,308 @@
...
+            if (item == null)
+            {
+                var metadata = new Dictionary<string, string>();
+
+                if (!string.IsNullOrEmpty(sourceUrl))
+                {
+                    metadata[SourceUrlAttrName] = sourceUrl;
+                }
+
+                project.AddElementWithAttributes(tagName, sourceFile, metadata);
```
JunTaoLuo: This is only used in one place, and you only ever have one metadata. Simplify by inlining the logic here.

src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
```diff
@@ -0,0 +1,308 @@
...
+            }
+            else
+            {
+                Out.Write($"A reference to '{sourceFile}' already exists in '{project.FullPath}'.");
+            }
+        }
+
+        internal async Task DownloadToFileAsync(string url, string destinationPath, bool overwrite)
+        {
+            Application application;
```
JunTaoLuo: This is a weird pattern... same when it's used for working directory.

JunTaoLuo: Do you ever change the working Directory or need a different DownloadProvider (also not sure why the Provider suffix, should be something like DownloadAsync)? If not, just resolve working dir when needed and use a static DownloadAsync method on the base class here.

src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
```diff
@@ -0,0 +1,308 @@
...
+            var content = await application.DownloadProvider(url);
+            await WriteToFile(content, destinationPath, overwrite);
+        }
+
+        internal void EnsurePackagesInProject(FileInfo projectFile, CodeGenerator codeGenerator)
+        {
+            var packages = GetServicePackages(codeGenerator);
+            foreach (var (packageId, version) in packages)
+            {
+                var args = new string[] {
```
JunTaoLuo: Seems a bit overkill no? Just call AddItem on the project and set the version on the XML element. No need to spin up a new process.

src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
```diff
@@ -0,0 +1,308 @@
...
+
+            return result;
+        }
+
+        private async Task WriteToFile(Stream content, string destinationPath, bool overwrite)
+        {
+            var destinationExists = File.Exists(destinationPath);
+            if (destinationExists && !overwrite)
+            {
+                await Out.WriteLineAsync($"Not overwriting existing file '{destinationPath}'.");
```
JunTaoLuo: Why the WriteLineAsync now? You've used WriteLine in a bunch of other places.

src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
JunTaoLuo: No I mean, there's no check of up-to-dateness here. I have the same logic in the gRPC tool but I thought this is a little concerning. I just hope the users are good with their version control and notices and fixes any roll-back that might occur.

src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
JunTaoLuo: Yea, it's hard to guarantee that you're leaving the file system the same way you found it. It's fine if this is what we expect, but just wanted to call out the behaviour.

src/Tools/dotnet-openapi/src/Commands/RefreshCommand.cs
```diff
@@ -0,0 +1,69 @@
...
+            return 0;
+        }
+
+
+        private string FindReferenceFromUrl(FileInfo projectFile, string url)
+        {
+            var project = LoadProject(projectFile);
+            var openApiReferenceItems = project.GetItems(OpenApiReference);
+
+            foreach (ProjectItem item in openApiReferenceItems)
```
JunTaoLuo: If you are already using foreach, might as well as go all in on Linq and use .Where()

src/Tools/dotnet-openapi/src/Commands/RemoveCommand.cs
```diff
@@ -0,0 +1,73 @@
...
+                }
+                File.Delete(sourceFile);
+            }
+
+            return Task.FromResult(0);
+        }
+
+        private void RemoveServiceReference(string tagName, FileInfo projectFile, string sourceFile)
+        {
+            var project = LoadProject(projectFile);
```
JunTaoLuo: Curious if you have tried this with multiple arguments? I thought loading the same project multiple times will fail since the static global project collection can only contain one copy of the project.

src/Tools/dotnet-openapi/src/Commands/RemoveCommand.cs
```diff
@@ -0,0 +1,73 @@
...
+                var include = item.EvaluatedInclude;
+                if (string.Equals(include, sourceFile, StringComparison.Ordinal))
+                {
+                    project.RemoveItem(item);
+                    project.Save();
+                    return;
+                }
+            }
+
+            Out.Write("No openapi reference was found with the given source file");
```
JunTaoLuo: Might want to log which file you couldn't find. Otherwise you might just get a bunch of these logs which isn't quite useful.

src/Tools/dotnet-openapi/src/Commands/RemoveCommand.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        private void RemoveServiceReference(string tagName, FileInfo projectFile, string sourceFile)
+        {
+            var project = LoadProject(projectFile);
+            var openApiReferenceItems = project.GetItems(tagName);
+
+            foreach (ProjectItem item in openApiReferenceItems)
+            {
+                var include = item.EvaluatedInclude;
+                if (string.Equals(include, sourceFile, StringComparison.Ordinal))
```
JunTaoLuo: So you always use the file as the id of the reference to remove? In the gRPC version, I allowed either the file or the url as the id, which may be more intuitive. Let's say you called add url <foo_url>, it's probably more intuitive to call remove <foo_url>

JunTaoLuo (commented):
Some minor comments.

src/Tools/dotnet-openapi/src/Commands/AddURLCommand.cs
ryanbrandenburg: Because if the file already exists then it likely means that another service reference is using it. In fact, I think I should change `WriteToFile` so that we fail and quit in the case that we aren't overwriting and the file already exists.

src/Tools/dotnet-openapi/src/DownloadFile.cs
ryanbrandenburg: It is not! good catch.

src/Tools/dotnet-openapi/src/Commands/AddURLCommand.cs
dougbu: What about the `refresh` case?

ryanbrandenburg: In the refresh case we pass `overwrite: true`, in which case we don't fail.

src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
ryanbrandenburg: I get weird errors about reloading equivalent projects if I don't use this particular setup. I tried a couple permutations of `new Project(string path)` and `Project.FromFile(string path)` that didn't work, but if you want to link me to your code I can see why they're different.

src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
ryanbrandenburg: I disagree. This method was purposefully an extension even though it's only used once because it's basic functionality that one would expect to be on the class anyway. Plus it gives semantic meaning to an operation which looks a bit weird if you haven't investigated the API.

src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
ryanbrandenburg: DownloadProvider is given a different value in the tests so that we're not reliant on an http request as per https://github.com/aspnet/AspNetCore/pull/10015#discussion_r281457154.

src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
ryanbrandenburg: Not true. Depending on the version you're using there's stuff like 
```
    <PackageReference Include="NSwag.ApiDescription.Client" Version="12.0.0">
      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
      <PrivateAssets>all</PrivateAssets>
    </PackageReference>
```

which is required to behave properly.

src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
ryanbrandenburg: I decided not to make methods async just for these WriteLines, so in those methods that didn't otherwise go async yes, it's mixed. I did a quick scrub for WriteLines violating this idea and found a couple.

src/Tools/dotnet-openapi/src/Commands/RefreshCommand.cs
ryanbrandenburg: That seems a bit hefty for a Linq query.

src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
JunTaoLuo: I'm curious where else the projects are loaded. Is it because of the dotnet add process?

src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
JunTaoLuo: Fair. But a dictionary for one metadata is overkill.

src/Tools/dotnet-openapi/src/Commands/RefreshCommand.cs
JunTaoLuo: How is it hefty? optionApiReferenceItems.Single/First(i => string.Equals(i.GetMetadataValue(SourceUrlAttrName), url, StringComparison.Ordinal)) is a lot concise than the loop.

src/Tools/dotnet-openapi/src/Commands/RemoveCommand.cs
ryanbrandenburg: I have tried it with multiple arguments (now added a test). As I recall it works for me because of the way I'm loading the projects (instead of just using the constructor.)

src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
ryanbrandenburg: The scenario that caused it was that my tests new-up an Application object then operate on that, so when I for example do two adds it's operating on the same object, so it tries to Load the project twice. Obviously that's not a "real" scenario, but it's the simplest way to test things, and it hardens us against having trouble if some future refactor should end up having use run LoadProject in two different places within one command, plus to my knowledge there's nothing "wrong" about it.

src/Tools/dotnet-openapi/src/Commands/BaseCommand.cs
ryanbrandenburg: As soon as we have more than one CodeGenerator we'll have to start storing that metadata as well.

src/Tools/dotnet-openapi/src/Commands/RefreshCommand.cs
ryanbrandenburg: That's not the complete logic it would be replacing though. To do it your way it would have to be:
```c#
var ref = openApiReferenceItems.FirstOrDefault(i => string.Equals(i.GetMetadataValue(SourceUrlAttrName), url, StringComparison.Ordinal))?.EvaluatedInclude;

if(ref == null)
{
    throw new ...
}
else{
   return ref;
}
```

Your way is certainly shorter in terms of LOC, but I think mine is easier to follow.

JunTaoLuo: Hehe, null coalescing operators 
```
optionApiReferenceItems.FirstOrDefault(i => string.Equals(i.GetMetadataValue(SourceUrlAttrName), url, StringComparison.Ordinal)) ?? throw new ArgumentException(...);
```

Just playing code golf at this point. Not gonna hold up this PR for it :)

JunTaoLuo (approved):
Feels like this is in decent shape. Good to check in when all tests pass.

JunTaoLuo:
FYI there's been an update to how we plan on working with package versions. To ensure consistency between the global tools and VS tooling, there is a proposal to use a version file (the format looks like this: https://devdiv.visualstudio.com/DevDiv/_git/WebTools/pullrequest/193332?_a=files&path=%2Fsrc%2FSettings%2FPackageVersions_2099561.json) to unify the versions of the packages we add. The file is intended to be published at the fwlink: https://go.microsoft.com/fwlink/?linkid=2099561 but has not been published yet. The expected logic will be to use the versions listed in this file if we are successful in parsing it. Otherwise fall back to our current logic.

ryanbrandenburg:
Need to re-target this for preview9, going to just close and open another rather than confusing the history here.

PR #10021, opened 2019-05-06
Move AuthZ policy types back into Policy and rejigger AddAuthorization

HaoK:
Old AddAuthorization => AddAuthorizationCore() [In base AuthZ package)
New AddAuthorization() [In Policy package, calls AddPolicyEvaluator and AddAuthorizationCore)
base AuthZ now targets netstandard 2.0


src/Security/Authorization/Policy/ref/Microsoft.AspNetCore.Authorization.Policy.netcoreapp3.0.cs
```diff
@@ -1,11 +1,57 @@
...
+        public static Microsoft.AspNetCore.Builder.IEndpointConventionBuilder RequireAuthorization(this Microsoft.AspNetCore.Builder.IEndpointConventionBuilder builder) { throw null; }
+        public static Microsoft.AspNetCore.Builder.IEndpointConventionBuilder RequireAuthorization(this Microsoft.AspNetCore.Builder.IEndpointConventionBuilder builder, params Microsoft.AspNetCore.Authorization.IAuthorizeData[] authorizeData) { throw null; }
+        public static Microsoft.AspNetCore.Builder.IEndpointConventionBuilder RequireAuthorization(this Microsoft.AspNetCore.Builder.IEndpointConventionBuilder builder, params string[] policyNames) { throw null; }
+    }
+}
 namespace Microsoft.Extensions.DependencyInjection
 {
     public static partial class PolicyServiceCollectionExtensions
     {
-        [System.ObsoleteAttribute("AddAuthorizationPolicyEvaluator is obsolete and will be removed in a future release. Use AddAuthorization instead.")]
```
rynowak: Is this still obsolete? We'd tell people to use `AddAuthorization` right?

src/SignalR/common/Http.Connections/src/ConnectionsDependencyInjectionExtensions.cs
```diff
@@ -22,7 +22,7 @@ public static class ConnectionsDependencyInjectionExtensions
         public static IServiceCollection AddConnections(this IServiceCollection services)
         {
             services.AddRouting();
-            services.AddAuthorization();
+            services.AddAuthorizationPolicyEvaluator();
```
rynowak: hmmm. Seems like a lot of stuff was partially udpated.

rynowak (commented):
This looks roughly like what I expected to see 😆 - looks like there's still some stuff to update.

src/SignalR/common/Http.Connections/src/ConnectionsDependencyInjectionExtensions.cs
HaoK: Yeah I'll double check things closer tomorrow, I just reverted the last PR and did a rename, appears that resulted in things in a weird state

src/Security/Authorization/Core/src/Microsoft.AspNetCore.Authorization.csproj
```diff
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
```
SteveSandersonMS: Can we change this assembly to be `netstandard2.0`? That will be necessary for it to work on WebAssembly.

If you want that to be a separate change that I do later, that's OK, but it would be handy if you could check that your refactoring will be compatible with this change.

src/Security/Authorization/Core/src/Microsoft.AspNetCore.Authorization.csproj
```diff
@@ -13,13 +13,7 @@ Microsoft.AspNetCore.Authorization.AuthorizeAttribute</Description>
   </PropertyGroup>
 
   <ItemGroup>
-    <Compile Include="$(SharedSourceRoot)SecurityHelper\**\*.cs" />
-  </ItemGroup>
-
-  <ItemGroup>
-    <Reference Include="Microsoft.AspNetCore.Authentication.Abstractions" />
     <Reference Include="Microsoft.AspNetCore.Http.Abstractions" />
```
SteveSandersonMS: One of the prerequisites for this working for client-side Blazor is removing the `Microsoft.AspNetCore.Http.Abstractions` reference. Will that be possible as part of your refactoring? It was something I did in my alternate PR #9997.

If you don't want to do that here, then I can do it in a follow-up PR.

src/Security/Authorization/Policy/ref/Microsoft.AspNetCore.Authorization.Policy.netcoreapp3.0.cs
HaoK: I guess its no longer obsolete since it just adds the evaluator, if for whatever reason they only wanted that

src/Security/Authorization/Core/src/Microsoft.AspNetCore.Authorization.csproj
HaoK: Yep will make that change now

src/Security/Authorization/Core/src/Microsoft.AspNetCore.Authorization.csproj
HaoK: Yep there was no ref to http in 2.2, should be safe to remove

HaoK:
So looks like there as a new IAllowAnonymous type that was added to Http Abstractions, can we flip this around and have this type live in Auth Z and have Http.Abstractions depend on AuthZ?

https://github.com/aspnet/AspNetCore/blob/84da613d2c03b6f1c0fa3c01828923ec3415d525/src/Http/Http.Abstractions/src/IAllowAnonymous.cs

HaoK:
Tests look green finally, making this a real PR instead of Draft

src/Http/Http.Abstractions/src/Microsoft.AspNetCore.Http.Abstractions.csproj
```diff
@@ -20,6 +20,7 @@ Microsoft.AspNetCore.Http.HttpResponse</Description>
   </ItemGroup>
 
   <ItemGroup>
+    <Reference Include="Microsoft.AspNetCore.Authorization" />
```
rynowak: This makes me go hmmm. Is this because of the attributes? What in this project is using them?

HaoK: Yeah we moved IAuthorizeData and IAllowAnonymous back to Authorization

rynowak: What depends on those types in `Http.Abstractions`? Was it like this before? 

@davidfowl @Tratcher do we want this dependency here?

HaoK: I think it comes from routing: https://github.com/aspnet/AspNetCore/blob/23efa15112ae2e7c9bf26e75f7173463b9d9a056/src/Http/Routing/src/EndpointMiddleware.cs#L45

Tratcher: I'd want to see the graph. Dependencies don't matter as much as the used to, but we still need to avoid a rats nest.

HaoK: AuthZ doesn't bring much, and nothing new, it just adds Options/Logging.Abstractions which are already referenced

rynowak: That would dictate routing referencing auth though, not http.abstractions

HaoK: I'll try moving the ref to routing

Tratcher: Much better

davidfowl:
> So looks like there as a new IAllowAnonymous type that was added to Http Abstractions, can we flip this around and have this type live in Auth Z and have Http.Abstractions depend on AuthZ?

I have to look in more detail but at first glance this doesn’t seem like a good idea. I’d like to keep the HttpAbstractions assembly near the bottom of the dependency graph

davidfowl (commented):
Feels like there should be an auth abstractions assembly no?

Tratcher:
@davidfowl why bother with new assemblies when everything is in the shared runtime together?

davidfowl:
> @davidfowl why bother with new assemblies when everything is in the shared runtime together?

Future proofing. I've heard this sentiment from @rynowak as well but we should *not* take how we ship today as a sign that we want to remove assembly boundaries unless it makes sense to do so from a layering POV. As far as I see it, we're not going to a place where we have a single assembly even though we ship that way for the most part (when using the shared framework). There are many benefits of having actual boundaries generally:
- It helps to validate dependency boundaries 
- It might make assembly trimming easier (when everything can't just reference everything)
- It's part of future proofing (we ship a shared framework and version everything together right now but we may not always ship this way).

TL;DR I don't want us to think we can't add new assemblies just because we're shipping a shared framework. If it's the "right thing to do" then lets just do it.

rynowak:
> TL;DR I don't want us to think we can't add new assemblies just because we're shipping a shared framework. If it's the "right thing to do" then lets just do it.

The problem is that *the right thing to do* is usually in the context of reacting to a single change or in the context of *today's requirements*. It makes it hard to judge. To a certain extent in this area we have to try and anticipate our future requirements OR just try and decouple things to the max.

---

Today we're responding to the requirement of needing to decouple the auth attributes from HTTP. 

I see a few different options:
- Create an *Microsoft.AspNetCore.Authorization.Metadata* assembly for these attributes to live in
- Create an *Microsoft.AspNetCore.Metadata* assembly for any metadata attributes that belong at the bottom of our layering like CORS or Auth. These types are just metadata and use simple types without coupling to an implementation
- Get rid of the dependency between routing and auth, and use an analyzer for this

*I didn't call these `.Abstractions` because I think that's a different thing.*

----

Option 1 or 2 both go after the idea that we want to make loosely coupled systems that can interoperate with declarative metadata. Put another way, there's no conflict with having multiple systems that care about the same *data*.  

Option 1 means that we'd do this as a tactical thing for solving this particular problem, and we could do it again in some other area if we choose. 

Option 2 means that we'd *want* to make this choice in other areas - and we'd start designing for it. We'd end up with a single assembly for our top-level metadata rather than a few different ones.

Option 3 would be to punt on this. The only reason that auth and routing are together today is because we wanted some kind of safeguard in place for our security-related scenarios.

----

I've already discussed option 2 in the past with @davidfowl so I know that he's not sold on it 😆

halter73:
> Create an Microsoft.AspNetCore.Metadata assembly

I kinda like the sound of option 2. At least that makes it extremely obvious where any metadata attributes should go.

HaoK:
Tests are all green now, should I merge and file a tracking issue for the final home of where these attributes should live?

davidfowl:
Nah I want the new assembly (that’s basically what this work is). We should mimic the features assembly here and stick cross cutting metadata in a single assembly with no deps

HaoK:
So create `Microsoft.AspNetCore.Authorization.Metadata ` and put `IAuthorizeData` and `IAllowAnonymous` there?  Or option 2 and create a top level `Microsoft.AspNetCore.MetaData` ?

rynowak:
`Microsoft.AspNetCore.Metadata`  IMO.

davidfowl:
Yes we’re going to put Cors there as well

HaoK:
Where should this metadata project live in our src...  `Features/Metadata`?  Or should it be a new top level folder?

rynowak:
I would expect `Metadata` in https://github.com/aspnet/AspNetCore/tree/master/src/Http

dougbu:
> base AuthZ now targets netstandard 2.0

What's driving this part of the change?

rynowak:
> What's driving this part of the change?

We need to use it in Blazor.

natemcmaster:
Remind me again…why doesn’t Blazor work for APIs that target netcoreapp3.0?

rynowak:
WebAssembly is netstandard2.0. **YES** it is.

HaoK:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=191784

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/2329
https://github.com/aspnet/AspNetCore-Internal/issues/2465

src/Mvc/Mvc.Formatters.Json/src/Properties/AssemblyInfo.cs
```diff
@@ -1,8 +0,0 @@
-// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
-using System.Runtime.CompilerServices;
-using Microsoft.AspNetCore.Mvc;
-
-[assembly: TypeForwardedTo(typeof(JsonResult))]
```
rynowak: Why is this being removed in this PR?

src/Security/Authorization/Core/src/Properties/AssemblyInfo.cs
```diff
@@ -1,9 +0,0 @@
-// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
-using System.Runtime.CompilerServices;
-using Microsoft.AspNetCore.Authorization;
-
-// Microsoft.AspNetCore.Http.Abstractions
-[assembly: TypeForwardedTo(typeof(IAuthorizeData))]
-[assembly: TypeForwardedTo(typeof(IAllowAnonymous))]
```
rynowak: Why remove these? We should still type-forward these so it's not a breaking change.

src/Mvc/Mvc.Formatters.Json/src/Properties/AssemblyInfo.cs
HaoK: bad rebase merge I think, i'll fix now

rynowak (approved):

Tratcher (approved):

SteveSandersonMS (changes requested):
Does this make policy evaluation runnable on `netstandard2.0`? It looks like, unlike [the proposal here](https://github.com/aspnet/AspNetCore/pull/9997), the policy evaluation is still in a `netcoreapp3.0` assembly after this change, so wouldn't be usable in WebAssembly.

In case it's unclear, we do want client-side Blazor to have the same notion of policies as server-side ASP.NET Core. The only part that would be different is that, on the client, there is no `HttpContext` hence pulling out the non-`HttpContext`-attached parts into an underlying layer in this [proposal](https://github.com/aspnet/AspNetCore/pull/9997).

I don't mind doing a follow-up PR to split the policy implementation in two, but that only makes sense if we have the right new assembly to put the shared (`netstandard2.0`+`netcoreapp3.0`) code, and it doesn't look like the new `Metadata` assembly is the right place for it given the description in the PR.

Marking as "request changes" just to be sure we talk about this before this is merged. Hope that's OK!

HaoK:
@SteveSandersonMS you don't actually need a policy evaluator, that's the thing that binds Http authentication to authorization.  For blazor, you should be able to directly use the AuthZ core package that is targeting netstandard2.0 to evaluate policies by using `IAuthorizationService`

HaoK:
The naming isn't great, but basically the policy package was added as a way for SignalR and MVC to reuse similar logic that was doing http authentication and authorization, its really closer to `Authorization.Http` conceptually.  If all you need is authorization and policies, that is all possible via the extension methods like:

https://github.com/aspnet/AspNetCore/blob/master/src/Security/Authorization/Core/src/AuthorizationServiceExtensions.cs#L103

SteveSandersonMS (approved):

SteveSandersonMS:
Thanks for clarifying, @HaoK! That makes sense now.

You're right - I just tried out the code in this PR and was able to define and run policies under WebAssembly.

Issue #6013, opened 2018-10-16
Have url_rewrite middleware resolve server variables from IIS when running ANCM in-proc


muratg:

mderriey:
Hi @muratg, @shirhatti and @jkotalik 👋

I'm keen to give this one a shot, but I'd appreciate some pointers on how to do this.
How does IIS expose server variables?

Thanks 🙇‍♂️

mderriey:
I've found [`IServerVariableFeature`](https://github.com/aspnet/AspNetCore/blob/3477daf3c4f530dff80f197e0642cb39a26fb07b/src/Servers/IIS/IIS/src/IServerVariableFeature.cs) and the associated [`GetIISServerVariable`](https://github.com/aspnet/AspNetCore/blob/3477daf3c4f530dff80f197e0642cb39a26fb07b/src/Servers/IIS/IIS/src/HttpContextExtensions.cs) extension method.

The current thinking is to create a new `PatternSegment` that gets values from the `IServerVariableFeature` while allowing fallback on the existing implementation if we're not using in-proc hosting.

Please let me know if that sounds about the direction you'd like this to take.

Cheers 🙏

jkotalik:
Yeah, that would be the right direction to take this!

mderriey:
I'm surprised the issue isn't closed given the associated PR was merged and the PR description mentions "Addresses #<number>".

Is that because I don't have permissions to close the issue?

anurse:
"Addresses" doesn't close the attached issue, you have to use "Closes" or "Fixes" :)

anurse:
But yeah, it's also because you don't have the permissions :). Thanks for this contribution!

anurse:
Fixed in https://github.com/aspnet/AspNetCore/pull/10022 (https://github.com/aspnet/AspNetCore/commit/4a2bd2adbf74b316f01eb062de3e83261562801d)

PR #10022, opened 2019-05-07
Source server variables from IIS when running ANCM in-proc

mderriey:
- Source server variables from IIS when running ANCM in-proc

Addresses #6013

### DIFF
--- a/src/Middleware/Rewrite/ref/Microsoft.AspNetCore.Rewrite.csproj
+++ b/src/Middleware/Rewrite/ref/Microsoft.AspNetCore.Rewrite.csproj
@@ -7,6 +7,7 @@
     <Compile Include="Microsoft.AspNetCore.Rewrite.netcoreapp3.0.cs" />
     <Reference Include="Microsoft.AspNetCore.Hosting.Abstractions"  />
     <Reference Include="Microsoft.AspNetCore.Http.Extensions"  />
+    <Reference Include="Microsoft.AspNetCore.Server.IIS"  />
     <Reference Include="Microsoft.Extensions.Configuration.Abstractions"  />
     <Reference Include="Microsoft.Extensions.FileProviders.Abstractions"  />
     <Reference Include="Microsoft.Extensions.Logging.Abstractions"  />
--- a/src/Middleware/Rewrite/src/IISUrlRewriteOptionsExtensions.cs
+++ b/src/Middleware/Rewrite/src/IISUrlRewriteOptionsExtensions.cs
@@ -19,8 +19,8 @@ namespace Microsoft.AspNetCore.Rewrite
         /// <param name="options">The <see cref="RewriteOptions"/></param>
         /// <param name="fileProvider">The <see cref="IFileProvider"/> </param>
         /// <param name="filePath">The path to the file containing UrlRewrite rules.</param>
-        /// <param name="useNativeIISServerVariables">If the application uses the in-process hosting model, server variables can be sourced natively from IIS. It is disabled by default.</param>
-        public static RewriteOptions AddIISUrlRewrite(this RewriteOptions options, IFileProvider fileProvider, string filePath, bool useNativeIISServerVariables = false)
+        /// <param name="useNativeIISServerVariables">Server variables are sourced natively from IIS if the application uses the in-process hosting model; use <c>false</c> to disable that behavior</param>
+        public static RewriteOptions AddIISUrlRewrite(this RewriteOptions options, IFileProvider fileProvider, string filePath, bool useNativeIISServerVariables = true)
         {
             if (options == null)
             {
@@ -45,8 +45,8 @@ namespace Microsoft.AspNetCore.Rewrite
         /// </summary>
         /// <param name="options">The <see cref="RewriteOptions"/></param>
         /// <param name="reader">The text reader stream.</param>
-        /// <param name="useNativeIISServerVariables">If the application uses the in-process hosting model, server variables can be sourced natively from IIS. It is disabled by default.</param>
-        public static RewriteOptions AddIISUrlRewrite(this RewriteOptions options, TextReader reader, bool useNativeIISServerVariables = false)
+        /// <param name="useNativeIISServerVariables">Server variables are sourced natively from IIS if the application uses the in-process hosting model; use <c>false</c> to disable that behavior</param>
+        public static RewriteOptions AddIISUrlRewrite(this RewriteOptions options, TextReader reader, bool useNativeIISServerVariables = true)
         {
             if (options == null)
             {
--- a/src/Middleware/Rewrite/src/IISUrlRewriteOptionsExtensions.cs
+++ b/src/Middleware/Rewrite/src/IISUrlRewriteOptionsExtensions.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.IO;
+using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Rewrite.Internal.IISUrlRewrite;
 using Microsoft.Extensions.FileProviders;
 
@@ -19,8 +20,8 @@ namespace Microsoft.AspNetCore.Rewrite
         /// <param name="options">The <see cref="RewriteOptions"/></param>
         /// <param name="fileProvider">The <see cref="IFileProvider"/> </param>
         /// <param name="filePath">The path to the file containing UrlRewrite rules.</param>
-        /// <param name="useNativeIISServerVariables">Server variables are sourced natively from IIS if the application uses the in-process hosting model; use <c>false</c> to disable that behavior</param>
-        public static RewriteOptions AddIISUrlRewrite(this RewriteOptions options, IFileProvider fileProvider, string filePath, bool useNativeIISServerVariables = true)
+        /// <param name="alwaysUseManagedServerVariables">Server variables are by default sourced from the server if it supports the <see cref="IServerVariablesFeature"/> feature. Use <c>true</c> to disable that behavior</param>
+        public static RewriteOptions AddIISUrlRewrite(this RewriteOptions options, IFileProvider fileProvider, string filePath, bool alwaysUseManagedServerVariables = false)
         {
             if (options == null)
             {
@@ -36,7 +37,7 @@ namespace Microsoft.AspNetCore.Rewrite
 
             using (var stream = file.CreateReadStream())
             {
-                return AddIISUrlRewrite(options, new StreamReader(stream), useNativeIISServerVariables);
+                return AddIISUrlRewrite(options, new StreamReader(stream), alwaysUseManagedServerVariables);
             }
         }
 
@@ -45,8 +46,8 @@ namespace Microsoft.AspNetCore.Rewrite
         /// </summary>
         /// <param name="options">The <see cref="RewriteOptions"/></param>
         /// <param name="reader">The text reader stream.</param>
-        /// <param name="useNativeIISServerVariables">Server variables are sourced natively from IIS if the application uses the in-process hosting model; use <c>false</c> to disable that behavior</param>
-        public static RewriteOptions AddIISUrlRewrite(this RewriteOptions options, TextReader reader, bool useNativeIISServerVariables = true)
+        /// <param name="alwaysUseManagedServerVariables">Server variables are by default sourced from the server if it supports the <see cref="IServerVariablesFeature"/> feature. Use <c>true</c> to disable that behavior</param>
+        public static RewriteOptions AddIISUrlRewrite(this RewriteOptions options, TextReader reader, bool alwaysUseManagedServerVariables = false)
         {
             if (options == null)
             {
@@ -58,7 +59,7 @@ namespace Microsoft.AspNetCore.Rewrite
                 throw new ArgumentException(nameof(reader));
             }
 
-            var rules = new UrlRewriteFileParser().Parse(reader, useNativeIISServerVariables);
+            var rules = new UrlRewriteFileParser().Parse(reader, alwaysUseManagedServerVariables);
 
             foreach (var rule in rules)
             {
--- a/src/Middleware/Rewrite/src/IISUrlRewriteOptionsExtensions.cs
+++ b/src/Middleware/Rewrite/src/IISUrlRewriteOptionsExtensions.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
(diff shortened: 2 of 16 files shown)


mderriey:
This is the first version to get something out there before asking too many questions.

Right now it doesn't honor the `UriMatchPart` parameter and returns IIS variable as-is.

Let me know if we'd like to support that or if you see anything else.

src/Middleware/Rewrite/ref/Microsoft.AspNetCore.Rewrite.csproj
```diff
@@ -7,6 +7,7 @@
     <Compile Include="Microsoft.AspNetCore.Rewrite.netcoreapp3.0.cs" />
     <Reference Include="Microsoft.AspNetCore.Hosting.Abstractions"  />
     <Reference Include="Microsoft.AspNetCore.Http.Extensions"  />
+    <Reference Include="Microsoft.AspNetCore.Server.IIS"  />
```
anurse: Hrm, it's unfortunate that we need to reference `.Server.IIS` just to get `IServerVariablesFeature`. Could we consider relocating that feature interface? It's very simple.

src/Middleware/Rewrite/test/IISUrlRewrite/ServerVariableTests.cs
```diff
@@ -53,7 +102,24 @@ private RewriteContext CreateTestHttpContext()
...
+        }
+
+        private RewriteContext CreateTestRewriteContext(HttpContext context = null)
+        {
+            return new RewriteContext { HttpContext = context ?? CreateTestHttpContext() };
+        }
+
+        private IServerVariablesFeature CreateTestServerVariablesFeature(Dictionary<string, string> variables)
+        {
+            var mockServerVariablesFeature = new Mock<IServerVariablesFeature>();
```
anurse: I'd prefer just creating a concrete `TestServerVariablesFeature` class that takes a dictionary and exposes it via the accessor. It's trivial and means we don't need more `Mock`s. In general, I lean towards using actual types rather than more dynamic code gen :)

```csharp
public class TestServerVariablesFeature: IServerVariablesFeature
{
    private readonly Dictionary<string, string> _variables;

    public TestServerVariablesFeature(IDictionary<string, string> variables)
    {
        _variables = variables;
    }

    public string this[string variableName]
    {
        get => _variables[variableName];
        set => _variables[variableName] = value;
    }
}
```

analogrelay (commented):
Looks OK to me. Minor suggestion around using a real type instead of a mock. Also not sure about adding the dependency. Now that it's one big happy `<FrameworkReference>`, it's probably not a big deal, but having to depend on `Servers.IIS` from `Rewrite` feels a little off.

analogrelay (commented):
Looks OK to me. Minor suggestion around using a real type instead of a mock. Also not sure about adding the dependency. Now that it's one big happy `<FrameworkReference>`, it's probably not a big deal, but having to depend on `Servers.IIS` from `Rewrite` feels a little off.

src/Middleware/Rewrite/ref/Microsoft.AspNetCore.Rewrite.csproj
jkotalik: I can go either way on this change. In one hand, the ServerVariable feature is pretty much purely an IIS feature, so it makes sense to live in that namespace. On the other hand, the feature is simple. 

I don't think it's a huge deal having Rewrite depend on IIS.

src/Middleware/Rewrite/src/Internal/PatternSegments/IISServerVariableSegment.cs
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Server.IIS;
+
+namespace Microsoft.AspNetCore.Rewrite.Internal.PatternSegments
+{
+    public class IISServerVariableSegment : PatternSegment
```
jkotalik: nit: let's make this internal.

src/Middleware/Rewrite/src/Internal/IISUrlRewrite/ServerVariables.cs
```diff
@@ -19,54 +19,79 @@ public static class ServerVariables
         /// <returns>The matching <see cref="PatternSegment"/></returns>
         public static PatternSegment FindServerVariable(string serverVariable, ParserContext context, UriMatchPart uriMatchPart)
         {
+            Func<PatternSegment> fallback = default;
```
jkotalik: This is a small a breaking change if one of these variables return a different value than the fallback did and there is no way to turn it off. We may need to add an option or something to not use IISServerVariables.

src/Middleware/Rewrite/ref/Microsoft.AspNetCore.Rewrite.csproj
mderriey: This your call. Would we relocate the interface to a common project and then have both `.Server.IIS` and `Rewrite` depend on it?

I'll leave it this way for now, but feel free to drop another comment here if you decide you don't want the current dependency.

Tratcher: How old is the feature? Moving it is likely breaking.

The dependencies aren't really a problem anymore for anything in the shared framework.

src/Middleware/Rewrite/src/Internal/IISUrlRewrite/ServerVariables.cs
mderriey: True.

Going up the layers, I think we can add an optional boolean parameter to the [`AddIISUrlRewrite`](https://github.com/aspnet/AspNetCore/blob/b0659f660b6e8151a10da1c6c16310ec0101c380/src/Middleware/Rewrite/src/IISUrlRewriteOptionsExtensions.cs) extension methods and let it flow down to `ServerVariables.FindServerVariable`.

mderriey: Rook a shot at introducing a way to opt out in [`2e90d8f3`](https://github.com/aspnet/AspNetCore/commit/2e90d8f3c9b71834196f854ca3f6d64ad48ef7fd)

src/Middleware/Rewrite/src/IISUrlRewriteOptionsExtensions.cs
```diff
@@ -19,7 +19,8 @@ public static class IISUrlRewriteOptionsExtensions
         /// <param name="options">The <see cref="RewriteOptions"/></param>
         /// <param name="fileProvider">The <see cref="IFileProvider"/> </param>
         /// <param name="filePath">The path to the file containing UrlRewrite rules.</param>
-        public static RewriteOptions AddIISUrlRewrite(this RewriteOptions options, IFileProvider fileProvider, string filePath)
+        /// <param name="useNativeIISServerVariables">If the application uses the in-process hosting model, server variables can be sourced natively from IIS. It is disabled by default.</param>
+        public static RewriteOptions AddIISUrlRewrite(this RewriteOptions options, IFileProvider fileProvider, string filePath, bool useNativeIISServerVariables = false)
```
jkotalik: I'd invert the bool here and have this on by default. I'm fine with this being a breaking change, however it would be bad if there wasn't a work around.

src/Middleware/Rewrite/ref/Microsoft.AspNetCore.Rewrite.csproj
anurse: I agree, I don't think the dependency issue is blocking. I just wanted to raise it. The shared framework does make this point mostly moot. There's no scenario where Rewrite is available and Server.IIS isn't.

Ok, let's go ahead then.

src/Middleware/Rewrite/src/IISUrlRewriteOptionsExtensions.cs
mderriey: Yeah, that makes sense. Did that in [`9fedfd53`](https://github.com/aspnet/AspNetCore/pull/10022/commits/9fedfd5330a8261cc3e2d970cdf9f32f7c3b11f0).

I was wondering whether to change the default of the parameter or rename it to `disableIISServerVariables`.

I ended up keeping the original name, change the default value and update the associated comment, but let me know if you think renaming the parameter would make more sense to you.

davidfowl:
I think we should move this feature to the features assembly in Http/*

src/Middleware/Rewrite/ref/Microsoft.AspNetCore.Rewrite.csproj
davidfowl: Lets move it and type forward.

davidfowl: It doesn't need to block this PR tho

mderriey:
Hey @davidfowl thanks for jumping in.

What feature are you referring to exactly?

src/Middleware/Rewrite/ref/Microsoft.AspNetCore.Rewrite.csproj
Tratcher: @davidfowl we're going to have trouble expanding the features model if we insist that every feature interface must go into a central dll. Also consider that feature interfaces can bring with them additional types specific to their scenario (e.g. IResponseCookiesFeature, IResponseCookies, CookieOptions).  Cases like this where the feature is component specific warrant locating the feature interface closer to that component. What about adding new packages like IIS.Abstractions? That gives you some separation but still give us a growth model. 

A more extreme example would be if a component that wasn't in the shared runtime wanted to expose a feature interface. It wouldn't make sense to add a feature to Http.Features for a component we wouldn't even bundle with the runtime.

davidfowl: I understand the concern and I share it. I don't think all features should go into there (for e.g. I'm not suggesting we move all of the kestrel internal features) but server variables are generic enough to imagine supporting them outside of IIS and dependency free enough that I think it's reasonable to move it to the features assembly. 

We don't have ANIs anymore...

Tratcher:
@mderriy he's talking about moving IServerVariablesFeature so we can avoid the IIS server dependency. We'd also need to move GetIISServerVariable, or not use it in this code path.

anurse:
That would be a breaking change, but I think I'm OK with it. It seems reasonable. Server Variables aren't actually an IIS-specific concept anyway. They date back to [CGI](https://en.wikipedia.org/wiki/Common_Gateway_Interface) so it seems reasonable that other servers may want to provide them.

If we do this, I'd consider renaming the extension method to `GetServerVariable` instead of `GetIISServerVariable`.

davidfowl:
> That would be a breaking change, but I think I'm OK with it. It seems reasonable. Server Variables aren't actually an IIS-specific concept anyway. They date back to CGI so it seems reasonable that other servers may want to provide them.

Exactly!

> If we do this, I'd consider renaming the extension method to GetServerVariable instead of GetIISServerVariable.

Right! I think that's what we want to do for 3.0. Lets not hold up this PR as a result tho. File an issue and refactor this since most of these changes are related to the rewrite middleware.

mderriey:
Thanks for the explanation/context.

I'd be happy to try and implement the final solution to expand my knowledge of the codebase.

Could you please tag me in the resulting issue when it's created?

anurse:
Filing a bug now :)

anurse:
https://github.com/aspnet/AspNetCore/issues/10167

Feel free to move discussion on the subject of moving/renaming the feature to that issue.

mderriey:
Great, thanks. I'll wait until this PR is merged to start the move.

Thanks again.

anurse:
Code Check failed but I can't get AzDO to load. Usually this means you've got reference assemblies that are out of date. See https://github.com/aspnet/AspNetCore/blob/master/docs/ReferenceAssemblies.md for instructions on updating.

anurse:
Got it to load, yep, just need to update the refasms.

```
error: Generated code is not up to date in src/Middleware/Rewrite/ref/Microsoft.AspNetCore.Rewrite.netcoreapp3.0.cs. You might need to regenerate the reference assemblies or project list (see docs/ReferenceAssemblies.md and docs/ReferenceResolution.md)
```

src/Middleware/Rewrite/src/IISUrlRewriteOptionsExtensions.cs
```diff
@@ -19,7 +19,8 @@ public static class IISUrlRewriteOptionsExtensions
         /// <param name="options">The <see cref="RewriteOptions"/></param>
         /// <param name="fileProvider">The <see cref="IFileProvider"/> </param>
         /// <param name="filePath">The path to the file containing UrlRewrite rules.</param>
-        public static RewriteOptions AddIISUrlRewrite(this RewriteOptions options, IFileProvider fileProvider, string filePath)
+        /// <param name="useNativeIISServerVariables">Server variables are sourced natively from IIS if the application uses the in-process hosting model; use <c>false</c> to disable that behavior</param>
+        public static RewriteOptions AddIISUrlRewrite(this RewriteOptions options, IFileProvider fileProvider, string filePath, bool useNativeIISServerVariables = true)
```
jkotalik:
```suggestion
        public static RewriteOptions AddIISUrlRewrite(this RewriteOptions options, IFileProvider fileProvider, string filePath, bool alwaysUseManagedServerVariables = false)
```

jkotalik: Generally we prefer bool defaults to be false.

src/Middleware/Rewrite/test/IISUrlRewrite/TestServerVariablesFeature.cs
```diff
@@ -0,0 +1,21 @@
+using System.Collections.Generic;
```
jkotalik: nit copyright.

jkotalik (approved):
Nits at this point 😄 

Looks good, we can handle moving the feature to Http in a second PR. 

I also want to do some E2E verification that this works when running inprocess.

src/Middleware/Rewrite/src/IISUrlRewriteOptionsExtensions.cs
mderriey: No worries, thanks for pointing that out. I pushed extra commits to address that.

src/Middleware/Rewrite/test/IISUrlRewrite/TestServerVariablesFeature.cs
mderriey: Fixed 😇

jkotalik:
There are still test failures and I believe you need to re update references. That's the reason we are working on making things internal.

mderriey:
Hey @jkotalik, sorry about that. I forgot to invert the boolean values after we changed the meaning of the parameter. I also regenerated reference assemblies.

Sorry that this one drags out so much, I keep forgetting about the ref assemblies.

Thanks for your time, I'll keep an eye on the outcomes of the builds.

jkotalik:
Thanks @mderriey

PR #10043, opened 2019-05-07
Synchronize writing to connection and aborting connection

### LINKED ISSUE #6701: "Writing is not allowed after writer was completed", with SignalR 1.1
### Describe the bug
I'm receiving "Writing is not allowed after writer was completed" on the latest release of SignalR with .NET Core 2.2. We previously used 1.0.0-rc1-final without any issues, and all of a sudden started receiving this issue with 1.1 on the client. We also receive a lot of "Task was cancelled" from the RedisHubLifetimeManager. Attaching screenshots below.

We ran SignalR 1.1 on the backend with 1.0.0-rc1-final on the client without any issues, but when we upgraded our client to 1.1, we started receiving this in our logs.

Another change we did while upgrading was to add "skipNegotiate: true" on the client, as well as forcing websockets. Not sure if this is related. I can't reproduce the error locally, since it doesn't happen often and seems related to only a few of our clients.

Also, the server seems to go down quite often after the upgrade. Can't seem to find any reason for this, but it was stable before we upped our clients.

### To Reproduce
Can't reproduce the error locally, we just see it in our production environment logs. 

### Expected behavior
Expected it to be gone with 1.1.0, since I saw previous reports of the problem on RC versions.

### Screenshots

Screenshot of writing is not allowed message:
image

Screenshot of task cancelled exception:
image

### Additional context
`dotnet --info` output:
image 
This is on our production environment. Locally we have the 2.2 runtime.

EDIT: If this issue is misplaced or inproper, feel free to close it. I didn't know where to turn.

BrennanConroy:
Fixes https://github.com/aspnet/AspNetCore/issues/6701

Verified perf didn't change.

### DIFF
--- a/src/SignalR/server/Core/src/HubConnectionContext.cs
+++ b/src/SignalR/server/Core/src/HubConnectionContext.cs
@@ -146,6 +146,7 @@ namespace Microsoft.AspNetCore.SignalR
 
             if (_connectionAborted)
             {
+                _writeLock.Release();
                 return default;
             }
 
@@ -181,6 +182,7 @@ namespace Microsoft.AspNetCore.SignalR
 
             if (_connectionAborted)
             {
+                _writeLock.Release();
                 return default;
             }
 
--- a/src/SignalR/server/Core/src/HubConnectionContext.cs
+++ b/src/SignalR/server/Core/src/HubConnectionContext.cs
@@ -41,7 +41,7 @@ namespace Microsoft.AspNetCore.SignalR
         private bool _receivedMessageThisInterval = false;
         private ReadOnlyMemory<byte> _cachedPingMessage;
         private bool _clientTimeoutActive;
-        private bool _connectedAborted;
+        private bool _connectionAborted;
         private int _streamBufferCapacity;
 
         /// <summary>
@@ -138,7 +138,7 @@ namespace Microsoft.AspNetCore.SignalR
 
         public virtual ValueTask WriteAsync(HubMessage message, CancellationToken cancellationToken = default)
         {
-            if (_connectedAborted)
+            if (_connectionAborted)
             {
                 return default;
             }
@@ -147,6 +147,11 @@ namespace Microsoft.AspNetCore.SignalR
             if (!_writeLock.Wait(0))
             {
                 return new ValueTask(WriteSlowAsync(message));
+            }
+
+            if (_connectionAborted)
+            {
+                return default;
             }
 
             // This method should never throw synchronously
@@ -173,7 +178,7 @@ namespace Microsoft.AspNetCore.SignalR
         /// <returns></returns>
         public virtual ValueTask WriteAsync(SerializedHubMessage message, CancellationToken cancellationToken = default)
         {
-            if (_connectedAborted)
+            if (_connectionAborted)
             {
                 return default;
             }
@@ -182,6 +187,11 @@ namespace Microsoft.AspNetCore.SignalR
             if (!_writeLock.Wait(0))
             {
                 return new ValueTask(WriteSlowAsync(message));
+            }
+
+            if (_connectionAborted)
+            {
+                return default;
             }
 
             // This method should never throw synchronously
@@ -259,10 +269,15 @@ namespace Microsoft.AspNetCore.SignalR
 
         private async Task WriteSlowAsync(HubMessage message)
         {
+            // Failed to get the lock immediately when entering WriteAsync so await until it is available
             await _writeLock.WaitAsync();
+
             try
             {
-                // Failed to get the lock immediately when entering WriteAsync so await until it is available
+                if (_connectionAborted)
+                {
+                    return;
+                }
 
                 await WriteCore(message);
             }
@@ -279,10 +294,15 @@ namespace Microsoft.AspNetCore.SignalR
 
         private async Task WriteSlowAsync(SerializedHubMessage message)
         {
+            // Failed to get the lock immediately when entering WriteAsync so await until it is available
+            await _writeLock.WaitAsync();
+
             try
             {
-                // Failed to get the lock immediately when entering WriteAsync so await until it is available
-                await _writeLock.WaitAsync();
+                if (_connectionAborted)
+                {
+                    return;
+                }
 
                 await WriteCore(message);
             }
@@ -313,6 +333,11 @@ namespace Microsoft.AspNetCore.SignalR
         {
             try
             {
+                if (_connectionAborted)
+                {
+                    return;
+                }
+
                 await _connectionContext.Transport.Output.WriteAsync(_cachedPingMessage);
 
                 Log.SentPing(_logger);
@@ -320,6 +345,7 @@ namespace Microsoft.AspNetCore.SignalR
             catch (Exception ex)
             {
                 Log.FailedWritingMessage(_logger, ex);
+                Abort();
             }
             finally
             {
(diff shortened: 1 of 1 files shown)


src/SignalR/server/Core/src/HubConnectionContext.cs
```diff
@@ -533,11 +557,15 @@ private void CheckClientTimeout()
             }
         }
 
-        private static void AbortConnection(object state)
+        private static async void AbortConnection(object state)
```
BrennanConroy: Yes I know... open to suggestions

davidfowl: Async void.....

src/SignalR/server/Core/src/HubConnectionContext.cs
```diff
@@ -533,11 +557,15 @@ private void CheckClientTimeout()
...
         {
             var connection = (HubConnectionContext)state;
+
+            await connection._writeLock.WaitAsync();
+
             try
             {
+                connection.Input.CancelPendingRead();
```
davidfowl: This doesn’t need to be synchronized

davidfowl:
Help me understand this change? The description of the fix is a bit light on explanation.

BrennanConroy:
Servers could get error logs saying "Writing is not allowed after writer was completed" if the connection was closed while writing, especially when writing the ping message. This synchronizes the abort with the writing so we wont see error logs when writing would be expected to fail.

src/SignalR/server/Core/src/HubConnectionContext.cs
halter73: More "async avoid", amirite?

I'll show myself out 🤦

src/SignalR/server/Core/src/HubConnectionContext.cs
```diff
@@ -355,19 +380,19 @@ private async Task WriteHandshakeResponseAsync(HandshakeResponseMessage message)
...
                 return;
             }
 
-            _connectedAborted = true;
-
             Input.CancelPendingRead();
 
             // We fire and forget since this can trigger user code to run
-            ThreadPool.QueueUserWorkItem(_abortedCallback, this);
+            _ = Task.Factory.StartNew(AbortConnection, this, cancellationToken: default, TaskCreationOptions.DenyChildAttach, TaskScheduler.Default);
```
davidfowl: Why?

JamesNK:
Unit test?

src/SignalR/server/Core/src/HubConnectionContext.cs
```diff
@@ -355,19 +380,19 @@ private async Task WriteHandshakeResponseAsync(HandshakeResponseMessage message)
...
                 return;
             }
 
-            _connectedAborted = true;
-
             Input.CancelPendingRead();
 
             // We fire and forget since this can trigger user code to run
-            ThreadPool.QueueUserWorkItem(_abortedCallback, this);
+            _ = Task.Run(AbortConnection);
```
davidfowl: This allocates

src/SignalR/server/Core/src/HubConnectionContext.cs
```diff
@@ -551,18 +577,28 @@ private void CheckClientTimeout()
...
-            }
-            catch (Exception ex)
-            {
-                Log.AbortFailed(connection._logger, ex);
-            }
-            finally
+
+            _ = InnerAbortConnection(connection);
+
+            async Task InnerAbortConnection(HubConnectionContext connection)
```
davidfowl: static

BrennanConroy: Didn't even realize that was a thing for local functions

src/SignalR/server/Core/src/HubConnectionContext.cs
```diff
@@ -551,18 +577,28 @@ private void CheckClientTimeout()
...
             {
-                // Communicate the fact that we're finished triggering abort callbacks
-                connection._abortCompletedTcs.TrySetResult(null);
+                await connection._writeLock.WaitAsync();
+
+                try
+                {
+                    connection._connectionAbortedTokenSource.Cancel();
```
davidfowl: This is calling user code under a lock.

src/SignalR/server/Core/src/HubConnectionContext.cs
```diff
@@ -561,8 +588,23 @@ private static void AbortConnection(object state)
...
             finally
             {
-                // Communicate the fact that we're finished triggering abort callbacks
-                connection._abortCompletedTcs.TrySetResult(null);
+                _ = InnerAbortConnection(connection);
+            }
+
+            static async Task InnerAbortConnection(HubConnectionContext connection)
+            {
+                await connection._writeLock.WaitAsync();
```
davidfowl: Add a comment here about why we're waiting

BrennanConroy:
>Unit test?

It's a race and there isn't anywhere to inject our own test code to force it.

src/SignalR/server/Core/src/HubConnectionContext.cs
```diff
@@ -173,7 +178,7 @@ public virtual ValueTask WriteAsync(HubMessage message, CancellationToken cancel
         /// <returns></returns>
         public virtual ValueTask WriteAsync(SerializedHubMessage message, CancellationToken cancellationToken = default)
         {
-            if (_connectedAborted)
+            if (_connectionAborted)
```
halter73: Is there much reason to check _connectionAborted before acquiring the lock when we check immediately after? Presumably, if the connection is aborted, whoever has the lock shouldn't for long. I prefer removing redundant checks if possible.

src/SignalR/server/Core/src/HubConnectionContext.cs
```diff
@@ -313,13 +333,19 @@ private async Task TryWritePingSlowAsync()
         {
             try
             {
+                if (_connectionAborted)
```
halter73: Just looking at this for the first time, I find it very funny that we have TryWritePingSlowAsync for the normal case when we actually write a ping, but save on the state machine allocation in the TryWritePingAsync failure case when it can't immediately acquire the write lock.

I think we might have officially jumped the shark with this fast/slow/awaited pattern 🤣

anurse: fonzie jumping the shark

src/SignalR/server/Core/src/HubConnectionContext.cs
```diff
@@ -355,15 +381,15 @@ private async Task WriteHandshakeResponseAsync(HandshakeResponseMessage message)
         /// </summary>
         public virtual void Abort()
         {
+            _connectionAborted = true;
```
halter73: Why was this moved? If _connectionAbortedTokenSource.IsCancellationRequested is true, doesn't that mean that Abort() must have been called?

Also, since _connectionAborted is set outside a lock and PipeReader.CancelPendingRead doesn't necessarily schedule the reader because the reader can observer cancellation synchronously, _connectionAborted should be `volatile`.

halter73: Forget about the `volatile`. My brain is still a little poisoned back from when Pipes/SocketInput locked less. [This lock](https://github.com/dotnet/corefx/blob/87b2e237fafbdfb01f03f16558f3fa17f8288f91/src/System.IO.Pipelines/src/System/IO/Pipelines/Pipe.cs#L600) should mean the volatile is unnecessary.

davidfowl (approved):

src/SignalR/server/Core/src/HubConnectionContext.cs
davidfowl: Agreed

davidfowl:
Do we need to backport this?

BrennanConroy:
>Do we need to backport this?

Worst case it's an erroneous error log. There is no functional issue here. Based on that I don't think we should, but open to other opinions.

davidfowl:
Looks like a legit test failures:
- Microsoft.AspNetCore.SignalR.Tests.HubConnectionHandlerTests.ConnectionTimesOutIfInitialPingAndThenNoMessages
- 
Microsoft.AspNetCore.SignalR.Tests.HubConnectionHandlerTests.AbortFromHubMethodForcesClientDisconnect

BrennanConroy:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=186945&view=logs

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/2511

PR #10049, opened 2019-05-07
Adding more streaming benchmark scenarios

mikaelm12:
Making some more streaming changes so I want to add these new benchmarks so we can track the effects.

One thing to note is that there are two variations of the same scenario with the `IAsyncEnumerable` where we await either `Task.Yield` or `Task.Completed` task. Any thoughts on which we should keep? As you'd expect the completed task variation performs better but is that what we want to compare to? [Here's the issue with the results](https://github.com/aspnet/AspNetCore/issues/9934#issuecomment-489735994)

### DIFF
--- a/src/SignalR/perf/Microbenchmarks/DefaultHubDispatcherBenchmark.cs
+++ b/src/SignalR/perf/Microbenchmarks/DefaultHubDispatcherBenchmark.cs
@@ -201,7 +201,7 @@ namespace Microsoft.AspNetCore.SignalR.Microbenchmarks
 
             public async Task UploadStreamIAsynEnumerable(IAsyncEnumerable<string> stream)
             {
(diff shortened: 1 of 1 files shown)


BrennanConroy (approved):
Lets keep both for now, they're *interesting*

halter73:
I doubt the ChannelWriter variations are actually going async. I would just keep the CompletedTask IAsyncEnumerable variation that doesn't actually dispatch.

halter73 (approved):
I'm starting to wonder how many scenarios is too many. Do we have scenarios that we're no longer interested in that can be removed?

mikaelm12:
We always just comment out all the scenarios that aren't relevant to the current investigation so I don't think having a lot is an issue. Always harder to add them back then to just comment them out for a second.

PR #10053, opened 2019-05-07
Handle a few other potentially null properties

jkotalik:
Finalizing https://github.com/aspnet/AspNetCore/issues/6075 and https://github.com/aspnet/AspNetCore/issues/9873.

Testing this via another test for application warmup.


jkotalik:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=183607

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/1929

src/Servers/IIS/IIS/src/Core/IISHttpContext.IHttpConnectionFeature.cs
```diff
@@ -81,15 +81,33 @@ string IHttpConnectionFeature.ConnectionId
         private void InitializeLocalEndpoint()
         {
             var localEndPoint = GetLocalEndPoint();
-            LocalIpAddress = localEndPoint.GetIPAddress();
-            LocalPort = localEndPoint.GetPort();
+            if (localEndPoint == null)
+            {
+                // ApplicationInitialization request might not set the local endpoint
+                LocalIpAddress = IPAddress.None;
```
anurse: Remind me: The potential issue here is the nullref when dereferencing `localEndPoint` right? If so, maybe just don't set `LocalIpAddress` and `LocalPort` if `localEndPoint` is `null` rather than setting them to bogus values?

jkotalik: I'm reviewing code now and I think you are right. If the calling code in our middleware (for example, the forwarding header middleware: https://github.com/aspnet/AspNetCore/blob/master/src/Middleware/HttpOverrides/src/ForwardedHeadersMiddleware.cs#L219) handles null correctly, then I think we are okay to return null.

jkotalik:
Going to verify this on a down level OS, per @Tratcher recommendation.

jkotalik:
Verified on Win8: 
image

Tratcher (approved):

PR #10056, opened 2019-05-07
[2.2] Port Auth Z Remove caching 

HaoK:
2.2 port of https://github.com/aspnet/AspNetCore/issues/9779

### DIFF
--- a/src/Mvc/Mvc.Core/src/Authorization/AuthorizeFilter.cs
+++ b/src/Mvc/Mvc.Core/src/Authorization/AuthorizeFilter.cs
@@ -24,7 +24,6 @@ namespace Microsoft.AspNetCore.Mvc.Authorization
     public class AuthorizeFilter : IAsyncAuthorizationFilter, IFilterFactory
     {
         private MvcOptions _mvcOptions;
-        private AuthorizationPolicy _effectivePolicy;
 
         /// <summary>
         /// Initializes a new <see cref="AuthorizeFilter"/> instance.
@@ -128,13 +127,7 @@ namespace Microsoft.AspNetCore.Mvc.Authorization
 
         private async Task<AuthorizationPolicy> GetEffectivePolicyAsync(AuthorizationFilterContext context)
         {
-            if (_effectivePolicy != null)
-            {
-                return _effectivePolicy;
-            }
-
             var effectivePolicy = await ComputePolicyAsync();
-            var canCache = PolicyProvider == null;
 
             if (_mvcOptions == null) 
             {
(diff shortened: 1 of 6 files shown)


HaoK:
Not quite ready for review, one of the new tests added fails in 2.2

HaoK:
Yowzers, didn't expect so many changes between 3.0 and 2.2

HaoK:
@aspnet/build are these signal r failures normal for release 2.2?  I reran them once already and they still fail.  I don't think I made any changes that would affect signal R at all in this PR

natemcmaster:
@HaoK  not sure - SignalR tests in 2.2 might just be flaky. Are they failing without your changes, too? 

cc @anurse

HaoK:
3rd retry did the trick, i guess its just flaky

anurse:
https://github.com/aspnet/AspNetCore-Internal/issues/2478

anurse:
Someone needs to provide some context (customer impact, fix description and risk) and/or rep this at shiproom to get it in. Nobody had the context today to approve it.

HaoK:
cc @rynowak this is the port for fixing the pretty bad authz bug with authorize filter + mvc, I'll open the PR for 2.1 as well.

Customer Impact: When mixing multiple authentication schemes and authorize attributes (on controller + methods), we end up caching incorrectly, so the wrong authorization policy is applied on subsequent requests.  This usually results in valid requests being rejected, I don't think there's any way this would let people who shouldn't be allowed access in.

Fix description: is to remove all the caching in MVC, so authorization policies are computed every time

Risk: Should be low (just removing caching logic, so just slower)

natemcmaster (changes requested):
Please update eng/PatchConfig.props.

dougbu:
> I'll open the PR for 2.1 as well.

2.1 and 2.2 should be aligned. Might be better to _only_ open a 2.1 PR and help it merge forward. Most likely difference would be in eng/PatchConfig.props but that can usually be resolved in the GitHub UI.

Please include the info from https://github.com/aspnet/AspNetCore/pull/10056#issuecomment-497443667 in the 2.1 PR whether or not you close this one.

HaoK:
2.1 PR is here: https://github.com/aspnet/AspNetCore/pull/10825

anurse:
@rynowak can you review? @HaoK please update eng/PatchConfig.props so this change actually ships :)

HaoK:
Ok I think @dougbu suggested that we probably close this, use the 2.1 PR and merge to 2.2

anurse:
Sounds good. I'll close and note that at shiproom.

HaoK:
Bad news, the code for authorize filter is different enough that I think we might need to just take 2 PRs

HaoK:
Closing for now

udlose:
@HaoK was this fix backported to .NET Core 2.2?  I am using 2.2 and facing this exact problem!

HaoK:
No given that there are workarounds and we had concerns with the fix, we decided not to backport this

udlose:
@HaoK  Can you please provide the workaround?

HaoK:
If setting the mvc option to false, and plugging in a custom policy provider both don't work, you should always be able to only have a single authorize attribute on each method (don't have any on the controller or global filters)

udlose:
@haok Thanks for your reply.

Can you clarify what "setting the mvc option to false" means? I'm not sure where I'd disable that.  My app is an API. Thank you for your time!

PR #10057, opened 2019-05-07
Set defaults for Json options

### LINKED ISSUE #9519: Apply API Review feedback to JsonHubProtocolOptions
Specifically, we should review the `JsonHubProtocolOptions` for the new System.Text.Json based protocol.

BrennanConroy:
Fixes https://github.com/aspnet/AspNetCore/issues/9519

### DIFF
--- a/src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
+++ b/src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
@@ -777,6 +777,10 @@ namespace Microsoft.AspNetCore.SignalR.Protocol
             options.IgnoreNullValues = false;
             options.IgnoreReadOnlyProperties = false;
             options.PropertyNamingPolicy = JsonNamingPolicy.CamelCase;
+            options.PropertyNameCaseInsensitive = true;
+            options.MaxDepth = 64;
+            options.DictionaryKeyPolicy = null;
+            options.DefaultBufferSize = 16 * 1024;
 
             return options;
         }
--- a/src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
+++ b/src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
@@ -777,7 +777,7 @@ namespace Microsoft.AspNetCore.SignalR.Protocol
             options.IgnoreNullValues = false;
             options.IgnoreReadOnlyProperties = false;
             options.PropertyNamingPolicy = JsonNamingPolicy.CamelCase;
-            options.PropertyNameCaseInsensitive = true;
+            options.PropertyNameCaseInsensitive = false;
             options.MaxDepth = 64;
             options.DictionaryKeyPolicy = null;
             options.DefaultBufferSize = 16 * 1024;


halter73 (approved):

mikaelm12 (approved):

src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs
```diff
@@ -777,6 +777,10 @@ internal static JsonSerializerOptions CreateDefaultSerializerSettings()
             options.IgnoreNullValues = false;
             options.IgnoreReadOnlyProperties = false;
             options.PropertyNamingPolicy = JsonNamingPolicy.CamelCase;
+            options.PropertyNameCaseInsensitive = true;
```
JamesNK: I believe MVC is leaving this as false. Should there be consistency between ASP.NET Core frameworks?

// @rynowak

rynowak: yeah, we were leaving this as false. I'm interested to know why folks think case-insensitive is a better option 😆

BrennanConroy: I believe it's because Json.NET is true by default.

JamesNK: Ryan is making a stand that Json.NET has the wrong default. I don't feel strongly either way other than SignalR and MVC should be consistent

*#TeamConsistency*

rynowak: I don't feel incredibly strongly about SignalR and MVC being consistent. I do feel strongly that case-insensitive is the wrong choice for building a REST api. 

I think it's a wierd idea to we would do case-insenstive because legacy.

anurse: As long as `JsonNamingPolicy.CamelCase` affects reading (IIRC it didn't in Json.NET?) I'm OK with leaving this false.

BrennanConroy: Looks like it affects reading. `{"someName":"value"}` will serialize into `class C { public string SomeName { get; set; }` with CamelCase set, and doesn't if it isn't.

JamesNK: The policy effects serializing and deserializing.

JamesNK (approved):

BrennanConroy:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=183985&view=logs

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/1929
https://github.com/aspnet/AspNetCore-Internal/issues/2250

BrennanConroy:
I am unable to merge this change because the blazor tests are too flaky...

@SteveSandersonMS I think your recent change to make tests more reliable might have made them more unreliable? https://github.com/aspnet/AspNetCore/commit/e0007f4f4bf6f0bd42d16c5d2e5cec2023b42a96

SteveSandersonMS:
@BrennanConroy Sorry for the inconvenience.

*Something* has changed to make the components E2E tests start being flaky since yesterday, but I don't know what. It started before I made the change you mention (my change was an attempt to address it, but didn't).

I've been investigating extensively and have got as far as [this](https://github.com/aspnet/AspNetCore/pull/10082#issuecomment-490872876), which *might* fix it.

PR #10058, opened 2019-05-07
Fix #6764 EndpointConventionBuilder API review

### LINKED ISSUE #6764: API Review: Review extension methods on IEndpointConventionBuilder

rynowak:
Fixes: #6764

### DIFF
--- a/src/Security/Authorization/Core/src/Policy/AuthorizationEndpointConventionBuilderExtensions.cs
+++ b/src/Security/Authorization/Core/src/Policy/AuthorizationEndpointConventionBuilderExtensions.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using Microsoft.AspNetCore.Authorization;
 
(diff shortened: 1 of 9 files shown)


src/Security/Authorization/Core/src/Policy/AuthorizationEndpointConventionBuilderExtensions.cs
```diff
@@ -13,38 +13,25 @@ namespace Microsoft.AspNetCore.Builder
     public static class AuthorizationEndpointConventionBuilderExtensions
     {
         /// <summary>
-        /// Adds authorization policies with the specified <see cref="IAuthorizeData"/> to the endpoint(s).
+        /// Adds the default authorization policy to the endpoint(s).
```
rynowak: @JamesNK - I reordered this.

I also made a different (and I think more inutitive decision) than what we discussed. Open to discuss further. 

Check out the docs on the other methods here.

JamesNK (approved):
Looks good to me :shipit:

src/Security/Authorization/Core/src/Policy/AuthorizationEndpointConventionBuilderExtensions.cs
rynowak: This will use the default policy if you don't specify any. So basically:

```
builder.RequireAuthorization() == builder.RequireAuthorization(Array.Empty<string>())
```

This made more sense to me then trying to force you to pass in a policy/data.

rynowak: Also will add a test for this behaviour since there aren't any.

rynowak:
FYI I need to sit on this for a day or two, it's blocked by a different Blazour PR.

PR #10062, opened 2019-05-08
Defer link interception until Router is initialized

### LINKED ISSUE #9834: Blazor Component Embedded in Razor Page
Repo: https://github.com/CarrotSolutions/WebApplication1.git

Please Help! We created a basic template razor page app.  We added Blazor components.  We have no problem getting this to work via replicating the steps in the 04/16 Blazor Update video or using the Blazor documentation.  However, once we add the Blazor component to a Razor page, all of our asp-page links no longer work.  The url changes in the address bar, but we remain on the same page.  The startup file is as follows:
```
        {
            services.Configure<CookiePolicyOptions>(options =>
            {
                options.CheckConsentNeeded = context => true;
            });


            services.AddRazorPages()
                .AddNewtonsoftJson();

            services.AddServerSideBlazor();
        }

        public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
        {
            if (env.IsDevelopment())
            {
            }
            else
            {
                app.UseExceptionHandler("/Error");
                app.UseHsts();
            }

            app.UseHttpsRedirection();
            app.UseStaticFiles();

            app.UseCookiePolicy();

            app.UseRouting();

            app.UseAuthorization();

            app.UseEndpoints(endpoints =>
            {
                endpoints.MapRazorPages();
                endpoints.MapBlazorHub();
            });
        }
```

Page with working Component and broken link
```

<a asp-page="./Index">Home</a>

<br /> 

<div id="Counter"> 
    @(await Html.RenderComponentAsync<Counter>()) 
    <script src="~/_framework/blazor.server.js">
</script> 

```

pranavkm:
Fixes https://github.com/aspnet/AspNetCore/issues/9834


src/Components/test/testassets/ComponentsApp.Server/Pages/RegularRazor.cshtml
```diff
@@ -0,0 +1,37 @@
+@page
```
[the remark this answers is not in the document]
pranavkm: I wanted to test two things - one we could get to it, and the other to make sure not having a router component does not get us stuck in a place where intercept a link but don't do anything with it. I added a comment stating why we reference the component there.

pranavkm:
🆙 📅

SteveSandersonMS:
Thanks for the update! Sorry I didn't get to review this today, but will make a priority of doing so tomorrow.

src/Components/Blazor/Blazor/src/Services/WebAssemblyUriHelper.cs
```diff
@@ -28,7 +28,7 @@ internal WebAssemblyUriHelper()
         protected override void EnsureInitialized()
         {
             WebAssemblyJSRuntime.Instance.Invoke<object>(
-                Interop.EnableNavigationInterception,
+                Interop.ListenForNavigationEvents,
```
SteveSandersonMS: Conceptually, this is so much clearer! 👍

src/Components/Browser.JS/src/Services/UriHelper.ts
```diff
@@ -1,69 +1,88 @@
...
   enableNavigationInterception,
   navigateTo,
   getBaseURI: () => document.baseURI,
   getLocationHref: () => location.href,
 };
 
-function enableNavigationInterception(assemblyName: string, functionName: string) {
-  if (hasRegisteredEventListeners || assemblyName === undefined || functionName === undefined) {
+function listenForNavigationEvents(assemblyName: string, functionName: string) {
+  if (hasRegisteredNavigationEventListeners || assemblyName === undefined || functionName === undefined) {
```
SteveSandersonMS: I just had a look and can't find any scenario where `assemblyName === undefined || functionName === undefined`. Unless you specifically know how those could happen, could we remove those clauses?

I know they were already there before your change, but maybe they can be cleaned away if they became redundant at some point in the past.

src/Components/Browser.JS/src/Services/UriHelper.ts
```diff
@@ -1,69 +1,88 @@
...
   } else {
     location.href = uri;
   }
 }
 
-function performInternalNavigation(absoluteInternalHref: string) {
+function performInternalNavigation(absoluteInternalHref: string, interceptedLink: boolean) {
   history.pushState(null, /* ignored title */ '', absoluteInternalHref);
-  handleInternalNavigation();
+  notifyLocationChanged(interceptedLink);
```
SteveSandersonMS: Just to check, is there any chance that this will result in duplicate notifications?

I see that we're calling `pushState` above, which is good. But won't that also trigger the `popstate` listener we are also registering? If so it seems like `notifyLocationChanged` would get called twice.

If you could clarify what actually happens (and maybe add a comment, if this code is correct) that would be great.

src/Components/Components/src/Microsoft.AspNetCore.Components.csproj
```diff
@@ -13,4 +13,8 @@
     <Reference Include="System.ComponentModel.Annotations" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Compile Include="..\..\Browser\src\BrowserUriHelperInterop.cs" LinkBase="Routing" />
```
SteveSandersonMS: Hopefully this might not be needed. See comment below.

src/Components/Components/src/Routing/Router.cs
```diff
@@ -79,41 +86,57 @@ protected virtual void Render(RenderTreeBuilder builder, Type handler, IDictiona
...
-        private void OnLocationChanged(object sender, string newAbsoluteUri)
+        Task IHandleAfterRender.OnAfterRenderAsync()
         {
-            _locationAbsolute = newAbsoluteUri;
-            if (_renderHandle.IsInitialized)
+            if (!_navigationInterceptionEnabled && ComponentContext.IsConnected)
             {
-                Refresh();
+                _navigationInterceptionEnabled = true;
+                return JSRuntime.InvokeAsync<object>(Interop.EnableNavigationInterception);
```
SteveSandersonMS: This is a bit awkward - sorry I didn't anticipate this when we discussed the design.

In general the intent is for the core `Microsoft.AspNetCore.Components` package not to make assumptions about the UI rendering technology. As such it doesn't assume that the UI renderer is a browser that would even have an associated JS runtime. The concept of "web browser" of that is in higher layers (e.g., `Microsoft.AspNetCore.Components.Browser`). The reason for this is to keep open the possibility that we allow completely different UI rendering technologies, such as native apps. So, AFAIK (and please correct me if I'm wrong), there aren't any other places in `Microsoft.AspNetCore.Components` where the framework makes JS interop calls.

To avoid making a JS interop call directly from here, we might need to add a new interface method. We could either:

 * Put a new method, `EnableNavigationInterception`, onto `IUriHelper` (since there are different implementations of that in different environments). It might be a bit weird for this to be public, though.
 * Or make a new internal DI service for enabling navigation interception. It could be a new interface `INavigationInterception` where the existing concrete `IUriHelper` implementations also implement that, so there's not much extra code or config to implement. The new interface would just be to segregate it from `IUriHelper` so it doesn't pollute intellisense.

Sorry that's extra hassle. Do you think this is viable?

src/Components/test/testassets/ComponentsApp.Server/Pages/RegularRazor.cshtml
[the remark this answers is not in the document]
SteveSandersonMS: Understood - thanks. If there's any ways of shrinking this file that would be great (e.g., removing all the CSS, the SignalR logging config, the footer, etc) that would be ideal. I'm only concerned that the duplication will make it another thing to remember to update when stuff changes.

SteveSandersonMS:
This looks really great!

One other E2E test that would be excellent to have would be showing that a component can observe the URL via `IUriHelper` even when there's no `Router`. Ideally it would also do JS interop to change the URL via `pushState` and see that it can still see the updated URL on the .NET side.

src/Components/Browser.JS/src/Services/UriHelper.ts
SteveSandersonMS: Oh wait, I misunderstood. Seems that [`pushState` does not trigger `popstate`](https://stackoverflow.com/a/26308536), so this would not cause duplicates.

src/Components/test/testassets/ComponentsApp.Server/Pages/RegularRazor.cshtml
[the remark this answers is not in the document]
SteveSandersonMS: > make sure not having a router component does not get us stuck in a place where intercept a link but don't do anything with it.

You could also do that easily from the existing E2E entrypoint, since it doesn't have a router except in the "router" tests. If you added a new test component like any of the non-router tests, that would represent the scenario where there can be `<a>` tags and no `<Router>`.

Then the other case is just "we could get to it", right? That could be a trivial static HTML file.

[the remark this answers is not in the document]
pranavkm: Maybe I'm missing something, but the App in the server side test always has a router?

src/Components/Browser.JS/src/Services/UriHelper.ts
pranavkm: Yup. I misunderstood how `pushstate` works until I had a look at the documentation for it. It would have been very strange for it to raise the `popstate` event when you called `pushState` now that I think of it 😄

src/Components/test/testassets/ComponentsApp.Server/Pages/RegularRazor.cshtml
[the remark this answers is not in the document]
SteveSandersonMS: It doesn't always have one. The only `<Router>` in `BasicTestApp` is in `RouterTest\TestRouter.razor`, which is not used except if you specifically pick "Router" from the test selector. Note how the only components with `@page` directives are the ones under `RouterTest\*`.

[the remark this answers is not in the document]
SteveSandersonMS: Oh wait, you're talking about `ComponentsApp.Server`. We're trying not to put E2E tests there (in fact we have [a bug tracking the work to take existing ones out](https://github.com/aspnet/AspNetCore/issues/8992)).

Is it possible to put all the E2E tests for this in `BasicTestApp`? This one works as I described, with the router being optional. Anything in that app can run on both the server and the client. AFAIK the "defer link interception" feature should work identically in server-side and client-side Blazor, so the same test case should apply to either.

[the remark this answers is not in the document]
pranavkm: > Is it possible to put all the E2E tests for this in BasicTestApp?

I guess we already have a scenario of ensuring non-Blazor links are followed in BasicTestApp. I wanted to add a test that captures the user reported scenario. The external link being Razor Pages isn't super relevant here I suppose.

pranavkm:
🈹 📅

src/Components/test/testassets/TestServer/Startup.cs
```diff
@@ -59,6 +60,7 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
 
                 subdirApp.UseEndpoints(endpoints =>
                 {
+                    endpoints.Map("NotAComponent.html", context => context.Response.WriteAsync("<div id=\"test-info\">Not a component!</div>"));
```
SteveSandersonMS: Is this line still needed, consider that you added `src/Components/test/testassets/BasicTestApp/wwwroot/NotAComponent.html`?

SteveSandersonMS (approved):
This is great, @pranavkm! Thanks for the updates. It all looks perfect now. I added one further comment about whether a certain line of code is strictly needed - please deal with or ignore this however you wish.

I think the client side router's behavior will make a lot more sense to people now. Allowing links to non-client-side URLs independently of `<base href>` concerns is extremely valuable.

PR #10063, opened 2019-05-08
Move obj and bin directories into repo root

### LINKED ISSUE #9928: Change the default repo layout to match dotnet/Arcade
Part of incremental changes to move to arcade. https://github.com/dotnet/arcade/issues/88

Arcade repo layout prescribes putting bin/ and obj/ folders in `$repoRoot/artifacts/`, and putting infrastructure stuff in `eng/`. We can update aspnet repos to use this layout now as a stepping stone to move to arcade. 

See https://github.com/dotnet/arcade/blob/master/Documentation/ArcadeSdk.md

```
- eng/
	+ *.props/targets
- artifacts/
	+ tmp/
	+ log/
	+ packages/
         - bin/
             + $(ProjectName)
	- obj/
              + $(ProjectName)
```

natemcmaster:
Resolves https://github.com/aspnet/AspNetCore/issues/9928
Part of https://github.com/aspnet/AspNetCore/issues/7280 (converging to Arcade)

The Arcade SDK requires that the obj/ and bin/ folders be placed in the top-level artifacts/ folder of the repo. https://github.com/dotnet/arcade/blob/master/Documentation/ArcadeSdk.md#single-build-output Although this PR does not complete our Arcade convergence, this is a step towards updating our repo to build with the Arcade SDK.

Changes:
* Set output path for build to artifacts/bin/$(ProjectName)/ 
* Set intermediate output path for build to artifacts/obj/$(ProjectName)/
* Cleanup .gitignore files (remove duplication between repo-root and tested gitignore files)
* Add code check which looks for project files that share the same name (could cause issues)
* Rename project files to have unique names (avoid race condition of build output)
* Update all locations which were hard-coded to expect bin/ and obj/ in the project directory
* Add overrides for tests which still assert test binaries exist in a given location relative to the source code


BrennanConroy:
@jkotalik Said something about installers

natemcmaster:
This is a draft PR. Lots of work to do before it's ready to merge. Among other this, hundreds of tests are written to assume the test binary is in a certain folder relative to the source code.

natemcmaster:
Test failures on this PR correspond exactly with tests that are disabled in Helix. The root of the problem appears to be the same -- the test code makes assumptions about the location of source code relative to the test binary. These assumptions are not true in Helix and not true in this PR where obj/ and bin/ directories have been moved.

Fixing this requires updating a lot of test code. @HaoK [opened issues to track most of these test projects](https://github.com/aspnet/AspNetCore/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+helix) months ago, but we haven't seen traction on this. I see some of this issues have been punted or backlogged.

@mkArtakMSFT @Eilon @anurse - to move forward on this PR, I see 3 viable options:

1. Kick the can down the road. Keep these tests disabled on Helix, and find a way to override the Arcade SDK to keep build output in the location the test code expects.
2. Disablepoolza. I disable the tests completely until they are updated.
3. Moar halp! We assign higher priority to fixing tests which are broken in Helix and this PR.

Thoughts?

By the way, here are the test projects that are broken:

image

anurse:
Sounds like it would be good to chat about this at a build team sync (unfortunately I'll be missing this Thursday's one...). I'd *like* to get them all fixed up and working (since Helix requires it too) but don't really have the resources right now to do it.

HaoK:
The good news is I'm somewhat close to finishing Components and Components E2e (those were mostly node issues). Also the two analyzer issues should be a cheap fix, as its just a max file path issue (rename a bunch of files).

So the big ones are really just: ProjectTemplates + hosting functionals left I think

natemcmaster:
@HaoK what are you doing about NodeJS? I'm wondering if we could re-use the approach I'm taking here to install mssql on-demand and extend it to include installing nodejs on-demand?

HaoK:
Yeah have to install nodjs on demand was basically their suggestion: https://github.com/dotnet/core-eng/issues/5302#issuecomment-490522509  I was going to try this today since they pinged me

natemcmaster:
Let's see if we can combine efforts on this. Will ping you internally.

natemcmaster:
We can discuss this more in manager sync today, but I'm favoring option 1

> Kick the can down the road. Keep these tests disabled on Helix, and find a way to override the Arcade SDK to keep build output in the location the test code expects.

My concern is that we've been kicking the can down the road for months, but as this "can" is scoped to only 10 projects, it's maybe not that big a deal.

natemcmaster:
I'm going to send a PR to the team once this merges. After it goes in, everyone who pulls this change will need to do a `git clean` to remove obj/ directories, otherwise they'll start hitting dozens of compiler error caused by the compiler starting to pick up generated source code.

> error CS0579: Duplicate 'System.Reflection.AssemblyCompanyAttribute' attribute

natemcmaster:
Flaky test: InputDateInteractsWithEditContext_NullableDateTimeOffset - appears to be unrelated to this change.

dougbu (commented):
This PR contains a number of non-overlapping changes, making it larger than it needs to be and harder to review. Suggest splitting (say) the C++ settings consolidation, project renames and perhaps .gitignore cleanup into a separate PR.

Directory.Build.props
```diff
@@ -144,7 +144,7 @@
     <CreateDirectory Include="$(VisualStudioSetupOutputPath)" />
   </ItemGroup>
 
-  <PropertyGroup Condition=" '$(OutputInRepoRoot)' == 'true' ">
+  <PropertyGroup>
```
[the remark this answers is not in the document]
natemcmaster: Intentionally chose to remove this. Arcade doesn't support a OutputInRepoRoot = false attribute. To reduce the churn of adopting Arcade, I chose to eliminate this variable as a part of this PR.

natemcmaster:
@dougbu while it's certainly possible to split this PR, the changes made here were done for related reasons - changing the location of build output made the content of some gitignore files irrelevant, required ensuring project names are unique to avoid overlap in artifacts/bin/, and using a shared file for C++ made it easier to ensure all .vcxproj files have the new setting.

natemcmaster:
In other words, not planning to split this PR unless this somehow blocking the ability to review it.

src/Components/Directory.Build.props
```diff
@@ -1,11 +1,14 @@
 <Project>
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory)..\, Directory.Build.props))\Directory.Build.props" />
+
   <PropertyGroup>
     <!-- Workaround for https://github.com/aspnet/AspNetCore/issues/5486 which requires the bin and obj directory be in the project directory -->
-    <OutputInRepoRoot>false</OutputInRepoRoot>
+    <BaseIntermediateOutputPath />
+    <IntermediateOutputPath />
+    <BaseOutputPath />
+    <OutputPath />
```
dougbu: Minor but why don't `$(BaseIntermediateOutputPath)` and `$(IntermediateOutputPath)` need to be cleared elsewhere?

src/Analyzers/Analyzers/test/Microsoft.AspNetCore.Analyzers.Test.csproj
```diff
@@ -4,8 +4,13 @@
     <TargetFramework>netcoreapp3.0</TargetFramework>
     <PreserveCompilationContext>true</PreserveCompilationContext>
     <RootNamespace>Microsoft.AspNetCore.Analyzers</RootNamespace>
+
+    <!-- Tests do not work on Helix or when bin/ directory is not in project directory due to undeclared dependency on test content. -->
     <!-- https://github.com/aspnet/AspNetCore/issues/6549 -->
     <BuildHelixPayload>false</BuildHelixPayload>
+    <BaseOutputPath />
+    <OutputPath />
```
dougbu: Do we have an issue or issues tracking fixing the tests and removing these workarounds?

src/DefaultBuilder/testassets/Directory.Build.props
```diff
@@ -1,7 +1,9 @@
 <Project>
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory)..\, Directory.Build.props))\Directory.Build.props" />
+
   <PropertyGroup>
-    <OutputInRepoRoot>true</OutputInRepoRoot>
+    <!-- Tests do not work on Helix or when bin/ directory is not in project directory due to undeclared dependency on test content. -->
+    <BaseOutputPath />
+    <OutputPath />
```
dougbu: To confirm: The referencing project will be moved as a unit into the Helix payload and only built on those agents? And, the same thing happens where `testassets/Directory.Build.props` files are needed?

src/Middleware/CORS/samples/Directory.Build.props
```diff
@@ -0,0 +1,9 @@
+<Project>
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory)..\, Directory.Build.props))\Directory.Build.props" />
+
+  <PropertyGroup>
+    <!-- Tests do not work on Helix or when bin/ directory is not in project directory due to undeclared dependency on test content. -->
```
dougbu: Suggest correcting this comment to comment on the _sample_.

src/Mvc/Extensions.ApiDescription.Client/src/Microsoft.Extensions.ApiDescription.Client.nuspec
```diff
@@ -19,7 +19,7 @@
   <files>
     <file src="build\*" target="build" />
     <file src="buildMultiTargeting\*" target="buildMultiTargeting" />
-    <file src="bin\$configuration$\net461\Microsoft.Extensions.ApiDescription.Tasks.*" target="tasks\net461" />
-    <file src="bin\$configuration$\netstandard2.0\Microsoft.Extensions.ApiDescription.Tasks.*" target="tasks\netstandard2.0" />
+    <file src="$baseOutputPath$\$configuration$\net461\Microsoft.Extensions.ApiDescription.Tasks.*" target="tasks\net461" />
+    <file src="$baseOutputPath$\$configuration$\netstandard2.0\Microsoft.Extensions.ApiDescription.Tasks.*" target="tasks\netstandard2.0" />
```
dougbu: 👍

src/Servers/IIS/AspNetCoreModuleV2/CommonLib/CommonLib.vcxproj
```diff
@@ -75,20 +75,15 @@
   <PropertyGroup Label="UserMacros" />
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
     <LinkIncremental>true</LinkIncremental>
-    <OutDir>$(MSBuildProjectDirectory)\bin\$(Configuration)\$(Platform)\</OutDir>
```
dougbu: Going forward, will VS restore these settings when we edit project settings? I've seen it undo changes similar to this&hellip;

dougbu (commented):
Think I'm about ready to sign off. Suggest a quick offline chat to make sure I'm not missing anything.

dougbu:
Should we be concerned about the failing Helix run?

natemcmaster:
Helix failed with

>  Work item IISExpress.FunctionalTests/netcoreapp3.0 in job c2f5f074-88e1-4e43-877f-781abfb5abee has failed.

I'll rebase on changes Justin merged yesterday to handle this.

src/Analyzers/Analyzers/test/Microsoft.AspNetCore.Analyzers.Test.csproj
natemcmaster: They are sort of generally tracked by https://github.com/aspnet/AspNetCore/issues/4248. Fixing tests which don't work on Helix will -- in most cases -- also solve this problem. The root cause of tests not working is the same - the test binary expects some files to exist in a relative location to the test binary.

src/Components/Directory.Build.props
natemcmaster: Blazor build tools and targets are the only thing which expect the obj/ directory to exist in the project dir. See  https://github.com/aspnet/AspNetCore/issues/5486. Everywhere else I tested this change, the projects and tests were okay with obj being moved to $repoRoot$/artifacts/obj/$projectName$

src/DefaultBuilder/testassets/Directory.Build.props
natemcmaster: Depends on how we decide to solve https://github.com/aspnet/AspNetCore/issues/10268.

src/Middleware/CORS/samples/Directory.Build.props
natemcmaster: This is one of the cases where the folder probably should have been named "testassets" instead of samples. The tests depend on these samples.

src/Servers/IIS/AspNetCoreModuleV2/CommonLib/CommonLib.vcxproj
natemcmaster: I opened VS and built the project and didn't see .vcxproj files change. What would trigger VS to undo this change?

dougbu (approved):

src/Servers/IIS/AspNetCoreModuleV2/CommonLib/CommonLib.vcxproj
dougbu: Change something in the project's property page while in VS. If that update to the project doesn't regurgitate these settings, life is grand.

natemcmaster: Seems to be okay. Didn't run into problems with VS when changing properties via UI.

natemcmaster:
@aspnet-hello  triage https://dev.azure.com/dnceng/public/_build/results?buildId=190143

Issue #9281, opened 2019-04-11
Implement detection logic that an app is using SignalR / server-side Blazor


danroth27:
This is needed to detect if an app needs the Azure SignalR Service.

mkArtakMSFT:
https://github.com/aspnet/AspNetCore/issues/5629

PR #10065, opened 2019-05-08
Fixes: #9281 - implement SignalR detection

rynowak:
This change adds detection of various SignalR configure-related calls to
the startup analysis infrastructure.

Also adds a shim that VS is going to call into to analyze the project
pre-publish.

### DIFF
--- a/src/Analyzers/Analyzers/src/CompilationFeatureDetector.cs
+++ b/src/Analyzers/Analyzers/src/CompilationFeatureDetector.cs
@@ -54,6 +54,5 @@ namespace Microsoft.AspNetCore.Analyzers
 
             return features.ToImmutable();
         }
-
     }
 }
--- a/src/Analyzers/Analyzers/src/CompilationFeatureDetector.cs
+++ b/src/Analyzers/Analyzers/src/CompilationFeatureDetector.cs
@@ -1,0 +1,59 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Operations;
+
+namespace Microsoft.AspNetCore.Analyzers
+{
+    internal static class CompilationFeatureDetector
+    {
+        public static async Task<IImmutableSet<string>> DetectFeaturesAsync(
+            Compilation compilation,
+            CancellationToken cancellationToken = default)
+        {
+            var symbols = new StartupSymbols(compilation);
+            if (!symbols.HasRequiredSymbols)
+            {
+                // Cannot find ASP.NET Core types.
+                return ImmutableHashSet<string>.Empty;
+            }
+
+            var features = ImmutableHashSet.CreateBuilder<string>();
+
+            // Find configure methods in the project's assembly
+            var configureMethods = ConfigureMethodVisitor.FindConfigureMethods(symbols, compilation.Assembly);
+            for (var i = 0; i < configureMethods.Count; i++)
+            {
+                var configureMethod = configureMethods[i];
+
+                // Handles the case where a method is using partial definitions. We don't expect this to occur, but still handle it correctly.
+                var syntaxReferences = configureMethod.DeclaringSyntaxReferences;
+                for (var j = 0; j < syntaxReferences.Length; j++)
+                {
+                    var semanticModel = compilation.GetSemanticModel(syntaxReferences[j].SyntaxTree);
+
+                    var syntax = await syntaxReferences[j].GetSyntaxAsync(cancellationToken).ConfigureAwait(false);
+                    var operation = semanticModel.GetOperation(syntax);
+
+                    // Look for a call to one of the SignalR gestures that applies to the Configure method.
+                    if (operation
+                        .Descendants()
+                        .OfType<IInvocationOperation>()
+                        .Any(op => StartupFacts.IsSignalRConfigureMethodGesture(op.TargetMethod)))
+                    {
+                        features.Add(WellKnownFeatures.SignalR);
+                    }
+                }
+            }
+
+            return features.ToImmutable();
+        }
+
+    }
+}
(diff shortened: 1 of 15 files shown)


src/Analyzers/Analyzers/src/WellKnownFeatures.cs
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Analyzers
+{
+    internal static class WellKnownFeatures
```
rynowak: The plan is to expand the set of features we know about. For now just SignalR on the menu.

src/Analyzers/shared/FeatureDetection/ProjectFeatureDetector.cs
```diff
@@ -0,0 +1,80 @@
...
+    // Be very careful making changes to this file. No project in this repo builds it.
+    //
+    // If you need to verify a change, make a local project (net472) and copy in everything included by this project.
+    //
+    // You'll also need some nuget packages like:
+    // - Microsoft.VisualStudio.LanguageServices
+    // - Microsoft.VisualStudio.Shell.15.0
+    // - Microsoft.VisualStudio.Threading
+    [Export(typeof(ProjectFeatureDetector))]
+    internal class ProjectFeatureDetector
```
abpiskunov: >ProjectFeatureDetector [](start = 19, length = 22)

Small ask, should we rename it to ProjectCompillationFeatureDetector? Or you plan to use it for any other types of features?

src/Analyzers/shared/FeatureDetection/ProjectFeatureDetector.cs
```diff
@@ -0,0 +1,80 @@
...
+    {
+        private readonly Lazy<VisualStudioWorkspace> _workspace;
+
+        [ImportingConstructor]
+        public ProjectFeatureDetector(Lazy<VisualStudioWorkspace> workspace)
+        {
+            _workspace = workspace;
+        }
+
+        public async Task<IImmutableSet<string>> DetectFeaturesAsync(string projectFullPath, CancellationToken cancellationToken = default)
```
abpiskunov: >DetectFeaturesAsync [](start = 49, length = 19)

The flow now is: we call this DetectFeaturesAsync when dependencies are refreshed on Publish overview page. It works fine , however there is a scenario when publish overview page stays open and user does not do any action that would trigger dependencies refresh. This way if he adds SignalR in code we would never be notified. I think at this time it is OK, this suggestion should not be really "real time", they will see it next time the open publish page then. However if we want to address this, your another approach with having real roslyn analyzer can help, since we can wait when Roslyn calls our analyzer on the code change, analyzer in turn notify ProjectFeatureDetector and that one can ping tooling etc..

src/Analyzers/shared/FeatureDetection/ProjectFeatureDetector.cs
abpiskunov: Let me know please where i could get a source package when this is merged.

---
In reply to: [282145188](https://github.com/aspnet/AspNetCore/pull/10065#discussion_r282145188) [](ancestors = 282145188)

rynowak: Yeah, I'm happy to change the name 👍

src/Analyzers/shared/FeatureDetection/ProjectFeatureDetector.cs
rynowak: Sure. I'd like to start simple and see if that comes up via feedback. It's something that can be done, but it will require many more moving parts. It also dictates a different API shape, since we now need to have events.

abpiskunov (approved):
:shipit:

src/Analyzers/Analyzers/src/CompilationFeatureDetector.cs
```diff
@@ -0,0 +1,59 @@
...
+                var configureMethod = configureMethods[i];
+
+                // Handles the case where a method is using partial definitions. We don't expect this to occur, but still handle it correctly.
+                var syntaxReferences = configureMethod.DeclaringSyntaxReferences;
+                for (var j = 0; j < syntaxReferences.Length; j++)
+                {
+                    var semanticModel = compilation.GetSemanticModel(syntaxReferences[j].SyntaxTree);
+
+                    var syntax = await syntaxReferences[j].GetSyntaxAsync(cancellationToken).ConfigureAwait(false);
+                    var operation = semanticModel.GetOperation(syntax);
```
pranavkm: More for my benefit, but is there a reason to prefer using operation over just looking for `InvocationExpressionSyntax`? You can effectively get to the Symbol using the latter.

src/Analyzers/Analyzers/src/WellKnownFeatures.cs
```diff
@@ -0,0 +1,10 @@
...
+    {
+        public static readonly string SignalR = nameof(SignalR);
```
pranavkm: `AspNetCoreSignalR`?

pranavkm (approved):

src/Analyzers/Analyzers/src/WellKnownFeatures.cs
rynowak: nah, this is fine for now.

src/Analyzers/Analyzers/src/CompilationFeatureDetector.cs
rynowak: Operation is the cool new thing. It's a little more high level, and includes the merge of symbols + syntax. Operation is also language independent.

src/Analyzers/shared/FeatureDetection/ProjectCompilationFeatureDetector.cs
```diff
@@ -0,0 +1,80 @@
...
+    {
+        private readonly Lazy<VisualStudioWorkspace> _workspace;
+
+        [ImportingConstructor]
+        public ProjectCompilationFeatureDetector(Lazy<VisualStudioWorkspace> workspace)
+        {
+            _workspace = workspace;
+        }
+
+        public async Task<IImmutableSet<string>> DetectFeaturesAsync(string projectFullPath, CancellationToken cancellationToken = default)
```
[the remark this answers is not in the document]
rynowak: He did

PR #10067, opened 2019-05-08
Remove EndpointSelectorContext

rynowak:
Doing the dirty work.


src/Http/Routing/src/EndpointRoutingMiddleware.cs
```diff
@@ -55,12 +56,10 @@ internal sealed class EndpointRoutingMiddleware
 
         public Task Invoke(HttpContext httpContext)
         {
-            var feature = new EndpointSelectorContext(httpContext);
-
             // There's already an endpoint, skip maching completely
-            if (feature.Endpoint != null)
+            if (httpContext.GetEndpoint() != null)
```
JamesNK: nit: Assign GetEndpoint to variable and reuse variable

JamesNK (approved):
I couldn't find any extra lines to complain about

:shipit:

src/Http/Routing/src/EndpointRoutingMiddleware.cs
rynowak: note that I'm still fixing 1.2 more thigns here.

davidfowl (approved):
Thanks. We should look at maybe exposing endpoint as a property or moving the ext method to a better ns

Issue #10068, opened 2019-05-08
blazor.boot.json error


nicksull123:
I'm not very familiar with aspnetcore/blazor internals, but I believe the MemoryFIleInfo.PhysicalPath property in src/Components/Server/src/DependencyInjection/ConfigureStaticFilesOptions.cs should be returning null.

The StaticFIleContext.SendAsync method ends up trying to send C:\_framework\blazor.boot.json when the IHttpSendFileFeature check is successful (which is the case for http.sys, kestrel doesn't seem to support?).

PR #10071, opened 2019-05-08
blazor.boot.json - don't set physicalpath

nicksull123:
#10068

### DIFF
--- a/src/Components/Server/src/DependencyInjection/ConfigureStaticFilesOptions.cs
+++ b/src/Components/Server/src/DependencyInjection/ConfigureStaticFilesOptions.cs
@@ -97,7 +97,7 @@ namespace Microsoft.AspNetCore.Components.Server
 
                 public long Length => _data.Length;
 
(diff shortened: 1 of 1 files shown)


SteveSandersonMS:
Thanks, @nicksull123!

I agree it shouldn't be hardcoded to `"/_framework/blazor.boot.json"`. What I'm less sure of is whether it makes a different whether it has the value `null` or whether we pass in a value for `PhysicalPath` to the constructor.

@javiercn You have a lot of background knowledge in this area. Do you know whether it makes a difference to the static files middleware whether a given `IFileInfo` has a valid `PhysicalPath` or not?

javiercn:
Seems like a valid concern to me. @Tratcher for confirmation.

javiercn (approved):

Tratcher (approved):
Null is the correct value here. PhysicalPath must either be a valid path on disk or null.

SteveSandersonMS:
Thanks for confirming, @Tratcher!

Also thanks for the contribution, @nicksull123! It's now merged.

PR #10081, opened 2019-05-08
Make GetOrdereredMetadata return something good

rynowak:


src/Http/Http.Abstractions/src/Routing/OrderedEndpointMetadataCollection.cs
```diff
@@ -0,0 +1,118 @@
...
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.Http
+{
+    /// <summary>
+    /// A collection of ordered, strongly-typed metadata associated with an <see cref="Endpoint"/>.
+    /// </summary>
+    /// <typeparam name="T">The metadata type.</typeparam>
+    public sealed class OrderedEndpointMetadataCollection<T> : IReadOnlyList<T>
```
rynowak: alternative to writing a class would be to just expose `IReadOnlyList<>` for that sweet sweet interface dispatch.

davidfowl: As much as I hate interface dispatch (tho it really isn’t that bad), exposing a custom collection type seems like overkill...

rynowak: Top 10 Anime Battles of all time:
10. James vs blank new lines
9.
8.
7.
6.
5.
4.
3.
2.
1. Me (with my custom collection types) vs everyone

rynowak: I'm ok with `IReadOnlyList<>` because I don't think we'll end up adding more functionality. I did this to start the discussion and I'm not convinced either

JamesNK (approved):

PR #10093, opened 2019-05-08
try v142 toolset again

### LINKED ISSUE #7548: Upgrade IIS C++ projects to build with platformtoolset v142 (VS2019)

analogrelay:
fixes #7548

Lets see what the CI says :). If this doesn't work, I'm ready to just close the issue and say "so what, it uses the old toolset, #dealwithit" ;).


anurse:
I might be stumped. It seems to build fine on my machine. The binlog shows that it is *indeed* building `iisca.vcxproj` using the `14.16.27023` version of the compiler (whereas everything else is built with `14.20.27508`), hence linking fails for `aspnetcoreCA` because the object files are built with different toolset versions.

The thing is, `iisca.vcxproj` has the `v142` toolset specified, so I'm not sure what is changing it. At this point, I'm not sure I care enough to follow through :). I'll move the issue to a late preview and we can look at it as clean-up later on.

PR #10094, opened 2019-05-08
suppress ExecutionContext by default in TestServer

### LINKED ISSUE #7975: HttpContextAccessor.HttpContext is null when using TestServer and .net 2.2.2
### Describe the bug
When using .net core 2.2.2 along with the `TestServer`(https://github.com/aspnet/AspNetCore/blob/master/src/Hosting/TestHost/src/TestServer.cs), `HttpContextAccessor.HttpContext` will return null in certain situations. We're experiencing this is the repro: https://github.com/Microsoft/fhir-server/

If you have 2.2.1 installed, this behavior is not exhibited. It appears related to the use of authentication and calls that are made on the JWT backchannel handler. 

### To Reproduce
1. Clone my repro repository here: https://github.com/brandonpollett/TestServerRepro
1. Execute the unit tests with .net 2.2.2 and you will see a test failure
1. Remove .net 2.2.2 and run with 2.2.1 and the tests will succeed

* Note: in this sample repo the null reference occurs at line 237 of https://github.com/IdentityServer/IdentityServer4/blob/master/src/Services/Default/DefaultUserSession.cs, but in the actual repro, the null reference is on line 62 of https://github.com/Microsoft/fhir-server/blob/master/src/Microsoft.Health.Fhir.Api/Features/Routing/UrlResolver.cs

### Expected behavior
The tests should pass while using 2.2.2 and 2.2.1

### Additional context
When the unit tests pass the output of `dotnet --info` is:
```
C:\projects\TestServerRepro [master ≡]> dotnet --info
 Version:   2.2.103
 Commit:    8edbc2570a

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.17763
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\2.2.103\

Host (useful for support):
  Version: 2.2.1
  Commit:  878dd11e62

.NET Core SDKs installed:
  1.0.4 [C:\Program Files\dotnet\sdk]
  1.1.0 [C:\Program Files\dotnet\sdk]
  2.0.0 [C:\Program Files\dotnet\sdk]
  2.0.2 [C:\Program Files\dotnet\sdk]
  2.1.2 [C:\Program Files\dotnet\sdk]
  2.1.4 [C:\Program Files\dotnet\sdk]
  2.1.100-preview-007354 [C:\Program Files\dotnet\sdk]
  2.1.100-preview-007363 [C:\Program Files\dotnet\sdk]
  2.1.100 [C:\Program Files\dotnet\sdk]
  2.1.103 [C:\Program Files\dotnet\sdk]
  2.1.104 [C:\Program Files\dotnet\sdk]
  2.1.200 [C:\Program Files\dotnet\sdk]
  2.1.201 [C:\Program Files\dotnet\sdk]
  2.1.202 [C:\Program Files\dotnet\sdk]
  2.1.402 [C:\Program Files\dotnet\sdk]
  2.1.403 [C:\Program Files\dotnet\sdk]
  2.1.503 [C:\Program Files\dotnet\sdk]
  2.2.103 [C:\Program Files\dotnet\sdk]

.NET Core runtimes installed:
  Microsoft.AspNetCore.All 2.1.4 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.1.5 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.1.7 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.2.1 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.App 2.1.4 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.1.5 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.1.7 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.2.1 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 1.0.5 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 1.1.2 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.0.0 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.0.3 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.0.5 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.0.6 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.0.7 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.0.9 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.4 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.5 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.7 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.2.1 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
```

When failing the output of `dotnet--info is`:
```
C:\projects\TestServerRepro [master ≡]> dotnet --info
 Version:   2.2.104
 Commit:    73f036d4ac

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.17763
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\2.2.104\

Host (useful for support):
  Version: 2.2.2
  Commit:  a4fd7b2c84

.NET Core SDKs installed:
  1.0.4 [C:\Program Files\dotnet\sdk]
  1.1.0 [C:\Program Files\dotnet\sdk]
  2.0.0 [C:\Program Files\dotnet\sdk]
  2.0.2 [C:\Program Files\dotnet\sdk]
  2.1.2 [C:\Program Files\dotnet\sdk]
  2.1.4 [C:\Program Files\dotnet\sdk]
  2.1.100-preview-007354 [C:\Program Files\dotnet\sdk]
  2.1.100-preview-007363 [C:\Program Files\dotnet\sdk]
  2.1.100 [C:\Program Files\dotnet\sdk]
  2.1.103 [C:\Program Files\dotnet\sdk]
  2.1.104 [C:\Program Files\dotnet\sdk]
  2.1.200 [C:\Program Files\dotnet\sdk]
  2.1.201 [C:\Program Files\dotnet\sdk]
  2.1.202 [C:\Program Files\dotnet\sdk]
  2.1.402 [C:\Program Files\dotnet\sdk]
  2.1.403 [C:\Program Files\dotnet\sdk]
  2.1.503 [C:\Program Files\dotnet\sdk]
  2.2.104 [C:\Program Files\dotnet\sdk]

.NET Core runtimes installed:
  Microsoft.AspNetCore.All 2.1.4 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.1.5 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.1.7 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.2.2 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.App 2.1.4 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.1.5 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.1.7 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.2.2 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 1.0.5 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 1.1.2 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.0.0 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.0.3 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.0.5 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.0.6 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.0.7 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.0.9 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.4 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.5 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.7 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.2.2 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
```

analogrelay:
fixes #7975

Currently, the `ExecutionContext` flows from client to server in `TestServer`. This seems odd because the server should be a separate instance. Also, it causes problems when you are running two different apps through `TestServer` (such as when you are testing IdentityServer and such). This change changes `TestServer`'s client logic to force a new `ExecutionContext` when issuing HTTP requests.

There is a `PreserveExecutionContext` property to turn the old behavior back on. Also I had to modify where `IHttpApplication.CreateContext` is called since that's what sets the `IHttpContextAccessor`. If it's set before we dispatch to the thread pool (thus dropping the EC) the accessor value will be lost when we abandon the EC!

I'm reliably informed that there is an MVC test or two depend on the AsyncLocal flow, so I'll fix those in this PR.

### DIFF
--- a/src/Hosting/TestHost/src/HttpContextBuilder.cs
+++ b/src/Hosting/TestHost/src/HttpContextBuilder.cs
@@ -62,11 +62,12 @@ namespace Microsoft.AspNetCore.TestHost
         {
             var registration = cancellationToken.Register(AbortRequest);
 
-            // Everything inside this function happens in the SERVER's ExecutionContext (unless PreserveExecutionContext is true)
+            // Everything inside this function happens in the SERVER's execution context (unless PreserveExecutionContext is true)
             async Task RunRequestAsync()
             {
                 // This will configure IHttpContextAccessor so it needs to happen INSIDE this function,
-                // since we are now inside the Server's ExecutionContext. If it happens out
+                // since we are now inside the Server's execution context. If it happens outside this context,
+                // it will be lost when we abandon the execution context.
                 _testContext = _application.CreateContext(_httpContext.Features);
 
                 try
--- a/src/Hosting/TestHost/src/HttpContextBuilder.cs
+++ b/src/Hosting/TestHost/src/HttpContextBuilder.cs
@@ -15,7 +15,7 @@ namespace Microsoft.AspNetCore.TestHost
     {
         private readonly IHttpApplication<Context> _application;
         private readonly HttpContext _httpContext;
-        
+
         private TaskCompletionSource<HttpContext> _responseTcs = new TaskCompletionSource<HttpContext>(TaskCreationOptions.RunContinuationsAsynchronously);
         private ResponseStream _responseStream;
         private ResponseFeature _responseFeature = new ResponseFeature();
@@ -57,15 +57,17 @@ namespace Microsoft.AspNetCore.TestHost
         /// Start processing the request.
         /// </summary>
         /// <returns></returns>
-        internal Task<HttpContext> SendAsync(CancellationToken cancellationToken)
+        internal Task<HttpContext> SendAsync(CancellationToken cancellationToken, bool preserveExecutionContext)
         {
             var registration = cancellationToken.Register(AbortRequest);
 
-            _testContext = _application.CreateContext(_httpContext.Features);
-
-            // Async offload, don't let the test code block the caller.
-            _ = Task.Factory.StartNew(async () =>
+            // Everything inside this function happens in the SERVER's ExecutionContext (unless PreserveExecutionContext is true)
+            async Task RunRequestAsync()
             {
+                // This will configure IHttpContextAccessor so it needs to happen INSIDE this function,
+                // since we are now inside the Server's ExecutionContext. If it happens out
+                _testContext = _application.CreateContext(_httpContext.Features);
+
                 try
                 {
                     await _application.ProcessRequestAsync(_testContext);
@@ -81,7 +83,17 @@ namespace Microsoft.AspNetCore.TestHost
                 {
                     registration.Dispose();
                 }
-            });
+            }
+
+            // Async offload, don't let the test code block the caller.
+            if (preserveExecutionContext)
+            {
+                _ = Task.Factory.StartNew(RunRequestAsync);
+            }
+            else
+            {
+                ThreadPool.UnsafeQueueUserWorkItem(_ => { _ = RunRequestAsync(); }, null);
+            }
 
             return _responseTcs.Task;
         }
--- a/src/Hosting/TestHost/src/HttpContextBuilder.cs
+++ b/src/Hosting/TestHost/src/HttpContextBuilder.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Globalization;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Hosting.Server;
@@ -92,7 +93,13 @@ namespace Microsoft.AspNetCore.TestHost
             }
             else
             {
-                ThreadPool.UnsafeQueueUserWorkItem(_ => { _ = RunRequestAsync(); }, null);
+                // Capture the current culture and restore it inside the server execution context.
+                var currentCulture = CultureInfo.CurrentCulture;
+                ThreadPool.UnsafeQueueUserWorkItem(_ =>
+                {
+                    CultureInfo.CurrentCulture = currentCulture;
+                    _ = RunRequestAsync();
+                }, null);
             }
 
             return _responseTcs.Task;
--- a/src/Hosting/TestHost/src/HttpContextBuilder.cs
+++ b/src/Hosting/TestHost/src/HttpContextBuilder.cs
@@ -15,6 +15,7 @@ namespace Microsoft.AspNetCore.TestHost
     internal class HttpContextBuilder : IHttpBodyControlFeature
     {
         private readonly IHttpApplication<Context> _application;
+        private readonly bool _preserveExecutionContext;
         private readonly HttpContext _httpContext;
 
         private TaskCompletionSource<HttpContext> _responseTcs = new TaskCompletionSource<HttpContext>(TaskCreationOptions.RunContinuationsAsynchronously);
@@ -24,10 +25,11 @@ namespace Microsoft.AspNetCore.TestHost
         private bool _pipelineFinished;
         private Context _testContext;
 
-        internal HttpContextBuilder(IHttpApplication<Context> application, bool allowSynchronousIO)
+        internal HttpContextBuilder(IHttpApplication<Context> application, bool allowSynchronousIO, bool preserveExecutionContext)
         {
             _application = application ?? throw new ArgumentNullException(nameof(application));
             AllowSynchronousIO = allowSynchronousIO;
+            _preserveExecutionContext = preserveExecutionContext;
             _httpContext = new DefaultHttpContext();
 
             var request = _httpContext.Request;
@@ -58,7 +60,7 @@ namespace Microsoft.AspNetCore.TestHost
         /// Start processing the request.
         /// </summary>
         /// <returns></returns>
-        internal Task<HttpContext> SendAsync(CancellationToken cancellationToken, bool preserveExecutionContext)
+        internal Task<HttpContext> SendAsync(CancellationToken cancellationToken)
         {
             var registration = cancellationToken.Register(AbortRequest);
 
@@ -66,7 +68,7 @@ namespace Microsoft.AspNetCore.TestHost
             async Task RunRequestAsync()
             {
                 // This will configure IHttpContextAccessor so it needs to happen INSIDE this function,
-                // since we are now inside the Server's execution context. If it happens outside this context,
+                // since we are now inside the Server's execution context. If it happens outside this cont
                 // it will be lost when we abandon the execution context.
                 _testContext = _application.CreateContext(_httpContext.Features);
 
@@ -88,7 +90,7 @@ namespace Microsoft.AspNetCore.TestHost
             }
 
             // Async offload, don't let the test code block the caller.
-            if (preserveExecutionContext)
+            if (_preserveExecutionContext)
             {
                 _ = Task.Factory.StartNew(RunRequestAsync);
             }
--- a/src/Hosting/TestHost/src/HttpContextBuilder.cs
+++ b/src/Hosting/TestHost/src/HttpContextBuilder.cs
@@ -94,11 +94,8 @@ namespace Microsoft.AspNetCore.TestHost
             }
             else
             {
-                // Capture the current culture and restore it inside the server execution context.
-                var currentCulture = CultureInfo.CurrentCulture;
                 ThreadPool.UnsafeQueueUserWorkItem(_ =>
                 {
-                    CultureInfo.CurrentCulture = currentCulture;
                     _ = RunRequestAsync();
                 }, null);
             }
--- a/src/Hosting/TestHost/src/TestServer.cs
+++ b/src/Hosting/TestHost/src/TestServer.cs
@@ -83,12 +83,12 @@ namespace Microsoft.AspNetCore.TestHost
         /// <remarks>
         /// Defaults to false.
         /// </remarks>
-        public bool AllowSynchronousIO { get; set; } = false;
+        public bool AllowSynchronousIO { get; set; }
 
         /// <summary>
         /// Gets or sets a value that controls if <see cref="ExecutionContext"/> and <see cref="AsyncLocal{T}"/> values are preserved from the client to the server.
         /// </summary>
-        public bool PreserveExecutionContext { get; set; } = false;
+        public bool PreserveExecutionContext { get; set; }
 
         private IHttpApplication<Context> Application
         {
--- a/src/Hosting/TestHost/src/TestServer.cs
+++ b/src/Hosting/TestHost/src/TestServer.cs
@@ -78,15 +78,12 @@ namespace Microsoft.AspNetCore.TestHost
         public IFeatureCollection Features { get; }
 
         /// <summary>
-        /// Gets or sets a value that controls whether synchronous IO is allowed for the <see cref="HttpContext.Request"/> and <see cref="HttpContext.Response"/>
+        /// Gets or sets a value that controls whether synchronous IO is allowed for the <see cref="HttpContext.Request"/> and <see cref="HttpContext.Response"/>. The default value is <see langword="false" />.
         /// </summary>
-        /// <remarks>
-        /// Defaults to false.
-        /// </remarks>
         public bool AllowSynchronousIO { get; set; }
 
         /// <summary>
-        /// Gets or sets a value that controls if <see cref="ExecutionContext"/> and <see cref="AsyncLocal{T}"/> values are preserved from the client to the server.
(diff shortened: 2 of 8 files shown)


rynowak:
This breaks our recipe for testing things that depend on localized resources in .NET.

src/Hosting/TestHost/src/HttpContextBuilder.cs
```diff
@@ -59,15 +59,17 @@ internal void Configure(Action<HttpContext> configureContext)
...
-            _testContext = _application.CreateContext(_httpContext.Features);
-
-            // Async offload, don't let the test code block the caller.
-            _ = Task.Factory.StartNew(async () =>
+            // Everything inside this function happens in the SERVER's ExecutionContext (unless PreserveExecutionContext is true)
+            async Task RunRequestAsync()
             {
+                // This will configure IHttpContextAccessor so it needs to happen INSIDE this function,
+                // since we are now inside the Server's ExecutionContext. If it happens out
+                _testContext = _application.CreateContext(_httpContext.Features);
```
rynowak: We should capture and set the culture in here. This is the only common reliance on the execution context that I'm aware of.

anurse:
> This breaks our recipe for testing things that depend on localized resources in .NET.

I'm mildly concerned that no tests seem to have failed despite this. Are there tests in this repo that should be catching this break?

As per our offline discussion, I'll capture and transfer the culture across.

src/Hosting/TestHost/src/HttpContextBuilder.cs
```diff
@@ -59,15 +59,17 @@ internal void Configure(Action<HttpContext> configureContext)
...
 
-            _testContext = _application.CreateContext(_httpContext.Features);
-
-            // Async offload, don't let the test code block the caller.
-            _ = Task.Factory.StartNew(async () =>
+            // Everything inside this function happens in the SERVER's ExecutionContext (unless PreserveExecutionContext is true)
+            async Task RunRequestAsync()
             {
+                // This will configure IHttpContextAccessor so it needs to happen INSIDE this function,
+                // since we are now inside the Server's ExecutionContext. If it happens out
```
anurse: Note to self: Don't forget to

anurse:
:up: :date: now with more culture!

src/Hosting/TestHost/src/HttpContextBuilder.cs
davidfowl: Hmm, for why?

rynowak: We use this in a bunch of places in tests where we're looking at something related to globlisation. That's the one dependency I think it's pretty reasonable to expect to work (setting culture).

src/Hosting/TestHost/src/TestServer.cs
```diff
@@ -85,6 +85,11 @@ public IWebHost Host
         /// </remarks>
         public bool AllowSynchronousIO { get; set; } = false;
 
+        /// <summary>
+        /// Gets or sets a value that controls if <see cref="ExecutionContext"/> and <see cref="AsyncLocal{T}"/> values are preserved from the client to the server.
+        /// </summary>
+        public bool PreserveExecutionContext { get; set; } = false;
```
rynowak: nit: init to false ins't needed here or on line 86.

rynowak (approved):

src/Hosting/TestHost/src/HttpContextBuilder.cs
davidfowl: Why not just make the default True then so we don't need to special case anything?

src/Hosting/TestHost/src/TestServer.cs
```diff
@@ -83,7 +83,12 @@ public IWebHost Host
         /// <remarks>
         /// Defaults to false.
         /// </remarks>
-        public bool AllowSynchronousIO { get; set; } = false;
+        public bool AllowSynchronousIO { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value that controls if <see cref="ExecutionContext"/> and <see cref="AsyncLocal{T}"/> values are preserved from the client to the server.
```
Tratcher: "The default value is..."

src/Hosting/TestHost/src/HttpContextBuilder.cs
Tratcher: Yeah, it's a bit odd to special case one thing. The broken case (IHttpContextAccessor) is rare, but hard to troubleshoot. The culture case seems more common in our tests (not sure if any customers use this approach), but when it's not working I'd expect the cause to be more obvious since people are more familiar with how culture flows.

If culture is a thing you regularly want to test in an application why not make that a first class property of the TestServer? The server can set that in its own execution context.

rynowak: Yeah, I think that's an even better idea.

davidfowl: I thought the culture setup in MVC sets it up on the middleware side, not the testserver side. https://github.com/aspnet/AspNetCore/blob/dcf49f257560130597b35f3b059b4ed34e8e3ead/src/Mvc/test/Mvc.FunctionalTests/Infrastructure/MvcTestFixture.cs#L21

https://github.com/aspnet/AspNetCore/blob/c565386a3ed135560bc2e9017aa54a950b4e35dd/src/Mvc/test/Mvc.FunctionalTests/Infrastructure/MvcWebApplicationBuilderExtensions.cs#L23

anurse: Our tests don't seem to care what we do here, so I think this is more about existing **customer** scenarios.

Let's lay out the situation and come to a conclusion. This is my proposal based on this thread)

1. Default `PreserveExecutionContext` to `false` (I still think this is the default behavior users will want)
1. Add a `public CultureInfo RequestCuluture { get; set; }` property (feel free to bikeshed the name) to `TestServer`
1. If `RequestCulture != null`, set `CultureInfo.CurrentCulture` to that value on the "server side"
1. If `RequestCulture == null`, set `CultureInfo.CurrentCulture` on "the server" to whatever value it had on "the client" (this feels necessary to avoid breaking every loc-related test in user apps until they set `RequestCuluture`... or am I misunderstanding the use case?)

anurse: Another question: If `PreserveExecutionContext == true` and `RequestCulture != null`, should we set `CultureInfo.CurrentCulture`? If we do, it will actually end up affecting the calling code, since we're in the same execution context.

davidfowl: I wouldn't add the culture until people asked for it. Lets do nothing here until somebody asks. Today its already possible with middleware in the pipeline (or a startup filter).

anurse: Ok with me. Our tests don't seem to be impacted at all by *not* flowing the culture. @rynowak ?

rynowak: I'm willing to just see what breaks then. If nothing broke when you tried this maybe we don't use it and we already do this in a more correct way.

src/Hosting/TestHost/src/TestServer.cs
anurse: ✔️ Also updating the other comment to move default value to `<summary>` since `<remarks>` don't show in IntelliSense (which is where you really want this info IMO).

src/Hosting/TestHost/src/HttpContextBuilder.cs
```diff
@@ -59,15 +60,18 @@ internal void Configure(Action<HttpContext> configureContext)
         /// Start processing the request.
         /// </summary>
         /// <returns></returns>
-        internal Task<HttpContext> SendAsync(CancellationToken cancellationToken)
+        internal Task<HttpContext> SendAsync(CancellationToken cancellationToken, bool preserveExecutionContext)
```
Tratcher: preserveExecutionContext should be a property like AllowSynchronousIO.

Tratcher (approved):

src/Hosting/TestHost/src/HttpContextBuilder.cs
anurse: I disagree. It's only used in this method and all the callers have a value they can provide here. `AllowSynchronousIO` is only a property because `HttpContextBuilder` implements `IHttpBodyControlFeature` and requires that the property be provided. In fact, no caller uses the setter on `HttpContextBuilder`, they only use the constructor argument.

Tratcher: And why not make it a constructor argument?

anurse: 🤷‍♀️

anurse:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=184916

aspnet-hello:

Amd3202:
I'm having the same/similar issue but i don't  think it's with TestServer. I have a .netstandard 2.0 class library which I have an IHttpContextAccessor. Reading and writing to the session works fine. Then i call a SOAP service through   Microsoft.VisualStudio.ConnectedService.Wcf in the class library the session gets (mostly) wiped out. Any key set before calling the webservice persists but the values are wiped out. I'm including the lib in the .netcore webapp Using 2.2 any advice would be greatly appreciated!

anurse:
@Amd3202 If you're having an issue or have a question I'd suggest filing a new issue and including a sample project that reproduces the issue. Otherwise it's likely we'll lose track of this thread since this is a merged PR. Open issues will show up in our triage process and we'll be able to track it better.

Amd3202:
Thanks for the quick reply, i thought it was related... haven't opened a thread on here before i'll see if i can figure that out! Thanks.

anurse:
Even if it is related, it's easier for us to track in a new open issue rather than as a comment on a closed issue or merged PR :)

Issue #526, opened 2015-05-01
nuget.org feed not updated with beta4 packages


conniey:
Hi,

I was looking through the stable release of ASP.NET 5 on nuget.org and noticed that the following packages were still on beta3 even though other packages were on beta4.  Is this something intentional?  The beta4 packages were available on your myget feed though, so I am a little confused.
* http://www.nuget.org/packages/Microsoft.AspNet.Security.Cookies/1.0.0-beta3
* http://www.nuget.org/packages/Microsoft.AspNet.Security.OpenIdConnect/1.0.0-beta3

davidfowl:
Those got renamed  `Microsoft.AspNet.Authentication.*`

Eilon:
Added an announcement: https://github.com/aspnet/Announcements/issues/12

PR #10096, opened 2019-05-08
HeaderPropagation Middleware: configuration per client

alefranz:
This is a follow up of #7921 and #9793 (related to #526).
The goal is to start a discussion with @rynowak (and who else is interested) on a possible approach to allow to configure the behavior per client.

The idea is that
- by default a new client will get all the propagated headers as per current behavior:
```.AddHeaderPropagation()```
- but you can choose to switch to the opt-in model, where only the specified headers are propagated:
```.AddHeaderPropagation(options => options.Headers.Add("X-BetaFeatures"))```
- eventually redefining the outbound header name
```.AddHeaderPropagation(options => options.Headers.Add("X-BetaFeatures", "X-Experiments"))```

After a few experiments I opted for splitting the configuration between the Middleware and the Handler; this feels more natural than evening a shared config but each part consuming only a portion of it.
I ended up renaming the outbound header name in the middleware as `captured` as it is the name used to store the value in the state. However this is only an internal detail and it doesn't affect the bahaviour improved by Ryan in #9793.

Questions:
- What do you think about this behavior and API design?
- The current implementation to provide the `Options` object to the Handler it's intentionally just a POC, however I am not sure what would be the recommended approach to have a different `Configure` of options per consumer.
- I am not able to run the sample project. Any idea on a possible reason for a missing registration of `Microsoft.AspNetCore.Hosting.Server.IServer` while activating the `WebHost`?

**Note: I am aware the XML docs is incorrect and it probably could use a few more tests. I will address these once we agree on the design**

Thank you,
 Alessio

### CODE REVIEW

**Comment by rynowak** [comment]
Sorry for the delay. Will take a look at this today.

File: src/Middleware/HeaderPropagation/src/DependencyInjection/HeaderPropagationHttpClientBuilderExtensions.cs [hunk 1]
```diff
@@ -22,7 +23,48 @@ public static IHttpClientBuilder AddHeaderPropagation(this IHttpClientBuilder bu
...
+                var middlewareOptions = services.GetRequiredService<IOptions<HeaderPropagationOptions>>();
+                for (int i = 0; i < middlewareOptions.Value.Headers.Count; i++)
+                {
+                    var header = middlewareOptions.Value.Headers[i];
+                    options.Headers.Add(header.CapturedHeaderName, header.CapturedHeaderName);
+                }
+                return new HeaderPropagationMessageHandler(Options.Options.Create(options), services.GetRequiredService<HeaderPropagationValues>());
+            });
+
+            return builder;
```
rynowak: What you want to do here is base this on named options using `builder.Name` as the key. https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-2.2#named-options-support-with-iconfigurenamedoptions

src/Middleware/HeaderPropagation/src/DependencyInjection/HeaderPropagationHttpClientBuilderExtensions.cs
```diff
@@ -22,7 +23,48 @@ public static IHttpClientBuilder AddHeaderPropagation(this IHttpClientBuilder bu
...
+            {
+                throw new ArgumentNullException(nameof(configureOptions));
+            }
+
+            builder.Services.AddHeaderPropagation();
+
+            builder.AddHttpMessageHandler(services=>
+            {
+                var options = new HeaderPropagationClientOptions();
+                configureOptions(options);
```
rynowak: This could be replaced with a call to `builder.Services.Configure<HeaderPropagationClientOptions>(builder.Name, configureOptions)`. 

Then you retrieve the value with `IOptionsMonitor<HeaderPropagationClientOptions>::Get(builder.Name)`.

rynowak:
I think the overall design of this works. 

> The idea is that by default a new client will get all the propagated headers as per current behavior:
>   .AddHeaderPropagation()
>    but you can choose to switch to the opt-in model, where only the specified headers are propagated:
>   .AddHeaderPropagation(options => options.Headers.Add("X-BetaFeatures"))

I think this is the right starting point. You want the default to have the most simple and obvious behavior. Next you want to give the ability to have full control (explicitly specify the headers).

This is the same pattern as middleware btw. Using no-args give you a default set of options that play with DI, and using an overload that takes options give you full control.

>    eventually redefining the outbound header name
>    .AddHeaderPropagation(options => options.Headers.Add("X-BetaFeatures", "X-Experiments"))

This is the part I want to resolve. 

We have the ability to *map* a header to another name during capturing, and now we have the same feature in the message handler. Having two ways to do it makes things harder to explain. I don't think it would be the end of the world, but it would simplify things if we could remove the complexity from one of these places.

Your thoughts?

rynowak:
Here's a commit that fixes the sample (without IIS) 020d5ce1735be8120442b17023268ddfca8a02ea

alefranz:
Hi Ryan,
thanks for taking the time to look into this draft.

If we want to support the ability of using different header names per multiple clients, I think we need to move the configuration of the name to the clients.
You can't work around it defining multiple entries using a different name in the case of using a non-deterministic value factory. (well, one options would be to allow to specify multiple names per the same header, to then pick in each client which one to keep), but it would be really confusing).

#### Moving the outbound name setup to the client configuration

The happy path (like the sample you included in the solution) would stay the same as you are forwarding a header without renaming.
If you want to rename the header it will happen in the client configuration:
```
client.AddHeaderPropagation(options => options.Headers.Add("X-BetaFeatures", "X-Experiments"))
```
However, this means that if you need to rename a header and apply it to every client you will have to specify the configuration for every client.
I don't think this is a big deal as you can easily factor it out.

```
services.AddHeaderPropagation(options =>
  options.Headers.Add("X-SpanId", "X-ParentSpanId")
);
```
will become 
```
services.AddHeaderPropagation(options =>
  options.Headers.Add("X-SpanId")
);
services.AddHttpClient("test"),AddHeaderPropagation(options =>
  options.Headers.Add("X-SpanId", "X-ParentSpanId")
);
```
Which makes me a bit sad.


#### Keeping the configuration in two places, but always refer to it with the inbound name

Another approach would be to keep the configuration in the two places, like in this PR, but change it back so that the header name defined in the middleware is the outbound name again, and in the client you eventually override it, but referring to the header with the inbound name.
So instead of the current behavior of this PR:
```
services.AddHeaderPropagation(options =>
  options.Headers.Add("A", "B")
);
services.AddHttpClient("course"),AddHeaderPropagation(options =>
  options.Headers.Add("B")  // the value in A will be propagated as B
);
services.AddHttpClient("student"),AddHeaderPropagation(options =>
  options.Headers.Add("B", C")  // the value in A will be propagated as C, quite confusing
);
```
changing to:
```
services.AddHeaderPropagation(options =>
  options.Headers.Add("A", "B")
);
services.AddHttpClient("course"),AddHeaderPropagation(options =>
  options.Headers.Add("A")  // the value in A will be propagated as B, this would be the confusing one
);
services.AddHttpClient("student"),AddHeaderPropagation(options =>
  options.Headers.Add("A", C")  // the value in A will be propagated as C
);
```
but it just move the problem.

*On top of this problem, I also don't like the need to have to specify the configuration in two different places, but I came to terms with it.
I would have liked to configure only the clients (as that give maximum flexibility) and then have the middleware automatically collect all the configurations to detect which headers need to be retrieved upfront, but that feels against the philosophy of AspNetCore and introduce a plethora of tricky scenarios to handle.*

Do you see another option?

Thank you,
 Alessio

rynowak:
After thinking about it a bit more and looking through some examples I think I like the design that's here best of these options. I think the examples you've pointed out are important.


 I'm going to go through this and start leaving some more detailed comments.

src/Middleware/HeaderPropagation/samples/HeaderPropagationSample/Startup.cs
```diff
@@ -71,19 +74,23 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env, IHttpCli
                         await context.Response.WriteAsync($"'/' Got Header '{header.Key}': {string.Join(", ", header.Value)}\r\n");
                     }
 
-                    await context.Response.WriteAsync("Sending request to /forwarded\r\n");
+                    var clientNames = new[] { "test", "anotehr" };
```
rynowak: typo `another`

src/Middleware/HeaderPropagation/src/DependencyInjection/HeaderPropagationHttpClientBuilderExtensions.cs
```diff
@@ -22,7 +23,48 @@ public static IHttpClientBuilder AddHeaderPropagation(this IHttpClientBuilder bu
 
             builder.Services.AddHeaderPropagation();
 
-            builder.AddHttpMessageHandler<HeaderPropagationMessageHandler>();
+            builder.AddHttpMessageHandler(services =>
+            {
+                var options = new HeaderPropagationClientOptions();
+                var middlewareOptions = services.GetRequiredService<IOptions<HeaderPropagationOptions>>();
+                for (int i = 0; i < middlewareOptions.Value.Headers.Count; i++)
```
rynowak: use var.

src/Middleware/HeaderPropagation/src/DependencyInjection/HeaderPropagationHttpClientBuilderExtensions.cs
```diff
@@ -22,7 +23,48 @@ public static IHttpClientBuilder AddHeaderPropagation(this IHttpClientBuilder bu
...
+                throw new ArgumentNullException(nameof(configureOptions));
+            }
+
+            builder.Services.AddHeaderPropagation();
+
+            builder.AddHttpMessageHandler(services=>
+            {
+                var options = new HeaderPropagationClientOptions();
+                configureOptions(options);
+                return new HeaderPropagationMessageHandler(Options.Options.Create(options), services.GetRequiredService<HeaderPropagationValues>());
```
rynowak: Based on the usage here, there probably isn't much of a reason to use `IOptions<>` at all for `HeaderPropagationClientOptions` - in both cases we're not getting the options from DI. We could try to change it so that it does play with `IOptions<>` in DI, but it's going to be messy and I'm not sure what it helps.

src/Middleware/HeaderPropagation/src/HeaderPropagationClientOptions.cs
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.HeaderPropagation
+{
+    /// <summary>
+    /// Provides configuration for the <see cref="HeaderPropagationMiddleware"/>.
+    /// </summary>
+    public class HeaderPropagationClientOptions
```
rynowak: `HeaderPropagationMessageHandlerOptions`?

src/Middleware/HeaderPropagation/src/HeaderPropagationClientOptions.cs
```diff
@@ -0,0 +1,22 @@
...
+        /// <summary>
+        /// Gets or sets the headers to be captured by the <see cref="HeaderPropagationMiddleware"/>
+        /// and to be propagated by the <see cref="HeaderPropagationMessageHandler"/>.
+        /// </summary>
+        /// <remarks>
+        /// Entries in <see cref="Headers"/> are processes in order while capturing headers inside
+        /// <see cref="HeaderPropagationMiddleware"/>. This can cause an earlier entry to take precedence
+        /// over a later entry if they have the same <see cref="HeaderPropagationEntry.CapturedHeaderName"/>.
+        /// </remarks>
+        public HeaderPropagationClientEntryCollection Headers { get; set; } = new HeaderPropagationClientEntryCollection();
```
rynowak: Docs are wrong 😆 - I know you already mentioned it was a copy-paste.

src/Middleware/HeaderPropagation/src/HeaderPropagationEntry.cs
```diff
@@ -13,34 +13,34 @@ public class HeaderPropagationEntry
...
+        ///  <param name="capturedHeaderName">
         /// The name of the header to be added by <see cref="HeaderPropagationMessageHandler"/>.
         /// </param>
         /// <param name="valueFilter">
         /// A filter delegate that can be used to transform the header value. May be null.
         /// </param>
         public HeaderPropagationEntry(
             string inboundHeaderName,
-            string outboundHeaderName,
+            string capturedHeaderName,
```
rynowak: 👍 - I was going to suggest this.

rynowak:
I'm ready for this to no longer be a draft 👍

src/Middleware/HeaderPropagation/samples/HeaderPropagationSample/Startup.cs
alefranz: Took me a while to find why the sample wasn't working as expected 😅

src/Middleware/HeaderPropagation/src/DependencyInjection/HeaderPropagationHttpClientBuilderExtensions.cs
alefranz: Is there a way to enable Visual Studio to highlight when certain code fixes are available? (I'm too used to ReSharper at work)

src/Middleware/HeaderPropagation/src/DependencyInjection/HeaderPropagationHttpClientBuilderExtensions.cs
alefranz: Agree. It would need something like AddHttpMessageHandler<> to automatically inject the builder name if the constructor accept a string to doing it cleanly.

src/Middleware/HeaderPropagation/src/HeaderPropagationClientOptions.cs
alefranz: What about `HeaderPropagationOptions`? is it ok to leave it like this as in the happy path you only deal with it? or should it be renamed to `HeaderPropagationMiddlewareOptions`?

alefranz:
I will take another look at this tomorrow with a fresh pair of eyes and remove the draft status once it is ready for review.

src/Middleware/HeaderPropagation/src/HeaderPropagationMessageHandlerEntry.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// </summary>
+    public class HeaderPropagationMessageHandlerEntry
+    {
+        /// <summary>
+        /// Creates a new <see cref="HeaderPropagationMessageHandlerEntry"/> with the provided <paramref name="capturedHeaderName"/>
+        /// and <paramref name="outboundHeaderName"/>.
+        /// </summary>
+        /// <param name="capturedHeaderName">
+        /// The name of the header to be used to lookup the headers captured by the <see cref="HeaderPropagationMiddleware"/>.
+        /// </param>
```
alefranz: Is this clear enough?

src/Middleware/HeaderPropagation/samples/HeaderPropagationSample/Startup.cs
```diff
@@ -49,6 +49,9 @@ public void ConfigureServices(IServiceCollection services)
                 .AddHttpClient("test")
                 .AddHeaderPropagation();
 
+            services
+                .AddHttpClient("another")
+                .AddHeaderPropagation(options => options.Headers.Add("X-BetaFeatures", "X-Experiments"));
```
alefranz: Do I also need more convoluted examples for all the possible scenarios we can handle now?
Or is it better to keep it nice and simple?

src/Middleware/HeaderPropagation/src/DependencyInjection/HeaderPropagationHttpClientBuilderExtensions.cs
```diff
@@ -11,6 +12,9 @@ public static class HeaderPropagationHttpClientBuilderExtensions
         /// <summary>
         /// Adds a message handler for propagating headers collected by the <see cref="HeaderPropagationMiddleware"/> to a outgoing request.
         /// </summary>
+        /// <remarks>
+        /// The headers will be named using the value specified in <see cref="HeaderPropagationEntry.CapturedHeaderName"/>.
```
alefranz: Does this need to go into more details?

src/Middleware/HeaderPropagation/src/HeaderPropagationMessageHandler.cs
```diff
@@ -17,23 +16,17 @@ namespace Microsoft.AspNetCore.HeaderPropagation
...
         {
-            if (options == null)
-            {
-                throw new ArgumentNullException(nameof(options));
-            }
-
-            _options = options.Value;
-
+            _options = options ?? throw new ArgumentNullException(nameof(options));
             _values = values ?? throw new ArgumentNullException(nameof(values));
```
alefranz: not sure if using a throw expression goes against the coding style

alefranz (commented):
Hi Ryan, this is now ready for review.

src/Middleware/HeaderPropagation/samples/HeaderPropagationSample/Startup.cs
rynowak: No, I really think we should keep the examples simple and approachable.

src/Middleware/HeaderPropagation/src/DependencyInjection/HeaderPropagationHttpClientBuilderExtensions.cs
rynowak: I think this comment could be left out. It doesn't really explain what's going on in a clear way.

rynowak: Actually I would change it to say something.

```
When using this method, all configured headers will be applied to outgoing HTTP requests.
```

src/Middleware/HeaderPropagation/src/DependencyInjection/HeaderPropagationHttpClientBuilderExtensions.cs
```diff
@@ -22,7 +26,49 @@ public static IHttpClientBuilder AddHeaderPropagation(this IHttpClientBuilder bu
...
+            });
+
+            return builder;
+        }
+
+        /// <summary>
+        /// Adds a message handler for propagating headers collected by the <see cref="HeaderPropagationMiddleware"/> to a outgoing request,
+        /// explicitly specifying which headers to propagate.
+        /// </summary>
+        /// <remarks>This also allows to redefine the name to use for a header in the outgoing request.</remarks>
```
rynowak: suggestion:
```
When using this method, the `HeaderPropagationMessageHandlerOptions` will determine which headers will be applied to outgoing HTTP requests.
```

src/Middleware/HeaderPropagation/src/DependencyInjection/HeaderPropagationHttpClientBuilderExtensions.cs
```diff
@@ -22,7 +26,49 @@ public static IHttpClientBuilder AddHeaderPropagation(this IHttpClientBuilder bu
...
+        /// <param name="builder">The <see cref="IHttpClientBuilder"/> to add the message handler to.</param>
+        /// <param name="configureOptions">A delegate used to configure the <see cref="HeaderPropagationMessageHandlerOptions"/>.</param>
+        /// <returns>The <see cref="IHttpClientBuilder"/> so that additional calls can be chained.</returns>
+        public static IHttpClientBuilder AddHeaderPropagation(this IHttpClientBuilder builder, Action<HeaderPropagationMessageHandlerOptions> configureOptions)
```
rynowak: nit: change parameter to `configure`

src/Middleware/HeaderPropagation/src/DependencyInjection/HeaderPropagationHttpClientBuilderExtensions.cs
```diff
@@ -22,7 +26,49 @@ public static IHttpClientBuilder AddHeaderPropagation(this IHttpClientBuilder bu
...
+            }
+
+            if (configureOptions == null)
+            {
+                throw new ArgumentNullException(nameof(configureOptions));
+            }
+
+            builder.Services.AddHeaderPropagation();
+
+            builder.AddHttpMessageHandler(services=>
```
rynowak: nit: space

src/Middleware/HeaderPropagation/src/HeaderPropagationMessageHandler.cs
rynowak: this is fine!

src/Middleware/HeaderPropagation/src/HeaderPropagationMessageHandlerEntry.cs
rynowak: Yes, I think so.

rynowak (changes requested):
This overall looks really good. Just a few small things to clean up.

alefranz:
Updated! Thanks

rynowak (approved):

rynowak:
@aspnet-hello  triage https://dev.azure.com/dnceng/public/_build/results?buildId=191755&view=logs

rynowak:
Merged. Thanks @alefranz

alefranz:
Cheers

PR #10099, opened 2019-05-09
Remove CancelableEnumerator

### LINKED ISSUE #7960: Remove CancelableAsyncEnumerable

Instead iterate the IAsyncEnumerable manually using GetAsyncEnumerator() and a while loop to avoid an allocation. See https://github.com/aspnet/AspNetCore/pull/6791#discussion_r259196518.

mikaelm12:
Fixes: https://github.com/aspnet/AspNetCore/issues/7960
Context: https://github.com/aspnet/AspNetCore/pull/6791#discussion_r259196518

### DIFF
--- a/src/SignalR/common/Shared/AsyncEnumerableAdapters.cs
+++ b/src/SignalR/common/Shared/AsyncEnumerableAdapters.cs
@@ -71,10 +71,6 @@ namespace Microsoft.AspNetCore.SignalR.Internal
 
             public IAsyncEnumerator<object> GetAsyncEnumerator(CancellationToken cancellationToken = default)
             {
-                // Assume that this will be iterated through with await foreach which always passes a default token.
-                // Instead use the token from the ctor.
-                Debug.Assert(cancellationToken == default);
-
                 var enumeratorOfT = _asyncEnumerable.GetAsyncEnumerator(_cancellationToken);
                 return enumeratorOfT as IAsyncEnumerator<object> ?? new BoxedAsyncEnumerator(enumeratorOfT);
             }
--- a/src/SignalR/common/Shared/AsyncEnumerableAdapters.cs
+++ b/src/SignalR/common/Shared/AsyncEnumerableAdapters.cs
@@ -51,36 +51,9 @@ namespace Microsoft.AspNetCore.SignalR.Internal
                     {
                         ((CancellationTokenSource)ctsState).Cancel();
                     }, _cts);
-
-                    return new CancelableEnumerator<TResult>(_asyncEnumerable.GetAsyncEnumerator(), registration);
                 }
 
                 return enumerator;
-            }
-
-            private class CancelableEnumerator<T> : IAsyncEnumerator<T>
-            {
-                private IAsyncEnumerator<T> _asyncEnumerator;
-                private readonly CancellationTokenRegistration _cancellationTokenRegistration;
-
-                public T Current => (T)_asyncEnumerator.Current;
-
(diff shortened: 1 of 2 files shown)


BrennanConroy:
Hmm, I'll give you change to review your own PR before commenting.

src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs
```diff
@@ -406,10 +406,11 @@ async Task ExecuteInvocation()
 
             try
             {
-                await foreach (var streamItem in enumerable)
+                var enumerator = enumerable.GetAsyncEnumerator(streamCts.Token);
```
campersau: Should the `enumerator` be disposed?

BrennanConroy:
We should make `CancelableTypedAsyncEnumerable` have a factory method that returns an `IAsyncEnumberable` that way we can avoid allocating the class if the token isn't cancelable.

src/SignalR/common/Shared/AsyncEnumerableAdapters.cs
```diff
@@ -51,37 +51,10 @@ public IAsyncEnumerator<TResult> GetAsyncEnumerator(CancellationToken cancellati
                     {
```
DAllanCarr: `registration` not referenced

BrennanConroy (approved):

src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs
davidfowl: Leave the await foreach and call .WithCancellation instead:

```C#
await foreach (var streamItem in enumerable.WithCancellation(streamCts.Token))
{
}
```

mikaelm12: Yes it should

src/SignalR/common/Shared/AsyncEnumerableAdapters.cs
mikaelm12: The class is still needed because we need to be able to cancel the inner stream operation from the token passed into `GetASyncEnumerator` but you bring up a good point about the registration. It should be disposed

mikaelm12:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=184617

aspnet-hello:

mikaelm12:
So in another PR we should work to remove more from the `AsyncEnumerableAdapters` class. There should be support for the [EnumeratorCancellation] parameter attribute which combines the tagged tokens passed into your `IAsyncEnumerable` returning methods and a token passed into `GetAsyncEnumerator`

src/SignalR/common/Shared/AsyncEnumerableAdapters.cs
```diff
@@ -98,10 +71,6 @@ public CancelableAsyncEnumerable(IAsyncEnumerable<T> asyncEnumerable, Cancellati
 
             public IAsyncEnumerator<object> GetAsyncEnumerator(CancellationToken cancellationToken = default)
             {
-                // Assume that this will be iterated through with await foreach which always passes a default token.
-                // Instead use the token from the ctor.
-                Debug.Assert(cancellationToken == default);
```
halter73: @mikaelm12 Why remove this Debug.Assert if we're still not using the cancellationToken being passed in? Don't we now need to link the token sources or assert the cancellationToken is not cancellable? Otherwise we're just ignoring it...

PR #10102, opened 2019-05-09
Betterize Hosting Log

### LINKED ISSUE #10097: Can the HostingRequestFinishedLog.ToString include the request url too?
[HostingRequestFinishedLog.ToString ](https://github.com/aspnet/AspNetCore/blob/master/src/Hosting/Hosting/src/Internal/HostingRequestFinishedLog.cs#L47) currently just has the elapsedTime, but not the request URL.

> "Request finished in {elapsedTime}ms {statusCode} {ContentType}",  

It is easier to match with the Request starting if it can include the URL like this.

> Request starting HTTP/1.1 GET http://myserver/myapi/myitem/123 [ 16984:92 ]
> Request finished HTTP/1.1 GET http://myserver/myapi/myitem/123 in 501.525ms 200 application/json; charset=utf-8 [ 16984:98 ]
> 

In my log files, sometimes it has a few Request finished around the same time and I cannot tell which one is which quickly without spending time to trace back the log.

This HostingRequestFinishedLog class is internal, I cannot find an easy way I can customize it to provide a custom ToString.

The HostingRequestStartingLog and HostingRequestFinishedLog are run in different thread, I cannot use the threadId to match the log.

### LINKED ISSUE #10556: HostingRequestStartingLog shouldn't capture HttpRequest for its values

### LINKED ISSUE #10557: HostingRequestFinishedLog shouldn't capture HttpContext for its values

benaadams:
Output something more useful for `HostingRequestFinishedLog` (which can get intermixed when multiple parallel requests); so it becomes the more interesting event given context; since its kinda junky currently unless you are also outputting full scopes (even then you need to do matching).

Before
```
info: Microsoft.AspNetCore.Hosting.Diagnostics[1]
      Request starting HTTP/1.1 GET http://ben-laptop/
info: Microsoft.AspNetCore.Hosting.Diagnostics[2]
      Request finished in 31.180400000000002ms 404
info: Microsoft.AspNetCore.Hosting.Diagnostics[1]
      Request starting HTTP/1.1 GET http://ben-laptop/index.html
info: Microsoft.AspNetCore.Hosting.Diagnostics[2]
      Request finished in 1.6021ms 200 text/html
info: Microsoft.AspNetCore.Hosting.Diagnostics[1]
      Request starting HTTP/1.1 GET http://ben-laptop/js/aoa.avatar.js
info: Microsoft.AspNetCore.Hosting.Diagnostics[2]
      Request finished in 2.6724ms 200 application/octet-stream
info: Microsoft.AspNetCore.Hosting.Diagnostics[1]
      Request starting HTTP/1.1 GET http://ben-laptop:26347/textures/spacedrop2048_right1.dds
info: Microsoft.AspNetCore.Hosting.Diagnostics[1]
      Request starting HTTP/1.1 GET http://ben-laptop:26347/textures/spacedrop2048_left2.dds
info: Microsoft.AspNetCore.Hosting.Diagnostics[1]
      Request starting HTTP/1.1 GET http://ben-laptop:26347/textures/spacedrop2048_top3.dds
info: Microsoft.AspNetCore.Hosting.Diagnostics[1]
      Request starting HTTP/1.1 GET http://ben-laptop:26347/textures/spacedrop2048_bottom4.dds
info: Microsoft.AspNetCore.Hosting.Diagnostics[1]
      Request starting HTTP/1.1 GET http://ben-laptop:26347/textures/spacedrop2048_back6.dds
info: Microsoft.AspNetCore.Hosting.Diagnostics[1]
      Request starting HTTP/1.1 GET http://ben-laptop:26347/textures/spacedrop2048_front5.dds
info: Microsoft.AspNetCore.Hosting.Diagnostics[2]
      Request finished in 46.6755ms 200 application/octet-stream
info: Microsoft.AspNetCore.Hosting.Diagnostics[2]
      Request finished in 2.5938000000000003ms 200 application/octet-stream
info: Microsoft.AspNetCore.Hosting.Diagnostics[2]
      Request finished in 2.7852ms 200 application/octet-stream
info: Microsoft.AspNetCore.Hosting.Diagnostics[2]
      Request finished in 246.4214ms 200 application/octet-stream
info: Microsoft.AspNetCore.Hosting.Diagnostics[2]
      Request finished in 280.962ms 200 application/octet-stream
info: Microsoft.AspNetCore.Hosting.Diagnostics[2]
      Request finished in 2.2681ms 200 application/octet-stream
```

After
```
info: Microsoft.AspNetCore.Hosting.Diagnostics[1]
      Request starting HTTP/1.1 GET http://ben-laptop/index.html - -
info: Microsoft.AspNetCore.Hosting.Diagnostics[2]
      Request finished HTTP/1.1 GET http://ben-laptop/index.html - - - 200 2348 text/html 76.8050ms
info: Microsoft.AspNetCore.Hosting.Diagnostics[1]
      Request starting HTTP/1.1 GET http://ben-laptop/js/aoa.avatar.js - -
info: Microsoft.AspNetCore.Hosting.Diagnostics[2]
      Request finished HTTP/1.1 GET http://ben-laptop/js/aoa.avatar.js - - - 200 39852 application/octet-stream 8.6488ms
info: Microsoft.AspNetCore.Hosting.Diagnostics[1]
      Request starting HTTP/1.1 POST http://ben-laptop/form application/x-www-form-urlencoded 19
info: Microsoft.AspNetCore.Hosting.Diagnostics[2]
      Request finished HTTP/1.1 POST http://ben-laptop/form application/x-www-form-urlencoded 19 - 404 0 - 8.1129ms
```

Also improve the more detailed semantic logged output

Before
image

After
image

Resolves #10097
Resolves https://github.com/aspnet/AspNetCore/issues/10556
Resolves https://github.com/aspnet/AspNetCore/issues/10557

/cc @davidfowl

### DIFF
--- a/src/Hosting/Hosting/src/Internal/HostingApplication.cs
+++ b/src/Hosting/Hosting/src/Internal/HostingApplication.cs
@@ -114,6 +114,7 @@ namespace Microsoft.AspNetCore.Hosting
             public HttpContext HttpContext { get; set; }
             public IDisposable Scope { get; set; }
             public Activity Activity { get; set; }
+            internal HostingRequestStartingLog StartLog { get; set; }
 
             public long StartTimestamp { get; set; }
             internal bool HasDiagnosticListener { get; set; }
@@ -125,6 +126,7 @@ namespace Microsoft.AspNetCore.Hosting
 
                 Scope = null;
                 Activity = null;
+                StartLog = null;
 
                 StartTimestamp = 0;
                 HasDiagnosticListener = false;
--- a/src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs
+++ b/src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs
@@ -187,7 +187,7 @@ namespace Microsoft.AspNetCore.Hosting
             // IsEnabled isn't checked in the caller, startTimestamp > 0 is used as a fast proxy check
             // but that may be because diagnostics are enabled, which also uses startTimestamp,
             // so check if we logged the start event
-            if (context.StartLog is object)
+            if (context.StartLog != null)
             {
                 var elapsed = new TimeSpan((long)(TimestampToTicks * (currentTimestamp - startTimestamp)));
 
--- a/src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs
+++ b/src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs
@@ -78,7 +78,7 @@ namespace Microsoft.AspNetCore.Hosting
                     }
 
                     // Non-inline
-                    LogRequestStarting(httpContext);
+                    LogRequestStarting(context);
                 }
             }
             context.StartTimestamp = startTimestamp;
@@ -97,7 +97,7 @@ namespace Microsoft.AspNetCore.Hosting
             {
                 currentTimestamp = Stopwatch.GetTimestamp();
                 // Non-inline
-                LogRequestFinished(httpContext, startTimestamp, currentTimestamp);
+                LogRequestFinished(context, startTimestamp, currentTimestamp);
             }
 
             if (_diagnosticListener.IsEnabled())
@@ -167,30 +167,34 @@ namespace Microsoft.AspNetCore.Hosting
         }
 
         [MethodImpl(MethodImplOptions.NoInlining)]
-        private void LogRequestStarting(HttpContext httpContext)
+        private void LogRequestStarting(HostingApplication.Context context)
         {
             // IsEnabled is checked in the caller, so if we are here just log
+            var startLog = new HostingRequestStartingLog(context.HttpContext);
+            context.StartLog = startLog;
+
             _logger.Log(
                 logLevel: LogLevel.Information,
                 eventId: LoggerEventIds.RequestStarting,
-                state: new HostingRequestStartingLog(httpContext),
+                state: startLog,
                 exception: null,
                 formatter: HostingRequestStartingLog.Callback);
         }
 
         [MethodImpl(MethodImplOptions.NoInlining)]
-        private void LogRequestFinished(HttpContext httpContext, long startTimestamp, long currentTimestamp)
+        private void LogRequestFinished(HostingApplication.Context context, long startTimestamp, long currentTimestamp)
         {
             // IsEnabled isn't checked in the caller, startTimestamp > 0 is used as a fast proxy check
-            // but that may be because diagnostics are enabled, which also uses startTimestamp, so check here
-            if (_logger.IsEnabled(LogLevel.Information))
+            // but that may be because diagnostics are enabled, which also uses startTimestamp,
+            // so check if we logged the start event
+            if (context.StartLog is object)
             {
                 var elapsed = new TimeSpan((long)(TimestampToTicks * (currentTimestamp - startTimestamp)));
 
                 _logger.Log(
                     logLevel: LogLevel.Information,
                     eventId: LoggerEventIds.RequestFinished,
-                    state: new HostingRequestFinishedLog(httpContext, elapsed),
+                    state: new HostingRequestFinishedLog(context, elapsed),
                     exception: null,
                     formatter: HostingRequestFinishedLog.Callback);
             }
--- a/src/Hosting/Hosting/src/Internal/HostingRequestFinishedLog.cs
+++ b/src/Hosting/Hosting/src/Internal/HostingRequestFinishedLog.cs
@@ -9,51 +9,56 @@ using Microsoft.AspNetCore.Http;
 
 namespace Microsoft.AspNetCore.Hosting
 {
+    using static HostingRequestStartingLog;
+
     internal class HostingRequestFinishedLog : IReadOnlyList<KeyValuePair<string, object>>
     {
         internal static readonly Func<object, Exception, string> Callback = (state, exception) => ((HostingRequestFinishedLog)state).ToString();
 
-        private readonly HttpContext _httpContext;
-        private readonly TimeSpan _elapsed;
+        private readonly HostingApplication.Context _context;
 
         private string _cachedToString;
+        public TimeSpan Elapsed { get; }
 
-        public int Count => 3;
+        public int Count => 11;
 
         public KeyValuePair<string, object> this[int index]
         {
             get
             {
-                switch (index)
+                var request = _context.HttpContext.Request;
+                var response = _context.HttpContext.Response;
+
+                return index switch
                 {
-                    case 0:
-                        return new KeyValuePair<string, object>("ElapsedMilliseconds", _elapsed.TotalMilliseconds);
-                    case 1:
-                        return new KeyValuePair<string, object>("StatusCode", _httpContext.Response.StatusCode);
-                    case 2:
-                        return new KeyValuePair<string, object>("ContentType", _httpContext.Response.ContentType);
-                    default:
-                        throw new IndexOutOfRangeException(nameof(index));
-                }
+                    0 => new KeyValuePair<string, object>("ElapsedMilliseconds", Elapsed.TotalMilliseconds),
+                    1 => new KeyValuePair<string, object>(nameof(response.StatusCode), response.StatusCode),
+                    2 => new KeyValuePair<string, object>(nameof(response.ContentType), response.ContentType),
+                    3 => new KeyValuePair<string, object>(nameof(response.ContentLength), response.ContentLength),
+                    4 => new KeyValuePair<string, object>(nameof(request.Protocol), request.Protocol),
+                    5 => new KeyValuePair<string, object>(nameof(request.Method), request.Method),
+                    6 => new KeyValuePair<string, object>(nameof(request.Scheme), request.Scheme),
+                    7 => new KeyValuePair<string, object>(nameof(request.Host), request.Host.ToString()),
+                    8 => new KeyValuePair<string, object>(nameof(request.PathBase), request.PathBase.ToString()),
+                    9 => new KeyValuePair<string, object>(nameof(request.Path), request.Path.ToString()),
+                    10 => new KeyValuePair<string, object>(nameof(request.QueryString), request.QueryString.ToString()),
+                    _ => throw new IndexOutOfRangeException(nameof(index)),
+                };
             }
         }
 
-        public HostingRequestFinishedLog(HttpContext httpContext, TimeSpan elapsed)
+        public HostingRequestFinishedLog(HostingApplication.Context context, TimeSpan elapsed)
         {
-            _httpContext = httpContext;
-            _elapsed = elapsed;
+            _context = context;
+            Elapsed = elapsed;
         }
 
         public override string ToString()
         {
             if (_cachedToString == null)
             {
-                _cachedToString = string.Format(
-                    CultureInfo.InvariantCulture,
-                    "Request finished in {0}ms {1} {2}",
-                    _elapsed.TotalMilliseconds,
-                    _httpContext.Response.StatusCode,
-                    _httpContext.Response.ContentType);
+                var response = _context.HttpContext.Response;
+                _cachedToString = $"Request finished {_context.StartLog.ToStringWithoutPreamble()} - {response.StatusCode.ToString(CultureInfo.InvariantCulture)} {ValueOrEmptyMarker(response.ContentLength)} {ValueOrEmptyMarker(response.ContentType)} {Elapsed.TotalMilliseconds.ToString("0.0000", CultureInfo.InvariantCulture)}ms";
             }
 
             return _cachedToString;
--- a/src/Hosting/Hosting/src/Internal/HostingRequestFinishedLog.cs
+++ b/src/Hosting/Hosting/src/Internal/HostingRequestFinishedLog.cs
@@ -38,10 +38,10 @@ namespace Microsoft.AspNetCore.Hosting
                     4 => new KeyValuePair<string, object>(nameof(request.Protocol), request.Protocol),
                     5 => new KeyValuePair<string, object>(nameof(request.Method), request.Method),
                     6 => new KeyValuePair<string, object>(nameof(request.Scheme), request.Scheme),
-                    7 => new KeyValuePair<string, object>(nameof(request.Host), request.Host.ToString()),
-                    8 => new KeyValuePair<string, object>(nameof(request.PathBase), request.PathBase.ToString()),
-                    9 => new KeyValuePair<string, object>(nameof(request.Path), request.Path.ToString()),
-                    10 => new KeyValuePair<string, object>(nameof(request.QueryString), request.QueryString.ToString()),
+                    7 => new KeyValuePair<string, object>(nameof(request.Host), request.Host.Value),
+                    8 => new KeyValuePair<string, object>(nameof(request.PathBase), request.PathBase.Value),
+                    9 => new KeyValuePair<string, object>(nameof(request.Path), request.Path.Value),
+                    10 => new KeyValuePair<string, object>(nameof(request.QueryString), request.QueryString.Value),
                     _ => throw new IndexOutOfRangeException(nameof(index)),
                 };
             }
@@ -58,7 +58,7 @@ namespace Microsoft.AspNetCore.Hosting
             if (_cachedToString == null)
             {
                 var response = _context.HttpContext.Response;
-                _cachedToString = $"Request finished {_context.StartLog.ToStringWithoutPreamble()} - {response.StatusCode.ToString(CultureInfo.InvariantCulture)} {ValueOrEmptyMarker(response.ContentLength)} {ValueOrEmptyMarker(response.ContentType)} {Elapsed.TotalMilliseconds.ToString("0.0000", CultureInfo.InvariantCulture)}ms";
+                _cachedToString = $"Request finished {_context.StartLog.ToStringWithoutPreamble()} - {response.StatusCode.ToString(CultureInfo.InvariantCulture)} {ValueOrEmptyMarker(response.ContentLength)} {EscapedValueOrEmptyMarker(response.ContentType)} {Elapsed.TotalMilliseconds.ToString("0.0000", CultureInfo.InvariantCulture)}ms";
             }
 
             return _cachedToString;
--- a/src/Hosting/Hosting/src/Internal/HostingRequestStartingLog.cs
+++ b/src/Hosting/Hosting/src/Internal/HostingRequestStartingLog.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Globalization;
+using System.Net;
 using Microsoft.AspNetCore.Http;
 
 namespace Microsoft.AspNetCore.Hosting
@@ -69,7 +70,8 @@ namespace Microsoft.AspNetCore.Hosting
             => ToString().Substring(LogPreamble.Length);
 
         internal static string EscapedValueOrEmptyMarker(string potentialValue)
-            => potentialValue?.Length > 0 ? Uri.EscapeDataString(potentialValue) : EmptyEntry;
+            // Encode space as +
+            => potentialValue?.Length > 0 ? potentialValue.Replace(' ', '+') : EmptyEntry;
 
         internal static string ValueOrEmptyMarker<T>(T? potentialValue) where T : struct, IFormattable
             => potentialValue?.ToString(null, CultureInfo.InvariantCulture) ?? EmptyEntry;
--- a/src/Hosting/Hosting/src/Internal/HostingRequestStartingLog.cs
+++ b/src/Hosting/Hosting/src/Internal/HostingRequestStartingLog.cs
@@ -29,10 +29,10 @@ namespace Microsoft.AspNetCore.Hosting
             2 => new KeyValuePair<string, object>(nameof(_request.ContentType), _request.ContentType),
             3 => new KeyValuePair<string, object>(nameof(_request.ContentLength), _request.ContentLength),
             4 => new KeyValuePair<string, object>(nameof(_request.Scheme), _request.Scheme),
-            5 => new KeyValuePair<string, object>(nameof(_request.Host), _request.Host.ToString()),
-            6 => new KeyValuePair<string, object>(nameof(_request.PathBase), _request.PathBase.ToString()),
-            7 => new KeyValuePair<string, object>(nameof(_request.Path), _request.Path.ToString()),
-            8 => new KeyValuePair<string, object>(nameof(_request.QueryString), _request.QueryString.ToString()),
+            5 => new KeyValuePair<string, object>(nameof(_request.Host), _request.Host.Value),
+            6 => new KeyValuePair<string, object>(nameof(_request.PathBase), _request.PathBase.Value),
+            7 => new KeyValuePair<string, object>(nameof(_request.Path), _request.Path.Value),
+            8 => new KeyValuePair<string, object>(nameof(_request.QueryString), _request.QueryString.Value),
             _ => throw new IndexOutOfRangeException(nameof(index)),
         };
 
@@ -46,7 +46,7 @@ namespace Microsoft.AspNetCore.Hosting
             if (_cachedToString == null)
             {
                 var request = _request;
-                _cachedToString = $"{LogPreamble}{request.Protocol} {request.Method} {request.Scheme}://{request.Host}{request.PathBase}{request.Path}{request.QueryString} {ValueOrEmptyMarker(request.ContentType)} {ValueOrEmptyMarker(request.ContentLength)}"; ;
+                _cachedToString = $"{LogPreamble}{request.Protocol} {request.Method} {request.Scheme}://{request.Host.Value}{request.PathBase.Value}{request.Path.Value}{request.QueryString.Value} {EscapedValueOrEmptyMarker(request.ContentType)} {ValueOrEmptyMarker(request.ContentLength)}"; ;
             }
 
             return _cachedToString;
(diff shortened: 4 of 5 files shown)


benaadams:
Also improve the more detailed logged output

Before
image

After
image

Tratcher:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=184378

aspnet-hello:

src/Hosting/Hosting/src/Internal/HostingRequestFinishedLog.cs
```diff
@@ -27,33 +29,49 @@ internal class HostingRequestFinishedLog : IReadOnlyList<KeyValuePair<string, ob
...
         {
             if (_cachedToString == null)
             {
-                _cachedToString = string.Format(
-                    CultureInfo.InvariantCulture,
-                    "Request finished in {0}ms {1} {2}",
-                    _elapsed.TotalMilliseconds,
-                    _httpContext.Response.StatusCode,
-                    _httpContext.Response.ContentType);
+                var logStart = _startLog.ToString().Substring(HostingRequestStartingLog.LogPreamble.Length);
```
halter73: What's the cause with all the space before the dash in the "After:" sample output? Is HostingRequestStartingLog.ToString() returning a lot of trailing spaces?

```
info: Microsoft.AspNetCore.Hosting.Diagnostics[1]
      Request starting HTTP/1.1 GET http://ben-laptop/
info: Microsoft.AspNetCore.Hosting.Diagnostics[2]
      Request finished HTTP/1.1 GET http://ben-laptop/   - 404 0  27.9983ms
info: Microsoft.AspNetCore.Hosting.Diagnostics[1]
      Request starting HTTP/1.1 GET http://ben-laptop/index.html
info: Microsoft.AspNetCore.Hosting.Diagnostics[2]
      Request finished HTTP/1.1 GET http://ben-laptop/index.html   - 200 2348 text/html 51.6009ms
i
```

benaadams: Space separated fields and they are empty.

Is why I `Uri.EscapeUriString` on content type so `application/json; charset=utf-8` would come out `application/json;+charset=utf-8` as IIS does:

image

halter73: Should we trim the string then?

```suggestion
                var logStart = _startLog.ToString().Substring(HostingRequestStartingLog.LogPreamble.Length).TrimEnd(' ');
```

halter73: Or better yet, have HostingRequestStartingLog.ToString() do that.

src/Hosting/Hosting/src/Internal/HostingRequestStartingLog.cs
```diff
@@ -51,25 +60,24 @@ internal class HostingRequestStartingLog : IReadOnlyList<KeyValuePair<string, ob
...
-                    _request.Protocol,
-                    _request.Method,
-                    _request.Scheme,
-                    _request.Host.Value,
-                    _request.PathBase.Value,
-                    _request.Path.Value,
-                    _request.QueryString.Value,
-                    _request.ContentType,
-                    _request.ContentLength);
+                _cachedToString = $"{LogPreamble}{Protocol} {Method} {Scheme}://{Host}{PathBase}{Path}{QueryString} {ContentType} {ContentLength?.ToString(CultureInfo.InvariantCulture) ?? string.Empty}";
```
halter73:
```suggestion
                _cachedToString = $"{LogPreamble}{Protocol} {Method} {Scheme}://{Host}{PathBase}{Path}{QueryString} {ContentType} {ContentLength?.ToString(CultureInfo.InvariantCulture) ?? string.Empty}".TrimEnd(' ');
```

src/Hosting/Hosting/src/Internal/HostingRequestFinishedLog.cs
benaadams: If you did a `string.Split(` `)` on it wouldn't you want the same number of fields in the same positions?

src/Hosting/Hosting/src/Internal/HostingRequestStartingLog.cs
benaadams: Would change the column positions of the fields based on which ones are filled in?

src/Hosting/Hosting/src/Internal/HostingRequestFinishedLog.cs
halter73: I'm not suggesting `string.Split(' ')`, but instead `string.TrimEnd(' ')`.

benaadams: I mean as the consumer of the logs; if the end (in this case the middle) was trimmed then space would no longer be a delimiter as the value in each column would be determined by the preceding fields and not by its position.

e.g. if this were instead csv it would look like

```
Request,finished,HTTP/1.1,GET,http://ben-laptop/textures/spacedrop2048_back6.dds,,,-,200,1445726,application/octet-stream,672.4511ms
```
Collating the commas would be problematic as the values would keep changing columns?

benaadams: e.g. a Post request would look like
```
Request finished HTTP/1.1 POST http://ben-laptop/myform application/json;+charset=utf-8 2567 - 200 1445726 application/octet-stream 672.4511ms
```
Or with commas
```
Request,finished,HTTP/1.1,POST,http://ben-laptop/myform,application/json;+charset=utf-8,2567,-,200,1445726,application/octet-stream,672.4511ms
```

halter73: I hold the principle that object.ToString() (in this case HostingRequestStartingLog.ToString()) should *never* have trailing spaces.

Whatever is interpolating object.ToString() into anther string (e.g. string.Format) should add any necessary surrounding whitespace.

halter73: I do see the point about parsing logs though. If we used commas instead of spaces at least it would be clear that there are empty columns. Now anyone reading the logs just sees what looks like bad formatting.

halter73: But anyone parsing the logs should be using semantic logged output, right?

benaadams: > But anyone parsing the logs should be using semantic logged output, right?

What people do, and what they should do, are often very far apart 😉 

These issues would suggest not always https://github.com/aspnet/AspNetCore/issues/5894, https://github.com/aspnet/AspNetCore/issues/10097

benaadams: Also I do know logging at Info in production is far more common than I thought it was,

halter73: Can we use something other than a space as the column separator? Maybe '|' or ','? We'd have to make sure whatever separator we chose keeps the logs easily readable, doesn't just look like bad formatting, and doesn't conflict with characters that could show up inside a column entry.

For a second I thought a ' ' in the ContentType column might ruin using ' ' as a column separator, but the Content-Type header gets url-encoded.

benaadams: Was looking at the w3c format and it goes with
> If a field is unused in a particular entry dash "-" marks the omitted field.

Might be better?

halter73: Yes! That's much better IMHO.

benaadams: Now outputs:
```
info: Microsoft.AspNetCore.Hosting.Diagnostics[1]
      Request starting HTTP/1.1 GET http://ben-laptop/index.html - -
info: Microsoft.AspNetCore.Hosting.Diagnostics[2]
      Request finished HTTP/1.1 GET http://ben-laptop/index.html - - - 200 2348 text/html 76.8050ms
info: Microsoft.AspNetCore.Hosting.Diagnostics[1]
      Request starting HTTP/1.1 GET http://ben-laptop/js/aoa.avatar.js - -
info: Microsoft.AspNetCore.Hosting.Diagnostics[2]
      Request finished HTTP/1.1 GET http://ben-laptop/js/aoa.avatar.js - - - 200 39852 application/octet-stream 8.6488ms
info: Microsoft.AspNetCore.Hosting.Diagnostics[1]
      Request starting HTTP/1.1 POST http://ben-laptop/form application/x-www-form-urlencoded 19
info: Microsoft.AspNetCore.Hosting.Diagnostics[2]
      Request finished HTTP/1.1 POST http://ben-laptop/form application/x-www-form-urlencoded 19 - 404 0 - 8.1129ms
```

src/Hosting/Hosting/src/Internal/HostingApplication.cs
```diff
@@ -57,12 +56,153 @@ public void DisposeContext(Context context, Exception exception)
...
+                    else
+                    {
+                        // Not null, so create the LoggingContext if it doesn't exist and set the value.
+                        LoggingContext.StartLog = value;
+                    }
+                }
+            }
+
+            // Get or create the _loggingContext
+            private LogContext LoggingContext => _loggingContext ??= new LogContext();
```
halter73: `??=` @pranavkm How many points for that?

src/Hosting/Hosting/src/Internal/HostingRequestFinishedLog.cs
```diff
@@ -1,24 +1,28 @@
...
 
 using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Globalization;
-using Microsoft.AspNetCore.Http;
 
 namespace Microsoft.AspNetCore.Hosting.Internal
 {
+    using static HostingRequestStartingLog;
```
halter73: More @pranavkm points 😄 I see this was introduced "way back" in C# 6, but I haven't seen this used much.

halter73 (approved):

src/Hosting/Hosting/src/Internal/HostingApplication.cs
pranavkm: 🥇

src/Hosting/Hosting/src/Internal/HostingApplication.cs
```diff
@@ -57,12 +56,153 @@ public void DisposeContext(Context context, Exception exception)
...
             public HttpContext HttpContext { get; set; }
-            public IDisposable Scope { get; set; }
-            public long StartTimestamp { get; set; }
-            public bool EventLogEnabled { get; set; }
-            public Activity Activity { get; set; }
-            internal bool HasDiagnosticListener { get; set; }
+            public IDisposable Scope
+            {
+                // Get set value or return null
+                get => _loggingContext?.Scope;
```
benaadams: I was going to make this `readonly get` but the version of C# AspNetCore uses is too old. 

I think updating it will make Code check go crazy as every auto-property getter gets the `[Readonly]` attribute

src/Hosting/Hosting/src/Internal/HostingApplication.cs
halter73: My remaining question is how often a Context is constructed without setting any of these properties to a non-default value.

This does seem like a *ton* of work to turn six properties into one lazy-allocated field with six properties.

Also, does this need to be thread safe? I hope not because `new LogContext()` could run multiple times if LoggingContext is accessed and initialized twice in a parallel.

benaadams: > My remaining question is how often a Context is constructed without setting any of these properties to a non-default value.

If you don't want an  `Activity` (152 bytes), 4 `ExecutionContext`s (160 bytes), 2 `string`s (76+66 bytes), 4 AsyncLocalMaps (160 bytes) and 3 Scope objects (144 bytes); => `758 bytes` per request; when you aren't outputting any of it because you aren't at `Information` level; you might set `"Microsoft.AspNetCore.Hosting.Diagnostics": "None"` which will do it.

e.g. 
```
"Logging": {
    "LogLevel": {
      "Default": "Information",
      "Microsoft": "Warning",
      "Microsoft.AspNetCore.Hosting.Diagnostics": "None"
    }
},
```

And if you are picking up that extra *758 bytes per request*, what's an extra object :)

Also if you have no logging as per the TE benchmarks; which also benefits from the struct being 64 bytes rather than 82 bytes when passed.

> Also, does this need to be thread safe?

No, its contained to before and after the request in hosting and not exposed to the user code in the application

benaadams: e.g. 
image


DateTimeFormat (`string`, `byte[]`) and ConcurrentDictionary (`enumerator`) are both heartbeat rather than request allocations so should be subtracted.

halter73: I figured there must be something more to it to go through all this effort. However, I still don't understand why lazy loading the private LogContext is necessary to reduce the allocations of `Activity` for example.

I don't understand why moving Activity from Context.Activity to Context. LoggingContext.Activity saves you from any allocations except for of the LogContext itself. None of this changes whether the Activity itself is allocated. If the Activity is allocated, it is allocated *before* the setter was ever called. And it's not like there are any structs on the LogContext...

150 lines to lazy allocating a POCO for longs, bools and reference properties seems kinda crazy.

benaadams: Yeah that doesn't save any allocations. Its because its a pass by value `struct` so is more about the size when its getting copied around the stack and to the heap in the statemachine.

Before
```
Type layout for 'Context'
Size: 48 bytes. Paddings: 6 bytes (%12 of empty space)
|======================================================================|
|   0-7: HttpContext <HttpContext>k__BackingField (8 bytes)            |
|----------------------------------------------------------------------|
|  8-15: IDisposable <Scope>k__BackingField (8 bytes)                  |
|----------------------------------------------------------------------|
| 16-23: Activity <Activity>k__BackingField (8 bytes)                  |
|----------------------------------------------------------------------|
| 24-31: HostingRequestStartingLog <StartLog>k__BackingField (8 bytes) |
|----------------------------------------------------------------------|
| 32-39: Int64 <StartTimestamp>k__BackingField (8 bytes)               |
|----------------------------------------------------------------------|
|    40: Boolean <EventLogEnabled>k__BackingField (1 byte)             |
|----------------------------------------------------------------------|
|    41: Boolean <HasDiagnosticListener>k__BackingField (1 byte)       |
|----------------------------------------------------------------------|
| 42-47: padding (6 bytes)                                             |
|======================================================================|
```

After
```
Type layout for 'Context'
Size: 16 bytes. Paddings: 0 bytes (%0 of empty space)
|===========================================================|
|   0-7: LogContext _logContext (8 bytes)                   |
|-----------------------------------------------------------|
|  8-15: HttpContext <HttpContext>k__BackingField (8 bytes) |
|===========================================================|
```


Its getting copied as `TContext` parameters in this interface
```csharp
public interface IHttpApplication<TContext>
{
    TContext CreateContext(IFeatureCollection contextFeatures);
    Task ProcessRequestAsync(TContext context);
    void DisposeContext(TContext context, Exception exception);
}
```
As well as copied to the heap in Kestrel's `ProcessRequests` statemachine as `TContext`

image

halter73: I completely forgot that the Context itself was a struct. Too bad it's public. It would have been sooo much easier to turn the Context into a class, but that's breaking.

benaadams: And if Logging is off for Hosting then the Activity and the LogContext will never be allocated.

halter73:
Actually, would it really be too breaking to make HostingApplication.Context a class? Context seems to have turned into an overly-complicated class anyway with almost everything being stored in the field for LogContext class.

@davidfowl

src/Hosting/Hosting/src/Internal/HostingApplication.cs
benaadams: > I completely forgot that the Context itself was a struct. Too bad it's public. It would have been sooo much easier to turn the Context into a class, but that's breaking.

Well `LogContext` is now a class if you can work out where to hang it rather than allocating it per request?

halter73: > And if Logging is off for Hosting then the Activity and the LogContext will never be allocated as they are guarded by the Log Enabled check

I figured, but that's unrelated to the LogContext.

> Well LogContext is now a class if you can work out where to hang it rather than allocating it per request? (when logging or diagnostics is enabled)

Pooling is an interesting idea. If Context itself was a class, could we pool that instead?

benaadams: The `HttpContext` is "pooled" via `IHttpContextFactory` (lives on connection); but the `HostingApplication.Context` is side-car'd over it and avoids allocation by living in the stack and being passed around by value.

benaadams:
> Actually, would it really be too breaking to make HostingApplication.Context a class?

Would need to introduce a service that the server implemented to not allocated it per request (as per HttpContext); so would be a lot more motion? And since it also contains the `HttpContext` would you want to bypass its pooling (rather than pooling both)?

Could go with a double interface; if server implements HostContext pooling use that, otherwise use only HttpContext pooling and allocate the HostContext?

benaadams:
> This PR changes the struct layout anyway which is why the reference assembly needed updating.

Only removes the attributes `[System.Runtime.CompilerServices.CompilerGeneratedAttribute]` which I'm not sure why Code Check cares about; and the dummy `int` field to indicate it also has non-objects for the C# compiler.

benaadams:
> Could go with a double interface; if server implements HostContext pooling use that, otherwise use only HttpContext pooling and allocate the HostContext?

Something like:
```
public interface IHostingContextContainer
{
    HostingContext HostingContext { get; }
}
```

benaadams:
Though... that might be part of a bigger issue. Like are some routing structures connection orientated rather than request; can scopes be reset and reused etc..?

benaadams:
Hmm.... The `LogContext` probably can't be pooled? If logging is delayed and happens outside the request, you don't want it to pick up the next request.

benaadams:
> Hmm.... The `LogContext` probably can't be pooled? If logging is delayed and happens outside the request, you don't want it to pick up the next request.

nvm. I'm thinking of the `HostingRequestStartingLog` and `HostingRequestFinishedLog` which couldn't be pooled. The Host Context could happily be resused.

benaadams:
Ahhhh... have idea!

benaadams:
With pooling the HostContext https://github.com/aspnet/AspNetCore/pull/10180

anurse:
@halter73 @Tratcher Good to merge? Everything seems nice and green :)

davidfowl:
I'd like to review this before merging. I haven't kept up to date with the changes.

benaadams:
Added WIP as it adds a cost; so should be rebased on (merged after) https://github.com/aspnet/AspNetCore/pull/10319 which removes that cost.

Also raised the issues https://github.com/aspnet/AspNetCore/issues/10556 and https://github.com/aspnet/AspNetCore/issues/10557 for the race conditions and incorrect reporting it addresses.

halter73:
@davidfowl Is this good to go in your opinion?

halter73:
Nvm. I now see @benaadams last comment.

benaadams:
Rebased on https://github.com/aspnet/AspNetCore/pull/11721 and https://github.com/aspnet/AspNetCore/pull/11723

/cc @david

benaadams:
Rebased on https://github.com/aspnet/AspNetCore/pull/11721 and https://github.com/aspnet/AspNetCore/pull/11723

/cc @davidfowl

benaadams:
Should be rebased after  #11721 and #11723 are merged and have the merge commits removed.

davidfowl:
@benaadams this needs a rebase. After recently looking at logs I think I want the activity printed in the begin and end request.

benaadams:
>  After recently looking at logs I think I want the activity printed in the begin and end request.

`Activity` doesn't look to override `.ToString()` so not sure what to output here?

benaadams:
Rebased

@davidfowl not sure what to include for Activity https://github.com/aspnet/AspNetCore/pull/10102#issuecomment-513978917

src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs
```diff
@@ -167,30 +167,34 @@ public void ContextDisposed(HostingApplication.Context context)
...
         [MethodImpl(MethodImplOptions.NoInlining)]
-        private void LogRequestFinished(HttpContext httpContext, long startTimestamp, long currentTimestamp)
+        private void LogRequestFinished(HostingApplication.Context context, long startTimestamp, long currentTimestamp)
         {
             // IsEnabled isn't checked in the caller, startTimestamp > 0 is used as a fast proxy check
-            // but that may be because diagnostics are enabled, which also uses startTimestamp, so check here
-            if (_logger.IsEnabled(LogLevel.Information))
+            // but that may be because diagnostics are enabled, which also uses startTimestamp,
+            // so check if we logged the start event
+            if (context.StartLog is object)
```
davidfowl: is object is too weird, use != null

benaadams: Done

anurse:
@davidfowl @halter73 what's the situation here? Are we ready to merge? Unfortunately it'll probably need another rebase :(.

halter73:
I'm fine with merging this. We'll want to keep a close eye on the benchmarks.

anurse:
With master targeting 5.0 now, I'm totally fine with that. I'll re-kick the build to force it to run against the latest master. If that works, I'm OK with merging without a rebase (since we'll rebase as it's merged).

davidfowl:
Imagine if we could run benchmarks before committing 😬

anurse:
Someday... someday our dreams will come true :).

@halter73 the rebuild looks to be fine, merge when you're ready.

src/Hosting/Hosting/src/Internal/HostingRequestStartingLog.cs
```diff
@@ -58,18 +45,8 @@ public override string ToString()
...
-                    _request.Method,
-                    _request.Scheme,
-                    _request.Host.Value,
-                    _request.PathBase.Value,
-                    _request.Path.Value,
-                    _request.QueryString.Value,
-                    _request.ContentType,
-                    _request.ContentLength);
+                var request = _request;
+                _cachedToString = $"{LogPreamble}{request.Protocol} {request.Method} {request.Scheme}://{request.Host.Value}{request.PathBase.Value}{request.Path.Value}{request.QueryString.Value} {EscapedValueOrEmptyMarker(request.ContentType)} {ValueOrEmptyMarker(request.ContentLength)}"; ;
```
halter73: @benaadams I just tried this locally before merging, and I noticed that the content-type in the request finished logged is getting URL-encoded. Should we really be escaping the content-type? Notice that `text/plain` becomes `text%2Fplain` below.

#### Current Master Log

```
info: Microsoft.AspNetCore.Hosting.Diagnostics[2]
      Request finished in 76.3405ms 200 text/plain
```

#### Betterized Hosting Log

```
info: Microsoft.AspNetCore.Hosting.Diagnostics[2]
      Request finished HTTP/1.1 GET http://localhost:5000/ - - - 200 14 text%2Fplain 50.6745ms
```

benaadams: Should be EscapeUriString

src/Hosting/Hosting/src/Internal/HostingRequestStartingLog.cs
```diff
@@ -87,5 +64,14 @@ IEnumerator IEnumerable.GetEnumerator()
         {
             return GetEnumerator();
         }
+
+        internal string ToStringWithoutPreamble()
+            => ToString().Substring(LogPreamble.Length);
+
+        internal static string EscapedValueOrEmptyMarker(string potentialValue)
+            => potentialValue?.Length > 0 ? Uri.EscapeDataString(potentialValue) : EmptyEntry;
```
benaadams:
```suggestion
            => potentialValue?.Length > 0 ? Uri.EscapeUriString(potentialValue) : EmptyEntry;
```

src/Hosting/Hosting/src/Internal/HostingRequestStartingLog.cs
benaadams: Hmm; looking at IIS logs it replaces space with `+` but EscapeUriString does `%20`

benaadams: Spaces are significant in the log; so `text/plain; charset=UTF-8` needs it space escaping `text/plain;+charset=UTF-8` checking what encoding does that.

benaadams: Looks like I need `WebUtility.UrlEncode`

benaadams: Oh, that makes it terrible in .NET Core
`text%2Fplain%3B+charset%3DUTF-8`

benaadams: @GrabYourPitchforks is there a better way of doing this? https://github.com/aspnet/AspNetCore/pull/10102/commits/2bdc4e706b06055e61f4d3d70332d9429965cbec

benaadams: Checking IIS logs it just converts space to `+` but leaves `+` as `+`

benaadams: @halter73 should be good now

benaadams:
Rebased

benaadams:
Didn't seem to restart the `AspNetCore-helix-test (Tests: Helix x64) ` test which failed in
`Microsoft.AspNetCore.SignalR.StackExchangeRedis.Tests`
So I'd assume unrelated?

halter73:
> Command '/usr/bin/docker logs redisTestContainer' failed with exit code '-1'

Doesn't look to be related to this change. The failure isn't related to the contents of the logs, but the `docker logs`	command exiting with a non-zero code.

PR #10105, opened 2019-05-09
Move service registration before hosting startups run

davidfowl:
- This allows IHostingStartup to see default service registrations and override them.
- Added a test to verify that hosting startup can see the IWebHostingEnvironment. Unfortunately this is used by MVC to get the application name before the container is baked. This was found when porting our workshop from 2.2 to 3.0

cc @DamianEdwards

### DIFF
--- a/src/Hosting/Hosting/src/GenericHost/GenericWebHostBuilder.cs
+++ b/src/Hosting/Hosting/src/GenericHost/GenericWebHostBuilder.cs
@@ -44,26 +44,7 @@ namespace Microsoft.AspNetCore.Hosting.Internal
                 ExecuteHostingStartups();
             });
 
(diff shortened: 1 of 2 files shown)


Tratcher:
Link to the MVC code that requires this?

davidfowl:
https://github.com/aspnet/AspNetCore/blob/258d34e3828a1870a16d13cd3c62d1b7a65acc4a/src/Mvc/Mvc.Core/src/DependencyInjection/MvcCoreServiceCollectionExtensions.cs#L81

Tratcher:
Sheish. AddMvcCore should have taken in the hosting environment or app name.

Tratcher (approved):

PR #10111, opened 2019-05-09
[Templating] Make Selenium tests run sequentially, run yarn once

javiercn:
* Makes Selenium E2E tests run sequentially based on @SteveSandersonMS investigation.
  * This should not have a great impact as things requiring npm are already sequential and the selenium bits should be relatively fast compared to the rest of the elements in the test.
* Avoids running npm as part of the E2E testing infrastructure if yarn was already run as part of CI builds.

### DIFF
--- a/src/ProjectTemplates/test/SpaTemplateTest/AngularTemplateTest.cs
+++ b/src/ProjectTemplates/test/SpaTemplateTest/AngularTemplateTest.cs
@@ -11,6 +11,7 @@ using Xunit.Abstractions;
 
 namespace Templates.Test.SpaTemplateTest
 {
+    [Collection("SpaTemplateTests")]
     public class AngularTemplateTest : SpaTemplateTestBase
     {
         public AngularTemplateTest(ProjectFactoryFixture projectFactory, BrowserFixture browserFixture, ITestOutputHelper output)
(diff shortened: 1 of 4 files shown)


SteveSandersonMS:
I tried using `[Collection]` in the components E2E tests and it didn't have any effect, at least not when running from VS. Does it work on these template tests when running from VS (and separately, on the CLI)? That's why I ended up having to use `[assembly: CollectionBehavior(DisableTestParallelization = true)]`.

Additionally, I've tracked down the components issue to an actual recently-introduced product bug now, so I'm no longer pointing the finger at Selenium! As such it definitely seems better to keep the parallelism on in the components case, since it actually helped us locate a real and potentially serious bug.

Whether it makes sense to disable parallelism for templates tests is a separate matter that I'm totally happy to defer to you on. How often do you think these are failing? When I've seen templates test failures in the CI logs recently it was to do with things like package restoration failing, not the Selenium runner.

javiercn:
> How often do you think these are failing?

Often enough that people ping me about it. If you see this does two things. One it disables parallelism within the spa tests (which collection should do) and prevents running yarn when we've already restored.

I made a change a while ago that ran custom restore steps as part of the global CI build restore. I'm not sure why it's not working anymore, so I'm looking to fix that, and then npm restore will happen at the beginning of the build (for csproj project files too) and won't run a second time as part of Build

natemcmaster:
> Avoids running npm as part of the E2E testing infrastructure if yarn was already run as part of CI builds.

Yarn should no-op in a second or less if it was already run. Are you seeing that this is not the case?

javiercn:
> Yarn should no-op in a second or less if it was already run. Are you seeing that this is not the case?

This restores within the build instead of after restores, it's what I'm trying to fix)

PR #10115, opened 2019-05-09
Fix: #10091 - Remove UseEndpointExecutor

### LINKED ISSUE #10091: Remove UseEndpointExecutor
We can add it back if it's needed but it doesn't seem needed.

cc @JamesNK @rynowak

rynowak:
Fixes: #10091

We added this middleware, but don't have a clear use case for it.


Tratcher:
I can has title, description?
At least put the issue number into the description where it becomes a link.

rynowak:
> 
> 
> I can has title, description?
> At least put the issue number into the description where it becomes a link.

Yeah these are all cool and good ideas 👍

davidfowl (approved):

PR #10121, opened 2019-05-09
Start using runtime properties to set StartupHook and EntrypointFilter

### LINKED ISSUE #9520: Use runtime property settings instead of environment variables for setting startup hook

The startup hook today is set via environment variable right now (https://github.com/aspnet/AspNetCore/blob/master/src/Servers/IIS/AspNetCoreModuleV2/InProcessRequestHandler/inprocessapplication.cpp#L241). Core-setup will eventually be adding an api to set runtime properties like https://github.com/dotnet/core-setup/pull/5336/files#diff-15fec1485f5f10c590d5b840300283ceR233. We should use these instead.

jkotalik:
Fixes https://github.com/aspnet/AspNetCore/issues/9520.

We still don't have support for stack size FYI.

See https://github.com/dotnet/core-setup/blob/master/Documentation/design-docs/native-hosting.md#initialize-host-context for more details of the APIs if curious.

### DIFF
--- a/src/Servers/IIS/AspNetCoreModuleV2/CommonLib/HostFxr.cpp
+++ b/src/Servers/IIS/AspNetCoreModuleV2/CommonLib/HostFxr.cpp
@@ -48,7 +48,6 @@ void HostFxr::Load(HMODULE moduleHandle)
         m_hostfxr_initialize_for_app_fn = ModuleHelpers::GetKnownProcAddress<hostfxr_initialize_for_app_fn>(moduleHandle, "hostfxr_initialize_for_app", true);
         m_hostfxr_set_runtime_property_value_fn = ModuleHelpers::GetKnownProcAddress<hostfxr_set_runtime_property_value_fn>(moduleHandle, "hostfxr_set_runtime_property_value", true);
         m_hostfxr_run_app_fn = ModuleHelpers::GetKnownProcAddress<hostfxr_run_app_fn>(moduleHandle, "hostfxr_run_app", true);
-        m_hostfxr_get_runtime_properties_fn = ModuleHelpers::GetKnownProcAddress<hostfxr_get_runtime_properties_fn>(moduleHandle, "hostfxr_get_runtime_properties", true);
     }
     catch (...)
     {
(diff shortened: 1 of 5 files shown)


src/Servers/IIS/AspNetCoreModuleV2/CommonLib/HostFxr.cpp
```diff
@@ -99,3 +107,40 @@ HostFxrErrorRedirector HostFxr::RedirectOutput(RedirectionOutput* writer) const
...
+    if (m_hostfxr_initialize_for_app_fn == nullptr)
+    {
+        return 0;
+    }
+
+    hostfxr_initialize_parameters params;
+    params.size = sizeof(hostfxr_initialize_parameters);
+    std::filesystem::path dotnetPath(dotnetExePath);
+
+
```
BrennanConroy:
```suggestion
```

src/Servers/IIS/AspNetCoreModuleV2/CommonLib/HostFxr.cpp
```diff
@@ -99,3 +107,40 @@ HostFxrErrorRedirector HostFxr::RedirectOutput(RedirectionOutput* writer) const
...
+    auto dotnetWstring = dotnetPath.parent_path();
+    params.dotnet_root = dotnetWstring.c_str();
+    params.host_path = L"";
+
+    RETURN_IF_NOT_ZERO(m_hostfxr_initialize_for_app_fn(argc - 1, &argv[1], nullptr, &params, &m_host_context_handle));
```
BrennanConroy: Need to call `hostfxr_close` at some point

src/Servers/IIS/AspNetCoreModuleV2/CommonLib/HostFxr.cpp
```diff
@@ -87,7 +88,14 @@ void HostFxr::SetMain(hostfxr_main_fn hostfxr_main_fn)
...
 int HostFxr::Main(DWORD argc, const PCWSTR* argv) const noexcept(false)
 {
-    return m_hostfxr_main_fn(argc, argv);
+    if (m_host_context_handle != nullptr && m_hostfxr_run_app_fn != nullptr)
+    {
+        return m_hostfxr_run_app_fn(m_host_context_handle);
+    }
+    else
+    {
+        return m_hostfxr_main_fn(argc, argv);
```
BrennanConroy: We're losing the environment variables that were set before, is that ok?

jkotalik: Yeah, it should be okay. All of this code lives in the handler, which is part of the aspnetcore runtime. Therefore, the only reason this function would be called is from our testing framework, where we set a static main method to be called here. Because everything lives in the handler, the handler version and the hostfxr version should align, meaning the handler will have access to all hostfxr functions necessary.

src/Servers/IIS/AspNetCoreModuleV2/CommonLib/HostFxr.cpp
```diff
@@ -99,3 +107,61 @@ HostFxrErrorRedirector HostFxr::RedirectOutput(RedirectionOutput* writer) const
...
+        RETURN_IF_NOT_ZERO(m_hostfxr_initialize_for_app_fn(argc - 1, &argv[1], nullptr, &params, &m_host_context_handle));
+
+    }
+    else
+    {
+        params.dotnet_root = L"";
+
+        // Initialize_for_app doesn't work with an exe name
+        std::filesystem::path applicationPath(argv[0]);
+        applicationPath.replace_extension(".dll");
```
jkotalik: Work around for https://github.com/dotnet/core-setup/issues/6388 for now.

jkotalik: Actually, I'd rather say we are blocked on it. There are a few other scenarios that still cause failing tests. StartsWithWebConfigVariationsPortable tests a bunch of different parameter configurations and 2 of them are failing. I can again work around it, but at this point it's getting quite bad.

jkotalik:
We met about this yesterday, I think we are going to hold off until there is an update to initailize_application which takes in command line parameters. This should be done in the preview7 timeframe.

jkotalik:
Specifically waiting on https://github.com/dotnet/core-setup/pull/6517

jkotalik:
I think this is ready for review now. @BrennanConroy <3

src/Servers/IIS/AspNetCoreModuleV2/InProcessRequestHandler/inprocessapplication.cpp
```diff
@@ -241,25 +241,19 @@ IN_PROCESS_APPLICATION::ExecuteApplication()
             LOG_INFOF(L"Setting current directory to %s", this->QueryApplicationPhysicalPath().c_str());
         }
 
-        if (m_pConfig->QueryCallStartupHook())
+        auto startupReturnCode = context->m_hostFxr.InitializeForApp(context->m_argc, context->m_argv.get(), m_dotnetExeKnownLocation);
+        if (startupReturnCode != 0)
         {
-            // Used to display developer exception page when there is an exception in main.
-            auto currentStartupHookEnv = Environment::GetEnvironmentVariableValue(DOTNETCORE_STARTUP_HOOK);
```
BrennanConroy: What happened to this?

BrennanConroy (approved):

src/Servers/IIS/AspNetCoreModuleV2/InProcessRequestHandler/inprocessapplication.cpp
jkotalik: It's still there: https://github.com/aspnet/AspNetCore/pull/10121/files#diff-ba7e6a78e2d291ac7cfef69259096922R252?

BrennanConroy: But you were adding before and now you aren't?

jkotalik: We are still setting the STARTUP_HOOK to a value, we are just using a runtime property instead of the environment variable. We don't need to append the startup hook as an environment variable anymore as we are using runtime properties.

PR #10123, opened 2019-05-09
Log stdout to event log on failure to start process for ANCM out of process

### LINKED ISSUE #10005: [ANCM] [out of process] Write stdout to EventLog when exit code is non-zero

To better diagnose pre-main errors, we should buffer whatever's written to stdout and write to EventLog in event of a non-zero exit code.

jkotalik:
Still thinking about how to simplify some event logs (kind of spammy right now). Will mark out of draft appropriately. 

Fixes https://github.com/aspnet/AspNetCore/issues/10005


jkotalik:
This be green now 😄

src/Servers/IIS/AspNetCoreModuleV2/OutOfProcessRequestHandler/serverprocess.cpp
```diff
@@ -1118,6 +1125,52 @@ SERVER_PROCESS::SetupStdHandles(
...
+void
+SERVER_PROCESS::ReadStdErrHandleInternal()
+{
+    std::string tempBuffer;
+    tempBuffer.resize(4096);
+    size_t m_charactersLeft = 30000;
+
+    // If ReadFile ever returns false, exit the thread
+    DWORD dwNumBytesRead = 0;
+    while (true)
```
BrennanConroy:
```suggestion
    while (m_charactersLeft > 0)
```

src/Servers/IIS/AspNetCoreModuleV2/OutOfProcessRequestHandler/serverprocess.cpp
```diff
@@ -1817,6 +1871,38 @@ SERVER_PROCESS::~SERVER_PROCESS()
         m_hStdoutHandle = NULL;
     }
 
+    // Forces ReadFile to cancel, causing the read loop to complete.
+    // Don't check return value as IO may or may not be completed already.
+    if (m_hReadThread != nullptr)
+    {
+        LOG_INFO(L"Canceling standard stream pipe reader");
```
BrennanConroy:
```suggestion
        LOG_INFO(L"Canceling standard stream pipe reader.");
```

src/Servers/IIS/AspNetCoreModuleV2/OutOfProcessRequestHandler/serverprocess.cpp
```diff
@@ -1118,6 +1125,52 @@ SERVER_PROCESS::SetupStdHandles(
...
+            nullptr))
+        {
+            auto text = to_wide_string(tempBuffer.substr(0, dwNumBytesRead), GetConsoleOutputCP());
+            auto const writeSize = min(m_charactersLeft, text.size());
+            m_output.write(text.c_str(), writeSize);
+            m_charactersLeft -= writeSize;
+        }
+        else
+        {
+            auto hr = GetLastError();
```
BrennanConroy: Should something be done here? Otherwise might as well just return without checking last error.

src/Servers/IIS/AspNetCoreModuleV2/OutOfProcessRequestHandler/serverprocess.cpp
```diff
@@ -1118,6 +1125,52 @@ SERVER_PROCESS::SetupStdHandles(
...
+    auto pLoggingProvider = static_cast<SERVER_PROCESS*>(pContext);
+    DBG_ASSERT(pLoggingProvider != NULL);
+    pLoggingProvider->ReadStdErrHandleInternal();
+}
+
+void
+SERVER_PROCESS::ReadStdErrHandleInternal()
+{
+    std::string tempBuffer;
+    tempBuffer.resize(4096);
```
BrennanConroy: nit: should be able to pass the initial size in the string constructor and save the initial small buffer allocation.

src/Servers/IIS/AspNetCoreModuleV2/CommonLib/resources.h
```diff
@@ -13,8 +13,8 @@
 
 #define ASPNETCORE_EVENT_PROCESS_START_SUCCESS_MSG           L"Application '%s' started process '%d' successfully and process '%d' is listening on port '%d'."
 #define ASPNETCORE_EVENT_RAPID_FAIL_COUNT_EXCEEDED_MSG       L"Maximum rapid fail count per minute of '%d' exceeded."
-#define ASPNETCORE_EVENT_PROCESS_START_ERROR_MSG             L"Application '%s' with physical root '%s' failed to start process with commandline '%s' at stage '%s', ErrorCode = '0x%x', assigned port %d, retryCounter '%d'."
-#define ASPNETCORE_EVENT_PROCESS_START_FAILURE_MSG           L"Application '%s' with physical root '%s' failed to start process with commandline '%s' with multiple retries. Failed to bind to port '%d'. See previous warnings for details."
+#define ASPNETCORE_EVENT_PROCESS_START_ERROR_MSG             L"Application '%s' with physical root '%s' failed to start process with commandline '%s' at stage '%s', ErrorCode = '0x%x', assigned port %d, retryCounter '%d'"
```
BrennanConroy: nit: Add the period back

src/Servers/IIS/AspNetCoreModuleV2/OutOfProcessRequestHandler/serverprocess.cpp
```diff
@@ -1118,6 +1125,52 @@ SERVER_PROCESS::SetupStdHandles(
...
+    size_t m_charactersLeft = 30000;
+
+    // If ReadFile ever returns false, exit the thread
+    DWORD dwNumBytesRead = 0;
+    while (true)
+    {
+        if (ReadFile(m_hStdoutHandle,
+            tempBuffer.data(),
+            4096,
```
BrennanConroy: nit: store 4096 in a local variable

src/Servers/IIS/AspNetCoreModuleV2/OutOfProcessRequestHandler/serverprocess.cpp
```diff
@@ -1118,6 +1125,52 @@ SERVER_PROCESS::SetupStdHandles(
...
+            &dwNumBytesRead,
+            nullptr))
+        {
+            auto text = to_wide_string(tempBuffer.substr(0, dwNumBytesRead), GetConsoleOutputCP());
```
BrennanConroy: super nit: so many allocations here, could make a `to_wide_string` function that takes the string and count so you don't need to `substr`

BrennanConroy (commented):
Should there be a test that verifies only the first 30kb of output is logged?

jkotalik:
> Should there be a test that verifies only the first 30kb of output is logged?

Yeah we should. Laziness on my part 🐋

src/Servers/IIS/AspNetCoreModuleV2/OutOfProcessRequestHandler/serverprocess.cpp
jkotalik: I think we shouldn't handle errors here. There are at least 3 different errors that could be returned that are false positives. We also cancel reading once the process exits.

src/Servers/IIS/AspNetCoreModuleV2/OutOfProcessRequestHandler/serverprocess.cpp
jkotalik: I wasn't able to find a constructor that took in just a size for string.

jkotalik:
🆙 📅

src/Servers/IIS/AspNetCoreModuleV2/OutOfProcessRequestHandler/serverprocess.cpp
```diff
@@ -1118,6 +1125,57 @@ SERVER_PROCESS::SetupStdHandles(
...
+    DWORD dwNumBytesRead = 0;
+    while (charactersLeft > 0)
+    {
+        if (ReadFile(m_hStdoutHandle,
+            tempBuffer.data(),
+            bufferSize,
+            &dwNumBytesRead,
+            nullptr))
+        {
+            auto text = to_wide_string(tempBuffer.substr(0, dwNumBytesRead), GetConsoleOutputCP());
```
BrennanConroy: Use the new overload!

BrennanConroy (approved):

jkotalik:
I would like to profile this before merging. I'm concerned that setting these to handles will have a perf impact if we are logging in a middleware.

jkotalik:
Verified locally, no performance regression when logging informational on localhost. I'll still like to do a second pass (and I'll add the accepted label eventually) after we merge this and it is available in the hosting bundle. I'm most concerned about performance on azure, so I'll make sure to verify it there.

I'll merge once this is green.

Issue #8843, opened 2019-03-26
Unexpected exception in TimingPipeFlusher.TimeFlushAsync when using HttpProtocols.Http1AndHttp2


silmon27:
I'm getting a couple dozen "Unexpected exception in TimingPipeFlusher.TimeFlushAsync" exceptions in my logs every day. This happens only when HTTP2 support is enabled in Kestrel.

```
System.InvalidOperationException: Can't GetResult unless awaiter is completed.
   at System.IO.Pipelines.ThrowHelper.ThrowInvalidOperationException_GetResultNotCompleted()
   at System.IO.Pipelines.Pipe.GetFlushAsyncResult()
   at System.IO.Pipelines.Pipe.DefaultPipeWriter.GetResult(Int16 token)
   at System.Threading.Tasks.ValueTask`1.ValueTaskSourceAsTask.<>c.<.cctor>b__4_0(Object state)
--- End of stack trace from previous location where exception was thrown ---
   at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure.TimingPipeFlusher.TimeFlushAsync(MinDataRate minRate, Int64 count, IHttpOutputAborter outputAborter, CancellationToken cancellationToken)
```

I can't seem to trigger it myself. The exceptions occur on a production server.

The app is running in Kubernetes from the mcr.microsoft.com/dotnet/core/aspnet:2.2.3-alpine image.

jkotalik:
To confirm, you are saying this only occurs when using HTTP/2, correct? The title and issue conflict.

silmon27:
The exceptions only occur when ListenOptions.Protocols is set to HttpProtocols.Http1AndHttp2, they don't occur when the default HttpProtocols.Http1 is used. I haven't tried forcing it to HTTP/2 only.

silmon27:
As expected forcing HttpProtocols.Http2 results in the same exception as HttpProtocols.Http1AndHttp2. Also not related to Alpine, same happens with regular image.
I will try to enable verbose logging and see if I can find out more.

silmon27:
I enabled verbose logging for a couple of days and found that most cases look something like this (shortened):

```
2019-04-04 09:39:32.629 +00:00 [DBG] Connection id "0HLLO7GQRO35H" started.
2019-04-04 09:39:32.695 +00:00 [VRB] Connection id "0HLLO7GQRO35H" sending "SETTINGS" frame for stream ID 0 with length 18 and flags "NONE"
[...]

PR #10126, opened 2019-05-09
Work around potential race in PipeWriter

halter73:
- Make sure we always await the last flush task before calling FlushAsync
  again instead of preemptively calling FlushAsync and checking to see
  if the ValueTask is incomplete before bothering to acquire the _flushLock
- This now acquires the _flushLock fore every call to Response.Body.WriteAsync
  whereas this only happened for truly async writes before.
- I don't think this is a big concern since this should normally be uncontested,
  and DefaultPipeWriter.FlushAsync/GetResult already acquire a lock.

Addresses #8843

@benaadams


JamesNK:
Is there a way to unit test this?

halter73:
> Is there a way to unit test this?

Yep. https://github.com/aspnet/AspNetCore/pull/10126/commits/18cae2800e8f4eadd17d545a2ee2f86ee259622c

benaadams:
This means its no longer allocation free?

src/Servers/Kestrel/Core/src/Internal/Infrastructure/TimingPipeFlusher.cs
```diff
@@ -111,5 +115,11 @@ private async ValueTask<FlushResult> TimeFlushAsync(MinDataRate minRate, long co
 
             return default;
         }
+
+        private async Task<FlushResult> AwaitLastFlushAndTimeFlushAsync(Task lastFlushTask, MinDataRate minRate, long count, IHttpOutputAborter outputAborter, CancellationToken cancellationToken)
```
benaadams: Any reason not `ValueTask<FlushResult>`? `Task<FlushResult>` will always allocate.

halter73: In this case we know the lastFlushTask is not complete so we'll allocate anyway. And it removes some calls to AsTask().

src/Servers/Kestrel/Core/src/Internal/Infrastructure/TimingPipeFlusher.cs
```diff
@@ -51,35 +51,39 @@ public ValueTask<FlushResult> FlushAsync(MinDataRate minRate, long count)
...
-
             // https://github.com/dotnet/corefxlab/issues/1334
             // Pipelines don't support multiple awaiters on flush.
-            // While it's acceptable to call PipeWriter.FlushAsync again before the last FlushAsync completes,
-            // it is not acceptable to attach a new continuation (via await, AsTask(), etc..). In this case,
-            // we find previous flush Task which still accounts for any newly committed bytes and await that.
             lock (_flushLock)
             {
-                if (_lastFlushTask == null || _lastFlushTask.IsCompleted)
+                if (_lastFlushTask != null && !_lastFlushTask.IsCompleted)
```
benaadams: `.IsCompletedSuccessfully`?

Want Cancellation and Exceptions to go down the awaited path?

benaadams: `.IsCompletedSuccessfully`?

Want Cancellation and Exceptions to go down the awaited path? (i.e. to throw and wrap their exceptions, rather than ignore them)

benaadams: nvm, see what's happening

benaadams (approved):

src/Servers/Kestrel/Core/src/Internal/Infrastructure/TimingPipeFlusher.cs
halter73: Yeah. I was thinking about this too, but my reasoning for not awaiting any completed Task is that if it is a persistent error, the next Flush should return the same error anyway. If not, all the better.

halter73: The remaining question I had is whether to swallow any exceptions that might be thrown from an awaited _lastFlushTask. Right now, I don't.

halter73:
> This means plaintext middleware is no longer allocation free?

I suspect we'll still be fine as long as the writes don't actually go async. There will be slightly more locking (PipeWriter itself was already locking al lot), but I doubt that will be too significant. I'll run the plaintext benchmark before merging though.

src/Servers/Kestrel/Core/test/TimingPipeFlusherTests.cs
```diff
@@ -0,0 +1,67 @@
...
+            Assert.False(flushTask2.IsCompleted);
+            Assert.Equal(1, pipeWriterFlushCallCount);
+
+            pipeWriterFlushTcsArray[0].SetResult(default);
+            await flushTask0.AsTask().DefaultTimeout();
+
+            Assert.True(flushTask0.IsCompleted);
+            Assert.False(flushTask1.IsCompleted);
+            Assert.False(flushTask2.IsCompleted);
+            Assert.True(pipeWriterFlushCallCount <= 2);
```
jkotalik: Why <= 2 rather than 2?

jkotalik (approved):

davidfowl:
> I suspect we still won't allocate as long as the writes don't actually go async. There will be slightly more locking (PipeWriter itself was already locking al lot), but I doubt that will be too significant. I'll run the plaintext benchmark before merging though.

Don't suspect, get numbers, run the benchmark.

davidfowl (changes requested):
The change looks good but we need to do a performance run to verify allocations and throughput.

src/Servers/Kestrel/Core/test/TimingPipeFlusherTests.cs
halter73: We already know the pipeWriterFlushCallCount is at least 1, so this is verifying the flush count is either 1 or 2.

We've only waited for the first TimingPipeFlusher flush to finish. It's possible that the second PipeWriter flush has already started, but it might not have. There's nothing to wait on to be sure the the 2nd PipeWriter flush has started except waiting for the second TimingPipeFlusher flush to finish at which point the 3rd PipeWriter flush may have already started.

We could set a Tcs in the mock PipeWriter.FlushAsync method and wait on that, but it doesn't seem necessary. We verify there were exactly 3 calls to flush at the end of the test.

PR #10127, opened 2019-05-09
Reduce Functional Test Stream Item Delay

mikaelm12:
This makes the functional tests run twice as fast.
A number of our tests use this method and when running all the variations it adds up.

### DIFF
--- a/src/SignalR/clients/csharp/Client/test/FunctionalTests/Hubs.cs
+++ b/src/SignalR/clients/csharp/Client/test/FunctionalTests/Hubs.cs
@@ -186,7 +186,7 @@ namespace Microsoft.AspNetCore.SignalR.Client.FunctionalTests
                 for (var i = 0; i < count; i++)
                 {
                     await channel.Writer.WriteAsync(i);
-                    await Task.Delay(100);
+                    await Task.Delay(10);
                 }
 
                 channel.Writer.TryComplete();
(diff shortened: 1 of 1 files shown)


halter73:
Is there any particular reason we need to delay any amount of time? What if we Thread.Yielded instead? Or just wrote to the ChannelWriter as fast as it would allow?

halter73:
I think we should also look at reducing the number of variations of our functional tests. In our C# functional HubConnectionTests.cs, we only have 40 test methods, but 768 test cases. I doubt that all these variations are really adding coverage.

mikaelm12:
> I think we should also look at reducing the number of variations of our functional tests. In our C# functional HubConnectionTests.cs, we only have 40 test methods, but 768 test cases. I doubt that all these variations are really adding coverage.

I think in the beginning when things were a lot smaller and we were just adding the new hub types and protocols it made more sense but I tend to have more of your mindset now.

mikaelm12:
> Is there any particular reason we need to delay any amount of time? What if we Thread.Yielded instead? Or just wrote to the ChannelWriter as fast as it would allow?

There are some streaming tests for cancellation that depended on having some delay. A better solution might be to find the tests that depend on that and add new hub methods for those and reduce the number of test variants. A lot of our tests current use the hub method that has a 100ms delay that don't particularly need it.

halter73 (approved):

halter73:
Either way, this seems to be an improvement.

PR #10133, opened 2019-05-10
Use sealed contexts for ResourceInvoker

benaadams:
They are all private fields; so introduce some `private` derived `sealed` contexts and use them as the variable types.

The public variety are still unsealed and when passed extrenally they are of the parent type; however internal to the class they can be used as the sealed type.

Since the Jit can see the type is sealed, it can devirtualize and even inline the virtual methods and properties.

e.g. before Rethrow has to make virtual calls to the properties `ExceptionHandled`, `ExceptionDispatchInfo` and `Exception`

image

With the sealed types it can inline `ExceptionHandled`, `ExceptionDispatchInfo` and the call to `Exception` is direct: 

image

And the public api surface remains the same.

/cc @rynowak 

This is why I'd like the Page classes to be generated sealed https://github.com/aspnet/AspNetCore/issues/5083 /cc @mkArtakMSFT, @pranavkm, @Eilon for the calls to `Write` etc to devirtualize.

### DIFF
--- a/src/Mvc/Mvc.Core/src/Infrastructure/ResourceInvoker.cs
+++ b/src/Mvc/Mvc.Core/src/Infrastructure/ResourceInvoker.cs
@@ -24,12 +24,12 @@ namespace Microsoft.AspNetCore.Mvc.Infrastructure
         protected readonly IFilterMetadata[] _filters;
         protected readonly IList<IValueProviderFactory> _valueProviderFactories;
 
-        private AuthorizationFilterContext _authorizationContext;
-        private ResourceExecutingContext _resourceExecutingContext;
-        private ResourceExecutedContext _resourceExecutedContext;
-        private ExceptionContext _exceptionContext;
-        private ResultExecutingContext _resultExecutingContext;
-        private ResultExecutedContext _resultExecutedContext;
+        private AuthorizationFilterContextSealed _authorizationContext;
+        private ResourceExecutingContextSealed _resourceExecutingContext;
+        private ResourceExecutedContextSealed _resourceExecutedContext;
+        private ExceptionContextSealed _exceptionContext;
+        private ResultExecutingContextSealed _resultExecutingContext;
+        private ResultExecutedContextSealed _resultExecutedContext;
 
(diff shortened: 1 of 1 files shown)


Eilon (approved):
I approve this message.

rynowak (approved):
Super cool and good

pranavkm:
Would we consider sealing the actual context types? While it's a breaking change, there aren't very compelling reasons for it to be unsealed.

benaadams:
> Would we consider sealing the actual context types? While it's a breaking change, there aren't very compelling reasons for it to be unsealed.

There are public interfaces that take them? e.g. `IAsyncActionFilter`

If you are happy to break the general purposefulness of these interfaces by sealing the type; then could save allocations by changing them to also accept state so a delegate doesn't get allocated each time they are called.

image

Could be something like
```diff
namespace Microsoft.AspNetCore.Mvc.Filters
{
-   public delegate Task<ActionExecutedContext> ActionExecutionDelegate();
+   public delegate ValueTask<ActionExecutedContext> ActionExecutionDelegate(object state);

    public interface IAsyncActionFilter : IFilterMetadata
    {
-       Task OnActionExecutionAsync(ActionExecutingContext context, ActionExecutionDelegate next);
+       Task OnActionExecutionAsync(ActionExecutingContext context, ActionExecutionDelegate next, object state);
    }
}
```

pranavkm:
I was thinking of either recommending users put the state on `HttpContext` or just having a settable `object state` on the context object. Changing the signature for filters is very obtrusive, that we would want to avoid.

benaadams:
Isn't user code that's allocating; is this line https://github.com/aspnet/AspNetCore/blob/0a148bfca4b9e26210242778a6d263d9b4b3d0a4/src/Mvc/Mvc.Core/src/Infrastructure/ControllerActionInvoker.cs#L124

That allocates both an `ActionExecutionDelegate` and a `Task<ActionExecutedContext>`

As `InvokeNextActionFilterAwaitedAsync` has to be converted to an `ActionExecutionDelegate` that has captured `this`; and then it needs to return a `Task<T>` with a unique result.

Issue #6880, opened 2019-01-20
TestHost with HTTP2 features


JamesNK:
Related: https://github.com/aspnet/AspNetCore/issues/6871

Attempting to use HTTP2 features in the test website with TestHost...

```cs
httpContext.Response.AppendTrailer("grpc-status", ((int)StatusCode.OK).ToString());
```

...results in an error: `Message: System.InvalidOperationException : Trailers are not supported for this response.`

This is the first error I encountered when using HTTP2 - I'm going to see if I can work around it in the test website - but I suspect there is other issues to fix to HTTP2-erize TestHost.

// @davidfowl @Tratcher

Tratcher:
https://github.com/aspnet/AspNetCore/blob/425c196cba530b161b120a57af8f1dd513b96f67/src/Http/Http.Abstractions/src/Extensions/ResponseTrailerExtensions.cs#L44-L47

This is a more interesting problem for TestServer than it was for just DefaultHttpContext. What client API are you using to send requests? HttpClient I assume. That doesn't have an object model to represent trailers yet. You'd have to use the TestServer API that lets you directly interact with the HttpContext in order to see any response trailers.

As to how TestServer should decide if it's in HTTP/2 mode or supports trailers (e.g. adds the feature) is another question. It could key it based on the protocol version on the request (which HttpClient now defaults to HTTP/2).

anurse:
Assuming we don't try to add support for HttpClient, is the work item here just to initialize the trailers feature so that the client can access that? Seems straightforward. There will likely be a reaction needed when (if?) HttpClient adds trailer support.

JamesNK:
HttpClient has already added trailer support for 3.0. It is in the latest SDK.

JamesNK:
https://github.com/dotnet/corefx/pull/36128 and https://github.com/dotnet/corefx/pull/36128

PR #10135, opened 2019-05-10
Return response trailers from test server

JamesNK:
Addresses #6880

### DIFF
--- a/src/Hosting/TestHost/src/ResponseStream.cs
+++ b/src/Hosting/TestHost/src/ResponseStream.cs
@@ -19,20 +19,21 @@ namespace Microsoft.AspNetCore.TestHost
         private bool _complete;
         private bool _aborted;
         private Exception _abortException;
-        private SemaphoreSlim _writeLock;
-
-        private Func<Task> _onFirstWriteAsync;
         private bool _firstWrite;
-        private Action _abortRequest;
-        private Func<bool> _allowSynchronousIO;
 
-        private Pipe _pipe = new Pipe();
+        private readonly SemaphoreSlim _writeLock;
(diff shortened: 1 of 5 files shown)


src/Hosting/TestHost/src/ResponseStream.cs
```diff
@@ -116,6 +117,7 @@ public async override Task<int> ReadAsync(byte[] buffer, int offset, int count,
                 if (result.Buffer.IsEmpty && result.IsCompleted)
                 {
                     _pipe.Reader.Complete();
+                    _readComplete();
```
Tratcher: ReadAsync can be called multiple times at the end and return 0 each time, no? Or would the pipe reader throw if you did that? You don't want _readComplete being called multiple times.

src/Hosting/TestHost/test/ClientHandlerTests.cs
```diff
@@ -88,6 +88,55 @@ public Task SingleSlashNotMovedToPathBase()
             return httpClient.GetAsync("https://example.com/");
         }
 
+        [Fact]
+        public async Task ServerTrailersSetOnResponseAfterContentRead()
+        {
+            var tcs = new TaskCompletionSource<object>();
```
Tratcher: RunContinuationAsync

Tratcher (approved):
Nice and clean.

src/Hosting/TestHost/src/ResponseStream.cs
JamesNK: The pipe reader throws. I've updated the stream to return 0 when the reader is complete.

Can you double check the way I've done it?

Tratcher (approved):

PR #10154, opened 2019-05-10
Remove custom stream->pipe wrappers and make BodyReader/BodyWriter read-only

### LINKED ISSUE #9223: Pipelines in ASP.NET Core 3.0
* Replace our default wrapping with the wrappers present in corefx
* Make the features read-only.

cc @davidfowl

analogrelay:
Just throwing this up for a first look and to see what (if anything) breaks.

Some `// REVIEW:` comments in the code to be looked at.

* Makes `HttpResponse.BodyReader` and `HttpRequest.BodyWriter` read-only. The **feature** is still read-write
* Deletes `StreamPipeReader` and `StreamPipeWriter` and uses `PipeReader.Create(Stream)` and `PipeWriter.Create(Stream)` from corefx instead.


src/Http/Http/src/Features/RequestBodyPipeFeature.cs
```diff
@@ -31,10 +33,13 @@ public PipeReader Reader
                 }
 
                 if (_internalPipeReader == null ||
-                    !object.ReferenceEquals(_internalPipeReader.InnerStream, _context.Request.Body))
+                    !ReferenceEquals(_streamInstanceWhenWrapped, _context.Request.Body))
                 {
-                    _internalPipeReader = new StreamPipeReader(_context.Request.Body);
-                    _context.Response.RegisterForDispose(_internalPipeReader);
+                    _streamInstanceWhenWrapped = _context.Request.Body;
+                    _internalPipeReader = PipeReader.Create(_context.Request.Body);
```
jkotalik: @davidfowl the new wrappers aren't disposable, correct?

anurse: Yeah, that's why the question. They aren't disposable but they seem to have resources that are cleaned up in `Complete`: https://github.com/dotnet/corefx/blob/master/src/System.IO.Pipelines/src/System/IO/Pipelines/StreamPipeWriter.cs#L206

Do we need to make sure `Complete` is called?

src/Http/Http/src/ReadOnlyPipeStream.cs
```diff
@@ -11,6 +11,7 @@ namespace System.IO.Pipelines
     /// <summary>
     /// Represents a read-only Stream backed by a PipeReader
     /// </summary>
+    // REVIEW: Could this be replaced or simplified with https://apisof.net/catalog/System.IO.Pipelines.PipeReader.AsStream() ?
```
jkotalik: It really should be able to. What issues did you run into here?

anurse: I didn't run in to any issues, I just didn't know exactly what this type was for :)

anurse: I can give it a try.

jkotalik: Oh nvrm I see the issue. Kestrel has a class that derives this. The stream types in corefx aren't public.

jkotalik:
I cloned the PR and see there are some benchmark projects that aren't compiling.

src/Http/Http/src/Features/RequestBodyPipeFeature.cs
jkotalik: I think we really should make sure Complete is called... maybe adding it to OnCompleted as a callback? Either way I hate it.

@davidfowl what should we do here?

davidfowl: OnCompleted

anurse: Which `OnCompleted`? `HttpResponse.OnCompleted`?

anurse: Is `Complete` idempotent?

jkotalik: Internally, `Complete` is called only once (bool to check if it has been called already).

src/Http/Http/src/Features/RequestBodyPipeFeature.cs
```diff
@@ -31,19 +34,31 @@ public PipeReader Reader
...
+
+                    _context.Response.OnCompleted((self) => ((RequestBodyPipeFeature)self).Complete(), this);
                 }
 
                 return _internalPipeReader;
             }
             set
             {
                 _userSetPipeReader = value ?? throw new ArgumentNullException(nameof(value));
+                // REVIEW: Should we do this (vvvv) or just get rid of the setter?
```
anurse: Thoughts?

src/Http/Http/src/Features/ResponseBodyPipeFeature.cs
```diff
@@ -31,19 +35,31 @@ public PipeWriter Writer
...
+
+                    _context.Response.OnCompleted((self) => ((ResponseBodyPipeFeature)self).Complete(), this);
                 }
 
                 return _internalPipeWriter;
             }
             set
             {
                 _userSetPipeWriter = value ?? throw new ArgumentNullException(nameof(value));
+                // REVIEW: Should we do this (vvvv) or just get rid of the setter?
```
anurse: Thoughts?

src/Http/Http/src/Features/ResponseBodyPipeFeature.cs
```diff
@@ -31,19 +35,31 @@ public PipeWriter Writer
...
                 // TODO set the response body Stream to an adapted pipe https://github.com/aspnet/AspNetCore/issues/3971
             }
         }
+
+        private Task Complete()
+        {
+            if(_internalPipeWriter != null && ReferenceEquals(_context.Response.Body, _streamInstanceWhenWrapped))
```
anurse: Should we complete the active `PipeWriter` or every `PipeWriter` wrapper we ever created? Right now this code only completes a `PipeWriter` if the stream it wraps is still the active body stream.

anurse:
> I cloned the PR and see there are some benchmark projects that aren't compiling.

Let's get the core implementation agreed on then I'll go and update the benchmarks.

src/Http/Http/src/Features/ResponseBodyPipeFeature.cs
anurse: Ping @davidfowl @jkotalik

jkotalik: I think we have to call complete on all wrappers created. Otherwise who else will clean it up?

src/Http/Http/src/ReadOnlyPipeStream.cs
jkotalik: Leave these types for now

src/Http/Http/src/Features/RequestBodyPipeFeature.cs
jkotalik: I think we should keep the setter. Kestrel needs to be able to set the optimized pipe writer somehow.

src/Http/Http/src/Features/ResponseBodyPipeFeature.cs
jkotalik: Same as above.

src/Http/Http/src/Features/ResponseBodyPipeFeature.cs
anurse: Yeah, that's my thought too.

src/Http/Http/src/Features/RequestBodyPipeFeature.cs
anurse: Can't Kestrel just implement the feature directly though? It doesn't need the setter to provide an optimized pipe.

davidfowl: Yea why would it need the setter

anurse: Should we just remove the setter entirely then? In the original plan we were keeping it (but making the property on `HttpRequest`/`HttpResponse` read-only). You can always replace the feature entirely so if we make the feature read-only it seems fine to me.

What's the contract here regarding completion? My suggestion is this:

1. If you provide this feature yourself, you complete the pipe sometime
1. If we provide this default feature implementation and someone accesses the pipe, we'll queue an `OnComplete` handler to complete the wrapper
1. If you want to "wrap" or "chain" a pipe, you replace the feature entirely and the replaced pipe will be completed (as per the previous two clauses).

jkotalik: Okay, I buy that this feature can be readonly.

I think your plan is spot on.

anurse: the camera slowly pans back and a man nods lightly

src/Http/Http/src/ReadOnlyPipeStream.cs
```diff
@@ -11,6 +11,7 @@ namespace System.IO.Pipelines
...
     public class ReadOnlyPipeStream : Stream
```
anurse: I want to remove this. It looks like Kestrel inherits from it in `HttpRequestStream` (and theres also a `WriteOnlyPipeStream` that Kestrel uses in `HttpResponseStream`). These are public so I want to yank them if we can. I don't think Kestrel needs to use them at all though. I think it can do all the work in it's own stream.

jkotalik: That sounds fine. I would do the following:
1. Move ReadOnlyPipeStream\WriteOnlyPipeStream to Kestrel project and make them internal
2. Merge HttpRequestStream with ReadOnlyPipeStream and visa versa.

src/Http/Http.Abstractions/src/HttpRequest.cs
```diff
@@ -104,9 +104,9 @@ public abstract class HttpRequest
         public abstract Stream Body { get; set; }
 
         /// <summary>
-        /// Gets or sets the request body pipe <see cref="PipeReader"/>.
+        /// Gets the request body pipe <see cref="PipeReader"/>.
         /// </summary>
-        public abstract PipeReader BodyReader { get; set; }
+        public abstract PipeReader BodyReader { get; }
```
halter73: I'm a huge fan of removing these setters. I think making replacing the entire feature the only way to replace the body PipeReader/PipeWriter, gets us off the hook for automagically replacing the body streams. After all, you'd only expect more experienced devs to try such a thing.

anurse:
@jkotalik making me look good ;). Thanks for taking this over!

jkotalik:
@anurse I'd prefer we get this in for preview6 as there are a few breaking changes between preview5 and preview6.

src/Http/Http.Abstractions/test/HttpResponseWritingExtensionsTests.cs
```diff
@@ -32,33 +30,6 @@ public async Task WritingText_MultipleWrites()
             Assert.Equal(22, context.Response.Body.Length);
         }
 
-        [Theory]
-        [MemberData(nameof(Encodings))]
-        public async Task WritingTextThatRequiresMultipleSegmentsWorks(Encoding encoding)
```
davidfowl: Why did this get removed?

src/Http/Http/test/Features/FormFeatureTests.cs
```diff
@@ -83,41 +83,6 @@ public async Task ReadFormAsync_SimpleData_ReturnsParsedFormCollection(bool buff
             await responseFeature.CompleteAsync();
         }
 
-        [Theory]
-        [InlineData(true)]
-        [InlineData(false)]
-        public async Task ReadFormAsync_SimpleData_ReplacePipeReader_ReturnsParsedFormCollection(bool bufferRequest)
```
davidfowl: I think we want to keep this test as well. Use the feature instead of the setter.

src/Http/Http/test/Features/RequestBodyPipeFeatureTests.cs
```diff
@@ -24,111 +19,7 @@ public void RequestBodyReturnsStreamPipeReader()
...
-            var expectedStream = new MemoryStream();
-            context.Request.Body = expectedStream;
-            pipeBody = feature.Reader;
-
-            Assert.True(pipeBody is StreamPipeReader);
-            Assert.Equal(expectedStream, (pipeBody as StreamPipeReader).InnerStream);
-        }
-
-        [Fact]
-        public async Task RequestBodyGetsDataFromSecondStream()
```
davidfowl: This test still looks valid.

src/Servers/Kestrel/test/InMemory.FunctionalTests/ResponseTests.cs
```diff
@@ -3913,181 +3883,6 @@ public async Task ResponseSetBodyToSameValueTwiceGetPipeMultipleTimesDifferentOb
...
-                        $"Date: {server.Context.DateHeaderValue}",
-                        "Content-Length: 0",
-                        "",
-                        "");
-                }
-            }
-        }
-
-        [Fact]
-        public async Task ResponseStreamWrappingWorks()
```
davidfowl: This test looks valid?

src/Http/Http.Abstractions/test/HttpResponseWritingExtensionsTests.cs
jkotalik: StreamPipeWriter doesn't exist anymore. I'll try rewriting this test to make it work with our current wrappers.

jkotalik:
🆙 📅

Tratcher (approved):

davidfowl:
Conflicts! This is good to go though

jkotalik:
> Conflicts! This is good to go though

Wonder who's fault that is <_<

src/Http/Http/src/Features/ResponseBodyPipeFeature.cs
```diff
@@ -25,25 +27,21 @@ public PipeWriter Writer
...
-                }
-
                 if (_internalPipeWriter == null ||
-                    !object.ReferenceEquals(_internalPipeWriter.InnerStream, _context.Response.Body))
+                    !ReferenceEquals(_streamInstanceWhenWrapped, _context.Response.Body))
                 {
-                    _internalPipeWriter = new StreamPipeWriter(_context.Response.Body);
-                    _context.Response.RegisterForDispose(_internalPipeWriter);
+                    _streamInstanceWhenWrapped = _context.Response.Body;
+                    _internalPipeWriter = PipeWriter.Create(_context.Response.Body);
```
halter73: Do we want to specify the leaveOpen option here if and when it becomes available? Or do we want to dispose whatever someone set Response.Body to on their available. I'm leaning towards the former.

halter73 (approved):

src/Http/Http/src/Features/ResponseBodyPipeFeature.cs
davidfowl: We’d probably set leaveOpen to true and not have it be pluggable.

JamesNK:
Broken gRPC TestServer test after this change - https://github.com/aspnet/AspNetCore/issues/10737

davidfowl:
Hah! Closing the Stream I guess.

PR #10166, opened 2019-05-10
Testing template build

pranavkm:
[Ignore]


ryanbrandenburg:
AZDO doesn't show us all the tests in the results view, as there's a failing test which doesn't show if you disable all filtering then look for its exact name. AZDO also reports "3,791 Tests not reported", whatever that means.

ryanbrandenburg:
testsnotreported

PR #10171, opened 2019-05-11
JSInterop + System.Text.Json

pranavkm:


src/Components/Blazor/Blazor/src/Http/WebAssemblyHttpMessageHandler.cs
```diff
@@ -168,14 +188,21 @@ public HttpResponseMessage ToResponseMessage(HttpContent content)
...
-                        contentHeaders?.TryAddWithoutValidation(pair[0], pair[1]);
+                        contentHeaders?.TryAddWithoutValidation(pair.Name, pair.Value);
                     }
                 }
 
                 return result;
             }
         }
+
+        private class Header
```
rynowak: readonly struct?

src/Components/Blazor/Blazor/src/Http/WebAssemblyHttpMessageHandler.cs
```diff
@@ -147,7 +167,7 @@ private class FetchOptions
         private class RequestInit
         {
             public string Credentials { get; set; }
-            public string[][] Headers { get; set; }
```
rynowak: this is a jagged array. YOU LIED to us in standup 🤣

src/Components/Blazor/Build/src/Core/BootJsonWriter.cs
```diff
@@ -40,7 +40,7 @@ public static string GetBootJsonContent(string assemblyFileName, string entryPoi
                 assemblyReferences,
                 embeddedContent,
                 linkerEnabled);
-            return Json.Serialize(data);
+            return JsonSerializer.ToString(data, new JsonSerializerOptions { PropertyNamingPolicy = JsonNamingPolicy.CamelCase });
```
rynowak: Worth making this a shared constant somewhere?

src/Components/Blazor/Build/test/RuntimeDependenciesResolverTest.cs
```diff
@@ -128,13 +121,21 @@ public void FindsReferenceAssemblyGraph_ForStandaloneApp()
...
             // Assert
+            if (contentNotFound.Any())
+            {
+                Assert.False(true, $"Expected content not found: {string.Join(", ", contentNotFound)}");
+            }
+
+            if (additionalContentFound.Any())
+            {
+                Assert.False(true, $"Unexpected content found: {string.Join(", ", additionalContentFound)}");
+            }
```
rynowak: this is less-good now right? It will only log one or the other. 

I always end up writing exceptions that derive from XunitException for things like this.

src/Components/Browser.JS/dist/Debug/blazor.webassembly.js
```diff
@@ -316,19 +316,15 @@ var DotNet;
             promise.catch(function (error) { return console.error(error); });
         };
         DotNetObject.prototype.serializeAsArg = function () {
-            return "__dotNetObject:" + this._id;
+            return { __dotNetObject: this._id };
```
rynowak: Is this file supposed to be checked in?

src/Components/Browser/src/RendererRegistryEventDispatcher.cs
```diff
@@ -30,34 +37,61 @@ private static UIEventArgs ParseEventArgsJson(string eventArgsType, string event
...
+            return JsonSerializer.Parse<T>(eventArgsJson, SerializerOptions);
+        }
+
+        private static UIChangeEventArgs DeserializeUIEventChangeArgs(string eventArgsJson)
+        {
+            var changeArgs = Deserialize<UIChangeEventArgs>(eventArgsJson);
+            var jsonElement = (JsonElement)changeArgs.Value;
+            switch (jsonElement.Type)
+            {
+                case JsonValueType.Null:
```
rynowak: should we allow undefined and treat the same as null? or not going to show up here?

src/Components/Components/ref/Microsoft.AspNetCore.Components.netstandard2.0.cs
```diff
@@ -91,11 +91,11 @@ public partial class DataTransfer
...
     }
-    [System.Runtime.InteropServices.StructLayoutAttribute(System.Runtime.InteropServices.LayoutKind.Sequential)]
-    public readonly partial struct ElementRef : Microsoft.JSInterop.Internal.ICustomArgSerializer
+    public sealed partial class ElementRef
     {
-        private readonly object _dummy;
-        object Microsoft.JSInterop.Internal.ICustomArgSerializer.ToJsonPrimitive() { throw null; }
+        internal ElementRef() { }
+        [System.Text.Json.Serialization.JsonPropertyNameAttribute("_blazorElementRef")]
+        public string Id { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
```
rynowak: This should be `EditorVisibleNever`? Is there anything meaningful a user could do with this?

src/Components/Components/src/HttpClientJsonExtensions.cs
```diff
@@ -21,10 +26,10 @@ public static class HttpClientJsonExtensions
         /// <param name="httpClient">The <see cref="HttpClient"/>.</param>
         /// <param name="requestUri">The URI that the request will be sent to.</param>
         /// <returns>The response parsed as an object of the generic type.</returns>
-        public static async Task<T> GetJsonAsync<T>(this HttpClient httpClient, string requestUri)
+        public static async ValueTask<T> GetJsonAsync<T>(this HttpClient httpClient, string requestUri)
         {
-            var responseJson = await httpClient.GetStringAsync(requestUri);
-            return Json.Deserialize<T>(responseJson);
+            var responseStream = await httpClient.GetStreamAsync(requestUri);
+            return await JsonSerializer.ReadAsync<T>(responseStream, HttpClientSerializationOptions);
```
rynowak: This isn't safe or correct. This will only work if it's UTF-8 😭

src/Components/Components/src/HttpClientJsonExtensions.cs
```diff
@@ -111,8 +116,8 @@ public static async Task<T> SendJsonAsync<T>(this HttpClient httpClient, HttpMet
             }
             else
             {
-                var responseJson = await response.Content.ReadAsStringAsync();
-                return Json.Deserialize<T>(responseJson);
+                var responseStream = await response.Content.ReadAsStreamAsync();
+                return await JsonSerializer.ReadAsync<T>(responseStream, HttpClientSerializationOptions);
```
rynowak: Same as above, only works for UTF8

src/Components/test/E2ETest/Tests/KeyTest.cs
```diff
@@ -19,6 +19,11 @@ namespace Microsoft.AspNetCore.Components.E2ETest.Tests
 {
     public class KeyTest : BasicTestAppTestBase
     {
+        private readonly JsonSerializerOptions _jsonSerializerOptions = new JsonSerializerOptions
+        {
+            PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
+        };
```
rynowak: So many copies of this. Would be nice to have this exposed somewhere.

src/Components/Components/src/HttpClientJsonExtensions.cs
```diff
@@ -13,6 +13,11 @@ namespace Microsoft.AspNetCore.Components
     /// </summary>
     public static class HttpClientJsonExtensions
     {
+        private static readonly JsonSerializerOptions HttpClientSerializationOptions = new JsonSerializerOptions
```
rynowak: Spicy thought - @SteveSandersonMS - do we have somewhere specific to client-side blazor to put these code?

src/Components/test/E2ETest/Tests/KeyTest.cs
```diff
@@ -33,7 +38,7 @@ protected override void InitializeAsyncCore()
             Navigate(ServerPathBase, noReload: !_serverFixture.UsingAspNetHost);
         }
 
-        [Fact]
+        [Fact(Skip = "https://github.com/dotnet/corefx/issues/37313")]
```
rynowak: This feels like we ought to block the integration on this fix. There are enough tests skipped that I have the heebie-jeebies.

src/Mvc/Mvc.ViewFeatures/src/Infrastructure/UnsupportedJavaScriptRuntime.cs
```diff
@@ -13,10 +13,5 @@ public Task<T> InvokeAsync<T>(string identifier, params object[] args)
         {
             throw new InvalidOperationException("JavaScript interop calls cannot be issued during server-side prerendering, because the page has not yet loaded in the browser. Prerendered components must wrap any JavaScript interop calls in conditional logic to ensure those interop calls are not attempted during prerendering.");
         }
-
-        public void UntrackObjectRef(DotNetObjectRef dotNetObjectRef)
```
rynowak: dead code? Isn't this called by interop?

src/Components/Blazor/Blazor/src/Http/WebAssemblyHttpMessageHandler.cs
```diff
@@ -114,10 +119,25 @@ public class WebAssemblyHttpMessageHandler : HttpMessageHandler
         private static byte[] AllocateArray(string length) => new byte[int.Parse(length)];
 #pragma warning restore IDE0051 // Remove unused private members
 
-        private string[][] GetHeadersAsStringArray(HttpRequestMessage request)
-            => (from header in request.Headers.Concat(request.Content?.Headers ?? Enumerable.Empty<KeyValuePair<string, IEnumerable<string>>>())
-                from headerValue in header.Value // There can be more than one value for each name
-                select new[] { header.Key, headerValue }).ToArray();
+        private static IReadOnlyList<Header> GetHeaders(HttpRequestMessage request)
```
SteveSandersonMS: What? You don't like my code-golfing LINQ expressions??? Look how many more lines we have now!

Oh sure your code is much easier to understand and debug, but think of the extra disk space this consumes!

src/Components/Components/ref/Microsoft.AspNetCore.Components.netstandard2.0.cs
SteveSandersonMS: Agreed - hide this if possible. Possibly give it an unfriendly name too. It's unfortunate that we have to make it public.

SteveSandersonMS: Seeing that you're using `[JsonPropertyName]`, we could definitely give it a name like `__internalId`.

src/Components/Components/src/ElementRef.cs
```diff
@@ -1,22 +1,27 @@
...
+using System.Text.Json.Serialization;
 using System.Threading;
 
 namespace Microsoft.AspNetCore.Components
 {
     /// <summary>
     /// Represents a reference to a rendered element.
     /// </summary>
-    public readonly struct ElementRef : ICustomArgSerializer
+    public sealed class ElementRef
```
SteveSandersonMS: Could you clarify why this is now becoming a `class`?

src/Components/Components/src/HttpClientJsonExtensions.cs
SteveSandersonMS: Which code? All of the `HttpClientJsonExtensions`? Not really because it's mean to be part of the programming model that works in either place (e.g., when sharing components across environments).

Please let me know if you meant something different.

src/Mvc/Mvc.ViewFeatures/src/Infrastructure/UnsupportedJavaScriptRuntime.cs
SteveSandersonMS: It's being removed in https://github.com/aspnet/Extensions/pull/1704

src/Components/Components/src/ElementRef.cs
pranavkm: As of when I tried, System.Text.Json did not support serializing value types. I think it's ongoing work.

src/Components/Browser.JS/dist/Debug/blazor.webassembly.js
pranavkm: Yes. We landed on checking this file in and adding a code-check to make sure it's up to date with the ts files rather than produce it as part of every build. Some of our build environments don't have `yarn` and there were some build from source concerns where we wanted to avoid `yarn`

src/Components/Blazor/Build/src/Core/BootJsonWriter.cs
pranavkm: User visible?

src/Components/Blazor/Blazor/src/Http/WebAssemblyHttpMessageHandler.cs
pranavkm: System.Text did not support value types when I tried this.

src/Components/Components/src/ElementRef.cs
SteveSandersonMS: Since this is part of the public API, the class-or-structness is going get fixed forever once we ship 3.0. It would be very sad if we ended up doing unnecessary allocations for every `ElementRef` forever!

Do you know if `System.Text.Json` is tracking fixing this on a particular timescale?

Could we have an issue in our repo to track reverting this on a suitable milestone?

pranavkm: It's currently out of scope for 3.0. https://github.com/dotnet/corefx/issues/37434. I'll file a tracking issue for this - maybe if we could make this work with a custom converter.

src/Components/Blazor/Blazor/src/Http/WebAssemblyHttpMessageHandler.cs
YohDeadfall: Now it supports value types as well as reference types.

ahsonkhan: Yep, its fixed: https://github.com/dotnet/corefx/issues/37434#issuecomment-494635210

pranavkm: Unless I'm mistaken, this using a `readonly struct` would not work since the serializer does not support deserializing in to ctor arguments. Plus it looks like you end up boxing the struct on deserialization, so it seems less useful. Might be more useful to switch to KVP if the serializer natively supports it: https://github.com/dotnet/corefx/issues/36643?

src/Components/Blazor/Build/src/Core/BootJsonWriter.cs
rynowak: I don't think it need to be.

src/Components/Blazor/Blazor/src/Http/WebAssemblyHttpMessageHandler.cs
rynowak: ok well let's leave it as-is then.

src/Components/Browser/src/RendererRegistryEventDispatcher.cs
pranavkm: @ahsonkhan \ @JeremyKuhne when would JsonElement return `Undefined`?

JeremyKuhne: When there was no value in the Json.

ahsonkhan: For example, if you had a default instance of `JsonElement` or if the underlying document was disposed.

I would consider that to be an exceptional case, so throwing is probably best.

cc @bartonjs

src/Components/Components/src/ElementRef.cs
pranavkm: This is back to being a reference type now.

pranavkm:
🆙 📅 

This PR is blocked on https://github.com/aspnet/Extensions/pull/1704, https://github.com/aspnet/Extensions/pull/1748. All the tests pass if I produce a fraken-build locally.

src/Components/Browser/src/RendererRegistryEventDispatcher.cs
pranavkm: Leaving this as-is for now. We can wait for a user report since none of the tests currently run in to an undefined value.

eng/Versions.props
```diff
@@ -109,8 +109,8 @@
     <MicrosoftExtensionsValueStopwatchSourcesPackageVersion>3.0.0-preview6.19274.1</MicrosoftExtensionsValueStopwatchSourcesPackageVersion>
     <MicrosoftExtensionsWebEncodersPackageVersion>3.0.0-preview6.19274.1</MicrosoftExtensionsWebEncodersPackageVersion>
     <MicrosoftInternalExtensionsRefsPackageVersion>3.0.0-preview6.19274.1</MicrosoftInternalExtensionsRefsPackageVersion>
-    <MicrosoftJSInteropPackageVersion>3.0.0-preview6.19274.1</MicrosoftJSInteropPackageVersion>
-    <MonoWebAssemblyInteropPackageVersion>3.0.0-preview6.19274.1</MonoWebAssemblyInteropPackageVersion>
+    <MicrosoftJSInteropPackageVersion>3.0.0-dev</MicrosoftJSInteropPackageVersion>
+    <MonoWebAssemblyInteropPackageVersion>3.0.0-dev</MonoWebAssemblyInteropPackageVersion>
```
dougbu: Don't check in changes to this file. These changes will only work in local builds

dougbu (changes requested):

src/Components/Components/src/ElementRef.cs
YohDeadfall: You can write a custom value converter which can be removed when the serializer supports constructors.

src/Components/Browser/src/RendererRegistryEventDispatcher.cs
bartonjs: Unless someone explicitly wrote `JsonElement element = default;` the only practical way they'd see a JsonElement whose Type was Undefined is reading the out-value from TryGetProperty when it returned false (because the API internally said `value = default;`).

SteveSandersonMS (approved):
This looks great to me now.

I don't know whether or not this feature as a whole is ready to merge for preview 6 - it depends on the readiness of System.Text.Json in general, and any other review comments you're still processing here. Perhaps we can confirm during our sync today.

PR #10179, opened 2019-05-11
Add localzied WXL files to bundles

joeloff:
Add localized .wxl files for shared framework and hosting bundle

### DIFF
--- a/src/Installers/Windows/SharedFrameworkBundle/1036/thm.wxl
+++ b/src/Installers/Windows/SharedFrameworkBundle/1036/thm.wxl
@@ -1,0 +1,61 @@
+<?xml version="1.0" encoding="utf-8"?>
+<WixLocalization Culture="en-us" Language="1033" xmlns="http://schemas.microsoft.com/wix/2006/localization">
+  <String Id="Caption">Installation de [BundleNameFull]</String>
+  <String Id="Title">[BundleNameShort]</String>
+  <String Id="SubTitle">[BundleNameSub]</String>
+  <String Id="InstallHeader">Bienvenue</String>
+  <String Id="InstallMessage">Le programme d'installation va installer [WixBundleName] sur votre ordinateur. Cliquez sur Installer pour continuer, sur Options pour définir le répertoire d'installation ou sur Fermer pour quitter le programme.</String>
+  <String Id="InstallVersion">Version de [WixBundleVersion]</String>
+  <String Id="ConfirmCancelMessage">Voulez-vous vraiment annuler ?</String>
+  <String Id="ExecuteUpgradeRelatedBundleMessage">Version précédente</String>
+  <String Id="HelpHeader">Aide à l'installation</String>
+  <String Id="HelpText">/install | /repair | /uninstall | /layout [répertoire] - installe, répare, désinstalle ou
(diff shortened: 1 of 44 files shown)


natemcmaster (approved):

src/Installers/Windows/SharedFrameworkBundle/1036/thm.wxl
```diff
@@ -0,0 +1,61 @@
...
+  <String Id="FailureInstallHeader">Échec de l'installation</String>
+  <String Id="FailureUninstallHeader">Échec de la désinstallation</String>
+  <String Id="FailureRepairHeader">Échec de la réparation</String> 
+  <String Id="FailureHyperlinkLogText">Un ou plusieurs problèmes sont à l'origine de l'échec de l'installation. Corrigez ces problèmes, puis recommencez l'installation. Pour plus d'informations, voir le &lt;a href="#"&gt;fichier journal&lt;/a&gt;.</String>
+  <String Id="FailureRestartText">Vous devez redémarrer votre ordinateur pour terminer l'opération de restauration du logiciel.</String>
+  <String Id="FailureRestartButton">&amp;Redémarrer</String>
+  <String Id="FailureCloseButton">&amp;Fermer</String>
+  <String Id="FilesInUseHeader">Fichiers en cours d'utilisation</String>
+  <String Id="FilesInUseLabel">Les applications suivantes utilisent des fichiers nécessitant une mise à jour :</String>
+  <String Id="FilesInUseCloseRadioButton">&amp;Fermer les applications essayez de les ouvrir de nouveau.</String>
```
PinpointTownes: FYI, this entry - that also appears in the server bundle installer - is not correct. The correct form would be:

```
Fermer les applications et essayer de les ouvrir de nouveau.
```

That said, it's a bit hard to read. Here's a much more natural version:

```
Fermer les applications et tenter de les redémarrer.
```

joeloff: Thank you for pointing this out. I take it you were referring to the "z"/"r" usage being inconsistent in the original translation which should have been Fermez instead?

I've updated the PR with your suggestion.

PinpointTownes: @joeloff thanks!

> I take it you were referring to the "z"/"r" usage being inconsistent in the original translation which should have been Fermez instead?

Actually, inconsistency wasn't the only issue. In French, a verb ending with `ez` without a personal pronoun (`vous` in this case) always indicates that a sentence uses the imperative mood with a second-person subject.

In this case, using the imperative mood would be - at beast - misleading, as it would imply that users have to close and restart the listed apps themselves before continuing (which would actually lead to the same error when trying to re-execute the installer, since the files would still be locked).

Here, the user is given a choice - allowing the installer to temporarily close apps holding locks on files it must update or restarting the computer - so an infinitive sentence is a much better option.

natemcmaster:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=187997

natemcmaster:
Looks like there was only one flaky test failure unrelated to this change. :shipit:

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/2423

Issue #8660, opened 2019-03-19
Add Pipelines-related http features in TestServer


anurse:
Epic https://github.com/aspnet/AspNetCore/issues/8829
Split from https://github.com/aspnet/AspNetCore/issues/7815

We should support the Pipe Goodness on TestServer.

We'd remove the stream-based implementation and use pipes. Stream-based implementation would then be a wrapper around the pipes.

JamesNK:
Something to test/implement with this issue: ensure that `Response.StartAsync` correctly sets `Response.HasStarted` to true in TestServer.

anurse:
Bumping out of preview4. I don't think this is a critical thing for preview 4, but feel free to debate :).

JamesNK:
I would like it. Does that make it critical? 🙃

Seriously though, I was looking into https://github.com/grpc/grpc-dotnet/issues/144 yesterday and TestServer never sets Response.HasStarted when the response content is flushed, and that makes our functional tests a little less trustworthy. For example, it is not an error to send headers after writing content. I'm considering moving some of the functional tests to use a spun up Kestrel instance because I don't trust how accurate TestServer is.

But I understand there is higher priority work.

anurse:
We just don't have the capacity for this in 3.0. I'll put this in 3.1 for now, and if someone feels so inspired, we can review a PR :)

anurse:
@JamesNK does your PR satisfy this need? Are we ready to close this?

JamesNK:
My PR satisfies my need to have StartAsync work correctly.

TestServer still needs to be refactored to use pipes.

anurse:
Ok, we'll keep this around. There may be capacity later when the unknown unknowns become known knowns.

anurse:
We don't expect to have the capacity in 3.0. Lots on the queue. This is a good candidate for 3.1 though.

alefranz:
I had a look at this and I believe that most of work has been done as part of https://github.com/dotnet/aspnetcore/issues/11598.

anurse:
Looks like you're correct and #11598 has a better listing of what is done and not yet done. I'm going to close this as a duplicate of it.

PR #10189, opened 2019-05-12
Improve TestServer support for Response.StartAsync

JamesNK:
Addresses #8660

re: https://github.com/aspnet/AspNetCore/issues/8660#issuecomment-476435564

### DIFF
--- a/src/Hosting/TestHost/src/ResponseFeature.cs
+++ b/src/Hosting/TestHost/src/ResponseFeature.cs
@@ -12,13 +12,15 @@ namespace Microsoft.AspNetCore.TestHost
 {
     internal class ResponseFeature : IHttpResponseFeature, IHttpResponseStartFeature
     {
+        private readonly HeaderDictionary _headers = new HeaderDictionary();
+        private readonly Action<Exception> _abort;
+
         private Func<Task> _responseStartingAsync = () => Task.FromResult(true);
         private Func<Task> _responseCompletedAsync = () => Task.FromResult(true);
-        private HeaderDictionary _headers = new HeaderDictionary();
         private int _statusCode;
         private string _reasonPhrase;
 
-        public ResponseFeature()
+        public ResponseFeature(Action<Exception> abort)
         {
             Headers = _headers;
             Body = new MemoryStream();
@@ -26,6 +28,7 @@ namespace Microsoft.AspNetCore.TestHost
             // 200 is the default status code all the way down to the host, so we set it
             // here to be consistent with the rest of the hosts when writing tests.
             StatusCode = 200;
+            _abort = abort;
         }
 
         public int StatusCode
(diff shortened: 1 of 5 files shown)


src/Hosting/TestHost/src/ResponseFeature.cs
```diff
@@ -98,14 +99,22 @@ public void OnCompleted(Func<object, Task> callback, object state)
...
         }
 
         public Task FireOnResponseCompletedAsync()
         {
             return _responseCompletedAsync();
         }
+
+        public Task StartAsync(CancellationToken token = default)
+        {
+            return FireOnSendingHeadersAsync();
```
Tratcher: If this throws then HasStarted doesn't get set and FireOnSendingHeadersAsync can be called again.

JamesNK: Done.

Is it correct that the response always gets returned when an error is thrown during StartAsync. I needed to do that to get `ExceptionFromOnStartingWithErrorHandlerIsReported` to pass. The test isn't doing what it says it is.

https://github.com/aspnet/AspNetCore/blob/11311d4f9d7eafcbe2ebafa7a5e8126244187a57/src/Hosting/TestHost/test/ClientHandlerTests.cs#L357-L380

src/Middleware/Diagnostics/test/UnitTests/ExceptionHandlerTest.cs
```diff
@@ -149,7 +150,7 @@ public async Task ClearsResponseBuffer_BeforeRequestIsReexecuted()
                     app.Run(async (context) =>
                     {
                         // Write some content into the response before throwing exception
-                        await context.Response.WriteAsync(new string('a', 100));
+                        await context.Response.Body.WriteAsync(Encoding.UTF8.GetBytes(new string('a', 100)));
```
Tratcher: Why did this change? These two lines should behave be identically.

Tratcher (commented):
Yeah, that's right. OnStarting exceptions are considered fatal in all servers, but in Test Server we make a point of flowing any useful exception directly to the client so we can debug it.

That test was added after we found that pattern causing a deadlock with large data filling the pipe. It was  real issue for the MVC tests. Feel free to clarify the naming but it's not wrong.

src/Middleware/Diagnostics/test/UnitTests/ExceptionHandlerTest.cs
JamesNK: Response.WriteAsync flushes. Exception handler can't run after flush.

Tratcher: I see, this is an odd test. The buffering prevented OnStarting from firing before, but now it fires directly from the WriteAsync extension. For this kind of buffering to really be effective it will need to suppress/wrapp the new StartAsync feature.

JamesNK:
Can this be merged as is or does it require more changes?

src/Middleware/Diagnostics/test/UnitTests/ExceptionHandlerTest.cs
JamesNK: What do you want to do? Can this be merged as is or does it require more changes?

Tratcher: Revert this line and then go to line 120 and remove IHttpResponseStartFeature from the feature collection. That's what you'd have to do to make this buffering-to-prevent-start scenario work.  Then :shipit:

Tratcher (approved):

PR #10204, opened 2019-05-13
Use UnsafeRegister for the handshake

davidfowl:


halter73 (approved):
Cool that this was added 👍

jkotalik (approved):

PR #10209, opened 2019-05-13
Danielmartin/update to jwt bearer events

danielDevelops:
Added new Jwt Bearer Event to allow for forbidden message capture.
 - Added new JwtBearerEvent that matches the pattern for the OnAuthenticationFailed event.
 - Implemented the same utilization for the override of the HandleForbiddenAsync as the HandleUnauthroizedAsync
- Will now allow for OnChallenge, OnFailedAuthorization, and OnFailedAuthentication


src/Security/Authentication/JwtBearer/ref/Microsoft.AspNetCore.Authentication.JwtBearer.netcoreapp3.0.cs
```diff
@@ -26,10 +26,12 @@ public partial class JwtBearerEvents
...
         public System.Func<Microsoft.AspNetCore.Authentication.JwtBearer.MessageReceivedContext, System.Threading.Tasks.Task> OnMessageReceived { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public System.Func<Microsoft.AspNetCore.Authentication.JwtBearer.TokenValidatedContext, System.Threading.Tasks.Task> OnTokenValidated { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public virtual System.Threading.Tasks.Task AuthenticationFailed(Microsoft.AspNetCore.Authentication.JwtBearer.AuthenticationFailedContext context) { throw null; }
+        public virtual System.Threading.Tasks.Task AuthorizationFailed(Microsoft.AspNetCore.Authentication.JwtBearer.AuthenticationFailedContext context) { throw null; }
```
Tratcher: "Authorization" is out of place in an Authentication component. Why not just call it "Forbidden"?

src/Security/Authentication/JwtBearer/src/JwtBearerEvents.cs
```diff
@@ -15,6 +15,10 @@ public class JwtBearerEvents
         /// Invoked if exceptions are thrown during request processing. The exceptions will be re-thrown after this event unless suppressed.
         /// </summary>
         public Func<AuthenticationFailedContext, Task> OnAuthenticationFailed { get; set; } = context => Task.CompletedTask;
+        /// <summary>
+        /// Invoked if exceptions are thrown during the authorzation process.  The exceptions will be re-thrown after this event unless supressed.
```
Tratcher: This text is wrong. 403's aren't the result of exceptions.

src/Security/Authentication/JwtBearer/src/JwtBearerHandler.cs
```diff
@@ -265,6 +265,15 @@ protected override async Task HandleChallengeAsync(AuthenticationProperties prop
...
 
+        protected override Task HandleForbiddenAsync(AuthenticationProperties properties)
+        {
+            var authenticationFailedContext = new AuthenticationFailedContext(Context, Scheme, Options)
+            {
+                Exception = new Exception("Forbidden")
+            };
+            Response.StatusCode = 403;
+            return Events.AuthorizationFailed(authenticationFailedContext);
+        }
```
Tratcher: Empty line between methods.
```suggestion
        }
        
```

src/Security/Authentication/JwtBearer/src/JwtBearerEvents.cs
```diff
@@ -32,6 +36,7 @@ public class JwtBearerEvents
         public Func<JwtBearerChallengeContext, Task> OnChallenge { get; set; } = context => Task.CompletedTask;
 
         public virtual Task AuthenticationFailed(AuthenticationFailedContext context) => OnAuthenticationFailed(context);
+        public virtual Task AuthorizationFailed(AuthenticationFailedContext context) => OnAuthorizationFailed(context);
```
Tratcher: Empty line between methods.
```suggestion

        public virtual Task AuthorizationFailed(AuthenticationFailedContext context) => OnAuthorizationFailed(context);
```

src/Security/Authentication/JwtBearer/ref/Microsoft.AspNetCore.Authentication.JwtBearer.csproj
```diff
@@ -1,11 +1,13 @@
+
 <!-- This file is automatically generated. -->
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFrameworks>netcoreapp3.0</TargetFrameworks>
   </PropertyGroup>
-  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp3.0'">
+  <ItemGroup Condition=" == netcoreapp3.0">
```
Tratcher: revert

src/Security/Authentication/JwtBearer/src/JwtBearerHandler.cs
```diff
@@ -265,6 +265,15 @@ protected override async Task HandleChallengeAsync(AuthenticationProperties prop
             }
         }
 
+        protected override Task HandleForbiddenAsync(AuthenticationProperties properties)
+        {
+            var authenticationFailedContext = new AuthenticationFailedContext(Context, Scheme, Options)
```
Tratcher: Don't re-use AuthenticationFailedContext, create a new context type.

src/Security/Authentication/JwtBearer/src/JwtBearerHandler.cs
```diff
@@ -265,6 +265,15 @@ protected override async Task HandleChallengeAsync(AuthenticationProperties prop
...
+            {
+                Exception = new Exception("Forbidden")
```
Tratcher: There's no reason to provide an exception here.

Tratcher (changes requested):
Needs a test.

What's your scenario for this new event? How would you customize the response?

src/Security/Authentication/JwtBearer/src/JwtBearerHandler.cs
danielDevelops: Reviewed and updated as suggested.

src/Security/Authentication/JwtBearer/ref/Microsoft.AspNetCore.Authentication.JwtBearer.netcoreapp3.0.cs
danielDevelops: Updated as suggested.

src/Security/Authentication/JwtBearer/src/JwtBearerEvents.cs
danielDevelops: Yes thank you.  I have updated this to follow the pattern for the OnChallenge

src/Security/Authentication/JwtBearer/src/JwtBearerHandler.cs
danielDevelops: Corrected.

src/Security/Authentication/JwtBearer/src/JwtBearerEvents.cs
danielDevelops: Corrected

src/Security/Authentication/JwtBearer/src/JwtBearerHandler.cs
danielDevelops: I have created a new context as suggested and applied this.

src/Security/Authentication/JwtBearer/src/JwtBearerHandler.cs
danielDevelops: removed this exception

blowdart:
Weren't we attaching PRs to issues to allow for discussion of PRs in issues first? 

What problem is this PR solving?

Tratcher:
Yes, we do prefer there be issues filed to track & design any significant feature work like this.

src/Security/Authentication/JwtBearer/ref/Microsoft.AspNetCore.Authentication.JwtBearer.csproj
danielDevelops: Corrected

danielDevelops:
> Yes, we do prefer there be issues filed to track & design any significant feature work like this.

I apologize, I didn't realize that was procedure.  This was an issue that I ran into a few weeks ago and thought it was a simple thing to add to the JwtBearerEvents.  

The specific use case is when there is an endpoint where the Authorization handler didn't allow the connection and you want to send back a custom response.  In my specific case I was running a mixed mode site with both Jwt and Cookie authentication.  The site would capture the 403 and do a redirect based to the error page.  

Ultimately this was to allow the capture of the 403 and return JSON data rather than a full page.

danielDevelops:
> Needs a test.
> 
> What's your scenario for this new event? How would you customize the response?

I have added tests as requested.

The scenario for the event would be to stop the 403 from performing any other redirects handled by other middleware.  This would be to capture that response and stop further processing writing a single message or JSON response.

Tratcher:
> The site would capture the 403 and do a redirect based to the error page

What was doing the redirect? The StatusCodes middleware?

danielDevelops:
> > The site would capture the 403 and do a redirect based to the error page
> 
> What was doing the redirect? The StatusCodes middleware?

Yes.  I was able to bypass that by preventing it on everything for a route but it made sense to me to be able to use this the same as you can with the cookie middleware and stop the further execution of the middleware in the event that a forbidden is captured.   

When working on developing a solution outside of this code update everything seemed very hacky and didn't have a simple solution.  By adding this here, it makes a much more clean solution to stop and grab that execution.

Tratcher:
Did you actually want JSON, or did you just want to suppress the Redirect/HTML? While there's nothing wrong with this new event there are existing solutions to that issue, some better documented than others.

- Using IStatusCodePagesFeature to opt out of status code pages. https://docs.microsoft.com/en-us/aspnet/core/fundamentals/error-handling?view=aspnetcore-2.2#disable-status-code-pages
- SkipStatusCodePagesAttribute - An MVC attribute for IStatusCodePagesFeature https://github.com/aspnet/AspNetCore/blob/c565386a3ed135560bc2e9017aa54a950b4e35dd/src/Mvc/Mvc.ViewFeatures/src/SkipStatusCodePagesAttribute.cs. I wonder if it runs before of after Authorize?
- UseStatusCodePagesWithReExecute - Doesn't redirect, it re-executes the request to generate a view. That view could use content negotiation to avoid sending HTML.

danielDevelops:
For my specific use case I am embedding it in a NuGet so having it done at the controller level didn't make sense since then it would require other developers to implement these features at another level.  I was just hoping to mirror what I can do with the Cookie Authorization as the Jwt Authorization.  I did find a solution to this by adding extending the JwtBearerHandler class and implementing my own at that level but then it wasn't passing the executing code down as an event in the same way as the Cookie Handler.  It just seemed as a gap to me in the functionality.

blowdart:
This is turning into "Make the auth pieces do UI actions", which we've tried very hard to move away from. This appears to be adding a coupling I'm not comfortable with.

Tratcher:
There's nothing wrong with having the event, we have one for every other code path. It would be a good place to set IStatusCodePagesFeature. Then it wouldn't have to do any UI stuff, just opt out.

src/Security/Authentication/test/JwtBearerTests.cs
```diff
@@ -678,7 +678,128 @@ public async Task EventOnChallengeSkip_ResponseNotModified()
             Assert.Empty(response.Response.Headers.WwwAuthenticate);
             Assert.Equal(string.Empty, response.ResponseText);
         }
+        [Fact]
```
Tratcher: Empty lines between tests.
```suggestion

        [Fact]
```

src/Security/Authentication/test/JwtBearerTests.cs
```diff
@@ -678,7 +678,128 @@ public async Task EventOnChallengeSkip_ResponseNotModified()
...
+                        return Task.FromResult(0);
+                    }
+                };
+            });
+            var newBearerToken = "Bearer " + tokenText;
+            var response = await SendAsync(server, "http://example.com/forbidden", newBearerToken);
+            Assert.Equal(HttpStatusCode.Forbidden, response.Response.StatusCode);
+        }
+        [Fact]
+        public async Task EventOnForbidden_ResponseNotModified()
```
Tratcher: How is this test different from the one above?

src/Security/Authentication/test/JwtBearerTests.cs
```diff
@@ -678,7 +678,128 @@ public async Task EventOnChallengeSkip_ResponseNotModified()
...
+            });
+            var newBearerToken = "Bearer " + tokenText;
+            var response = await SendAsync(server, "http://example.com/forbidden", newBearerToken);
+            Assert.Equal(HttpStatusCode.Forbidden, response.Response.StatusCode);
+        }
+        [Fact]
+        public async Task EventOnForbidden_ResponseModified()
+        {
+            var key = new SymmetricSecurityKey(Encoding.UTF8.GetBytes(new string('a', 128)));
+            var creds = new SigningCredentials(key, SecurityAlgorithms.HmacSha256);
```
Tratcher: Can you pull most of this token creation logic out to a shared method?

src/Security/Authentication/test/JwtBearerTests.cs
```diff
@@ -678,7 +678,128 @@ public async Task EventOnChallengeSkip_ResponseNotModified()
...
+                {
+                    ValidIssuer = "issuer.contoso.com",
+                    ValidAudience = "audience.contoso.com",
+                    IssuerSigningKey = key,
+                };
+                o.Events = new JwtBearerEvents()
+                {
+                    OnForbidden = context => 
+                    {
+                        context.Response.StatusCode = 403;
```
Tratcher: Test an alternate status code.
```suggestion
                        context.Response.StatusCode = 418;
```

src/Security/Authentication/JwtBearer/src/JwtBearerEvents.cs
```diff
@@ -15,6 +15,10 @@ public class JwtBearerEvents
         /// Invoked if exceptions are thrown during request processing. The exceptions will be re-thrown after this event unless suppressed.
         /// </summary>
         public Func<AuthenticationFailedContext, Task> OnAuthenticationFailed { get; set; } = context => Task.CompletedTask;
+        /// <summary>
```
Tratcher:
```suggestion

        /// <summary>
```

Tratcher (commented):
Code looks about ready, just some aesthetic cleanup.

Tratcher:
Not sure what's up with Code check. I'll check on that one for you.

src/Security/Authentication/test/JwtBearerTests.cs
danielDevelops: Updated

src/Security/Authentication/test/JwtBearerTests.cs
danielDevelops: Thanks for the catch, it should have not had the event in one of them.

src/Security/Authentication/test/JwtBearerTests.cs
danielDevelops: Done

src/Security/Authentication/test/JwtBearerTests.cs
danielDevelops: Agreed and done

src/Security/Authentication/JwtBearer/src/JwtBearerEvents.cs
danielDevelops: Done

src/Security/Authentication/JwtBearer/src/ForbiddenContext.cs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.AspNetCore.Http;
+
+namespace Microsoft.AspNetCore.Authentication.JwtBearer
+{
+    public class ForbiddenContext : ResultContext<JwtBearerOptions>
```
Tratcher:
```suggestion
    public class ForbiddenContext : BaseContext<JwtBearerOptions>
```

Tratcher (approved):

src/Security/Authentication/JwtBearer/ref/Microsoft.AspNetCore.Authentication.JwtBearer.csproj
```diff
@@ -1,3 +1,4 @@
+
```
Tratcher: I think CodeCheck is complaining about the extra lines in this file. Revert.

danielDevelops: done

anurse:
Looks like we've still got a code check issue. I'll re-kick the build but if you're still interested in contributing this you'll need to re-run the reference assembly generation code (as described in https://github.com/aspnet/AspNetCore/blob/master/docs/ReferenceAssemblies.md)

Otherwise we won't be able to take this in 3.0

danielDevelops:
> Looks like we've still got a code check issue. I'll re-kick the build but if you're still interested in contributing this you'll need to re-run the reference assembly generation code (as described in https://github.com/aspnet/AspNetCore/blob/master/docs/ReferenceAssemblies.md)
> 
> Otherwise we won't be able to take this in 3.0

I did the required code check process, but when i ran it, the only change it tried to make was to add two lines to the csproj file change the target framework

 <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp3.0'"> to <ItemGroup Condition=" == netcoreapp3.0">

The first time i checked in after running the required scripts, i was told to revert this back (rightly so).

Tratcher:
I'll give it a try.

Tratcher:
Replaced by https://github.com/aspnet/AspNetCore/pull/12133 with CodeCheck fixes.

PR #10214, opened 2019-05-13
Do not throw when a status code with a codefix appears multiple times…

pranavkm:
… in the method body

Fixes https://github.com/aspnet/AspNetCore/issues/4480

## Impact
This is a bug fix in an analyzer that currently crashes when running in to a fairly common scenario.

Workaround
No workaround is available outside of not using the analyzer.

Risk
Since no user code \ production scenario is impacted, and since users have to manually upgrade the package to get this change, this is a fairly low risk change.


pranavkm:
Le sigh. GitHub really doesn't want to forget about the old PR

Issue #4744, opened 2018-06-20
Support HTTP/2 response headers with END_STREAM, no body


Tratcher:
If the response completes without a body there's no reason to send a data frame. Flush the headers with an End_Stream flag.

anurse:
We don't expect to have capacity for this in 3.0. If gRPC wants it to improve performance, you're welcome to have it :) (@jamesnk @juntaoluo @Eilon)

anurse:
Backlogging, as per my comment above. Again, if gRPC has higher priorities here, let us know and we can revisit.

JamesNK:
I believe this is needed to properly support gRPC "Trailers-Only" responses. Some clients require that gRPC status is in the HEADERS frame with end stream set. Kestrel having a separate DATA frame with end stream set makes this not possible.

https://github.com/grpc/grpc-dotnet/issues/241

Tratcher:
Those aren't trailers anymore...

JamesNK:
That's why I air-quoted it 😋

anurse:
@shirhatti confirms that this was verified against a nightly build in gRPC interop tests (cc @JamesNK)

PR #10219, opened 2019-05-14
HTTP2: Include END_STREAM with HEADERS frame to avoid empty DATA frame

JamesNK:
Addresses #4744

### DIFF
--- a/src/Servers/Kestrel/Core/src/Internal/Http2/Http2OutputProducer.cs
+++ b/src/Servers/Kestrel/Core/src/Internal/Http2/Http2OutputProducer.cs
@@ -143,15 +143,16 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
                     return;
                 }
 
-                // If the responseHeaders will be written to the final frame then set
-                // END_STREAM on the HEADERS frame. This avoids the need to write an
-                // empty DATA frame with END_STREAM
+                // If the responseHeaders will be written as the final HEADERS frame then
+                // set END_STREAM on the HEADERS frame. This avoids the need to write an
+                // empty DATA frame with END_STREAM.
                 //
-                // The headers will be the final frame if there is no content and there
-                // is no trailing HEADERS frame.
+                // The headers will be the final frame if:
+                // 1. There is no content
+                // 2. There is no trailing HEADERS frame.
                 Http2HeadersFrameFlags http2HeadersFrame;
 
-                if (appCompleted && responseHeaders.ContentLength == 0 && (_stream.Trailers == null || _stream.Trailers.Count == 0))
+                if (appCompleted && !_startedWritingDataFrames && (_stream.Trailers == null || _stream.Trailers.Count == 0))
                 {
                     _streamEnded = true;
                     http2HeadersFrame = Http2HeadersFrameFlags.END_STREAM;
@@ -326,6 +327,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
                     }
                     else if (readResult.IsCompleted && _streamEnded)
                     {
+                        Debug.Assert(readResult.Buffer.Length == 0);
+
                         // Headers have already been written and there is no other content to write
                         flushResult = await _frameWriter.FlushAsync(outputAborter: null, cancellationToken: default);
                     }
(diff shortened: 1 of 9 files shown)


src/Servers/Kestrel/Core/src/Internal/Http2/Http2OutputProducer.cs
```diff
@@ -296,13 +315,20 @@ private async ValueTask<FlushResult> ProcessDataWrites()
...
                         {
                             flushResult = await _frameWriter.WriteDataAsync(_streamId, _flowControl, readResult.Buffer, endStream: false);
                         }
 
                         flushResult = await _frameWriter.WriteResponseTrailers(_streamId, _stream.Trailers);
                     }
+                    else if (readResult.IsCompleted && _streamEnded)
+                    {
+                        // Headers have already been written and there is no other content to write
+                        flushResult = await _frameWriter.FlushAsync(outputAborter: null, cancellationToken: default);
```
jkotalik: Where else is FlushAsync called? Can we add a method to FrameWriter that takes no parameters instead?

jkotalik (commented):
We should try to get TechEmpower benchmarks for this 😄 

Besides that the change looks clean.

src/Servers/Kestrel/Core/src/Internal/Http2/Http2OutputProducer.cs
```diff
@@ -142,7 +143,25 @@ public void WriteResponseHeaders(int statusCode, string ReasonPhrase, HttpRespon
...
-                _frameWriter.WriteResponseHeaders(_streamId, statusCode, responseHeaders);
+                // If the responseHeaders will be written to the final frame then set
+                // END_STREAM on the HEADERS frame. This avoids the need to write an
+                // empty DATA frame with END_STREAM
+                //
+                // The headers will be the final frame if there is no content and there
+                // is no trailing HEADERS frame.
+                Http2HeadersFrameFlags http2HeadersFrame;
+
+                if (appCompleted && responseHeaders.ContentLength == 0 && (_stream.Trailers == null || _stream.Trailers.Count == 0))
```
Tratcher: Why rely on ContentLength 0? Wouldn't it be better to check if the app completed without ever starting to write to the body?

src/Servers/Kestrel/Core/src/Internal/Http2/Http2OutputProducer.cs
```diff
@@ -296,13 +315,20 @@ private async ValueTask<FlushResult> ProcessDataWrites()
...
                         if (readResult.Buffer.Length > 0)
                         {
                             flushResult = await _frameWriter.WriteDataAsync(_streamId, _flowControl, readResult.Buffer, endStream: false);
                         }
 
                         flushResult = await _frameWriter.WriteResponseTrailers(_streamId, _stream.Trailers);
                     }
+                    else if (readResult.IsCompleted && _streamEnded)
+                    {
+                        // Headers have already been written and there is no other content to write
```
Tratcher: Assert readResult.Buffer.Length == 0

src/Servers/Kestrel/Core/src/Internal/Http2/Http2OutputProducer.cs
JamesNK: This is needed for the case where someone has written to a span on the underlying PipeReader, then set the content length.

> Wouldn't it be better to check if the app completed without ever starting to write to the body?

How do I do that?

src/Servers/Kestrel/Core/src/Internal/Http2/Http2OutputProducer.cs
JamesNK: Everything calls down to TimingPipeFlusher.FlushAsync with these arguments in the end. I think it could change to `Http2OutputProducer.FlushAsync` but that will eventually call back to this method anyway 🤷‍♂

JamesNK:
> We should try to get TechEmpower benchmarks for this 😄

I don't know how to do that. And I thought TechEmpower was HTTP/1 only?

src/Servers/Kestrel/Core/src/Internal/Http2/Http2OutputProducer.cs
JamesNK: I guess you mean _startedWritingDataFrames. I've changed to use that.

JamesNK:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=188904

aspnet-hello:

Tratcher:
> We should try to get TechEmpower benchmarks for this

I doubt it will show anything, the benchmarks don't cover these scenarios.

Tratcher (approved):

halter73 (approved):

halter73:
I like how this turned into a net-negative lines of code change because of all the expected response frames it removed from our tests 👍

PR #10227, opened 2019-05-14
Components auth: basic services and components

SteveSandersonMS:
The first step in implementing the auth spec, https://gist.github.com/SteveSandersonMS/60ca3a5f70a7f42fba14981add7e7f79

This PR contains:

 * The underlying DI service, `IAuthenticationStateProvider`, and its server-side implementation that simply passes through the state from `HttpContext.User`
 * The `AuthenticationStateProvider` component that adapts `IAuthenticationStateProvider` into a cascading value, so components can easily observe changes to the authentication state and optionally use it in procedural logic
 * (**Partial**) The `AuthorizeView` component that consumes the auth state cascading value and lets developers easily show different UI states depending on authorization conditions purely using declarative code.
   * This is a partial implementation, because it's waiting for #10021 to be merged. Until then, the authorization rule is purely "you're authorized if authenticated", as it lacks parameters to control authorization conditions based on policies/roles.

More auth PRs will follow.


SteveSandersonMS:
Review note: I'll be adding E2E tests later in process of doing the whole auth thing, once the feature set is more complete. So not in this PR.

src/Components/Components/src/Auth/AuthenticationStateProvider.razor
```diff
@@ -0,0 +1,43 @@
+@namespace Microsoft.AspNetCore.Components
+@implements IDisposable
+
+<CascadingValue T="IAuthenticationState" Value="@_currentAuthenticationState" ChildContent="@ChildContent" />
+
+@functions {
+    private IAuthenticationState _currentAuthenticationState;
+
+    [Inject] private IAuthenticationStateProvider AuthStateProviderService { get; set; }
```
javiercn: nit: ProviderService makes me sad. It's like UtilityHelper. I would either pick Provider or Service

src/Components/Components/src/Auth/IAuthenticationStateProvider.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// </summary>
+    public interface IAuthenticationStateProvider
+    {
+        /// <summary>
+        /// Gets an <see cref="IAuthenticationState"/> instance that describes
+        /// the current user.
+        /// </summary>
+        /// <param name="forceRefresh">If true, instructs the provider to re-determine the user's authentication state, which can be an expensive operation. If false, the provider may reuse data cached for the current user.</param>
+        /// <returns>An <see cref="IAuthenticationState"/> instance that describes the current user.</returns>
+        Task<IAuthenticationState> GetAuthenticationStateAsync(bool forceRefresh);
```
javiercn: What would be a good scenario for ForceRefresh = true?

src/Components/Components/src/Auth/PendingAuthenticationState.cs
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Security.Claims;
+
+namespace Microsoft.AspNetCore.Components
+{
+    internal class PendingAuthenticationState : IAuthenticationState
+    {
```
javiercn: I would call this AnonymousAuthenticationState as this represents a user that hasn't been authenticated.

src/Components/Components/src/Auth/AuthenticationStateProvider.razor
SteveSandersonMS: OK, this name isn't part of the public API surface but I'm happy to change it.

src/Components/Components/src/Auth/IAuthenticationStateProvider.cs
SteveSandersonMS: When you're performed some action that you think may have invalidated the state (e.g., issued a logout request), then you need some way of checking how the state has now changed. This is only relevant in the client-side Blazor case (or custom `IAuthenticationStateProvider`) because that's the only place where it's even possible to change auth state without reloading the page.

The actual API may change. It looks like this in the spec but it could end up being a separate method (`RefreshAuthenticationState()`) by the time we review the end product.

src/Components/Components/src/Auth/PendingAuthenticationState.cs
SteveSandersonMS: TBH I prefer `PendingAuthenticationState` because it doesn't *only* represent a user that hasn't been authenticated: it represents a subset of those, ones who not only haven't been authenticated, but also have an authentication process currently in flight.

In any case this is not part of the public API surface.

src/Components/Server/src/Circuits/DefaultCircuitFactory.cs
```diff
@@ -41,8 +41,10 @@ internal class DefaultCircuitFactory : CircuitFactory
             var encoder = scope.ServiceProvider.GetRequiredService<HtmlEncoder>();
             var jsRuntime = (RemoteJSRuntime)scope.ServiceProvider.GetRequiredService<IJSRuntime>();
             var componentContext = (RemoteComponentContext)scope.ServiceProvider.GetRequiredService<IComponentContext>();
+            var authenticationStateProvider = (FixedAuthenticationStateProvider)scope.ServiceProvider.GetRequiredService<IAuthenticationStateProvider>();
             jsRuntime.Initialize(client);
             componentContext.Initialize(client);
+            authenticationStateProvider.Initialize(httpContext.User);
```
javiercn: Side note: We keep running into these type of patterns where we need to initialize stuff at some point in the lifecycle. I'm wondering if we are missing an abstraction here (that can potentially also be useful for customers).

It feels to me that we are missing some sort of platform "initializer" with maybe individual initializers for different platforms. Something that we and people can implement and that will get resolved from DI and called in at the right time to initialize platform dependent stuff like Auth, the UriHelper or any primitive that is context dependent like browser abstractions.

SteveSandersonMS: If the need for some new abstraction arises I'm sure we can.

I hope that for customers, our existing DI APIs cover most requirements. For example, to have an `IAuthorizationService` that obtains data from some other DI service (e.g., an `IHttpContextAccesstor`), the existing DI APIs allow the construction of a scoped service in terms of another scoped/singleton service.

src/Components/Components/src/Auth/AuthorizeView.razor
```diff
@@ -0,0 +1,63 @@
...
+        return result;
+    }
+
+    private bool IsAuthorized()
+    {
+        // TODO: Support various authorization condition parameters, equivalent to those offered
+        // by the [Authorize] attribute, e.g., "Roles" and "Policy". This is on hold until we're
+        // able to reference the policy evaluator APIs from this package.
+
+        return CurrentAuthenticationState.User?.Identity?.IsAuthenticated == true;
```
javiercn: It is a bit weird that this doesn't check for !CurrentAuthenticationState.IsPending although you check on the code above.

src/Components/Components/ref/Microsoft.AspNetCore.Components.netstandard2.0.cs
```diff
@@ -306,6 +333,16 @@ public static partial class HttpClientJsonExtensions
         [System.Diagnostics.DebuggerStepThroughAttribute]
         public static System.Threading.Tasks.Task<T> SendJsonAsync<T>(this System.Net.Http.HttpClient httpClient, System.Net.Http.HttpMethod method, string requestUri, object content) { throw null; }
     }
+    public partial interface IAuthenticationState
+    {
+        bool IsPending { get; }
+        System.Security.Claims.ClaimsPrincipal User { get; }
```
javiercn: Why not represent these two things as `Task<ClaimsPrincipal>`. I get why you are doing this. (In fact our react template does it this way) but in our stack we already respect the notion of Task as a way to represent a "future" value.

If this just becomes a Task, then I can simply await it and react appropriately when it changes.

I think what motivated you to go this way is to provide an idiomatic implementation in AuthorizeView, but I would have used the provider directly on that one (where I want a great degree of control) and used the cascading value in other views (like a menu) where I only care about displaying a current value (like the user name).

In that case, AuthorizeView.razor would be something like
```
if(_currentView != null)
{
    @_currentView
}

Task OnInitAsync()
{
    _currentView = Authorizing;
    var user = await _currentAuthenticationState;
    if(user.IsAuthenticated)
    {
        _currentView = Authorized;
    }
   else
   {
       _currentView = NotAuthorized;
   }
}
```

src/Components/Server/src/Circuits/DefaultCircuitFactory.cs
javiercn: The Fowler will come to haunt you in your dreams if you use IHttpContextAccessor. 
But I don't think DI is enough for this because different platforms have different requeriments, like interactive circuits vs statically rendered components, and DI doesn't enable you to have both registered at the same time. (And also DI only solves the where to get things from, but not the when is the right time to initialize things).

I'm just saying that things like these keep piling up.

src/Components/Components/src/Auth/IAuthenticationStateProvider.cs
javiercn: I think that is implicit in your `IAuthenticationStateProvider.cs` implementation. Your IAuthenticationStateProvider should be responsible for determining with absolute certainty whether you are authenticated or not and hide that from the consumers.

I think the better model is for components to be reactive and to subscribe to the event that IAuthenticationStateProvider give you to get notified of changes. Whatever the implementation is for IAuthenticationStateProvider, that component should be the single source of truth for login and logout operations.

In the case you are using cookies, the cookie expiration determines how long you will be authenticated. (And there's no out of the box way for us to track that other than setting up some sort of timer or checking the cookie expiration on every call (Which we should do anyways as we don't want to provide a stale value when we know its no longer valid))

In the case you are using oAuth/JWTs, the token can be invalidated between when you check and when the actual API call/Rendered view happens, so you need to plan for that case anyways.

What I'm trying to convene here, is that components/services should react to the changes on the auth state and plan for the value being possibly stale rather than proactively checking against a source to see that things are up to date.

src/Components/Components/ref/Microsoft.AspNetCore.Components.netstandard2.0.cs
SteveSandersonMS: > Why not represent these two things as Task<ClaimsPrincipal>
> If this just becomes a Task, then I can simply await it and react appropriately when it changes.

I don't think it would really simplify the developer's code, because the state is more akin to an observable than a task. The value can change any number of times, at any time, rather than being something you just wait for once then get the final result for all time.

So even if we used `Task`, the framework has to notify you through some non-`Task` based API each time it changes again. Given that we're doing that, we might as well supply the information in a synchronous form. This simplifies developers' code (eliminates asynchrony, which in the default server-side case would always be a no-op anyway), and is especially useful if you're using the state info during rendering independently of `AuthorizeView`.

The alternative form you've posted for `AuthorizeView.razor` is essentially the same as the form in the PR, except substituting `if (_currentAuthenticationState.IsPending)` for `await _currentAuthenticationState`.

src/Components/Components/src/Auth/IAuthenticationStateProvider.cs
SteveSandersonMS: > Your IAuthenticationStateProvider should be responsible for determining with absolute certainty whether you are authenticated or not and hide that from the consumers.

I disagree for the client-side case. The client isn't the source of truth (the server is). We can't re-query the server on every call to determine this, and even if we did, the state might already have changed again since. On the client, the `IAuthenticationStateProvider` is tracking what the server has most recently told us.

> I think the better model is for components to be reactive and to subscribe to the event that IAuthenticationStateProvider give you to get notified of changes.

That already does happen.

> What I'm trying to convene here, is that components/services should react to the changes on the auth state and plan for the value being possibly stale rather than proactively checking against a source to see that things are up to date.

Perhaps I was unclear in my description. It's not expected that typical UI components are responsible for proactively querying for updated authentication state. Typical UI components just receive the cascading parameter and react to it.

The only time the developer would be concerned about invalidating auth state is:

 * They are running client-side, or have a custom `IAuthenticationStateProvider`
 * They have taken some action that, due to application-specific rules that only they can know about, may have changed something about the `ClaimsPrincipal`. For example maybe the user just subscribed to "gold tier" service and now may have a new claim value, but only the server really knows for sure. The UI would like to get refreshed info, in case doing so causes some other components to display new output (e.g., different menu entries).

Does this make sense?

src/Components/Components/ref/Microsoft.AspNetCore.Components.netstandard2.0.cs
SteveSandersonMS: I've been thinking about this more, and have reached a view that we *should* make use of `Task`, but a bit differently than discussed above. This will not affect usage of `<AuthorizeView>`; it will only affect consumption of the cascading value from `<AuthenticationStateProvider>`.

Reasoning:

Previously, the goal was to make consumption of the cascading value convenient. However, I think there are basically two usage scenarios:

 1. Using the auth state in **procedural logic**
    * For example, in `OnParametersSetAsync`, fetching some data and including some token in the outbound request, or fetching different data depending on who's logged in. The action should be re-triggered whenever the auth state changes.
    * Or for example, in a button click handler, invoking some business-domain operation that depends on the auth state info.
 2. Using the auth state in **rendering output**
    * E.g., `<div>Hello, @AuthState.User.Identity.Name!</div>`. This needs to update if the user logs in or out.
    * Normally you'd use `<AuthorizeView>` instead for this, but there's nothing to stop you from getting the cascaded value directly and using it

Supplying a `Task` of some kind is fine in case 1, because you can trivially `await`. It's even advantageous, because it forces developers to understand that sometimes the info acquisition is async, and they shouldn't just throw "not authorized" errors instantly if they see the user isn't authenticated, because maybe the info is still pending. Another advantage of `Task` here is that it gives some way for "getting auth state failed" errors to surface (i.e., if the `Task` resolves with an exception).

Supplying a `Task` of some kind is disadvantageous in case 2, because it complicates rendering logic. However it doesn't complicate it very much: we already have excellent support for using `await` during lifecycle methods, and people who do async data access already understand how to show "loading..." states while those tasks are in flight.

Since case 2 is unusual anyway (you'd normally use `<AuthorizeView>` for this), it makes sense to give more weight to case 1, which gives a preference for some `Task` instead of an `IsPending` flag.

The other thing that bothered me about the earlier proposal is that it if `IAuthenticationState` contained a `Task<ClaimsPrincipal>`, this would also complicate things for `<AuthorizeView>`. The `context` is of type `IAuthenticationState`, so you'd be in a real mess if you were trying to access something from the claims principal during rendering. However, this isn't really where we'd want to put the task. It's not only the `ClaimsPrincipal` that can change, but rather the *entire* `IAuthenticationState`, which includes any future properties we attach there (we might put antiforgery tokens there too).

So I think we should drop the `IsPending` flag from `IAuthenticationState` but otherwise leave it alone. And then we modify `<AuthenticationStateProvider>` so that the type of the cascading value changes from `IAuthenticationState` to `Task<IAuthenticationState>`. This means that anyone consuming the cascading value is forced to recognise that, sometimes, the state arrives asynchronously. But it doesn't change anything for people using `<AuthorizeView>`, as this already resolves the succession of tasks into a succession of render calls anyway.

javiercn: I think I agree with what you mention here for the most part. I'm not sure there is an scenario for antiforgery tokens here as:
* For server-side rendering the actions/forms/etc would go through the websocket.
* For client-side rendering it is not recommended to use cookies for auth but JWTs which don't require you to use antiforgery tokens.
* In case you need antiforgery tokens:
  * They could surface as claims on the principal.
  * They could be offered by an antiforgery token service, and that way we avoid polluting a type for an uncommon use case.

For these two additional use cases, (access tokens and antiforgery tokens) it would be interesting to see if we could make it trivial for HttpClient to retrieve the required pieces of data and to attach them to outgoing requests automatically for you.

SteveSandersonMS: > They could surface as claims on the principal

That was my initial design too, but Barry said they should be separate. However you are probably right that we might never even need to expose them in this kind of way.

src/Components/Components/ref/Microsoft.AspNetCore.Components.netstandard2.0.cs
```diff
@@ -3,6 +3,33 @@
 
 namespace Microsoft.AspNetCore.Components
 {
+    public delegate void AuthenticationStateChangedHandler(System.Threading.Tasks.Task<Microsoft.AspNetCore.Components.IAuthenticationState> newAuthenticationStateTask);
+    public partial class AuthenticationStateProvider : Microsoft.AspNetCore.Components.ComponentBase, System.IDisposable
```
rynowak: Should this be sealed? Is there anything reasonable to do with it by subclassing?

src/Components/Components/src/Auth/AuthenticationStateProvider.razor
```diff
@@ -0,0 +1,37 @@
+@namespace Microsoft.AspNetCore.Components
+@implements IDisposable
+
+<CascadingValue T="Task<IAuthenticationState>" Value="@_currentAuthStateTask" ChildContent="@ChildContent" />
+
+@functions {
+    private Task<IAuthenticationState> _currentAuthStateTask;
+
+    [Inject] private IAuthenticationStateProvider AuthStateProvider { get; set; }
```
rynowak: Why does this need to be written out longhand instead of using `@inject`? Is it to control the visibility?

src/Components/Components/src/Auth/AuthorizeView.razor
```diff
@@ -0,0 +1,68 @@
...
+}
+else
+{
+    @NotAuthorized
+}
+
+@functions {
+    private IAuthenticationState currentAuthState;
+
+    [CascadingParameter] public Task<IAuthenticationState> AuthStateTask { get; private set; }
```
rynowak: We should write out `Authentication` in all of this places. I'd also prefer if we can drop `Task` from the name.

src/Components/Components/src/Auth/AuthorizeView.razor
```diff
@@ -0,0 +1,68 @@
...
+    @Authorizing
+}
+else if (IsAuthorized())
+{
+    @((Authorized ?? ChildContent)?.Invoke(currentAuthState))
+}
+else
+{
+    @NotAuthorized
+}
```
rynowak: This is dope 👍

src/Components/Components/src/Auth/AuthorizeView.razor
rynowak: Actually is there a reason why this is public?

src/Components/Components/src/Auth/AuthorizeView.razor
```diff
@@ -0,0 +1,68 @@
...
+    /// The content that will be displayed while asynchronous authorization is in progress.
+    /// </summary>
+    [Parameter] public RenderFragment Authorizing { get; private set; }
+
+    protected override async Task OnParametersSetAsync()
+    {
+        // We allow 'ChildContent' for convenience in basic cases, and 'Authorized' for symmetry
+        // with 'NotAuthorized' in other cases. Besides naming, they are equivalent. To avoid
+        // confusion, explicitly prevent the case where both are supplied.
+        if (ChildContent != null && Authorized != null)
```
rynowak: If we want, we can make this a language feature. It would be pretty easy to do if we do it before 3.0.

src/Components/Components/src/Auth/IAuthenticationStateProvider.cs
```diff
@@ -0,0 +1,33 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Components
+{
+    /// <summary>
+    /// Provides information about the authentication state of the current user.
+    /// </summary>
+    public interface IAuthenticationStateProvider
```
rynowak: This is me showing up (as expected) to ask if we're sure why an interface is better than an abstract class for this. I could see us adding more methods/overloads to this. I always get suspicious about maintainablility when I see a bool method parameter.

So with that out of the way. I'll do what I'm here to do.

Are we sure this wouldn't be better off as an abstract class? It feels more like a class to me.

src/Components/Components/test/Auth/AuthenticationStateProviderTest.cs
```diff
@@ -0,0 +1,234 @@
...
+            var receiveAuthStateId = batch.GetComponentFrames<ReceiveAuthStateComponent>().Single().ComponentId;
+            var receiveAuthStateDiff = batch.DiffsByComponentId[receiveAuthStateId].Single();
+            Assert.Collection(receiveAuthStateDiff.Edits, edit =>
+            {
+                Assert.Equal(RenderTreeEditType.PrependFrame, edit.Type);
+                AssertFrame.Text(
+                    batch.ReferenceFrames[edit.ReferenceFrameIndex],
+                    "Authenticated: True; Name: Bert; Pending: False; Renders: 1");
+            });
+        }
```
rynowak: As we build more of these type of components, would it make sense for us to build more of a framework for testing these kinds of details. 


Nice to have for sure, but there's a maintenance cost of writing tests that are so coupled to the low level implementation. I'm worried about things like trying to make changes to the render tree or compiler and getting nuked from orbit by our own tests.

src/Components/Server/src/Circuits/DefaultCircuitFactory.cs
rynowak: Personally I think the current state is the **best possible** solution. 

It's not overdesigned. It's very clear to see what happens and in what order. If the ordering changes, or the dependencies change, then that will be reflected in a clear diff.

I don't think we should turn this into a filter pipeline, or property bag, or anything like that. To me simple is good.

src/Components/Components/ref/Microsoft.AspNetCore.Components.netstandard2.0.cs
```diff
@@ -306,6 +333,15 @@ public static partial class HttpClientJsonExtensions
         [System.Diagnostics.DebuggerStepThroughAttribute]
         public static System.Threading.Tasks.Task<T> SendJsonAsync<T>(this System.Net.Http.HttpClient httpClient, System.Net.Http.HttpMethod method, string requestUri, object content) { throw null; }
     }
+    public partial interface IAuthenticationState
```
rynowak: Can we turn this into a sealed class? This just seems like homework for someone to have to implement, it's just data and we want to keep it that.

The provider already has to be stateful, so there's really no reason why you'd want to attach extra state to this type.

src/Components/Components/src/Auth/AuthorizeView.razor
```diff
@@ -0,0 +1,68 @@
...
+    /// If you specify a value for this parameter, do not also specify a value for <see cref="ChildContent"/>.
+    /// </summary>
+    [Parameter] public RenderFragment<IAuthenticationState> Authorized { get; private set; }
+
+    /// <summary>
+    /// The content that will be displayed while asynchronous authorization is in progress.
+    /// </summary>
+    [Parameter] public RenderFragment Authorizing { get; private set; }
+
+    protected override async Task OnParametersSetAsync()
```
rynowak: This will eventually have to accept an auth policy name or auth policy. I assume you're waiting for that relayering to blow over first :)

rynowak (commented):
Looks good. Looks like what I expected to see 👍

src/Components/Components/ref/Microsoft.AspNetCore.Components.netstandard2.0.cs
SteveSandersonMS: I didn't think we had a way to mark `.razor`-generated classes as `sealed`. Please correct me if I'm wrong!

src/Components/Components/ref/Microsoft.AspNetCore.Components.netstandard2.0.cs
SteveSandersonMS: I'm fine with turning it into a class, but marking it as `sealed` seems pretty harsh. You might want to attach extra state, for example to hold authentication tokens that were returned by the server when this is used on the client.

I know there are other theoretically possible ways to attach extra state (`WeakDictionary`, or embedding them in the claims) but I would think we can make it easier by subclassing.

src/Components/Components/src/Auth/AuthenticationStateProvider.razor
SteveSandersonMS: Good point - fixed. The visibility remains the same with `@inject`.

src/Components/Components/src/Auth/AuthorizeView.razor
SteveSandersonMS: Renamed.

Good point about visibility. It was just following the pattern of other params here, but you're right - the tooling limitation doesn't apply to CascadingParameter. Fixed.

src/Components/Components/src/Auth/AuthorizeView.razor
SteveSandersonMS: I'm not sure this will be very common, and we could still add it without breaking any existing apps later if it was something components explicitly opted into. Example:

```cs
[Parameter, ParameterAlias("Authorized")] public RenderFragment<IAuthenticationState> ChildContent { get; private set; }
```

So while I think it's a cool idea, I'd prioritise other things.

src/Components/Components/src/Auth/AuthorizeView.razor
SteveSandersonMS: Yes, exactly. This is also mentioned in the `IsAuthorized` logic below.

src/Components/Components/src/Auth/IAuthenticationStateProvider.cs
SteveSandersonMS: To help me track remaining todos, I'll mark this as resolved since it seems like the conversation concluded.

src/Components/Components/src/Auth/IAuthenticationStateProvider.cs
```diff
@@ -0,0 +1,33 @@
...
+        /// has changed. For example, this event may be raised if a user logs in or out.
+        /// </summary>
+        event AuthenticationStateChangedHandler AuthenticationStateChanged;
+    }
+
+    /// <summary>
+    /// A handler for the <see cref="IAuthenticationStateProvider.AuthenticationStateChanged"/> event.
+    /// </summary>
+    /// <param name="newAuthenticationStateTask">A <see cref="Task"/> that supplies the updated <see cref="IAuthenticationState"/>.</param>
+    public delegate void AuthenticationStateChangedHandler(Task<IAuthenticationState> newAuthenticationStateTask);
```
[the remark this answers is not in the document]
SteveSandersonMS: Done

src/Components/Components/test/Auth/AuthenticationStateProviderTest.cs
SteveSandersonMS: 100% yes. This has been one of the "big rocks" since the beginning: https://github.com/aspnet/AspNetCore/issues/5458

Each time I write these kinds of unit tests, I'm doing so with a view to learn what patterns turn out to be useful. My hope is that eventually, we extract a component unit testing library from our real-world experience of testing our own components.

It's pretty cool to have the chance to design the unit testing system through this approach instead of just abstractly in our heads for a change.

> tests that are so coupled to the low level implementation

I share that concern somewhat. But I'm also really glad that they are coupled to details like "exactly how many renders occurred, and in what order" because it forces us to be aware of certain perf characteristics and would blow up if we regressed it. For "regression", I include things like "taking stuff that was synchronous and making it asynchronous".

SteveSandersonMS: BTW in case you were wondering, I'd imagine the unit testing mechanism we eventually ship would make your tests look vastly simpler than these, because it would include something like CSS-style selectors for traversing the rendered document (represented as a whole, not separate trees for each component). You wouldn't be coding in terms of render tree frames.

In most cases you'd probably write assertions in terms of the final document shape, not in terms of diffs like we do here. But it is releveant to observe individual diffs if you're being really precise.

src/Components/Server/src/Circuits/DefaultCircuitFactory.cs
SteveSandersonMS: Marking resolved because I think we're concluded that we're not changing this in this PR.

src/Components/Components/src/Auth/AuthorizeView.razor
SteveSandersonMS: The logic here is very temporary. It will be replaced entirely when we add specific auth policies, roles, etc.

src/Components/Components/src/Auth/IAuthenticationStateProvider.cs
SteveSandersonMS: I was wondering whether default interface methods would be the future for avoiding this regular decision point. However since we're not in a position to use them on Mono WebAssembly (`netstandard2.0`) currently, and we can't be sure how soon they will become available, it's probably not the solution for now.

As such, yes, I'll convert this to an ABC.

I was thinking of doing this before because I wanted to add some convenience properties, such as a top-level `bool IsAuthenticated` and `string Name` that got their data from the default identity. It's much more obvious and pleasant to write `AuthState.IsAuthenticated` than `AuthState.User.Identity.IsAuthenticated`.

I won't add the convenience properties just now because I'd like this feature and its usage patterns to bake a bit more first. It's something we can consider later by the time we're making templates and samples.

src/Components/Components/ref/Microsoft.AspNetCore.Components.netstandard2.0.cs
SteveSandersonMS: I was thinking of making it a class before, because I wanted to add some convenience properties, such as a top-level `bool IsAuthenticated` and `string Name` that got their data from the default identity. It's much more obvious and pleasant to write `AuthState.IsAuthenticated` than `AuthState.User.Identity.IsAuthenticated`.

I won't add the convenience properties just now because I'd like this feature and its usage patterns to bake a bit more first. It's something we can consider later by the time we're making templates and samples.

src/Components/Components/src/Auth/IAuthenticationStateProvider.cs
SteveSandersonMS: OK, done. One consequence of this is that we had a name clash: we had both this service and an associated component that want to be called `AuthenticationStateProvider`. It's awkward because that is the best name for both of them.

I've resolved this by renaming the component to `<CascadingAuthenticationState>`. In a way that is a bit more explicit.

However, I don't really like the fact that `<CascadingAuthenticationState>` is going to show up in people's code at all. It just feels like boilerplate in the template. I'm considering having some way of including it by default even without it being in app code, since if you're not using auth, it doesn't do anything anyway. Two ways this could be done:

 * We change the renderer to put an implicit `<CascadingAuthenticationState>` at the root of the app tree. Similarly, for any future things like this, they all get wrapped around each other and the app tree.
 * **(Preferred)** We don't interfere with the render tree (as it has perf and debugging implications), we just amend the cascading parameter resolution logic so that there are some default framework-provided values at the root. It doesn't stop you from overriding them. We would have an `IAuthenticationState` at the root (along with any future built-in intrinsics), for which the framework would obtain the value from whatever `AuthenticationStateProvider` is registered in DI.

I spoke with @javiercn about this and our conclusion was that, for the short term, we do nothing. We don't want to introduce new concepts until we know there are multiple usage scenarios. So all the templates will have a `<CascadingAuthenticationState>` around the router in `App.razor`.

@rynowak I'd be interested in whether you have views on this or can anticipate other places where we would benefit from a concept like this.

javiercn: > default interface methods

This is not a good solution for this problem I think. Default Interface Methods only allow you to add methods, but you still can not add state, so for things that we want to be open for future extension, ABCs are the superior choice as you aren't limited to what things you can add.

src/Components/Components/ref/Microsoft.AspNetCore.Components.netstandard2.0.cs
SteveSandersonMS: Changed it to be a class.

@rynowak Let me know if you disagree about leaving it unsealed.

src/Components/Components/src/Auth/AuthenticationStateProvider.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// </summary>
+    public abstract class AuthenticationStateProvider
+    {
+        /// <summary>
+        /// Gets an <see cref="AuthenticationState"/> instance that describes
+        /// the current user.
+        /// </summary>
+        /// <param name="forceRefresh">If true, instructs the provider to re-determine the user's authentication state, which can be an expensive operation. If false, the provider may reuse data cached for the current user.</param>
+        /// <returns>An <see cref="AuthenticationState"/> instance that describes the current user.</returns>
+        public abstract Task<AuthenticationState> GetAuthenticationStateAsync(bool forceRefresh);
```
SteveSandersonMS: We might now choose not to have the `forceRefresh` param on the base class method, but rather make this part of the semantics of specific subclasses.

For example, the client-side `ServerAuthenticationStateProvider`, which works by making an HTTP to the server, could have its own `RefreshState` method that re-issues the request to the server and triggers `AuthenticationStateChanged` using the result.

SteveSandersonMS: Since I can't think of any reason why the *framework* needs to request a refresh in a polymorphic way, this no longer needs to be part of the base class. Developers can still trigger a refresh on their own subclasses, via their own APIs, if they want.

So this is now done.

src/Components/Components/src/Auth/IAuthenticationStateProvider.cs
rynowak: I would suggest doing option 1, but making it based on DI. So you can register things that you want to be injected as *cascading value providers*.

SteveSandersonMS: These options are open to us in the future, but I'd like to keep it out of scope for this PR. It would be a whole new feature which we can't really determine the design for at this stage.

rynowak: Agreed. I don't expect any action right now.

src/Components/Components/ref/Microsoft.AspNetCore.Components.netstandard2.0.cs
rynowak: Yeah, I think you're right 😢 we don't have that.

src/Components/Components/src/Auth/AuthorizeView.razor
rynowak: We could add the feature in the future, we wouldn't be able to use it here without a binary breaking change.

rynowak: Ultimately I'm fine with ignoring this for now.

src/Components/Components/test/Auth/AuthenticationStateProviderTest.cs
rynowak: Agreed on all of these points.

src/Components/Components/ref/Microsoft.AspNetCore.Components.netstandard2.0.cs
rynowak: I don't feel super strongly about being unsealed. 

> You might want to attach extra state, for example to hold authentication tokens that were returned by the server when this is used on the client.

I think that people are going to ask for this one way or another, but the current design really neither helps nor harms it. I think it's fine to address this in a future release.

src/Components/Components/ref/Microsoft.AspNetCore.Components.netstandard2.0.cs
```diff
@@ -3,6 +3,34 @@
...
+    }
+    public delegate void AuthenticationStateChangedHandler(System.Threading.Tasks.Task<Microsoft.AspNetCore.Components.AuthenticationState> task);
+    public abstract partial class AuthenticationStateProvider
+    {
+        protected AuthenticationStateProvider() { }
+        public event Microsoft.AspNetCore.Components.AuthenticationStateChangedHandler AuthenticationStateChanged { add { } remove { } }
+        public abstract System.Threading.Tasks.Task<Microsoft.AspNetCore.Components.AuthenticationState> GetAuthenticationStateAsync();
+        protected void NotifyAuthenticationStateChanged(System.Threading.Tasks.Task<Microsoft.AspNetCore.Components.AuthenticationState> task) { }
+    }
+    public partial class AuthorizeView : Microsoft.AspNetCore.Components.ComponentBase
```
rynowak: There's a high level question that came up for me. I'm going to mention it, and I see it as more of an API review thing rather than something to address right now.

We should develop the vocabulary we want to use for components that we provide. Like this guy is the `AuthorizeView`, it could also be the `AuthorizeDisplay` or something else. 

I indexed on the work `View` because that already means something in MVC. Anyways, just for discussion 👍

src/Components/Components/src/Auth/AuthenticationState.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    /// <summary>
+    /// Provides information about the currently authenticated user, if any.
+    /// </summary>
+    public class AuthenticationState
+    {
+        /// <summary>
+        /// Gets a <see cref="ClaimsPrincipal"/> that describes the current user.
+        /// </summary>
+        public ClaimsPrincipal User { get; }
```
rynowak: nit: fields -> constructors -> properties

src/Components/Components/src/Auth/AuthenticationStateProvider.cs
```diff
@@ -0,0 +1,49 @@
...
+    /// <summary>
+    /// Provides information about the authentication state of the current user.
+    /// </summary>
+    public abstract class AuthenticationStateProvider
+    {
+        /// <summary>
+        /// Gets an <see cref="AuthenticationState"/> instance that describes
+        /// the current user.
+        /// </summary>
+        /// <returns>An <see cref="AuthenticationState"/> instance that describes the current user.</returns>
```
rynowak: technically not correct, it returns "a task that blah blah blah"

src/Components/Components/src/Auth/AuthenticationStateProvider.cs
```diff
@@ -0,0 +1,49 @@
...
+        public abstract Task<AuthenticationState> GetAuthenticationStateAsync();
+
+        /// <summary>
+        /// An event that provides notification when the <see cref="AuthenticationState"/>
+        /// has changed. For example, this event may be raised if a user logs in or out.
+        /// </summary>
+#pragma warning disable 0067 // "Never used" (it's only raised by subclasses)
+        public event AuthenticationStateChangedHandler AuthenticationStateChanged;
+#pragma warning restore 0067
```
rynowak: Is this actually needed now that `NotifyAuthenticationStateChanged` is there?

src/Components/Components/src/Auth/AuthenticationStateProvider.cs
```diff
@@ -0,0 +1,49 @@
...
+        protected void NotifyAuthenticationStateChanged(Task<AuthenticationState> task)
+        {
+            if (task == null)
+            {
+                throw new ArgumentNullException(nameof(task));
+            }
+
+            AuthenticationStateChanged?.Invoke(task);
+        }
+    }
```
rynowak: Food for thought. It appears that we want to (and support) raising this event on any thread. Do we want to document that as supported? Or do we want to try and make it enforced that you raise this on the sync context.

rynowak (approved):

src/Components/Components/src/Auth/AuthenticationStateProvider.cs
javiercn: I don’t think we want that. The callback should be responsible for calling Invoke or InvokeAsync when handling the event.

I’m wondering how good our support story is when a component calls Invoke

src/Components/Components/ref/Microsoft.AspNetCore.Components.netstandard2.0.cs
javiercn: I think that’s a hood idea, although I don’t necessarily subscribe your specific suggestion.

SteveSandersonMS: Likewise I agree. TBH I think the name `AuthorizeDisplay` would be fine. However we can leave this as-is until the API review.

src/Components/Components/src/Auth/AuthenticationState.cs
SteveSandersonMS: OK. Rather than go through another multi-hour CI process to make this tweak, I'm going to include it in my next auth PR.

src/Components/Components/src/Auth/AuthenticationStateProvider.cs
SteveSandersonMS: OK. Rather than go through another multi-hour CI process to make this tweak, I'm going to include it in my next auth PR.

src/Components/Components/src/Auth/AuthenticationStateProvider.cs
SteveSandersonMS: Well spotted. Rather than go through another multi-hour CI process to make this tweak, I'm going to include it in my next auth PR.

src/Components/Components/ref/Microsoft.AspNetCore.Components.netstandard2.0.cs
```diff
@@ -3,6 +3,34 @@
...
+    {
+        public AuthorizeView() { }
+        [Microsoft.AspNetCore.Components.ParameterAttribute]
```
rynowak: @SteveSandersonMS - something else I thought of late (sorry). 

These need to go into the manually maintained section of the ref assembly. The reason why is that ref assemblies don't preserve the setter if it's non-public. You can use https://github.com/aspnet/AspNetCore/commit/cc1b2941f90fda48b38212b3d402c2869f410756 as an example.

SteveSandersonMS: Great reminder - thanks. This should be covered in today's PR in this commit: https://github.com/aspnet/AspNetCore/pull/10293/commits/ad9ac558beae16ea708ed36adc5e416ee8ae73ed

PR #10236, opened 2019-05-14
Make RouteAttribute non-inherited

### LINKED ISSUE #5529: Router does not find inherited RouteAttribute
Since component is not a partial class the issue is that when i specify route (e.g. `/interop`) in base class and i won't use `@page` in the view but use `@inherits` - i've got issue in console in the browser: 
```
MonoPlatform.ts:70
 Uncaught (in promise) Error: System.InvalidOperationException: Failed to compare two elements in the array. ---> System.InvalidOperationException: The following routes are ambiguous:
'interop' in 'blazorfirststandaloneapplication.Base.InteropBase'
'interop' in 'blazor_first_standalone_application.Pages.Interop'
```

If it's inheritance - in my opinion it should derive Route attribute.

rynowak:
Fixes: #5529

Inheriting and looking for inherited route attributes will cause nothing
but trouble. We had a bug tracking what to do about this and we decided
to make it really clear that routes are not inherited.

Previously the attribute was marked as inherited, but we woulnd't look
for inherited routes.

### DIFF
--- a/src/Components/Components/src/RouteAttribute.cs
+++ b/src/Components/Components/src/RouteAttribute.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
@@ -8,7 +8,7 @@ namespace Microsoft.AspNetCore.Components
     /// <summary>
     /// Indicates that the associated component should match the specified route template pattern.
     /// </summary>
-    [AttributeUsage(AttributeTargets.Class, AllowMultiple = true, Inherited = true)]
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = true, Inherited = false)]
     public class RouteAttribute : Attribute
     {
         /// <summary>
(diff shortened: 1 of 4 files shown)


SteveSandersonMS (approved):
Would be great to have an extra unit test in `RouteTableTests` to represent this.

Besides that, looks great!

rynowak:
Added some tests

src/Components/Components/src/RouteAttribute.cs
```diff
@@ -8,7 +8,7 @@ namespace Microsoft.AspNetCore.Components
     /// <summary>
     /// Indicates that the associated component should match the specified route template pattern.
     /// </summary>
-    [AttributeUsage(AttributeTargets.Class, AllowMultiple = true, Inherited = true)]
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = true, Inherited = false)]
     public class RouteAttribute : Attribute
```
pranavkm: Make it sealed? I think you're allowed to redefine `AttributeUsage` on derived types so you could get in to the same problem.

rynowak: Hmm that's wierd that we didn't do this already. Since github didn't show your comment until after I merged, I'm going to add this to the API review item.

Protagores:
@rynowak One possible legitimate scenario with inherited route _templates_ I could see, is when the API scheme is using API version in the URL path, and one would like to avoid duplicating the same `[Route("api/v{version:apiVersion}/[controller]")]` attribute in every versioned API controller in order to uphold the DRY principle.

Right now I’m having to resort to a `IEndpointRouteBuilder.MapControllerRoute(...)` call in my `Startup` class to configure something like that, which just isn’t that flexible when my controllers would have vastly different sets of action endpoints.

rynowak:
This issue was related to Blazor. What you're suggesting with work fine with MVC.

PR #10241, opened 2019-05-14
Publish the shared framework MSI to artifacts

joeloff:
Publish the shared framework MSI to artifacts. This will enable the SDK to consume it directly for Visual Studio changes we're making.

### DIFF
--- a/src/Installers/Windows/SharedFramework/SharedFramework.wixproj
+++ b/src/Installers/Windows/SharedFramework/SharedFramework.wixproj
@@ -15,6 +15,7 @@
     <HarvestDirectorySuppressSpecificWarnings>5150;5151</HarvestDirectorySuppressSpecificWarnings>
     <HarvestDirectorySuppressRegistry>true</HarvestDirectorySuppressRegistry>
     <HarvestSource>$(IntermediateOutputPath)fx\</HarvestSource>
+    <IsProductInstaller>true</IsProductInstaller>
     <DefineConstants>$(DefineConstants);AspNetCoreSharedFrameworkSource=$(HarvestSource)</DefineConstants>
     <NamespaceGuid>$(SharedFrameworkNamespaceGuid)</NamespaceGuid>
     <SchemaVersion>2.0</SchemaVersion>


joeloff:
@johnbeisner FYI, I think once this is merged it should unblock you on the VS side for the work you're doing right now

natemcmaster:
Does this mean we can stop producing wixlib files in favor of .msi?

joeloff:
@natemcmaster not yet, but eventually. We're still including the MSI in 3 bundles, so the wixlib guarantees consistency. This is just for VS where we want to directly consume items to simplify some servicing scenarios

natemcmaster:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=188739

natemcmaster (approved):

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/2327

dougbu:
@joeloff the official build has been failing since this change went in. Specifically, the SharedFramework project now places `sfx_x86.cab` and `xfx_x64.cab` in the `artifacts/bin/SharedFramework/x64/Release/en-US/` folder but doesn't copy them into `artifacts/installers/Release/`. Nonetheless the `Microsoft.DotNet.SignCheck` tool is looking for the .cab files in the installers folder for some reason:
```
Could not find file 'F:\workspace\_work\1\s\artifacts\installers\Release\sfx_x64.cab'.
Could not find file 'F:\workspace\_work\1\s\artifacts\installers\Release\sfx_x86.cab'.
```
There's no other mention of these .cab files in the `SignCheck` build or its binary log.

What needs to change to fix this?

dougbu:
@aspnet/build I can't find an issue about these failures in aspnet/AspNetCore-Internal though the bot has been running reliably. What gives?

joeloff:
Well, SignCheck would complain because it's now picking up the MSI, but the CABs aren't being published to artifacts.

Looking at the targets file, it's only copying the output, which for a wixproj is the MSI typically, not the CABs.

dougbu:
@joeloff :ok: but what's the correct fix?

joeloff:
@dougbu  Copy the CAB files. I sent a PR for that and added you to it. Ideally I'd like to see what the agent's raw build process looks like on disk. Downside is we'd scan the CABs twice or maybe more (because we'd be scanning the EXE (which includes the MSI and the CABs) and then scan the individual MSI+CABs as well

dougbu:
@joeloff I doubt the recan #10340 adds will make a significant difference. And, fixing things here will get the builds running again much faster than a `Microsoft.DotNet.SignCheck` change for this particular case.

PR #10247, opened 2019-05-14
Accessibility

ryanbrandenburg:
All of these except for the `<html lang="en"` ones are to provide higher contrast. The fact that Bootstrap classes don't have sufficient contrast by default with a white background is a known thing that they don't seem to have plans to fix.

Something I learned as a result of this is that IdentityUI has it's own layout and styling which don't seem to integrate with the SPA frameworks (because they're Razor). The result is that changes to SPA templates are not reflected in IdentityUI pages at all. For example if you were to make the background of your SPA site black instead of white then go to the login page, the background of the login page would still be  white. That seems less than ideal.


Eilon (approved):
Looks alright to me, but I don't know this code very well. One of @javiercn or @HaoK should review as well.

javiercn:
Can you post images with the visual changes?

javiercn:
There isn’t anything we can do about the SPA/Identity layout impedance as they are built with completely different technologies.

ryanbrandenburg:
countercontrast
contrast

Here are some examples from the Angular template. Basically the offending elements were just made a couple shades darker.

src/ProjectTemplates/Web.Spa.ProjectTemplates/content/React-CSharp/ClientApp/src/App.js
```diff
@@ -10,6 +10,8 @@ import ApiAuthorizationRoutes from './components/api-authorization/ApiAuthorizat
 import { ApplicationPaths } from './components/api-authorization/ApiAuthorizationConstants';
 ////#endif
 
+import './custom.css'
```
javiercn: Don’t name this custom. Look up the common react convention for this and if not either name it site.css or app.css to match the file.

javiercn: Wait, seems like custom was what we were doing. Why do we change it from scss to css (as I expect scss to be processed at build)

javiercn (approved):
Looks good, I want to better understand the move from scss to css.

src/ProjectTemplates/Web.Spa.ProjectTemplates/content/React-CSharp/ClientApp/src/App.js
ryanbrandenburg: The scss file that existed before was both empty and unused. If someone had tried to import it as I do here they would have gotten an error like `You attempted to import ../scss/custom.scss which falls outside of the project src/ directory. Relative imports outside of src/ are not supported.`. If we try just moving the scss file within the src directory we instead get `To import Sass files, you first need to install node-sass.
Run `npm install node-sass` or `yarn add node-sass` inside your workspace.`. I figured that it wasn't worth it to add an additional OSS dependency when the styles we're applying makes no use of the additional features from scss, so I made it a css file.

ryanbrandenburg:
@aspnet-hell triage https://dev.azure.com/dnceng/public/_build/results?buildId=192570

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/2535

ryanbrandenburg:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=192575

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/2432
https://github.com/aspnet/AspNetCore-Internal/issues/2445

Issue #9658, opened 2019-04-22
Handle the null value scenario in JsonDocumentAuthExtensions.GetString


Tratcher:
When moving Auth over to JsonDocument we added a GetString extension to make life easier.
https://github.com/aspnet/AspNetCore/pull/7105/files#diff-6628f97b66ced342630a3798d4e68eb9R8

We missed a scenario where `{ "foo" : null } ` returns string.Empty rather than null. This is likely what caused the MSA regression https://github.com/aspnet/AspNetCore/issues/9083.

Check if the value type is null before calling ToString. See if we can remove the MSA mitigation as well.

Tratcher:
Verified with SDK 3.0.100-preview6-012234

PR #10252, opened 2019-05-15
Handle `null`s in the `JsonDocumentAuthExtensions.GetString` extension method

mderriey:
- Returns `null` instead of empty string if the JSON property has a `null` value

Addresses #9658

### DIFF
--- a/src/Security/Authentication/Core/src/JsonDocumentAuthExtensions.cs
+++ b/src/Security/Authentication/Core/src/JsonDocumentAuthExtensions.cs
@@ -7,8 +7,14 @@ namespace Microsoft.AspNetCore.Authentication
 {
     public static class JsonDocumentAuthExtensions
     {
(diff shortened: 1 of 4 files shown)


mderriey:
I also had a look at reverting the fix you put in for the MSA regression in [`0d45fe73`](https://github.com/aspnet/AspNetCore/commit/0d45fe73d75ca82b6968e2548f70bc59fe1ce9a1).

I think it'll work most of the time unless we get an `email` claim which value is an empty string, in which case we'll return that one. The probability is low, but I thought it's a case we _might_ need to consider.

If you think we can ignore that edge case, then I believe the commit can be reverted safely; happy to do that on that PR if you give me the 👍.

Tratcher:
Yes, please revert the MSA changes. Thanks.

Tratcher:
Thanks

PR #10258, opened 2019-05-15
Add components solution filter file

SteveSandersonMS:
I'm still figuring out whether there are any drawbacks or problems with this, but so far it seems to be working great. Opening this PR now in case others have insight into the practicality of it.

Previously, if dependency projects were not loaded in the solution, the build would fail in VS. However with the current version of VS (16.1.0 Preview 3), it now appears to handle this scenario just fine, assuming you've already done at least one command-line build.

This PR adds a [solution filter file](https://docs.microsoft.com/en-us/visualstudio/ide/filtered-solutions?view=vs-2019) `ComponentsNoDeps.slnf` you can optionally use instead of `Components.sln`. This makes VS only load the 28 projects under `src/Components` and *not* the whole set of 94 projects if transitive project references are included. For me, this means:

 * VS open in a couple of seconds compared with about 30 seconds before
 * A cold (first-time) full-solution build is now about 15-20 seconds (mostly spent in the Mono linker) versus about 1-2 minutes before.
 * A warm (not much changed) full-solution build is now about 10 seconds versus about 30 seconds before
 * Test explorer and debugging still work and are much quicker starting/updating

Am I really behind the curve and everyone else was already doing this? Or has this only just become possible?

### DIFF
--- a/src/Components/ComponentsNoDeps.slnf
+++ b/src/Components/ComponentsNoDeps.slnf
@@ -1,0 +1,35 @@
+{
+  "solution": {
+    "path": "Components.sln",
+    "projects": [
+      "Analyzers\\src\\Microsoft.AspNetCore.Components.Analyzers.csproj",
+      "Analyzers\\test\\Microsoft.AspNetCore.Components.Analyzers.Tests.csproj",
+      "Blazor\\Blazor\\src\\Microsoft.AspNetCore.Blazor.csproj",
+      "Blazor\\Blazor\\test\\Microsoft.AspNetCore.Blazor.Tests.csproj",
+      "Blazor\\Build\\src\\Microsoft.AspNetCore.Blazor.Build.csproj",
+      "Blazor\\Build\\test\\Microsoft.AspNetCore.Blazor.Build.Tests.csproj",
+      "Blazor\\DevServer\\src\\Microsoft.AspNetCore.Blazor.DevServer.csproj",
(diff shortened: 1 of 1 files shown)


javiercn:
Solution filters are a new VS2019 feature. It was already possible to do this if you built from the command-line before.

src/Components/ComponentsNoDeps.slnf
```diff
@@ -0,0 +1,35 @@
+{
+  "solution": {
+    "path": "Components.sln",
+    "projects": [
```
javiercn: What is this json madness. Give me back my unintelligible format full of GUIDs

javiercn (approved):

src/Components/ComponentsNoDeps.slnf
SteveSandersonMS: I know. Who will think of our job security? With a file format like this, *anyone* could just come along and understand it 😱

natemcmaster (approved):

natemcmaster:
I totally support adding slnf to the repo in whatever what each team thinks is useful. It's too bad slnf is a UI only feature...there is no way to use it in a command line build afaik. But if this helps with VS load time, go for it!

rynowak:
>  assuming you've already done at least one command-line build.

I think if there is a drawback, it's this. We don't explicitly have this as a requirement to use our repo.

PR #10271, opened 2019-05-15
Use attributes for application part discovery

pranavkm:


src/Mvc/Mvc.Core/src/ApplicationParts/ApplicationPartAttribute.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Mvc.ApplicationParts
+{
+    [AttributeUsage(AttributeTargets.Assembly, AllowMultiple = true)]
+    public sealed class ApplicationPartAttribute : Attribute
```
pranavkm: Naming suggestions.

mkArtakMSFT: What's the concern with current one (ApplicationPartAttribute)?

src/Mvc/Mvc.Core/src/ApplicationParts/ApplicationPartManager.cs
```diff
@@ -52,18 +52,57 @@ public void PopulateFeature<TFeature>(TFeature feature)
...
         }
+
+        private static IEnumerable<Assembly> GetApplicationPartAssemblies(string entryAssemblyName)
+        {
+            var entryAssembly = Assembly.Load(new AssemblyName(entryAssemblyName));
+
+            // Use ApplicationPartAttribute to get the closure of direct or transitive dependencies
+            // that reference MVC.
+            var assembliesFromAttributes = entryAssembly.GetCustomAttributes<ApplicationPartAttribute>()
+                .Select(name => Assembly.Load(name.AssemblyName))
```
mkArtakMSFT: So this will result in the assemblies being actually loaded? Is that intended to happen at this phase?
If not, consider using `Assembly.ReflectionOnlyLoad` instead.
If, however this is the intention, the method name should probably be somewhat cleaner - to reflect on that.

src/Mvc/Mvc.Core/src/ApplicationParts/ApplicationPartAttribute.cs
pranavkm: Nothing in particular. People generally tend to have strong opinions about names

src/Mvc/Mvc.Core/src/ApplicationParts/ApplicationPartManager.cs
pranavkm: >  Is that intended to happen at this phase?

Yup, we've already decided that these assemblies are of interest to MVC and will be turned in to `ApplicationPart` instances.

src/Mvc/Mvc.Core/src/ApplicationParts/ApplicationPartAttribute.cs
rynowak: Is the idea of this new attribute that we want to do something other than what our existing attribute does? I think it's called `RelatedPartAttribute`?

src/Mvc/Mvc.Core/src/ApplicationParts/ApplicationPartManager.cs
```diff
@@ -52,18 +52,57 @@ public void PopulateFeature<TFeature>(TFeature feature)
...
-            foreach (var assembly in applicationAssemblies)
+            foreach (var assembly in assemblies)
             {
+                if (!seenAssemblies.Add(assembly))
+                {
+                    // Assembly discovery may be duplicated (e.g. if a user explicitly specified an ApplicationPartAAttribute
+                    // that was also codegened), but we do not want duplicate ApplicationParts.
+                    // Avoid duplicates. Note that we prefer doing this over Distinct since the latter isn't
+                    // guaranteed to preserve the original ordering which is important here.
+                    continue;
```
rynowak: this comment is kinda redundant with itself 😆 I think it's seen some surgery.

src/Mvc/Mvc.Core/src/ApplicationParts/ApplicationPartManager.cs
```diff
@@ -52,18 +52,57 @@ public void PopulateFeature<TFeature>(TFeature feature)
...
+            yield return assembly;
+
+            var relatedAssemblies = RelatedAssemblyAttribute.GetRelatedAssemblies(assembly, throwOnError: false)
+                .OrderBy(assembly => assembly.FullName, StringComparer.Ordinal);
+
+            foreach (var relatedAssembly in relatedAssemblies)
+            {
+                yield return relatedAssembly;
+            }
+        }
```
rynowak: Are we able to remove a dependency on dependency context? (yes please 😁)

src/Mvc/Mvc.Core/src/ApplicationParts/ApplicationPartAttribute.cs
pranavkm: We could re-use it here. We generally don't look for `RelatedAssemblyPart` on assemblies that we found using `RelatedAssemblyPart`. However in this case we would need to do for things we discover off of the entry-assembly:

```C#
// MyApp.dll
[RelatedAssemblyPart("MyApp.Views")]
[RelatedAssemblyPart("Rcl")]
[RelatedAssemblyPart("Identity.UI")]
[RelatedAssemblyPart("Identity.UI.Views")]

// Rcl.dll
[RelatedAssemblyPart("Rcl.Views")]

// Identity.UI.dll
[RelatedAssemblyPart("Identity.UI.Views")]
```

The way this would work is - look for `RelatedAssemblyPart` for entry assembly (`MyApp`) and then look for `RelatedAssemblyPart` on those assemblies, but nothing further than that.

rynowak: I'm happy with the new attribute. I wanted to make sure you thought through why it's needed 🤣

src/Mvc/Mvc.Core/ref/Microsoft.AspNetCore.Mvc.Core.csproj
```diff
@@ -15,7 +15,6 @@
     <Reference Include="Microsoft.AspNetCore.Routing.Abstractions"  />
     <Reference Include="Microsoft.AspNetCore.Routing"  />
     <Reference Include="Microsoft.Extensions.DependencyInjection"  />
-    <Reference Include="Microsoft.Extensions.DependencyModel"  />
```
pranavkm: @natemcmaster we could remove `Microsoft.Extensions.DependencyModel` from the shared fx now. None of the other MVC assemblies that reference it are part of Microsoft.AspNetCore.App. Should I remove it?

rynowak: verified S P I C Y

natemcmaster: Let's run this by Damian and Fowler. Can you send mail on this? I'm personally okay with it, but let's make sure we think measure the impact first.

eng/SharedFramework.External.props
```diff
@@ -56,9 +56,6 @@
     <ExternalAspNetCoreAppReference Include="System.Security.Cryptography.Xml"                             Version="$(SystemSecurityCryptographyXmlPackageVersion)" />
     <ExternalAspNetCoreAppReference Include="System.Threading.Channels"                                    Version="$(SystemThreadingChannelsPackageVersion)" />
 
-    <!-- Dependencies from dotnet/core-setup -->
-    <ExternalAspNetCoreAppReference Include="Microsoft.Extensions.DependencyModel"                         Version="$(MicrosoftExtensionsDependencyModelPackageVersion)" />
```
pranavkm: cc @natemcmaster

pranavkm:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=192647&view=logs

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/2267

src/Mvc/Mvc.Razor.RuntimeCompilation/src/AssemblyPartExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+using Microsoft.Extensions.DependencyModel;
+
+namespace Microsoft.AspNetCore.Mvc.ApplicationParts
+{
+    public static class AssemblyPartExtensions
+    {
+        /// <inheritdoc />
+        public static IEnumerable<string> GetReferencePaths(this AssemblyPart assemblyPart)
+        {
+            var assembly = assemblyPart?.Assembly ?? throw new ArgumentNullException(nameof(assemblyPart));
```
rynowak: This should do a cast to the interface. Was the interface removed?

rynowak: IMO it's better if the interface is a detail of this extension method.

pranavkm: I don't follow. What does this look like?

rynowak: Nevermind my brain is crazy.

rynowak (approved):

eng/SharedFramework.External.props
natemcmaster: Can you update this list? https://github.com/aspnet/AspNetCore/issues/3755

davidfowl:
👏 👏

davidfowl:
@pranavkm Can/Did we turn off PreserveCompilationContext now in the Razor SDK?

src/Mvc/Mvc.Core/src/ApplicationParts/ApplicationPartManager.cs
```diff
@@ -52,18 +52,56 @@ public void PopulateFeature<TFeature>(TFeature feature)
 
         internal void PopulateDefaultParts(string entryAssemblyName)
```
davidfowl: @pranavkm it would be good to profile this to understand how this particular piece of code scales (allocations wise) with the number of assemblies. I know it only happens once per app but I'm curious mostly because of the amount of LINQ

pranavkm: Ok. We used to have a ton of Linq in the Deps file code too which is primarily why I didn't think it was a big deal: https://github.com/aspnet/AspNetCore/blob/release/3.0-preview5/src/Mvc/Mvc.Core/src/ApplicationParts/ApplicationAssembliesProvider.cs#L48-L79. We could switch it to vanilla for loops if you think this might be something to be concerned about.

Issue #4135, opened 2018-04-03
Templates do not work in Antares Linux due to missing scheme forwarders


Tratcher:
The 2.1 templates include UseHttpsRedirection and UseHsts by default. These stop working and put your site into an infinite loop if you deploy to Antares Linux (or behind any other reverse proxy besides IIS). The problem is that TLS is terminated by the reverse proxy and Kestrel doesn't know the correct request scheme. OAuth and OIDC also fail in this configuration as they generate incorrect redirects. 

The fix is to add and configure the ForwardedHeaders middleware to fix up the scheme as forwarded from the proxy. UseIISIntegration adds and configures this middleware when running behind IIS but we have no matching lightup and config for Linux e.g. UseApacheIntegration, UseNginxIntegration, etc.. 

Related:
https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/proxy-load-balancer
https://github.com/aspnet/Docs/issues/5806

Proposal:
See what it would take to detect and light up in Antares Linux, and add that code to CreateDefaultBuilder. Also see if this can be generalized to other Apache and NGinx scenarios.

@DamianEdwards for 2.1 rc1 consideration

muratg:
Moving out since preview2 already shipped.

bradygaster:
@muratg i tried this tonight using 2.1 templates to antares linux and it lit up right away (hero path was clean). i will also try on AKS, as that uses, if memory serves me correctly, nginx, and update this with the result. if this also works i'd like to make sure i understand what's awry to help make sure the experiences (kubernetes with nginx, or an AKS out-of-the-box deployment and app service linux).

muratg:
@bradygaster interesting. The template does have UseHttpsRedirection, right? I wonder if there's no reverse proxy involved in antares linux then (would be surprising on its own.) 

On AKS, I believe you'd pass through Azure load balancer. (You may also install an internal reverse proxy, e.g. via nginx-ingress.) Not sure if those would be broken or not.

bradygaster:
@muratg yes, they do have UseHttpsRedirection in Startup.cs. The e2e "hero scenario" to go from dotnet new razor to a working https://dotnetcoreonlinux.azurewebsites.net/ (real url) took about 10 minutes. i'll try it again for kicks now using the new `az webapp up` command, just for kicks. 👍 

[...]

PR #10273, opened 2019-05-15
Add ForwardedHeaders to CreateDefaultBuilder

Tratcher:
#4135 This enables apps to conditionally enable ForwardedHeaders when published to places like Azure Web Apps Linux. The forwarders are needed to correctly set the request scheme that's used when generating links, doing remote auth logins, etc..

Changes:
- Makes `UseForwardedHeaders()` idempotent (but not `UseForwardedHeaders(options)`)
- Adds code to `CreateDefaultBuilder` that enables forwarded headers if environment variable `ASPNETCORE_FORWARDEDHEADERS_ENABLED=true`
- That environment variable was recently added to the App Service Linux container images and should be generally available in a few weeks: Azure-App-Service/dotnetcore-template#1

@bradygaster


src/DefaultBuilder/samples/SampleApp/Program.cs
```diff
@@ -16,19 +16,16 @@ public class Program
...
         }
 
+        public static IHostBuilder CreateHostBuilder(string[] args) =>
+            Host.CreateDefaultBuilder(args)
+                .ConfigureWebHostDefaults(webBuilder =>
+                {
+                    webBuilder.UseStartup<Startup>();
+                });
+
         private static void HelloWorld()
```
JunTaoLuo: How relevant are these samples anymore?

src/DefaultBuilder/src/WebHost.cs
```diff
@@ -224,6 +229,20 @@ internal static void ConfigureWebDefaults(IWebHostBuilder builder)
 
                 services.AddTransient<IStartupFilter, HostFilteringStartupFilter>();
 
+                if (string.Equals("true", hostingContext.Configuration["ForwardedHeaders_Enabled"], StringComparison.OrdinalIgnoreCase))
```
JunTaoLuo: We might want to consolidate these setting strings somewhere instead.

JunTaoLuo (approved):

src/DefaultBuilder/src/WebHost.cs
davidfowl: Can you do some performance tests? More default middleware in the request pipeline means it needs to be scrutinized for performance and allocations...

Tratcher: @davidfowl I'll file a separate task. Any improvements to the middleware itself would not overlap with the changes in this PR.

Tratcher: https://github.com/aspnet/AspNetCore/issues/10276

davidfowl: Dude you can’t introduce allocations into the mainline. As long as you know you’re not doing that I’ll be fine. I’ll be very unhappy otherwise

Tratcher: ? This isn't the first time we've added this middleware, it has been added by default for IIS out-of-proc since 1.0. This change only adds it when you publish to Azure Linux containers.

On a perf note the middleware is [synchronous](https://github.com/aspnet/AspNetCore/blob/1dbc203e197eaa2dc7f1e243ed74cf3ca03efcfb/src/Middleware/HttpOverrides/src/ForwardedHeadersMiddleware.cs#L140-L144), it does not add an async state machine. Its primary allocations are for header [splitting](https://github.com/aspnet/AspNetCore/blob/1dbc203e197eaa2dc7f1e243ed74cf3ca03efcfb/src/Middleware/HttpOverrides/src/ForwardedHeadersMiddleware.cs#L158), and most of the time there will only be a single header entry so no allocation will be required there. Eventually it allocates one IPAddress and one IPEndPoint.

davidfowl: Will it only allocate if those headers exist?

davidfowl: So I did the work to profile and it allocates even if there's nothing to do:

image

Please fix it so that it doesn't allocate unless it *has* to.

Tratcher: > it allocates even if there's nothing to do

In the scenarios where it's enabled, there's always something to do. These headers are not optional in these environments. For a given deployment all requests are expected to have similar headers.

davidfowl: I don't have many hills but this one I'm willing to die on. The needless allocations need to be fixed.

anurse: To be clear, right now this only goes on by default if you explicitly turn it on with an environment variable. The only place we set that is in App Service for Linux which has a front-end that will always provide these headers.

@davidfowl so you want to block this PR on this work rather than handling it inhttps://github.com/aspnet/AspNetCore/issues/10276 ? Just want to make sure we're all on the same page. I do agree that we need to do the work when we introduce these new defaults. My main concern is understanding the true "do nothing" scenario here. If the environment variable is set, the middleware should _never_ "do nothing".

davidfowl: Spoke to Chris, I didn't realize it wasn't on by default. We did discuss that using the middleware without the header is "invalid" yet it just passes through and could maybe fix that. No need to block the PR anymore.

Tratcher: Resolved offline. He didn't understand when it would be enabled.

PR #10275, opened 2019-05-15
start of request queue

DylanDmitri:
- Got VS working
 - Tested semaphore behavior


src/Middleware/RequestQueue/src/Microsoft.AspNetCore.RequestQueue.csproj
```diff
@@ -0,0 +1,20 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <Description>ASP.NET Core middleware for caching HTTP responses on the server.</Description>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <IsAspNetCoreApp>true</IsAspNetCoreApp>
```
natemcmaster:
```suggestion
```
Remove this setting for now. We're not ready to add this assembly to the shared framework.

src/Middleware/RequestQueue/src/Microsoft.AspNetCore.RequestQueue.csproj
```diff
@@ -0,0 +1,20 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <Description>ASP.NET Core middleware for caching HTTP responses on the server.</Description>
```
natemcmaster:
```suggestion
    <Description>ASP.NET Core middleware for something?</Description>
```

Maybe? Not sure what this middleware is for, but let's update the description.

src/Middleware/RequestQueue/src/Microsoft.AspNetCore.RequestQueue.csproj
```diff
@@ -0,0 +1,20 @@
...
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <IsAspNetCoreApp>true</IsAspNetCoreApp>
+    <NoWarn>$(NoWarn);CS1591</NoWarn>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <PackageTags>aspnetcore;cache;caching</PackageTags>
```
natemcmaster:
```suggestion
    <PackageTags>aspnetcore</PackageTags>
```

src/Middleware/RequestQueue/src/Microsoft.AspNetCore.RequestQueue.csproj
```diff
@@ -0,0 +1,20 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <Description>ASP.NET Core middleware for caching HTTP responses on the server.</Description>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <IsAspNetCoreApp>true</IsAspNetCoreApp>
+    <NoWarn>$(NoWarn);CS1591</NoWarn>
```
natemcmaster:
```suggestion
    <WarningsNotAsErrors>$(WarningsNotAsErrors);CS1591</WarningsNotAsErrors>
```

Instead of suppressing undocumented API, I recommend just making this an error instead. This will helps you find undocumented API without failing the build.

natemcmaster (commented):
Only reviewed the .csproj file. It looks like it was copied from another assembly. Here are a few suggestions for how to adjust it for a new project.

src/Middleware/RequestQueue/src/Microsoft.AspNetCore.RequestQueue.csproj
anurse:
```suggestion
    <Description>ASP.NET Core middleware for queuing incoming HTTP requests.</Description>
```

src/Middleware/RequestQueue/src/Microsoft.AspNetCore.RequestQueue.csproj
anurse: Yeah, let's start things off right and document the public APIs as we go :)

src/Middleware/RequestQueue/src/SemaphoreWrapper.cs
```diff
@@ -0,0 +1,33 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.RequestQueue
+{
+    public class SemaphoreWrapper
```
anurse:
```suggestion
    internal class SemaphoreWrapper
```

Let's make this internal and use `InternalsVisibleTo` (@halter73 can help) since this kind of type wouldn't be a public type in the final product.

src/Middleware/RequestQueue/sample/Startup.cs
```diff
@@ -0,0 +1,51 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.HttpsPolicy;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+
+namespace HttpsSample
```
anurse:
```suggestion
namespace RequestQueueSample
```

analogrelay (commented):
Looks like a good start! You need to update the reference assemblies (@halter73 may already have told you about that but if not he can fill you in) and update the sample name but this seems like a good framework to start working in.

src/Middleware/RequestQueue/ref/Microsoft.AspNetCore.RequestQueue.netcoreapp3.0.cs
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Builder
+{
+    public static partial class HstsBuilderExtensions
```
rynowak: It's wierd that this stuff is showing up for this assembly no?

src/Middleware/RequestQueue/src/Microsoft.AspNetCore.RequestQueue.csproj
```diff
@@ -0,0 +1,20 @@
...
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
```
blowdart: You don't have unsafe code, so this isn't needed. Remove.

rynowak: alternative proposal, add unsafe code 😈

src/Middleware/RequestQueue/ref/Microsoft.AspNetCore.RequestQueue.netcoreapp3.0.cs
halter73: It think this was copied. I'll work with @DylanDmitri to regen.

src/Middleware/RequestQueue/test/SemaphoreWrapperTests.cs
```diff
@@ -0,0 +1,56 @@
...
+        public async Task TestWaiting()
+        {
+            var s = new SemaphoreWrapper(1);
+            await s.EnterQueue();
+
+            var waitingTask = s.EnterQueue();
+            Assert.False(waitingTask.IsCompleted);
+
+            s.LeaveQueue();
+            await waitingTask.TimeoutAfter(TimeSpan.FromSeconds(1));
```
halter73: This might sound crazy, but I caution against using a timeout of less than 30 seconds if you can avoid it. While you hope it doesn't happen, you never know when this test might be run in a resource-starved environment. 30 seconds is the default timeout for most SignalR and Kestrel tests.

src/Middleware/RequestQueue/src/SemaphoreWrapper.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        private static SemaphoreSlim semaphore;
+
+        public SemaphoreWrapper(int qlength)
```
davidfowl: queueLength

src/Middleware/RequestQueue/src/SemaphoreWrapper.cs
```diff
@@ -0,0 +1,33 @@
...
+using System.Collections.Generic;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.RequestQueue
+{
+    public class SemaphoreWrapper
+    {
+        private static SemaphoreSlim semaphore;
```
davidfowl: Why is this static?

src/Middleware/RequestQueue/src/SemaphoreWrapper.cs
davidfowl: I missed the memo, why does this type exist?

halter73: Because we wanted to start the process of submitting a PR early to get the experience. At some point, this will be renamed to RequestQueue.

src/Middleware/RequestQueue/src/SemaphoreWrapper.cs
halter73: Good catch. Add a regression test.

src/Middleware/RequestQueue/src/SemaphoreWrapper.cs
davidfowl: Ah, well there's a problem with having the namespace and a type name the same. So lets fix that now as it'll be much harder to change later and it'll make coding annoying.

src/Middleware/RequestQueue/test/SemaphoreWrapperTests.cs
anurse: Agreed. Our tests run at such a high degree of parallelism that they themselves are sometimes subject to thread-pool starvation (perhaps we need a test queuing system 😈)

src/Middleware/RequestQueue/sample/Startup.cs
```diff
@@ -0,0 +1,51 @@
+using System;
```
jkotalik: Nit: namespace.

src/Middleware/RequestQueue/sample/Startup.cs
```diff
@@ -0,0 +1,51 @@
...
+    {
+        // This method gets called by the runtime. Use this method to add services to the container.
+        // For more information on how to configure your application, visit https://go.microsoft.com/fwlink/?LinkID=398940
+        public void ConfigureServices(IServiceCollection services)
+        {
+        }
+
+        public void Configure(IApplicationBuilder app, IWebHostEnvironment environment)
+        {
+
```
jkotalik: nit: new line

src/Middleware/RequestQueue/src/SemaphoreWrapper.cs
```diff
@@ -0,0 +1,33 @@
+using System;
```
jkotalik: nit namespace

src/Middleware/RequestQueue/test/SemaphoreWrapperTests.cs
```diff
@@ -0,0 +1,56 @@
...
+using System;
+using System.Runtime.CompilerServices;
+using Microsoft.AspNetCore.Testing;
+
+namespace Microsoft.AspNetCore.RequestQueue.Tests
+{
+    public class SemaphoreWrapperTests
+    {
+        [Fact]
+        public async Task TestBehavior()
```
jkotalik: I'd rename some of these tests to better describe the behavior they are testing. For example, you could rename this test to `EnterAndLeaveChangesSpotsLeftCorrectly` or something to that effect.

src/Middleware/RequestQueue/test/SemaphoreWrapperTests.cs
jkotalik: I think we use DefaultTimeout (which is 2 minutes) for tests we are concerned about timeouts.

jkotalik (commented):
Good start. I like the direction @halter73

src/Middleware/RequestQueue/src/Microsoft.AspNetCore.RequestQueue.csproj
anurse: Even better :)

```suggestion
    <PackageTags>aspnetcore;queue;queuing</PackageTags>
```

src/Middleware/RequestQueue/src/SemaphoreWrapper.cs
```diff
@@ -0,0 +1,33 @@
...
+        {
+            return semaphore.WaitAsync();
+        }
+
+        public void LeaveQueue()
+        {
+            semaphore.Release();
+        }
+
+        public int SpotsLeft
```
anurse: "Spots" is a pretty casual term for a property name IMO. I'd suggest something like this:

```suggestion
        public int Count
```

or

```suggestion
        public int QueueLength
```

src/Middleware/RequestQueue/test/SemaphoreWrapperTests.cs
```diff
@@ -0,0 +1,56 @@
...
+
+            await s.EnterQueue();
+            Assert.Equal(0, s.SpotsLeft);
+
+            s.LeaveQueue();
+            Assert.Equal(1, s.SpotsLeft);
+        }
+
+        [Fact]
+        public void TestQueueLength()
```
anurse: As @jkotalik mentioned, we like our test names to be nice and descriptive. When the test fails we want someone who sees only the test name to have a relatively good idea of what's going on.

For example here it would be: "EnterQueueBlocksWhenNoQueueSpaceIsAvailable"

Don't be afraid to have long method names when it comes to tests, within some reason. For example "ThisTestEnsuresThatIfYouCallEnterQueueAndThereAreNoSpotsTheTaskReturnedIsNotCompleted" would be **too** verbose 😝

src/Middleware/RequestQueue/src/SemaphoreWrapper.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Microsoft.AspNetCore.RequestQueue
+{
+    public class SemaphoreWrapper
+    {
+        private static SemaphoreSlim semaphore;
+
+        public SemaphoreWrapper(int qlength)
+        {
+            semaphore = new SemaphoreSlim(qlength);
```
anurse: `SemaphoreSlim` is `IDisposable`, so we should be disposing it in this wrapper and properly using `using` in our tests (good time to try out the new C# 8 feature: `using var wrapper = new SemaphoreWrapper(...)` ;))

src/Middleware/RequestThrottling/test/SemaphoreWrapperTests.cs
```diff
@@ -0,0 +1,66 @@
...
+            await waitingTask.TimeoutAfter(TimeSpan.FromSeconds(30));
+        }
+
+        [Fact]
+        public async Task IsEncapsulated()
+        {
+            using var s1 = new SemaphoreWrapper(1);
+            using var s2 = new SemaphoreWrapper(1);
+
+            await s1.EnterQueue();
```
halter73: Everything you await in a test should have a timeout. I would create an `OrTimeout()` extension method that defaults to 30 seconds.

DylanDmitri: Oh yeah I thought since I can assume certain conditions based on past tests (ie that something can enter) that I wouldn't need to validate that again. But the timeouts aren't about validation or logical consistency, they are about automated tests actually running to completion.

src/Middleware/RequestThrottling/test/SemaphoreWrapperTests.cs
```diff
@@ -10,6 +10,13 @@
 
 namespace Microsoft.AspNetCore.RequestThrottling.Tests
 {
+    public static class TaskExtensions
```
halter73: Nit: put this in its own file

src/Middleware/RequestThrottling/sample/Startup.cs
```diff
@@ -26,7 +26,9 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment environment)
         {
             app.Run(async context =>
             {
+                Console.WriteLine("HEWWWO?");
                 await context.Response.WriteAsync("Hello world!");
+                Console.WriteLine("GOODBWWYWE!");
```
halter73: Let's try to leave this stuff out of the PR.

halter73:
Can you add the solution filter file to this PR?

src/Middleware/RequestThrottling/test/SemaphoreWrapperTests.cs
anurse: The timeouts are basically there just in case you get something wrong, so that the test doesn't just hang indefinitely, but fails at some point :).

We have code in a bunch of places to do this (Kestrel, SignalR, etc.). I wonder if we should look at centralizing that in `src/Shared`. For now just copy the one from SignalR I think.

src/Middleware/RequestThrottling/test/SemaphoreWrapperTests.cs
halter73: Copy SignalR's TaskExtensions instead like @anurse suggests and we'll look at sharing the source later.

src/Middleware/RequestThrottling/src/Microsoft.AspNetCore.RequestThrottling.csproj
```diff
@@ -0,0 +1,11 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <Description>ASP.NET Core middleware for queuing incoming HTTP requests, to avoid threadpool starvation.</Description>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <PackageTags>aspnetcore;queue;queuing</PackageTags>
+    <WarningsNotAsErrors>$(WarningsNotAsErrors);CS1591</WarningsNotAsErrors>
```
halter73: Can we remove this line?

halter73 (approved):
🎉

src/Middleware/RequestThrottling/sample/RequestThrottlingSample.csproj
```diff
@@ -0,0 +1,13 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Reference Include="Microsoft.AspNetCore.Server.Kestrel" />
+    <Reference Include="Microsoft.Extensions.Logging.Console" />
+    <Reference Include="Microsoft.AspNetCore.HttpsPolicy" />
```
Tratcher: You should be able to remove HttpsPolicy from here and the slnf

Tratcher (approved):

src/Middleware/RequestThrottling/RequestThrottling.slnf
```diff
@@ -0,0 +1,25 @@
+{
+  "solution": {
```
davidfowl: What is this? @natemcmaster

anurse: + @halter73

Tratcher: It's a saved VS solution filter that only loads the parts of the sln listed here. The Middleware.sln is too large to load and work with, but creating an sln for each middleware is also overkill. Creating a filter for each middleware looks like a promising compromise.
https://docs.microsoft.com/en-us/visualstudio/ide/filtered-solutions?view=vs-2019

natemcmaster: +1. I've actually been thinking maybe we should only have one .sln which includes everything, and then just use a bunch of .slnf files instead....but that's not at the top of my priority list right now.

natemcmaster:
Going to close and re-open. Attempting to re-trigger the CLA bot triggers.

BrennanConroy:
Ouch, next time click squash and merge

DylanDmitri:
woops sorry

Issue #4339, opened 2018-11-30
Switch package versions to SemVer 2.0


natemcmaster:
In 3.0, most other .NET Core packages have started using semver 2.0. This solves a common problem with semver v1, which sorts pre-release labels alphabetically instead of numerically. We've worked around this by zero padding, but still occasionally get this wrong.

e.g. (semver 2) `3.0.0-preview.18572.1` vs (semver 1) `3.0.0-preview-18572-01`

Depends on https://github.com/dotnet/core-setup/issues/4795

**Requirements**

- [ ] We used semver v1 in 3.0 preview 1. Ensure NuGet, npm, and Maven would sort preview 2 packages higher. e.g. ensure `3.0.0-preview.18580.1` > `3.0.0-preview-18579-0056`
- [x] Make sure corehost can handle semver2 - tracking https://github.com/dotnet/core-setup/issues/4795
- [ ] Ensure Maven supports semver2.  If not, keep versioning .jar's with semver v1   (Update: it does not, so keep semver1.)
- [x] Ensure npm supports semver2. If not, keep versioning npm packages with semver v1

natemcmaster:
I think it's too late to do this. At least according to MyGet sorting, SemVer v2 packages would sort as older than SemVer v1 packages.

![image](https://user-images.githubusercontent.com/2696087/49529553-18704480-f86b-11e8-87d7-b18691a4adce.png)

natemcmaster:
Now that we include preview numbers, we could consider doing this. Re-opening for consideration.

natemcmaster:
Based on what I see in https://docs.oracle.com/middleware/1212/core/MAVEN/maven_version.htm#MAVEN8855, Maven does not support semver2, so we'll keep the .jar packages using semver1. (cc @anurse) I'm going to update everything else to use semver2.

anurse:
Yeah, Maven isn't technically semver at all. However it does seem to support `.` and do proper integer sorting now: https://stackoverflow.com/questions/13004443/how-does-maven-sort-version-numbers/31482463#31482463.

Might be worth a try, we can do a quick test of the updated package and see if it works in our core environments (Android Studio, etc.).

anurse:
cc @mikaelm12

PR #10279, opened 2019-05-15
Use Arcade version suffixes for local (3.0.0-dev) and PR builds (3.0.0-ci)

natemcmaster:
Part of #7280 (Convert to Arcade)

This updates our versioning to better match Arcades https://github.com/dotnet/arcade/blob/master/Documentation/CorePackages/Versioning.md

Effective changes
* Local builds will build and produce packages as 3.0.0-dev
* PR builds will build and produce packages as 3.0.0-ci

Everything else should remain the same

### DIFF
--- a/version.props
+++ b/version.props
@@ -17,10 +17,10 @@
     <AspNetCoreModuleVersionRevision>$(AspNetCorePatchVersion)</AspNetCoreModuleVersionRevision>
 
     <OfficialBuildId Condition="'$(OfficialBuildId)' == ''">$(BUILD_BUILDNUMBER)</OfficialBuildId>
-    <OfficialBuildId Condition="'$(OfficialBuildId)' == '' AND '$(IsOfficialBuild)' == 'true'">$([System.DateTime]::Now.ToString('yyyyMMdd')).1</OfficialBuildId>
+    <OfficialBuild Condition="'$(OfficialBuildId)' != '' AND '$(BUILD_REASON)' != 'PullRequest'">true</OfficialBuild>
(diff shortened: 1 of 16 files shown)


version.props
```diff
@@ -40,15 +40,17 @@
     <BuildNumberSuffix>$(_BuildNumberShortDate)-$(VersionSuffixBuildOfTheDayPadded)</BuildNumberSuffix>
 
     <FileRevisionVersion>$(_BuildNumberShortDate)</FileRevisionVersion>
+    <VersionSuffix>$(PreReleaseLabel)-$(BuildNumberSuffix)</VersionSuffix>
```
dougbu: Would changing the dash to a period resolve the format difference in releases between AspNetCoreRuntime.3.0.x64.3.0.0-preview-19081-0488 and Microsoft.Extensions.Logging.3.0.0-preview5.19227.9 ?

dougbu (approved):

version.props
natemcmaster: Yes it would resolve the format difference, but it would create more problems. See https://github.com/aspnet/AspNetCore/issues/4339

dougbu: That issue was written before we moved to include the preview number in our versions. So, we could make the switch when we next update the branding in this repo. Suggest reopening #4339.

natemcmaster: Re opened. I'll bring it up during our area-platform triage. It's not really a build team decision to do this or not because this impacts the appearance and behavior of the product, so I won't attempt to address that in this PR.

PR #10282, opened 2019-05-16
Write Premain errors to the response and write different status codes

### LINKED ISSUE #10008: Create new IIS sub-status code handler load failure

Today we use a 500.0 to indicate a handler load failure. Let's create a new sub-status code for better SEO for folks running into this issue.

### LINKED ISSUE #10011: [ANCM] [in proc] Write pre-main errors to the browser when running in development
Epic #8833 

For pre-main errors, in addition to writing to the event log, surface these errors in the browser when running in a development environment.

jkotalik:
Fixes https://github.com/aspnet/AspNetCore/issues/10011 and will eventually fix https://github.com/aspnet/AspNetCore/issues/10008.

Here is so backstory behind how ANCM writes error pages, and my initial thoughts on how to design dynamic vs static responses.

When ANCM hits a startup exception in the shim or in the request handler, an error "Application" is created, which serves static HTML or the Hosting Startup Error page. This needs to be expanded now to serve dynamic content from hostfxr or potentially from ANCM. Along with checking if disableStartupErrorPage is set, writing text/plain vs text/html, status codes, substatus codes, etc, having a single application to handle multiple failure points has become unmanageable. 

I think we should probably redesign how we server error responses. My initial idea was to create an ENUM or key off of HRESULT to know what type of error occurred, and change the type of application/handler to create. An example of this is what I did with hostfxr errors, where I create a new handler and application. I think the best approach is to create a unique application/handler per type of error which just serves a single type of failure. Then, per each failure type, we can add more dynamic data (like stack trace, hresult, doc link, etc.) when running in development. There may be a good amount of duplication, but trying to cram everything into the same class is getting worse and worse (see ServerApplication.cpp).

If this plan looks good, I'll go ahead and follow through. I just want to confirm with _someone_ before I do a decently size refactoring.


src/Servers/IIS/AspNetCoreModuleV2/AspNetCore/ShimOptions.cpp
```diff
@@ -41,4 +42,16 @@ ShimOptions::ShimOptions(const ConfigurationSource &configurationSource) :
...
+    // Reading environment variables can be added here, but it adds more code to the shim.
+    const auto detailedErrors = Environment::GetEnvironmentVariableValue(L"ASPNETCORE_DETAILEDERRORS").value_or(L"");
+    const auto aspnetCoreEnvironment = Environment::GetEnvironmentVariableValue(L"ASPNETCORE_ENVIRONMENT").value_or(L"");
+    const auto dotnetEnvironment = Environment::GetEnvironmentVariableValue(L"DOTNET_ENVIRONMENT").value_or(L"");
+
+    auto detailedErrorsEnabled = equals_ignore_case(L"1", detailedErrors) || equals_ignore_case(L"true", detailedErrors);
+    auto aspnetCoreEnvironmentEnabled = equals_ignore_case(L"Development", aspnetCoreEnvironment);
+    auto dotnetEnvironmentEnabled = equals_ignore_case(L"Development", dotnetEnvironment);
+
+    m_fIsDevelopment = detailedErrorsEnabled || aspnetCoreEnvironmentEnabled || dotnetEnvironmentEnabled;
```
Tratcher:
```suggestion
    m_fShowDetailedErrors = detailedErrorsEnabled || aspnetCoreEnvironmentEnabled || dotnetEnvironmentEnabled;
```

jkotalik:
Some feedback I received from @shirhatti is that the error pages need work with regards to looking different than the default error page (they look too similar). Currently errors look like this:
Capture

This does have the necessary information, but it isn't presented nicely. We can consider doing more work on this in either this PR or a separate PR.

src/Servers/IIS/AspNetCoreModuleV2/AspNetCore/HandlerResolver.cpp
```diff
@@ -130,6 +149,13 @@ HandlerResolver::GetApplicationFactory(const IHttpApplication &pApplication, std
         // Multiple in-process apps
         if (m_loadedApplicationHostingModel == HOSTING_IN_PROCESS && m_loadedApplicationId != pApplication.GetApplicationId())
         {
+            errorContext.detailedErrorContent = to_multi_byte_string(format(ASPNETCORE_EVENT_MIXED_HOSTING_MODEL_ERROR_MSG, pApplication.GetApplicationId(), options.QueryHostingModel()), CP_UTF8);
```
BrennanConroy: `ASPNETCORE_EVENT_MIXED_HOSTING_MODEL_ERROR_MSG` seems like the wrong error message

src/Servers/IIS/AspNetCoreModuleV2/AspNetCore/HandlerResolver.cpp
```diff
@@ -214,14 +241,25 @@ try
...
+    {
+        m_hHostFxrDll.Load(hostfxrOptions.GetHostFxrLocation());
+    }
+    catch (...)
+    {
+        errorContext.detailedErrorContent = "Could not load hostfxr.dll.";
+        errorContext.statusCode = 500i16;
+        errorContext.subStatusCode = 32i16;
+        errorContext.generalErrorType = "ANCM Failed to Load dll";
+        errorContext.solution = "Most likely, the application was published for a different bitness than w3wp.exe/iisexpress.exe is running as.";
```
BrennanConroy: Is it possible to know for sure what went wrong?

```suggestion
        errorContext.solution = "The application was likely published for a different bitness than w3wp.exe/iisexpress.exe is running as.";
```

src/Servers/IIS/AspNetCoreModuleV2/CommonLib/ServerErrorApplication.h
```diff
@@ -9,16 +9,13 @@
 class ServerErrorApplication : public PollingAppOfflineApplication
 {
 public:
-    ServerErrorApplication(const IHttpApplication& pApplication, HRESULT hr, HINSTANCE moduleInstance)
-        : ServerErrorApplication(pApplication, hr, moduleInstance, true /* disableStartupPage*/, 0 /* page */)
-    {
-    }
-
-    ServerErrorApplication(const IHttpApplication& pApplication, HRESULT hr, HINSTANCE moduleInstance, bool disableStartupPage, int page)
+    ServerErrorApplication(const IHttpApplication& pApplication, HRESULT hr, bool disableStartupPage, std::string responseContent, USHORT status, USHORT substatus, std::string statusText)
```
BrennanConroy: nit: `const &` the `std::string`'s

src/Servers/IIS/AspNetCoreModuleV2/InProcessRequestHandler/StartupExceptionApplication.h
```diff
@@ -13,15 +13,18 @@ class StartupExceptionApplication : public InProcessApplicationBase
     StartupExceptionApplication(
         IHttpServer& pServer,
         IHttpApplication& pApplication,
-        HINSTANCE moduleInstance,
         BOOL disableLogs,
         HRESULT hr,
-        std::vector<byte>&& errorPageContent
-        )
+        std::string errorPageContent,
```
BrennanConroy: nit: `const &`

src/Servers/IIS/AspNetCoreModuleV2/InProcessRequestHandler/StartupExceptionApplication.h
```diff
@@ -13,15 +13,18 @@ class StartupExceptionApplication : public InProcessApplicationBase
...
+        USHORT statusCode,
+        USHORT subStatusCode,
+        std::string statusText)
```
BrennanConroy: nit: `const &`

src/Servers/IIS/AspNetCoreModuleV2/InProcessRequestHandler/StartupExceptionApplication.h
```diff
@@ -30,15 +33,19 @@ class StartupExceptionApplication : public InProcessApplicationBase
 
     HRESULT CreateHandler(IHttpContext* pHttpContext, IREQUEST_HANDLER** pRequestHandler)
     {
-        *pRequestHandler = new ServerErrorHandler(*pHttpContext, 500, 30, "Internal Server Error", m_HR, m_moduleInstance, m_disableLogs, IN_PROCESS_RH_STATIC_HTML, m_errorPageContent);
+        // TOOD this should now take std::string
+        // TODO make this take in status and substatus codes.
```
BrennanConroy: done?

src/Servers/IIS/AspNetCoreModuleV2/InProcessRequestHandler/dllmain.cpp
```diff
@@ -120,6 +121,10 @@ CreateApplication(
         g_fInProcessApplicationCreated = true;
 
         std::unique_ptr<IN_PROCESS_APPLICATION, IAPPLICATION_DELETER> inProcessApplication;
+
+        // TODO not sure how easy it will be to untangle errors here
+        // ErrorContext errorContext;
```
BrennanConroy: Did you figure out if it's possible?

src/Servers/IIS/AspNetCoreModuleV2/OutOfProcessRequestHandler/forwardinghandler.cpp
```diff
@@ -302,7 +304,19 @@ FORWARDING_HANDLER::ExecuteRequestHandler()
...
     else if (fFailedToStartKestrel && !m_pApplication->QueryConfig()->QueryDisableStartUpErrorPage())
     {
-        ServerErrorHandler handler(*m_pW3Context, 502, 5, "Bad Gateway", hr, g_hOutOfProcessRHModule, m_pApplication->QueryConfig()->QueryDisableStartUpErrorPage(), OUT_OF_PROCESS_RH_STATIC_HTML);
+        static std::string htmlResponse = FILE_UTILITY::GetHtml(g_hOutOfProcessRHModule,
+            OUT_OF_PROCESS_RH_STATIC_HTML,
+            502,
+            5,
+            "ANCM Out-Of-Process Startup Failure",
+            "<ul><li> The application process failed to start </li><li> The application process started but then stopped </li><li> The application process started but failed to listen on the configured port </li></ul>");
+        ServerErrorHandler handler(*m_pW3Context,
```
BrennanConroy: nit: newline

src/Servers/IIS/AspNetCoreModuleV2/CommonLib/ServerErrorHandler.h
```diff
@@ -9,107 +9,53 @@
...
-    IHttpContext& m_pContext;
     HRESULT m_HR;
     bool m_disableStartupPage;
-    int m_page;
-    HINSTANCE m_moduleInstance;
     USHORT m_statusCode;
     USHORT m_subStatusCode;
     std::string m_statusText;
-    std::vector<byte> m_ExceptionInfoContent;
+    std::string& m_ExceptionInfoContent;
```
BrennanConroy: I'd just like to publicly say "I'm not a fan of this!"

BrennanConroy (commented):
First pass

src/Servers/IIS/AspNetCoreModuleV2/CommonLib/ServerErrorHandler.h
jkotalik: Allocationsssss

src/Servers/IIS/AspNetCoreModuleV2/InProcessRequestHandler/dllmain.cpp
jkotalik: It is possible, but it isn't an important scenario. There are a few errors that can occur here, but they are very rare because we already invoked hostfxr earlier. After main has started, we would return the developer exception page.

jkotalik:
🆙 📅

src/Servers/IIS/AspNetCoreModuleV2/CommonLib/file_utility.cpp
```diff
@@ -167,3 +168,51 @@ FILE_UTILITY::EnsureDirectoryPathExist(
     return hr;
 }
 
+std::string FILE_UTILITY::GetHtml(HMODULE module, int page, USHORT statusCode, USHORT subStatusCode, std::string speicificReasonPhrase, std::string solution)
```
BrennanConroy: `std::string` -> `const std::string&` in both `GetHtml` calls.

src/Servers/IIS/AspNetCoreModuleV2/InProcessRequestHandler/dllmain.cpp
```diff
@@ -129,7 +134,20 @@ CreateApplication(
...
+                *pHttpApplication,
+                options->QueryDisableStartUpErrorPage(),
+                hr,
+                FILE_UTILITY::GetHtml(g_hServerModule,
+                    IN_PROCESS_RH_STATIC_HTML,
+                    500i16,
+                    30i16,
+                    "ANCM In-Process Start Failure",
+                    "<ul><li>The application failed to start</li><li>The application started but then stopped</li><li>The application started but threw an exception during startup</li></ul>",
+                    std::move(g_errorPageContent)),
```
BrennanConroy: Don't `std::move` a global. And since you should change the `GetHtml` function based on my other comment this wont be needed to save the copy.

BrennanConroy (approved):
Looks good to me. Should anyone else be reviewing?

PR #10283, opened 2019-05-16
Reduce HttpSysServer Allocations

benaadams:
Reduce plaintext allocations by 79%

HttpSys for 1M requests:
* Allocates 34M objects rather than 54M objects (-20M, -37%)
* Allocates 1.64GB rather than 7.83GB (-6.19GB, -79%)

Before
image

After
image


src/Servers/HttpSys/src/FeatureContext.cs
```diff
@@ -600,20 +722,24 @@ internal Task OnCompleted()
...
-                return;
+                return Task.CompletedTask;
             }
 
-            actions.Reverse();
-            // Execute last to first. This mimics a stack unwind.
-            foreach (var actionPair in actions)
+            return Awaited(actions);
+
+            static async Task Awaited(List<(Func<object, Task> callback, object state)> actions)
```
halter73: Nit: since this is duplicated in the same class, you might as well make it a method.

src/Servers/HttpSys/src/MessagePump.cs
```diff
@@ -125,7 +125,8 @@ public Task StartAsync<TContext>(IHttpApplication<TContext> application, Cancell
 
             Contract.Assert(application != null);
 
-            _application = new ApplicationWrapper<TContext>(application);
+            _application = new ApplicationWrapper<TContext>(this, application);
+            _processRequest = new Action<RequestInitalizationContext>((ctx) => _ = _application.ProcessRequestAsync(ctx));
```
halter73:
```suggestion
            _processRequest = (ctx) => _ = _application.ProcessRequestAsync(ctx);
```

Does the above not work?

src/Servers/HttpSys/src/MessagePump.cs
```diff
@@ -172,7 +173,7 @@ private async void ProcessRequestsWorker()
                 }
                 try
                 {
-                    Task ignored = Task.Factory.StartNew(_processRequest, requestContext);
+                    ThreadPool.UnsafeQueueUserWorkItem(_processRequest, requestContext, preferLocal: false);
```
halter73: 👍

davidfowl: Nah, we need to implement IThreadPoolWorkItem on ApplicationWrapper to make it 0 alloc 😄

src/Shared/CoreLib/System/Collections/Concurrent/ConcurrentQueueSegment.cs
```diff
@@ -0,0 +1,339 @@
...
+
+namespace System.Collections.Concurrent
+{
+    /// <summary>
+    /// Provides a multi-producer, multi-consumer thread-safe bounded segment.  When the queue is full,
+    /// enqueues fail and return false.  When the queue is empty, dequeues fail and return null.
+    /// These segments are linked together to form the unbounded <see cref="ConcurrentQueue{T}"/>. 
+    /// </summary>
+    [DebuggerDisplay("Capacity = {Capacity}")]
+    internal sealed class ConcurrentQueueSegment<T>
```
halter73: Why copy this in instead of using the higher-level ConcurrentQueue?

src/Servers/HttpSys/src/MessagePump.cs
benaadams: Changed to implement IThreadPoolWorkItem

src/Servers/HttpSys/src/MessagePump.cs
halter73: Even better. Stopping the FIFO dispatching by itself is really nice though.

src/Shared/CoreLib/System/Collections/Concurrent/ConcurrentQueueSegment.cs
benaadams: ConcurrentQueue is unbounded; whereas the ConcurrentQueueSegment is bounded (and ConcurrentQueue chains them to become unbounded); so it has the characteristics we are after. If we used ConcurrentQueue we'd have to add an additional layer to limit its growth.

Alas exposing it directly was rejected by api review https://github.com/dotnet/corefx/issues/24365

src/Servers/HttpSys/src/MessagePump.cs
benaadams: Added the IThreadPoolWorkItem change https://github.com/aspnet/AspNetCore/pull/10283/commits/073ee5f55373e1bfa41bfa9349e76fe7d3169a39

src/Servers/HttpSys/src/FeatureContext.cs
```diff
@@ -15,12 +15,13 @@
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Http.Features.Authentication;
-using Microsoft.AspNetCore.HttpSys.Internal;
 using Microsoft.Net.Http.Headers;
 
 namespace Microsoft.AspNetCore.Server.HttpSys
 {
-    internal class FeatureContext :
+    using Microsoft.AspNetCore.HttpSys.Internal;
```
davidfowl: intentional?

src/Servers/HttpSys/src/FeatureContext.cs
```diff
@@ -63,29 +65,142 @@ internal class FeatureContext :
...
+            _initializedFields = Fields.None;
+
+            _enableResponseCaching = _requestContext.Server.Options.EnableResponseCaching;
+        }
+
+        internal void Reset()
+        {
+            // Reset fields in sequential order according to layout for better memory/cache access behaviour
+            /*
+            Type layout for 'FeatureContext`1'
```
davidfowl: Remove this, it'll get out of date.

src/Shared/CoreLib/System/Collections/Concurrent/ConcurrentQueueSegment.cs
halter73: That's too bad. How bad do you think the additional layer on ConcurrentQueue would be? I'm thinking TryEnqueue would Interlocked.Increment before Enqueuing into the underlying ConcurrentQueue and TryDequeue would decrement.

While not free, that doesn't sound to bad given one enqueue and dequeue per request. I'm not sure how much ConcurrentQueue itself adds in terms of cost on top of ConcurrentQueueSegment. Hopefully not much.

src/Servers/HttpSys/src/FeatureContext.cs
```diff
@@ -63,29 +65,142 @@ internal class FeatureContext :
 
         private Fields _initializedFields;
 
-        private List<Tuple<Func<object, Task>, object>> _onStartingActions = new List<Tuple<Func<object, Task>, object>>();
-        private List<Tuple<Func<object, Task>, object>> _onCompletedActions = new List<Tuple<Func<object, Task>, object>>();
+        private readonly List<(Func<object, Task> callback, object state)> _onStartingActions = new List<(Func<object, Task> callback, object state)>();
```
davidfowl: Use a stack

src/Servers/HttpSys/src/HttpSysListener.cs
```diff
@@ -264,7 +264,7 @@ private void DisposeInternal()
         /// <summary>
         /// Accept a request from the incoming request queue.
         /// </summary>
-        public Task<RequestContext> AcceptAsync()
+        public Task<RequestInitalizationContext> AcceptAsync()
```
davidfowl: ValueTask?

src/Servers/HttpSys/src/MessagePump.cs
```diff
@@ -12,6 +12,7 @@
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Microsoft.AspNetCore.HttpSys.Internal;
+using System.Collections.Concurrent;
```
davidfowl: sort

src/Servers/HttpSys/src/FeatureContext.cs
benaadams: It was; now its just left over detritus

src/Servers/HttpSys/src/MessagePump.cs
```diff
@@ -121,11 +119,11 @@ public Task StartAsync<TContext>(IHttpApplication<TContext> application, Cancell
             }
 
             // Can't call Start twice
-            Contract.Assert(_application == null);
+            Contract.Assert(_getRequestWorker == null);
 
             Contract.Assert(application != null);
 
-            _application = new ApplicationWrapper<TContext>(application);
+            _getRequestWorker = (server, nativeRequestContext) => ApplicationWrapper<TContext>.RentWrapper(server, nativeRequestContext, this, application);
```
benaadams: @davidfowl has pointed out; cognitively this is probably better as a factory class rather than a capturing lambda (as per the version in IIS)

src/Shared/CoreLib/System/Collections/Concurrent/ConcurrentQueueSegment.cs
davidfowl: The segment is fine, corefx also has a copy in the codebased used in a bunch of places so it doesn't feel bad copying it into our shared code here.

cc @stephentoub for FYI

stephentoub: > corefx also has a copy

Yes and no.  There is a copy of the code, but it's part of the folder that's mirrored between the other repos, where any change in one is automatically propagated to the others.

davidfowl: ASP.NET Core will get it a couple weeks later 😄

src/Servers/HttpSys/src/RequestProcessing/Response.cs
```diff
@@ -32,21 +31,64 @@ internal Response(RequestContext requestContext)
...
+            Headers.IsReadOnly = false;
+            Headers.Clear();
+            _expectedBodyLength = 0;
+            _responseState = ResponseState.Created;
+            _authChallenges = AuthenticationSchemes.None;
+            _boundaryType = BoundaryType.None;
             _cacheTtl = null;
-            _authChallenges = RequestContext.Server.Options.Authentication.Schemes;
+
+            _nativeResponse = default; // 568 bytes!
```
benaadams: Could either be smarter with clearing this 568 bytes or just clear the counts and not clear the headers

src/Shared/CoreLib/System/Collections/Concurrent/ConcurrentQueueSegment.cs
benaadams: > How expensive do you think the additional layer on ConcurrentQueue would be? I'm thinking TryEnqueue would Interlocked.Increment before Enqueuing into the underlying ConcurrentQueue and TryDequeue would decrement, or vice versa if you like counting down.

I think, it would need to be a compare-and-swap loop to force bounding; increment/decrement would only achieve best endeavours as it wouldn't be atomic with the enqueue/dequeue.

But since @stephentoub is here he can probably provide more insight 😉

benaadams: > > corefx also has a copy

> Yes and no. There is a copy of the code, but it's part of the folder that's mirrored between the other repos, where any change in one is automatically propagated to the others.

Does ASP.NET do any mirroring that could be used? (i.e. for WebSocket; Http/2)

halter73: When we discussed this for HTTP/2 we decided we'd mirror manually if at all.

halter73: Also, I really doubt you need to loop to force bounding when you're able to just fail the TryEnqueue call.

benaadams: Could look to see how a ThreadStatic queue per thread would work; would have lower contention than a single count value.

halter73: This makes me wonder what the threadpool threadcount vs. concurrent request ratio is for a typical web server under load.

I'd hold off and hear what @davidfowl has to think. I'm interpreting "ASP.NET Core will get it a couple weeks later 😄" to mean he has faith in the manual mirroring method.

davidfowl: Yes this is fine. Lets keep the queue segment around.

src/Servers/HttpSys/src/AsyncAcceptContext.cs
```diff
@@ -195,8 +177,8 @@ internal void AllocateNativeRequest(uint? size = null, ulong requestId = 0)
             //Debug.Assert(size != 0, "unexpected size");
 
             // We can't reuse overlapped objects
-            uint newSize = size.HasValue ? size.Value : DefaultBufferSize;
-            var backingBuffer = new byte[newSize + AlignmentPadding];
+            int newSize = checked((int)((size.HasValue ? size.Value : DefaultBufferSize) + AlignmentPadding));
+            var backingBuffer = ArrayPool<byte>.Shared.Rent(newSize);
```
Tratcher: Addresses https://github.com/aspnet/AspNetCore/issues/5849

Tratcher: The other common problem with these request buffers is that they can get pinned for a long time and cause memory fragmentation. Pool helps, but only to a point. Ideally we could use a large contiguous slab that was always pinned.

src/Servers/HttpSys/src/HttpSysListener.cs
Tratcher: Yes it is called in parallel.

src/Servers/HttpSys/src/AsyncAcceptContext.cs
Tratcher: Are these rented buffers guaranteed to be zero'd out?

benaadams: No, is it important? Can zero them out if needed

davidfowl: Might be better to use the kestrel pool, like IIS does

src/Servers/HttpSys/src/MessagePump.cs
benaadams: Fxied

src/Servers/HttpSys/src/MessagePump.cs
benaadams: Fixed

src/Servers/HttpSys/src/FeatureContext.cs
benaadams: Fixed

src/Servers/HttpSys/src/FeatureContext.cs
benaadams: Removed

src/Servers/HttpSys/src/FeatureContext.cs
benaadams: Fixed

src/Servers/HttpSys/src/FeatureContext.cs
benaadams: Done

src/Servers/HttpSys/src/RequestProcessing/Response.cs
benaadams: Couldn't work out if not nulling the pointers in the KnownHeaders struct would cause an issue; so left as is.

src/Servers/HttpSys/src/RequestProcessing/RequestContextOfT.cs
```diff
@@ -0,0 +1,29 @@
...
+
+using Microsoft.AspNetCore.HttpSys.Internal;
+
+namespace Microsoft.AspNetCore.Server.HttpSys
+{
+    internal sealed class RequestContext<TContext> : RequestContext
+    {
+        public FeatureContext<TContext> FeatureContext { get; set; }
+
+        public RequestContext() : base()
```
gfoidl:
```suggestion
        public RequestContext()
```

src/Shared/CoreLib/Internal/Padding.cs
```diff
@@ -0,0 +1,26 @@
...
+#if ARM64
+        internal const int CACHE_LINE_SIZE = 128;
+#else
+        internal const int CACHE_LINE_SIZE = 64;
+#endif
+    }
+
+    /// <summary>Padding structure used to minimize false sharing</summary>
+    [StructLayout(LayoutKind.Explicit, Size = PaddingHelpers.CACHE_LINE_SIZE - sizeof(int))]
+    internal struct PaddingFor32
```
gfoidl: Isn't used anywhere. Or is this here because the whole file is copied over from CoreLib?

gfoidl (commented):
Some nits.

src/Servers/HttpSys/src/MessagePump.cs
```diff
@@ -138,24 +136,34 @@ private void ActivateRequestProcessingLimits()
         {
             for (int i = _acceptorCounts; i < _maxAccepts; i++)
             {
-                ProcessRequestsWorker();
+                _ = ProcessRequestsWorker();
```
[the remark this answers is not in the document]
benaadams: As Roslyn will give a CS4014 error about an unawaited task without it

> Error CS4014
> Because this call is not awaited, execution of the current method continues before the call is completed. Consider applying the 'await' operator to the result of the call.

src/Servers/HttpSys/src/MessagePump.cs
```diff
@@ -301,28 +238,138 @@ public void Dispose()
...
+                }
+
+                public void Reset()
+                {
+                    _requestContext.Reset();
+                    _messagePump = null;
+                    _application = null;
+                }
+
+                public void Execute() => _ = ProcessRequestAsync();
```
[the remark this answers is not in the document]
benaadams: CS4014 error without it (and the alternative of `async void` is a very bad code path)

src/Shared/CoreLib/Internal/Padding.cs
benaadams: Just copied whole files, easier for syncing

src/Servers/HttpSys/src/HttpSysListener.cs
benaadams: Not reusing, but using ValueTask which means can dump the TCS https://github.com/aspnet/AspNetCore/pull/10283/commits/fd0eab7ea2ee5aed8389d5adf29a57188e5536a5

src/Servers/HttpSys/src/AsyncAcceptContext.cs
Tratcher: @benaadams how does it look if you measure this pooling change in isolation? I assume this is 90% of the impact from this PR?

The scope of this change makes it quite risky and we want to avoid destabilizing HttpSys when we're not doing a lot of active development and testing on it.

And yes, assume you need to zero out the buffer until proven otherwise.

benaadams: Zero'd it out.

>  how does it look if you measure this pooling change in isolation? I assume this is 90% of the impact from this PR?

Its not actually the pooling I care about though...

So the HostContext is a very chunky struct and gets passed around and stored in async statemachines; its already increased in size for 3.0 and improving the Hosting Logs https://github.com/aspnet/AspNetCore/pull/10102 would make that struct even bigger. So the idea is to additionally pool that on the "connection" and shrink it to a pointer size however as @davidfowl pointed out that would make all servers worse other than Kestrel https://github.com/aspnet/AspNetCore/pull/10180#issuecomment-491540286

The problem is the other servers have no concept of the "connection" so no where to store it; so if we change to pooling request objects they can then store both it and the HttpContext on the reusable requests; this was my main aim.

As it turns out the other servers are also a hotbed of allocations, so saving 48 bytes seems strange when HttpSys is also allocating 5KB+ a request hence the additional reduction.

Mainly I care about the HostContext though...

halter73: > Its not actually the pooling I care about though...

I think we should also look at how this additional pooling affects memory usage when the server idles after load.

Making Hosting's Context struct smaller is nice, and we should look into making it smaller even if we don't end up pooling. It would likely create more GC pressure, but that might be an OK tradeoff.

Tratcher: 90/10. If we get 90% of the improvements in Http.Sys via the array pooling and only 10% from the HostContext, we'd rather take the pooling change here. It's simpler, lower risk, and far more impactful.

benaadams: > Making Hosting's Context struct smaller is nice, and we should look into making it smaller even if we don't end up pooling.

Then it moves to allocating.

It currently doesn't show up as an allocation as its stack copied and then "pool"ed in the `ProcessRequestsAsync` and `ProcessRequests` statemachines (i.e. its a field in the AsyncStateMachineBox so its allocation is amortized for the connection); however it pays a copy cost through each method that receives it due to its size and in the copying to the statemachines,

Moving to a pooled struct wrapped* object means its passed by register and has essentially no copy costs. (*struct wrapped to devirtualize the methods). However, the now heap context has to be stored somewhere to be reused; for Kestrel we can easily do that on the connection, but for other servers there isn't currently anywhere (which is also the same for `DefaultHttpContext` which is currently stored on the connection in Kestrel)

benaadams: > 90/10. If we get 90% of the improvements in Http.Sys via the array pooling and only 10% from the HostContext, we'd rather take the pooling change here.

Sounds like a trade ;)

halter73: > Then it moves to allocating.

Correct. That's why I was saying it might be an OK tradeoff.

benaadams: Before this change, HttpSys currently boxes the struct anyway so.... ¯\\_(ツ)_/¯

Will do the HostContext change with Kestrel benefits, and the ArrayPool change for HttpSys as two PRs

halter73: Ah, when you said you didn't care about the pooling, I thought you were saying the benefit of this was avoiding struct copies. I didn't realize HttpSys was already boxing anyway.

benaadams: > Ah, when you said you didn't care about the pooling

I might have been a bit strong :)

Boxing a single field + allocation of object *may* be similar to boxing large struct? Though haven't measured it.

benaadams: Context changes  #10319

halter73: I'm not against pooling. It seems that even just pooling the request buffers would be a big win. I do want to understand which optimizations can be made independently of one another. If @Tratcher is good with fast tracking just request buffer pooling part of the change (with or without the kestrel pool) maybe we should just get that in now.

src/Servers/HttpSys/src/MessagePump.cs
```diff
@@ -301,28 +238,142 @@ public void Dispose()
...
-        private class ApplicationWrapper<TContext> : IHttpApplication<object>
+        private interface IWorkerFactory
         {
-            private readonly IHttpApplication<TContext> _application;
+            IThreadPoolWorkItem Get(HttpSysListener server, NativeRequestContext requestContext);
+        }
+
+        private sealed class WorkerFactory<TContext> : IWorkerFactory
+        {
+            private const int _maxPooledContexts = 512;
```
halter73: Picking 512 for the _maxPooledContexts seems pretty arbitrary. I feel that it's tuned to perform well in benchmarks that will never establish more that 512 concurrent connections.

At the same time, it seems way too implementation-specific to be adding a knob for this.

halter73: BTW I'm not suggesting we change this necessarily. I just wanted to bring it up so everyone reviewing the PR can consider it.

Tratcher:
Closing this as we don't want to take this much churn in HttpSys.

The main piece of this that's interesting is the ArrayPool for the request buffer. It's only a fraction of the churn here but has the majority of the impact. Let us know if you want to revisit that part.

Issue #8825, opened 2019-03-26
Update visibility of component parameters


rynowak:
# Summary

We have to make a final decision about the visibility of component parameters. We've run into many issues trying to use non-public properties, and I don't think these are going to stop.

## Problems and Goals

- We want to make it hard or impossible to set a component parameter property directly. This is never correct when done *in application code*
- We've tried using non-public properties and found a bunch of things that don't work
   - Reference docs require public
   - Reference assemblies require public
   - Our metadata analysis has (possibly fragile) workarounds in place that we use to look at non-public properties
- We want whatever we choose to work with AOT, but we don't exactly know the requirements. It's generally assumed in AOT that unused non-public code is dead code and can be removed.
- We want whatever we choose to work with testing, but we don't know the requirements of that yet

**TLDR** we have given non-public properties a try and the results aren't great. We got here because we now have to maintain ref assembly code for our components manually. It would be great if the guidance for users worked for us as well.

## Some options

We could recommend that users use public get properties, and we will call the setter regardless of its visibility.

We could recommend that users use public get/set properties and we will write an analyzer to tell you not to set component parameter properties.

I don't like the option that we'd keep on doing analysis over non-public properties. We don't do this anywhere else I can think of, and I don't have confidence that this will play out in our favor wrt tooling.

## Conclusion

We have a conclusion here! 

**Preview 8**
- [x] Reference the component analyzer package from `.Components`
- [x] Add analyzer to warn if a parameter is set in code outside of the inheritance hierarchy or a nested class
- [x] Update our samples
- [x] Verify `[CascadingParameter]s` do not get Analyzesr warnings impacted
- [x] Don't pass analyzers to the declaration compilation target
- [x] Add our component analyzer package to the SDK
- [x] Add a .targets file to our analyzer package to avoid duplicates when included by both the SDK and package
- [x] Make parameters public get/set
- [x] Add analyzer to warn if a property is not public get/set and has `[Parameter]`
- [x] Update our tests

**Before final release**
- Remove workarounds in TagHelper discovery - Tracked by #12291
- Change from warning to error in the analyzer - Tracked by #12293 and #12294

We can leave `[CascadingParameter]` properties alone. They don't interact with tooling, don't want documentation, and don't represent a public contract.

egil:
[...]

PR #10293, opened 2019-05-16
Components auth step 2

SteveSandersonMS:
This PR is mainly about E2E tests for the previous auth work (#10227). It also contains a few commits that address remaining CR feedback from #10227.

Things to note:

 * I had to do a bit of refactoring on how E2E tests decide whether to run client-side or server-side. Previously, the "hosting environment" (dev server, or the TestServer app) and "execution environment" (client-side or server-side) were coupled together as concepts. But the new AuthTests have to do both client-side and server-side execution while running on TestServer, since that's what supplies the auth state. The net result is that the E2E test code is now just more explicit about execution environment in many places, so this is good anyway.
 * From looking at the `ServerAuthenticationStateProvider`, `ClientSideAuthenticationStateData`, and `UserController` classes, you can now get a sense of how this feels when doing client-side Blazor. It's more involved than for server-side Blazor, because we have to fetch the authentication state from the server via an API request.

### DIFF
--- a/eng/GenAPI.exclusions.txt
+++ b/eng/GenAPI.exclusions.txt
@@ -4,6 +4,8 @@ T:Microsoft.AspNetCore.Components.RenderTree.RenderTreeFrame
 T:Microsoft.AspNetCore.Mvc.ApplicationModels.PageParameterModel
 T:Microsoft.AspNetCore.Mvc.ApplicationModels.PagePropertyModel
 # Manually implemented - https://github.com/aspnet/AspNetCore/issues/8825
+T:Microsoft.AspNetCore.Components.AuthorizeView
+T:Microsoft.AspNetCore.Components.CascadingAuthenticationState
 T:Microsoft.AspNetCore.Components.CascadingValue`1
 T:Microsoft.AspNetCore.Components.Forms.DataAnnotationsValidator
 T:Microsoft.AspNetCore.Components.Forms.EditForm
--- a/src/Components/Blazor/Blazor/test/WebAssemblyUriHelperTest.cs
+++ b/src/Components/Blazor/Blazor/test/WebAssemblyUriHelperTest.cs
@@ -29,6 +29,9 @@ namespace Microsoft.AspNetCore.Blazor.Services.Test
         [InlineData("scheme://host/path/", "scheme://host/path/", "")]
         [InlineData("scheme://host/path/", "scheme://host/path/more", "more")]
         [InlineData("scheme://host/path/", "scheme://host/path", "")]
+        [InlineData("scheme://host/path/", "scheme://host/path#hash", "#hash")]
+        [InlineData("scheme://host/path/", "scheme://host/path/#hash", "#hash")]
+        [InlineData("scheme://host/path/", "scheme://host/path/more#hash", "more#hash")]
         public void ComputesCorrectValidBaseRelativePaths(string baseUri, string absoluteUri, string expectedResult)
         {
             var actualResult = _uriHelper.ToBaseRelativePath(baseUri, absoluteUri);
--- a/src/Components/test/testassets/BasicTestApp/AuthTest/AuthRouter.razor
+++ b/src/Components/test/testassets/BasicTestApp/AuthTest/AuthRouter.razor
@@ -1,0 +1,29 @@
+@using Microsoft.AspNetCore.Components.Routing
+@inject IUriHelper UriHelper
+
+@*
+    This router is independent of any other router that may exist within the same project.
+    It exists so that (1) we can easily have multiple test cases that depend on the
+    CascadingAuthenticationState, and (2) we can test the integration between the router
+    and @page authorization rules.
+*@
+
+<CascadingAuthenticationState>
+    <Router AppAssembly=typeof(BasicTestApp.Program).Assembly />
+</CascadingAuthenticationState>
+
+<hr />
+<Links />
+
+@functions {
+    protected override void OnInit()
+    {
+        // Start at AuthHome, not at any other component in the same app that happens to
+        // register itself for the route ""
+        var relativeUri = UriHelper.ToBaseRelativePath(UriHelper.GetBaseUri(), UriHelper.GetAbsoluteUri());
+        if (relativeUri == string.Empty)
+        {
+            UriHelper.NavigateTo("AuthHome");
+        }
+    }
+}
--- a/src/Components/test/testassets/BasicTestApp/AuthTest/AuthRouter.razor
+++ b/src/Components/test/testassets/BasicTestApp/AuthTest/AuthRouter.razor
@@ -20,7 +20,8 @@
     {
         // Start at AuthHome, not at any other component in the same app that happens to
         // register itself for the route ""
-        var relativeUri = UriHelper.ToBaseRelativePath(UriHelper.GetBaseUri(), UriHelper.GetAbsoluteUri());
+        var absoluteUriPath = new Uri(UriHelper.GetAbsoluteUri()).GetLeftPart(UriPartial.Path);
+        var relativeUri = UriHelper.ToBaseRelativePath(UriHelper.GetBaseUri(), absoluteUriPath);
         if (relativeUri == string.Empty)
         {
             UriHelper.NavigateTo("AuthHome");
--- a/src/Components/test/testassets/BasicTestApp/AuthTest/ServerAuthenticationStateProvider.cs
+++ b/src/Components/test/testassets/BasicTestApp/AuthTest/ServerAuthenticationStateProvider.cs
@@ -1,0 +1,43 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Linq;
+using System.Net.Http;
+using System.Security.Claims;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Components;
+
+namespace BasicTestApp.AuthTest
+{
+    // This is intended to be similar to the authentication stateprovider included by default
+    // with the client-side Blazor "Hosted in ASP.NET Core" template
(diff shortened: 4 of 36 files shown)


src/Components/Blazor/Blazor/test/WebAssemblyUriHelperTest.cs
```diff
@@ -29,6 +29,9 @@ public void ComputesCorrectBaseUri(string baseUri, string expectedResult)
         [InlineData("scheme://host/path/", "scheme://host/path/", "")]
         [InlineData("scheme://host/path/", "scheme://host/path/more", "more")]
         [InlineData("scheme://host/path/", "scheme://host/path", "")]
+        [InlineData("scheme://host/path/", "scheme://host/path#hash", "#hash")]
+        [InlineData("scheme://host/path/", "scheme://host/path/#hash", "#hash")]
+        [InlineData("scheme://host/path/", "scheme://host/path/more#hash", "more#hash")]
```
SteveSandersonMS: Support for these cases was needed for the E2E tests. It's because we use a `#server` suffix to trigger server-side execution (and always have done - that's not new), and it turns out the `ToBaseRelativePath` logic didn't understand the case that's now on line 32 here.

src/Components/test/testassets/BasicTestApp/AuthTest/ServerAuthenticationStateProvider.cs
```diff
@@ -0,0 +1,43 @@
...
+        private readonly HttpClient _httpClient;
+
+        public ServerAuthenticationStateProvider(HttpClient httpClient)
+        {
+            _httpClient = httpClient;
+        }
+
+        public override async Task<AuthenticationState> GetAuthenticationStateAsync()
+        {
+            var uri = new Uri(_httpClient.BaseAddress, "/api/User");
```
javiercn: I would add a comment here saying this is not the recomended way to do this in client applications.

src/Components/test/testassets/TestServer/Controllers/UserController.cs
```diff
@@ -0,0 +1,28 @@
...
+using Microsoft.AspNetCore.Mvc;
+
+namespace Components.TestServer.Controllers
+{
+    [Route("api/[controller]")]
+    public class UserController : Controller
+    {
+        // GET api/user
+        [HttpGet]
+        public ClientSideAuthenticationStateData Get()
```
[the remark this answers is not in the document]
javiercn: Yes, you would do that with disablecors, but its extremely discouraged to have an endpoint like that. On SPA the recommendation is to use id tokens or the user endpoint which is a protected resource. I would add a comment in here to make sure people don't get confused.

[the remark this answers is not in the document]
SteveSandersonMS: What is “the user endpoint”?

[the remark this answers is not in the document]
javiercn: https://openid.net/specs/openid-connect-core-1_0.html#UserInfo

Point being, its ok to implement your "own" auth for tests, but it shouldn't be done in prod.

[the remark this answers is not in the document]
SteveSandersonMS: Let’s have a call about this tomorrow. I think we have a mismatch of expectations here.

src/Components/test/testassets/TestServer/Pages/Authentication.cshtml
```diff
@@ -0,0 +1,72 @@
...
+</head>
+<body>
+    <h1 id="authentication">Authentication</h1>
+    <p>
+        This is a completely fake login mechanism for automated test purposes.
+        It accepts any username, with no password.
+    </p>
+    <p>
+        <strong>Obviously you should not do this in real applications.</strong>
+        See also: <a href="https://docs.microsoft.com/aspnet/core/security">documentation on configuring a real login system.</a>
```
DAllanCarr: Tbh I don't like this "Do not do this in a real application..." It would be good to build out a test example which shows are a near real world example.

eng/GenAPI.exclusions.txt
```diff
@@ -4,6 +4,8 @@ T:Microsoft.AspNetCore.Components.RenderTree.RenderTreeFrame
 T:Microsoft.AspNetCore.Mvc.ApplicationModels.PageParameterModel
 T:Microsoft.AspNetCore.Mvc.ApplicationModels.PagePropertyModel
 # Manually implemented - https://github.com/aspnet/AspNetCore/issues/8825
+T:Microsoft.AspNetCore.Components.AuthorizeView
+T:Microsoft.AspNetCore.Components.CascadingAuthenticationState
```
dougbu: Just curious: Are these additions and the related movement between ref assemblies about catching up with previous changes made in product assemblies?

src/Components/test/testassets/TestServer/Pages/Authentication.cshtml
SteveSandersonMS: This is an automated test, not a sample.

eng/GenAPI.exclusions.txt
SteveSandersonMS: This was left over from #10227. We have to write out the ref assembly sources for these components manually because the generator doesn't retain private setters. They have to be retained otherwise the tooling won't work.

natemcmaster: Which tooling? Is there an issue tracking this? I'm not familiar with this problem, but I find it surprising that we need private members in a ref assembly. So far, we've really only meant for ref assemblies to represent public API.

SteveSandersonMS: The Razor editor shipped in VS needs to identify component parameters, and one of the conditions it's looking for is that there's a setter (regardless of visibility). If the ref assemblies claim it's a true read-only property, then the parameters don't show up in intellisense.

The tracking issue is the one mentioned on line 6, https://github.com/aspnet/AspNetCore/issues/8825

Longer term, maybe we could relax the requirement for there to be a setter, and just look for the `[Parameter]` attribute. We could also have an analyzer that warns people not to create `[Parameter]` properties that are truly read-only. @rynowak is the expert for this - what's your expectation of the long-term solution, Ryan?

SteveSandersonMS: Actually, for any follow-up discussion on this topic, please use #8825 :)

Marking as resolved.

src/Components/test/testassets/TestServer/Controllers/UserController.cs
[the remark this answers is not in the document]
SteveSandersonMS: We discussed this offline.

eng/GenAPI.exclusions.txt
rynowak: My preference would be that we make the properties public and change the analyzer to flag *property set* operations.

We've got workarounds in 2-3 different systems now to support the properties being non-public.

src/Components/test/testassets/TestServer/Pages/Authentication.cshtml
```diff
@@ -0,0 +1,72 @@
...
+        if (!string.IsNullOrEmpty(username))
+        {
+            var claims = new List<Claim>
+            {
+                new Claim(ClaimTypes.Name, username),
+                new Claim("test-claim", "Test claim value"),
+            };
+
+            await HttpContext.SignInAsync(
+                new ClaimsPrincipal(new ClaimsIdentity(claims, "FakeAuthenticationType")));
```
rynowak: Should we use the same scheme here? `CookieAuthenticationDefaults.AuthenticationScheme` - not sure that it matters.

src/Components/test/testassets/BasicTestApp/AuthTest/AuthRouter.razor
```diff
@@ -0,0 +1,29 @@
+@using Microsoft.AspNetCore.Components.Routing
+@inject IUriHelper UriHelper
+
+@*
+    This router is independent of any other router that may exist within the same project.
+    It exists so that (1) we can easily have multiple test cases that depend on the
+    CascadingAuthenticationState, and (2) we can test the integration between the router
+    and @page authorization rules.
+*@
```
rynowak: 👍

rynowak (approved):
Seems good.

I'm starting to wonder about the scalability of using a single app for all of these tests, but that's a conversation for another day. I realize this test has some new requirements we haven't had in the past.

src/Components/test/testassets/TestServer/Pages/Authentication.cshtml
SteveSandersonMS: I slightly prefer having specifying `FakeAuthenticationType` here because it clarifies that we're indifferent to the scheme name. But overall I agree it doesn't matter much!

src/Components/test/testassets/BasicTestApp/AuthTest/ServerAuthenticationStateProvider.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Linq;
+using System.Net.Http;
+using System.Security.Claims;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Components;
+
+namespace BasicTestApp.AuthTest
+{
+    // This is intended to be similar to the authentication stateprovider included by default
+    // with the client-side Blazor "Hosted in ASP.NET Core" template
```
SteveSandersonMS: Now I've written out this code, it makes me think we should consider making this more built-in, because it doesn't feel very friendly right now. We could ship a standard `ServerAuthenticationStateProvider` which developers explicitly add to their client-side Blazor DI config, along with some option about what URL it hits to get the state data. We'd also ship a standard `ClientSideAuthenticationStateData` DTO type (probably with a better name). Then the amount of boilerplate code in apps/templates would be pretty minimal (just some DI config plus `UserController` on the server), plus it's still possible to make your own entirely custom `AuthenticationStateProvider` if you don't like ours. Thoughts on this?

PR #10297, opened 2019-05-16
Upload native symbols for ANCM shim.

jkotalik:
Goal is to have a package in the nonshipping folder which contains the symbols.nupkg. I verified if I run dotnet pack --include-symbols, this generates the correct output, now need to verify on CI.


src/Servers/IIS/AspNetCoreModuleV2/Symbols/Microsoft.AspNetCore.ANCMSymbols.csproj
```diff
@@ -0,0 +1,27 @@
...
+    <NativeProjectReference Include="$(MSBuildThisFileDirectory)..\..\AspNetCoreModuleV2\OutOfProcessRequestHandler\OutOfProcessRequestHandler.vcxproj" Platform="Win32" />
+    <NativeProjectReference Include="$(MSBuildThisFileDirectory)..\..\AspNetCoreModuleV2\AspNetCore\AspNetCore.vcxproj" Platform="x64" />
+    <NativeProjectReference Include="$(MSBuildThisFileDirectory)..\..\AspNetCoreModuleV2\AspNetCore\AspNetCore.vcxproj" Platform="Win32" />
+  </ItemGroup>
+
+  <Target Name="AddPackNativeComponents" BeforeTargets="_GetPackageFiles" Condition="$(PackNativeAssets) == 'true'">
+    <ItemGroup>
+      <None Include="%(ShimComponents.DllLocation)"
+            Pack="true"
+            PackagePath="runtimes/win-%(ShimComponents.Platform)/nativeassets/$(NativeAssetsTargetFramework)" />
```
anurse: Let's make this a quasi-real package if we're going to do it. `nativeassets` doesn't seem like the right folder. The CLR packages are in `runtimes\linux-x64\native`

natemcmaster:
Where is the new _._ file used?

src/Servers/IIS/AspNetCoreModuleV2/Symbols/Microsoft.AspNetCore.ANCMSymbols.csproj
```diff
@@ -0,0 +1,27 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <PackageId>Microsoft.AspNetCore.ANCMSymbols</PackageId>
```
natemcmaster:
```suggestion
    <IncludeBuildOutput>false</IncludeBuildOutput>
```

We don't really want to start producing Microsoft.AspNetCore.ANCMSymbols.dll. Consider also overriding the compile targets to prevent this file from getting produced in the first place. See this example: https://github.com/aspnet/AspNetCore/blob/d83bb8efa2398897dc671a98d89b8ad4a87ccb11/src/Analyzers/shared/FeatureDetection/Microsoft.AspNetCore.Analyzers.FeatureDetection.Sources.csproj#L63-L67

jkotalik:
Yay, the package is present in the non shipping packages! https://dev.azure.com/dnceng/public/_build/results?buildId=191464&view=results

natemcmaster:
Weird. Your PR check picked up partial changes from master. CodeCheck.ps1 now checks for java, but the matching azure YML change didn't get applied. Might need to actually merge your PR with master to make AzDO work right.

src/Servers/IIS/AspNetCoreModuleV2/Symbols/Microsoft.AspNetCore.ANCMSymbols.csproj
```diff
@@ -0,0 +1,31 @@
...
+            Pack="true"
+            PackagePath="runtimes/win-%(ShimComponents.Platform)/native/$(NativeAssetsTargetFramework)" />
+      <None Include="%(ShimComponents.PdbLocation)"
+            Pack="true"
+            PackagePath="runtimes/win-%(ShimComponents.Platform)/native/$(NativeAssetsTargetFramework)" />
+    </ItemGroup>
+  </Target>
+
+  <!-- This project does not produce a binary. Disable the compiler. -->
+  <Target Name="Compile" />
```
natemcmaster: Can you verify there isn't actually a .dll getting produced? I suspect that due to implicit top-bottom imports from Microsoft.NET.Sdk, you are actually still compiling a .dll.

If that's the case, the solution is to go from implicit to explicit imports from Microsoft.NET.Sdk so that you can override the compiler targets.

https://github.com/aspnet/AspNetCore/blob/117dd8cff4231c1567938d9835a22ff9c525a9ae/src/Analyzers/shared/FeatureDetection/Microsoft.AspNetCore.Analyzers.FeatureDetection.Sources.csproj#L1-L3
https://github.com/aspnet/AspNetCore/blob/117dd8cff4231c1567938d9835a22ff9c525a9ae/src/Analyzers/shared/FeatureDetection/Microsoft.AspNetCore.Analyzers.FeatureDetection.Sources.csproj#L63-L67

jkotalik: No dll: 
image

 I do need to readd _._, otherwise I get the error `NuGet.Packaging.Core.PackagingException: Cannot create a package that has no dependencies nor content.`

anurse: Weird that you'd get that issue. I'd have thought the `pdb` was sufficient content for that check.

jkotalik:
The only thing that isn't good is that both the nupkg and symbols.nupkg have the same content (pdbs are included in the nupkg). Thoughts on how to fix this?

src/Servers/IIS/AspNetCoreModuleV2/Symbols/Microsoft.AspNetCore.ANCMSymbols.csproj
```diff
@@ -0,0 +1,31 @@
...
+  </ItemGroup>
+
+  <Target Name="AddPackNativeComponents" BeforeTargets="_GetPackageFiles" Condition="$(PackNativeAssets) == 'true'">
+    <ItemGroup>
+      <None Include="%(ShimComponents.DllLocation)"
+            Pack="true"
+            PackagePath="runtimes/win-%(ShimComponents.Platform)/native/$(NativeAssetsTargetFramework)" />
+      <None Include="%(ShimComponents.PdbLocation)"
+            Pack="true"
+            PackagePath="runtimes/win-%(ShimComponents.Platform)/native/$(NativeAssetsTargetFramework)" />
```
anurse: Wouldn't this mean it ends up in `runtimes/win-Win32/native/...`? It should be `win-x86` for 32-bit and `win-x64` for 64-bit. I know that these are mostly bogus packages, but I really think we should align to the conventions anyway, especially if Arcade is going to consume these at some point. We don't know what restrictions they may put on it and while we can always have input, it's better to just *be* a correct NuGet package to start with :).

If I'm wrong about the folder path, then there's no problem :)

jkotalik:
Also, if I can run an AzDo build to verify that symbols are actually loaded to symbols server, that would be great. @natemcmaster ?

natemcmaster:
> The only thing that isn't good is that both the nupkg and symbols.nupkg have the same content (pdbs are included in the nupkg). Thoughts on how to fix this?

NuGet's packaging tasks are meant to tie into some well-known MSbuild item groups that represent the different kinds of build output. `<None ... Pack="true">` is treated as static content. If you want .pdbs to only appear in the symbols.nupkg, you need to use BuiltProjectOutputGroupOutput instead. Microsoft.AspNetCore.App.Runtime.csproj does this...see these snippets:

https://github.com/aspnet/AspNetCore/blob/117dd8cff4231c1567938d9835a22ff9c525a9ae/src/Framework/src/Microsoft.AspNetCore.App.Runtime.csproj#L169-L172
https://github.com/aspnet/AspNetCore/blob/117dd8cff4231c1567938d9835a22ff9c525a9ae/src/Framework/src/Microsoft.AspNetCore.App.Runtime.csproj#L250-L263

> Also, if I can run an AzDo build to verify that symbols are actually loaded to symbols server, that would be great. 

As long as your .symbols.nupkg appears in the build artifacts page on this PR, you're good to go. Symbols aren't publishing for master branch at all right now, but Doug is working on fixing that sometime soon.

src/Servers/IIS/AspNetCoreModuleV2/Symbols/Microsoft.AspNetCore.ANCMSymbols.csproj
jkotalik: No, it's win-x86.

src/Servers/IIS/AspNetCoreModuleV2/Symbols/Microsoft.AspNetCore.ANCMSymbols.csproj
natemcmaster: Do you see the dll in artifacts/bin/Microsoft.AspNetCore.ANCMSymbols/ ?

src/Servers/IIS/AspNetCoreModuleV2/Symbols/Microsoft.AspNetCore.ANCMSymbols.csproj
anurse: 🎉

src/Servers/IIS/AspNetCoreModuleV2/Symbols/Microsoft.AspNetCore.ANCMSymbols.csproj
```diff
@@ -0,0 +1,31 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <IncludeBuildOutput>false</IncludeBuildOutput>
```
natemcmaster:
```suggestion
    <IncludeBuildOutput>false</IncludeBuildOutput>
    <IsPackable>$(PackNativeAssets)</IsPackable>
```
Should fix this for non-Windows builds

jkotalik:
With some offline feedback, we decided to do this a "better" way with NugetPackInput and BuiltProjectOutputGroupOutput. This is what I have right now locally:
```

  <PropertyGroup>
    <IsPackable>$(PackNativeAssets)</IsPackable>
    <BuildOutputTargetFolder>runtimes/$(TargetRuntimeIdentifier)/native/$(NativeAssetsTargetFramework)</BuildOutputTargetFolder>
  </PropertyGroup>

  <Import Project="..\..\build\assets.props" />

  <ItemGroup Condition="'$(BuildIisNativeProjects)' == 'true' AND '$(VCTargetsPath)' != ''">
    <NativeProjectReference Include="$(MSBuildThisFileDirectory)..\..\AspNetCoreModuleV2\OutOfProcessRequestHandler\OutOfProcessRequestHandler.vcxproj" Platform="$(NativeVCPlatform)" />
    <NativeProjectReference Include="$(MSBuildThisFileDirectory)..\..\AspNetCoreModuleV2\AspNetCore\AspNetCore.vcxproj" Platform="$(NativeVCPlatform)" />
  </ItemGroup>

  <Target Name="AddPackNativeComponents" AfterTargets="_CalculateInputsOutputsForPack" Condition="$(PackNativeAssets) == 'true'">
    <ItemGroup>
      <!-- RunShimComponents is based on the current runtime (NativePlatform) which should be the same as TargetRuntimeIdentifier -->
      <BuiltProjectOutputGroupOutput Include="%(RunShimComponents.DllLocation)"/>
      <BuiltProjectOutputGroupOutput Include="%(RunShimComponents.PdbLocation)"/>
      <NuGetPackInput Include="@(BuiltProjectOutputGroupOutput)" />
    </ItemGroup>
  </Target>

  <!-- This project does not produce a binary. Disable the compiler. -->
  <Target Name="Compile" />
  <Target Name="CopyFilesToOutputDirectory" />
</Project>

```
I see the native components being added to NugetPackInput, but I'm getting an error saying the package still doesn't have contents `error NU5017: Cannot create a package that has no dependencies nor content.`

natemcmaster:
You might be running into a target ordering issue. Try adding this to a propertygroup

```xml
 <TargetsForTfmSpecificBuildOutput> 
   $(TargetsForTfmSpecificBuildOutput); 
   AddPackNativeComponents
 </TargetsForTfmSpecificBuildOutput> 
```

natemcmaster:
Also, make this change:
```diff
- <BuildOutputTargetFolder>runtimes/$(TargetRuntimeIdentifier)/native/</BuildOutputTargetFolder>
+ <BuildOutputTargetFolder>runtimes/$(TargetRuntimeIdentifier)/native/$(NativeAssetsTargetFramework)</BuildOutputTargetFolder>
```
NuGet is going to append the value of the TargetFramework property to BuildOutputTargetFolder automatically

natemcmaster:
Oops, that was backwards. I meant
```diff
- <BuildOutputTargetFolder>runtimes/$(TargetRuntimeIdentifier)/native/$(NativeAssetsTargetFramework)</BuildOutputTargetFolder>
+ <BuildOutputTargetFolder>runtimes/$(TargetRuntimeIdentifier)/native/</BuildOutputTargetFolder>
```

jkotalik:
No dice.

jkotalik:
It works 😄

natemcmaster (approved):

PR #10301, opened 2019-05-16
Update log directory

JunTaoLuo:
Addresses https://github.com/aspnet/AspNetCore/issues/7316. 

The issue is that the log directory for functional test logs is by default [`$(RepoRoot)artifacts\log\`](https://github.com/aspnet/Extensions/blob/f41cfded3c12eec0efea89ece1dafe43afa9c6b8/src/Logging/Logging.Testing/src/build/Microsoft.Extensions.Logging.Testing.props#L6) but our ci configuration publishes `artifacts\logs\` instead.

There's a bunch of different ways to fix this:
1. Change the directory in ci.yml
2. Change the default functional test log from artifacts\log to artifacts\logs
3. Set an environment variable to override the default log output dir.

I opted for 1. but can easily go with another approach, especially if artifacts/logs/ is a better conventionl.

### DIFF
--- a/.azure/pipelines/ci.yml
+++ b/.azure/pipelines/ci.yml
@@ -46,14 +46,14 @@ jobs:
     # if they have already been signed. This results in slower builds due to re-submitting the same .nupkg many times for signing.
     # The sign settings have been configured to
 
-    - script: ./eng/scripts/cibuild.cmd -arch x64 /p:DisableCodeSigning=true /bl:artifacts/logs/build.x64.binlog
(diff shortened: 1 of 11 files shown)


natemcmaster:
+1, this contributes to #7280 (convert to arcade) as artifacts/log/ is the Arcade convention.

Can we take this a step further and update _everything_ to use this folder? Add this to build/repo.props to override the KoreBuild default:
```xml
<LogOutputDir>$(ArtifactsDir)log\</LogOutputDir>
```

natemcmaster (approved):

PR #10306, opened 2019-05-16
Update install JDK script an instructions to make it easier to use for local development

natemcmaster:
Resolves https://github.com/aspnet/AspNetCore-Internal/issues/2163

Like the PR description say...the goal here is to make acquiring the JDK much easier for our contributors and devs.

cc @Tratcher

### DIFF
--- a/.azure/pipelines/jobs/default-build.yml
+++ b/.azure/pipelines/jobs/default-build.yml
@@ -146,10 +146,7 @@ jobs:
         command: custom
         arguments: 'locals all -clear'
   - ${{ if and(eq(parameters.installJdk, 'true'), eq(parameters.agentOs, 'Windows')) }}:
-    - powershell: |
-        ./eng/scripts/InstallJdk.ps1 '11.0.1'
-        Write-Host "##vso[task.prependpath]$env:JAVA_HOME\bin"
-
+    - powershell: ./eng/scripts/InstallJdk.ps1
       displayName: Install JDK 11
     - ${{ if eq(parameters.isTestingJob, true) }}:
       - powershell: |
--- a/.azure/pipelines/jobs/default-build.yml
+++ b/.azure/pipelines/jobs/default-build.yml
@@ -121,7 +121,7 @@ jobs:
     DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
     TeamName: AspNetCore
     ${{ if and(eq(parameters.installJdk, 'true'), eq(parameters.agentOs, 'Windows')) }}:
-      JAVA_HOME: $(Agent.BuildDirectory)\.tools\jdk
+      JAVA_HOME: $(Agent.BuildDirectory)\.tools\jdk\win-x64
     ${{ if or(ne(parameters.codeSign, true), ne(variables['System.TeamProject'], 'internal')) }}:
       _SignType: ''
     ${{ if and(eq(parameters.codeSign, true), eq(variables['System.TeamProject'], 'internal')) }}:
--- a/build.ps1
+++ b/build.ps1
@@ -306,7 +306,13 @@ $MSBuildArguments += "/p:TargetOsName=win"
 if (($All -or $BuildJava) -and -not $NoBuildJava) {
     $foundJdk = $false
     $javac = Get-Command javac -ErrorAction Ignore -CommandType Application
-    if ($env:JAVA_HOME) {
+    $localJdkPath = "$PSScriptRoot\.tools\jdk\win-x64\"
+    if (Test-Path "$localJdkPath\bin\javac.exe") {
+        $foundJdk = $true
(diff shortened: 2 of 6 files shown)


build.ps1
```diff
@@ -306,7 +306,13 @@ $MSBuildArguments += "/p:TargetOsName=win"
 if (($All -or $BuildJava) -and -not $NoBuildJava) {
     $foundJdk = $false
     $javac = Get-Command javac -ErrorAction Ignore -CommandType Application
-    if ($env:JAVA_HOME) {
+    $localJdkPath = "$PSScriptRoot\.tools\jdk\win-x64\"
```
Tratcher: That's an interesting solution. We'd want a `clean.cmd` script to go with this that doesn't delete `.dotnet` or `.tools` (or the SignalR js client 😆).

natemcmaster: Could also be a simple git command.
```
git config --global alias.cleanme 'clean -xfd -e .dotnet -e .tools -e src/SignalR/clients/ts/FunctionalTests/node_modules/'
```

Tratcher: "simple" 😁 that's why I suggested a cmd, this affects the whole team.

natemcmaster: Clean makes me a little nervous. Can we discuss this one separately? This is the first request for a clean command I've heard. As clean is potentially destructive, I want to make sure we capture the right intent here so we don't end up with a script that deletes your unstaged code.

natemcmaster: https://github.com/aspnet/AspNetCore-Internal/issues/2532

dougbu (approved):
Just questions&hellip;

eng/scripts/InstallJdk.ps1
```diff
@@ -1,27 +1,52 @@
...
-mkdir (split-path -parent $env:JAVA_HOME) -ea ignore | out-null
-Write-Host "Installing JDK to $env:JAVA_HOME"
-Move-Item "$tempDir/jdk/jdk-${jdkVersion}" $env:JAVA_HOME
-Write-Host "Done installing JDK to $env:JAVA_HOME"
+Write-Host "Installing JDK to $installDir"
+Move-Item "$tempDir/jdk/jdk-${JdkVersion}/*" $installDir
+Write-Host "Done installing JDK to $installDir"
+
+if ($env:TF_BUILD) {
+    Write-Host "##vso[task.prependpath]$installDir\bin"
```
javiercn: I imagine this is the bit that does the magic to put java inside the PATH?

javiercn:
I’m ok with whatever you do here so long as things keep working and that it doesn’t become harder to enable E2E tests for Mac and Linux in the future.

javiercn (commented):
Forgot to submit the review, but essentially what my top level comment said.

.azure/pipelines/jobs/default-build.yml
```diff
@@ -121,7 +121,7 @@ jobs:
     DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
     TeamName: AspNetCore
     ${{ if and(eq(parameters.installJdk, 'true'), eq(parameters.agentOs, 'Windows')) }}:
-      JAVA_HOME: $(Agent.BuildDirectory)\.tools\jdk
+      JAVA_HOME: $(Agent.BuildDirectory)\.tools\jdk\win-x64
```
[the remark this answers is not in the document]
natemcmaster: I was planning to add a linux version later which would install into .tools/jdk/linux-x64

.azure/pipelines/jobs/default-build.yml
```diff
@@ -146,10 +146,7 @@ jobs:
         command: custom
         arguments: 'locals all -clear'
   - ${{ if and(eq(parameters.installJdk, 'true'), eq(parameters.agentOs, 'Windows')) }}:
-    - powershell: |
-        ./eng/scripts/InstallJdk.ps1 '11.0.1'
-        Write-Host "##vso[task.prependpath]$env:JAVA_HOME\bin"
-
+    - powershell: ./eng/scripts/InstallJdk.ps1
```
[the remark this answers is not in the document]
natemcmaster: Nothing changed, these just moved into the script.

eng/scripts/InstallJdk.ps1
natemcmaster: Yes. We could also use build.ps1/sh to prepend PATH with JAVA_HOME for local builds.

.azure/pipelines/jobs/default-build.yml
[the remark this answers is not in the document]
dougbu: That doesn't require changing this directory because the Linux SDK won't be useful on Windows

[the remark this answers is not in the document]
natemcmaster: Linux SDKs might be installed if you are using Bash for Windows, like I do :)

Issue #9349, opened 2019-04-13
Expose CipherSuitesPolicy on Kestrel


Tratcher:
SslStream has a new API for controlling ciphers that Kestrel should surface.
https://github.com/dotnet/corefx/pull/36775/files#diff-bd768ea3857573965e853e2ce2563434R134

HTTP/2 requires restricting the ciphers used. Do we do this by default or leave it as an option?

Tratcher:
@davidfowl:

> it would be good if we could expose this entire type (if it makes sense) so we could avoid the property by property copy when new features are added.

Since we need to create one per connection you could expose a `Action<SslServerAuthenticationOptions>` to let people tweak the configuration, but it's not the simplest type to configure:

https://github.com/aspnet/AspNetCore/blob/7bd1e5ea2022039e23d0f83f0535a7570891176b/src/Servers/Kestrel/Core/src/Internal/HttpsConnectionAdapter.cs#L135-L171

anurse:
Let's add the configuration callback.

Wayne-Xiong:
Is there an example showing how to set the cipher suites in Kestrel? We are running our .NET core service in a Linux container and would love to see how we can make our service compliant. Thanks!

Tratcher:
@Wayne-Xiong no, the in-app feature hasn't been built yet. For now you need to configure your ciphers directly on the machine in OpenSsl.

Wayne-Xiong:
Thanks Chris. Do we have an estimate on when we will have this feature ready? Also, do you know how to config that in OpenSSL? Is there a sample somewhere we can take a look?

Tratcher:
This is scheduled for 3.0.0-preivew6, our next milestone.

@blowdart for OpenSsl config suggestions.

blowdart:
AFAIK You can't configure it in OpenSSL without recompiling a custom version. Instead you'd block in nginx or apache (https://medium.com/@davetempleton/tls-configuration-cipher-suites-and-protocols-a01ee7005778)

Wayne-Xiong:
Thanks guys. @Tratcher, when is the 3.0.0-preview6 release scheduled? Would love to understand about the timeline a little bit. Thanks!

Tratcher:
@shirhatti for schedule questions.

anurse:
> when is the 3.0.0-preview6 release scheduled?

[...]

PR #10310, opened 2019-05-16
Expose the SslServerAuthenticationOptions via a callback

Tratcher:
This covers most of #9349 by exposing SslServerAuthenticationOptions so the developer can customize the options without us exposing each individually. 

The second half of #9349 will be to experiment with the new CipherSuitesPolicy on SslServerAuthenticationOptions to see how to use it to meet the HTTP/2 spec requirements. It's unclear yet if that will be a sample, kestrel change, etc..


src/Servers/Kestrel/test/InMemory.FunctionalTests/HttpsTests.cs
```diff
@@ -380,6 +380,42 @@ public async Task ClientAttemptingToUseUnsupportedProtocolIsLoggedAsDebug()
...
+                new TestServiceContext(LoggerFactory),
+                listenOptions =>
+                {
+                    listenOptions.UseHttps(httpsOptions =>
+                    {
+                        httpsOptions.ServerCertificate = testCert;
+                        httpsOptions.OnAuthenticate = (connectionContext, authOptions) =>
+                        {
+                            Assert.Same(testCert, authOptions.ServerCertificate);
+                            onAuthenticateCalled = true;
```
halter73: Would it work if you didn't set `httpsOptions.ServerCertificate` and just set in in OnAuthenticate? Can we test that?

halter73 (approved):

src/Servers/Kestrel/test/InMemory.FunctionalTests/HttpsTests.cs
Tratcher: Almost. We have argument validation on HttpsConnectionAdapter that says you must provide ServerCertificate or ServerCertificateSelector. I can make a test work.

PR #10311, opened 2019-05-16
Mark ReactTemplate_IndividualAuth_NetCore_LocalDb flaky

Tratcher:

This has been failing on all of my PRs today and looks like a regression in @halter73's change yesterday.

### DIFF
--- a/src/ProjectTemplates/test/SpaTemplateTest/ReactTemplateTest.cs
+++ b/src/ProjectTemplates/test/SpaTemplateTest/ReactTemplateTest.cs
@@ -28,6 +28,7 @@ namespace Templates.Test.SpaTemplateTest
             => SpaTemplateImplAsync("reactindividual", "react", useLocalDb: false, usesAuth: true);
 
         [Fact]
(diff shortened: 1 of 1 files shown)


natemcmaster (approved):

halter73 (approved):

halter73:
@anurse We might want to reprioritize https://github.com/aspnet/AspNetCore-Internal/issues/2348 since it looks like the test has gotten a lot more flaky since I merged https://github.com/aspnet/AspNetCore/pull/10165.

If we cannot figure out the root cause, we might want to look at reverting my PR is it isn't reverted already be then for perf reasons.

Tratcher:
@halter73 now AngularTemplate_IndividualAuth_Works is failing with the same error. Time to revert.

ryanbrandenburg:
Oh is THAT what causes the 
```
[Severe] https://localhost:14615/Identity/lib/bootstrap/dist/js/bootstrap.bundle.js - Failed to load resource: net::ERR_CONNECTION_CLOSED
```
stuff? I was operating under the assumption that it was something wrong with the certificates.

anurse:
@halter73 Sounds good, thanks for volunteering :).

PR #10312, opened 2019-05-16
[Design]Force certificate usage

ryanbrandenburg:
I'm using this PR to try to tease out failures on VSTS agents due to certificates.


ryanbrandenburg:
It turned out that the problem I was trying to solve here was actually caused by a Kestrel change.

PR #10314, opened 2019-05-17
Support a few more complex types with DefaultTempDataSerializer

### LINKED ISSUE #9540: Add support for collection of primitives in DefaultTempDataSerializer
The TempDataSerializer based on Newtonsoft.Json supports some collection types. We should add parity features to the System.Text.Json based serializer.

pranavkm:
Fixes https://github.com/aspnet/AspNetCore/issues/9540


src/Mvc/Mvc.ViewFeatures/src/Infrastructure/DefaultTempDataSerializer.cs
```diff
@@ -150,7 +232,10 @@ public override bool CanSerializeType(Type type)
                 type == typeof(string) ||
                 type == typeof(bool) ||
                 type == typeof(DateTime) ||
-                type == typeof(Guid);
+                type == typeof(Guid) ||
+                typeof(ICollection<int>).IsAssignableFrom(type) ||
+                typeof(ICollection<string>).IsAssignableFrom(type) ||
+                typeof(IDictionary<string, string>).IsAssignableFrom(type);
```
pranavkm: Json.Net supports a bunch more collection types like `List<DateTime>, and `Dictionary<string, Guid>`, but we have to write our own dictionary deserialization for all of these scenarios with System.Text.Json. I'm limiting it to a couple of meaningful collection types. We can always tell users to continue using Json.NET if they run in to issues.

src/Identity/testassets/Identity.DefaultUI.WebSite/StartupBase.cs
```diff
@@ -50,8 +50,7 @@ public virtual void ConfigureServices(IServiceCollection services)
                 .AddRoles<IdentityRole>()
                 .AddEntityFrameworkStores<TContext>();
 
-            services.AddMvc()
-                .AddNewtonsoftJson();
+            services.AddMvc();
```
pranavkm: https://github.com/aspnet/AspNetCore/issues/10139

src/Mvc/Mvc.NewtonsoftJson/test/BsonTempDataSerializerTest.cs
```diff
@@ -235,6 +235,34 @@ public override void RoundTripTest_GuidToString()
             Assert.Equal(value.ToString(), roundTripValue);
         }
 
+        [Fact]
+        public void RoundTripTest_ListOfDateTime()
+        {
+            // Documents the behavior of round-tripping a Guid value as a string
```
rynowak: comment looks like a bad copy-pate

src/Mvc/Mvc.ViewFeatures/src/Infrastructure/DefaultTempDataSerializer.cs
rynowak: Or even better they can use strings 😆

rynowak: I don't really want to set up us to add a million different options here. like, I don't think it's a success condition if users have to ask to rebuild the JSON serializer piecemeal. 

Ultimately we're going to get support for the serializer to read from the reader directly. Does that solve this issue once and for all for us or will there still be limitations?

pranavkm: > Does that solve this issue once and for all for us or will there still be limitations?

In some ways. The larger problem is that we have to inspect the value and determine what the value intended to stash in TempData`. We essentially need a limited form of `TypeNameHandling`

pranavkm:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=192053

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/2329

src/Mvc/Mvc.ViewFeatures/src/Infrastructure/DefaultTempDataSerializer.cs
rynowak: Right. I remember the issue now. Exposing TempData as `IDictionary<string, object>` is just fundamentally problematic. The API contract implies that we will round-trip values with their original runtime type.

src/Mvc/Mvc.ViewFeatures/src/Infrastructure/DefaultTempDataSerializer.cs
```diff
@@ -73,6 +81,53 @@ internal class DefaultTempDataSerializer : TempDataSerializer
...
+
+                return array.ToArray();
+            }
+            else if (arrayElement[0].Type == JsonValueType.Number)
+            {
+                var array = new List<int>();
+
+                foreach (var item in arrayElement.EnumerateArray())
+                {
+                    array.Add(item.GetInt32());
```
mkArtakMSFT: What if the elements are actually strings. If that string for the first element will be a number, this will eventually fail. Is there a model, where the user can specify explicitly, what type of data to expect or just enforce `string` array somehow?

pranavkm: Not super sure I follow. Is this if the collection is heterogeneous? We control what goes in here so we only allow collection of strings or collection of ints.

rynowak (approved):

Issue #8630, opened 2019-03-18
[Tooling] Using TagHelpers in void returning methods at design time does not cause an error


NTaylorMullen:
### Describe the bug
TagHelpers aren't indicated as an error inside of a void returning method at design time. This is due to how we change rendering at design time vs. runtime.

i.e. no errors:
![image](https://user-images.githubusercontent.com/2008729/54554913-dc24a800-4972-11e9-9ac7-6e7f832ae171.png)

NTaylorMullen:
An interesting case here will be handling things like `~/` which don't "TagHelperify" at design time but do at runtime.

NickCraver:
Is there a chance of `~/` *not* being async in nature overall? It seems...very odd to experience. As a user, I'd just assume replace it with a static somewhere to avoid all the pain it causes downstream today.

rynowak:
I super duper agree and it's under consideration. The perf implications of this being a tag helper are quite large.

NTaylorMullen:
202ccebb0d259585cdcb3c30d34cd93c0d1bde3b

NTaylorMullen:
This does not cover `~/` so filed https://github.com/aspnet/AspNetCore/issues/10734

PR #10316, opened 2019-05-17
Add analyzer to detect TagHelpers in code blocks.

NTaylorMullen:
- We detect `__tagHelperRunner.RunAsync` method calls in lambdas, anonymous methods, local functions, and method declarations. Upon detection we verify that the methods are `async Task` returning and if not log an error.
- Analyzed to the best of my knowledge 💪, let me know if I messed up anywhere big.

#8630

image


src/Mvc/Mvc.Analyzers/src/TagHelpersInCodeBlocksAnalyzer.cs
```diff
@@ -0,0 +1,157 @@
...
+                    return;
+                }
+
+                InitializeWorker(context, symbolCache);
+            });
+        }
+
+        private void InitializeWorker(CompilationStartAnalysisContext context, SymbolCache symbolCache)
+        {
+            context.RegisterOperationBlockStartAction(startBlockContext =>
```
NTaylorMullen: Need this in order to retrieve method symbols for `IMethodBodyOperation`s

rynowak:
Just here to say that this is D O P E. /cc @NickCraver

src/Mvc/Mvc.Analyzers/src/SymbolNames.cs
```diff
@@ -72,5 +72,13 @@ internal static class SymbolNames
         public const string ConfigureServicesMethod = "ConfigureServices";
 
         public const string ConfigureMethod = "Configure";
+
+        public const string RunAsyncMethodName = "RunAsync";
+
+        public const string TaskTypeName = "System.Threading.Tasks.Task";
+
+        public const string TagHelperRunnerFieldName = "__tagHelperRunner";
```
NTaylorMullen: Remove

src/Mvc/Mvc.Analyzers/src/TagHelpersInCodeBlocksAnalyzer.cs
```diff
@@ -0,0 +1,157 @@
...
+using System.Collections.Immutable;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Operations;
+
+namespace Microsoft.AspNetCore.Mvc.Analyzers
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class TagHelpersInCodeBlocksAnalyzer : DiagnosticAnalyzer
```
NTaylorMullen: @pranavkm / @rynowak This PR is in design because it's my first take at adding Anaylzers and I want to make sure I'm not doing anything catastrophically wrong design wise.

src/Mvc/Mvc.Analyzers/src/TagHelpersInCodeBlocksAnalyzer.cs
```diff
@@ -0,0 +1,157 @@
...
+            SupportedDiagnostics = ImmutableArray.Create(new[] { TagHelperInCodeBlockDiagnostic });
+        }
+
+        private DiagnosticDescriptor TagHelperInCodeBlockDiagnostic { get; }
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; }
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze | GeneratedCodeAnalysisFlags.ReportDiagnostics);
```
rynowak: This deserves some comments, because it's not a typical thing to do in an analyzer.

src/Mvc/Mvc.Analyzers/src/TagHelpersInCodeBlocksAnalyzer.cs
```diff
@@ -0,0 +1,157 @@
...
+                    if (parent == null)
+                    {
+                        return;
+                    }
+
+                    var methodSymbol = (IMethodSymbol)(parent switch
+                    {
+                        ILocalFunctionOperation localFunctionOperation => localFunctionOperation.Symbol,
+                        IAnonymousFunctionOperation anonymousFunctionOperation => anonymousFunctionOperation.Symbol,
+                        IMethodBodyOperation methodBodyOperation => startBlockContext.OwningSymbol,
```
rynowak: so there is a way. Nice.

src/Mvc/Mvc.Analyzers/src/TagHelpersInCodeBlocksAnalyzer.cs
```diff
@@ -0,0 +1,157 @@
...
+                    if (methodSymbol == null)
+                    {
+                        // Unsupported operation type.
+                        return;
+                    }
+
+                    if (!methodSymbol.IsAsync ||
+                        !symbolCache.TaskType.IsAssignableFrom(methodSymbol.ReturnType))
+                    {
+                        if (parent.Syntax.GetDiagnostics().Any(diagnostic => diagnostic.Id == TagHelperInCodeBlockDiagnostic.Id))
```
rynowak: Did you actually see this happen? I would be really surprised if this works. I wouldn't expect the syntax tree to update DURING your analyzer running.

src/Mvc/Mvc.Analyzers/src/TagHelpersInCodeBlocksAnalyzer.cs
```diff
@@ -0,0 +1,157 @@
...
+
+                InitializeWorker(context, symbolCache);
+            });
+        }
+
+        private void InitializeWorker(CompilationStartAnalysisContext context, SymbolCache symbolCache)
+        {
+            context.RegisterOperationBlockStartAction(startBlockContext =>
+            {
+                startBlockContext.RegisterOperationAction(context =>
```
rynowak: It might be useful to check the `startBlockContext` up front to prevent doing a bunch of work on async methods. If that's super hard them I'm not worried about it.

src/Mvc/Mvc.Analyzers/src/TagHelpersInCodeBlocksAnalyzer.cs
```diff
@@ -0,0 +1,157 @@
...
+                    if (!methodSymbol.IsAsync ||
+                        !symbolCache.TaskType.IsAssignableFrom(methodSymbol.ReturnType))
+                    {
+                        if (parent.Syntax.GetDiagnostics().Any(diagnostic => diagnostic.Id == TagHelperInCodeBlockDiagnostic.Id))
+                        {
+                            // Diagnostic has already been logged on this syntax item, no-op.
+                            return;
+                        }
+
+                        context.ReportDiagnostic(
```
rynowak: The typical way to de-dupe stuff btw is the following:

1. Register an operation block start operation action
2. Create a collection, or some data to track state
3. Register an operation action (use this to collect the data you need)
4. Register an operation block end operation action (use this to report diagnostics)

This way you only report a diagnostic once per block.

src/Mvc/Mvc.Analyzers/src/TagHelpersInCodeBlocksAnalyzer.cs
```diff
@@ -0,0 +1,157 @@
...
+                            Diagnostic.Create(
+                                TagHelperInCodeBlockDiagnostic,
+                                parent.Syntax.GetLocation()));
+                    }
+
+                }, OperationKind.Await);
+            });
```
rynowak: Is there any useful parameter you can pass to this the OperationBlockStartAction?

src/Mvc/Mvc.Analyzers/src/TagHelpersInCodeBlocksAnalyzer.cs
```diff
@@ -0,0 +1,157 @@
...
+        }
+
+        private bool IsTagHelperRunnerRunAsync(IMethodSymbol method, SymbolCache symbolCache)
+        {
+            if (method.IsGenericMethod)
+            {
+                return false;
+            }
```
rynowak: Does this check really buy you anything?

src/Mvc/Mvc.Analyzers/src/TagHelpersInCodeBlocksAnalyzer.cs
```diff
@@ -0,0 +1,157 @@
...
+
+            return false;
+        }
+
+        private readonly struct SymbolCache
+        {
+            public SymbolCache(Compilation compilation)
+            {
+                var tagHelperRunnerType = compilation.GetTypeByMetadataName(SymbolNames.TagHelperRunnerTypeName);
+                var members = tagHelperRunnerType.GetMembers(SymbolNames.RunAsyncMethodName);
```
rynowak: This needs a null check!

src/Mvc/Mvc.Analyzers/src/TagHelpersInCodeBlocksAnalyzer.cs
NTaylorMullen: Nah, just being dumb 😆

src/Mvc/Mvc.Analyzers/src/TagHelpersInCodeBlocksAnalyzer.cs
NTaylorMullen: Didn't fully test this part. You be right though. This totally fails to work as expected 😄 

image

src/Mvc/Mvc.Analyzers/src/TagHelpersInCodeBlocksAnalyzer.cs
NTaylorMullen: I could do it for methods in `@functions` but I feel like that separation of logic would be a little wonky to have separated here. Everything else I couldn't because of how operations work on nested methods aka
```
@{
    void Foo1()
    {
        <p>Hello WOrld</p>

        Func<Task> func2 = () =>
        {
            <distributed-cache name="foo"></distributed-cache>
        };
    }
}
```

src/Mvc/Mvc.Analyzers/src/TagHelpersInCodeBlocksAnalyzer.cs
NTaylorMullen: Nope

src/Mvc/Mvc.Analyzers/src/TagHelpersInCodeBlocksAnalyzer.cs
NTaylorMullen: > 4\. Register an operation block end operation action (use this to report diagnostics)

Do you mean a compilation end action?

src/Mvc/Mvc.Analyzers/src/TagHelpersInCodeBlocksAnalyzer.cs
```diff
@@ -0,0 +1,157 @@
...
+        private DiagnosticDescriptor TagHelperInCodeBlockDiagnostic { get; }
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; }
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.ConfigureGeneratedCodeAnalysis(GeneratedCodeAnalysisFlags.Analyze | GeneratedCodeAnalysisFlags.ReportDiagnostics);
+            context.RegisterCompilationStartAction(context =>
+            {
+                var symbolCache = new SymbolCache(context.Compilation);
```
pranavkm: I am trying out a `TryCreate` pattern inspired by a different PR that @rynowak sent out: (https://github.com/aspnet/AspNetCore/pull/10375/files#diff-873a6b7499a94ad1cace9d0ecb095fa2R26). Might be useful to consider switching this here.

pranavkm:
```C#
if (!SymbolCache.TryCreate(context.Compilation, out var symbolCache))
{
                    // No-op if we can't find types we care about.
                    return;
```

src/Mvc/Mvc.Analyzers/src/TagHelpersInCodeBlocksAnalyzer.cs
pranavkm: Think `RegisterOperationBlockAction` would be the corresponding end operation.

src/Mvc/Mvc.Analyzers/src/TagHelpersInCodeBlocksAnalyzer.cs
```diff
@@ -0,0 +1,157 @@
...
+            {
+                return false;
+            }
+
+            return true;
+        }
+
+        private bool IsParentMethod(IOperation operation)
+        {
+            if (operation.Kind == OperationKind.LocalFunction)
```
pranavkm: Wouldn't the `OperationKind.MethodBody` catch this? LocalFunctions would have bodies too, no?

pranavkm (commented):
Looks neat. We have a way to unit test these and we've checked in "generated" files in the MVC repo in the past: https://github.com/aspnet/AspNetCore/blob/master/src/Mvc/test/WebSites/RazorBuildWebSite.Views/Pages/Rzc/Page.cs

src/Mvc/Mvc.Analyzers/src/TagHelpersInCodeBlocksAnalyzer.cs
NTaylorMullen: Talked offline, gist, no it's different 😄

NTaylorMullen:
🆙 📅  Addressed all comments and added tests.

NickCraver:
Looks like builds above went kaboom:
```
src\Mvc\Mvc.Analyzers\src\TagHelpersInCodeBlocksAnalyzer.cs(73,40): error CS8600: Converting null literal or possible null value to non-nullable type. [src\Mvc\Mvc.Analyzers\src\Microsoft.AspNetCore.Mvc.Analyzers.csproj]
```

src/Mvc/Mvc.Analyzers/src/DiagnosticDescriptors.cs
```diff
@@ -53,7 +53,15 @@ public static class DiagnosticDescriptors
                 isEnabledByDefault: true,
                 helpLinkUri: "https://aka.ms/AA20pbc");
 
-        
         // MVC1005 reserved for startup
+
+        public static readonly DiagnosticDescriptor MVC1006_FunctionsContainingTagHelpersMustBeAsyncAndReturnTask =
+            new DiagnosticDescriptor(
+                "MVC1006",
+                "Functions containing TagHelpers must be async and return Task.",
```
pranavkm: super nit: Should `Functions` -> `Methods`? What does the compiler do?

src/Mvc/Mvc.Analyzers/test/Mvc.Analyzers.Test.csproj
```diff
@@ -16,6 +16,7 @@
   <ItemGroup>
     <None Include="xunit.runner.json" CopyToOutputDirectory="PreserveNewest" />
     <Compile Include="$(SharedSourceRoot)test\SkipOnHelixAttribute.cs" />
+    <Compile Remove="TestFiles\TagHelpersInCodeBlocksAnalyzerTest\*.*" />
```
pranavkm: ?

src/Mvc/Mvc.Analyzers/test/TagHelpersInCodeBlocksAnalyzerTest.cs
```diff
@@ -0,0 +1,93 @@
...
+        private async Task VerifyDefault(TestSource testSource)
+        {
+            // Arrange
+            var expectedLocation = testSource.DefaultMarkerLocation;
+
+            // Act
+            var result = await Executor.GetDiagnosticsAsync(testSource.Source);
+
+            // Assert
+            // We want to ignore C# diagnostics because they'll include diagnostics for not awaiting bits correctly. The purpose of this analyzer is to
```
pranavkm: 👍

pranavkm (approved):

src/Mvc/Mvc.Analyzers/test/Mvc.Analyzers.Test.csproj
NTaylorMullen: The C# code in the file is invalid, without this the project attempts to compile the test files which are purposely non-compileable.

src/Mvc/Mvc.Analyzers/src/DiagnosticDescriptors.cs
NTaylorMullen: L:ooks like C# uses `methods`, will change.

NTaylorMullen:
> Looks like builds above went kaboom:

Looks like C# 8 nullability enforcement was enabled after I started my changes 😄. That's what I get for last minute rebasing -> pushing

NTaylorMullen:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=207076

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/2521
https://github.com/aspnet/AspNetCore-Internal/issues/1989

PR #10319, opened 2019-05-17
Improved Context pooling

### LINKED ISSUE #10556: HostingRequestStartingLog shouldn't capture HttpRequest for its values

### LINKED ISSUE #10557: HostingRequestFinishedLog shouldn't capture HttpContext for its values

### LINKED ISSUE #11069: [Discussion] NullReferenceException at Microsoft.AspNetCore.Http.Internal.DefaultHttpRequest.get_Query()
This topic came from several other:
#2806
#3219
#3219

They all a bit different, but the reason (I think) the same: concurent access to HttpContext from different threads.

Our story is pretty similar: we have project on asp.net 1.1, and we widely using HttpContext and HttpContextContextAccessor for different things (mainly for custom authorization and access to Session info), and I'm not sure if all what we're doing is really safe.

My first question is: there was a several mentions in issues above that middleware is kind of safe place, but what about ActionFilters? 
We have a pretty easy one like this:
```
public class PartialViewFilter : IActionFilter
    {
        public void OnActionExecuted(ActionExecutedContext context)
        {
            
        }

        public void OnActionExecuting(ActionExecutingContext context)
        {
            context.HttpContext.Items["_partial"] = context.HttpContext.Request.Query.Keys.Contains("partial");
        }
    }
```
and catches NullReferenceException sometimes (not very often, less than 5% requests aprox).
Exception:

> System.NullReferenceException: Object reference not set to an instance of an object.
> at Microsoft.AspNetCore.Http.Internal.DefaultHttpRequest.get_Query()
> at ****.PartialViewFilter.OnActionExecuting
> …

The most interesting part is that problem start to be occurs only when we moved from 1.0 to 1.1 (no other changes at all). 

And now my second question: is any Best Practices how to avoid this? Or at least how to detect such things?

And the last one; what about to generate more accurate error? NullReferenceException is not really nice.

benaadams:
3 commits so the changes are easier to follow

1. More generalized `DefaultHttpContext` pooling (move creation+reset to Hosting rather than Kestrel) https://github.com/aspnet/AspNetCore/pull/10319/commits/c617b8d0a16746b50058e01730684c01b8d77bd2
2. Fix HttpContext use after request end https://github.com/aspnet/AspNetCore/pull/10319/commits/b52f3bee85181438188f7e79572af41ac192da3d
    * Captured in a couple places rather than its properties captured; (1.) now causes this to explicitly fail.
    * Throw ObjectDisposed exception when this happens (added in 1.) added test for it here.
3. Reduce HostContext costs/pool (struct wrapper over object rather than large struct; pool the object) https://github.com/aspnet/AspNetCore/pull/10319/commits/2f903d41ab051bd1aa4e315a7bc6316dcfe49be4

/cc @davidfowl @halter73 @Tratcher this is purely scoped to the changes to contexts that I wanted to make - can follow up with the improved log https://github.com/aspnet/AspNetCore/pull/10102

Resolves #10556 
Resolves #10557
Resolves https://github.com/aspnet/AspNetCore/issues/11069


benaadams:
(3.) Is for shaving 32 bytes off the size of the `ProcessRequests` statemachine
```diff
 ; Assembly listing for method HttpProtocol:ProcessRequests(ref):ref:this
 ;
 ;  V00 this         [V00,T00] (  3,  3   )     ref  ->  rcx         this class-hnd
 ;  V01 arg1         [V01,T01] (  3,  3   )     ref  ->  rdx         class-hnd
-;  V02 loc0         [V02    ] (  7,  6.25)  struct (136) [rsp+0x20]   do-not-enreg[XSFB] must-init addr-exposed ld-addr-op
+;  V02 loc0         [V02    ] (  7,  6.25)  struct (104) [rsp+0x20]   do-not-enreg[XSFB] must-init addr-exposed ld-addr-op
 ;* V03 loc1         [V03    ] (  0,  0   )  struct ( 8) zero-ref    ld-addr-op
 ;  V04 OutArgs      [V04    ] (  1,  1   )  lclBlk (32) [rsp+0x00]   "OutgoingArgSpace"
 ;* V05 tmp1         [V05    ] (  0,  0   )  struct ( 8) zero-ref    ld-addr-op "Inline ldloca(s) first use temp"
 ;  V06 tmp2         [V06,T02] (  2,  4   )     ref  ->  rax         class-hnd "dup spill"
 ;  V07 tmp3         [V07,T03] (  4,  3.25)     ref  ->  rax        
 ;* V08 tmp4         [V08    ] (  0,  0   )     ref  ->  zero-ref    V03.m_task(offs=0x00) P-INDEP "field V03.m_task (fldOffset=0x0)"
 ;* V09 tmp5         [V09    ] (  0,  0   )     ref  ->  zero-ref    V05.m_task(offs=0x00) P-INDEP "field V05.m_task (fldOffset=0x0)"
 ;
 ; Lcl frame size = 168
 
 G_M39292_IG01:
        57                   push     rdi
        56                   push     rsi
-       4881ECA8000000       sub      rsp, 168
+       4881EC88000000       sub      rsp, 136
        488BF1               mov      rsi, rcx
        488D7C2420           lea      rdi, [rsp+20H]
-       B922000000           mov      ecx, 34
+       B91A000000           mov      ecx, 26
        33C0                 xor      rax, rax
        F3AB                 rep stosd 
```
And saving a bunch of stack copies from it
image

and

image

src/Http/Http.Features/src/FeatureReferences.cs
```diff
@@ -108,5 +109,16 @@ public void Initalize(IFeatureCollection collection, int revision)
...
+
+        private static int ContextDisposed()
+        {
+            ThrowContextDisposed();
+            return 0;
+        }
+
+        private static void ThrowContextDisposed()
+        {
+            throw new ObjectDisposedException(nameof(HttpContent), $"Request has finished and {nameof(HttpContent)} disposed.");
```
benaadams:
```suggestion
            throw new ObjectDisposedException(nameof(HttpContext), $"Request has finished and {nameof(HttpContext)} disposed.");
```

src/Http/Http/src/DefaultHttpContext.cs
```diff
@@ -162,6 +163,16 @@ public override void Abort()
...
         }
+        private static IFeatureCollection ContextDisposed()
+        {
+            ThrowContextDisposed();
+            return null;
+        }
+
+        private static void ThrowContextDisposed()
+        {
+            throw new ObjectDisposedException(nameof(HttpContent), $"Request has finished and {nameof(HttpContent)} disposed.");
```
benaadams:
```suggestion
            throw new ObjectDisposedException(nameof(HttpContext), $"Request has finished and {nameof(HttpContext)} disposed.");
```

benaadams:
Rebased

benaadams:
Rebased

benaadams:
Rebased

src/Hosting/Hosting/src/Internal/HostingApplication.cs
```diff
@@ -53,16 +65,87 @@ public void DisposeContext(Context context, Exception exception)
...
-            public bool EventLogEnabled { get; set; }
-            public Activity Activity { get; set; }
-            internal bool HasDiagnosticListener { get; set; }
+            // Single field wrapper over an object to keep pass-by-value sematics in a single pointer register
+            // rather than involving stack copies.
+            private InnerContext _context;
+
+            internal static Context Create()
+            {
+                return new Context() { _context = new InnerContext() };
```
halter73: Is this only only place a new Context is created? If so, why have the Initialize method?

halter73:
Does this fix both #10556 and #10557?

benaadams:
> Does this fix both #10556 and #10557?

Yes; though it doesn't change/improve the output for end logging as per https://github.com/aspnet/AspNetCore/pull/10102

src/Hosting/Hosting/src/Internal/HostingApplication.cs
benaadams: If `IHostContextContainer<Context>` is implemented it never calls `Create()` and only calls `Initialize()`

benaadams:
Rebased

src/Hosting/Hosting/src/Internal/HostingApplication.cs
benaadams: Mainly I thought having `Context.Create()` was less strange then `deafult(Context).Initialize()`

Suppose could do `(new Context()).Initialize()` but again seems messy?

halter73: I see. I missed that Http1Connection/Http2Stream aren't calling Create when implementing `IHostContextContainer<Context>`. I think it's fine as is then.

halter73 (approved):

halter73 (approved):

benaadams:
Am not sure what the test timeouts are; is that CI being dodgy or something from this change?

benaadams:
Hmm...

benaadams:
Hmm, is seems to jam around
```
  [xUnit.net 00:00:04.34]     Microsoft.AspNetCore.Blazor.Build.Test.ComponentRenderingRazorIntegrationTest.Render_HtmlBlock_Integration [SKIP]
  Skipped  Microsoft.AspNetCore.Blazor.Build.Test.ComponentRenderingRazorIntegrationTest.Render_HtmlBlock_Integration
  [xUnit.net 00:00:04.48]     Microsoft.AspNetCore.Blazor.Build.Test.ComponentRenderingRazorIntegrationTest.Render_Component_HtmlEncoded [SKIP]
  Skipped  Microsoft.AspNetCore.Blazor.Build.Test.ComponentRenderingRazorIntegrationTest.Render_Component_HtmlEncoded
```

benaadams:
Interesting

benaadams:
Need to track down why its being weird; even when I switch off the actual pooling

jkotalik:
Yeah, it definitely seems to be something in this PR.

benaadams:
ClientHandlerTests we using same `Internal.HostingApplication.Context` but not initalizing it correctly

benaadams:
Looks to have fixed it

benaadams:
Rebased

benaadams:
Raised issue for Helix issue https://github.com/aspnet/AspNetCore/issues/11038

benaadams:
Rebased instead :P

benaadams:
Rebased

benaadams:
> AspNetCore-ci (Test: Windows Server 2016 x64)
```
[plugin]: Error during loading "F:\workspace\_work\1\s\src\SignalR\clients\ts\FunctionalTests\node_modules/karma-sauce-launcher" plugin:
    Cannot find module '../build/safe-execute'
INFO [karma-server]: Writing browser console to file: F:\workspace\_work\1\s\artifacts\log\browserlogs.console.log
INFO [karma-server]: Karma v3.1.4 server started at http://0.0.0.0:9876/
INFO [launcher]: Launching browsers ChromeHeadlessNoSandbox with concurrency unlimited
ERROR [karma-server]: Error: Found 1 load error
      at Server.webServer.listen (F:\workspace\_work\1\s\src\SignalR\clients\ts\FunctionalTests\node_modules\karma\lib\server.js:182:27)
      at Object.onceWrapper (events.js:286:20)
      at Server.emit (events.js:203:15)
      at Server.EventEmitter.emit (domain.js:448:20)
      at emitListeningNT (net.js:1313:10)
      at process._tickCallback (internal/process/next_tick.js:63:19)
F:\workspace\_work\1\s\src\SignalR\clients\ts\FunctionalTests\SignalR.Npm.FunctionalTests.npmproj(22,5): error : Command failed with exit code 1.
```
Could someone restart just this job (I can always rebase to restart everything :P)

davidfowl:
Ben I’m cold on this change, tell me why I should want it

benaadams:
> Ben I’m cold on this change, tell me why I should want it

1. Moves creating and reseting pooled `HttpContext`s from Kestrel to `HttpContextFactory`; since 
   i. logically that owns the lifetime rather than the server and 
   ii. other servers can then also make use of it without copypasta (other than implementing 1 vanilla interface property)
2. Throws `ObjectDisposedException`s rather than `NullReferenceException`s when features are accessed between requests 
3. No longer captures `HttpContext` into the log objects and uses properties off that when doing logging (e.g. ConsoleLogger queues will it be reporting on the same context that the log was created for when it actually logs it?)
4. Having the host `Context` as a struct both prevents a per request allocation and devitualizes the generics; however its large at 36 bytes; which is then copied both in and through the parameters of 5 methods prerequest. This keeps it as the struct for devirtualized access, but shrinks it to passing by register. To do this it then needs to be backed by a class; which is then pooled to prevent pre-request allocation.

The test changes are because the tests were capturing the HttpContext and then inspecting post-request completion, which now throws `ObjectDisposedException`s which it didn't before because Kestrel was handling the disposal but wasn't involved in the tests; whereas hosting now manages the Disposal and is involved in the tests; so they need to capture the values they want to inspect prior to the end of the "request".

benaadams:
Also then I can follow up with https://github.com/aspnet/AspNetCore/pull/10102 which increases the host Context by another pointer to 44 bytes without feeling bad that I'm hitting performance doing so.

davidfowl:
> Moves creating and reseting pooled HttpContexts from Kestrel to HttpContextFactory; since
i. logically that owns the lifetime rather than the server and
ii. other servers can then also make use of it without copypasta (other than implementing 1 vanilla interface property)

By introducing a pretty obscure interface (that really needs to go into the public namespace) `IHostContextContainer<T>` which forces more generic types into the servers (but I guess we have that already with `IHttpApplication<T>`).

> Throws ObjectDisposedExceptions rather than NullReferenceExceptions when features are accessed between requests

That piece can be separate from this PR.

> No longer captures HttpContext into the log objects and uses properties off that when doing logging (e.g. ConsoleLogger queues will it be reporting on the same context that the log was created for when it actually logs it?)

The console logger stringifies before queuing to the background thread. That issue you filed is not a problem in practice. https://github.com/aspnet/Extensions/blob/531db4e313e577192cdbca3931c4298b6db1e611/src/Logging/Logging.Console/src/ConsoleLogger.cs#L59-L64

>

benaadams:
> The console logger stringifies before queuing to the background thread. That issue you filed is not a problem in practice. 

Ok, maybe not the best example, but loggers that use the `IReadOnlyList<KeyValuePair<string, object>>` interface to extract the properties from `HostingRequestStartingLog` and `HostingRequestFinishedLog` **must** currently look up all those properties inline with the request because they are not immutable as they lookup on a HttpContext with connection lifetime; rather than a per-request lifetime. 

So either the per-request `HostingRequestStartingLog` and `HostingRequestFinishedLog` should resolve their properties during the request so they become inert (as per this PR) or they also might as well be pooled.

davidfowl:
> Ok, maybe not the best example, but loggers that use the IReadOnlyList<KeyValuePair<string, object>> interface to extract the properties from HostingRequestStartingLog and HostingRequestFinishedLog must currently look up all those properties inline with the request because they are not immutable as they lookup on a HttpContext with connection lifetime; rather than a per-request lifetime.

They should be doing that anyways, they don't own the object lifetime. Not saying we don't have a potential problem with poorly written loggers but this fix conflates like 4 unrelated things IMO.

benaadams:
>  but this fix conflates like 4 unrelated things IMO.

K, so you'd want them as separate PRs?

davidfowl:
Yes please.

benaadams:
Split into https://github.com/aspnet/AspNetCore/pull/11721, https://github.com/aspnet/AspNetCore/pull/11722 and https://github.com/aspnet/AspNetCore/pull/11723

PR #10321, opened 2019-05-17
Implement new bedrock listener abstraction and re-plat Kestrel on top

### LINKED ISSUE #4752: Consider letting the transport own the creation of IDuplexPipe
Today the transport is responsible for exposing properties on the connection that can be used by Kestrel to create a duplex pipe. It does this today for a few reasons:

- It wants to directly apply the MaxResponseBufferSize and MaxRequestBuffer size to pipe options
- It communicates "fin" via cancelling the transport's output reader (this is super hacky and we should get rid of it anyways).
- It wants to control where "user code" runs. This includes:
  - ConnectionAdapters (soon to be deprecated)
  - CancellationToken callbacks
  - the request processing delegate

Delegating to the transport has some other benefits:
- Transports aren't tied to a specific pipe implementation.
- We can potentially delete ITransportSchedulerFeature, IApplicationTransportFeature 

To make this change though, we need to pass the relevant settings from kestrel to the transport without coupling. My current though is to pass a settings object to ITransportFactory.Create.

See https://github.com/geoffkizer/KestrelHttpServer/blob/7bd21cdb1581fc3559479842262a24deecf171bb/src/Kestrel.Transport.Sockets/Internal/SocketConnection.cs for an example of the socket transport with a direct pipe implementation.

/cc @tmds @benaadams (only other people that wrote transports 😄)

### LINKED ISSUE #10308: Client/server networking abstractions
Epic https://github.com/aspnet/AspNetCore/issues/10869

The goal of this issue is to specify abstractions which the .NET community can use to build robust, correct, high-performance client & server applications. This is a follow-up to #4772 (Project Bedrock) and is largely a summary of a discussion with @davidfowl, @anurse, @halter73, @shirhatti, & @jkotalik.

## Initiating & accepting connections

Clients initiate connections to an endpoint using `IConnectionFactory`. Servers bind to an endpoint using `IConnectionListenerFactory` and then accept incoming connections using the `IConnectionListener` returned at binding time.

``` C#
public abstract class ConnectionContext : IAsyncDisposable
{
  // unchanged except for IAsyncDisposable base
}

public interface IConnectionListenerFactory
{
    ValueTask<IConnectionListener> BindAsync(System.Net.EndPoint endpoint, CancellationToken cancellationToken = default);
}

public interface IConnectionListener : IAsyncDisposable
{
    EndPoint EndPoint { get; }
    ValueTask<ConnectionContext> AcceptAsync(CancellationToken cancellationToken = default);
    ValueTask UnbindAsync(CancellationToken cancellationToken = default);
}

public interface IConnectionFactory
{
    ValueTask<ConnectionContext> ConnectAsync(System.Net.EndPoint endpoint, CancellationToken cancellationToken = default);
}
```

A framework might want to support multiple different transports, such as file + socket. In that case, it can use multiple `IConnectionListenerFactory` implementations.

The shared abstractions should *not* dictate how these factories are are specified or configured: frameworks should continue to follow a framework-specific builder pattern. Eg, `KestrelServerOptions` has `List<ListenOptions>` where each entry is a different endpoint to listen on + middleware to handle connections to that endpoint.

Both `IConnectionListenerFactory.BindAsync` & `IConnectionFactory.ConnectAsync` accept a `System.Net.EndPoint` to bind/connect to. The BCL has `EndPoint` implementations for `IPEndPoint`, `DnsEndPoint`, and `UnixDomainSocketEndPoint`. Users can add new transports which accept custom sub-classes of `EndPoint`, eg `MyOverlayNetworkEndPoint`.

Servers bind to endpoints using `IConnectionListenerFactory`, resulting in a `IConnectionListener`. The listener's `AcceptAsync()` is then called in a loop, each time resulting in a `ConnectionContext` instance which represent a newly accepted connection. When the server wants to stop accepting new connections, the listener is disposed by calling `listener.DisposeAsync()`. Connections which were previously accepted continue to operate until each is individually terminated.

For implementations of `IConnectionListener` which are backed by a callback system (eg, libuv), decoupling the lifetime of `IConnectionListener` and the (many) resulting `ConnectionContext`s poses a challenge. We need to consider whether or not this minimal interface allows graceful shutdown (including draining connections) to still be cleanly implemented in those cases.

## Handling connections

The client/server decides how it handles the `ConnectionContext`. Typically it will be decorated (eg, via `ConnectionDelegate`) and pumped until the connection is terminated. Connections can be terminated by calling `DisposeAsync()` (from `IAsyncDisposable`) but can also be terminated out-of-band (eg, because the remote endpoint terminated it). In either case, clients/servers can subscribe to the `ConnectionContext`'s `IConnectionLifetimeFeature` feature to learn about termination and take necessary action (eg, cleanup).

## Connection metadata

Currently, `IHttpConnectionFeature` is used to expose local/remote endpoint information. We will add a new feature, `IConnectionEndPointFeature` as a uniform way to access this information.

``` C#
public interface IConnectionEndPointFeature
{
  EndPoint RemoteEndPoint { get; set; }
  EndPoint LocalEndPoint { get; set; }
}
```

davidfowl:
This is a massive set of changes to Kestrel to remove the existing pubternal transport layer and implement a public facing API for listeners and clients, see the details here https://github.com/aspnet/AspNetCore/issues/10308.

This change only has the server side pieces of the story as I don't want to add the client APIs without having ported SignalR to use them. Here are the highlights:

- Transport.Abstractions is empty (will be removed in a separate PR as it requires removing it from a *ton* of places)
- `TransportConnection` has been moved to Connection.Abstractions (we can decide if we need to consolidate with `DefaultConnectionContext` in a later PR)
- Added `FileHandleEndPoint` which allows binding to a file handle (could be a pipe or tcp handle)
- ListenOptions has been gutted for most pubternal API and returns various types of binding information . The source of truth is the `EndPoint` instance.
- Cleaned up a bunch of libuv tests decoupling them from Kestrel.Core

## Breaking Changes
- Removing pubternal API is itself a breaking change but one that we already planned to do.
- We've removed the ability to set the scheduling mode on Kestrel, this will be a transport knob instead (https://github.com/aspnet/AspNetCore/pull/10321/commits/77d3c27c468d69e2bda0195b2a64d68a921bf783) cc @sebastienros (I know our benchmarks support this)
- DisposeAsync was added to ConnectionContext (it will implement IAsyncDisposable in a later change, it supports netstandard 2.0 so that requires a bit more work) cc @anurse @BrennanConroy as this affects SignalR (https://github.com/aspnet/AspNetCore/pull/10321/commits/4977f1b5f237a84a85ca0e0d4cd4eb7f85cd3e51)
- NoDelay is a noop, on `ListenOptions`, we should consider deprecating it. This has been moved to each of the transports. One major difference though is that it's no longer localized per endpoint but is global. We'd need a derived `EndPoint` type (or maybe extend `IPEndPoint`) to store both the socket options and the binding information.

@halter73 this is ready for review now.

Fixes #10308 
Fixes #4752


src/Servers/Kestrel/Transport.Libuv/src/Internal/LibuvConnectionListener.cs
```diff
@@ -119,9 +123,39 @@ public async Task BindAsync()
...
+                var elem = slots[i];
+                if (elem == null)
+                {
+                    slots[i] = AcceptAsync(_listeners[i], i);
+                }
+            }
+
+            while (true)
+            {
+                (LibuvConnection connection, int slot) = await await Task.WhenAny(slots);
```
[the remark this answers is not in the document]
davidfowl: But the double await look so cool.

src/Servers/Kestrel/Core/src/ListenOptions.cs
```diff
@@ -90,6 +90,8 @@ public FileHandleType HandleType
             }
         }
 
+        public EndPoint Endpoint => IPEndPoint;
```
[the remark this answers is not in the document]
davidfowl: There are more breaks that will need to be made, I'm still feeling it out.

src/Servers/Connections.Abstractions/src/IConnectionListener.cs
```diff
@@ -0,0 +1,14 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Connections
+{
+    public interface IConnectionListener
+    {
+        ValueTask<ConnectionContext> AcceptAsync();
+
+        ValueTask DisposeAsync();
```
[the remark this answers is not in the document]
davidfowl: Yes I just hacked it because this is ns2.0.

src/Servers/Kestrel/Core/src/Internal/ConnectionDispatcher.cs
```diff
@@ -28,23 +28,45 @@ public ConnectionDispatcher(ServiceContext serviceContext, ConnectionDelegate co
...
+                    while (true)
+                    {
+                        var connection = await listener.AcceptAsync();
 
-            // This *must* be set before returning from OnConnection
-            connection.Application = pair.Application;
+                        // TODO: We need a bit of command and control to do connection management and that requires
+                        // that we have access to a couple of methods that only exists on TransportConnection
+                        // (specifically RequestClose, TickHeartbeat and CompleteAsync. This dependency needs to be removed or
+                        // the return value of AcceptAsync needs to change.
```
ReubenBond: Changing the return value & adding a `CancellationToken` to `AcceptAsync` works for me. In-band cancellation is ideal (i.e, cancellation via return value, not exception).

src/Servers/Connections.Abstractions/src/ConnectionContext.cs
```diff
@@ -18,6 +20,9 @@ public abstract class ConnectionContext
 
         public abstract IDuplexPipe Transport { get; set; }
 
+        public abstract EndPoint LocalEndpoint { get; set; }
+        public abstract EndPoint RemoteEndpoint { get; set; }
+
```
ReubenBond: If we have the Endpoints here (lgtm) then it seems we don't need `IConnectionEndPointFeature` - should we axe it?

Tratcher: The features are still useful upstack where ConnectionContext isn't available.

davidfowl: Generally, the ConnectionContext isn't the source of truth, features are and the implementation should be using the feature. This enables decoration of features by middleware without having to make a derived ConnectionContext

src/Servers/Kestrel/Transport.Libuv/src/Internal/LibuvConnectionListener.cs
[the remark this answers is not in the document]
gfoidl: Agree 😄

src/Servers/Connections.Abstractions/src/IConnectionListener.cs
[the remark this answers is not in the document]
gfoidl: Ahh...thx.

src/Servers/Kestrel/Transport.Libuv/src/Internal/ListenerContext.cs
```diff
@@ -3,39 +3,52 @@
...
         /// <summary>
         /// Creates a socket which can be used to accept an incoming connection.
         /// </summary>
         protected UvStreamHandle CreateAcceptSocket()
         {
-            switch (EndPointInformation.Type)
+            switch (EndPoint)
             {
-                case ListenType.IPEndPoint:
+                case IPEndPoint ip:
```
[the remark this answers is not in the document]
DAllanCarr: Discards are ugly!

src/Servers/Kestrel/Core/src/Internal/ConnectionDispatcher.cs
davidfowl: Then we can make null signify no more connections accept. I have another idea to try for the return value

src/Servers/Kestrel/Transport.Sockets/src/Internal/SocketConnection.cs
```diff
@@ -101,8 +121,22 @@ public override void Abort(ConnectionAbortedException abortReason)
...
 
         // Only called after connection middleware is complete which means the ConnectionClosed token has fired.
-        public void Dispose()
+        public override async ValueTask DisposeAsync()
         {
+            if (_task != null)
+            {
+                // TODO: Make this timeout configurable, this gives the task time to flush the data to the socket
+                // before timing out
+                var result = await Task.WhenAny(_task, Task.Delay(5000));
```
davidfowl: This doesn't *have* to timeout, the previous logic just disposed and assumed timeout logic would be handled by the upper application layers. It could just await the task without aborting.

src/Servers/Kestrel/Transport.Sockets/src/SocketConnectionListener.cs
```diff
@@ -0,0 +1,244 @@
...
+            if (listenSocket != null)
+            {
+                // Unbind the listen socket, so no new connections can come in
+                listenSocket.Dispose();
+
+                // Wait for all pending accepts to drain
+                var spin = new SpinWait();
+                while (_pendingAccepts > 0)
+                {
+                    spin.SpinOnce();
```
davidfowl: This could Task.Delay instead but it should be very fast.

src/Servers/Kestrel/Core/src/Internal/Infrastructure/KestrelConnection.cs
```diff
@@ -1,25 +1,158 @@
...
 using System.Threading.Tasks;
-using Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.Internal;
+using Microsoft.AspNetCore.Connections;
+using Microsoft.AspNetCore.Connections.Features;
+using Microsoft.Extensions.Logging;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure
 {
-    internal class KestrelConnection
+    internal class KestrelConnection : IConnectionHeartbeatFeature, IConnectionCompleteFeature, IConnectionLifetimeNotificationFeature
```
davidfowl: @halter73 check this out. This basically moves the logic that Kestrel needs to trigger into kestrel itself instead of being in the transport layer.

src/Servers/Kestrel/Core/test/ConnectionDispatcherTests.cs
```diff
@@ -50,28 +51,6 @@ public void OnConnectionCreatesLogScopeWithConnectionId()
             Assert.True(((TestKestrelTrace)serviceContext.Log).Logger.Scopes.IsEmpty);
         }
 
-        [Fact]
-        public async Task OnConnectionCompletesTransportPipesAfterReturning()
```
davidfowl: The transport does this now.

src/SignalR/clients/csharp/Http.Connections.Client/src/HttpConnection.cs
```diff
@@ -234,7 +234,7 @@ private async Task StartAsyncCore(TransferFormat transferFormat)
         /// A connection cannot be restarted after it has stopped. To restart a connection
         /// a new instance should be created using the same options.
         /// </remarks>
-        public async Task DisposeAsync()
+        public override async ValueTask DisposeAsync()
```
davidfowl: cc @BrennanConroy @anurse

davidfowl:
AHHHHH

src/Servers/Kestrel/Core/src/Internal/Http2/Http2OutputProducer.cs
```diff
@@ -363,7 +362,7 @@ private static Pipe CreateDataPipe(MemoryPool<byte> pool)
                 pauseWriterThreshold: 1,
                 resumeWriterThreshold: 1,
                 useSynchronizationContext: false,
-                minimumSegmentSize: KestrelMemoryPool.MinimumSegmentSize
+                minimumSegmentSize: pool.GetMinimumAllocSize()
```
davidfowl: This should be GetMinimumSegmentSize

src/Servers/Kestrel/Core/src/Internal/Http2/Http2Stream.cs
```diff
@@ -494,7 +493,7 @@ private Pipe CreateRequestBodyPipe(uint windowSize)
                 pauseWriterThreshold: windowSize + 1,
                 resumeWriterThreshold: windowSize + 1,
                 useSynchronizationContext: false,
-                minimumSegmentSize: KestrelMemoryPool.MinimumSegmentSize
+                minimumSegmentSize: _context.MemoryPool.GetMinimumAllocSize()
```
davidfowl: This should be GetMinimumSegmentSize()

src/Servers/Kestrel/Core/test/AddressBinderTests.cs
```diff
@@ -65,21 +65,20 @@ public void ParseAddressDefaultsToAnyIPOnInvalidIPAddress(string host)
...
             Assert.IsType<LocalhostListenOptions>(listenOptions);
-            Assert.Equal(ListenType.IPEndPoint, listenOptions.Type);
+            Assert.IsType<IPEndPoint>(listenOptions.EndPoint);
             Assert.Equal(IPAddress.Loopback, listenOptions.IPEndPoint.Address);
             Assert.Equal(80, listenOptions.IPEndPoint.Port);
             Assert.False(https);
         }
 
-        [Fact]
+        [OSSkipCondition(OperatingSystems.Windows, WindowsVersions.Win7, WindowsVersions.Win8, WindowsVersions.Win81, WindowsVersions.Win2008R2, SkipReason = "UnixDomainSocketEndPoint is not supported on older versions of Windows")]
```
davidfowl: This is breaking change as well, you can no longer construct a `UnixDomainSocketEndPoint` on an OS that doesn't support it.

src/Servers/Kestrel/Core/src/Internal/Infrastructure/KestrelConnection.cs
```diff
@@ -1,25 +1,168 @@
...
+                    _heartbeatHandlers = new List<(Action<object> handler, object state)>();
+                }
+
+                _heartbeatHandlers.Add((action, state));
+            }
+        }
+
+        void IConnectionCompleteFeature.OnCompleted(Func<object, Task> callback, object state)
+        {
+            if (_completed)
```
jkotalik: Don't we have this same pattern with Request OnCompleted? Can we simplify/consolidate this logic somehow?

src/Servers/Kestrel/Core/src/Internal/Infrastructure/KestrelConnection.cs
```diff
@@ -1,25 +1,168 @@
...
+
+        public void RequestClose()
+        {
+            try
+            {
+                _connectionClosingCts.Cancel();
+            }
+            catch (ObjectDisposedException)
+            {
+                // There's a race where the token could be disposed
```
jkotalik: Why/Where is there a race?

src/Servers/Kestrel/Core/src/Internal/MemoryPoolExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+                return 4096;
+            }
+
+            return Math.Min(4096, pool.MaxBufferSize);
+        }
+
+        public static int GetMinimumAllocSize(this MemoryPool<byte> pool)
+        {
+            // 1/2 of a segment
+            return pool.GetMinimumSegmentSize() / 2;
```
jkotalik: Wasn't this 1/4th before? Why change it?

src/Servers/Kestrel/Core/src/KestrelServer.cs
```diff
@@ -79,27 +77,11 @@ private static ServiceContext CreateServiceContext(IOptions<KestrelServerOptions
...
-                default:
-                    throw new NotSupportedException(CoreStrings.FormatUnknownTransportMode(serverOptions.ApplicationSchedulingMode));
-            }
-
             return new ServiceContext
             {
                 Log = trace,
                 HttpParser = new HttpParser<Http1ParsingHandler>(trace.IsEnabled(LogLevel.Information)),
-                Scheduler = scheduler,
+                Scheduler = PipeScheduler.ThreadPool,
```
jkotalik: You said:
> We've removed the ability to set the scheduling mode on Kestrel, this will be a transport knob instead 

Where is this set?

src/Servers/Kestrel/Core/src/KestrelServer.cs
```diff
@@ -182,8 +167,10 @@ public async Task StopAsync(CancellationToken cancellationToken)
                 var tasks = new Task[_transports.Count];
                 for (int i = 0; i < _transports.Count; i++)
                 {
-                    tasks[i] = _transports[i].UnbindAsync();
+                    (IConnectionListener listener, Task acceptLoop) = _transports[i];
+                    tasks[i] = Task.WhenAll(listener.UnbindAsync(cancellationToken).AsTask(), acceptLoop);
```
jkotalik: I'm not a fan of having to do Task.WhenAll here between unbinding and the accept loop. Can you clarify why you need to wait on both? How would someone else using these interfaces handle this?

src/Servers/Kestrel/Core/test/AddressBinderTests.cs
jkotalik: This is a rare case, correct?

src/Servers/Kestrel/Transport.Abstractions/ref/Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.netcoreapp3.0.cs
```diff
@@ -1,113 +1,3 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.Internal
```
jkotalik: I didn't realize you already deleted all of this already. How much more work is it to remove it entirely?

src/Servers/Kestrel/Transport.Libuv/src/Internal/LibuvConnection.cs
```diff
@@ -31,31 +30,54 @@ internal partial class LibuvConnection : TransportConnection, IDisposable
...
+
+        public PipeReader Output => Application.Input;
+
         public LibuvOutputConsumer OutputConsumer { get; set; }
         private ILibuvTrace Log { get; }
         private LibuvThread Thread { get; }
         public override MemoryPool<byte> MemoryPool => Thread.MemoryPool;
-        public override PipeScheduler InputWriterScheduler => Thread;
-        public override PipeScheduler OutputReaderScheduler => Thread;
+        public void Start()
```
jkotalik: nit space.

jkotalik (commented):
Got through half of it. Will add some more feedback on the transports later.

src/Servers/Kestrel/Core/src/Internal/ConnectionDispatcher.cs
```diff
@@ -28,26 +26,46 @@ public ConnectionDispatcher(ServiceContext serviceContext, ConnectionDelegate co
...
-            var outputOptions = GetOutputPipeOptions(_serviceContext, connection.MemoryPool, connection.OutputReaderScheduler);
+            ThreadPool.UnsafeQueueUserWorkItem(StartAcceptingConnectionsCore, listener, preferLocal: false);
+            return _acceptLoopTcs.Task;
+        }
 
-            var pair = DuplexPipe.CreateConnectionPair(inputOptions, outputOptions);
+        private void StartAcceptingConnectionsCore(IConnectionListener listener)
+        {
+            // REVIEW: Multiple accept loops in parallel?
+            _ = AcceptConnectionsAsync();
```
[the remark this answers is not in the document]
davidfowl: Throughput. It’s not explicitly required unless we see a connection ramp up throughput problem

src/Servers/Kestrel/Core/src/Internal/ConnectionDispatcher.cs
```diff
@@ -28,26 +26,46 @@ public ConnectionDispatcher(ServiceContext serviceContext, ConnectionDelegate co
...
+                            break;
+                        }
 
-            return Execute(new KestrelConnection(connection));
+                        _ = Execute(new KestrelConnection(connection, _serviceContext.Log));
+                    }
+                }
+                catch (Exception)
+                {
+                    // REVIEW: Today the only way to end the accept loop is an exception
```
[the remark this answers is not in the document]
davidfowl: Yes I forgot to log here and update the comment

src/Servers/Kestrel/Core/src/Internal/Infrastructure/KestrelConnection.cs
davidfowl: This isn’t new code. It’s just moved, and it’s the same logic. If you have comments on what you’d like to simplify, it can be applied to both patterns in a later commit

src/Servers/Kestrel/Core/src/Internal/MemoryPoolExtensions.cs
davidfowl: Because it’s 1/2 at the transport level I don’t know why it was 1/4 here. I made them all the same

src/Servers/Kestrel/Core/src/KestrelServer.cs
davidfowl: It’s not exposed in this PR. I’m not convinced we should yet.

src/Servers/Kestrel/Core/src/KestrelServer.cs
davidfowl: To dispose, we need to first unbind so no new connections can enter the accept loop, then we need to await the accept loop so that it drains existing connections before disposing.

src/Servers/Kestrel/Transport.Abstractions/ref/Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.netcoreapp3.0.cs
davidfowl: I was going to remove it but it’s part of every solution because we have all of the dependencies there. I can do it less properly and just remove it from this solution (my guess is that nobody else did this yet?)

src/Servers/Kestrel/Core/src/Internal/Infrastructure/KestrelConnection.cs
davidfowl: Originally it was because the token was disposed in a different place than it was consumed. We can do this without the catch now.

src/Servers/Kestrel/Core/test/AddressBinderTests.cs
davidfowl: Correct, it would have failed later anyways.

src/Servers/Connections.Abstractions/src/Features/IConnectionEndpointFeature.cs
```diff
@@ -0,0 +1,10 @@
+using System.Net;
```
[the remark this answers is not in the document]
davidfowl: Yea I always forget these

src/Servers/Kestrel/Core/src/Internal/ConnectionDispatcher.cs
```diff
@@ -28,26 +26,46 @@ public ConnectionDispatcher(ServiceContext serviceContext, ConnectionDelegate co
...
-            // Set the transport and connection id
-            connection.ConnectionId = CorrelationIdGenerator.GetNextId();
-            connection.Transport = pair.Transport;
+            async Task AcceptConnectionsAsync()
+            {
+                try
+                {
+                    while (true)
+                    {
+                        var connection = await listener.AcceptAsync();
```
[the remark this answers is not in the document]
davidfowl: What’s being boxed? These are all reference types.

src/Servers/Kestrel/Core/src/Internal/Infrastructure/KestrelConnection.cs
davidfowl: I take that back, the race is between shutting down and the connection ending. A call to RequestClose can be made “at the same time” a connection is being disposed.

anurse:
This feels like it would be good to get in for preview 6. What kind of e2e verification do we need here? We want to land everything in a candidate build by EOD Friday.

davidfowl:
> This feels like it would be good to get in for preview 6 because it's breaking. What kind of e2e verification do we need here? We want to land everything in a candidate build by EOD Friday.


Yep. Not much should be affected up stack but there are some settings that no longer work that exists on `KestrelSeverOptions`

src/Servers/Kestrel/Core/src/Internal/ConnectionDispatcher.cs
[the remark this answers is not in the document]
halter73: I'm assuming you mean captured by the local function. `this` is captured too. You could make the local function static and pass in both the listener an this as arguments, but this is only called once to start the accept loop.

src/Servers/Kestrel/Core/src/Internal/ConnectionDispatcher.cs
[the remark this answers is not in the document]
halter73: I'd say we should leave it to the transports to do the right thing. If having multiple loops is helpful, a transport can do that itself and merge the results into a single channel or something like libuv.

src/Servers/Kestrel/Transport.Libuv/src/Internal/ListenerContext.cs
```diff
@@ -2,46 +2,70 @@
...
-using Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.Internal;
+using Microsoft.AspNetCore.Connections;
 using Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal.Networking;
 using Microsoft.Extensions.Logging;
 
 namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal
 {
     internal class ListenerContext
     {
+        private readonly Channel<LibuvConnection> _acceptQueue = Channel.CreateUnbounded<LibuvConnection>();
```
halter73: Should this be single reader?

```suggestion
        private readonly Channel<LibuvConnection> _acceptQueue = Channel.CreateUnbounded<LibuvConnection>(new UnboundedChannelOptions { SingleReader = true });
```

halter73:
> TransportConnection has been moved to Connection.Abstractions (we can decide if we need to consolidate with DefaultConnectionContext in a later PR)

It seems like you want to do this. It's probably a good idea. Do you want to file an issue?

halter73:
> NoDelay is a noop, on ListenOptions, we should consider deprecating it.

Deprecate it.

davidfowl:
> Deprecate it.

I'm removing it, prior art is MVC which did a similar thing for options that no longer work.

src/Servers/Kestrel/Transport.Libuv/test/LibuvConnectionTests.cs
```diff
@@ -19,167 +19,153 @@ public class LibuvConnectionTests
...
-                    listenerContext.TransportContext.ConnectionDispatcher.OnConnection(connection);
-                    connectionTask = connection.Start();
+                    listenerContext.HandleConnection(socket);
 
                     mockLibuv.AllocCallback(socket.InternalGetHandle(), 2048, out var ignored);
                     mockLibuv.ReadCallback(socket.InternalGetHandle(), 5, ref ignored);
 
                 }, null);
 
+                var connection = await listenerContext.AcceptAsync();
```
halter73: It's nice how simplified this is.

src/Servers/Kestrel/Transport.Libuv/test/LibuvTransportTests.cs
```diff
@@ -23,71 +23,142 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Tests
...
+            var transport = new LibuvConnectionListener(transportContext, new IPEndPoint(IPAddress.Loopback, 0));
 
             await transport.BindAsync();
             await transport.UnbindAsync();
-            await transport.StopAsync();
+            await transport.DisposeAsync();
         }
 
-        [Theory]
-        [MemberData(nameof(ConnectionAdapterData))]
```
halter73: Why do you hate my variations‽ /s

src/Servers/Kestrel/Transport.Libuv/test/ListenerPrimaryTests.cs
```diff
@@ -270,44 +237,73 @@ public async Task PipeConnectionsWithWrongMessageAreLoggedAndIgnored()
...
-                    Assert.Contains(response, allowed);
+                    expected ??= listenerPrimary.AcceptAsync().AsTask();
+                    currentListener = listenerPrimary;
                 }
 
-                await Task.Delay(retryDelay);
-            }
+                using var socket = await HttpClientSlim.GetSocket(address);
+
+                await using var connection = await expected.DefaultTimeout();
```
halter73: Yeah, you did take some artistic license on this test. It would be nice if there was a quicker way we could verify HttpClientSlim actually has the same socket listenerPrimary does. Before, the "Primary" vs. "Secondary" responses gave you a quick error instead of this test failing with a timeout.

davidfowl: It can't though. We need to verify what's in the queues, there's no way to have different "applications" for different listeners, the fact that it worked that way before was a bit of a fluke because we had a different dispatcher per listener and we took advantage of that.

src/Servers/Connections.Abstractions/ref/Microsoft.AspNetCore.Connections.Abstractions.netstandard2.0.cs
```diff
@@ -30,12 +30,16 @@ public static partial class ConnectionBuilderExtensions
     public abstract partial class ConnectionContext
```
halter73: File an issue to make IDisposable.

src/Servers/Connections.Abstractions/ref/Microsoft.AspNetCore.Connections.Abstractions.netstandard2.0.cs
```diff
@@ -70,40 +74,97 @@ public partial class ConnectionResetException : System.IO.IOException
         public ConnectionResetException(string message) { }
         public ConnectionResetException(string message, System.Exception inner) { }
     }
-    public partial class DefaultConnectionContext : Microsoft.AspNetCore.Connections.ConnectionContext, Microsoft.AspNetCore.Connections.Features.IConnectionIdFeature, Microsoft.AspNetCore.Connections.Features.IConnectionItemsFeature, Microsoft.AspNetCore.Connections.Features.IConnectionLifetimeFeature, Microsoft.AspNetCore.Connections.Features.IConnectionTransportFeature, Microsoft.AspNetCore.Connections.Features.IConnectionUserFeature, System.IDisposable
+    public partial class DefaultConnectionContext : Microsoft.AspNetCore.Connections.ConnectionContext, Microsoft.AspNetCore.Connections.Features.IConnectionEndPointFeature, Microsoft.AspNetCore.Connections.Features.IConnectionIdFeature, Microsoft.AspNetCore.Connections.Features.IConnectionItemsFeature, Microsoft.AspNetCore.Connections.Features.IConnectionLifetimeFeature, Microsoft.AspNetCore.Connections.Features.IConnectionTransportFeature, Microsoft.AspNetCore.Connections.Features.IConnectionUserFeature, System.IDisposable
```
halter73: Should we remove the synchronous IDisposable from this?

davidfowl: I'll file it under the same issue as IAsyncDisposable, I think that'll break some signalr tests.

src/Servers/Connections.Abstractions/src/IConnectionListenerFactory.cs
```diff
@@ -0,0 +1,17 @@
...
+using System;
+using System.Collections.Generic;
+using System.Net;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Connections
+{
+    public interface IConnectionListenerFactory
```
halter73: How nice would it be to avoid the word factory? Maybe IEndpointBinder?

src/Servers/Kestrel/Transport.Sockets/src/SocketTransportOptions.cs
```diff
@@ -17,6 +17,18 @@ public class SocketTransportOptions
...
+        /// <remarks>
+        /// Defaults to true.
+        /// </remarks>
+        public bool NoDelay { get; set; } = true;
+
+        public long? MaxReadBufferSize { get; set; } = PipeOptions.Default.PauseWriterThreshold;
+
+        public long? MaxWriteBufferSize { get; set; } = PipeOptions.Default.PauseWriterThreshold;
+
+        internal Func<MemoryPool<byte>> MemoryPoolFactory { get; set; } = System.Buffers.MemoryPoolFactory.Create;
```
halter73: How many people do you expect to look at this and think "Oh look, *that's* where they wire up Kestrel's SlabMemoryPool"?

src/Shared/Buffers.MemoryPool/MemoryPoolFactory.cs
```diff
@@ -1,12 +1,9 @@
...
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
-using System.Buffers;
-
-namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Abstractions.Internal
+namespace System.Buffers
 {
-    public static class KestrelMemoryPool
+    internal static class MemoryPoolFactory
```
halter73: Can we give this abetter name and/or namespace?

src/SignalR/SignalR.sln
```diff
@@ -147,6 +147,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.AspNetCore.HttpOv
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.AspNetCore.SignalR.Protocols.Json", "common\Protocols.Json\src\Microsoft.AspNetCore.SignalR.Protocols.Json.csproj", "{BB52C0FB-19FD-485A-9EBD-3FC173ECAEA0}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.AspNetCore.Metadata", "..\Http\Metadata\src\Microsoft.AspNetCore.Metadata.csproj", "{2E107FBB-2387-4A9F-A2CA-EFDF2E4DD64D}"
```
halter73: You're adding IAllowAnonymous and IAuthorizeData to the SignalR solution in this PR?

src/Servers/Kestrel/Transport.Sockets/src/Internal/SocketConnection.cs
```diff
@@ -32,44 +30,61 @@ internal sealed class SocketConnection : TransportConnection, IDisposable
         private readonly object _shutdownLock = new object();
         private volatile bool _socketDisposed;
         private volatile Exception _shutdownReason;
-
-        internal SocketConnection(Socket socket, MemoryPool<byte> memoryPool, PipeScheduler scheduler, ISocketsTrace trace)
+        private Task _task;
```
halter73: Come up with a better name please.

src/Servers/Kestrel/Transport.Libuv/src/Internal/LibuvConnection.cs
```diff
@@ -123,19 +161,25 @@ public async Task Start()
...
+            Transport.Input.Complete();
+            Transport.Output.Complete();
+
+            if (_task != null)
+            {
+                await _task;
+            }
+
             _connectionClosedTokenSource.Dispose();
-            _connectionClosingCts.Dispose();
```
halter73: Why did this get removed?

src/Shared/Buffers.MemoryPool/MemoryPoolFactory.cs
davidfowl: Suggest a name, this is basically unnecessary churn at this point.

src/SignalR/SignalR.sln
davidfowl: I needed it to compile, so pretend its ok. We added a new package but didn't add it to all the places. I had to make sure SignalR was working so I included this in the PR

src/Servers/Kestrel/Transport.Libuv/src/Internal/LibuvConnection.cs
davidfowl: The transport no longer owns this but it's not being disposed right now. It's in KestrelConnection.

src/Servers/Kestrel/Transport.Sockets/src/SocketTransportOptions.cs
davidfowl: Maybe like 4 people, 5 if we're lucky 😄

src/Servers/Connections.Abstractions/src/IConnectionListenerFactory.cs
davidfowl: Not changing the name in this PR, we agreed to this one in the meeting, we can iterate on the issue.

davidfowl: That said, I don't disagree

src/Servers/Connections.Abstractions/ref/Microsoft.AspNetCore.Connections.Abstractions.netstandard2.0.cs
davidfowl: https://github.com/aspnet/AspNetCore/issues/10607

src/Servers/Kestrel/Transport.Libuv/src/Internal/LibuvConnection.cs
davidfowl: Fixed

src/Servers/Kestrel/Transport.Libuv/test/ListenerPrimaryTests.cs
halter73: You could write "Primary" or "Secondary" to the LibuvConnection after accepting it, no?

src/Shared/Buffers.MemoryPool/MemoryPoolFactory.cs
halter73: KestrelMemoryPoolFactory

halter73: in the old namespace

davidfowl: Nah

halter73: Do you want it approved or not? 😉

halter73 (approved):
Approved pending finding and fixing the root cause of the UvLoopHandle finalized. It would also be nice if you wrote "Primary"/"Secondary" to the accepted pipes in ListenerPrimaryTests and verified the response.

src/Servers/Kestrel/Transport.Libuv/test/ListenerPrimaryTests.cs
davidfowl: Sure, I can do that but the real test is the fact that it actually is en-queued in the right queue, not the writing to the queue.

davidfowl:
> Approved pending finding and fixing the root cause of the UvLoopHandle finalized.

OK I spent a few hours on this and I'm still stumped and haven't been able to write a test to cause the condition to happen. I make a few tweaks to use await using in a couple of places just to make sure we're calling dispose in the right spots.

What if we just add a log [here](https://github.com/aspnet/AspNetCore/blob/master/src/Servers/Kestrel/Transport.Libuv/src/Internal/LibuvThread.cs#L303) so that we can know if this is a resource leak or a bad unit test?

halter73:
> 
What if we just add a log here  so that we can know if this is a resource leak or a bad unit test?

Seems reasonable. We should probably log something anytime OnStopRude() let alone OnStopImmediate() is called. Maybe we could make those logs fail tests like the ungraceful shutdown log does.

src/Servers/Kestrel/Transport.Sockets/src/SocketTransportOptions.cs
```diff
@@ -17,6 +17,18 @@ public class SocketTransportOptions
...
-        internal Func<MemoryPool<byte>> MemoryPoolFactory { get; set; } = () => KestrelMemoryPool.Create();
+        /// <summary>
+        /// Set to false to enable Nagle's algorithm for all connections.
+        /// </summary>
+        /// <remarks>
+        /// Defaults to true.
+        /// </remarks>
+        public bool NoDelay { get; set; } = true;
+
+        public long? MaxReadBufferSize { get; set; } = PipeOptions.Default.PauseWriterThreshold;
```
jkotalik: Nit: doc comments?

PR #10340, opened 2019-05-17
CopyCabs

joeloff:
Copy CAB files to artifacts


joeloff:
This should copy CAB files if there are any

@natemcmaster, I'm assuming that signing is done by the time we hit this target.

dougbu (approved):
Straightforward 😺

dougbu:
@JunTaoLuo and @natemcmaster I believe this is just FYI for you two.

@joeloff, yes, the cabs are signed before the copy you're changing.

JunTaoLuo (approved):

dougbu:
@joeloff the required checks are done. Let's get this. Unless, that is, you're waiting for something specific in the Helix tests?

PR #10341, opened 2019-05-17
Update BuildFromSource.md to add JAVA_HOME instructions

rynowak:
@natemcmaster - I had to manually set JAVA_HOME with the latest Oracle JDK


natemcmaster:
Try pulling latest changes from master. I don't think this should be necessary anymore. There was a bug in the script which I fixed this morning as a part of https://github.com/aspnet/AspNetCore/pull/10306

natemcmaster:
The underlying issue here should have been fixed with https://github.com/aspnet/AspNetCore/pull/10369. Please let aspnet/build know if you still have issues with the JDK.

PR #10352, opened 2019-05-17
Disable test reporting for flaky tests on Helix

JunTaoLuo:
While I look into https://github.com/aspnet/AspNetCore-Internal/issues/2482. We might want to disable test reporting of flaky Helix tests so the PR checks are more meaningful.

WIP don't review yet.

### DIFF
--- a/eng/helix/content/runtests.cmd
+++ b/eng/helix/content/runtests.cmd
@@ -52,7 +52,7 @@ if errorlevel 1 (
 
 set FLAKY_FILTER="Flaky:All=true|Flaky:Helix:All=true|Flaky:Helix:Queue:All=true|Flaky:Helix:Queue:%HELIX%=true"
 echo Running known-flaky tests.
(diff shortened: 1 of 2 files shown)


jkotalik (approved):

JunTaoLuo:
aspnet/AspNetCore-Internal#2482 will likely take some feature work to resolve so I think we probably want to merge this and make the helix test check marks more meaningful. @natemcmaster @dougbu . Note that even this test run isn't green but that's because there are flaky tests that haven't been marked as such.

natemcmaster:
> Note that even this test run isn't green but that's because there are flaky tests that haven't been marked as such.

Let's apply `[Flaky]`.

natemcmaster (approved):
Fine a workaround for now. Let's make sure to work with the dnceng team to see if we can make Helix reporting understand the concept of flaky tests.

JunTaoLuo:
Yup @ @alexperovich  is looking into devops features and workarounds.

JunTaoLuo:
Seems like I mis-spoke the helix test failures are not flakiness, they are consistently failing. Will file followup issues.

Issue #7635, opened 2019-02-15
Disable request body data rate limits per HTTP/2 stream


JunTaoLuo:
Epic https://github.com/aspnet/AspNetCore/issues/8858

Currently we can disable the request body data rate limit for the entire server with
```c#
                    webBuilder.ConfigureKestrel(options =>
                    {
                        options.Limits.MinRequestBodyDataRate = null;
                    });
```

However, we want to be able to control this on a finer scale of per request/stream. We need this in since certain gRPC endpoints are client streaming methods which may not satisfy the request body data rate while all other endpoints of the app should still be subject to the request body data rate. Websocket scenarios will also benefit from this feature.

To implement this feature, it's likely we'll need to overhaul the accounting for request data rates. @halter73 has more context about the implementation here. 

cc @Tratcher @davidfowl

halter73:
There are two options we discussed:

1. A stopgap measure where we allow disabling rate limit for specific HTTP/2 streams/requests, but not modifying the rate limit to a non-null value.

2. A more thorough rearchitecture of HTTP/2 rate limiting which tracks and limits data rates at the stream level instead of the connection level. One potential wrinkle with this idea is that the default rate limit for an HTTP/2 connection would become 240 bytes/sec times the number of streams actively reading from the request body instead of just 240 bytes/sec if the app is actively reading from the request body like it is for HTTP/1.

blowdart:
The stopgap seems a deeply undesirable idea.

halter73:
@blowdart Why is that? It's basically what we do for WebSocket requests. If you have an endpoint where you know the client will use the request body to send intermittent messages, the stopgap measure gives you the behavior you want. Even the more thorough rearchitecture will allow you to get the behavior the stopgap solution give you. It just gives you more complete control.

blowdart:
Because it's a half implementation? And more worrying if that feature is there then you known someone, somewhere is going to turn that on for normal requests.

JamesNK:
@muratg @shirhatti Please prioritize this.

muratg:
Do we need a design meeting for this, @shirhatti?

JunTaoLuo:
ping @anurse we spoke offline and we want to take @halter73 suggestion number 1:
> A measure where we allow disabling rate limit for specific HTTP/2 streams/requests, but not modifying the rate limit to a non-null value.

Can we get a milestone commitment for this work?

JunTaoLuo:
[...]

PR #10355, opened 2019-05-17
Disable request body data rate limits per HTTP/2 stream

halter73:
Addresses #7635

### DIFF
--- a/src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.FeatureCollection.cs
+++ b/src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.FeatureCollection.cs
@@ -28,6 +28,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                                           IHttpRequestIdentifierFeature,
                                           IHttpBodyControlFeature,
                                           IHttpMaxRequestBodySizeFeature,
+                                          IHttpMinRequestBodyDataRateFeature,
                                           IHttpResponseStartFeature,
                                           IEndpointFeature,
                                           IRouteValuesFeature
@@ -225,6 +226,12 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 
                 MaxRequestBodySize = value;
             }
+        }
+
+        MinDataRate IHttpMinRequestBodyDataRateFeature.MinDataRate
+        {
+            get => MinRequestBodyDataRate;
+            set => MinRequestBodyDataRate = value;
         }
 
         PipeWriter IResponseBodyPipeFeature.Writer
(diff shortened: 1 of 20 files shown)


src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.FeatureCollection.cs
```diff
@@ -227,6 +228,12 @@ bool IHttpBodyControlFeature.AllowSynchronousIO
             }
         }
 
+        MinDataRate IHttpMinRequestBodyDataRateFeature.MinDataRate
```
jkotalik: Can this be in the Http1Connection class still?

jkotalik (approved):

src/Servers/Kestrel/Core/src/Internal/Http2/Http2Stream.FeatureCollection.cs
```diff
@@ -30,5 +34,19 @@ IHeaderDictionary IHttpResponseTrailersFeature.Trailers
         }
 
         int IHttp2StreamIdFeature.StreamId => _context.StreamId;
+
+        MinDataRate IHttpMinRequestBodyDataRateFeature.MinDataRate
+        {
+            get => throw new NotSupportedException(CoreStrings.Http2MinDataRateNotSupported);
```
Tratcher: The getter should return null if that's the only allowed value.

Tratcher (approved):

JamesNK:
I'd like to understand how we'll use this in gRPC.

I believe we only want to disable the rate limit on gRPC requests. We could be more specific and only disable the rate limit on client or duplex streaming request. Would that be done by setting this feature on the HttpContext with a null MinDataRate on only gRPC client/duplex streaming requests?

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.FeatureCollection.cs
halter73: Yep.

src/Servers/Kestrel/Core/src/Internal/Http2/Http2Stream.FeatureCollection.cs
halter73: The thing is we generally are enforcing a rate limit for HTTP/2, it's just per connection, not per stream. So I don't think it makes sense to return null, but I don't think it makes sense to return the connection limit either.

halter73:
> Would that be done by setting this feature on the HttpContext with a null MinDataRate on only gRPC client/duplex streaming requests?

Exactly. Something like https://github.com/aspnet/AspNetCore/pull/10355/files#diff-2d9b5c655816cc6e1ea41260d95cd33eR815

JamesNK:
Ah, the `IHttpMinRequestBodyDataRateFeature` will always be set on a HTTP/2 request, and it is a matter of getting it and setting it to null in the appropriate requests.

halter73:
> What happens if you have a HTTP/2 request and someone defines there own IHttpMinRequestBodyDataRateFeature implementation and has a non-null MinDataRate?

That's the prerogative of whoever defines their own IHttpMinRequestBodyDataRateFeature. I doubt a custom IHttpMinRequestBodyDataRateFeature would be able to do much with a non-null MinDataRate and an HTTP/2 request unless is measured the rate of the request body itself.

halter73:
I expect most people who replace a IHttpMinRequestBodyDataRateFeature would be replacing it with a facade that would call the "real" MinDataRate setter which would throw a NotSupportedException given a non-null rate with HTTP/2.

JamesNK:
Ok. I'm not super familiar with this space, just want to make sure it is thought about.

JamesNK (approved):

PR #10356, opened 2019-05-17
Certsign

HaoK:
See if KeyUsage CertSign fixes test failures on linux


HaoK:
Replaced by https://github.com/aspnet/AspNetCore/pull/9756

Issue #8336, opened 2019-03-08
[Components] `EventCallback<T>` changes the `RenderTreeBuilder.AddAttribute(object)` behavior


chucker:
### Describe the bug
The introduction of `EventCallback<T>` in #6351 changes the behavior of `RenderTreeBuilder.AddAttribute(object)` such that calls that were previously treated as delegates — an `onclick` attribute, for example — now fall back to a `ToString()` call.

This breaks code between Blazor 0.8 and 0.9, and doesn't appear to be the intended behavior of the `AddAttribute(object)` overload. [There still _exists_ a case for delegates](https://github.com/aspnet/AspNetCore/blob/98fe8a8328a0b41ac776a1ccdc71943de8093440/src/Components/Components/src/RenderTree/RenderTreeBuilder.cs#L386), but it doesn't seem to be used any more:

```csharp
// BUG: Action<UIMouseEventArgs> in Blazor 0.8 _was_ a MulticastDelegate
// but EventCallback<UIMouseEventArgs> is not!
else if (value is MulticastDelegate)
{
    Append(RenderTreeFrame.Attribute(sequence, name, value));
}
```

### To Reproduce
1. Using this version of ASP.NET Core: 3.0.0-preview3-19153-02
2. Run this code:

```csharp
        protected override void BuildRenderTree(RenderTreeBuilder builder)
        {
            base.BuildRenderTree(builder);
            builder.OpenElement(0, "MyCustomTag");

            var param = new KeyValuePair<string, object>("onclick", 
                EventCallback.Factory.Create<UIMouseEventArgs>(this, s => { }));

            builder.AddAttribute(2, param.Key, param.Value);
            builder.CloseElement();
        }
```

### Expected behavior
`param.Value` should be recognized as a delegate.

### Actual behavior
`RenderTreeBuilder` falls back to treating `param.Value` as an unrecognized object, and calls `ToString()` on it.

Therefore, in the generated DOM in the browser looks something like:

```html
<button
onclick="Microsoft.AspNetCore.Components.EventCallback`1[[Microsoft.AspNetCore.Components.UIMouseEventArgs,
Microsoft.AspNetCore.Components,
Version=3.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60]]">
```

### Additional context
This came up in the _BlazorStrap_ project, and is tracked there as https://github.com/chanan/BlazorStrap/issues/42.

mkArtakMSFT:
[...]

Issue #5071, opened 2018-05-02
Syntax for writing a dictionary of attributes onto an element


chanan:
This issue is being opened by request of @SteveSandersonMS Please see the full issue details https://github.com/aspnet/Blazor/issues/735 in the Blazor Repo.

The summary is that will be nice to be able to pass arbitrary attributes from a custom component down to an HTML element. For example:

`<MyComponent id="myid" />`

in the component render it as:

`<button id="myid" />`

without having to code every single attribute possible. Not mention allowing attributes such data-* to be passed down.

rynowak:
@SteveSandersonMS - added some design notes, would like your feedback on this.

SteveSandersonMS:
This basically all looks great to me and matches what I think we've discussed before.

> the only syntactic conflict is with directive attributes

One further syntactic possibility is to copy JS/TS and be explicit that we're merging contents by prefixing with `...` as JS does. So the closest Razorish equivalent to the JS/TS syntax would be:

    <somelem normal-attribute="something" @...myAttributes />

You'd then get a compile error if you just tried to put `@myAttributes` inside an element, except if that's the name of a directive attribute.

I'm not 100% determined that we should do this, but am interested in what you think.

chucker:
I'm not sure JS-esque `@...` i quite the right 'idiomatic' syntax for C#, but I do favor such explicitness, given recent issues like https://github.com/aspnet/AspNetCore/issues/9786 which arose because the Razor syntax _isn't_ particularly explicit.

rynowak:
> One further syntactic possibility is to copy JS/TS and be explicit that we're merging contents by prefixing with ... as JS does. So the closest Razorish equivalent to the JS/TS syntax would be:
> <somelem normal-attribute="something" @...myAttributes />

This would be possible to do, but could be tricky. I think this would require parser changes since we do shenanigans with `.`.

Originally I was thinking about something like `<someelem normal-attribute="something" @splat="myAttributes" />` or `<someelem normal-attribute="something" @attrs="myAttributes" />`

rynowak:
There's a bit of a tricky thing to figure out here, because the behaviour of HTML is different from what's described here. 

If you write the following:
```html
<div class="test1" class="test2">Peasy Cheese</div>
```

The result (in the DOM) is `<div class="test1">Peasy Cheese</div>`. That's right, static HTML does the opposite of what we've specced out here.

[...]

PR #10357, opened 2019-05-17
Add runtime support for Blazor attribute splatting

rynowak:
Implements: #5071

See design notes here for details: https://github.com/aspnet/AspNetCore/issues/5071#issuecomment-386045586

Opening a PR now to get early feedback and discussions. This will require checkins and updates from the tooling repo to fix the runtime behaviour of markup blocks and to add support for splatting as a language features (syntax not yet locked).


src/Components/Components/src/Rendering/HtmlRenderer.cs
```diff
@@ -189,6 +193,29 @@ private int RenderChildren(List<string> result, ArrayRange<RenderTreeFrame> fram
...
+            // We need to de-duplicate attributes when rendering to HTML. We allow *last-attribute-wins* when writing
+            // Blazor code, but HTML is the opposite.
+            //
+            // To implement this we do two passes through the attributes:
+            // - On the first pass, record the index of the *last* occurrence of the attribute.
+            // - On the second pass, write each attribute *if* it's the last occurrence.
+            //
+            // This process is done by the compiler for markup blocks (fully static HTML), so we don't need to repeat
+            // it there.
+            _currentAttributes.Clear();
```
rynowak: @SteveSandersonMS @javiercn - to implement "last attribute wins" behaviour we either need to do this:
- inside the `RenderTreeBuilder` by having it buffer all attributes
- inside the renderer implementations

I'm interested in ideas about which one we should do and why. I started with the *inside the renderer* approach because it already "just works" except for the static html case.

However, this is precedent for the `RenderTreeBuilder` to be *smart*. It already knows to omit bool false values.

SteveSandersonMS: If we can do it cheaply enough inside the builder that would keep things maximally consistent, but I’d hope it doesn’t add cost in the non-splatting cases. For example we could go into buffered mode only once we see the first add-splatted-attributes call, if it’s not too complex. 

Otherwise I’m unsure how the diff algo will handle it. It doesn’t expect to see duplicate attributes, and since it uses a dictionary-based merge for attributes sometimes, that will need to be checked carefully to ensure things get overwritten in the correct order.

rynowak: Hmm. There's actually a problem with this approach in general. Currently the RTB omits frames for things like `bool false` attributes. The technique used here is going to have subtle bugs unless all calls to `AddAttribute` result in a frame. 

```
<div @attributes disable="@disableMyDiv">...</div>
```

We want the effect of `disable="@disableMyDiv"` to supersede whatever is in `@myAttribtues`.

I'll try to the opposite approach and see what happens. I have an idea of how to do without any additional data structures.

rynowak:
@SteveSandersonMS @javiercn @pranavkm - ready for your review.

I need to merge the `AddMultipleAttributes` changes in order to make progress on the compiler. I'd also really like it if we could lock on the attribute/parameter terminology now.

src/Components/Components/src/Rendering/HtmlRenderer.cs
rynowak: @SteveSandersonMS - please take a look at how this is handled inside the RTB now.

src/Components/Analyzers/src/ComponentFacts.cs
```diff
@@ -0,0 +1,89 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+
+namespace Microsoft.AspNetCore.Components.Analyzers
+{
+    internal static class ComponentFacts
```
rynowak: This is part of the design that I've settled on for most analyzers we're building in ASP.NET Core.

Basically:
1. Big chunky analyzers that implement many features allows for efficient data sharing
2. Checking for symbols up front, and using symbols rather than syntax is productive and simple
3. Pulling logic into static methods based on symbols keeps code free of clutter and duplication

src/Components/Analyzers/test/ComponentsTestDeclarations.cs
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Components.Analyzers
+{
+    public static class ComponentsTestDeclarations
+    {
+        public static readonly string Source = $@"
+    namespace {typeof(ParameterAttribute).Namespace}
```
rynowak: This is a looming maintenance problem, but I chose not to solve it in this PR. In our other analyzers we use a more mature test infrastructure, and we use the actual product code for testing.

src/Components/Components/src/EventCallback.cs
```diff
@@ -9,7 +9,7 @@ namespace Microsoft.AspNetCore.Components
     /// <summary>
     /// A bound event handler delegate.
     /// </summary>
-    public readonly struct EventCallback
+    public readonly struct EventCallback : IEventCallback
```
rynowak: I also fixed #8336 in this PR because I was looking at it 😆

src/Components/Components/src/ParameterAttribute.cs
```diff
@@ -11,5 +13,24 @@ namespace Microsoft.AspNetCore.Components
...
+        /// <para>
+        /// <see cref="CaptureExtraAttributes"/> can be used on at most one parameter per component.
+        /// </para>
+        /// <para>
+        /// <see cref="CaptureExtraAttributes"/> should only be applied to parameters of a type that
+        /// can be used with <see cref="RenderTreeBuilder.AddMultipleAttributes{T}(int, System.Collections.Generic.IEnumerable{System.Collections.Generic.KeyValuePair{string, T}})"/>
+        /// such as <see cref="Dictionary{String, Object}"/>.
+        /// </para>
+        /// </remarks>
+        public bool CaptureExtraAttributes { get; set; }
```
rynowak: Suggestions for name?

I feel like making this an option on `[Parameter]` is the best choice. We want this to be settable as a normal parameter as well.

src/Components/Components/src/RenderTree/RenderTreeBuilder.cs
```diff
@@ -425,6 +447,38 @@ public void AddAttribute(int sequence, in RenderTreeFrame frame)
...
+            // Calling this up-front just to make sure we validate before mutating anything.
+            AssertCanAddAttribute();
+
+            if (attributes != null)
+            {
+                foreach (var attribute in attributes)
+                {
+                    // This will call the AddAttribute(int, string, object) overload.
+                    //
+                    // This is fine because we try to make the object overload behave identically
```
rynowak: This is where I ended up deciding to fix the eventcallback bug 🤣

src/Components/Components/src/RenderTree/RenderTreeBuilder.cs
```diff
@@ -590,13 +644,42 @@ public void Clear()
...
             _entries.Append(frame);
 
-            var frameType = frame.FrameType;
             if (frameType != RenderTreeFrameType.Attribute)
             {
                 _lastNonAttributeFrameType = frame.FrameType;
             }
         }
+
+        private void ClearAttributesWithName(string name)
```
rynowak: Options I see for this:
1. What I did here (clear dupes of each attribute as it's added) - simple and O(N^2)
2. Clear dupes of *all* attributes for an element/component when it's closed - somewhat simple and also O(N^2)
3. Use a HashSet - simple, requires extra data/hashing and still kinda O(N^2) before removing from an array is O(N)

Basically I felt like keeping the data in the *always consistent* state was the best thing overall. Without doing dedicated perf work around this specific feature, I don't have a clear insight into which of these are the best. That seems like a good thing to do some other time.

src/Components/Components/src/EventCallback.cs
SteveSandersonMS: Sounds good.

Is there any value in `EventCallback` being a `struct`? Are there any common use cases where it *won't* get boxed?

src/Components/Components/src/ParameterAttribute.cs
SteveSandersonMS: I'd prefer we call these *parameters*, not attributes, since the use cases for this feature is more general than just splatting attributes. For example you might want to capture all extra supplied parameters programmatically then run some procedural logic over that dictionary, and not necessarily splat them back out at all.

So, I'd be open to any of:

 1. `[Parameter(CaptureExtraParameters = true)]`
 2. `[Parameter(CaptureUnboundParameters = true)]`
 3. `[Parameter(CaptureUnmatchedParameters = true)]`
 4. `[UnboundParameters]` (where `UnboundParameters` inherits from `[Parameter]`, so you can still pass it explicitly)
 5. `[UnmatchedParameters]` (where `UnmatchedParameters` inherits from `[Parameter]`, so you can still pass it explicitly)

I think I prefer them in reverse order of that list (i.e., `[UnmatchedParameters]` as my favourite). If you think it's problematic to inherit from `[Parameter]` then a bool flag is fine with me too.

Most .NET devs would think of "unbound" as more intuitive than "unmatched", but it's always annoying to overload termonology hence my reluctance to reuse the notion of "binding".

SteveSandersonMS: I expect it's clear, but just in case, I think it's legitimate to call these "unmatched parameters" as opposed to "unmatched arguments" based on the "actual parameters vs formal parameters" distinction rather than the "arguments vs parameters" distinction.

SteveSandersonMS: One last option, in case the "parameters" vs "arguments" thing is too problematic:

```
[Parameter(CaptureUnmatchedValues = true)]
```

src/Components/Components/src/ParameterCollectionExtensions.cs
```diff
@@ -41,17 +42,68 @@ public static class ParameterCollectionExtensions
                 _cachedWritersByType[targetType] = writers;
             }
 
+            // The logic is a little convoluted now that we have "Extra" parameters. We want to avoid allocations where
```
SteveSandersonMS: This is interesting. I didn't expect the splatting resolution to happen here in `SetParameterProperties`. This means that splatting becomes a feature of `SetParameterProperties`, so if someone tries to write manual logic to iterate over the parameter collection (as we do in various built-in components), it won't work because they will see the actual dictionaries of parameters to be splatted.

We need to choose whether splat resolution is meant to be visible to receivers. What use cases are there for knowing what specific dictionaries were supplied, instead of just seeing the result of combining them all? If we choose only to make visible the results of combining all the dictionaries, then ideally this would be resolved at an earlier phase, either:

 * During the render tree builder itself (for each "set multiple parameters" call, we iterate over the dictionaries at that point so the output is indistinguishable from a normal series of "set attribute" calls, and to implement the overwriting logic, we scan backwards through the attribute frames and replace any with the same name). This is what I was expecting.
 * Or, inside `ParameterCollectionExtensions`, implementing the equivalent of what you have here, but in the custom iterator. Then in any typical use case you'd only observe the results of combining the dictionaries, and `SetParameterProperties` would get it automatically.

Do you think either of those would be possible?

SteveSandersonMS: Update: Please ignore the above comment. Now I've read the actual code here and more of the implementation later, I see you are already doing what I was suggesting. Sorry for the confusion.

src/Components/Components/src/ParameterCollectionExtensions.cs
```diff
@@ -41,17 +42,68 @@ public static class ParameterCollectionExtensions
...
+            // 2. Using an unknown parameter when there is an "Extra" parameter defined *AND* explicitly
+            //    providing a value for the extra parameter.
+            //
+            // The second case has to be an error because we want to allow users to set the "Extra" parameter
+            // explicitly, but we don't ever want to mutate a value the user gives us. We also don't want to
+            // implicitly copy a value the user gives us. Either one of those implementation choices would do
+            // something unexpected.
+
+            var isExtraParameterSetExplicitly = false;
+            Dictionary<string, object> extras = null;
```
SteveSandersonMS: I know it's a bit nitpicky and unquantified-micro-optimizationy, but could we skip most of these extra per-param tests inside the loop if we know the receiver component doesn't have any "extra attributes" parameter (e.g., if `writers.CaptureExtraAttributesPropertyName == null`)?

If we checked for that once up front, could we eliminate the many of the checks inside the loop, even if it comes at the cost of a bit of extra code duplication?

It would be cool if this feature was basically zero cost in the case where you're not using it.

src/Components/Components/src/RenderTree/RenderTreeBuilder.cs
SteveSandersonMS: I'm not at all keen on the `ClearAttributesWithName` technique. It might be just about tolerable if it only applied when we know you're splatting, but to do it in all cases seems very troublesome. Adding a linear search to find matches, including the case-insensitive string comparisons, makes all parameters everywhere more expensive.

So most importantly, I think we only need to handle duplicates at all once we observe the first call to `AddMultipleAttributes` within a given container.

As for the deduplication implementation, what do you think of the following slight variation of combining your options 2 and 3?

 * Add all the attributes without attention to dupes initially, but track a flag to say whether there's been at least one call to `AddMultipleAttributes`
 * When closing an element/component, if the "is splatting" flag is set, then iterate backwards over the parameter frames to build a `Set<string>` containing all the parameter names. For each duplicate, don't delete it from the array - just overwrite it with a marker frame to say this frame should be ignored. For instance, it could be an attribute frame whose `AttributeName` is null, or it could be a new frame type entirely. Then we update any code that iterates over attribute frames so it knows to skip such items (there will only be a few places where we do this, as most just uses the custom iterator in `ParameterCollectionExtensions`). This reduces the O(N^2) to O(N).

SteveSandersonMS: Hopefully then we wouldn't also need the massive number of unit tests dealing with deduplication of different types, as the deduplication logic would be the same code paths for all types.

src/Components/test/testassets/BasicTestApp/DuplicateAttributesComponent.razor
```diff
@@ -0,0 +1,5 @@
+<div id="duplicate-on-component">
+    <DuplicateAttributesChildComponent Message="test1" Message="test2" />
```
SteveSandersonMS: I'm hoping we don't need this feature. Duplicate attributes are only useful in conjunction with splatting - I think we can consider it either erroneous or undefined-behavior to have duplicate explicit (non-splatted) attributes. Then we can avoid most of the extra runtime costs.

Do you think that would be reasonable?

src/Components/Components/src/ParameterCollectionExtensions.cs
```diff
@@ -41,17 +42,68 @@ public static class ParameterCollectionExtensions
...
+                    // Case 1: There is nowhere to put this value.
                     ThrowForUnknownIncomingParameterName(targetType, parameterName);
+                    throw null; // Unreachable
+                }
+
+                isExtraParameter = writer == null;
+
+                if (isExtraParameter)
+                {
+                    extras ??= new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase);
```
pranavkm: `Ordinal`? The behavior's odd even if you do `OrdinalIgnoreCase` since it just knocks out the previous value.

src/Components/Components/src/ParameterCollectionExtensions.cs
```diff
@@ -120,8 +203,34 @@ public WritersForType(Type targetType)
...
+                        // This is an "Extra" parameter.
+                        //
+                        // There should only be one of these.
+                        if (CaptureExtraAttributesWriter != null)
+                        {
+                            ThrowForMultipleCaptureExtraAttributesParameters(targetType);
+                        }
+
+                        // It must be able to hold a Dictionary<string, object> since that's what we create.
+                        if (!propertyInfo.PropertyType.IsAssignableFrom(typeof(Dictionary<string, object>)))
```
pranavkm: Any reason this isn't an `IDictionary<string, object>`?

src/Components/Components/src/RenderTree/ArrayBuilder.cs
```diff
@@ -97,6 +97,17 @@ internal int Append(T[] source, int startIndex, int length)
         public void Overwrite(int index, in T value)
             => _items[index] = value;
 
+        /// <summary>
+        /// Removes the item at the specified index.
+        /// </summary>
+        /// <param name="index">The index of the item to remove.</param>
+        public void RemoveAt(int index)
+        {
+            Array.Copy(_items, index + 1, _items, index, _itemsInUse - 1 - index);
```
pranavkm: Will `index + 1` throw if it's attempting to remove the last element in the array?

src/Components/Components/src/EventCallback.cs
rynowak: Yes, in many of the common cases it's not boxed. Specifically places like `<input type="button" onclick="@OnClick" />`. Generally we have to box with a component, but not with an element.

src/Components/Components/src/ParameterAttribute.cs
rynowak: > I expect it's clear, but just in case, I think it's legitimate to call these "unmatched parameters" as opposed to "unmatched arguments" based on the "actual parameters vs formal parameters" distinction rather than the "arguments vs parameters" distinction.

I don't think you and I will ever agree 100% on the debate between parameters and arguments - and I'm not trying to start that discussion again 😆 I still really prefer terminology about `Values` vs `Parameters` and maybe that's a point of common ground.

--

The things that I don't like about `[UnmatchedParameters]` is that inheritance is problematic with the way we use these attributes (and there are examples of this in TagHelpers). 

We expect to add more API surface to these attributes that we will statically analyze with the compiler/analyzers and other tools. To make that work it's really important that we don't let the user **run code** to affect the settings we care about, it has to be purely declarative. The constructor of a defined class (if we unseal the attribute) is the only place a user can do this. 

People did this with TagHelpers, and found that it worked in the runtime, as long as you use runtime compilation. It did not work in the compiler or at build time (which we added later).

So I guess we could tell people that's not supported, but there's other wierdness with unsealed *and* non-abstract attribute types that I'd like to avoid.

----

Basically I think the best option of these is:

```C#
[Parameter(CaptureUnmatchedValues = true)]
```

Unless we find a much better idea than this, I'm going to go with it for now. We can debate this more once we can start writing code with it.

src/Components/test/testassets/BasicTestApp/DuplicateAttributesComponent.razor
rynowak: Discussed offline and fixed in the tooling repo.

src/Components/Components/src/ParameterCollectionExtensions.cs
rynowak: I'll try this out. It might even be a lot simpler to read this way.

rynowak: Take a look now, it's vastly simplified by breaking it up into two loops 👍

src/Components/Analyzers/test/ComponentCaptureExtraAttributesParameterMustBeUniqueTest.cs
```diff
@@ -0,0 +1,68 @@
...
+        public void IgnoresPropertiesWithCaptureExtraAttributesFalse()
+        {
+            var test = $@"
+    namespace ConsoleApplication1
+    {{
+        using System.Collections.Generic;
+        using {typeof(ParameterAttribute).Namespace};
+        class TypeName
+        {{
+            [Parameter(CaptureExtraAttributes = false)] string MyProperty {{ get; set; }}
```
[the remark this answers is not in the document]
rynowak: It will work in the correct way. Processing of the diagnostics is separate. We process the property types per-property, then we process the uniqueness per-type.

src/Components/Components/src/ParameterCollectionExtensions.cs
rynowak: Attribute matching in HTML is case-insensitive.

src/Components/Components/src/ParameterCollectionExtensions.cs
rynowak: Because that's more restrictive than required.

src/Components/Components/src/RenderTree/ArrayBuilder.cs
rynowak: No, because it will be 0-length.

src/Components/Analyzers/src/ComponentsApi.cs
```diff
@@ -1,18 +1,28 @@
...
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-namespace Microsoft.AspNetCore.Components.Shared
+namespace Microsoft.AspNetCore.Components.Analyzers
 {
     // Constants for type and method names used in code-generation
     // Keep these in sync with the actual definitions
     internal static class ComponentsApi
     {
+        public static readonly string SystemCollectionsGenericDictionary = "System.Collections.Generic.Dictionary`2";
```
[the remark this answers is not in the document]
rynowak: Are you worried that the type name is going to change?

src/Components/Components/src/RenderTree/RenderTreeBuilder.cs
rynowak: > Hopefully then we wouldn't also need the massive number of unit tests dealing with deduplication of different types, as the deduplication logic would be the same code paths for all types.

We could theoretically reduce the number of tests, but each code path where we call `AddAttribute` but don't create a frame needs to have special behaviour. This was my rationale for writing all of these tests, and I'm not totally convinced that it's safe to get rid of them.

I like your proposal and I'm going to try it out, but we're going to end up with *some* logic in each place where we have a *ghost frame*. This is the same concern that led me to writing a million tests. It would be really easy to introduce bugs in these cases - and its a concern that you need to be aware of every time we add another overload to AddAttribute.

rynowak: > Update: We could also make the "false bool" handling work by not only skipping adding the frame, but also by putting the name into the Set<string> at that point so that cleared values get overwritten during the deduplication pass if there is one.

This can work if we use a dictionary and track the position as well.

Imagine the following:
```
<div @foo disabled="@isDisabled" @bar>
```

When we process `disabled` we have *seen* an `AddMultipleAttributes` call (from `@foo`) so we track that we have seen `disabled`. 

Now when we process the `AddMultipleAttributes` for `@bar` we have already *seen* `disabled`. So if `@bar` also contains `disabled` then it would be skipped. Therefore set membership isn't quite enough, we need the position as well.

SteveSandersonMS: A `Dictionary<string, int>` sounds good to me.

It also occurs to me that, once we know which frames are going to be ghosted (e.g., once we've actually ghosted them), we could also do a compaction step where in a single forwards linear pass, we remove all the ghosted items (like [this](https://www.geeksforgeeks.org/move-zeroes-end-array/)).

Then it's not necessary to find and update any other code that walks through attributes, so this *might* be less work overall (and avoids introducing a new concept we have to remember in the future).

I'm not fixated on whether or not we do it this way - up to you.

rynowak:
@javiercn @pranavkm could one of you take a look? Steve and I already discussed the details of this works but he's on vactaion.

javiercn:
Looking at it.

rynowak:
@javiercn @pranavkm ping

src/Components/Components/src/ParameterCollectionExtensions.cs
```diff
@@ -120,8 +208,34 @@ public WritersForType(Type targetType)
...
+                        // This is an "Extra" parameter.
+                        //
+                        // There should only be one of these.
+                        if (CaptureUnmatchedValuesWriter != null)
+                        {
+                            ThrowForMultipleCaptureUnmatchedValuesParameters(targetType);
+                        }
+
+                        // It must be able to hold a Dictionary<string, object> since that's what we create.
+                        if (!propertyInfo.PropertyType.IsAssignableFrom(typeof(Dictionary<string, object>)))
```
pranavkm: Is the condition inverted here or am I reading this incorrectly:

` if (!typeof(Dictionary<string, object>).IsAssignableFrom(propertyInfo.PropertyType))`

src/Components/Components/src/RenderTree/RenderTreeBuilder.cs
```diff
@@ -590,13 +699,111 @@ public void Clear()
...
+        {
+            Debug.Assert(_hasSeenAddMultipleAttributes);
+
+            // When AddMultipleAttributes method has been called, we need to postprocess attributes while closing
+            // the element/component. However, we also don't know the end index we should look at because it
+            // will contain nested content.
+            var buffer = _entries.Buffer;
+            var last = _entries.Count - 1;
+
+            for (var i = first; i <= last; i++ )
```
pranavkm: Formatting (extra space after `i++`)

src/Components/Components/src/RenderTree/RenderTreeBuilder.cs
```diff
@@ -590,13 +699,111 @@ public void Clear()
...
+            {
+                if (buffer[i].FrameType != RenderTreeFrameType.Attribute)
+                {
+                    last = i - 1;
+                    break;
+                }
+            }
+
+            // Now that we've found the last attribute, we can iterate backwards and process duplicates.
+            var seenAttributeNames = (_seenAttributeNames ??= new Dictionary<string, int>(StringComparer.OrdinalIgnoreCase));
```
pranavkm: Could this be short-circuited if a component has no attributes i.e. `last` == `first`?

src/Components/Components/src/RenderTree/RenderTreeBuilder.cs
```diff
@@ -590,13 +699,111 @@ public void Clear()
...
 
-            var frameType = frame.FrameType;
             if (frameType != RenderTreeFrameType.Attribute)
             {
                 _lastNonAttributeFrameType = frame.FrameType;
             }
         }
+
+        // Internal for testing
+        internal void ProcessDuplicateAttributes(int first)
```
pranavkm: I understand some of this.

pranavkm (approved):

src/Components/Components/src/ParameterCollectionExtensions.cs
rynowak: It's correct and it has tests that are passing. We don't need to assign the property type **into** a Dictionary<>, we need to assign a Dictionary<> into the property type.

src/Components/Components/src/RenderTree/RenderTreeBuilder.cs
rynowak: Sure.

src/Components/Components/src/RenderTree/RenderTreeBuilder.cs
rynowak: It understands you pretty well.

src/Components/Components/src/RenderTree/RenderTreeBuilder.cs
rynowak: This isn't super meaningful to optimize. We won't get into this code path unless someone called AddMultipleAttributes.

benaadams:
Breaks on unixy style OSs I think its the embedded crlf in the test string; have added fix https://github.com/aspnet/AspNetCore/pull/10610

PR #10362, opened 2019-05-18
unify the `PasswrodOptions` Required or Require

wangkanai:
`AddDefaultIdentity()` PasswordOptions should have a way unify the grammer

image


jkotalik:
@HaoK @blowdart  thoughts? This is a breaking change.

HaoK:
This doesn't seem to be worth it, doesn't add much value at all

blowdart:
No, it's grammatically correct as it is anyway.  So no, I'm afraid not.

PR #10369, opened 2019-05-18
Build: add support for JDK 8

alefranz:
This fix a bug in detecting the JDK path via registry.
It also adds support to detect JDK 8 as it uses a different registry key (see https://github.com/ojdkbuild/ojdkbuild/wiki/InstallerFeatures)

/cc @natemcmaster

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -329,12 +329,19 @@ if ($RunBuild -and ($All -or $BuildJava) -and -not $NoBuildJava) {
     }
     else {
         try {
-            $jdkVersion = (Get-Item HKLM:\SOFTWARE\JavaSoft\JDK | Get-ItemProperty -name CurrentVersion).CurrentVersion
-            $javaHome = (Get-Item HKLM:\SOFTWARE\JavaSoft\JDK\$jdkVersion | Get-ItemProperty -Name JavaHome).JavaHome
-            if (Test-Path "$javaHome\bin\javac.exe") {
-                $env:JAVA_HOME = $javaHome
-                Write-Host -f Magenta "Detected JDK $jdkVersion in $env:JAVA_HOME (via registry)"
-                $foundJdk = $true
+            $jdkRegistryKeys = @(
+                "HKLM:\SOFTWARE\JavaSoft\JDK",  # for JDK 10+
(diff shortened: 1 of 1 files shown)


build.ps1
```diff
@@ -323,12 +323,19 @@ if (($All -or $BuildJava) -and -not $NoBuildJava) {
     }
     else {
         try {
-            $jdkVersion = (Get-Item HKLM:\SOFTWARE\JavaSoft\JDK | Get-ItemProperty -name CurrentVersion).CurrentVersion
-            $javaHome = (Get-Item HKLM:\SOFTWARE\JavaSoft\JDK\$jdkVersion | Get-ItemProperty -Name JavaHome).JavaHome
-            if (Test-Path "${env:JAVA_HOME}\bin\java.exe") {
```
alefranz: ☝️ This was the bug

natemcmaster: Thanks, I noticed the same problem this morning and merged the same fix in a different PR.

natemcmaster:
Can you confirm that building this code with JDK 8 works? I was under the impression we had started using features which requires a newer JDK 10+, like SHA2 code signing.

Cc @brennanconroy @anurse

anurse:
> I was under the impression we had started using features which requires a newer JDK 10+, like SHA2 code signing.

If SHA2 signing is the only thing that requires JDK 10, I don't see a problem with falling back to JDK 8. As long as our **offical** builds continue to use JDK 10 and SHA2 sign we should be OK. Also, our JDK doesn't do the signing, ESRP does.

alefranz:
Hi Nate,

I don't have errors related to Java in the build.
Is there something in particular I should try?

Anyway I don't have an opinion on supporting Java 8. Updating the Readme (and eventually build errors) to state that Java 10 is required is good as well.

Thank you!

natemcmaster:
Thanks for confirming. CI will keep using JDK 11. Our docs currently say we require JDK 11 or newer (https://github.com/aspnet/AspNetCore/blob/master/docs/BuildFromSource.md#install-pre-requistes), and I think we should keep that as the baseline for now. 

But I suppose there is no harm in allowing contributors to build with JDK 8 if that is what they prefer to use.

PR #10371, opened 2019-05-18
jdk 12.0.1

BenCoden:
I had to run this url to get the jdk

 - InstallJdk.ps1 did not seem to have the right url
 - I could not find a jdk url that would let me hack the version number, so i made a script just for 12.0.1

### DIFF
--- a/eng/scripts/InstallJdk12.0.1.ps1
+++ b/eng/scripts/InstallJdk12.0.1.ps1
@@ -1,0 +1,23 @@
+
+
+$ErrorActionPreference = 'Stop'
+$ProgressPreference = 'SilentlyContinue' # Workaround PowerShell/PowerShell#2138
+
+Set-StrictMode -Version 1
+
+if (-not $env:JAVA_HOME) {
+    throw 'You must set the JAVA_HOME environment variable to the destination of the JDK.'
+}
+
+$repoRoot = Resolve-Path "$PSScriptRoot/../.."
+$tempDir = "$repoRoot/obj"
+mkdir $tempDir -ea Ignore | out-null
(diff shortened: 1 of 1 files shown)


dougbu (changes requested):
This looks redundant given #10306. It also installs Java into the repo's /obj directory and that doesn't make sense.

natemcmaster:
Thanks for the suggestion @BenCoden. At the moment, we have intentionally chosen to stick with JDK 11 because it is the current JDK LTS. If there is a reason we should update to JDK 12, please open an issue and tag the @aspnet/build team.

Closing as we are not planning to take this PR at the moment. If we did update to JDK 12, we would go about implementing this by updating global.json instead and ensuring the JDK 12 is mirrored into our 'netcorenativeassets' blob store. But thank you for the contribution!
-Nate

BenCoden:
Let me leave some words. I was following [BuildFromSource.md](https://github.com/aspnet/AspNetCore/blob/master/docs/BuildFromSource.md) and was running into an issue when trying to run .\restore.cmd, because I didn't have JDK up. I was downloading the JDK from the links on the doc. Then I found [InstallJdk.ps1](https://github.com/aspnet/AspNetCore/blob/master/eng/scripts/InstallJdk.ps1)
but the script was failing, because the URL in it is bad. So maybe we don't need 12.0.1 but a new script is still needed. 
The environment I am using is win10 and VS 19 preview.  Fresh install all across the board.

natemcmaster:
I made changes Thursday that caused these problems. I merged some fixups this morning. Give the latest master branch a try and let me know if you are still having trouble.

PR #10375, opened 2019-05-19
 Fix more null refs in Api Analyzers

pranavkm:
* Make attributes used by analyzer public
 * Compile analyzer tests against ref assemblies
 * Turn on Nullable reference types and harden some null-checks


src/Mvc/Mvc.Analyzers/src/Microsoft.AspNetCore.Mvc.Analyzers.csproj
```diff
@@ -9,6 +9,7 @@
     <IncludeBuildOutput>false</IncludeBuildOutput>
     <GenerateDocumentationFile>false</GenerateDocumentationFile>
     <NuspecFile>$(MSBuildProjectName).nuspec</NuspecFile>
+    <NullableContextOptions>enable</NullableContextOptions>
```
rynowak: Also set `<Nullable>enable</Nullable>` the compiler has already renamed this property, and it will just start failing the build due to not recognizing `IPropertySymbol?` when it hits us.

src/Mvc/Mvc.Analyzers/src/TopLevelParameterNameAnalyzer.cs
```diff
@@ -185,20 +184,44 @@ internal static bool SpecifiesModelType(in SymbolCache symbolCache, IParameterSy
 
         internal readonly struct SymbolCache
         {
+#pragma warning disable RS1008 // Avoid storing per-compilation data into the fields of a diagnostic analyzer.
+            private readonly IMethodSymbol? _idisposableDispose;
```
rynowak: IMO just use the autoprop and `!` assign it.

src/Mvc/Mvc.Analyzers/test/Mvc.Analyzers.Test.csproj
```diff
@@ -21,7 +21,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\..\Mvc.Analyzers\src\Microsoft.AspNetCore.Mvc.Analyzers.csproj" />
-    <Reference Include="Microsoft.AspNetCore.Mvc" />
+    <ProjectReference Include="..\..\Mvc\ref\Microsoft.AspNetCore.Mvc.csproj" />
```
rynowak: spicy

src/Mvc/Mvc.Api.Analyzers/src/ActualApiResponseMetadata.cs
```diff
@@ -26,10 +27,10 @@ public ActualApiResponseMetadata(ReturnStatementSyntax returnStatement, int stat
 
         public ReturnStatementSyntax ReturnStatement { get; }
 
-        public int StatusCode => _statusCode.Value;
+        public int StatusCode => _statusCode ?? throw new ArgumentException("Status code is not available when IsDefaultResponse is true");
```
rynowak: This is kinda a wierd pattern no?

rynowak: Why not make this nullable?

src/Mvc/Mvc.Core/src/Infrastructure/ActionResultStatusCodeAttribute.cs
```diff
@@ -20,7 +20,7 @@ namespace Microsoft.AspNetCore.Mvc.Infrastructure
     /// StatusCodeResult([ActionResultStatusCodeParameter] int statusCode)
     /// </example>
     [AttributeUsage(AttributeTargets.Parameter, AllowMultiple = false, Inherited = false)]
-    internal sealed class ActionResultStatusCodeAttribute : Attribute
+    public sealed class ActionResultStatusCodeAttribute : Attribute
```
rynowak: why tho?

src/Mvc/Mvc.Core/src/Infrastructure/ActionResultObjectValueAttribute.cs
```diff
@@ -21,7 +21,7 @@ namespace Microsoft.AspNetCore.Mvc.Infrastructure
     /// ObjectResult { [ActionResultObjectValueAttribute] public object Value { get; set; } }
     /// </example>
     [AttributeUsage(AttributeTargets.Parameter | AttributeTargets.Property, AllowMultiple = false, Inherited = false)]
-    internal sealed class ActionResultObjectValueAttribute : Attribute
+    public sealed class ActionResultObjectValueAttribute : Attribute
```
rynowak: why tho?

src/Mvc/Mvc.Core/src/Infrastructure/ActionResultStatusCodeAttribute.cs
pranavkm: We look at these attributes as part of the Api Analyzer which requires these to be part of the ref assemblies. I'm going to slap an EditorBrowseable Never on them

pranavkm (commented):
🆙 📅

src/Mvc/Mvc.Analyzers/src/TopLevelParameterNameAnalyzer.cs
```diff
@@ -185,20 +183,100 @@ internal static bool SpecifiesModelType(in SymbolCache symbolCache, IParameterSy
...
+                    return false;
+                }
+
+
+                if (!TryGetType(SymbolNames.ControllerAttribute, out var controllerAttribute))
+                {
+                    return false;
+                }
+
+
```
rynowak: @JamesNK

src/Mvc/Mvc.Analyzers/src/TopLevelParameterNameAnalyzer.cs
```diff
@@ -185,20 +183,100 @@ internal static bool SpecifiesModelType(in SymbolCache symbolCache, IParameterSy
...
+            public static bool TryCreate(Compilation compilation, out SymbolCache symbolCache)
+            {
+                symbolCache = default;
+
+                if (!TryGetType(SymbolNames.BindAttribute, out var bindAttribute))
+                {
+                    return false;
+                }
+
+
```
rynowak: @JamesNK

src/Mvc/Mvc.Api.Analyzers/src/AddResponseTypeAttributeCodeFixAction.cs
```diff
@@ -43,7 +44,14 @@ public AddResponseTypeAttributeCodeFixAction(Document document, Diagnostic diagn
...
         protected override async Task<Document> GetChangedDocumentAsync(CancellationToken cancellationToken)
         {
-            var context = await CreateCodeActionContext(cancellationToken).ConfigureAwait(false);
+            var nullableContext = await CreateCodeActionContext(cancellationToken).ConfigureAwait(false);
+            if (nullableContext == null)
+            {
+                return _document;
+            }
+
+            var context = nullableContext.Value;
```
rynowak: FYI this shouldn't be needed. The flow analysis should do the right thing after it sees the null check at L48.

rynowak: Example: https://github.com/aspnet/AspNetCore/pull/10375/files#diff-51adbcdc12291a2342784a001a9f502bR57

src/Mvc/Mvc.Core/src/Infrastructure/ActionResultStatusCodeAttribute.cs
rynowak: I don't think that's necessary. These should probably just be public.

rynowak (approved):

src/Mvc/Mvc.Analyzers/src/TopLevelParameterNameAnalyzer.cs
JamesNK: image

JamesNK (approved):

src/Mvc/Mvc.Analyzers/src/TopLevelParameterNameAnalyzer.cs
pranavkm: Why is aquaman sad?

JamesNK: Thanos collected the one ring and now aquaman has to leave Hogwarts. Poor aquaman 😢

src/Mvc/Mvc.Api.Analyzers/src/AddResponseTypeAttributeCodeFixAction.cs
pranavkm:
```C#
error CS1061: 'AddResponseTypeAttributeCodeFixAction.CodeActionContext?' does not contain a definition for 'SymbolCache' and no accessible extension method 'SymbolCache' accepting a first argument of type 'AddResponseTypeAttributeCodeFixAction.CodeActionContext?' could be found
```

Doesn't seem to work if I remove it. Perhaps because it's a value type.

src/Mvc/Mvc.Analyzers/src/TopLevelParameterNameAnalyzer.cs
pranavkm: I'll clean up the extra new lines in a follow up.

PR #10385, opened 2019-05-20
fix: increase z-index for reconnect modal

Stamo-Gochev:
Use the maximum value for z-index as I think it should be shown above any other custom modal (e.g. from a 3rd party library that can have randomly big z-index value).


Stamo-Gochev:
@javiercn Can you provide some guidelines how to fix the error in:
https://dev.azure.com/dnceng/public/_build/results?buildId=194153

I don't see how this is connected with the change, so probably something else is causing the problem.

javiercn:
The change looks good to me. But there are a couple of things I'm not sure about:
* Do we have to make it the maximum or just some very high value. If not people can run in the opposite scenario where this dialog hides some critical functionality. I don't think this is likely, but is good to think about it.
* Could this be fixed in the application with just a custom CSS selector? Alternatively, you could (and likely should) customize the UI for your app so that the style is inline with the appearance of the rest of your app.

@SteveSandersonMS thoughts?

Stamo-Gochev:
Adding my thoughts on the question - the idea of the maximum possible value is that showing the reconnect modal means there is a critical problem with the app - the user is no longer to interact with the page in a blazor-related way. I don't think there can be anything more critical that is displayed by a third-party modal. 

I also don't think that adding a custom selector will work if you don't know that this modal exists (although adding the CSS rule will work). The way I hit this problem is that I had an element that had a higher z-index and the reconnect modal wasn't visible at all. This means that developers can deploy their app and experience a crash without an easy to spot message why this is happening. I was able to find the problem by inspecting the dev tools console, but for end users, they will probably report a "crash"without any error message, which is hard to investigate.

Stamo-Gochev:
@SteveSandersonMS Any update on the suggestion?

SteveSandersonMS:
It's a reasonable suggestion. Just FYI we're already at max capacity for delivering committed work items for 3.0, so anything not already committed will have to happen after 3.0 (including incoming PRs). Hope you understand!

mkArtakMSFT:
Given that we haven't heard similar complaints about the current z-index value, we've decided not to move forward with this change. An easy alternative is to override the `z-index` in your own CSS rule.

PR #10394, opened 2019-05-20
Add InputPassword form component. Fixes #9179

### LINKED ISSUE #9179: Razor Components feature request: Microsoft.AspNetCore.Components.Forms.InputPassword
### Is your feature request related to a problem? Please describe.

I am trying to use `Microsoft.AspNetCore.Components.Forms.InputText` with `type="password"`.  As far as I can tell though there is no way to add `type="password"`

1. Allow specifying `type` in `Microsoft.AspNetCore.Components.Forms.InputText`

or

2. Add a  `Microsoft.AspNetCore.Components.Forms.InputPassword` class

### Describe alternatives you've considered

For now I plan to just extend `Microsoft.AspNetCore.Components.Forms.InputText` and add `type="password"`.  But this seems like something a lot of apps would need, so it'd be nice to put it in the standard library.

### Other info

Currently using .NET Core 3.0 Preview 3

SteveSandersonMS:


src/Components/Components/src/Forms/InputComponents/InputPassword.cs
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Components.RenderTree;
+
+namespace Microsoft.AspNetCore.Components.Forms
+{
```
SteveSandersonMS: This could be implemented as a `.razor` file now, but I'm keeping it as `.cs` for consistency with the others until we get round to updating them all.

src/Components/Components/src/Forms/InputComponents/InputPassword.cs
```diff
@@ -0,0 +1,25 @@
...
+    public class InputPassword : InputText
+    {
+        /// <inheritdoc />
+        protected override void BuildRenderTree(RenderTreeBuilder builder)
+        {
+            builder.OpenElement(0, "input");
+            builder.AddAttribute(1, "id", Id);
+            builder.AddAttribute(2, "class", CssClass);
+            builder.AddAttribute(3, "value", BindMethods.GetValue(CurrentValue));
+            builder.AddAttribute(4, "onchange", BindMethods.SetValueHandler(__value => CurrentValue = __value, CurrentValue));
```
rynowak: Should this be `EventCallbackFactory`?

rynowak: As part of the API review I'd like to audit all of the things the compiler users. We still have `BindMethods` sitting around, but I think it only gets used now for the `GetValue` stuff.

SteveSandersonMS: The code here is identical to the equivalent in `InputText`. If `EventCallbackFactory` is required, we should update all the forms components. If so let's do that separately from this PR.

rynowak: sure!

rynowak (approved):

RemiBou:
Maybe we could add a InputText with type (it can be an enum) attribute instead ? I did this here https://github.com/RemiBou/Toss.Blazor/blob/master/Toss.Client/Shared/Forms/MyInput.razor. I don't see the added value of a specific class.

SteveSandersonMS:
That's a good point, @RemiBou. We have another feature being built at the moment that will allow arbitrary attribute values to be passed through, so the same thing could be achieved using:

```cshtml
<InputText type="password" bind-Value="@SomeProperty" />
```

@rynowak @danroth27 Do you have strong opinions on whether `<InputPassword />` should also exist as a standalone thing? My preference would be not to expand the set of built-in primitives when there's (about to be) such a simple and discoverable way to make an exactly-equivalent password field.

danroth27:
> Do you have strong opinions on whether <InputPassword /> should also exist as a standalone thing?

I'm fine with the more minimalistic approach.

javiercn:
I'm fine with not adding InputPassword

SteveSandersonMS:
Great, closing as not required.

Issue #5947, opened 2018-11-14
Expose `TestServer` as public property on `RequestBuilder`


carusology:
In order to make our integration tests cleaner, I've been all about adding extension methods to the [`RequestBuilder`](https://github.com/aspnet/Hosting/blob/2.1.1/src/Microsoft.AspNetCore.TestHost/RequestBuilder.cs). For example, to prevent issues around everyone learning for the first time that [HTTP methods are case-sensitive](https://tools.ietf.org/html/rfc2616#section-5.1.1), we can add a simple extension to make sure developers use method casing consistently.

```cs
public static class RequestBuilderExtensions {

    public static Task<HttpResponseMessage> PatchAsync(this RequestBuilder request) {
        if (request == null) {
            throw new ArgumentNullException(nameof(request));
        }

        return request.SendAsync("PATCH");
    }

    public static Task<HttpResponseMessage> PutAsync(this RequestBuilder request) {
        if (request == null) {
            throw new ArgumentNullException(nameof(request));
        }

        return request.SendAsync("PUT");
    }

    public static Task<HttpResponseMessage> DeleteAsync(this RequestBuilder request) {
        if (request == null) {
            throw new ArgumentNullException(nameof(request));
        }

        return request.SendAsync("DELETE");
    }

}
```

These are easy enough to do when the state of the `TestServer` is irrelevant for the construction of the `HttpRequestMessage`, but that's not always the case. For example, what if I want to create an extension method that sets a `Bearer` token in the `Authorization` header for a given `User`.

Here's an example xUnit test:

```cs
[Fact]
public void GetResource_ValidUser() {

    // Arrange

    var path = "path/to/resource";
    var user = new User { EmailAddress = "me@example.com", Id = "i-am-unique" };
    var request = this.TestServer.CreateRequest(path).SetUser(user);

    // Act

    var response = await request.GetAsync();

    // Assert

    Assert.Equal(HttpStatusCode.OK, response.StatusCode);

}
```

In order to create a valid `Bearer` token, I need to be able to fetch the authentication configuration that is currently being used within the `TestServer`, but because `RequestBuilder` doesn't expose it, I can't do this. Here's what I _want_ to do:

```cs
public static class RequestBuilderExtensions {

    public static RequestBuilder SetUser(this RequestBuilder request, User user) {
        if (user == null) {
            request = request.And(message => message.Headers.Authorization = null);
        } else {
            // "request.TestServer" property does not exist at this time.
[...]

PR #10396, opened 2019-05-20
Expose TestServer as public property on RequestBuilder

huysentruitw:
Expose TestServer as public property on RequestBuilder.

Addresses #5947

### DIFF
--- a/src/Hosting/TestHost/ref/Microsoft.AspNetCore.TestHost.netcoreapp3.0.cs
+++ b/src/Hosting/TestHost/ref/Microsoft.AspNetCore.TestHost.netcoreapp3.0.cs
@@ -28,8 +28,8 @@ namespace Microsoft.AspNetCore.TestHost
     {
         public TestServer(Microsoft.AspNetCore.Hosting.IWebHostBuilder builder) { }
         public TestServer(Microsoft.AspNetCore.Hosting.IWebHostBuilder builder, Microsoft.AspNetCore.Http.Features.IFeatureCollection featureCollection) { }
-        public TestServer(Microsoft.AspNetCore.Http.Features.IFeatureCollection featureCollection, System.IServiceProvider services) { }
         public TestServer(System.IServiceProvider services) { }
+        public TestServer(System.IServiceProvider services, Microsoft.AspNetCore.Http.Features.IFeatureCollection featureCollection) { }
         public bool AllowSynchronousIO { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public System.Uri BaseAddress { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public Microsoft.AspNetCore.Http.Features.IFeatureCollection Features { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
--- a/src/Hosting/TestHost/ref/Microsoft.AspNetCore.TestHost.netcoreapp3.0.cs
+++ b/src/Hosting/TestHost/ref/Microsoft.AspNetCore.TestHost.netcoreapp3.0.cs
@@ -29,12 +29,14 @@ namespace Microsoft.AspNetCore.TestHost
         public TestServer() { }
         public TestServer(Microsoft.AspNetCore.Hosting.IWebHostBuilder builder) { }
         public TestServer(Microsoft.AspNetCore.Hosting.IWebHostBuilder builder, Microsoft.AspNetCore.Http.Features.IFeatureCollection featureCollection) { }
-        public TestServer(Microsoft.AspNetCore.Http.Features.IFeatureCollection featureCollection) { }
+        public TestServer(Microsoft.AspNetCore.Http.Features.IFeatureCollection featureCollection, System.IServiceProvider services) { }
+        public TestServer(System.IServiceProvider services) { }
         public bool AllowSynchronousIO { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public System.Uri BaseAddress { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public Microsoft.AspNetCore.Http.Features.IFeatureCollection Features { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
         public Microsoft.AspNetCore.Hosting.IWebHost Host { get { throw null; } }
         public bool PreserveExecutionContext { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
+        public System.IServiceProvider Services { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
         public System.Net.Http.HttpClient CreateClient() { throw null; }
         public System.Net.Http.HttpMessageHandler CreateHandler() { throw null; }
         public Microsoft.AspNetCore.TestHost.RequestBuilder CreateRequest(string path) { throw null; }
--- a/src/Hosting/TestHost/ref/Microsoft.AspNetCore.TestHost.netcoreapp3.0.cs
+++ b/src/Hosting/TestHost/ref/Microsoft.AspNetCore.TestHost.netcoreapp3.0.cs
@@ -26,7 +26,6 @@ namespace Microsoft.AspNetCore.TestHost
     }
     public partial class TestServer : Microsoft.AspNetCore.Hosting.Server.IServer, System.IDisposable
     {
-        public TestServer() { }
         public TestServer(Microsoft.AspNetCore.Hosting.IWebHostBuilder builder) { }
         public TestServer(Microsoft.AspNetCore.Hosting.IWebHostBuilder builder, Microsoft.AspNetCore.Http.Features.IFeatureCollection featureCollection) { }
         public TestServer(Microsoft.AspNetCore.Http.Features.IFeatureCollection featureCollection, System.IServiceProvider services) { }
--- a/src/Hosting/TestHost/ref/Microsoft.AspNetCore.TestHost.netcoreapp3.0.cs
+++ b/src/Hosting/TestHost/ref/Microsoft.AspNetCore.TestHost.netcoreapp3.0.cs
@@ -17,6 +17,7 @@ namespace Microsoft.AspNetCore.TestHost
     public partial class RequestBuilder
     {
         public RequestBuilder(Microsoft.AspNetCore.TestHost.TestServer server, string path) { }
+        public Microsoft.AspNetCore.TestHost.TestServer TestServer { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
         public Microsoft.AspNetCore.TestHost.RequestBuilder AddHeader(string name, string value) { throw null; }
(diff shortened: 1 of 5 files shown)


Tratcher:
The code seems fine, but there's a problem with the intent. You did this so you could get access to the IWebHost and IServiceCollection. However, when we [refactored](https://github.com/aspnet/AspNetCore/commit/241d2c13df468d18149f2245efb0d6dbcfc7a60d) TestServer for generic host it no longer holds a reference to either of those.

You might need to re-think the larger design here.

carusology:
Thanks @huysentruitw for picking up this issue!

@Tratcher, I'm not sure I follow your comment.

When I look at [the changeset you linked to](https://github.com/aspnet/AspNetCore/commit/241d2c13df468d18149f2245efb0d6dbcfc7a60d#diff-221e453471a4add687f84426a987a1b2R69), I see that [there's still an `public IWebHost Host` property on L69](https://github.com/aspnet/AspNetCore/blob/241d2c13df468d18149f2245efb0d6dbcfc7a60d/src/Microsoft.AspNetCore.TestHost/TestServer.cs#L69) both after that changeset as well as [in the latest pre-release of v3.0](https://github.com/aspnet/AspNetCore/blob/5502cf46706a25c39ce78a57b9191714c0eb40c2/src/Hosting/TestHost/src/TestServer.cs#L69). When I look at [the latest pre-release of the `IWebHost` definition](https://github.com/aspnet/AspNetCore/blob/5502cf46706a25c39ce78a57b9191714c0eb40c2/src/Hosting/Abstractions/src/IWebHost.cs#L14), it still has a reference to the `IServiceProvider`.

With this changeset, this should complete the chain from `RequestBuilder` => `TestServer` => `IWebHost` => `IServiceProvider`. Is there something I'm missing? Are the constructors on `TestServer` that allow the injection of an `IWebHostBuilder`, and therefore a reference to its `IWebHost`, going to go away?

Tratcher:
New 3.0 apps don't use IWebHost anymore, they've moved to IHost instead (which doesn't have a property on TestServer due to the redesign).

The new 3.0 templates:
https://github.com/aspnet/AspNetCore/blob/master/src/ProjectTemplates/Web.ProjectTemplates/content/WebApi-CSharp/Program.cs

carusology:
Are [the current v3.0 preview integration docs](https://docs.microsoft.com/en-us/aspnet/core/test/integration-tests?view=aspnetcore-3.0) reflective of how we should be writing integration tests using the `TestServer` in ASP.NET Core v3? Assuming so, as long as we do not port our v2.X web application's `TEntrypoint` to use `HostBuilder` over `WebHostBuilder`, we will still have access to [the `TestServer` via the public property on `WebApplicationFactory`](https://github.com/aspnet/AspNetCore/blob/0f0a388eab7e9bcb67a1db8911246e5fb1aa00f7/src/Mvc/Mvc.Testing/src/WebApplicationFactory.cs#L73), and [the `TestServer` will still be hydrated with an `IWebHostBuilder`](https://github.com/aspnet/AspNetCore/blob/0f0a388eab7e9bcb67a1db8911246e5fb1aa00f7/src/Mvc/Mvc.Testing/src/WebApplicationFactory.cs#L138), and we'll continue to be able to `RequestBuilder` that is being mutated as part of this merge request, right? That certainly seems preferable to porting all of our existing integration tests to use `HttpClient` as well.

Assuming my above reading is correct, is there some benefit to using `HttpClient` over `RequestBuilder` to perform integration tests that we would be missing out on?

Tratcher:
The WebApplicationFactory has been updated to work with [either](https://github.com/aspnet/AspNetCore/blob/26b53236b1e694090d6fb67bc37813ba4c393c34/src/Mvc/Mvc.Testing/src/WebApplicationFactory.cs#L90) IHost or IWebHost. It's useful for testing larger components, especially full MVC apps.

TestServer is designed for unit testing individual middleware.

> is there some benefit to using HttpClient over RequestBuilder

HttpClient has an ecosystem of libraries built around it for crafting requests and reading responses. That can be helpful for some scenarios. RequestBuilder is more helpful when you need to seed some of the server side structures directly.

When what you're really after here is the services, we should revisit making those available somehow on TestServer and RequestBuilder. We don't need the IHost to do that, we could inject the IServiceCollection into the constructor.

huysentruitw:
Anyone fancy to review my updates from 24 days ago? 😁

src/Hosting/TestHost/ref/Microsoft.AspNetCore.TestHost.netcoreapp3.0.cs
```diff
@@ -28,12 +29,14 @@ public partial class TestServer : Microsoft.AspNetCore.Hosting.Server.IServer, S
         public TestServer() { }
         public TestServer(Microsoft.AspNetCore.Hosting.IWebHostBuilder builder) { }
         public TestServer(Microsoft.AspNetCore.Hosting.IWebHostBuilder builder, Microsoft.AspNetCore.Http.Features.IFeatureCollection featureCollection) { }
-        public TestServer(Microsoft.AspNetCore.Http.Features.IFeatureCollection featureCollection) { }
```
Tratcher: We need to keep all of the old constructors for back compat. Add this overload back.

src/Hosting/TestHost/test/TestServerTests.cs
```diff
@@ -253,6 +252,32 @@ public void TestServerConstructorWithNullFeatureCollectionThrows()
...
+        {
+            // Arrange & Act
+            var testServer = new TestServer();
+
+            // Assert
+            Assert.NotNull(testServer.Services);
+        }
+
+        [Fact]
+        public void TestServerConstructorShouldProvideServicesFromWebHost()
```
Tratcher: Add a variant for HostBuilder.

Tratcher (approved):
Sorry, hadn't noticed the updates. Looks better this time.

src/Hosting/TestHost/src/TestServer.cs
```diff
@@ -21,20 +21,31 @@ public class TestServer : IServer
         private IHttpApplication<Context> _application;
 
         /// <summary>
-        /// For use with IHostBuilder or IWebHostBuilder.
+        /// For use with IHostBuilder.
         /// </summary>
         public TestServer()
-            : this(new FeatureCollection())
+            : this(new ServiceCollection().BuildServiceProvider())
```
davidfowl: Whats the point of this constructor?

src/Hosting/TestHost/ref/Microsoft.AspNetCore.TestHost.netcoreapp3.0.cs
```diff
@@ -28,12 +29,14 @@ public partial class TestServer : Microsoft.AspNetCore.Hosting.Server.IServer, S
...
+        public TestServer(Microsoft.AspNetCore.Http.Features.IFeatureCollection featureCollection, System.IServiceProvider services) { }
+        public TestServer(System.IServiceProvider services) { }
```
davidfowl: Whats the scenario for this?

src/Hosting/TestHost/src/TestServer.cs
huysentruitw: I'm not sure if it can be removed. The parameter-less constructor already existed before this PR.

davidfowl: Where is the constructor with the IServiceProvider used?

Tratcher: This was the one used by DI in the HostBuilder scenario. We could obsolete it with a warning if you like @davidfowl?

Tratcher: The IServiceProvider ctor is the new one used by DI in the HostBuilder scenario.

Tratcher: @huysentruitw don't wait for David, he's traveling. We can obsolete old things later if we need to.

davidfowl: We should obsolete them if they aren't useful anymore (or don't work), otherwise it's fine keeping it.

Tratcher: Looking back at 2.2, these weren't even there, they were added in 3.0 for HostBuilder. Ok, we can remove them.
https://github.com/aspnet/AspNetCore/blob/release/2.2/src/Hosting/TestHost/src/TestServer.cs

I'll comment on the individual APIs to remove.

src/Hosting/TestHost/ref/Microsoft.AspNetCore.TestHost.netcoreapp3.0.cs
```diff
@@ -28,12 +29,14 @@ public partial class TestServer : Microsoft.AspNetCore.Hosting.Server.IServer, S
         public TestServer() { }
```
Tratcher: `TestServer()` can be removed, it wasn't present in 2.2.

src/Hosting/TestHost/ref/Microsoft.AspNetCore.TestHost.netcoreapp3.0.cs
Tratcher: Nevermind, this one doesn't need to come back, it wasn't present in 2.2 either.

Tratcher:
@huysentruitw were you going to be able to finish this up?

huysentruitw:
@Tratcher I have a problem with [this test](https://github.com/aspnet/AspNetCore/pull/10396/files#diff-a7375c1f860702ceffe950efc8d85398R27) when removing the parameter-less constructor for `TestServer`.

Now I need to pass a `serviceProvider` to `new TestServer(serviceProvider)`, but it feels a bit weird to pass an empty service provider, while the `.UseServer` extension method adds the test server to an other service collection...

I think we need to remove this test as it probably demonstrates a usage we don't want to support. Or we could change the `UseServer` extension method so it takes a lambda instead, so we can do:

`UseServer(serviceProvider => new TestServer(serviceProvider));`

What do you think?

Tratcher:
@huysentruitw Replace the UseServer call with `AddSingleton<IServer>(services =>  new TestServer(serviceProvider));`

Tratcher (approved):

src/Hosting/TestHost/test/TestServerTests.cs
```diff
@@ -278,6 +280,58 @@ public void TestServerConstructorWithNullFeatureCollectionThrows()
...
+            // Assert
+            Assert.Equal(testService, testServer.Services.GetService<TestService>());
+        }
+
+        [Fact]
+        public async Task TestServerConstructorShouldProvideServicesFromHostBuilder()
+        {
+            // Arrange
+            var testService = new TestService();
+            var host = await new HostBuilder()
```
Tratcher: You're hitting an issue where not disposing the IHost can cause the tests to hang. We're tracking that over at  https://github.com/aspnet/Extensions/issues/1363
```suggestion
            using var host = await new HostBuilder()
```

src/Hosting/TestHost/ref/Microsoft.AspNetCore.TestHost.netcoreapp3.0.cs
```diff
@@ -25,15 +26,16 @@ public partial class RequestBuilder
     }
     public partial class TestServer : Microsoft.AspNetCore.Hosting.Server.IServer, System.IDisposable
     {
-        public TestServer() { }
         public TestServer(Microsoft.AspNetCore.Hosting.IWebHostBuilder builder) { }
         public TestServer(Microsoft.AspNetCore.Hosting.IWebHostBuilder builder, Microsoft.AspNetCore.Http.Features.IFeatureCollection featureCollection) { }
-        public TestServer(Microsoft.AspNetCore.Http.Features.IFeatureCollection featureCollection) { }
+        public TestServer(Microsoft.AspNetCore.Http.Features.IFeatureCollection featureCollection, System.IServiceProvider services) { }
```
Tratcher: Inconsistent parameter ordering
```suggestion
        public TestServer(System.IServiceProvider services, Microsoft.AspNetCore.Http.Features.IFeatureCollection featureCollection) { }
```

Tratcher:
Thanks

carusology:
> Thanks

No, **thank YOU.** This is a wonderful addition to `TestServer` that enables fluent API-styled extension methods.

I'm looking even more forward to ASP.NET Core v3!

h/t @Tratcher @huysentruitw

PR #10405, opened 2019-05-20
Initialize eng/common/ folder from dotnet/arcade

natemcmaster:
Part of #7280 (Convert to Arcade SDK)

Arcade currently requires that you copy these 5,694 lines of code into your repo to use it.

Reviewers: there is no point in reviewing the contents of eng/common/ as we're supposed to mirror this code exactly from https://github.com/dotnet/arcade/. The only non eng/common/ change in this PR is an update to CODEOWNERS

cc @markwilkie


JunTaoLuo (approved):
That's a lotta boiler plate.

.github/CODEOWNERS
```diff
@@ -6,6 +6,7 @@
 /.config/                       @aspnet/build
 /build/                         @aspnet/build
 /eng/                           @aspnet/build
+/eng/common/                    @dotnet-maestro-bot
```
dougbu: Suggest including my, myself or I for the same reason I'm included on the next two lines

PR #10410, opened 2019-05-20
Put request trailers in a separate collection

Tratcher:
#4771 Having request headers and trailers in the same collection creates some ambiguity and potential spoofing concerns (e.g. headers changing after validation). In the HTTP/2 scenario there are also concurrency issues where all headers and trailers are processed on the connection thread, not the request thread. This PR adds a new feature and collection for trailers to keep them isolated. 

I ended up using a bool Available to track state rather than returning a null connection because it let me use the main collection property to store incomplete trailers without worrying if anyone would access them (e.g. HTTP/2 concurrency issues). It also simplified re-using the collection across requests, it can just be cleared out.

The most interesting detail is that Http/2 trailers aren't processed as part of the request body like HTTP/1.1. The connection loop reads them as soon as they arrive. That means you may not have to read all the way to the end of the request body before they're available, only far enough that the remainder fit in the buffer and the client's window size unblocked the sending of trailers. Consistency with HTTP/1.1 is maintained by ensuring that the trailers are always available if you have read to the end of the request body.

### DIFF
--- a/src/Http/Http.Abstractions/src/Extensions/RequestTrailerExtensions.cs
+++ b/src/Http/Http.Abstractions/src/Extensions/RequestTrailerExtensions.cs
@@ -13,9 +13,14 @@ namespace Microsoft.AspNetCore.Http
     /// </summary>
     public static class RequestTrailerExtensions
     {
+        /// <summary>
+        /// Gets the request "Trailer" header that lists which trailers to expect after the body.
+        /// </summary>
+        /// <param name="request"></param>
+        /// <returns></returns>
         public static StringValues GetDeclaredTrailers(this HttpRequest request)
         {
-            return request.Headers[HeaderNames.Trailer];
+            return request.Headers.GetCommaSeparatedValues(HeaderNames.Trailer);
         }
 
         /// <summary>
@@ -36,7 +41,7 @@ namespace Microsoft.AspNetCore.Http
         /// <returns></returns>
         public static bool CheckTrailersAvailable(this HttpRequest request)
         {
-            return request.HttpContext.Features.Get<IHttpRequestTrailersFeature>()?.Trailers != null;
+            return request.HttpContext.Features.Get<IHttpRequestTrailersFeature>()?.Available == true;
         }
 
         /// <summary>
@@ -51,11 +56,7 @@ namespace Microsoft.AspNetCore.Http
             var feature = request.HttpContext.Features.Get<IHttpRequestTrailersFeature>();
             if (feature == null)
             {
-                throw new NotSupportedException("This server does not support request trailers.");
-            }
-            if (feature.Trailers == null)
-            {
-                throw new InvalidOperationException("The trailers are not available yet. Did you finish reading the request body?");
+                throw new NotSupportedException("This request does not support trailers.");
             }
 
             return feature.Trailers[trailerName];
--- a/src/Http/Http.Abstractions/src/Extensions/RequestTrailerExtensions.cs
+++ b/src/Http/Http.Abstractions/src/Extensions/RequestTrailerExtensions.cs
@@ -24,7 +24,7 @@ namespace Microsoft.AspNetCore.Http
         }
 
         /// <summary>
-        /// Indicates if the server supports receiving trailer headers.
+        /// Indicates if the request supports receiving trailer headers.
         /// </summary>
         /// <param name="request"></param>
         /// <returns></returns>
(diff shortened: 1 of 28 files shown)


src/Http/Http.Features/src/IHttpRequestTrailersFeature.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// <summary>
+    /// This feature exposes HTTP request trailer headers, either for HTTP/1.1 chunked bodies or HTTP/2 trailing headers.
+    /// </summary>
+    public interface IHttpRequestTrailersFeature
+    {
+        /// <summary>
+        /// Indicates if the <see cref="Trailers"/> are available yet. They may not be available until the
+        /// request body is fully read.
+        /// </summary>
+        bool Available { get; }
```
halter73: Would this be better as a Task? It could be lazy initialized.

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
```diff
@@ -524,11 +528,35 @@ public void OnHeader(Span<byte> name, Span<byte> value)
...
+            // Trailers still count towards the limit.
+            _requestHeadersParsed++;
+            if (_requestHeadersParsed > ServerOptions.Limits.MaxRequestHeaderCount)
+            {
+                BadHttpRequestException.Throw(RequestRejectionReason.TooManyHeaders);
+            }
+
+            string key = name.GetHeaderName();
+            var valueStr = value.GetAsciiOrUTF8StringNonNullCharacters();
+            RequestTrailers.Append(key, valueStr);
```
halter73: Why not pass the spans directly to Append like OnHeader?

src/Http/Http.Features/src/IHttpRequestTrailersFeature.cs
Tratcher: A Task isn't a good fit, just waiting for it will rarely be enough, you'll deadlock yourself. If it's false then you need to take action (read the request body) before it will become true.

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
Tratcher: RequestTrailers is a vanilla IHeaderDictionary, not a custom collection. It's not clear that most of the header optimizations we've done elsewhere are relevant to request trailers.

src/Http/Http.Features/src/IHttpRequestTrailersFeature.cs
halter73: I now remember this discussion. It's been too long.

halter73 (approved):

Tratcher:
@halter73 updated to always support request trailers.

src/Servers/Kestrel/tools/CodeGenerator/HttpProtocolFeatureCollection.cs
```diff
@@ -26,6 +26,7 @@ public static string GenerateFile()
             var commonFeatures = new[]
             {
                 "IHttpAuthenticationFeature",
+                "IHttpRequestTrailersFeature",
```
halter73: I think all the `implementedFeatures` should also be `alwaysFeatures` instead of `commonFeatures`.

src/Http/Http.Abstractions/src/Extensions/RequestTrailerExtensions.cs
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        /// <summary>
+        /// Indicates if the request supports receiving trailer headers.
+        /// </summary>
+        /// <param name="request"></param>
+        /// <returns></returns>
+        public static bool SupportsTrailers(this HttpRequest request)
+        {
+            return request.HttpContext.Features.Get<IHttpRequestTrailersFeature>() != null;
```
halter73: This is now true even for requests that don't support request bodies and HTTP/1.1 Content-Length requeusts too?

I don't feel too strongly about this, but maybe `SupportsTrailers` should be a property on IHttpRequestTrailersFeature like `Available` is. That way SupportsTrailers could still be false despite IHttpRequestTrailersFeature despite being present on every request in Kestrel.

That being said, I think this is OK as is for now. We can add a default interface method later if customers complain the SupportsTrailers() extension method returns true too often.

halter73 (approved):
Approved again.

Issue #10250, opened 2019-05-15
Add EnumeratorCancellation attribute to CancellationTokens in async iterators


anurse:
Anywhere we have an async iterator (see below) and take a `CancellationToken` as a parameter, we need to apply the `[EnumeratorCancellation]` attribute to the `CancellationToken` parameter to avoid `CS8425`.

```csharp
public async IAsyncEnumerable<string> GetStringsAsync(
		[EnumeratorCancellation] CancellationToken cancellationToken)
{
	yield break;
}
```

SignalR does this and it triggered an issue in CTI validation (specifically the tests that validate that the latest VS can build our stuff): https://devdiv.visualstudio.com/DevDiv/_workitems/edit/892482?src=WorkItemMention&src-action=artifact_link

It looks like [`EnumeratorCancellationAttribute`](https://github.com/dotnet/corefx/blob/master/src/System.Runtime/src/System/Runtime/CompilerServices/EnumeratorCancellationAttribute.cs) was added 24 days ago so we should be able to depend on it now.

**NOTE** It appears the warning *only repros* in VS 16.2 pre-release builds.

BrennanConroy:
The warning is a really annoying "feature" when someone writes an `IAsyncEnumerable<T>` hub method that accepts a `CancellationToken` (which means SignalR will create a CTS internally for the hub method and cancel it appropriately).

So now customers will be forced to add the attribute or ignore the warning.

anurse:
Yep, that's more of a question on how the C# compiler wants to handle this though. I think our best way forward is to react and see how customers feel about it.

BrennanConroy:
Done via https://github.com/aspnet/AspNetCore/pull/10411

BrennanConroy:
**Acceptance checklist** (check one item)
* [ ] We decided not to take this fix.
* [ ] The fix is tests-only.
* [x] The fix contains product changes (**check all items below**).
    * [x] Relevant XML documentation comments for new public APIs are present.
    * [x] Narrative docs (docs.microsoft.com) are updated. (**check one item below**)
        * [ ] The change requires a new article. An issue with an outline has been filed here: [ISSUE LINK]
        * [ ] The change requires a change to an existing article. A **closed** docs PR with these changes is linked here: [PR LINK]
        * [x] The change requires no docs changes.
    * [x] Verification has been completed. (**check one item below**)
        * [x] The change is in the shared framework and was verified against the following versions 
[...]

PR #10411, opened 2019-05-20
Fix build failures with new VS

BrennanConroy:
Can't build locally anymore because of this.

Not sure this "fixes" https://github.com/aspnet/AspNetCore/issues/10250 because needing the attribute on the hub method is pretty ugly.


anurse:
This will fix the main issue in #10250, so we should close it when this PR closes. Are you concerned about Hub methods needing the attribute? They only need it if they choose to have a CT and it's a warning by default (so it depends on how they've configured their build).

analogrelay (approved):

BrennanConroy:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=194806

PR #10413, opened 2019-05-21
Dylan/request throttle

DylanDmitri:
- new middleware to limit concurrent active requests 
 - extra requests are buffered server-side behind a semaphore

Initial behavior is here. Future additions include:
 - limit on queue length (start rejecting requests after a point)
 - benchmarks
 - more tests (and tests with TestServer)
 - better sample (feature to send X requests at once)


src/Middleware/RequestThrottling/src/RequestThrottlingExtensions.cs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Aspnetcore.RequestThrottling;
+
+namespace Microsoft.AspNetCore.Builder
+{
+    public static class RequestThrottlingExtensions
```
Tratcher: Doc comments for all public classes and members, even if they say obvious things. You should be seeing warnings/errors about these missing.

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
```diff
@@ -0,0 +1,39 @@
...
+            _options = options.Value;
+            _semaphore = new SemaphoreWrapper(_options.MaxConcurrentRequests); 
+        }
+
+        public async Task Invoke(HttpContext context)
+        {
+            await _semaphore.EnterQueue();
+            _logger.LogDebug("Entered Queue");
+
+            await _next(context);
```
Tratcher: put the call to next in a try block and LeaveQueue in a finally block. That way you don't leak anything if next throws an Exception. This would be a good test case.

src/Middleware/RequestThrottling/test/MiddlewareTests.cs
```diff
@@ -0,0 +1,62 @@
...
+            var blocker = new TaskCompletionSource<string>(TaskCreationOptions.RunContinuationsAsynchronously);
+            var firstRequest = true;
+
+            var middleware = TestUtils.CreateTestMiddleWare(
+                maxConcurrentRequests: 1,
+                next: httpContext =>
+                {
+                    if (firstRequest)
+                    {
+                        firstRequest = false;
```
Tratcher: You can use another TaskCompletionSource to ensure the first request was received before sending the second. You can await it after getting t1 below, and complete it here after setting firstRequest.

If you had enough async operations going on at once you wouldn't be able to guarantee that t1 and t2 didn't get re-ordered. I don't expect such in a simple case like this but it's better to have a positive signal from the app that the first request was received before sending the second.

Tratcher (changes requested):

src/Middleware/RequestThrottling/src/RequestThrottlingExtensions.cs
anurse: I believe most of the other middleware have example messages you could borrow from.

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
anurse: Or `EnterQueue` could return an `IDisposable` that leaves the queue when disposed. Then you can use `using`.

src/Middleware/RequestThrottling/test/MiddlewareTests.cs
anurse: This looks like a job for... [`SyncPoint`](https://github.com/aspnet/AspNetCore/blob/master/src/SignalR/common/testassets/Tests.Utils/SyncPoint.cs)!

An animated superman flying through the sky

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
```diff
@@ -0,0 +1,39 @@
...
+
+            _semaphore.LeaveQueue();
+            _logger.LogDebug($"request finished, semaphore count at: {_semaphore.Count}");
```
Eilon: Today's lesson: Structured logging! Your buddies will teach you all about it.

src/Middleware/RequestThrottling/src/RequestThrottlingOptions.cs
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.RequestThrottling
+{
+    public class RequestThrottlingOptions
+    {
+        /// <summary>
+        /// Maximum number of concurrent requests. Any extras will be queued on the server.
```
Eilon: Do we usually document the value of the default value? I think we do in some places, and I rather like it.

src/Middleware/RequestThrottling/src/SemaphoreWrapper.cs
```diff
@@ -1,12 +1,10 @@
 using System;
-using System.Collections.Generic;
-using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 
 namespace Microsoft.AspNetCore.RequestThrottling
 {
-    internal class SemaphoreWrapper : IDisposable
+    public class SemaphoreWrapper : IDisposable
```
Eilon: Is this meant to be public? Do we want this API surface?

src/Middleware/RequestThrottling/src/SemaphoreWrapper.cs
```diff
@@ -1,12 +1,10 @@
 using System;
```
Eilon: Missing copyright/license header here (elsewhere too?).

Eilon (changes requested):

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
DAllanCarr:
```suggestion
            _logger.LogDebug("Request finished, semaphore count at: {semaphoreCount}", _semaphore.Count);
```

anurse: We should turn on the logging analyzers here if possible. They should give you a squiggle here for using string interpolation. (cc @halter73)

src/Middleware/RequestThrottling/src/RequestThrottlingOptions.cs
anurse: Agreed, let's put the default value in.

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
DylanDmitri: Having implemented both, I'm going with the Try/Finally because the purpose is more obvious and it does not require an additional class.

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
DylanDmitri: I'm going to procrastinated this a bit, and add proper logging after implementing some of the other suggestions, to prevent having to go back and change stuff.

src/Middleware/RequestThrottling/test/MiddlewareTests.cs
DylanDmitri: For more complicated tests, `SyncPoint` will be useful. This test is okay, because the code that retrieves the tasks for t1 and t2 runs synchronously, so it can't get out of order.

src/Middleware/RequestThrottling/src/SemaphoreWrapper.cs
DylanDmitri: good catch

src/Middleware/RequestThrottling/test/MiddlewareTests.cs
halter73: @anurse I showed @DylanDmitri SyncPoint yesterday because that's what I immediately though of too. Useful type there 😉

src/Middleware/RequestThrottling/src/RequestThrottlingOptions.cs
halter73: I've also started to avoid using `<remarks>` tags for this. Otherwise, the default values will only show up on the MSDN page specific to the property and not in the class listing. For example with KestrelServerOptions.NoDelay, you can see the default value of true in the remarks section of the [MSDN NoDelay page](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.server.kestrel.kestrelserveroptions.nodelay?view=aspnetcore-1.1#Microsoft_AspNetCore_Server_Kestrel_KestrelServerOptions_NoDelay), but the default value doesn't show up on the [MSDN KestrelServerOptions page](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.server.kestrel.kestrelserveroptions?view=aspnetcore-1.1) that more people look at.

We might want to look at moving the existing `<remarks>` doc comments to the `<summary>` section en masse.

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// <summary>
+        /// Invokes the logic of the middleware.
+        /// </summary>
+        /// <param name="context">The <see cref="HttpContext"/>.</param>
+        /// <returns>A <see cref="Task"/> that completes when the request leaves.</returns>
+        public async Task Invoke(HttpContext context)
+        {
+            try
+            {
+                await _semaphore.EnterQueue();
```
Tratcher: Careful, EnterQueue goes outside the try block in case it throws before acquiring the semaphore.

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
```diff
@@ -0,0 +1,68 @@
...
+                _semaphore.LeaveQueue();
+                _logger.LogDebug($"request finished, semaphore count at: {_semaphore.Count}");
+            }
+        }
+
+        /// <summary>
+        /// The number of live requests that are downstream from this middleware.
+        /// Cannot exceeed <see cref="RequestThrottlingOptions.MaxConcurrentRequests"/>.
+        /// </summary>
+        public int ConcurrentRequests
```
Tratcher: internal

DylanDmitri: yeah I guess this is basically for unit testing

src/Middleware/RequestThrottling/src/Internal/LoggerExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            _requestEnqueued = LoggerMessage.Define(
+                logLevel: LogLevel.Debug,
+                eventId: new EventId(1, "Request Enqueued"),
+                formatString: "Server is busy; queuing request in middleware."
+                );
+            _requestDequeued = LoggerMessage.Define(
+                logLevel: LogLevel.Debug,
+                eventId: new EventId(2, "Request Dequeued"),
+                formatString: "Space availible on server; request is leaving queue."
```
halter73: I think we normally just define these inline with the declaration, otherwise these classes get big pretty quickly.

See [SignalR's DefaultHubDispatcher.Log.cs](https://github.com/aspnet/AspNetCore/blob/01a87989364dea1b18743a9fe3757eccc220b2ca/src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.Log.cs ) for an example.

src/Middleware/RequestThrottling/src/Internal/LoggerExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+                );
+        }
+
+        internal static void RequestEnqueued(this ILogger logger)
```
halter73: Internal or not, I think it's a bad idea to make log methods extension methods because it makes it too easy to log messages to the wrong logging namespace.

I would either just make these normal static methods like [DefaultHubDispatcher.Log.cs](https://github.com/aspnet/AspNetCore/blob/01a87989364dea1b18743a9fe3757eccc220b2ca/src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.Log.cs) or create a logger object that has a private field reference the correct logger like [KestrelTace.cs](https://github.com/aspnet/AspNetCore/blob/01a87989364dea1b18743a9fe3757eccc220b2ca/src/Servers/Kestrel/Core/src/Internal/Infrastructure/KestrelTrace.cs).

src/Middleware/RequestThrottling/src/Internal/LoggerExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        private static readonly Action<ILogger, Exception> _requestEnqueued;
+        private static readonly Action<ILogger, Exception> _requestDequeued;
+
+        static LoggerExtensions()
+        {
+            _requestEnqueued = LoggerMessage.Define(
+                logLevel: LogLevel.Debug,
+                eventId: new EventId(1, "Request Enqueued"),
+                formatString: "Server is busy; queuing request in middleware."
```
halter73: I think logging the number of currently queued requests here makes sense as part of this PR.

halter73: @davidfowl @Tratcher @anurse What's our feeling about including in the TraceIdentifier in the log message as a structure parameter? I think we should. I know this can be pulled out of the scope, but this doesn't help you if you use console logging in its default configuration like we often do in our templates and samples.

[KestrelTrace.cs](https://github.com/aspnet/AspNetCore/blob/01a87989364dea1b18743a9fe3757eccc220b2ca/src/Servers/Kestrel/Core/src/Internal/Infrastructure/KestrelTrace.cs#L61) includes a TraceIdentifier structure parameter in logs where it's possible and uses ConnectionId otherwise. This is even done for logging when the app starts and stops reading the request body, so it's not like all these logs are being done outside of the request log scope and that's the reason for doing this.

src/Middleware/RequestThrottling/src/Internal/LoggerExtensions.cs
Tratcher: @halter73? Extension methods are the standard pattern we use.
https://github.com/aspnet/AspNetCore/blob/master/src/Middleware/StaticFiles/src/LoggerExtensions.cs
https://github.com/aspnet/AspNetCore/blob/master/src/Mvc/Mvc.Formatters.Xml/src/LoggerExtensions.cs
https://github.com/aspnet/AspNetCore/blob/master/src/Servers/Kestrel/Core/src/Internal/LoggerExtensions.cs

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
```diff
@@ -0,0 +1,81 @@
...
+
+namespace Microsoft.Aspnetcore.RequestThrottling
+{
+    /// <summary>
+    /// Limits the flow of requests through your application,
+    /// hopefully decreasing congestion by preventing threadpool starvation.
+    /// </summary>
+    public class RequestThrottlingMiddleware
+    {
+        private SemaphoreWrapper _semaphore;
```
halter73: Nit: readonly

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
```diff
@@ -0,0 +1,81 @@
...
+                {
+                    _logger.RequestDequeued();
+                }
+
+                await _next(context);
+            }
+            finally
+            {
+                _semaphore.LeaveQueue();
+                _logger.LogDebug($"request finished, semaphore count at: {_semaphore.Count}");
```
halter73: This should be removed or turned into a structured log.

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
halter73: It might also be nice to track QueuedRequests using an internal property for testing.

src/Middleware/RequestThrottling/src/RequestThrottlingOptions.cs
halter73: I see this was resolved, but I don't see the default value in the doc comment.

src/Middleware/RequestThrottling/test/MiddlewareTests.cs
```diff
@@ -0,0 +1,75 @@
...
+            var middleware = TestUtils.CreateTestMiddleWare(
+                maxConcurrentRequests: 1,
+                next: httpContext =>
+                {
+                    throw new DivideByZeroException();
+                });
+
+            Assert.Equal(0, middleware.ConcurrentRequests);
+
+            await Assert.ThrowsAsync<DivideByZeroException>(() =>
```
halter73: .OrTimeout()

Nit: Make this one line or use brackets for the lambda.

src/Middleware/RequestThrottling/src/Internal/LoggerExtensions.cs
davidfowl: No, need it’s already part of the scope.

src/Middleware/RequestThrottling/src/Internal/LoggerExtensions.cs
halter73: Gross. Even in Kestrel.Core! 😭 I think I should have made you put those methods in KestrelTrace.

Clearly we're inconsistent with whether we use extension methods on `ILogger` for structured logging or not. I vote for not using extension methods for this purpose going forward.

On an unrelated note, it looks like the consumers of Kestrel's LoggerExtensions are logging to `Microsoft.AspNetCore.Server.Kestrel.Core.KestrelServer` instead of `Microsoft.AspNetCore.Server.Kestrel` like the rest of the Kestrel.Core assembly. I guess at least we don't have EventId conflicts that way... Do you think it's worth consolidating these namespaces before 3.0, or is it not worth the break?

src/Middleware/RequestThrottling/src/Internal/LoggerExtensions.cs
halter73: Should we be removing ConnectionId and TraceId parameters from the Kestrel traces made inside the connection/request scope? Technically, that's a breaking change, right?

davidfowl: We can talk about fixing it for console logging so that it is on by default

src/Middleware/RequestThrottling/src/Internal/LoggerExtensions.cs
davidfowl: We changed the pattern but didn’t update all the places, nor was it really communicated broadly on the team.

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
davidfowl: This might be useful as an API actually. We can discuss it when we start to do metrics

davidfowl:
Now that we have end point routing it might be interesting to think about what type of per route options we would want to support.

cc @rynowak

src/Middleware/RequestThrottling/src/Internal/LoggerExtensions.cs
Tratcher: > We changed the pattern but didn’t update all the places, nor was it really communicated broadly on the team.

To the extension method pattern, right?

Tratcher:
> Now that we have end point routing it might be interesting to think about what type of per route options we would want to support.

Being able to opt out of throttling per-request or per-endpoint might be interesting. And what about a do-not-queue option (e.g. 503 immediately if the concurrent request limit is hit)?

I don't think you could do anything more complex like prioritization without making the queue implementation really complicated.

Tratcher:
> Now that we have end point routing it might be interesting to think about what type of per route options we would want to support.

Though this assumes you put the throttling after routing. We'll need to take a hard look at what middleware make sense before and after this one.
https://docs.microsoft.com/en-us/aspnet/core/fundamentals/middleware/?view=aspnetcore-2.2#built-in-middleware

src/Middleware/RequestThrottling/src/Internal/LoggerExtensions.cs
DylanDmitri: Inlined the stuff. I was copying another logging file :(

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
DylanDmitri: Removed; we discussed that logging already exists for request entry/outry and want to avoid duplicating that.

src/Middleware/RequestThrottling/test/MiddlewareTests.cs
```diff
@@ -0,0 +1,76 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.SignalR.Tests;
```
BrennanConroy: Change the namespace of `SyncPoint`

src/Middleware/RequestThrottling/src/Internal/LoggerExtensions.cs
davidfowl: Nope. Static method

src/Middleware/RequestThrottling/test/MiddlewareTests.cs
DylanDmitri: Should I make a new namespace? Maybe `Microsoft.AspNetCore.SyncPoint`?

BrennanConroy: I would use `Microsoft.AspNetCore.Internal`

src/Middleware/RequestThrottling/src/Internal/LoggerExtensions.cs
Tratcher: @davidfowl MAKE UP YOUR MINDS. Put it in https://github.com/aspnet/AspNetCore/wiki/Engineering-guidelines or never speak of this again.

davidfowl: @rynowak do the needful

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
```diff
@@ -33,7 +36,14 @@ public RequestThrottlingMiddleware(RequestDelegate next, ILoggerFactory loggerFa
             _next = next;
             _logger = loggerFactory.CreateLogger<RequestThrottlingMiddleware>();
             _options = options.Value;
-            _semaphore = new SemaphoreWrapper(_options.MaxConcurrentRequests); 
+
+            if (_options.MaxConcurrentRequests == null)
+            {
+                // TODO :: Localization
+                throw new ArgumentException(string.Format("Option must be provided {0}", nameof(_options.MaxConcurrentRequests)));
```
Eilon: A few notes:
1. Loc is not really a concern at this point, but it's fine if you want to use a RESX.
2. The exception message itself must be a complete sentence (proper grammar, end with a period, etc.)
3. One of the parameters to `ArgumentException` is the parameter name. In this case the parameter name is `options` (even though the problem is a sub-property of it), so make sure to pass that in.

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
```diff
@@ -33,7 +36,14 @@ public RequestThrottlingMiddleware(RequestDelegate next, ILoggerFactory loggerFa
...
+            }
+
+            _semaphore = new SemaphoreWrapper((int)_options.MaxConcurrentRequests);
```
Eilon: Instead of casting, use `.Value`.

src/Middleware/RequestThrottling/src/RequestThrottlingOptions.cs
```diff
@@ -12,7 +12,9 @@ public class RequestThrottlingOptions
     {
         /// <summary>
         /// Maximum number of concurrent requests. Any extras will be queued on the server. 
+        /// We do not provide a default since reasonable values can vary from 500 to 500 thousand.
+        /// Generally, highly async systems (and systems with higher backend latency) do better with more concurrent requests.
         /// </summary>
-        public int MaxConcurrentRequests { get; set; } = 10;
+        public int? MaxConcurrentRequests { get; set; } = null;
```
Eilon: No need to set `null` explicitly - it's already the default.

src/Middleware/RequestThrottling/src/RequestThrottlingOptions.cs
```diff
@@ -12,7 +12,9 @@ public class RequestThrottlingOptions
     {
         /// <summary>
         /// Maximum number of concurrent requests. Any extras will be queued on the server. 
+        /// We do not provide a default since reasonable values can vary from 500 to 500 thousand.
```
Eilon: There's no `we` here. This sounds like more of a regular `//` comment to devs, not a comment to an end user. In what way would this help them?

src/Middleware/RequestThrottling/test/MiddlewareTests.cs
```diff
@@ -72,5 +72,14 @@ public async Task QueuedRequestsContinueWhenSpaceBecomesAvailible()
             await t1.OrTimeout();
             await t2.OrTimeout();
         }
+
+        [Fact]
+        public void InvalidArgumentIfMaxConcurrentRequestsIsNull()
+        {
+            Assert.Throws<ArgumentException>(() =>
```
Eilon: Once you update the parameters passed into the exception (see earlier comment), I believe you can add a check to this assertion that the right parameter name was passed in (`options`).

Eilon (changes requested):

src/Middleware/RequestThrottling/src/Internal/LoggerExtensions.cs
anurse: https://github.com/aspnet/AspNetCore/wiki/Engineering-guidelines#logging-patterns

src/Middleware/RequestThrottling/src/RequestThrottlingOptions.cs
Tratcher: This is null by default because the correct value is application specific. This option must be configured by the application.

src/Middleware/RequestThrottling/test/MiddlewareTests.cs
Tratcher: Yes, Thows returns the exception object and you can inspect its properties.

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
```diff
@@ -0,0 +1,93 @@
...
+            }
+        }
+
+        /// <summary>
+        /// The number of live requests that are downstream from this middleware.
+        /// Cannot exceeed <see cref="RequestThrottlingOptions.MaxConcurrentRequests"/>.
+        /// </summary>
+        internal int ConcurrentRequests
+        {
+            get => ((int)_options.MaxConcurrentRequests - _semaphore.Count);
```
Tratcher: Why isn't ConcurrentRequests a property on semaphore?

src/Middleware/RequestThrottling/src/Internal/SemaphoreWrapper.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        public void Release()
+        {
+            _semaphore.Release();
+        }
+
+        public int Count
+        {
+            get => _semaphore.CurrentCount;
+        }
```
halter73: Nit: newiline

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
halter73: I agree that would be better layering. @DylanDmitri I would just store the queueLength as a readonly field in SemaphoreWrapper and move this property there.

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
```diff
@@ -0,0 +1,93 @@
...
+        /// </summary>
+        /// <param name="context">The <see cref="HttpContext"/>.</param>
+        /// <returns>A <see cref="Task"/> that completes when the request leaves.</returns>
+        public async Task Invoke(HttpContext context)
+        {
+            var waitInQueueTask = _semaphore.EnterQueue();
+
+            try
+            {
+                var needsToWaitOnQueue = !waitInQueueTask.IsCompleted;
```
halter73: Nit:

```suggestion
                var needsToWaitOnQueue = !waitInQueueTask.IsCompletedSuccessfully;
```

src/Middleware/RequestThrottling/src/Internal/SemaphoreWrapper.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        public async Task EnterQueue()
+        {
+            lock (_waitingRequestsLock)
+            {
+                _waitingRequests++;
+            }
+
+            await _semaphore.WaitAsync();
```
halter73: We should save this task and check `IsCompletedSuccessfully` like we do in the middleware. If WaitAsync immediately completed successfully, you can save yourself from acquiring the lock twice to increment and then immediately decrement.

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
DylanDmitri: Would this be okay?
```
throw new ArgumentNullException(nameof(_options.MaxConcurrentRequests), "The parameter 'options.MaxConcurrentRequets' must be specified.");
```

src/Middleware/RequestThrottling/src/Internal/SemaphoreWrapper.cs
DylanDmitri: I've made the change: if the semaphore immediately goes, the lock section is skipped.

Having duplicate structures immediately next to each other (Middleware -> RequestQueue) seems fishy. Theoretically I would like to move the logging into the RequestQueue and simplify the middleware logic, but for now it's probably more useful for logging to be at the higher level.

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
```diff
@@ -0,0 +1,88 @@
...
+        {
+            var waitInQueueTask = _requestQueue.EnterQueue();
+
+            try
+            {
+                var needsToWaitOnQueue = !waitInQueueTask.IsCompletedSuccessfully;
+                if (needsToWaitOnQueue)
+                {
+                    RequestThrottlingLog.RequestEnqueued(_logger, WaitingRequests);
+                    await waitInQueueTask;
```
Tratcher: Move above the try.

src/SignalR/common/Http.Connections/test/Microsoft.AspNetCore.Http.Connections.Tests.csproj
```diff
@@ -1,11 +1,12 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFramework>netcoreapp3.0</TargetFramework>
   </PropertyGroup>
 
   <ItemGroup>
     <Compile Include="$(SharedSourceRoot)Buffers.Testing\**\*.cs" />
+    <Compile Include="$(SharedSourceRoot)test\SyncPoint.cs" />
```
Tratcher:
```suggestion
    <Compile Include="$(SharedSourceRoot)SyncPoint\SyncPoint.cs" />
```

Tratcher (approved):

src/Middleware/RequestThrottling/src/Internal/RequestThrottlingLog.cs
```diff
@@ -0,0 +1,27 @@
...
+
+namespace Microsoft.AspNetCore.RequestThrottling.Internal
+{
+    internal static class RequestThrottlingLog
+    {
+        private static readonly Action<ILogger, int, Exception> _requestEnqueued =
+            LoggerMessage.Define<int>(LogLevel.Debug, new EventId(1, "Request Enqueued"), "Server is busy; request queued in middleware. Current queue length: {queuedRequests}.");
+
+        private static readonly Action<ILogger, int, Exception> _requestDequeued =
+            LoggerMessage.Define<int>(LogLevel.Debug, new EventId(2, "Request Dequeued"), "Space availible on server; request has left queue. Current queue length: {queuedRequests}.");
```
anurse:
```suggestion
            LoggerMessage.Define<int>(LogLevel.Debug, new EventId(2, "Request Dequeued"), "Request dequeued. Current queue length: {queuedRequests}.");
```

"Space" is a vague concept here :). I think it's sufficient to just say it's dequeued.

src/Middleware/RequestThrottling/src/Internal/RequestThrottlingLog.cs
```diff
@@ -0,0 +1,27 @@
...
+
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.RequestThrottling.Internal
+{
+    internal static class RequestThrottlingLog
+    {
+        private static readonly Action<ILogger, int, Exception> _requestEnqueued =
+            LoggerMessage.Define<int>(LogLevel.Debug, new EventId(1, "Request Enqueued"), "Server is busy; request queued in middleware. Current queue length: {queuedRequests}.");
```
anurse:
```suggestion
            LoggerMessage.Define<int>(LogLevel.Debug, new EventId(1, "Request Enqueued"), "Concurrent request limit reached, queuing request. Current queue length: {queuedRequests}.");
```

Let's clarify what "busy" means here. It means the concurrent request limit was reached.

src/Middleware/RequestThrottling/src/Internal/RequestQueue.cs
```diff
@@ -0,0 +1,62 @@
...
+            var waitInQueueTask = _semaphore.WaitAsync();
+
+            var needsToWaitOnQueue = !waitInQueueTask.IsCompletedSuccessfully;
+            if (needsToWaitOnQueue)
+            {
+                lock (_waitingRequestsLock)
+                {
+                    WaitingRequests++;
+                }
+                await waitInQueueTask;
```
halter73: Nit: newlines around the await.

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
```diff
@@ -0,0 +1,88 @@
...
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.RequestThrottling;
+using Microsoft.AspNetCore.RequestThrottling.Internal;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Aspnetcore.RequestThrottling
```
anurse:
```suggestion
namespace Microsoft.AspNetCore.RequestThrottling
```

src/Middleware/RequestThrottling/src/RequestThrottlingExtensions.cs
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Aspnetcore.RequestThrottling;
```
anurse:
```suggestion
using Microsoft.AspNetCore.RequestThrottling;
```

halter73 (approved):

src/Middleware/RequestThrottling/src/RequestThrottlingExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+
+namespace Microsoft.AspNetCore.Builder
+{
+    /// <summary>
+    /// Extension methods for adding the <see cref="RequestThrottlingMiddleware"/> to an application.
+    /// </summary>
+    public static class RequestThrottlingExtensions
+    {
+        /// <summary>
+        /// Adds the <see cref="RequestThrottlingMiddleware"/> to limit the number of active requests within
```
anurse: I think the terminology of being "within" the server doesn't really make sense here. All the requests are "within" the server, even the queued ones. They're just being held up in the queue. What we're doing is limiting the number of concurrent requests. Perhaps something like this:

```suggestion
        /// Adds the <see cref="RequestThrottlingMiddleware"/> to limit the number of concurrently-executing requests.
```

src/Middleware/RequestThrottling/src/RequestThrottlingExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// your server.
```
anurse: Given the above comment, you wouldn't need this line at all.

```suggestion
```

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
```diff
@@ -0,0 +1,88 @@
...
+{
+    /// <summary>
+    /// Limits the flow of requests through your application,
```
anurse: "Flow" is another somewhat vague term. Better to be clear about precisely what is limited:

```suggestion
    /// Limits the number of concurrent requests allowed in the application,
```

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
```diff
@@ -0,0 +1,88 @@
...
+    /// hopefully decreasing congestion by preventing threadpool starvation.
```
anurse: "Hopefully" is another phrasing choice I'd avoid here. Let's be concrete about what this does. In fact, I think it's sufficient to omit this sentence since in the class summary we just need to focus on what this class actually **does**. We'll have more documentation on the why and how elsewhere.

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
```diff
@@ -0,0 +1,88 @@
...
+        /// <param name="context">The <see cref="HttpContext"/>.</param>
+        /// <returns>A <see cref="Task"/> that completes when the request leaves.</returns>
+        public async Task Invoke(HttpContext context)
+        {
+            var waitInQueueTask = _requestQueue.EnterQueue();
+
+            try
+            {
+                var needsToWaitOnQueue = !waitInQueueTask.IsCompletedSuccessfully;
+                if (needsToWaitOnQueue)
```
anurse: Let's also log in an `else` clause here. It's useful to know that the queue wasn't needed.

src/Middleware/RequestThrottling/src/Internal/RequestThrottlingLog.cs
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.RequestThrottling.Internal
+{
+    internal static class RequestThrottlingLog
```
anurse: Unless this is used in other places, our pattern is generally to make this a nested class *inside* `RequestThrottlingMiddleware`

analogrelay (commented):
Some word smithing. I'm loving this so far!

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
anurse: This one is a little more than wordsmithing. It's actually putting this class in a completely different namespace because namespaces are case-sensitive.

src/Middleware/RequestThrottling/src/Internal/RequestQueue.cs
```diff
@@ -0,0 +1,64 @@
...
+        public async Task EnterQueue()
+        {
+            var waitInQueueTask = _semaphore.WaitAsync();
+
+            var needsToWaitOnQueue = !waitInQueueTask.IsCompletedSuccessfully;
+            if (needsToWaitOnQueue)
+            {
+                lock (_waitingRequestsLock)
+                {
+                    WaitingRequests++;
```
omariom: Why not `Interlocked.Increment`?

BrennanConroy: Not needed since the variable is only accessed inside a lock

omariom: And the lock only protects the variable. Interlocked.Increment would be more efficient.

omariom:
The number of the requests waiting due to throttling is an important metric. 
It would be nice to have it reported somehow.

davidfowl:
> The number of the requests waiting due to throttling is an important metric.
It would be nice to have it reported somehow.

Metrics are planned

src/Middleware/RequestThrottling/src/Internal/RequestQueue.cs
DylanDmitri: Good catch! Interlocked has better performance, and I'll go add it in the next PR

PR #10431, opened 2019-05-21
Allow receipts for render batches later than the current entry.

### LINKED ISSUE #9901: Blazor (server-side) System.InvalidOperationException: Received a notification for a rendered batch when not expecting it.
grafik

pranavkm:
Fixes https://github.com/aspnet/AspNetCore/issues/9901

### DIFF
--- a/src/Components/Server/src/Circuits/RemoteRenderer.cs
+++ b/src/Components/Server/src/Circuits/RemoteRenderer.cs
@@ -3,13 +3,11 @@
 
 using System;
 using System.Collections.Concurrent;
-using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Text.Encodings.Web;
 using System.Threading;
 using System.Threading.Tasks;
-using MessagePack;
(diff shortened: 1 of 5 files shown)


pranavkm:
It looks like we can address this by hardening ACKs on the server. Previously we required these to be in order, versus if we treat it more like a sliding window, it makes this fairly robust. It was trivial to reproduce the original error using Long polling before the change:

Before

Once this change was in, I couldn't reproduce it (this ran for a few minutes, it's sped up and trimmed):

After

cc @BrennanConroy

SteveSandersonMS:
Could you clarify why this is the right solution? My understanding, which might be incorrect, is that if the underlying transport preserves ordering then this issue would never exist. I'd consider that a much more preferable resolution than making it "fairly robust" by being less strict.

Do you know of some situation where it's possible, even just theoretically, for us to get an unexpected receipt number if the underlying transport is preserving delivery order?

SteveSandersonMS:
Cool gifs BTW

Issue #9327, opened 2019-04-12
Too many "Failed connection handshake" errors in our logs


rfrancisco:
### Describe the bug
In each of our servers logs we see dozens of "Failed connection handshake" errors every second. Each of our signalR servers is running in an AWS EC2 instance behind an AWS Application Load Balancer. 
We have taken a look at prevously reported issues related to this error message, and the solutions they presented had to do with NGINX configuration. We are not using nginx, since we rely on AWS ALB (with sticky sessions enabled) to route traffic to the servers, so we tried setting up a server that nginx (using the recommended settings), and removing the AWS ALB from the equation, but the problem persisted.

The clients connect to our servers using a javascript that uses @aspnet/signalr 1.1.4. In our local tests, in the most common desktop and mobile browsers, we don't find any problems connecting to signalR and we have hundreds of thousand os users connected, but we also don't know how many may not be connecting due to this error.

### To Reproduce
Can't reproduce the error locally, we just see it in our production environment logs.

### Expected behavior
Expect not to have these errors happening. Or at least not so frequently.

### Screenshots
![ubuntu_ip-172-31-54-19___opt_insurads_log](https://user-images.githubusercontent.com/1242411/56038430-d9c32d00-5d29-11e9-9e8b-d676c57d4fbd.png)

### Additional context
![ubuntu_ip-172-31-54-19___opt_insurads_log](https://user-images.githubusercontent.com/1242411/56037552-746e3c80-5d27-11e9-9033-f4264c85bd70.png)

Here are some logs taken with "Microsoft.AspNet.SignalR" at Debug log level: [log.zip](https://github.com/aspnet/AspNetCore/files/3073503/log.zip)

Also, the client script that users use to connect is built with @aspnet/signalr-1.1.4 but we've had these error messages even before we upgraded from 1.0.0-rc-final.

Let me know if there's any other information i can provide to help troubleshoot this.

mikaelm12:
Can you share any client logic you're using to connect? Just a hunch but are you trying to implement any reconnect logic?

rfrancisco:
Yes. We do have reconnect logic implemented on the client. 

Here is the related code: [messaging.ts.zip](https://github.com/aspnet/AspNetCore/files/3076093/messaging.ts.zip)

anurse:
Thanks for the logs, we'll look at this in our upcoming milestone. If you are able to capture any network traces from the server side when this error occurs, that would be very helpful.

[...]

PR #10433, opened 2019-05-22
 Log clearer handshake failures in SignalR .NET client

halter73:
Addresses #9327

### DIFF
--- a/src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
+++ b/src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
@@ -1018,14 +1018,17 @@ namespace Microsoft.AspNetCore.SignalR.Client
             if (sendHandshakeResult.IsCompleted)
             {
                 // The other side disconnected
-                throw new InvalidOperationException("The server disconnected before the handshake was completed");
+                var ex = new IOException("The server disconnected before the handshake could be started.");
+                Log.ErrorReceivingHandshakeResponse(_logger, ex);
+                throw ex;
             }
 
             var input = startingConnectionState.Connection.Transport.Input;
 
+            using var handshakeCts = new CancellationTokenSource(HandshakeTimeout);
+
             try
             {
-                using (var handshakeCts = new CancellationTokenSource(HandshakeTimeout))
                 // cancellationToken already contains _state.StopCts.Token, so we don't have to link it again
                 using (var cts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken, handshakeCts.Token))
                 {
@@ -1057,6 +1060,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
                                             $"Unable to complete handshake with the server due to an error: {message.Error}");
                                     }
 
+                                    Log.HandshakeComplete(_logger);
                                     break;
                                 }
                             }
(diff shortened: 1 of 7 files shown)


src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -1015,14 +1015,17 @@ private async Task HandshakeAsync(ConnectionState startingConnectionState, Cance
...
                 // The other side disconnected
-                throw new InvalidOperationException("The server disconnected before the handshake was completed");
+                var ex = new IOException("The server disconnected before the handshake could be started.");
+                Log.ErrorReceivingHandshakeResponse(_logger, ex);
+                throw ex;
             }
 
             var input = startingConnectionState.Connection.Transport.Input;
 
+            using var handshakeCts = new CancellationTokenSource(HandshakeTimeout);
```
BrennanConroy: Curious, why did you move this?

src/SignalR/clients/csharp/Client/test/UnitTests/HubConnectionTests.ConnectionLifecycle.cs
```diff
@@ -166,20 +154,43 @@ Task<ConnectionContext> ConnectionFactory(TransferFormat format)
...
+
+                using (StartVerifiableLog(ExpectedErrors))
+                {
+                    var testConnection = new TestConnection(autoHandshake: false);
+                    await AsyncUsing(CreateHubConnection(testConnection, loggerFactory: LoggerFactory), async connection =>
+                    {
+                        testConnection.Transport.Input.Complete();
+                        try
+                        {
+                            await connection.StartAsync();
```
BrennanConroy: nit: OrTimeout

src/SignalR/clients/csharp/Client/test/UnitTests/HubConnectionTests.Reconnect.cs
```diff
@@ -17,896 +18,899 @@ namespace Microsoft.AspNetCore.SignalR.Client.Tests
 {
     public partial class HubConnectionTests
     {
-        [Fact]
-        public async Task ReconnectIsNotEnabledByDefault()
+        public class Reconnect : VerifiableLoggedTest
```
BrennanConroy: Is this the only change to this file?

BrennanConroy (approved):

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
halter73: I log different messages when the user manually cancels the handshake using a CancellationToken or by calling StopAsync() than I do when the handshake simply times out. I moved this out of the try scope so I could check `handshakeCts.IsCancellationRequested` in the catch block below to determine what to log.

src/SignalR/clients/csharp/Client/test/UnitTests/HubConnectionTests.ConnectionLifecycle.cs
halter73: Good catch. I cleaned up most of HubConnectionTests.Protocol but missed this test despite modifying it.

src/SignalR/clients/csharp/Client/test/UnitTests/HubConnectionTests.Reconnect.cs
halter73: https://github.com/aspnet/AspNetCore/pull/10433/commits/5a97797b0112ba2f6b5bf45a047e0964c2cb4493#diff-071f9acf91eca9ec3e4b318ccfea3037 and removing "Reconnect" from the beginning of the test methods are the only other changes in the file. If you add "?w=1" to the URL or click the "No Whitespace" button.

Issue #2360, opened 2018-01-01
Add ConsoleReporter


aspnet-hello:
_From @ryanbrandenburg on Wednesday, October 25, 2017 3:07:28 PM_

Add the `ConsoleReporter` as seen in [BuildTools](https://github.com/aspnet/DotNetTools/blob/8c327312824457b4dd1fccd814380894edf36eab/shared/ConsoleReporter.cs) to `Microsoft.Extensions.CommandLineUtils.Sources`.

_Copied from original issue: aspnet/Common#276_

aspnet-hello:
_From @ryanbrandenburg on Wednesday, October 25, 2017 3:08:10 PM_

After adding it here, remove any copies that have popped up in repos like Infrastructure.

aspnet-hello:
_From @Eilon on Tuesday, November 7, 2017 10:06:21 AM_

Go for it 😄

Eilon:
Naah.

PR #10439, opened 2019-05-22
Update Mono WebAssembly for Blazor

SteveSandersonMS:
Mmm, freshly baked Mono. See https://github.com/aspnet/Blazor/pull/1807 for more.

This PR also updates the ws-proxy code that we consume in a shared-source style from Mono. **There is no need to code-review the contents of the ws-proxy directory**, as we didn't write that code. Basically the only thing relevant to code-review here are the change to `Versions.props` and `Blazor.MonoRuntime.props`.

This build is going to fail at first because it won't be able to restore the `Microsoft.AspNetCore.Blazor.Mono` package. We can't get a CI build of that until we have updated code-signing approval for that build.


javiercn (approved):
Not a lot of context in what's going on here, but I see nothing wrong.

build/sources.props
```diff
@@ -27,6 +27,11 @@
       $(RestoreSources);
       https://dotnet.myget.org/F/aspnetcore-dev/api/v3/index.json;
     </RestoreSources>
+    <!-- TODO remove this once we move Microsoft.AspNetCore.Blazor.Mono to a non-myget feed -->
+    <RestoreSources>
+      $(RestoreSources);
+      https://dotnet.myget.org/F/blazor-dev/api/v3/index.json;
+    </RestoreSources>
```
SteveSandersonMS: Flagging this to @natemcmaster @dougbu 

Looks like Microsoft.AspNetCore.Blazor.Mono doesn't have a new home on dotnetfeed.blob.core.windows.net yet.

Since this follows the same pattern as for other un-rehomed packages above, it looks reasonable to unblock ourselves by adding this `blazor-dev` source here. Not sure which feed it should ultimately be moved to, or whether it needs a new feed of its own.

natemcmaster: I will address this as a part of aspnet/AspNetCore-Internal
#2360. We've been moving away from myget feeds due to perf and reliability reasons, so this will get updated with a static feed instead.

PR #10444, opened 2019-05-22
Server-side Blazor template with authentication

SteveSandersonMS:
Adds all the same authentication options as in the MVC template, i.e.:

 * NoAuth
 * Individual (both SQLite and SQL Server variants)
 * IndividualB2C
 * SingleOrg
 * MultiOrg
 * Windows

### Caveat

On the IndividualB2C and the MultiOrg templates, when you log out, it takes you to a page that says "You're now logged out" but it has no layout, so it's just black Times New Roman on a white screen.

For the IndividualB2C template, we *could* fix this by overriding the `SignedOut.cshtml` page with one that just immediately redirects to '/', but the same fix doesn't work for MultiOrg because that would then immediately redirect you on to the login screen, which is weird when you just logged out

It would be better if we added a minimal Razor Pages layout somehow. I'm guessing this can be done by laying down extra files like `Areas/AzureB2C/Pages/_ViewStart.cshtml`, but this hasn't been working for reasons that neither I nor @javiercn have tracked down yet.

If it's possible to configure these authentication types to have a different return URL after sign out, that would solve the IndividualB2C case (redirect to homepage) and maybe also the MultiOrg case (don't redirect - just stay on Azure's "You're signed out" page).

As a final alternative, this could be a known issue for this release. It just looks rough - it's not functionally defective.

### Follow-up: client-side Blazor templates

I'm hoping we can find someone else to tackle the client-side Blazor equivalent to this PR next week. I'll post an issue with some guidance on the implementation design, since this has already been designed in detail and I have a working prototype.


src/ProjectTemplates/Web.ProjectTemplates/content/RazorComponentsWeb-CSharp/App.razor
```diff
@@ -1,5 +1,8 @@
-﻿@*
-    The Router component displays whichever component has a @page
-    directive matching the current URI.
-*@
-<Router AppAssembly="typeof(Startup).Assembly" />
+﻿<CascadingAuthenticationState>
```
SteveSandersonMS: I'm proposing we put `<CascadingAuthenticationState>` here in all cases, even NoAuth, since there's no drawback and it will help people if they try to use `<AuthenticationView>` later.

src/ProjectTemplates/Web.ProjectTemplates/content/RazorComponentsWeb-CSharp/Shared/LoginDisplay.razor
```diff
@@ -0,0 +1,57 @@
...
+<AuthorizeView>
+    <Authorized>
+        <a href="Identity/Account/Manage">Hello, @context.User.Identity.Name!</a>
+        <a href="Identity/Account/LogOut">Log out</a>
+    </Authorized>
+    <NotAuthorized>
+        <a href="Identity/Account/Register">Register</a>
+        <a href="Identity/Account/Login">Log in</a>
+    </NotAuthorized>
+</AuthorizeView>
```
javiercn: I like this

src/ProjectTemplates/Web.ProjectTemplates/content/RazorComponentsWeb-CSharp/Startup.cs
```diff
@@ -2,24 +2,127 @@
...
     {
+#if (OrganizationalAuth || IndividualAuth)
+        public Startup(IConfiguration configuration)
+        {
+            Configuration = configuration;
+        }
+
+        public IConfiguration Configuration { get; }
+
+#endif
```
javiercn: You can avoid the `#if` block here and just add the ctor unconditionally, as that is what other templates do and we want to be aligned.

javiercn (commented):
Looks great!

A couple of small comments, but overall is a great start.
Do you plan to add tests for the auth configurations?

You need to include tests for the template baselines at the very least to validate we don't leave files behind, and I think restore/build tests are also important. (I think they get done as part of the baseline test, but can't remember).

It might also help if you post a screenshot of the template, to get an idea of the visual changes.

We also need CTI issues filed for these, as we want vendors to cover these new scenarios on VS. @ajaybhargavb can help you through that.

ryanbrandenburg (changes requested):
> I think they get done as part of the baseline test, but can't remember

[They do not](https://github.com/aspnet/AspNetCore/blob/master/src/ProjectTemplates/test/BaselineTest.cs#L73).

As @javiercn said, these scenarios need baseline tests and scenario tests. You also need to rebase to avoid the compile errors that accidentally got merged yesterday.

src/ProjectTemplates/Web.ProjectTemplates/content/RazorComponentsWeb-CSharp/Startup.cs
SteveSandersonMS: Seems reasonable. Done.

SteveSandersonMS:
OK, I've added baseline tests for all the auth types. I've also added a couple of new scenario tests, which reaches roughly the same coverage as for other project types (i.e., just individual auth, and not automating any usage of the identity UI).

Once CI passes this should be ready now.

src/ProjectTemplates/test/RazorComponentsTemplateTest.cs
```diff
@@ -73,6 +73,56 @@ public async Task RazorComponentsTemplateWorks()
             }
         }
 
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        [Flaky("https://github.com/aspnet/AspNetCore-Internal/issues/2407", FlakyOn.AzP.Windows)]
```
ryanbrandenburg: This issue is closed. Did you mean to link https://github.com/aspnet/AspNetCore-Internal/issues/2329?

ryanbrandenburg (approved):

src/ProjectTemplates/test/RazorComponentsTemplateTest.cs
SteveSandersonMS: I was just copying the same link from another test. I guess multiple need to be updated.

PR #10451, opened 2019-05-22
Fix RendererSyncContext.Post()

### LINKED ISSUE #9683: Blazor signalr connection breaks after syncing large dom / dom element
### Describe the bug
On login in our application we load a logo which is displayed. Normally this is quite small but since it is user specific it sometimes can be large. With preview 4 the server connection is broken when this is too large. The client then "freezes" - i.e. does not respond to clicks. The clicks do generate websocket messages but no responses. After a timeout the auto reconnect establishes a new connection and everything is then ok (until the next large dom sync).
 
This worked fine in preview 3.
### To Reproduce
I tried to reproduce with a small update to the template project but could not. Let me know if I should try again. It is 100% reproducible in our application.

I've artificially generated various size images and it seems to break when the image is between 25440 and 25450 bytes. Tracing in chrome the websocket message for this dom sync is between 43772 and 43784 bytes (working / non-working).

Please let me know if there is anything I can do to get more diagnostics on this for the case where it fails.

rynowak:
Fixes: #9683 - SignalR connection breaks on large DOM

The root cause here is a misbehaving sync context. It's not legal for a
Post() implementation to run a callback synchronously. We want that
behavior for most of the functionality Blazor calls directly, but Post()
should always go async or else various threading primitives are broken.

-------------

/cc @stephentoub @BrennanConroy @halter73 

-----

This won't pass the CI yet because we have a number (8 or so) of Blazor **tests** that were based on this flawed behavior. All of these tests follow the pattern of:

```
var tcs = new TaskCompletionSource<object>();

// Do a bunch of stuff

tcs.SetResult(null);

// Verify a bunch of stuff
```

The flaw is that we're not awaiting anything after setting the result - and baking in the assumption that the continuations of the non-completed task will run synchronously. This relies on the bug we're fixing, and so these tests need to be improved to await for something.

I'm not aware of another underlying product bug in this area yet, and my attempts to investigate these tests have been stymied by debugger issues.

### DIFF
--- a/src/Components/Components/src/Rendering/RendererSynchronizationContext.cs
+++ b/src/Components/Components/src/Rendering/RendererSynchronizationContext.cs
@@ -40,7 +40,7 @@ namespace Microsoft.AspNetCore.Components.Rendering
         public Task Invoke(Action action)
         {
             var completion = new TaskCompletionSource<object>();
-            Post(_ =>
+            ExecuteSynchronouslyIfPossible(_ =>
             {
                 try
                 {
@@ -63,7 +63,7 @@ namespace Microsoft.AspNetCore.Components.Rendering
         public Task InvokeAsync(Func<Task> asyncAction)
         {
             var completion = new TaskCompletionSource<object>();
-            Post(async (_) =>
+            ExecuteSynchronouslyIfPossible(async (_) =>
             {
                 try
                 {
@@ -86,7 +86,7 @@ namespace Microsoft.AspNetCore.Components.Rendering
         public Task<TResult> Invoke<TResult>(Func<TResult> function)
         {
             var completion = new TaskCompletionSource<TResult>();
-            Post(_ =>
+            ExecuteSynchronouslyIfPossible(_ =>
             {
                 try
                 {
@@ -109,7 +109,7 @@ namespace Microsoft.AspNetCore.Components.Rendering
         public Task<TResult> InvokeAsync<TResult>(Func<Task<TResult>> asyncFunction)
         {
             var completion = new TaskCompletionSource<TResult>();
-            Post(async (_) =>
+            ExecuteSynchronouslyIfPossible(async (_) =>
             {
                 try
                 {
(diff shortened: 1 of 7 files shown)


javiercn (approved):

src/Components/Components/src/Rendering/RendererSynchronizationContext.cs
```diff
@@ -63,7 +63,7 @@ public Task Invoke(Action action)
         public Task InvokeAsync(Func<Task> asyncAction)
         {
             var completion = new TaskCompletionSource<object>();
-            Post(async (_) =>
+            ExecuteSynchronouslyIfPossible(async (_) =>
```
stephentoub: Unrelated to this change, but how important is perf on these code paths?  Right now this is allocating a delegate and a closure on every call, in order to capture both asyncAction and completion.  You could avoid both of those by creating your own TCS-derived type, e.g.
```C#
internal sealed class StatefulTaskCompletionSource<TState, TResult> : TaskCompletionSource<TResult>
{
    public TState State;
}
```
and then passing that object into the callback via its `object state`, e.g.
```C#
var completion = new StatefulTaskCompletionSource<Func<Task>, object>() { State = asyncAction };
ExecuteSynchronouslyIfPossible(async state =>
{
    var local = (StatefulTaskCompletionSource<Func<Task>, object>)state;
    try
    {
        await local.State();
        local.SetResult(null);
    }
    catch (OperationCanceledException)
    {
        local.SetCanceled();
    }
    catch (Exception exception)
    {
        local.SetException(exception);
    }
}, completion);
return completion.Task;
```

rynowak: We'd love to improve the perf of this. I'll give your idea a try.

SteveSandersonMS:
@rynowak I expect this is all good, but to what extent could you describe the effects of this change in terms of user-visible effects? In what cases might (say) a UI update be completed synchronously before but it becomes async now? Is it possible for this to change anything about (say) the order in which components are rendered, either on the server or on WebAssembly?

rynowak:
> I expect this is all good, but to what extent could you describe the effects of this change in terms of user-visible effects? In what cases might (say) a UI update be completed synchronously before but it becomes async now? 

I'm not aware of any product behavior changes as a result of this. We use `IDispatcher.Invoke` which hasn't had its behavior change. 

Where users could have had an issue would be in their own code - if they are using task completion source and scheduling their own continuations on in. This would have behaved correctly in client-side and buggily in server-side.

SteveSandersonMS (approved):

PR #10453, opened 2019-05-22
React to Authorization refactor

Tratcher:
Fixes the build after overlapping PR merges.
https://github.com/aspnet/AspNetCore-Internal/issues/2554

Please merge if you don't see me, I've got an appointment.


HaoK (approved):

natemcmaster:
I see the macOS step already passed. That's good enough confirmation of this change for me. Master is already completely busted with compiler errors, so merging now without waiting the full hour for test validation.

Issue #10291, opened 2019-05-16
No handling for bad request when sent multipart form data request


rosieks:
### Describe the bug
If I send multipart/form-data request which has invalid headers then I receive 500 error. I would rather expect 400 as the request is invalid. In my case if content type doesn't have boundary or if there is no content disposition header

### To Reproduce
1. Using this version of ASP.NET Core '2.2.0'
2. Run this code: https://gist.github.com/rosieks/c26c64ee1f0eb19efb07e5783909c20b
3. With these arguments: Just open '/api/values' in the browser

### Expected behavior
Right now I receive error 500, while I would expect that incorrect request will be handled by ASP.NET Core and returned as 400 - Bad Request.

### Additional context
Include the output of `dotnet --info`

.NET Core SDK (reflecting any global.json):
 Version:   2.2.203
 Commit:    e5bab63eca

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.16299
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\2.2.203\

Host (useful for support):
  Version: 2.2.4
  Commit:  f95848e524

.NET Core SDKs installed:
  1.1.0 [C:\Program Files\dotnet\sdk]
  2.0.0 [C:\Program Files\dotnet\sdk]
  2.0.2 [C:\Program Files\dotnet\sdk]
  2.0.3 [C:\Program Files\dotnet\sdk]
  2.1.2 [C:\Program Files\dotnet\sdk]
  2.1.4 [C:\Program Files\dotnet\sdk]
  2.1.100 [C:\Program Files\dotnet\sdk]
  2.1.101 [C:\Program Files\dotnet\sdk]
  2.1.103 [C:\Program Files\dotnet\sdk]
  2.1.200 [C:\Program Files\dotnet\sdk]
  2.1.201 [C:\Program Files\dotnet\sdk]
  2.1.202 [C:\Program Files\dotnet\sdk]
  2.1.300 [C:\Program Files\dotnet\sdk]
  2.1.402 [C:\Program Files\dotnet\sdk]
  2.1.403 [C:\Program Files\dotnet\sdk]
  2.1.500 [C:\Program Files\dotnet\sdk]
  2.1.503 [C:\Program Files\dotnet\sdk]
  2.1.505 [C:\Program Files\dotnet\sdk]
  2.1.602 [C:\Program Files\dotnet\sdk]
[...]

PR #10456, opened 2019-05-22
Fix for crashing during parsing `multipart/form-data` requests in FormValueProvider

rosieks:
This PR adds 
* test that prove that `FormValueProviderFactory` crash if *content-type* miss boundary
* fix that catch all parsing errors (`InvalidDataException`)

Addresses #10291

### DIFF
--- a/src/Mvc/Mvc.Core/src/ModelBinding/FormValueProviderFactory.cs
+++ b/src/Mvc/Mvc.Core/src/ModelBinding/FormValueProviderFactory.cs
@@ -2,8 +2,11 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.Globalization;
+using System.IO;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
 
 namespace Microsoft.AspNetCore.Mvc.ModelBinding
 {
@@ -33,9 +36,20 @@ namespace Microsoft.AspNetCore.Mvc.ModelBinding
         private static async Task AddValueProviderAsync(ValueProviderFactoryContext context)
         {
             var request = context.ActionContext.HttpContext.Request;
+            IFormCollection form;
+            try
+            {
+                form = await request.ReadFormAsync();
+            }
+            catch (InvalidDataException ex)
+            {
+                form = FormCollection.Empty;
+                context.ActionContext.ModelState.TryAddModelException(string.Empty, ex);
+            }
+
             var valueProvider = new FormValueProvider(
                 BindingSource.Form,
-                await request.ReadFormAsync(),
+                form,
                 CultureInfo.CurrentCulture);
 
             context.ValueProviders.Add(valueProvider);
--- a/src/Mvc/Mvc.Core/test/ModelBinding/FormValueProviderFactoryTest.cs
+++ b/src/Mvc/Mvc.Core/test/ModelBinding/FormValueProviderFactoryTest.cs
@@ -3,6 +3,8 @@
 
 using System.Collections.Generic;
 using System.Globalization;
+using System.IO;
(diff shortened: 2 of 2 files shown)


src/Mvc/Mvc.Core/src/ModelBinding/FormValueProviderFactory.cs
```diff
@@ -33,9 +36,20 @@ public Task CreateValueProviderAsync(ValueProviderFactoryContext context)
         private static async Task AddValueProviderAsync(ValueProviderFactoryContext context)
         {
             var request = context.ActionContext.HttpContext.Request;
+            IFormCollection form;
+            try
```
davidfowl: Don't catch the exception, use `request .HasFormContentType `

rosieks: I would like not to catch exception. As far as I remember `HasFormContentType` returns true for "multipart/form-data" without boundary. Also `ReadFormAsync` throws `InvalidDataException` in many other cases where it has problem with parsing form.

rosieks: I also thought about introducing new method var errors = request.TryReadFormAsync(out form) - would that be a way?

pranavkm: @Tratcher

src/Mvc/Mvc.Core/src/ModelBinding/FormValueProviderFactory.cs
```diff
@@ -33,9 +36,20 @@ public Task CreateValueProviderAsync(ValueProviderFactoryContext context)
...
+            {
+                form = await request.ReadFormAsync();
+            }
+            catch (InvalidDataException ex)
+            {
+                form = FormCollection.Empty;
+                context.ActionContext.ModelState.TryAddModelException(string.Empty, ex);
```
pranavkm: Using `string.Empty` here is really problematic, but also incorrect in this case. Can we explore the other alternative?

rosieks: Maybe it shouldn't be ModelState at all but maybe something different to add error. Eventually it's not an issue with modelstate but rather the whole request.

src/Mvc/Mvc.Core/src/ModelBinding/FormValueProviderFactory.cs
Tratcher: TryReadFormAsync would require a whole new call stack / form parsing implementation. Some of those exceptions come from the very bottom. Since forms are rarely invalid I think you're better off catching the exception.

davidfowl: Agreed. After looking at this catching might be best

mkArtakMSFT:
Closing as there has been no activity on this PR for quite a while.

rosieks:
@mkArtakMSFT - what can be done to reactivate it? @pranavkm - could you explain why `string.Empty` is incorrect/problematic and what should error should be returned instead?

pranavkm:
@rosieks the issue is that it marks the entire input space as invalid (including model values that are bound from query, body, header etc) which would the wrong outcome. There was another report in the same area where reading the form would throw if the client disconnects - https://github.com/aspnet/AspNetCore/issues/13119. My guess is we need a better way to handle these in general, rather than handle this in a one-off way.

pranavkm:
@rosieks we ultimately went with using `string.Empty` for a lack of a better alternative. The changes were merged as part of a different PR and the underlying issue was resolved for the 3.1.0 release. Closing this since the actual issue is resolved. Thanks for sending the PR!

Issue #10421, opened 2019-05-21
Update Blazor VSIX to only install on d16.2 (preview) or later


danroth27:
We are getting a lot of issues where customers are trying to work with Blazor in the release version of VS2019, which is not supported. The latest Visual Studio Preview (https://visualstudio.com/preview) is *required*. We should update the Blazor VSIX so that it can only be installed on d16.2 (preview) or later.

danroth27:
@mkArtakMSFT We should do this for Preview 6.

PR #10458, opened 2019-05-22
Updating minimum required VS version to 16.2

mkArtakMSFT:
Addresses #10421

### DIFF
--- a/src/Components/Blazor/BlazorExtension/src/source.extension.vsixmanifest
+++ b/src/Components/Blazor/BlazorExtension/src/source.extension.vsixmanifest
@@ -12,7 +12,7 @@
         <PreviewImage>Content\WebConfiguration.png</PreviewImage>
     </Metadata>
     <Installation AllUsers="true">
-         <InstallationTarget Id="Microsoft.VisualStudio.Community" Version="[16.0,)" />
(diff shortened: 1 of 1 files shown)


mkArtakMSFT (commented):
@rynowak, @NTaylorMullen if I want to target VS 16.2, should I also update the Prerequisite for `Microsoft.VisualStudio.Component.Web` to the same version too? (see line 23)

NTaylorMullen:
> @rynowak, @NTaylorMullen if I want to target VS 16.2, should I also update the Prerequisite for `Microsoft.VisualStudio.Component.Web` to the same version too? (see line 23)

As of today we don't technically require a newer version of them (they haven't made any changes to require that). However, in practice they will always be 16.2+ so it doesn't hurt.

NTaylorMullen (approved):

rynowak:
Make sure you test this, by doing a build and then double-clicking the VSIX and going through the whole install process.

mkArtakMSFT:
Sure, I've done that already, @rynowak.

mkArtakMSFT:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=198009&view=ms.vss-test-web.build-test-results-tab

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/2559

PR #10460, opened 2019-05-22
Usability cleanup in IIS tests

jkotalik:
- Combine each websites together into the same test site
- Remove x86 profile as it isn't used by IIS tests (TODO consider adding a single test for this scenario)
- Consolidates tests into better namespaces
- Move tests into the right folders
- Fixes https://github.com/aspnet/AspNetCore-Internal/issues/2222 by deleting the test; we may just want to run this as part of build on helix anyways 😄


natemcmaster:
Does this mean we can also stop installing the x86 runtime for all Windows builds? (See https://github.com/aspnet/AspNetCore/blob/f26b87aa42ee646c1e65532fbbb53dba2104196e/build/runtimes.props#L11-L15)

jkotalik:
> Does this mean we can also stop installing the x86 runtime for all Windows builds? (See
> 
> https://github.com/aspnet/AspNetCore/blob/f26b87aa42ee646c1e65532fbbb53dba2104196e/build/runtimes.props#L11-L15
> 

I have no idea. I didn't realize it was a thing to begin with.

natemcmaster:
I think IIS x86 tests were the only reason we were installing this. It's not an issue if we need to keep it or add it back later.

src/Servers/IIS/IISIntegration.sln
```diff
@@ -123,10 +123,10 @@ Global
 		{8B3446E8-E6A8-4591-AA63-A95837C6E97C}.Debug|x86.Build.0 = Debug|Any CPU
 		{8B3446E8-E6A8-4591-AA63-A95837C6E97C}.Release|Any CPU.ActiveCfg = Debug|Any CPU
 		{8B3446E8-E6A8-4591-AA63-A95837C6E97C}.Release|Any CPU.Build.0 = Debug|Any CPU
-		{8B3446E8-E6A8-4591-AA63-A95837C6E97C}.Release|x64.ActiveCfg = Release|Any CPU
-		{8B3446E8-E6A8-4591-AA63-A95837C6E97C}.Release|x64.Build.0 = Release|Any CPU
-		{8B3446E8-E6A8-4591-AA63-A95837C6E97C}.Release|x86.ActiveCfg = Release|Any CPU
-		{8B3446E8-E6A8-4591-AA63-A95837C6E97C}.Release|x86.Build.0 = Release|Any CPU
+		{8B3446E8-E6A8-4591-AA63-A95837C6E97C}.Release|x64.ActiveCfg = Debug|Any CPU
```
natemcmaster: Intentional? This means switching to release mode in VS still builds the project in debug mode.

jkotalik: Sigh no, I have a strong dislike for VS property pages...

natemcmaster (approved):
LGTM

natemcmaster:
There are some test failures in Helix which seem to be the result of this change, btw.

image

PR #10471, opened 2019-05-22
AuthorizeHelper will no-op if endpoint routing is used

### LINKED ISSUE #9216: [SignalR]: AuthorizeHelper should no-op if endpoint routing has already executed auth
Similar to the solution in #8387, we need to avoid re-running auth in SignalR if the middleware previously ran.

BrennanConroy:
Fixes #9216 

If I understand correctly, Auth will run on the Endpoint metadata (which we set the authorize metadata correctly) so internally SignalR doesn't need to run Auth again in the Endpoint code path.


src/SignalR/server/SignalR/test/EndToEndTests.cs
```diff
@@ -532,6 +532,84 @@ private async Task ServerClosesConnectionWithErrorIfHubCannotBeCreated(HttpTrans
...
+                    logger.LogInformation("Disposing Connection");
+                    await connection.DisposeAsync().OrTimeout();
+                    logger.LogInformation("Disposed Connection");
+                }
+            }
+        }
+
+        [Fact]
+        [LogLevel(LogLevel.Trace)]
+        public async Task UnauthorizedHubConnectionDoesNotConnect()
```
pranavkm: Should this have a test that calling it with the requisite auth works?

src/SignalR/common/Http.Connections/src/ConnectionEndpointRouteBuilderExtensions.cs
```diff
@@ -48,20 +48,14 @@ public static class ConnectionEndpointRouteBuilderExtensions
...
         {
             var options = new HttpConnectionDispatcherOptions();
-            // REVIEW: WE should consider removing this and instead just relying on the
-            // AuthorizationMiddleware
-            var attributes = typeof(TConnectionHandler).GetCustomAttributes(inherit: true);
-            foreach (var attribute in attributes.OfType<AuthorizeAttribute>())
-            {
-                options.AuthorizationData.Add(attribute);
-            }
             configureOptions?.Invoke(options);
```
pranavkm: In MVC we ended up combining the `AuthorizeAttribute` decorated on the controller with metadata on endpoint routing: https://github.com/aspnet/AspNetCore/blob/master/src/Mvc/Mvc.Core/src/Authorization/AuthorizeFilter.cs#L148-L152

The concern was that you could have done something like so:

```C#
[Authorize("Some-Claim")]
public class UserController
{

}

...
mvcOptions.Filters.Add(new AuthorizeFilter(new AuthorizeAttribute("Different-Claim")));
```

And you may have a auth policy that only makes sense when it observes both `Some-Claim` and `Different-Claim` together. In the SignalR case, you kinda get in to a similar place if the user configures `HttpConnectionDispatcherOptions`:

```C#
[Authorize("Some-Claim")]
public class MyHub { ..}

MapHub(..., options => options.AuthorizeData.Add(new AuthorizeAttribute("Different"));
```

We have a couple of options here:

1) Copy `HttpConnectionDispatcherOptions.AuthorizeData` to EndpointMetadata. The auth will run for the full set once and then again for the values specified in `HttpConnectionDispatcherOptions.AuthorizeData`

2) Add annotated attributes to `HttpConnectionDispatcherOptions.AuthorizeData` if the user configured it

3) Throw if the user configured  `HttpConnectionDispatcherOptions.AuthorizeData` when using endpoint routing and direct them to use endpoint routing's extension methods to configure auth.

pranavkm (commented):
cc @HaoK \ @rynowak

src/SignalR/server/SignalR/test/EndToEndTests.cs
BrennanConroy: Yeah probably, I was going to and then forgot.

src/SignalR/common/Http.Connections/src/ConnectionEndpointRouteBuilderExtensions.cs
BrennanConroy: For 1, couldn't we copy the data over and then replace the options with an empty list so it only runs once?

pranavkm: Sure, that's an option. As long the user doesn't get to look at the collection later on.

HaoK: Yeah the issue is that all the attributes play an important role determining which schemes to authenticate, so it potentially has a big impact on whether authorization succeeds

BrennanConroy: This should be fixed now, turns out we were already copying the data over to EndPoints but now I'm clearing the list so we don't run Auth again.

src/SignalR/common/Shared/TaskCache.cs
```diff
@@ -3,11 +3,11 @@
 
 using System.Threading.Tasks;
 
-namespace Microsoft.AspNetCore.SignalR.Internal
+namespace Microsoft.AspNetCore.Internal
```
halter73: Nit: Why keep this under `/src/SignalR` but remove `SignalR` from the namespace? Is it used by non-SignalR projects?

halter73: I get it now... It's used by Microsoft.AspNetCore.Http.Connections.

src/SignalR/common/Http.Connections/src/ConnectionEndpointRouteBuilderExtensions.cs
```diff
@@ -86,6 +80,19 @@ public static class ConnectionEndpointRouteBuilderExtensions
...
+            // Copy the list if there are items, we're going to clear the list on the options
+            // so that EndPoints is the only one running Auth and we don't run it again in HttpConnectionDispatcher
+            if (options.AuthorizationData.Count > 0)
+            {
+                authorizationData = new List<IAuthorizeData>(options.AuthorizationData.Count);
+                for (var i = 0; i < options.AuthorizationData.Count; i++)
+                {
+                    authorizationData.Add(options.AuthorizationData[i]);
+                }
+                options.AuthorizationData.Clear();
```
halter73: Isn't a little weird to mutate the options object? What if someone tries to reuse HttpConnectionDispatcherOptions? Is there an alternative way to ensure HttpConnectionDispatcher doesn't run auth again?

pranavkm: Or like one of my other suggestions, disallow this entirely and get users to use the endpoint routing authorization extensions: https://github.com/aspnet/AspNetCore/blob/16a47948f80fede807fabe3c291d793590e8fd17/src/Security/Authorization/Policy/src/AuthorizationEndpointConventionBuilderExtensions.cs#L14

While `HttpConnectionDispatcherOptions.AuthorizeData` is old API, the endpoint routing based extension methods are new. It doesn't seem terrible to say this feature does not work when using endpoint routing. We could also use the Startup analyzer to guide users to not use this property.

src/SignalR/common/Http.Connections/src/ConnectionsAppBuilderExtensions.cs
```diff
@@ -27,14 +24,13 @@ public static IApplicationBuilder UseConnections(this IApplicationBuilder app, A
...
-
-            var routes = new RouteBuilder(app);
-
-            configure(new ConnectionsRouteBuilder(routes, dispatcher));
-
             app.UseWebSockets();
-            app.UseRouter(routes.Build());
+            app.UseRouting();
+            app.UseAuthorization();
+            app.UseEndpoints(endpoints =>
```
BrennanConroy: @davidfowl 
Have any thoughts on this? Converting `app.UseSignalR(r => r.Map...())` to use Endpoints internally. This lets us get rid of our custom auth stuff in `HttpConnectionDispatcher`.

And side-effect, I believe will allow apps that port to 3.0 but don't change to endpoints to use the service automagically without `UseAzureSignalR`.

BrennanConroy:
Ping reviewers, I basically rewrote the change so we no longer do Auth inside `HttpConnectionDispatcher` and rely on Auth through Endpoint routing.

The one remaining thing is that `app.UseSignalR(routes => routes.MapHub<MyHub>("/path"));` doesn't set the `HubMetadata` on the Endpoint produced.

pranavkm (commented):
Auth changes kinda makes sense, but would be great if somebody more familiar with SignalR's routing signed off on this.

halter73:
> The one remaining thing is that app.UseSignalR(routes => routes.MapHub<MyHub>("/path")); doesn't set the HubMetadata on the Endpoint produced.

Can you explain why this is the case?

BrennanConroy:
>Can you explain why this is the case?

I've just made a commit with a fix, I didn't do it initially because it's a bit ugly, but I don't have a better solution right now.

src/SignalR/common/Http.Connections/test/HttpConnectionDispatcherTests.cs
```diff
@@ -1347,358 +1347,6 @@ public async Task TransferModeSet(HttpTransportType transportType, TransferForma
...
-
-                // would get stuck if EndPoint was running
-                await dispatcher.ExecuteAsync(context, options, app).OrTimeout();
-
-                Assert.Equal(StatusCodes.Status401Unauthorized, context.Response.StatusCode);
-            }
-        }
-
-        [Fact]
-        public async Task AuthenticatedUserWithoutPermissionCausesForbidden()
```
halter73: Do we still have end-to-end coverage for cases like this?

BrennanConroy: We do have some end-to-end tests for auth, for example: https://github.com/aspnet/AspNetCore/pull/10471/files#diff-c65c848cc599ae1724a72656ff60bedaR453

But we don't need to verify it as much because we now rely on the underlying auth system instead of implementing our own version.

halter73: Do we still manually verify authorization for hub method invocations to account for method-level auth attributes over a WebSocket for example? Where is that?

BrennanConroy: Yes we still do. This PR doesn't touch any of that.

Test located [here](https://github.com/aspnet/AspNetCore/blob/a26c4e936b1a775b890a8fb2d33fdefcece60518/src/SignalR/server/SignalR/test/HubConnectionHandlerTests.cs#L2156)

halter73 (approved):

BrennanConroy:
@aspnet/build What's going on?

BrennanConroy:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=206259&view=logs

dougbu:
@BrennanConroy looks like AzDO missed your comment or hit a temporary glitch.

Later, the Hello bot ignored your triage request. That bot seems to get confused when the URL doesn't exactly match a known format (see aspnet/AspNetCore-Internal#2425). In this case, `&view=logs` probably busted it.

src/SignalR/common/Http.Connections/src/ConnectionsAppBuilderExtensions.cs
rynowak: This to me seems super spicy. Does it make sense to deprecate this API now that it's literally equivalent to something else? It's goodness if we can converge all of the code that we show in samples/docs.

src/SignalR/common/Http.Connections/src/Internal/AuthorizeHelper.cs
```diff
@@ -1,69 +0,0 @@
...
-using System.Collections.Generic;
-using System.Threading.Tasks;
-using Microsoft.AspNetCore.Authentication;
-using Microsoft.AspNetCore.Authorization;
-using Microsoft.AspNetCore.Authorization.Policy;
-using Microsoft.Extensions.DependencyInjection;
-
-namespace Microsoft.AspNetCore.Http.Connections.Internal
-{
-    internal static class AuthorizeHelper
```
rynowak: It makes me super happy to see this go away. We basically have 3 version of this code rn, and this cuts it down to two.

src/SignalR/common/Http.Connections/test/HttpConnectionDispatcherTests.cs
rynowak: Yeah, I can't speak to the specific tests that are being removed, but I'd strongly encourage testing a variety of auth scenarios and interactions as a user would write them.

src/SignalR/common/Http.Connections/test/MapConnectionHandlerTests.cs
```diff
@@ -34,39 +34,80 @@ public MapConnectionHandlerTests(ITestOutputHelper output)
...
                 options => authCount += options.AuthorizationData.Count))
             {
-                builder.Start();
+                host.Start();
+
+                var dataSource = host.Services.GetRequiredService<EndpointDataSource>();
+                // We register 2 endpoints (/negotiate and /)
+                Assert.Equal(2, dataSource.Endpoints.Count);
+                Assert.Equal(1, dataSource.Endpoints[0].Metadata.GetOrderedMetadata<IAuthorizeData>().Count);
+                Assert.Equal(1, dataSource.Endpoints[1].Metadata.GetOrderedMetadata<IAuthorizeData>().Count);
```
rynowak: You can use `Assert.Collection` for this (with a sort) and verify the details that are in the comments. It's a little more verbose, but also a lot more precise.

src/SignalR/server/SignalR/src/HubRouteBuilder.cs
```diff
@@ -25,6 +28,11 @@ public HubRouteBuilder(ConnectionsRouteBuilder routes)
             _routes = routes;
         }
 
+        internal HubRouteBuilder(IEndpointRouteBuilder endpoints)
```
rynowak: Worth a comment explaining why and when this can be null?

src/SignalR/server/SignalR/test/AuthHub.cs
```diff
@@ -0,0 +1,12 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Authorization;
+
+namespace Microsoft.AspNetCore.SignalR.Tests
+{
+    [Authorize]
+    class AuthHub : Hub
+    {
+    }
```
rynowak: So is the expectation that method-level auth doesn't play with endpoints or with the authZ middleware?

rynowak (commented):
I left a few teeny comments, but this looks like what I expected to see, and this seems like valuable simplification.

BrennanConroy:
@dougbu I doubt it: https://github.com/aspnet/AspNetCore/pull/10425#issuecomment-495394631

dougbu:
@BrennanConroy that was a whole week ago ❕ 

More seriously, we don't know why the bot sometimes ignores triage requests.

src/SignalR/common/Http.Connections/src/ConnectionsAppBuilderExtensions.cs
BrennanConroy: I would be ok deprecating the old stuff
@anurse @davidfowl ?

src/SignalR/common/Http.Connections/test/MapConnectionHandlerTests.cs
```diff
@@ -126,9 +259,45 @@ void ConfigureRoutes(IEndpointRouteBuilder endpoints)
...
+            {
+                host.Start();
+
+                var dataSource = host.Services.GetRequiredService<EndpointDataSource>();
+                // We register 2 endpoints (/negotiate and /)
+                Assert.Collection(dataSource.Endpoints,
+                    endpoint =>
+                    {
+                        Assert.Equal("/path/negotiate", endpoint.DisplayName);
+                        Assert.NotNull(endpoint.Metadata.GetMetadata<IEnableCorsAttribute>());
```
BrennanConroy: Look @anurse, I did a thing!

src/SignalR/common/Http.Connections/test/MapConnectionHandlerTests.cs
BrennanConroy: I wasn't a fan when I first saw this, but after writing some, I kind of like it.

src/SignalR/common/Http.Connections/src/ConnectionsAppBuilderExtensions.cs
anurse: I'm OK marking `UseSignalR` as obsolete if that's what we're talking about. We should have done that already anyway, since we want people to use `UseEndpoints`.

src/SignalR/common/Http.Connections/test/MapConnectionHandlerTests.cs
anurse: But do you have a test that verifies CORS headers are actually set? :)

src/SignalR/common/Http.Connections/src/ConnectionsAppBuilderExtensions.cs
BrennanConroy: I would be obsoleting:
`UseConnections`
`ConnectionsRouteBuilder`
`HubRouteBuilder`
`UseSignalR`

davidfowl: Sounds good.

BrennanConroy:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=207741&view=logs

PR #10483, opened 2019-05-23
Fix TLS protocols used in tests

JunTaoLuo:
Fixes https://github.com/aspnet/AspNetCore-Internal/issues/2490

Disabling tests failing in https://github.com/aspnet/AspNetCore-Internal/issues/2490.

### DIFF
--- a/src/Servers/Kestrel/test/InMemory.FunctionalTests/HttpsTests.cs
+++ b/src/Servers/Kestrel/test/InMemory.FunctionalTests/HttpsTests.cs
@@ -381,7 +381,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.InMemory.FunctionalTests
         }
 
         [ConditionalFact]
-        [SkipOnHelix("https://github.com/aspnet/AspNetCore-Internal/issues/2490", Queues = "Windows.7.Amd64.Open")]
         public async Task OnAuthenticate_SeesOtherSettings()
         {
             var loggerProvider = new HandshakeErrorLoggerProvider();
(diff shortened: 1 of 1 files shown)


Tratcher:
This is actually a bit concerning, these tests aren't doing anything that should be OS specific.

The only difference I see is that they specify SslProtocols.None rather than SslProtocols.Tls11 | SslProtocols.Tls12.

Win7 defaults topped out at TLS 1.0. https://support.microsoft.com/en-us/help/3154520/support-for-tls-system-default-versions-included-in-the-net-framework. Kestrel defaults to 1.1 & 1.2 so this may explain the conflict.
https://github.com/aspnet/AspNetCore/blob/c2e2d7d1359ef8caf5460562adfe46b7f4f68df0/src/Servers/Kestrel/Core/src/HttpsConnectionAdapterOptions.cs#L27

Try changing the test to SslProtocols.Tls11 | SslProtocols.Tls12 instead.

src/Servers/Kestrel/test/InMemory.FunctionalTests/HttpsTests.cs
```diff
@@ -380,7 +380,7 @@ public async Task ClientAttemptingToUseUnsupportedProtocolIsLoggedAsDebug()
             Assert.Equal(LogLevel.Debug, loggerProvider.FilterLogger.LastLogLevel);
         }
 
-        [Fact]
+        [ConditionalFact]
```
BrennanConroy: Change back to `Fact` and fix those compile errors

JunTaoLuo: Gah. Should have ran it locally.

halter73 (approved):
LGTM. This matches the default value for SslProtocols in HttpsConnectionAdapterOptions, so it makes sense.

anurse:
I updated the PR to close https://github.com/aspnet/AspNetCore-Internal/issues/2490 when merged since we believe it will solve the issue.

JunTaoLuo:
Looks like the test is fixed! Thanks @Tratcher.

JunTaoLuo:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=199321

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/2570

PR #10487, opened 2019-05-23
Integrate AuthorizeView with actual authorization

SteveSandersonMS:


src/Components/Blazor/Blazor/src/Services/WebAssemblyConsoleLogger.cs
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.Blazor.Services
+{
+    internal class WebAssemblyConsoleLogger<T> : ILogger<T>
```
SteveSandersonMS: This is intended as a minimal way of meeting the requirements for the built-in default `IAuthorizationService`, without bringing in the full-fledged `Microsoft.AspNetCore.Extensions.Logging`.

The full logging package adds about 100KB to the app size and a couple extra assembly downloads, so if we can have a bare-bones one without it that's preferable. People who want to use the full thing (e.g., to control logging levels, have multiple categories, etc.) can still reference it and add it to their DI config.

src/Components/Blazor/Blazor/src/Hosting/WebAssemblyHostBuilder.cs
```diff
@@ -102,6 +104,10 @@ private void CreateServiceProvider()
                 };
             });
 
+            // Needed for authorization
+            services.AddOptions();
+            services.TryAdd(ServiceDescriptor.Singleton(typeof(ILogger<>), typeof(WebAssemblyConsoleLogger<>)));
```
rynowak: I think you'll also have to support `ILoggerFactory`. You're going to find code in our stack that expects that, and it should be easy to add.

src/Components/Blazor/Blazor/src/Services/WebAssemblyConsoleLogger.cs
rynowak: >The full logging package adds about 100KB to the app size and a couple extra assembly downloads, so if we can have a bare-bones one without it that's preferable. People who want to use the full thing (e.g., to control logging levels, have multiple categories, etc.) can still reference it and add it to their DI config.

To me this seems like a great thing to chase down in the future when we have a TFM for web assembly. Right now if we wanted to offer a thinner profile of these features for Blazor we have no real way to do so.

src/Components/Components/src/Auth/AuthorizeView.razor
```diff
@@ -54,15 +83,17 @@ else
...
     {
-        // TODO: Support various authorization condition parameters, equivalent to those offered
-        // by the [Authorize] attribute, e.g., "Roles" and "Policy". This is on hold until we're
-        // able to reference the policy evaluator APIs from this package.
-
-        return currentAuthenticationState.User?.Identity?.IsAuthenticated == true;
+        var policy = await AuthorizationPolicy.CombineAsync(
+            AuthorizationPolicyProvider, selfAsAuthorizeData);
+        var result = await AuthorizationService.AuthorizeAsync(user, Resource, policy);
+        return result.Succeeded;
```
rynowak: @HaoK

src/Components/Components/src/Microsoft.AspNetCore.Components.csproj
```diff
@@ -10,6 +10,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <Reference Include="Microsoft.AspNetCore.Authorization" />
```
rynowak: I don't have a better idea right now, but are we really OK with baking this in at the lowest level. We're saying **all** components present and future have access to the ASP.NET Authorization library

rynowak (approved):
:shipit: from me. it would be good if @HaoK can look as well.

src/Components/Blazor/Blazor/src/Hosting/WebAssemblyHostBuilder.cs
SteveSandersonMS: I know it is fairly easy to add, but we don't have any code paths that hit it currently, and I'm trying to be minimal with this. We're always free to add more default services later.

If you have reasons to think it's going to be problematic if we don't, let me know, otherwise I'll leave this as-is.

src/Components/Components/src/Microsoft.AspNetCore.Components.csproj
SteveSandersonMS: Yep, I agree with both the uneasiness and the conclusion that this is the best outcome currently. 

Considering that all the Components/Blazor packages have `AspNetCore` in their name, it doesn't really change anything to reference `Microsoft.AspNetCore.Authorization` as compared with if we had `Microsoft.AspNetCore.Components.Authorization`.

src/Components/Components/src/Auth/AuthorizeView.razor
SteveSandersonMS: This is the same as what `DefaultHubDispatcher` does, which looks more or less equivalent in terms of its requirements and its position in the stack.

src/Components/Blazor/Blazor/src/Hosting/WebAssemblyHostBuilder.cs
rynowak: I can't really articulate a specific scenario. This just feels lazy - because we know what we would do if people complain. We could prevent them from having to log and issue and wait for us to fix it by just doing a little thing now.

SteveSandersonMS:
**TODO** Manual ref-assembly code for the new params on `AuthorizeView`

src/Components/Components/src/Auth/AuthorizeView.razor
HaoK: So just to confirm I'm reading this correctly, selfAsAuthorizeData is total set of authorize attributes for this view that you want to use for the authZ check.

SteveSandersonMS: Yes (except this info doesn’t come from attributes in this case- that’s a separate feature on the Blazor router).

HaoK (approved):
This looks pretty reasonable

src/Components/Blazor/Blazor/src/Hosting/WebAssemblyHostBuilder.cs
javiercn: Yeah, we want to have a similar set of services in both environments, otherwise when I try to use a component that does logging, then I end up having to register it myself, making the experience different in the two environments.

src/Components/Components/src/Microsoft.AspNetCore.Components.csproj
javiercn: Yeah, this makes me a bit sad on the inside. I'm wondering if we could use some of the infrastructure for single file publish coming in 3.0 to separate the core of the platform from the baked-in components that we give you into two assemblies and simply ILMerge them at publish time.

javiercn (commented):
Looked at the changes, they seem fine except for a couple of comments. Still have to look at the tests

src/Components/Blazor/Blazor/src/Hosting/WebAssemblyHostBuilder.cs
SteveSandersonMS: OK, I've added a super-minimal `ILoggerFactory`.

> Yeah, we want to have a similar set of services in both environments

I'd be cautious about thinking about it like that. The environments have very different capabilities, goals, and characteristics. Most of the server-side services are for server-side application-server concerns, so a large proportion of them won't be present on WebAssembly by design.

Issue #10317, opened 2019-05-17
SkipStatusCodePagesAttribute should run before AuthorizeAttribute


huan086:
### Is your feature request related to a problem? Please describe.
`[SkipStatusCodePages]` is meant to be used in actions that are API calls, so that the StatusCodePagesMiddleware does not interfere with the response status code and body.

API actions are almost always decorated with `[Authorize]`. When user is not authorized, AuthorizeFilter short circuits and returns 401. Due to the short circuit, IResourceFilter, which SkipStatusCodePagesAttribute inherits, does not run, thus StatusCodePagesMiddleware runs and modifies the status code and body. The API caller does not receive 401 with empty body.

### Describe the solution you'd like
Ideally, the StatusCodePagesMiddleware does not run when `[SkipStatusCodePages]`, thus the API caller receives 401 with empty body.

This can be achieved by having SkipStatusCodePagesAttribute inherit from IAlwaysRunResultFilter instead.

### Describe alternatives you've considered
Modifying the middleware pipeline with custom middleware. But this dissociates the action that needs SkipStatusCodePages from the code that does the work

mkArtakMSFT:
Thanks for contacting us, @huan086. Would you like to send a PR for this? We'd happily consider it!

huan086:
I'm thinking we need to implement IOrderedFilter as well, so that when multiple IAlwaysRunResultFilter are present and some of them short-circuits, it'll be possible to make sure SkipStatusCodePagesAttribute runs first. @mkArtakMSFT what do you think?

kevinchalet:
@mkArtakMSFT is there a chance this bug could be fixed in 5.0? It sadly makes `[SkipStatusCodePages]` hard to use in mixed API/views applications using token authentication.

mkArtakMSFT:
@kevinchalet, as much as I wish we could, I don't think we will get to this.

trannamtrung1st:
Any update on this? Currently I'm using a custom middleware. It seems a little bit weird but it's working. I can pass my custom options into the middleware constructor.

```
public class ApiStatusCodeMiddleware
    {
        private readonly RequestDelegate _next;

        public ApiStatusCodeMiddleware(RequestDelegate next)
        {
            _next = next;
        }

        public async Task InvokeAsync(HttpContext context)
        {
            await _next(context);
            var isApiRequest = context.Request.Path.StartsWithSegments("/api");
[...]

PR #10490, opened 2019-05-23
Change SkipStatusCodePagesAttribute to inherit from IAlwaysRunResultFilter

huan086:
Summary of the changes
 - Change SkipStatusCodePagesAttribute to inherit from IAlwaysRunResultFilter instead of IResourceFilter
-  Add IOrderedFilter to SkipStatusCodePagesAttribute so that callers can ensure it executes first

Addresses #10317

### DIFF
--- a/src/Mvc/Mvc.ViewFeatures/src/SkipStatusCodePagesAttribute.cs
+++ b/src/Mvc/Mvc.ViewFeatures/src/SkipStatusCodePagesAttribute.cs
@@ -11,8 +11,15 @@ namespace Microsoft.AspNetCore.Mvc
     /// A filter that prevents execution of the StatusCodePages middleware.
     /// </summary>
     [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, AllowMultiple = false, Inherited = true)]
-    public class SkipStatusCodePagesAttribute : Attribute, IAlwaysRunResultFilter
+    public class SkipStatusCodePagesAttribute : Attribute, IAlwaysRunResultFilter, IOrderedFilter
     {
+        /// <inheritdoc />
+        /// <summary>
+        /// Gets or sets the order value for determining the order of execution of filters. Filters execute in
+        /// ascending numeric value of the <see cref="Order" /> property.
+        /// </summary>
+        public int Order { get; set; } = -1000;
+
         /// <inheritdoc />
         public void OnResultExecuted(ResultExecutedContext context)
         {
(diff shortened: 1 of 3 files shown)


src/Mvc/Mvc.ViewFeatures/src/SkipStatusCodePagesAttribute.cs
```diff
@@ -11,15 +11,23 @@ namespace Microsoft.AspNetCore.Mvc
     /// A filter that prevents execution of the StatusCodePages middleware.
     /// </summary>
     [AttributeUsage(AttributeTargets.Class | AttributeTargets.Method, AllowMultiple = false, Inherited = true)]
-    public class SkipStatusCodePagesAttribute : Attribute, IResourceFilter
+    public class SkipStatusCodePagesAttribute : Attribute, IAlwaysRunResultFilter, IOrderedFilter
```
pranavkm: While I think this change makes sense, this does 

a) is a breaking change
b) does not work with endpoint routing

We should leave the `IResourceFilter` on there since it does not hurt to invoke the filter in multiple stages and allows it to be non-breaking. In addition, as of 3.0, Auth happens as part of the middleware pipeline, so this change really doesn't help there. If you'd like to solve it for the latter, one solution would be to:

a) Introduce an `ISkipStatusCodePagesMetadata`
b) Use the presence of the metadata to skip the status code page if you're doing endpoint routing. Here's a similar pattern: https://github.com/aspnet/AspNetCore/blob/master/src/Middleware/CORS/src/Infrastructure/CorsMiddleware.cs#L140-L159

pranavkm (changes requested):

src/Mvc/Mvc.ViewFeatures/src/SkipStatusCodePagesAttribute.cs
pranavkm: If you want to consider skipping solving this for endpoint metadata, let me know and I can file a separate issue to track it. cc @JamesNK \ @rynowak \ @Tratcher

huan086:
Cool, let me revert the changes and follow the way CorsMiddleware does it

mkArtakMSFT:
Thanks for your effort, @huan086 .
Do you plan to resolve @pranavkm 's comments here, or should this be closed?

huan086:
I haven't have the time to make the necessary changes yet

pranavkm:
Feel free to re-send this PR once you've had a chance to update it per PR comments.

Issue #10445, opened 2019-05-22
Router displays NotFoundContent without any layout


SteveSandersonMS:
Since we changed `Router` to have a `NotFoundContent` fragment instead of referencing a "fallback page" type, we've lost the convenience of having the content rendered inside a layout. It just renders very plainly (typically, black Times New Roman on a white screen).

This might not seem super important for the "not found" message, but we're also about to add "authorizing..." and "not authorized" content here too, and those really should be rendered inside a layout or the app will look flickery and weird as it jumps between displays styles.

Possible solution:

 * We add an extra `DefaultLayout` property to `Router` (of type `System.Type`, just like we used to for `FallbackComponent`).
    * Usage in template: `<Router ... DefaultLayout="typeof(MainLayout)">`
 * We also add the same `DefaultLayout` property to the `LayoutDisplay` component, and have `Router` automatically pass the value through
 * Then we update the rendering logic for `LayoutDisplay` so that, when the target doesn't define any layout, we use the `DefaultLayout` type
 * Finally, we make `LayoutDisplay` able to receive a `RenderFragment` as an alternative to a page type, i.e., it supports either. Then `Router` can render its `NotFoundContent` by passing the render fragment to `LayoutDisplay`

SteveSandersonMS:
This is fixed (or at least, it's no longer applicable) with #12800

PR #10491, opened 2019-05-23
Integrate authorization into Blazor router

SteveSandersonMS:
~~Tests are in progress, but this does work.~~ **Update: tests are done and this is ready.** This builds on my previous PR, #10487

Posting the PR a bit early in case anyone sounds any alarms about the way it determines what authorization attributes to apply (I've done it in a very simple way, but it's hard to compare against the giant metadata system used for controllers/pages).


src/Components/Components/src/Auth/AuthorizeView.cs
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Authorization;
+
+namespace Microsoft.AspNetCore.Components
+{
+#pragma warning disable PUB0001 // Pubternal type in public API
```
SteveSandersonMS: Because `.razor` files can't generate either `abstract` or `internal` yet, so this is the best I can do to discourage use.

src/Components/Components/src/Layouts/PageDisplay.cs
```diff
@@ -0,0 +1,141 @@
...
+using Microsoft.AspNetCore.Components.Layouts;
+using Microsoft.AspNetCore.Components.RenderTree;
+
+namespace Microsoft.AspNetCore.Components
+{
+    /// <summary>
+    /// Displays the specified page component, rendering it inside its layout
+    /// and any further nested layouts, plus applying any authorization rules.
+    /// </summary>
+    public class PageDisplay : IComponent
```
SteveSandersonMS: Renamed from `LayoutDisplay` to `PageDisplay`, because this now deals with all the concerns of displaying a `@page` component that you reached via routing. It doesn't just wrap the result in the chain of layouts; it now also respects any authorization metadata on the page.

SteveSandersonMS:
**TODO** Update the manual ref-assembly code

SteveSandersonMS:
@rynowak @danroth27 

On a design point, I want to emphasise that when authorization fails, although we don't display the page **we do still use the layouts nominated by that page**. This is because what other layout would we use? We can either:

 * Decide this is sensible and carry on
 * Decide we think the page's layout should not be used because it might contain sensitive content, and instead have a `DefaultLayout` on the `Router`, and use that instead. We also have a need for a `DefaultLayout` to fix #10445, so maybe it's the right way forwards.

I'd lean towards the second option, but not blocking the merging of this PR for it.

src/Components/Components/src/Auth/AuthorizeView.cs
javiercn: Is that in the roadmap?

src/Components/Components/src/Auth/AuthorizeViewCore.razor
```diff
@@ -1,4 +1,8 @@
-@namespace Microsoft.AspNetCore.Components
+@*
+    Ideally, this would be an internal (and probably abstract) class. So temporarily
+    this is in an "internal" namespace and should become actually internal in the future.
+*@
+@namespace Microsoft.AspNetCore.Components.Internal
```
javiercn: Is this is causing us so much "grief". Why don't we author this as a plain c# class here? The markup is simple and avoids all the problems you mention.

src/Components/Components/src/Routing/Router.cs
```diff
@@ -78,9 +88,11 @@ private string StringUntilAny(string str, char[] chars)
...
         protected virtual void Render(RenderTreeBuilder builder, Type handler, IDictionary<string, object> parameters)
         {
-            builder.OpenComponent(0, typeof(LayoutDisplay));
-            builder.AddAttribute(1, LayoutDisplay.NameOfPage, handler);
-            builder.AddAttribute(2, LayoutDisplay.NameOfPageParameters, parameters);
+            builder.OpenComponent(0, typeof(PageDisplay));
+            builder.AddAttribute(1, nameof(PageDisplay.Page), handler);
+            builder.AddAttribute(2, nameof(PageDisplay.PageParameters), parameters);
+            builder.AddAttribute(3, nameof(PageDisplay.NotAuthorizedContent), NotAuthorizedContent);
+            builder.AddAttribute(4, nameof(PageDisplay.AuthorizingContent), AuthorizingContent);
```
javiercn: This starts to feel a lot like "middleware". I think that longer term (before 3.0) we should find a better design that frees the router from these concerns.

src/Components/Components/src/Auth/AuthorizeView.cs
SteveSandersonMS: I remember seeing an issue for it somewhere, but it's not commited to any ETA.

src/Components/Components/src/Routing/Router.cs
SteveSandersonMS: I felt exactly the same, and posted a proposal about this yesterday: https://github.com/aspnet/AspNetCore/issues/10493

src/Components/Components/src/Auth/AuthorizeViewCore.razor
SteveSandersonMS: You're right - this is better as a plain C# class. Done, and removed the pubternalness since it's now `abstract`.

src/Components/Components/src/Auth/AuthorizeView.cs
rynowak: yeah it's not tracked for this release.

CornedBee:
This moved PageLayout out of the Layouts namespace, but failed to update the file location or the ref assembly.

danroth27:
@CornedBee Thanks for letting us know! Could you please open an issue for this so that we can track it?

CornedBee:
@danroth27 #11240 is essentially this, except the reporter apparently found a workaround for his use case and closed the bug. Maybe you can reopen it? If not, I can create a new one.

danroth27:
@CornedBee Thanks for bringing this to our attention! I've reopened #11240 so that we can address this.

Issue #10307, opened 2019-05-16
Fix MessagePack JS package peer dependency


anurse:
The `@aspnet/signalr-protocol-msgpack` package has a peer dependency on `@aspnet/signalr`. This is good, but the value is static and [hasn't been updated since `1.0.0-preview3`](https://unpkg.com/@aspnet/signalr-protocol-msgpack@1.1.0/package.json). We should either add automation to update the peer dependency on every build, or just remove it, or maybe make it a real dependency.

anurse:
It should almost definitely just be a real dependency, in which case we can use a path link and I think the build output will be correct.

OzBob:
Confirmed.
It also, stops MessagePack messages between JavaScript client and server.

npm i @aspnet/signalr@next
//which installs v "^3.0.0-preview5-19227-01"
npm i @aspnet/signalr-protocol-msgpack@next
//which installs v "^3.0.0-preview5-19227-01"
BUT WARNS:

> "npm i @aspnet/signalr-protocol-msgpack@3.0.0-preview5-19227-01 requires a peer of @aspnet/signalr@^1.0.0-preview3 but none is installed. You must install peer dependencies yourself."

that version is NOT available on https://www.npmjs.com/package/@aspnet/signalr

BrennanConroy:
The warning can be ignored for now, we'll be fixing it up soon.

Those 2 packages should be compatible. Could you share a repro app that shows the issue?

anurse:
This is more complicated than I hoped. If you set it to `link:../signalr` or `file:../signalr`, that's exactly what goes in to the package.json, which isn't what we want.

To do this right I think we need to rewrite the package.json during build... I'll tinker with that.

anurse:
@pranavkm is looking at fixing this as part of his send queue change.

anurse:
Should be fixed in #10506. TODO: Verify in output build.

BrennanConroy:
Reopening as the fix was reverted due to build issues

anurse:
Assigning you to this @pranavkm but only because you've been looking at it in your PR. Let me know if you need help or someone to take this over!

pranavkm:
Fixed as part of https://github.com/aspnet/AspNetCore/commit/d935ea2d2553d48887b968638726470adf0c2b3c

OzBob:
npm i @aspnet/signalr-protocol-msgpack@next

//still has similar warning about compatibility:

[...]

PR #10506, opened 2019-05-23
Add an option to SignalR to guarantee ordering of send events

### LINKED ISSUE #9901: Blazor (server-side) System.InvalidOperationException: Received a notification for a rendered batch when not expecting it.

pranavkm:
Fixes https://github.com/aspnet/AspNetCore/issues/9901


src/SignalR/clients/ts/signalr/src/HubConnection.ts
```diff
@@ -854,7 +854,7 @@ export class HubConnection {
 
     private isObservable(arg: any): arg is IStreamResult<any> {
         // This allows other stream implementations to just work (like rxjs)
-        return arg.subscribe && typeof arg.subscribe === "function";
+        return arg && arg.subscribe && typeof arg.subscribe === "function";
```
pranavkm: Null refs

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -147,7 +148,13 @@ export class HttpConnection implements IConnection {
         }
 
         // Transport will not be null if state is connected
-        return this.transport!.send(data);
+        const send = () => this.transport!.send(data);
```
pranavkm: Are there other places that needs to be fixed?

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -147,7 +148,13 @@ export class HttpConnection implements IConnection {
...
+        if (this.options.ensureMessageSendOrder) {
+            this.sendQueue = this.sendQueue.then(send, send);
```
pranavkm: How should this work if the send fails?

src/Components/Browser.JS/src/Boot.Server.ts
```diff
@@ -110,8 +110,10 @@ async function initializeConnection(options: Required<BlazorOptions>, circuitHan
   (hubProtocol as unknown as { name: string }).name = 'blazorpack';
 
   const connectionBuilder = new signalR.HubConnectionBuilder()
-    .withUrl('_blazor')
-    .withHubProtocol(hubProtocol);
+    .withUrl('_blazor', {
+      ensureMessageSendOrder: true
```
halter73: Are we making this non-default because of perf considerations? If so, that could probably be mitigated by batching multiple sends per post requests if we start falling behind.

I like making this the default, because it provides more consistency between the WebSocket transport vs. SSE/LP.

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
halter73: I think we should continue dequeuing after send failures. The error should be reported back to the caller though.

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
halter73: I don't think so.

halter73: I wonder if it wouldn't be better to put the data itself in a queue and create an async send loop that dequeues the data. If there's no data available, the loop could start waiting on a promise that gets tripped on the next send.

The benefit of this is that batching sends into a single post should be made much easier. I talked to @davidfowl about this, and I think we can just concatenate/append all the data and send it in a single call to transport.send which should result in a single post.

src/SignalR/clients/ts/signalr/src/HubConnection.ts
halter73: Good catch. Is there a regression test?

pranavkm: I haven't written any tests as yet. I'll get a few done once I had the design out of the way

src/Components/Browser.JS/src/Boot.Server.ts
pranavkm: @BrennanConroy suggested making this opt-in due to perf considerations. If we do want to make this default:

* Is the value "true" in the options by default
* Or is there an inverse option and the user has a gesture to turn it off?  `disableHttpTransportSendQueue: true`?

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
pranavkm: That seems reasonable. What do I do to send a batch of data - send an array?

jkotalik:
Should blazor.js be there? I can't review this PR at all...

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
anurse: Hrm. A send loop seems tricky to do right in the single-threaded JavaScript environment

src/Components/Browser.JS/package.json
```diff
@@ -11,8 +11,7 @@
     "test": "jest"
   },
   "devDependencies": {
-    "@aspnet/signalr": "^1.0.0",
-    "@aspnet/signalr-protocol-msgpack": "^1.0.0",
```
rynowak: should `@aspnet/signalr` still be here?

src/Components/test/testassets/ComponentsApp.App/Shared/NavMenu.razor
```diff
@@ -37,11 +37,6 @@
                 <span class="oi oi-list-rich" aria-hidden="true"></span> Error
             </NavLink>
         </li>
-        <li class="nav-item px-3">
-            <NavLink class="nav-link" href="links">
-                <span class="oi oi-list-rich" aria-hidden="true"></span> Links
-            </NavLink>
-        </li>
```
rynowak: ???

src/Components/Browser.JS/src/Boot.Server.ts
halter73: Ideally, we get the perf good enough with this change that we don't need an option to disable it. If that doesn't work, I think opt-out is better than opt-in.

BrennanConroy: I don't remember complaining about perf here.

The main reason for opt-in is that it's a change in behavior, although it's a change in behavior to be more correct.
I'm ok with it being on by default. cc @anurse

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
BrennanConroy: Now that this is a delegate you can't safely assume `this.transport` is not null

src/Components/Browser.JS/package.json
pranavkm: We get it transitively from signalr-protocol-msgpack

src/Components/test/testassets/ComponentsApp.App/Shared/NavMenu.razor
pranavkm: The link doesn't exist. When I was trying out my super awesome script to test this: https://github.com/aspnet/AspNetCore/pull/10431#issuecomment-494602536, navigating to this link results in a 404 so I ended up removing it.

src/SignalR/clients/ts/signalr-protocol-msgpack/package.json
```diff
@@ -41,10 +41,8 @@
     "dist/**/*",
     "src/**/*"
   ],
-  "peerDependencies": {
-    "@aspnet/signalr": "^1.0.0-preview3"
-  },
   "dependencies": {
+    "@aspnet/signalr": "file:../signalr",
```
pranavkm: @anurse since you were investigating this.

anurse: We can't put a `file:` reference in here as-is. NPM doesn't translate this at all during packing (it ends up in the final package as a `file:` link to a path that doesn't exist). If we do this, we have to rewrite the dependency manually at pack time (in `eng/targets/Npm.targets`)

I couldn't find anything built-in to help us out here :(.

anurse: There are tools like [Lerna](https://github.com/lerna/lerna) that some repos use to handle this, but I'm hesitant to bring a whole new tool into play here for this...

anurse:
Preview 6 deadline is looming and we still want to validate this change. @pranavkm do you expect to be able to land this today or early tomorrow?

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -501,3 +502,66 @@ export class HttpConnection implements IConnection {
...
+        }
+
+        return this.sendCore(transport, data);
+    }
+
+    private async sendCore(transport: ITransport, data: string | ArrayBuffer) {
+        this.sending = true;
+
+        let resolve: () => void;
+        this.promiseQueue = new Promise((r) => resolve = r);
```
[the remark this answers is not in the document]
halter73: As long as failures get reported to the caller, which it looks like it does, I think it's fine for the send queue to continue on failure.

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -501,3 +502,70 @@ export class HttpConnection implements IConnection {
...
+
+        try {
+            await transport.send(data);
+        } finally {
+            this.sending = false;
+            this.promiseQueue = undefined;
```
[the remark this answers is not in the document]
pranavkm: We should set up a new promise in `send` if we aren't sending. Nothing really would be using this once this is no longer buffering.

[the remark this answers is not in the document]
halter73: I agree. I don't know why you'd clear this.

An a tangentially related note, it might be nicer if it wasn't nullable, and it was initialized to a completed promise.

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -501,3 +502,70 @@ export class HttpConnection implements IConnection {
...
+
+    private bufferData(transport: ITransport, data: string | ArrayBuffer) {
+        if (this.buffer.length) {
+            if (typeof(this.buffer[0]) !== typeof(data)) {
+                throw new Error(`Expected data to be of type ${typeof(this.buffer)} but was of type ${typeof(data)}`);
+            }
+        } else {
+            // For the first entry, update the promise to queue up a buffer send
+            this.promiseQueue = this.promiseQueue!
+                .then(() => this.sendBufferedData(transport));
```
halter73: I think this would be easier to follow if this was written as an async loop instead of using `.then()`.

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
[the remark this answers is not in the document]
BrennanConroy: Are `.then` callbacks run inline when a promise finishes?

If no, then there is a race here when a callback has been registered and another send comes in right after a send completes.

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -501,3 +502,70 @@ export class HttpConnection implements IConnection {
...
+
+        let resolve: () => void;
+        this.promiseQueue = new Promise((r) => resolve = r);
+
+        try {
+            await transport.send(data);
+        } finally {
+            this.sending = false;
+            this.promiseQueue = undefined;
+            resolve!();
```
halter73: The next call to sendBufferedData won't be called inline. Effectively it's called with a 0 setTimeout. If there's an intervening call to TransportSendQueue.send() it will see that this.sending is false and allow concurrent/out-of-order data.

The fix for this is usually to unset the flag (if necessary since more data could have been buffered while you were being scheduled) after the continuation awaiting the promise starts. In this case that would be sendBufferedData, but it's possible that there's no queued call to sendBufferedData.

That's one reason writing the async loop might be cleaner.

pranavkm (commented):
🆙 📅

src/SignalR/clients/ts/signalr-protocol-msgpack/package.json
pranavkm: I updated the Npm.targets to support rewriting the file. Unfortunately `ReadAllLines` returns a sequence of trimmed lines, so the formatting for the package.json is lost. Is that something to be concerned about?

eng/targets/Npm.Common.targets
```diff
@@ -53,6 +53,23 @@
 
     <Copy SourceFiles="$(PackageJson)" DestinationFiles="$(_BackupPackageJson)" />
 
+    <ReadLinesFromFile File="$(PackageJson)" Condition="@(ReplacePackageContent->Count()) != 0">
```
pranavkm: cc @natemcmaster

src/SignalR/clients/ts/signalr-protocol-msgpack/package.json
anurse: I am not worried about that. It's not ideal, but it's OK.

eng/targets/Npm.Common.targets
anurse: This is awesome @pranavkm. I'm not a huge fan of how we have to do it, but I'm glad to knock #10307 out as part of this.

For context @natemcmaster : Neither NPM nor Yarn seem to have any mechanism to have a "file" reference (i.e. Project Reference) and then convert it to a version reference during pack. So we have to do it in MSBuild...

There are tools ([Lerna](https://github.com/lerna/lerna)) that can make this better, but they are complicated and I don't think adding new tools is a great idea ;).

src/SignalR/clients/ts/signalr-protocol-msgpack/signalr-protocol-msgpack.npmproj
```diff
@@ -12,5 +12,9 @@
     <ProjectReference Include="..\signalr\signalr.npmproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <ReplacePackageContent Include="file:../signalr" ReplaceWith="$(PackageVersion)" />
```
anurse: We should use an npm version range here (see [npm-semver for syntax](https://docs.npmjs.com/misc/semver)). If we want the same behavior we have in NuGet, we need to explicitly specify `>=`. Alternatively we could use `~` which only allows the patch version to float. I think for now, let's match our NuGet behavior:

```suggestion
    <ReplacePackageContent Include="file:../signalr" ReplaceWith=">=$(PackageVersion)" />
```

analogrelay (commented):
Go ahead and mark this PR as fixing #10307 as well! Two birds with one stone!

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -501,3 +511,96 @@ export class HttpConnection implements IConnection {
...
     return !requestedTransport || ((actualTransport & requestedTransport) !== 0);
 }
+
+export class TransportSendQueue {
+    private buffer: any[] = [];
+    private sendBufferedData: PromiseSource;
+    private executing: boolean = true;
+    private transportResult: PromiseSource;
+
+    constructor(private transport: ITransport) {
```
BrennanConroy: Does this make a member variable on the class and automatically assign it?

Not sure how I feel about it...

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -501,3 +511,96 @@ export class HttpConnection implements IConnection {
...
+    private bufferData(data: string | ArrayBuffer): void {
+        if (this.buffer.length && typeof(this.buffer[0]) !== typeof(data)) {
+            throw new Error(`Expected data to be of type ${typeof(this.buffer)} but was of type ${typeof(data)}`);
+        }
+
+        this.buffer.push(data);
+        this.sendBufferedData.resolve();
+    }
+
+    private async sendLoop(): Promise<void> {
```
BrennanConroy: Nice! This does looks a lot cleaner

src/SignalR/clients/ts/signalr/tests/HttpConnection.test.ts
```diff
@@ -844,3 +845,180 @@ describe("HttpConnection", () => {
...
+
+        const queue = new TransportSendQueue(transport);
+
+        const first = queue.send("Hello");
+        // This should allow first to enter transport.send
+        promiseSource1.resolve();
+        // Wait until we're inside transport.send
+        await promiseSource2;
+
+        // These two operations should get queued.
```
BrennanConroy: nit: comment is slightly wrong

src/SignalR/clients/ts/signalr/tests/HttpConnection.test.ts
```diff
@@ -844,3 +845,180 @@ describe("HttpConnection", () => {
...
```
BrennanConroy: nit: comment is slightly wrong

src/SignalR/clients/ts/signalr/tests/HttpConnection.test.ts
```diff
@@ -844,3 +845,180 @@ describe("HttpConnection", () => {
         });
     });
 });
+
+describe("TransportSendQueue", () => {
```
BrennanConroy: Add a test for trying to queue two different types, `string` and `ArrayBuffer`

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -501,3 +511,96 @@ export class HttpConnection implements IConnection {
...
+        for (const item of arrayBuffers) {
+            result.set(new Uint8Array(item), offset);
+            offset += item.byteLength;
+        }
+
+        return result;
+    }
+}
+
+class PromiseSource {
```
BrennanConroy: We already have this class somewhere

pranavkm: It's in the test `Utiils`.  I do feel slightly icky about deriving from `Promise` when technically what we need is closer to `SemaphoreSlim.WaitAsync`

pranavkm:
🆙 📅

BrennanConroy (approved):
Should add a test for `queue.stop()`

eng/targets/Npm.Common.targets
natemcmaster: Unfortunately, I don't think this will work on macOS/Linux due to https://github.com/microsoft/msbuild/issues/1024.

Minimal repro of the problem -- run this on Linux. The output has the backslash flipped to a forward slash. `{ "key": "value /" with a quote" }`.
```xml
<Project>
    <Target Name="Build">
        <PropertyGroup>
            <Content>{ "key": "value \" with a quote" }</Content>
        </PropertyGroup>

        <WriteLinesToFile Lines="$(Content)" File="out1.txt" Overwrite="true" />
    </Target>
</Project>
```

pranavkm: Is there a way to write custom tasks with arcade? Additionally, could this wait for the next release? I don't think we publish npm packages built from xplat environments to npmjs, do we?

src/Components/Browser.JS/Microsoft.AspNetCore.Components.Browser.JS.npmproj
```diff
@@ -7,23 +7,11 @@
...
-  <!-- Override the default 'Build' target from eng/targets/Npm.Common.targets. -->
-  <Target Name="Build"
-          Inputs="@(WebpackInputs)"
-          Outputs="@(WebPackOutputs)">
-    <RemoveDir Directories="dist" />
-
-    <Yarn Command="run build:debug" />
-    <Yarn Command="run build:production" />
-  </Target>
+  <Target Name="Restore" />
```
SteveSandersonMS: I'm sure this is all for a good reason but could you explain what it is?

Previously the `WebpackInputs`/`WebPackOutputs` were for incremental build. Is that handled in another way now, or are we losing that feature?

Is the `ProjectReference` to `signalr-protocol-msgpack.npmproj` just a way of ensuring the SignalR library builds first, or are we changing to a source reference instead of an NPM reference now?

src/Components/Browser.JS/package.json
```diff
@@ -5,14 +5,16 @@
...
+    "prebuild": "yarn run clean && yarn install --mutex network",
     "build": "yarn run build:debug && yarn run build:production",
     "build:debug": "cd src && webpack --mode development --config ./webpack.config.js",
     "build:production": "cd src && webpack --mode production --config ./webpack.config.js",
     "test": "jest"
   },
   "devDependencies": {
-    "@aspnet/signalr": "^1.0.0",
-    "@aspnet/signalr-protocol-msgpack": "^1.0.0",
+    "@aspnet/signalr-protocol-msgpack": "file:../../SignalR/clients/ts/signalr-protocol-msgpack",
```
SteveSandersonMS: Would this be problematic if we choose to publish the Blazor package to NPM? When someone references our package, where is NPM going to pull this dependency from?

NPM has other (but less convenient) ways of doing source-to-source references during development that interact properly with package references in production, i.e., `npm link`. I'm not saying that's necessarily what we need to do right now, but let's be sure this is not creating a problem for ourselves.

src/Components/Browser.JS/utils/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "common",
+  "private": true,
+  "version": "0.0.1",
+  "description": "Common dependencies used during dev time. DO NOT PUBLISH",
+  "devDependencies": {
+    "rimraf": "^2.6.2"
+  }
+}
```
SteveSandersonMS: I don't know why this is being added. Let's discuss.

src/Components/Browser.JS/package.json
SteveSandersonMS: OK, I see from another comment lower down that there's some kind of `package.json`-rewriting trickery going on to deal with this. Having so much customization over the build process is a worrisome trap we keep falling into, but I don't necessarily have a better suggestion in this case.

If you could at least confirm that the rewriting is what avoids this problem that would be good to know!

src/Components/Browser.JS/Microsoft.AspNetCore.Components.Browser.JS.npmproj
pranavkm: Sorry, did not see your comments until after I merged. There's a couple of issues here: since we're compiling SignalR source files, we have to also include that closure to determine incrementalism. However these project files currently don't have a way to specify that. Secondly, restoring a `file://` results in the directory being copied to the `node_modules`. Arcade discovers these directories and attempts to build them, which works out poorly since none of the other files \ folders required to build these project files exist. 

Building this is much slower than earlier, is that something we want to try track and address?

src/Components/Browser.JS/utils/package.json
pranavkm: Another issue of the `file://` and maybe I'm just going about it wrong. On Windows, a reference with `file://` simply gets copied to the `node_modules` directory. Changing SignalR source files doesn't cause `yarn install` to invalidate or update it. The build currently cleans up the node_modules and runs `yarn install` as part building, this file exists to install `rimraf`

src/Components/Browser.JS/Microsoft.AspNetCore.Components.Browser.JS.npmproj
SteveSandersonMS: > Building this is much slower than earlier, is that something we want to try track and address?

Yes. Obviously not for preview 6, but this is worth following up on.

We're taking on a lot of technical debt around NPM. The amount of nonstandardness in our build system, and its tendency to always increase over time, is problematic.

As soon as preview 6 ships and hence SignalR's NPM package is on the public npmjs.org, can we go back to having a regular package dependency on the published package? These two areas of the product should be pretty decoupled and it would help if we could eliminate the `package.json`-rewriting and source-to-source nonstandardness.

src/Components/Browser.JS/utils/package.json
SteveSandersonMS: > On Windows, a reference with file:// simply gets copied to the node_modules directory.

In that case it doesn't seem like a valid way of having a dependency at all. An NPM package, once published, isn't necessarily in the same shape as its pre-publish source files. We wouldn't know whether our packages are really going to work when published.

How pressing is our need to do source-to-source references for NPM packages?

 * If this is the only case, we could sidestep the issue entirely (at least for the forseeable future) by waiting for the SignalR package to be published (preview 6) then reverting to a package reference.
 * Longer term, if we can accept non-atomic updates, we could continue using package references but with whatever private NPM registry feed our CI pushes to. Some PRs would have to be two-phase, which I know isn't ideal, but keeps us completely in the standard happy path for NPM/Yarn.
 * If we absolutely have to have atomic repo-wide PRs, we could have the build publish NPM packages (`.tgz` files) to some artifacts location, presumably with a timestamp as the version, and reference them locally from there. It's not standard, but it's closer to standard than `file://`+hacks.

eng/targets/Npm.Common.targets
natemcmaster: Arcade doesn't support a 'repotasks' convention. We don't publish npm packages from xplat, but we would like for the repo to build correctly on all platforms.

src/Components/Browser.JS/utils/package.json
anurse: > * If we absolutely have to have atomic repo-wide PRs, we could have the build publish NPM packages (`.tgz` files) to some artifacts location, presumably with a timestamp as the version, and reference them locally from there. It's not standard, but it's closer to standard than `file://`+hacks.

That is certainly an ideal way to do it, unfortunately my understanding is that it would take a lot of hacking Arcade to do it. I believe pack phase that generates the `.tgz` happens after the build phase for all the projects completes.

> How pressing is our need to do source-to-source references for NPM packages?

Since you embed the source directly, not doing a source-to-source reference would effectively mean you'd always be at least a build behind (since you'd only be able to use the publish outputs from the previous build at best). If we did the `.tgz` hack above we could get something a little cleaner and a little better, but I'm not sure what it would take to get there.

> An NPM package, once published, isn't necessarily in the same shape as its pre-publish source files.

This is true, but generally only in relation to `npmignore` files or `package.json`'s `files` element. NPM in fact requires that the source effectively be a superset of what is packaged, by the time `npm pack` is run. The only thing you can really do is filter **out** certain files via an exclude list (`npmignore`) or an include list ([`files`](https://docs.npmjs.com/files/package.json#files)).

SteveSandersonMS: > Since you embed the source directly, not doing a source-to-source reference would effectively mean you'd always be at least a build behind

Yep, that's what I meant with "non-atomic updates" (sorry for being obscure).

While not perfect, I think this is a better situation to be in than fighting against both NPM/Yarn and Arcade and having weird hacks like extra `package.json` roots to work around restore issues.

Perhaps longer term we could request the functionality we need in Arcade to support multiple NPM packages properly (dropping `.tgz` artifacts as each one is built, so they can take package dependencies on each other). Or we go all-in on lerna or a similar alternative that addresses this in a sane way.

anurse: > While not perfect, I think this is a better situation to be in

I'm not sure I agree. How would you get a coherent final build in this scenario? You'd be shipping 3.0 RTM while embedding the previous build of SignalR. In order to fix a Blazor-blocking bug in SignalR you'd either have to do two build cycles (somehow disabling Blazor in the first cycle) or manually flow the dependency (which feels hackier than what we're doing here). Not to mention the fact that we have extra lag introduced in even detecting possibly-breaking bugs. Blazor test failures could be due to a change introduced in an earlier build, causing confusion when investigating them.

It seems really odd to me to be shipping as a single product unit in the same repo and yet have dependent versions drifting between components within that same repo.

Also, it's not at all uncommon to have to use extra tools to manage this stuff. Angular's packages are [full of placeholder versions](https://github.com/angular/angular/blob/815443313044e0ec0ec87af588d09eda6b88f0c8/packages/platform-server/package.json#L15) that get replaced. They do things a little differently, but the premise is the same.

SteveSandersonMS: OK, you're probably right that the drawbacks of that approach outweigh the benefits. I'll be interested to see how the alternative `link://`-based approach that @pranavkm has changed this to works out.

Issue #8235, opened 2019-03-06
Make it so Helix runs can be rerun and preserve history


jkotalik:
Right now, if you rerun a helix build after a build failure, the previous build that failed will be deleted. On VSTS, you see both attempts, but helix overrides it. 

cc @anurse

anurse:
cc @HaoK

natemcmaster:
_Build Team Triage:_ seems like this should be possible. If there is a build variable for the iteration number, we could add it to the Helix job name to make it unique from prior iterations. Putting into the next sprint for further investigation.

anurse:
IIRC there isn't a variable for iteration number but we had the though of capturing a timestamp at the beginning of the run and using that timestamp as part of the job name.

PR #10510, opened 2019-05-24
Make it so Helix runs can be rerun and preserve history R2

JunTaoLuo:
Addresses #8235

### DIFF
--- a/eng/helix/helix.proj
+++ b/eng/helix/helix.proj
@@ -19,7 +19,7 @@
   <PropertyGroup Condition=" '$(CI)' == 'true' ">
     <HelixType>ci</HelixType>
     <Creator>aspnetcore</Creator>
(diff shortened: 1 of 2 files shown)


eng/helix/helix.proj
```diff
@@ -19,7 +19,7 @@
   <PropertyGroup Condition=" '$(CI)' == 'true' ">
     <HelixType>ci</HelixType>
     <Creator>aspnetcore</Creator>
-    <HelixBuild>$(BUILD_BUILDNUMBER)</HelixBuild>
+    <HelixBuild>$(BUILD_BUILDNUMBER).$(BUILD_BUILDID)</HelixBuild>
```
JunTaoLuo: Doesn't work since the BUILDID doesn't increment between re-runs. Will look into another mechanism.

dougbu (approved):

PR #10511, opened 2019-05-24
Increase stack size for IIS Inprocess

### LINKED ISSUE #9394: Increase stack size for host when running in IIS In-Proc

Blocked by https://github.com/dotnet/coreclr/issues/21450

Once there's an API/Property to increase the stack size, we need to react to it in IIS in two ways:

1. Increase the default to something more appropriate for our scenario
2. Make it configurable in web.config (if possible).

jkotalik:
Fixes https://github.com/aspnet/AspNetCore/issues/9394. Adds a handler setting to override it.

The only thing I'd consider changing is making it so we preserve the CS_* environment variables if they are already set.

### DIFF
--- a/src/Servers/IIS/AspNetCoreModuleV2/InProcessRequestHandler/InProcessOptions.h
+++ b/src/Servers/IIS/AspNetCoreModuleV2/InProcessRequestHandler/InProcessOptions.h
@@ -112,6 +112,12 @@ public:
         return m_bindingInformation;
     }
 
+    std::wstring
+    QueryStackSize() const
+    {
+        return m_strStackSize;
+    }
+
     InProcessOptions(const ConfigurationSource &configurationSource, IHttpSite* pSite);
 
     static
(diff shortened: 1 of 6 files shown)


src/Servers/IIS/AspNetCoreModuleV2/InProcessRequestHandler/InProcessOptions.h
```diff
@@ -112,6 +112,12 @@ class InProcessOptions: NonCopyable
         return m_bindingInformation;
     }
 
+    std::wstring
```
BrennanConroy: nit: return `const std::wstring &`

BrennanConroy (approved):

src/Servers/IIS/IIS/test/testassets/InProcessWebSite/Startup.cs
```diff
@@ -696,6 +696,29 @@ private async Task ShutdownStopAsyncWithCancelledToken(HttpContext ctx)
...
+        private async Task StackSizeLarge(HttpContext ctx)
+        {
+            // This would normally stackoverflow if we didn't increase the stack size per thread.
+            RecursiveFunction(30000);
+            await ctx.Response.WriteAsync("Hello World");
+        }
+
+        private void RecursiveFunction(int i)
+        {
+            if (i == 0)
```
anurse: Psh, I feel like you should at least calculate Pi or some Fibonacci numbers or something ;P

PR #10528, opened 2019-05-24
Changing `@functions` block to `@code` block

mkArtakMSFT:
Addresses https://github.com/aspnet/AspNetCore/issues/10527


ryanbrandenburg:
Can we get a bit more background here? Is it the case that both will work (for a time) but we recommend `@code` over `@functions` for functional or stylistic reasons?

mkArtakMSFT:
@ryanbrandenburg yes, both are supported, and we want to use simpler `@code` block for Blazor templates and leave `@functions` block for razor scenarios

ryanbrandenburg (commented):
Looks fine to me, definitely have someone with Blazor knowledge review also though.

rynowak:
>Can we get a bit more background here? Is it the case that both will work (for a time) but we recommend @code over @functions for functional or stylistic reasons?

`@functions` will also be supported forever. This is old Razor feature that's been around for years and we have no reason to break it.

rynowak:
I think you may have missed the client side templates in src/Components/Blazor/Templates/src/content

mkArtakMSFT:
Double checked and apparently I had missed just two files, but the Blazor templates have been already updated 🆙📅

BrennanConroy:
Bad rebase?

mkArtakMSFT:
BAD REBASE. PL:EASE HOLD OFF FROM REVIEWING - FIXING

PR #10540, opened 2019-05-25
Target netstandard2.1 for SignalR client

### LINKED ISSUE #9525: Update SignalR client to Target .Net Standard 2.1
We are currently cross targeting to netcoreapp3.0 for `IAsyncEnmerable` support. It's included in .Net Standard https://github.com/dotnet/standard/pull/1017 we should move to cross target that instead

https://github.com/aspnet/AspNetCore/blob/507a765dfb078f446c445318e7b55ad9f7f5cbe0/src/SignalR/clients/csharp/Client.Core/src/Microsoft.AspNetCore.SignalR.Client.Core.csproj#L5

### LINKED ISSUE #10607: Implement IAsyncDisposable in Connection.Abstractions
Epic https://github.com/aspnet/AspNetCore/issues/10869

This requires adding a dependency on the new Microsoft.Bcl.AsyncInterfaces on netstandard 2.0 and not having that dependency on netstandard 2.1.

- [ ] - ConnectionContext
- [ ] - IConnectionListener 

As part of this we want to **remove IDisposable from DefaultConnectionContext**

BrennanConroy:
Fixes https://github.com/aspnet/AspNetCore/issues/9525
Fixes https://github.com/aspnet/AspNetCore/issues/10607


src/SignalR/clients/csharp/Http.Connections.Client/src/HttpConnection.cs
```diff
@@ -586,22 +587,15 @@ private void CheckDisposed()
 
         private static bool IsWebSocketsSupported()
         {
-#if NETCOREAPP3_0
-            // .NET Core 2.1 and above has a managed implementation
-            return true;
-#else
-            var isWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
-            if (!isWindows)
+            try
```
BrennanConroy: I don't think we can do a compile time check here, for example, someone running on netcoreapp3.0 but targeting the netstandard2.0 library on windows 7 will have access to the managed websockets impl, but if we just hardcoded netstandard2.0 as not having websockets then a scenario that should work will not work.

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -49,7 +49,7 @@ public partial class HubConnection
         };
 
         private static readonly MethodInfo _sendStreamItemsMethod = typeof(HubConnection).GetMethods(BindingFlags.NonPublic | BindingFlags.Instance).Single(m => m.Name.Equals(nameof(SendStreamItems)));
-#if NETCOREAPP3_0
+#if NETSTANDARD2_1
```
BrennanConroy: Maybe all the `#if NETSTANDARD2_1` checks should be `#if !NETSTANDARD2_0`, that way we might never have to update them in the future.

src/SignalR/common/Shared/AsyncEnumerableAdapters.cs
```diff
@@ -24,13 +24,26 @@ public static IAsyncEnumerable<T> MakeCancelableTypedAsyncEnumerable<T>(IAsyncEn
             return new CancelableTypedAsyncEnumerable<T>(asyncEnumerable, cts);
         }
 
+#if NETCOREAPP3_0
         public static async IAsyncEnumerable<object> MakeAsyncEnumerableFromChannel<T>(ChannelReader<T> channel, [EnumeratorCancellation] CancellationToken cancellationToken = default)
         {
             await foreach (var item in channel.ReadAllAsync(cancellationToken))
```
BrennanConroy: `channel.ReadAllAsync` isn't available on `netstandard2.1`, is this something we should ask CoreFX to provide or...?

BrennanConroy (commented):
I didn't change a couple libraries like `Http.Connections.Common` because the reason we cross-targeted that one was to not pull in the `System.Text.Json` package on `netcoreapp3.0`

src/SignalR/common/Shared/AsyncEnumerableAdapters.cs
davidfowl: Yes

BrennanConroy: We might get this for free with https://github.com/dotnet/corefx/issues/34963

Directory.Build.targets
```diff
@@ -81,9 +81,12 @@
...
-      <DefaultRuntimeFrameworkVersion Condition="'$(IsServicingBuild)' != 'true'">$(MicrosoftNETCoreAppPackageVersion)</DefaultRuntimeFrameworkVersion>
-      <TargetingPackVersion Condition="'$(IsServicingBuild)' != 'true'">$(MicrosoftNETCoreAppPackageVersion)</TargetingPackVersion>
+      <!-- Always update the 'latest version', whether the repo is servicing or not. -->
+      <LatestRuntimeFrameworkVersion Condition="'%(TargetFramework)' == '$(DefaultNetCoreTargetFramework)'">$(MicrosoftNETCoreAppPackageVersion)</LatestRuntimeFrameworkVersion>
+      <!-- Only update the default runtime version for preview builds. -->
+      <DefaultRuntimeFrameworkVersion Condition="'%(TargetFramework)' == '$(DefaultNetCoreTargetFramework)' and '$(IsServicingBuild)' != 'true'">$(MicrosoftNETCoreAppPackageVersion)</DefaultRuntimeFrameworkVersion>
+      <!-- Only update the targeting pack version for preview builds. -->
+      <TargetingPackVersion Condition="'%(TargetFramework)' == '$(DefaultNetCoreTargetFramework)' and '$(IsServicingBuild)' != 'true'">$(MicrosoftNETCoreAppPackageVersion)</TargetingPackVersion>
```
dougbu: Suggest moving the consistent `Condition="'%(TargetFramework)' == '$(DefaultNetCoreTargetFramework)'"` to the `<KnownFrameworkReference>` element as is done below.

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
dougbu: Doubt it matters. We can't be sure at this point when we'll drop .NET Standard 2.0 support, update 2.1 to something newer, add a new TFM, &hellip;

JamesNK: What you can do is have ifdefs based on the available feature.

For example, this test could be `#if HAS_ASYNCENUMERABLE`. Then in the csproj against `NETSTANDARD2_1` you would define `HAS_ASYNCENUMERABLE`.

Now when a new target is added in the future, e.g. netcoreapp50, you just add it to the csproj and set the required flags. No need to update individual source code files.

JamesNK: Example: https://github.com/JamesNK/Newtonsoft.Json.Bson/blob/a1db92678e7e72eb733f37079f3a93bfb6215338/Src/Newtonsoft.Json.Bson/Newtonsoft.Json.Bson.csproj#L36-L47

BrennanConroy: We know that all targets after netstandard2.0 will have these new APIs so I don't see why we'd add extra upkeep when a simple `!NETSTANDARD2_0` check will do. And in the future if we remove netstandard2.0 compatibility then we just remove the `#if` entirely.

If for some crazy reason we do have future targets that remove features like `IAsyncEnumerable` then we can revisit and add feature flags.

src/SignalR/clients/csharp/Client.Core/ref/Microsoft.AspNetCore.SignalR.Client.Core.csproj
```diff
@@ -11,8 +11,8 @@
     <Reference Include="Microsoft.Extensions.Logging"  />
     <Reference Include="System.Threading.Channels"  />
   </ItemGroup>
-<ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp3.0'">
```
davidfowl: We might want all 3.

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
JamesNK: Future targets will never remove IAsyncEnumerable but future targets will add new features that you will likely want to use. With multiple targets things start getting complicated quickly...

A new feature in 3.0: `#if !(NETSTANDARD2_0 || NETSTANDARD2_1)`
A new feature in 4.0: `#if !(NETSTANDARD2_0 || NETSTANDARD2_1 || NETSTANDARD3_0)`
etc

Meanwhile using feature flags are a single test that and easy to read and all visible in the csproj to see what each target supports.

src/SignalR/clients/csharp/Client.Core/ref/Microsoft.AspNetCore.SignalR.Client.Core.csproj
BrennanConroy: Haha, ok why?

BrennanConroy:
We might be delaying this until preview7.

I'll probably spin up a PR with just the infrastructure changes.

anurse:
At this point unless we're super clear on what we're doing, deferring to preview 7 seems safest :).

BrennanConroy:
🆙 📅 This should be good now

src/SignalR/clients/csharp/Http.Connections.Client/ref/Microsoft.AspNetCore.Http.Connections.Client.netstandard2.1.cs
```diff
@@ -0,0 +1,93 @@
...
+    {
+        public NoTransportSupportedException(string message) { }
+    }
+    public partial class TransportFailedException : System.Exception
+    {
+        public TransportFailedException(string transportType, string message, System.Exception innerException = null) { }
+        public string TransportType { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
+    }
+}
+namespace Microsoft.AspNetCore.Http.Connections.Client.Internal
```
davidfowl: We need to make these internal 😄

src/SignalR/clients/csharp/Http.Connections.Client/src/HttpConnection.cs
davidfowl: This kinda sucks....

src/SignalR/clients/csharp/Http.Connections.Client/ref/Microsoft.AspNetCore.Http.Connections.Client.netstandard2.1.cs
BrennanConroy: https://github.com/aspnet/AspNetCore/issues/10169 !

BrennanConroy:
Reviewers?

My main concern with this is that we only test the netcoreapp3.0 versions of the binaries.

src/SignalR/common/Shared/AsyncEnumerableAdapters.cs
```diff
@@ -24,13 +24,26 @@ public static IAsyncEnumerable<T> MakeCancelableTypedAsyncEnumerable<T>(IAsyncEn
             return new CancelableTypedAsyncEnumerable<T>(asyncEnumerable, cts);
         }
 
+#if NETCOREAPP3_0
```
anurse: Is this not possible in `NETSTANDARD2_1`?

analogrelay (commented):
I only see one place where we are conditional on `netcoreapp3.0` **without** also supporting `netstandard2.1`, what's the value of keeping that TFM in place? Is there actually different code we want to write?

src/SignalR/common/Shared/AsyncEnumerableAdapters.cs
BrennanConroy: `channel.ReadAllAsync` isn't in netstandard2.1 currently. And they seem to not be interested in moving corefx to netstandard2.1 currently. https://github.com/dotnet/corefx/issues/34963

BrennanConroy:
>what's the value of keeping that TFM in place?

@davidfowl is wanting to put the client in the shared framework.

anurse:
> is wanting to put the client in the shared framework.

Does that require the `netcoreapp3.0` TFM? I wouldn't think it should.

src/SignalR/common/Shared/AsyncEnumerableAdapters.cs
JamesNK: If you want an API added to netstandard2.1 you should request it on dotnet/standard, e.g. https://github.com/dotnet/standard/issues/1144

anurse: Channels isn't in `netstandard` at all, it's a library that targets `netstandard2.0` (and doesn't cross-target 2.1).

My understanding of that issue though is that corefx isn't going to bulk-update to target 2.1. I think it doesn't preclude targeting 2.1 in a particular assembly.

If we really think `netstandard2.1` is worth it here then I guess this is why we need to triple-target. It just seems counter-productive to keep doing this.

davidfowl:
> Does that require the netcoreapp3.0 TFM? I wouldn't think it should.

I want no unnecessary deps when running on netcoreapp3.0 as well.

src/SignalR/common/Shared/AsyncEnumerableAdapters.cs
JamesNK: This is what the world of developing libraries that run everywhere is like.

btw I still think you should use feature flags. If makes this stuff easier https://github.com/aspnet/AspNetCore/pull/10540#discussion_r287878654

#VoiceOfExperience

anurse:
> I want no unnecessary deps when running on netcoreapp3.0 as well.

What unnecessary deps are there?

davidfowl:
> What unnecessary deps are there?

1.  System.Threading.Channels is in the 3.0 shared framework.

I was also assuming we'd use the Microsoft.Bcl.AsyncInterfaces on ns2.0

anurse:
Right but why would embedding this package in the shared framework be a problem if all it's deps are in the shared fx? I don't understand how a separate build actually solves what seems to be a deployment problem. When this is embedded in the shared framework, it's dependencies will already be there so it should only add this single assembly. Even if we use the `netstandard2.1` version that would be true.

The comment above about `ChannelReader.ReadAllAsync` not being present in `netstandard2.1` may be the real reason we should triple-target but I don't see how your concerns warrant triple-targeting...

eng/SharedFramework.External.props
```diff
@@ -78,6 +78,10 @@
     <ExternalAspNetCoreAppReference Include="@(_TransitiveExternalAspNetCoreAppReference)" />
   </ItemGroup>
 
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+    <_CompilationOnlyReference Include="Microsoft.Bcl.AsyncInterfaces" />
```
BrennanConroy: @natemcmaster Here is a fun one, should I put the same comment as below, but modified slightly because the package is irrelevant on `>netstandard2.0`?

natemcmaster: I don't know what this assembly is. Is it like a polyfill for IAsyncEnumerable on ns2.0?

BrennanConroy: Yeah

natemcmaster: Then yes, it's probably ok the classify this the same as the other compilation-only refs.

src/SignalR/common/Shared/AsyncEnumerableAdapters.cs
```diff
@@ -32,9 +32,11 @@ await foreach (var item in channel.ReadAllAsync(cancellationToken))
             }
         }
 #else
+        // System.Threading.Channels.ReadAllAsync() is not available on netstandard2.0 and netstandard2.1
+        // But this is the exact same code that it uses
```
anurse: Was just going to ask for this after our convo 👍

BrennanConroy:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=222484

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/2200

davidfowl (approved):

src/SignalR/clients/csharp/Http.Connections.Client/src/HttpConnection.cs
```diff
@@ -586,20 +583,18 @@ private void CheckDisposed()
...
             {
-                // Assume other OSes have websockets
+                new ClientWebSocket().Dispose();
                 return true;
             }
-            else
+            catch
```
halter73: Do we know what get thrown in practice?

BrennanConroy: `NotSupportedException`

halter73 (approved):

src/SignalR/clients/csharp/Http.Connections.Client/src/HttpConnection.cs
halter73: I figured. This is gross, but catching only the `NotSupportedException` is probably even worse.

BrennanConroy:
@aspnet/build Can someone merge please. Seems some kind of AzDo and GitHub integration is broken, everything is passing if you manually check.

src/SignalR/clients/csharp/Http.Connections.Client/src/HttpConnection.cs
davidfowl: Do we cache this now?

PR #10541, opened 2019-05-25
Fix broken build

BrennanConroy:
Guess we shouldn't be merging 3 month old PRs without rebuilding 😆 
https://github.com/aspnet/AspNetCore/pull/7227

### DIFF
--- a/src/Servers/Kestrel/samples/PlaintextApp/Startup.cs
+++ b/src/Servers/Kestrel/samples/PlaintextApp/Startup.cs
@@ -27,7 +27,7 @@ namespace PlaintextApp
                 response.ContentType = "text/plain";
                 response.ContentLength = payload.Length;
 
-                return response.BodyPipe.WriteAsync(payload).GetAsTask();
+                return response.BodyWriter.WriteAsync(payload).GetAsTask();
             });
         }
 


jkotalik (approved):

Pilchie:
Thanks for being on top of it @BrennanConroy

jkotalik:
> Guess we shouldn't be merging 3 month old PRs without rebuilding 😆

This is a legitimate problem though, even with shorter time gaps. There are many times where we have merged a PR shortly after another PR has been merged and it causes the build to break, even though the PR is green. 

A few general ideas on how we can make this experience better:

- Have a bot that can detect when there is broken build and immediately merge a revert commit.
- Have a PR check that is retriggered every time something is committed to master. It only does builds though and not tests.
- Have a check that says a branch is stale and blocks merging the PR until the check has been rerun.

anurse:
Ah crap, my bad. I respun a different PR and thought that I had done so with this one.

anurse:
I also feel like we should try to do a better job of closing/handling stale PRs. For the most part, after a month or so without merging a (non-draft) PR, it should probably just be closed. If the change is still desired, it's easy to either re-open or open a new PR.

Pilchie:
> Have a check that says a branch is stale and blocks merging the PR until the check has been rerun.

GitHub has a built-in option for this, but in practice, it ends up being very frustrating to enable, unless PR checks complete in a *very* short time.

anurse:
I think what GitHub has is that it invalidates PR checks whenever something new is pushed to the target branch. I agree that is too intense here. Something that says "you can't merge a week-old/month-old/etc PR unless you respin all the checks" might be reasonable if it could be done.

Pilchie:
At one point @dpoeschl had a Chrome Ext that warned when the checks were stale, but agree that just a bit of diligence is probably enough

Tratcher:
I've hit this several times with compiler breaks and behavior breaks. Preventing both kinds requires very invasive mechanics and would make normal development almost imposible. As the fixes have always been trivial, quick notification and reaction should be sufficient. Can the official build automatically email the committer on failure?

BrennanConroy:
It should already do that. At least I get them when I merge a PR and it fails.

anurse:
Yep, it's all configurable in AzDO settings. The default should be emailing you on failure though.

natemcmaster:
We have ways of working around GitHub limitations. We could, for example, use the aspnet-github-bot account to add a required PR check that asserts the PR check passed within the last 7 days.

jkotalik:
>  We could, for example, use the aspnet-github-bot account to add a required PR check that asserts the PR check passed within the last 7 days.

That was exactly my thought 👍

anurse:
Yep, that would be the way to move forward if we felt like it added enough value to prioritize it over other stuff. I'm just not sure it's the main issue we're facing right now ;P.

PR #10551, opened 2019-05-27
Enable mono url workaround

lewing:
This works around mono/mono#14630 for client side Blazor without requiring a new mono bump by turning off a System.Uri difference some platforms.


SteveSandersonMS (approved):
Thanks Larry!

PR #10560, opened 2019-05-27
Added support for unix domain sockets to the sockets transport

### LINKED ISSUE #4764: Allow Unix Sockets on Windows
😮 https://blogs.msdn.microsoft.com/commandline/2017/12/19/af_unix-comes-to-windows/

### LINKED ISSUE #4768: Consider supporting unix domain sockets with SocketTransport
Seems like it's possible. 

https://github.com/dotnet/corefx/blob/958511d023231f120109967484fb63ae1837e6b8/src/System.IO.Pipes/src/System/Net/Sockets/UnixDomainSocketEndPoint.cs

https://github.com/dotnet/corefx/blob/b5d0c3277dc384a23fee2e57a0b5c232c99dbd88/src/System.IO.Pipes/src/System/IO/Pipes/NamedPipeServerStream.Unix.cs#L290-L301

It would be cleaner if we kept using the socket APIs and could just copy that end point logic over to kestrel. Ideally we wouldn't have to use the pipe API to support this.'

This might be the only problem https://github.com/dotnet/corefx/blob/958511d023231f120109967484fb63ae1837e6b8/src/System.IO.Pipes/src/System/Net/Sockets/UnixDomainSocketEndPoint.cs#L25

/cc @stephentoub

davidfowl:
- Added a test for UnixDomainSockets on both transports

Fixes #4768
Fixes #4764

PS: This is on top of #10321. This will be rebased once that's in.

### DIFF
--- a/src/Servers/Kestrel/Transport.Sockets/src/SocketConnectionListener.cs
+++ b/src/Servers/Kestrel/Transport.Sockets/src/SocketConnectionListener.cs
@@ -31,10 +31,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets
             SocketTransportOptions options,
             ISocketsTrace trace)
         {
-            Debug.Assert(endpoint != null);
-            Debug.Assert(endpoint is IPEndPoint);
-            Debug.Assert(trace != null);
-
             EndPoint = endpoint;
             _trace = trace;
             _options = options;
@@ -66,9 +62,12 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets
                 throw new InvalidOperationException(SocketsStrings.TransportAlreadyBound);
             }
 
-            // TODO: Add support for UnixDomainSocket
+            Socket listenSocket;
 
-            var listenSocket = new Socket(EndPoint.AddressFamily, SocketType.Stream, ProtocolType.Tcp);
+            // Unix domain sockets are unspecified
+            var protocolType = EndPoint is UnixDomainSocketEndPoint ? ProtocolType.Unspecified : ProtocolType.Tcp;
+
+            listenSocket = new Socket(EndPoint.AddressFamily, SocketType.Stream, protocolType);
 
             // Kestrel expects IPv6Any to bind to both IPv6 and IPv4
             if (EndPoint is IPEndPoint ip && ip.Address == IPAddress.IPv6Any)
@@ -99,7 +98,12 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets
                 try
                 {
                     var acceptSocket = await _listenSocket.AcceptAsync();
-                    acceptSocket.NoDelay = _options.NoDelay;
+
+                    // Only apply no delay to Tcp based endpoints
+                    if (acceptSocket.LocalEndPoint is IPEndPoint)
+                    {
+                        acceptSocket.NoDelay = _options.NoDelay;
+                    }
 
                     var connection = new SocketConnection(acceptSocket, _memoryPool, _schedulers[_schedulerIndex], _trace, _options.MaxReadBufferSize, _options.MaxWriteBufferSize);
 
--- a/src/Servers/Kestrel/test/FunctionalTests/UnixDomainSocketsTests.cs
+++ b/src/Servers/Kestrel/test/FunctionalTests/UnixDomainSocketsTests.cs
@@ -1,0 +1,117 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Buffers;
+using System.IO;
+using System.Linq;
+using System.Net.Sockets;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Builder;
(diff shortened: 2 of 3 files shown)


src/Servers/Kestrel/Transport.Sockets/src/SocketConnectionListener.cs
```diff
@@ -66,9 +62,16 @@ internal void Bind()
                 throw new InvalidOperationException(SocketsStrings.TransportAlreadyBound);
             }
 
-            // TODO: Add support for UnixDomainSocket
+            Socket listenSocket;
```
gfoidl: Maybe like this
```c#
Socket listenSocket = new Socket(EndPoint.AddressFamily, SocketType.Stream, EndPoint is UnixDomainSocketEndPoint ? ProtocolType.Unspecified : ProtocolType.Tcp);
```
?

davidfowl: Yep, I was thinking about that after I saw this PR 😄

anurse:
@davidfowl Do you expect to be able to get this in and verified e2e by Friday? Assuming not, we should wait until preview 7 (opens next week) to merge.

davidfowl:
> @davidfowl Do you expect to be able to get this in and verified e2e by Friday? Assuming not, we should wait until preview 7 (opens next week) to merge.

Yes, I expect it to be in today or tomorrow.

src/Servers/Kestrel/test/FunctionalTests/UnixDomainSocketsTests.cs
```diff
@@ -0,0 +1,117 @@
...
+                var hostBuilder = TransportSelector.GetWebHostBuilder()
+                    .UseKestrel(o =>
+                    {
+                        o.ListenUnixSocket(path, builder =>
+                        {
+                            builder.Run(EchoServer);
+                        });
+                    })
+                    .ConfigureServices(AddTestLogging)
+                    .Configure(c => { });
```
halter73: Out of curiosity, is this line necessary?

src/Servers/Kestrel/test/FunctionalTests/UnixDomainSocketsTests.cs
```diff
@@ -0,0 +1,117 @@
...
+                        socket.Connect(new UnixDomainSocketEndPoint(path));
+
+                        var data = Encoding.ASCII.GetBytes("Hello World");
+                        socket.Send(data);
+
+                        var buffer = new byte[data.Length];
+                        var read = 0;
+                        while (read < data.Length)
+                        {
+                            read += socket.Receive(buffer, read, buffer.Length - read, SocketFlags.None);
```
halter73: I know it's just a test, but I'd avoid blocking. We've seen threadpool issues in the past...

src/Servers/Kestrel/Transport.Sockets/src/SocketConnectionListener.cs
```diff
@@ -31,10 +31,6 @@ internal sealed class SocketConnectionListener : IConnectionListener
             SocketTransportOptions options,
             ISocketsTrace trace)
         {
-            Debug.Assert(endpoint != null);
```
jkotalik: `endpoint != null` is still a valid assert I believe.

src/Servers/Kestrel/test/FunctionalTests/UnixDomainSocketsTests.cs
davidfowl: Yea, the webhost throws without it.

src/Servers/Kestrel/test/FunctionalTests/UnixDomainSocketsTests.cs
```diff
@@ -0,0 +1,117 @@
...
+    {
+#if LIBUV
+        [OSSkipCondition(OperatingSystems.Windows, SkipReason = "Libuv does not support unix domain sockets on Windows.")]
+#else
+        [OSSkipCondition(OperatingSystems.Windows, WindowsVersions.Win7, WindowsVersions.Win8, WindowsVersions.Win81, WindowsVersions.Win2008R2, SkipReason = "UnixDomainSocketEndPoint is not supported on older versions of Windows")]
+#endif
+        [ConditionalFact]
+        public async Task TestUnixDomainSocket()
+        {
+            var path = Path.GetTempFileName();
```
halter73: This seems to be creating the file.

davidfowl:
@halter73 @jkotalik is this good to go?

src/Servers/Kestrel/test/FunctionalTests/UnixDomainSocketsTests.cs
jkotalik: Yeah this is a bit jank.

jkotalik (approved):

src/Servers/Kestrel/test/FunctionalTests/UnixDomainSocketsTests.cs
anurse: Especially when you're using it just to get a name and then immediately deleting it. This is a bit of a flake-farm waiting to happen... why not just use `Path.Combine(Path.GetTempPath(), $"{Guid.NewGuid():N}.socket")` or something?

PR #10572, opened 2019-05-28
Remove use of GetRuntimeInterfaceMap in IsTypeAwaitable

MichalStrehovsky:
`GetRuntimeInterfaceMap` is a rather expensive API and it doesn't look like the use here is warranted.

The replacement should have equivalent behavior.


MichalStrehovsky:
Cc @jkotas

davidfowl:
So this should actually improve the performance of this API 😄 ?

MichalStrehovsky:
> So this should actually improve the performance of this API 😄 ?

We can call it a perf fix - it removes some useless array allocations and the extra work that `GetInterfaceMap` does and we don't need...

PR #10586, opened 2019-05-28
Add inproc handler to shared framework

jkotalik:
The inproc handler wasn't being included in the shared framework. I _believe_ this will fix it, but will verify for sure.

Reported by CTI. Adding a check to make sure that CI always includes the native assets.

### DIFF
--- a/build/repo.props
+++ b/build/repo.props
@@ -3,6 +3,13 @@
     <TargetRuntimeIdentifier Condition="'$(TargetRuntimeIdentifier)' == ''">$(TargetOsName)-$(TargetArchitecture)</TargetRuntimeIdentifier>
   </PropertyGroup>
 
+  <PropertyGroup Condition=" '$(BuildAllProjects)' == 'true' ">
+    <BuildNative>true</BuildNative>
+    <BuildManaged>true</BuildManaged>
+    <BuildNodeJS>true</BuildNodeJS>
+    <BuildJava>true</BuildJava>
+  </PropertyGroup>
+
   <PropertyGroup>
     <!-- This repo does not have solutions to build -->
     <DisableDefaultTargets>true</DisableDefaultTargets>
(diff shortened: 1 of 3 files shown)


src/Framework/src/Microsoft.AspNetCore.App.Runtime.csproj
```diff
@@ -202,6 +202,12 @@ This package is an internal implementation of the .NET Core SDK and is not meant
     <Warning Text="This project has native dependencies which were not built. Without this, tests may not function correctly. Run `build.cmd -BuildNative -BuildManaged` to build both C# and C++." />
   </Target>
 
+    <Target Name="_ErrorAboutUnbuiltNativeDependenciesOnCI"
+          BeforeTargets="Build"
+          Condition=" '$(BuildIisNativeProjects)' != 'true' AND $(CI) == 'true' AND '$(TargetOsName)' == 'win' ">
```
dougbu: s/$(CI)/'$(CI)'/

jkotalik:
This is working. The check I added was faulty on other OSs.

build/repo.props
```diff
@@ -27,13 +34,6 @@
...
 
     <SharedSourcesFolder>$(RepoRoot)src\Shared\</SharedSourcesFolder>
-  </PropertyGroup>
-
-  <PropertyGroup Condition=" '$(BuildAllProjects)' == 'true' ">
-    <BuildNative>true</BuildNative>
-    <BuildManaged>true</BuildManaged>
-    <BuildNodeJS>true</BuildNodeJS>
-    <BuildJava>true</BuildJava>
     <BuildIisNativeProjects Condition="'$(TargetOsName)' == 'win' AND ('$(TargetArchitecture)' == 'x86' OR '$(TargetArchitecture)' == 'x64')">true</BuildIisNativeProjects>
```
jkotalik: This diff looks odd. The fix is to check BuildIisNativeProjects always rather than just when BuildAllProjects is true.

jkotalik:
@aspnet/build _bump_

BrennanConroy (approved):

jkotalik:
If this last commit doesn't work, I'll just revert the previous commit that broke this.

PR #10590, opened 2019-05-28
Dylan/queue limit

DylanDmitri:
- middleware takes a "queue limit"
 - any requests over the limit are rejected

### DIFF
--- a/src/Middleware/RequestThrottling/src/Internal/RequestQueue.cs
+++ b/src/Middleware/RequestThrottling/src/Internal/RequestQueue.cs
@@ -11,12 +11,15 @@ namespace Microsoft.AspNetCore.RequestThrottling.Internal
     {
         private SemaphoreSlim _semaphore;
         private object _waitingRequestsLock = new object();
+
         public readonly int MaxConcurrentRequests;
+        public readonly int RequestQueueLimit;
         public int WaitingRequests { get; private set; }
 
-        public RequestQueue(int maxConcurrentRequests)
+        public RequestQueue(int maxConcurrentRequests, int requestQueueLimit)
         {
             MaxConcurrentRequests = maxConcurrentRequests;
+            RequestQueueLimit = requestQueueLimit;
             _semaphore = new SemaphoreSlim(maxConcurrentRequests);
         }
 
@@ -37,6 +40,17 @@ namespace Microsoft.AspNetCore.RequestThrottling.Internal
                 lock (_waitingRequestsLock)
                 {
                     WaitingRequests--;
+                }
+            }
+        }
+
+        public bool QueueLengthExceeded
+        {
+            get
+            {
+                lock (_waitingRequestsLock)
+                {
+                    return WaitingRequests > RequestQueueLimit;
                 }
             }
         }
--- a/src/Middleware/RequestThrottling/src/Internal/RequestQueue.cs
+++ b/src/Middleware/RequestThrottling/src/Internal/RequestQueue.cs
@@ -10,49 +10,37 @@ namespace Microsoft.AspNetCore.RequestThrottling.Internal
     internal class RequestQueue : IDisposable
     {
         private SemaphoreSlim _semaphore;
-        private object _waitingRequestsLock = new object();
+        private int _waitingRequests;
 
-        public readonly int MaxConcurrentRequests;
-        public readonly int RequestQueueLimit;
-        public int WaitingRequests { get; private set; }
+        private readonly int _maxConcurrentRequests;
+        private readonly int _requestQueueLimit;
 
         public RequestQueue(int maxConcurrentRequests, int requestQueueLimit)
         {
-            MaxConcurrentRequests = maxConcurrentRequests;
-            RequestQueueLimit = requestQueueLimit;
+            _maxConcurrentRequests = maxConcurrentRequests;
+            _requestQueueLimit = requestQueueLimit;
             _semaphore = new SemaphoreSlim(maxConcurrentRequests);
         }
 
-        public async Task EnterQueue()
+        public RequestQueue(int maxConcurrentRequests) : this(maxConcurrentRequests, 5000) { }
+
+        public async Task<bool> TryEnterQueueAsync()
         {
+            // first, check the queue limits
+            if (WaitingRequests > _requestQueueLimit)
+            {
+                return false;
+            }
+
             var waitInQueueTask = _semaphore.WaitAsync();
-
-            var needsToWaitOnQueue = !waitInQueueTask.IsCompletedSuccessfully;
-            if (needsToWaitOnQueue)
+            if (!waitInQueueTask.IsCompletedSuccessfully)
             {
-                lock (_waitingRequestsLock)
-                {
-                    WaitingRequests++;
-                }
-
+                Interlocked.Increment(ref _waitingRequests);
                 await waitInQueueTask;
-
-                lock (_waitingRequestsLock)
-                {
-                    WaitingRequests--;
-                }
+                Interlocked.Decrement(ref _waitingRequests);
             }
-        }
 
-        public bool QueueLengthExceeded
-        {
-            get
-            {
-                lock (_waitingRequestsLock)
-                {
-                    return WaitingRequests > RequestQueueLimit;
-                }
-            }
+            return true;
         }
 
         public void Release()
@@ -67,7 +55,12 @@ namespace Microsoft.AspNetCore.RequestThrottling.Internal
 
         public int ConcurrentRequests
         {
-            get => MaxConcurrentRequests - _semaphore.CurrentCount;
+            get => _maxConcurrentRequests - _semaphore.CurrentCount;
+        }
+
+        public int WaitingRequests
+        {
+            get => _waitingRequests;
         }
 
         public void Dispose()
--- a/src/Middleware/RequestThrottling/src/Internal/RequestQueue.cs
+++ b/src/Middleware/RequestThrottling/src/Internal/RequestQueue.cs
@@ -11,7 +11,8 @@ namespace Microsoft.AspNetCore.RequestThrottling.Internal
     {
         private readonly int _maxConcurrentRequests;
         private readonly int _requestQueueLimit;
-        private readonly SemaphoreSlim _semaphore;
+        private readonly SemaphoreSlim _serverSemaphore;
+        //private readonly SemaphoreSlim _queueSemaphore;
 
         private object _waitingRequestsLock = new object();
         private int _waitingRequests;
@@ -20,14 +21,17 @@ namespace Microsoft.AspNetCore.RequestThrottling.Internal
         {
             _maxConcurrentRequests = maxConcurrentRequests;
             _requestQueueLimit = requestQueueLimit;
-            _semaphore = new SemaphoreSlim(maxConcurrentRequests);
+            //_queueSemaphore = new SemaphoreSlim(_requestQueueLimit);
+            _serverSemaphore = new SemaphoreSlim(_maxConcurrentRequests);
         }
+
+        // check total server capacity; only reject if it's all full
 
         public async Task<bool> TryEnterQueueAsync()
         {
             lock (_waitingRequestsLock)
             {
-                if (_waitingRequests >= _requestQueueLimit)
+                if (_waitingRequests + ConcurrentRequests >= _requestQueueLimit + _maxConcurrentRequests)
                 {
                     return false;
                 }
@@ -35,7 +39,7 @@ namespace Microsoft.AspNetCore.RequestThrottling.Internal
                 _waitingRequests++;
             }
 
-            await _semaphore.WaitAsync();
+            await _serverSemaphore.WaitAsync();
 
             lock (_waitingRequestsLock)
             {
@@ -44,30 +48,47 @@ namespace Microsoft.AspNetCore.RequestThrottling.Internal
 
             return true;
         }
+        //{
+        // a return value of 'false' indicates that the request is rejected
+        // a return value of 'true' indicates that the request may proceed
+        // _serverSemaphore.Release is *not* called in this method, it is called externally when requests leave the server
+
+        // must check the queue before checking the concurrency limit, can't let requests skip the 
+
+        //var enterQueueTask = _queueSemaphore.WaitAsync();
+        //if (!enterQueueTask.IsCompletedSuccessfully)
+        //{
+        //    return false;
+        //}
+
+        //await _serverSemaphore.WaitAsync();
+        //_queueSemaphore.Release();
+        //return true;
+        //}
+
+        public bool TryEnterServerWithNoQueue()
+        {
+            return _serverSemaphore.WaitAsync().IsCompletedSuccessfully;
+        }
 
         public void Release()
         {
-            _semaphore.Release();
-        }
-
-        public int Count
-        {
-            get => _semaphore.CurrentCount;
+            _serverSemaphore.Release();
         }
 
         public int ConcurrentRequests
         {
-            get => _maxConcurrentRequests - _semaphore.CurrentCount;
+            get => _maxConcurrentRequests - _serverSemaphore.CurrentCount;
         }
 
         public int WaitingRequests
         {
-            get => _waitingRequests;
+            get => _waitingRequests; // _requestQueueLimit - _queueSemaphore.CurrentCount;
         }
 
         public void Dispose()
         {
-            _semaphore.Dispose();
+            _serverSemaphore.Dispose();
         }
     }
 }
--- a/src/Middleware/RequestThrottling/src/Internal/RequestQueue.cs
+++ b/src/Middleware/RequestThrottling/src/Internal/RequestQueue.cs
@@ -9,11 +9,12 @@ namespace Microsoft.AspNetCore.RequestThrottling.Internal
 {
     internal class RequestQueue : IDisposable
     {
-        private SemaphoreSlim _semaphore;
-        private int _waitingRequests;
-
         private readonly int _maxConcurrentRequests;
         private readonly int _requestQueueLimit;
+        private readonly SemaphoreSlim _semaphore;
+
+        private object _waitingRequestsLock = new object();
+        private int _waitingRequests;
 
         public RequestQueue(int maxConcurrentRequests, int requestQueueLimit)
         {
@@ -22,22 +23,23 @@ namespace Microsoft.AspNetCore.RequestThrottling.Internal
             _semaphore = new SemaphoreSlim(maxConcurrentRequests);
         }
 
-        public RequestQueue(int maxConcurrentRequests) : this(maxConcurrentRequests, 5000) { }
-
         public async Task<bool> TryEnterQueueAsync()
         {
-            // first, check the queue limits
-            if (WaitingRequests > _requestQueueLimit)
+            lock (_waitingRequestsLock)
             {
-                return false;
+                if (_waitingRequests >= _requestQueueLimit)
+                {
+                    return false;
+                }
+
+                _waitingRequests++;
             }
 
-            var waitInQueueTask = _semaphore.WaitAsync();
-            if (!waitInQueueTask.IsCompletedSuccessfully)
+            await _semaphore.WaitAsync();
+
+            lock (_waitingRequestsLock)
             {
-                Interlocked.Increment(ref _waitingRequests);
-                await waitInQueueTask;
-                Interlocked.Decrement(ref _waitingRequests);
+                _waitingRequests--;
             }
 
             return true;
--- a/src/Middleware/RequestThrottling/src/Internal/RequestQueue.cs
+++ b/src/Middleware/RequestThrottling/src/Internal/RequestQueue.cs
@@ -46,12 +46,12 @@ namespace Microsoft.AspNetCore.RequestThrottling.Internal
 
         public void Release()
         {
+            _serverSemaphore.Release();
+
             lock (_totalRequestsLock)
             {
                 TotalRequests--;
             }
-
-            _serverSemaphore.Release();
         }
 
         public void Dispose()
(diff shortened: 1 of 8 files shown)


src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
```diff
@@ -46,8 +46,14 @@ public RequestThrottlingMiddleware(RequestDelegate next, ILoggerFactory loggerFa
         /// <returns>A <see cref="Task"/> that completes when the request leaves.</returns>
         public async Task Invoke(HttpContext context)
         {
-            var waitInQueueTask = _requestQueue.EnterQueue();
+            if (_requestQueue.QueueLengthExceeded)
+            {
+                context.Response.StatusCode = 503;
+                context.Abort();
```
Tratcher: Abort not required

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
```diff
@@ -46,8 +46,14 @@ public RequestThrottlingMiddleware(RequestDelegate next, ILoggerFactory loggerFa
...
+                return;
+            }
 
+            var waitInQueueTask = _requestQueue.EnterQueue();
```
Tratcher: `Task<bool> TryEnterQueueAsync`

src/Middleware/RequestThrottling/src/RequestThrottlingOptions.cs
```diff
@@ -15,5 +15,11 @@ public class RequestThrottlingOptions
         /// This is null by default because the correct value is application specific. This option must be configured by the application.
         /// </summary>
         public int? MaxConcurrentRequests { get; set; }
+
+        /// <summary>
+        /// Maximum number of queued requests before the server starts rejecting connections.
+        /// The server will respond with a 503 if this limit is exceeeded.
+        /// </summary>
+        public int RequestQueueLimit { get; set; } = 5000;
```
davidfowl: 5000 * number of CPUs

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
DAllanCarr: The 503 seems a bit dramatic.

DAllanCarr: The 503 seems a bit dramatic.

src/Middleware/RequestThrottling/src/RequestThrottlingOptions.cs
DylanDmitri: The queue limit was fixed at 5000 the old one, so I'm just copying that. The number of concurrently running requests was 5000*CPUs

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
```diff
@@ -46,8 +46,14 @@ public RequestThrottlingMiddleware(RequestDelegate next, ILoggerFactory loggerFa
         /// <returns>A <see cref="Task"/> that completes when the request leaves.</returns>
         public async Task Invoke(HttpContext context)
         {
-            var waitInQueueTask = _requestQueue.EnterQueue();
+            if (_requestQueue.QueueLengthExceeded)
+            {
+                context.Response.StatusCode = 503;
```
Tratcher: StatusCodes.Status503ServiceUnavailable
https://github.com/aspnet/AspNetCore/blob/3eca32965d699e08692d074b449d1148bb626703/src/Http/Http.Abstractions/src/StatusCodes.cs#L70

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
Tratcher: @DAllanCarr how so? What else would we respond with?

DAllanCarr: This is not rate limiting I assume. I was thinking 503 maybe misinterpreted.

Tratcher: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status
"The server is not ready to handle the request. Common causes are a server that is down for maintenance or that is overloaded."

src/Middleware/RequestThrottling/src/Internal/RequestQueue.cs
```diff
@@ -11,12 +11,15 @@ internal class RequestQueue : IDisposable
     {
         private SemaphoreSlim _semaphore;
         private object _waitingRequestsLock = new object();
+
         public readonly int MaxConcurrentRequests;
+        public readonly int RequestQueueLimit;
```
jkotalik:
```suggestion
        private readonly int _requestQueueLimit;
```

jkotalik: And update MaxConcurrentRequests to private too.

src/Middleware/RequestThrottling/src/RequestThrottlingOptions.cs
jkotalik: Add a comment in a remarks sections below summary mentioning that the limit was copied from IIS. Maybe include https://blogs.msdn.microsoft.com/benjaminperkins/2011/11/14/modify-the-request-queue-limit-requestqueuelimit-or-queue-length-in-iis-7/ or a similar doc as a reference.

src/Middleware/RequestThrottling/test/MiddlewareTests.cs
```diff
@@ -82,5 +82,20 @@ public void InvalidArgumentIfMaxConcurrentRequestsIsNull()
             });
             Assert.Equal("options", ex.ParamName);
         }
+
+        [Fact]
+        public async void RequestsBlockedIfQueueFull()
```
jkotalik: I think we need to add a few more tests here. Specifically, one that has multiple requests and verifying the status code of 503.

src/Middleware/RequestThrottling/test/RequestQueueTests.cs
```diff
@@ -12,7 +12,7 @@ public class RequestQueueTests
         [Fact]
         public async Task LimitsIncomingRequests()
         {
-            using var s = new RequestQueue(1);
+            using var s = new RequestQueue(maxConcurrentRequests: 1, requestQueueLimit: 999);
```
jkotalik: Let's add a second constructor that only has a parameter for maxConcurrentRequests and default requestQueueLimit to 5000. Then have some pretty constructor chaining.

src/Middleware/RequestThrottling/test/TestUtils.cs
```diff
@@ -6,16 +6,18 @@
...
 using Microsoft.Extensions.Logging.Abstractions;
 using Microsoft.Extensions.Options;
+using Microsoft.AspNetCore.RequestThrottling.Internal;
 
 namespace Microsoft.AspNetCore.RequestThrottling.Tests
 {
     public static class TestUtils
     {
-        public static RequestThrottlingMiddleware CreateTestMiddleWare(int? maxConcurrentRequests, RequestDelegate next = null)
+        public static RequestThrottlingMiddleware CreateTestMiddleWare(int? maxConcurrentRequests, int? requestQueueLimit = null, RequestDelegate next = null)
```
jkotalik:
```suggestion
        public static RequestThrottlingMiddleware CreateTestMiddleWare(int? maxConcurrentRequests, int requestQueueLimit = 999, RequestDelegate next = null)
```

src/Middleware/RequestThrottling/test/RequestQueueTests.cs
Tratcher: Meh, then you're defining the default in multiple places.

src/Middleware/RequestThrottling/src/RequestThrottlingOptions.cs
Tratcher: > Add a comment in a remarks sections below summary mentioning that the limit was copied from IIS. Maybe include https://blogs.msdn.microsoft.com/benjaminperkins/2011/11/14/modify-the-request-queue-limit-requestqueuelimit-or-queue-length-in-iis-7/ or a similar doc as a reference.

Not needed. That's interesting to us but nobody else. A code comment would be sufficient.

src/Middleware/RequestThrottling/test/RequestQueueTests.cs
jkotalik: Maybe we can make the default an internal static int within options?

src/Middleware/RequestThrottling/src/RequestThrottlingOptions.cs
jkotalik: Sure. @DylanDmitri here is an example: https://github.com/aspnet/AspNetCore/blob/master/src/Servers/Kestrel/Core/src/KestrelServerLimits.cs#L27

src/Middleware/RequestThrottling/src/Internal/RequestQueue.cs
```diff
@@ -10,35 +10,37 @@ namespace Microsoft.AspNetCore.RequestThrottling.Internal
...
-        {
-            var waitInQueueTask = _semaphore.WaitAsync();
+        public RequestQueue(int maxConcurrentRequests) : this(maxConcurrentRequests, 5000) { }
 
-            var needsToWaitOnQueue = !waitInQueueTask.IsCompletedSuccessfully;
-            if (needsToWaitOnQueue)
+        public async Task<bool> TryEnterQueueAsync()
+        {
+            // first, check the queue limits
+            if (WaitingRequests > _requestQueueLimit)
```
halter73: Do the increment/decrement thing we talked about in person.

src/Middleware/RequestThrottling/src/Internal/RequestQueue.cs
```diff
@@ -10,35 +10,37 @@ namespace Microsoft.AspNetCore.RequestThrottling.Internal
...
     {
         private SemaphoreSlim _semaphore;
-        private object _waitingRequestsLock = new object();
-        public readonly int MaxConcurrentRequests;
-        public int WaitingRequests { get; private set; }
+        private int _waitingRequests;
 
-        public RequestQueue(int maxConcurrentRequests)
+        private readonly int _maxConcurrentRequests;
+        private readonly int _requestQueueLimit;
```
halter73: nit: readonly fields generally come before mutable fields

src/Middleware/RequestThrottling/src/Internal/RequestQueue.cs
```diff
@@ -10,35 +10,37 @@ namespace Microsoft.AspNetCore.RequestThrottling.Internal
     internal class RequestQueue : IDisposable
     {
         private SemaphoreSlim _semaphore;
```
halter73: Not new, but can't this be readonly?

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
```diff
@@ -46,11 +47,21 @@ public RequestThrottlingMiddleware(RequestDelegate next, ILoggerFactory loggerFa
...
-                RequestThrottlingLog.RequestRunImmediately(_logger);
+                var requestPassedThrough = await waitInQueueTask;
+                if (requestPassedThrough)
+                {
+                    RequestThrottlingLog.RequestRunImmediately(_logger);
+                }
+                else
+                {
+                    // the queue is full
+                    context.Response.StatusCode = StatusCodes.Status503ServiceUnavailable;
```
halter73: This should definitely be logged.

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
```diff
@@ -46,11 +47,21 @@ public RequestThrottlingMiddleware(RequestDelegate next, ILoggerFactory loggerFa
         /// <returns>A <see cref="Task"/> that completes when the request leaves.</returns>
         public async Task Invoke(HttpContext context)
         {
-            var waitInQueueTask = _requestQueue.EnterQueue();
+            var waitInQueueTask = _requestQueue.TryEnterQueueAsync();
 
             if (waitInQueueTask.IsCompletedSuccessfully)
             {
-                RequestThrottlingLog.RequestRunImmediately(_logger);
+                var requestPassedThrough = await waitInQueueTask;
```
halter73: I know this can currently only be false in the IsCompletedSuccessfully case, but it is kinda weird to not check whether TryEnterQueueAsync's return value is true or false in the async changes. What if the implementation changes?

src/Middleware/RequestThrottling/test/MiddlewareTests.cs
halter73: Yeah. While not strictly necessary for testing this, it's probably also a good idea to start TestServer-based tests.

src/Middleware/RequestThrottling/test/RequestQueueTests.cs
```diff
@@ -15,7 +15,7 @@ public async Task LimitsIncomingRequests()
             using var s = new RequestQueue(1);
             Assert.Equal(1, s.Count);
 
-            await s.EnterQueue().OrTimeout();
+            await s.TryEnterQueueAsync().OrTimeout();
```
halter73: Assert the return value is true for all the current calls to TryEnterQueueAsync.

src/Middleware/RequestThrottling/src/Internal/RequestQueue.cs
```diff
@@ -10,35 +10,37 @@ namespace Microsoft.AspNetCore.RequestThrottling.Internal
...
-            MaxConcurrentRequests = maxConcurrentRequests;
+            _maxConcurrentRequests = maxConcurrentRequests;
+            _requestQueueLimit = requestQueueLimit;
             _semaphore = new SemaphoreSlim(maxConcurrentRequests);
         }
 
-        public async Task EnterQueue()
-        {
-            var waitInQueueTask = _semaphore.WaitAsync();
+        public RequestQueue(int maxConcurrentRequests) : this(maxConcurrentRequests, 5000) { }
```
Tratcher: Is this constructor only for test convenience? If so, replace it with a helper method in the test class. We don't want defaults hard-coded in places like this, or test helpers in production code.

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
```diff
@@ -7,6 +7,7 @@
 using Microsoft.AspNetCore.RequestThrottling.Internal;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
+using Microsoft.AspNetCore.Http.Abstractions;
```
Tratcher: Sort

src/Middleware/RequestThrottling/src/Internal/RequestQueue.cs
Tratcher: This is still racy. We need to re-think how to check this.

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
davidfowl: This should be a 429 or something

jkotalik: ASP.NET returned 503.2. I don't think 429 is correct because it isn't an individual client sending too many requests.

Tratcher: Agreed, a 429 would unfairly blame the client. 503 is right for a server issue.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status
429 Too Many Requests
The user has sent too many requests in a given amount of time ("rate limiting").
503 Service Unavailable
The server is not ready to handle the request. Common causes are a server that is down for maintenance or that is overloaded.

halter73: Yeah. @Tratcher and I were discussing this earlier. 503 is definitely the right status code, but eventually we can add an option to specify a custom request queue full RequestDelegate that could set a Retry-After header, change the status code to something else or whatever.

Tratcher: Yeah, `Task OnRejected(HttpContext)` where you can provide your own response.

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
anurse: @DAllanCarr no, this isn't rate limiting, though it's something we have on the radar for this project eventually. Right now the scope is just about having a single queue for all requests coming in to the server to help manage resources better in certain environments.

src/Middleware/RequestThrottling/src/Internal/RequestQueue.cs
DylanDmitri: I've gone back to boring locks

DAllanCarr: I prefer this. Go back to lock free if it's necessary.

src/Middleware/RequestThrottling/src/Internal/RequestQueue.cs
```diff
@@ -9,36 +9,40 @@ namespace Microsoft.AspNetCore.RequestThrottling.Internal
...
+        public async Task<bool> TryEnterQueueAsync()
         {
-            var waitInQueueTask = _semaphore.WaitAsync();
-
-            var needsToWaitOnQueue = !waitInQueueTask.IsCompletedSuccessfully;
-            if (needsToWaitOnQueue)
+            lock (_waitingRequestsLock)
             {
-                lock (_waitingRequestsLock)
+                if (_waitingRequests >= _requestQueueLimit)
```
Tratcher: So a _requestQueueLimit of 0 still breaks this, right? _waitingRequests will be 0 and all requests will be rejected even if the _semaphore has space.

That's fine, but we'd want to put argument validation on _requestQueueLimit.

Tratcher (approved):

src/Middleware/RequestThrottling/src/Internal/RequestQueue.cs
jkotalik: @DylanDmitri and I discussed this yesterday. The fix will be to check both _waitingRequests and concurrentRequests to both limits. Specifically: `_waitingRequests + ConcurrentRequests >= _requestQueueLimit + _maxConcurrentRequests`

Tratcher: That would help. Except ConcurrentRequests isn't controlled under the lock, it's controlled by the _semaphore so your results are going to be inconsistent. e.g. I can get 10k requests past this lock if they pause before acquiring the semaphore. It's unlikely to get that off balance, but it will certainly drift.

I expect this will need to abandon the semaphore soon, it's already outgrowing it.

jkotalik: > it's controlled by the _semaphore so your results are going to be inconsistent.

Bleh you are right. I didn't realize we used _semaphore.CurrentCount rather than counting outside of the semaphore.

halter73: We could manage our own count independently of the semaphore for now. Our counter would increment before and decrement after the semaphore's internal counter does, that way you'd still never let through more than `_requestQueueLimit + _maxConcurrentRequests` through at a time.

src/Middleware/RequestThrottling/src/Internal/RequestQueue.cs
```diff
@@ -9,36 +9,40 @@ namespace Microsoft.AspNetCore.RequestThrottling.Internal
...
                 {
-                    WaitingRequests++;
+                    return false;
                 }
 
-                await waitInQueueTask;
+                _waitingRequests++;
+            }
+
+            await _semaphore.WaitAsync();
```
halter73: One optimization I'd be fine with you doing now (or saving in your back pocket for when you start benchmarking) is to store the Task returned by WaitAsync. If it's already complete, you can avoid taking the _waitingRequestsLock once let alone twice, since you wouldn't need to increment or decrement _waitingRequests.

src/Middleware/RequestThrottling/src/Internal/RequestQueue.cs
DylanDmitri: I'll try three different implementations and then see what feels best:
- no semaphores just locks and counts
- two semaphores, one for queue one for server
- something involving a concurrent queue

DylanDmitri: wow the first one is bad defining a queue yourself is not the way to go

DylanDmitri: I am feeling tied to the log messages I already defined; in order to log the proper stuff in the proper places there's considerable gymnastics. Will revise which types of messages exist in these 'experiment' branches.

halter73: > I am feeling tied to the log messages I already defined; in order to log the proper stuff in the proper places there's considerable gymnastics.

Consider that it's not just logging. Going forward, we'll probably want to increment/decrement performance counters in the same places, potentially call user-defined callbacks, etc...

Tratcher: Making ConcurrentRequests a local field rather than checking _semaphore.CurrentCount makes sense. Just make sure you add another lock in Release when derecementing it.

DylanDmitri:
The code enforces that: (# queued requests) + (# active requests) < threshold

The number of queued requests aren't consistent, and can drift a bit past the queue limit. I decided to combine both metrics into a single `TotalRequests` field. This ensures correctness in diagnostics and reporting and reduces the number of locks/unlocks.

Tratcher:
Clever, :shipit:

PR #10598, opened 2019-05-28
HeaderPropagation middleware: small cleanup

alefranz:
- Removed `MessageHandler` DI registration as not needed.
- Fixed two parameters XML descriptions.
- Cleaned up `using`s in the example.

/cc @rynowak


rynowak (approved):
Thanks!

PR #10600, opened 2019-05-29
React to breaking change in dotnet/arcade#2753

natemcmaster:
cc @davidfowl

### DIFF
--- a/eng/targets/ReferenceAssembly.targets
+++ b/eng/targets/ReferenceAssembly.targets
@@ -55,7 +55,7 @@
     </ItemGroup>
 
     <PropertyGroup>
-      <_GenAPICommand Condition="'$(MSBuildRuntimeType)' == 'core'">$(ToolHostCmd) --roll-forward-on-no-candidate-fx 2 "$(_GenAPIPath)"</_GenAPICommand>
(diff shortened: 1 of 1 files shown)


eng/targets/ReferenceAssembly.targets
```diff
@@ -55,7 +55,7 @@
     </ItemGroup>
 
     <PropertyGroup>
-      <_GenAPICommand Condition="'$(MSBuildRuntimeType)' == 'core'">$(ToolHostCmd) --roll-forward-on-no-candidate-fx 2 "$(_GenAPIPath)"</_GenAPICommand>
+      <_GenAPICommand Condition="'$(MSBuildRuntimeType)' == 'core'">"$(DotNetTool)" --roll-forward-on-no-candidate-fx 2 "$(_GenAPIPath)"</_GenAPICommand>
```
dougbu: Have we already taken an Arcade dependency update with this change in it? If so, how did validation succeed for that PR?

dougbu (approved):

eng/targets/ReferenceAssembly.targets
natemcmaster: PR validation runs with `'$(MSBuildRuntimeType)' == 'Full'`

PR #10610, opened 2019-05-29
Fix linux CI helix break

benaadams:
Test comparison with newline doesn't match the platform newlines

/cc @rynowak


BrennanConroy (approved):
:+1:

rynowak (approved):
Thanks YO 👍

PR #10613, opened 2019-05-29
Make new HttpContext properties virtual

### LINKED ISSUE #10594: Make pipe-related properties on Http abstractions non-abstract
We shouldn't add `abstract` members to `HttpRequest`/`HttpResponse`/etc. unless we really really need it. It breaks creators of derived classes.

We should change the following `abstract` members to `virtual` and have them throw `NotImplementedException`:
* `HttpRequest.BodyReader`
* `HttpResponse.BodyWriter`
* `HttpResponse.StartAsync`

jkotalik:
Fixes https://github.com/aspnet/AspNetCore/issues/10594


Tratcher:
Ref assemblies...

Tratcher (approved):

PR #10618, opened 2019-05-29
Clear ShutdownTestRun flaky state

Tratcher:
https://github.com/aspnet/AspNetCore-Internal/issues/1687

The original issue was fixed. It's not clear why there was flakyness for several weeks after, but there hasn't been any in 2 months now.


jkotalik:
Are we concerned about this failure? https://github.com/aspnet/AspNetCore-Internal/issues/2577.

jkotalik (approved):

Tratcher:
@jkotalik yes, but...
- it doesn't look related to any of the prior failures (or at least the ones that weren't deleted), 
- it's only happened once so far
- it's an AV so we want that to call attention to it if it happens again, nothing in this test should be able to cause an AV.

PR #10632, opened 2019-05-29
Runtime integration for static web assets

javiercn:
* Adds an extension method to IWebHostBuilder to plug-in static web
  assets.
* Plugs-in static web assets in development by default.


src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsWebHostBuilderExtensions.cs
```diff
@@ -0,0 +1,30 @@
...
+    {
+        /// <summary>
+        /// Configures the <see cref="IWebHostEnvironment.WebRootFileProvider"/> to use static web assets
+        /// defined by referenced projects and packages.
+        /// </summary>
+        /// <param name="builder">The <see cref="IWebHostBuilder"/>.</param>
+        /// <returns>The <see cref="IWebHostBuilder"/>.</returns>
+        public static IWebHostBuilder UseStaticWebAssets(this IWebHostBuilder builder)
+        {
+            builder.ConfigureAppConfiguration((context, configBuilder) =>
```
javiercn: We do it in ConfigureAppConfiguration to avoid adding a new lifecycle method in hosting. This happens early enough in the configuration process that makes it really hard to observe the wrong WebRootFileProvider

src/DefaultBuilder/src/WebHost.cs
```diff
@@ -206,6 +206,13 @@ public static IWebHostBuilder CreateDefaultBuilder(string[] args)
 
         internal static void ConfigureWebDefaults(IWebHostBuilder builder)
         {
+            builder.ConfigureAppConfiguration((ctx,cb) =>
+            {
+                if (ctx.HostingEnvironment.IsDevelopment())
+                {
+                    ctx.HostingEnvironment.UseStaticWebAssets();
```
javiercn: We automatically plug-in the static web assets during development, similar to user secrets

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsFileProvider.cs
```diff
@@ -0,0 +1,69 @@
...
+using System.Runtime.InteropServices;
+using Microsoft.Extensions.FileProviders;
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNetCore
+{
+    /// <summary>
+    /// A <see cref="IFileProvider"/> for serving static web assets during development.
+    /// </summary>
+    internal class StaticWebAssetsFileProvider : IFileProvider
```
javiercn: All these classes will go into a separate package. Adding them to the default builder for now as we want to avoid delaying the feature from preview6 and all these are an implementation detail

src/DefaultBuilder/src/Microsoft.AspNetCore.csproj
```diff
@@ -21,6 +21,7 @@
     <Reference Include="Microsoft.Extensions.Configuration.Json" PrivateAssets="None" />
     <Reference Include="Microsoft.Extensions.Configuration.CommandLine" PrivateAssets="None" />
     <Reference Include="Microsoft.Extensions.Configuration.UserSecrets" PrivateAssets="None" />
+    <Reference Include="Microsoft.Extensions.FileProviders.Composite" PrivateAssets="None" />
```
benaadams: Need to update ref assemblies also
```
##[error]src\DefaultBuilder\ref\Microsoft.AspNetCore.csproj(,): error :
 Generated code is not up to date in src/DefaultBuilder/ref/Microsoft.AspNetCore.csproj. 

You might need to regenerate the reference assemblies or project list (see docs/ReferenceAssemblies.md and docs/ReferenceResolution.md)
error : Generated code is not up to date in src/DefaultBuilder/ref/Microsoft.AspNetCore.csproj. You might need to regenerate the reference assemblies or project list (see docs/ReferenceAssemblies.md and docs/ReferenceResolution.md) [F:\workspace\_work\1\s\src\DefaultBuilder\ref\Microsoft.AspNetCore.csproj]
```

```diff --git a/src/DefaultBuilder/ref/Microsoft.AspNetCore.csproj b/src/DefaultBuilder/ref/Microsoft.AspNetCore.csproj
index 12795e2033..dd42476a01 100644
--- a/src/DefaultBuilder/ref/Microsoft.AspNetCore.csproj
+++ b/src/DefaultBuilder/ref/Microsoft.AspNetCore.csproj
@@ -17,6 +17,7 @@
     <Reference Include="Microsoft.Extensions.Configuration.Json"  />
     <Reference Include="Microsoft.Extensions.Configuration.CommandLine"  />
     <Reference Include="Microsoft.Extensions.Configuration.UserSecrets"  />
+    <Reference Include="Microsoft.Extensions.FileProviders.Composite"  />
     <Reference Include="Microsoft.Extensions.Logging"  />
     <Reference Include="Microsoft.Extensions.Logging.Configuration"  />
     <Reference Include="Microsoft.Extensions.Logging.Console"  />
```

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsFileProvider.cs
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        public PhysicalFileProvider InnerProvider { get; }
+
+        public string BasePath { get; }
+
+        /// <inheritdoc />
+        public IDirectoryContents GetDirectoryContents(string subpath)
+        {
+            if (!subpath.StartsWith(BasePath, FileSystemBasePathComparisonMode))
```
Tratcher: This should only match on whole segments.
"/foobar" starts with "/foo" but shouldn't match.

anurse:
Can someone catch me up on the quick ~2 sentence reason for this new integration? Are static web assets part of single-file deployment or something else?

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsFileProvider.cs
rynowak: Why does this justify a separate package?

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsFileProvider.cs
```diff
@@ -0,0 +1,69 @@
...
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.AspNetCore
+{
+    /// <summary>
+    /// A <see cref="IFileProvider"/> for serving static web assets during development.
+    /// </summary>
+    internal class StaticWebAssetsFileProvider : IFileProvider
+    {
+        private static readonly StringComparison FileSystemBasePathComparisonMode = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ?
```
rynowak: `FilePathComparison`?

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsFileProvider.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// Initializes a new instance of <see cref="StaticWebAssetsFileProvider"/>.
+        /// </summary>
+        /// <param name="pathPrefix">The path prefix under which the files in the <paramref name="contentRoot"/> folder will
+        /// be mapped.</param>
+        /// <param name="contentRoot">The absolute path to the content root associated with the static web assets.</param>
+        public StaticWebAssetsFileProvider(string pathPrefix, string contentRoot)
+        {
+            BasePath = pathPrefix.StartsWith("/") ? pathPrefix : "/" + pathPrefix;
+            InnerProvider = new PhysicalFileProvider(contentRoot);
+        }
```
rynowak: I think this deserves some comments about why we need to write a new file provider. We could also consider adding support for *base paths* to `PhysicalFileProvider` which would result in this being more efficient.

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsHelper.cs
```diff
@@ -0,0 +1,137 @@
...
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Xml.Linq;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.FileProviders;
+
+namespace Microsoft.AspNetCore
+{
+    internal static class StaticWebAssetsHelper
```
rynowak: gross. Can we think of better ways to organise and name this functionality?

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsHelper.cs
```diff
@@ -0,0 +1,137 @@
...
+    {
+        private const string StaticWebAssetsManifestName = "Microsoft.AspNetCore.StaticWebAssets.xml";
+        private const string ManifestRootElementName = "StaticWebAssets";
+        private const string VersionAttributeName = "Version";
+        private const string ContentRootElementName = "ContentRoot";
+
+        internal static void UseStaticWebAssets(this IWebHostEnvironment environment)
```
rynowak: re: naming. I would **not** expected to find extension methods on a top-level type like this in this randomly named class.

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsHelper.cs
```diff
@@ -0,0 +1,137 @@
...
+                environment.WebRootFileProvider = new CompositeFileProvider(staticWebAssetContentRoots);
+            }
+        }
+
+        internal static Stream ResolveManifest(IWebHostEnvironment environment)
+        {
+            var assembly = Assembly.Load(environment.ApplicationName);
+            if (assembly.GetManifestResourceNames().Any(a => a == StaticWebAssetsManifestName))
+            {
+                return assembly.GetManifestResourceStream(StaticWebAssetsManifestName);
```
rynowak: Do we need this check? I think `GetManifestResourceStream` would just do the right thing.

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsHelper.cs
```diff
@@ -0,0 +1,137 @@
...
+                }
+            }
+        }
+
+        internal static void UseStaticWebAssetsCore(this IWebHostEnvironment environment, Stream manifest)
+        {
+            var staticWebAssetsFileProvider = new List<IFileProvider>();
+            var webRootFileProvider = environment.WebRootFileProvider;
+
+            var staticWebAssetContentRoots = Parse(manifest)
```
rynowak: This is wierd because it's a list of file-providers not a list of roots.

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsHelper.cs
```diff
@@ -0,0 +1,137 @@
...
+        {
+            var assembly = Assembly.Load(environment.ApplicationName);
+            if (assembly.GetManifestResourceNames().Any(a => a == StaticWebAssetsManifestName))
+            {
+                return assembly.GetManifestResourceStream(StaticWebAssetsManifestName);
+            }
+            else
+            {
+                // Fallback to physical file as we plan to use a file on disk instead of the embedded resource.
+                var filePath = Path.Combine(Path.GetDirectoryName(new Uri(assembly.CodeBase).LocalPath), $"{environment.ApplicationName}.StaticWebAssets.xml");
```
rynowak: Can we resolve what we're doing here before checking this in? I'm really not sold on the consequences of making this an embedded resource. 

I would also expect to see `.Location` here instead of codebase and URI if this is needed can you elaborate?

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsHelper.cs
rynowak: suggestion: `StaticWebAssetsReader` and move the file-provider logic elsewhere. I also don't think there's really a justifcation for these to be extension methods.

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsHelper.cs
```diff
@@ -0,0 +1,137 @@
...
+                throw new InvalidOperationException($"Invalid manifest format. Manifest root must be '{ManifestRootElementName}'");
+            }
+
+            var version = document.Root.Attribute(VersionAttributeName);
+            if (version == null)
+            {
+                throw new InvalidOperationException($"Invalid manifest format. Manifest root element must contain a version '{VersionAttributeName}' attribute");
+            }
+
+            if (version.Value != "1.0")
```
rynowak: One thought I have ... re:versioning. I think it's good that you put a version in the file-format, but that isn't really a get out of jail free card. SDKs support **ALL** versions of .NET, so we're not really free to *just* make breaking changes to the file-format, it has to be conscious of the version of the runtime.

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsWebHostBuilderExtensions.cs
rynowak: @Tratcher @davidfowl - if we need to replace the file provider, what's the best way to do that?

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsHelper.cs
rynowak: I don't expect you to resolve this in this PR 😆 - just felt like mentioning it.

src/DefaultBuilder/src/WebHost.cs
rynowak: Consider just making this not be an extension method. There's no reason it has to be, and it's not user-facing.

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsHelper.cs
```diff
@@ -0,0 +1,137 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Xml.Linq;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.FileProviders;
+
+namespace Microsoft.AspNetCore
```
rynowak: Folder should match namespace (I'd recommend dropping the folder).

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsHelper.cs
```diff
@@ -0,0 +1,137 @@
...
+        {
+            var attribute = element.Attribute(attributeName);
+            if (attribute == null)
+            {
+                throw new InvalidOperationException($"Invalid manifest format. Missing {attributeName} attribute in '{ContentRootElementName}' element.");
+            }
+            return attribute.Value;
+        }
+
+        internal class ContentRootMapping
```
rynowak: readonly struct?

src/DefaultBuilder/test/Microsoft.AspNetCore.Tests/StaticWebAssets/StaticWebAssetsFileProviderTests.cs
```diff
@@ -0,0 +1,72 @@
...
+
+namespace Microsoft.AspNetCore.Tests.StaticWebAssets
+{
+    public class StaticWebAssetsFileProviderTests
+    {
+        [Fact]
+        public void StaticWebAssetsFileProvider_ConstructorThrows_WhenPathIsNotFound()
+        {
+            // Arrange, Act & Assert
+            var provider = Assert.Throws<DirectoryNotFoundException>(() => new StaticWebAssetsFileProvider("/prefix", "/nonexisting"));
```
rynowak: What does this error message look like? Is this the default exception thrown by .NET? Should we try to provide some additional context about why the user is seeing this error?

src/DefaultBuilder/test/Microsoft.AspNetCore.Tests/StaticWebAssets/StaticWebAssetsFileProviderTests.cs
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        [Fact]
+        public void StaticWebAssetsFileProvider_Constructor_PrependsPrefixWithSlashIfMissing()
+        {
+            // Arrange & Act
+            var provider = new StaticWebAssetsFileProvider("_content", AppContext.BaseDirectory);
+
+            // Assert
+            Assert.Equal("/_content", provider.BasePath);
```
rynowak: I don't really have an context about how to write these tests in a way that's friendly to helix, but it seems worthy to figure out. (if you haven't already).

src/DefaultBuilder/test/Microsoft.AspNetCore.Tests/StaticWebAssets/StaticWebAssetsHelperTests.cs
```diff
@@ -0,0 +1,237 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Xml;
+using Microsoft.AspNetCore.Hosting.Internal;
+using Microsoft.Extensions.FileProviders;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Tests.StaticWebAssets
```
rynowak: `Microsoft.AspNetCore` - here and other places.

rynowak:
> Can someone catch me up on the quick ~2 sentence reason for this new integration? Are static web assets part of single-file deployment or something else?

This is part of an E2E for multi-project development *with static assets* for Razor/Blazor. For us to grow a library ecosystem for Blazor, we need provide a way for developers to package assets like css or js with their libraries.

This works by:
- Copying files to the output directory on publish
- Providing a set of extra paths to server for local builds (run without publish)

The mechanism in this case is a file that's produced by the build, but not included in publishing. 

The experience with this is really pretty nice because it will *just work*:
- As a package author
- As a person who uses multiple projects for *any* ASP.NET Core development
- As a person who uses rando nuget packages

That's more than two sentence tho 😢

src/DefaultBuilder/src/WebHost.cs
rynowak: Thinking about this more, I'm a little concerned about this being development only. Does this feature get used for files in nuget packages as well? If so, I don't think we can key this on `IsDevelopment()`.

anurse:
Cool, that's exactly what I was looking for!

src/DefaultBuilder/test/Microsoft.AspNetCore.Tests/StaticWebAssets/StaticWebAssetsFileProviderTests.cs
javiercn: No. The original exception contains all the information and we would simply obfuscate things if we provided our own.

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsFileProvider.cs
javiercn: Maybe not. I just didn’t want to cobble it together in here as this package is meant to be absolutely minimal.

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsWebHostBuilderExtensions.cs
davidfowl: We need to add something, there's nothing for this today.

javiercn: @davidfowl I kept it minimalistic because I wanted your input on this. Originally I had a ConfigureWebHostEnvironment addition to CreateAppConfiguration, but I didn't find it justified enough as the only thing to configure there are the file providers.

I also didn't want to invent anything based on settings or similar nor bake any concept into hosting, as it would be hard to make something generic based on settings for what we need.

src/DefaultBuilder/src/WebHost.cs
javiercn: We totally can, I think explaining the different cases might clarify things.
* `dotnet run` builds the manifest and places it on the output folder. 
  * Under development (`dotnet run`) we will find the manifest and load the related assets.
  * If you want to do `dotnet run` and use the prod environment, you can simply call the extension method to enable it.
* `dotnet <<dll>>` (after publish) the files coming from packages and referenced projects will already be in their final locations, so it will work for development or production.
* Even if we were to unconditionally enable this, it will work based on the presence of the manifest (which we never publish) so it would work for dotnet run and not for dotnet publish.
* The reason to key it off development is to make it clear that this is a development feature.

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsHelper.cs
javiercn: We do. I think it returns a bogus/empty stream. My tests were failing without it.

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsHelper.cs
javiercn: @rynowak The embedded resource part is going away, but I won't have time for all the dependencies to flow from the Razor SDK before I check this in, so this is just a tactical implementation until I can take the embedded resource part in preview7 and simply rely on the physical file.

https://blogs.msdn.microsoft.com/suzcook/2003/06/26/assembly-codebase-vs-assembly-location/

https://github.com/aspnet/AspNetCore/blob/master/src/Mvc/Mvc.Core/src/ApplicationParts/RelatedAssemblyAttribute.cs#L116

I will change the implementation to do what we do in Razor

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsHelper.cs
javiercn: If we evolve the SDK we will evolve this code in sync. I think the SDK would include logic to generate a version of the manifest that is compatible with the given runtime being targeted.

src/DefaultBuilder/src/WebHost.cs
javiercn: > Consider just making this not be an extension method. There's no reason it has to be, and it's not user-facing.

It was leftover from a previous implementation, didn't thought it matters given that it was internal, but I cleaned it up.

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsFileProvider.cs
javiercn: Updated it to use ` PathString.StartsWithSegments` and added a test for it. Thanks @Tratcher!

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsFileProvider.cs
javiercn: done

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsFileProvider.cs
javiercn: This is very specialized implementation for this scenario so I wouldn't pollute PhysicalFileProvider with this concept, if the need arises, we can always integrate it at a later point.

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsHelper.cs
javiercn: Done

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsHelper.cs
javiercn: done

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsHelper.cs
javiercn: changed the name

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsHelper.cs
javiercn: Updated

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsHelper.cs
javiercn: done

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsHelper.cs
javiercn: done

src/DefaultBuilder/src/StaticWebAssetsFileProvider.cs
```diff
@@ -0,0 +1,65 @@
...
+
+        /// <inheritdoc />
+        public IChangeToken Watch(string filter)
+        {
+            return InnerProvider.Watch(filter);
+        }
+
+        private bool StartsWithBasePath(string subpath, out PathString rest)
+        {
+            return new PathString(subpath).StartsWithSegments(BasePath, FilePathComparison, out rest);
```
Tratcher: Careful, the implicit PathString<->string converters change the encoding. BasePath needs to be a PathString too.

src/DefaultBuilder/src/StaticWebAssetsLoader.cs
```diff
@@ -0,0 +1,86 @@
...
+            else
+            {
+                additionalFiles.Insert(0, webRootFileProvider);
+                environment.WebRootFileProvider = new CompositeFileProvider(additionalFiles);
+            }
+        }
+
+        internal static Stream ResolveManifest(IWebHostEnvironment environment)
+        {
+            var assembly = Assembly.Load(environment.ApplicationName);
```
Tratcher: This is a poor assumption that the app name is an assembly. It could be anything.

This needs error handling.

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsFileProvider.cs
Tratcher: Are these ever going to be called from outside of CreateDefaultBuilder? If so then they need a different home.

Adding a new package is not a significant cost. Ask @aspnet/build if you need help.

src/DefaultBuilder/src/WebHost.cs
Tratcher: > Even if we were to unconditionally enable this, it will work based on the presence of the manifest (which we never publish) so it would work for dotnet run and not for dotnet publish.

Then `IsDevelopment` isn't actually needed if the feature behaves the same with or without it.

> The reason to key it off development is to make it clear that this is a development feature.

Insufficient. If there's not a beahavioral difference then don't check IsDevelopment.

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsHelper.cs
```diff
@@ -0,0 +1,137 @@
...
+            else
+            {
+                staticWebAssetContentRoots.Insert(0, webRootFileProvider);
+                environment.WebRootFileProvider = new CompositeFileProvider(staticWebAssetContentRoots);
+            }
+        }
+
+        internal static Stream ResolveManifest(IWebHostEnvironment environment)
+        {
+            var assembly = Assembly.Load(environment.ApplicationName);
```
Tratcher: You can't assume ApplicationName is always an assembly. You should provide an alternate way to locate the manifest that doesn't depend on ApplicationName.

This needs error handling.

src/DefaultBuilder/src/StaticWebAssetsFileProvider.cs
javiercn: Ah, could you spill the exact magic to use here? Maybe I can add a test that check encodings, but doesn't the url already arrive decoded?

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsWebHostBuilderExtensions.cs
Tratcher: ConfigureAppConfiguration is your best option for now. StaticWebAssets are cumulative so it's unlikely to break anyone else. Also, I've seen very few examples of anyone setting WebRootFileProvider so it's unlikely anyone else will break you.

Tratcher (changes requested):

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsFileProvider.cs
Tratcher: Alternatively these could be added to Hosting.

src/DefaultBuilder/src/StaticWebAssetsLoader.cs
javiercn: This is going away as we are moving the manifest to be a file, but this is also how MVC finds and loads the entry point, and I don't remember getting complaints about it.

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsHelper.cs
javiercn: The embedded resource piece is going away. Loading the assembly this way is what MVC does to find the entry point, and I don't remember hearing complains about it. I'll add a try-catch here for completeness, but I won't be adding a test as this code-path (embedded resource) is going away in preview7

javiercn:
> Insufficient. If there's not a beahavioral difference then don't check IsDevelopment.

I think of this feature in the same way I think of user secrets and user secrets is keyed of development. I'm not sure how I feel about the app using this feature in a non development environment.

We don't want to enable this in publish if people copy the file to the publish output by carelessly updating a `<content />` item in their project (or by some other SDK/props/targets including it by mistake)

Keyeing it off development ensures we don't run into those situations.

javiercn:
Are these ever going to be called from outside of ConfigureWebDefaults? If so then they need a different home.

No that I'm aware off (there is only a single extension method on `IWebHostBuilder` if you want to enable the feature unconditionally). 

I'm not opposed to move this to a new package (in fact it was my first thought) but I won't be doing this on preview6 as its risky. Based on @rynowak lets wait until we find a good reason to move it into a new package.

javiercn:
@HaoK Helix is failing due to a completely unrelated test. Do we care?

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsFileProvider.cs
javiercn: I wouldn't move these to hosting, as they are kind of an "opinion" about how we do things. Lets leave them here for now as they are an implementation detail. We can move them to hosting in future without causing a breaking change as they are all internal.

(I believe we could even typeforward the class with the extension methods if we wanted to)

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsWebHostBuilderExtensions.cs
javiercn: Yeah, I don't think this will break people or that people will be breaking this by changing it. It will be documented, so if someone runs into this, they can simply account for it.

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsFileProvider.cs
javiercn: Also, regarding efficiency, this is for development scenarios only (it doesn't affect published apps) so i'm not super concerned about the performance.

Even if you plug this in, static files will be considered first, so it is unlikely you see a hit here.

src/DefaultBuilder/src/WebHost.cs
javiercn: > Then `IsDevelopment` isn't actually needed if the feature behaves the same with or without it.

It prevents it from being used on published apps if you copy the manifest by mistake (msbuild things happen)

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsFileProvider.cs
Tratcher: The only opinion here is the line in ConfigureWebDefaults that adds this by default. Everything else should be moved elsewhere. This assembly does not provide features, only opinions.

src/DefaultBuilder/src/StaticWebAssetsFileProvider.cs
Tratcher: You don't need FilePathComparison, BasePath isn't considered as part of the file path, it's treated as a case insensitive prefix.
https://github.com/aspnet/AspNetCore/blob/08ed1afbbc1d1e11115eafcf4bd2c771c00ffd90/src/Middleware/StaticFiles/src/Helpers.cs#L40

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsFileProvider.cs
javiercn: @Tratcher Fair enough. I'm going to be in Redmond next week, so lets leave it here for now and discuss it at that time. I don't want to be creating a package 2 days from code complete and this is an implementation detail that can be handled at the beginning of preview7.

Does that sound right?

src/DefaultBuilder/src/StaticWebAssetsFileProvider.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// <inheritdoc />
+        public IFileInfo GetFileInfo(string subpath)
+        {
+            if (!StartsWithBasePath(subpath, out var physicalPath))
+            {
+                return new NotFoundFileInfo(subpath);
+            }
+            else
+            {
+                return InnerProvider.GetFileInfo(physicalPath);
```
Tratcher:
```suggestion
                return InnerProvider.GetFileInfo(physicalPath.Value);
```

src/DefaultBuilder/src/StaticWebAssetsFileProvider.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// <inheritdoc />
+        public IDirectoryContents GetDirectoryContents(string subpath)
+        {
+            if (!StartsWithBasePath(subpath, out var physicalPath))
+            {
+                return NotFoundDirectoryContents.Singleton;
+            }
+            else
+            {
+                return InnerProvider.GetDirectoryContents(physicalPath);
```
Tratcher: Avoid encoding again
```suggestion
                return InnerProvider.GetDirectoryContents(physicalPath.Value);
```

src/DefaultBuilder/src/StaticWebAssets/StaticWebAssetsFileProvider.cs
javiercn: I'll file an issue for it.

javiercn: https://github.com/aspnet/AspNetCore/issues/10656

src/DefaultBuilder/src/StaticWebAssetsLoader.cs
javiercn: Added a try-catch and silently ignore if there is an error.

HaoK:
@javiercn I don't believe helix-tests are a required check so I don't think you need to block on those, cc @aspnet/build to confirm that

javiercn:
@Tratcher I can't seem to be able to comment on the right thing
> You don't need FilePathComparison, BasePath isn't considered as part of the file path, it's treated as a case insensitive prefix.

We do need to maintain the case sensitivity of the OS to avoid discrepancies with the published outcome.

At publish time the files end up as physical files under wwwroot in the published app. For example, if the base path is `_content/library` and there is a file in `<<library>>\wwwroot\js\file.js`. The file ends up in `wwwroot\_content\library\js\file.js` on the published app folder. 

If we are in a case-sensitive file system and we don't respect the sensitivity, then during development a relative path to `/_CONTENT/LIBRARY/js/file.js` will work, and when the app its published and deployed it will start failing. To avoid that, we need to replicate the case sensitivity of the OS.

Keep in mind that the base path here represents a prefix to where the file ends up on disk, and not a base path in the URL sense.

@Tratcher Does that help clarify things?

Is there anything else to address or is it good to merge?

Tratcher:
> We do need to maintain the case sensitivity of the OS to avoid discrepancies with the published outcome. 

That assumes your dev and production environments use the same OS and casing restrictions. Even with this check it could work fine on windows and fail when published to Linux. This check only ends up protecting Unix -> Unix, but I guess some safety is better than none.

> Keep in mind that the base path here represents a prefix to where the file ends up on disk, and not a base path in the URL sense.

Ah, nothing in this PR conveys the meaning of BasePath, it's an arbitrary field in the manifest. You should add some doc comments to that effect.

javiercn:
> That assumes your dev and production environments use the same OS and casing restrictions. Even with this check it could work fine on windows and fail when published to Linux. This check only ends up protecting Unix -> Unix, but I guess some safety is better than none.

That's the point, it offers the same guarantees. I've added a doc comment explaining the purpose and workings of the base path, and there will be documentation in the official docs

Tratcher (approved):

src/DefaultBuilder/ref/Microsoft.AspNetCore.netcoreapp3.0.cs
```diff
@@ -3,6 +3,10 @@
 
 namespace Microsoft.AspNetCore
 {
+    public static partial class StaticWebAssetsWebHostBuilderExtensions
+    {
+        public static Microsoft.AspNetCore.Hosting.IWebHostBuilder UseStaticWebAssets(this Microsoft.AspNetCore.Hosting.IWebHostBuilder builder) { throw null; }
```
davidfowl: UseStaticAssets

src/DefaultBuilder/src/StaticWebAssetsLoader.cs
```diff
@@ -0,0 +1,96 @@
...
+                    // return early. Misconfigurations will be uncommon given that the entire process is automated
+                    // at build time.
+                    return null;
+                }
+            }
+        }
+
+        internal static string GetAssemblyLocation(Assembly assembly)
+        {
+            if (Uri.TryCreate(assembly.CodeBase, UriKind.Absolute, out var result) &&
```
davidfowl: Why? Just use the location?

src/DefaultBuilder/src/StaticWebAssetsReader.cs
```diff
@@ -0,0 +1,75 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Xml.Linq;
+
+namespace Microsoft.AspNetCore
+{
+    internal static class StaticWebAssetsReader
```
davidfowl: It's super weird that we use xml....

src/DefaultBuilder/src/WebHost.cs
```diff
@@ -206,6 +206,13 @@ public static IWebHostBuilder CreateDefaultBuilder(string[] args)
 
         internal static void ConfigureWebDefaults(IWebHostBuilder builder)
         {
+            builder.ConfigureAppConfiguration((ctx,cb) =>
```
davidfowl: nit: formatting

src/DefaultBuilder/src/StaticWebAssetsLoader.cs
```diff
@@ -0,0 +1,96 @@
...
+            Assembly assembly = null;
+            try
+            {
+                assembly = Assembly.Load(environment.ApplicationName);
+            }
+            catch (Exception)
+            {
+            }
+
+            if (assembly != null && assembly.GetManifestResourceNames().Any(a => a == StaticWebAssetsManifestName))
```
davidfowl: When are things embedded attributes? Identity?

src/DefaultBuilder/src/StaticWebAssetsLoader.cs
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        internal static Stream ResolveManifest(IWebHostEnvironment environment)
+        {
+            // We plan to remove the embedded file resolution code path in
+            // a future preview.
+            Assembly assembly = null;
+            try
+            {
+                assembly = Assembly.Load(environment.ApplicationName);
```
davidfowl: Why swallow the failure? Can we not?

javiercn: As mentioned in the comment, this code-path is going away in preview7.

src/DefaultBuilder/src/StaticWebAssetsLoader.cs
javiercn: No, this is going away in preview7 as we moved to use a file. Just need it here until the Razor SDK flows.

src/DefaultBuilder/src/StaticWebAssetsReader.cs
javiercn: Its friendlier on tooling because its available in netstandard. Otherwise It would have required to create a custom tool to bundle the json dependency so that the task would invoke a tool so that visual studio msbuild (full .net could call it) without having to load the json dependency into the build. This avoids unnecessary pain.

src/DefaultBuilder/src/StaticWebAssetsLoader.cs
javiercn: Because when you run tests and do things like shadow-copying (like in tests), the location points to the wrong place. This is what MVC does to find related assemblies as part of application parts, and it has been improved a couple of times based on user feedback.

davidfowl: So you’re telling me everywhere we use location is incorrect?

javiercn: @davidfowl It is correct in most cases, but it doesn't work with things like shadow-copying. I posted a link above to an MSDN blog explaining it, and as I said, this is the result of our own experience.

minhhieugma:
This PR is on Preview 7 now, but I cannot find out any instructions to use it. Could someone help us?

javiercn:
@minhhieugma https://docs.microsoft.com/en-us/aspnet/core/razor-pages/ui-class?view=aspnetcore-3.0&tabs=visual-studio#create-an-rcl-with-static-assets

PR #10634, opened 2019-05-30
Use SemVer2 in packaging

### LINKED ISSUE #4339: Switch package versions to SemVer 2.0

natemcmaster:
Resolves #4339 

Internal validation build: https://dev.azure.com/dnceng/internal/_build/results?buildId=205179&_a=summary

### DIFF
--- a/version.props
+++ b/version.props
@@ -34,13 +34,11 @@
     <_BuildNumberShortDate>$([MSBuild]::Add($([MSBuild]::Add($([MSBuild]::Multiply($(_BuildNumberYY), 1000)), $([MSBuild]::Multiply($(_BuildNumberMM), 50)))), $(_BuildNumberDD)))</_BuildNumberShortDate>
 
     <VersionSuffixBuildOfTheDay>$([System.Convert]::ToInt32($(_BuildNumberR)))</VersionSuffixBuildOfTheDay>
-    <VersionSuffixBuildOfTheDayPadded>$(VersionSuffixBuildOfTheDay.PadLeft(2, '0'))</VersionSuffixBuildOfTheDayPadded>
 
-    <!-- TODO: consider using semver 2.0 instead, when/if https://github.com/dotnet/core-setup/issues/4795 is resolved -->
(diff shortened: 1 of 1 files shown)


anurse:
I'll grab the JARs and publish them to a private feed so we can do a quick sanity check on them.

anurse:
Though I see the PR build doesn't use these build numbers :). Still, it's trivial to hack it up so I can still run a quick test.

natemcmaster:
You can get the jar from https://dev.azure.com/dnceng/internal/_build/results?buildId=205179&_a=summary

anurse:
The main thing I actually want to test is that it's treated like SemVer 2 (i.e. integer comparisons), so I need to hack something up anyway. Fortunately it's pretty trivial to do.

anurse:
It looks good (Maven can download a version with `.`s in the pre-release tag) based on a MyGet feed I tried (https://myget.org/F/maven-semver-test/maven/). MyGet doesn't produce the right metadata to check that `3.0.0-foo.10` is offered as an update for `3.0.0-foo.2` but I know how we could test that. I don't feel a need to block this PR on that though.

Java packages look good to me!

analogrelay (approved):

PR #10635, opened 2019-05-30
 Don't remove auth header when stopping HubConnection

### LINKED ISSUE #10575: Java SignalR client clears authorization header after reconnect
Hello. 
My signalr hubs uses authentication based on Authorization header value.
For some reasons I am not using bearer token authentication scheme, so I cant use HttpHubConnectionBuilder.withAccessTokenProvider() method for authentication. 
Instead of this I use something like HubConnectionBuilder.create("https...").withHeader("Authorization", "somestring")

When disconnect occurred I am trying to perform reconnect to the server:

    connection = HubConnectionBuilder.create("...")
        .shouldSkipNegotiate(true)
        .withHeader("Authorization", "...")
        .build()

    connection.onClosed {
        connect()
    }
    ...
    fun connect(){
        connection.start()
    }

After reconnect server cant authenticate this client because of empty Authorization header.
It looks like issue in [HubConnection.java#L496](https://github.com/aspnet/AspNetCore/blob/master/src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java#L496)
I think that there should be check about token source and clearing authorization header only if its value was populated via access token provider

mikaelm12:
Fixes: https://github.com/aspnet/AspNetCore/issues/10575

We were removing the Authorization header after the connection was stopped. This causes issues when users don't set the header via the `AccessTokenProvider`. 
An alternative for the linked issue would be to have an option that would allow the user to specify something other than bearer auth with the `AccessTokenProvider`.

### DIFF
--- a/src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
+++ b/src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
@@ -57,7 +57,6 @@ public class HubConnection {
     private Map<String, Observable> streamMap = new ConcurrentHashMap<>();
     private TransportEnum transportEnum = TransportEnum.ALL;
     private String connectionId;
-    private boolean tokenSetByNegotiate;
     private final Logger logger = LoggerFactory.getLogger(HubConnection.class);
 
     /**
@@ -275,7 +274,6 @@ public class HubConnection {
                 this.redirectAccessTokenProvider = Single.just(negotiateResponse.getAccessToken());
                 // We know the Single is non blocking in this case
                 // It's fine to call blockingGet() on it.
-                this.tokenSetByNegotiate = true;
                 String token = this.redirectAccessTokenProvider.blockingGet();
                 this.localHeaders.put("Authorization", "Bearer " + token);
             }
@@ -336,7 +334,7 @@ public class HubConnection {
                         transport = new LongPollingTransport(localHeaders, httpClient, tokenProvider);
                         break;
                     default:
-                        transport = new WebSocketTransport(localHeaders, httpClient, tokenProvider);
+                        transport = new WebSocketTransport(localHeaders, httpClient);
                 }
             }
 
--- a/src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
+++ b/src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
@@ -41,7 +41,7 @@ public class HubConnection {
     private final boolean skipNegotiate;
     private Single<String> accessTokenProvider;
     private Single<String> redirectAccessTokenProvider;
-    private Map<String, String> headers;
+    private final Map<String, String> headers;
     private final Map<String, String> localHeaders = new HashMap<>();
     private ConnectionState connectionState = null;
     private HttpClient httpClient;
@@ -148,10 +148,7 @@ public class HubConnection {
             this.handshakeResponseTimeout = handshakeResponseTimeout;
         }
 
-        if (headers != null) {
-            this.headers = headers;
-        }
-
+        this.headers = headers;
         this.skipNegotiate = skipNegotiate;
 
         this.callback = (payload) -> {
@@ -304,7 +301,7 @@ public class HubConnection {
         handshakeResponseSubject = CompletableSubject.create();
         handshakeReceived = false;
         CompletableSubject tokenCompletable = CompletableSubject.create();
-        if(headers != null) {
+        if (headers != null) {
             this.localHeaders.putAll(headers);
         }
 
--- a/src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
+++ b/src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
@@ -56,6 +56,7 @@ public class HubConnection {
     private Map<String, Observable> streamMap = new ConcurrentHashMap<>();
     private TransportEnum transportEnum = TransportEnum.ALL;
     private String connectionId;
+    private boolean tokenSetByNegotiate;
     private final Logger logger = LoggerFactory.getLogger(HubConnection.class);
 
     /**
@@ -273,6 +274,7 @@ public class HubConnection {
                 this.redirectAccessTokenProvider = Single.just(negotiateResponse.getAccessToken());
                 // We know the Single is non blocking in this case
                 // It's fine to call blockingGet() on it.
+                this.tokenSetByNegotiate = true;
                 String token = this.redirectAccessTokenProvider.blockingGet();
                 this.headers.put("Authorization", "Bearer " + token);
             }
@@ -493,7 +495,9 @@ public class HubConnection {
             redirectAccessTokenProvider = null;
             connectionId = null;
             transportEnum = TransportEnum.ALL;
-            this.headers.remove("Authorization");
+            if (tokenSetByNegotiate) {
+                this.headers.remove("Authorization");
+            }
         } finally {
             hubConnectionStateLock.unlock();
         }
--- a/src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
+++ b/src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
@@ -331,7 +331,7 @@ public class HubConnection {
                         transport = new LongPollingTransport(headers, httpClient, tokenProvider);
                         break;
                     default:
-                        transport = new WebSocketTransport(headers, httpClient);
+                        transport = new WebSocketTransport(headers, httpClient, tokenProvider);
                 }
             }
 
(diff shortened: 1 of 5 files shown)


analogrelay (commented):
My only minor concern is that it's likely we were removing this header for a reason. What happens if you start the connection once, get an `accessToken` in the `/negotiate` response, then re-start the connection and get a **new** `accessToken` (or no token at all) in the negotiate payload? Will we end up with duplicate `Authorization` headers?

mikaelm12:
Yeah ,that is the concern. I think we should make it possible for users to use auth schemes other than bearer with the accessTokenProvider rather than just not clearing out headers when restarting the connection. Otherwise we'd need some internal state to determine if the auth header was set by the user or via the negotiate response and clear conditionally.

anurse:
> I think we should make it possible for users to use auth schemes other than bearer with the accessTokenProvider

TBH there aren't really other auth schemes that make sense here. The vast majority of auth models used here would be: Cookies (which don't use that header at all), Windows/Kerberos (which is automatically handled by the platform) and Bearer auth. Other schemes would represent completely custom logic. If a user needs that they can use either the existing header extensibility or a custom HTTP client (in case they need to "refresh" the token).

If we want to support custom authentication schemes, I'm not opposed, but it wouldn't be through `.withAccessTokenProvider` since "access tokens" are explicitly a Bearer auth concept. The likely design would be a "credential provider" which is given the chance to do whatever it wants to the request (adding headers, etc.).

The scenario I think we **do** need a test for is the one I specified above, a "reconnect" where the `accessToken` returned by `/negotiate` changes (or is removed) in the second connection. That I would block the PR on since it could be a major break for existing apps.

Time is a bit tight to get this in for preview 6, but as long as the verification is there I'm OK taking it. It's also fine if it doesn't land in time for preview 6, we can do it first thing after branching.

mikaelm12:
> The scenario I think we do need a test for is the one I specified above, a "reconnect" where the accessToken returned by /negotiate changes (or is removed) in the second connection

Agreed

mikaelm12:
⬆️ 📆

src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
```diff
@@ -493,7 +495,9 @@ private void stopConnection(String errorMessage) {
             redirectAccessTokenProvider = null;
             connectionId = null;
             transportEnum = TransportEnum.ALL;
-            this.headers.remove("Authorization");
+            if (tokenSetByNegotiate) {
+                this.headers.remove("Authorization");
```
BrennanConroy: Why are we removing the header in the first place? I don't think we do anything like this for the other clients.
I looked at when we added it and the issue was that we weren't saving the original `accessTokenProvider`.

I think all of our clients (at least C# and Java) have a problem if someone sets the `Authorization` header via options and then either provides an `accessTokenProvider` or uses the service. Because the original `Authorization` header will be replaced and never used again. Which could cause the client to never be able to reconnect if the server is doing custom auth based on the original `Authorization` header. We should probably consider a broader fix for that class of issues.

halter73: We should not be adding the result of `redirectAccessTokenProvider` directly to `this.headers` only in negotiate and relying on the headers dictionary to keep the state for the duration of the connection. Instead, we should have a prepare requests method that calls `redirectAccessTokenProvider` for every request and re-adds the token to the to a dictionary used only for that request.

mikaelm12: We actually had the correct behavior in the LongPolling transport. Before any request we we call an `updateHeaderToken` method. For the Websockets transport we only care about updating the token right before we connect, so we shouldn't need extra logic in the transport, right? Updating the token in the `HubConnection` 's start should be enough.

mikaelm12: The main issue still is how do we want to handle users explicitly setting the Auth header, because currently they wouldn't be able to use the SignalR service. The reported issue would be fixed because we don't clear the user set auth header when restarting a connection

halter73: It's just difficult to manually undo changes you make to a dictionary. It's better to shallow-copy the original dictionary, add the entry you want, and then make the single websocket/longpolling/whatever request with the shallow-copied dictionary. That way you're sure you never end up in a state were the original dictionary is in an invalid state.

mikaelm12:
I updated this to have 2 header collections. One that stores the headers initialized from the user, and another collection that gets cleared every time the connection is restarted. That collection gets the accessTokens and is passed into the transports

src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
```diff
@@ -41,7 +41,8 @@
     private final boolean skipNegotiate;
     private Single<String> accessTokenProvider;
     private Single<String> redirectAccessTokenProvider;
-    private final Map<String, String> headers = new HashMap<>();
+    private Map<String, String> headers;
```
halter73: Since this isn't assigned outside of the ctor, can this still be marked final?

src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
```diff
@@ -56,6 +57,7 @@
     private Map<String, Observable> streamMap = new ConcurrentHashMap<>();
     private TransportEnum transportEnum = TransportEnum.ALL;
     private String connectionId;
+    private boolean tokenSetByNegotiate;
```
halter73: Do you still need this field?

halter73 (approved):
LGTM pending nits

src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
mikaelm12: Nope!

src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
mikaelm12: It can't because it's conditionally set

halter73: You could just set it unconditionally though, right?

src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
```diff
@@ -303,9 +304,13 @@ public Completable start() {
         handshakeResponseSubject = CompletableSubject.create();
         handshakeReceived = false;
         CompletableSubject tokenCompletable = CompletableSubject.create();
+        if(headers != null) {
```
BrennanConroy:
```suggestion
        if (headers != null) {
```

src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/WebSocketTransport.java
```diff
@@ -15,17 +17,15 @@
     private OnReceiveCallBack onReceiveCallBack;
     private TransportOnClosedCallback onClose;
     private String url;
-    private final HttpClient client;
-    private final Map<String, String> headers;
-
+    private HttpClient client;
```
BrennanConroy: nit: revert

src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/WebSocketTransport.java
```diff
@@ -5,6 +5,8 @@
 
 import java.util.Map;
 
+import io.reactivex.Single;
+import io.reactivex.subjects.CompletableSubject;
```
BrennanConroy: nit: remove

src/SignalR/clients/java/signalr/src/test/java/com/microsoft/signalr/HubConnectionTest.java
```diff
@@ -1885,6 +1931,34 @@ public void headersAreSetAndSentThroughBuilder() {
         assertEquals("ExampleValue", header.get());
     }
 
+    @Test
+    public void headersAreNotClearedWhenConnectionIsRestarted() {
```
BrennanConroy: I think there should be another test that is a combination of the two just added. Where user sets an Authorization header explicitly and then there is a negotiate redirect with a new header value, and then stopping and starting again preserves the original user header

BrennanConroy (approved):

src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
halter73: Or just set the headers field unconditionally

src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
mikaelm12: Sure

mikaelm12:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=216832

PR #10636, opened 2019-05-30
Add SignInWithClaims overload and set amr claim appropriately

### LINKED ISSUE #5783: AuthenticationMethod for 2FA
I wonder if the the `SignInManager.DoTwoFactorSignInAsync` should be adding an explicit `AuthenticationMethod` to the `SignInMAnager.SignInAsync` call to indicate that the login used 2FA.

Currently, it is difficult for an application using aspnet Identity to determine, whether 2FA was used during the sign in. The `amr` claim contains only `password`, regardless of whether the login was 1FA or 2FA.

HaoK:
Fixes https://github.com/aspnet/AspNetCore/issues/5783

New SignInWithClaimsAsync overload that adds arbitrary claims to the cookie
Also amr: pwd and mfa claims are added for PasswordSignIn vs TwoFactorSignIn

### DIFF
--- a/src/Identity/Core/src/SignInManager.cs
+++ b/src/Identity/Core/src/SignInManager.cs
@@ -169,8 +169,19 @@ namespace Microsoft.AspNetCore.Identity
         public virtual async Task RefreshSignInAsync(TUser user)
         {
             var auth = await Context.AuthenticateAsync(IdentityConstants.ApplicationScheme);
-            var authenticationMethod = auth?.Principal?.FindFirstValue(ClaimTypes.AuthenticationMethod);
-            await SignInAsync(user, auth?.Properties, authenticationMethod);
+            var claims = new List<Claim>();
+            var authenticationMethod = auth?.Principal?.FindFirst(ClaimTypes.AuthenticationMethod);
+            if (authenticationMethod != null)
+            {
+                claims.Add(authenticationMethod);
+            }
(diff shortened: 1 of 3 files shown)


src/Identity/Core/src/SignInManager.cs
```diff
@@ -438,9 +473,13 @@ private async Task DoTwoFactorSignInAsync(TUser user, TwoFactorAuthenticationInf
             // When token is verified correctly, clear the access failed count used for lockout
             await ResetLockout(user);
 
+            var claims = new List<Claim>();
+            claims.Add(new Claim("amr", "mfa"));
```
blowdart: @brockallen The spec says if we have mfa we can also have specific authentication methods used, so would be expect to also see two more mfa claims of pwd, and otp here?

HaoK: It would require some more plumbing to flow things through, plus its user token provider specific what that is, its not always authenticator for example, that just happens to be what our token provider uses.  Unless we want to add that to the interface, its going to be a bit difficult

brockallen: Sure, you can have any number that mean whatever they need to. IdentityServer will also accept any number of `amr` claims, and in addition will interpret the ClaimsIdentity's AuthenticationType property as an `amr` as well.

ajcvickers (approved):

Issue #10508, opened 2019-05-23
Missing symbols for the Microsoft.Extensions.ApiDescription.Client package


dougbu:
Symbol check fails consistently for the assemblies in this package.

See https://dnceng.visualstudio.com/internal/_releaseProgress?_a=release-environment-logs&releaseId=7988&environmentId=22139 Details:
```
2019-05-23T22:21:37.0497796Z Validating Microsoft.Extensions.ApiDescription.Client.0.3.0-preview6-19272-03.nupkg 
2019-05-23T22:21:38.8856205Z 	 Checking file E:\A\_work\r1\a\checkSymbols\3dbc120b-8427-478b-90c3-3e27525d6ced\tasks\net461\Microsoft.Extensions.ApiDescription.Tasks.dll ... No symbols found on MSDL or SymWeb!
2019-05-23T22:21:40.7486530Z 	 Checking file E:\A\_work\r1\a\checkSymbols\3dbc120b-8427-478b-90c3-3e27525d6ced\tasks\netstandard2.0\Microsoft.Extensions.ApiDescription.Tasks.dll ... No symbols found on MSDL or SymWeb!
2019-05-23T22:21:40.8294325Z CheckSymbolsAvailable : Missing symbols for 2 modules in the package 
```

dougbu:
25672336f92accfa8d24823d11d0bb5528c9ad4d

Issue #6582, opened 2019-01-10
Add mechanism to disable project build for ProjectServiceReference


JunTaoLuo:
From the meeting, it seems like a ProjectServiceReference requires the project referenced to be built. We'd like to add a mechanism to disable this requirement since it would be redundant in GRPC scenarios.

cc @dougbu @rynowak

mkArtakMSFT:
@glennc and @dougbu please sync up regarding this during your next sync up meeting.

dougbu:
We sync'd up.

This property (renamed to `$(OpenApiGenerateOnBuild)`) remains useful after the client / server split. Will control whether projects with (say) an NSwag.AspNetCore dependency get generated JSON files in their project. When `false`, the documents will not be saved to disk unless the user manually runs one of the CLI tools or hooks an `<Exec/>` of a tool into their project.

dougbu:
This will come just after #8242

dougbu:
25672336f92accfa8d24823d11d0bb5528c9ad4d

Issue #4924, opened 2018-10-07
Make it possible to change when client code generation runs


dougbu:
#### Is this a Bug or Feature request?
Enhancement that's part of #4896

#### Description of the problem
Currently, any project that references the Microsoft.Extensions.ApiDescription.Client will have the `ServiceFileReferenceGenerator` and related targets run unconditionally before the `BeforeCompile` target. Make this conditional so that users can configure the targets to run elsewhere in the build pipeline. For example, use `$(BuildDependsOn)`, `$(CoreBuildDependsOn)`, `$(CompileDependsOn)` or `$(CoreCompileDependsOn)` and make it conditional on a documented property.

Not positive where hook should land but prepending it into `$(BuildDependsOn)` seems better than current spot because the code generation targets do not rely on other `Build` or `Compile` work.

#### Version of `Microsoft.AspNetCore.Mvc` or `Microsoft.AspNetCore.App` or `Microsoft.AspNetCore.All`
See the [features/client.code.generation](https://github.com/aspnet/Mvc/tree/feature/client.code.generation) branch. That work is intended for the 2.2 Preview 3 milestone.

dougbu:
25672336f92accfa8d24823d11d0bb5528c9ad4d

Issue #4923, opened 2018-10-07
Clean up TODOs and duplicated code in client code generation


dougbu:
#### Is this a Bug or Feature request?
Enhancement that's part of #4896

#### Description of the problem
- [ ] replace `Microsoft.DotNet.Cli.CommandLine` code with `Microsoft.Extensions.CommandLineUtils.Sources` package (#13323)
- [x] replace custom lookup with `Internal.WebHostBuilderFactory.Sources` package (4c8ca0b08067f0b821d728b1d3f22559f34fdfe7)
- [ ] perhaps, replace console writing code with ~~`Microsoft.Extensions.Logging`~~ (#13323)
  - not positive because adding dependencies to the tool seems perilous
  - instead used `Reporter` class shared with other tools in this repo

#### Version of `Microsoft.AspNetCore.Mvc` or `Microsoft.AspNetCore.App` or `Microsoft.AspNetCore.All`
See the [features/client.code.generation](https://github.com/aspnet/Mvc/tree/feature/client.code.generation) branch. That work is intended for the 2.2 Preview 3 milestone.

dougbu:
New bullets from #10641 discussions:
- remove `net461` tasks assembly and stick with the `netstandard2.0` one
   - see https://github.com/aspnet/AspNetCore/pull/10641/files#r289548486
- rename `GetCurrentItems` to something less generic in client targets file
  - see https://github.com/aspnet/AspNetCore/pull/10641/files#r289549577
- add an escape hatch for the properties removed when calling Web API projects: `%(OpenApiProjectReference.GlobalPropertiesToRemove)`
  - see https://github.com/aspnet/AspNetCore/pull/10641/files#r289551199

dougbu:
New bullet from a #10669 discussion:
- [ ] add `@(FileWrites)` items for code and document files as well as the OpenApiFiles.cache file
  - see https://github.com/aspnet/AspNetCore/pull/10669/files#r289567963

dougbu:
Couple more bullets
- [ ] avoid NU1704 build warnings and other client / server incompatibility issues
- [ ] change default `$(OpenApiGenerateDocuments)` to `false` in projects using .NET Core 2.0 or earlier
  - not supported by `dotnet-getdocument` tool in the Microsoft.Extensions.ApiDescription.Server package
[...]

PR #10641, opened 2019-05-30
Add more properties controlling service reference code generation

dougbu:
- `$(OpenApiGenerateBeforeCompile)` controls if targets run before compile targets
  - #4924
  - also correct multiple invocations when project has multiple target frameworks
- `$(OpenApiBuildReferencedProjects)` controls whether `@(OpenApiProjectReference)` items build automatically
  - #6582

also:
- add symbols for Microsoft.Extensions.ApiDescription.Client task assembly
  - #10508
- unconditionally run `OpenApiGetDocuments` target in referenced projects
  - corrects compilation in design-time builds
  - no longer uses `@(ProjectReferenceWithConfiguration)`; referenced project chooses all property values

nits:
- don't remove `$(TargetFrameworks)` when invoking `OpenApiGetDocuments` target (not necessary)
- consolidate into a single `$(GenerateOpenApiReferenceCodeDependsOn)` property
- rename task assembly and namespaces in Microsoft.Extensions.ApiDesription.Client to match the project
- allow `OpenApiGetDocuments` targets to run in parallel if `$(BuildInParallel)` is enabled
- remove `$(OpenApiDefaultOutputDirectory)` normalization; never concatenated with anything else
- remove dangling `GenerateServiceFileReferenceCode` names


src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.targets
```diff
@@ -30,19 +27,16 @@
...
 
     <MSBuild Targets="OpenApiGetDocuments"
-        Projects="@(ProjectReferenceWithConfiguration)"
-        Condition="'%(ProjectReferenceWithConfiguration.OpenApiReference)' == 'true'"
-        Properties="%(ProjectReferenceWithConfiguration.SetConfiguration); %(ProjectReferenceWithConfiguration.SetPlatform); %(ProjectReferenceWithConfiguration.SetTargetFramework)"
+        BuildInParallel="$(BuildInParallel)"
+        Projects="@(OpenApiProjectReference)"
         RebaseOutputs="true"
-        RemoveProperties="%(ProjectReferenceWithConfiguration.GlobalPropertiesToRemove);TargetFrameworks;RuntimeIdentifier">
+        RemoveProperties="Configuration;Platform;RuntimeIdentifier;TargetFramework;TargetFrameworks">
```
dougbu: @rainersigwald is removing these properties sufficient?

And, am I missing something important in the shift from `@(ProjectReferenceWithConfiguration)` to using `@(OpenApiProjectReference)` items directly? (Without this, design-time builds would still have been broken and `'$(OpenApiBuildReferencedProjects)' == 'false'` would not work.)

dougbu:
@natemcmaster I'd of course appreciate any input you have on the MSBuild changes

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.props
```diff
@@ -1,35 +1,58 @@
...
   </PropertyGroup>
-  <UsingTask TaskName="GetCurrentItems" AssemblyFile="$(_ApiDescriptionTasksAssemblyPath)" />
-  <UsingTask TaskName="GetFileReferenceMetadata" AssemblyFile="$(_ApiDescriptionTasksAssemblyPath)" />
+  <UsingTask TaskName="GetCurrentItems" AssemblyFile="$(_ApiDescriptionClientAssemblyPath)" />
+  <UsingTask TaskName="GetFileReferenceMetadata" AssemblyFile="$(_ApiDescriptionClientAssemblyPath)" />
 
   <!--
     Settings users may update as they see fit.
   -->
   <PropertyGroup>
```
dougbu: @glennc do you have any input on the names of these user-visible properties?

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.targets
```diff
@@ -30,19 +27,16 @@
     </ProjectReference>
   </ItemGroup>
 
-  <Target Name="_CreateOpenApiReferenceItemsForOpenApiProjectReferences"
-      AfterTargets="ResolveProjectReferences"
-      Condition="'$(DesignTimeBuild)' != 'true' AND '$(BuildingProject)' == 'true'">
+  <Target Name="_CreateOpenApiReferenceItemsForOpenApiProjectReferences" DependsOnTargets="ResolveProjectReferences">
```
pranavkm: Should this have `Condition="@(OpenApiProjectReference->Count()) != 0"`?

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.props
```diff
@@ -1,35 +1,58 @@
...
     <!--
-      If 'true', will generate code for OpenApiReference items during design-time builds. Otherwise, generate code only
-      for output files that do not yet exist during design-time builds.
+      If 'true' (the default), generate code for @(OpenApiReference) and @(OpenApiProjectReference) items before the
+      BeforeCompile target.
+
+      If 'false', the GenerateOpenApiReferenceCode target is not part of the build (by default) but can run when
+      explicitly referenced. That is, the target may be invoked from the command line or tied in through another target.
+    -->
+    <OpenApiGenerateBeforeCompile Condition="'$(OpenApiGenerateBeforeCompile)' == ''">true</OpenApiGenerateBeforeCompile>
```
pranavkm: `OpenApiGenerateOnBuild`? That's similar to `RazorCompileOnBuild`. Also removes having this be tied to `BeforeCompile` which is essentially an implementation detail.

src/Mvc/Extensions.ApiDescription.Client/src/buildMultiTargeting/Microsoft.Extensions.ApiDescription.Client.targets
```diff
@@ -1,9 +1,14 @@
...
+  <Target Name="GenerateOpenApiReferenceCode">
+    <MSBuild Projects="$(MSBuildProjectFile)"
+        Targets="GenerateOpenApiReferenceCode"
         Properties="TargetFramework=$(TargetFrameworks.Split(';')[0])"
-        RemoveProperties="TargetFrameworks;RuntimeIdentifier" />
+        RemoveProperties="RuntimeIdentifier" />
   </Target>
+
+  <Target Name="_TieInGenerateOpenApiReferenceCode"
+      BeforeTargets="BeforeCompile"
```
pranavkm: Should this be `CoreCompile`. Would that allow users to do user things ™️ in `BeforeCompile`?

pranavkm (approved):
Looks good to me. I wouldn't block on any of my comments.

JunTaoLuo (approved):

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.props
dougbu: This property is specifically about moving execution elsewhere in the build. So, I like the current name. But, I defer to @glennc here.

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.targets
dougbu: It really doesn't matter because everything in this target no-ops when there's nothing worth doing. Let me know if you feel differently

src/Mvc/Extensions.ApiDescription.Client/src/buildMultiTargeting/Microsoft.Extensions.ApiDescription.Client.targets
dougbu: `CoreCompile` is late because it comes after the MSBuild targets cache the list of references and source files involved in the compilation

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.props
glennc: OpenApiGenerateCodeOnBuild

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.props
```diff
@@ -1,35 +1,58 @@
...
-        Condition="'$(MSBuildRuntimeType)' == 'Core'">netstandard2.0</_ApiDescriptionTasksAssemblyTarget>
-    <_ApiDescriptionTasksAssemblyTarget
-        Condition="'$(MSBuildRuntimeType)' != 'Core'">net461</_ApiDescriptionTasksAssemblyTarget>
-    <_ApiDescriptionTasksAssemblyPath>$(MSBuildThisFileDirectory)/../tasks/$(_ApiDescriptionTasksAssemblyTarget)/Microsoft.Extensions.ApiDescription.Tasks.dll</_ApiDescriptionTasksAssemblyPath>
-    <_ApiDescriptionTasksAssemblyTarget />
+    <_ApiDescriptionClientAssemblyTarget
+        Condition="'$(MSBuildRuntimeType)' == 'Core'">netstandard2.0</_ApiDescriptionClientAssemblyTarget>
+    <_ApiDescriptionClientAssemblyTarget
+        Condition="'$(MSBuildRuntimeType)' != 'Core'">net461</_ApiDescriptionClientAssemblyTarget>
+    <_ApiDescriptionClientAssemblyPath>$(MSBuildThisFileDirectory)/../tasks/$(_ApiDescriptionClientAssemblyTarget)/Microsoft.Extensions.ApiDescription.Client.dll</_ApiDescriptionClientAssemblyPath>
```
rynowak: Is this really relevant anymore? MSBuild supports netstandard2.0 everywhere .NET Core runs.

eng/ProjectReferences.props
```diff
@@ -8,7 +8,7 @@
     <ProjectReferenceProvider Include="Microsoft.AspNetCore.Identity.Specification.Tests" ProjectPath="$(RepoRoot)src\Identity\Specification.Tests\src\Microsoft.AspNetCore.Identity.Specification.Tests.csproj"  />
     <ProjectReferenceProvider Include="Microsoft.Web.Xdt.Extensions" ProjectPath="$(RepoRoot)src\SiteExtensions\Microsoft.Web.Xdt.Extensions\src\Microsoft.Web.Xdt.Extensions.csproj"  />
     <ProjectReferenceProvider Include="Microsoft.AspNetCore.DeveloperCertificates.XPlat" ProjectPath="$(RepoRoot)src\Tools\FirstRunCertGenerator\src\Microsoft.AspNetCore.DeveloperCertificates.XPlat.csproj"  />
-    <ProjectReferenceProvider Include="Microsoft.Extensions.ApiDescription.Tasks" ProjectPath="$(RepoRoot)src\Mvc\Extensions.ApiDescription.Client\src\Microsoft.Extensions.ApiDescription.Client.csproj"  />
+    <ProjectReferenceProvider Include="Microsoft.Extensions.ApiDescription.Client" ProjectPath="$(RepoRoot)src\Mvc\Extensions.ApiDescription.Client\src\Microsoft.Extensions.ApiDescription.Client.csproj"  />
```
rynowak: I might be missing some necessary context here, but I don't really get the point of renaming an assembly containing tasks to not be suffixed with `.Tasks`. 

Is the metapoint here that we want to separate the tasks used for client-gen from the tasks used inside server-projects?

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.props
```diff
@@ -1,35 +1,58 @@
...
+    <_ApiDescriptionClientAssemblyTarget />
   </PropertyGroup>
-  <UsingTask TaskName="GetCurrentItems" AssemblyFile="$(_ApiDescriptionTasksAssemblyPath)" />
-  <UsingTask TaskName="GetFileReferenceMetadata" AssemblyFile="$(_ApiDescriptionTasksAssemblyPath)" />
+  <UsingTask TaskName="GetCurrentItems" AssemblyFile="$(_ApiDescriptionClientAssemblyPath)" />
+  <UsingTask TaskName="GetFileReferenceMetadata" AssemblyFile="$(_ApiDescriptionClientAssemblyPath)" />
```
rynowak: random observation: `GetCurrentItems` is a pretty generic name given that `item` is a common term in MSBuild. We might want to choose more specific names. I don't think `GetFileReferenceMetadata` is bad.

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.props
```diff
@@ -1,35 +1,58 @@
...
 
   <!--
     Settings users may update as they see fit.
   -->
   <PropertyGroup>
+    <!--
+      Options added to the code generator command line by default. Provides the default %(Options) metadata of
+      @(OpenApiReference) and @(OpenApiProjectReference) items.
+    -->
```
rynowak: I'm always really happy to see quality docs in MSBuild 👍

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.props
```diff
@@ -1,35 +1,58 @@
...
+      BeforeCompile target.
+
+      If 'false', the GenerateOpenApiReferenceCode target is not part of the build (by default) but can run when
+      explicitly referenced. That is, the target may be invoked from the command line or tied in through another target.
+    -->
+    <OpenApiGenerateBeforeCompile Condition="'$(OpenApiGenerateBeforeCompile)' == ''">true</OpenApiGenerateBeforeCompile>
+
+    <!--
+      If 'true' (the default), generate code for @(OpenApiReference) and @(OpenApiProjectReference) items during
+      design-time builds. Otherwise, generate code only during a full build.
```
rynowak: If you're going with the name `OpenApiGenerateCodeOnBuild` would this become `OpenApiGenerateCodeOnDesignTimeBuild`?

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.props
rainersigwald: If the assembly has any dependencies that aren't in netstandard _as implemented in .NET 4.7.2_, I'd leave this in. Netstandard task assemblies are probably much more viable now that VS is on 4.7.2, but they've historically been a problem and splitting into core and full implementations has been required.

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.targets
rynowak: Simple is better, I agree with Doug.

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.targets
rainersigwald: I'd definitely want to keep something like `%(OpenApiProjectReference.GlobalPropertiesToRemove)` in there as an escape hatch. This list looks ok to me, but there's always a risk of things changing (like when multitargeting was added) or user projects having their own custom dimensions.

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.targets
```diff
@@ -85,10 +79,9 @@
       Condition="$(OpenApiGenerateAtDesignTime) OR ('$(DesignTimeBuild)' != 'true' AND '$(BuildingProject)' == 'true')"
       Inputs="@(OpenApiReference)"
       Outputs="%(OutputPath)">
-    <MSBuild BuildInParallel="$(BuildInParallel)"
-        Projects="$(MSBuildProjectFullPath)"
+    <MSBuild Projects="$(MSBuildProjectFullPath)"
```
rynowak: Is this idea with this call that we only want to call this once even with multi-targeting? If that's the case should we have a comment?

Basically I'm not sure why this is a separate invocation of MSBuild, a comment would help.

rynowak (approved):
I left some minor comments here but nothing blocking.

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.props
rynowak: I'm mentioning it because we've been doing ns2.0 in Razor for a while now with no problems. 

Since this is new code it should be fine as long as we test it in VS right?

rainersigwald: Yeah, as long as it works in VS, command line msbuild.exe, and command line msbuild.exe from a Build Tools install of VS, you should be good to go.

dougbu: I'll file an issue deferring this cleanup and testing in to Preview 7

eng/ProjectReferences.props
dougbu: That was my original point but I ended up not needing tasks in the server project and decided I preferred using the same name everywhere.

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.props
dougbu: I've added this to #4923 since the extra bits do no harm. This is all an implementation detail and it'll take time to do the necessary testing&hellip;

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.props
dougbu: I added the suggestion to #4923

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.props
dougbu: Thought about it but the sense is a bit different. @glennc and I decided this one was fine as-is. Could of course change it if users disagree.

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.targets
dougbu: Great suggestion! I added this to #4923 because I have other bits to do short-term.

src/Mvc/Extensions.ApiDescription.Client/src/build/Microsoft.Extensions.ApiDescription.Client.targets
dougbu: Yes, this is about running things once. And, sure, I'll add a comment

dougbu:
:up::date: to add the comment @rynowak suggested. Also rebased on 'master' but deferred most other suggestions to #4923

PR #10650, opened 2019-05-30
EndpointMiddleware does not invoke endpoint

### LINKED ISSUE #10649: EndpointMiddleware will not work if SuppressCheckForUnhandledSecurityMetadata is true
When `RouteOptions.SuppressCheckForUnhandledSecurityMetadata` is set to true, `EndpointMiddleware` would not execute any Endpoint.

I think the reason is `EndpointMiddleware` must skip checking for Auth and Cors Middleware, instead of return a CompletedTask.

This is the current code
```
if (_routeOptions.SuppressCheckForUnhandledSecurityMetadata)
{
    // User opted out of this check.
    return Task.CompletedTask;
}

if (endpoint.Metadata.GetMetadata<IAuthorizeData>() != null &&
!httpContext.Items.ContainsKey(AuthorizationMiddlewareInvokedKey))
{
    ThrowMissingAuthMiddlewareException(endpoint);
}

if (endpoint.Metadata.GetMetadata<ICorsMetadata>() != null &&
!httpContext.Items.ContainsKey(CorsMiddlewareInvokedKey))
{
    ThrowMissingCorsMiddlewareException(endpoint);
}
```

I think change it to this would solve the issue
```
if (!_routeOptions.SuppressCheckForUnhandledSecurityMetadata)
{
    if (endpoint.Metadata.GetMetadata<IAuthorizeData>() != null &&
    !httpContext.Items.ContainsKey(AuthorizationMiddlewareInvokedKey))
    {
        ThrowMissingAuthMiddlewareException(endpoint);
    }

    if (endpoint.Metadata.GetMetadata<ICorsMetadata>() != null &&
    !httpContext.Items.ContainsKey(CorsMiddlewareInvokedKey))
    {
        ThrowMissingCorsMiddlewareException(endpoint);
    }
}
```

rynowak:
Fixes: #10649

The bug is that the endpoint delegate is not invoked when
SuppressCheckForUnhandledSecurityMetadata is set to true. This option is
provided so that can you suppress the security checks done by routing,
but right now what it does is suppress the entire middleware.

We had tests for the supression cases, but they didn't actually validate
that the middleware did any work, just that we don't throw. Fixed that.

/cc @Kahbazi

### DIFF
--- a/src/Http/Routing/src/EndpointMiddleware.cs
+++ b/src/Http/Routing/src/EndpointMiddleware.cs
@@ -36,22 +36,19 @@ namespace Microsoft.AspNetCore.Routing
             var endpoint = httpContext.Features.Get<IEndpointFeature>()?.Endpoint;
             if (endpoint?.RequestDelegate != null)
             {
-                if (_routeOptions.SuppressCheckForUnhandledSecurityMetadata)
+                if (!_routeOptions.SuppressCheckForUnhandledSecurityMetadata)
                 {
-                    // User opted out of this check.
-                    return Task.CompletedTask;
-                }
+                    if (endpoint.Metadata.GetMetadata<IAuthorizeData>() != null &&
+                        !httpContext.Items.ContainsKey(AuthorizationMiddlewareInvokedKey))
+                    {
+                        ThrowMissingAuthMiddlewareException(endpoint);
+                    }
 
-                if (endpoint.Metadata.GetMetadata<IAuthorizeData>() != null &&
-                    !httpContext.Items.ContainsKey(AuthorizationMiddlewareInvokedKey))
-                {
-                    ThrowMissingAuthMiddlewareException(endpoint);
-                }
-
-                if (endpoint.Metadata.GetMetadata<ICorsMetadata>() != null &&
-                    !httpContext.Items.ContainsKey(CorsMiddlewareInvokedKey))
-                {
-                    ThrowMissingCorsMiddlewareException(endpoint);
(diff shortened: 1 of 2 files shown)


Kahbazi:
Since `Invoke_WithEndpoint_DoesNotThrowIfUnhandledCorsAttributesWereFound_ButSuppressedViaOptions` and `Invoke_WithEndpoint_WorksIfCorsMetadataWasFound_AndCorsMiddlewareInvoked` is changed, should `Invoke_WithEndpoint_DoesNotThrowIfUnhandledAuthAttributesWereFound_ButSuppressedViaOptions` and `Invoke_WithEndpoint_WorksIfAuthAttributesWereFound_AndAuthMiddlewareInvoked` also change?

rynowak:
Ah, you're right. I missed that one. I'm going to go ahead and update all of these tests to run the endpoint and verify it.

rynowak:
I've hardened these tests to make the expectations clearer about what delegates run and what don't.

src/Http/Routing/src/EndpointMiddleware.cs
```diff
@@ -36,22 +36,19 @@ public Task Invoke(HttpContext httpContext)
             var endpoint = httpContext.Features.Get<IEndpointFeature>()?.Endpoint;
             if (endpoint?.RequestDelegate != null)
             {
-                if (_routeOptions.SuppressCheckForUnhandledSecurityMetadata)
+                if (!_routeOptions.SuppressCheckForUnhandledSecurityMetadata)
```
pranavkm: Oops

pranavkm (approved):

PR #10655, opened 2019-05-30
Remove CaptureStartupErrors from ANCM out of proc

jkotalik:
Follow up to https://github.com/aspnet/AspNetCore/issues/9667


jkotalik:
Actually, I'm closing this after discussions with @Tratcher . This isn't the right fix for out of process as it would disable the hosting startup error page in development (and production). We need to verify the experience today and see if hosting already logs a startup exception to the eventlog if hosting throws.

PR #10659, opened 2019-05-30
Don't inject ANCM managed exception page into the static page

jkotalik:
Another thing noticed by verification. I actually added the ANCM startup page into the static page.


BrennanConroy (approved):
Should have asked for screen-shots in the original PR :D

jkotalik:
Yeah probably, but it significantly easier to get error pages when doing verification. 

Capture

PR #10660, opened 2019-05-30
Update Identity tests for EF breaking changes

ajcvickers:
Specifically:
* Removal of Relational()
* Query changes

There is some kind of flakiness still impacting these tests. It may be a race condition in EF query generation, or it may be an issue with the test infrastructure

Also, one database error page test is currently disabled.

### DIFF
--- a/src/Identity/EntityFrameworkCore/src/UserOnlyStore.cs
+++ b/src/Identity/EntityFrameworkCore/src/UserOnlyStore.cs
@@ -250,6 +250,7 @@ namespace Microsoft.AspNetCore.Identity.EntityFrameworkCore
         {
             cancellationToken.ThrowIfCancellationRequested();
             ThrowIfDisposed();
+
             return Users.FirstOrDefaultAsync(u => u.NormalizedUserName == normalizedUserName, cancellationToken);
         }
 
@@ -311,7 +312,8 @@ namespace Microsoft.AspNetCore.Identity.EntityFrameworkCore
                 throw new ArgumentNullException(nameof(user));
             }
 
-            return await UserClaims.Where(uc => uc.UserId.Equals(user.Id)).Select(c => c.ToClaim()).ToListAsync(cancellationToken);
+            // TODO: Fix once EF query works Issue #10668
+            return (await UserClaims.Where(uc => uc.UserId.Equals(user.Id)).ToListAsync(cancellationToken)).Select(c => c.ToClaim()).ToList();
         }
 
         /// <summary>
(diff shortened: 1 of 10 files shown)


smitpatel (approved):

src/Identity/EntityFrameworkCore/src/UserOnlyStore.cs
```diff
@@ -311,7 +312,8 @@ public async override Task<IList<Claim>> GetClaimsAsync(TUser user, Cancellation
                 throw new ArgumentNullException(nameof(user));
             }
 
-            return await UserClaims.Where(uc => uc.UserId.Equals(user.Id)).Select(c => c.ToClaim()).ToListAsync(cancellationToken);
+            // TODO: Fix once EF query works
```
Tratcher: Include an issue link with every TODO

ajcvickers: Issues don't exist yet, and it's more urgent to get this in. I will make sure there are issues.

HaoK (approved):
I'm assuming most of these will be addressed by EF in preview7?

ajcvickers:
@HaoK Yes, or before.

ajcvickers:
@HaoK @natemcmaster I can't figure out how to get the ref assemblies failure to pass. I can't see any differences, so I don't understand why it is failing. Any ideas?

HaoK:
Try running `dotnet msbuild /t:GenerateReferenceSource` from the src directories of the refs you want to update, if that works, it usually does the trick

HaoK:
Okay I see the issue, you made something not async, so you need to remove the 

```
```

HaoK:
Remove  `[System.Diagnostics.DebuggerStepThroughAttribute]` from
```
         public override System.Threading.Tasks.Task<System.Collections.Generic.IList<System.Security.Claims.Claim>> GetClaimsAsync(TUser user, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
``` 
in src/Identity/EntityFrameworkCore/ref/Microsoft.AspNetCore.Identity.EntityFrameworkCore.netcoreapp3.0.cs.

ajcvickers:
@natemcmaster @dougbu @JunTaoLuo Now this is merged we should be good to take the latest EF dependency here.

/cc @Pilchie @smitpatel @divega

natemcmaster:
Thanks!

dougbu:
I created #10682 as a draft PR to see how things go.

@ajcvickers EF Core is behind on its CoreFx, core-setup and Extensions dependencies. Any objection to EveryBuild there? (We left it and EF6 when we sped everything else up.)

smitpatel:
Every build is too much noise.

ajcvickers:
@dougbu Yes, every build is fine while we are working towards coherence for the preview.

@smitpatel We will switch it off once we get to where we need to be for preview.

dougbu:
Done

PR #10663, opened 2019-05-30
Dylan/double semaphores experiment

DylanDmitri:
Redid the structure.

No more underlying RequestQueue, the middleware directly uses 2 SemaphoreSlims

Is an approach like this okay? Or is it too tricky to scale?


src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
```diff
@@ -16,9 +15,11 @@ namespace Microsoft.AspNetCore.RequestThrottling
     /// </summary>
     public class RequestThrottlingMiddleware
     {
-        private readonly RequestQueue _requestQueue;
+        private readonly SemaphoreSlim _queueSemaphore;
```
halter73: Why isn't this just a counter with a lock? I see the call _queueSemaphore.WaitAsync(), but it's never actually awaited. You just reject the request if the returned Task is incomplete, but you could do the exact same thing much more simply with an integer and lock object.

src/Middleware/RequestThrottling/src/Internal/RequestQueue.cs
```diff
@@ -1,73 +0,0 @@
-// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
-using System;
-using System.Threading;
-using System.Threading.Tasks;
-
-namespace Microsoft.AspNetCore.RequestThrottling.Internal
-{
-    internal class RequestQueue : IDisposable
```
halter73: Please keep the RequestQueue and the middleware segregated.

src/Middleware/RequestThrottling/test/MiddlewareTests.cs
```diff
@@ -103,7 +103,7 @@ public async void FullQueueResultsIn503Error()
         {
             var middleware = TestUtils.CreateTestMiddleware(
                 maxConcurrentRequests: 0,
-                requestQueueLimit: -1);
+                requestQueueLimit: 0);
```
halter73: This part is good.

src/Middleware/RequestThrottling/src/Internal/RequestQueue.cs
DylanDmitri: My problem was that I wanted to forget about layers and focus on eliminating redundancy. I am re-prioritizing these for future decisions.

Issue #8242, opened 2019-03-06
Add Web Application project support for build-time OpenAPI document generation


mkArtakMSFT:
@glennc, the experience for this should be defined.
Here are the list of items we need clarity on:
- [x] Define the changes in `csproj` file indicating that the document should be auto-generated
  - no .csproj changes should be needed by default e.g. projects that include the new project must opt **out** of document generation during builds
- [x] Define the details regarding where the generated document should be placed (defaults)
  - will use the `obj/` directory by default
- [x] Define user's options for overriding our defaults
  - minimal
- [x] Define options for generating additional document(s)
  - no new options for this, build getting the document names list into `dotnet-getdocument` tool

mkArtakMSFT:
Here are the list of items we need clarity on:
- [ ] Define the changes in `csproj` file indicating that the document should be auto-generated
- [ ] Define the details regarding where the generated document should be placed (defaults)
- [ ] Define user's options for overriding our defaults
- [ ] Define options for generating additional document(s)

dougbu:
We have enough of a spec that I'll get a design PR out in this milestone. Reassigning to myself and placing the completed installation in Preview6.

dougbu:
Work to do:
- [ ] create new Microsoft.Extensions.ApiDescription.Server project / package containing the bits of the old Microsoft.Extensions.ApiDescription.Design project / package specific to getting documents
- [ ] change MSBuild infrastructure to work on the _current_ project
- [ ] enhance `dotnet-getdocument` tool to use new `GetDocumentNames(...)` method in `IDocumentProvider` interface
  - main point is to avoid user ever having to duplicate document names in .csproj and `Startup` class
  - cache document names somewhere for use in MSBuild infrastructure
- [ ] as a follow up to #4945, place documents in `$(BaseIntermediateOutputPath)`
- [ ] implement `OpenApiGetDocuments` target
- [ ] add a project capability

related issues:
- #4912
- #4941
- #6582

dougbu:
part 1: 4c8ca0b08067f0b821d728b1d3f22559f34fdfe7

dougbu:
part 2 (of 2): b3ddd65a15456d9fdf4642d5f7174418f4e8b9c5

Issue #4914, opened 2018-09-08
Add tests of client code generation components


dougbu:
#### Is this a Bug or Feature request?
Enhancement that's part of #4896

#### Description of the problem
Have previously been manually testing the client code generation feature. Need to automate this, including tests of the tasks, the generic document tool, and the MSBuild targets.

Tasks should be relatively simple to unit test. But, the remainder will likely need functional tests that (unfortunately) write to disk. Will have a look at aspnet/DotNetTools, aspnet/EntityFrameworkCore, and aspnet/websdk to see how those repos test tools and MSBuild targets. May also discuss potential MSBuild and tool testing approaches with @pranavkm and @ryanbrandenburg.

#### Version of `Microsoft.AspNetCore.Mvc` or `Microsoft.AspNetCore.App` or `Microsoft.AspNetCore.All`
See the [features/client.code.generation branch](https://github.com/aspnet/Mvc/tree/feature/client.code.generation) branch. That work is intended for the 2.2 Preview 3 milestone.

dougbu:
Part 1 in 04b1adbc2e49

Pilchie:
@dougbu I'm assuming any further tests coming wouldn't be in the Preview 9 branch, right?  Can we update the milestone for this?

dougbu:
@Pilchie your assumption was correct 😺

dougbu:
More added in 24fc9c000865

PR #10667, opened 2019-05-30
Rewrite dotnet-getdocument and GetDocument.Insider to retrieve all documents in one go

dougbu:
- #8242 1 of 2
- save a cache file listing all retrieved documents
- remove .NET Core App 2.0 support
- remove ServiceProjectReferenceMetadata.targets, related `Project` class, and searches for a project
  - tools will run within a project and get needed information from project on command line
- roll framework forward in both tools to expand their applicability

nits:
- refactor methods in `GetDocumentCommandWorker`
- reorder option addition for consistency and to place `--help` at the top of help output
- consolidate information about method signatures at top of `GetDocumentCommandWorker`
- consolidate `try` / `catch` blocks in `GetDocumentCommandWorker`


src/Mvc/GetDocumentInsider/src/Commands/GetDocumentCommandWorker.cs
```diff
@@ -22,200 +46,325 @@ public static int Process(GetDocumentCommandContext context)
                 return 2;
             }
 
-            var services = GetServices(entryPointType, context.AssemblyPath, context.AssemblyName);
-            if (services == null)
+            static string FormatException(Exception exception) => $"{exception.GetType().FullName}: {exception.Message}";
```
rynowak: consider putting this after the return statement to separate it a bit. It's a little jarring to see a single-line local function defined in the middle of a method.

javiercn (approved):
Pranav asked me to help him look at this.

Overall I think looks good. I left a few comments. Mostly I would strongly recommend you to reuse the logic for finding the app entry point, as we share that everywhere.

I'm not sure about the constraints this tool has, but there are other things that are a bit concerning to me, like reading the deps file or the runtime.config.json file.

I also don't see any tests for this PR, not sure if its because there are already tests and didn't require changes or because there aren't any. If its the latter, I think there should be at least one, as this tool does a lot of things that are prone to integration errors.

That said, I don't see anything that strikes me as wrong here.

src/Mvc/GetDocumentInsider/src/Commands/GetDocumentCommandWorker.cs
rynowak: I would really really strongly encourage the use of `Exception.ToString()` instead of hiding the information behind a verbose logging switch. 

By doing this you can optimize for users getting an experience that they troubleshoot themselves (failure in their own app), or can use to immediately report a problem to us (copy-paste the call stack).

src/Mvc/GetDocumentInsider/src/Commands/GetDocumentCommandWorker.cs
```diff
@@ -22,200 +46,325 @@ public static int Process(GetDocumentCommandContext context)
...
+            }
+
+            var builder = (IWebHostBuilder)InvokeMethod(methodInfo, instance: null, arguments: CreateArguments);
+            if (builder == null)
+            {
+                return null;
+            }
+
+            return builder.Build().Services;
+        }
```
rynowak: /cc @javiercn

src/Mvc/GetDocumentInsider/src/Commands/GetDocumentCommandWorker.cs
```diff
@@ -22,200 +46,325 @@ public static int Process(GetDocumentCommandContext context)
...
-                var service = services.GetService(serviceType);
-                if (service == null)
+                filePathList.Add(filePath);
+            }
+
+            // Write out the cache file.
+            var stream = File.Create(context.FileListPath);
+            using var writer = new StreamWriter(stream) { AutoFlush = true };
+            writer.WriteLine(string.Join(Environment.NewLine, filePathList));
+            stream.Flush();
```
rynowak: Is there a really strong need for either `AutoFlush` or the explict call to `flush`? the stream will flush when it's disposed.

src/Mvc/GetDocumentInsider/src/Commands/GetDocumentCommandWorker.cs
```diff
@@ -22,200 +46,325 @@ public static int Process(GetDocumentCommandContext context)
...
+            string documentName,
+            string projectName,
+            string outputDirectory,
+            MethodInfo generateMethod,
+            object service)
+        {
+            Reporter.WriteInformation(Resources.FormatRetrievingDocument(documentName));
+
+            var stream = new MemoryStream();
+            using var writer = new StreamWriter(stream) { AutoFlush = true };
```
rynowak: Similar question about `AutoFlush`

src/Mvc/GetDocumentInsider/src/Commands/GetDocumentCommandWorker.cs
```diff
@@ -22,200 +46,325 @@ public static int Process(GetDocumentCommandContext context)
...
+            var resultTask = (Task)InvokeMethod(generateMethod, service, new object[] { documentName, writer });
+            if (resultTask == null)
+            {
+                return null;
+            }
+
+            var finishedIndex = Task.WaitAny(resultTask, Task.Delay(TimeSpan.FromMinutes(1)));
```
rynowak: Could this be simplified using https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.wait?view=netframework-4.8#System_Threading_Tasks_Task_Wait_System_TimeSpan_

src/Mvc/GetDocumentInsider/src/Commands/GetDocumentCommandWorker.cs
rynowak: I've noticed that this is really prevalent all over the code, I'd be interested to learn if this is here for any important reason.

src/Mvc/GetDocumentInsider/src/Commands/GetDocumentCommandWorker.cs
```diff
@@ -22,200 +46,325 @@ public static int Process(GetDocumentCommandContext context)
...
+                if (stream.Length == 0L)
                 {
-                    Reporter.WriteWarning(Resources.FormatServiceNotFound(serviceName));
-                    return false;
+                    Reporter.WriteError(Resources.FormatMethodWroteNoContent(
+                        GenerateMethodName,
+                        DocumentService,
+                        documentName));
+
+                    return null;
```
rynowak: Is there a reason for this to be inside the `try/catch`? This seems like it belongs higher up.

src/Mvc/GetDocumentInsider/src/Commands/GetDocumentCommandWorker.cs
```diff
@@ -22,200 +46,325 @@ public static int Process(GetDocumentCommandContext context)
...
-                foreach (var innerException in ex.Flatten().InnerExceptions)
+                // Sanitize the document name because it may contain almost any character, including illegal filename
+                // characters such as '/' and '?' and the string "..". Do not treat slashes as folder separators.
+                var sanitizedDocumentName = string.Join("_", documentName.Split(InvalidFilenameCharacters));
+                while (sanitizedDocumentName.Contains(InvalidFilenameStrings[0]))
                 {
-                    Reporter.WriteWarning(FormatException(innerException));
+                    sanitizedDocumentName = string.Join(
+                        ".",
+                        sanitizedDocumentName.Split(InvalidFilenameStrings, StringSplitOptions.None));
```
rynowak: Could this simplified and and made more obvious using `string.Replace`?

src/Mvc/dotnet-getdocument/src/Commands/InvokeCommand.cs
```diff
@@ -16,155 +16,123 @@ internal class InvokeCommand : HelpCommandBase
...
-                            Resources.FormatUnsupportedFramework(project.ProjectName, targetFramework.Identifier));
+                            Resources.FormatUnsupportedFramework(projectName, targetFramework.Identifier));
                 }
 
                 args.AddRange(_args);
                 args.Add("--assembly");
-                args.Add(project.TargetPath);
+                args.Add(assemblyPath);
+                args.Add("--project");
+                args.Add(projectName);
```
rynowak: I'm probably not the best person to review the inside-man stuff. I've never really known how this works 😕

rynowak:
The general spirit of the changes here makes sense to me and sees like a good enhancement. Due to the diffs being what they are, I'm not really sure if I'm reviewing old code or new code, but I still found some things that seem worthy of a look.

dougbu:
@javiercn I realize I missed your high-level comments, sorry:
>  things that are a bit concerning to me, like reading the deps file or the runtime.config.json file.

None of the code reads either of these files though it does scan the project assets file for probing paths. The `dotnet` command line doesn't have a `--please-also-use-these-assets` option.

See similar code in the `dotnet-ef` tool: https://github.com/aspnet/EntityFrameworkCore/blob/078449fd1363bc3d79778a4f726262e2ced30ae5/src/dotnet-ef/RootCommand.cs#L127-L139 So, fairly battle-tested

> I also don't see any tests for this PR

This feature has some engineering debt I'll be catching up on in the next milestone. #4914 covers the test part of that.

dougbu (commented):
:up::date; to address more comments than I expected

src/Mvc/GetDocumentInsider/src/Commands/GetDocumentCommandWorker.cs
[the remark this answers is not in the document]
dougbu: FYI I had to do this with the other `StreamWriter` because the `MemoryStream` was used before it was disposed. Did the more obvious thing of giving those two objects separate lifetimes. Cleaned this instance up.

PR #10675, opened 2019-05-31
Improve shutdown of SignalR .NET client sample

halter73:
Small SignalR .NET client sample cleanup that I made when @BrennanConroy and I were investigating potential logging issues.

This basically ensure that the sample can be stopped reliably using Ctrl+C even if in the middle of connecting/reconnecting.

### DIFF
--- a/src/SignalR/samples/ClientSample/HubSample.cs
+++ b/src/SignalR/samples/ClientSample/HubSample.cs
@@ -29,7 +29,9 @@ namespace ClientSample
         public static async Task<int> ExecuteAsync(string baseUrl)
         {
             var uri = baseUrl == null ? new Uri("net.tcp://127.0.0.1:9001") : new Uri(baseUrl);
+
             Console.WriteLine("Connecting to {0}", uri);
+
             var connectionBuilder = new HubConnectionBuilder()
                 .ConfigureLogging(logging =>
                 {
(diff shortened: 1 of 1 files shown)


anurse:
Scene from the movie "Austin Powers" in which a security guard is about to be run over by a steamroller and throws his hand out in a "stop" gesture

src/SignalR/samples/ClientSample/HubSample.cs
```diff
@@ -53,58 +55,57 @@ public static async Task<int> ExecuteAsync(string baseUrl)
...
             // Set up handler
             connection.On<string>("Send", Console.WriteLine);
 
-            CancellationTokenSource closedTokenSource = null;
-
             connection.Closed += e =>
             {
-                // This should never be null by the time this fires
                 closedTokenSource.Cancel();
+                connection.DisposeAsync().GetAwaiter().GetResult();
```
davidfowl: Why isn't this being awaited?

anurse:
Kinda cutting it close... as of noon this change will need QB approval to get in to preview 6

src/SignalR/samples/ClientSample/HubSample.cs
```diff
@@ -52,88 +53,91 @@ public static async Task<int> ExecuteAsync(string baseUrl)
...
-
-                Console.WriteLine("Connection closed...");
-                return Task.CompletedTask;
-            };
-
+                Console.CancelKeyPress += (sender, a) =>
+                {
+                    a.Cancel = true;
+                    closedTokenSource.Cancel();
+                    connection.StopAsync().GetAwaiter().GetResult();
```
davidfowl: @halter73 fix this please.

halter73: fix your glasses 😆

src/SignalR/samples/ClientSample/HubSample.cs
davidfowl: NM, this callback is sync

PR #10678, opened 2019-05-31
Unify C# and TS SignalR client auto-reconnect APIs

### LINKED ISSUE #10128: Unify TS and C# auto-reconnect APIs
In 3.0, auto-reconnect support is being added to the TS and C# SignalR clients. The APIs are mostly the same except for two notable differences.

In the TS client, we have:

```TS
IReconnectPolicy.nextRetryDelayInMilliseconds(previousRetryCount: number, elapsedMilliseconds: number): number | null;
```

And in the C# client, we have:

```C#
IRetryPolicy.NextRetryDelay(RetryContext retryContext);

public sealed class RetryContext
{
    public long PreviousRetryCount { get; set; }
    public TimeSpan ElapsedTime { get; set; }
    public Exception RetryReason { get; set; }
}
```

The first difference is `IReconnectPolicy` vs `IRetryPolicy`. I prefer changing the TS client to use `IRetryPolicy` since the interface could be used for more than just reconnects in the future (e.g. retrying starting the connection initially).

The second difference is having a `RetryContext` instead of individual parameters. I like having the `RetryContext` since you don't need to add new overload methods to the interface to add more information. I wouldn't mind renaming `RetryContext` to `RetryInfo` in both clients though. I think "Info" is more familiar than "Context" in JS/TS, and I don't think it's too out-of-place in .NET.

The TS client changes would break anyone who wrote a custom IReconnectPolicy using the preview. I'm not sure if this warrants a breaking change announcement.

@anurse

halter73:
Fixes #10128.

This changes the older TS IReconnectPolicy to be more like the newer C# IRetryPolicy which takes a context instead of individual arguments. This will break custom IReconnectPolicy's that might have been written using an earlier preview.

One thing that's unchanged is the "InMilliseconds" suffix on nextRetryDelayInMilliseconds since TS/JS don't have a standard specialized TimeSpan type. Everyone just uses milliseconds.


src/SignalR/clients/ts/signalr/src/IRetryPolicy.ts
```diff
@@ -0,0 +1,30 @@
...
+    /** Called after the transport loses the connection.
+     *
+     * @param {RetryContext} retryContext Details related to the retry event to help determine how long to wait for the next retry.
+     *
+     * @returns {number | null} The amount of time in milliseconds to wait before the next retry. `null` tells the client to stop retrying.
+     */
+    nextRetryDelayInMilliseconds(retryContext: RetryContext): number | null;
+}
+
+export interface RetryContext {
```
halter73: I looked at making this a class with private setters, but I think this is better. It's how we define all our HubMessage types for example.

BrennanConroy (approved):

anurse:
Let's get this in before noon.

BrennanConroy:
>Let's get this in before noon.

It will need to target branch release/3.0-preview6

anurse:
Also **NOTE** we have branched so this needs to be retargetted to `release/3.0-preview6` to get in to that milestone.

mikaelm12 (approved):

anurse:
Do we feel this is essential for preview6? We need QB approval if so.

mikaelm12:
It's a breaking change so it would have been nice for preview6 but I don't think we need to force it.

BrennanConroy:
FYI @halter73, whenever making SignalR TS client changes we now also need to build the blazor project to update the "ref assembly" for "blazor.server.js".

halter73:
> Do we feel this is essential for preview6? We need QB approval if so.

Not particularly, but as @mikaelm12 it breaks people writing custom nextRetryDelayInMilliseconds callbacks. I don't know if (m)any other than me has done so though.

anurse:
I get it, but I think we need to limit the churn we take now that the coherency process is starting up.

PR #10683, opened 2019-05-31
Loop is not neccessary for checking createSelectorForSilentRouteProviders when first routeProvider is not IActionHttpMethodProvider

wu-yafeng:
 Loop is not neccessary for checking createSelectorForSilentRouteProviders when first routeProvider is not IActionHttpMethodProvider,break the loop if the there are routeProvider is IActionHttpMethodProvider before


rynowak (approved):
I agree, this is an improvement to the clarity of this code. Thanks!

PR #10688, opened 2019-05-31
Better circuit ids

javiercn:
* Generates cryptographically strong circuit ids.
  * 32 bits of entropy from a PRNG.
  * DataProtected for authenticated encryption.

This is a simplified version that contains most of the things we want in this space. Only missing piece is to use a cookie to store the Id and use a hash of the cookie as the token that roundtrips

Quick summary for @blowdart 
* I create 32 bytes of entropy with a PRNG.
* I data protect those bytes.
* I base64urlencode the result.

On validation:
* I decode the base64urlencoded payload.
* I unprotect the bytes.
  * Unprotecting the bytes is proof that the id was generated by our system.

Notes:
* The DataProtector instance can be configured through options.
* I don't think the PRNG needs to be configurable (as I haven't seen that done through the stack).


src/Components/Server/src/CircuitOptions.cs
```diff
@@ -45,5 +46,10 @@ public class CircuitOptions
         /// Defaults to <c>3 minutes</c>.
         /// </value>
         public TimeSpan DisconnectedCircuitRetentionPeriod { get; set; } = TimeSpan.FromMinutes(3);
+
+        /// <summary>
+        /// Gets the <see cref="IDataProtector"/> used to protect the generated circuit Ids.
+        /// </summary>
+        public IDataProtector CircuitIdProtector { get; set; }
```
javiercn: This is needed for crypto-agility

src/Components/Server/src/Circuits/CircuitIdFactory.cs
```diff
@@ -0,0 +1,65 @@
...
+
+namespace Microsoft.AspNetCore.Components.Server.Circuits
+{
+    // This is a singleton instance
+    // Generates strong criptographic ids for circuits that are protected with authenticated encryption.
+    internal class CircuitIdFactory
+    {
+        // Follow up with blowdart/levi. Create should be good enough, but not sure if we need
+        // to do anything additional for FIPS compliance.
+        private readonly RandomNumberGenerator _generator = RandomNumberGenerator.Create();
```
javiercn: I've searched through the codebase and this is how we do it in all places, so this must be the correct way to create it. The only reason I made this comment in the first place is because for some crypto algorithms there are special ways to create them in order to achieve FIPS compliance. This is not the case with the PRNG I believe

src/Components/Server/src/Circuits/CircuitIdFactory.cs
```diff
@@ -0,0 +1,65 @@
...
+        public CircuitIdFactory(IOptions<CircuitOptions> options)
+        {
+            _protector = options.Value.CircuitIdProtector;
+        }
+
+        // Generates a circuit id that is produced from a strong cryptographic random number generator
+        // we don't care about the underlying payload, other than its uniqueness and the fact that we
+        // authenticate encrypt it using data protection.
+        // For validation, the fact that we can unprotect the payload is guarantee enough.
+        public string CreateCircuitId()
```
javiercn: In the future we can pass additional information that we can validate, but this should be good enough as 32 bytes of entropy ensure us uniqueness

src/Components/Server/src/Microsoft.AspNetCore.Components.Server.csproj
```diff
@@ -52,10 +53,7 @@
   <ItemGroup>
     <EmbeddedResource Include="..\..\Browser.JS\dist\$(Configuration)\blazor.server.js" LogicalName="_framework/%(Filename)%(Extension)" />
 
-    <EmbeddedResource
-      Include="..\..\Browser.JS\dist\$(Configuration)\blazor.server.js.map"
-      LogicalName="_framework/%(Filename)%(Extension)"
-      Condition="Exists('..\..\Browser.JS\dist\$(Configuration)\blazor.server.js.map')" />
+    <EmbeddedResource Include="..\..\Browser.JS\dist\$(Configuration)\blazor.server.js.map" LogicalName="_framework/%(Filename)%(Extension)" Condition="Exists('..\..\Browser.JS\dist\$(Configuration)\blazor.server.js.map')" />
```
javiercn: I'm not gonna fight visual studio formatting.

src/Components/Server/src/Circuits/PostConfigureCircuitOptionsCircuitIdDataprotector.cs
```diff
@@ -0,0 +1,23 @@
...
+        private const string CircuitIdProtectorPurpose = "Microsoft.AspNetCore.Components.Server";
+        private readonly IDataProtectionProvider _dataProtectionProvider;
+
+        public PostConfigureCircuitOptionsCircuitIdDataprotector(IDataProtectionProvider dataProtectionProvider)
+        {
+            _dataProtectionProvider = dataProtectionProvider;
+        }
+        public void PostConfigure(string name, CircuitOptions options)
+        {
+            options.CircuitIdProtector ??= _dataProtectionProvider.CreateProtector(CircuitIdProtectorPurpose);
```
javiercn: This is how other parts of the stack initialize the protector

src/Components/Server/src/CircuitOptions.cs
```diff
@@ -45,5 +46,10 @@ public class CircuitOptions
         /// Defaults to <c>3 minutes</c>.
         /// </value>
         public TimeSpan DisconnectedCircuitRetentionPeriod { get; set; } = TimeSpan.FromMinutes(3);
+
+        /// <summary>
+        /// Gets the <see cref="IDataProtector"/> used to protect the generated circuit Ids.
```
campersau:
```suggestion
        /// Gets or sets the <see cref="IDataProtector"/> used to protect the generated circuit Ids.
```

src/Components/Server/src/Circuits/CircuitIdFactory.cs
blowdart: PRNGs aren't part of FIPS, so you're ok here

src/Components/Server/src/Circuits/PostConfigureCircuitOptionsCircuitIdDataprotector.cs
blowdart: Yup, you're just copying session and that's the right way to go about it.

blowdart (approved):
LGTM, thanks for doing it

src/Components/Server/src/CircuitOptions.cs
rynowak: Is the idea that a developer who wants to audit/standardize their use of crypto will be able to replace this with whatever they want?

src/Components/Server/src/Circuits/CircuitHost.cs
```diff
@@ -78,7 +80,7 @@ public static void SetCurrentCircuitHost(CircuitHost circuitHost)
             Renderer.UnhandledSynchronizationException += SynchronizationContext_UnhandledException;
         }
 
-        public string CircuitId { get; } = Guid.NewGuid().ToString();
+        public string CircuitId { get; }
```
rynowak: Is the circuitId considered a secret? IE can we log it? @blowdart 

We will **need** to have identifiers that we can use for correlation and troubleshooting.

src/Components/Server/src/Circuits/CircuitIdFactory.cs
```diff
@@ -0,0 +1,63 @@
...
+using System.Security.Cryptography;
+using Microsoft.AspNetCore.DataProtection;
+using Microsoft.AspNetCore.WebUtilities;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.AspNetCore.Components.Server.Circuits
+{
+    // This is a singleton instance
+    // Generates strong criptographic ids for circuits that are protected with authenticated encryption.
+    internal class CircuitIdFactory
```
rynowak: <3

src/Components/Server/src/CircuitOptions.cs
rynowak: What are the expectations if this is set to null?

src/Components/Server/test/Circuits/CircuitIdFactoryTest.cs
```diff
@@ -0,0 +1,100 @@
...
+using Xunit;
+
+namespace Microsoft.AspNetCore.Components.Server.Tests.Circuits
+{
+    public class CircuitIdFactoryTest
+    {
+        [Fact]
+        public void CreateCircuitId_Generates_NewRandomId()
+        {
+            CircuitIdFactory factory = CreateTestFactory();
```
rynowak: var

src/Components/Server/test/Circuits/CircuitIdFactoryTest.cs
```diff
@@ -0,0 +1,100 @@
...
+
+using System;
+using System.Linq;
+using Microsoft.AspNetCore.Components.Server.Circuits;
+using Microsoft.AspNetCore.DataProtection;
+using Microsoft.AspNetCore.WebUtilities;
+using Microsoft.Extensions.Options;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Components.Server.Tests.Circuits
```
rynowak: remove `.Tests`

It might be a good idea to just set the base namespace of the project so we don't keep fighting it.

src/Components/Server/test/Circuits/CircuitIdFactoryTest.cs
```diff
@@ -0,0 +1,100 @@
...
+        {
+            CircuitIdFactory factory = CreateTestFactory();
+
+            // Act
+            var id = factory.CreateCircuitId();
+
+            // Assert
+            Assert.NotNull(id);
+            // This is the magic data protection header that validates its protected
+            Assert.StartsWith("CfDJ", id);
```
rynowak: Do we really want this hardcoded here? This seems bad. Do we have other options?

src/Components/Server/test/Circuits/CircuitIdFactoryTest.cs
```diff
@@ -0,0 +1,100 @@
...
+            // Act
+            var isValid = factory.ValidateCircuitId(tamperedId);
+
+            // Assert
+            Assert.False(isValid, "Accepted a tampered payload");
+        }
+
+        private static CircuitIdFactory CreateTestFactory()
+        {
+            // Arrange
```
rynowak: ???

src/Components/Server/test/Circuits/CircuitRegistryTest.cs
```diff
@@ -34,8 +35,13 @@ public void Register_AddsCircuit()
         public async Task ConnectAsync_TransfersClientOnActiveCircuit()
         {
             // Arrange
-            var registry = CreateRegistry();
-            var circuitHost = TestCircuitHost.Create();
+            var circuitIdFactory = new CircuitIdFactory(Options.Create<CircuitOptions>(new CircuitOptions
+            {
+                CircuitIdProtector = new EphemeralDataProtectionProvider().CreateProtector("Test")
+            }));
```
rynowak: Should we extract this into a reusable method? like `TestCircuitIdFactory.Create()`? I could see this used in a bunch of places and it would pay to clean it up preemptively.

src/Components/Server/test/Circuits/CircuitIdFactoryTest.cs
javiercn: Its quite the opposite. That is a good test that things are being data protected. You'll see that prefix on every data protected payload.

src/Components/Server/test/Circuits/CircuitIdFactoryTest.cs
javiercn: Refactor issues

rynowak:
Code changes here look really solid.

**Question:** now that we're using data-protection, will users be impacted by key-rotation? Like, what's the MAX lifetime of a circuitId?

rynowak (approved):

javiercn:
> now that we're using data-protection, will users be impacted by key-rotation?

Data protection keys last for 3 months (90 days) and the key starts to rotate early. (There is a time where the old key is still valid but the payloads are signed with the new, I believe).

So (I'm saying this top of my head) you won't have any problem whatsoever if your circuit lasts less than 45 days up.

src/Components/Server/src/Circuits/CircuitHost.cs
javiercn: I don't necessarily think of them as a secret, but I also don't feel super confortable logging them.
I say we can change this to a struct and have two fields, one with the actual payload and one with a string version of the random bytes for logging purposes.

The random bytes are ok to use as the identifier for display/logging purposes as they are data protected and someone would need access to the dp key to actually generate the over the wire format.

src/Components/Server/src/CircuitOptions.cs
Tratcher: Why not provide it via DI?

src/Components/Server/src/Circuits/CircuitIdFactory.cs
```diff
@@ -0,0 +1,63 @@
...
+            {
+                _generator.GetBytes(buffer);
+                var payload = _protector.Protect(buffer);
+
+                return Base64UrlTextEncoder.Encode(payload);
+            }
+            finally
+            {
+                // Its good practice to clear secrets from memory as soon as we are done with them.
+                Array.Clear(buffer, 0, 32);
```
Tratcher: ^-^ There's nothing secret about this value. It's protected to prevent tampering.

src/Components/Server/src/Circuits/CircuitHost.cs
blowdart: No, it's not, and, because its sent over a data protected payload even if you knew it, you'd also have to know the signing keys in order to abuse it and hop circuits.

Tratcher (commented):
The design is a little strange in that you never use the decrypted value. The encrypted value is bulkier.

src/Components/Server/src/Circuits/CircuitIdFactory.cs
javiercn: Fair enough, I just get on paranoid mode when I write this type of code. You are right in that we don't need to necessarily clear it.

src/Components/Server/src/CircuitOptions.cs
javiercn: Just follows the pattern used in other places across auth stuff. I believe cookies also do it this way.

javiercn: > What are the expectations if this is set to null?

The expectation is that you'll get a runtime exception. This is an advanced setting, so if you are changing this, the expectation is that you know what you are doing, so I'm not concerned about it.

Tratcher: That's a legacy pattern leftover from Katana. Use DI.

PR #10690, opened 2019-05-31
Exposes the underlying HttpConnectionDispatcherOptions in MapBlazorHub

javiercn:
These overloads are missing and are necessary to configure things like the message size.

https://github.com/aspnet/AspNetCore/issues/9570


src/Components/Server/src/Builder/ComponentEndpointRouteBuilderExtensions.cs
```diff
@@ -28,6 +29,28 @@ public static ComponentEndpointConventionBuilder MapBlazorHub(this IEndpointRout
             return new ComponentEndpointConventionBuilder(endpoints.MapHub<ComponentHub>(ComponentHub.DefaultPath));
         }
 
+        /// <summary>
+        /// Maps the SignalR <see cref="ComponentHub"/> to the path <see cref="ComponentHub.DefaultPath"/>.
```
pranavkm: `ComponentHub` is internal no? We should avoid including it here since users wouldn't know what it is.

> Maps the Blazor `<see cref="Hub" />` to the default path.

pranavkm: Same update elsewhere

pranavkm (approved):

src/Components/Server/ref/Microsoft.AspNetCore.Components.Server.netcoreapp3.0.cs
```diff
@@ -10,10 +10,15 @@ public static partial class ComponentEndpointConventionBuilderExtensions
...
+        public static Microsoft.AspNetCore.Components.Server.ComponentEndpointConventionBuilder MapBlazorHub<TComponent>(this Microsoft.AspNetCore.Routing.IEndpointRouteBuilder endpoints, string selector, System.Action<Microsoft.AspNetCore.Http.Connections.HttpConnectionDispatcherOptions> configureOptions) where TComponent : Microsoft.AspNetCore.Components.IComponent { throw null; }
         public static Microsoft.AspNetCore.Components.Server.ComponentEndpointConventionBuilder MapBlazorHub<TComponent>(this Microsoft.AspNetCore.Routing.IEndpointRouteBuilder endpoints, string selector, string path) where TComponent : Microsoft.AspNetCore.Components.IComponent { throw null; }
+        public static Microsoft.AspNetCore.Components.Server.ComponentEndpointConventionBuilder MapBlazorHub<TComponent>(this Microsoft.AspNetCore.Routing.IEndpointRouteBuilder endpoints, string selector, string path, System.Action<Microsoft.AspNetCore.Http.Connections.HttpConnectionDispatcherOptions> configureOptions) where TComponent : Microsoft.AspNetCore.Components.IComponent { throw null; }
```
rynowak: I'm thinking we should reduce the set of overloads here by using optional parameters. Imagine we add another thing, now there's like 16 overloads. I think this is still a good change for now, but we should revisit as part of API review. 

Another approach would be to expose this as another call on the builder. I guess I expected us to get this for free from `MapHub`. @anurse @BrennanConroy

src/Components/Server/src/Builder/ComponentEndpointRouteBuilderExtensions.cs
```diff
@@ -29,16 +31,37 @@ public static ComponentEndpointConventionBuilder MapBlazorHub(this IEndpointRout
...
+            {
+                throw new ArgumentNullException(nameof(endpoints));
+            }
+
+            if (configureOptions == null)
+            {
+                throw new ArgumentNullException(nameof(configureOptions));
+            }
+
+            return new ComponentEndpointConventionBuilder(endpoints.MapHub<ComponentHub>(ComponentHub.DefaultPath, configureOptions));
```
rynowak: Is it possible to have this call through to one of the longer overloads? I always get nervous when there isn't a single implementation of these builder methods. We have a history of missing scenarios in tests.

rynowak: Similar comment about L286 - could you do a pass on these and try to reduce the duplication?

rynowak (changes requested):
I'd really like to see at least one or two unit tests for this area. They don't have to be E2E - just verify that things are wired up

src/Components/Server/src/Builder/ComponentEndpointRouteBuilderExtensions.cs
javiercn: Updated, there are only now two places where is terminal

javiercn:
Updated

PR #10695, opened 2019-05-31
Fix helix test

jkotalik:
Currently helix is failing on the test TargedDifferenceSharedFramework_FailedToFindNativeDependenciesErrorInResponse. 

The build that contained my changes on helix was https://mc.dot.net/#/user/aspnetcore/pr~2Faspnet~2Faspnetcore/ci/20190530.15. This seemed to pass with my changes just fine, and it passed locally. However, this test was incredibly flaky (I called shutdown on the server and then sent an HttpRequest to the server after shutdown). 

The thing that confuses me the most is that the test is failing hard on other builds: https://dev.azure.com/dnceng/public/_build/results?buildId=206419&view=ms.vss-test-web.build-test-results-tab. Like after 3 runs the test is failing consistently. 

I'm going to investigate if the build that contained my changes _actually_ contained my changes.


jkotalik:
NVRM the helix build I linked wasn't associated with my build at all, it was another PR I made around the same time. Helix just didn't trigger at all 😢

BrennanConroy (approved):
Lets make sure Helix runs :trollface:

PR #10712, opened 2019-05-31
Add unmatched attributes support to forms

rynowak:
Part of https://github.com/aspnet/AspNetCore/issues/5071

This change makes it possible to use attributes that aren't directly supported by our components. 

Ex:
```razor
<InputText @bind="person.Name" placeholder="Enter your name" />
```

`placeholder` isn't something we define or handle inside `InputText` - it just gets passed through to the underlying input element.

This support is limited to components implemented in C# for now, and we'll be doing a language feature to light it up for components written in Razor in preview 7 (see #5071)

I'm logging a separate issue to follow up on two things I didn't address in this PR because I think they need a discussion. 
- Should we remove `Id` and `Class` from these components now that we support pass through
- Should we rewrite `NavLink` to use this feature

### DIFF
--- a/src/Components/Components/src/Forms/InputComponents/InputCheckbox.cs
+++ b/src/Components/Components/src/Forms/InputComponents/InputCheckbox.cs
@@ -24,11 +24,12 @@ namespace Microsoft.AspNetCore.Components.Forms
         protected override void BuildRenderTree(RenderTreeBuilder builder)
         {
             builder.OpenElement(0, "input");
-            builder.AddAttribute(1, "type", "checkbox");
-            builder.AddAttribute(2, "id", Id);
-            builder.AddAttribute(3, "class", CssClass);
-            builder.AddAttribute(4, "checked", BindMethods.GetValue(CurrentValue));
-            builder.AddAttribute(5, "onchange", BindMethods.SetValueHandler(__value => CurrentValue = __value, CurrentValue));
+            builder.AddMultipleAttributes(1, AdditionalAttributes);
+            builder.AddAttribute(2, "type", "checkbox");
+            builder.AddAttribute(3, "id", Id);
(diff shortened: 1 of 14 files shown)


javiercn (approved):
Looks good!

SteveSandersonMS:
> I'm logging a separate issue to follow up on two things I didn't address in this PR because I think they need a discussion.

I didn't see the other issue but my opinions are:

> Should we remove Id and Class from these components now that we support pass through

Yes

> Should we rewrite NavLink to use this feature

Yes

src/Components/Components/src/Forms/InputComponents/InputCheckbox.cs
```diff
@@ -24,11 +24,12 @@ public class InputCheckbox : InputBase<bool>
...
-            builder.AddAttribute(2, "id", Id);
-            builder.AddAttribute(3, "class", CssClass);
-            builder.AddAttribute(4, "checked", BindMethods.GetValue(CurrentValue));
-            builder.AddAttribute(5, "onchange", BindMethods.SetValueHandler(__value => CurrentValue = __value, CurrentValue));
+            builder.AddMultipleAttributes(1, AdditionalAttributes);
+            builder.AddAttribute(2, "type", "checkbox");
+            builder.AddAttribute(3, "id", Id);
+            builder.AddAttribute(4, "class", CssClass);
+            builder.AddAttribute(5, "checked", BindMethods.GetValue(CurrentValue));
+            builder.AddAttribute(6, "onchange", BindMethods.SetValueHandler(__value => CurrentValue = __value, CurrentValue));
```
SteveSandersonMS: At some point, we should think about somehow combining any explicitly-provided `onchange` with the one that `bind` is using (e.g., calling theirs from ours).

SteveSandersonMS:
This is such a big improvement!

PR #10716, opened 2019-05-31
Application restarts on changes to Razor files

### LINKED ISSUE #9644: Ensure razor and cshtml files are watched for changes by dotnet-watch and VS
Related to #9014

pranavkm:
* .cshtml and .razor files cause dotnet-watch to reload by default
* .cshtml files do not cause VS or dotnet-watch to reload when using RuntimeCompilation

Fixes https://github.com/aspnet/AspNetCore/issues/9644

### DIFF
--- a/src/Mvc/Mvc.Razor.RuntimeCompilation/src/build/netcoreapp3.0/Microsoft.AspNetCore.Mvc.Razor.RuntimeCompilation.targets
+++ b/src/Mvc/Mvc.Razor.RuntimeCompilation/src/build/netcoreapp3.0/Microsoft.AspNetCore.Mvc.Razor.RuntimeCompilation.targets
@@ -8,5 +8,7 @@
 
     <!-- Prevent VS from restarting the application when cshtml files change since this package handles it -->
     <RazorUpToDateReloadFileTypes>$(RazorUpToDateReloadFileTypes.Replace('.cshtml', ''))</RazorUpToDateReloadFileTypes>
+
+    <AddCshtmlFilesToDotNetWatchList>false</AddCshtmlFilesToDotNetWatchList>
   </PropertyGroup>
 </Project>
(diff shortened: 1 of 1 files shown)


src/Tools/dotnet-watch/src/assets/DotNetWatch.targets
```diff
@@ -55,6 +55,11 @@ Returns: @(Watch)
       <Watch Include="%(Compile.FullPath)" Condition="'%(Compile.Watch)' != 'false'" />
       <Watch Include="%(EmbeddedResource.FullPath)" Condition="'%(EmbeddedResource.Watch)' != 'false'"/>
       <Watch Include="$(MSBuildProjectFullPath)" />
+      <Watch Include="%(Content.FullPath)"
+        Condition="'%(Content.Extension)' == '.razor' AND '%(Content.Watch)' != 'false'" />
+      <Watch Include="%(Content.FullPath)"
+        Condition="'%(Content.Extension)' == '.cshtml' AND '$(RazorLangVersion)' >= '3.0'  AND '%(Content.Watch)' != 'false'" />
```
rynowak: Why 3.0?

natemcmaster (changes requested):
I don't like that we're special casing Razor from inside dotnet-watch. The layering feels wrong, especially because this requires also knowing about the version of Razor used. I would prefer to invert the control here and have the Razor SDK instead include this line:
```xml
<ItemGroup>
  <Watch Include="*.razor" />
</ItemGroup>
```

pranavkm:
> because this requires also knowing about the version of Razor used.

We could remove the version check, I realize that's not necessary. Would that remove the need to put this in the RazorSDK?

natemcmaster:
It still seems wrong to me that this would put app-model specific information into dotnet-watch. Although dotnet-watch is (currently) owned by our team, it was designed to work in all .NET project types -- test projects, WPF/WinForms, class libraries, console apps, ASP.NE Core, etc. I think project types that want to alter the behavior should use the `<Watch>` item group on their own to suit their needs.

pranavkm:
Fair enough. I'll get the dotnet-watch specific changes later once I've had a chance to test it.

SteveSandersonMS:
@pranavkm I don't know what these changes to the NPM project structure are about. Could you explain why there should be this new "utils" package and also "jsutils"?

pranavkm:
@SteveSandersonMS wrong branch. Fixed it up.

rynowak (approved):

PR #10727, opened 2019-06-01
Make JSON case-insensitive

### LINKED ISSUE #10723: Change MVC's use of System.Text.Json to case-insensitive
**Edit: if you arrived here via twitter then Hi! Thanks for coming. I wrote the initial bug report last night pretty late so some adding some information and context.**

## History and context

MVC gets its existing JSON support through Newtonsoft.Json (thanks @JamesNK). Newtonsoft.Json provides options to configure canonicalization of names and case-sensitivity. Newtonsoft.Json's default settings are to leave your property names as-is and to compare property names case-insensitively.

MVC however configures Newtonsoft.Json to canonicalize property names as camelCase. The effect of this is:
- JSON output produces camelCase property names
- JSON input matches property names using case-insensitive comparisons, supporting camelCase or PascalCase.

MVC's choice of a different default than what Newtonsoft.Json users makes sense to us because MVC is for the web (web sites, APIs). If you're using JSON on the web, you're doing to so either:
- interop with javascript for a web site
- interop with *any* other language for an API

In either of these cases, choosing camelCase is the least surprising default. 

## Enter System.Text.Json

We're adding a new serializer to dotnet/CoreFx - which is a good opportunity to question any and all assumptions about what we're doing. Using a new serializer as the default will mean breaking changes in behavior, so if there are things we want to change in this are it's our one chance.

I originally said we should use camelCase and be case-sensitive because it's the more correct default. JSON is fundamentally a case-sensitive format so it seems like the best default.

## The problem

The project is that this breaks webapi client - which uses Newtonsoft.Json's default settings and preserves the case of your C# properties. In fact, for the immediate future, any .NET code that calls your APIs probably is using Newtonsoft.Json and probably is using its default settings (PascalCase).

So they send:
```js
{
    "FirstName": "James",
    "LastName": "Newton-King",
}
```

But we would only accept:
```js
{
    "firstName": "James",
    "lastName": "Newton-King",
}
```

This gets worse because we also want to tolerate/ignore unmatched properties by default. The result is that things don't work and you don't know why.

## The solution

The solution is opt-in to case-insensitivity. This gives us the same behavior as we had with Newtonsoft.Json. 

I'm personally convinced that this is the right default because it's going to *just work* for most users, and it's also our current behavior. Again, this is a discussion about our default, so it's not about what's the most correct, it's about what's going to work well for most users. 

If I were writing an API, I would consider changing this setting, which is easy to do.
```C#
services.AddControllers().AddJsonOptions(options => options.JsonSerializerOptions.PropertyNameCaseInsensitive = false);
```

/cc @anurse @BrennanConroy @pranavkm

rynowak:
Fixes: #10723

The rationale for this change is that existing .NET client code for the
most part uses JSON.NET with its default settings (preserve property
casing). This includes the WebAPI client - which we're encouraging
everyone to use. It's not really reasonable for us to break everyone
using webapi client.

### DIFF
--- a/src/Mvc/Mvc.Core/ref/Microsoft.AspNetCore.Mvc.Core.netcoreapp3.0.cs
+++ b/src/Mvc/Mvc.Core/ref/Microsoft.AspNetCore.Mvc.Core.netcoreapp3.0.cs
@@ -814,6 +814,11 @@ namespace Microsoft.AspNetCore.Mvc
     public partial interface IRequestSizePolicy : Microsoft.AspNetCore.Mvc.Filters.IFilterMetadata
     {
     }
+    public partial class JsonOptions
+    {
+        public JsonOptions() { }
+        public System.Text.Json.Serialization.JsonSerializerOptions JsonSerializerOptions { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
+    }
     public partial class JsonResult : Microsoft.AspNetCore.Mvc.ActionResult, Microsoft.AspNetCore.Mvc.IActionResult, Microsoft.AspNetCore.Mvc.Infrastructure.IStatusCodeActionResult
     {
         public JsonResult(object value) { }
@@ -881,7 +886,6 @@ namespace Microsoft.AspNetCore.Mvc
         public bool RequireHttpsPermanent { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public bool RespectBrowserAcceptHeader { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public bool ReturnHttpNotAcceptable { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
-        public System.Text.Json.Serialization.JsonSerializerOptions SerializerOptions { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
         public int? SslPort { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public bool SuppressAsyncSuffixInActionNames { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
         public bool SuppressImplicitRequiredAttributeForNonNullableReferenceTypes { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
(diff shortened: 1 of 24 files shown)


src/Mvc/Mvc.Core/src/MvcOptions.cs
```diff
@@ -337,16 +337,21 @@ public int MaxModelBindingRecursionDepth
         }
 
         /// <summary>
-        /// Gets the <see cref="JsonSerializerOptions"/> used by <see cref="SystemTextJsonInputFormatter"/> and
+        /// Gets the <see cref="System.Text.Json.Serialization.JsonSerializerOptions"/> used by <see cref="SystemTextJsonInputFormatter"/> and
         /// <see cref="SystemTextJsonOutputFormatter"/>.
         /// </summary>
-        public JsonSerializerOptions SerializerOptions { get; } = new JsonSerializerOptions
+        public JsonSerializerOptions JsonSerializerOptions { get; } = new JsonSerializerOptions
```
rynowak: @pranavkm - this seemed right to change. It felt right at the time to make it `SerializerSettings` but I think this is better. when a user looks for this they are thinking "JSON" not "Serializer".

Thoguhts?

src/Mvc/Mvc.Core/ref/Microsoft.AspNetCore.Mvc.Core.netcoreapp3.0.cs
```diff
@@ -814,6 +814,11 @@ public partial interface IRequestFormLimitsPolicy : Microsoft.AspNetCore.Mvc.Fil
     public partial interface IRequestSizePolicy : Microsoft.AspNetCore.Mvc.Filters.IFilterMetadata
     {
     }
+    public partial class JsonOptions
+    {
+        public JsonOptions() { }
+        public System.Text.Json.Serialization.JsonSerializerOptions JsonSerializerOptions { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
```
rynowak: @pranavkm thoughts? Folwer and I like it because we get to write `.AddControllers().AddJsonOptions(...)` which feels a little better than putting it on mvc options

pranavkm: Seems fine. It's closer to how you configure other formatters \ `AddXmlOptions`

src/Mvc/Mvc.Core/src/JsonOptions.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            // Limit the object graph we'll consume to a fixed depth. This prevents stackoverflow exceptions
+            // from deserialization errors that might occur from deeply nested objects.
+            // This value is the same for model binding and Json.Net's serialization.
+            MaxDepth = MvcOptions.DefaultMaxModelBindingRecursionDepth,
+
+            // We're using case-insensitive because there's a TON of code that there that does uses JSON.NET's default
+            // settings (preserve case) - including the WebAPIClient. This worked when we were using JSON.NET + camel casing
+            // because JSON.NET is case-insensitive by default.
+            PropertyNameCaseInsensitive = true,
```
pranavkm: ezgif-2-c0511db22947

pranavkm (approved):

davidfowl (approved):

src/Mvc/Mvc.Core/src/JsonOptions.cs
rynowak: nah. You and @davidfowl are right but for the wrong reason.

rynowak:
>  Azure Pipelines successfully started running 1 pipeline(s).

AMAZING THANK YOU

JamesNK (approved):

rynowak:
yeth

PR #10728, opened 2019-06-01
Clean up some pipes things

davidfowl:
- Use the Kestrel memory pool
- Fix bad casts

### DIFF
--- a/src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.FeatureCollection.cs
+++ b/src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.FeatureCollection.cs
@@ -109,11 +109,11 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                 if (!ReferenceEquals(_requestStreamInternal, RequestBody))
                 {
                     _requestStreamInternal = RequestBody;
-                    RequestBodyPipeReader = PipeReader.Create(RequestBody);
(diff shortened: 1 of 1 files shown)


src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.FeatureCollection.cs
```diff
@@ -109,11 +109,11 @@ PipeReader IRequestBodyPipeFeature.Reader
                 if (!ReferenceEquals(_requestStreamInternal, RequestBody))
                 {
                     _requestStreamInternal = RequestBody;
-                    RequestBodyPipeReader = PipeReader.Create(RequestBody);
+                    RequestBodyPipeReader = PipeReader.Create(RequestBody, new StreamPipeReaderOptions(_context.MemoryPool, _context.MemoryPool.GetMinimumSegmentSize(), _context.MemoryPool.GetMinimumAllocSize()));
 
                     OnCompleted((self) =>
                     {
-                        ((PipeWriter)self).Complete();
+                        ((PipeReader)self).Complete();
```
rynowak: ouch

rynowak: Why isn't this defined like `OnCompleted<T>(Func<T, Task>, T)`?

jkotalik (approved):
Oops.

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.FeatureCollection.cs
davidfowl: We messed up

rynowak: I guess that's a good reason

PR #10730, opened 2019-06-01
Add support for TypeConverter

### LINKED ISSUE #8385: EventCallback not working as expected
### Describe the bug
I'm having trouble with the `EventCallback` in new RC. I have a `SelectEdit` component with `SelectedValue` and `SelectedValueChanged` parameters. I want for my component to support two way binding (`bind-SelectedValue`) and event handlers (`SelectedValueChanged=@"OnValueChanged"`). This has worked with `Action`, but now in the new RazorComponents when I convert to EventCallback I get the errors:

> CS1503 Argument 6: cannot convert from 'Microsoft.AspNetCore.Components.EventCallback' to 'Microsoft.AspNetCore.Components.EventCallback<string>'

> Error	CS1660	Cannot convert lambda expression to type 'EventCallback<string>' because it is not a delegate type

**SelectEdit.cshtml** signature
```
[Parameter] protected TValue SelectedValue { get; set; }
[Parameter] protected EventCallback<TValue> SelectedValueChanged { get; set; }
```

### Expected behavior

Users should be able to use both ways of handling the component values, and not just the **bind-** attributes.  Custom actions should also be able to use if necessary.

```
<SelectEdit SelectedValue="@someValue" SelectedValueChanged="@((v) => someValue=v)">
```

or

```
<SelectEdit SelectedValue="@someValue" SelectedValueChanged="@OnSomeValueChanged">
@functions{
    void OnSomeValueChanged(string v)
    {
        someValue = v;
        Console.WriteLine(someValue);
        // do some other actions
    }
}
```

### Additional context

I have another component that is also using EventCallback but this one is not generic. On that components I can use events handler without any error. eg.

**SimpleButton.cshtml**
```
[Parameter] protected EventCallback Clicked { get; set; }
```

**Usage**
```
<SimpleButton Clicked="@RaiseBasicCounter">Count</SimpleButton>
@functions{
    int basicEventCounter = 0;

    void RaiseBasicCounter()
    {
        basicEventCounter++;
    }
}
```

### LINKED ISSUE #9632: typeparam bind error
Hi, in Blazor 0.10 such code stopped working

```
@typeparam TItem

<input bind="@line" />

@functions{
    [Parameter]
    TItem line { get; set; }
}
```

This is a build error of version 0.10 or is the ability to typeparam bind removed ?, is there an alternative solution to the problem?

### LINKED ISSUE #10077: EventCallback<T> Error cannot convert from 'method group' to 'EventCallback'
### Describe the bug
Using the following two components, the compiler emits an error in the Parent component  in the OnSelect event parameter.

Error cannot convert from 'method group' to 'EventCallback'

```
<Child Data="@Data" OnSelect="@OnSelect">
</Child>
```

Is this an issue or I am doing this wrong ?

### To Reproduce

```
@*Parent.razor*@

@page "/Parent"

<div>
    <Child Data="@Data" OnSelect="@OnSelect">
    </Child>
</div>

@functions {
    List<Item> Data = new List<Item>();

    private void OnSelect(Item item)
    {        
    }
}
```
```
@*Child.razor*@

@typeparam TData

<div>
    @foreach (var item in Data)
    {
        <button onclick="@(() => OnSelect.InvokeAsync(item))"></button>
    }
</div>

@functions {
    [Parameter]
    public IEnumerable<TData> Data { get; set; }

    [Parameter]
    private EventCallback<TData> OnSelect { get; set; }
}
```
```
// Item.cs
public class Item
{
     public Item()
     {

     }
}
```

rynowak:
Fixes: #9632
Fixes: #8385
Fixes: #10077

This fix adds support for type converters as well as a few other minor
things we were missing from binding. The key thing about supporting
conversions is that we new can support arbitrary types with `@bind`.
This means you can use it with generics, which is something many users
have tried.

Along with type converters we get Guid and TimeSpan from the BCL. The
BCL also includes converters for types we're less interested in like
`short`.

Note: this change is targeting preview 7.


src/Components/Components/src/EventCallbackFactoryBinderExtensions.cs
```diff
@@ -2,13 +2,27 @@
...
+    //
+    // NOTE: for number parsing, the HTML5 spec dictates that <input type="number"> the DOM will represent
+    // number values as floating point numbers using `.` as the period separator. This is NOT culture senstive.
+    // Put another way, the user might see `,` as their decimal separator, but the value available in events
+    // to JS code is always simpilar to what .NET parses with InvariantCulture.
+    //
+    // See: https://www.w3.org/TR/html5/sec-forms.html#number-state-typenumber
+    // See: https://www.w3.org/TR/html5/infrastructure.html#valid-floating-point-number
+    //
+    // For now we're not necessarily handling this correctly since we parse the same way for number and text.
```
rynowak: This is a comment I had written about the current state of globalization and our conversions before I impleemented type converter support. I want to leave it in because it's good info to keep close to the code.

src/Components/Components/src/EventCallbackFactoryBinderExtensions.cs
```diff
@@ -53,7 +67,7 @@ private static bool ConvertToIntCore(object obj, out int value)
                 return false;
             }
 
-            if (!int.TryParse(text, out var converted))
+            if (!int.TryParse(text, NumberStyles.Any, CultureInfo.CurrentCulture, out var converted))
```
rynowak: I think it's a nice idea to update these in general to use the longest overload so that it's self-documenting. We *do* care about these details.

src/Components/Components/src/EventCallbackFactoryBinderExtensions.cs
```diff
@@ -246,28 +260,83 @@ private static bool ConvertToNullableDecimalCore(object obj, out decimal? value)
             return true;
         }
 
-        private static class EnumConverter<T> where T : struct, Enum
+        private static BindConverter<DateTime> ConvertToDateTime = ConvertToDateTimeCore;
+        private static BindConverter<DateTime?> ConvertToNullableDateTime = ConvertToNullableDateTimeCore;
```
rynowak: Added nullable datetime support (without format strings). We have a separate issue tracking adding format strings broadly.

src/Components/Components/src/EventCallbackFactoryBinderExtensions.cs
```diff
@@ -246,28 +260,83 @@ private static bool ConvertToNullableDecimalCore(object obj, out decimal? value)
...
+            {
+                value = default;
+                return false;
+            }
+
+            value = converted;
+            return true;
+        }
+
+        private static bool ConvertToNullableEnum<T>(object obj, out Nullable<T> value) where T : struct, Enum
```
rynowak: Added support for nullable enums without boxing. Type converters can do this, but by special casing we can avoid the boxing.

src/Components/Components/src/EventCallbackFactoryBinderExtensions.cs
```diff
@@ -489,15 +557,6 @@ public static bool ConvertCore(object obj, out T value)
...
         {
-            Func<object, decimal?> converter = (obj) =>
-            {
-                if (decimal.TryParse((string)obj, out var value))
-                {
-                    return value;
-                }
-
-                return null;
-            };
```
rynowak: This was dead code

src/Components/Components/src/EventCallbackFactoryBinderExtensions.cs
```diff
@@ -592,26 +664,9 @@ public static bool ConvertCore(object obj, out T value)
             this EventCallbackFactory factory,
             object receiver,
             Action<T> setter,
-            T existingValue) where T : struct, Enum
+            T existingValue)
         {
-            return CreateBinderCore<T>(factory, receiver, setter, EnumConverter<T>.Convert);
-        }
-
-        private static DateTime ConvertDateTime(object obj, string format)
```
rynowak: Made this a local function because it's used in one place.

src/Components/Components/src/EventCallbackFactoryBinderExtensions.cs
```diff
@@ -642,5 +697,139 @@ private static DateTime ConvertDateTime(object obj, string format)
...
+
+            private static MethodInfo _convertToEnum;
+            private static MethodInfo _convertToNullableEnum;
+
+            public static BindConverter<T> Get<T>()
+            {
+                if (!_cache.TryGetValue(typeof(T), out var converter))
+                {
+                    // We need to replicate all of the primitive cases that we handle here so that they will behave the same way.
+                    // The result will be cached.
```
rynowak: I'm not super happy about this detail, but I think we mostly have covered the set of things we want to special case already.

rynowak: We only really want to special case something if we need to behave in a special way:
- supporting a format string, which type converters don't support
- supporting something without a type converter (nothing comes to mind, we could do Guid if we were really motivated)
- avoiding boxing (maybe DateTimeOfffset?)

src/Components/Components/src/EventCallbackFactoryBinderExtensions.cs
PinpointTownes: It's worth noting this also results in a subtle behavior change, as the `int.TryParse` overload that doesn't take an enum uses `NumberStyles.Integer` and not `NumberStyles.Any`. Concretely, this means that things like parenthesis, currency symbols or exponents will now be considered valid.

The same is true for the other types, but they often use different `NumberStyles` flags (e.g `decimal` uses `NumberStyles.Number`).

rynowak: Ah that's a really good point. I was trying to find that info and I found it wrongly it looks like. I don't intend to change the behavior of this without more discussion so I'll update it. 👍

rynowak:
> In EventCallbackFactoryBinderExtensions class it is parsing numbers using the CurrentCulture, and that can give different results depending on the machine where the app is ran. Shouldn't it use InvariantCulture?

Sadly, it's not that simple. See: https://github.com/aspnet/AspNetCore/issues/9386 which tracks all of the work in this area. InvariantCulture will result in the correct behavior for an `<input type="number">` but not for other field types like <input type="text">`

This PR doesn't cover fixing that issue, and as far as I know just makes the current behavior more clearly explicit. We definitely plan to address this before release.

src/Components/test/testassets/BasicTestApp/BindGenericComponent.razor
```diff
@@ -0,0 +1,9 @@
+@typeparam TValue
+
+<input id="@Id" type="text" value="@Value" @onchange="@(EventCallback.Factory.CreateBinder<TValue>(this, _ => ValueChanged.InvokeAsync(_), Value))" />
```
SteveSandersonMS: The manual use of `CreateBinder` here looks pretty hardcore. Is your expectation that customers will do this? I'm not sure I'd feel confident writing this code and being sure it was right without copy/pasting some example.

Out of interest, since I can't run this code right now, does the following simplified syntax (relying on generic type inference and methodgroup-to-delegate conversion) produce the same result?

```razor
<input id="@Id" type="text" value="@Value" @onchange="@(EventCallback.Factory.CreateBinder(this, ValueChanged.InvokeAsync, Value))" />
```

src/Components/Components/src/EventCallbackFactoryBinderExtensions.cs
```diff
@@ -642,5 +697,139 @@ private static DateTime ConvertDateTime(object obj, string format)
...
+                {
+                    var text = (string)obj;
+                    if (string.IsNullOrEmpty(text))
+                    {
+                        value = default;
+                        return true;
+                    }
+
+                    // We intentionally close-over the TypeConverter to cache it. The TypeDescriptor infrastructure is slow.
+                    var converted = typeConverter.ConvertFromString(context: null, CultureInfo.InvariantCulture, text);
```
SteveSandersonMS: Why use `CultureInfo.InvariantCulture` here? The other parsing cases use `CultureInfo.CurrentCulture` - shouldn't the typeconverter be allowed to be culture-sensitive too?

src/Components/Components/src/EventCallbackFactoryBinderExtensions.cs
```diff
@@ -577,6 +632,23 @@ public static bool ConvertCore(object obj, out T value)
...
             return factory.Create<UIChangeEventArgs>(receiver, callback);
+
+            static DateTime ConvertDateTime(object obj, string format)
+            {
+                var text = (string)obj;
+                if (string.IsNullOrEmpty(text))
+                {
+                    return default;
+                }
+                else if (format != null && DateTime.TryParseExact(text, format, CultureInfo.InvariantCulture, DateTimeStyles.RoundtripKind, out var value))
```
SteveSandersonMS: Why `InvariantCulture` for dates but `CurrentCulture` for (e.g.) numbers?

SteveSandersonMS (approved):
Other than my questions about `InvariantCulture` vs `CurrentCulture`, this looks great!

Approving now in case it's possible to resolve those last details and merge in time for preview 6.

rynowak:
Sadly we're out of time, this isn't going to make it for preview 6.

src/Components/test/testassets/BasicTestApp/BindGenericComponent.razor
rynowak: I'll try this and see. I expect it does.

rynowak: It does not work because all of these methods accept `Action<>`

src/Components/Components/src/EventCallbackFactoryBinderExtensions.cs
rynowak: This is the existing code for dates with formats. I'm not going to message with the existing beahvior because we have another issue for that.

Issue #5826, opened 2018-10-10
Show errors in the log from VerifyUserTokenAsync


LindaLawton:
I am getting a nice warning in my logs when the code is invalid  from [L1712](https://github.com/aspnet/Identity/blob/c7276ce2f76312ddd7fccad6e399da96b9f6fae1/src/Core/UserManager.cs#L1712)

>warn: Microsoft.AspNetCore.Identity.UserManager[9]
>     VerifyUserTokenAsync() failed with purpose: ResetPassword for user 21248592.

However it does not tell me what the error is in the error message is.   I have had to add this in my own code.

>dbug:IdentityServer.Controllers.Account.AccountController[2150]
>     User invite failed for [User: 21248592] [error: Invalid token.]

Which basically means i have two log entries for the same issue.  It would be nice to get a list of the errors from the result into the warning message.

Can i suggest something like this?  I would be happy to submit a pull request if you like

```
if (!result)
            {
               var errors = string.Join(", ", result.Errors.Select(a => a.Description));
                Logger.LogWarning(9, "VerifyUserTokenAsync() failed with purpose: {purpose} for user {userId}. 
 Errors {errors }", purpose, await GetUserIdAsync(user), errors );
            }
```

guylando:
+1 makes it much harder to understand what is the problem because of the lack of logs (debug level logs seem appropriate)

dpr-dev:
Hello, can I create PR for this issue ? 

@HaoK, @blowdart

HaoK:
sure go ahead

HaoK:
Moving to preview 8 as PR is almost there but won't make it in today

HaoK:
Fixed by @dpr-dev in #10739

PR #10739, opened 2019-06-02
Add debugg logging for DataProtectorTokenProvider.ValidationAsync

dpr-dev:
changes [#5826](https://github.com/aspnet/AspNetCore/issues/5826)

### DIFF
--- a/src/Identity/Core/src/DataProtectorTokenProvider.cs
+++ b/src/Identity/Core/src/DataProtectorTokenProvider.cs
@@ -173,8 +173,10 @@ namespace Microsoft.AspNetCore.Identity
                     return stampIsEmpty;
                 }
             }
+            // ReSharper disable once EmptyGeneralCatchClause
             catch
             {
+                // Do not leak exception
                 Logger.UnhandledException();
             }
 
--- a/src/Identity/Core/src/DataProtectorTokenProvider.cs
+++ b/src/Identity/Core/src/DataProtectorTokenProvider.cs
@@ -5,6 +5,7 @@ using System;
 using System.IO;
 using System.Text;
 using System.Threading.Tasks;
+
 using Microsoft.AspNetCore.DataProtection;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
@@ -125,7 +126,7 @@ namespace Microsoft.AspNetCore.Identity
                     var expirationTime = creationTime + Options.TokenLifespan;
                     if (expirationTime < DateTimeOffset.UtcNow)
                     {
-                        Logger.LogWarning(0, "ValidateAsync failed: the expiration time is invalid.");
+                        Logger.InvalidExpirationTime();
                         return false;
                     }
 
@@ -133,21 +134,21 @@ namespace Microsoft.AspNetCore.Identity
                     var actualUserId = await manager.GetUserIdAsync(user);
                     if (userId != actualUserId)
                     {
-                        Logger.LogWarning(1, $"ValidateAsync failed: the expected and actual UserId aren't equal");
+                        Logger.UserIdsNotEquals();
                         return false;
                     }
 
                     var purp = reader.ReadString();
                     if (!string.Equals(purp, purpose))
                     {
-                        Logger.LogWarning(2, $"ValidateAsync failed: the expected and actual purpose aren't equal.");
+                        Logger.PurposeNotEquals();
                         return false;
                     }
 
                     var stamp = reader.ReadString();
                     if (reader.PeekChar() != -1)
                     {
-                        Logger.LogWarning(2, $"ValidateAsync failed: unexpected end of input.");
+                        Logger.UnexpectedEndOfInput();
                         return false;
                     }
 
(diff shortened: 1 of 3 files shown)


src/Identity/Core/src/DataProtectorTokenProvider.cs
```diff
@@ -110,38 +125,58 @@ public virtual async Task<bool> ValidateAsync(string purpose, string token, User
...
                     }
-                    return stamp == "";
+
+                    return stampIsEmpty; 
                 }
             }
-            // ReSharper disable once EmptyGeneralCatchClause
-            catch
-            {
-                // Do not leak exception
```
HaoK: @blowdart how important is it that we don't leak exceptions here?  Given the comment do we wante to err on the side of not logging the exact exception message?

blowdart: Yes, this needs to return. We cannot return the exception message, only that an exception occured.

src/Identity/Core/src/DataProtectorTokenProvider.cs
```diff
@@ -110,38 +125,58 @@ public virtual async Task<bool> ValidateAsync(string purpose, string token, User
                     var expirationTime = creationTime + Options.TokenLifespan;
                     if (expirationTime < DateTimeOffset.UtcNow)
                     {
+                        Logger.LogWarning(0, "ValidateAsync failed: the expiration time is invalid.");
```
HaoK: I know its a pain, but the pattern we are supposed to use for logging going forward is structured logging, see: https://github.com/aspnet/AspNetCore/blob/master/src/Security/Authentication/Core/src/LoggingExtensions.cs

Can you put these new log messages into a new LoggingExtensions class like that?

dpr-dev: Yes of course. thank. I did not know about it

src/Identity/Core/src/DataProtectorTokenProvider.cs
dpr-dev: > Yes, this needs to return. We cannot return the exception message, only that an exception occured.

okay. Should we write a debug message when an exception occurs ? Or not ?

HaoK: No I think we should just log that an exception has occurred without any of the details in the exception, basically go back to the general catch

dpr-dev:
@HaoK, please, review new commits

src/Identity/Core/src/LoggingExtensions.cs
```diff
@@ -0,0 +1,91 @@
...
+        {
+            _invalidExpirationTime = LoggerMessage.Define(
+                eventId: new EventId(0, "InvalidExpirationTime"),
+                logLevel: LogLevel.Information,
+                formatString: "ValidateAsync failed: the expiration time is invalid.");
+
+            _userIdsNotEquals = LoggerMessage.Define(
+                eventId: new EventId(1, "UserIdsNotEquals"),
+                logLevel: LogLevel.Debug,
+                formatString: "ValidateAsync failed: the expected and actual UserId aren't equal");
```
HaoK: nit: "ValidateAsync failed: did not find expected UserId."  Note there's also a period missing from this one

src/Identity/Core/src/LoggingExtensions.cs
```diff
@@ -0,0 +1,91 @@
...
+
+            _purposeNotEquals = LoggerMessage.Define(
+                eventId: new EventId(2, "PurposeNotEquals"),
+                logLevel: LogLevel.Debug,
+                formatString: "ValidateAsync failed: the expected and actual purpose aren't equal.");
```
HaoK: Nit: "ValidateAsync failed: did not find expected purpose."

src/Identity/Core/src/LoggingExtensions.cs
```diff
@@ -0,0 +1,91 @@
...
+
+            _unexpectedEndOfInput = LoggerMessage.Define(
+                eventId: new EventId(3, "UnexpectedEndOfInput"),
+                logLevel: LogLevel.Debug,
+                formatString: "ValidateAsync failed: unexpected end of input.");
+
+            _securityStampNotEquals = LoggerMessage.Define(
+                eventId: new EventId(4, "SecurityStampNotEquals"),
+                logLevel: LogLevel.Debug,
+                formatString: "ValidateAsync failed: the expected and actual stamp aren't equal.");
```
HaoK: Nit: "ValidateAsync failed: did not find expected security stamp."

HaoK:
@blowdart can we log additional info for some of the tokens?  i.e. expected/actual security stamps/user ids/purposes, those would help a lot with debugging...  purpose doesn't really have any PII, security stamp is just a random GUID, user id might be more of an issue.

blowdart:
Purpose isn't a secret so sure, you can log that. The rest is PII.

src/Identity/Core/src/DataProtectorTokenProvider.cs
```diff
@@ -110,38 +126,60 @@ public virtual async Task<bool> ValidateAsync(string purpose, string token, User
...
                     if (userId != actualUserId)
                     {
+                        Logger.UserIdsNotEquals();
                         return false;
                     }
+
                     var purp = reader.ReadString();
                     if (!string.Equals(purp, purpose))
                     {
+                        Logger.PurposeNotEquals();
```
HaoK: Can you pass in purp and purpose to the log and print that out in the log, that should help diagnose when the wrong token is used

src/Identity/Core/src/LoggingExtensions.cs
```diff
@@ -0,0 +1,91 @@
...
+        {
+            _invalidExpirationTime(logger, null);
+        }
+
+        public static void UserIdsNotEquals(this ILogger logger)
+        {
+            _userIdsNotEquals(logger, null);
+        }
+
+        public static void PurposeNotEquals(this ILogger logger)
```
HaoK: Take in expectedPurpose and actualPurpose and display those to help debugging

dpr-dev: Ok, I will do it. Thank you for your suggestions

HaoK:
Almost there looks like you just need to update the ref assembly to account for the new logger property, see https://dev.azure.com/dnceng/public/_build/results?buildId=227250&view=logs&jobId=2fe95fc5-ae3d-58fd-c2b8-46bb81541875

HaoK:
Thanks @dpr-dev !

PR #10740, opened 2019-06-02
Use the defaults that existed in Kestrel before

davidfowl:
Preserving the existing settings but now at the transport level.


davidfowl:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=208641

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/1950

jkotalik (approved):

Tratcher:
@davidfowl you know this missed preview6, right?

davidfowl:
Yep. @anurse told me it was too late to change these values :trollface: . If there's no performance regression, it's fine.

Issue #10318, opened 2019-05-17
OpenIdConnectHandler should also return no-store


huan086:
### Is your feature request related to a problem? Please describe.
The OpenIdConnectHandler currently returns no-cache only. This does not prevent caching by the back-button cache. Thus, after switching to `OpenIdConnectOptions.AuthenticationMethod = OpenIdConnectRedirectBehavior.FormPost;`, it would still be possible for an attacker to retrieve the id_token containing first/last name etc by pressing the back button

### Describe the solution you'd like
Ensure that id_token is not cached or in the browsing history in any way. In these lines
https://github.com/aspnet/AspNetCore/blob/master/src/Security/Authentication/OpenIdConnect/src/OpenIdConnectHandler.cs#L259
https://github.com/aspnet/AspNetCore/blob/master/src/Security/Authentication/OpenIdConnect/src/OpenIdConnectHandler.cs#L445

Change them to
`Response.Headers[HeaderNames.CacheControl] = "no-cache no-store"`

to prevent back button caching

### Describe alternatives you've considered
Encrypting the id_token. However, many OpenID Connect providers do not support it.

### Additional context
https://blog.55minutes.com/2011/10/how-to-defeat-the-browser-back-button-cache/

anurse:
Seems reasonable. Marking 'help wanted' because it's nice and simple. @huan086 any interest in submiting a PR? If not, we'll try to look at this in preview 7.

huan086:
Reproducing issue to ensure fix works.

On IdP page, press back. Blank page appears. View source

```html
<html>

<head>
    <title>Working...</title>
</head>

<body>
    <form method="POST" name="hiddenform" action="https://[redacted]/connect/authorize">
        <input type="hidden" name="client_id" value="[redacted]" />
        <input type="hidden" name="redirect_uri" value="https://localhost:44300/[redacted]" />
        <input type="hidden" name="response_type" value="code" />
        <input type="hidden" name="scope" value="openid profile offline_access [redacted list of requested privileges]" />
        <input type="hidden" name="response_mode" value="form_post" />
        <input type="hidden" name="nonce" value="[redacted]" />
        <input type="hidden" name="state" value="[redacted]" />
        <noscript>
            <p>Script is disabled. Click Submit to continue.</p>
            <input type="submit" value="Submit" />
        </noscript>
    </form>
    <script language="javascript">
        window.setTimeout('document.forms[0].submit()', 0);
    </script>
</body>

</html>
```

[...]

PR #10741, opened 2019-06-02
Add no-store to OpenIdConnectHandler cache-control response headers

huan086:
 - Add no-store to OpenIdConnectHandler cache-control response headers, in order to prevent browsers with bfcache from storing id_token_hint and old values of nonce and state.

Addresses #10318


Tratcher (approved):

Tratcher:
Thanks

PR #10745, opened 2019-06-02
Use GetEndpoint extension method

Kahbazi:


rynowak (approved):
Thanks for the cleanup. Looks like the test failures weren't cause by this change, I'm retying.

Issue #7955, opened 2019-02-26
WsFederation SaveTokens is not working


mindaugasdirg:
Hello,

I have tried setting `WsFederationOptions.SaveTokens = true` to get token in `AuthenticationProperties` but it is not being set in Asp.Net Core 2.0 or 2.1.

According to the documentation ([WsFederationOptions documentation](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.authentication.wsfederation.wsfederationoptions?view=aspnetcore-2.1))  it should work, but it looks like this functionality is not implemented. In other Authentication models (OpenIdConnect, Oauth, JwtBearer) there is functionality to save tokens if `SaveTokens` is set to `true`. 

I can create pull request to implement it if you will be willing to review it. Or is there a reason why WsFederation does not have it?

blowdart:
What tokens do you expect there to be in WSFed? The token is the identity, there aren't refresh or access tokens in WSFed.

Tratcher:
This has been discussed before: https://github.com/aspnet/Security/issues/1716

mindaugasdirg:
In app I was expecting to get the same token which `WsFederationMessage.GetToken()` returns.

If you prefer to leave SaveTokens options as it is in WSFederation now, maybe it would be good to clarify in WsFederationOptions documentation that this option is ignored.

Eilon:
Let's override the property and:
1. Set `[EditorBrowsable(None)]`
2. Update the doc comment to say it's not supported

PR #10763, opened 2019-06-03
Indicates that SaveTokens is not supported for WsFederation

Kahbazi:
#7955


Tratcher:
Does this really prevent the base class intellisense from showing up?

The reference assemblies need to be updated because you added public API.
```
error : Generated code is not up to date in src/Security/Authentication/WsFederation/ref/Microsoft.AspNetCore.Authentication.WsFederation.netcoreapp3.0.cs. You might need to regenerate the reference assemblies or project list (see docs/ReferenceAssemblies.md ...)
```

anurse:
> Does this really prevent the base class intellisense from showing up?

I believe it will as long as the variable you are dereferencing is actually of the derived type (`WsFederationOptions`)

@Kahbazi could you verify this and post a quick screenshot?

Kahbazi:
It's working!
image

Tratcher:
Nice. And if you type in SaveTokens manually and hover over it does the right test come up?

Kahbazi:
yep
image

Tratcher (approved):

analogrelay (approved):
Very nice! Thanks @Kahbazi ! We should be able to take it from here.

PinpointTownes:
Why not implementing proper `SaveTokens` support rather than hiding the config' property? 😄

Tratcher:
@PinpointTownes What would you save? I was under the impression that the wsfed token wasn't useful beyond sign-in.

anurse:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=210299

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/2620

Tratcher:
@PinpointTownes Reading through the prior thread (https://github.com/aspnet/Security/issues/1716#issuecomment-384528952) it ended because the tokens were too large and caused the cookie to hit header size limits.

I'm merging this change for now and we can revisit if there's enough interest in those tokens.

PR #10764, opened 2019-06-03
Reduce the number of rows in the class, easy to read

YPTopMan:
 - Can the code part be returned directly? English translation
 - The method is streamlined body => syntax, if you can? 

--Translate Chinese into English via Google Translate


pranavkm (approved):
Thanks for the PR!

PR #10770, opened 2019-06-03
Blazor template fixes

SteveSandersonMS:
Issues found during initial verification for preview 6:

* In the server-side .csproj, `<None Update="app.db" CopyToOutputDirectory="PreserveNewest" />` appears if you're using LocalDB, but not if you aren't. This condition is the opposite to what it should be.
* None of the conditionals in any of the `.razor` files are applied when the template is executed from VS, although they do work correctly on the command line. This means the app does nothing useful, as all the conditionally-included code is just left commented out.
* In the client-side Blazor templates,
  * There's a compilation error: it tries to call `UseBlazorClientSideFiles`; should be `UseClientSideBlazorFiles`.
  * If you visit a "not found" URL (e.g., any made-up URL), instead of displaying "not found", it does an infinite redirect loop
  * The csproj still adds references to MyGet feeds which are no longer updated

This PR includes fixes to all the above. Notes:

 * For the "conditionals are not applied in `.razor` files", who knows who to follow up with about this? Why does VS differ from the command line for this? My workaround in this PR is kind of awkward - it would be good if we could revert it and go back to using conditionals.
 * For the redirect-loop in client-side Blazor, my workaround is to add some `NotFoundContent`, because the logic in `Router` is to do a redirect if there is no `NotFoundContent`. But why do we have this logic? I know it was [added recently here](https://github.com/aspnet/AspNetCore/blob/master/src/Components/Components/src/Routing/Router.cs#L118) but it doesn't cover this scenario. cc @pranavkm for thoughts on this, as he worked on that PR.

### Other issues

In verification I also found the following, which I don't think we can address for preview 6:

 * The client-side Blazor app has expanded back to 2.3MB (same as before the linker improvements) because we now reference a load more assemblies from `Microsoft.Extensions` (e.g., `Primitives`, `Options`, `Logging.Abstractions`) as well as `Microsoft.AspNetCore.Authorization` even if you aren't using it.
   * This is problematic because people will keep adding stuff to those `Microsoft.Extensions` packages without being aware of the consequences for client-side Blazor app size.
   * We should look at either having the linker treat `Microsoft.AspNetCore.*` as part of the BCL (i.e., subject to linking out), or try to find ways to break the dependency chains
* In my version of VS, the `@attribute` directive still doesn't show up. Not sure what version of VS I should be using.

### DIFF
--- a/src/Components/Blazor/Templates/src/SetPackageProperties.targets
+++ b/src/Components/Blazor/Templates/src/SetPackageProperties.targets
@@ -11,15 +11,11 @@
 
   <Target Name="SetTemplateJsonSymbolReplacements">
     <PropertyGroup>
-      <IncludeCustomRestoreSourcesValue>true</IncludeCustomRestoreSourcesValue>
-      <IncludeCustomRestoreSourcesValue Condition=" '$(IsFinalBuild)' == 'true' AND ('$(PreReleaseLabel)' == 'servicing' OR '$(PreReleaseLabel)' == 'rtm')">false</IncludeCustomRestoreSourcesValue>
-
       <!--
         Properties here will be injected into the template config *.json files
         during the build, replacing tokens of the form ${PropertyName}
       -->
       <GeneratedContentProperties>
-        IncludeCustomRestoreSources=$(IncludeCustomRestoreSourcesValue);
         TemplateBlazorVersion=$(PackageVersion);
         TemplateComponentsVersion=$(ComponentsPackageVersion);
         RepositoryCommit=$(SourceRevisionId);
--- a/src/Components/Blazor/Templates/src/content/BlazorHosted-CSharp/BlazorHosted-CSharp.Client/App.razor
+++ b/src/Components/Blazor/Templates/src/content/BlazorHosted-CSharp/BlazorHosted-CSharp.Client/App.razor
@@ -1,1 +1,5 @@
-﻿<Router AppAssembly="typeof(Program).Assembly" />
+﻿<Router AppAssembly="typeof(Program).Assembly">
+    <NotFoundContent>
+        <p>Sorry, there's nothing at this address.</p>
+    </NotFoundContent>
+</Router>
--- a/src/ProjectTemplates/Web.ProjectTemplates/content/RazorComponentsWeb-CSharp/App.razor
+++ b/src/ProjectTemplates/Web.ProjectTemplates/content/RazorComponentsWeb-CSharp/App.razor
@@ -1,7 +1,6 @@
 ﻿<CascadingAuthenticationState>
     <Router AppAssembly="typeof(Startup).Assembly">
         <NotFoundContent>
-            <h3>Page not found</h3>
             <p>Sorry, there's nothing at this address.</p>
         </NotFoundContent>
     </Router>
--- a/src/ProjectTemplates/Web.ProjectTemplates/content/RazorComponentsWeb-CSharp/Shared/LoginDisplay.IndividualB2CAuth.razor
+++ b/src/ProjectTemplates/Web.ProjectTemplates/content/RazorComponentsWeb-CSharp/Shared/LoginDisplay.IndividualB2CAuth.razor
@@ -1,15 +1,3 @@
-@*#if (IndividualLocalAuth)
-<AuthorizeView>
-    <Authorized>
-        <a href="Identity/Account/Manage">Hello, @context.User.Identity.Name!</a>
-        <a href="Identity/Account/LogOut">Log out</a>
-    </Authorized>
-    <NotAuthorized>
-        <a href="Identity/Account/Register">Register</a>
-        <a href="Identity/Account/Login">Log in</a>
-    </NotAuthorized>
-</AuthorizeView>
-#elseif (IndividualB2CAuth)
 @using Microsoft.AspNetCore.Authentication.AzureADB2C.UI
 @using Microsoft.Extensions.Options
 @inject IOptionsMonitor<AzureADB2COptions> AzureADB2COptions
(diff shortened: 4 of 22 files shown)


src/ProjectTemplates/Web.ProjectTemplates/content/RazorComponentsWeb-CSharp/Shared/LoginDisplay.IndividualB2CAuth.razor
```diff
@@ -1,15 +1,3 @@
-@*#if (IndividualLocalAuth)
```
rynowak: @seancpeters do you know why this wouldn't work with `.razor`? Does this work with `.cshtml`?

rynowak (approved):

src/ProjectTemplates/Web.ProjectTemplates/content/RazorComponentsWeb-CSharp/Shared/LoginDisplay.IndividualB2CAuth.razor
rynowak: It looks like the templating engine does understand this https://github.com/dotnet/templating/blob/c758b031c16827a6d760f125705945c4cf5f9b9d/src/Microsoft.TemplateEngine.Orchestrator.RunnableProjects/SimpleConfigModel.cs#L414

SteveSandersonMS: Yes, it works fine from the command line. It just doesn't work inside VS. I'm not sure why they would be different.

src/Components/Blazor/Templates/src/SetPackageProperties.targets
```diff
@@ -11,15 +11,11 @@
...
     <PropertyGroup>
-      <IncludeCustomRestoreSourcesValue>true</IncludeCustomRestoreSourcesValue>
-      <IncludeCustomRestoreSourcesValue Condition=" '$(IsFinalBuild)' == 'true' AND ('$(PreReleaseLabel)' == 'servicing' OR '$(PreReleaseLabel)' == 'rtm')">false</IncludeCustomRestoreSourcesValue>
-
       <!--
         Properties here will be injected into the template config *.json files
         during the build, replacing tokens of the form ${PropertyName}
       -->
       <GeneratedContentProperties>
-        IncludeCustomRestoreSources=$(IncludeCustomRestoreSourcesValue);
```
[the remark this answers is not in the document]
SteveSandersonMS: Yes, I removed everything to do with emitting the config for the MyGet restore sources, so this is no longer used.

src/Components/Blazor/Templates/src/content/BlazorHosted-CSharp/BlazorHosted-CSharp.Client/App.razor
```diff
@@ -1 +1,5 @@
-﻿<Router AppAssembly="typeof(Program).Assembly" />
+﻿<Router AppAssembly="typeof(Program).Assembly">
```
ryanbrandenburg: This is because this component is route-able, but doesn't really have any meaning?

ryanbrandenburg (commented):
One question for my own education. You definitely want to rely on the Blazor folks for your Approval though.

I'm also interested to learn why it doesn't work in VS.

src/ProjectTemplates/Web.ProjectTemplates/content/RazorComponentsWeb-CSharp/App.razor
```diff
@@ -1,7 +1,6 @@
 ﻿<CascadingAuthenticationState>
```
danroth27: Is leaving in this cascading authentication state intentional even when no authentication is enabled?

src/Components/Blazor/Templates/src/content/BlazorHosted-CSharp/BlazorHosted-CSharp.Client/App.razor
SteveSandersonMS: The `AppAssembly` parameter specifies which assembly to look in to find the routable page components, and it's not necessarily the same as your app entrypoint assembly.

src/ProjectTemplates/Web.ProjectTemplates/content/RazorComponentsWeb-CSharp/App.razor
SteveSandersonMS: Yes, it's intentional, because there's no drawback to having it here if it's not doing anything.

However we had a whole discussion before about whether it could be eliminated from the app source code entirely, even if you are using authentication. I personally would still like it if we did that, but it's not urgent - we can ship 3.0 without doing that, and it would be a back-compatible change in the future to make it there by default with duplicates being a no-op.

danroth27 (approved):
LGTM

src/ProjectTemplates/Web.ProjectTemplates/content/RazorComponentsWeb-CSharp/Shared/LoginDisplay.IndividualB2CAuth.razor
seancpeters: The reason for this not working in VS would seem to be that the latest bits of dotnet/templating are not shipped with VS. I'm looking into this - and will try to get the bits updated in VS.

SteveSandersonMS:
@mkArtakMSFT Can this be merged into the preview 6 branch, or is some further approval needed?

PR #10776, opened 2019-06-03
Save condition backcollection between runs

### LINKED ISSUE #10768: Apache mod_rewrite query string matches from RewriteCond not available in RewriteRule substitution when QSD flag used
Using the rewrite rule below, read in from the mod_rewrites.txt. The resulting redirect does not have the report id value as expected at the end of the url. The %1 should have the value matched from the backreference to the RewriteCond line. This rule works as I expect on Apache

### To Reproduce
1. Using ASP.NET Core 2.2

2. Add mod_rewrites.txt with this rewrite rule in it
RewriteCond %{REQUEST_URI} /home
RewriteCond %{QUERY_STRING} report_id=(.+)
RewriteRule (.*) http://localhost:64782/home/report/%1 [R=301,L,QSD]

3. Add this call in Startup.cs
```
using (var sr = File.OpenText("mod_rewrites.txt")) {
    var rwo = new RewriteOptions().AddApacheModRewrite(sr);
    app.UseRewriter(rwo);
}
```
4. Add HomeController action
```
[HttpGet("/home/report/{id}")]
        public IActionResult Report(int id)
        {
            return Content("rewritten id = " + id);
        }
```
5. Run and navigate to http://localhost:64782/home?report_id=123

### Expected behavior
Rewritten URL is http://localhost:64782/home/report/123

### Actual behavior
Rewritten URL is http://localhost:64782/home/report/

### Additional context
This rewrite rule works as expected in Apache

Issue previously submitted as doc bug [here](https://github.com/aspnet/AspNetCore.Docs/issues/12613)

jkotalik:
Fixes https://github.com/aspnet/AspNetCore/issues/10768

### DIFF
--- a/src/Middleware/Rewrite/src/Internal/ApacheModRewrite/ApacheModRewriteRule.cs
+++ b/src/Middleware/Rewrite/src/Internal/ApacheModRewrite/ApacheModRewriteRule.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


mikaelm12 (approved):
👍

mikaelm12:
I'm sad that this bug existed for so long though. There's just so many configuration options which makes it difficult to make sure everything is tested thoroughly.

PR #10807, opened 2019-06-03
Revert "Add an option to SignalR to guarantee ordering of send events…

pranavkm:
… (#10506)"

This reverts commit 4646c4a5e4412dbff6e40ade887792241f5a2c94.


BrennanConroy:
Well... we can keep the change in SignalR and just remove all the infrastructure stuff.

pranavkm:
Just keeping this prepped. I'd like to take a stab at fixing the build if I can with the change.

BrennanConroy:
>Just keeping this prepped. I'd like to take a stab at fixing the build if I can with the change.

Sure, but a better prep is to not remove the feature you made :D

pranavkm:
@aspnet/build could you merge this in? The official build is currently broken due to this change.

cc @Pilchie \ @mkArtakMSFT

natemcmaster (approved):
:shipit: We haven't had a successful build since the original PR went in, and it's blocking preview6 dependency flow. Let's avoid rushing in a fix and target preview7.

Issue #10707, opened 2019-05-31
Add `OnRejected` callback to Request Queuing middleware to allow user code to handle rejections


anurse:
Epic #10702 

Right now, when a request is rejected by the queuing middleware, we just 503. We should provide a user callback to allow user code to handle it specially.

anurse:
Done in https://github.com/aspnet/AspNetCore/pull/10817 !

PR #10817, opened 2019-06-03
Add OnRejected to Request Throttling middleware

Kahbazi:
#10707

### DIFF
--- a/src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
+++ b/src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
@@ -18,6 +18,7 @@ namespace Microsoft.AspNetCore.RequestThrottling
     {
         private readonly RequestQueue _requestQueue;
         private readonly RequestDelegate _next;
+        private readonly RequestThrottlingOptions _requestThrottlingOptions;
         private readonly ILogger _logger;
 
         /// <summary>
@@ -37,7 +38,17 @@ namespace Microsoft.AspNetCore.RequestThrottling
                 throw new ArgumentException("The value of 'options.RequestQueueLimit' must be a positive integer.", nameof(options));
             }
 
+            if (options.Value.OnRejected == null)
+            {
+                options.Value.OnRejected = context =>
+                {
+                    context.Response.StatusCode = StatusCodes.Status503ServiceUnavailable;
+                    return Task.CompletedTask;
+                };
+            }
+
             _next = next;
+            _requestThrottlingOptions = options.Value;
             _logger = loggerFactory.CreateLogger<RequestThrottlingMiddleware>();
             _requestQueue = new RequestQueue(
                 options.Value.MaxConcurrentRequests.Value,
@@ -55,7 +66,7 @@ namespace Microsoft.AspNetCore.RequestThrottling
             if (waitInQueueTask.IsCompletedSuccessfully && !waitInQueueTask.Result)
             {
                 RequestThrottlingLog.RequestRejectedQueueFull(_logger);
-                context.Response.StatusCode = StatusCodes.Status503ServiceUnavailable;
+                await _requestThrottlingOptions.OnRejected(context);
                 return;
             }
             else if (!waitInQueueTask.IsCompletedSuccessfully)
--- a/src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
+++ b/src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
@@ -29,30 +29,27 @@ namespace Microsoft.AspNetCore.RequestThrottling
         /// <param name="options">The <see cref="RequestThrottlingOptions"/> containing the initialization parameters.</param>
         public RequestThrottlingMiddleware(RequestDelegate next, ILoggerFactory loggerFactory, IOptions<RequestThrottlingOptions> options)
         {
-            if (options.Value.MaxConcurrentRequests == null)
+            _requestThrottlingOptions = options.Value;
+
+            if (_requestThrottlingOptions.MaxConcurrentRequests == null)
             {
                 throw new ArgumentException("The value of 'options.MaxConcurrentRequests' must be specified.", nameof(options));
             }
-            if (options.Value.RequestQueueLimit < 0)
+            if (_requestThrottlingOptions.RequestQueueLimit < 0)
             {
                 throw new ArgumentException("The value of 'options.RequestQueueLimit' must be a positive integer.", nameof(options));
             }
 
-            if (options.Value.OnRejected == null)
+            if (_requestThrottlingOptions.OnRejected == null)
             {
-                options.Value.OnRejected = context =>
-                {
-                    context.Response.StatusCode = StatusCodes.Status503ServiceUnavailable;
-                    return Task.CompletedTask;
-                };
+                throw new ArgumentException("The value of 'options.OnRejected' must not be null.", nameof(options));
             }
 
             _next = next;
-            _requestThrottlingOptions = options.Value;
             _logger = loggerFactory.CreateLogger<RequestThrottlingMiddleware>();
             _requestQueue = new RequestQueue(
-                options.Value.MaxConcurrentRequests.Value,
-                options.Value.RequestQueueLimit);
+                _requestThrottlingOptions.MaxConcurrentRequests.Value,
+                _requestThrottlingOptions.RequestQueueLimit);
         }
 
         /// <summary>
--- a/src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
+++ b/src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
@@ -63,6 +63,7 @@ namespace Microsoft.AspNetCore.RequestThrottling
             if (waitInQueueTask.IsCompletedSuccessfully && !waitInQueueTask.Result)
             {
                 RequestThrottlingLog.RequestRejectedQueueFull(_logger);
+                context.Response.StatusCode = StatusCodes.Status503ServiceUnavailable;
                 await _requestThrottlingOptions.OnRejected(context);
                 return;
             }
--- a/src/Middleware/RequestThrottling/src/RequestThrottlingOptions.cs
+++ b/src/Middleware/RequestThrottling/src/RequestThrottlingOptions.cs
@@ -26,7 +26,7 @@ namespace Microsoft.AspNetCore.RequestThrottling
         public int RequestQueueLimit { get; set; } = 5000;
 
         /// <summary>
-        /// 
+        /// Optional callback; processes requests rejected by the middleware.
         /// </summary>
         public RequestDelegate OnRejected { get; set; } = context =>
         {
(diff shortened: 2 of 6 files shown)


anurse:
Thanks for the submission @Kahbazi ! One of the parts of this middleware I didn't mention clearly is that it is a project our summer intern is working on. However, I think this would be a great opportunity to introduce him to working with external contributors ;).

@DylanDmitri @halter73 Can you both work to see if we can get @Kahbazi 's change iterated on and merged into the repo?

@Kahbazi The first major bit of feedback I can see already is that we're going to need unit tests for this functionality before we could merge it in?

Kahbazi:
@anurse  I'm often looking for bits that I can contribute, and this middleware seems a good one to me. 

> it is a project our summer intern is working on

Does this mean that I shouldn't contribute for this middleware?

Tratcher:
> Does this mean that I shouldn't contribute for this middleware?

You're welcome to contribute but keep two things in mind:
- This component is under active development so be careful not to duplicate work (already assigned issues).
- We'll have our intern @DylanDmitri review and accept the contributions (with guidance from the rest of us).

Kahbazi:
I See. Thanks.

DylanDmitri (changes requested):

DylanDmitri:
Also, the build is complaining about reference assemblies, which is fixed by going to `/eng/scripts/` and running `.\GenerateRefenceAssemblies.ps1`

anurse:
https://github.com/aspnet/AspNetCore/blob/master/docs/ReferenceAssemblies.md has some docs about how to generate the reference assemblies :). You can actually do it with an MSBuild command.

DylanDmitri:
The MSBuild command was broken for me for a while so I didn't recommend it. It works now though!

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
```diff
@@ -37,7 +38,17 @@ public RequestThrottlingMiddleware(RequestDelegate next, ILoggerFactory loggerFa
...
+                    context.Response.StatusCode = StatusCodes.Status503ServiceUnavailable;
+                    return Task.CompletedTask;
+                };
+            }
+
             _next = next;
+            _requestThrottlingOptions = options.Value;
             _logger = loggerFactory.CreateLogger<RequestThrottlingMiddleware>();
             _requestQueue = new RequestQueue(
                 options.Value.MaxConcurrentRequests.Value,
```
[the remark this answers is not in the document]
Kahbazi: Do you mean that I should only store `OnRejected`?

src/Middleware/RequestThrottling/test/MiddlewareTests.cs
```diff
@@ -127,5 +127,24 @@ public void MultipleRequestsFillUpQueue()
...
+                maxConcurrentRequests: 0,
+                requestQueueLimit: 0,
+                onRejected: httpContext =>
+                {
+                    onRejectedInvoked = true;
+                    return Task.CompletedTask;
+                });
+
+            var context = new DefaultHttpContext();
+            await middleware.Invoke(context);
```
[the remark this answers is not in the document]
Kahbazi: There are other tests which don't have OrTimeout after await. Should I add it to those too?

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
```diff
@@ -55,7 +63,7 @@ public async Task Invoke(HttpContext context)
             if (waitInQueueTask.IsCompletedSuccessfully && !waitInQueueTask.Result)
             {
                 RequestThrottlingLog.RequestRejectedQueueFull(_logger);
-                context.Response.StatusCode = StatusCodes.Status503ServiceUnavailable;
+                await _requestThrottlingOptions.OnRejected(context);
```
halter73: What if this throws? As it turns out, `_requestQueue.Release()` never gets called meaning the server will eventually stop accepting any new requests. This is ultimately the fault of not putting the start of the try/finally block in this method immediately after the call to `_requestQueue.TryEnterQueueAsync()`, but that's not the only thing that should be fixed.

Any time we call user-specified callbacks, we always wrap the call in a generic try/catch that swallows and logs any user-thrown exceptions. You can see an example of this where [Kestrel invokes the middleware pipeline.](https://github.com/aspnet/AspNetCore/blob/81f2d46660ec6ea92549edd55486de78a7597c58/src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs#L628-L651)

Tratcher: Kestrel's exception handling is special, it's at the top of the stack. A user callback throwing on the main request thread is fine, it gets logged by the server or exception handler and becomes a 500. No need to do anything extra so long as it the middleware state is maintained.

Tratcher: In fact, swallowing the exception could lead to false 200 responses.

Kahbazi: So I don't need to change anything, right?

DylanDmitri: With the first part of your comment @halter73, the code sure is super misleading. Also when implementing the other queuing strategy (that rejects at dequeue time) I needed to refactor this method in middleware to deal with the new flow, and then refactored it back after putting that on hold. Basically it's pretty fragile and I need to go back and work on it.

halter73: > In fact, swallowing the exception could lead to false 200 responses.

That would need to be accounted for of course. That's one reason why I recommended adding a throwing OnRejected test which we should add anyway, but it might not make too much sense to do so until we add more functional testing to verify we get a 503/500 instead of a 200 response.

After in person discussion with @Tratcher, we decided that since this exception will still get thrown from the main middleware pipeline, and that should still get logged with a helpful StackTrace, we're fine not catching here.

I personally still think it's more correct to respond with a 503 instead of a 500 and swallow/log the exception, but I can understand not wanting to add the complexity.

> So I don't need to change anything, right?

No. Hopefully the above explains the reasoning.

DylanDmitri:
Alright @Kahbazi there's one last change and then it's all good to go! This is adding the unit test that Stephen mentioned above. 

Throw an exception during the `OnRejected` callback, and verify that the server replies appropriately.

src/Middleware/RequestThrottling/test/MiddlewareTests.cs
```diff
@@ -146,5 +146,19 @@ public async void FullQueueInvokesOnRejected()
             await middleware.Invoke(context).OrTimeout();
             Assert.True(onRejectedInvoked);
         }
+
+        [Fact]
+        public async void ExceptionThrownDuringOnRejected()
```
Kahbazi: @halter73 @DylanDmitri I'm not sure if I quite understand the assertion part of the test that is needed. Is this what you had in mind?

DylanDmitri: looks good!

src/Middleware/RequestThrottling/test/MiddlewareTests.cs
```diff
@@ -127,5 +127,38 @@ public void MultipleRequestsFillUpQueue()
...
+            var context = new DefaultHttpContext();
+            await middleware.Invoke(context).OrTimeout();
+            Assert.True(onRejectedInvoked);
+        }
+
+        [Fact]
+        public async void ExceptionThrownDuringOnRejected()
+        {
+            var middleware = TestUtils.CreateTestMiddleware(
+                maxConcurrentRequests: 0,
```
halter73: This is a good start, but can we expand this test a little? I recommend the following.

1. Make `maxConcurrentRequests` equal to 1
2. Pass in a `next` RequestDelegate that returns the Task from a TCS (TaskCompletionSource) that you define at the top of the test.
3. Invoke the middleware and store the returned Task without awaiting it.
4. Invoke the middleware a second time and assert the DivideByZeroException like you're already doing.
5. Complete the TCS used by the `next`.
6. Assert that the Task returned from the first middleware invocation completes successfully.
7. Invoke the middleware a third time and Assert that it completes successfully since there's no other active requests and the Task returned by `next` is now completed.

You can also assert the `ActiveRequestCount` values like @DylanDmitri suggested in a previous comment.

DylanDmitri: Another behavior change: we are afraid that people will overwrite `OnRejected` to log a message, and then the server will return an http 200.

The fix here is having the middleware set the statuscode to 503, and then invoking the user provided callback. The user can still set custom status codes, but by default it 503s.

Also, this means that the default `OnRejected` (in the options file) is now an empty lambda, since its current functionality gets moved to the middleware.

halter73 (approved):
Thanks @Kahbazi!

src/Middleware/RequestThrottling/src/RequestThrottlingOptions.cs
```diff
@@ -22,5 +24,13 @@ public class RequestThrottlingOptions
         /// Defaults to 5000 queued requests.
         /// </summary>
         public int RequestQueueLimit { get; set; } = 5000;
+
+        /// <summary>
+        ///
```
halter73: We still need add a doc comment for this though.

DylanDmitri: maybe "Optional callback; processes requests rejected by the middleware."

DylanDmitri (approved):

src/Middleware/RequestThrottling/src/RequestThrottlingOptions.cs
halter73: Everything has defaults, so I don't think we should bother pointing out it's optional. How about this?

"A <see cref="RequestDelegate"/> that handles requests rejected by this middleware. If it doesn't modify the response, an empty 503 response will be written."

halter73: Sorry to make you iterate so much @Kahbazi. @DylanDmitri I'm fine with you merging this as is and fixing up the doc comment later.

Kahbazi: no, it's ok. I can change it again when you decided what you want it to be. I think maybe "A callback that handles rejected request by this middleware." is good enough since 503 status code is now part  of the middleware and not OnRejected.

halter73: I do like making it clear that it's a RequestDelegate that can modify the response that is chooses to. I just realized my last comment wasn't formatted properly so you couldn't see the `<see cref="RequestDelegate"/>`'s. It was meant to be:

`"A <see cref="RequestDelegate"/> that handles requests rejected by this middleware. If it doesn't modify the response, an empty 503 response will be written."`

@DylanDmitri @Kahbazi Does this seem reasonable to you?

DylanDmitri: Stephen's description is better and more professional

PR #10819, opened 2019-06-04
Add mime types in summary

Kahbazi:
I think it is good to be able to see the default mime types by checking the summary instead of source code or at runtime.


anurse:
I don't think there's a good way to render this that won't be kinda ugly in the tool-tip view VS provides. This is better rendered in non-API docs. Thanks for the contribution, but we're going to decline this change.

PR #10823, opened 2019-06-04
If libuv doesn't unwind gracefully log a warning

davidfowl:
- We added 3 critical logs to see if there was a resource leak happeing during the loop thread shutdown, turns out that during testing its very common for the thread to not end immediately because of the latency of thread pool dispatching (the latency between queuing and execution of the work item). As a result, we log a warning instead of an error.

Fixes https://github.com/aspnet/AspNetCore-Internal/issues/2606

### DIFF
--- a/src/Servers/Kestrel/Transport.Libuv/src/Internal/LibuvThread.cs
+++ b/src/Servers/Kestrel/Transport.Libuv/src/Internal/LibuvThread.cs
@@ -112,7 +112,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Libuv.Internal
                 Post(t => t.AllowStop());
                 if (!await WaitAsync(_threadTcs.Task, stepTimeout).ConfigureAwait(false))
                 {
-                    _log.LogCritical($"{nameof(LibuvThread)}.{nameof(StopAsync)} failed to terminate libuv thread, {nameof(AllowStop)}");
+                    _log.LogWarning($"{nameof(LibuvThread)}.{nameof(StopAsync)} failed to terminate libuv thread, {nameof(AllowStop)}");
 
                     Post(t => t.OnStopRude());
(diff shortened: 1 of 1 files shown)


jkotalik (approved):

src/Servers/Kestrel/Transport.Libuv/src/Internal/LibuvThread.cs
```diff
@@ -112,7 +112,7 @@ public async Task StopAsync(TimeSpan timeout)
                 Post(t => t.AllowStop());
                 if (!await WaitAsync(_threadTcs.Task, stepTimeout).ConfigureAwait(false))
                 {
-                    _log.LogCritical($"{nameof(LibuvThread)}.{nameof(StopAsync)} failed to terminate libuv thread, {nameof(AllowStop)}");
+                    _log.LogWarning($"{nameof(LibuvThread)}.{nameof(StopAsync)} failed to terminate libuv thread, {nameof(AllowStop)}");
```
halter73: This kinda feels like a cheap way to avoid reverting your change rather than finding the real regression...

halter73: My bad. I thought this was removing the line on 134. I need to pay more attention. Carry on.

halter73 (approved):

src/Servers/Kestrel/Transport.Libuv/src/Internal/LibuvThread.cs
halter73: This is pretty rare, right? We think this is due to a thread starved environment?

davidfowl: That's my current guess.

anurse:
QB Approval was granted for this (just noting for tracking).

PR #10824, opened 2019-06-04
Disable OSX 10.14 tests on Helix by default and only run select tests

JunTaoLuo:
Addresses https://github.com/aspnet/AspNetCore-Internal/issues/2566


src/Security/test/AuthSamples.FunctionalTests/AuthSamples.FunctionalTests.csproj
```diff
@@ -6,6 +6,10 @@
     <PlatformTarget>AnyCPU</PlatformTarget>
   </PropertyGroup>
 
+  <ItemGroup>
```
JunTaoLuo: This is just for testing purposes. Need to actually identify a list of tests that need to be run on OSX 10.14

src/Security/test/AuthSamples.FunctionalTests/AuthSamples.FunctionalTests.csproj
```diff
@@ -4,6 +4,7 @@
     <TargetFramework>netcoreapp3.0</TargetFramework>
     <IsSampleProject>false</IsSampleProject>
     <PlatformTarget>AnyCPU</PlatformTarget>
+    <RunOnOSX1014>true</RunOnOSX1014>
```
JunTaoLuo: This is just for testing purposes. Need to actually identify a list of tests that need to be run on OSX 10.14

jkotalik: I say we identify the list of tests after we unblock helix.

jkotalik (approved):

src/Security/test/AuthSamples.FunctionalTests/AuthSamples.FunctionalTests.csproj
JunTaoLuo: Yup, and the results before I accidentally closed the PR illustrates that this mechanism of enabling test runs on OSX.10.14  works: https://mc.dot.net/#/user/aspnetcore/pr~2Faspnet~2Faspnetcore/ci/20190604.32.

eng/targets/Helix.Common.props
```diff
@@ -28,6 +28,10 @@
     <!--  TODO: re-enable Debian.9.Arm64.Open and Ubuntu.1804.Arm64.Open    -->
   </ItemGroup>
 
+  <ItemGroup Condition="'$(RunOnOSX1014)' == 'true'">
```
dougbu: What sets `$(RunOnOSX1014)`?

dougbu (commented):
Looks :ok: but&hellip;

dougbu (approved):
Meant 2 approve :shipit: if property is useful

eng/targets/Helix.Common.props
JunTaoLuo: That's why I left a comment on a test 😄 https://github.com/aspnet/AspNetCore/pull/10824#pullrequestreview-245248769

src/Security/test/AuthSamples.FunctionalTests/AuthSamples.FunctionalTests.csproj
JunTaoLuo: Okay actually there must be at least one test that runs on each queue. I'm going to just turn this test on as a placeholder until we determine which the set of tests that actually need to be tested on this platform.

jkotalik: Doesn't that defeat the purpose of this change? This will still cause Helix to hang for 4 hours.

eng/targets/Helix.Common.props
```diff
@@ -16,7 +16,7 @@
     <HelixAvailableTargetQueue Include="Windows.7.Amd64.Open" Platform="Windows" />
     <HelixAvailableTargetQueue Include="Windows.10.Amd64.EnterpriseRS3.ASPNET.Open" Platform="Windows" EnableByDefault="false" />
     <HelixAvailableTargetQueue Include="OSX.1012.Amd64.Open" Platform="OSX" />
-    <HelixAvailableTargetQueue Include="OSX.1014.Amd64.Open" Platform="OSX" />
+    <HelixAvailableTargetQueue Include="OSX.1014.Amd64.Open" Platform="OSX" EnableByDefault="false" />
```
jkotalik: Can you see what happens when you remove this line entirely? I think that's the reason you are getting an error on helix.

JunTaoLuo: This list is used to list all the helix queues that are used as well as which helix queues are used to test per each project. If we want to completely remove this platform that would work, but if we want to enable this queue for some subset of tests, then we need to keep this here and mark it with EnableByDefault="false". This is why `Windows.10.Amd64.EnterpriseRS3.ASPNET.Open` looks the same.

I'm making the changes with the assumption that we actually have a set of tests to run on OSX 10.14. If there are no such tests then yes, we should remove the queue completely.

src/Security/test/AuthSamples.FunctionalTests/AuthSamples.FunctionalTests.csproj
JunTaoLuo: The point of this change is to run a smaller set of tests so we don't overwhelm the queue. But it doesn't mean we want to completely remove coverage.

eng/targets/Helix.Common.props
jkotalik: I'm advocating for what unblocks helix and removes the red X entirely. My concern is that having even just one agent enabled will cause the test run not to complete as, presumably, other teams are still using these mac agents too.

>I'm making the changes with the assumption that we actually have a set of tests to run on OSX 10.14. If there are no such tests then yes, we should remove the queue completely.

Eventually yes, we should run all tests on OSX 10.14, but until we have more agents, we should remove the workload entirely.

natemcmaster:
Looks like we're still getting timeouts waiting for available 10.14 agents. Let's remove this test queue altogether until we get positive confirmation from dnceng that they've added more 10.14 agents.

Btw, we might try running in 10.13 instead. I think they have a lot more capacity in that queue.

natemcmaster (approved):

PR #10825, opened 2019-06-04
[2.1] Remove AuthZ caching from AuthorizeFilter

HaoK:
2.1 Port of https://github.com/aspnet/AspNetCore/issues/9779

Customer Impact: When mixing multiple authentication schemes and authorize attributes (on controller + methods), we end up caching incorrectly, so the wrong authorization policy is applied on subsequent requests. This usually results in valid requests being rejected, I don't think there's any way this would let people who shouldn't be allowed access in.

Fix description: is to remove all the caching of authorization policies in MVC, so authorization policies are computed every time

Risk: Should be low (just removing caching logic, so just slower)


anurse:
@rynowak @mkArtakMSFT can we get a review from someone on MVC (since this changes a bunch of AuthorizeFilter stuff). It's mostly a port from 3.0.

eng/PatchConfig.props
```diff
@@ -35,5 +35,10 @@ Later on, this will be checked using this condition:
       Microsoft.AspNetCore.SignalR.Redis;
     </PackagesInPatch>
   </PropertyGroup>
+  <PropertyGroup Condition=" '$(VersionPrefix)' == '2.1.12' ">
+    <PackagesInPatch>
+      Microsoft.AspNetCore.Mvc.Core;
```
anurse: @aspnet/build please double-check this when you can :).

HaoK:
Hold off on the review for a bit, looks like there were some 2.2 to 2.1 differences that I need to resolve, ah backporting fun

rynowak:
OK. I'm ready to look at this when you are :)

dougbu:
Do not merge until / unless this goes back to shiproom and is approved.

src/Mvc/Mvc.Core/src/Authorization/AuthorizeFilter.cs
```diff
@@ -140,14 +152,13 @@ private async Task<AuthorizationPolicy> GetEffectivePolicyAsync(AuthorizationFil
                     
                     if (context.Filters[i] is AuthorizeFilter authorizeFilter)
                     {
-                        builder = builder ?? new AuthorizationPolicyBuilder(effectivePolicy);
-                        builder.Combine(authorizeFilter.Policy);
```
rynowak: This logic is a little bit different right? `.Policy` is different stuff than `ComputePolicyAsync`

src/Mvc/test/Mvc.IntegrationTests/AuthorizeFilterIntegrationTest.cs
```diff
@@ -72,18 +140,18 @@ private static ApplicationModelProviderContext GetProviderContext(Type controlle
...
-            serviceCollection.AddMvc();
+            serviceCollection.AddMvc(o => o.AllowCombiningAuthorizeFilters = combineAuthorize);
             serviceCollection
                 .AddTransient<ILoggerFactory, LoggerFactory>()
                 .AddTransient<ILogger<DefaultAuthorizationService>, Logger<DefaultAuthorizationService>>()
                 .AddSingleton<ObjectPoolProvider, DefaultObjectPoolProvider>();
             return serviceCollection.BuildServiceProvider();
         }
-
+
```
rynowak: Functional test with the repro for the bug that's being fixed?

src/Mvc/Mvc.Core/src/Authorization/AuthorizeFilter.cs
rynowak: I guess this is the same as what's in 2.2/3.0. Is the difference between these something to be concerned about?

HaoK: Yeah 2.1 was different from 2.2/3.0 to start, I am trying to make all of them the same with this PR

rynowak: I get that this changes the code to be more consistent, but does this change the observable behavior?

HaoK: So maybe we shouldn't port this change to 2.1, it looks like we never ported this 2.2 fix to 2.1 which is why the code bases are so different...  https://github.com/aspnet/Mvc/commit/a5083d525b76ed8699d2b94e91d7ef52106d65aa#diff-fbf649a01b0521186bc5d75f07c20d8f

eng/PatchConfig.props
natemcmaster: PatchConfig.props looks good now.

HaoK:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=213310

HaoK:
I'm seeing some failures on Linux for 5 MVC Modelbinding tests (this should be totally unrelated to my changes except for maybe adding a reference to Mvc.Core).  Rerun didn't work either

aspnet-hello:

HaoK:
Ok looks like the linux model binding failures are unrelated

src/Mvc/test/Mvc.IntegrationTests/AuthorizeFilterIntegrationTest.cs
HaoK: I will try to port the new functional tests in 3.0 that pranav added for this scenario to 2.1/2.2 https://github.com/aspnet/AspNetCore/blob/master/src/Mvc/test/Mvc.FunctionalTests/AuthMiddlewareAndFilterTestBase.cs

HaoK:
Closing for now

udlose:
@HaoK was this ever fixed in .NET Core 2.2?  I am using 2.2 and experiencing this exact problem.

PR #10830, opened 2019-06-04
Collect dumps for tests that seem to be flaky

davidfowl:
Lets see if we can find any blocking going on when these tests fail with timeouts.


halter73 (approved):

davidfowl:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=210986

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/2303

anurse:
Does this... work now?

davidfowl:
Only on windows. But it *can* also work on linux if we switch to event pipe (https://github.com/dotnet/diagnostics/blob/049d377673644e073b560da4a856532c58a09054/src/Microsoft.Diagnostics.Tools.RuntimeClient/DiagnosticHelpers.cs#L34)

anurse:
Swanky

Issue #10789, opened 2019-06-03
Changes needed to microsoft.extensions.apidescription.client targets and props files to work in Visual Studio


BillHiebert:
### Describe the bug
1.	Add AvailableItemName Items for OpenApiReference and OpenApiReferenceProject. Without these CPS doesn’t allow me to add items with this type. I manually modified the Microsoft.Extensions.ApiDescription.Client.props file with:
```
 <!--
    Well-known metadata of the code generator item groups.
  -->
  <ItemGroup>
    <AvailableItemName Include="OpenApiReference" />
    <AvailableItemName Include="OpenApiProjectReference" />
  </ItemGroup>
```

2.  We really need xaml rules defined for these item types. This provides a mechamism to be notified when a property on one of these items changes,  and provides F4 (property grid) support for defined properties. CPS auto-generates properties from the rules files and makes them available in the property gird – this  allows the customer to change values we don’t specifically have UI for. I created a placeholder file to test it out and wired it in by add a “tools” target file (see attached files, renamed with .txt). I imported the “tools” targets from the *.client.targets as follows:
```
  <!-- OpenApiProjectReference support. -->

  <Import Project ="$(MSBuildThisFileDirectory)Microsoft.Extensions.ApiDescription.Client.Tools.targets" Condition="Exists('$(MSBuildThisFileDirectory)Microsoft.Extensions.ApiDescription.Client.Tools.targets')" />
  <ItemGroup>
    <!--
```
[Microsoft.Extensions.ApiDescription.Client.Tools.targets.txt](https://github.com/aspnet/AspNetCore/files/3249026/Microsoft.Extensions.ApiDescription.Client.Tools.targets.txt)
[OpenApiItemsSchema.xaml.txt](https://github.com/aspnet/AspNetCore/files/3249027/OpenApiItemsSchema.xaml.txt)

dougbu:
@BillHiebert is there a typo in your .targets change? Looks like it should be `Include="$(MSBuildThisFileDirectory)OpenApiItemsSchema.xaml"` rather than importing itself. The attached .targets.txt file looks closer though the condition isn't necessary if I'm understanding correctly. That is, we should ship the .xaml file in the Microsoft.Extensions.ApiDescription.Client package, thereby ensuring it exists. Or, should we look for the .xaml file somewhere else?

BillHiebert:
[...]

PR #10851, opened 2019-06-04
Add metadata needed for tooling of service reference features

dougbu:
- #10789

### DIFF
--- a/src/Mvc/Extensions.ApiDescription.Client/src/buildMultiTargeting/Microsoft.Extensions.ApiDescription.Client.targets
+++ b/src/Mvc/Extensions.ApiDescription.Client/src/buildMultiTargeting/Microsoft.Extensions.ApiDescription.Client.targets
@@ -1,5 +1,11 @@
 ﻿<?xml version="1.0" encoding="utf-8" standalone="no"?>
 <Project>
+  <ItemGroup>
+    <PropertyPageSchema Include="../build/OpenApiItemsSchema.xaml">
+      <Context>File;BrowseObject</Context>
+    </PropertyPageSchema>
+  </ItemGroup>
+
   <Target Name="GenerateOpenApiCode">
     <MSBuild Projects="$(MSBuildProjectFile)"
         Targets="GenerateOpenApiCode"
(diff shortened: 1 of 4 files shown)


dougbu:
@BillHiebert I made a couple of space changes to the OpenApiItemsSchema.xaml file you provided and added the words "file" or "project" in a couple of spots. Also removed the conditions and the indirections. Is there anything I should test locally before we check this in?

src/Mvc/Extensions.ApiDescription.Client/src/buildMultiTargeting/Microsoft.Extensions.ApiDescription.Client.targets
```diff
@@ -1,5 +1,11 @@
 ﻿<?xml version="1.0" encoding="utf-8" standalone="no"?>
 <Project>
+  <ItemGroup>
```
BillHiebert: @dougbu  Will this cause the schema file to be imported twice? Are both of these targets files imported?

dougbu: No double import. The `build` targets file is imported only in inner builds i.e. when `TargetFrameworks` isn't set and the `buildMultiTargeting` one is imported only in outer builds. Basically the duplication ensures the items show up no matter what &hellip but only once 😄

dougbu:
Also kicked off a trial official build: https://dnceng.visualstudio.com/internal/_build/results?buildId=221516

dougbu:
@anurse we have a lot of `ServerTests` issues in aspnet/AspNetCore-Internal. But, `Server_MultipleOutstandingSyncRequests_Success` isn't specifically mentioned. Is it likely that Helix failure on Windows 10 AMD64 is just another example of flakiness in the test class?

dougbu:
Sorry to the many people now subscribed to this PR. GitHub added loads of code owners when I changed the target branch ☹️ I've pared things back again.

anurse:
@dougbu Yes, there are a few of these TaskCancelledException failures already tracked in RaaS. I think this is just the same.

dougbu:
Ping reviewers. All checks have passed.

BillHiebert (approved):

PR #10852, opened 2019-06-04
Mark ShutdownTestRun as Flaky

Tratcher:
https://github.com/aspnet/AspNetCore-Internal/issues/2577


analogrelay (approved):
Sad but true :)

Tratcher:
Which I can't merge because of BindModel_ReturnsModel_IfAttemptedValueIsValid_FrenchThirtyTwoThousandPointOne, which is being addressed in another PR. 
@aspnet/build admin merge override please?

anurse:
Or we can just get both flaky attributes into the same PR... let's just do that, it's easy and doesn't require an override :).

natemcmaster:
BindModel_ReturnsModel_IfAttemptedValueIsValid_FrenchThirtyTwoThousandPointOne was marked as flaky already in master, and merged while your PR validation was in-flight, so using the red button to merge :)

PR #10858, opened 2019-06-04
Guarantee that ANCM package dlls are used for back compat tests

jkotalik:
Fixes https://github.com/aspnet/AspNetCore-Internal/issues/2620. Also renames a few projects to avoid confusion.

TL;DR we have tests that use an older version of the AspNetCoreModule, which are copied to a bin folder eventually. These references can race with build outputs from the project build itself. We had a check to look a the full path to the dll and remove all ANCM references from the project build. This broke with https://github.com/aspnet/AspNetCore/commit/dc90e11c7aa0f88ffe2bbd729743e040f34ebb8b, which caused flakiness with copying, causing weird test failures like shown above.

@natemcmaster I'm not a fan at all of the solution I came up with. Fundamentally, this all stems from this specific test group referencing M.A.Servers.IntegrationTesting.IIS. Maybe I can conditionally depend on the native components in this specific project instead, such that this race never happens.


src/Servers/IIS/IIS/test/IIS.NewHandler.FunctionalTests/IIS.NewHandler.FunctionalTests.csproj
```diff
@@ -41,7 +38,12 @@
 
   <Target Name="RemoveSourceOverride" BeforeTargets="_CopyOutOfDateSourceItemsToOutputDirectory">
     <ItemGroup>
-      <SourceItemsToRemove Include="@(_SourceItemsToCopyToOutputDirectory)" Condition="$([System.String]::Copy(%(_SourceItemsToCopyToOutputDirectory.FullPath)).Contains('IIS\AspNetCoreModule')) == 'true'" />
+    <!-- There is a race between which dll is copied to the bin directory between ANCM in a package and the built ANCM. 
+        We want to guarantee the ANCM in the nuget package is used, so remove all paths that contain paths in the artifacts directory.
+        -->
+      <SourceItemsToRemove Include="@(_SourceItemsToCopyToOutputDirectory)" Condition="$([System.String]::Copy(%(_SourceItemsToCopyToOutputDirectory.FullPath)).Contains('AspNetCoreModuleShim')) == 'true'" />
```
natemcmaster: Why do these files end up in _SourceItemsToCopyToOutputDirectory in the first place? Could we prevent that from happening?

natemcmaster:
>  Maybe I can conditionally depend on the native components in this specific project instead, such that this race never happens.

Sounds reasonable to me.

src/Servers/IIS/IIS/test/IIS.NewHandler.FunctionalTests/IIS.NewHandler.FunctionalTests.csproj
jkotalik: The reason they are there is that we reference Microsoft.AspNetCore.Server.IntegrationTesting.IIS which includes native assets: https://github.com/aspnet/AspNetCore/blob/master/src/Servers/IIS/IntegrationTesting.IIS/src/Microsoft.AspNetCore.Server.IntegrationTesting.IIS.csproj#L30. We also reference `<PackageReference Include="Microsoft.AspNetCore.AspNetCoreModuleV2" Version="2.2.0">` above. These cause both to be copied to
_SourceItemsToCopyToOutputDirectory. I still would like to continue referencing Microsoft.AspNetCore.Server.IntegrationTesting.IIS, maybe I can set a property that causes those components not to be referenced.

jkotalik: @natemcmaster I think I came up with a better fix: https://github.com/aspnet/AspNetCore/pull/10858/files#diff-0ec7ab070eb9a3b8d1eac9206a279cddL30

BrennanConroy (approved):

jkotalik:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=213636

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/2629
https://github.com/aspnet/AspNetCore-Internal/issues/2572

PR #10880, opened 2019-06-05
Initial counter implementations

davidfowl:
- Added RPS counter - Requests per second, Total Requests, Current Requests and Failed Requests.
- Added EventSource tests that can run in parallel.
- Added IHttpCounters so that the diagnostic middleware can increment the failed request count without a direct reference to the hosting event source and without their own event source.

Here's how it looks:

image

Notes:
- I'm not in love with `IHttpCounters`, it feels kinda smelly as it needs to be a read/write interface so outside consumers can increment it (like the diagnostic middleware).
- The event source needs to be in the container now so that we can get the right instance of `IHttpCounters`
- In speaking to various people, we're going with the approach that each subsystem has its own set of counters. This means you'll need to explicitly enable them to see all of the ASP.NET counters. We won't have as much as logging categories though so it might scale just fine.

PS: If we can live without the diagnostic middleware incrementing the main failed request counter, it would be much simpler 😄. 

cc @shirhatti @DamianEdwards


davidfowl:
Just had an ephipany!

PR #10884, opened 2019-06-05
Added basic event counters for hosting

### LINKED ISSUE #9486: Create an RPS EventCounter
Let's add an RPS EventCounter in ASP.NET for use with the [dotnet-counters](https://github.com/dotnet/diagnostics/tree/master/src/Tools/dotnet-counters) tool.

Previous attempt at adding EventCounters to Hosting- https://github.com/aspnet/Hosting/pull/886

Relevant docs
- https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.Tracing/tests/BasicEventSourceTest/TestEventCounter.cs
- https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.Tracing/documentation/EventCounterTutorial.md

davidfowl:
- Added Request Rate counter (RPS), Total Requests counter, current requests and failed requests.
- Counters do nothing until enabled
- Added tests for counters and improved tests to use unique event source names
- Renamed EventSource to match the convention in the runtime Microsoft.AspNetCore.Hosting

Fixes #9486 

How it looks:

image

cc @shirhatti @DamianEdwards

### DIFF
--- a/src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs
+++ b/src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs
@@ -136,10 +136,19 @@ namespace Microsoft.AspNetCore.Hosting.Internal
                 StopActivity(httpContext, activity, context.HasDiagnosticListener);
             }
 
-            if (context.EventLogEnabled && exception != null)
+            if (context.EventLogEnabled)
             {
-                // Non-inline
-                HostingEventSource.Log.UnhandledException();
+                if (exception != null)
+                {
+                    // Non-inline
+                    HostingEventSource.Log.UnhandledException();
+                }
+
+                // Count 500 as failed requests
+                if (httpContext.Response.StatusCode >= 500)
+                {
+                    HostingEventSource.Log.RequestFailed();
+                }
             }
 
             // Logging Scope is finshed with
--- a/src/Hosting/Hosting/src/Internal/HostingEventSource.cs
+++ b/src/Hosting/Hosting/src/Internal/HostingEventSource.cs
@@ -1,17 +1,37 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using System.Diagnostics.Tracing;
 using System.Runtime.CompilerServices;
+using System.Threading;
 
 namespace Microsoft.AspNetCore.Hosting.Internal
 {
-    [EventSource(Name = "Microsoft-AspNetCore-Hosting")]
     public sealed class HostingEventSource : EventSource
     {
         public static readonly HostingEventSource Log = new HostingEventSource();
 
-        private HostingEventSource() { }
+        private IncrementingPollingCounter _requestsPerSecondCounter;
+        private PollingCounter _totalRequestsCounter;
+        private PollingCounter _failedRequestsCounter;
+        private PollingCounter _currentRequestsCounter;
+
+        private long _totalRequests;
+        private long _currentRequests;
+        private long _failedRequests;
+
+        internal HostingEventSource()
+            : this("Microsoft.AspNetCore.Hosting")
+        {
+
+        }
+
+        // Used for testing
+        internal HostingEventSource(string eventSourceName)
+            : base(eventSourceName)
+        {
+        }
 
         // NOTE
         // - The 'Start' and 'Stop' suffixes on the following event names have special meaning in EventSource. They
@@ -35,6 +55,8 @@ namespace Microsoft.AspNetCore.Hosting.Internal
         [Event(3, Level = EventLevel.Informational)]
         public void RequestStart(string method, string path)
         {
+            Interlocked.Increment(ref _totalRequests);
+            Interlocked.Increment(ref _currentRequests);
             WriteEvent(3, method, path);
         }
 
@@ -42,6 +64,7 @@ namespace Microsoft.AspNetCore.Hosting.Internal
         [Event(4, Level = EventLevel.Informational)]
         public void RequestStop()
         {
+            Interlocked.Decrement(ref _currentRequests);
             WriteEvent(4);
         }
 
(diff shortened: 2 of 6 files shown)


src/Hosting/Hosting/src/Internal/HostingEventSource.cs
```diff
@@ -51,5 +74,40 @@ public void UnhandledException()
...
         }
+
+        internal void RequestFailed()
+        {
+            Interlocked.Increment(ref _failedRequests);
+        }
+
+        protected override void OnEventCommand(EventCommandEventArgs command)
+        {
+            if (command.Command == EventCommand.Enable)
```
Tratcher: What about Disable?

Tratcher (commented):
Clean

src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs
```diff
@@ -136,10 +136,19 @@ public void RequestEnd(HttpContext httpContext, Exception exception, HostingAppl
                 StopActivity(httpContext, activity, context.HasDiagnosticListener);
             }
 
-            if (context.EventLogEnabled && exception != null)
+            if (context.EventLogEnabled)
```
anurse: First time I'm seeing this code, so I know this is already "how it's been" but `EventLog` != `EventSource` so this name struck me as odd. I backtracked to where it's initialized and confirmed it was right but the naming seems off.

It's internal so a rename would be safe if you want to be a little more consistent... 🤷‍♂

src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs
```diff
@@ -136,10 +136,19 @@ public void RequestEnd(HttpContext httpContext, Exception exception, HostingAppl
...
             {
-                // Non-inline
-                HostingEventSource.Log.UnhandledException();
+                if (exception != null)
+                {
+                    // Non-inline
+                    HostingEventSource.Log.UnhandledException();
```
anurse: Is there a reason we don't put exception details in the message?

src/Hosting/Hosting/src/Internal/HostingEventSource.cs
anurse: From below this line:

> `// They aren't disabled afterwards...`

Do you have more context on why?

src/Hosting/Hosting/src/Internal/HostingEventSource.cs
```diff
@@ -35,13 +55,16 @@ public void HostStop()
         [Event(3, Level = EventLevel.Informational)]
         public void RequestStart(string method, string path)
         {
+            Interlocked.Increment(ref _totalRequests);
```
anurse: This is cool stuff. Is it worth investigating using `ThreadLocal<int>` and summing the `.Values` in the counter delegate? That would save a bit of time in the hot path by not locking the processor cores. Though I'm probably optimizing a bit prematurely ;)

src/Hosting/Hosting/test/Internal/HostingEventSourceTests.cs
```diff
@@ -216,5 +263,36 @@ protected override void OnEventWritten(EventWrittenEventArgs eventData)
                 }
             }
         }
+
+        private class CounterListener : EventListener
```
anurse: This could be useful for testing all counters. We'll have to remember this when adding other counters and promote this to `src/Shared` (or you could just do it now...)

analogrelay (commented):
Swanky!

src/Hosting/Hosting/test/Internal/HostingEventSourceTests.cs
anurse: Or maybe add this to https://github.com/aspnet/Extensions/blob/master/src/TestingUtils/Microsoft.AspNetCore.Testing/src/Tracing/EventSourceTestBase.cs ?

src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs
davidfowl: 1. This isn't new code
2. You can't pass the exception itself (that isn't supported in EventSource writes), you could pass the text though.

src/Hosting/Hosting/src/Internal/HostingEventSource.cs
davidfowl: Yea I had a version with that though I'll do it after we measure this which I haven't since it involves turning on logging/eventsource which has other overhead. We also still need to interlocked.increment/decrement current requests which also made me go meh.

src/Hosting/Hosting/test/Internal/HostingEventSourceTests.cs
davidfowl: Yep. I have a thread with corefx about how to test these things. As I add more counters I'll move it to shared.

src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs
anurse: > 1. This isn't new code

Yes, I am aware.

src/Hosting/Hosting/src/Internal/HostingEventSource.cs
anurse: Measure first is best

src/Hosting/Hosting/src/Internal/HostingEventSource.cs
davidfowl: It's racy to disable counters and dispose and re-create them, that would need to be synchronized. The other way to turn it off is to stop pumping the data (that can be done via a different set of arguments), which turns the timer off so allocating the counter itself is fine.

src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs
davidfowl: It's public unfortunately. That's why I never renamed it. Can do it later, I think we should make all this stuff internal anyways.

Tratcher (approved):

PR #10901, opened 2019-06-05
Propagate cancellation token IAsyncEnumerators

mikaelm12:
Will hopefully fix: https://github.com/aspnet/AspNetCore-Internal/issues/2572
We weren't passing through a cancellation token passed into upload stream methods.

### DIFF
--- a/src/SignalR/common/Shared/AsyncEnumerableAdapters.cs
+++ b/src/SignalR/common/Shared/AsyncEnumerableAdapters.cs
@@ -45,7 +45,7 @@ namespace Microsoft.AspNetCore.SignalR.Internal
 
             public IAsyncEnumerator<TResult> GetAsyncEnumerator(CancellationToken cancellationToken = default)
             {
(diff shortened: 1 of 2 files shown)


src/SignalR/common/Shared/AsyncEnumerableAdapters.cs
```diff
@@ -45,15 +45,15 @@ public CancelableTypedAsyncEnumerable(IAsyncEnumerable<TResult> asyncEnumerable,
 
             public IAsyncEnumerator<TResult> GetAsyncEnumerator(CancellationToken cancellationToken = default)
             {
-                var enumerator = _asyncEnumerable.GetAsyncEnumerator();
+                var enumerator = _asyncEnumerable.GetAsyncEnumerator(cancellationToken);
```
BrennanConroy: I think this should be `_cts`

mikaelm12: It definitely should

BrennanConroy (approved):

PR #10904, opened 2019-06-05
fix #7548 by updating to v142 toolset (again)

### LINKED ISSUE #7548: Upgrade IIS C++ projects to build with platformtoolset v142 (VS2019)

analogrelay:
This time, it's personal.

Trying again to update to v142. We'll see how this goes ;).

We can probably also remove the v141 toolset from the vs.json files as a result of this, but I'd like to wait until we know this works to do that :). If we get a green check, we can do a second pass to clean-up (in this PR, but just after we know it works)

Before merge checklist:
* [x] Remove v141 toolset install from `korebuild.json`, `eng\scripts\vs.buildtools.json` and `eng\scripts\vs.json` files

### DIFF
--- a/src/Servers/IIS/AspNetCoreModuleV2/CommonLib/HandleWrapper.cpp
+++ b/src/Servers/IIS/AspNetCoreModuleV2/CommonLib/HandleWrapper.cpp
@@ -4,5 +4,6 @@
 #include "HandleWrapper.h"
 
 // Workaround for VC++ bug https://developercommunity.visualstudio.com/content/problem/33928/constexpr-failing-on-nullptr-v141-compiler-regress.html
+// REVIEW: This is fixed maybe? We could try removing it now that we're on v142. Is there test coverage?
 const HANDLE InvalidHandleTraits::DefaultHandle = INVALID_HANDLE_VALUE;
 const HANDLE FindFileHandleTraits::DefaultHandle = INVALID_HANDLE_VALUE;
--- a/src/Servers/IIS/AspNetCoreModuleV2/CommonLib/HandleWrapper.cpp
+++ b/src/Servers/IIS/AspNetCoreModuleV2/CommonLib/HandleWrapper.cpp
@@ -1,9 +1,0 @@
-// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
-
-#include "HandleWrapper.h"
-
-// Workaround for VC++ bug https://developercommunity.visualstudio.com/content/problem/33928/constexpr-failing-on-nullptr-v141-compiler-regress.html
-// REVIEW: This is fixed maybe? We could try removing it now that we're on v142. Is there test coverage?
-const HANDLE InvalidHandleTraits::DefaultHandle = INVALID_HANDLE_VALUE;
-const HANDLE FindFileHandleTraits::DefaultHandle = INVALID_HANDLE_VALUE;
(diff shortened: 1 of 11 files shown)


natemcmaster:
I believe this means we can also delete these two lines: https://github.com/aspnet/AspNetCore/blob/d1138f2cdeda8e04755ffec138b9514d0ffd43a8/korebuild.json#L19-L20

natemcmaster (approved):
If it works, :shipit:

cc @dougbu - this would unblock https://github.com/aspnet/AspNetCore-Internal/issues/2538

anurse:
After this change, nothing should be on toolset v142 (we do have it in our VS install scripts and an unfortunate coincidence in some SVG files ;))

image

jkotalik:
> we do have it in our VS install scripts 

Which we can remove now?

src/Servers/IIS/AspNetCoreModuleV2/CommonLib/HandleWrapper.h
```diff
@@ -9,7 +9,7 @@
 struct InvalidHandleTraits
 {
     using HandleType = HANDLE;
-    static const HANDLE DefaultHandle;
+    static constexpr HANDLE DefaultHandle = nullptr;
```
anurse: Being optimistic that the [compiler bug that prevented this](https://developercommunity.visualstudio.com/content/problem/33928/constexpr-failing-on-nullptr-v141-compiler-regress.html) is fixed now. We'll see :).

src/Servers/IIS/AspNetCoreModuleV2/CommonLib/HandleWrapper.h
```diff
@@ -35,16 +35,16 @@ struct ModuleHandleTraits
 };
 
 // Code analysis doesn't like nullptr usages via traits
-#pragma warning( push )
-#pragma warning ( disable : 26477 ) // disable  Use 'nullptr' rather than 0 or NULL (es.47).
+#pragma warning(push)
+#pragma warning(disable : 26477) // disable  Use 'nullptr' rather than 0 or NULL (es.47).
```
anurse: VSCode wanted to reformat all the things.

src/Servers/IIS/AspNetCoreModuleV2/CommonLib/HandleWrapper.cpp
```diff
@@ -4,5 +4,6 @@
 #include "HandleWrapper.h"
 
 // Workaround for VC++ bug https://developercommunity.visualstudio.com/content/problem/33928/constexpr-failing-on-nullptr-v141-compiler-regress.html
+// REVIEW: This is fixed maybe? We could try removing it now that we're on v142. Is there test coverage?
```
jkotalik: If this is fixed yeah. I think it will fail to build if this isn't fixed, so we have "test coverage"

jkotalik (approved):
Hoping for the best.

anurse:
image

image

image

image

src/Servers/IIS/AspNetCoreModuleV2/CommonLib/HandleWrapper.cpp
anurse: Yep, I realized it was literally a compiler bug, so it felt safe to try :)

anurse:
image

Success!

anurse:
Going to push to an AzDO branch to confirm it's all good on official agents too.

anurse:
https://dev.azure.com/dnceng/internal/_build/results?buildId=213480 is the official build I kicked off, for reference.

natemcmaster:
🎉 nice!

jkotalik:
@anurse just to confirm, you did verify this locally, correct?

anurse:
`./build -BuildNative` is happy on my box. I still have the `v141` toolset on though (need it for other things). Both the test official and PR validation builds worked.

I'll leave it for @jkotalik to hit the final merge button here since I will be much less available over the next couple days and don't want to merge and bail ;). At least if @jkotalik merges he knows what to revert!

jkotalik:
I’ll take liability.

PR #10907, opened 2019-06-05
Overhead Benchmark for Request Throttling

DylanDmitri:
There's a benchmark project!

The single included benchmark captures the overhead of requests passing directly through the middleware (which happens the server has extra room). The baseline represents calling the rest of the pipeline directly without passing through the middleware. Comparing these two numbers, we can get an estimate for the overhead.

|   Method  |     Mean |     Error |    StdDev |      Op/s |  Gen 0 | Allocated |
|------------|----------:|----------:|----------:|----------:|-------:|----------:|
| Baseline | 1.697 us | 0.0337 us | 0.0904 us | 589,400.1 | 3.9063 |     296 B |
 WithEmptyQueueOverhead |  2.080 us | 0.0556 us | 0.1585 us | 480,863.7 | 7.8125 |     440 B |


src/Middleware/RequestThrottling/perf/Microbenchmarks/Overhead.cs
```diff
@@ -0,0 +1,69 @@
...
+            // this one should not change with MaxConcurrentRequests
+            for (int i = 0; i < _numRequests; i++)
+            {
+                await _middleware.Invoke(null);
+            }
+        }
+
+        private async Task NextDelay(HttpContext context)
+        {
+            for (int i = 0; i < 1000; i++) { }
```
jkotalik: I think you need a better way to represent work. spoopy.

DylanDmitri:
Current runs:

|Method | YieldsThreadInternally |         Mean |      Error |     StdDev |          Op/s |    Gen 0 | Allocated |
|------------ |----------------------- |-------------:|-----------:|-----------:|--------------:|---------:|----------:|
| Baseline |              False |     4.599 ns |  0.0968 ns |  0.2145 ns | 217,423,611.1 |        - |       0 B |
|WithEmptyQueueOverhead |                      False |   158.331 ns |  3.1055 ns |  5.1025 ns |   6,315,868.6 |        - |       0 B |
|Baseline |                        True | 1,154.286 ns | 22.1071 ns | 24.5720 ns |     866,336.0 |  31.2500 |     296 B |
| WithEmptyQueueOverhead |                        True | 1,596.948 ns | 32.8273 ns | 92.5899 ns |     626,194.5 | 125.0000 |       0 B |

DylanDmitri:
@halter73 am I good to merge here? 

Also I've snuck a `IRequestQueue` interface into here, to support different queue implementations in the future. It shouldn't technically be in this PR sorry.

src/Middleware/RequestThrottling/perf/Microbenchmarks/QueueFullOverhead.cs
```diff
@@ -0,0 +1,79 @@
...
+    public class QueueFullOverhead
+    {
+        private const int _numRequests = 2000;
+        private int _requestCount = 0;
+        private TaskCompletionSource<bool> tcs;
+
+        private RequestThrottlingMiddleware _middleware;
+
+        [Params(1,8,64)]
+        public int MaxConcurrentRequests;
```
halter73: This variation applies to all scenarios. Only put scenarios that vary on MaxConcurrentRequests in this class.

halter73: If this is flaky with variations, remove the variations for now.

src/Middleware/RequestThrottling/perf/Microbenchmarks/QueueFullOverhead.cs
```diff
@@ -0,0 +1,79 @@
...
+                    loggerFactory: NullLoggerFactory.Instance,
+                    options: Options.Create(options)
+                );
+        }
+
+        [IterationSetup]
+        public void Setup()
+        {
+            _requestCount = 0;
+            tcs = new TaskCompletionSource<bool>();
```
halter73: Nit: we usually use `object` as `T` when it doesn't matter

If this isn't being awaited, we can just use a ManualResetEventSlim, then you don't need to allocate in Setup().

src/Middleware/RequestThrottling/src/Internal/TailDrop.cs
```diff
@@ -0,0 +1,62 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.RequestThrottling.Internal
+{
+    internal class TailDrop : RequestQueue
```
halter73: We shouldn't just change the default to TailDrop. Keep all the existing tests fro the base RequestQueue and add special tests for TailDrop.

src/Middleware/RequestThrottling/src/Internal/RequestQueue.cs
```diff
@@ -1,62 +1,17 @@
...
 using System;
+using System.Collections.Generic;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 
 namespace Microsoft.AspNetCore.RequestThrottling.Internal
 {
-    internal class RequestQueue : IDisposable
+    interface RequestQueue : IDisposable
```
halter73: IRequestQueue?

IRequestQueue.cs

src/Middleware/RequestThrottling/src/Internal/RequestQueue.cs
```diff
@@ -1,62 +1,17 @@
-// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
```
halter73: Still need this too

src/Middleware/RequestThrottling/src/Internal/RequestQueue.cs
```diff
@@ -1,62 +1,17 @@
...
-
-            await _serverSemaphore.WaitAsync();
-
-            return true;
-        }
-
-        public void Release()
-        {
-            _serverSemaphore.Release();
+        public int TotalRequests { get; }
```
halter73: You can't put public in interface members

halter73: I guess you can... but don't.

src/Middleware/RequestThrottling/test/TestUtils.cs
```diff
@@ -16,10 +15,12 @@ public static RequestThrottlingMiddleware CreateTestMiddleware(int? maxConcurren
         {
             var options = new RequestThrottlingOptions
             {
-                MaxConcurrentRequests = maxConcurrentRequests,
+                MaxConcurrentRequests = maxConcurrentRequests == 0 ? 999 : maxConcurrentRequests,
                 RequestQueueLimit = requestQueueLimit
             };
 
+            options.ServerAlwaysBlocks = maxConcurrentRequests == 0;
```
halter73: This check should be in options or the middleware. Right now this only work if you use TestUtils.

halter73: I think I understand now. What you have is fine, just use a different method for the ServerAlwaysBlocks case.

src/Middleware/RequestThrottling/perf/Microbenchmarks/QueueFullOverhead.cs
DylanDmitri: reduced to only one param, this also helps with flaky tests

src/Middleware/RequestThrottling/src/Internal/TailDrop.cs
DylanDmitri: Hopefully will have multiple queue implementations soon, will run the existing tests on all of them.

src/Middleware/RequestThrottling/perf/Microbenchmarks/QueueFullOverhead.cs
```diff
@@ -0,0 +1,79 @@
...
+using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.AspNetCore.RequestThrottling.Microbenchmarks
+{
+    public class QueueFullOverhead
+    {
+        private const int _numRequests = 2000;
+        private int _requestCount = 0;
+        private ManualResetEventSlim mres = new ManualResetEventSlim();
```
halter73:
```suggestion
        private ManualResetEventSlim _mres = new ManualResetEventSlim();
```

src/Middleware/RequestThrottling/sample/Startup.cs
```diff
@@ -15,6 +15,8 @@ namespace RequestThrottlingSample
 {
     public class Startup
     {
+        private int totalLoads = 0;
```
halter73:
```suggestion
        private int _totalLoads = 0;
```

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
```diff
@@ -51,9 +51,15 @@ public RequestThrottlingMiddleware(RequestDelegate next, ILoggerFactory loggerFa
 
             _next = next;
             _logger = loggerFactory.CreateLogger<RequestThrottlingMiddleware>();
-            _requestQueue = new RequestQueue(
-                _requestThrottlingOptions.MaxConcurrentRequests.Value,
-                _requestThrottlingOptions.RequestQueueLimit);
+
+            if (_requestThrottlingOptions.ServerAlwaysBlocks)
```
halter73: Nit: comment here too that this is for testing only

halter73 (approved):

DylanDmitri:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=220646&view=logs&jobId=f2f63060-d9d6-52d0-adee-b97db5a9ab91

aspnet-hello:

PR #10914, opened 2019-06-05
Fix a write past the end of buffer in IntegerEncoder

### LINKED ISSUE #10913: HPACK multi-byte integer encoding writes past end of buffer
The HPACK integer encoder will write past the end of a too-short buffer when encoding an integer into multiple bytes, on line 47 below:

https://github.com/aspnet/AspNetCore/blob/9cd2ad15ff2a306bab9894dc7c1926c196b3fa2e/src/Servers/Kestrel/Core/src/Internal/Http2/HPack/IntegerEncoder.cs#L36-L47

scalablecory:
Resolves #10913

Fixes a write past end of buffer when an integer requires multiple bytes to encode, and the buffer does not have enough space. Adds a test to confirm this.
Add some basic asserts to constrain parameters to HPACK spec.
Add some additional tests.

### DIFF
--- a/src/Servers/Kestrel/Core/test/HPackIntegerTests.cs
+++ b/src/Servers/Kestrel/Core/test/HPackIntegerTests.cs
@@ -1,6 +1,8 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
+using System.Collections.Generic;
 using System.Linq;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2.HPack;
 using Xunit;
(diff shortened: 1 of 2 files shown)


scalablecory:
CI failures are in unrelated code.

src/Servers/Kestrel/Core/test/HPackIntegerTests.cs
```diff
@@ -35,5 +37,55 @@ public void IntegerEncoderDecoderRoundtrips()
                 }
             }
         }
+
+        [Theory]
+        [MemberData(nameof(IntegerCodecSamples))]
+        public void EncodeSamples(int value, int bits, byte[] expectedResult)
+        {
+            Span<byte> actualResult = new byte[64];
```
jkotalik: nit: I think this and `DecodeSamples` are a bit redundant with IntegerEncoderDecoderRoundtrips. I believe the combinations you listed would eventually be tested. the only difference is that you are verifying the expectedResult.

jkotalik (approved):

jkotalik:
Thanks!

Issue #7734, opened 2019-02-20
Consider adding Code + PKCE support in the OpenID Connect handler


leastprivilege:
Traditional code flow in OAuth 2 has the code substitution problem (https://tools.ietf.org/id/draft-ietf-oauth-security-topics-06.html#rfc.section.3.5). This was one of the motivations to create the OIDC hybrid flow (aka code id_token), where the id_token contains the hash of the code (aka detached signature) to protect against this attack. The downside of this approach is, that an id_token must be sent over the front channel.

PKCE (Proof Key for Code Exchange) solves the same problem in a different and simpler way. The spec (https://tools.ietf.org/html/rfc7636) reads like this is only useful for native client applications, but it turned out it is useful in many more cases.

The advantages of using PKCE are:

* can go back to code response type which means no id_token must be sent over the front-channel which eliminates privacy concerns (e.g. leaking the subject id over the browser)
* since there is no id_token on the front channel, no JWT validation must be done
* since the id_token comes over the HTTPS  back-channel only, JWT validation is optional
* the client implementation is considerably simpler

You might have followed the latest discussions around OIDC/OAuth security considerations and also the SPA guidance. There is a movement to simplify the process around selecting the right flows - basically it boils down to 

* use client credentials for non-interactive server to server
* use code + PKCE for the rest

It is not hard to add this feature today - I created a POC here:

https://github.com/leastprivilege/AspNetCoreSecuritySamples/tree/aspnetcore21/OidcCodePkce

especially 

https://github.com/leastprivilege/AspNetCoreSecuritySamples/blob/aspnetcore21/OidcCodePkce/src/AspNetCoreSecurity/Startup.cs#L77-L118

Something to consider for 3.0

Eilon:
Adding to the list of potential features for 3.1.

@jmprieur - out of curiosity, if we had this, is this something that would work on AzureAD?

brockallen:
you're welcome @leastprivilege

jmprieur:
@Eilon :MSAL.NET uses it in desktop/mobile apps. Adding @brentschmaltz for his opinion for Web apps.

Tratcher:
The OAuth base class should also support this. See #9448, #9092

brentschmaltz:
This seems like a valuable feature.
For the OIDC handler, the sample that @leastprivilege provided shows that a simple property OpenIdConnectOptions should trigger the scenario.

[...]

PR #10928, opened 2019-06-05
Add PKCE support in OIDC & OAuth

Tratcher:
#7734 (Best reviewed with Ignore Whitespace due to formatting improvements).

I've also cleaned up the samples quite a bit and added IdentityServer to them.

Considerations:
- For OIDC how important is it to scope PKCE to only code flow? Why not use it for everything? @brockallen @leastprivilege 
- I've enabled this by default for OIDC. Tested with ADD and IdentityServer. Servers that don't support it should silently ignore it.
- I have not enabled it by default for OAuth. Only MSA and IdentityServer supported it, and OAuth servers tend to be far less standards compliant. Individual providers can opt-in.
- This required an API breaking change for OAuth and we know of at least 9 providers in aspnet-contrib that will have to update. @PinpointTownes 
  - Note all of these providers were already broken by the Json changes, the update is still in progress: https://github.com/aspnet-contrib/AspNet.Security.OAuth.Providers/pull/280/files
  - https://github.com/aspnet/AspNetCore/pull/9448/files will no longer require a breaking change.
- This doubles the size of the State for OAuth and OIDC, which may contribute to long url issues.

### DIFF
--- a/src/Security/Authentication/Core/src/RemoteAuthenticationHandler.cs
+++ b/src/Security/Authentication/Core/src/RemoteAuthenticationHandler.cs
@@ -186,6 +186,17 @@ namespace Microsoft.AspNetCore.Authentication
         protected override Task HandleForbiddenAsync(AuthenticationProperties properties)
             => Context.ForbidAsync(SignInScheme);
 
+        /// <summary>
+        /// Creates a 32 bit random Id and Base64Url encodes it.
+        /// </summary>
+        /// <returns></returns>
+        protected virtual string GenerateUniqueId()
+        {
+            var bytes = new byte[32];
+            CryptoRandom.GetBytes(bytes);
+            return Base64UrlTextEncoder.Encode(bytes);
+        }
+
         protected virtual void GenerateCorrelationId(AuthenticationProperties properties)
         {
             if (properties == null)
@@ -193,9 +204,7 @@ namespace Microsoft.AspNetCore.Authentication
                 throw new ArgumentNullException(nameof(properties));
             }
 
-            var bytes = new byte[32];
-            CryptoRandom.GetBytes(bytes);
-            var correlationId = Base64UrlTextEncoder.Encode(bytes);
+            var correlationId = GenerateUniqueId();
 
             var cookieOptions = Options.CorrelationCookie.Build(Context, Clock.UtcNow);
 
--- a/src/Security/Authentication/Core/src/RemoteAuthenticationHandler.cs
+++ b/src/Security/Authentication/Core/src/RemoteAuthenticationHandler.cs
@@ -186,17 +186,6 @@ namespace Microsoft.AspNetCore.Authentication
         protected override Task HandleForbiddenAsync(AuthenticationProperties properties)
             => Context.ForbidAsync(SignInScheme);
 
-        /// <summary>
-        /// Creates a 32 bit random Id and Base64Url encodes it.
-        /// </summary>
-        /// <returns></returns>
-        protected virtual string GenerateUniqueId()
-        {
-            var bytes = new byte[32];
-            CryptoRandom.GetBytes(bytes);
-            return Base64UrlTextEncoder.Encode(bytes);
-        }
-
         protected virtual void GenerateCorrelationId(AuthenticationProperties properties)
         {
             if (properties == null)
@@ -204,7 +193,9 @@ namespace Microsoft.AspNetCore.Authentication
                 throw new ArgumentNullException(nameof(properties));
             }
 
-            var correlationId = GenerateUniqueId();
+            var bytes = new byte[32];
+            CryptoRandom.GetBytes(bytes);
+            var correlationId = Base64UrlTextEncoder.Encode(bytes);
 
             var cookieOptions = Options.CorrelationCookie.Build(Context, Clock.UtcNow);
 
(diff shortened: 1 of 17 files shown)


src/Security/Authentication/MicrosoftAccount/src/MicrosoftAccountOptions.cs
```diff
@@ -21,6 +21,7 @@ public MicrosoftAccountOptions()
             AuthorizationEndpoint = MicrosoftAccountDefaults.AuthorizationEndpoint;
             TokenEndpoint = MicrosoftAccountDefaults.TokenEndpoint;
             UserInformationEndpoint = MicrosoftAccountDefaults.UserInformationEndpoint;
+            UsePkse = true;
```
HaoK: Is this supposed to be PKCE?  (Just going off title of issue)

brockallen:
> For OIDC how important is it to scope PKCE to only code flow? Why not use it for everything? 

implicit there is no use of the token endpoint, thus no need.
hybrid uses the nonce and c_hash from the spec as the protection, so that's redundant (albeit more complicated for the client).

src/Security/Authentication/MicrosoftAccount/src/MicrosoftAccountOptions.cs
Tratcher: Yes 😁

PinpointTownes:
> This required an API breaking change for OAuth and we know of at least 9 providers in aspnet-contrib that will have to update. @PinpointTownes

@martincostello more breaking changes for your PR 😄 

> I have not enabled it by default for OAuth. Only MSA and IdentityServer supported it, and OAuth servers tend to be far less standards compliant. Individual providers can opt-in.

Sounds safer, indeed 😄

src/Security/Authentication/Core/src/RemoteAuthenticationHandler.cs
```diff
@@ -186,16 +186,25 @@ protected override async Task<AuthenticateResult> HandleAuthenticateAsync()
         protected override Task HandleForbiddenAsync(AuthenticationProperties properties)
             => Context.ForbidAsync(SignInScheme);
 
+        /// <summary>
+        /// Creates a 32 bit random Id and Base64Url encodes it.
+        /// </summary>
+        /// <returns></returns>
+        protected virtual string GenerateUniqueId()
```
halter73: Who overrides this? Is there a test that overrides this?

Tratcher: Nobody at the moment, but this base class is intended to allow as much extensibility as is practical.

HaoK: Its probably worth considering just moving it all the way up to AuthenticationHandler then, as its just a general purpose helper at that point

src/Security/Authentication/OAuth/src/Events/OAuthCodeExchangeContext.cs
```diff
@@ -0,0 +1,39 @@
...
+
+namespace Microsoft.AspNetCore.Authentication.OAuth
+{
+    /// <summary>
+    /// Contains information used to perform the code exchange.
+    /// </summary>
+    public class OAuthCodeExchangeContext
+    {
+        /// <summary>
+        /// Initializes a new <see cref="OAuthCreatingTicketContext"/>.
```
HaoK: Nit: wrong class

src/Security/Authentication/OAuth/src/OAuthConstants.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    /// <summary>
+    /// Constants used in the OAuth protocol
+    /// </summary>
+    public static class OAuthConstants
+    {
+        /// <summary>
+        /// code_verifier defined in https://tools.ietf.org/html/rfc7636
+        /// </summary>
+        public static readonly string CodeVerifierKey = "code_verifier";
```
HaoK: Nit: should these be const, they cannot ever change right?

src/Security/Authentication/OpenIdConnect/samples/OpenIdConnectSample/Startup.cs
```diff
@@ -45,16 +45,22 @@ public void ConfigureServices(IServiceCollection services)
...
+                o.ClientSecret = "secret"; // for code flow
+                o.Authority = "https://demo.identityserver.io/";
 
                 o.ResponseType = OpenIdConnectResponseType.CodeIdToken;
                 o.SaveTokens = true;
                 o.GetClaimsFromUserInfoEndpoint = true;
                 o.AccessDeniedPath = "/access-denied-from-remote";
 
-                o.ClaimActions.MapAllExcept("aud", "iss", "iat", "nbf", "exp", "aio", "c_hash", "uti", "nonce");
+                // o.ClaimActions.MapAllExcept("aud", "iss", "iat", "nbf", "exp", "aio", "c_hash", "uti", "nonce");
```
HaoK: need to clean up sample?  Remove commented out lines?

src/Security/Authentication/Core/src/RemoteAuthenticationHandler.cs
brockallen: There are specific requirements around it for PKCE, so not sure a general unique id base API would always work. FWIW.

HaoK (approved):

src/Security/Authentication/Core/src/RemoteAuthenticationHandler.cs
HaoK: If that's the case maybe add a specific GeneratePKCEId method in OAuthHandler

Tratcher: I had generalized this one because it was already fulfilling the PKCE requirements. I can revert that part and duplicate the code to be more explicit.

src/Security/Authentication/OAuth/src/OAuthConstants.cs
Tratcher: As a rule we only use public const where it's required for use in attributes.

src/Security/Authentication/OpenIdConnect/samples/OpenIdConnectSample/Startup.cs
Tratcher: I wanted to keep them so it would be easy to switch back and forth.

halter73 (approved):

PR #10929, opened 2019-06-05
Lazy-init formatter resolvers in MessagePackHubProtocolOptions

### LINKED ISSUE #10309: Lazy-initialize MessagePackHubProtocolOptions
We should lazy-init the formatter resolvers in MessagePackHubProtocolOptions. The default set of resolvers trigger a static constructor that throws on AOT environments (IL2CPP for Unity for example). If we lazy init it, an AOT-based app can call the setter to replace the list of formatters with an AOT-friendly set without risking an exception due to trying to initialize the non-AOT-friendly ones.

mikaelm12:
Fixes: https://github.com/aspnet/AspNetCore/issues/10309

### DIFF
--- a/src/SignalR/common/Protocols.MessagePack/src/MessagePackHubProtocolOptions.cs
+++ b/src/SignalR/common/Protocols.MessagePack/src/MessagePackHubProtocolOptions.cs
@@ -9,6 +9,23 @@ namespace Microsoft.AspNetCore.SignalR
 {
     public class MessagePackHubProtocolOptions
     {
-        public IList<IFormatterResolver> FormatterResolvers { get; set; } = MessagePackHubProtocol.CreateDefaultFormatterResolvers();
+        private IList<IFormatterResolver> _formatterResolvers;
+
+        public IList<IFormatterResolver> FormatterResolvers
+        {
+            get
+            {
+                if (_formatterResolvers == null)
+                {
+                    _formatterResolvers = MessagePackHubProtocol.CreateDefaultFormatterResolvers();
+                }
+
+                return _formatterResolvers;
+            }
+            set
+            {
+                _formatterResolvers = value;
+            }
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


src/SignalR/common/Protocols.MessagePack/src/MessagePackHubProtocolOptions.cs
```diff
@@ -9,6 +9,23 @@ namespace Microsoft.AspNetCore.SignalR
...
-        public IList<IFormatterResolver> FormatterResolvers { get; set; } = MessagePackHubProtocol.CreateDefaultFormatterResolvers();
+        private IList<IFormatterResolver> _formatterResolvers;
+
+        public IList<IFormatterResolver> FormatterResolvers
+        {
+            get
+            {
+                if (_formatterResolvers == null)
+                {
+                    _formatterResolvers = MessagePackHubProtocol.CreateDefaultFormatterResolvers();
```
BrennanConroy: Lets add a comment explaining why we're doing this

BrennanConroy (approved):

src/SignalR/common/Protocols.MessagePack/src/MessagePackHubProtocolOptions.cs
```diff
@@ -9,6 +9,25 @@ namespace Microsoft.AspNetCore.SignalR
...
+                    // The default set of resolvers trigger a static constructor that throws on AOT environments.
+                    // This gives users the chance to use an AOT friendly formatter.
+                    _formatterResolvers = MessagePackHubProtocol.CreateDefaultFormatterResolvers();
+                }
+
+                return _formatterResolvers;
+            }
+            set
+            {
+                _formatterResolvers = value;
```
halter73: I'm assuming there's no reason that anyone would actually want to set this to null, right?

halter73 (approved):

mikaelm12:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=213777&view=logs

src/SignalR/common/Protocols.MessagePack/src/MessagePackHubProtocolOptions.cs
mikaelm12: > I'm assuming there's no reason that anyone would actually want to set this to null, right?

Right. I don't think that's a use case we'll see.

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/2614
https://github.com/aspnet/AspNetCore-Internal/issues/2629

PR #10932, opened 2019-06-05
Localization handback for missing strings

joeloff:
- Updated handback that includes some strings that were missed during the first round of localization


jkotalik (approved):

joeloff:
As discussed with @Pilchie we'll take this for Preview 7 and disable the installer loc UI for Preview 6 instead.

PR #10934, opened 2019-06-05
Update Helix Sd

JunTaoLuo:
Addresses https://github.com/aspnet/AspNetCore-Internal/issues/2482

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -392,17 +392,17 @@
       <Uri>https://github.com/aspnet/Extensions</Uri>
       <Sha>ef5d488c054c3aa850c29c10d851fcdc8a004eed</Sha>
     </Dependency>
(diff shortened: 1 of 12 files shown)


natemcmaster (commented):
I'm not sure it completely addresses https://github.com/aspnet/AspNetCore-Internal/issues/2482 yes. We need to make changes to runtests.cmd/sh to re-enable the test reporting for flaky tests, right?

JunTaoLuo:
Yea. I think I'll also need to tell the team about the workflow of making tests as "Flaky" on AzDO. Not a big fan of it so I want to talk to someone to see if there's something better or plans to improve it. I'll update this once I have more.

JunTaoLuo:
Turns out it's not recommended to use the Flaky button. I guess we'll need wait for another mechanism to deal with flaky tests.

PR #10937, opened 2019-06-06
[Static Web Assets] Various fixes

javiercn:
* Remove support for the embedded manifest.
* Move Static Web Assets into their own package.


Tratcher (approved):

eng/SharedFramework.Local.props
```diff
@@ -29,6 +29,7 @@
     <AspNetCoreAppReference Include="Microsoft.AspNetCore.Hosting.Abstractions" />
     <AspNetCoreAppReference Include="Microsoft.AspNetCore.Hosting" />
     <AspNetCoreAppReference Include="Microsoft.AspNetCore.Hosting.Server.Abstractions" />
+    <AspNetCoreAppReference Include="Microsoft.AspNetCore.StaticWebAssets" />
```
dougbu: Appreciate a bit more context:
- Why do we need a new assembly for just 5 classes?
- Have we got agreement on adding this assembly to the shared framework? (I assume yes because the classes were already in Microsoft.AspNetCore.App but want to confirm.)

src/Hosting/StaticWebAssets/src/Microsoft.AspNetCore.StaticWebAssets.csproj
```diff
@@ -0,0 +1,22 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <Description>ASP.NET Core infrastructure for static web assets.</Description>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <IsAspNetCoreApp>true</IsAspNetCoreApp>
+    <NoWarn>$(NoWarn);CS1591</NoWarn>
```
natemcmaster:
```suggestion
```

We should try to eliminate this NoWarn. Docs are a good thing and we should know where we have undocumented API. For all new API, I recommend removing this suppression.

eng/SharedFramework.Local.props
natemcmaster: We don't really have guidance on when to create a new assembly or not, so it's sort of a gut thing. Historically, the approach has been to create lots and lots of new assemblies, but I'm not sure if that's a practice we should continue in .NET Core 3. @davidfowl typically helps us make decisions on these kinds of design choices, so would like his input on this.

That said, I don't see a compelling reason to create a new assembly, especially given that this assembly is shared-framework only. @javiercn can you share your thinking on why this API needs to be split out from Microsoft.AspNetCore.dll?

Tratcher: It wasn't supposed to go into Microsoft.AspNetCore.dll in the first place. That library only contains our opinionated template APIs.

javiercn: What @Tratcher said.

davidfowl: Why not put it into hosting?

Tratcher: An alternative home was suggested in Microsoft.AspNetCore.Hosting.

javiercn: If anything, being on the shared framework makes it less important to be its own assembly. We have plenty of assemblies that are like this (for example user secrets) and I wouldn't like to put a feature that is meant to be used only during development into hosting.

davidfowl: Put it in Hosting.

javiercn: Hosting it its.

PR #10946, opened 2019-06-06
Inject ILogger to RequestLocalization Middleware

Kahbazi:
https://github.com/aspnet/AspNetCore/issues/6532

### DIFF
--- a/src/Middleware/Localization/src/RequestLocalizationMiddleware.cs
+++ b/src/Middleware/Localization/src/RequestLocalizationMiddleware.cs
@@ -43,7 +43,7 @@ namespace Microsoft.AspNetCore.Localization
             }
 
             _next = next ?? throw new ArgumentNullException(nameof(next));
-            _logger = loggerFactory?.CreateLogger<RequestLocalizationMiddleware>() ?? throw new ArgumentNullException(nameof(next));
+            _logger = loggerFactory?.CreateLogger<RequestLocalizationMiddleware>() ?? throw new ArgumentNullException(nameof(loggerFactory));
             _options = options.Value;
         }
 
--- a/src/Middleware/Localization/src/RequestLocalizationMiddleware.cs
+++ b/src/Middleware/Localization/src/RequestLocalizationMiddleware.cs
@@ -11,6 +11,7 @@ using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Localization.Internal;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
 using Microsoft.Extensions.Options;
 using Microsoft.Extensions.Primitives;
 
@@ -26,7 +27,7 @@ namespace Microsoft.AspNetCore.Localization
 
         private readonly RequestDelegate _next;
         private readonly RequestLocalizationOptions _options;
-        private readonly ILogger<RequestLocalizationMiddleware> _logger;
+        private readonly ILogger _logger;
 
         /// <summary>
         /// Creates a new <see cref="RequestLocalizationMiddleware"/>.
@@ -35,6 +36,7 @@ namespace Microsoft.AspNetCore.Localization
         /// <param name="options">The <see cref="RequestLocalizationOptions"/> representing the options for the
         /// <param name="loggerFactory">The <see cref="ILoggerFactory"/> used for logging.</param>
         /// <see cref="RequestLocalizationMiddleware"/>.</param>
+        [ActivatorUtilitiesConstructor]
         public RequestLocalizationMiddleware(RequestDelegate next, IOptions<RequestLocalizationOptions> options, ILoggerFactory loggerFactory)
         {
             if (options == null)
@@ -44,6 +46,25 @@ namespace Microsoft.AspNetCore.Localization
 
             _next = next ?? throw new ArgumentNullException(nameof(next));
             _logger = loggerFactory?.CreateLogger<RequestLocalizationMiddleware>() ?? throw new ArgumentNullException(nameof(loggerFactory));
+            _options = options.Value;
+        }
+
+        /// <summary>
+        /// Creates a new <see cref="RequestLocalizationMiddleware"/>.
+        /// </summary>
+        /// <param name="next">The <see cref="RequestDelegate"/> representing the next middleware in the pipeline.</param>
+        /// <param name="options">The <see cref="RequestLocalizationOptions"/> representing the options for the
+        /// <see cref="RequestLocalizationMiddleware"/>.</param>
+        [Obsolete("This is obsolete and will be removed in a future version. Use RequestLocalizationMiddleware(RequestDelegate next, IOptions<RequestLocalizationOptions> options, ILoggerFactory loggerFactory) instead")]
+        public RequestLocalizationMiddleware(RequestDelegate next, IOptions<RequestLocalizationOptions> options)
+        {
+            if (options == null)
+            {
+                throw new ArgumentNullException(nameof(options));
+            }
+
+            _next = next ?? throw new ArgumentNullException(nameof(next));
+            _logger = NullLogger.Instance;
             _options = options.Value;
         }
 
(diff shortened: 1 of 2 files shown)


src/Middleware/Localization/src/RequestLocalizationMiddleware.cs
```diff
@@ -26,22 +26,24 @@ public class RequestLocalizationMiddleware
...
 
         /// <summary>
         /// Creates a new <see cref="RequestLocalizationMiddleware"/>.
         /// </summary>
         /// <param name="next">The <see cref="RequestDelegate"/> representing the next middleware in the pipeline.</param>
         /// <param name="options">The <see cref="RequestLocalizationOptions"/> representing the options for the
+        /// <param name="loggerFactory">The <see cref="ILoggerFactory"/> used for logging.</param>
         /// <see cref="RequestLocalizationMiddleware"/>.</param>
-        public RequestLocalizationMiddleware(RequestDelegate next, IOptions<RequestLocalizationOptions> options)
+        public RequestLocalizationMiddleware(RequestDelegate next, IOptions<RequestLocalizationOptions> options, ILoggerFactory loggerFactory)
```
gfoidl: Why did you take the change in https://github.com/aspnet/AspNetCore/pull/10946/commits/16739e0cfdbff8c4b50ea3398da4b7d9f585c561?

So to not inject the `ILogger<RequestLocalizationMiddleware>`.

Kahbazi: I saw that other middlewares inject `ILoggerFactory` so I wanted to keep the same pattern

src/Middleware/Localization/src/RequestLocalizationMiddleware.cs
```diff
@@ -26,22 +26,24 @@ public class RequestLocalizationMiddleware
...
         {
             if (options == null)
             {
                 throw new ArgumentNullException(nameof(options));
             }
 
             _next = next ?? throw new ArgumentNullException(nameof(next));
+            _logger = loggerFactory?.CreateLogger<RequestLocalizationMiddleware>() ?? throw new ArgumentNullException(nameof(next));
```
Tratcher:
```suggestion
            _logger = loggerFactory?.CreateLogger<RequestLocalizationMiddleware>() ?? throw new ArgumentNullException(nameof(loggerFactory));
```

src/Middleware/Localization/src/RequestLocalizationMiddleware.cs
Tratcher: The breaking change guidelines are to add a new constructor and obsolete the old one. For weird DI reasons the new constructor should be placed above the old one.

Kahbazi: So how should we create logger object for the old constructor. Maybe I should put back the EnsureLogger!!?

Kahbazi: So how should we create logger object for the old constructor. Maybe I should put back the EnsureLogger!!?

src/Middleware/Localization/src/RequestLocalizationMiddleware.cs
Kahbazi: Oops

src/Middleware/Localization/src/RequestLocalizationMiddleware.cs
rynowak: > For weird DI reasons the new constructor should be placed above the old one.

Does this actually matter to DI? the order used by reflection is not a stable guarantee 😢

Tratcher: It does for the ActivatorUtilities library we use to construct middleware. Middleware aren't in the DI container so we build them manually and pull their dependencies from DI.

pranavkm: There's an attribute that you can use to specify the constructor to be used - `ActivatorUtilitiesConstructorAttribute`

pranavkm: @Kahbazi here's what I'd do:

1) Add a new ctor which accepts a `ILoggerFactory` \ `ILogger<T>` instance. I personally prefer the latter, but there isn't a clear guidance for one over the other.
2) Add `[Obsolete]` attribute to the current attribute. Add `ActivatorUtilitiesConstructorAttribute` to the new ctor
3) Change the current constructor to call the other ctor. Have it pass a `NullLogger` instance. This should avoid the need to have `EnsureLogger`

Tratcher: @Kahbazi here's an example, we use a NullLogger for the other constructor:
https://github.com/aspnet/AspNetCore/blob/62351067ff4c1401556725b401478e648b66acdc/src/Security/CookiePolicy/src/CookiePolicyMiddleware.cs#L20-L32

Kahbazi: So the log will be disabled if someone directly uses the old constructor, is that ok?

Tratcher: Yes, it's quite unlikely anyone will use the old one, but we leave it for binary compat.

src/Middleware/Localization/src/RequestLocalizationMiddleware.cs
```diff
@@ -26,14 +27,35 @@ public class RequestLocalizationMiddleware
...
+        }
+
+        /// <summary>
+        /// Creates a new <see cref="RequestLocalizationMiddleware"/>.
+        /// </summary>
+        /// <param name="next">The <see cref="RequestDelegate"/> representing the next middleware in the pipeline.</param>
+        /// <param name="options">The <see cref="RequestLocalizationOptions"/> representing the options for the
+        /// <see cref="RequestLocalizationMiddleware"/>.</param>
+        [Obsolete("This is obsolete and will be removed in a future version. Use RequestLocalizationMiddleware(RequestDelegate next, IOptions<RequestLocalizationOptions> options, ILoggerFactory loggerFactory) instead")]
         public RequestLocalizationMiddleware(RequestDelegate next, IOptions<RequestLocalizationOptions> options)
```
pranavkm: Could you update this ctor to call the other one:

```C#
public RequestLocalizationMiddleware(RequestDelegate next, IOptions<RequestLocalizationOptions> options)
   : this(next, options, NullLoggerFactory.Instance)
```

src/Middleware/Localization/src/RequestLocalizationMiddleware.cs
```diff
@@ -26,14 +27,35 @@ public class RequestLocalizationMiddleware
...
+            _options = options.Value;
+        }
+
+        /// <summary>
+        /// Creates a new <see cref="RequestLocalizationMiddleware"/>.
+        /// </summary>
+        /// <param name="next">The <see cref="RequestDelegate"/> representing the next middleware in the pipeline.</param>
+        /// <param name="options">The <see cref="RequestLocalizationOptions"/> representing the options for the
+        /// <see cref="RequestLocalizationMiddleware"/>.</param>
+        [Obsolete("This is obsolete and will be removed in a future version. Use RequestLocalizationMiddleware(RequestDelegate next, IOptions<RequestLocalizationOptions> options, ILoggerFactory loggerFactory) instead")]
```
pranavkm: `[Obsolete("This constructor is obsolete and will be removed in a future version. Use RequestLocalizationMiddleware(RequestDelegate next, IOptions<RequestLocalizationOptions> options, ILoggerFactory loggerFactory) instead")]`

src/Middleware/Localization/src/RequestLocalizationMiddleware.cs
Kahbazi: Done

Tratcher:
So far so good. @pranavkm can you track merging this?

pranavkm:
@Kahbazi 👍 

```
src\Middleware\Localization\ref\Microsoft.AspNetCore.Localization.netcoreapp3.0.cs(,): error : Generated code is not up to date in src/Middleware/Localization/ref/Microsoft.AspNetCore.Localization.netcoreapp3.0.cs. You might need to regenerate the reference assemblies or project list (see docs/ReferenceAssemblies.md and docs/ReferenceResolution.md)
```

Kahbazi:
@pranavkm Done!

pranavkm:
Thanks for the PR!

PR #10959, opened 2019-06-06
Add TeamName to MicroBuildCleanup@1

JunTaoLuo:
Might help with failures https://dev.azure.com/dnceng/internal/_build/results?buildId=214844

Not sure why we didn't need this before though.


natemcmaster:
I think a better solution is further up in the file.  https://github.com/aspnet/AspNetCore/blob/aed62cdd9f4e0c9e832d17d65cc9e0c5d2480e48/.azure/pipelines/jobs/default-build.yml#L98-L100

Move TeamName outside of the `if` condition so it is always set globally for all tasks.

dougbu (approved):

natemcmaster (approved):

PR #10972, opened 2019-06-06
Return an empty list instead of null for IHeaderDictionary extensions…

Kahbazi:
https://github.com/aspnet/AspNetCore/issues/3486

### DIFF
--- a/src/Http/Http.Extensions/src/HeaderDictionaryTypeExtensions.cs
+++ b/src/Http/Http.Extensions/src/HeaderDictionaryTypeExtensions.cs
@@ -204,7 +204,7 @@ namespace Microsoft.AspNetCore.Http
 
             if (StringValues.IsNullOrEmpty(values))
             {
-                new List<T>();
+                return Array.Empty<T>();
             }
 
             if (KnownListParsers.TryGetValue(typeof(T), out temp))
@@ -281,7 +281,7 @@ namespace Microsoft.AspNetCore.Http
             {
                 return (IList<T>)parameters[1];
             }
-            return new List<T>();
+            return Array.Empty<T>();
         }
     }
 }
--- a/src/Http/Http.Extensions/src/HeaderDictionaryTypeExtensions.cs
+++ b/src/Http/Http.Extensions/src/HeaderDictionaryTypeExtensions.cs
@@ -161,11 +161,11 @@ namespace Microsoft.AspNetCore.Http
 
         private static IDictionary<Type, object> KnownListParsers = new Dictionary<Type, object>()
         {
-            { typeof(MediaTypeHeaderValue), new Func<IList<string>, IList<MediaTypeHeaderValue>>(value => { IList<MediaTypeHeaderValue> result; return MediaTypeHeaderValue.TryParseList(value, out result) ? result : null; })  },
-            { typeof(StringWithQualityHeaderValue), new Func<IList<string>, IList<StringWithQualityHeaderValue>>(value => { IList<StringWithQualityHeaderValue> result; return StringWithQualityHeaderValue.TryParseList(value, out result) ? result : null; })  },
-            { typeof(CookieHeaderValue), new Func<IList<string>, IList<CookieHeaderValue>>(value => { IList<CookieHeaderValue> result; return CookieHeaderValue.TryParseList(value, out result) ? result : null; })  },
-            { typeof(EntityTagHeaderValue), new Func<IList<string>, IList<EntityTagHeaderValue>>(value => { IList<EntityTagHeaderValue> result; return EntityTagHeaderValue.TryParseList(value, out result) ? result : null; })  },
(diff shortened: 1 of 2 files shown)


src/Http/Http.Extensions/src/HeaderDictionaryTypeExtensions.cs
```diff
@@ -204,7 +204,7 @@ internal static IList<T> GetList<T>(this IHeaderDictionary headers, string name)
 
             if (StringValues.IsNullOrEmpty(values))
             {
-                return null;
+                new List<T>();
```
benaadams: Can use Array.Empty instead of `new List` to not allocate
```suggestion
                Array.Empty<T>();
```

src/Http/Http.Extensions/src/HeaderDictionaryTypeExtensions.cs
```diff
@@ -281,7 +281,7 @@ private static IList<T> GetListViaReflection<T>(StringValues values)
             {
                 return (IList<T>)parameters[1];
             }
-            return null;
+            return new List<T>();
```
benaadams: Here to
```suggestion
            Array.Empty<T>();;
```

src/Http/Http.Extensions/src/HeaderDictionaryTypeExtensions.cs
Kahbazi: @benaadams I think I Should change it back to `new List`, because you get a exception when calling `Add` on `Array.Empty`. What's your thought?

benaadams:
Also need to update 3 tests to expect not null instead of null; and add check for count is 0
```
HeaderDictionaryTypeExtensionsTest.GetListT_UnknownTypeWithTryParseListAndInvalidValue_Null()
 in /_/src/Http/Http.Extensions/test/HeaderDictionaryTypeExtensionsTest.cs:line 141

HeaderDictionaryTypeExtensionsTest.GetListT_UnknownTypeWithTryParseListAndMissingValue_Null()
 in /_/src/Http/Http.Extensions/test/HeaderDictionaryTypeExtensionsTest.cs:line 150

HeaderDictionaryTypeExtensionsTest.GetListT_KnownTypeWithMissingValue_Null()
 in /_/src/Http/Http.Extensions/test/HeaderDictionaryTypeExtensionsTest.cs:line 109
```

Kahbazi:
@benaadams I think I Should change it back to new List, because you get a exception when calling Add on Array.Empty. What's your thought?

benaadams:
> I think I Should change it back to new List, because you get a exception when calling Add on Array.Empty. What's your thought?

Don't see a need. Previously you'd get a null reference exception; the `IList<T>`'s property `IsReadOnly` will return true; it isn't a live list (adding an element won't update the headers), and its primary use is for reading the elements.

Tratcher:
Agreed, Array.Empty is fine.

davidfowl:
Tests?

src/Http/Http.Extensions/test/HeaderDictionaryTypeExtensionsTest.cs
```diff
@@ -41,7 +41,7 @@ public void GetT_KnownTypeWithInvalidValue_Null()
 
             var result = context.Request.GetTypedHeaders().Get<MediaTypeHeaderValue>(HeaderNames.ContentType);
 
-            Assert.Null(result);
+            Assert. Null(result);
```
davidfowl: Revert

src/Http/Http.Extensions/test/HeaderDictionaryTypeExtensionsTest.cs
```diff
@@ -99,24 +99,24 @@ public void GetListT_KnownTypeWithValidValue_Success()
...
         [Fact]
-        public void GetListT_KnownTypeWithMissingValue_Null()
+        public void GetListT_KnownTypeWithMissingValue_EmptyList()
         {
             var context = new DefaultHttpContext();
 
             var result = context.Request.GetTypedHeaders().GetList<MediaTypeHeaderValue>(HeaderNames.Accept);
 
-            Assert.Null(result);
+            Assert.Equal(0, result.Count);
```
davidfowl: Assert.Empty(result)

src/Http/Http.Extensions/test/HeaderDictionaryTypeExtensionsTest.cs
```diff
@@ -99,24 +99,24 @@ public void GetListT_KnownTypeWithValidValue_Success()
...
-        public void GetListT_KnownTypeWithInvalidValue_Null()
+        public void GetListT_KnownTypeWithInvalidValue_EmptyList()
         {
             var context = new DefaultHttpContext();
             context.Request.Headers[HeaderNames.Accept] = "invalid";
 
             var result = context.Request.GetTypedHeaders().GetList<MediaTypeHeaderValue>(HeaderNames.Accept);
 
-            Assert.Null(result);
+            Assert.Equal(0, result.Count);
```
davidfowl: Ditto

davidfowl (approved):

Tratcher:
Thanks

PR #10780, opened 2019-06-03
Guarantee ordering for SignalR client HttpConnection.send

pranavkm:
Resolves https://github.com/aspnet/AspNetCore/issues/9901

PR #10987, opened 2019-06-07
Guarantee ordering for SignalR client HttpConnection.send #10780

### LINKED ISSUE #9901: Blazor (server-side) System.InvalidOperationException: Received a notification for a rendered batch when not expecting it.

pranavkm:
Resolves #9901


eng/targets/Npm.Common.targets
```diff
@@ -47,13 +48,13 @@
...
   <Target Name="Pack" Condition="'$(IsPackable)' == 'true'" DependsOnTargets="$(PackDependsOn)">
     <PropertyGroup>
-      <_BackupPackageJson>$(IntermediateOutputPath)$(MSBuildProjectName).package.json.bak</_BackupPackageJson>
       <_PackageTargetPath>$(MSBuildProjectDirectory)\$(PackageFileName)</_PackageTargetPath>
     </PropertyGroup>
 
     <Copy SourceFiles="$(PackageJson)" DestinationFiles="$(_BackupPackageJson)" />
 
     <Yarn Command="version --no-git-tag-version --new-version $(PackageVersion)" />
+    <Exec Command="node $(MSBuildThisFileDirectory)..\scripts\update-packagejson-links.js $(PackageJson) $(PackageVersion)" />
```
pranavkm: @anurse

src/Components/Browser.JS/Microsoft.AspNetCore.Components.Browser.JS.npmproj
```diff
@@ -7,23 +7,10 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <WebpackInputs Include="src\**\*.ts" />
```
pranavkm: I considered adding build incrementalism to the root Npm.Common.targets and got something working fairly well. Might be something to consider to make inner loops with `dotnet build` be nice.

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -501,3 +515,103 @@ export class HttpConnection implements IConnection {
...
+        if (!this.transportResult) {
+            this.transportResult = new PromiseSource();
+        }
+        return this.transportResult.promise;
+    }
+
+    public stop(): Promise<void> {
+        this.executing = false;
+        this.sendBufferedData.resolve();
+        return this.sendLoopPromise;
```
halter73: Thanks for remembering.

src/SignalR/clients/ts/signalr/tests/HttpConnection.test.ts
```diff
@@ -844,3 +845,232 @@ describe("HttpConnection", () => {
...
+                await promiseSource;
+            });
+        transport.send = sendMock;
+
+        const queue = new TransportSendQueue(transport);
+
+        const send = queue.send("Test");
+        await queue.stop();
+
+        await expect(send).rejects.toBe("Connection stopped.");
```
halter73: Awesomesauce

halter73 (approved):

halter73:
I only evaluated the runtime changes, not the build changes.

src/Components/Browser.JS/package.json
```diff
@@ -5,14 +5,15 @@
   "description": "",
   "main": "index.js",
   "scripts": {
+    "prebuild": "yarn install --mutex network",
```
natemcmaster: Curious - why was this added? While I think it's normal for someone who is used to `dotnet build` to expect `yarn run build` to also restore, I don't know if this is something that's normally done in NodeJS projects. Yarn is pretty good about no-op and caching, but let's be careful about unintended consequences for running restore during build.

eng/scripts/update-packagejson-links.js
```diff
@@ -0,0 +1,12 @@
...
+const fs = require("fs");
+
+const args = process.argv.slice(2);
+const packageJson = args[0];
+const packageVersion = args[1];
+
+const fileContent = fs.readFileSync(packageJson);
+const updatedContent = fileContent.toString().replace(/\"link:.*\"/g, `">=${packageVersion}"`);
+if (fileContent != updatedContent) {
+    fs.writeFileSync(packageJson, updatedContent);
```
natemcmaster: 👏  much better than trying to do this with MSBuild.

build/repo.props
```diff
@@ -126,7 +126,6 @@
         <NodeJsProjects Include="
                           $(RepoRoot)src\SignalR\**\*.npmproj;
                           $(RepoRoot)src\Middleware\**\*.npmproj;
-                          $(RepoRoot)src\Components\Browser.JS\**\*.npmproj;
```
natemcmaster: Why is this being removed from the build?

pranavkm: I'm not a 100% certain since I didn't investigate a great deal as to why, but from what I could guess building the SignalR functional test project and the Browser.JS project in parallel would party over the same set of linked directories. However, we don't actually need to build the Browser.JS project on the CI - we have a CodeCheck to verify it is up to date and the generated files are checked in.

eng/scripts/update-packagejson-links.js
pranavkm: My only concern was that we now invoke `node` in MSBuild. But presumably we already do that as part of `yarn`

src/Components/Browser.JS/package.json
pranavkm: Think I could remove it now. When I was messing around with getting the linked build, I experimented with differing the `yarn install` until after project discovery. I'll try and undo.

src/SignalR/clients/ts/signalr/package.json
```diff
@@ -46,6 +46,7 @@
   ],
   "devDependencies": {
     "@types/eventsource": "^1.0.2",
+    "@types/jest": "^23.3.2",
```
pranavkm: Btw, this gives me typing and completion for `jest` APIs in the SignalR test projects.

PR #10990, opened 2019-06-07
Prefer File.Move for DataProtection key creation + Fallback

BrennanConroy:
We changed from `File.Move` (which is atomic) to `File.Copy` because `File.Move` has issues when using NFS.
You can read the original here: https://github.com/aspnet/AspNetCore/issues/2941

The proposed change is to keep the preferred atomic operation and fallback to `Copy` when `Move` doesn't work.

This does mean there can still be issues on NFS because we are using `Copy` but the only solution (I know of) to that would be to write our own Pinvoke code to use rename on Unix systems instead of link/unlink (which is what `File.Move` does).

### DIFF
--- a/src/DataProtection/DataProtection/src/Repositories/FileSystemXmlRepository.cs
+++ b/src/DataProtection/DataProtection/src/Repositories/FileSystemXmlRepository.cs
@@ -143,8 +143,16 @@ namespace Microsoft.AspNetCore.DataProtection.Repositories
                 // Renames are atomic operations on the file systems we support.
                 _logger.WritingDataToFile(finalFilename);
 
-                // Use File.Copy because File.Move on NFS shares has issues in .NET Core 2.0
-                File.Copy(tempFilename, finalFilename);
+                try
+                {
+                    // Prefer the atomic move operation to avoid multi-process startup issues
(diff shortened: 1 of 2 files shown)


src/DataProtection/DataProtection/src/Repositories/FileSystemXmlRepository.cs
```diff
@@ -143,8 +143,16 @@ private void StoreElementCore(XElement element, string filename)
...
-                // Use File.Copy because File.Move on NFS shares has issues in .NET Core 2.0
-                File.Copy(tempFilename, finalFilename);
+                try
+                {
+                    // Prefer the atomic move operation to avoid multi-process startup issues
+                    File.Move(tempFilename, finalFilename);
+                }
+                catch (IOException)
+                {
+                    // Use File.Copy because File.Move on NFS shares has issues in .NET Core 2.0
```
Tratcher: Issue link for history?

natemcmaster (approved):

HaoK:
Yay this was causing so many flaky tests!

BrennanConroy:
@aspnet/build Is the ~494 failing tests on ubuntu18 known and being tracked somewhere?

Issue #4682, opened 2016-02-24
RemoteError should expose raw error details


PinpointTownes:
Currently, `FailureContext` - used by the `RemoteError` event - doesn't expose the raw protocol error details but an exception, which makes determining the root cause of the error much harder (one may want to detect `access_denied` errors, for instance)

We should consider adding the error code to this class. Sadly, `RemoteAuthenticationHandler` and the `RemoteError` event are supposed to be protocol-agnostic and things like `error`/`error_description`/`error_uri` are OAuth2-specific, so adding `Error`/`ErrorDescription`/`ErrorUri` to `FailureContext` will likely break the existing layering.

PR #11002, opened 2019-06-07
Add error details for OAuth and OIDC remote errors

Tratcher:
#4682 We include individual fields in the Exception.Data Dictionary so the developer can inspect them without parsing out the message. @PinpointTownes

### DIFF
--- a/src/Security/Authentication/OAuth/src/OAuthHandler.cs
+++ b/src/Security/Authentication/OAuth/src/OAuthHandler.cs
@@ -91,7 +91,12 @@ namespace Microsoft.AspNetCore.Authentication.OAuth
                     failureMessage.Append(";Uri=").Append(errorUri);
                 }
 
-                return HandleRequestResult.Fail(failureMessage.ToString(), properties);
+                var ex = new Exception(failureMessage.ToString());
+                ex.Data["error"] = error.ToString();
+                ex.Data["error_description"] = errorDescription.ToString();
+                ex.Data["error_uri"] = errorUri.ToString();
+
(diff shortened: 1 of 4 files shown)


HaoK (approved):

src/Security/Authentication/OAuth/src/OAuthHandler.cs
```diff
@@ -91,7 +91,12 @@ protected override async Task<HandleRequestResult> HandleRemoteAuthenticateAsync
                     failureMessage.Append(";Uri=").Append(errorUri);
                 }
 
-                return HandleRequestResult.Fail(failureMessage.ToString(), properties);
+                var ex = new Exception(failureMessage.ToString());
+                ex.Data["error"] = error.ToString();
+                ex.Data["error_description"] = errorDescription.ToString();
```
PinpointTownes: IIRC, `StringValues.ToString()` returns an empty string, so you could potentially end up with an empty entry in the data dictionary. Would that be worth refactoring to only add `error_description`/`error_uri` when they are present in the query string?

PinpointTownes: As part of the automatic `access_denied` error handling, we also call `HandleRequestResult.Fail` at line 78. Would it make sense to do a similar thing to preserve the error description/uri?

Tratcher: I'm not too worried about the empty strings, they're harmless and omitting them would complicate the code.

Tratcher (commented):
Updated for AccessDenied.

PR #11003, opened 2019-06-07
Ensure MaxConcurrentRequests is a positive number

Kahbazi:
@DylanDmitri

### DIFF
--- a/src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
+++ b/src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
@@ -35,6 +35,10 @@ namespace Microsoft.AspNetCore.RequestThrottling
             {
                 throw new ArgumentException("The value of 'options.MaxConcurrentRequests' must be specified.", nameof(options));
             }
+            if (_requestThrottlingOptions.MaxConcurrentRequests < 0)
+            {
+                throw new ArgumentException("The value of 'options.MaxConcurrentRequests' must be a positive integer.", nameof(options));
+            }
             if (_requestThrottlingOptions.RequestQueueLimit < 0)
             {
                 throw new ArgumentException("The value of 'options.RequestQueueLimit' must be a positive integer.", nameof(options));


src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
```diff
@@ -35,6 +35,10 @@ public RequestThrottlingMiddleware(RequestDelegate next, ILoggerFactory loggerFa
             {
                 throw new ArgumentException("The value of 'options.MaxConcurrentRequests' must be specified.", nameof(options));
             }
+            if (_requestThrottlingOptions.MaxConcurrentRequests < 0)
+            {
+                throw new ArgumentException("The value of 'options.MaxConcurrentRequests' must be a positive integer.", nameof(options));
+            }
```
Eilon: I feel this is the wrong place to do validation. The setter on the property should throw for invalid values:

https://github.com/aspnet/AspNetCore/blob/master/src/Middleware/RequestThrottling/src/RequestThrottlingOptions.cs#L19

Eilon (dismissed):

DylanDmitri:
If `MaxConcurrentRequests` is negative (or zero), there's currently no feedback and the server never responds to any requests. This is BAD behavior and we want to prevent this. 

However, some tests rely on using 0 for `MaxConcurrentRequests` to make sure that nothing sneaks out of the queue. Adding the validation check breaks those tests, so it's all trickier than expected.

Kahbazi:
The check allows `MaxConcurrentRequests` to be 0 so the tests will pass. Also if `MaxConcurrentRequests` is negative we have a exception anyway https://github.com/aspnet/AspNetCore/blob/master/src/Middleware/RequestThrottling/src/Internal/RequestQueue.cs#L23

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
Tratcher: Note we don't currently use that pattern in this middleware (or any other middleware).

Recommendation: take this change as is and then follow up with a change to move this all to Options Validation.
https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-2.2#options-validation

DylanDmitri:
@Kahbazi good catch on the issue!

PR #11004, opened 2019-06-07
Move Database Error Page tests to SQLite

ajcvickers:
In an attempt to make them more reliable.

See also https://github.com/aspnet/EntityFrameworkCore/issues/15997

### DIFF
--- a/src/Middleware/Diagnostics.EntityFrameworkCore/src/DatabaseErrorPageMiddleware.cs
+++ b/src/Middleware/Diagnostics.EntityFrameworkCore/src/DatabaseErrorPageMiddleware.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Builder;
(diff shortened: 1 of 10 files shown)


smitpatel (approved):

src/Middleware/Diagnostics.EntityFrameworkCore/src/DatabaseErrorPageMiddleware.cs
```diff
@@ -131,6 +132,12 @@ public virtual async Task Invoke(HttpContext httpContext)
                             {
                                 var databaseExists = await relationalDatabaseCreator.ExistsAsync();
 
+                                if (databaseExists)
+                                {
+                                    // Also check if the database is completely empty - see https://github.com/aspnet/EntityFrameworkCore/issues/15997
+                                    databaseExists = (bool)typeof(RelationalDatabaseCreator).GetMethod("HasTables", BindingFlags.Instance | BindingFlags.NonPublic).Invoke(relationalDatabaseCreator, null);
```
AndriySvyryd: Cache the `MethodInfo`

ajcvickers: This is only temporary until I get the fix for https://github.com/aspnet/EntityFrameworkCore/issues/15997 checked in.

AndriySvyryd: Ok and you can remove the cached value when that is fixed.

AndriySvyryd (approved):

Issue #10167, opened 2019-05-10
Move IServerVariables feature and friends to Http.Abstractions


anurse:
The IIS server adds an HTTP feature `IServerVariablesFeature`. Server Variables are not an IIS-exclusive concept and date all the way back to CGI so it seems reasonable that a server may want to provide them. We should move this type out of Servers.IIS and in to HttpAbstractions.

Specific changes:
* Move [`IServerVariablesFeature`](https://github.com/aspnet/AspNetCore/blob/master/src/Servers/IIS/IIS/src/IServerVariableFeature.cs) to [Http.Features](https://github.com/aspnet/AspNetCore/tree/3477daf3c4f530dff80f197e0642cb39a26fb07b/src/Http/Http.Features/src)
* Add `GetServerVariables` extension method in [Http.Features](https://github.com/aspnet/AspNetCore/tree/3477daf3c4f530dff80f197e0642cb39a26fb07b/src/Http/Http.Features/src).
* Change [`GetIISServerVariables`](https://github.com/aspnet/AspNetCore/blob/master/src/Servers/IIS/IIS/src/HttpContextExtensions.cs) extension method to just call `GetServerVariables` (and mark as `[Obsolete]`?)
* Update all [reference assemblies](https://github.com/aspnet/AspNetCore/blob/master/docs/ReferenceAssemblies.md)
* Add [`TypeForwardedToAttribute`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.typeforwardedtoattribute?view=netcore-3.0) to [`Servers.IIS`'s `AssemblyInfo.cs`](https://github.com/aspnet/AspNetCore/blob/3477daf3c4f530dff80f197e0642cb39a26fb07b/src/Servers/IIS/IIS/src/AssemblyInfo.cs) pointing at the new type.

davidfowl:
Maybe type forwarding is pointless since it won’t be compatible anyways

Tratcher:
Type forwarding IServerVariablesFeature would be compatible, no? Unless you change the namespace.

GetIISServerVariables can stay for compat, GetServerVariables would be a new extension method. Note GetServerVariables does not go in Http.Features, it would go into Http.Extensions.

davidfowl:
Ah it’s an extension method, I didn’t look at the interface

mderriey:
Cool, thanks for the clear instructions. I'll wait until #10022 is merged to take a stab at it.

anurse:
[...]

PR #11007, opened 2019-06-07
Move IServerVariablesFeature to Http.Features

jkotalik:
- Move `IServerVariablesFeature` to `Http.Features` and forward in `Servers.IIS`
- Create `GetServerVariable` in `Http.Extensions`
- Mark `GetIISServerVariable` obsolete in `Servers.IIS`

Addresses #10167 

(ported from https://github.com/aspnet/AspNetCore/pull/10374)

@aspnet/build big question for you. We have a scenario with IIS tests that require us to test a 2.2 app with a new ASP.NET Core Module. Before, this app was taking a single dependency from 2.2 (Microsoft.AspNetCore.Server.IIS). With this change, there was a type conflict `error CS0433: The type 'IServerVariablesFeature' exists in both 'Microsoft.AspNetCore.Http.Features, Version=3.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60' and 'Microsoft.AspNetCore.Server.IIS, Version=2.2.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'` where two types with the same name existed in the same namespace. Instead of trying to resolve that, I did the "right" thing and made app actually target netcoreapp2.2. However, I'd imagine this could be a build nightmare (though I don't care about updating the versions). What are your thoughts?


natemcmaster:
Re: type conflict. Given that Microsoft.AspNetCore.Http.Features is both shared framework and in a nuget package, it's conceivable a customer could run into issues. I believe the only reason Http.Features is a package is because the SignalR 3.0 client uses it.

One idea: cross-compile Microsoft.AspNetCore.Http.Features for netcoreapp3.0, and exclude the IServerVariablesFeature when building for netstandard2.0.

cc @davidfowl

Tratcher:
@natemcmaster the type conflict was caused by referencing 2.2 and 3.0 AspNetCore assemblies from the same application which as a rule we do not support. I asked him to change the backcompat test asset project to use only AspNetCore 2.2 dependencies and that fixed the issue.

src/Http/Http.Extensions/src/ContextExtensions.cs
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Http.Features;
+
+namespace Microsoft.AspNetCore.Http
+{
+    public static class ContextExtensions
```
Tratcher: Extension method classes are supposed to be named [ObjectBeingExtended]+[FeatureAreaBeingAdded] to avoid conflicts with other extension methods on the same type. Especially when they all use the same namespace.
```suggestion
    public static class HttpContextServerVariableExtensions
```

Tratcher (approved):

Tratcher:
Hmm, Helix didn't like this.
```
F:\workspace\_work\1\s\.dotnet\x64\sdk\3.0.100-preview5-011568\Sdks\Microsoft.NET.Sdk\targets\Microsoft.PackageDependencyResolution.targets(233,5): error NETSDK1005: Assets file 'F:\workspace\_work\1\s\artifacts\obj\InProcessNewShimWebSite\project.assets.json' doesn't have a target for '.NETCoreApp,Version=v3.0'. Ensure that restore has run and that you have included 'netcoreapp3.0' in the TargetFrameworks for your project. [F:\workspace\_work\1\s\src\Servers\IIS\IIS\test\testassets\InProcessNewShimWebSite\InProcessNewShimWebSite.csproj]
```

That's an understandable assumption on Helix's part. Any way to override it?

davidfowl:
> Re: type conflict. Given that Microsoft.AspNetCore.Http.Features is both shared framework and in a nuget package, it's conceivable a customer could run into issues. I believe the only reason Http.Features is a package is because the SignalR 3.0 client uses it.

> One idea: cross-compile Microsoft.AspNetCore.Http.Features for netcoreapp3.0, and exclude the IServerVariablesFeature when building for netstandard2.0.

Hmm this is an interesting problem, we may need to move IFeatureCollection to a lower assembly and type forward.

jkotalik:
As @Tratcher mentioned, this conflict was only occurring due to this test site referencing 2.2 and 3.0 assets. Doing that hasn't been a problem up until now.
Regarding:
```
F:\workspace\_work\1\s\.dotnet\x64\sdk\3.0.100-preview5-011568\Sdks\Microsoft.NET.Sdk\targets\Microsoft.PackageDependencyResolution.targets(233,5): error NETSDK1005: Assets file 'F:\workspace\_work\1\s\artifacts\obj\InProcessNewShimWebSite\project.assets.json' doesn't have a target for '.NETCoreApp,Version=v3.0'. Ensure that restore has run and that you have included 'netcoreapp3.0' in the TargetFrameworks for your project. [F:\workspace\_work\1\s\src\Servers\IIS\IIS\test\testassets\InProcessNewShimWebSite\InProcessNewShimWebSite.csproj]
```

IIRC this was the main reason I was hesitant on making this change: we don't target the 2.2 tfm anywhere else. @aspnet/build thoughts?

Tratcher:
@jkotalik could you use netcoreapp3.0? It's primarily the 2.2 package dependencies that matter here, no?

dev-thinks:
Because of this conflict, am not able to get IServerVariablesFeature in .net standard 2.0/2.1 and available in core app 3.0+ ?

davidfowl:
ASP.NET Core 3.x is .NET Core only. See https://docs.microsoft.com/en-us/aspnet/core/migration/22-to-30?view=aspnetcore-3.1&tabs=visual-studio#migrate-libraries-via-multi-targeting

Issue #3774, opened 2018-10-31
Consider moving ASP.NET Core Data Protection to Microsoft.Extensions.*


PinpointTownes:
[@davidfowl asked for concrete examples](https://github.com/aspnet/AspNetCore/issues/3753#issuecomment-434253160) of things people would like to use outside of ASP.NET Core, that will be limited to .NET Core starting with 3.0.

Some of the OWIN/Katana apps I've worked on rely on the `Microsoft.Owin.Security.Interop` backcompat' package to be able to share things like cookies and authentication tokens with more recent ASP.NET Core apps. This package depends on the shiny ASP.NET Core Data Protection, which currently lives in the `Microsoft.AspNetCore.DataProtection` namespace.

My request is simple: please consider moving the DP components to `Microsoft.Extensions.DataProtection` and keeping them compatible with .NET Standard (and .NET Framework).

blowdart:
I'm confused why you, of all people, don't just share logins via OIDC, you wrote an entire system for that. The back compat shim was meant as a temporary backstop, and honestly I'd like to kill it off in 3.0.

While there's no technical reason not to, data protection was meant for asp.net core use, and for ephemeral data at that. Moving it out would encourage uses outside of what we'd be willing to support, and when that use can end in data loss it seems to me to be a bad idea to encourage it.

blowdart:
(Tagging @davidfowl because he started this)

PinpointTownes:
> I'm confused why you, of all people, don't just share logins via OIDC, you wrote an entire system for that.

I do use OIDC to "share logins", but this standard alone is not the only one involved in the token authentication dance. Not everyone wants to always use JWT for their access tokens: I’m one of those guys.

In the "system I wrote" tokens are by default created using ASP.NET Core Data Protection, which has many advantages over JWT, like an almost configuration-less experience, a very high security level and a huge perf boost (between +15% and +20% req/sec compared to HMACed JWTs encrypted using a symmetric AES key).

With the compat package, we can have OWIN/Katana/ASP.NET 4.x API projects that accept DP access tokens issued by an ASP.NET Core app.

philmcmillan:
[...]

PR #11008, opened 2019-06-07
Move all of DataProtection and Extensions.Identity to netstandard2.0

HaoK:
First part of https://github.com/aspnet/AspNetCore/issues/3774 this will allow Extensions.Identity.Core Stores, and a few parts of data protection to target netstandard 2.0


HaoK:
@natemcmaster @blowdart this seems a bit scary as the NetCorePbkdf2Provider provider is using a netcore DeriveBytes API, I only started with a few of the DP.abstractions packages and KeyDerivation which is the main thing that Identity needs for the password hasher.  

I also reverted the Apr 26 PR that switched the password hasher to use CryptographicOperations.FixedTimeEquals  (2b2aeee9f869ec43f65c188a66d9285858823628)

src/DataProtection/Cryptography.KeyDerivation/src/Microsoft.AspNetCore.Cryptography.KeyDerivation.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <Description>ASP.NET Core utilities for key derivation.</Description>
-    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
natemcmaster:
```suggestion
    <TargetFrameworks>netstandard2.0;netcoreapp3.0</TargetFrameworks>
```

src/DataProtection/Cryptography.KeyDerivation/src/PBKDF2/NetCorePbkdf2Provider.cs
```diff
@@ -1,63 +1,63 @@
-// Copyright (c) .NET Foundation. All rights reserved.
-// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+//// Copyright (c) .NET Foundation. All rights reserved.
+//// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
-using System.Diagnostics;
-using System.Security.Cryptography;
-using System.Text;
+//using System;
```
natemcmaster: Undo this and use `#if` to only compile this for netcoreapp.

src/Identity/Extensions.Core/src/PasswordHasher.cs
```diff
@@ -222,7 +242,7 @@ private static bool VerifyHashedPasswordV2(byte[] hashedPassword, string passwor
 
             // Hash the incoming password and verify it
             byte[] actualSubkey = KeyDerivation.Pbkdf2(password, salt, Pbkdf2Prf, Pbkdf2IterCount, Pbkdf2SubkeyLength);
-            return CryptographicOperations.FixedTimeEquals(actualSubkey, expectedSubkey);
+            return ByteArraysEqual(actualSubkey, expectedSubkey);
```
natemcmaster:
```suggestion
#if NETSTANDARD2_0
            return ByteArraysEqual(actualSubkey, expectedSubkey);
#elif NETCOREAPP3_0
            return CryptographicOperations.FixedTimeEquals(actualSubkey, expectedSubkey);
#else
#error Update target frameworks
#endif
```

Let's not undo https://github.com/aspnet/AspNetCore/pull/9798

src/Identity/Extensions.Core/src/Microsoft.Extensions.Identity.Core.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <Description>ASP.NET Core Identity is the membership system for building ASP.NET Core web applications, including membership, login, and user data. ASP.NET Core Identity allows you to add login features to your application and makes it easy to customize data about the logged in user.</Description>
-    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
natemcmaster:
```suggestion
    <TargetFrameworks>netstandard2.0;netcoreapp3.0</TargetFrameworks>
```

src/Identity/Extensions.Core/src/PasswordHasher.cs
```diff
@@ -65,6 +65,26 @@ public PasswordHasher(IOptions<PasswordHasherOptions> optionsAccessor = null)
             _rng = options.Rng;
         }
 
+        // Compares two byte arrays for equality. The method is specifically written so that the loop is not optimized.
```
natemcmaster: Would recommend wrapping this in `#if NETSTANDARD2_0` to avoid usage in netcoreapp3.0, where there is a better API.

natemcmaster (changes requested):
Good start, but let's use cross compiling to keep existing improvements.

src/DataProtection/Cryptography.KeyDerivation/src/Microsoft.AspNetCore.Cryptography.KeyDerivation.csproj
natemcmaster: On second thought, this should probably stay the same as it was in 2.x and target `netstandard2.0;netcoreapp2.0`. By changing the TFMs to `netstandard2.0;netcoreapp3.0`, anyone who uses the 3.0 package on netcoreapp2.x will end up with different behavior because they will get the netstandard DLL, not the netcore DLL. This will have a negative perf impact on Linux/macOS.

cref https://github.com/aspnet/AspNetCore/commit/e552b5861a9f35a3daa3079d0e56b011f907daed

HaoK:
Okay so I took a stab at switching DP to netstandard and the issue is the IStartupFilter that gets added as part of `AddDataProtection()`

```
           services.TryAddEnumerable(ServiceDescriptor.Singleton<IStartupFilter, DataProtectionStartupFilter>());
```

Was there any prior discussion about how this should be resolved?

HaoK:
@natemcmaster @ajcvickers what do you guys think about merging the identity + some of dataprotection changes since this at least enables extensions identity + dp abstractions / keyderivations to be used on netstandard 2.0.  I can open another PR addressing the gnarly hosting + DP when we have agreed on an approach

natemcmaster:
Discussed briefly with @ajcvickers.

I'm okay merging the changes in this PR as long as you update the description to remove the "fixes #3774" description. The goal of https://github.com/aspnet/AspNetCore/issues/3774 was to support packages which are both .NET Standard _and_ do not have a dependency on the .NET Core shared framework, and this PR doesn't yet satisfy that criteria. To complete that issue, you'll need to address the DP -> Hosting dependency. Here are my thoughts on ways you might solve this: https://github.com/aspnet/AspNetCore/issues/3774#issuecomment-500563552.

Also, to make the packages available as packages, you need to set `<IsShippingPackage>` in the .csproj file, otherwise the assemblies will still be shared-framework only.

HaoK:
Removed fix comment and added `<IsShippingPackage>true</IsShippingPackage`

src/Shared/WebEncoders/WebEncoders.cs
```diff
@@ -2,7 +2,9 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+#if NETCOREAPP3_0
```
HaoK: FYi @BrennanConroy I had to add this guard around the using buffers introduced from the spanify web encoders change last week since it isn't available in netstandard 2.0

HaoK:
This should now fix https://github.com/aspnet/AspNetCore/issues/3774 pending the resolution of IHostedService not running before the server starts :(

src/Shared/WebEncoders/WebEncoders.cs
BrennanConroy: Makes sense 👍

src/DataProtection/DataProtection/src/Microsoft.AspNetCore.DataProtection.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <Description>ASP.NET Core logic to protect and unprotect data, similar to DPAPI.</Description>
-    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <TargetFramework>netstandard2.0</TargetFramework>
     <IsAspNetCoreApp>true</IsAspNetCoreApp>
```
dougbu: Suggest we also need to ship this as a package i.e.
```suggestion
    <IsAspNetCoreApp>true</IsAspNetCoreApp>
    <IsShippingPackage>true</IsShippingPackage>
```

dougbu (changes requested):

Tratcher:
> This should now fix #3774 pending the resolution of IHostedService not running before the server starts :(

In WebHost IHostedServices are started just after the server. In generic Host the web app is an IHostedService and they're started in registration order. Try testing with generic host to make sure you can get the ordering right. For WebHost it looks like IHostedServices is better than first-request, but not quite as good as IStartupFilter.

HaoK:
@Tratcher Cool, so the key ring startup test does pass using the generic host, but is this 'good enough'?  The webhost test fails, from @natemcmaster comment in the main issue, it doesn't seem safe for the key ring to not be intialized before server start, @blowdart time is running out for preview 7 do we need to resolve the key loading issue with WebHost now, or is that issue something we can tackle in a follow up PR (most likely early preview 8)?

HaoK:
@Tratcher I went ahead and tried starting IHostingService in WebHost before server start as well, I had to updated 3 tests as well, no idea if this is 'correct' but it makes the data protection web host test pass as well 🤷‍♂

src/Hosting/Hosting/src/Internal/WebHost.cs
```diff
@@ -158,8 +162,6 @@ public virtual async Task StartAsync(CancellationToken cancellationToken = defau
             // Fire IApplicationLifetime.Started
             _applicationLifetime?.NotifyStarted();
 
-            // Fire IHostedService.Start
-            await _hostedServiceExecutor.StartAsync(cancellationToken).ConfigureAwait(false);
```
Tratcher: Can't change this, it's quite breaking.

src/DataProtection/DataProtection/test/HostingTests.cs
```diff
@@ -46,6 +47,32 @@ public async Task LoadsKeyRingBeforeServerStarts()
...
+                .Returns(Mock.Of<IKeyRing>())
+                .Callback(() => tcs.TrySetResult(null));
+
+            var builder = new HostBuilder()
+                .ConfigureServices(s =>
+                    s.AddDataProtection()
+                    .Services
+                    .Replace(ServiceDescriptor.Singleton(mockKeyRing.Object))
+                    .AddSingleton<IServer>(
+                        new FakeServer(onStart: () => tcs.TrySetException(new InvalidOperationException("Server was started before key ring was initialized")))));
```
Tratcher: This IServer isn't actually getting started, you didn't add the webhostedservice.

Tratcher (commented):
As to if this is good enough for webhost, I honestly couldn't say. It's certainly not as good as the startup filter. I'll follow with you directly to bainstorm.

src/DataProtection/DataProtection/test/HostingTests.cs
HaoK: You mean adding a `.ConfigureWebHost(b => b.UseStartup<TestStartup>());` right?

src/Hosting/Hosting/src/Internal/WebHost.cs
HaoK: The only reason I tried this was that @davidfowl seemed open to changing WebHost in https://github.com/aspnet/AspNetCore/issues/3774#issuecomment-501794264

> Another option might be an interface like IHostedService? We can change the WebHost to make sure these run before the server starts (I believe it doesn't today).

davidfowl: Can't? We can discuss how breaking it is.

HaoK:
@aspnet/build any ideas how i can update SharedFramework.Local.props manually since I'm unable to run the script since my local restore/build is borked, it gave me a diff before when I updated it previously in this PR, but its not clear to me what the delta is between what it expects and what is checked in right now

```

eng\SharedFramework.Local.props(,): error : Generated code is not up to date in eng/SharedFramework.Local.props. You might need to regenerate the reference assemblies or project list (see docs/ReferenceAssemblies.md and docs/ReferenceResolution.md)
```

HaoK:
Figured out the diff was trailing whitespace in the sharedframework.local...grrr

HaoK:
Per offline thread with @Tratcher and @davidfowl this seems good enough for preview 7, will tweak in subsequent PRs as needed

PR #11010, opened 2019-06-08
Update HubConnection url in Java client

mikaelm12:
Part of issue: https://github.com/aspnet/AspNetCore/issues/5342

### DIFF
--- a/src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
+++ b/src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
@@ -29,7 +29,7 @@ public class HubConnection {
     private static final List<Class<?>> emptyArray = new ArrayList<>();
     private static final int MAX_NEGOTIATE_ATTEMPTS = 100;
 
-    private final String baseUrl;
+    private String baseUrl;
     private Transport transport;
     private OnReceiveCallBack callback;
     private final CallbackMap handlers = new CallbackMap();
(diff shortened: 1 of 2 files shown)


src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
```diff
@@ -290,6 +290,27 @@ public HubConnectionState getConnectionState() {
...
+     * Sets a new url for the HubConnection.
+     * @param url The url to connect to.
+     */
+    public void setBaseUrl(String url) {
+        if (url == null || url.isEmpty()) {
+            throw new IllegalArgumentException("The HubConnection url must be a valid url.");
+        }
+
+        if (hubConnectionState != HubConnectionState.DISCONNECTED) {
+            throw new IllegalStateException("The HubConnection must be in the disconnected state to change the url.");
```
anurse: Are we guaranteed that the connection will be in the `DISCONNECTED` state during the onClosed callback? I'd like to see a test confirming that works (changing the URL in the `onClosed` callback, and doing so **without** waiting for `.stop` to complete, to make sure the asynchrony is working right).

mikaelm12:
> Are we guaranteed that the connection will be in the DISCONNECTED state during the onClosed callback? I'd like to see a test confirming that works

Added that test and ran in in a loop locally 10 thousand times.

mikaelm12:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=219084

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/2200
https://github.com/aspnet/AspNetCore-Internal/issues/2656

BrennanConroy (approved):

PR #11018, opened 2019-06-09
Only run selenium tests when build java is specified

JunTaoLuo:
Addresses https://github.com/aspnet/AspNetCore/issues/10504


dougbu (approved):

dougbu:
More Ubuntu `inotify` failures in Helix runs ☹️

JunTaoLuo:
Yea there are a few tracking issues on the inotify issue already. cc @javiercn are you looking into those?

javiercn:
Yes, I have a PR out, although looking at my changes I don’t know how they caused it.

pranavkm:
I'm doing the egregious sin of commenting on a closed PR. But with this change, I can't do `dotnet test` on the Component's E2E tests anymore since `BuildJava` isn't defined by default. Should the test here be:

``$(BuildJava)` != "false"` so it's only skipped when we explicitly say we do not require Java? Right now, there's a risk that the tests never run if the flag isn't set. What do you folks think?

dougbu:
`dotnet test /p:BuildJava=true`

PR #11020, opened 2019-06-09
Reduce allocations when Activity is enabled

davidfowl:
- Remove string allocations caused by DiagnosticSource.Stop/StartActivity
- Pass the HttpContext directly as the object for StartActivity and StopActivity to avoid the anonymous object allocation.
- Though it's a bit ugly, added an HttpContext property to DefaultHttpContext to avoid breaking back-compat (which had to do reflection to get the HttpContext property anyways)

Contributes to #9594

Before:
image

After:
image

### DIFF
--- a/src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs
+++ b/src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs
@@ -302,7 +302,7 @@ namespace Microsoft.AspNetCore.Hosting.Internal
             }
         }
 
-        // These are versions of DiagnosticSource.Star/StopActivity that don't allocate strings per call (see https://github.com/dotnet/corefx/issues/37055)
+        // These are versions of DiagnosticSource.Start/StopActivity that don't allocate strings per call (see https://github.com/dotnet/corefx/issues/37055)
         private Activity StartActivity(Activity activity, HttpContext httpContext)
         {
             activity.Start();
--- a/src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs
+++ b/src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs
@@ -16,7 +16,8 @@ namespace Microsoft.AspNetCore.Hosting.Internal
         private static readonly double TimestampToTicks = TimeSpan.TicksPerSecond / (double)Stopwatch.Frequency;
 
         private const string ActivityName = "Microsoft.AspNetCore.Hosting.HttpRequestIn";
-        private const string ActivityStartKey = "Microsoft.AspNetCore.Hosting.HttpRequestIn.Start";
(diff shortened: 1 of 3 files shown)


src/Http/Http/src/DefaultHttpContext.cs
```diff
@@ -157,6 +157,11 @@ public override ISession Session
             }
         }
 
+        // This property exists because of backwards compatibility.
+        // We send an anonymous object with an HttpContext property
+        // via DiagnosticSource in various events throughout the pipeline. Instead
+        // we just send the HttpContext to avoid extra allocations
+        public HttpContext HttpContext => this;
```
Tratcher: EditorBrowsableNever?

davidfowl: Ah nice!

src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs
```diff
@@ -293,12 +294,32 @@ private void StopActivity(HttpContext httpContext, Activity activity, bool hasDi
...
-                _diagnosticListener.StopActivity(activity, new { HttpContext = httpContext });
+                StopActivity(activity, httpContext);
             }
             else
             {
                 activity.Stop();
             }
         }
+
+        // These are versions of DiagnosticSource.Star/StopActivity that don't allocate strings per call (see https://github.com/dotnet/corefx/issues/37055)
```
watfordgnf:
```suggestion
        // These are versions of DiagnosticSource.Start/StopActivity that don't allocate strings per call (see https://github.com/dotnet/corefx/issues/37055)
```

Tratcher (approved):

davidfowl:
cc @lmolkova @SergeyKanzhelev as an FYI, this should be non breaking though.

src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs
```diff
@@ -278,7 +279,7 @@ private Activity StartActivity(HttpContext httpContext, out bool hasDiagnosticLi
             if (_diagnosticListener.IsEnabled(ActivityStartKey))
             {
                 hasDiagnosticListener = true;
-                _diagnosticListener.StartActivity(activity, new { HttpContext = httpContext });
+                StartActivity(activity, httpContext);
```
lmolkova: I like this and we may be able to update listener in time for 3.0 though. But it is breaking (i.e. there is no HttpContext property on HttpContext type), everyone who attempts to get it will fail.

lmolkova: ooops, you've added property on HttpConetxt :)

davidfowl: HttpContext.HttpContext 😄

PR #11028, opened 2019-06-10
Upgrade syntax is ?? to reduce the amount of code

YPTopMan:
- Parameter empty judgment, upgrade syntax ??
 - 
Upgrade the syntax and the user reduces the number of lines of code.Hope to get merged into the main branch


anurse:
We don't generally take bulk style update PRs like this. We also don't have a clear agreement on the team that `?? throw` is the syntax we prefer.

PR #11029, opened 2019-06-10
[Identity] Move to use static web assets support.

javiercn:
* Replaces the embedded file support inside identity with support for the new static web assets feature.
* Removes the need to specify the bootstrap framework at runtime.
* Removes the need to embed static assets into the assembly.
* Removes the need to hijack the static file options to plug in the static assets.
* Copies only the assets for the selected version to the output folder on publish.


src/Identity/UI/src/build/Microsoft.AspNetCore.Identity.UI.props
```diff
@@ -0,0 +1,31 @@
...
+  <ItemGroup Condition="'$(IdentityUIFrameworkVersion)' == 'Bootstrap4'">
+    <StaticWebAsset Include="$(MSBuildThisFileDirectory)..\staticwebassets\V4\**">
+      <SourceType>Package</SourceType>
+      <SourceId>Microsoft.AspNetCore.Identity.UI</SourceId>
+      <ContentRoot>$([MSBuild]::NormalizePath('$(MSBuildThisFileDirectory)..\staticwebassets\V4'))</ContentRoot>
+      <BasePath>/Identity</BasePath>
+      <RelativePath>%(RecursiveDir)%(FileName)%(Extension)</RelativePath>
+    </StaticWebAsset>
+  </ItemGroup>
+</Project>
```
javiercn: I plan to individually enumerate the files here so that we can have a some extra options.
* Do not publish assets if identity is configured to use the app layout. (We can have a pretty good guess with `Exists(
Areas\Identity\Pages\_ViewStart.cshtml')`
* Only publish minified assets by default.
* Disable publishing any static asset

src/Identity/UI/src/Microsoft.AspNetCore.Identity.UI.csproj
```diff
@@ -7,23 +7,30 @@
...
+    <IdentityUIFrameworkVersion Condition="'$(IdentityUIFrameworkVersion)' == ''">Bootstrap4</IdentityUIFrameworkVersion>
+
   </PropertyGroup>
 
   <ItemGroup>
-    <EmbeddedResource Include="wwwroot/**/*" />
-    <EmbeddedResource Remove="wwwroot/**/LICENSE*" />
-    <None Remove="wwwroot/**/LICENSE*" />
+    <Content Remove="@(Content)" />
+    <Content Include="wwwroot\**\*" Pack="true" PackagePath="staticwebassets" />
```
mkArtakMSFT: Is this ("staticwebassets") going to appear in the path here (C:\Users*.nuget\packages\microsoft.aspnetcore.identity.ui\3.0.0-preview7.19320.1\**staticwebassets**\V4\Identity\lib\jquery-validation\dist\jquery.validate.min.js)? If so, maybe we should choose a shorter option `swa`? I understand we can't completely eliminate the `path-too-long problem`, but we can do our best to minimize the chance.

javiercn: We can't actually, due to the way this works during development. We add a static files provider pointing to `C:\Users*.nuget\packages\microsoft.aspnetcore.identity.ui\3.0.0-preview7.19320.1*staticwebassets`. We can only shorten the staticwebassets piece, but everything else must stay.

The issue with https://github.com/aspnet/AspNetCore/issues/11529 is that given there is no proper support for SWA today in this package, it ends up using the default configuration, which publishes them to `_content/microsoft.aspnetcore.identity.ui/V4\Identity\lib\jquery-validation\dist\jquery.validate.min.js`

With proper support for SWA, the publish path becomes 
`identity\lib\jquery-validation\dist\jquery.validate.min.js`

Which is way shorter. We could also get away with shortening the paths even more, as these only appear in the default layout. When the layout is overriden (most common case) it can simply use the assets in the template.

mkArtakMSFT: That's good and my suggestion here was simply shortening that even further, nothing fancy

src/ProjectTemplates/test/IdentityUIPackageTest.cs
```diff
@@ -0,0 +1,190 @@
...
+            Assert.True(0 == migrationsResult.ExitCode, ErrorMessages.GetFailedProcessMessage("run EF migrations", Project, migrationsResult));
+            Project.AssertEmptyMigration("razorpages");
+
+            using (var aspNetProcess = Project.StartBuiltProjectAsync())
+            {
+                Assert.False(
+                    aspNetProcess.Process.HasExited,
+                    ErrorMessages.GetFailedProcessMessageOrEmpty("Run built project", Project, aspNetProcess.Process));
+
+                var response = await aspNetProcess.SendRequest("/Identity/lib/bootstrap/dist/css/bootstrap.css");
```
ryanbrandenburg: So the test here is to confirm not just that the files ended up in the produced result, but that that result is then accessible via HTTP?

ryanbrandenburg (approved):
The ProjectTemplates parts of this seem fine, I'll leave reviewing the IdentityUI stuff to someone with more context.

src/ProjectTemplates/test/IdentityUIPackageTest.cs
javiercn: Yes

src/Identity/UI/ref/Microsoft.AspNetCore.Identity.UI.netcoreapp3.0.cs
```diff
@@ -6,20 +6,15 @@ namespace Microsoft.AspNetCore.Identity
     public static partial class IdentityBuilderUIExtensions
     {
         public static Microsoft.AspNetCore.Identity.IdentityBuilder AddDefaultUI(this Microsoft.AspNetCore.Identity.IdentityBuilder builder) { throw null; }
-        public static Microsoft.AspNetCore.Identity.IdentityBuilder AddDefaultUI(this Microsoft.AspNetCore.Identity.IdentityBuilder builder, Microsoft.AspNetCore.Identity.UI.UIFramework framework) { throw null; }
     }
 }
 namespace Microsoft.AspNetCore.Identity.UI
 {
-    public partial class DefaultUIOptions
```
pranavkm: Could we leave these types in and obsolete them with guidance on how to use the correct API instead of removing it? Are there any consequences to leaving it around, but not do anything?

src/Identity/UI/src/IdentityBuilderUIExtensions.cs
```diff
@@ -131,5 +125,31 @@ private static void AddRelatedParts(IdentityBuilder builder, UIFramework framewo
                     }
                 });
         }
+
+        private static Assembly GetApplicationAssembly(IdentityBuilder builder)
+        {
+            // Whis is the same logic that MVC follows to find the application assembly.
```
pranavkm: ~~Whis~~ This

/cc @davidfowl since this is your favourite piece of code

src/Identity/UI/src/Microsoft.AspNetCore.Identity.UI.csproj
```diff
@@ -40,47 +47,27 @@
...
-      SkipUnchangedFiles="$(SkipCopyUnchangedFiles)"
-      OverwriteReadOnlyFiles="$(OverwriteReadOnlyFiles)"
-      Retries="$(CopyRetryCount)"
-      RetryDelayMilliseconds="$(CopyRetryDelayMilliseconds)"
-      UseHardlinksIfPossible="$(CreateHardLinksForCopyFilesToOutputDirectoryIfPossible)"
-      UseSymboliclinksIfPossible="$(CreateSymbolicLinksForCopyFilesToOutputDirectoryIfPossible)">
-
-      <Output TaskParameter="DestinationFiles" ItemName="FileWrites"/>
-      <Output TaskParameter="DestinationFiles" ItemName="_RazorAssembly"/>
+    <MSbuild Projects="$(MSBuildThisFile)" Targets="BuildForUI" Properties="&#xD;&#xA;             BuildProjectReferences=false;&#xD;&#xA;             RazorCompileOnBuild=true;&#xD;&#xA;             UIFrameworkVersion=%(UIFrameworkVersionMoniker.Identity);&#xD;&#xA;             IntermediateOutputPath=$(IntermediateOutputPath)%(UIFrameworkVersionMoniker.Identity)\;" />
```
pranavkm: Why did this change?

src/Identity/UI/src/build/Microsoft.AspNetCore.Identity.UI.props
```diff
@@ -0,0 +1,31 @@
...
+    <IdentityUIFrameworkVersion Condition="'$(IdentityUIFrameworkVersion)' == ''">Bootstrap4</IdentityUIFrameworkVersion>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <AssemblyAttribute Include="Microsoft.AspNetCore.Identity.UI.UIFrameworkAttribute">
+      <_Parameter1>$(IdentityUIFrameworkVersion)</_Parameter1>
+    </AssemblyAttribute>
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(IdentityUIFrameworkVersion)' == 'Bootstrap3'">
```
pranavkm: Have we considered having two packages instead of doing this?

src/Identity/UI/src/UIFramework.cs
```diff
@@ -1,21 +1,13 @@
...
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 namespace Microsoft.AspNetCore.Identity.UI
 {
-    /// <summary>
-    /// The list of supported presentation frameworks for the default UI
-    /// </summary>
-    public enum UIFramework
+    internal enum UIFramework
```
pranavkm: Can we avoid making a breaking change here?

src/Identity/UI/src/UIFrameworkAttribute.cs
```diff
@@ -0,0 +1,29 @@
...
+
+using System;
+
+namespace Microsoft.AspNetCore.Identity.UI
+{
+    /// <summary>
+    /// The UIFramework Identity UI will use on the application.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Assembly, Inherited = false, AllowMultiple = false)]
+    public class UIFrameworkAttribute : Attribute
```
pranavkm: Does this need to be public? This should also be declared sealed.

src/Identity/UI/src/IdentityBuilderUIExtensions.cs
```diff
@@ -131,5 +125,31 @@ private static void AddRelatedParts(IdentityBuilder builder, UIFramework framewo
...
+            var metadata = assembly.GetCustomAttributes<UIFrameworkAttribute>()
+                .SingleOrDefault()?.UIFramework; // Bootstrap4 is the default
+            if (metadata == null)
+            {
+                uiFramework = default;
+                return false;
+            }
+
+            // If we find the metadata there must be a valid framework here.
+            uiFramework = Enum.Parse<UIFramework>(metadata, ignoreCase: true);
```
pranavkm: Could this throw a meaningful error if the value is unsupported?

src/Identity/UI/src/IdentityBuilderUIExtensions.cs
javiercn: I'm getting rid of it

src/Identity/UI/src/Microsoft.AspNetCore.Identity.UI.csproj
javiercn: VS mangling

src/Identity/UI/src/UIFrameworkAttribute.cs
javiercn: I'm trying to get rid of this attribute too

src/Identity/UI/ref/Microsoft.AspNetCore.Identity.UI.netcoreapp3.0.cs
javiercn: We want to get rid of the types as its confusing to keep them around when the mechanics have completely changed.

src/Identity/samples/IdentitySample.DefaultUI/IdentitySample.DefaultUI.csproj
```diff
@@ -34,4 +37,23 @@
...
+    </AssemblyAttribute>
+  </ItemGroup>
+
+  <PropertyGroup>
+    <ResolveStaticWebAssetsInputsDependsOn>_SetBootstrapFrameworkVersion</ResolveStaticWebAssetsInputsDependsOn>
+  </PropertyGroup>
+
+  <Target Name="_SetBootstrapFrameworkVersion">
+    <ItemGroup>
+      <_StaticWebAssetsProjectReference Condition="'%(FileName)' == 'Microsoft.AspNetCore.identity.UI'">
```
HaoK: nit: casing on "AspNetCore.identity"

HaoK (approved):
Removing the UIFramework runtime options is nice

src/Identity/UI/src/Microsoft.AspNetCore.Identity.UI.csproj
javiercn: I undid this

src/Identity/UI/src/UIFramework.cs
javiercn: We want to take all these types out now that we can.

src/Identity/UI/src/UIFrameworkAttribute.cs
javiercn: Couldn't remove it.

It needs to be public because it gets stamped into the app assembly. I marked it sealed, though

src/Identity/samples/IdentitySample.DefaultUI/IdentitySample.DefaultUI.csproj
javiercn: done

src/Identity/UI/src/IdentityBuilderUIExtensions.cs
javiercn: done

src/Identity/UI/src/IdentityBuilderUIExtensions.cs
javiercn: couldn't get rid of it, but unimportant as we only poke the collection in testing scenarios. (otherwise app.getentryassembly works fine)

PR #11042, opened 2019-06-10
Add support for static web assets in blazor client-side

javiercn:


SteveSandersonMS:
This looks good to me, but did you say yesterday it doesn't work for some reason? What's not working about it?

javiercn:
Some piece when the client is hosted on an asp.net core app. The index ends up in the wrong place. As Static Web Assets treats it as candidate content and puts it under _content/clientlib

SteveSandersonMS:
So you're saying this change breaks `MapFallbackToClientSideBlazor` somehow?

Or are you just saying it's redundant to copy the client app's `wwwroot` into `_content`, even though it doesn't break anything? If so I agree it's redundant to do that, since we are serving the client app's `wwwroot` folder at the **root URL** via `UseClientSideBlazorFiles` (we need to continue doing this).

javiercn:
@SteveSandersonMS I fixed publish on hosted scenarios, which was the last thing that was broken

src/Components/Blazor/Build/src/ReferenceFromSource.props
```diff
@@ -28,7 +28,7 @@
   <PropertyGroup>
     <RunCommand>dotnet</RunCommand>
     <_BlazorCliLocation>$(MSBuildThisFileDirectory)../../DevServer/src/bin/$(Configuration)/netcoreapp3.0/blazor-devserver.dll</_BlazorCliLocation>
-    <RunArguments>exec &quot;$(_BlazorCliLocation)&quot; serve $(AdditionalRunArguments)</RunArguments>
+    <RunArguments>exec &quot;$(_BlazorCliLocation)&quot; serve $(OutputPath)$(TargetFileName) $(AdditionalRunArguments)</RunArguments>
```
javiercn: This also improves the DevServer scenario and will fix a bug users have been having with Windows 7. I chose to include this fix because it actually causes issues that are hard to diagnose when running the functional E2E tests

src/Components/Blazor/DevServer/src/Commands/ServeCommand.cs
```diff
@@ -20,11 +23,28 @@ public ServeCommand(CommandLineApplication parent)
...
             OnExecute(Execute);
         }
 
+        public CommandArgument ApplicationPath { get; private set; }
+
         private int Execute()
         {
+            if (string.IsNullOrWhiteSpace(ApplicationPath.Value))
+            {
+                throw new InvalidOperationException($"Invalid path {ApplicationPath.Value}.");
```
SteveSandersonMS: This error will be pretty unclear if it's actually an empty string. It will just say `Invalid path `.

```suggestion
                throw new InvalidOperationException($"Invalid value for parameter '{nameof(ApplicationPath)}'. Value supplied: '{ApplicationPath.Value}'");
```

src/Components/Blazor/Build/src/ReferenceFromSource.props
SteveSandersonMS: That's cool. Looks like it will also fix https://github.com/aspnet/AspNetCore/issues/5486, right?

Might it be worth passing the absolute path to the binary here though? I haven't tried this to check it, but I'd guess if you did `dotnet run --project someotherpath` then this would fail because `$(OutputPath)` would be relative to the `.csproj`, but at runtime it would try to resolve it relative to the CWD. Or am I misreading this?

SteveSandersonMS: Also I think the `$(OutputPath)$(TargetFileName)` part should be surrounded in `&quot;` since either part of that could contain a space.

src/Components/Blazor/DevServer/src/build/Microsoft.AspNetCore.Blazor.DevServer.props
```diff
@@ -1,7 +1,7 @@
-﻿<Project>
+<Project>
   <PropertyGroup>
     <_BlazorDevServerDll>$(MSBuildThisFileDirectory)../tools/blazor-devserver.dll</_BlazorDevServerDll>
     <RunCommand>dotnet</RunCommand>
-    <RunArguments>&quot;$(_BlazorDevServerDll)&quot; serve</RunArguments>
+    <RunArguments>&quot;$(_BlazorDevServerDll)&quot; serve $(OutputPath)$(TargetFileName)</RunArguments>
```
SteveSandersonMS: Same comments I posted at https://github.com/aspnet/AspNetCore/pull/11042/files#r296587897

src/Components/Browser.JS/src/Boot.Server.ts
```diff
@@ -45,11 +45,6 @@ async function boot(userOptions?: Partial<BlazorOptions>): Promise<void> {
   const circuitHandlers: CircuitHandler[] = [new AutoReconnectCircuitHandler(logger)];
   window['Blazor'].circuitHandlers = circuitHandlers;
 
-  // In the background, start loading the boot config and any embedded resources
-  const embeddedResourcesPromise = fetchBootConfigAsync().then(bootConfig => {
-    return loadEmbeddedResourcesAsync(bootConfig);
-  });
```
SteveSandersonMS: There's more involved in removing this feature:

 * We should also remove the equivalent code from `Boot.WebAssembly.ts`
 * We should also remove any underlying infrastructure that used to support these old-style embedded resources, e.g., other MSBuild targets
 * We should update the templates and E2E tests that rely on it

Altogether my guess is you won't want to do all that as part of this PR. If that's correct, maybe we could leave out the changes in this file and remove all this stuff one time later.

SteveSandersonMS (approved):
Some minor comments but altogether great to see this!

src/Components/Blazor/Build/src/ReferenceFromSource.props
javiercn: Yes, and the windows 7 issue. And an issue where the E2E tests don't run proper if you built from command-line (which is why I ended up fixing it, cuz it cost me 1h of my life)

javiercn: You might be right about relative/full, but that's an easy fix.

src/Components/Browser.JS/src/Boot.Server.ts
javiercn: I would like to start getting rid of this on the server-side part to make sure it doesn't make it to 3.0. I'll file a bug to finish cleaning-up the blazor client-side pieces of it.

SteveSandersonMS: OK cool!

src/Components/Blazor/DevServer/src/build/Microsoft.AspNetCore.Blazor.DevServer.props
javiercn: done

src/Components/Blazor/Build/src/ReferenceFromSource.props
javiercn: I did add the full path

javiercn: and validated calling `dotnet run --project projectfilepath`

PR #11045, opened 2019-06-10
Use Ubuntu 18.04 LTS instead of 18.10

JunTaoLuo:
18.10 is going out of support next month. I'm assuming we want to test on 18.04 LTS as we probably want some coverage on the latest LTS instead of removing the queue completely.


dougbu (approved):

natemcmaster (approved):

JunTaoLuo:
Looks like the ubuntu 18.04 tests are running on Helix.

PR #11046, opened 2019-06-10
Add PlatformManifest to runtime package

JunTaoLuo:
Addresses https://github.com/aspnet/AspNetCore/issues/10474

I verified on a local build that data/PlatformManifest.txt is now included in Microsoft.AspNetCore.App.Runtime.*.nupkg. I see a PackageOverrides.txt, is that required for the runtime pack as well or is it just a ref pack thing?

### DIFF
--- a/src/Framework/src/Microsoft.AspNetCore.App.Runtime.csproj
+++ b/src/Framework/src/Microsoft.AspNetCore.App.Runtime.csproj
@@ -265,8 +265,18 @@ This package is an internal implementation of the .NET Core SDK and is not meant
           BeforeTargets="_GetPackageFiles;GenerateNuspec"
           DependsOnTargets="ResolveReferences">
     <ItemGroup>
-      <!-- Native files are included in _PackageFile instead because they belong in a separate package folder.r -->
+      <!-- Native files are included in _PackageFile instead because they belong in a separate package folder. -->
       <Content Include="@(ReferenceCopyLocalPaths)" PackagePath="$(NativeAssetsPackagePath)" Condition=" '%(ReferenceCopyLocalPaths.IsNativeImage)' == 'true' AND '%(ReferenceCopyLocalPaths.Extension)' != '.pdb' " />
+    </ItemGroup>
+  </Target>
+
+  <Target Name="_ResolveRuntimePackContent"
+          BeforeTargets="_GetPackageFiles"
+          DependsOnTargets="GenerateSharedFxDepsFile">
+    <ItemGroup>
+      <RuntimePackContent Include="$(PlatformManifestOutputPath)" PackagePath="$(ManifestsPackagePath)" />
+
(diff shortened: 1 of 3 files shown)


src/Framework/src/Microsoft.AspNetCore.App.Runtime.csproj
```diff
@@ -265,11 +265,21 @@ This package is an internal implementation of the .NET Core SDK and is not meant
...
+      <!-- Native files are included in _PackageFile instead because they belong in a separate package folder. -->
       <Content Include="@(ReferenceCopyLocalPaths)" PackagePath="$(NativeAssetsPackagePath)" Condition=" '%(ReferenceCopyLocalPaths.IsNativeImage)' == 'true' AND '%(ReferenceCopyLocalPaths.Extension)' != '.pdb' " />
     </ItemGroup>
   </Target>
 
+  <Target Name="_ResolveRuntimePackContent"
+          BeforeTargets="_GetPackageFiles"
+          DependsOnTargets="GenerateSharedFxDepsFile">
+    <ItemGroup>
+      <RuntimePackContent Include="$(PlatformManifestOutputPath)" PackagePath="$(ManifestsPackagePath)" />
```
dougbu: Why bother w/ this indirection?

src/Framework/src/Microsoft.AspNetCore.App.Runtime.csproj
```diff
@@ -265,11 +265,21 @@ This package is an internal implementation of the .NET Core SDK and is not meant
...
+
+      <_PackageFiles Include="@(RuntimePackContent)" />
```
dougbu: Odd to update an internal item group of another target. Suggest using an item group `_GetPackageFiles` reads from e.g. `@(Content)` or `@(None)`.

dougbu (dismissed):

src/Framework/src/Microsoft.AspNetCore.App.Runtime.csproj
JunTaoLuo: I was copying the logic from App.Ref.csproj. Seems like the indirection was introduced in https://github.com/aspnet/AspNetCore/commit/cc065f0055b72401a666f9eb2bb579bcc37870ce. @natemcmaster any reason for this? If not I'll just use the Content itemgroup.

natemcmaster: This implementation is fine, no compelling reason to use content or none item groups IMO.

natemcmaster (approved):

src/Framework/src/Microsoft.AspNetCore.App.Runtime.csproj
JunTaoLuo: I think Nate and I feel the current implementation is okay.

PR #11048, opened 2019-06-10
[Identity] Disable polling on file providers

javiercn:


src/Identity/testassets/Identity.DefaultUI.WebSite/StartupBase.cs
```diff
@@ -42,10 +42,11 @@ public virtual void ConfigureServices(IServiceCollection services)
             services.AddDbContext<TContext>(options =>
                 options
                     .ConfigureWarnings(b => b.Log(CoreEventId.ManyServiceProvidersCreatedWarning))
-                    .UseSqlServer(
-                        Configuration.GetConnectionString("DefaultConnection"),
-                        sqlOptions => sqlOptions.MigrationsAssembly("Identity.DefaultUI.WebSite")
-                    ));
+                    //.UseSqlServer(
```
HaoK: Nit: cleanup?

HaoK (approved):

src/Identity/testassets/Identity.DefaultUI.WebSite/StartupBase.cs
javiercn: I left it commented on purpose so that its easy to test with SQL Server. I need to follow up with @ajcvickers on why SetupTestDatabase wasn't overriding it, which is what I was expecting to happen.

PR #11049, opened 2019-06-10
Make sure to release connection lock

mikaelm12:
This is specifically to fix https://github.com/aspnet/AspNetCore-Internal/issues/2465
When acquiring the lock, if the connection wasn't in an active state we would get the lock and not release it. This was causing the timeouts when we tried to dispose.

### DIFF
--- a/src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
+++ b/src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
@@ -1948,6 +1948,7 @@ namespace Microsoft.AspNetCore.SignalR.Client
 
                 if (CurrentConnectionStateUnsynchronized == null || CurrentConnectionStateUnsynchronized.Stopping)
                 {
+                    ReleaseConnectionLock();
                     throw new InvalidOperationException($"The '{methodName}' method cannot be called if the connection is not active");
                 }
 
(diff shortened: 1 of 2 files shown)


src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -1948,6 +1948,7 @@ public async Task<ConnectionState> WaitForActiveConnectionAsync(string methodNam
 
                 if (CurrentConnectionStateUnsynchronized == null || CurrentConnectionStateUnsynchronized.Stopping)
                 {
+                    ReleaseConnectionLock();
```
BrennanConroy: Add `methodName`

BrennanConroy: Add a comment to the method about not putting it in a try/finally

BrennanConroy: Remove the flaky test attribute!

BrennanConroy (approved):

analogrelay (approved):

src/SignalR/clients/csharp/Client/test/FunctionalTests/HubConnectionTests.cs
```diff
@@ -770,7 +770,6 @@ async IAsyncEnumerable<string> clientStreamData()
         }
 
         [Theory]
-        [Flaky("https://github.com/aspnet/AspNetCore-Internal/issues/2465", FlakyOn.All)]
         [MemberData(nameof(HubProtocolsAndTransportsAndHubPaths))]
```
jkotalik: nit: consider adding [Repeat(int)] here to help diagnose faster if the test is flaky.

mikaelm12: Talked irl

PR #11051, opened 2019-06-10
Fix typo in condition for Windows Auth with LongPolling

BrennanConroy:
Just something I found while reading through code.

### DIFF
--- a/src/SignalR/common/Http.Connections/src/Internal/HttpConnectionDispatcher.cs
+++ b/src/SignalR/common/Http.Connections/src/Internal/HttpConnectionDispatcher.cs
@@ -504,7 +504,7 @@ namespace Microsoft.AspNetCore.Http.Connections.Internal
                     // We specifically clone the identity on first poll if it's a windows identity
                     // If we swapped the new User here we'd have to dispose the old identities which could race with the application
                     // trying to access the identity.
-                    if (context.User.Identity is WindowsIdentity)
+                    if (context.User.Identity is WindowsIdentity == false)
                     {
                         existing.User = context.User;
                     }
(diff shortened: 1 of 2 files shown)


src/SignalR/common/Http.Connections/src/Internal/HttpConnectionDispatcher.cs
```diff
@@ -504,7 +504,7 @@ private async Task<bool> EnsureConnectionStateAsync(HttpConnectionContext connec
                     // We specifically clone the identity on first poll if it's a windows identity
                     // If we swapped the new User here we'd have to dispose the old identities which could race with the application
                     // trying to access the identity.
-                    if (context.User.Identity is WindowsIdentity)
+                    if (context.User.Identity is WindowsIdentity == false)
```
anurse:
```suggestion
                    if (context.User.Identity isn't WindowsIdentity)
```

If only... :) (to be clear, this doesn't exist I'm just :trollface: ing)

mikaelm12: We should actually put parens around the first clause though.

BrennanConroy: Should we? There isn't a question of order of operations here, there is only one evaluation of the if statement.

mikaelm12: You don't *have* to, but I think it's a good idea to put parens around individual logical clauses when chaining. I wouldn't block the pr on it though.

halter73: I like the parens here. I think it just reads better. Or what about the following?

```suggestion
                    if (!(context.User.Identity is WindowsIdentity))
```

halter73 (approved):

src/SignalR/common/Http.Connections/src/Internal/HttpConnectionDispatcher.cs
anurse: I like this better than `== false` I think. Someday we'll get the `is not` operator...

BrennanConroy:
@aspnet/build Can someone merge please! The build hang isn't caused by this PR

PR #11053, opened 2019-06-10
React to System.Text.Json namespace changes

sebastienros:


sebastienros:
This app doesn't impact the CI, if someone wants to merge it so it doesn't take too much resources, and it unblocks the benchmarks, then thanks.

benaadams:
This is why Platform benchmarks are broken?

sebastienros:
Yep

sebastienros:
Can someone merge it? Not sure if I can't because there is not review of the failing helix step.

davidfowl (approved):

PR #11062, opened 2019-06-10
Small fixes for DatabaseErrorPageMiddleware

ajcvickers:
The main change here is to also allow public `HasTables` to be found. This will be removed once we get the new EF merged.

The change to the migration makes it align with what we would generate, but I don't think has any functional impact on the tests.

See also https://github.com/aspnet/EntityFrameworkCore/pull/16024 which (I really hope!) fixes the other failures once it makes it here.

### DIFF
--- a/src/Middleware/Diagnostics.EntityFrameworkCore/src/DatabaseErrorPageMiddleware.cs
+++ b/src/Middleware/Diagnostics.EntityFrameworkCore/src/DatabaseErrorPageMiddleware.cs
@@ -135,7 +135,7 @@ namespace Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore
                                 if (databaseExists)
                                 {
                                     // Also check if the database is completely empty - see https://github.com/aspnet/EntityFrameworkCore/issues/15997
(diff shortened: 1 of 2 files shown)


smitpatel (approved):

dougbu (approved):
Is all of this going away once we have an official build containing the aspnet/EntityFrameworkCore#16024 fix? If so, I'd just as soon wait because we can manually run `darc` on the #10939 branch. Otherwise, looks fine.

ajcvickers:
@dougbu It's a chicken-and-egg situation. The current code in master won't build with the new EF, but we need the new EF to fix the other issues. By merging this to master, it should mean that everything works when the new EF flows through.

We could wait until that happens and then edit that PR, but this way we don't need to do that.

dougbu:
:ok: might save us an hour or two before 'master' builds cleanly&hellip; (https://dnceng.visualstudio.com/internal/_build/results?buildId=219152 hasn't even started its Windows job, I need to react, and #10939 needs to be PR validated)

AndriySvyryd (approved):

PR #11063, opened 2019-06-10
Make CORS functional tests select different ports

jkotalik:
Fixes https://github.com/aspnet/AspNetCore-Internal/issues/2641. Also makes new test assets to preserve samples.


src/Middleware/CORS/test/testassets/TestDestination/Program.cs
```diff
@@ -0,0 +1,46 @@
+// Copyright (c) .NET Foundation. All rights reserved.
```
pranavkm: You could move them over instead of copying them. There isn't inherent value in the samples.

src/Middleware/CORS/test/FunctionalTests/test.js
```diff
@@ -1,8 +1,11 @@
-﻿const puppeteer = require('puppeteer');
+const puppeteer = require('puppeteer');
 const os = require("os");
-const hostname = os.hostname();
+const hostname = "localhost";
+const originPortVar = process.env.ORIGIN_PORT;
```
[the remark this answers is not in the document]
jkotalik: How would we know which defaults to use? The servers would need to listen to a port, which would be random, and that port needs to get propagated here.

src/Middleware/CORS/test/testassets/TestDestination/Program.cs
jkotalik: Cool if I delete the test.htm file then too? Means you can't run these locally without the deployers.

src/Middleware/CORS/test/FunctionalTests/test.js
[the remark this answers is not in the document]
Eilon: Total hack: When the .NET server starts, write it to disk (or somewhere), then read from this JS code? No idea if that would even work.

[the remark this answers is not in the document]
jkotalik: > Total hack: When the .NET server starts, write it to disk (or somewhere), then read from this JS code? No idea if that would even work.

 I mean yes that would work, but the underlying (and hard) problem is making sure the server starts up correctly. 

Thought: Keep the samples and have test.js fallback to 9000/9001 if they aren't set?

src/Middleware/CORS/test/testassets/TestDestination/Program.cs
pranavkm: Could you move that too? Yeah, that should be ok.

src/Middleware/CORS/test/FunctionalTests/test.js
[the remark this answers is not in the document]
pranavkm: If it's too complicated, I'd just leave it be. I can suffer through the needless pain and misery you inflict upon me 👍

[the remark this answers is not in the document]
jkotalik: > If it's too complicated, I'd just leave it be. I can suffer through the needless pain and misery you inflict upon me 👍

 To any outsiders, this is hyperbole 😉

jkotalik:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=219341&view=logs

aspnet-hello:

jkotalik:
Idk why I'm continuously hitting timeouts with this PR. It seems to be hanging arbitrarily. 

@pranavkm I think you said you were good with the change, but I could use a format approval 😉

src/Middleware/CORS/test/testassets/TestDestination/StartupWithoutEndpointRouting.cs
```diff
@@ -12,7 +13,7 @@ namespace SampleDestination
 {
     public class StartupWithoutEndpointRouting
     {
-        private static readonly string DefaultAllowedOrigin = $"http://{Dns.GetHostName()}:9001";
+        private static readonly string DefaultAllowedOrigin = $"http://localhost:{Environment.GetEnvironmentVariable("ORIGIN_PORT")}";
```
pranavkm: Missed this bit in my initial review. Could we use `Dns.GetHostName()`? Chrome will sometimes not do Cors if it's running on localhost. Using the hostname avoids this.

src/Middleware/CORS/test/testassets/TestOrigin/Program.cs
```diff
@@ -11,15 +13,10 @@ public class Program
...
-            var host = new WebHostBuilder()
-                .UseKestrel()
-                .UseUrls("http://+:9001", "http://+:9002")
-                .UseContentRoot(Directory.GetCurrentDirectory())
-                .ConfigureLogging(factory => factory.AddConsole())
-                .UseStartup<Startup>()
-                .Build();
-
-            host.Run();
+            using (var host = WebHost.CreateDefaultBuilder(args).UseStartup<Startup>().Build())
```
pranavkm: Part of the reason we use `http://+` was to force Kestrel to use the local host name (i..e. `http://mymachine:...` versus `http://localhost`). I think `CreateDefaultBuilder` does not give us the same behavior. Could you confirm?

Tratcher: Kestrel doesn't use machine names, it uses IPs. The difference between http://localhost and http://+ is that localhost binds to loopback (::1), and + binds to any machine IP. CreateDefaultBuilder doesn't set a default URL, Kestrel defaults to http://localhost:5000 of nothing is specified.

jkotalik:
@pranavkm _ping_

pranavkm (approved):

jkotalik:
@aspnet/build can you merge this? I've tried rerunning template tests a few times now and they keep failing due to timeout.

dougbu:
Sure 😺

PR #11065, opened 2019-06-10
Work around potential race in PipeWriter Redux

halter73:
Maybe we'll address #8843 this time.

### DIFF
--- a/src/Servers/Kestrel/Core/src/Internal/Infrastructure/TimingPipeFlusher.cs
+++ b/src/Servers/Kestrel/Core/src/Internal/Infrastructure/TimingPipeFlusher.cs
@@ -51,35 +51,39 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure
 
         public ValueTask<FlushResult> FlushAsync(MinDataRate minRate, long count, IHttpOutputAborter outputAborter, CancellationToken cancellationToken)
         {
-            var flushValueTask = _writer.FlushAsync(cancellationToken);
-
-            if (minRate != null)
-            {
-                _timeoutControl.BytesWrittenToBuffer(minRate, count);
-            }
-
-            if (flushValueTask.IsCompletedSuccessfully)
-            {
-                return new ValueTask<FlushResult>(flushValueTask.Result);
-            }
-
             // https://github.com/dotnet/corefxlab/issues/1334
             // Pipelines don't support multiple awaiters on flush.
-            // While it's acceptable to call PipeWriter.FlushAsync again before the last FlushAsync completes,
-            // it is not acceptable to attach a new continuation (via await, AsTask(), etc..). In this case,
-            // we find previous flush Task which still accounts for any newly committed bytes and await that.
             lock (_flushLock)
             {
-                if (_lastFlushTask == null || _lastFlushTask.IsCompleted)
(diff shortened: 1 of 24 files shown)


jkotalik:
Can you try to rerun `AngularTemplate_Works` multiple times as that test repro'd the race fairly consistently?

jkotalik:
> /home/vsts/work/1/s/src/Servers/Kestrel/perf/Kestrel.Performance/Http1LargeWritingBenchmark.cs(14,53): error CS0234: The type or namespace name 'Abstractions' does not exist in the namespace 'Microsoft.AspNetCore.Server.Kestrel.Transport' (are you missing an assembly reference?) [/home/vsts/work/1/s/src/Servers/Kestrel/perf/Kestrel.Performance/Microsoft.AspNetCore.Server.Kestrel.Performance.csproj]

src/Servers/Kestrel/Core/src/Internal/Infrastructure/TimingPipeFlusher.cs
```diff
@@ -22,7 +22,13 @@ internal class TimingPipeFlusher
         private readonly IKestrelTrace _log;
 
         private readonly object _flushLock = new object();
-        private Task<FlushResult> _lastFlushTask = null;
+
+        // This ValueTask that was either:
+        // 1. The default value which "IsCompleted"
```
jkotalik:
```suggestion
        // 1. The default value where "IsCompleted" is true
```

jkotalik: Also, comment on how _lastFlushTask must only be get/set under the _flushLock.

jkotalik (approved):
Really make sure `AngularTemplate_Works` passes multiple times.

ryanbrandenburg (commented):
Templates parts seem fine. Someone with Kestrel knowledge should give final approval.

src/ProjectTemplates/test/SpaTemplateTest/AngularTemplateTest.cs
```diff
@@ -17,8 +17,13 @@ public AngularTemplateTest(ProjectFactoryFixture projectFactory, BrowserFixture
             : base(projectFactory, browserFixture, output) { }
 
         [Fact]
-        public Task AngularTemplate_Works()
```
jkotalik: Next time, just add [Repeat(10)] 😉 .

As long as you revert this after it all passes, 🚢 🇮🇹

jkotalik:
I got a feeling the template tests are not fond of being run multiple times in a row.

src/Shared/Buffers.MemoryPool/MemoryPoolFactory.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace System.Buffers
 {
-    internal static class MemoryPoolFactory
+    internal static class SlabMemoryPoolFactory
```
davidfowl: LOL

halter73: You can't say I didn't warn you :trollface:

davidfowl (approved):
What's the performance delta?

halter73:
> What's the performance delta?

I can see no reason for it to be any different than https://github.com/aspnet/AspNetCore/pull/10165#issuecomment-491469419, and that took a while to run the first time.

PR #11066, opened 2019-06-10
Fix Complete in Message Bodies

### LINKED ISSUE #9600: Calling HttpContext.Request.BodyReader.Complete causes an exception
In ASP.NET Core 3.0 Preview 4 calling `HttpContext.Request.BodyReader.Complete()` eventually triggers an exception with the following call stack:

```
Http1ContentLengthMessageBody.ThrowIfCompleted() in Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http, Microsoft.AspNetCore.Server.Kestrel.Core.dll
Http1ContentLengthMessageBody.TryRead() in Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http, Microsoft.AspNetCore.Server.Kestrel.Core.dll
Http1MessageBody.OnConsumeAsync() in Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http, Microsoft.AspNetCore.Server.Kestrel.Core.dll
Http1ContentLengthMessageBody.ConsumeAsync() in Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http, Microsoft.AspNetCore.Server.Kestrel.Core.dll
async HttpProtocol.ProcessRequests<Microsoft.AspNetCore.Hosting.Internal.HostingApplication.Context>() in Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http, Microsoft.AspNetCore.Server.Kestrel.Core.dll
...
```

jkotalik:
Fixes https://github.com/aspnet/AspNetCore/issues/9600
This will still cause an application exception if you set one.

### DIFF
--- a/src/Servers/Kestrel/Core/src/Internal/Http/Http1UpgradeMessageBody.cs
+++ b/src/Servers/Kestrel/Core/src/Internal/Http/Http1UpgradeMessageBody.cs
@@ -14,7 +14,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
     /// </summary>
     internal sealed class Http1UpgradeMessageBody : Http1MessageBody
     {
-        public bool _completed;
         public Http1UpgradeMessageBody(Http1Connection context)
             : base(context)
         {
@@ -77,6 +76,16 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
         public override Task StopAsync()
         {
             return Task.CompletedTask;
+        }
+
+        public override bool TryReadInternal(out ReadResult readResult)
+        {
+            return _context.Input.TryRead(out readResult);
+        }
+
+        public override ValueTask<ReadResult> ReadAsyncInternal(CancellationToken cancellationToken = default)
+        {
+            return _context.Input.ReadAsync(cancellationToken);
         }
     }
 }
--- a/src/Servers/Kestrel/Core/test/MessageBodyTests.cs
+++ b/src/Servers/Kestrel/Core/test/MessageBodyTests.cs
@@ -1252,8 +1252,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Tests
 
                 input.Add("asdf");
 
-                reader.Complete();
                 reader.AdvanceTo(readResult.Buffer.End);
+                reader.Complete();
 
                 await body.ConsumeAsync();
             }
@@ -1284,7 +1284,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Tests
                 await body.StopAsync();
             }
         }
-
 
         [Fact]
         public async Task CompleteForChunkedAllowsConsumeToWork()
--- a/src/Servers/Kestrel/Core/test/MessageBodyTests.cs
+++ b/src/Servers/Kestrel/Core/test/MessageBodyTests.cs
@@ -1236,6 +1236,30 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Tests
         }
 
         [Fact]
+        public async Task CompleteForContentLengthAllowsConsumeToWork()
+        {
+            using (var input = new TestInput())
+            {
+                var body = Http1MessageBody.For(HttpVersion.Http11, new HttpRequestHeaders { HeaderContentLength = "5" }, input.Http1Connection);
+                var reader = new HttpRequestPipeReader();
+                reader.StartAcceptingReads(body);
+
+                input.Add("a");
+
+                Assert.True(reader.TryRead(out var readResult));
+
+                Assert.False(readResult.IsCompleted);
+
+                input.Add("asdf");
+
+                reader.Complete();
+                reader.AdvanceTo(readResult.Buffer.End);
+
+                await body.ConsumeAsync();
+            }
+        }
+
+        [Fact]
         public async Task CompleteForContentLengthDoesNotCompleteConnectionPipeMakesReadReturnThrow()
         {
             using (var input = new TestInput())
@@ -1258,6 +1282,31 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Tests
                 Assert.Throws<InvalidOperationException>(() => reader.TryRead(out readResult));
 
                 await body.StopAsync();
+            }
+        }
+
+
+        [Fact]
+        public async Task CompleteForChunkedAllowsConsumeToWork()
+        {
+            using (var input = new TestInput())
+            {
+                var body = Http1MessageBody.For(HttpVersion.Http11, new HttpRequestHeaders { HeaderTransferEncoding = "chunked" }, input.Http1Connection);
+                var reader = new HttpRequestPipeReader();
+                reader.StartAcceptingReads(body);
+
+                input.Add("5\r\nHello\r\n");
+
+                Assert.True(reader.TryRead(out var readResult));
+
+                Assert.False(readResult.IsCompleted);
+                reader.AdvanceTo(readResult.Buffer.End);
+
+                input.Add("1\r\nH\r\n0\r\n\r\n");
+
+                reader.Complete();
+
+                await body.ConsumeAsync();
             }
         }
 
@@ -1312,7 +1361,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Tests
                 await body.StopAsync();
             }
         }
-
 
         [Fact]
         public async Task CompleteForZeroByteBodyDoesNotCompleteConnectionPipeNoopsReads()
--- a/src/Servers/Kestrel/test/InMemory.FunctionalTests/ChunkedRequestTests.cs
+++ b/src/Servers/Kestrel/test/InMemory.FunctionalTests/ChunkedRequestTests.cs
@@ -1007,7 +1007,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.InMemory.FunctionalTests
                 var readResult = await request.BodyReader.ReadAsync();
                 request.BodyReader.AdvanceTo(readResult.Buffer.End);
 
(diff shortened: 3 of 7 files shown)


src/Servers/Kestrel/Core/test/MessageBodyTests.cs
```diff
@@ -1236,7 +1236,7 @@ public async Task OnWriterCompletedForNoContentLengthNoop()
         }
 
         [Fact]
-        public async Task CompleteForContentLengthDoesNotCompleteConnectionPipeMakesReadReturnThrow()
+        public async Task CompleteForContentLengthDoesNotCompleteConnectionPipeMakes()
```
Tratcher:
```suggestion
        public async Task CompleteForContentLengthDoesNotCompleteConnectionPipe()
```

src/Servers/Kestrel/test/InMemory.FunctionalTests/RequestTests.cs
```diff
@@ -57,6 +57,36 @@ public async Task StreamsAreNotPersistedAcrossRequests()
             }
         }
 
+        [Fact]
+        public async Task Test()
```
Tratcher: ?

jkotalik: It's been a long day.

davidfowl:
Part of me thinks it’s fine to throw when a user does ReadAsync after completing but you want to avoid throwing inside of kestrel on drain. We also probably won’t encourage people to complete anyways so maybe it’s not a big deal

jkotalik:
🆙 📅

src/Servers/Kestrel/test/InMemory.FunctionalTests/RequestTests.cs
```diff
@@ -57,6 +58,37 @@ public async Task StreamsAreNotPersistedAcrossRequests()
...
         }
 
+        [Fact]
+        public async Task CallingCompleteDoesNotLogGreaterThanInfo()
+        {
+            var testContext = new TestServiceContext(LoggerFactory);
+
+            await using (var server = new TestServer(async context =>
+            {
+                // This will hang if 0 content length is not assumed by the server
```
halter73: What will hang when?

src/Servers/Kestrel/Core/test/MessageBodyTests.cs
```diff
@@ -1255,14 +1255,12 @@ public async Task CompleteForContentLengthDoesNotCompleteConnectionPipeMakesRead
                 reader.Complete();
                 reader.AdvanceTo(readResult.Buffer.End);
 
-                Assert.Throws<InvalidOperationException>(() => reader.TryRead(out readResult));
```
halter73: Why remove this line? Shouldn't we be verifying that read doesn't trhow after calling Complete()?

src/Servers/Kestrel/Core/test/MessageBodyTests.cs
```diff
@@ -1281,14 +1279,12 @@ public async Task CompleteForChunkedDoesNotCompleteConnectionPipeMakesReadThrow(
 
                 reader.Complete();
 
-                Assert.Throws<InvalidOperationException>(() => reader.TryRead(out readResult));
```
halter73: ditto

src/Servers/Kestrel/Core/test/MessageBodyTests.cs
```diff
@@ -1307,13 +1303,10 @@ public async Task CompleteForUpgradeDoesNotCompleteConnectionPipeMakesReadThrow(
...
```
halter73: ditto

src/Servers/Kestrel/Core/test/MessageBodyTests.cs
jkotalik: Yeah probably, but ¯\\_(ツ)_/¯

src/Servers/Kestrel/test/InMemory.FunctionalTests/RequestTests.cs
jkotalik: This comment was a bunch of spaghetti. Let me rethink this.

halter73: That said, I don't think you need to post a body in this test. Kestrel will create a log warning the app didn't consume the request body. Today, I think it is info-level, but that could change. If it started to be logged as a warning, this test would start failing.

src/Servers/Kestrel/test/InMemory.FunctionalTests/RequestTests.cs
```diff
@@ -57,6 +58,37 @@ public async Task StreamsAreNotPersistedAcrossRequests()
...
+            {
+                context.Request.BodyReader.Complete();
+                await Task.CompletedTask;
+            }, testContext))
+            {
+                using (var connection = server.CreateConnection())
+                {
+                    await connection.Send(
+                        "POST / HTTP/1.0",
+                        "Content-Length: 11",
```
halter73: Just send a GET so any logs about not consuming the request body don't interfere with this test.

src/Servers/Kestrel/test/InMemory.FunctionalTests/RequestTests.cs
jkotalik: I think I do here because I'm verifying the behavior of ConsumeAsync which I believe requires a request body to be there.

src/Servers/Kestrel/test/InMemory.FunctionalTests/RequestTests.cs
jkotalik: I do need to send a post. Why is that log not consistent? Shouldn't it either always or never be logged? 

If it isn't consistent, I can change the max loglevel to warning.

halter73: I'm talking specifically about the RequestBodyNotEntirelyRead info-level log. I'm just pointing out that could be made higher severity later. That said, it now occurs to me that this test might be verifying the RequestBodyNotEntirelyRead log itself isn't higher than info level. Is that the case?

jkotalik: The goal of this test is to make sure that an exception wasn't thrown from ConsumeAsync if the pipe is completed. I'd be fine with this test breaking if we increase the severity later.

halter73: I don't think I read this correctly the first time. I think this is OK.

src/Servers/Kestrel/Core/test/MessageBodyTests.cs
```diff
@@ -1261,6 +1285,31 @@ public async Task CompleteForContentLengthDoesNotCompleteConnectionPipeMakesRead
             }
         }
 
+
```
davidfowl: nit: Extra space.

src/Servers/Kestrel/Core/test/MessageBodyTests.cs
```diff
@@ -1235,6 +1235,30 @@ public async Task OnWriterCompletedForNoContentLengthNoop()
...
+                input.Add("a");
+
+                Assert.True(reader.TryRead(out var readResult));
+
+                Assert.False(readResult.IsCompleted);
+
+                input.Add("asdf");
+
+                reader.Complete();
+                reader.AdvanceTo(readResult.Buffer.End);
```
davidfowl: AdvanceTo after complete is weird.

davidfowl:
Can you add a functional test as well? In memory is fine.

jkotalik:
I had one before that I forced pushed over... I'll revive it shortly.

src/Servers/Kestrel/test/InMemory.FunctionalTests/ChunkedRequestTests.cs
```diff
@@ -991,6 +992,62 @@ public async Task ChunkedRequestCallCompleteThrowsExceptionOnRead()
...
+            {
+                var request = httpContext.Request;
+
+                // This read may receive all data, but what we care about
+                // is that ConsumeAsync is called and doesn't error. Calling
+                // TryRead before would always fail.
+                var readResult = await request.BodyReader.ReadAsync();
+                request.BodyReader.AdvanceTo(readResult.Buffer.End);
+
+                httpContext.Request.BodyReader.Complete();
```
davidfowl: nit: Use the request local?

jkotalik:
```suggestion
                request.BodyReader.Complete();
```

src/Servers/Kestrel/Core/src/Internal/Http/Http1UpgradeMessageBody.cs
```diff
@@ -78,5 +77,15 @@ public override Task StopAsync()
         {
             return Task.CompletedTask;
         }
+
+        public override bool TryReadInternal(out ReadResult readResult)
```
halter73: It might be clearer to call this unchecked instead of internal

halter73 (approved):

src/Servers/Kestrel/Core/src/Internal/Http/Http1UpgradeMessageBody.cs
jkotalik: I've seen the Internal pattern with a higher frequency. If you feel strongly about it I can change it, but I personally prefer Internal.

halter73: I don't feel strongly at all. I agree Internal usually implies no validation, so I'm fine as is.

PR #11067, opened 2019-06-10
Make the test input a bit more reasonable

pranavkm:


sebastienros (approved):

dougbu:
@pranavkm what's the context here? How does changing the test input make its execution less flaky?

sebastienros:
This is not a valid French number on Linux apparently. Maybe should file another issue on corefx?

dougbu:
> This is not a valid French number on Linux apparently. Maybe should file another issue on corefx?

I've filed similar issues in the past. CoreFx generally relies on the underlying platform for globalization information. You would probably just get a redirect to ICU (https://github.com/unicode-org/icu and http://site.icu-project.org/) -- the Ubuntu component providing number formats (I think). If the information isn't in there, it's in CLDR (http://cldr.unicode.org/index).

More generally, feel free to let _someone_ know France is more forgiving than something thinks (about numbers).

pranavkm:
I'll file an issue with CoreFx. That said, jQuery Validation does not allow it, neither does browser validation (browsers remove spaces if you paste in a number with spaces). I'm questioning the need for MVC to test for validating numbers with spaces in them.

mkArtakMSFT (approved):

Issue #2711, opened 2018-01-02
Unable to upload file greater than 2GB in size


aspnet-hello:
_From @kagoyal on Wednesday, February 1, 2017 3:11:34 PM_

I am trying to upload file greater than 2GB using multipart formdata. 

The issue I am observing is after 2GB upload, read of input stream gets stuck and process hangs.
Microsoft.AspNetCore.Server.Kestrel version we are using is "1.1.0".

HTTP request’s response never come back and timesout after one hour (execution time we have set).
I have modified IIS post limits for upto 4GB, also in ConfigureServices –

                services.Configure<FormOptions>(x =>
                {
                    x.MultipartBodyLengthLimit = 4294967296; 
                });

I am following the sample here - https://github.com/aspnet/Docs/tree/master/aspnetcore/mvc/models/file-uploads/sample/FileUploadSample to read multipart data.

Please let me know if any more information is required to debug this, or if this is expected behavior.

_Copied from original issue: aspnet/HttpAbstractions#766_

aspnet-hello:
_From @benaadams on Wednesday, February 1, 2017 4:11:23 PM_

What's the upload speed/transfer rate?

aspnet-hello:
_From @benaadams on Wednesday, February 1, 2017 4:13:31 PM_

/cc @Eilon

aspnet-hello:
_From @kagoyal on Wednesday, February 1, 2017 4:15:33 PM_

I am able to upload approx. 1.8 GB of data in about 30-35 seconds.

aspnet-hello:
_From @blowdart on Wednesday, February 1, 2017 4:18:46 PM_

Well the timeout is working, so not a security concern, just a weird bug. @muratg

aspnet-hello:
_From @JunTaoLuo on Wednesday, March 1, 2017 2:46:26 PM_

I am able to upload 4GB files when using Kestrel directly but I ran into problems when running it on IIS. I was not able to produce a hang but I do see a 502 failure response. It seems like the connection is being reset and it could be due to ANCM. I filed https://github.com/aspnet/AspNetCoreModule/issues/78.

aspnet-hello:
_From @JunTaoLuo on Wednesday, March 1, 2017 7:26:39 PM_

[...]

PR #11068, opened 2019-06-11
Handle Content Length header greater than 2GB on uploads

ryanski44:
Summary of the changes
 - Use atoll instead of atol when converting the string content length to the value (solves 2GB limit)
 - If content length > 4 GB call WinHttpSendRequest with 0 (WINHTTP_IGNORE_REQUEST_TOTAL_LENGTH) for the content length to indicate that the content length is greater than 4 GB. See https://docs.microsoft.com/en-us/windows/desktop/api/winhttp/nf-winhttp-winhttpsendrequest

Addresses #2711

### DIFF
--- a/src/Servers/IIS/AspNetCoreModuleV2/OutOfProcessRequestHandler/forwardinghandler.cpp
+++ b/src/Servers/IIS/AspNetCoreModuleV2/OutOfProcessRequestHandler/forwardinghandler.cpp
@@ -216,19 +216,26 @@ FORWARDING_HANDLER::ExecuteRequestHandler()
     //
     // Calculate the bytes to receive from the content length.
     //
-    DWORD cbContentLength = 0;
+    DWORD cbContentLength = WINHTTP_IGNORE_REQUEST_TOTAL_LENGTH;
     PCSTR pszContentLength = pRequest->GetHeader(HttpHeaderContentLength);
     if (pszContentLength != NULL)
     {
-        cbContentLength = m_BytesToReceive = atol(pszContentLength);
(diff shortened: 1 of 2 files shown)


jkotalik:
The PR itself looks good, but I'd prefer to have an automated test that we can run against this change. Can you try to add a test where you upload a payload larger than 2gb? See https://github.com/aspnet/AspNetCore/blob/master/src/Servers/IIS/IIS/test/Common.FunctionalTests/OutOfProcess/GlobalVersionTests.cs#L30 for an example where we publish/deploy ANCM out of process.

Rubenisme:
Would be great to get this. @jkotalik will we see this soon? I can imagine you're super busy with the .NET Core 3.0 release you merged the other day, so no pressure but just curious on where we are at with this.

jkotalik:
@Rubenisme we shall see. This probably won't land in the 3.0 timeframe, potentially in the 5.0 timeframe. This PR is stale at this point, so if it gets revived/ someone on our team has bandwidth, we may take over it ourselves.

anurse:
I'm going to go ahead and close this since it's gone stale since we last requested some changes. @ryanski44 if you'd like to pick it up again, feel free to re-open or open a new one. @Rubenisme if you're interested in taking it on, feel free to open a PR against `master`.

> This probably won't land in the 3.0 timeframe

Yeah, it's very unlikely it'll land for 3.0, but 5.0 is nice and open!

ryanski44:
@Rubenisme,

I haven't had any free time to write tests for this.

PR #11070, opened 2019-06-11
[SignalR] Pass a resource into IPolicyEvaluator for Hub method auth

### LINKED ISSUE #5279: Should SignalR pass a resource into IPolicyEvaluator?
MVC passes in the AuthorizationFilterContext.  SignalR should decide what resource is appropriate (if any)

BrennanConroy:
Fixes https://github.com/aspnet/AspNetCore/issues/5279

I'd like to get some feedback on what we should be passing into the resource. Currently I'm using the `HubCallerContext` and the hub method name and hub method arguments.


src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs
```diff
@@ -479,22 +479,22 @@ private static async Task<object> ExecuteHubMethod(ObjectMethodExecutor methodEx
         private void InitializeHub(THub hub, HubConnectionContext connection)
         {
             hub.Clients = new HubCallerClients(_hubContext.Clients, connection.ConnectionId);
-            hub.Context = new DefaultHubCallerContext(connection);
+            hub.Context = connection.HubCallerContext;
```
davidfowl: YESSS

src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs
```diff
@@ -479,22 +479,22 @@ private static async Task<object> ExecuteHubMethod(ObjectMethodExecutor methodEx
...
+        private Task<bool> IsHubMethodAuthorized(IServiceProvider provider, HubConnectionContext hubConnectionContext, IList<IAuthorizeData> policies, string hubMethodName, object[] hubMethodArguments)
         {
             // If there are no policies we don't need to run auth
             if (!policies.Any())
             {
                 return TaskCache.True;
             }
 
-            return IsHubMethodAuthorizedSlow(provider, principal, policies);
+            return IsHubMethodAuthorizedSlow(provider, hubConnectionContext.User, policies, new AuthHubConnectionContext(hubConnectionContext, hubMethodName, hubMethodArguments));
```
davidfowl: Noo, allocations...

BrennanConroy: Haha, I know, it's draft for a reason

src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs
```diff
@@ -570,4 +570,16 @@ public override IReadOnlyList<Type> GetParameterTypes(string methodName)
...
+    public class AuthHubConnectionContext : DefaultHubCallerContext
+    {
+        public AuthHubConnectionContext(HubConnectionContext connection, string hubMethodName, object[] hubMethodArguments) : base(connection)
+        {
+            HubMethodName = hubMethodName;
+            HubMethodArguments = hubMethodArguments;
+        }
+
+        public string HubMethodName { get; }
+        public IReadOnlyList<object> HubMethodArguments { get; }
```
anurse: Open question (which @BrennanConroy and I talked about earlier): Should the arguments be included?

src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs
anurse: The tricky thing is figuring out how to avoid this allocation. We could pool them maybe but that's a bunch of work. What about skipping this whole path if there are no policies?

BrennanConroy: >What about skipping this whole path if there are no policies?

We already do that https://github.com/aspnet/AspNetCore/pull/11070#discussion-diff-292252496R489

anurse: Does that temper your concern about allocations @davidfowl ? If there is no auth policy on a hub method we won't allocate.

src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs
```diff
@@ -503,7 +503,7 @@ private static async Task<bool> IsHubMethodAuthorizedSlow(IServiceProvider provi
             // AuthorizationPolicy.CombineAsync only returns null if there are no policies and we check that above
             Debug.Assert(authorizePolicy != null);
 
-            var authorizationResult = await authService.AuthorizeAsync(principal, authorizePolicy);
+            var authorizationResult = await authService.AuthorizeAsync(principal, resource, authorizePolicy);
```
anurse: `resource` is an `object` though right? So this will just box it anyway.

BrennanConroy: That is correct :(

javiercn: Can I get to the HttpContext from here?

javiercn: I want to use the query string from the original negotiate request, which I think its fair game?

BrennanConroy: >Can I get to the HttpContext

Yes.

src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs
javiercn: The methods will be very useful for cases where the authorization depends on something that is not on the original negotiate request.

src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs
anurse: The resource gives you `HubCallerContext` which has a `.GetHttpContext` extension method.

src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs
BrennanConroy: So far no one has said anything about arguments.

Maybe we shouldn't include them.

src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs
halter73: Is there some standard way that IAuthorizationHandler can use an HttpContext today? Is the `resource` normally the HttpContext itself?

src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs
halter73: Why is this exciting?

BrennanConroy:
Because this is removing one of our allocations from the hot path.
________________________________

src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs
BrennanConroy: Because this is removing one of our allocations from the hot path.

src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs
anurse: > Is there some standard way that IAuthorizationHandler can use an HttpContext today?

Not unless the handler uses `IHttpContextAccessor`.

> Is the resource normally the HttpContext itself?

Depends on what best describes the resource being authenticated. In MVC it's an [`AuthorizationFilterContext`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.filters.authorizationfiltercontext?view=aspnetcore-2.2) which includes `HttpContext` but also the resolved action information.

src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs
BrennanConroy: Made it no longer allocate, but it does rely on serial hub invocations

anurse: Honestly I prefer the allocation but I'm not @davidfowl ;). We already know we don't want to have serial hub invocations long-term, so adding more things that depend on that behavior is just setting us up to regress perf later.

anurse: I'd rather do the "correct" thing in terms of our design plans and then **pro**gress perf later by finding a way to reduce the allocations.

davidfowl: Better to allocate.

src/SignalR/server/SignalR/test/HubConnectionHandlerTests.cs
```diff
@@ -2218,6 +2218,11 @@ public Task HandleAsync(AuthorizationHandlerContext context)
             }
         }
 
+        internal class HttpContextFeatureImpl : IHttpContextFeature
```
davidfowl: Move this to the bottom of the file.

BrennanConroy:
This is ready for review now

halter73 (approved):

davidfowl:
I'm not a fan of this context being tied to auth. I imagine we'd do something like this once we have the hub invocation pipeline again. Do we need to call it an auth context?

BrennanConroy:
Other names welcome :)

anurse:
Call it `HubInvocationContext`. Then we could use it in the pipeline too when we have it. It's accurate and descriptive since this is contextual information about a Hub invocation.

BrennanConroy:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=233873&view=logs

aspnet-hello:

I've triaged the above build.

Issue #4081, opened 2018-11-19
Kestrel default certificate store is invalid


miloush:
### Describe the bug
The JSON examples in [Kestrel web server implementation in ASP.NET Core]https://docs.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel?view=aspnetcore-2.1#listenoptionsusehttps section *ListenOptions.UseHttps* suggest the following:

```JSON
"Store": "<cert store; defaults to My>",
```

However, that is not the case. If "Store" is left out from the app settings, `Internal.Cryptography.CryptoThrowHelper.WindowsCryptographicException: 'The parameter is incorrect'` exception is thrown. Stack suggests that the store's default is not "My":

`Microsoft.AspNetCore.Server.Kestrel.Core.dll!Microsoft.AspNetCore.Server.Kestrel.Https.Internal.CertificateLoader.LoadFromStoreCert(string subject = "...", string storeName = null, System.Security.Cryptography.X509Certificates.StoreLocation storeLocation = CurrentUser, bool allowInvalid = true)`

### Expected behavior

Fix the configuration loader to supply "My" as default for "Store" or eventually fix the documentation so that it says something like
```JSON
"Store": "<cert store; required>",
```

### Additional context
```
.NET Core SDK (reflecting any global.json):
 Version:   2.1.500
 Commit:    b68b931422

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.18272
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\2.1.500\

Host (useful for support):
  Version: 2.1.6
  Commit:  3f4f8eebd8

.NET Core SDKs installed:
  1.0.0-preview3-003788 [C:\Program Files\dotnet\sdk]
  2.1.202 [C:\Program Files\dotnet\sdk]
  2.1.500-preview-009404 [C:\Program Files\dotnet\sdk]
  2.1.500 [C:\Program Files\dotnet\sdk]

.NET Core runtimes installed:
  Microsoft.AspNetCore.All 2.1.1 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.1.5 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.1.6 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
[...]

PR #11074, opened 2019-06-11
Resolve Kestrel default certificate store is invalid

wu-yafeng:
Resolve Kestrel default certificate store is invalid

Addresses #4081


jkotalik (approved):
Change itself seems good to me, but I _think_ I would like an automated test for this change. @Tratcher @halter73 thoughts?

Tratcher:
No, testing the cert store is too dependent on machine state. This is fine so long as we manually verify it.

halter73 (approved):

Tratcher:
Thanks

PR #11076, opened 2019-06-11
Don't strip amr and auth_time in OIDC by default

### LINKED ISSUE #10494: Remove AMR and auth_time from the OIDC filter list
https://github.com/aspnet/AspNetCore/blob/2948c81aea2e5a947dc80fe4a62fa7150a9c6f94/src/Security/Authentication/OpenIdConnect/src/OpenIdConnectOptions.cs#L51-L58

These are used to gauge the strength and freshness of the login.

They are also not large values so they don't contribute significantly to the cookie size.

@brockallen

HaoK:
Fixes https://github.com/aspnet/AspNetCore/issues/10494


analogrelay (approved):
Such code. Very change. Many remove. :)

Tratcher (approved):

PR #10674, opened 2019-05-31
Replace Internal.AspNetCore.Sdk with Microsoft.DotNet.Arcade.Sdk

natemcmaster:
This is part of #7280. 

Here are the high-level changes:
* Change location of dotnet.exe to `$repoRoot/dotnet.exe` instead of `$repoRoot/x64/dotnet.exe`
* Remove our own implementation of versioning properties and targets
* Remove licenseUrl from all nuget packages and use PackageLicenseExpression instead
* Remove a bunch of properties and targets from our code that are provided by Microsoft.DotNet.Arcade.Sdk (or are unnecessary now)
* Add a handful of workaround to override defaults from Arcade which don't work for this repo

Work remaining to complete #7280 (will be done in a separate PR)
* Replace KoreBuild usages in build scripts with Arcade scripts
* Update Azure pipelines to use Arcade YAML templates

FYI - marking this as a draft PR until I can get CI to pass. Local builds are working, but still need to validate CI output and tests don't change substantially.

/cc @Pilchie @markwilkie @tmat @aspnet/build

PR #11077, opened 2019-06-11
Set official build ID

natemcmaster:
Fix regression caused by #10674. Before this, CI picked up BUILD_BUILDNUMBER automatically from CI.

Arcade, on the other hand, requires that you explicitly set `/p:OfficialBuildId=$(Build.BuildNumber)` in order to get builds like `3.0.0-preview7.build.number`


natemcmaster:
Internal test build: https://dev.azure.com/dnceng/internal/_build/results?buildId=219621&view=logs

dougbu (approved):
If internal test build works&hellip;

PR #11079, opened 2019-06-11
Add TryAddScheme to IAuthenticationSchemeProvider

### LINKED ISSUE #4666: IAuthenticationSchemeProvider needs a TryAddScheme()
Looking at this example/test to dynamically add authentication schemes at runtime:

https://github.com/aspnet/Security/blob/0959c941b40086b131b1e01d304fc23bb887fdc8/test/Microsoft.AspNetCore.Authentication.Test/DynamicSchemeTests.cs#L137-L143

If the endpoint is invoked twice, it will result in an `InvalidOperationException`: 
https://github.com/aspnet/HttpAbstractions/blob/dev/src/Microsoft.AspNetCore.Authentication.Core/AuthenticationSchemeProvider.cs#L121

Instead perhaps this method should be `bool TryAddScheme(AuthenticationScheme scheme)` so users don't have to wrap it in a try-catch.

Also it appears that the underlying `_map` should be a `ConcurrentDictionary`?

HaoK:
Fixes https://github.com/aspnet/AspNetCore/issues/4666

### DIFF
--- a/src/Http/Authentication.Abstractions/src/IAuthenticationSchemeProvider.cs
+++ b/src/Http/Authentication.Abstractions/src/IAuthenticationSchemeProvider.cs
@@ -76,7 +76,17 @@ namespace Microsoft.AspNetCore.Authentication
         /// </summary>
         /// <param name="scheme">The scheme.</param>
         /// <returns>true if the scheme was added successfully.</returns>
-        bool TryAddScheme(AuthenticationScheme scheme) => false;
+        bool TryAddScheme(AuthenticationScheme scheme)
+        {
+            try
+            {
+                AddScheme(scheme);
+                return true;
+            }
+            catch {
+                return false;
+            }
+        }
 
         /// <summary>
         /// Removes a scheme, preventing it from being used by <see cref="IAuthenticationService"/>.
--- a/src/Http/Authentication.Abstractions/src/IAuthenticationSchemeProvider.cs
+++ b/src/Http/Authentication.Abstractions/src/IAuthenticationSchemeProvider.cs
@@ -72,6 +72,13 @@ namespace Microsoft.AspNetCore.Authentication
         void AddScheme(AuthenticationScheme scheme);
 
         /// <summary>
+        /// Registers a scheme for use by <see cref="IAuthenticationService"/>. 
+        /// </summary>
+        /// <param name="scheme">The scheme.</param>
+        /// <returns>true if the scheme was added successfully.</returns>
+        bool TryAddScheme(AuthenticationScheme scheme) => false;
+
+        /// <summary>
         /// Removes a scheme, preventing it from being used by <see cref="IAuthenticationService"/>.
         /// </summary>
         /// <param name="name">The name of the authenticationScheme being removed.</param>
--- a/src/Http/Authentication.Core/src/AuthenticationSchemeProvider.cs
+++ b/src/Http/Authentication.Core/src/AuthenticationSchemeProvider.cs
@@ -133,17 +133,18 @@ namespace Microsoft.AspNetCore.Authentication
         /// Registers a scheme for use by <see cref="IAuthenticationService"/>. 
         /// </summary>
         /// <param name="scheme">The scheme.</param>
-        public virtual void AddScheme(AuthenticationScheme scheme)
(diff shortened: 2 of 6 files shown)


src/Http/Authentication.Abstractions/src/IAuthenticationSchemeProvider.cs
```diff
@@ -71,6 +71,13 @@ public interface IAuthenticationSchemeProvider
         /// <param name="scheme">The scheme.</param>
         void AddScheme(AuthenticationScheme scheme);
 
+        /// <summary>
+        /// Registers a scheme for use by <see cref="IAuthenticationService"/>. 
+        /// </summary>
+        /// <param name="scheme">The scheme.</param>
+        /// <returns>true if the scheme was added successfully.</returns>
+        bool TryAddScheme(AuthenticationScheme scheme) => false;
```
Tratcher: You didn't want to call AddScheme and return true? It works fine most of the time.

src/Http/Authentication.Core/src/AuthenticationSchemeProvider.cs
```diff
@@ -152,6 +153,26 @@ public virtual void AddScheme(AuthenticationScheme scheme)
...
+            }
+        }
+
+        /// <summary>
+        /// Registers a scheme for use by <see cref="IAuthenticationService"/>. 
+        /// </summary>
+        /// <param name="scheme">The scheme.</param>
+        public virtual void AddScheme(AuthenticationScheme scheme)
+        {
+            if (_schemes.ContainsKey(scheme.Name))
```
Tratcher: This check is redundant, AddScheme will succeed most of the time so there's no point optimizing for the lock.

Tratcher (approved):

Tratcher:
Hmm, doesn't compile on the CI? Is this our first use of default interface methods?

src/Http/Authentication.Abstractions/src/IAuthenticationSchemeProvider.cs
anurse: That depends on what the contract of this method is. What kind of "Try" is it? Is it "try and if there's an exception, return false instead" or is it "add it if it hasn't already been added"?

anurse: From the [original issue](https://github.com/aspnet/AspNetCore/issues/4666):

> Instead perhaps this method should be bool TryAddScheme(AuthenticationScheme scheme) so users don't have to wrap it in a try-catch.

Maybe the default implementation should be `try { AddScheme(scheme); return true } catch { return false; }` ?

HaoK: Yeah try catch is the correct behavior, this is mostly to prevent exceptions

anurse:
Crossgen appears to be failing on the DIM. We are a little behind on SDKs if I remember correctly, but we should have preview 5 (which I thought had DIMs).

anurse:
Are we still seeing blocking issues here? We need to land preview7 stuff soon, so perhaps this won't make it?

HaoK:
Yeah I'm still seeing crossgen issues using the preview6 sdk, I can try preview 7 but doesn't look like this will make it today

anurse:
Ok, let's plan to punt to preview8

anurse:
@Pilchie we're getting crossgen failures using Default Interface Methods here. Who can we loop in to help diagnose this? If crossgen is failing with DIMs that might be a problem :)

anurse:
First thing: It's not the SDK that brings crossgen, it's the runtime. I'm going to rebase this and see what happens.

Second: I'll start an email thread to see if we can sort out the crossgen issues.

anurse:
No dice on the first part :(

JunTaoLuo:
This PR has been open for quite a while, what are we doing about this?

HaoK:
I just need to rebase and try it again to see if cross gen is fixed now

HaoK:
Replaced by https://github.com/aspnet/AspNetCore/pull/14448 but cross gen is still failing in release/3.1 which might not have the fix yet, the expected fix for the cross gen is this PR: https://github.com/dotnet/coreclr/pull/25839

PR #11080, opened 2019-06-11
Try the sync context again maybe

davidfowl:
Trying something


anurse:
The big problem the sync context added for us was timeouts. IIRC the xunit sync context isn't about limiting the number of **tests** running at once but rather the number of available **threads** which means we can overload the queues if we have a bunch of active background tasks.

davidfowl:
yea I remember now, we need to figure out the starvation issue then.

Issue #11009, opened 2019-06-07
NegotiateHandlerFunctionalTests fails on Helix Win7 and Win8


natemcmaster:
src\Security\Authentication\Negotiate\test\Negotiate.FunctionalTest\Microsoft.AspNetCore.Authentication.Negotiate.FunctionalTest.csproj

This project was added a couple weeks ago, but never actually ran successfully on Helix due to it's project name not matching expected test name conventions until https://github.com/aspnet/AspNetCore/pull/10674.

cc @Tratcher

Tratcher:
Note to self: Most of the failures we saw could be attributed to HTTP/2 specific tests and those platforms not having ALPN. Those test can be skipped on those OS versions.  However, more tests were failing on Win7 than on Win8, and these additional tests did not give a clear error message. I'll need to debug those and see why.

PR #11101, opened 2019-06-11
Fix Negotiate tests downlevel

Tratcher:
#11009 These tests weren't running on Helix until recently due to a project naming convention issue. A few of them needed more skips due to downlevel support. All of the others gracefully fall back to HTTP/1.1.


natemcmaster (approved):

jkotalik (approved):
If it works!

Issue #5342, opened 2018-10-01
Url can't be changed for easy reconnect


benaadams:
Using a client-side cursor (e.g. `lastId` processed); for a reconnect this wants to be passed back to the server in the query string (as cors doesn't allow custom headers and cors is always applicable to websockets).

However while `signalR.HubConnectionBuilder()` provides easy construction of the connection; the url cannot be changed after building.

This means if the connection disconnects and you need to pass a cursor (e.g. id); the connection has to be rebuilt and all the events re-registered to the new connection.

It would be easier if the url could be updated and only `.start` needed to be called again.

muratg:
cc @halter73 because I saw the word "reconnect" :smiley:

halter73:
This wouldn't affect automatic reconnects. If we want this, we should probably set up a design meeting to determine things like how the URL should be updated (e.g. a property or `.start` param), when it can be updated, etc...

anurse:
Triage decision: Design this in the preview 6 time period.

anurse:
Proposal: Make `Url` settable, but it doesn't take affect until the next `StartAsync`/`start` call. We'd capture the current value in the connection state. We'd do this in the HubConnection layer (since `HttpConnection` is not reusable anyway).

benaadams:
Your proposal is acceptable... {insert "men in black" gif}

anurse:
Still planning to get this done... eventually 😫 

With the reconnect stuff, this should be doable in the `Reconnecting` event as well. @halter73 wants to try throwing `InvalidOperationException` if you try to set it while the connection is active (not reconnecting or disconnected). So sure, give that a whirl :)

anurse:
This is done in Java and JS. The .NET work is blocked on #10872 because we'll be refactoring that logic.

rjvdboon:
> 
> 
> The .NET work is blocked on #10872 because we'll be refactoring that logic.

It seems #10872 is done now

mikaelm12:
Yeah, #10872 came in right at the end of preview 7 so this will now go into the preview 8 milestone. I can go ahead and finish this up.

anurse:
We missed the 3.0 train. We'll revisit for 3.1

coonmoo:
Any news on getting the .NET SignalR client to support changing the url in the reconnecting event?
I see in the JS client this is already supported

PR #11102, opened 2019-06-11
Add ability to change url in TS client

mikaelm12:
Part of issue: #5342


src/SignalR/clients/ts/signalr/src/HubConnection.ts
```diff
@@ -127,6 +127,22 @@ export class HubConnection {
         return this.connection ? (this.connection.connectionId || null) : null;
     }
 
+    get baseUrl(): string | null {
+        return this.connection ? (this.connection.baseUrl || null) : null;
```
halter73: Can connection actually be null/undefined? Can baseUrl be?

src/SignalR/clients/ts/signalr/src/HubConnection.ts
```diff
@@ -127,6 +127,22 @@ export class HubConnection {
...
+    }
+
+    set baseUrl(newUrl: string | null) {
```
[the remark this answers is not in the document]
mikaelm12: This has to match the getter signature. I figured we'd want the getter to be able to return null as well. If not we can change it

src/SignalR/clients/ts/signalr/src/HubConnection.ts
mikaelm12: Connection shouldn't  be null. For baseUrl the check was there for non http scenarios but not really needed.

halter73: We can at least get rid of the ternary expression.

halter73: If we do eventually support non-http scenarios with the TS-client, we can just leave baseUrl as an empty string.

src/SignalR/clients/ts/signalr/src/HubConnection.ts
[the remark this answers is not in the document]
anurse: It has to match it exactly?

Disappointed

src/SignalR/clients/ts/signalr/tests/HubConnection.test.ts
```diff
@@ -41,6 +41,97 @@ describe("HubConnection", () => {
...
+                    hubConnection.baseUrl = "http://newurl.com";
+                    expect(hubConnection.baseUrl).not.toBe(originalUrl);
+                    expect(hubConnection.baseUrl).toBe("http://newurl.com");
+                } finally {
+                    await hubConnection.stop();
+                }
+            });
+        });
+
+        it("can change url in onclose", async () => {
```
halter73: Should we add a test to verify you cannot change the url in onreconnecting?

src/SignalR/clients/ts/signalr/src/HubConnection.ts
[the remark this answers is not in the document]
halter73: Just make it `string` instead of `string | null` and we can use empty string if/when we need to in the future.

src/SignalR/clients/ts/signalr/tests/HubConnection.test.ts
mikaelm12: I thought we did want to be able to change the while in the reconnecting event?

From the issue
> this should be doable in the Reconnecting event as well. @halter73 wants to try throwing InvalidOperationException if you try to set it while the connection is active (not reconnecting or disconnected).

src/SignalR/clients/ts/signalr/src/HubConnection.ts
```diff
@@ -127,6 +127,27 @@ export class HubConnection {
         return this.connection ? (this.connection.connectionId || null) : null;
     }
 
+    /** Indicates the url of the {@link HubConnection} to the server. */
+    get baseUrl(): string {
+        return this.connection.baseUrl || "";
+    }
+
+    /**
+     * Sets a new url for the HubConnection.
```
BrennanConroy: Should have some text explaining that this should only be set when the connection isn't active.

Also, not a fan of "newUrl" as the parameter in public API.

src/SignalR/clients/ts/signalr/tests/HubConnection.test.ts
BrennanConroy: Then we should have a test for during reconnecting?

halter73: Yeah my bad. We should add a test to verify you *can* change the url in onreconnecting.

src/SignalR/clients/ts/signalr/src/HubConnection.ts
```diff
@@ -127,6 +127,28 @@ export class HubConnection {
...
 
+    /** Indicates the url of the {@link HubConnection} to the server. */
+    get baseUrl(): string {
+        return this.connection.baseUrl || "";
+    }
+
+    /**
+     * Sets a new url for the HubConnection. Note that the url can only be changed when the connection is in either the Disconnected or
+     * Reconnecting states.
+     * @param {string} newUrl The url to connect to.
```
BrennanConroy:
```suggestion
     * @param {string} url The url to connect to.
```

BrennanConroy (approved):

PR #11116, opened 2019-06-11
Add option for only packing runtime-specific packages

natemcmaster:
A necessary code-cleanup to prepare for #7280


dougbu (approved):

natemcmaster:
:up: :date: it turns out there where several places that depended IsPackable=true, which probably shouldn't have. I've updated to make the following changes:

* Suppress producing Microsoft.AspNetCore.Blazor.Templates.dll. This was a dummy .dll file which was getting packed into the Blazor VSIX, even though it contained no code
* Decouple some build targets from IsPackable=true that shouldn't have depended on it
* Fixup ANCMSymbols

PR #11117, opened 2019-06-11
fix #11114 by setting WebHostService to 'Code' designer category

### LINKED ISSUE #11114: Mark WebHostService with DesignerCategory("Code")

cc @Pilchie @Tratcher

analogrelay:
Verified that this fixes the icon in VS and makes it load in the code editor instead of the component designer. I also verified that it works transitively for derived classes (which is how users actually hit this issue).

Addresses #11114 in 3.0.


anurse:
Actually we can do this in 2.1 and have the bot flow it forward.

PR #11118, opened 2019-06-11
Eagerly read IAsyncEnumerable{object} instances before formatting

### LINKED ISSUE #4833: Add IAsyncEnumerable support to MVC
**Edit: @rynowak hijacking top post for great justice**

This would include adding support for returning `IAsyncEnumerable<>` from an action method and letting MVC do the buffering for you because giving the data to the formatter.

This would require integrating with JsonResult and ObjectResult

pranavkm:
Fixes https://github.com/aspnet/AspNetCore/issues/4833

### DIFF
--- a/src/Mvc/Mvc.Core/src/Infrastructure/AsyncEnumerableReader.cs
+++ b/src/Mvc/Mvc.Core/src/Infrastructure/AsyncEnumerableReader.cs
@@ -1,0 +1,99 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Reflection;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc.Core;
+using Microsoft.Extensions.Internal;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.AspNetCore.Mvc.Infrastructure
+{
+    using ReaderFunc = Func<IAsyncEnumerable<object>, Task<ICollection>>;
+
+    /// <summary>
+    /// Type that reads an <see cref="IAsyncEnumerable{T}"/> instance into a
+    /// generic collection instance.
+    /// </summary>
+    /// <remarks>
+    /// This type is used to create a strongly typed synchronous <see cref="ICollection{T}"/> instance from
+    /// an <see cref="IAsyncEnumerable{T}"/>. An accurate <see cref="ICollection{T}"/> is required for XML formatters to 
+    /// correctly serialize.
+    /// </remarks>
+    public sealed class AsyncEnumerableReader
+    {
+        private readonly MethodInfo Converter = typeof(AsyncEnumerableReader).GetMethod(
+            nameof(ReadInternal),
+            BindingFlags.NonPublic | BindingFlags.Instance);
+
+        private readonly ConcurrentDictionary<Type, ReaderFunc> _asyncEnumerableConverters =
+            new ConcurrentDictionary<Type, ReaderFunc>();
+        private readonly MvcOptions _mvcOptions;
+
+        /// <summary>
+        /// Initializes a new instance of <see cref="AsyncEnumerableReader"/>.
+        /// </summary>
+        /// <param name="mvcOptions">Accessor to <see cref="MvcOptions"/>.</param>
+        public AsyncEnumerableReader(IOptions<MvcOptions> mvcOptions)
+        {
+            _mvcOptions = mvcOptions?.Value ?? throw new ArgumentNullException(nameof(mvcOptions));
+        }
+
+        /// <summary>
+        /// Reads a <see cref="IAsyncEnumerable{T}"/> into an <see cref="ICollection{T}"/>.
+        /// </summary>
+        /// <param name="value">The <see cref="IAsyncEnumerable{T}"/> to read.</param>
+        /// <returns>The <see cref="ICollection"/>.</returns>
+        public Task<ICollection> ReadAsync(IAsyncEnumerable<object> value)
+        {
+            if (value == null)
+            {
+                throw new ArgumentNullException(nameof(value));
+            }
+
+            var type = value.GetType();
+            if (!_asyncEnumerableConverters.TryGetValue(type, out var result))
+            {
+                var enumerableType = ClosedGenericMatcher.ExtractGenericInterface(type, typeof(IAsyncEnumerable<>));
+                Debug.Assert(enumerableType != null);
+
+                var enumeratedObjectType = enumerableType.GetGenericArguments()[0];
+
+                var converter = (ReaderFunc)Converter
+                    .MakeGenericMethod(enumeratedObjectType)
+                    .CreateDelegate(typeof(ReaderFunc), this);
+
+                _asyncEnumerableConverters.TryAdd(type, converter);
+                result = converter;
+            }
+
+            return result(value);
+        }
+
+        private async Task<ICollection> ReadInternal<T>(IAsyncEnumerable<object> value)
+        {
+            var asyncEnumerable = (IAsyncEnumerable<T>)value;
+            var result = new List<T>();
+            var count = 0;
+
+            await foreach (var item in asyncEnumerable)
+            {
+                if (count++ >= _mvcOptions.MaxIAsyncEnumerableBufferLimit)
+                {
+                    throw new InvalidOperationException(Resources.FormatObjectResultExecutor_MaxEnumerationExceeded(
+                        nameof(AsyncEnumerableReader),
+                        value.GetType()));
+                }
+
+                result.Add(item);
+            }
+
+            return result;
+        }
+    }
+}
--- a/src/Mvc/Mvc.Core/src/Infrastructure/AsyncEnumerableReader.cs
+++ b/src/Mvc/Mvc.Core/src/Infrastructure/AsyncEnumerableReader.cs
@@ -10,9 +10,12 @@ using System.Reflection;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Mvc.Core;
 using Microsoft.Extensions.Internal;
-using Microsoft.Extensions.Options;
 
+#if JSONNET
+namespace Microsoft.AspNetCore.Mvc.NewtonsoftJson
+#else
 namespace Microsoft.AspNetCore.Mvc.Infrastructure
+#endif
 {
     using ReaderFunc = Func<IAsyncEnumerable<object>, Task<ICollection>>;
 
@@ -22,10 +25,10 @@ namespace Microsoft.AspNetCore.Mvc.Infrastructure
     /// </summary>
     /// <remarks>
     /// This type is used to create a strongly typed synchronous <see cref="ICollection{T}"/> instance from
-    /// an <see cref="IAsyncEnumerable{T}"/>. An accurate <see cref="ICollection{T}"/> is required for XML formatters to 
+    /// an <see cref="IAsyncEnumerable{T}"/>. An accurate <see cref="ICollection{T}"/> is required for XML formatters to
     /// correctly serialize.
     /// </remarks>
-    public sealed class AsyncEnumerableReader
+    internal sealed class AsyncEnumerableReader
     {
         private readonly MethodInfo Converter = typeof(AsyncEnumerableReader).GetMethod(
             nameof(ReadInternal),
@@ -39,9 +42,9 @@ namespace Microsoft.AspNetCore.Mvc.Infrastructure
         /// Initializes a new instance of <see cref="AsyncEnumerableReader"/>.
         /// </summary>
         /// <param name="mvcOptions">Accessor to <see cref="MvcOptions"/>.</param>
-        public AsyncEnumerableReader(IOptions<MvcOptions> mvcOptions)
+        public AsyncEnumerableReader(MvcOptions mvcOptions)
         {
-            _mvcOptions = mvcOptions?.Value ?? throw new ArgumentNullException(nameof(mvcOptions));
+            _mvcOptions = mvcOptions;
         }
 
         /// <summary>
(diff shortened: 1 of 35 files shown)


src/Mvc/Mvc.Core/src/Infrastructure/ObjectResultExecutor.cs
```diff
@@ -18,16 +23,43 @@ namespace Microsoft.AspNetCore.Mvc.Infrastructure
     /// </summary>
     public class ObjectResultExecutor : IActionResultExecutor<ObjectResult>
     {
+        private delegate Task<object> ReadAsyncEnumerableDelegate(object value);
```
rynowak: Do we need this delegate type? Why not just `Func<object, Task<object>>`

src/Mvc/Mvc.Core/src/Infrastructure/ObjectResultExecutor.cs
```diff
@@ -138,5 +190,50 @@ private static void InferContentTypes(ActionContext context, ObjectResult result
...
         }
+
+        private Task<object> EnumerateAsyncEnumerable(IAsyncEnumerable<object> value)
+        {
+            var type = value.GetType();
+            if (!_asyncEnumerableConverters.TryGetValue(type, out var result))
+            {
+                var enumerableType = ClosedGenericMatcher.ExtractGenericInterface(type, typeof(IAsyncEnumerable<>));
+                result = null;
+                if (enumerableType != null)
```
rynowak: Should this ever fail? This seems like you should just let it blow up right?

src/Mvc/Mvc.Core/src/Infrastructure/ObjectResultExecutor.cs
pranavkm: It was ugly to repeat it. But I can do a `using`

src/Mvc/Mvc.Core/src/Infrastructure/ObjectResultExecutor.cs
```diff
@@ -138,5 +190,50 @@ private static void InferContentTypes(ActionContext context, ObjectResult result
                 result.ContentTypes.Add("application/problem+xml");
             }
         }
+
+        private Task<object> EnumerateAsyncEnumerable(IAsyncEnumerable<object> value)
+        {
+            var type = value.GetType();
+            if (!_asyncEnumerableConverters.TryGetValue(type, out var result))
+            {
+                var enumerableType = ClosedGenericMatcher.ExtractGenericInterface(type, typeof(IAsyncEnumerable<>));
```
rynowak: What happens if you implement multiple `IAsyncEnumerable<>` interfaces? implementation-defined? Crash with a bad error?

src/Mvc/Mvc.Core/src/Infrastructure/ObjectResultExecutor.cs
pranavkm: Yup. Left over from before. I'll Debug.Assert it.

src/Mvc/Mvc.Core/src/Infrastructure/ObjectResultExecutor.cs
```diff
@@ -138,5 +190,50 @@ private static void InferContentTypes(ActionContext context, ObjectResult result
...
+                result = null;
+                if (enumerableType != null)
+                {
+                    var enumeratedObjectType = enumerableType.GetGenericArguments()[0];
+
+                    var converter = (ReadAsyncEnumerableDelegate)Converter
+                        .MakeGenericMethod(enumeratedObjectType)
+                        .CreateDelegate(typeof(ReadAsyncEnumerableDelegate), this);
```
rynowak: cheeky.

src/Mvc/Mvc.Core/src/Infrastructure/ObjectResultExecutor.cs
```diff
@@ -138,5 +190,50 @@ private static void InferContentTypes(ActionContext context, ObjectResult result
...
+        private Task<object> EnumerateAsyncEnumerable(IAsyncEnumerable<object> value)
+        {
+            var type = value.GetType();
+            if (!_asyncEnumerableConverters.TryGetValue(type, out var result))
+            {
+                var enumerableType = ClosedGenericMatcher.ExtractGenericInterface(type, typeof(IAsyncEnumerable<>));
+                result = null;
+                if (enumerableType != null)
+                {
+                    var enumeratedObjectType = enumerableType.GetGenericArguments()[0];
```
pranavkm: Btw, we need to do this because XML serializer needs to be told the type of the list. Giving it a vanilla `List<object>` causes them to not be serialized correctly.

pranavkm: I should probably add a note

src/Mvc/Mvc.Core/src/Resources.resx
```diff
@@ -503,5 +503,8 @@
   </data>
   <data name="Property_MustBeInstanceOfType" xml:space="preserve">
     <value>Property '{0}.{1}' must be an instance of type '{2}'.</value>
-  </data>  
-</root>
+  </data>
+  <data name="ObjectResultExecutor_MaxEnumerationExceeded" xml:space="preserve">
+    <value>{0} execeeded the maximum configured enumeration limit '{1}' when enumerating async enumerable type '{2}'.</value>
```
rynowak: Should we add more details here?

This doesn't address how a user would solve a problem if they have one.

I think it would be nice to say something like:

```
The '{formatter}' reached the configured maximum size of the buffer when an enumerating a value of type `FooType'. This limit is in place to prevent infinite streams of `IAsyncEnumerable<>` from continuing indefinitely. If this is not a programming mistake, consider ways to reduce the collection or size, or consider manually converting 'FooType' into a list rather than increasing the limit. 
```

src/Mvc/Mvc.Core/src/Infrastructure/ObjectResultExecutor.cs
```diff
@@ -2,14 +2,19 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Concurrent;
```
rynowak: JsonResult? You might want to convert all of this new logic to a shared file.

src/Mvc/Mvc.Core/src/Infrastructure/ObjectResultExecutor.cs
rynowak: There's no difference between this and a using. My point is that I find `Func<>` a lot more readable than custom delegate types. If you don't agree then feel free to ignore.

src/Mvc/Mvc.Core/src/Infrastructure/ObjectResultExecutor.cs
pranavkm: yup yup. Totally forgot about Json

src/Mvc/Mvc.Core/src/Infrastructure/ObjectResultExecutor.cs
mburbea: I'd argue that the using doesn't create an extra type, so it's slightly better.

pranavkm:
🆙 📅

src/Mvc/Mvc.Core/src/Infrastructure/AsyncEnumerableReader.cs
```diff
@@ -0,0 +1,99 @@
...
+    /// <summary>
+    /// Type that reads an <see cref="IAsyncEnumerable{T}"/> instance into a
+    /// generic collection instance.
+    /// </summary>
+    /// <remarks>
+    /// This type is used to create a strongly typed synchronous <see cref="ICollection{T}"/> instance from
+    /// an <see cref="IAsyncEnumerable{T}"/>. An accurate <see cref="ICollection{T}"/> is required for XML formatters to 
+    /// correctly serialize.
+    /// </remarks>
+    public sealed class AsyncEnumerableReader
```
rynowak: Does this really need to be public? I don't have a strong objection, but it seems like infrastructure, rather than something we expect users to need.

src/Mvc/Mvc.Core/src/Infrastructure/AsyncEnumerableReader.cs
```diff
@@ -0,0 +1,99 @@
...
+            if (value == null)
+            {
+                throw new ArgumentNullException(nameof(value));
+            }
+
+            var type = value.GetType();
+            if (!_asyncEnumerableConverters.TryGetValue(type, out var result))
+            {
+                var enumerableType = ClosedGenericMatcher.ExtractGenericInterface(type, typeof(IAsyncEnumerable<>));
+                Debug.Assert(enumerableType != null);
```
rynowak: This is a public API, so this should probably throw instead of assert. Someone could pass in any type here.

src/Mvc/Mvc.Core/src/Infrastructure/ObjectResultExecutor.cs
```diff
@@ -138,5 +177,7 @@ private static void InferContentTypes(ActionContext context, ObjectResult result
                 result.ContentTypes.Add("application/problem+xml");
             }
         }
+
+
```
rynowak: @JamesNK

src/Mvc/Mvc.Core/src/Infrastructure/ObjectResultExecutor.cs
```diff
@@ -87,16 +107,35 @@ public virtual Task ExecuteAsync(ActionContext context, ObjectResult result)
...
+                return ExecuteAsyncEnumerable(context, result, asyncEnumerable);
+            }
+
+            return ExecuteAsyncCore(context, result, objectType, value);
+        }
+
+        private async Task ExecuteAsyncEnumerable(ActionContext context, ObjectResult result, IAsyncEnumerable<object> asyncEnumerable)
+        {
+            var enumerated = await _asyncEnumerableReader.ReadAsync(asyncEnumerable);
+            await ExecuteAsyncCore(context, result, enumerated.GetType(), enumerated);
```
rynowak: Should we be logging here?

src/Mvc/Mvc.Core/src/Infrastructure/SystemTextJsonResultExecutor.cs
```diff
@@ -123,11 +134,22 @@ private static class Log
                 new EventId(1, "JsonResultExecuting"),
                 "Executing JsonResult, writing value of type '{Type}'.");
 
+            private static readonly Action<ILogger, string, Exception> _eagerlyReadingAsyncEnumerable = LoggerMessage.Define<string>(
+               LogLevel.Debug,
+               new EventId(2, "EagerReadAsyncEnumerable"),
+               "Eagerly reading IAsyncEnumerable instance of type '{Type}'.");
```
rynowak: Can we say something like `Eagerly buffering` instead of reading? I think it's a little confusing to say that we're reading when we're about to write. I realize that it's technically correct, but still a little misleading.

src/Mvc/Mvc.Core/src/Infrastructure/SystemTextJsonResultExecutor.cs
```diff
@@ -123,11 +134,22 @@ private static class Log
...
+
             public static void JsonResultExecuting(ILogger logger, object value)
             {
                 var type = value == null ? "null" : value.GetType().FullName;
                 _jsonResultExecuting(logger, type, null);
             }
+
+            public static void EagerlyReadingAsyncEnumerable(ILogger logger, object value)
+            {
+                var type = value == null ? "null" : value.GetType().FullName;
```
rynowak: The value can't really be null here right? This would fail the `is`-cast.

src/Mvc/Mvc.Core/src/Resources.resx
```diff
@@ -503,5 +503,8 @@
   </data>
   <data name="Property_MustBeInstanceOfType" xml:space="preserve">
     <value>Property '{0}.{1}' must be an instance of type '{2}'.</value>
-  </data>  
-</root>
+  </data>
+  <data name="ObjectResultExecutor_MaxEnumerationExceeded" xml:space="preserve">
+    <value>'{0}' reached the configured maximum size of the buffer when an enumerating a value of type `{1}'. This limit is in place to prevent infinite streams of `IAsyncEnumerable&lt;&gt;` from continuing indefinitely. If this is not a programming mistake, consider ways to reduce the collection or size, or consider manually converting '{1}' into a list rather than increasing the limit.</value>
```
rynowak: `collection or size` => `collection size`

src/Mvc/Mvc.NewtonsoftJson/src/NewtonsoftJsonResultExecutor.cs
```diff
@@ -131,7 +134,13 @@ await using (var writer = _writerFactory.CreateWriter(responseStream, resolvedCo
                     jsonWriter.AutoCompleteOnClose = false;
 
                     var jsonSerializer = JsonSerializer.Create(jsonSerializerSettings);
-                    jsonSerializer.Serialize(jsonWriter, result.Value);
+                    var value = result.Value;
+                    if (result.Value is IAsyncEnumerable<object> asyncEnumerable)
+                    {
+                        value = await _asyncEnumerableReader.ReadAsync(asyncEnumerable);
```
rynowak: Should we log here?

rynowak (approved):

src/Mvc/Mvc.Core/src/Infrastructure/AsyncEnumerableReader.cs
pranavkm: I could defer reading this from DI or have duplicated caching if your app uses a mix of `ObjectResult` and `JsonResult` returning `IAsyncEnumerable`. I guess the latter is rare and easily fixable if it turns out to be a problem in the future.

rynowak: I don't understand your answer. What's the reason why this type is public?

pranavkm: `ObjectResultExecutor` is public. If we have to share the same singleton instance between the JsonResultExecutor(s) and ObjectResultExecutor, we either have to make the type public for ctor DI or read it in the body of `ExecuteAsync`.

Issue #11061, opened 2019-06-10
Fail on startup for Http/2 HTTPS on Mac


Tratcher:
Mac doesn't support ALPN: https://github.com/dotnet/corefx/issues/33016

In 3.0 HTTPS and HTTP/2 (and HTTP/1) are enabled by default. The GRPC template is going to disable HTTP/1. HTTP/2 per spec requires ALPN when used over HTTPS. Today this fails per request and you have to dig into the server logs for the error.

Proposal: Check for HTTPS + HTTP/2-only on Mac at startup and fail early.

@shirhatti @JunTaoLuo

JunTaoLuo:
FYI @JamesNK

Tratcher:
@JunTaoLuo for verification.

JunTaoLuo:
Partly verified in https://github.com/aspnet/AspNetCore/pull/11473#issuecomment-505287505. I'll try it out on a preview 7 SDK.

PR #11120, opened 2019-06-11
Fail on startup for Http/2 HTTPS on Mac

Tratcher:
#11061 Testing via Helix


Tratcher:
It worked, and failed a bunch of other tests like Templates.Test.GrpcTemplateTest.GrpcTemplate 😁.

src/ProjectTemplates/test/GrpcTemplateTest.cs
```diff
@@ -35,6 +36,13 @@ public async Task GrpcTemplate()
             var buildResult = await Project.RunDotNetBuildAsync();
             Assert.True(0 == buildResult.ExitCode, ErrorMessages.GetFailedProcessMessage("build", Project, buildResult));
 
+            // These templates are HTTPS + HTTP/2 only which is not supported on Mac due to missing ALPN support.
```
ryanbrandenburg: Shouldn't this be:
```c#
[ConditionalFact]
[OSSkipCondition(OperatingSystems.MacOSX)]
```
Instead? This pattern gives the impression that the test ran on Mac when it didn't do anything interesting. Either that or we should verify the behavior you get on a mac (I assume it gives a semi-useful error message?).

src/Servers/Kestrel/Kestrel/test/KestrelConfigurationBuilderTests.cs
```diff
@@ -371,11 +384,23 @@ public void DefaultConfigSectionCanSetProtocols(string input, HttpProtocols expe
             Assert.True(ran3);
         }
 
+        [Theory]
+        [InlineData("http1", HttpProtocols.Http1)]
+        // [InlineData("http2", HttpProtocols.Http2)] // Not supported due to missing ALPN support. https://github.com/dotnet/corefx/issues/33016
```
ryanbrandenburg: Too bad you can't mark InlineData as skipped.

ryanbrandenburg (approved):
This is good to go once the ProjectTemplate test either uses OSSkipCondition or checks the behavior that was changed here.

src/ProjectTemplates/test/GrpcTemplateTest.cs
Tratcher: Added a check for the error message.

PR #11123, opened 2019-06-12
fix #11114 by setting DesignerCategory for WebHostService

### LINKED ISSUE #11114: Mark WebHostService with DesignerCategory("Code")

analogrelay:
This time for reals. We're doing the 2.1/2.2 fixes via fixing the doc (https://github.com/aspnet/AspNetCore.Docs/pull/12823).

Fixes #11114

### DIFF
--- a/src/Hosting/WindowsServices/ref/Microsoft.AspNetCore.Hosting.WindowsServices.netcoreapp3.0.cs
+++ b/src/Hosting/WindowsServices/ref/Microsoft.AspNetCore.Hosting.WindowsServices.netcoreapp3.0.cs
@@ -3,6 +3,7 @@
 
 namespace Microsoft.AspNetCore.Hosting.WindowsServices
 {
+    [System.ComponentModel.DesignerCategoryAttribute("Code")]
     public partial class WebHostService : System.ServiceProcess.ServiceBase
     {
         public WebHostService(Microsoft.AspNetCore.Hosting.IWebHost host) { }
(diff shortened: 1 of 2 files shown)


Pilchie (approved):

anurse:
From @Pilchie in [a different thread](https://github.com/aspnet/AspNetCore.Docs/pull/12823):

> Just a note that the reason this is surfacing now is that the .NET Core Project system didn't used to obey the [DesignerCategory] attribute, but it now needs to in order for Winforms to work properly in 3.0, so that was done in VS 2019.

Yes, I realize this type is no longer the intended way to do this in 3.0 (use Worker Service instead). The [doc does reflect that in the 3.0 pivot](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/windows-service?view=aspnetcore-3.0&tabs=visual-studio). Still, this type is here and we don't want to break people if they use it.

Tratcher:
```
src\Hosting\WindowsServices\ref\Microsoft.AspNetCore.Hosting.WindowsServices.netcoreapp3.0.cs(,): error : Generated code is not up to date in src/Hosting/WindowsServices/ref/Microsoft.AspNetCore.Hosting.WindowsServices.netcoreapp3.0.cs. You might need to regenerate the reference assemblies or project list (see docs/ReferenceAssemblies.md and docs/ReferenceResolution.md)
```

anurse:
Boo! Yeah, ref-asms do something with attributes so I guess that makes sense.

anurse:
@aspnet/build it seems unlikely that my change cause the build to start hanging (especially only the Windows x86 build)... is there a known issue with hanging builds right now?

dougbu:
> is there a known issue with hanging builds right now?

Yes, we've seen this a lot and it may be limited to this repo. @JunTaoLuo is looking into the issue.

anurse:
Merging as Helix shouldn't be affected by this.

PR #11124, opened 2019-06-12
Fix Platform benchmark

### LINKED ISSUE #11100: PlatformBenchmark broken
Due to this change:
https://github.com/aspnet/AspNetCore/commit/f10680a37ab34c2b92829bc5135dde253260614a#diff-d6a9ae754775b6cb72445c3b8db610c2

At this line:
https://github.com/aspnet/AspNetCore/blob/master/src/Servers/Kestrel/perf/PlatformBenchmarks/BenchmarkApplication.HttpConnection.cs#L102

@benaadams @davidfowl

benaadams:
Resolves #11100

/cc @sebastienros I tested this time :)


benaadams:
Updated to a more simplified change

sebastienros:
I tested it and it works, thanks

sebastienros (approved):

PR #11130, opened 2019-06-12
Convert SignalR EventSource usage to new pattern

BrennanConroy:
And sneak in a new counter: CurrentConnections

### DIFF
--- a/src/SignalR/common/Http.Connections/src/Internal/HttpConnectionsEventSource.cs
+++ b/src/SignalR/common/Http.Connections/src/Internal/HttpConnectionsEventSource.cs
@@ -105,7 +105,10 @@ namespace Microsoft.AspNetCore.Http.Connections.Internal
                     DisplayName = "Current Connections",
                 };
 
-                _connectionDuration ??= new EventCounter("connections-duration", this);
+                _connectionDuration ??= new EventCounter("connections-duration", this)
+                {
+                    DisplayName = "Average Connection Duration",
+                };
             }
         }
 
--- a/src/SignalR/common/Http.Connections/src/Internal/HttpConnectionsEventSource.cs
+++ b/src/SignalR/common/Http.Connections/src/Internal/HttpConnectionsEventSource.cs
@@ -38,12 +38,13 @@ namespace Microsoft.AspNetCore.Http.Connections.Internal
         [NonEvent]
         public void ConnectionStop(string connectionId, ValueStopwatch timer)
         {
+            Interlocked.Increment(ref _connectionsStopped);
+            Interlocked.Decrement(ref _currentConnections);
+
             if (IsEnabled())
             {
                 var duration = timer.IsActive ? timer.GetElapsedTime().TotalMilliseconds : 0.0;
-                _connectionDuration.WriteMetric((float)duration);
-                Interlocked.Increment(ref _connectionsStopped);
-                Interlocked.Decrement(ref _currentConnections);
+                _connectionDuration.WriteMetric(duration);
 
                 if (IsEnabled(EventLevel.Informational, EventKeywords.None))
                 {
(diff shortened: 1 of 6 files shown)


davidfowl:
Nice! The pattern where we check IsEnabled() is broken because you can end up with bad counts if the counters are enabled midway.

BrennanConroy:
So should we move all `Interlocked` to outside the `IsEnabled` check?

halter73 (approved):

BrennanConroy:
@aspnet/build How does one reset the AspNetCore-ci check without closing and re-opening the PR?

dougbu:
@BrennanConroy what do you mean by "reset" in this context?

BrennanConroy:
There is a red X still even though I have restarted the job.

Looks like a similar issue to another PR where the check isn't being updated on a successful build: https://github.com/aspnet/AspNetCore/pull/10540

davidfowl:
Before you merge, a screenshot with the counters please 😄

src/SignalR/common/Http.Connections/src/Internal/HttpConnectionsEventSource.cs
```diff
@@ -1,40 +1,50 @@
...
+        private EventCounter _connectionDuration;
 
-        private HttpConnectionsEventSource()
+        private long _connectionsStarted;
+        private long _connectionsStopped;
+        private long _connectionsTimedOut;
+        private long _currentConnections;
+
+        internal HttpConnectionsEventSource()
+            : base("Microsoft-AspNetCore-Http-Connections")
```
davidfowl: Rename to `Microsoft.AspNetCore.Http.Connections`?

BrennanConroy: Should we also normalize the counter names? The hosting ones follow the pattern of "current-requests", and we currently use "CurrentConnections".

davidfowl: Yes, lets unify the patterns.

anurse: EventSources have always been traditionally `Microsoft-Things-Separated-By-Dashes`, why would we change that?

Also, we should update the hosting ones since these have already shipped with these names. Plus `snake-case` is so un-.NET ;)

davidfowl: https://github.com/dotnet/coreclr/blob/e2b3130f4d0c9f1b759a769f4d344ee72a55b771/src/System.Private.CoreLib/src/System/Diagnostics/Eventing/RuntimeEventSource.cs#L12

Inconsistent AF. This is something we should resolve as part of preview7 (Hosting is updated BTW).

cc @noahfalk

anurse: Well, it's unfortunate that CoreFX didn't follow our _completely_ undocumented design pattern from 2.2 here 😝. Hard to fix now though. I'm OK with establishing `snake-case` as the convention if that's the case.

If we do that we have two options:
* Leave the SignalR counters alone. Nobody is "broken" but they look different (though the Display Name is what is generally shown I believe)
* Change the existing SignalR counters to match the new pattern. Anybody depending on the exact name is broken. There weren't great tools for collecting these before 3.0 though so I don't feel too bad breaking that (as long as we Announcements it)

davidfowl: There's only one option, use the new pattern.

BrennanConroy:
image

davidfowl:
Whats up with the duration counter? Why does it look like that?

BrennanConroy:
>Whats up with the duration counter? Why does it look like that?

I was going to ask if you knew how to set the display name of a "normal" event counter.

BrennanConroy:
Figured it out, didn't realize `EventCounter` inherited from `DiagnosticCounter`

image

src/SignalR/common/Http.Connections/src/Internal/HttpConnectionsEventSource.cs
```diff
@@ -65,14 +73,42 @@ public ValueStopwatch ConnectionStart(string connectionId)
...
+                    DisplayName = "Total Connections Timed Out",
+                };
+                _currentConnectionsCounter ??= new PollingCounter("current-connections", this, () => _currentConnections)
+                {
+                    DisplayName = "Current Connections",
+                };
+
+                _connectionDuration ??= new EventCounter("connections-duration", this)
+                {
+                    DisplayName = "Average Connection Duration",
```
halter73: If we're renaming anyway, this should probably be called "connections-duration-milliseconds".

If we decide that's to verbose, we should still document the units somewhere. The DisplayName should include "milliseconds" or "ms" somewhere.

davidfowl (approved):

src/SignalR/common/Http.Connections/src/Internal/HttpConnectionsEventSource.cs
davidfowl: There’s a new display units field coming

BrennanConroy:
@aspnet/build Please merge :)

PR #11132, opened 2019-06-12
Remove dependency on pipe events in HttpConnection

davidfowl:
- Refactored the HttpConnection to not depend on OnReaderCompleted and OnWriterCompleted. Instead we use ConnectionClosed to detect the FIN that we propagate via ConnectedAborted.
- Fire ConnectionClosed when a FIN is received from both transports.
- Remove pipe completion from Http1Connection and Http1OutputProducer. Instead just return from request processing.
- Flow the CancellaitonToken property in RawStream

### DIFF
--- a/src/Servers/Kestrel/Core/src/Adapter/Internal/AdaptedPipeline.cs
+++ b/src/Servers/Kestrel/Core/src/Adapter/Internal/AdaptedPipeline.cs
@@ -15,7 +15,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal
     {
         private readonly int _minAllocBufferSize;
 
-        private readonly IDuplexPipe _transport;
+        private Task _inputTask;
+        private Task _outputTask;
 
         public AdaptedPipeline(IDuplexPipe transport,
                                Pipe inputPipe,
@@ -23,12 +24,14 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal
                                IKestrelTrace log,
                                int minAllocBufferSize)
         {
-            _transport = transport;
+            TransportStream = new RawStream(transport.Input, transport.Output, throwOnCancelled: true);
             Input = inputPipe;
             Output = outputPipe;
             Log = log;
             _minAllocBufferSize = minAllocBufferSize;
         }
+
+        public RawStream TransportStream { get; }
 
         public Pipe Input { get; }
 
@@ -40,21 +43,31 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal
 
         PipeWriter IDuplexPipe.Output => Output.Writer;
 
-        public async Task RunAsync(Stream stream)
+        public void RunAsync(Stream stream)
         {
-            var inputTask = ReadInputAsync(stream);
-            var outputTask = WriteOutputAsync(stream);
-
-            await outputTask;
-            // Yield the current request
-            _transport.Input.CancelPendingRead();
-            await inputTask;
+            _inputTask = ReadInputAsync(stream);
+            _outputTask = WriteOutputAsync(stream);
         }
 
-        public void Complete()
+        public async Task CompleteAsync()
         {
             Output.Writer.Complete();
             Input.Reader.Complete();
+
+            if (_outputTask == null)
+            {
+                return;
+            }
+
+            // Wait for the output task to complete, this ensures that we've copied
+            // the application data to the underlying stream
+            await _outputTask;
+
+            // Cancel the underlying stream so that the input task yields
+            TransportStream.CancelPendingRead();
+
+            // The input task should yield now that we've cancelled it
+            await _inputTask;
         }
 
         private async Task WriteOutputAsync(Stream stream)
--- a/src/Servers/Kestrel/Core/src/Adapter/Internal/AdaptedPipeline.cs
+++ b/src/Servers/Kestrel/Core/src/Adapter/Internal/AdaptedPipeline.cs
@@ -4,7 +4,9 @@
 using System;
 using System.IO;
 using System.IO.Pipelines;
+using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Connections;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure;
 using Microsoft.Extensions.Logging;
 
@@ -15,6 +17,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal
         private readonly int _minAllocBufferSize;
 
         private readonly IDuplexPipe _transport;
+        private readonly CancellationTokenSource _readingTokenSource = new CancellationTokenSource();
 
         public AdaptedPipeline(IDuplexPipe transport,
                                Pipe inputPipe,
@@ -44,8 +47,15 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal
             var inputTask = ReadInputAsync(stream);
             var outputTask = WriteOutputAsync(stream);
 
-            await inputTask;
             await outputTask;
+            _readingTokenSource.Cancel();
+            await inputTask;
+        }
+
+        public void Complete()
+        {
+            Output.Writer.Complete();
+            Input.Reader.Complete();
         }
 
         private async Task WriteOutputAsync(Stream stream)
@@ -97,7 +107,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal
             finally
             {
                 Output.Reader.Complete();
-                _transport.Output.Complete();
             }
         }
 
@@ -115,9 +124,8 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal
 
                 while (true)
                 {
-
                     var outputBuffer = Input.Writer.GetMemory(_minAllocBufferSize);
-                    var bytesRead = await stream.ReadAsync(outputBuffer);
+                    var bytesRead = await stream.ReadAsync(outputBuffer, _readingTokenSource.Token);
                     Input.Writer.Advance(bytesRead);
 
                     if (bytesRead == 0)
@@ -134,6 +142,11 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal
                     }
                 }
             }
+            catch (OperationCanceledException ex)
+            {
+                // Propagate the exception if it's ConnectionAbortedException
+                error = ex is ConnectionAbortedException ? ex : null;
+            }
             catch (Exception ex)
             {
                 // Don't rethrow the exception. It should be handled by the Pipeline consumer.
@@ -142,9 +155,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Adapter.Internal
             finally
             {
                 Input.Writer.Complete(error);
-                // The application could have ended the input pipe so complete
-                // the transport pipe as well
-                _transport.Input.Complete();
             }
         }
     }
--- a/src/Servers/Kestrel/Core/src/Adapter/Internal/AdaptedPipeline.cs
+++ b/src/Servers/Kestrel/Core/src/Adapter/Internal/AdaptedPipeline.cs
@@ -4,7 +4,6 @@
 using System;
 using System.IO;
 using System.IO.Pipelines;
-using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Connections;
 using Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Infrastructure;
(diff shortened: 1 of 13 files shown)


src/Servers/Kestrel/Core/src/Adapter/Internal/AdaptedPipeline.cs
```diff
@@ -134,6 +142,11 @@ private async Task ReadInputAsync(Stream stream)
                     }
                 }
             }
+            catch (OperationCanceledException ex)
+            {
+                // Propagate the exception if it's ConnectionAbortedException
+                error = ex is ConnectionAbortedException ? ex : null;
```
gfoidl:
```suggestion
                error = ex as ConnectionAbortedException;
```

jkotalik:
I'd also prefer to have this for https://github.com/aspnet/AspNetCore/pull/11109

src/Servers/Kestrel/Core/src/Internal/HttpConnection.cs
```diff
@@ -160,19 +160,20 @@ public async Task ProcessRequestsAsync<TContext>(IHttpApplication<TContext> http
                         }
                     }
 
-                    _context.Transport.Input.OnWriterCompleted(
```
jkotalik: 👏

jkotalik (approved):
The change itself seems good, but I think we should go over this in person before merging. Also to figure out how we can sync this with https://github.com/aspnet/AspNetCore/pull/11109/checks?check_run_id=146487478. I'd rather merge this first before merging my PR as I don't want to have the pipe scheduler be thread pool.

src/Servers/Kestrel/Core/src/Adapter/Internal/AdaptedPipeline.cs
```diff
@@ -115,9 +124,8 @@ private async Task ReadInputAsync(Stream stream)
 
                 while (true)
                 {
-
                     var outputBuffer = Input.Writer.GetMemory(_minAllocBufferSize);
-                    var bytesRead = await stream.ReadAsync(outputBuffer);
+                    var bytesRead = await stream.ReadAsync(outputBuffer, _readingTokenSource.Token);
```
halter73: But the perf! JK. looking forward to the update.

src/Servers/Kestrel/Core/src/Adapter/Internal/AdaptedPipeline.cs
```diff
@@ -44,8 +45,16 @@ public async Task RunAsync(Stream stream)
             var inputTask = ReadInputAsync(stream);
             var outputTask = WriteOutputAsync(stream);
 
-            await inputTask;
             await outputTask;
+            // Yield the current request
+            _transport.Input.CancelPendingRead();
```
halter73: Shouldn't we wait for the connection middleware pipeline before closing the input? I think this breaks half-duplex if that matters.

src/Servers/Kestrel/Core/src/Adapter/Internal/RawStream.cs
```diff
@@ -105,14 +107,19 @@ public override Task FlushAsync(CancellationToken cancellationToken)
...
             while (true)
             {
-                var result = await _input.ReadAsync();
+                var result = await _input.ReadAsync(cancellationToken);
                 var readableBuffer = result.Buffer;
                 try
                 {
+                    if (_throwOnCancelled && result.IsCanceled)
+                    {
+                        throw new OperationCanceledException();
```
halter73: We need a flag to know *we're* the once that canceled the read.

src/Servers/Kestrel/Core/src/Adapter/Internal/AdaptedPipeline.cs
```diff
@@ -134,6 +141,11 @@ private async Task ReadInputAsync(Stream stream)
                     }
                 }
             }
+            catch (OperationCanceledException ex)
+            {
+                // Propagate the exception if it's ConnectionAbortedException
+                error = ex as ConnectionAbortedException;
```
halter73: Why not make this an empty `catch (ConnectionAbortedException) { }` block?

halter73: I misread. You are only storing this if it *is* a ConnectionAbortedException. In that case having a non-empty ConnectionAbortedException catch then an empty OperationCanceledException catch after is still probably better.

davidfowl: This is less code.

src/Servers/Kestrel/Transport.Sockets/src/Internal/SocketConnection.cs
```diff
@@ -187,6 +173,10 @@ private async Task DoReceive()
             {
                 // If Shutdown() has already bee called, assume that was the reason ProcessReceives() exited.
                 Input.Complete(_shutdownReason ?? error);
+
+                FireConnectionClosed();
```
halter73: This makes me wish we called to ConnectionClosed token FinReceived or something like that, because the connection might be only half closed at this point.

src/Servers/Kestrel/Transport.Sockets/src/Internal/SocketConnection.cs
```diff
@@ -307,6 +297,26 @@ private async Task ProcessSends()
             }
         }
 
+        private void FireConnectionClosed()
+        {
+            // Guard against scheduling this multiple times
+            if (_connectionClosed)
```
halter73: I see why you added this bool for the libuv transport, but it seems completely unnecessary in the socket transport. FireConnectionClosed can only ever be called once.

src/Servers/Kestrel/Transport.Libuv/src/Internal/LibuvConnection.cs
```diff
@@ -246,6 +239,26 @@ private void OnRead(UvStreamHandle handle, int status)
             }
         }
 
+        private void FireConnectionClosed()
+        {
+            // Guard against scheduling this multiple times
+            if (_connectionClosed)
```
halter73: I see you're taking advantage of the single-threaded nature of libuv's event loop 👍

src/Servers/Kestrel/Transport.Libuv/src/Internal/LibuvConnection.cs
```diff
@@ -135,21 +137,10 @@ private async Task StartCore()
                     // We're done with the socket now
                     _socket.Dispose();
 
-                    // Fire the connection closed token and wait for it to complete
-                    var waitForConnectionClosedTcs = new TaskCompletionSource<object>(TaskCreationOptions.RunContinuationsAsynchronously);
+                    // Ensure this always fires
+                    FireConnectionClosed();
```
halter73: If we're going to continue firing ConnectionClosed for libuv output errors, we should do the same for the socket transport.

src/Servers/Kestrel/test/InMemory.FunctionalTests/RequestTests.cs
```diff
@@ -856,6 +852,7 @@ await using (var server = new TestServer(TestApp.EchoAppChunked, testContext))
                     await connection.Send(
                         "POST / HTTP/1.1");
                     connection.ShutdownSend();
+                    await connection.TransportConnection.WaitForCloseTask;
```
halter73: Why is this necessary?

src/Servers/Kestrel/Core/src/Adapter/Internal/RawStream.cs
```diff
@@ -61,17 +63,17 @@ public override int Read(byte[] buffer, int offset, int count)
...
+        public override Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken = default)
         {
-            return ReadAsyncInternal(new Memory<byte>(buffer, offset, count)).AsTask();
+            return ReadAsyncInternal(new Memory<byte>(buffer, offset, count), cancellationToken).AsTask();
         }
 
         public override ValueTask<int> ReadAsync(Memory<byte> destination, CancellationToken cancellationToken = default)
         {
-            return ReadAsyncInternal(destination);
+            return ReadAsyncInternal(destination, cancellationToken);
```
halter73: Now that we're flowing the cancellationToken through RawStream, do we have a regression test?

src/Servers/Kestrel/Transport.Sockets/src/Internal/SocketConnection.cs
davidfowl: Yes that is true. I noticed that when I was copying code.

src/Servers/Kestrel/Transport.Libuv/src/Internal/LibuvConnection.cs
davidfowl: What do you mean? Aren't we doing that? It always fires no?

src/Servers/Kestrel/test/InMemory.FunctionalTests/RequestTests.cs
halter73: Discussed this offline. Calling ReceiveEnd unblocks the ProcessRequest loop allowing to observe the FIN and write a 4XX before the transport is aborted.

src/Servers/Kestrel/Transport.Libuv/src/Internal/LibuvConnection.cs
halter73: I'm now thinking about making this more like the socket transport instead. The important think is consistency.

Instead of calling FireConnectionClosed here, just don't dispose _socket here. Instead, close the uv_stream_handle only after both the output loop completed and

a) you've received a terminal read callback or
b) the connection was aborted

src/Servers/Kestrel/Transport.Libuv/src/Internal/LibuvConnection.cs
```diff
@@ -234,6 +225,8 @@ private void OnRead(UvStreamHandle handle, int status)
                 if (status == LibuvConstants.EOF)
                 {
                     Log.ConnectionReadFin(ConnectionId);
+
+                    FireConnectionClosed();
```
halter73: Why not fire this for any terminal status?

halter73: Really we should be calling this every time we call Input.Complete().

src/Servers/Kestrel/Core/src/Adapter/Internal/RawStream.cs
davidfowl: Done

src/Servers/Kestrel/Transport.Sockets/src/Internal/SocketConnection.cs
davidfowl: You could argue the same about HttpContext.ConnectionAborted, I don't think it makes a difference.

davidfowl:
@aspnet/build What's the deal with all the PRs, do we merge, do we wait? I know it's being looked at but what is the mitigation to unblock work?

src/Servers/Kestrel/Core/src/Adapter/Internal/RawStream.cs
```diff
@@ -14,11 +14,20 @@ internal sealed class RawStream : Stream
...
+        public RawStream(PipeReader input, PipeWriter output, bool throwOnCancelled = false)
         {
             _input = input;
             _output = output;
+            _throwOnCancelled = throwOnCancelled;
+        }
+
+        public void CancelPendingRead()
+        {
+            _cancelCalled = true;
```
jkotalik: Does CancelPendingRead need to be synchronized?

davidfowl: Not for our usage.

dougbu:
@davidfowl as a few people have done, I suggest making a polite request for an admin to merge PRs where build success isn't being reported correctly or individual jobs hang. Looks like both of those issues have hit this PR but is it ready to merge?

davidfowl:
> @davidfowl as a few people have done, I suggest making a polite request for an admin to merge PRs where build success isn't being reported correctly or individual jobs hang. Looks like both of those issues have hit this PR but is it ready to merge?

I can merge I just want to know the protocol. Yes this is ready to merge.

dougbu:
:ok:, ping if you're blocked

src/Servers/Kestrel/Transport.Sockets/src/Internal/SocketConnection.cs
halter73: You absolutely could, but at least that's HTTP-specific so it makes a bit more sense.

src/Servers/Kestrel/Core/src/Adapter/Internal/RawStream.cs
halter73: I don't even think _cancelCalled *needs* to be volatile since CancelPendingRead acquires and releases the Pipe's `_sync` lock right afterwards.

PR #11133, opened 2019-06-12
HttpMethodAttribute: fix documentation and parameter check

Thiez:
The second constructor of the `HttpMethodAttribute` class claims that the `template` parameter may not be `null`, and no such claim is present about the `httpMethods` parameter. However, when `httpMethods` is `null`, an exception is thrown, while `template` *is* allowed to be null (as can be seen in the `HttpMethodAttribute(IEnumerable<string>)` constructor, which passes `null`.

In addition, the `HttpMethodAttribute(IEnumerable<string>)` constructor contains a `null`-check for the `httpMethods` parameter, but since this check is *also* in the `HttpMethodAttribute(IEnumerable<string>, string)` constructor this check is dead code.


pranavkm (approved):

pranavkm:
Thanks for the PR.

Thiez:
My pleasure :-)

PR #11150, opened 2019-06-12
Set rollForward to LatestPatch

JunTaoLuo:
Addresses https://github.com/aspnet/AspNetCore/issues/10567

### DIFF
--- a/src/Framework/test/SharedFxTests.cs
+++ b/src/Framework/test/SharedFxTests.cs
@@ -65,6 +65,7 @@ namespace Microsoft.AspNetCore
 
             Assert.Equal("Microsoft.NETCore.App", (string)runtimeConfig["runtimeOptions"]["framework"]["name"]);
             Assert.Equal(_expectedTfm, (string)runtimeConfig["runtimeOptions"]["tfm"]);
+            Assert.Equal("LatestPatch", (string)runtimeConfig["rollForward"]);
 
             Assert.Equal(TestData.GetMicrosoftNETCoreAppPackageVersion(), (string)runtimeConfig["runtimeOptions"]["framework"]["version"]);
         }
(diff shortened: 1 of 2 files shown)


natemcmaster:
Can you add an assertion to the unit test, too?

https://github.com/aspnet/AspNetCore/blob/402394347d5a9b3add7decf52af85cbb4e7866b3/src/Framework/test/SharedFxTests.cs#L56-L57

src/Framework/test/SharedFxTests.cs
```diff
@@ -137,5 +138,57 @@ public void ItContainsVersionFile()
             Assert.Equal(TestData.GetRepositoryCommit(), lines[0]);
             Assert.Equal(TestData.GetSharedFxVersion(), lines[1]);
         }
+
+        [Fact]
+        public void PlatformManifestListsAllFiles()
```
JunTaoLuo: Wasn't aware we ran tests for SharedFx, this test is ported from TargetingPackTests and covers the PlatformManifest I added to the runtime pack.

natemcmaster (approved):

src/Framework/test/SharedFxTests.cs
natemcmaster: FYI @jkotalik  

Yes, we have some limited testing for the manifests that we generate. Errors in the manifests can cause big issues, and we've had regressions in this area before, so automated testing is super helpful.

jkotalik: Is this tangential to the rollforward change?

JunTaoLuo: This test is actually following up to my previous change https://github.com/aspnet/AspNetCore/commit/7a8487e2bdbe72c0aa3146ca0b82842da99eec0f. But I'm including it here since I don't want to make a separate PR.

jkotalik: Cool. Should we be responsible for adding a test to make sure roll forward works?

natemcmaster:
FYI @vitek-karas

JunTaoLuo:
Hmm the tests are failing on the CI but I ran them locally then they passed. Need to investigate.

src/Framework/test/SharedFxTests.cs
JunTaoLuo: Seems like there are some clarification needed in this area. I'll hold off on adding the tests for now. As for the roll-forward behaviour tests, I think that belongs in core-host that actually interprets this setting. It is sufficient to test that the setting exists in our tests.

JunTaoLuo:
A few issues with github checks not updating but the required builds/tests are passing. I'll leave this PR open so AzDO infrastructure can take a look but otherwise I'll merge this tomorrow.

JunTaoLuo:
Looks like the issue is tracked in https://github.com/dotnet/core-eng/issues/6416#event-2411148101.

JunTaoLuo:
Also not going to wait for Helix tests for this change either since it should not be affected. The benchmarking-pr check should not have been run so I'm going to ignore that failure as well. Looks like it's been broken for a while but we can look into fixing it separately.

PR #11151, opened 2019-06-12
Add logging to debug CI hangs

JunTaoLuo:
This might help with our hang investigation.

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -1,3 +1,4 @@
 @ECHO OFF
 SETLOCAL
-PowerShell -NoProfile -NoLogo -ExecutionPolicy unrestricted -Command "[System.Threading.Thread]::CurrentThread.CurrentCulture = ''; [System.Threading.Thread]::CurrentThread.CurrentUICulture = ''; try { & '%~dp0build.ps1' %*; exit $LASTEXITCODE } catch { write-host $_; exit 1 }"
+PowerShell -NoProfile -NoLogo -ExecutionPolicy Bypass -Command "[System.Threading.Thread]::CurrentThread.CurrentCulture = ''; [System.Threading.Thread]::CurrentThread.CurrentUICulture = ''; try { & '%~dp0build.ps1' %*; exit $LASTEXITCODE } catch { write-host $_; exit 1 }"
+ECHO build.cmd completed
--- a/build.ps1
+++ b/build.ps1
@@ -124,6 +124,9 @@ param(
     # MSBuild for .NET Core
     [switch]$ForceCoreMsbuild,
 
+    # Diagnostics
+    [switch]$DumpProcesses, # Capture all running processes and dump them to a file.
+
     # Other lifecycle targets
     [switch]$Help, # Show help
 
@@ -243,8 +246,11 @@ $env:DOTNET_HOME = $DotNetHome
 
 # Execute
 
-# Dump running processes
-Start-Job -Name DumpProcesses -FilePath $PSScriptRoot\eng\scripts\dump_process.ps1 -ArgumentList $PSScriptRoot
+if ($DumpProcesses)
+{
+    # Dump running processes
+    Start-Job -Name DumpProcesses -FilePath $PSScriptRoot\eng\scripts\dump_process.ps1 -ArgumentList $PSScriptRoot
+}
 
 $korebuildPath = Get-KoreBuild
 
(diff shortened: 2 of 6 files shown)


build.cmd
```diff
@@ -1,3 +1,4 @@
 @ECHO OFF
 SETLOCAL
-PowerShell -NoProfile -NoLogo -ExecutionPolicy unrestricted -Command "[System.Threading.Thread]::CurrentThread.CurrentCulture = ''; [System.Threading.Thread]::CurrentThread.CurrentUICulture = ''; try { & '%~dp0build.ps1' %*; exit $LASTEXITCODE } catch { write-host $_; exit 1 }"
+PowerShell -NoProfile -NoLogo -ExecutionPolicy Bypass -Command "[System.Threading.Thread]::CurrentThread.CurrentCulture = ''; [System.Threading.Thread]::CurrentThread.CurrentUICulture = ''; try { & '%~dp0build.ps1' %*; exit $LASTEXITCODE } catch { write-host $_; exit 1 }"
```
JunTaoLuo: I looked at the build scripts of CoreFX and everywhere else in arcade and it seems like ByPass is used as the execution policy instead of unrestricted. I'm not sure if this will help but it should be safe to try.

dougbu (approved):

JunTaoLuo:
Hmm seems like it didn't help with the build hang. In fact build.ps1 didn't complete.

eng/scripts/dump_process.ps1
```diff
@@ -0,0 +1,6 @@
+Set-Location $args[0]
+
+while ($true) {
+    Get-Process > artifacts/log/processes/runningProcesses.$(get-date -f yyyy-MM-dd-HH-mm-ss).txt
```
dougbu: We don't really need anything but the last copy of this file and our artifacts bundles are already unreadable. Repeatedly overwrite the same file, append to a single file, or delete all but the last one or two copies of this file. (The specific choice doesn't matter to me.)

dougbu (changes requested):

build.cmd
dougbu: FYI ByPass tends to work better in non-interactive cases because `powershell` will never prompt a non-existing user to see if it should continue. But, I doubt this change will help _complete_ scripts more reliably.

dougbu (approved):

dougbu:
Thanks @JunTaoLuo ❕

jkotalik:
Will this PR add process termination as a finalizer?

build.ps1
```diff
@@ -372,4 +381,16 @@ finally {
...
     Remove-Item env:KOREBUILD_KEEPGLOBALJSON
+
+    if ($DumpProcesses -or $CI)
+    {
+        Stop-Job -Name DumpProcesses
+        Remove-Job -Name DumpProcesses
+    }
+
+    if ($CI) {
+        & "$PSScriptRoot/eng/scripts/KillProcesses.ps1"
```
JunTaoLuo: @jkotalik Yes right here.

JunTaoLuo:
Note going to wait on helix to merge this change.

build.cmd
```diff
@@ -1,3 +1,4 @@
...
+ECHO build.cmd completed
```
dougbu: @JunTaoLuo I just had a look at the build for this PR and notice the Windows x64 build step (https://dev.azure.com/dnceng/public/_build/results?buildId=224899&view=logs&j=f31c9f97-4411-58e7-49ac-fc73f645e6b6) should have failed but did not. This line loses the exit code from the PowerShell invocation.

And, jobs are still timing out even though your change is in. For example the internal build https://dnceng.visualstudio.com/internal/_build/results?buildId=224691&view=logs&j=f31c9f97-4411-58e7-49ac-fc73f645e6b6 shows:
```
2019-06-14T02:35:06.4970654Z     84 Warning(s)
2019-06-14T02:35:06.4971845Z     16 Error(s)
2019-06-14T02:35:06.4972524Z
2019-06-14T02:35:06.4974642Z Time Elapsed 00:39:21.51
2019-06-14T04:52:55.8328920Z ##[error]The operation was canceled.
2019-06-14T04:52:55.8691848Z ##[section]Finishing: Build x64
```

The above looks very much like MSBuild itself is not exiting after it wraps up, not PowerShell.

Let me know if you can't tackle this almost immediately.

JunTaoLuo: Let me remove the logging commands and investigate

JunTaoLuo: I don't have access to the diagnostics logs since they were not uploaded in time after the timeout grace period. I'm going to move that build step earlier and see if it helps.

JunTaoLuo: Turns out it's not really configurable. Artifacts are uploaded after test results are uploaded. Hmm.... this mechanism might not work.

PR #11157, opened 2019-06-12
Ensure CircuitRegistry evicts CircuitHost entries after configured

pranavkm:
duration

* Use an active expiration token to trigger expiration
* Add logging during host state transitions

Fixes https://github.com/aspnet/AspNetCore/issues/9893

### DIFF
--- a/src/Components/Server/src/Circuits/CircuitHost.cs
+++ b/src/Components/Server/src/Circuits/CircuitHost.cs
@@ -194,6 +194,8 @@ namespace Microsoft.AspNetCore.Components.Server.Circuits
 
         private async Task OnCircuitOpenedAsync(CancellationToken cancellationToken)
         {
+            Log.CircuitOpened(_logger, Circuit.Id);
+
             for (var i = 0; i < _circuitHandlers.Length; i++)
             {
                 var circuitHandler = _circuitHandlers[i];
@@ -210,6 +212,8 @@ namespace Microsoft.AspNetCore.Components.Server.Circuits
 
         public async Task OnConnectionUpAsync(CancellationToken cancellationToken)
         {
+            Log.ConnectionUp(_logger, Circuit.Id, Client.ConnectionId);
+
             try
             {
                 await HandlerLock.WaitAsync(cancellationToken);
@@ -235,6 +239,8 @@ namespace Microsoft.AspNetCore.Components.Server.Circuits
 
         public async Task OnConnectionDownAsync(CancellationToken cancellationToken)
         {
+            Log.ConnectionDown(_logger, Circuit.Id, Client.ConnectionId);
+
             try
             {
                 await HandlerLock.WaitAsync(cancellationToken);
@@ -265,6 +271,8 @@ namespace Microsoft.AspNetCore.Components.Server.Circuits
 
         private async Task OnCircuitDownAsync()
         {
+            Log.CircuitClosed(_logger, Circuit.Id);
+
             for (var i = 0; i < _circuitHandlers.Length; i++)
             {
                 var circuitHandler = _circuitHandlers[i];
@@ -283,13 +291,13 @@ namespace Microsoft.AspNetCore.Components.Server.Circuits
         {
             Log.DisposingCircuit(_logger, CircuitId);
 
-            await Renderer.InvokeAsync((Func<Task>)(async () =>
+            await Renderer.InvokeAsync(async () =>
             {
                 await OnConnectionDownAsync(CancellationToken.None);
                 await OnCircuitDownAsync();
                 Renderer.Dispose();
                 _scope.Dispose();
-            }));
+            });
         }
 
         private void AssertInitialized()
@@ -314,11 +322,19 @@ namespace Microsoft.AspNetCore.Components.Server.Circuits
         {
             private static readonly Action<ILogger, Type, string, string, Exception> _unhandledExceptionInvokingCircuitHandler;
             private static readonly Action<ILogger, string, Exception> _disposingCircuit;
+            private static readonly Action<ILogger, string, Exception> _onCircuitOpened;
+            private static readonly Action<ILogger, string, string, Exception> _onConnectionUp;
+            private static readonly Action<ILogger, string, string, Exception> _onConnectionDown;
+            private static readonly Action<ILogger, string, Exception> _onCircuitClosed;
 
             private static class EventIds
             {
                 public static readonly EventId ExceptionInvokingCircuitHandlerMethod = new EventId(100, "ExceptionInvokingCircuitHandlerMethod");
                 public static readonly EventId DisposingCircuit = new EventId(101, "DisposingCircuitHost");
+                public static readonly EventId OnCircuitOpened = new EventId(102, "OnCircuitOpened");
+                public static readonly EventId OnConnectionUp = new EventId(103, "OnConnectionUp");
+                public static readonly EventId OnConnectionDown = new EventId(104, "OnConnectionDown");
+                public static readonly EventId OnCircuitClosed = new EventId(105, "OnCircuitClosed");
             }
 
             static Log()
@@ -332,6 +348,26 @@ namespace Microsoft.AspNetCore.Components.Server.Circuits
                     LogLevel.Trace,
                     EventIds.DisposingCircuit,
                     "Disposing circuit with identifier {CircuitId}");
+
+                _onCircuitOpened = LoggerMessage.Define<string>(
+                    LogLevel.Debug,
+                    EventIds.OnCircuitOpened,
+                    "Opening circuit with id {CircuitId}.");
+
+                _onConnectionUp = LoggerMessage.Define<string, string>(
+                    LogLevel.Debug,
+                    EventIds.OnConnectionUp,
+                    "Circuit id {CircuitId} connected using connection {ConnectionId}.");
+
+                _onConnectionDown = LoggerMessage.Define<string, string>(
+                    LogLevel.Debug,
+                    EventIds.OnConnectionDown,
+                    "Circuit id {CircuitId} disconnected from connection {ConnectionId}.");
+
+                _onCircuitClosed = LoggerMessage.Define<string>(
+                   LogLevel.Debug,
+                   EventIds.OnCircuitClosed,
+                   "Closing circuit with id {CircuitId}.");
             }
 
             public static void UnhandledExceptionInvokingCircuitHandler(ILogger logger, CircuitHandler handler, string handlerMethod, Exception exception)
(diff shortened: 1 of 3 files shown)


src/Components/Server/src/Circuits/CircuitRegistry.cs
```diff
@@ -104,44 +107,60 @@ public virtual Task DisconnectAsync(CircuitHost circuitHost, string connectionId
...
         {
+            var cancellationTokenSource = new CancellationTokenSource(_options.DisconnectedCircuitRetentionPeriod);
             var entryOptions = new MemoryCacheEntryOptions
             {
-                AbsoluteExpiration = DateTimeOffset.UtcNow.Add(_options.DisconnectedCircuitRetentionPeriod),
                 Size = 1,
                 PostEvictionCallbacks = { _postEvictionCallback },
+                ExpirationTokens =
+                {
+                    new CancellationChangeToken(cancellationTokenSource.Token),
```
pranavkm: cc @Tratcher

src/Components/Server/src/Circuits/CircuitHost.cs
```diff
@@ -194,6 +194,8 @@ public async void BeginInvokeDotNetFromJS(string callId, string assemblyName, st
 
         private async Task OnCircuitOpenedAsync(CancellationToken cancellationToken)
         {
+            Log.CircuitOpened(_logger, Circuit.Id);
```
pranavkm: What do you think about using a display string here instead of the actual Id. The Id is really long and it's marginally useful for users to know what it's actual value is versus being able to distinguish circuits in some way.

>  Connection up for CfDJ8DP9EAkPGfZAsPNtItRjMVWZzd4_ED00FlFBkVSfA4PoLIQIGWGUpVewG5VAw5opZNYSWQePn-gmwHgsXJyfAir1vmy_9Gl3ty4cjsXNKgw_y_g1sKSvOjFUB22MG7RGChX5opeXICO0VaQcO3KS7GZDDXQXVMu0CLoWoHSpbB8P.

javiercn: This will change with my next PR, so don't worry too much about it. I can shorten it a bit more if necessary.

src/Components/Server/src/Circuits/CircuitRegistry.cs
```diff
@@ -190,36 +211,45 @@ public virtual async Task<CircuitHost> ConnectAsync(string circuitId, IClientPro
...
                 // The host is still active i.e. the server hasn't detected the client disconnect.
                 // However the client reconnected establishing a new connection.
-                circuitHost.Client.Transfer(clientProxy, connectionId);
-                return (circuitHost, true);
+                connectedCircuitHost.Client.Transfer(clientProxy, connectionId);
+                return (connectedCircuitHost, true);
             }
 
-            if (DisconnectedCircuits.TryGetValue(circuitId, out circuitHost))
+            if (DisconnectedCircuits.TryGetValue(circuitId, out DisconnectedCircuitEntry disconnectedEntry))
```
SteveSandersonMS: Nit:

```suggestion
            if (DisconnectedCircuits.TryGetValue(circuitId, out var disconnectedEntry))
```

src/Components/Server/src/Circuits/CircuitRegistry.cs
```diff
@@ -190,36 +211,45 @@ public virtual async Task<CircuitHost> ConnectAsync(string circuitId, IClientPro
...
 
             return default;
         }
 
-        private void OnEntryEvicted(object key, object value, EvictionReason reason, object state)
+        protected virtual void OnEntryEvicted(object key, object value, EvictionReason reason, object state)
         {
             switch (reason)
             {
                 case EvictionReason.Expired:
```
SteveSandersonMS: Is the `Expired` case still possible?

src/Components/Server/src/Circuits/CircuitRegistry.cs
```diff
@@ -232,42 +262,182 @@ private void OnEntryEvicted(object key, object value, EvictionReason reason, obj
...
+        private void DisposeTokenSource(DisconnectedCircuitEntry entry)
+        {
+            try
+            {
+                entry.TokenSource.Dispose();
+            }
+            catch (Exception ex)
+            {
+                Log.ExceptionDisposingTokenSource(_logger, ex);
+            }
```
SteveSandersonMS: This is interesting. For what reasons do you expect it's possible that `entry.TokenSource.Dispose()` would throw? Does this trigger a synchronous callback in our code somewhere?

SteveSandersonMS (approved):
Besides minor questions, this looks good to me!

src/Components/Server/src/Circuits/CircuitRegistry.cs
```diff
@@ -232,42 +262,182 @@ private void OnEntryEvicted(object key, object value, EvictionReason reason, obj
...
 
-        private async Task DisposeCircuitHost(CircuitHost circuitHost)
+        private async Task DisposeCircuitEntry(DisconnectedCircuitEntry entry)
         {
+            DisposeTokenSource(entry);
+
             try
             {
-                await circuitHost.DisposeAsync();
+                await entry.CircuitHost.DisposeAsync();
```
javiercn: Doesn't this need to happen within the sync context?

src/Components/Server/test/Circuits/CircuitRegistryTest.cs
```diff
@@ -313,16 +314,108 @@ public async Task DisconnectWhenAConnectIsInProgress()
...
+                mre.Set();
+            };
+            var circuitHost = TestCircuitHost.Create();
+
+            registry.RegisterDisconnectedCircuit(circuitHost);
+
+            // Act
+            // Verify it's present in the dictionary.
+            Assert.True(registry.DisconnectedCircuits.TryGetValue(circuitHost.CircuitId, out var _));
+            await Task.Run(() => Assert.True(mre.Wait(TimeSpan.FromSeconds(10))));
```
javiercn: Is there a more reliable way to do this other than using an mre. Maybe simple set a flag in OnAfterEntryEvicted and do await Task.Delay(4) before checking the flag was set?

src/Components/Server/test/Circuits/CircuitRegistryTest.cs
```diff
@@ -313,16 +314,108 @@ public async Task DisconnectWhenAConnectIsInProgress()
...
+                registry.RegisterDisconnectedCircuit(hosts[i]);
+            }
+
+            // Assert
+            for (var i = 0; i < maxCircuits; i++)
+            {
+                Assert.True(registry.DisconnectedCircuits.TryGetValue(hosts[i].CircuitId, out var _));
+            }
+
+            // Additional circuits do not get registered.
```
javiercn: We need to rethink this behavior.

src/Components/Server/test/Circuits/CircuitRegistryTest.cs
SteveSandersonMS: Out of interest, why would that be more reliable? I would have thought the mre was preferable and more reliable, because it continues immediately after the flag is set, so we can give it up to 10 seconds in case code on one thread happens to run much faster than code on another thread.

src/Components/Server/src/Circuits/CircuitRegistry.cs
```diff
@@ -104,44 +107,60 @@ public virtual Task DisconnectAsync(CircuitHost circuitHost, string connectionId
...
             RegisterDisconnectedCircuit(circuitHost);
+
+            Log.CircuitMarkedDisconnected(_logger, circuitId);
+
             return true;
         }
 
         public void RegisterDisconnectedCircuit(CircuitHost circuitHost)
         {
+            var cancellationTokenSource = new CancellationTokenSource(_options.DisconnectedCircuitRetentionPeriod);
```
javiercn: Will this create 1 timer per circuit? Is that something we should care about? Should this instead be handled by something that executes at regular intervals and clears circuits past some expiration time?

src/Components/Server/test/Circuits/CircuitRegistryTest.cs
javiercn: I meant simpler. Reliable is probably not the right way. I'm still waking up :)

javiercn (approved):

src/Components/Server/src/Circuits/CircuitRegistry.cs
pranavkm: This unfortunately doesn't work. `MemoryCache` returns an `object` instance by default (yes we're boxing these) and you have to specify the type to use the `TryGetValue<T>`  overload

src/Components/Server/src/Circuits/CircuitRegistry.cs
pranavkm: We shouldn't. I'll remove it so we know if it's being weird

src/Components/Server/test/Circuits/CircuitRegistryTest.cs
pranavkm: Yup. Maybe the cache needs a LRU policy. Right now it effectively always rejects newer adds: https://github.com/aspnet/Extensions/blob/master/src/Caching/Memory/src/MemoryCache.cs#L156.

src/Components/Server/test/Circuits/CircuitRegistryTest.cs
pranavkm: I remember @Tratcher changing all MRE's recently to TCS's for reliability. I'll make that change and await the TCS.Task instead here.

src/Components/Server/test/Circuits/CircuitRegistryTest.cs
pranavkm: @Tratcher \ @JunTaoLuo \ @davidfowl  what do you think about having a eviction strategy for capacity in the cache? All entries being equal priority, we would want the cache to do knock things based on LRU. Should we add a feature to the cache or roll our own thing?

Tratcher: LRU is already part of the eviction strategy.

Tratcher: https://github.com/aspnet/Extensions/blob/f12d709976e382672ce100bc00381a8847f06489/src/Caching/Memory/src/MemoryCache.cs#L460-L461

src/Components/Server/src/Circuits/CircuitRegistry.cs
davidfowl: > so it should be fairly efficient.

😄 Famous last words. While it is efficient. If most of your objects have the same interval then it might be better to have a single recurring timer that loops over all of your circuits. SignalR tries really hard to avoid making too many timers.

pranavkm: Let's profile this first and we can figure it out

PR #11161, opened 2019-06-12
Use a valid canceled task

pranavkm:
Task.FromCanceled throws if the token isn't cancelled. From the docs: 

> Exceptions
ArgumentOutOfRangeException
Cancellation has not been requested for cancellationToken; its IsCancellationRequested property is false.

### DIFF
--- a/src/Components/Server/src/Circuits/RemoteRenderer.cs
+++ b/src/Components/Server/src/Circuits/RemoteRenderer.cs
@@ -3,13 +3,11 @@
 
 using System;
 using System.Collections.Concurrent;
-using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Text.Encodings.Web;
 using System.Threading;
 using System.Threading.Tasks;
-using MessagePack;
 using Microsoft.AspNetCore.Components.Rendering;
 using Microsoft.AspNetCore.Components.Server.Circuits;
 using Microsoft.AspNetCore.SignalR;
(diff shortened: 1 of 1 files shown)


davidfowl (approved):

SteveSandersonMS (approved):

javiercn (approved):
#perf

src/Components/Server/src/Circuits/RemoteRenderer.cs
```diff
@@ -114,7 +114,7 @@ protected override Task UpdateDisplayAsync(in RenderBatch batch)
             if (_disposing)
             {
                 // We are being disposed, so do no work.
-                return Task.FromCanceled<object>(CancellationToken.None);
+                return CanceledTask;
```
rynowak: Was there a repro for this issue ? or just an random improvment?

pranavkm: I was messing around with circuit expiry and saw the ArgOutOfRange exception pop up in the console. I haven't super investigated it

rynowak: Sounds good, and this is the exactly the kind of issues we should fixing during the reliablity/capacity/perf push.

rynowak (approved):
**super** approved

pranavkm:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=224856&view=logs&jobId=bfccef3f-45d4-504d-9a7f-a8a74922e6bf

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/2674

PR #11162, opened 2019-06-12
Clean up Pubternal Types in SignalR

mikaelm12:
Issue: https://github.com/aspnet/AspNetCore/issues/10169

Some things to note
- `ServerSentEventsMessageFormatter` used in the Microbenchmark project but that causes a bunch of issues
- `HubOptionsSetup` and `HubOptionsSetup<T>` is used in `SignalRDependencyInjectionExtensions` so Ieft those as pubternal. Didn't want to use internals visible to in this case. Should this just be public?


halter73:
I think `HubOptionsSetup` and `HubOptionsSetup<T>` should be public if they're referenced across shipped assemblies.

halter73 (approved):

halter73:
The approval assumes `HubOptionsSetup` and `HubOptionsSetup<T>` are the last pubternal types and will be made public.

mikaelm12:
I agree about the HubOptions. The `ServerSentEventsMessageFormatter` is currently still pubternal because of the Microbenchmarking project. I still need to resolve that

anurse:
> The `ServerSentEventsMessageFormatter` is currently still pubternal because of the Microbenchmarking project. I still need to resolve that

It seems OK to use `InternalsVisibleTo` there, does it not work? It turns out Arcade is the one that provides the `InternalsVisibleTo` MSBuild item and we **do** support it in AspNetCore. See https://github.com/aspnet/AspNetCore/blob/f28cf2bbc8e97a05be575a61af08f01edf0ac9c9/src/Middleware/HeaderPropagation/src/Microsoft.AspNetCore.HeaderPropagation.csproj#L12-L14

anurse:
> * `HubOptionsSetup` and `HubOptionsSetup<T>` is used in `SignalRDependencyInjectionExtensions` so Ieft those as pubternal.

Seems unfortunate but it would be pretty funky to refactor. We could put a method in `...SignalR.Core` somewhere that takes an `IServiceProvider` and registers `HubOptionsSetup` in it, but that gets a bit messy. Make sure only the constructor (because DI requires `public` constructors) and interface implementations are `public`

davidfowl:
The ref assembly is the only way to review this change.

src/SignalR/common/Http.Connections/src/Internal/HttpConnectionContext.cs
```diff
@@ -323,7 +323,7 @@ private async Task WaitOnTasks(Task applicationTask, Task transportTask, bool cl
             }
         }
 
-        public bool TryActivatePersistentConnection(
+        internal bool TryActivatePersistentConnection(
```
BrennanConroy: Why only this method?

Also, this class is still pubternal.

mikaelm12:
> It seems OK to use InternalsVisibleTo there, does it not work?

I should have elaborated in the description. Just adding internals visible to doesn't doesn't quite solve the issue. It causes some ambiguous references to types like `MemoryBufferWriter`.  That's the last thing I need to work through.

BrennanConroy (commented):
We should also try to switch to the new `<InternalsVisibleTo>` tag in .csproj's!

src/SignalR/common/Http.Connections/src/Internal/HttpConnectionContext.cs
davidfowl: Why is this class pubternal still?

mikaelm12: > Why only this method?

Because `IHttpTransport` is  internal now. 

>Also, this class is still pubternal.

Yeah, I updated the issue with the new namespace that was missed and I'm updating those types now.

src/SignalR/server/Core/ref/Microsoft.AspNetCore.SignalR.Core.netcoreapp3.0.cs
```diff
@@ -167,6 +167,14 @@ public partial class HubConnectionStore
             public void Reset() { }
         }
     }
+    public abstract partial class HubDispatcher<THub> where THub : Microsoft.AspNetCore.SignalR.Hub
+    {
+        protected HubDispatcher() { }
+        public abstract System.Threading.Tasks.Task DispatchMessageAsync(Microsoft.AspNetCore.SignalR.HubConnectionContext connection, Microsoft.AspNetCore.SignalR.Protocol.HubMessage hubMessage);
+        public abstract System.Collections.Generic.IReadOnlyList<System.Type> GetParameterTypes(string name);
```
davidfowl: This is super unfortunate as a public API @anurse. Can we file an issue to see if we can do anything tactical for 3.0?

src/SignalR/common/Http.Connections/src/Internal/HttpConnectionContext.cs
halter73: From the issue.

> Edit More Pubternal Types
> Microsoft.AspNetCore.Http.Connections.Internal.

> HttpConnectionContext
> HttpConnectionDispatcher
> HttpConnectionManager
> HttpConnectionStatus

Looking at the [ref assembly](https://github.com/aspnet/AspNetCore/blob/017782ccc576c92164bd31b49db7df4f517fd21a/src/SignalR/common/Http.Connections/ref/Microsoft.AspNetCore.Http.Connections.netcoreapp3.0.cs) it looks like we should add AwaitableThreadPool to the list.

davidfowl: That’s public?

halter73: I think it was pubternal so it could be called from Http.Connections and Client.Core, but we're much better off just copying it.

src/SignalR/server/Core/ref/Microsoft.AspNetCore.SignalR.Core.netcoreapp3.0.cs
halter73: Why is a HubDispatcher registered with DI to begin with? Is it just so we don't have to construct in manually in HubConnectionHandler? It looks like we don't actually use anything other than the DefaultHubDispatcher even in our tests. There are no other HubConnectionHandler implementations at all, not even test mocks.

I even checked https://github.com/Azure/azure-signalr to see if they might have taken a dependency on it somehow, but it doesn't seem so. We should just construct a DefaultHubDispatcher manually in HubConnectionHandler.

src/SignalR/common/Http.Connections/src/Internal/HttpConnectionContext.cs
davidfowl: Shared codeZ

mikaelm12: >  like we should add AwaitableThreadPool to the list

Yeah, I made it internal but forgot to add it to the issue.

src/SignalR/server/Core/ref/Microsoft.AspNetCore.SignalR.Core.netcoreapp3.0.cs
anurse: It was to provide a way for users to do something like what hub pipeline modules let you do in ASP.NET SignalR. I'm OK with removing the extensibility point for now and considering "real" pipeline modules (https://github.com/aspnet/AspNetCore/issues/5353)

halter73: @mikaelm12 I think we're in agreement. For now, lets make `HubDispatcher<THub>` internal and remove it from DI. I think you just need manually construct a DefaultHubDispatcher in the HubConnectionHandler ctor.

davidfowl: Seems fine. Also, we're doing hub pipeline modules next release.

mikaelm12:
Had to revert the change to use the `InternalsVisibleTo` msbuild tag because it doesn't work with Moq
Filed an issue with arcade: https://github.com/dotnet/arcade/issues/3068

mikaelm12:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=227887&view=logs

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/2303

Pilchie:
Roslyn is able to use `InternalsVisibleTo` in MSBuild with Moq (see [here](https://github.com/dotnet/roslyn/blob/master/src/VisualStudio/Core/Def/Microsoft.VisualStudio.LanguageServices.csproj#L111))

BrennanConroy:
https://github.com/dotnet/roslyn/blob/d4afaea39dc21a45ac119dd2bec9be678ce5081d/src/VisualStudio/Core/Def/Microsoft.VisualStudio.LanguageServices.csproj#L120

Ah, they had to provide the key

mikaelm12:
Ah okay, I'll try that out. Thanks!

src/SignalR/server/Core/src/HubConnectionHandler.cs
```diff
@@ -59,10 +59,16 @@ HubDispatcher<THub> dispatcher
...
 
             _enableDetailedErrors = _hubOptions.EnableDetailedErrors ?? _globalHubOptions.EnableDetailedErrors ?? false;
             _maximumMessageSize = _hubOptions.MaximumReceiveMessageSize ?? _globalHubOptions.MaximumReceiveMessageSize;
+
+            _dispatcher = new DefaultHubDispatcher<THub>(
+                serviceScopeFactory,
+                new HubContext<THub>(lifetimeManager),
+                hubOptions,
+                globalHubOptions,
+                new Logger<DefaultHubDispatcher<THub>>(loggerFactory));
```
davidfowl: Wow this is kinda gross 😄

davidfowl (approved):
I'll follow up with why ServerSentEventsMessageFormatter is public.

src/SignalR/server/Core/src/HubConnectionHandler.cs
BrennanConroy: Should we also remove `HubDispatcher` from DI now that it's no longer used in DI?

mikaelm12:
Going to follow this up with another clean up  pr for `InternalsVisibleTo` and SSEMessageFormatter

parkycai:
Is there a new way to implement custom HubDispatcher and inject it?

PR #11164, opened 2019-06-13
Attempt to reload if reconnecting to a prerendered circuit fails

pranavkm:
A before and after to help explain what's happening. If I close a tab, and revive it (Ctrl+Shift+T) after the circuit has been disposed, the server rejects the initial attempt to reconnect if you have pre-rendered content. Reloading the application seemed like as a good a solution as any other, but I'm open to alternatives:

## Before
Before

## After
After


src/Components/Browser.JS/src/Boot.Server.ts
```diff
@@ -95,7 +95,16 @@ async function boot(userOptions?: Partial<BlazorOptions>): Promise<void> {
...
   }
 
-  await reconnectTask;
+  if (!await reconnectTask) {
+    // reconnection for the initialConnection will only fail when the application is prerendered but the server no longer has state for the
+    // Circuit. One fairly trivial way to get in to this state is to have the browser reopen a closed tab with the initial render rehydrated
+    // from cache.
+    // In this event, the best thing to try would be to attempt reloading the page.
+    console.assert(!renderingFailed, 'Initial connection should not have encountered a rendering failure.');
+    console.assert(circuits.length > 0, 'Initial render should only fail if there are circuits available.');
```
SteveSandersonMS: I'm not so sure about using `console.assert` in this kind of way. These messages make sense to *us*, but they are phrased very oddly from the point of view of developers using the framework. I think customers would get on better with messages phrased like:

```
Initial connection failed during rendering.
```

or

```
Initial connection failed, even though no prerendered components were discovered in the page.
```

These still aren't super useful messages (it doesn't provide any indication about what you could do to fix it) but at least it's more direct about saying what the problem is.

src/Components/Browser.JS/src/Boot.Server.ts
```diff
@@ -95,7 +95,16 @@ async function boot(userOptions?: Partial<BlazorOptions>): Promise<void> {
     circuits.push(circuit);
   }
 
-  await reconnectTask;
+  if (!await reconnectTask) {
```
javiercn: We should return something from the circuit to specifically signal that the circuit is gone and then do different things based on whether we were able to ever connect to the circuit.

* If we prerendered, closed and reopened, then it's fair game to reload.
* If not, we should simply display different UI explaining what happened and offering the user to reload the app.
  * Its a very bad user experience if you go through a tunnel or close your laptop for 1 minute and then suddenly your 
    app gets refreshed without you knowing why and potentially loosing data/etc.
  * Apps should be able to customize what happens in this case, for example to save form fields in local storage and so on before refreshing.

SteveSandersonMS:
**Mostly** this looks like a good thing to me. I'd expect the primary case where it gets used is during development:

1. the developer compiles some changes
1. the old server goes down
1. a new server comes up
1. the client tries to reconnect
1. the new server doesn't know the old circuit ID so rejects the connection
1. the client reloads
1. the developer carries on with their work

Do you agree that sequence is what happens during development? If so that's a big improvement over seeing "connection failed" errors and having to reload manually.

**However**, my one concern about this technique is: does it risk an infinite-reload loop? If there's some underlying reason why the client can never reconnect, it will just go into a spin, possibly trying to reload dozens of times per second, making it really hard to see what the error is, right?

I don't know if you can think of any elegant solution to that. For me a better situation would be something like: *only* reload if we know for sure the error was the server saying it doesn't know that circuit ID. Don't reload for other types of error. Do you think that's possible?

javiercn (changes requested):
I think this is a positive change, but that we need to tweak it a bit as I think the proposed behavior will lead to a bad user experience in some common cases and that application developers must have a say in when the application reloads.

SteveSandersonMS:
Oh, looks like @javiercn and I are writing up similar ideas simultaneously!

src/Components/Browser.JS/src/Boot.Server.ts
SteveSandersonMS: > Apps should be able to customize what happens in this case, for example to save form fields in local storage and so on before refreshing.

That sounds great, but I'd be OK with considering it out of scope for this PR. It sounds like a bigger feature change. This PR doesn't regress the existing situation, since there isn't yet a way of doing that anyway.

javiercn: As long as we can detect in a deterministic way that the circuit is gone (the server says so) we shouldn't reload automatically. Other than that, although I would like to see app developers having a choice about when the reload happens, I'm ok with simply reloading for now (only when we know for sure the circuit is gone).

src/Components/Browser.JS/src/Boot.Server.ts
javiercn: I also wouldn't use console.xxx directly. We have a logger for these things.

pranavkm:
This change specifically affects the initial connect \ reconnect when you have pre-rendering on. 

None of the behavior around `reconnect` when when the client is disconnected has changed. It's entirely up to the user to determine what happens when `reconnect` fails. I'm not entirely convinced it's meaningful to let the user determine state transitions when pre-rendering is meant to be entirely a feature of the framework. 

> However, my one concern about this technique is: does it risk an infinite-reload loop?

There is a potential for an infinite loop when the circuit registry disallows setting up circuits for pre-rendered content. As I noted here https://github.com/aspnet/AspNetCore/pull/11157#discussion_r293457423, the registry rejects adding new disconnected circuits when it's at capacity. Consequently if you have pre-rendering on, but too many disconnected circuits, we will not register the circuit for the pre-rendered content. We need to fix this behavior, but that's the one immediate scenario that I know is broken.

As for the scenario you listed, the scenario is only improved if:

1) Server0 sends pre-rendered content
2) Server0 goes down
3) Client attempts to connect to server
4) Server1 comes up just in time for the client reconnect to come through
5) Server1 rejects the reconnect since it has no idea what it is.

If step 3 had already completed, none of the behavior would change.

That said, if this issue is turning out to be contentious, I'll file an issue so we can triage it for an appropriate milestone. I don't want to spend more cycles for a scenario that's not mainline.

pranavkm:
Tracking this issue via https://github.com/aspnet/AspNetCore/issues/11187

javiercn:
No! I think this change is good. 😢 Don't give up on it!

PR #11165, opened 2019-06-13
Work around potential race in PipeWriter (IIS Edition)

halter73:
We should wait until #11065 (the Kestrel version of this) is merged and stable. If everything looks good for the Kestrel change, and it didn't have to detrimental of an impact to our benchmarks, I think we should take this despite the lack of IIS-related customer reports so far.

@jkotalik @davidfowl @pakrym if you're still interested.

### DIFF
--- a/src/Servers/IIS/IIS/src/Core/OutputProducer.cs
+++ b/src/Servers/IIS/IIS/src/Core/OutputProducer.cs
@@ -22,16 +22,15 @@ namespace Microsoft.AspNetCore.Server.IIS.Core
         private readonly Pipe _pipe;
 
         // https://github.com/dotnet/corefxlab/issues/1334
-        // Pipelines don't support multiple awaiters on flush
-        // this is temporary until it does
-        private TaskCompletionSource<object> _flushTcs;
+        // https://github.com/aspnet/AspNetCore/issues/8843
+        // Pipelines don't support multiple awaiters on flush. This is temporary until it does.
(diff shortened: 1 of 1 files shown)


jkotalik:
Got some build errors there 😮

halter73:
Maybe I should try building before submitting my PR next time 😆

anurse:
But that's what The ☁️ is for ;)

src/Servers/IIS/IIS/src/Core/OutputProducer.cs
```diff
@@ -90,35 +87,27 @@ public Task WriteAsync(ReadOnlyMemory<byte> buffer, CancellationToken cancellati
...
+                    FlushNowAsync(pipeWriter, cancellationToken) :
+                    AwaitLastFlushAndThenFlushAsync(_lastFlushTask, pipeWriter, cancellationToken);
+
+                return _lastFlushTask;
             }
-            return FlushAsyncAwaited(awaitable, cancellationToken);
         }
 
-        private async Task FlushAsyncAwaited(ValueTask<FlushResult> awaitable, CancellationToken cancellationToken)
+        private Task FlushNowAsync(PipeWriter pipeWriter, CancellationToken cancellationToken)
```
jkotalik: FlushNow is... an interesting name. I can't think of anything better.

jkotalik (approved):
Fix looks good, let's make sure the Kestrel changes are stable.

PR #11166, opened 2019-06-13
Add API to configure CircuitOptions

pranavkm:


src/Components/Server/src/DependencyInjection/ServerSizeBlazorBuilderExtensions.cs
```diff
@@ -12,6 +12,29 @@ namespace Microsoft.Extensions.DependencyInjection
     /// </summary>
     public static class ServerSizeBlazorBuilderExtensions
     {
+        /// <summary>
+        /// Adds options to configure circuits.
+        /// </summary>
+        /// <param name="builder">The <see cref="IServerSideBlazorBuilder"/>.</param>
+        /// <param name="configure">A callback to configure <see cref="CircuitOptions"/>.</param>
+        /// <returns>The <see cref="IServerSideBlazorBuilder"/>.</returns>
+        public static IServerSideBlazorBuilder AddCircuitOptions(this IServerSideBlazorBuilder builder, Action<CircuitOptions> configure)
```
pranavkm: Makes it a bit more discoverable. This is fairly undiscoverable now. How do we feel about `AddServerSideBlazor` optionally accepting a callback to configure `CircuitOptions`?

src/Components/Server/ref/Microsoft.AspNetCore.Components.Server.netcoreapp3.0.cs
```diff
@@ -103,6 +103,7 @@ public partial interface IServerSideBlazorBuilder
     }
     public static partial class ServerSizeBlazorBuilderExtensions
```
SteveSandersonMS: I know it predates your PR, but there's a typo in the class name. Server "size"?

SteveSandersonMS: The typo is also in the filename. Would you mind fixing it as part of this PR?

SteveSandersonMS (approved):
Looks like a good improvement to me.

Without this PR, is the only way to configure `CircuitOptions` calling `services.Configure<CircuitOptions>(...)` manually? If so I agree that's extremely hard to discover.

src/Components/Server/ref/Microsoft.AspNetCore.Components.Server.netcoreapp3.0.cs
javiercn: lol. We should change the product to be ServerSizeBlazor 😄

src/Components/Server/src/DependencyInjection/ServerSizeBlazorBuilderExtensions.cs
javiercn: I think that would be good. Its a bit jarring to see AddServer**Size**Blazor and the options type name to be different, but I can live with that.

javiercn (approved):

pranavkm:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=223774&view=ms.vss-test-web.build-test-results-tab&runId=5514352&resultId=112852&paneView=debug

aspnet-hello:

PR #11167, opened 2019-06-13
Dylan/request queue di

DylanDmitri:
Decoupled the specific queuing policy from the middleware. This means the middleware is now policy agnostic; it should be able to support different approaches (previously had some hard-coded assumptions). Added a bunch of DI plumbing, renamed some things for clarity (mostly IRequestQueue -> IQueuePolicy), and redid the tests.

This is a useful decoupling, because next I want to iterate on the queuing policy logic, and having a swappable internal component enables benchmarking and direct comparisons. Additionally, it lets customers or other contributors implement their own policies in the future.

The tests got simpler (hopefully) because the logic is now split between two simple components, rather than one moderately complex one. The tests might also have gotten simpler because I removed testing functionality. This is something I have checked, but I would appreciate if you could double-check that there's no missing coverage.


src/Middleware/RequestThrottling/perf/Microbenchmarks/QueueEmptyOverhead.cs
```diff
@@ -24,17 +21,10 @@ public void GlobalSetup()
...
-            };
-
-            _middleware = new RequestThrottlingMiddleware(
-                    next: _restOfServer,
-                    loggerFactory: NullLoggerFactory.Instance,
-                    options: Options.Create(options)
-                );
+            _middleware = TestUtils.CreateTestMiddleware_TailDrop(
+                maxConcurrentRequests: 8,
+                requestQueueLimit: 2000,
```
halter73:
```suggestion
                requestQueueLimit: _numRequests,
```

src/Middleware/RequestThrottling/perf/Microbenchmarks/QueueFullOverhead.cs
```diff
@@ -24,16 +23,10 @@ public class QueueFullOverhead
...
-            };
-
-            _middleware = new RequestThrottlingMiddleware(
-                    next: (RequestDelegate)_incrementAndCheck,
-                    loggerFactory: NullLoggerFactory.Instance,
-                    options: Options.Create(options)
+            _middleware = TestUtils.CreateTestMiddleware_TailDrop(
+                maxConcurrentRequests: MaxConcurrentRequests,
+                requestQueueLimit: _numRequests,
+                next: (RequestDelegate)_incrementAndCheck
```
halter73: I just saw that `_incrementAndCheck` is a method not a field. Make it `IncrementAndCheck`.

If you were going to pass around an IncrementAndCheck RequestDelegate a lot and it captured `this` it would make sense to store it in a field in the constructor to avoid allocating each time. But since you only pass it as an argument once during global setup, the implicit closure is fine.

src/Middleware/RequestThrottling/src/IQueuePolicy.cs
```diff
@@ -0,0 +1,27 @@
...
+        /// When it returns 'true' the request procedes to the server.
+        /// When it returns 'false' the request is rejected immediately.
+        /// </summary>
+        Task<bool> TryEnterAsync();
+
+        /// <summary>
+        /// Called after successful requests have been returned from the server.
+        /// Does NOT get called for rejected requests.
+        /// </summary>
+        void OnExit();
```
halter73: @davidfowl This is what I was talking to you about this afternoon. I think this is a good first step, but it doesn't allow the policy implementation track the time taken by individual requests. At best you can deduce the average time taken.

We talked about having TryEnterAsync return an interface that implements IDisposable instead. We also talked about queuing and dequeuing separately which would leave the middleware completely responsible for enforcing the concurrent request limit, though the policy could indirectly observe what the limit is.

The IDisposable idea doesn't seem too bad, but it's tricky to implement efficiently. The separate queue and dequeue seems like a major rearchitecture without any real concrete benefits other than it seems more pure and separates concerns better.

src/Middleware/RequestThrottling/src/Policies/TailDrop.cs
```diff
@@ -4,10 +4,11 @@
 using System;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Extensions.Options;
 
-namespace Microsoft.AspNetCore.RequestThrottling.Internal
+namespace Microsoft.AspNetCore.RequestThrottling.Policies
 {
-    internal class TailDrop : IRequestQueue
+    internal class TailDrop : IQueuePolicy, IDisposable
```
halter73: Good job separating these IQueuePolicy and IDisposable

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
```diff
@@ -70,48 +49,43 @@ public RequestThrottlingMiddleware(RequestDelegate next, ILoggerFactory loggerFa
...
         }
-
+        
         /// <summary>
-        /// The number of requests currently on the server.
-        /// Cannot exceeed the sum of <see cref="RequestThrottlingOptions.RequestQueueLimit"> and </see>/><see cref="RequestThrottlingOptions.MaxConcurrentRequests"/>.
+        /// The total number of requests (queued and active) within the server
         /// </summary>
-        public int ActiveRequestCount
+        public int TotalRequestCount
```
halter73: Hosting already know the TotalRequestCount. Counting the request currently sitting in the queue or the counting the requests currently executing is more interesting.

src/Middleware/RequestThrottling/test/MiddlewareTests.cs
```diff
@@ -96,52 +71,39 @@ public async void RequestsBlockedIfQueueFull()
...
+            Assert.Equal(1, middleware.TotalRequestCount);
 
             _ = middleware.Invoke(new DefaultHttpContext());
-            Assert.Equal(2, middleware.ActiveRequestCount);
+            Assert.Equal(2, middleware.TotalRequestCount);
         }
 
         [Fact]
-        public async void FullQueueInvokesOnRejected()
+        public async Task CleanupHappensEvenIfNextErrors()
```
halter73: I'm liking the more generic names. This is the new version of SemaphoreStatePreservedIfRequestsError, right?

src/Middleware/RequestThrottling/test/RequestQueueTests.cs
```diff
@@ -2,62 +2,50 @@
...
 
 using System.Threading.Tasks;
-using Microsoft.AspNetCore.RequestThrottling.Internal;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Internal;
 using Xunit;
 
 namespace Microsoft.AspNetCore.RequestThrottling.Tests
 {
     public class RequestQueueTests
```
halter73: Should this be renamed to TailDropQueueTests?

src/Middleware/RequestThrottling/test/TestUtils.cs
```diff
@@ -1,53 +1,96 @@
...
+                MaxConcurrentRequests = maxConcurrentRequests,
+                RequestQueueLimit = requestQueueLimit
+            });
 
-            return BuildFromOptions(options, onRejected, next);
+            return new TailDrop(options);
         }
+    }
+
+    public class TestStrategy : IQueuePolicy
```
halter73: @anurse How do we feel about moq?

anurse: I prefer not using it unless absolutely necessary :)

anurse: This is a edgey case though. I prefer not using Moq because I prefer building "real fake" implementations of things (see `TestConnection` in SignalR, etc.) rather than just making enough things work so that the method doesn't throw. Here, the test strategy is basically just a way of pairing up two delegates into an `IQueuePolicy` so it's not much beyond Moq.

Unless we're using it elsewhere in this project, it still doesn't seem like this warrants adding a dependency.

src/Middleware/RequestThrottling/perf/Microbenchmarks/QueueEmptyOverhead.cs
DylanDmitri: Technically the queue length does not matter in this benchmark, since each request is awaited to completion before the next request is allowed to arrive. I will change the values to `maxConcurrentRequests: 1, requestsQueueLimit: 0` to better reflect the test behavior.

src/Middleware/RequestThrottling/src/IQueuePolicy.cs
DylanDmitri: After implementing some queueing policies, I will better understand what features the middleware needs, and can always go back to add them.

Having the middleware enforce the concurrent limit separately is a good separation of responsibility. However, the semaphore currently both enforces the limit and provides the queue.  Separating the responsibility would mean turning one semaphore into two, for the same end result.

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
DylanDmitri: Yeah I mostly choose `TotalRequests` because its convenient in tests. Switching to a `QueuedRequests` would be more useful, and then we could subtract from the Hosting value to get the active requests.

src/Middleware/RequestThrottling/test/MiddlewareTests.cs
DylanDmitri: yes

src/Middleware/RequestThrottling/src/IQueuePolicy.cs
halter73: Yeah. This is perfectly fine for now. I just wanted more eyes on this in case there are some other good ideas we haven't considered yet.

halter73 (approved):
Pending feedback ofc

PR #11190, opened 2019-06-13
JSON error handling

### LINKED ISSUE #8474: System.Text.Json based formatters - error handling
Follow up to #7256, we need to do better error handling for parse and Required fields when deserializing. This would likely require features in the serializer to support our use case.

ryanbrandenburg:
Fixes https://github.com/aspnet/AspNetCore/issues/8474.

@pranavkm to ensure I've covered all the scenarios he had in mind.

### DIFF
--- a/src/Mvc/Mvc.Core/src/Formatters/SystemTextJsonInputFormatter.cs
+++ b/src/Mvc/Mvc.Core/src/Formatters/SystemTextJsonInputFormatter.cs
@@ -8,6 +8,8 @@ using System.Text.Json;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc.Formatters.Json;
+using Microsoft.AspNetCore.Mvc.ModelBinding;
+using Microsoft.Extensions.Logging;
 
 namespace Microsoft.AspNetCore.Mvc.Formatters
 {
@@ -16,13 +18,19 @@ namespace Microsoft.AspNetCore.Mvc.Formatters
     /// </summary>
     public class SystemTextJsonInputFormatter : TextInputFormatter, IInputFormatterExceptionPolicy
     {
+        private readonly ILogger<SystemTextJsonInputFormatter> _logger;
+
         /// <summary>
         /// Initializes a new instance of <see cref="SystemTextJsonInputFormatter"/>.
         /// </summary>
         /// <param name="options">The <see cref="JsonOptions"/>.</param>
-        public SystemTextJsonInputFormatter(JsonOptions options)
+        /// <param name="logger">The <see cref="ILogger"/>.</param>
+        public SystemTextJsonInputFormatter(
+            JsonOptions options,
+            ILogger<SystemTextJsonInputFormatter> logger)
         {
             SerializerOptions = options.JsonSerializerOptions;
+            _logger = logger;
 
             SupportedEncodings.Add(UTF8EncodingWithoutBOM);
             SupportedEncodings.Add(UTF16EncodingLittleEndian);
@@ -67,6 +75,26 @@ namespace Microsoft.AspNetCore.Mvc.Formatters
             {
                 model = await JsonSerializer.ReadAsync(inputStream, context.ModelType, SerializerOptions);
             }
+            catch (JsonException jsonException)
+            {
+                var path = jsonException.Path;
+                if (path.StartsWith("$.", StringComparison.Ordinal))
+                {
+                    path = path.Substring(2);
+                }
+
+                // Handle path combinations such as ""+"Property", "Parent"+"Property", or "Parent"+"[12]".
+                var key = ModelNames.CreatePropertyModelName(context.ModelName, path);
+
+                var formatterException = new InputFormatterException(jsonException.Message, jsonException);
+
+                var metadata = GetPathMetadata(context.Metadata, path);
+                context.ModelState.TryAddModelError(key, formatterException, metadata);
+
+                Log.JsonInputException(_logger, jsonException);
+
+                return InputFormatterResult.Failure();
+            }
             finally
             {
                 if (inputStream is TranscodingReadStream transcoding)
@@ -97,6 +125,69 @@ namespace Microsoft.AspNetCore.Mvc.Formatters
             }
 
             return new TranscodingReadStream(httpContext.Request.Body, encoding);
+        }
+
+        // Keep in sync with NewtonsoftJsonInputFormatter.GetPatMetadata
+        private ModelMetadata GetPathMetadata(ModelMetadata metadata, string path)
+        {
+            var index = 0;
+            while (index >= 0 && index < path.Length)
+            {
+                if (path[index] == '[')
+                {
+                    // At start of "[0]".
+                    if (metadata.ElementMetadata == null)
+                    {
+                        // Odd case but don't throw just because ErrorContext had an odd-looking path.
+                        break;
+                    }
+
+                    metadata = metadata.ElementMetadata;
+                    index = path.IndexOf(']', index);
+                }
+                else if (path[index] == '.' || path[index] == ']')
+                {
+                    // Skip '.' in "prefix.property" or "[0].property" or ']' in "[0]".
+                    index++;
+                }
+                else
+                {
+                    // At start of "property", "property." or "property[0]".
+                    var endIndex = path.IndexOfAny(new[] { '.', '[' }, index);
+                    if (endIndex == -1)
+                    {
+                        endIndex = path.Length;
+                    }
+
+                    var propertyName = path.Substring(index, endIndex - index);
+                    if (metadata.Properties[propertyName] == null)
+                    {
+                        // Odd case but don't throw just because ErrorContext had an odd-looking path.
+                        break;
+                    }
+
+                    metadata = metadata.Properties[propertyName];
+                    index = endIndex;
+                }
+            }
+
+            return metadata;
+        }
+
+        private static class Log
+        {
+            private static readonly Action<ILogger, Exception> _jsonInputFormatterException;
+
+            static Log()
+            {
+                _jsonInputFormatterException = LoggerMessage.Define(
+                   LogLevel.Debug,
+                   new EventId(1, "SystemTextJsonInputException"),
+                   "JSON input formatter threw an exception.");
+            }
+
+            public static void JsonInputException(ILogger logger, Exception exception) 
+                => _jsonInputFormatterException(logger, exception);
         }
     }
 }
--- a/src/Mvc/Mvc.Core/src/Formatters/SystemTextJsonInputFormatter.cs
+++ b/src/Mvc/Mvc.Core/src/Formatters/SystemTextJsonInputFormatter.cs
@@ -83,6 +83,11 @@ namespace Microsoft.AspNetCore.Mvc.Formatters
                     path = path.Substring(2);
                 }
 
+                if (path.StartsWith("$[", StringComparison.Ordinal))
+                {
+                    path = path.Substring(1);
+                }
+
                 // Handle path combinations such as ""+"Property", "Parent"+"Property", or "Parent"+"[12]".
                 var key = ModelNames.CreatePropertyModelName(context.ModelName, path);
 
--- a/src/Mvc/Mvc.Core/src/Formatters/SystemTextJsonInputFormatter.cs
+++ b/src/Mvc/Mvc.Core/src/Formatters/SystemTextJsonInputFormatter.cs
@@ -132,7 +132,7 @@ namespace Microsoft.AspNetCore.Mvc.Formatters
             return new TranscodingReadStream(httpContext.Request.Body, encoding);
         }
 
-        // Keep in sync with NewtonsoftJsonInputFormatter.GetPatMetadata
+        // Keep in sync with NewtonsoftJsonInputFormatter.GetPathMetadata
         private ModelMetadata GetPathMetadata(ModelMetadata metadata, string path)
         {
             var index = 0;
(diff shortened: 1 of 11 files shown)


src/Mvc/Mvc.Core/src/Formatters/SystemTextJsonInputFormatter.cs
```diff
@@ -98,5 +131,68 @@ private Stream GetInputStream(HttpContext httpContext, Encoding encoding)
 
             return new TranscodingReadStream(httpContext.Request.Body, encoding);
         }
+
+        // Keep in sync with NewtonsoftJsonInputFormatter.GetPatMetadata
```
Eilon: Tpyo in GetPathMetadata in comment

src/Mvc/Mvc.Core/src/Formatters/SystemTextJsonInputFormatter.cs
```diff
@@ -98,5 +131,68 @@ private Stream GetInputStream(HttpContext httpContext, Encoding encoding)
 
             return new TranscodingReadStream(httpContext.Request.Body, encoding);
         }
+
+        // Keep in sync with NewtonsoftJsonInputFormatter.GetPathMetadata
+        private ModelMetadata GetPathMetadata(ModelMetadata metadata, string path)
```
rynowak: This is a fair-bit of detailed code to duplicate. Does it make sense to extra this to a separate shared file, or add as an internal API to `ModelStateDictionary` or `ModelNames`?

src/Mvc/Mvc.Core/src/Formatters/SystemTextJsonInputFormatter.cs
```diff
@@ -98,5 +131,68 @@ private Stream GetInputStream(HttpContext httpContext, Encoding encoding)
...
+        private static class Log
+        {
+            private static readonly Action<ILogger, Exception> _jsonInputFormatterException;
+
+            static Log()
+            {
+                _jsonInputFormatterException = LoggerMessage.Define(
+                   LogLevel.Debug,
+                   new EventId(1, "SystemTextJsonInputException"),
+                   "JSON input formatter threw an exception.");
```
rynowak: Should we log more details here? What type of object failed to deserialize? Or the actual error message?

Also, should we log success as well?

src/Mvc/Mvc.Core/src/Formatters/SystemTextJsonInputFormatter.cs
pranavkm: `JsonPathToModelMetadataHelper` 🙈

src/Mvc/Mvc.Core/src/Formatters/SystemTextJsonInputFormatter.cs
pranavkm: This is pretty much what we do with Json.NET. If we were to log a success message, a better place would be in `BodyModelBinder`: https://github.com/aspnet/AspNetCore/blob/master/src/Mvc/Mvc.Core/src/ModelBinding/Binders/BodyModelBinder.cs#L157-L183

src/Mvc/Mvc.Core/test/Formatters/JsonInputFormatterTestBase.cs
```diff
@@ -227,19 +249,18 @@ public virtual async Task ReadAsync_InvalidArray_AddsOverflowErrorsToModelState(
             // Arrange
             var formatter = GetInputFormatter();
 
-            var content = "[0, 23, 300]";
```
pranavkm: Guessing System.Text doesn't like byte arrays?

src/Mvc/Mvc.Core/test/Formatters/JsonInputFormatterTestBase.cs
```diff
@@ -227,19 +249,18 @@ public virtual async Task ReadAsync_InvalidArray_AddsOverflowErrorsToModelState(
...
+            var formatterContext = CreateInputFormatterContext(typeof(short[]), httpContext);
 
             // Act
             var result = await formatter.ReadAsync(formatterContext);
 
             // Assert
-            Assert.True(result.HasError);
-            Assert.Equal("The supplied value is invalid.", formatterContext.ModelState["[2]"].Errors[0].ErrorMessage);
-            Assert.Null(formatterContext.ModelState["[2]"].Errors[0].Exception);
+            Assert.True(result.HasError, "Model should have produced an error!");
```
pranavkm: 👏

src/Mvc/Mvc.Core/test/Formatters/SystemTextJsonInputFormatterTest.cs
```diff
@@ -1,40 +1,80 @@
...
 using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
 using Xunit;
 
 namespace Microsoft.AspNetCore.Mvc.Formatters
 {
     public class SystemTextJsonInputFormatterTest : JsonInputFormatterTestBase
     {
-        [Fact(Skip = "https://github.com/aspnet/AspNetCore/issues/8474")]
+        [Fact]
```
pranavkm: While it's generally not useful to have these methods here, it's way more useful to be able to have overriden methods for debugging. In the absence of this you have to use Test Explorer to be able debug. I'm ok with leaving this in.

src/Mvc/Mvc.Core/test/Formatters/SystemTextJsonInputFormatterTest.cs
```diff
@@ -1,40 +1,80 @@
...
         }
 
-        [Fact(Skip = "https://github.com/aspnet/AspNetCore/issues/8474")]
+        [Fact]
         public override Task ReadAsync_UsesTryAddModelValidationErrorsToModelState()
         {
             return base.ReadAsync_UsesTryAddModelValidationErrorsToModelState();
         }
 
+        [Fact(Skip = "https://github.com/dotnet/corefx/issues/38492")]
```
pranavkm: 👍

src/Mvc/Mvc.Core/test/Formatters/SystemTextJsonInputFormatterTest.cs
```diff
@@ -1,40 +1,80 @@
...
+            var contentBytes = Encoding.UTF8.GetBytes(content);
+            var httpContext = GetHttpContext(contentBytes);
+
+            var formatterContext = CreateInputFormatterContext(typeof(List<int>), httpContext);
+
+            // Act
+            await formatter.ReadAsync(formatterContext);
+
+            // Assert
+            foreach(var modelState in formatterContext.ModelState)
```
pranavkm: Could we do 


```C#
Assert.Collection(
   modelState.OrderBy(k => k),
   kvp =>
   {
          Assert.Equal("[0]", kvp.Key);
          Assert.Empty(kvp.Value.Errors);
   },
  kvp =>
   {
          Assert.Equal("[1]", kvp.Key);
          var error = Assert.Single(kvp.Value.Errors);
         Assert.StartsWith("''' is an invalid start of a value", error.ErrorMessage);
   });
```

src/Mvc/Mvc.NewtonsoftJson/src/NewtonsoftJsonInputFormatter.cs
```diff
@@ -225,7 +225,8 @@ void ErrorHandler(object sender, Newtonsoft.Json.Serialization.ErrorEventArgs ev
                         }
                         else
                         {
-                            addMember = !path.EndsWith("." + member, StringComparison.Ordinal);
+                            addMember = !path.EndsWith($".{member}", StringComparison.Ordinal)
```
pranavkm: Interesting. Did we miss this scenario with json.net?

src/Mvc/Mvc.NewtonsoftJson/test/NewtonsoftJsonInputFormatterTest.cs
```diff
@@ -235,6 +235,45 @@ public Task ReadAsync_WithInputThatDeserializesToNull_SetsModelOnlyIfAllowingEmp
...
+            var contentBytes = Encoding.UTF8.GetBytes(content);
+            var httpContext = GetHttpContext(contentBytes);
+
+            var formatterContext = CreateInputFormatterContext(typeof(List<int>), httpContext);
+
+            // Act
+            var result = await formatter.ReadAsync(formatterContext);
+
+            // Assert
+            foreach (var modelState in formatterContext.ModelState)
```
pranavkm: Could we use `Assert.Collection`?

src/Mvc/Mvc.NewtonsoftJson/test/NewtonsoftJsonInputFormatterTest.cs
```diff
@@ -235,6 +235,45 @@ public Task ReadAsync_WithInputThatDeserializesToNull_SetsModelOnlyIfAllowingEmp
...
+        [InlineData("[5, 'seven', 3, 'notnum']", 2)]
+        public async Task ReadAsync_AllowMultipleErrors(string content, int failCount)
+        {
+            // Arrange
+            var failTotal = 0;
+            var serializerSettings = new JsonSerializerSettings
+            {
+                Error = delegate (object sender, ErrorEventArgs args)
+                {
+                    args.ErrorContext.Handled = true;
```
pranavkm: I don't think you need to do this. The formatter should already be doing this on your behalf.

src/Mvc/Mvc.Core/src/Formatters/SystemTextJsonInputFormatter.cs
```diff
@@ -98,5 +131,68 @@ private Stream GetInputStream(HttpContext httpContext, Encoding encoding)
...
```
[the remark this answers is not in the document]
Eilon: No. It's either `JSON input formatter` or `JsonInputFormatter`. But never `Json input formatter` (JSON is not a regular word, it only gets lower-cased in identifiers.)

src/Mvc/Mvc.Core/src/Formatters/SystemTextJsonInputFormatter.cs
```diff
@@ -98,5 +131,68 @@ private Stream GetInputStream(HttpContext httpContext, Encoding encoding)
...
             return new TranscodingReadStream(httpContext.Request.Body, encoding);
         }
+
+        // Keep in sync with NewtonsoftJsonInputFormatter.GetPathMetadata
+        private ModelMetadata GetPathMetadata(ModelMetadata metadata, string path)
+        {
+            var index = 0;
+            while (index >= 0 && index < path.Length)
+            {
+                if (path[index] == '[')
```
JamesNK: What about escaped strings in a path? e.g.

`$.DictionaryContainer.['it\'s a lot of work to handle all situations'].Name`

This can happen when the path has a dictionary key in it and the dictionary key requires escaping.

JamesNK: Logic in serializer: https://github.com/dotnet/corefx/blob/94c5f1384c9b6477d164050de964cf6dcf99b2a9/src/System.Text.Json/src/System/Text/Json/Serialization/ReadStack.cs#L105-L123

JamesNK: Hmm, I think you're going to need to duplicate logic for reading an escaped JSON string. Maybe you could reuse Utf8JsonReader but you'd need to parse the string anyway to find where it is closed so you had a slice of text to pass to it.

src/Mvc/Mvc.NewtonsoftJson/src/NewtonsoftJsonInputFormatter.cs
JamesNK: The JSON path logic in S.T.J is slightly different from Json.NET

JamesNK: btw I know this isn't a hot path but string interpolation instead of concatenating two strings together is so wasteful.

src/Mvc/Mvc.Core/src/Formatters/SystemTextJsonInputFormatter.cs
rynowak: I'm not making the argument that we have good logging elsewhere, I'm making the argument that we should have good logging everywhere.

src/Mvc/Mvc.Core/src/Formatters/SystemTextJsonInputFormatter.cs
rynowak: I'm just here so I won't get fined... and to say that the existing behaviour (trying to map JSON path to specific .NET properties) isn't useful for anyone.

src/Mvc/Mvc.Core/test/Formatters/JsonInputFormatterTestBase.cs
ryanbrandenburg: Yeah. Further [context here](https://github.com/dotnet/corefx/issues/38494#issuecomment-501571451).

src/Mvc/Mvc.Core/test/Formatters/SystemTextJsonInputFormatterTest.cs
ryanbrandenburg: Yes, and we should!

src/Mvc/Mvc.Core/src/Formatters/SystemTextJsonInputFormatter.cs
pranavkm: Ok. I'm ok with us starting off without the mapping behavior for System.Text.Json. It's different from Json.NET, and we can consider putting that behind a flag.

Removing this would mean there's less of this weird code around and the ModelState \ `ValidationProblemDetails` that's returned has JSONPath in it. That sound reasonable?

PR #11191, opened 2019-06-13
[TS] Remove extra negotiates from common scenarios

### LINKED ISSUE #11160: [SignalR] Avoid over-negotiating in client
There are 2 cases where we might over-negotiate in the Typescript client (should audit the other clients too).

1. Client specifies a transport and server sends a larger list of supported transports, the client will still evaluate transports higher in the list if there are any, and cause a re-negotiate to happen
1. Client doesn't support WebSockets or EventSource and we evaluate that too late and cause a re-negotiate

BrennanConroy:
Fixes https://github.com/aspnet/AspNetCore/issues/11160


src/SignalR/clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -401,32 +404,32 @@ export class HttpConnection implements IConnection {
         }
     }
 
-    private resolveTransport(endpoint: IAvailableTransport, requestedTransport: HttpTransportType | undefined, requestedTransferFormat: TransferFormat): HttpTransportType | null {
+    private resolveTransport(endpoint: IAvailableTransport, requestedTransport: HttpTransportType | undefined, requestedTransferFormat: TransferFormat): HttpTransportType | Error {
```
halter73: I see C/C++ has poisoned your mind to the point that you don't like throwing exceptions. I'm OK with this since this is fairly normal control flow.

I would like renaming this to tryResloveTransport, resolveTransportOrError or something like that. At least TS should prevent the caller from just assuming resolveTransport can only return a HttpTransportType.

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -353,15 +353,18 @@ export class HttpConnection implements IConnection {
         const transports = negotiateResponse.availableTransports || [];
         for (const endpoint of transports) {
             try {
-                const transport = this.resolveTransport(endpoint, requestedTransport, requestedTransferFormat);
-                if (typeof transport === "number") {
-                    this.transport = this.constructTransport(transport);
+                const transportOrError = this.resolveTransport(endpoint, requestedTransport, requestedTransferFormat);
+                if (typeof transportOrError === "number") {
+                    this.transport = this.constructTransport(transportOrError);
                     if (!negotiateResponse.connectionId) {
```
halter73: I see, so we always want to renegotiate for transport!.connect failures, because we don't want to risk reusing the same connectionid. But in the case that the client doesn't support the transport, there's no reason not to use the connectionid we already have, because we have tried using it yet.

My question is if getNegotiationResponse itself fails, not the transport, will the client move on to the next transport in the list and try renegotiating? That seems super broken. If any negotiate request fails, start should probably just fail. I'm fine leaving this PR as is and filing a separate issue though.

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -353,15 +353,18 @@ export class HttpConnection implements IConnection {
         const transports = negotiateResponse.availableTransports || [];
         for (const endpoint of transports) {
             try {
-                const transport = this.resolveTransport(endpoint, requestedTransport, requestedTransferFormat);
-                if (typeof transport === "number") {
-                    this.transport = this.constructTransport(transport);
+                const transportOrError = this.resolveTransport(endpoint, requestedTransport, requestedTransferFormat);
```
halter73: Should this be moved outside the try? It being the first statement in the try makes me think it can throw. If it cannot throw, moving it outside of the try makes the code easier to read/modify/review.

halter73 (approved):

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
BrennanConroy: Good observation, we should probably jump out of the loop in that case

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
mikaelm12: +1  to something like `resolveTransportOrError `
It would feel weird to have a "try"  prefixed method not return a bool.
Imo you should always be able to do 

```
if (tryFoo()) {
    // do something if Foo succeeded
}
else { 
    // do something if foo failed
}
```

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -401,32 +409,32 @@ export class HttpConnection implements IConnection {
         }
     }
 
-    private resolveTransport(endpoint: IAvailableTransport, requestedTransport: HttpTransportType | undefined, requestedTransferFormat: TransferFormat): HttpTransportType | null {
+    private tryResolveTransport(endpoint: IAvailableTransport, requestedTransport: HttpTransportType | undefined, requestedTransferFormat: TransferFormat): HttpTransportType | Error {
```
mikaelm12: It feels weird to have a "try"  prefixed method not return a bool.
Imo you should always be able to do 

```
if (tryFoo()) {
    // do something if Foo succeeded
}
else { 
    // do something if foo failed
}
```

BrennanConroy:
@pranavkm I can't get the "correct" blazor.server.js file built locally...

pranavkm:
@BrennanConroy seems like it worked?

BrennanConroy:
Yeah, had to try it on my Mac.

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
```diff
@@ -352,12 +352,20 @@ export class HttpConnection implements IConnection {
...
                     if (!negotiateResponse.connectionId) {
-                        negotiateResponse = await this.getNegotiationResponse(url);
+                        try {
+                            negotiateResponse = await this.getNegotiationResponse(url);
+                        } catch (ex) {
+                            return Promise.reject(ex);
+                        }
                         connectUrl = this.createConnectUrl(url, negotiateResponse.connectionId);
                     }
                     await this.transport!.connect(connectUrl, requestedTransferFormat);
```
halter73: Can anything other than this line throw? If not, I'd like to keep the try/catch local to the transport.connect call.

halter73 (approved):
Looks good. I'd like to make the try block smaller if possible though.

src/SignalR/clients/ts/signalr/src/HttpConnection.ts
BrennanConroy: `constructTransport` can throw if we can't successfully create the requested transport.

I can move the try/catch inside the if check though

halter73: I think that makes sense. We probably also don't need to null out negotiateResponse.connectionId and renegotiate when constructTransport throws.

BrennanConroy:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=227442

BrennanConroy:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=227443&view=logs

aspnet-hello:

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/2577

Issue #10886, opened 2019-06-05
Customizing HTTP/2 abort response


JamesNK:
In gRPC there is the concept of a deadline. The deadline is sent by the client to the server with a gRPC call in a header. It is the client saying to the server "you have X seconds to finish this call, after-which give up".

* If the server finished the call in less than X sends then the response completes as usual.
* If the server exceeds the X seconds then it immediately finishes the HTTP/2 response, sending a RST_STREAM. The user's gRPC method code will likely still be executing after this. If they are good then it will be using a cancellation token, and they stop whatever they're doing quickly. But they might not, and continue doing stuff after the response is complete. This is fine. It is also fine if an error eventually gets thrown because they tried to do an HTTP thing.

Today the gRPC server is finishing the connection when the deadline is hit by calling `HttpContext.Abort()`. This immediately sends RST_STREAM with an INTERNAL_ERROR error code. The request delegate with the user's code is still running. Using `Abort` is not ideal, but its response seems to be acceptable to the various gRPC clients of the world.

Today that response on Abort looks very hard-coded in Http2Stream:

```cs
protected override void ApplicationAbort()
{
    var abortReason = new ConnectionAbortedException(CoreStrings.ConnectionAbortedByApplication);
    ResetAndAbort(abortReason, Http2ErrorCode.INTERNAL_ERROR);
}
```

I think the current abort behavior is fine for gRPC to use with deadline, the one thing that could be improved is the abort response over HTTP/2. We're sending INTERNAL_ERROR, and no status trailers. Another gRPC implementation I looked at immediately ends the response, but it sends RST_STREAM (NO_ERROR), along with some HTTP/2 trailing headers.

It would be nice to have a hook in ASP.NET Core to modify the response when an HTTP/2 abort occurs, such as changing the `Http2ErrorCode` used with RST_STREAM, or appending trailers. 

Thoughts around the area?
Or alternatives?

Tratcher:
> it sends RST_STREAM (NO_ERROR), along with some HTTP/2 trailing headers.

How does it do that? It should only work if they send the trailers first as trailers after RST would be invalid.

> It would be nice to have a hook in ASP.NET Core to modify the response when an HTTP/2 abort occurs, such as changing the Http2ErrorCode used with RST_STREAM, or appending trailers.

I could image having an HTTP Feature and response extension for providing a Reset method with a custom error code (and another one for GoAway). Also including trailers would be complicated.

[...]

PR #11193, opened 2019-06-13
Implement Http/2 CompleteAsync

Tratcher:
#10886 Adding IHttpResponseCompletionFeature.CompleteAsync for HTTP/2 only.

- Allows for the graceful completion of a response message (from the client's perspective) without waiting for the request delegate to unwind.
- It makes sure all remaining response headers, data, and trailers are sent before it returns.
- It may throw if the response is in an invalid state such as a Content-Length mismatch.
- RequestAborted will not fire after this completes. Abort can still be used though.

Retrofitting this for HTTP/1.1 shouldn't be hard but I wanted to unblock the gRPC scenarios first.

While testing this I fixed a bug in response trailers where they were not getting cleared if the app threw an exception and the server converted it to a 500 response. You'd end up with a 500 + trailers.

### DIFF
--- a/src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
+++ b/src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
@@ -210,6 +210,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
         public bool RequestTrailersAvailable { get; set; }
         public Stream RequestBody { get; set; }
         public PipeReader RequestBodyPipeReader { get; set; }
+        public HttpResponseTrailers ResponseTrailers { get; set; }
 
         private int _statusCode;
         public int StatusCode
@@ -287,7 +288,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 
         public bool HasResponseStarted => _requestProcessingStatus >= RequestProcessingStatus.HeadersCommitted;
 
-        public bool HasFlushedHeaders => _requestProcessingStatus == RequestProcessingStatus.HeadersFlushed;
+        public bool HasFlushedHeaders => _requestProcessingStatus >= RequestProcessingStatus.HeadersFlushed;
+
+        public bool HasResponseCompleted => _requestProcessingStatus == RequestProcessingStatus.ResponseCompleted;
 
         protected HttpRequestHeaders HttpRequestHeaders { get; }
 
@@ -632,9 +635,9 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                     // Run the application code for this request
                     await application.ProcessRequestAsync(context);
 
-                    if (!_connectionAborted)
+                    if (!_connectionAborted && !VerifyResponseContentLength(out var lengthException))
                     {
-                        VerifyResponseContentLength();
+                        ReportApplicationError(lengthException);
                     }
                 }
                 catch (BadHttpRequestException ex)
@@ -898,7 +901,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             }
         }
 
-        protected void VerifyResponseContentLength()
+        protected bool VerifyResponseContentLength(out Exception ex)
         {
             var responseHeaders = HttpResponseHeaders;
 
@@ -915,9 +918,13 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
                     _keepAlive = false;
                 }
 
-                ReportApplicationError(new InvalidOperationException(
-                    CoreStrings.FormatTooFewBytesWritten(_responseBytesWritten, responseHeaders.ContentLength.Value)));
+                ex = new InvalidOperationException(
+                    CoreStrings.FormatTooFewBytesWritten(_responseBytesWritten, responseHeaders.ContentLength.Value));
+                return false;
             }
+
+            ex = null;
+            return true;
         }
 
         public void ProduceContinue()
@@ -935,7 +942,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
             }
         }
 
-        public Task InitializeResponseAsync(int firstWriteByteCount)
+        public Task InitializeResponseAsync(int firstWriteByteCount, bool appCompleted = false)
         {
             var startingTask = FireOnStarting();
             // If return is Task.CompletedTask no awaiting is required
@@ -946,7 +953,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 
             VerifyInitializeState(firstWriteByteCount);
 
-            ProduceStart(appCompleted: false);
+            ProduceStart(appCompleted: appCompleted);
 
             return Task.CompletedTask;
         }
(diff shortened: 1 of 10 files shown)


src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
```diff
@@ -915,9 +918,13 @@ protected void VerifyResponseContentLength()
...
 
-                ReportApplicationError(new InvalidOperationException(
-                    CoreStrings.FormatTooFewBytesWritten(_responseBytesWritten, responseHeaders.ContentLength.Value)));
+                ex = new InvalidOperationException(
+                    CoreStrings.FormatTooFewBytesWritten(_responseBytesWritten, responseHeaders.ContentLength.Value));
+                return false;
             }
+
+            ex = null;
+            return true;
```
halter73: You could just return the Exception normally. Null could mean false.

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
```diff
@@ -946,7 +953,7 @@ public Task InitializeResponseAsync(int firstWriteByteCount)
 
             VerifyInitializeState(firstWriteByteCount);
 
-            ProduceStart(appCompleted: false);
+            ProduceStart(appCompleted: appCompleted);
```
halter73: Nit:

```suggestion
            ProduceStart(appCompleted);
```

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
```diff
@@ -1064,7 +1076,7 @@ private Task WriteSuffix()
 
             if (!HasFlushedHeaders)
             {
-                _requestProcessingStatus = RequestProcessingStatus.HeadersFlushed;
+                _requestProcessingStatus = RequestProcessingStatus.ResponseCompleted;
```
halter73: Did this incorrectly setting just HeadersFlushed instead of ResponseCompleted result in any bad behavior prior to this change?

halter73: Never mind. Now I see this state didn't exist before. That makes more sense.

src/Servers/Kestrel/test/InMemory.FunctionalTests/Http2/Http2StreamTests.cs
```diff
@@ -1839,6 +1839,32 @@ public async Task ResponseTrailers_WithoutData_Sent()
             Assert.Equal("Custom Value", _decodedHeaders["CustomName"]);
         }
 
+        [Fact]
+        public async Task ResponseTrailers_WithExeption500_Cleared()
+        {
+            await InitializeConnectionAsync(context =>
+            {
+                context.Response.AppendTrailer("CustomName", "Custom Value");
```
halter73: Do we have any tests where the App attempts to call AppendTrailer after CompleteAsync. That should throw an InvalidOperationException due to being read-only at that point, right?

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
Tratcher: ResponseCompleted is new. This flush operation flushes and completes so I had it skip the flush state.

src/Servers/Kestrel/test/InMemory.FunctionalTests/Http2/Http2StreamTests.cs
halter73: I would test both code paths. One where a trailer was added before CompleteAsync, and another where no trailers were written so the empty read-only trailer header collection is lazy-initialized.

Tratcher: These tests check IsReadOnly on the trailers collection. I could also try adding in one of them.

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
Tratcher: The try works better for the callers.

src/Servers/Kestrel/Core/src/Internal/Http2/Http2Stream.FeatureCollection.cs
```diff
@@ -48,5 +53,30 @@ MinDataRate IHttpMinRequestBodyDataRateFeature.MinDataRate
                 MinRequestBodyDataRate = value;
             }
         }
+
+        async Task IHttpResponseCompletionFeature.CompleteAsync()
+        {
```
halter73: What would happen if this method was just verifying the content-length and call ProduceEnd()? It should initialize the response for you if the headers haven't been committed yet and will return the task from WriteSuffix(). This might help get rid of an unnecessary asyncstatemachine.

src/Servers/Kestrel/test/InMemory.FunctionalTests/Http2/Http2StreamTests.cs
halter73: It's fine either way. I didn't see those tests.

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
```diff
@@ -632,9 +635,18 @@ private async Task ProcessRequests<TContext>(IHttpApplication<TContext> applicat
...
 
-                    if (!_connectionAborted)
+                    // Trigger OnStarting if it hasn't been called yet and the app hasn't
+                    // already failed. If an OnStarting callback throws we can go through
+                    // our normal error handling in ProduceEnd.
+                    // https://github.com/aspnet/KestrelHttpServer/issues/43
+                    if (!HasResponseStarted && _applicationException == null && _onStarting?.Count > 0)
                     {
-                        VerifyResponseContentLength();
+                        await FireOnStarting();
```
Tratcher: FireOnStarting was moved to prevent users from modifying ContentLength after the check.

src/Servers/Kestrel/test/InMemory.FunctionalTests/Http2/Http2StreamTests.cs
```diff
@@ -3307,5 +3333,698 @@ public async Task ResponseBodyPipeCompleteWithExceptionThrows()
...
+            _hpackDecoder.Decode(headersFrame.PayloadSequence, endHeaders: false, handler: this);
+
+            Assert.Equal(3, _decodedHeaders.Count);
+            Assert.Contains("date", _decodedHeaders.Keys, StringComparer.OrdinalIgnoreCase);
+            Assert.Equal("500", _decodedHeaders[HeaderNames.Status]);
+            Assert.Equal("0", _decodedHeaders[HeaderNames.ContentLength]);
+        }
+
+        [Fact]
+        public async Task CompleteAsync_AfterBodyStarted_SendsBodyWithEndStream()
```
JamesNK: I believe you can write to the body pipe without starting the response. Should that be tested?

halter73: Are you suggesting we test what happens when we call ResponseBodyPipe.GetMemory(), then ResponseBodyPipe.Advance(), then CompleteAsync(), and then finally ResponseBodyPipe.FlushAsync()?

I expect FlushAsync() would throw and InvalidOperationException in that case. That seems like a reasonable test.

halter73 (approved):

src/Servers/Kestrel/test/InMemory.FunctionalTests/Http2/Http2StreamTests.cs
JamesNK: I was thinking:

1. GetMemory
2. Advance
3. CompleteAsync
4. Assert that written data is sent (CompleteAsync would match FlushAsync in this regard)

FlushAsync erroring after CompleteAsync is probably already tested.

jkotalik: Agree with @JamesNK. That code is spicy and deserves a test.

jkotalik (approved):

JamesNK (approved):

PR #11198, opened 2019-06-14
Always include ANCM in build output

jkotalik:
Note, this is purely a build only change. I'm currently working on verifying this works with separate branches on aspnetci.

## Impact
After we stopped pushing to MyGet, ANCM transfer packages have stopped being uploaded. Builds are now failing with an error saying ANCM could not be found as an artifact dependency as well as the latest version of ANCM being 2.2.2 on myget. This change makes it so ANCM always builds, regardless of if there was a change to ANCM or not.

The reason this is fine is:
- ANCM is shipped globally and isn't put in a package.
- We are already doing this in 3.0.

## Risk
Very low risk. We are rebuilding ANCM every patch version, which will update dll/msi versions. However, the source code being used is the same.

### DIFF
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -11,6 +11,7 @@
     <!-- Always include framework metapackages in patch updates. -->
     <IsPackageInThisPatch Condition="'$(IsFrameworkMetapackage)' == 'true' OR '$(IsSharedSourcePackage)' == 'true' ">true</IsPackageInThisPatch>
     <IsPackageInThisPatch Condition="'$(IsPackageInThisPatch)' == ''">$(PackagesInPatch.Contains(' $(PackageId);'))</IsPackageInThisPatch>
+    <IsPackageInThisPatch Condition="'$(IsANCMPackage)' == 'true'">true</IsPackageInThisPatch>
   </PropertyGroup>
 
   <PropertyGroup Condition=" '$(IsPackable)' != 'false' AND '$(IsServicingBuild)' == 'true' ">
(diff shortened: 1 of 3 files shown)


Directory.Build.targets
```diff
@@ -11,6 +11,7 @@
     <!-- Always include framework metapackages in patch updates. -->
     <IsPackageInThisPatch Condition="'$(IsFrameworkMetapackage)' == 'true' OR '$(IsSharedSourcePackage)' == 'true' ">true</IsPackageInThisPatch>
     <IsPackageInThisPatch Condition="'$(IsPackageInThisPatch)' == ''">$(PackagesInPatch.Contains(' $(PackageId);'))</IsPackageInThisPatch>
+    <IsPackageInThisPatch Condition="'$(IsANCMPackage)' == 'true'">true</IsPackageInThisPatch>
```
JunTaoLuo: Add a comment explaining why this must be rebuilt for every patch.

JunTaoLuo (approved):
Yes this is a fix that we need in order to build our 2.2 patch. @anurse what kind of approval process is needed for this change?

anurse:
> Yes this is a fix that we need in order to build our 2.2 patch.

👍 . Make sure `netcoreship` is **aware** of the change, but if it's necessary to build, we gotta take it so go ahead.

jkotalik:
Verified here: http://aspnetci/viewLog.html?buildId=655457&buildTypeId=Releases_22xPublic_WindowsInstallers&tab=buildLog&_focus=481#_state=441

vivmishra:
Approved for 2.2.6

JunTaoLuo:
I'm going to merge this one since we are running into build hangs on 2.2 right now and it's unlikely that the tests will pass. The linux failure is unrelated.

PR #11199, opened 2019-06-14
Always include ANCM in build output

jkotalik:
Note, this is purely a build only change. I'm currently working on verifying this works with separate branches on aspnetci.

Same change as https://github.com/aspnet/AspNetCore/pull/11198, except just for 2.1.


JunTaoLuo (approved):
@anurse Similarly, 2.1 patching needs this fix as well.

anurse:
Ok by me. Since this is build only, please email `netcoreship` with the information (for both PRs).

If this is **necessary** to build the July patch, go ahead. Otherwise, hold until shiproom (via email) confirms it can be merged. We're trying to get the final build by end-of-week. I did see some emails going around about this though and I'm still catching up so if I'm behind a bit, I trust you to do the needful in conversations with shiproom ;P.

jkotalik:
> If this is necessary to build the July patch, go ahead. 

Yep this is necessary 😄

vivmishra:
Approved for 2.1.12

PR #11202, opened 2019-06-14
Did API diff from 2.2 to preview7 for scrutiny and research

davidfowl:
- I removed EF, identity model, roslyn, remotion linq basically any 3rd party dep
- Blazor is all new so it shows up as everything new
- There's still a bunch of pubternal API hanging around.
- I didn't remove things that were removed from the shared framework but still exist in packages. They show up as all deleted.
- This is by namespace.
- Use the markdown view to look at namespaces of interest.

This is a good entry point; https://github.com/aspnet/AspNetCore/blob/ab7f1caaf21630f500837224a3fd8a1fc53448d1/api-diffs/api-diffs.md

cc @aspnet/aspnet-push


natemcmaster:
Cool stuff! Do you think it would be worthwhile putting these diffs in the ref/* folder of the relevant assembly? Or maybe under $repoRoot/docs/3.0/?

davidfowl:
> Cool stuff! Do you think it would be worthwhile putting these diffs in the ref/* folder of the relevant assembly? Or maybe under $repoRoot/docs/3.0/?

I think we should put these where corefx puts their API preview docs. https://github.com/dotnet/core/tree/master/release-notes/3.0/preview/api-diff

They have packages and shared framework. At RTM there should be an API diff for 2.2 to 3.0.

halter73:
Unfortunately, especially for reviewing the removal of pubternal APIs, just seeing the diff is insufficient for review because you cannot see what *wasn't* removed from the namespace.

I think it would be far better to have two commits, one with the 2.2 API, and a second commit on top of it replacing the 2.2 API with the 3.0 API.

You could then push both commits to separate branches, e.g. davidfowl/api-structure-2.2 and davidfowl/api-structure-3.0-p7, and create a PR from davidfowl/api-structure-3.0-p7 based on davidfowl/api-structure-2.2.

Then reviewers could use their normal git and github tools to see a more interactive diff where I can see the other APIs in the namespace/class whatever that haven't change.

halter73:
Of course, going forward, if we put all our reference assemblies in a single forward we could just do `git diff /ref` or something.

halter73:
Also, did you mention what tool you used to generate this and what parameters you used? I must have missed it.

davidfowl:
I chose to remove what’s already there because of noise.

davidfowl:
@halter73 updated it

davidfowl:
Updated!

Issue #5769, opened 2018-12-18
UX issues related to Identity


Eilon:
_From @Eilon on October 24, 2017 21:51_

_From @SteveSandersonMS on May 9, 2016 16:32_

It's debatable to what extent any of these are "bugs", since the template is just a starting point and not meant to be a finished, polished app. But many of them can be fixed easily.
1. **Default "External Logins" UX is confusing**
   
   It says "External logins: 0 [Manage]". When you click 'Manage', it takes you to a blank page (there's just a heading - no content). What's an end user supposed to do with this? There should at least be some instructions here for the developer, like on other pages.
2. **Default 'Manage your account' layout is bad**
   
   Screenshot:
   ![screen shot 2016-05-09 at 17 20 10](https://cloud.githubusercontent.com/assets/1101362/15119713/1fffe7a0-160a-11e6-8a0d-37431e1c4c33.png)
   
   The text "Two-Factor Authentic..." is truncated for no good reason, and its counterpart in the next column isn't correctly vertically aligned with it.
3. **Change password UI is rough**
   
   The heading "Change Password." is inconsistently capitalised (on other pages like "Log in." and "Forgot your password?" we don't capitalise the first letter of each word). The subheading, "Change Password Form", is completely pointless. Suffixing headings like "Change Password." with period characters is weird and ugly in my option, though that's just a matter of personal preference.
4. **The account management page is hard to discover**
   
   To reach it, you have to click the part of the heading that says "Hi [username]!". There's no clear indication that this is how to manage your account.
5. **When using confirmation URLs, the registration flow provides no indication of what's going on**
   
   If you set up account confirmation [as per the docs](http://docs.asp.net/en/latest/security/authentication/accconfirm.html), then when the user registers an account, they just get redirected back to the home page. They are not logged in, and there's no message to say their registration succeeded. It would be nice to display some sort of message like "Please check your email and click the link we sent to you.", otherwise the user may think their registration was rejected.
6. **Login failed message is unhelpful**
   
   When you enter the wrong username or password, it says "Login attempt failed", without indicating whether it's a bad username or password. I know from reading the code that this is deliberate, but I don't know what sort of benefit it's supposed to provide. An attacker who wants to know whether a given username is registered can easily find out (just try to register a new account with it), so why withhold genuinely useful information from legit users?
7. **Account-not-confirmed error message makes little sense**
   
[...]

PR #11204, opened 2019-06-14
Ext Auth Dev Msg Remove

OzBob:
Remove developer message: "There are no external authentication services configured."

 - Removing from v3 and v4 identity page templates, the dev message when there are no External auth providers
 - Pointless to have to scaffold and override Login.cshtml when only supplying local auth to customer.

Addresses #5769


HaoK:
This is tracked by https://github.com/dotnet/aspnetcore/issues/18164 thanks for the PR but we aren't going this direction currently

PR #11211, opened 2019-06-14
Hoist activity fields to the logging scope

### LINKED ISSUE #9491: Consider tracing span id and trace id separately in the log
Seems to like open census:

https://github.com/census-instrumentation/opencensus-specs/blob/c3fd41b28fed96ceffc8c35af280d74cacea3ab2/trace/LogCorrelation.md#key-names-for-tracing-data

cc: @SergeyKanzhelev @lmolkova

davidfowl:
- Expose SpanId, TraceId and ParentId to logging scope properties
- Added tests to verify the Hierarchical ID format

Fixes #9491

PS: I need to spend some time looking at allocations before I merge this.

cc @benaadams 

cc @lmolkova @SergeyKanzhelev for review

### DIFF
--- a/src/Hosting/Hosting/src/Internal/HostingLoggerExtensions.cs
+++ b/src/Hosting/Hosting/src/Internal/HostingLoggerExtensions.cs
@@ -151,7 +151,7 @@ namespace Microsoft.AspNetCore.Hosting.Internal
                 {
                     _traceId = activity.TraceId.ToHexString();
                     _spanId = activity.SpanId.ToHexString();
-                    _parentId = activity.ParentId;
+                    _parentId = activity.ParentSpanId.ToHexString();
                 }
                 else
                 {
--- a/src/Hosting/Hosting/src/Internal/HostingLoggerExtensions.cs
+++ b/src/Hosting/Hosting/src/Internal/HostingLoggerExtensions.cs
@@ -123,15 +123,15 @@ namespace Microsoft.AspNetCore.Hosting.Internal
                     }
                     else if (index == 2)
                     {
-                        return new KeyValuePair<string, object>("SpanId", _activity.GetSpanId() ?? string.Empty);
+                        return new KeyValuePair<string, object>("SpanId", _activity.GetSpanId());
                     }
                     else if (index == 3)
                     {
-                        return new KeyValuePair<string, object>("TraceId", _activity.GetTraceId() ?? string.Empty);
+                        return new KeyValuePair<string, object>("TraceId", _activity.GetTraceId());
                     }
                     else if (index == 4)
                     {
-                        return new KeyValuePair<string, object>("ParentId", _activity.GetParentId() ?? string.Empty);
+                        return new KeyValuePair<string, object>("ParentId", _activity.GetParentId());
                     }
 
                     throw new ArgumentOutOfRangeException(nameof(index));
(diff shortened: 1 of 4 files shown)


src/Hosting/Hosting/src/Internal/HostingLoggerExtensions.cs
```diff
@@ -122,20 +125,40 @@ public int Count
...
                 _path = (httpContext.Request.PathBase.HasValue 
                          ? httpContext.Request.PathBase + httpContext.Request.Path 
                          : httpContext.Request.Path).ToString();
-                _activityId = activityId;
+
+                if (activity.IdFormat == ActivityIdFormat.W3C)
+                {
+                    _traceId = activity.TraceId.ToHexString();
+                    _spanId = activity.SpanId.ToHexString();
+                    _parentId = activity.ParentId;
```
lmolkova:
```suggestion
                    _parentId = activity.ParentSpanId;
```

davidfowl: Allocations everywhere. Will this return null if ParentSpanId is null?

lmolkova: nope :( it will be "0000000000000000". Need null check on ParentId before accessing it to avoid extra allocation. 

I guess we should fix it in `ActivitySpanId`. If `id1 == 0` return const "0000000000000000", similar for TraceId

https://github.com/dotnet/corefx/blob/master/src/System.Diagnostics.DiagnosticSource/src/System/Diagnostics/Activity.cs#L1229

benaadams:
Do you need to capture the fields to `HostingLogScope` directly? Is it safe to capture the Activity then output the values from that when needed?

davidfowl:
> Do you need to capture the fields to HostingLogScope directly? Is it safe to capture the Activity then output the values from that when needed?

I was going to but storing the async local felt strange. I'll need to experiment with that.

benaadams:
> I was going to but storing the async local felt strange. 

That's just a tangental bit of Activity? Hosting creates the Scope, Activity and starts it; as well as Disposing the Scope and ending the Activity, and the Activity doesn't get reused, so should be ok?

src/Hosting/Hosting/test/HostingApplicationTests.cs
```diff
@@ -53,7 +53,36 @@ public void CreateContextWithEnabledLoggerCreatesActivityAndSetsActivityIdInScop
...
+            Assert.Null(pairs["ParentId"]?.ToString());
+        }
+
+        [Fact]
+        public void CreateContextWithEnabledLoggerAndRequestIdCreatesActivityAndSetsActivityInScope()
+        {
+            // Arrange
+
+            // Generate an id we can use for the request id header (in the correct format)
+            var activity = new Activity("IncomingRequest");
```
davidfowl: BTW @lmolkova this why I'd like this feature https://github.com/dotnet/corefx/issues/36999. It's makes unit testing possible.

davidfowl:
> That's just a tangental bit of Activity? Hosting creates the Scope, Activity and starts it; as well as Disposing the Scope and ending the Activity, and the Activity doesn't get reused, so should be ok?

Yes, you're right, it would just be storing the activity object on 2 async locals (Activity.Current and the logging scope). It does mean it would have similar issues to holding onto the HttpContext but maybe it won't be as big a problem since not much is stored on the activity currently.

davidfowl:
🆙 📅

lmolkova (approved):

Tratcher (approved):

PR #11213, opened 2019-06-14
Add explicit content types to 'no formatter' logging

### LINKED ISSUE #11173: NoFormatter log message does not include ContentTypes from result
When MVC fails to select an output formatter based on `ObjectResult.ContentTypes` (`Response.ContentType` has not been set), it logs the following warning:

> No output formatter was found for content type '' to write the response.

This is pretty unhelpful and confusing. I suggest the warning also includes the content types considered from the `ObjectResult`.

khellang:
Closes #11173

### DIFF
--- a/src/Mvc/Mvc.Core/src/Infrastructure/DefaultOutputFormatterSelector.cs
+++ b/src/Mvc/Mvc.Core/src/Infrastructure/DefaultOutputFormatterSelector.cs
@@ -147,14 +147,11 @@ namespace Microsoft.AspNetCore.Mvc.Infrastructure
                 }
             }
 
-            if (selectedFormatter == null)
+            if (selectedFormatter != null)
             {
-                // No formatter supports this.
-                _logger.NoFormatter(context);
-                return null;
+                _logger.FormatterSelected(selectedFormatter, context);
             }
 
-            _logger.FormatterSelected(selectedFormatter, context);
             return selectedFormatter;
         }
 
--- a/src/Mvc/Mvc.Core/src/Infrastructure/ObjectResultExecutor.cs
+++ b/src/Mvc/Mvc.Core/src/Infrastructure/ObjectResultExecutor.cs
@@ -105,7 +105,7 @@ namespace Microsoft.AspNetCore.Mvc.Infrastructure
             if (selectedFormatter == null)
             {
                 // No formatter supports this.
-                Logger.NoFormatter(formatterContext);
+                Logger.NoFormatter(formatterContext, result.ContentTypes);
 
                 context.HttpContext.Response.StatusCode = StatusCodes.Status406NotAcceptable;
                 return Task.CompletedTask;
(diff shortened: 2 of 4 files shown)


src/Mvc/Mvc.Core/src/Infrastructure/DefaultOutputFormatterSelector.cs
```diff
@@ -149,8 +149,6 @@ public override IOutputFormatter SelectFormatter(OutputFormatterCanWriteContext
 
             if (selectedFormatter == null)
             {
-                // No formatter supports this.
-                _logger.NoFormatter(context);
```
khellang: This resulted in the log message being repeated twice:

```
warn: Microsoft.AspNetCore.Mvc.Infrastructure.DefaultOutputFormatterSelector[1]
      No output formatter was found for content type '' to write the response.
warn: Microsoft.AspNetCore.Mvc.Infrastructure.ObjectResultExecutor[1]
      No output formatter was found for content type '' to write the response.
```

I opted to remove so `ObjectResultExecutor` would still log it, even if `DefaultOutputFormatterSelector` has been swapped with a different service.

src/Mvc/Mvc.Core/src/Infrastructure/ObjectResultExecutor.cs
```diff
@@ -105,7 +105,7 @@ public virtual Task ExecuteAsync(ActionContext context, ObjectResult result)
             if (selectedFormatter == null)
             {
                 // No formatter supports this.
-                Logger.NoFormatter(formatterContext);
```
rynowak: So one thing to mention about this is that the list logged here won't include anything from the accept header. The logic to iterate the accept header (or not in the case of `result.ContentType`) lives inside the formatter selector. What's the logging like when you have an accept header for `application/xml` and only have the JSON formatter registered?

khellang: > So one thing to mention about this is that the list logged here won't include anything from the accept header.

That's right, but there are other (debug-level) log statements that include the accept header:

> Attempting to select an output formatter based on Accept header '{AcceptHeader}'.
> Attempting to select an output formatter without using a content type as no explicit content types were specified for the response.

or

> Attempting to select an output formatter based on Accept header '{AcceptHeader}' and explicitly specified content types '{ExplicitContentTypes}'. The content types in the accept header must be a subset of the explicitly set content types.
> Attempting to select the first output formatter in the output formatters list which supports a content type from the explicitly specified content types '{ExplicitContentTypes}'.

depending on whether `result.ContentTypes` is set.

@rynowak Do you think Accept-header values should be included in this (`NoFormatter`) statement as well?

khellang:
@ryanbrandenburg?

src/Mvc/Mvc.Core/src/Infrastructure/ObjectResultExecutor.cs
rynowak: This seems ok then. It's an improvment, if we need to improve it more in the future we can.

rynowak (approved):

rynowak:
Thanks @khellang - this is going to get into preview 8 - I'll merge when the branch is open again

rynowak:
Nevermind, I lied. Thanks!

Issue #11032, opened 2019-06-10
Adding HTTP header with empty string value yields confusing error


bjorn-ali-goransson:
### Describe the bug
Some legacy code in .NET Framework has an HTTP header with an empty string as value. When porting the same code to .NET Core, the exception System.ArgumentException: 'Value does not fall within the expected range.' is thrown.

### To Reproduce
1. Using a blank project with ASP.NET Core 2.2
2. Run this code (replace the WriteAsync Hello World):

            app.Run(async (context) =>
            {
                context.Response.Headers.Add("test", "");
            });

4. See error below

### Expected behavior
The header was perhaps unnecessary and maybe even invalid. But the message "Value does not fall within the expected range" isn't really helpful, the offending header name isn't even mentioned, and if the response is sent outside of your own code (like the example above - do a write and you'll get it, though) you don't even get a meaningful stacktrace and you cannot possible know what the relevant request was. So the issue gets really hard to track down.

I had to decompile Microsoft.AspNetCore.Server.IIS.dll and put the generated PDB alongside it in the nuget cache to get on my way.

### Screenshots

![image](https://user-images.githubusercontent.com/668160/59190782-e7583000-8b7d-11e9-9ae2-a6feef7aab3f.png)

### Additional context

Perhaps a guard statement could be added here: to avoid confusing any future developers.

https://github.com/aspnet/AspNetCore/blob/aaaaf572fdf11088ee0efd205ff6834e16a36804/src/Servers/IIS/IIS/src/Core/IISHttpContext.cs#L388

But in the same time, it feels like the exception is a native one, and thus this may be the wrong repo to discuss it in.

anurse:
Native exceptions like this a converted from error codes returned by the native code. The error code returned here is `E_INVALIDARG`, which gets converted to this generic argument exception (because the code is all we have to go on).

One thing we could do is "pre-check" the argument and throw a more useful `ArgumentException` with a message (like `Header value cannot be an empty string`) **before** calling the native code.

Putting this in the backlog and marking "help wanted"/"good first issue" since this would be a great opportunity for a new contributor!

PR #11217, opened 2019-06-14
Adding HTTP header with empty string value yields confusing error #11032

Yogendra0Sharma:
 - Pre-Check **headerValueBytes** length 
 - if header value is empty than throw exception with custom message

Addresses #11032

### DIFF
--- a/src/Servers/IIS/IIS/src/Core/IISHttpContext.cs
+++ b/src/Servers/IIS/IIS/src/Core/IISHttpContext.cs
@@ -385,8 +385,12 @@ namespace Microsoft.AspNetCore.Server.IIS.Core
                             var headerNameBytes = Encoding.UTF8.GetBytes(headerPair.Key);
                             fixed (byte* pHeaderName = headerNameBytes)
                             {
-                                NativeMethods.HttpResponseSetUnknownHeader(_pInProcessHandler, pHeaderName, pHeaderValue, (ushort)headerValueBytes.Length, fReplace: isFirst);
-                            }
+								if(headerValueBytes.Length == 0){
+									throw new ArgumentException("Header value cannot be an empty string");
+								}else{
+									NativeMethods.HttpResponseSetUnknownHeader(_pInProcessHandler, pHeaderName, pHeaderValue, (ushort)headerValueBytes.Length, fReplace: isFirst);
+								}
+							}
                         }
                         else
                         {


src/Servers/IIS/IIS/src/Core/IISHttpContext.cs
```diff
@@ -385,8 +385,12 @@ public unsafe void SetResponseHeaders()
                             var headerNameBytes = Encoding.UTF8.GetBytes(headerPair.Key);
                             fixed (byte* pHeaderName = headerNameBytes)
                             {
-                                NativeMethods.HttpResponseSetUnknownHeader(_pInProcessHandler, pHeaderName, pHeaderValue, (ushort)headerValueBytes.Length, fReplace: isFirst);
-                            }
+								if(headerValueBytes.Length == 0){
+									throw new ArgumentException("Header value cannot be an empty string");
+								}else{
```
anurse: This will need to be reformatted to follow our code style. We use the standard VS style, which would be more like this:

```csharp
if (headerValueBytes.Length == 0)
{
	// ...
}
else
}
	// ...
}
``

Note: spaces around keywords and braces get their own line.

File: src/Servers/IIS/IIS/src/Core/IISHttpContext.cs [hunk 2]
```diff
@@ -385,8 +385,12 @@ public unsafe void SetResponseHeaders()
                             var headerNameBytes = Encoding.UTF8.GetBytes(headerPair.Key);
                             fixed (byte* pHeaderName = headerNameBytes)
                             {
-                                NativeMethods.HttpResponseSetUnknownHeader(_pInProcessHandler, pHeaderName, pHeaderValue, (ushort)headerValueBytes.Length, fReplace: isFirst);
-                            }
+								if(headerValueBytes.Length == 0){
+									throw new ArgumentException("Header value cannot be an empty string");
```
anurse: `ArgumentException` may not be the most appropriate here (I know I suggested it but I missed the full context 😄). Thoughts @Tratcher @jkotalik ?

Also, in the bug it was mentioned that having the header name be present in the message would be helpful.

analogrelay (commented):
Looks OK to me. @Tratcher @davidfowl I'm wondering if we should guard against empty values in the actual Header dictionary. If we don't think they make sense, and if they cause errors like the linked bug when present, perhaps that's where the guard should be.

davidfowl:
This also needs some tests

src/Servers/IIS/IIS/src/Core/IISHttpContext.cs
halter73: HttpResponseSetUnknownHeader was already throwing an ArgumentException in this case, right? So is this basically just changing the exception message to make it clearer?

Kestrel throws InvalidOperationExceptions for header names and values that it rejects (though it *does* allow empty header values), so it's probably better to use an InvalidOperationException here too for consistency.

I'm hoping this exception gets bubbled up through the response header dictionary's indexer setter and Add methods, but seeing that SetResponseHeaders is called by ProduceStart which is delayed until the headers are flushed, I'm guessing that's not the case.

We should ensure we throw an exception as early as possible for invalid empty header values and we should add a unit test for it similar to what we [have in Kestrel](https://github.com/aspnet/AspNetCore/blob/a79bb2a79d76a39d31809b71335e5a20e0747ff3/src/Servers/Kestrel/Core/test/HttpResponseHeadersTests.cs#L100).

Tratcher:
There's a case in the request headers where a Host header must be allowed with an empty string value. Otherwise no, empty headers don't make much sense.

Note I'm pretty sure HttpSys an Kestrel already silently ignore empty headers. Throwing isn't great because headers are usually sent indirectly via operations like WriteAsync.

Tratcher:
Kestrel skips StringValues with 0 entries and any individual null values in the list. It does not skip empty strings.
https://github.com/aspnet/AspNetCore/blob/a79bb2a79d76a39d31809b71335e5a20e0747ff3/src/Servers/Kestrel/Core/src/Internal/Http/HttpResponseHeaders.cs#L43-L56

HttpSys skips StringValues with 0 entries and converts null entries to empty strings.
https://github.com/aspnet/AspNetCore/blob/a79bb2a79d76a39d31809b71335e5a20e0747ff3/src/Servers/HttpSys/src/RequestProcessing/Response.cs#L507

src/Servers/IIS/IIS/src/Core/IISHttpContext.cs
Tratcher: This isn't worth throwing for, just skip the entry. I think you need to check for null or empty before encoding on line 380.

Tratcher (changes requested):
Throwing here isn't worth it, skip the empty values.

halter73:
I'm also fine with ignoring empty string header values.

Considering that setting an empty string header value would cause the header to be sent in Kestrel instead of ignored, we should probably write an info-level log or something when we ignore an empty string header value in IIS.

jkotalik:
@Tratcher @anurse we should revitalize this PR for master when we get the chance. Putting in the 5.0.0-preview1 milestone for now.

anurse:
We actually fixed this earlier by simply no-oping if the header value is empty. I think this isn't relevant any more.

PR #11218, opened 2019-06-14
[Fixes #11183] Race condition in RouteBase.EnsureLoggers

### LINKED ISSUE #1183: Issues of VS's integration with client side code building tools
There are some tool sets for client side code building, e.g.  r.js, webpack, and Fis3. Among these, I found Fis3 the most comprehensive and powerful one: https://github.com/fex-team/fis3. Briefly, Fis3 can do the following:  uglify and/or merge js/css resources external or inline, csssprite  and merge,  compact html, inline resources (like small images, templates etc).  With these capability, client side code can be optimized greatly.
The issue here is that, Fis3 aggressively modify the html files ( and the aspx, and cshtml files likewise), so its integration with vs build is an issue. .Net 5 and DNX have a separate www folder under the project root, this step can be helpful to incorporate Fis3 into vs build flow, but a more closer and careful study is needed for better and smooth integration of client side build tool into VS build.

### LINKED ISSUE #11183: RouteBase.EnsureLoggers race condition
### Describe the bug
I'm on the Azure Functions team and we see a very rare NullRef with this stack:
```
System.ArgumentNullException : Value cannot be null.
Parameter name: logger
   at Microsoft.AspNetCore.Routing.RouteConstraintMatcher.Match(IDictionary`2 constraints,RouteValueDictionary routeValues,HttpContext httpContext,IRouter route,RouteDirection routeDirection,ILogger logger)
   at Microsoft.AspNetCore.Routing.RouteBase.RouteAsync(RouteContext context)
```

I've looked into the issue (I thought it was a race on our end) but found that `_constraintLogger` (in `RouteBase`) ends up being null, even with the call to `EnsureLoggers()` that should be making sure it's initialized. 

```
private void EnsureLoggers(HttpContext context)
{
    if (_logger == null)
    {
        var factory = context.RequestServices.GetRequiredService<ILoggerFactory>();
        _logger = factory.CreateLogger(typeof(RouteBase).FullName);
        _constraintLogger = factory.CreateLogger(typeof(RouteConstraintMatcher).FullName);
    }
}
```

Imagine if the second `CreateLogger()` call here runs slowly (for any reason: Functions runs on a single core, file locks, etc -- it’d be rare, but possible). Now you’ve got `_logger != null`, but `_constraintLogger == null`. Any other threads calling this code will skip over this `if` statement and go on with their execution -- with a null `_constraintLogger`. If those two `CreateLogger()` lines were switched, it should fix it.

### To Reproduce
See my test -- required some sempahore coordination, but proved that it's possible, but rare: https://gist.github.com/brettsam/6f60bfccba80d1515be9cfb13d0d985b

javiercn:
Reorders lines to prevent race condition
#11183

### DIFF
--- a/src/Http/Routing/src/RouteBase.cs
+++ b/src/Http/Routing/src/RouteBase.cs
@@ -3,20 +3,19 @@
 
 using System;
 using System.Collections.Generic;
-using System.Text.Encodings.Web;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
-using Microsoft.AspNetCore.Routing.Internal;
 using Microsoft.AspNetCore.Routing.Logging;
 using Microsoft.AspNetCore.Routing.Template;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
-using Microsoft.Extensions.ObjectPool;
 
 namespace Microsoft.AspNetCore.Routing
 {
     public abstract class RouteBase : IRouter, INamedRouter
     {
+        private readonly object _loggersLock = new object();
+
         private TemplateMatcher _matcher;
         private TemplateBinder _binder;
         private ILogger _logger;
@@ -259,11 +258,25 @@ namespace Microsoft.AspNetCore.Routing
 
         private void EnsureLoggers(HttpContext context)
         {
+            // We check first using the _logger to see if the loggers have been initialized to avoid taking
+            // the lock on the most common case.
             if (_logger == null)
             {
-                var factory = context.RequestServices.GetRequiredService<ILoggerFactory>();
-                _constraintLogger = factory.CreateLogger(typeof(RouteConstraintMatcher).FullName);
-                _logger = factory.CreateLogger(typeof(RouteBase).FullName);
+                // We need to lock here to ensure that _constraintLogger and _logger get initialized atomically.
+                lock (_loggersLock)
+                {
+                    if (_logger != null)
+                    {
+                        // Multiple threads might have tried to accquire the lock at the same time. Technically
+                        // there is nothing wrong if things get reinitialized by a second thread, but its easy
+                        // to prevent by just rechecking and returning here.
+                        return;
+                    }
+
+                    var factory = context.RequestServices.GetRequiredService<ILoggerFactory>();
+                    _constraintLogger = factory.CreateLogger(typeof(RouteConstraintMatcher).FullName);
+                    _logger = factory.CreateLogger(typeof(RouteBase).FullName);
+                }
             }
         }
 
--- a/src/Http/Routing/src/RouteBase.cs
+++ b/src/Http/Routing/src/RouteBase.cs
@@ -262,8 +262,8 @@ namespace Microsoft.AspNetCore.Routing
             if (_logger == null)
             {
                 var factory = context.RequestServices.GetRequiredService<ILoggerFactory>();
-                _logger = factory.CreateLogger(typeof(RouteBase).FullName);
                 _constraintLogger = factory.CreateLogger(typeof(RouteConstraintMatcher).FullName);
+                _logger = factory.CreateLogger(typeof(RouteBase).FullName);
             }
         }
 


rynowak (approved):

src/Http/Routing/src/RouteBase.cs
```diff
@@ -262,8 +262,8 @@ private void EnsureLoggers(HttpContext context)
             if (_logger == null)
             {
                 var factory = context.RequestServices.GetRequiredService<ILoggerFactory>();
-                _logger = factory.CreateLogger(typeof(RouteBase).FullName);
                 _constraintLogger = factory.CreateLogger(typeof(RouteConstraintMatcher).FullName);
+                _logger = factory.CreateLogger(typeof(RouteBase).FullName);
```
anurse: Perhaps I'm wrong or misunderstanding how things work but it was my understanding that without `volatile`, it's entirely possible for lines like this to be reordered by the CPU. The only way to guarantee that every core/thread will see `_constraintLogger` *definitely assigned* before `_logger` is would be locking or using `volatile` fields.

If this seems to fix the race, then maybe it's moot, but I don't know that we can **assert** strongly that the ordering issue is resolved by a change like this.

rynowak: Re-ordering virtual method calls?

rynowak: If we *want* to be super clean and clear about this, I'd just slap a lock around the init, or use `LazyInitializer` with a valuetuple. 

I think I went through the same steps you did and then saw that these are iface calls. The compiler can't reason about that.

javiercn: I believe this should be enough. I don't think we mark things volatile almost anywhere in the stack. There are a bunch of volatile things in Kestrel, but id rather use https://docs.microsoft.com/en-us/dotnet/api/system.threading.volatile.write?view=netframework-4.8#System_Threading_Volatile_Write__1___0____0_ instead as this is a one-off. volatile has the potential to have perf implications

javiercn: I've updated it to use Volatile.Write cuz it's cheap

anurse: > Re-ordering virtual method calls?

Maybe not, I'm really just being conservative here and don't entirely know what I'm talking about 😛.


> If we _want_ to be super clean and clear about this, I'd just slap a lock around the init, or use `LazyInitializer` with a valuetuple.

If this isn't a hot path, that seem safest.

> I've updated it to use Volatile.Write cuz it's cheap

🤷‍♂ 	probably ok?

davidfowl:
Please add a comment.

src/Http/Routing/src/RouteBase.cs
davidfowl: > If we want to be super clean and clear about this, I'd just slap a lock around the init, or use LazyInitializer with a valuetuple.

That would be better IMO.

cc @stephentoub PTAL

stephentoub: You'd ideally also have a Volatile.Read on the read of _logger; that is, for example, what LazyInitializer.EnsureInitialized itself does, to avoid any concerns on weaker memory models around reordered, introduced, or coalesced reads.  I agree with @davidfowl, @anurse, and @rynowak that it'd be better to use a lock or Lazy or LazyInitializer; they exist in part to discourage devs from writing their own lock-free lazy initialization.

javiercn: Fair enough, I'll change this to use a LazyInitializer then.
@stephentoub thanks!

javiercn:
> or use LazyInitializer with a valuetuple.

Updated to use a lock, I wasn't sure how to accomplish that with LazyInitializer. Maybe I'm missing something. The alternative would have been to put each initialization inside LazyInitializer calls, which I imagine would have prevented reordering and have done the trick.

src/Http/Routing/src/RouteBase.cs
```diff
@@ -3,20 +3,19 @@
...
 using Microsoft.AspNetCore.Routing.Template;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
-using Microsoft.Extensions.ObjectPool;
 
 namespace Microsoft.AspNetCore.Routing
 {
     public abstract class RouteBase : IRouter, INamedRouter
     {
+        private object _loggersLock = new object();
```
pranavkm: `private readonly object _loggersLock = new object();`

pranavkm (approved):

PR #11227, opened 2019-06-14
Move SPA flows to use code

### LINKED ISSUE #9869: Use Code+PKCE OAuth2 flow for SPA applications
Current SPA templates are using API request to load [oidc-client](https://github.com/IdentityModel/oidc-client-js) configuration which boils down to following
[source code](https://github.com/aspnet/AspNetCore/blob/master/src/Identity/ApiAuthorization.IdentityServer/src/Extensions/DefaultClientRequestParametersProvider.cs#L41). As you can see it only supports implicit code flow which is less secure than alternatives.

There is a very good article by @brockallen called [The State of the Implicit Flow in OAuth2](https://brockallen.com/2019/01/03/the-state-of-the-implicit-flow-in-oauth2/) which describes more secure way of handling authentication in browser based applications using authorization code flow with Proof Key for Code Exchange by OAuth Public Clients (also known as PKCE).

[oidc-client](https://github.com/IdentityModel/oidc-client-js) which is used in all ASP.NET Core SPA templates already [supports](https://github.com/IdentityModel/oidc-client-js/commit/c68ceaa78a43c92ebe6198a07f3bab6b916d559a) Code + PKCE. Here is an [example](https://github.com/IdentityServer/IdentityServer4/blob/master/samples/Quickstarts/9_Combined_AspId_and_EFStorage/src/JavaScriptClient/wwwroot/app.js#L25).

javiercn:
Move SPA flows to use Code+PKCE 
[Fixes #9869]


brockallen:
BTW, I have a release soon of oidc-client-js that is a slimmed down version with RSA only that cuts 200k fromt he size of the library. I will try to find time this week to release. I'd like to get it in ASAP for you -- what's a deadline?

javiercn:
@brockallen It shouldn't be a problem. I'll start a thread.

brockallen:
I just released 1.8.0 -- have a try. You'll need to change to using the 'oidc-client.rsa256.slim.min.js` version.

javiercn:
Thanks @brockallen

javiercn:
@SteveSandersonMS What was the process for updating the NPM dependencies. I tried this:
```
npm install package@version on the ClientApp folder of the react and angular templates.
```

It generated a package-lock.json that contained a bunch of changes I wasn't sure about.
* `optional: true` was added to a bunch of places.
* Some angular dependencies were downgraded from 7.2.8 to 7.2.5

Also, when I pushed, the CI complained the package-lock.json being corrupted/wrong format. Do you have any idea?

javiercn:
@brockallen I would have expected to get the size reduction for free given that we are using webpack in both react and angular templates, so I imagine any trimming we need will happen there. 

Is that not the case?

javiercn:
@SteveSandersonMS You can take a look at the changes in this PR where I've isolated them for angular
https://github.com/aspnet/AspNetCore/pull/11284/files

javiercn:
@SteveSandersonMS Nevermind, I must have done something wrong, as when I tried again it has succeded.

pranavkm (approved):

mkrtich-m:
I think this change will break existing SPA template flow. 
Oidc-client automatically uses reload to exchange code to token. In this case angular code will try to manually redirect to token endpoint which will cause unauthorized code exception cause code was already used to get token.

cc @brockallen

javiercn:
That doesn't seem to be an issue. We have functional tests covering the authentication process.

mkrtich-m:
Where can I see the results of those test ?

javiercn:
Not sure if you have access to https://dev.azure.com/dnceng/public/_build/results?buildId=226305&view=ms.vss-test-web.build-test-results-tab

mkrtich-m:
I can see them.
Thanks

mkrtich-m:
I cannot find any JavaScript tests. Seems like all tests are C# based.
What I'm describing is JavaScript related issue.

javiercn:
I understand. We use Selenium for our E2E tests, which cover authentication.

PR #11235, opened 2019-06-14
Add file manifest for runtime and targeting pack

### LINKED ISSUE #10215: Add manifest to ASP.NET targeting pack
The targeting pack for ASP.NET Core should have a manifest listing all of the reference assemblies along with their assembly and file versions.  This should be in the `data\FrameworkList.xml` file, and the format should be like this:

```xml
<FileList Name=".NET Core 3.0" TargetFrameworkIdentifier=".NETCore" TargetFrameworkVersion="3.0" FrameworkName="Microsoft.NETCore.App">
  <File AssemblyName="Microsoft.CSharp" PublicKeyToken="..." AssemblyVersion="..." FileVersion="..." />
  <File AssemblyName="Microsoft.VisualBasic" PublicKeyToken="..." AssemblyVersion="..." FileVersion="..." />
  <File AssemblyName="System" PublicKeyToken="..." AssemblyVersion="..." FileVersion="..." />
  <!-- etc. -->
</FileList>
```

Related issue for the other targeting packs: https://github.com/dotnet/core-setup/issues/5074

@natemcmaster @pranavkm

jkotalik:
Fixes https://github.com/aspnet/AspNetCore/issues/10215 and https://github.com/aspnet/AspNetCore/issues/10216

This seems to be working locally for win-x64. I want to confirm things look good on CI before I do some finalization.

Few open questions that I'm working on solving:
- For the runtime pack manifest, make sure manifests are in the right location and have the right names (may need clarification here as it wasn't clear in the issues)
- Determine which dlls should be in which manifest (should native dlls be in the targeting pack?)

I'd also like to add a few automated tests once I confirm things are correct.

### DIFF
--- a/eng/tools/RepoTasks/CreateFrameworkListFile.cs
+++ b/eng/tools/RepoTasks/CreateFrameworkListFile.cs
@@ -1,15 +1,14 @@
-// Licensed to the .NET Foundation under one or more agreements.
-// The .NET Foundation licenses this file to you under the MIT license.
-// See the LICENSE file in the project root for more information.
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using Microsoft.Build.Framework;
-using Microsoft.Build.Utilities;
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Xml.Linq;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
 
 namespace RepoTasks
 {
@@ -26,7 +25,7 @@ namespace RepoTasks
 
         /// <summary>
         /// Extra attributes to place on the root node.
-        /// 
+        ///
         /// %(Identity): Attribute name.
         /// %(Value): Attribute value.
         /// </summary>
--- a/eng/tools/RepoTasks/CreateFrameworkListFile.cs
+++ b/eng/tools/RepoTasks/CreateFrameworkListFile.cs
@@ -53,7 +53,6 @@ namespace RepoTasks
                     IsNative = item.GetMetadata("IsNativeImage") == "true",
                     IsSymbolFile = item.GetMetadata("IsSymbolFile") == "true"
                 })
-                .Where(a => a.Filename.Contains("aspnetcorev2"))
                 .Where(f =>
                     !f.IsSymbolFile &&
                     (f.Filename.EndsWith(".dll", StringComparison.OrdinalIgnoreCase) || f.IsNative))
--- a/eng/tools/RepoTasks/CreateFrameworkListFile.cs
+++ b/eng/tools/RepoTasks/CreateFrameworkListFile.cs
@@ -34,10 +34,6 @@ namespace RepoTasks
 
         public override bool Execute()
         {
-            //while (!Debugger.IsAttached)
-            //{
-
-            //}
             XAttribute[] rootAttributes = RootAttributes
                 ?.Select(item => new XAttribute(item.ItemSpec, item.GetMetadata("Value")))
                 .ToArray();
@@ -54,9 +50,10 @@ namespace RepoTasks
                     TargetPath = item.GetMetadata("TargetPath"),
                     AssemblyName = FileUtilities.GetAssemblyName(item.ItemSpec),
                     FileVersion = FileUtilities.GetFileVersion(item.ItemSpec),
-                    IsNative = item.GetMetadata("IsNative") == "true",
+                    IsNative = item.GetMetadata("IsNativeImage") == "true",
                     IsSymbolFile = item.GetMetadata("IsSymbolFile") == "true"
                 })
+                .Where(a => a.Filename.Contains("aspnetcorev2"))
                 .Where(f =>
                     !f.IsSymbolFile &&
                     (f.Filename.EndsWith(".dll", StringComparison.OrdinalIgnoreCase) || f.IsNative))
--- a/eng/tools/RepoTasks/CreateFrameworkListFile.cs
+++ b/eng/tools/RepoTasks/CreateFrameworkListFile.cs
@@ -1,0 +1,170 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Xml.Linq;
+
+namespace RepoTasks
+{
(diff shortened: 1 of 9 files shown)


jkotalik:
@dsplaisted I think I need to run this PR by you to make sure I'm producing the right manifests in the correct locations.

build/tasks/CreateFrameworkListFile.cs
```diff
@@ -0,0 +1,108 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
```
dougbu: This isn't our usual C# license header and it's aligned with neither `<PackageLicenseExpression>Apache-2.0</PackageLicenseExpression>` in the root Directory.Build.props file nor the content of our LICENSE.txt file.

jkotalik: This file was copied from Core-setup. I'll change the header.

build/tasks/CreateFrameworkListFile.cs
```diff
@@ -0,0 +1,107 @@
...
+        /// </summary>
+        public ITaskItem[] RootAttributes { get; set; }
+
+        public override bool Execute()
+        {
+            XAttribute[] rootAttributes = RootAttributes
+                ?.Select(item => new XAttribute(item.ItemSpec, item.GetMetadata("Value")))
+                .ToArray();
+
+            var frameworkManifest = new XElement("FileList", rootAttributes);
```
dougbu: Doesn't this blow up if `rootAttributes` (or `RootAttributes`) is `null`? Suggest you remove the `null`-conditional in the `rootAtrributes` declaration (which can then use `var`) and make `RootAttributes` `[Required]`.

dougbu (approved):

build/tasks/CreateFrameworkListFile.cs
jkotalik: This is copied code from corefx. I think there is an overall goal of unifying these tasks in arcade (Can't find the issue, but it's there 😄 ).

IMO RootAttributes should be Required and we remove nullables.

jkotalik:
Verified the targeting pack and the runtime pack work locally. For the runtime pack, I change the KnownFrameworkReference to a local dev build of the M.A.App.Runtime.win-x64 and made sure restore/build worked. dotnet run failed from `System.IO.FileNotFoundException: Could not load file or assembly 'Microsoft.Bcl.AsyncInterfaces, Version=1.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'. The system cannot find the file specified.`, however, I think this is due to me not targeting an equivalent version of Microsoft.NetCore.App.Runtime.win-x64.

I'll be on top of verifying this once it gets merged into core-sdk and can react quickly.

eng/tools/RepoTasks/CreateFrameworkListFile.cs
```diff
@@ -0,0 +1,107 @@
...
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Xml.Linq;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+
+namespace RepoTasks
+{
+    public class CreateFrameworkListFile : Task
```
JunTaoLuo: Is this analogous to https://github.com/dotnet/core-setup/blob/5cfc2cab25bfc431ef89572aa2705725a95aca32/tools-local/tasks/CreateFrameworkListFile.cs? Should we move this to arcade and reduce the duplication?

eng/tools/RepoTasks/FileUtilities.cs
```diff
@@ -7,41 +7,42 @@
 using System.IO;
 using System.Reflection;
 
-namespace Microsoft.DotNet.Build.Tasks
+namespace RepoTasks
```
JunTaoLuo: Changes here also seem similar to https://github.com/dotnet/core-setup/blob/5cfc2cab25bfc431ef89572aa2705725a95aca32/tools-local/tasks/FileUtilities.cs. That's the intention right?

src/Framework/src/Microsoft.AspNetCore.App.Runtime.csproj
```diff
@@ -100,6 +100,9 @@ This package is an internal implementation of the .NET Core SDK and is not meant
     <RuntimePackageRoot>$([MSBuild]::EnsureTrailingSlash('$(RuntimePackageRoot)'))</RuntimePackageRoot>
     <CrossgenToolPath>$([System.IO.Path]::Combine('$(RuntimePackageRoot)', 'tools', '$(CrossgenToolPackagePath)'))</CrossgenToolPath>
     <AssetTargetFallback>$(AssetTargetFallback);native,Version=0.0</AssetTargetFallback>
+
+    <FrameworkListFileName>RuntimeList.xml</FrameworkListFileName>
+    <FrameworkListOutputPath>$(ArtifactsObjDir)$(FrameworkListFileName)</FrameworkListOutputPath>
```
JunTaoLuo: nit: FrameworkList => RuntimeList. these are different files so the property names should probably follow suite.

src/Framework/src/Microsoft.AspNetCore.App.Runtime.csproj
```diff
@@ -448,4 +450,19 @@ This package is an internal implementation of the .NET Core SDK and is not meant
       Condition="'$(ArchiveExtension)' == '.tar.gz'" />
   </Target>
 
+  <Target Name="IncludeFrameworkListFile"
```
JunTaoLuo: Is there a way to shared the implementation of this target between the runtime pack and ref pack? Just to reduce duplication, not strictly necessary.

eng/tools/RepoTasks/CreateFrameworkListFile.cs
jkotalik: https://github.com/dotnet/arcade/issues/2704

eng/tools/RepoTasks/FileUtilities.cs
jkotalik: https://github.com/dotnet/arcade/issues/2704

jkotalik: Yes, eventually this will be shared.

src/Framework/src/Microsoft.AspNetCore.App.Runtime.csproj
jkotalik: I think it would cause more pain in the long run than help. Differences include RefPackContent vs SharedFxContent, the target file name, different DependsOnTargets, etc. Anyways, all this target is doing is invoking another target and putting stuff into package files.

src/Framework/src/Microsoft.AspNetCore.App.Runtime.csproj
jkotalik: How much do you care about this? PR is green and I'd rather not wait another 2 hours for a rebuild on a rename like this if you have no other feedback.

PR #11245, opened 2019-06-15
Correct EnableEndpointRouting spelling

dannyBies:
Fixed spelling of EnableEndpointRouting that could show up on the intellisense for app.UseMvc


analogrelay (approved):
Looks good to me, though this isn't the IntelliSense docs, it's the error message produced by the MVC Roslyn Analyzers. Still a misspelling we should fix, but I wanted to make sure that if you *intended* to fix IntelliSense, this won't do that ;). If you just mean that the error message doesn't match what would show up in IntelliSense then this is all good!

dannyBies:
Yeah my bad, I meant that the error message didn't match what does up in intellisense!

anurse:
Thanks for the contribution!

Issue #11246, opened 2019-06-15
JsonResult and Controller.Json() do not work without calling AddNewtonsoftJson() in 3.0 preview 6


martincostello:
### Describe the bug

If `AddNewtonsoftJson()` from `Microsoft.AspNetCore.Mvc.NewtonsoftJson` is not called while configuring MVC in ASP.NET Core 3.0 preview 6, then rendering the result of a call to the `Controller.Json()` method fails with an `InvalidOperationException` and an HTTP 500 status.

```
InvalidOperationException: No service for type 'Microsoft.AspNetCore.Mvc.Infrastructure.IActionResultExecutor`1[Microsoft.AspNetCore.Mvc.JsonResult]' has been registered.
Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService(IServiceProvider provider, Type serviceType)
Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService<T>(IServiceProvider provider)
Microsoft.AspNetCore.Mvc.JsonResult.ExecuteResultAsync(ActionContext context)
Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.InvokeResultAsync(IActionResult result)
Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.ResultNext<TFilter, TFilterAsync>(ref State next, ref Scope scope, ref object state, ref bool isCompleted)
Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.InvokeNextResultFilterAsync<TFilter, TFilterAsync>()
Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.Rethrow(ResultExecutedContextSealed context)
Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.ResultNext<TFilter, TFilterAsync>(ref State next, ref Scope scope, ref object state, ref bool isCompleted)
Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.InvokeResultFilters()
Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.<InvokeNextResourceFilter>g__Awaited|24_0(ResourceInvoker invoker, Task lastTask, State next, Scope scope, object state, bool isCompleted)
Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.Rethrow(ResourceExecutedContextSealed context)
[...]

PR #11247, opened 2019-06-15
Register SystemTextJsonResultExecutor

martincostello:
Register `SystemTextJsonResultExecutor` as part of MVC core services so that `JsonResult` works without Newtonsoft.Json being registered via `AddNewtonsoftJson()`.

I wasn't sure on the best approach to use for the tests, so I forked the existing `JsonResultTest` functional test into two as `JsonResultWithNewtonsoftJsonTest` and `JsonResultWithSystemTextJsonTest`.

Addresses #11246.


pranavkm (approved):
Thanks @martincostello

PR #11253, opened 2019-06-15
More read interface IList.Count once rather than per iteration + miscellany

martincostello:
* Inspired by #9962 I had a look through some of the less hot-path areas to see if there were anymore `IList.Count` `for` loops that could benefit from the same change. There's also some for `ReadOnlyCollection<T>` as looking [at the code](https://github.com/dotnet/corefx/blob/c4c1985531cc1f25c805bbc4801b85c98e0dfee0/src/Common/src/CoreLib/System/Collections/ObjectModel/ReadOnlyCollection.cs#L29) that just calls through to `IList.Count` anyway, so would benefit the same way(?).
  * Use `nameof()` on an enum value instead of `ToString()` to make a compile-time constant.
  * Initialise a dictionary with a fixed number of members with a capacity equal to the number of members.
  * Override `DisposeAsync` on `FileBufferingReadStream` to call `DisposeAsync()` on the wrapped `Stream`.
  * Use `FileBufferingReadStream.DisposeAsync()` in a number of input formatters.
  * Use `GetValueOrDefault()` on a `long?` instead of calling `HasValue` once and `Value` up to three times in a number of input formatters.

### DIFF
--- a/src/Http/WebUtilities/ref/Microsoft.AspNetCore.WebUtilities.netcoreapp3.0.cs
+++ b/src/Http/WebUtilities/ref/Microsoft.AspNetCore.WebUtilities.netcoreapp3.0.cs
@@ -54,6 +54,8 @@ namespace Microsoft.AspNetCore.WebUtilities
         public override long Position { get { throw null; } set { } }
         public string TempFileName { get { throw null; } }
         protected override void Dispose(bool disposing) { }
+        [System.Diagnostics.DebuggerStepThroughAttribute]
+        public override System.Threading.Tasks.ValueTask DisposeAsync() { throw null; }
         public override void Flush() { }
         public override int Read(byte[] buffer, int offset, int count) { throw null; }
         [System.Diagnostics.DebuggerStepThroughAttribute]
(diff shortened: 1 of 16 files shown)


martincostello:
I wasn't sure exactly what tool was used in #9962 to generate the call counts, so I haven't crunched the before and after numbers.

jkotalik (approved):
Looks good to me. Thanks!

jkotalik:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=226845

jkotalik:
@davidfowl You added VerifyCountersFireWithCorrectValues, I'll disable if this continues to fail.

@halter73 also a reconnect test failure: https://dev.azure.com/dnceng/public/_build/results?buildId=226845&view=ms.vss-test-web.build-test-results-tab&runId=5614518&resultId=133322&paneView=debug

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/1979

pranavkm (approved):
Thanks for the PR!

PR #11265, opened 2019-06-16
Install Visual Studio: add support for preview

alefranz:
Currently the `eng/scripts/InstallVisualStudio.ps1` script, if only the preview version is installed, will install a release version instead of updating the preview one.
This add the ability to target the preview version.
It also allow to decide which channel to target if both are installed (I should be able to test it tomorrow on a different machine) or allow to install a different channel in addition to it.
(Note: I doesn't support installing from scratch an additional preview 'Edition', as the automatic path selection will point to the existing Preview, however I am not sure if it is a supported scenario, I had issue in the past with the VS installer when attempting it).

@natemcmaster let me know if I should take a different approach.
Also the download url has been guessed by trial and error, so although it works I am not sure it is the supported one.

Thank you,
 Alessio

### DIFF
--- a/eng/scripts/InstallVisualStudio.ps1
+++ b/eng/scripts/InstallVisualStudio.ps1
@@ -9,6 +9,10 @@
         Community
         Professional
         Enterprise (the default)
+.PARAMETER Channel
+    Selects which channel of Visual Studio to install. Must be one of these values:
+        Release (the default)
+        Preview
 .PARAMETER InstallPath
     The location on disk where Visual Studio should be installed or updated. Default path is location of latest
     existing installation of the specified edition, if any. If that VS edition is not currently installed, default
@@ -28,6 +32,8 @@
 param(
     [ValidateSet('BuildTools','Community', 'Professional', 'Enterprise')]
     [string]$Edition = 'Enterprise',
+    [ValidateSet('Release', 'Preview')]
+    [string]$Channel = 'Release',
     [string]$InstallPath,
     [switch]$Passive,
     [switch]$Quiet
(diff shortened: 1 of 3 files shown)


eng/scripts/InstallVisualStudio.ps1
```diff
@@ -78,7 +93,11 @@ if (-not $InstallPath) {
 }
 
 if (-not $InstallPath) {
-    $InstallPath = "${env:ProgramFiles(x86)}\Microsoft Visual Studio\2019\$Edition"
+    if ("$Channel" -eq "Preview") {
+        $InstallPath = "${env:ProgramFiles(x86)}\Microsoft Visual Studio\2019\Preview"
```
dougbu: This is going to lead to weird errors if anyone installs multiple preview editions accidentally or on purpose. Suggest
```suggestion
        $InstallPath = "${env:ProgramFiles(x86)}\Microsoft Visual Studio\2019\${Edition}_Pre"
```
or perhaps
```suggestion
        $InstallPath = "${env:ProgramFiles(x86)}\Microsoft Visual Studio\2019_Pre\$Edition"
```

dougbu (changes requested):
Not really sure it's necessary to update this convenience script but we'll take it (with a small change)

dougbu:
> Also the download url has been guessed by trial and error, so although it works I am not sure it is the supported one.

This is fine. FYI the Preview channel is at 16.2.0-pre.2.0+29006.145 and Release is at 16.1.3+29009.5.

eng/scripts/InstallVisualStudio.ps1
alefranz: Hi Doug,
thanks for reviewing.

I used `Preview` as that is the default for the Preview version if you use the installer.
If you have more then one version installer, launching regularly the installer you are presented with an error:
> Visual Studio Enterprise 2019 is already installed in this folder. You’ll need to choose another location.
You are then forced to manually change the path:

image

I tried to emulate the behavior, however when the `--installPath` parameter is provided to the installer it just try to modify the installation at that path, even if the installer Edition doesn't match (which feels like a bug in the installer).

About your suggestions, I would prefer the first one as all the edition are installed in the `2019` folder. Probably I would spell out `Preview` instead of `Pre`. What do you think?
Is it worth to add some logic to detect if Preview exists and use this name only as fallback when preview is in use but contains a different edition?

Thank you,
 Alessio

dougbu: This script doesn't need to be perfect because the installer will normally run interactively. So, go with a simple approach that avoids this expected problem i.e. include `$Edition` in the installation location. (The script will update the 2019\Preview folder if the edition installed there matches. But, that "special" default isn't necessary.)

> I would spell out Preview instead of Pre.

My only concern is the length of the path but four characters shouldn't make or break things. If spelling out "Preview" works with the Professional edition (the longest edition name of interest since TestController won't help much), go for it.

alefranz: That's a valid point.
I took your suggestion to stay on the safe side.

dougbu (approved):

dougbu:
I'll get this in once the CI checks pass.

alefranz:
Thank you!

dougbu:
Thanks @alefranz❕

PR #11268, opened 2019-06-16
TestServer sends remaining content in pipe writer on complete

JamesNK:
Change TestServer to match Kestrel's behavior by sending unflushed content in pipe writer when completing request.

### DIFF
--- a/src/Hosting/TestHost/src/HttpContextBuilder.cs
+++ b/src/Hosting/TestHost/src/HttpContextBuilder.cs
@@ -127,6 +127,9 @@ namespace Microsoft.AspNetCore.TestHost
 
         internal async Task CompleteResponseAsync()
         {
+            // Flush any pending content in writer
+            await _httpContext.Response.BodyWriter.FlushAsync();
+
             _pipelineFinished = true;
             await ReturnResponseMessageAsync();
             _responseStream.CompleteWrites();
--- a/src/Hosting/WindowsServices/src/Microsoft.AspNetCore.Hosting.WindowsServices.csproj
+++ b/src/Hosting/WindowsServices/src/Microsoft.AspNetCore.Hosting.WindowsServices.csproj
@@ -16,9 +16,7 @@
     <Reference Include="System.ServiceProcess.ServiceController" />
   </ItemGroup>
   <ItemGroup>
-    <Compile Update="WebHostService.cs">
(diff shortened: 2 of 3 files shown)


src/Hosting/WindowsServices/src/Microsoft.AspNetCore.Hosting.WindowsServices.csproj
```diff
@@ -16,9 +16,7 @@
     <Reference Include="System.ServiceProcess.ServiceController" />
   </ItemGroup>
   <ItemGroup>
-    <Compile Update="WebHostService.cs">
-      <SubType>Component</SubType>
-    </Compile>
+    <Compile Update="WebHostService.cs" />
```
JamesNK: VS changed this automatically 🤷‍♂

jkotalik (approved):

src/Hosting/TestHost/src/HttpContextBuilder.cs
```diff
@@ -127,6 +127,9 @@ internal void AbortRequest()
 
         internal async Task CompleteResponseAsync()
         {
+            // Flush any pending content in writer
+            await _httpContext.Response.BodyWriter.FlushAsync();
```
Tratcher: Isn't this redundant with `PipeWriter.Complete()` that's called from `CompleteWrites()`? Do you just need to reverse the calls to ReturnResponseMessageAsync and CompleteWrites?

halter73: Yeah. I'm not sure I understand this change. If the PipeWriter is completed, that should flush any pending content. The reader might see the pending data until after Complete() returns, but that's also true for FlushAsync() with a PauseWriterThreshold of anything other than 1.

JamesNK: It isn't redundant because the unit test fails if FlushAsync isn't here.

Is there a better way to do this?

Tratcher:
Recommend closing this PR. We realized this is an issue with the PipeWriter polyfill, not with the server. I'll file an issue.

JamesNK:
Fix is either https://github.com/aspnet/AspNetCore/issues/11305 or changing TestServer to use pipes instead of Streams.

PR #11276, opened 2019-06-17
Do not overwrite `$LASTEXITCODE` in `build.ps1`'s `finally` block

dougbu:
- part of aspnet/AspNetCore-Internal#2665
- should surface errors we're currently hiding


dougbu:
Expect Windows PR validation to fail until the rest of aspnet/AspNetCore-Internal#2665 is fixed ☹️

JunTaoLuo (approved):

dougbu:
Needs more work -- Windows x64 build failed in MSBuild with 16 errors (as usual) but that step exited successfully.

Issue #10495, opened 2019-05-23
Missing symbols for Microsoft.Web.Xdt.Extensions.dll in 3.0 builds


natemcmaster:
Symcheck errors fail consistently when checking the AspNetCoreRuntime.3.0.x64 package. 

> Checking file E:\A\_work\r1\a\checkSymbols\c14973bc-38c2-4465-b39e-2027cbd88420\content\Microsoft.Web.Xdt.Extensions.dll ... No symbols found on MSDL or SymWeb!

This .dll is only found in AspNetCoreRuntime.3.0.x64, and it appears we are missing symbols for it.

dougbu:
Once this and #10297 are done, will be able to remove an exclusion from our release pipeline's "Publish and verify symbols" deployment

dougbu:
4f67ff921673 and a slight change in the release pipeline

PR #11277, opened 2019-06-17
Deploy Microsoft.Web.Xdt.Extensions.pdb

dougbu:
- #10495
- add a general mechanism to publish "loose" PDBs

also, correct a typo from af2f647d635d: always publish the log artifacts

### DIFF
--- a/.azure/pipelines/ci.yml
+++ b/.azure/pipelines/ci.yml
@@ -131,15 +131,17 @@ jobs:
     artifacts:
     - name: Windows_Logs
       path: artifacts/log/
+      publishOnError: true
     - name: Windows_Packages
       path: artifacts/packages/
+    - name: Windows_Symbols
+      path: artifacts/symbols/
     - name: Windows_VSIX
       path: artifacts/VSSetup/
     - name: Windows_Manifests
       path: artifacts/manifests/
     - name: Windows_Installers
       path: artifacts/installers/
-      publishOnError: true
 
 # Build Windows ARM
 - template: jobs/default-build.yml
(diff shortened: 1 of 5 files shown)


dougbu:
@jkotalik FYI, this might allow you to simplify Microsoft.AspNetCore.ANCMSymbols.csproj

dougbu:
@JunTaoLuo trial internal runs:
- build https://dnceng.visualstudio.com/internal/_build/results?buildId=227139
- release https://dnceng.visualstudio.com/internal/_releaseProgress?releaseId=9329

jkotalik:
Interesting. I'd be curious to see if this works with native projects!

Just to be sure I understand, we do want to publish symbols for this package, correct? IIRC, this package is used by the aspnetcore site extension to change the applicationhost.config file.

dougbu:
> Interesting. I'd be curious to see if this works with native projects!

Symbol .nupkg files are just a transport mechanism. This approach avoids the extra wrapper when the actual package isn't needed.

> we do want to publish symbols for this package, correct? IIRC, this package is used by the aspnetcore site extension to change the applicationhost.config file.

We should publish symbols for every assembly we publish. In this case, it may be unlikely any user would debug the code but that doesn't matter much. (For one thing, checking symbol deployment doesn't allow exclusions and this assembly is shipped in much larger bundles.)

JunTaoLuo (approved):

dougbu:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=227209

dougbu:
Regular AspNetCore-ci build completed successfully but was not reported back to GitHub

aspnet-hello:

dougbu:
As the Hello bot just confirmed, Helix failure was due to known-flaky tests

PR #11284, opened 2019-06-17
Update OIDC dependency

javiercn:
Updates OIDC to 1.8.0.

Doesn't update any script tag yet. Simply upates the version for now to see how the CI behaves.

### DIFF
--- a/src/ProjectTemplates/Web.Spa.ProjectTemplates/content/Angular-CSharp/ClientApp/package-lock.json
+++ b/src/ProjectTemplates/Web.Spa.ProjectTemplates/content/Angular-CSharp/ClientApp/package-lock.json
@@ -1,5 +1,5 @@
 {
-  "name": "Company.WebApplication1",
+  "name": "company.webapplication1",
   "version": "0.0.0",
   "lockfileVersion": 1,
   "requires": true,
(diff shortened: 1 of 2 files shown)


src/ProjectTemplates/Web.Spa.ProjectTemplates/content/Angular-CSharp/ClientApp/package-lock.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "Company.WebApplication1",
+  "name": "company.webapplication1",
```
javiercn: This needs to be all lowercase

src/ProjectTemplates/Web.Spa.ProjectTemplates/content/Angular-CSharp/ClientApp/package-lock.json
```diff
@@ -854,7 +767,8 @@
     "@types/selenium-webdriver": {
       "version": "3.0.15",
       "resolved": "https://registry.npmjs.org/@types/selenium-webdriver/-/selenium-webdriver-3.0.15.tgz",
-      "integrity": "sha512-5nh8/K2u9p4bk95GGCJB7KBvewaB0TUziZ9DTr+mR2I6RoO4OJVqx7rxK83hs2J1tomwtCGkhiW+Dy8EUnfB+Q=="
+      "integrity": "sha512-5nh8/K2u9p4bk95GGCJB7KBvewaB0TUziZ9DTr+mR2I6RoO4OJVqx7rxK83hs2J1tomwtCGkhiW+Dy8EUnfB+Q==",
+      "optional": true
```
javiercn: Not sure why this is getting added. I'm using npm 6.9.0 and node 10.16.0. It might be that we used an older node version before or that the format changed with a later version of node.

In any case, we should migrate it to this format as angular8 will require node10 (current LTS) and doesn't support previous versions.

javiercn:
Got merged in a separate PR

PR #11286, opened 2019-06-17
Update oidc dependency (angular and react)

javiercn:


ryanbrandenburg (commented):
Nothing wrong with the code but the new version needs to go through OSS and security review. @mkArtakMSFT to follow up on that and give the approval when its done.

javiercn:
That's handled automatically by our internal processes AFAIK.

ryanbrandenburg:
I don't believe so, but I'd be happy to be proven wrong and be able to merge the Angular update PR.

mkArtakMSFT:
All of these are updates to existing dependencies. So let's take these, as ideally we get higher security with patched dependencies.
As for the process, we should be find updating existing dependencies and component governance should take care of this. The OSS and security review process is for bigger dependencies like `Angular8` and `IdentityServer4` which I'm following up on.

mkArtakMSFT (approved):

PR #11289, opened 2019-06-17
Complete and make ignitor more extensible

javiercn:
I added a bunch stuff that I'm gonna be needing and I've moved the bootstrap code into a separate class library so that it can be reused. I would like the flow of this to be something like:
* I can create a regular functional test (like the ones we have)
* I new up an instance of `BlazorClient` (ignitor).
* I start issuing commands and perform my test.

Here are the things that I've added:

* Add support for sending dotnet calls from "JS"
* Add support for acknowledging render batches.
* Extract logic into a separate `BlazorClient` class so that it can be reused.
* Support non prerendered pages.
* Allow reacting to incoming JSInterop calls.
* Allow reacting to incoming RenderBatch.

### DIFF
--- a/src/Components/test/testassets/Ignitor/Error.cs
+++ b/src/Components/test/testassets/Ignitor/Error.cs
@@ -1,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Ignitor
+{
+    internal class Error
+        {
+            public string Stack { get; set; }
+        }
+}
--- a/src/Components/test/testassets/Ignitor/Error.cs
+++ b/src/Components/test/testassets/Ignitor/Error.cs
@@ -4,7 +4,7 @@
 namespace Ignitor
 {
     internal class Error
-        {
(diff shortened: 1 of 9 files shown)


src/Components/test/testassets/Ignitor/Error.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Ignitor
+{
+    internal class Error
+        {
```
NTaylorMullen: formatting

NTaylorMullen (approved):
Looks great!

> * Add support for sending dotnet calls from "JS"

I think I missed this part of the PR. Could you elaborate.

javiercn:
>Looks great!

>Add support for sending dotnet calls from "JS"
>I think I missed this part of the PR. Could you elaborate.

https://github.com/aspnet/AspNetCore/pull/11289/files#diff-ee4bf3b51cfe67e04c3fc29b8a44d0e7R124

This allows to call .net methods on the server from the client.

Issue #11267, opened 2019-06-16
Blazor Template Site CSS Breaks Bootstrap Tabs


xJonathanLEI:
### Describe the bug
The following lines from `site.css` breaks Bootstrap tabs:

    .nav-item:first-of-type {
        padding-top: 1rem;
    }

    .nav-item:last-of-type {
        padding-bottom: 1rem;
    }

Bootstrap tabs are broken as shown in the following screenshot:

<img width="235" alt="BootstrapTabBug" src="https://user-images.githubusercontent.com/19556359/59570139-22022100-90c6-11e9-97d5-69e73752ca08.png">

### To Reproduce
1. Create a client-side Blazor project with `dotnet new blazor`
2. Replace the body element in `wwwroot/index.html` with the following:

        <body>
            <nav>
                <div class="nav nav-tabs" id="nav-tab" role="tablist">
                    <a class="nav-item nav-link active" id="nav-home-tab" data-toggle="tab" href="#nav-home" role="tab"
                        aria-controls="nav-home" aria-selected="true">Home</a>
                    <a class="nav-item nav-link" id="nav-profile-tab" data-toggle="tab" href="#nav-profile" role="tab"
                        aria-controls="nav-profile" aria-selected="false">Profile</a>
                    <a class="nav-item nav-link" id="nav-contact-tab" data-toggle="tab" href="#nav-contact" role="tab"
                        aria-controls="nav-contact" aria-selected="false">Contact</a>
                </div>
            </nav>
            <div class="tab-content" id="nav-tabContent">
                <div class="tab-pane fade show active" id="nav-home" role="tabpanel" aria-labelledby="nav-home-tab">...</div>
                <div class="tab-pane fade" id="nav-profile" role="tabpanel" aria-labelledby="nav-profile-tab">...</div>
                <div class="tab-pane fade" id="nav-contact" role="tabpanel" aria-labelledby="nav-contact-tab">...</div>
            </div>
        </body>

3. Run the project

### Expected behavior
The tabs should have looked like this:

<img width="195" alt="BootstrapTabExpected" src="https://user-images.githubusercontent.com/19556359/59570202-fb90b580-90c6-11e9-8ee8-29692d3fafd4.png">

### Solution
Simply changed the aforementioned CSS lines to the following:

    li.nav-item:first-of-type {
        padding-top: 1rem;
    }

    li.nav-item:last-of-type {
        padding-bottom: 1rem;
    }

The problem goes away ;)

### Additional context
[...]

PR #11292, opened 2019-06-17
Fix CSS Bug in Blazor Template that Breaks Other Components

xJonathanLEI:
 - Added `.sidebar` selectors to `.nav-item` styles in CSS files

Addresses #11267


mkArtakMSFT:
Thanks for your effort.
Please resolve the merge conflicts so that we can move forward with this change.

mkArtakMSFT:
Hi. We're closing this issue as we have heard no response from you for some time now. If you have more details and are encountering this issue please add a new reply and re-open the issue.

PR #11293, opened 2019-06-17
Dylan/stack policy

DylanDmitri:
Added a new LIFO-based policy.

For servers under congestion, LIFO policies result in better outcomes. As an example, take a server with throughput for a million requests per second receiving two million requests per second. In all cases, the server will serve half the incoming requests and reject the other half. However, in the FIFO case all served requests must wait through the queue, while rejects happen instantly on arrival. In the LIFO case this is reversed: rejects have a delay added to them while page loads happen with minimal latency.

Microbenchmarks show:
 - under heavy async load: 1100 ns -> 1050 ns
 - singular async request: 1500 ns -> 1350 ns
 - singular sync request: 150 ns -> 100 ns

Biggest performance hit is likely allocating a `TaskCompletionSource` in an array for each queued request. Looking to optimize this in the future.


src/Middleware/RequestThrottling/src/Policies/StackPolicy.cs
```diff
@@ -0,0 +1,76 @@
+using System;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.AspNetCore.RequestThrottling.Policies
+{
+    class StackPolicy : IQueuePolicy
+    {
+        private TaskCompletionSource<bool>[] _buffer;
```
JamesNK: readonly

src/Middleware/RequestThrottling/src/Policies/StackPolicy.cs
```diff
@@ -0,0 +1,76 @@
...
+        private int _head; 
+        private int _queueLength;
+
+        private Task<bool> _trueTask = Task.FromResult(true);
```
JamesNK: static readonly

src/Middleware/RequestThrottling/src/Policies/StackPolicy.cs
```diff
@@ -0,0 +1,76 @@
...
+
+        private object _bufferLock = new Object();
```
JamesNK: readonly

src/Middleware/RequestThrottling/src/Policies/StackPolicy.cs
```diff
@@ -0,0 +1,76 @@
...
+
+        private int _freeServerSpots;
```
JamesNK: readonly

src/Middleware/RequestThrottling/src/Policies/StackPolicy.cs
```diff
@@ -0,0 +1,76 @@
+using System;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.AspNetCore.RequestThrottling.Policies
+{
+    class StackPolicy : IQueuePolicy
```
JamesNK: internal

src/Middleware/RequestThrottling/src/Policies/StackPolicy.cs
```diff
@@ -0,0 +1,76 @@
...
+
+        private Task<bool> _trueTask = Task.FromResult(true);
+
+        private object _bufferLock = new Object();
+
+        private int _freeServerSpots;
+
+        public StackPolicy(IOptions<TailDropOptions> options)
+        {
+            _buffer = new TaskCompletionSource<bool>[options.Value.RequestQueueLimit];
```
JamesNK: Should _buffer be a `List<T>` instead of an array? An array will immediately allocate up to the limit, and if someone specifies `TailDropOptions.RequestQueueLimit = int.MaxValue` then they're going to get an OutOfMemoryException.

DylanDmitri: My thought process was that I'm trying to optimize for the worst-case scenario. When the server is under maximum load, the entire buffer will be allocated anyways, so I would rather error early than during a traffic spike.

halter73: I agree that it makes sense to lazily grow the collection, and List is probably the best way to do that.

The List should be a lot better in cases with a large RequestQueueLimit but not many queued requests in practice. Even in the worst case when the queue fills up, a List shouldn't be that much less efficient than an array. After all, List is [backed by a single array](https://github.com/dotnet/corefx/blob/04e05d7d0a02f43ce5e6df455052dbfc0862dff7/src/Common/src/CoreLib/System/Collections/Generic/List.cs#L115) anyway. The array just gets reallocated [with double the previous capacity](https://github.com/dotnet/corefx/blob/04e05d7d0a02f43ce5e6df455052dbfc0862dff7/src/Common/src/CoreLib/System/Collections/Generic/List.cs#L410) and copied as the list grows. So there should be at most log(N) reallocations/copies where N is the RequestQueueLimit, and it should be as fast as an array once the queue stops growing.

src/Middleware/RequestThrottling/src/Policies/StackPolicy.cs
```diff
@@ -0,0 +1,77 @@
...
+            lock (_bufferLock)
+            {
+                if (_queueLength == 0)
+                {
+                    _freeServerSpots++;
+                    return;
+                }
+
+                // step backwards and launch a new task
+                if (_head-- == 0)
```
Tratcher: Split

src/Middleware/RequestThrottling/src/Policies/StackPolicy.cs
```diff
@@ -0,0 +1,77 @@
+using System;
+using System.Threading.Tasks;
+using System.Threading.Tasks.Sources;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.AspNetCore.RequestThrottling.Policies
+{
+    internal class StackPolicy : IQueuePolicy
```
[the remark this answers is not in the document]
DylanDmitri: going with StackQueuePolicy because just StackQueue makes me sad

src/Middleware/RequestThrottling/src/Policies/QueuePolicyServiceCollectionExtensions.cs
```diff
@@ -22,13 +22,20 @@ public static class QueuePolicyServiceCollectionExtensions
...
-        public static IServiceCollection AddTailDropQueue(this IServiceCollection services, Action<TailDropOptions> configure)
+        public static IServiceCollection AddTailDropQueue(this IServiceCollection services, Action<QueuePolicyOptions> configure)
         {
-            services.Configure<TailDropOptions>(configure);
+            services.Configure(configure);
             services.AddSingleton<IQueuePolicy, TailDrop>();
             return services;
         }
+
+        public static IServiceCollection AddStackPolicy(this IServiceCollection services, Action<QueuePolicyOptions> configure)
```
Tratcher:
```suggestion
        public static IServiceCollection AddStackQueue(this IServiceCollection services, Action<QueuePolicyOptions> configure)
```

DylanDmitri: oh dear

src/Middleware/RequestThrottling/src/QueuePolicies/StackQueuePolicy.cs
```diff
@@ -0,0 +1,104 @@
...
+            _maxConcurrentRequests = options.Value.MaxConcurrentRequests;
+            _freeServerSpots = options.Value.MaxConcurrentRequests;
+        }
+
+        public Task<bool> TryEnterAsync()
+        {
+
+            lock (_bufferLock)
+            {
+                if (_freeServerSpots > 0)
```
Kahbazi: Is it a good idea a this part is moved back to the middleware, so the `IQueuePolicy` only handles the queue? So if some one wants to implement a custom `IQueuePolicy` does not need to repeat this.

src/Middleware/RequestThrottling/src/QueuePolicies/StackQueuePolicy.cs
```diff
@@ -0,0 +1,104 @@
...
+                if (_head == 0)
+                {
+                    _head = _maxQueueCapacity - 1;
+                }
+                else
+                {
+                    _head--;
+                }
+
+                _buffer[_head].SetResult(true);
```
Kahbazi: Is setting `_buffer[_head] = null;` helps `GC` to collect `TaskCompletionSource<bool>` sooner rather than waiting for another request to replace `_buffer[_head]`?
```suggestion
                _buffer[_head].SetResult(true);
                _buffer[_head] = null;
```

src/Middleware/RequestThrottling/src/QueuePolicies/QueuePolicyOptions.cs
```diff
@@ -5,9 +5,9 @@
 namespace Microsoft.AspNetCore.RequestThrottling.Policies
 {
     /// <summary>
-    /// Specifies options for the <see cref="TailDrop"/>
+    /// Specifies options for the <see cref="TailDropQueuePolicy"/>
```
Kahbazi: Since this option is for any kind of `IQueuePolicy`, should this cref change?
```suggestion
    /// Specifies options for the <see cref="IQueuePolicy"/>
```

DylanDmitri: good catch, fixed

src/Middleware/RequestThrottling/src/QueuePolicies/StackQueuePolicy.cs
DylanDmitri: testing this now

src/Middleware/RequestThrottling/src/QueuePolicies/StackQueuePolicy.cs
DylanDmitri: I thought about doing this, and in an ideal world would split the logic this way. 

In the other implementation, a single SemaphoreSlim controls both flow into the server, and the queue on top. In this implementation, the queue is separate, and flow is control by this semaphore baked into the queue logic. I could go for more abstraction by moving this logic up to the middleware, but I'm trying to keep policy design open-ended and prevent extra lock congestion.

src/Middleware/RequestThrottling/src/QueuePolicies/StackQueuePolicy.cs
DylanDmitri: It makes the full queue benchmark slightly slower, and the one-at-a-time benchmark slightly faster. The effective change is that garbage collection happens as the queue empties, rather than as it fills back up. This shifts work earlier, and into times where the server has more capacity. Good change.

src/Middleware/RequestThrottling/sample/Startup.cs
```diff
@@ -14,40 +14,41 @@ namespace RequestThrottlingSample
...
             services.AddTailDropQueue((options) =>
             {
-                options.MaxConcurrentRequests = 4;
-                options.RequestQueueLimit = 0;
+                options.MaxConcurrentRequests = Math.Max(1, _config.GetValue<int>("maxCores"));
+                options.RequestQueueLimit = Math.Max(1, _config.GetValue<int>("maxQueue"));
             });
 
-            services.AddLogging();
+            //services.AddLogging();
```
halter73: Remove or uncomment.

src/Middleware/RequestThrottling/src/QueuePolicies/StackQueuePolicy.cs
halter73: @Kahbazi I like the way you think. [We've discussed this](https://github.com/aspnet/AspNetCore/pull/11167#discussion_r293175216), but we've decided that the extra purity isn't worth extra development effort yet.

One thing I don't like about leaving the concurrent request limit completely up to the middleware is that the IQueuePolicy could be left unaware how many requests the middleware will try to dequeue in the near future. If the middleware is always trying to dequeue every request all the time, that gives more control to the IQueuePolicy, but that forces the IQueuePolicy to manage a concurrent request limit itself.

src/Middleware/RequestThrottling/src/QueuePolicies/StackQueuePolicy.cs
```diff
@@ -0,0 +1,104 @@
...
+        {
+            _buffer = new List<TaskCompletionSource<bool>>();
+            _maxQueueCapacity = options.Value.RequestQueueLimit;
+            _maxConcurrentRequests = options.Value.MaxConcurrentRequests;
+            _freeServerSpots = options.Value.MaxConcurrentRequests;
+        }
+
+        public Task<bool> TryEnterAsync()
+        {
+
```
halter73:
```suggestion
```

src/Middleware/RequestThrottling/perf/Microbenchmarks/QueueFullOverhead.cs
```diff
@@ -11,19 +11,26 @@ namespace Microsoft.AspNetCore.RequestThrottling.Microbenchmarks
...
 
         [GlobalSetup]
         public void GlobalSetup()
         {
-            _middleware = TestUtils.CreateTestMiddleware_TailDrop(
+            _middleware_FIFO = TestUtils.CreateTestMiddleware_TailDrop(
+                maxConcurrentRequests: MaxConcurrentRequests,
+                requestQueueLimit: _numRequests,
+                next: IncrementAndCheck
+                );
```
halter73: I don't care a whole lot, but this is pretty non-standard for our codebase. Normally we'd just put this on the end of the previous line.

halter73 (approved):

src/Middleware/RequestThrottling/src/QueuePolicies/StackQueuePolicy.cs
halter73: Eventually, we'll probably pool a resettable TCS of sorts to avoid these GC cycles altogether.

src/Middleware/RequestThrottling/src/QueuePolicies/QueuePolicyServiceCollectionExtensions.cs
```diff
@@ -22,12 +22,19 @@ public static class QueuePolicyServiceCollectionExtensions
...
-        public static IServiceCollection AddTailDropQueue(this IServiceCollection services, Action<TailDropOptions> configure)
+        public static IServiceCollection AddTailDropQueue(this IServiceCollection services, Action<QueuePolicyOptions> configure)
         {
-            services.Configure<TailDropOptions>(configure);
-            services.AddSingleton<IQueuePolicy, TailDrop>();
+            services.Configure(configure);
+            services.AddSingleton<IQueuePolicy, TailDropQueuePolicy>();
+            return services;
+        }
+
```
Kahbazi:
```suggestion
        /// </summary>
        /// <param name="services">The <see cref="IServiceCollection"/> to add services to.</param>
        /// <param name="configure">Set the options used by the queue.
        /// Mandatory, since <see cref="QueuePolicyOptions.MaxConcurrentRequests"></see> must be provided.</param>
        /// <returns></returns>

```

DylanDmitri: added

src/Middleware/RequestThrottling/sample/Startup.cs
DylanDmitri: I had made several changes to the sample project, since I was running it on the benchmark server. Now that I'm adding real scenarios to the benchmarks repo, I reverted those changes here.

src/Middleware/RequestThrottling/src/Policies/StackPolicy.cs
DylanDmitri: There is now a list.

DylanDmitri:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=232772

DylanDmitri:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=232772&view=logs&jobId=1bca1137-cedc-5e44-af9b-a27ff406bb70

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/2225
https://github.com/aspnet/AspNetCore-Internal/issues/2314

src/Middleware/RequestThrottling/src/QueuePolicies/QueuePolicyOptions.cs
```diff
@@ -5,9 +5,9 @@
 namespace Microsoft.AspNetCore.RequestThrottling.Policies
```
Kahbazi:
```suggestion
namespace Microsoft.AspNetCore.RequestThrottling.QueuePolicies
```

src/Middleware/RequestThrottling/src/QueuePolicies/StackQueuePolicy.cs
Kahbazi: @halter73 Can we use `ValueTask<bool>` instead of `Task<bool>` and use `ManualResetValueTaskSource` for doing the reset?

src/Middleware/RequestThrottling/src/QueuePolicies/QueuePolicyOptions.cs
DylanDmitri: fixed, thank you

src/Middleware/RequestThrottling/src/QueuePolicies/StackQueuePolicy.cs
halter73: @Kahbazi That's the way we'd likely do it.

Kahbazi: @halter73 Can I do that? is this PR worthy?

PR #11300, opened 2019-06-17
Implement IHttpResetFeature in Kestrel

Tratcher:
#10886 We already had the ResetAndAbort functionality, we just needed a feature to expose it.

### DIFF
--- a/src/Http/Http.Features/src/IHttpResetFeature.cs
+++ b/src/Http/Http.Features/src/IHttpResetFeature.cs
@@ -8,11 +8,10 @@ namespace Microsoft.AspNetCore.Http.Features
     /// </summary>
     public interface IHttpResetFeature
     {
-
         /// <summary>
         /// Send a reset message with the given error code. The request will be considered aborted.
         /// </summary>
-        /// <param name="errorCode"></param>
+        /// <param name="errorCode">The error code to send in the reset message.</param>
         void Reset(int errorCode);
     }
 }
--- a/src/Http/Http.Features/src/IHttpResetFeature.cs
+++ b/src/Http/Http.Features/src/IHttpResetFeature.cs
@@ -1,0 +1,18 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Http.Features
+{
+    /// <summary>
+    /// Used to send reset messages for protocols that support them such as HTTP/2 or HTTP/3.
+    /// </summary>
+    public interface IHttpResetFeature
+    {
+
+        /// <summary>
+        /// Send a reset message with the given error code. The request will be considered aborted.
+        /// </summary>
+        /// <param name="errorCode"></param>
+        void Reset(int errorCode);
+    }
+}
--- a/src/Servers/Kestrel/Core/src/Internal/Http2/Http2Stream.FeatureCollection.cs
+++ b/src/Servers/Kestrel/Core/src/Internal/Http2/Http2Stream.FeatureCollection.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Connections;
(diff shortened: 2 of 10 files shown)


src/Servers/Kestrel/Core/src/Internal/Http2/Http2Stream.FeatureCollection.cs
```diff
@@ -73,5 +75,11 @@ async Task IHttpResponseCompletionFeature.CompleteAsync()
                 await ProduceEnd();
             }
         }
+
+        void IHttpResetFeature.Reset(int errorCode)
+        {
+            var abortReason = new ConnectionAbortedException(CoreStrings.FormatHttp2StreamResetByApplication((Http2ErrorCode)errorCode));
```
JamesNK: Do we want to validate the errorCode is a valid Http2ErrorCode?

src/Http/Http.Features/src/IHttpResetFeature.cs
```diff
@@ -0,0 +1,18 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Http.Features
+{
+    /// <summary>
+    /// Used to send reset messages for protocols that support them such as HTTP/2 or HTTP/3.
+    /// </summary>
+    public interface IHttpResetFeature
+    {
+
```
JamesNK: nit: liney

src/Http/Http.Features/src/IHttpResetFeature.cs
```diff
@@ -0,0 +1,18 @@
...
+        /// <summary>
+        /// Send a reset message with the given error code. The request will be considered aborted.
+        /// </summary>
+        /// <param name="errorCode"></param>
```
JamesNK: Docs?

src/Servers/Kestrel/Core/src/Internal/Http2/Http2Stream.FeatureCollection.cs
JamesNK: What happens if Reset is called multiple times for a request?

What happens if Reset is called after the request has already been aborted? (e.g. the client sends RST_STREAM)

Tratcher: No, they can add new ones at any time. Also the server (and in most cases the client) does not behave any differently based on the value.

Tratcher: ResetAndAbort is guarded against multiple invocations.
https://github.com/aspnet/AspNetCore/blob/9c4faf86e5534dba490dc0e2f634ae291aaa2cfd/src/Servers/Kestrel/Core/src/Internal/Http2/Http2Stream.cs#L452-L460

JamesNK:
Should this raise the RequestAborted cancellation token in TestServer when called?

Tratcher:
@JamesNK if we implemented it in TestServer... It's a specialized Abort.
That said, this is a weird one to implement in TestServer, there's no way to surface the reset to the client.

JamesNK:
Abort (and CompleteAsync) in TestServer should probably return the HttpResponseMessage to the client immediately instead of waiting for the request delegate to complete. That would be a non-minor change.

I think at least we can make it raise the CT like Abort does - https://github.com/aspnet/AspNetCore/pull/9975

Tratcher:
TestServer should already be returning the response message when you write/flush the response body. Making CompleteAsync work on top of that should be trivial, it just needs to additionally close the response body.

Abort on TestServer should throw on the client, not return a response message.

I'd be tempted to only expose IHttpResetFeature in TestServer for request that set the request protocol to HTTP/2.

IHttpResetFeature in TestServer behaving the same as Abort seems like a good start.

I can do a follow-up PR for TestServer CompleteAsync and Reset.

JamesNK (approved):

halter73 (approved):

Issue #4412, opened 2018-12-05
ParameterAttribute Inherited = true


nenaaki:
Can 'Inherited' of ParameterAttribute be set to 'true'?
It is useful to be able to use parameters with inherited classes.

mkArtakMSFT:
Thanks for contacting us, @nenaaki.

mkArtakMSFT:
@nenaaki, feel free to send a PR for this.

mkArtakMSFT:
Closing as done.

PR #11343, opened 2019-06-18
Set Inherited = true for ParameterAttribute

Kahbazi:
#4412


rynowak:
@Kahbazi - can you add an E2E test that relies on this change? I want to make sure that it really works, including intellisense, etc.

mkArtakMSFT:
Thanks for your effort. We're closing this PR as this change has already been done as part of some other work.

PR #11345, opened 2019-06-18
Generate .nupkg files that contain aspnetcore-runtime MSI for insertion into Visual Studio

jkotalik:
Part of https://github.com/aspnet/AspNetCore/issues/11071

Packages will be put in the NonShipping folder and uploaded to the blob store. 

@johnbeisner what do we need to do to verify this E2E?


src/Installers/Windows/SharedFramework/SharedFrameworkPackage.nuspec
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
+    <metadata>
+        <id>VS.Redist.Common.NetCore.Toolset.$ARCH$</id>
+        <version>$MAJOR_VERSION$.$MINOR_VERSION$</version>
```
jkotalik: I'm concerned about this version here as nuget will add a patch version after, making the package name VS.Redist.Common.AspNetCore.TargetingPack.x64.3.0.0.nupkg and VS.Redist.Common.NetCore.Toolset.x64.3.0.0.nupkg. Should we have patch/preview version numbers here @johnbeisner ?

src/Installers/Windows/GenerateNugetPack.ps1
```diff
@@ -0,0 +1,18 @@
+# Copyright (c) .NET Foundation and contributors. All rights reserved.
+# Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+param(
```
dougbu: Not a blocker but I'd rather have such a generic script in eng/scripts/

src/Installers/Windows/Wix.props
```diff
@@ -25,6 +25,7 @@
   <PropertyGroup>
     <DefineConstants>$(DefineConstants);files=$(MSBuildThisFileDirectory)files</DefineConstants>
     <DefineConstants>$(DefineConstants);Culture=$(Cultures)</DefineConstants>
+    <GenerateNupkgPowershellScript>$(RepoRoot)\src\Installers\Windows\GenerateNugetPack.ps1</GenerateNupkgPowershellScript>
```
dougbu: Where is this copy of the property used?

dougbu (approved):
Looks good but please wait to hear back from @johnbeisner before merging

src/Installers/Windows/Wix.props
jkotalik: https://github.com/aspnet/AspNetCore/pull/11345/files#diff-d09a1ae45b24853c16c713e250339fd3R89

src/Installers/Windows/GenerateNugetPack.ps1
jkotalik: It isn't very generic, they are both used for specifically generating a nuget file with mis/cabs. I'll rename the script though to be more specific.

johnbeisner:
@jkotalik
You should go to the VS feed and download/inspect at the existing Asp.NET NuPkgs for VS: 
_https://devdiv.visualstudio.com/DevDiv/_packaging?_a=feed&feed=VS_

If you filter on "VS.Redist.Common.AspNet" you will see your team's VS NuPkgs - I feel we should stay with these naming/ID/versioning conventions at least thru the first insertion of 3.0 SDK into VS.

johnbeisner:
@jkotalik
"what do we need to do to verify this E2E?"
1. produce an official Asp.NET build with the NuPkgs
2. upload the four 3.0 NuPkgs to the VS feed
3. Modify the current VS insertion PR to consume these new Asp.NET NuPkgs
_https://devdiv.visualstudio.com/DevDiv/_git/VS/pullrequest/186888_

jkotalik:
Need to figure out how to ignore signing this package.

johnbeisner:
@jkotalik 
Pretty sure the MSIs need to be signed.

joeloff:
The nupkg is just a delivery mechanism for getting external binaries into the VS build. As long as the MSIs are signed, we're good. The nupkg signature doesn't matter. It can either be signed or unsigned.

jkotalik:
Chatted with @joeloff offline. The SignTool only signs nuget packages in the Build Windows Installers (not msis). Msis are signed in a separate step within the wixprojs themselves. This is the first time we are including a msi within a nuget package, so we should ignore trying to sign it. Also, we shouldn't attempt to sign it too as then we could have multiple msis with the same binaries with different signatures. 

I'll verify that the contents are in fact signed. Unfortunately, this either would mean I need to make an internal PR or push and verify after I pushed.

src/Installers/Windows/SharedFramework/SharedFrameworkPackage.nuspec
```diff
@@ -0,0 +1,19 @@
...
+    <metadata>
+        <id> VS.Redist.Common.AspNetCore.SharedFramework.$ARCH$.$MAJOR$.$MINOR$</id>
+        <version>1.0.0</version>
+        <title> VS.Redist.Common.AspNetCore.SharedFramework.$ARCH$.$MAJOR$.$MINOR$</title>
+        <authors>Microsoft</authors>
+        <owners>Microsoft</owners>
+        <licenseUrl>https://www.microsoft.com/net/dotnet_library_license.htm</licenseUrl>
+        <projectUrl>https://github.com/aspnet/aspnetcore</projectUrl>
+        <requireLicenseAcceptance>true</requireLicenseAcceptance>
+        <description>3.0 ASP.NET Core TargetingPack (x86) Windows Installer MSI as a .nupkg for internal Visual Studio build consumption</description>
```
johnbeisner: Nit: the "3.0" should be from $MAJOR$.$MINOR$

src/Installers/Windows/TargetingPack/TargetingPackPackage.nuspec
```diff
@@ -0,0 +1,18 @@
...
+    <metadata>
+        <id>VS.Redist.Common.AspNetCore.TargetingPack.$ARCH$.$MAJOR$.$MINOR$</id>
+        <version>1.0.0</version>
+        <title> VS.Redist.Common.AspNetCore.TargetingPack.$ARCH$.$MAJOR$.$MINOR$</title>
+        <authors>Microsoft</authors>
+        <owners>Microsoft</owners>
+        <licenseUrl>https://www.microsoft.com/net/dotnet_library_license.htm</licenseUrl>
+        <projectUrl>https://github.com/aspnet/aspnetcore</projectUrl>
+        <requireLicenseAcceptance>true</requireLicenseAcceptance>
+        <description>3.0 ASP.NET Core TargetingPack (x64) Windows Installer MSI as a .nupkg for internal Visual Studio build consumption</description>
```
johnbeisner: Nit: same as above...

ghost (approved):

src/Installers/Windows/TargetingPack/TargetingPackPackage.nuspec
```diff
@@ -0,0 +1,18 @@
...
+    <metadata>
+        <id>VS.Redist.Common.AspNetCore.TargetingPack.$ARCH$</id>
+        <version>$MAJOR_VERSION$.$MINOR_VERSION$</version>
+        <title> VS.Redist.Common.AspNetCore.TargetingPack.$ARCH$</title>
+        <authors>Microsoft</authors>
+        <owners>Microsoft</owners>
+        <licenseUrl>https://www.microsoft.com/net/dotnet_library_license.htm</licenseUrl>
+        <projectUrl>https://github.com/aspnet/aspnetcore</projectUrl>
+        <requireLicenseAcceptance>true</requireLicenseAcceptance>
+        <description>.NET Core SDK Toolset ($ARCH$) Windows Installer MSI as a .nupkg for internal Visual Studio build consumption</description>
```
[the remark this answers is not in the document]
jkotalik: I think the three comments above were all outdated.

johnbeisner:
@jkotalik 
I do not see the NuPkgs at blob storage - could I be looking at the wrong versions?

jkotalik:
I see it being uploaded. It should be on the aspnet-aspnetcore feed. Example build version number: VS.Redist.Common.AspNetCore.SharedFramework.x64.3.0.3.0.0-preview7.19351.4.nupkg and internal build.

Issue #9556, opened 2019-04-19
Inconsistent "wrong password" errors in identity UI


Eilon:
The error message for a bad password in `https://localhost:XYZ/Identity/Account/Manage/ChangePassword` is `Incorrect password.`

![image](https://user-images.githubusercontent.com/202643/56433127-ba298880-6284-11e9-927d-81a363a5e2f9.png)

But the error message for a bad password in `https://localhost:XYZ/Identity/Account/Manage/DeletePersonalData` is `Password not correct.`

![image](https://user-images.githubusercontent.com/202643/56433181-e218ec00-6284-11e9-8226-4220788bb944.png)

These should at least be consistent, because they're the same error. And the first one (`Incorrect password`) sounds better, so let's use that.

HaoK:
Merged thanks wenz!

PR #11346, opened 2019-06-18
Make incorrect password error message consistent (#9556)

HaoK:
Rebased version of https://github.com/aspnet/AspNetCore/pull/9561


HaoK:
FIxes Addresses #9556

PR #11355, opened 2019-06-18
Skip CORS tests

jkotalik:
I think my PR caused tests to hang indefinitely... for example: https://dev.azure.com/dnceng/public/_build/results?buildId=229386 is hanging forever. Disabling for now and will investigate when I have time why jest isn't being stopped.


JunTaoLuo (approved):

anurse:
Force-merging to unblock the build.

Issue #8204, opened 2019-03-05
Components bind-value-oninput can lose cursor position with fast typing


SteveSandersonMS:
Repro:

 * `<input bind-value-oninput="@SomeString" />`
 * Type `X` into the input
 * Position the cursor at the start (before the `X`)
 * Mash the keyboard

Expected: Keystrokes typed should appear only at the cursor position, so the `X` should remain at the end

Actual: A few keystrokes appear before the `X`, but then the cursor jumps to the end, and subsequent keystrokes are placed after the `X`

I think this happens because of how `bind` does a 2-way binding. On each keystroke, it not only (1) reads the updated value and writes it to `SomeString`, but also (2) tries to re-apply the value from `SomeString` to the input. When it does (2), as long as the value in the input already matches the string, it's a no-op and gets skipped, so normally we don't do anything that moves the cursor.

*However*, because of the asynchrony with Razor Components, there might be more keystrokes between (1) and (2), so the value we're applying in (2) no longer matches the contents of the input. We don't lose the keystrokes (we already did some work on that), but the fact that `input.value !== SomeString` means we do write both the previous and then current value to the input. The fact that we're no longer no-opping this write means now the cursor implicitly moves to the end.

**Suggested fix:** In `BrowserRenderer.ts`, in `tryApplyValueProperty`, we should add more logic to preserve the selectionStart/selectionEnd properties of input/textarea when assigning value. This would also improve things in other scenarios, like if you had a collaborative textarea bound to a value that multiple users could edit simultaneously.

SteveSandersonMS:
This issue was assigned to me so I would write up some thoughts on resolution, not with the expectation of a fix in preview 6. So here are my thoughts.

@javiercn has already written up some thoughts on the general issue, but in this post I want summarise it from the point of view of typing, since that's probably the worst case.

### The problem

This is a general problem affecting any case where the DOM can get out of sync with the underlying rendertree (e.g., because the user modified the state of a form input).

But it's worse for server-side Blazor, because given the async-ness, there are extra opportunities to be out of sync due to time spent in flight. Common example which is at the root of this issue here:

1. There's an `<input bind-value-oninput=@MyProperty />` where `MyProperty` is just a normal `{ get; set; }` property.
2. User presses "a"; client sends "a" to server
3. User presses"b"; client sends "ab" to server
[...]

PR #11375, opened 2019-06-19
On RenderTreeBuilder, add SetUpdatesAttributeName

SteveSandersonMS:
This is part of #8204

Because aspnetcore-tooling is a separate repo, we need a three-phase commit:

 1. First we add `SetUpdatesAttributeName` in this repo
 2. Then we can merge the PR in aspnetcore-tooling that generates calls to `SetUpdatesAttributeName`. If we didn't do (1) first, then we'd start getting build errors when the aspnetcore-tooling changes were received here.
 3. Then we can implement the behavior for `SetUpdatesAttributeName` and it will actually be used.

### DIFF
--- a/src/Components/Components/src/RenderTree/RenderTreeBuilder.cs
+++ b/src/Components/Components/src/RenderTree/RenderTreeBuilder.cs
@@ -517,6 +517,24 @@ namespace Microsoft.AspNetCore.Components.RenderTree
         }
 
         /// <summary>
+        /// <para>
+        /// Indicates that the preceding attribute represents an event handler
+        /// whose execution updates the attribute with name <paramref name="updatesAttributeName"/>.
+        /// </para>
+        /// <para>
+        /// This information is used by the rendering system to determine whether
+        /// to accept a value update for the other attribute when receiving a
+        /// call to the event handler.
+        /// </para>
+        /// </summary>
+        /// <param name="updatesAttributeName">The name of another attribute whose value can be updated when the event handler is executed.</param>
+        public void SetUpdatesAttributeName(string updatesAttributeName)
+        {
+            // TODO: This will be implemented in a later PR, once aspnetcore-tooling
+            // is updated to call this method.
+        }
+
+        /// <summary>
         /// Appends a frame representing a child component.
         /// </summary>
         /// <typeparam name="TComponent">The type of the child component.</typeparam>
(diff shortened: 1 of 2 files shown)


javiercn (approved):

src/Components/Components/src/RenderTree/RenderTreeBuilder.cs
```diff
@@ -516,6 +516,24 @@ public void AddMultipleAttributes<T>(int sequence, IEnumerable<KeyValuePair<stri
...
+        /// whose execution updates the attribute with name <paramref name="updatesAttributeName"/>.
+        /// </para>
+        /// <para>
+        /// This information is used by the rendering system to determine whether
+        /// to accept a value update for the other attribute when receiving a
+        /// call to the event handler.
+        /// </para>
+        /// </summary>
+        /// <param name="updatesAttributeName">The name of another attribute whose value can be updated when the event handler is executed.</param>
+        public void SetUpdatesAttributeName(string updatesAttributeName)
```
pranavkm: Need to regen the reference sources

src/Components/Components/src/RenderTree/RenderTreeBuilder.cs
```diff
@@ -516,6 +516,24 @@ public void AddMultipleAttributes<T>(int sequence, IEnumerable<KeyValuePair<stri
...
 
+        /// <summary>
+        /// <para>
+        /// Indicates that the preceding attribute represents an event handler
+        /// whose execution updates the attribute with name <paramref name="updatesAttributeName"/>.
+        /// </para>
+        /// <para>
+        /// This information is used by the rendering system to determine whether
+        /// to accept a value update for the other attribute when receiving a
+        /// call to the event handler.
```
rynowak: @SteveSandersonMS - are you concerned that this won't really be possible to do with *manual* two-way binding?

SteveSandersonMS: Yes, and I'd be happy to discuss this whole design with you next week.

My thought was that later on, we'd also add another parameter to event handlers, e.g., `@oninput:updateField="value"` so that it is possible to do this with a manual two-way binding. It would be optional, since it's up to you whether or not the binding enforces consistency synchronously when the event is processed.

I also thought it would be acceptable for this to be a `@bind`-only feature for preview 7, and maybe longer, since adding `@onevent:updateField` is not urgent.

Issue #4050, opened 2018-11-16
BLAZOR API review


danroth27:
**Edit: @rynowak hijacking top post for great justice**

# Framing Blazor API reviews

I'm suggesting that we bucket API review work into areas based on the user-impact and expected usage of APIs, and from there prioritize and time-box some of the decision making. I'm hoping to use this document to paint a high-level picture rather than "do the work" for everything at once.

I don't think it's super reasonable to have us sit in a room and "do a pass" given how many of these APIs we've built up over time - the discussion would quickly get unfocused. I want everyone's input and involvement in making the first release of Blazor awesome - so I'm going to suggest that we do initial reviews and discussions on github or email and then have meetings as-needed.

Here's a quick shout-out for all of the ideas that were explicitly mentioned in the github [issue](https://github.com/aspnet/AspNetCore/issues/4050). 
 - Review all of our startup API for both client-side and server-side
 - Startup Code (various suggestions)
 - Mark all attributes sealed (and review `AttributeUsage`)
 - Develop a *style-guide* for components and make the built-in ones conform
 - Figure out what to do with our HttpClient extension methods
 - Do we want to develop a naming convention for types that are intended for JS interop only (like is commonly done with P/Invoke)

 It's been a long time since some of us started work on Blazor (over a year and a half for Steve). We should look at the details of everything important with fresh eyes, and make sure it's as good as it can be. We've gone up until this point doing *the best we can right now*, but we're about to cross the threshold where all of these decisions become **permanent**. It's scary! But at least we're doing it together.

## What are we doing?

I think there are probably two beneficiaries of an API review - us, and unsurprisingly our users. While it might seem like we want different things, I think we usually want the same things in this area. If we can craft APIs with good ergonomics, it's easy to talk about with customers, or write documentation for. If we create good APIs it's easy to scale them up as we work on the framework.

- Simplicity (few namespaces, overloads you want are easy to find, avoid tricky patterns)
- Consistency (predictability, portability)
- Horizontal Scalability (API and overloads can grow as functionality is added)

Often this means that we need to do hard work so that users have a good experience. If you've worked on MVC for a while I'm sure you can think of a few examples where we wish we did better :wink:.

## Areas of focus

Areas are in rough order of priority and attempt to call out of some of things that are important to review and decide. Please feel free to suggest your own ideas or disagreements - but keep in mind that we will have to time-box this work (both reviewing and changing).

I'm not attempting to make a list of everything that has to be reviewed. 

[...]

PR #11376, opened 2019-06-19
Javiercn/blazor improved circuit is

javiercn:
* Makes the circuit id a csrf token:
  * We generate a key with a PRNG.
  * We produce a request token by hashing the key.
  * We produce a cookie token by dataprotecting the key.
    * The cookie name includes a prefix of the request token to distinguish the cookie among all possible options.
    * If this is a problem, we can attach an additional prefix to both the request token and cookie for matching purposes.
* Uses an endpoint policy matcher that attaches an identity with a claim to the request principal.
  * SignalR service copies a limited amount of information to pass in.
  * The claims from the existing principal are among those pieces of information.
  * The policy matcher wraps the hub endpoints, and gets applied after the signalr service matcher.
  * We use the circuit id to compare it against the claim we setup in the principal.
* Updates the bootstrapping process to limit 1 circuit per connection.
  * Prerendered and non-prerendered components render on the same circuit.
  * When there are prerendered components, we reconnect to the circuit first and then render non prerendered components.
  * When there are no prerendered components we simply create a new circuit and render the non prerendered components.

Pending
- [x] Cleanup stale cookies.
- [x] Use an ephemeral data protector.
- [x] Reject cross-origin requests when generating cookie.
- [x] Set cache headers when setting the cookie.
- [x] Unit tests
- [x] Hide the request for generating a new circuit id inside the signalr connection protocol.
  - We can't hide the call to generate a new circuit id within the hub protocol due to the fact that the client needs to get a circuit id back it can reuse for reconnection and to start the circuit in the first place.


eng/Dependencies.props
```diff
@@ -41,6 +41,7 @@ and are generated based on the last package release.
     <LatestPackageReference Include="Microsoft.AspNetCore.Razor.Language" Version="$(MicrosoftAspNetCoreRazorLanguagePackageVersion)" />
     <LatestPackageReference Include="Microsoft.AspNetCore.Testing" Version="$(MicrosoftAspNetCoreTestingPackageVersion)" />
     <LatestPackageReference Include="Microsoft.Azure.KeyVault" Version="$(MicrosoftAzureKeyVaultPackageVersion)" />
+    <LatestPackageReference Include="Microsoft.Azure.SignalR" Version="1.1.0-preview1-10426" />
```
javiercn: Included here so that its easier to test with the rest of the repository.

src/Components/Server/src/Authentication/CircuitAuthenticationHandler.cs
```diff
@@ -0,0 +1,206 @@
...
+using Microsoft.Extensions.Primitives;
+using Microsoft.Net.Http.Headers;
+
+namespace Microsoft.AspNetCore.Components.Server
+{
+    // Authentication handler for circuit ids.
+    // We pair each circuit id with a cookie.
+    // We use an authentication handler to enable smooth integration with signalr service, as it preserves limited
+    // information about the connection context (amongs it the Principal), that's why we do it that way.
+    internal class CircuitAuthenticationHandler : SignInAuthenticationHandler<CircuitAuthenticationOptions>
```
Tratcher: This is the weirdest authentication handler I have ever seen.
- No challenge.
- Signin ignores the user
- No signout

How exactly does this "enable smooth integration with signalr service"?

How well does this stack when signalr really is using auth?

javiercn: > This is the weirdest authentication handler I have ever seen.
> 
> * No challenge.

What would the challenge would be? Simply return a 401?

> * Signin ignores the user

We are not authenticating a specific user but a specific circuit id. I could have build a principal and passed in the properties through claims on it. That might be more "correct".

> * No signout

We might implement that if we end up offering an endpoint to close the circuit.


> How exactly does this "enable smooth integration with signalr service"?

The information flows transparently through the principal whether you are using signalr service or not. Signalr service copies a limited amount of information, if we were to do something different it would require them to do something specific for blazor, which we want to avoid.

> How well does this stack when signalr really is using auth?

It stacks fine. There are two identities, one for whatever auth system you use and another one for the connection. When using the singnalr service all identities get merged. The hub authentication is based on a policy and by default it gets configured to use the default scheme + our scheme, resulting in what you have configured + our scheme. Its also available for you to tweak it to your heart's contents.

javiercn: For the record, I'm killing the auth hander as I found a better approach

src/Components/Browser.JS/src/Boot.Server.ts
```diff
@@ -51,9 +53,15 @@ async function boot(userOptions?: Partial<BlazorOptions>): Promise<void> {
...
-
   const circuits = discoverPrerenderedCircuits(document);
+  if (circuits.length > 1){
+    throw new Error('Can\'t have multiple circuits per connection');
+  }
+
+  const circuitId = circuits.length > 0 ? circuits[0].circuitId : await getNewCircuitId(options);
+
+  const initialConnection = await initializeConnection(options, circuitHandlers, circuitId, logger);
+
```
javiercn: Further cleanup can be done to this file, but I'm not going to do it in this PR

eng/Dependencies.props
SteveSandersonMS: When this is merged, I'm guessing you'll remove this, and ensure the CTI scripts give us adequate coverage of the SignalR Service scenarios?

src/Components/Browser.JS/src/Boot.Server.ts
```diff
@@ -89,13 +91,15 @@ async function boot(userOptions?: Partial<BlazorOptions>): Promise<void> {
...
-    circuits.push(circuit);
-  }
+  // We render any additional component after all prerendered components have
+  // re-stablished the connection with the circuit.
+  const renderedComponents = await startCircuit(circuitId, initialConnection);
 
-  await reconnectTask;
+  if (!renderedComponents) {
+    logger.log(LogLevel.Information, 'No preregistered components to render.');
+  }
```
SteveSandersonMS: It's very difficult to figure out by reading the PR diff how this is going to interact with reconnection cases. Do you feel we have adequate E2E (or if needed CTI) test coverage of reconnection?

src/Components/Server/ref/Microsoft.AspNetCore.Components.Server.netcoreapp3.0.cs
```diff
@@ -11,6 +11,8 @@ public static partial class ComponentEndpointRouteBuilderExtensions
     {
         public static Microsoft.AspNetCore.Components.Server.ComponentEndpointConventionBuilder MapBlazorHub(this Microsoft.AspNetCore.Routing.IEndpointRouteBuilder endpoints) { throw null; }
         public static Microsoft.AspNetCore.Components.Server.ComponentEndpointConventionBuilder MapBlazorHub(this Microsoft.AspNetCore.Routing.IEndpointRouteBuilder endpoints, System.Action<Microsoft.AspNetCore.Http.Connections.HttpConnectionDispatcherOptions> configureOptions) { throw null; }
+        public static Microsoft.AspNetCore.Components.Server.ComponentEndpointConventionBuilder MapBlazorHub(this Microsoft.AspNetCore.Routing.IEndpointRouteBuilder endpoints, string path) { throw null; }
+        public static Microsoft.AspNetCore.Components.Server.ComponentEndpointConventionBuilder MapBlazorHub(this Microsoft.AspNetCore.Routing.IEndpointRouteBuilder endpoints, string path, System.Action<Microsoft.AspNetCore.Http.Connections.HttpConnectionDispatcherOptions> configureOptions) { throw null; }
```
SteveSandersonMS: As an API review point, we should consider wrapping both the `path` and `HttpConnectionDispatcherOptions` into some new type `BlazorHubOptions` so that we're free to keep adding more options in the future without an exponentially increasing number of overload combinations. Adding a note to #4050

src/Components/Server/src/CircuitOptions.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using Microsoft.AspNetCore.Authorization;
```
SteveSandersonMS: This line probably isn't required on its own.

src/Components/Browser.JS/src/Boot.Server.ts
```diff
@@ -104,13 +108,23 @@ async function boot(userOptions?: Partial<BlazorOptions>): Promise<void> {
...
 
-async function initializeConnection(options: Required<BlazorOptions>, circuitHandlers: CircuitHandler[], logger: ILogger): Promise<signalR.HubConnection> {
+async function getNewCircuitId(options: BlazorOptions): Promise<string> {
+  const response = await fetch(`${options.serviceUrl}/start`, {
+    method: 'POST',
+  });
+
+  const responseBody = await response.json() as { id: string };
+
+  return responseBody.id;
```
SteveSandersonMS: In the TS-side code, could we rename all references to "circuit ID" to something like `circuitIdRequestToken`, or whatever its more accurate name now is? Otherwise it's hard to keep track of what aspect of the circuit ID we're sharing with the client.

src/Components/Server/src/Circuits/CircuitHost.cs
```diff
@@ -149,8 +149,15 @@ public async Task InitializeAsync(CancellationToken cancellationToken)
                     SetCurrentCircuitHost(this);
                     _initialized = true; // We're ready to accept incoming JSInterop calls from here on
 
-                    await OnCircuitOpenedAsync(cancellationToken);
-                    await OnConnectionUpAsync(cancellationToken);
+                    if (!existingCircuit)
+                    {
+                        // If the circuit we are using was created during prerendering don't run
+                        // the lifecycle events again at this time.
```
SteveSandersonMS: Was this a bug before? I'm surprised we need to change that behavior unless it was a bug before.

SteveSandersonMS: I am a bit uncertain about the meanings of these lifecycle events now. Previously I would have expected:

 * `OnCircuitOpenedAsync` to occur during prerendering
 * `OnConnectionUpAsync` to occur after the client opens the socket connection and is ready for interactivity

... but it seems with this change that there wouldn't be a way to know when the socket connection was opened. Is this right? I'm not sure we've got a formal definition of when these lifecycle events should fire, but now would be a good time to be explicit about it.

src/Components/Server/src/Circuits/CircuitIdFactory.cs
```diff
@@ -2,56 +2,114 @@
...
+        }
+
+        private static byte[] GetHash(byte[] buffer)
+        {
+            var hashAlgorithm = SHA256.Create();
+            var hash = hashAlgorithm.ComputeHash(buffer);
+            return hash;
+        }
+
+        public bool ValidateCircuitId(string circuitId, string cookie, out CircuitId id)
```
SteveSandersonMS: Nit: I find the namings of the `circuitId` parameter and the locals in this method hard to follow. It's not obvious to me that `circuitId` should mean "the token from the querystring but not the whole ID" or `rawRequestId` should mean "the bit we got from the querystring" or that `id` should mean "from the cookie". I'd find something like the following easier to interpret at a glance:

```cs
var circuitIdFromCookie = FromCookieValue(cookie);
var requestTokenFromQueryStringBytes = Base64UrlTextEncoder.Decode(requestTokenFromQueryString);
var requestTokenFromCookieBytes = Base64UrlTextEncoder.Decode(circuitIdFromCookie.RequestToken);
if (CryptographicOperations.FixedTimeEquals(requestTokenFromQueryStringBytes, requestTokenFromCookieBytes))
```

Otherwise, interpreting the meanings involves tracing backwards to remember how each data item was derived.

src/Components/Server/src/Circuits/CircuitIdFactory.cs
```diff
@@ -2,56 +2,114 @@
...
+            }
+            else
+            {
+                id = default;
+                return false;
+            }
         }
 
-        public bool ValidateCircuitId(string circuitId)
+        public bool ValidateCircuitId(string circuitId, ClaimsPrincipal user)
```
SteveSandersonMS: Out of interest, have we got to the point of verifying that the current user's "name" claim matches the existing one on the circuit when reconnecting? Or is that a future enhancement?

src/Components/Server/src/Circuits/CircuitIdFactory.cs
```diff
@@ -2,56 +2,114 @@
...
+
+        internal CircuitId FromCookieValue(string value)
+        {
+            var payload = Base64UrlTextEncoder.Decode(value);
+            var id = _protector.Unprotect(payload);
+            return new CircuitId(id, GetHash(id), payload);
+        }
+    }
+
+    internal struct CircuitId
```
SteveSandersonMS: Looks like this could be `readonly` for clarity

src/Components/Server/src/Circuits/CircuitRegistry.cs
```diff
@@ -105,9 +132,14 @@ public virtual Task DisconnectAsync(CircuitHost circuitHost, string connectionId
             return circuitHandlerTask;
         }
 
+        internal bool ContainsCircuit(string circuitId)
```
SteveSandersonMS: Calling it `circuitId` here seems really confusing. The dictionary key is now the request token, not the circuit ID, isn't it?

src/Components/Server/src/Circuits/DefaultCircuitFactory.cs
```diff
@@ -4,7 +4,10 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Security.Claims;
+using System.Security.Cryptography.Xml;
 using System.Text.Encodings.Web;
+using Microsoft.AspNetCore.Authentication;
```
SteveSandersonMS: Not sure the changes in this file are needed

src/Components/Server/src/ComponentHub.cs
```diff
@@ -65,10 +68,28 @@ public override Task OnDisconnectedAsync(Exception exception)
...
+        //   * OnConnectionUpAsync runs before all components get rendered.
+        //     * Will run inside ConnectCircuit
+        // If there were only prerendered components on the page start circuit will do nothing.
+        // If there were no prerendered components on the page:
+        //   * A new circuit will be created with the given circuit id.
+        public string StartCircuit(string circuitId, string uriAbsolute, string baseUriAbsolute)
         {
             var circuitClient = new CircuitClientProxy(Clients.Caller, Context.ConnectionId);
-            if (DefaultCircuitFactory.ResolveComponentMetadata(Context.GetHttpContext(), circuitClient).Count == 0)
+            var preregisterredDescriptors = DefaultCircuitFactory.ResolveComponentMetadata(
```
SteveSandersonMS: Nit: typo
```suggestion
            var preregisteredDescriptors = DefaultCircuitFactory.ResolveComponentMetadata(
```

src/Components/Server/src/ComponentHub.cs
```diff
@@ -79,33 +100,57 @@ public string StartCircuit(string uriAbsolute, string baseUriAbsolute)
...
             // Fire-and-forget the initialization process, because we can't block the
             // SignalR message loop (we'd get a deadlock if any of the initialization
             // logic relied on receiving a subsequent message from SignalR), and it will
             // take care of its own errors anyway.
-            _ = circuitHost.InitializeAsync(Context.ConnectionAborted);
 
-            _circuitRegistry.Register(circuitHost);
+            // For existing circuits we will simply trigger the rendering of non prerendered
+            // components. Otherwise we do a full initialization. Once the reconnection gets
+            // stablished the rest of the lifecycle events will run and the circuit will
```
SteveSandersonMS:
```suggestion
            // established the rest of the lifecycle events will run and the circuit will
```

src/Components/Server/src/DependencyInjection/ComponentServiceCollectionExtensions.cs
```diff
@@ -47,6 +49,9 @@ public static IServerSideBlazorBuilder AddServerSideBlazor(this IServiceCollecti
                 options.SupportedProtocols.Add(BlazorPackHubProtocol.ProtocolName);
             });
 
+            // Matcher policy to configure csrf protection on the negotiate endpoint
```
SteveSandersonMS: It's not really CSRF, is it? My understanding was that we're protecting against circuit ID exfiltration via XSS.

src/Components/Server/src/Endpoints/CircuitMatcherPolicy.cs
```diff
@@ -0,0 +1,232 @@
...
+using Microsoft.Net.Http.Headers;
+
+namespace Microsoft.AspNetCore.Components.Server
+{
+    internal class CircuitMatcherPolicy : MatcherPolicy, IEndpointSelectorPolicy
+    {
+        internal const string IdClaimType = "bcid";
+        internal const string CookiePrefix = "Circuit.";
+
+        private const int PrefixLenght = 4;
```
SteveSandersonMS:
```suggestion
        private const int PrefixLength = 4;
```

SteveSandersonMS: Also, are we relying on non-collision of the IDs up to this prefix length for a given user? If so we might be better adding another couple of characters to the prefix length.

(e.g., if these are base64 chars, the 4-char prefix space size is 64^4=16 million)

src/Components/Server/src/Endpoints/CircuitMatcherPolicy.cs
```diff
@@ -0,0 +1,232 @@
...
+using Microsoft.AspNetCore.Routing.Matching;
+using Microsoft.AspNetCore.SignalR;
+using Microsoft.Extensions.Primitives;
+using Microsoft.Net.Http.Headers;
+
+namespace Microsoft.AspNetCore.Components.Server
+{
+    internal class CircuitMatcherPolicy : MatcherPolicy, IEndpointSelectorPolicy
+    {
+        internal const string IdClaimType = "bcid";
```
SteveSandersonMS: Maybe I'm being overly-cautious here, but might it be valuable to give this a more unique name? It's not hard to imagine that people might try using the claim name `bcid` to mean something completely different in their application.

src/Components/Server/src/Endpoints/CircuitMatcherPolicy.cs
```diff
@@ -0,0 +1,232 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
```
SteveSandersonMS: Would definitely like to see @rynowak's review on all the endpoint-related stuff, as it's quite far outside my area of expertise.

src/Components/test/E2ETest/ServerExecutionTests/ConnectionQualityTests.cs
```diff
@@ -0,0 +1,122 @@
...
+
+        [Fact]
+        public void RefreshingThePageMultipleTimesCreatesNewCircuitsWithoutIssues()
+        {
+            Navigate("/connectionqualityhost");
+
+            var previousIdentifier = CaptureScopeIdentifier();
+            for (int i = 0; i < 50; i++)
+            {
+                Browser.Equal(i + 3, () => Browser.Manage().Cookies.AllCookies.Where(c => c.Name.StartsWith("Circuit.")).Count());
```
SteveSandersonMS: Should there be a call to `Browser.Manage().Cookies.DeleteAllCookies();` before we get here? Otherwise it seems like this could be flaky.

src/Components/test/E2ETest/ServerExecutionTests/RenderingModesTests.cs
```diff
@@ -0,0 +1,102 @@
...
+            var buttons = Browser.FindElements(By.CssSelector(".reveal"));
+            foreach (var button in buttons)
+            {
+                button.Click();
+            }
+
+            var ids = Browser.FindElements(By.CssSelector(".circuit-id")).Select(e => e.Text).ToArray();
+
+            Assert.Equal(2, ids.Length);
+            Assert.Single(ids.Distinct());
```
SteveSandersonMS: I'm not sure this test is doing anything to wait until the UI updates before reading the displayed circuit IDs.

SteveSandersonMS: Same with the other two tests below.

src/Components/test/testassets/ComponentsApp.Server/ComponentsApp.Server.csproj
```diff
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <TargetFramework>netcoreapp3.0</TargetFramework>
+    <AspNetCoreHostingModel>OutOfProcess</AspNetCoreHostingModel>
```
SteveSandersonMS: Out of interest, why is this needed?

src/Components/test/testassets/TestServer/Pages/ConnectionQualityHost.cshtml
```diff
@@ -0,0 +1,40 @@
...
+        So that E2E tests can make assertions about both the prerendered and
+        interactive states, we only load the .js file when told to.
+    *@
+    <hr />
+
+    <button id="load-boot-script" onclick="start()">Load boot script</button>
+
+    <script src="_framework/blazor.server.js" autostart="false"></script>
+    <script>
+        // Used by InteropOnInitializationComponent
```
SteveSandersonMS: Copy-paste artifact?

Also it would be great not to be duplicating the boilerplate we see here from L22-L38. Could this be factored out into some static `.js` file shared among the E2E test cases that use it?

SteveSandersonMS: Same with file below

src/Components/test/testassets/TestServer/Startup.cs
```diff
@@ -74,12 +73,14 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
...
             });
 
+            app.UseStaticFiles();
             app.UseRouting();
 
             app.UseEndpoints(endpoints =>
             {
                 endpoints.MapControllers();
                 endpoints.MapRazorPages();
+                endpoints.MapBlazorHub();
```
SteveSandersonMS: If this is now here, do we still also need L97?

SteveSandersonMS:
A couple of more high-level questions:

1. **What am I doing wrong?**

I tried running this with the Azure SignalR service, and found it was working fine *only on the first connection per browser instance*. After connecting once, pressing F5 in the browser to reload would fail with this:

image

To create this situation, I did the following:

 * In ComponentsApp.Server.csproj, added `<Reference Include="Microsoft.Azure.SignalR" />`
 * In its `Index.cshtml`, eliminated prerendering (i.e., reduced `<app>` to `<app>Loading...</app>`)
 * In its `Startup.cs`, added in ConfigureServices `services.AddSignalR().AddAzureSignalR("<connection string>");` and changed the `MapBlazorHub` call to `endpoints.MapBlazorHub<ComponentsApp.App.App>("app");`

Am I doing something wrong, or is this a bug? I'm not sure where to find the underlying error since all it says in the browser is that it couldn't call `StartCircuit`. I tried putting a breakpoint inside that method, and it wasn't getting hit at all when this error occurs (though does get hit when the error doesn't occur when I use a fresh new browser).

2. **Can XSS code hijack a circuit using SignalR's `negotiate` endpoint?**

I see that the protocol to connect is that we first make a POST request to `_blazor/negotiate`, passing the `circuitId` on the querystring, plus obviously any cookies go with the request too. The response gives a URL and access token for connecting to the SignalR service.

So, what's to stop XSS code from also doing a POST request to the same endpoint from your origin? It will also be sent with the cookie that includes the data protection version of the circuit identifier. Then the XSS code can read the response and exfiltrate the URL and access token.

Isn't that enough for an attacker then to connect to the SignalR service with your access token and assume control over your circuit?

src/Components/Server/src/Endpoints/CircuitMatcherPolicy.cs
```diff
@@ -0,0 +1,232 @@
...
+        // Blazor requires stickyness as the circuit is held in memory.
+        // That means that on scale-out scenarios there must be a session cookie and all circuits for a given
+        // session will end up in the same server.
+        // With this information, we can simply check the cookies the client sends us and discard any cookie for
+        // which we can't find a circuit, either connected or disconnected, as when the user closes the browser
+        // all sessions (and whatever affinity cookie was set) will go away.
+        // Circuits for which there isn't an already created circuit will contain an additional KeepAlive cookie
+        // that will be used as a filter to avoid deleting the circuit id cookie. The KeepAlive cookie for a given
+        // circuit will go away automatically.
+        private void CleanupStaleCookies(HttpContext context)
```
SteveSandersonMS: I didn't spot any E2E (or unit) tests covering this. Did I miss this? Will some be added?

javiercn: It does work (I verified manually) but I chose not to add one because it takes until the circuit goes away to take effect, so I didn't want to add a test that waits for 4 minutes.

In retrospective I think I might be able to add an endpoint that pings the server and tells it to dispose a circuit out of band, and that would allow for a test this in an E2E fashion

src/Components/test/testassets/TestServer/Pages/ConnectionQualityHost.cshtml
javiercn: That's fair.

src/Components/test/testassets/ComponentsApp.Server/ComponentsApp.Server.csproj
javiercn: I need to revert this one, but I wasn't able to run InProc IIS Express, (I haven't for a while) so I switched it to the old OutOfProccess

eng/Dependencies.props
javiercn: This doesn't hurt CTI, it only puts the reference in our build infrastructure so that we can do `<Reference Import="Microsoft.Azure.SignalR" />` in our test projects when we want to test something against Azure SignalR

src/Components/Browser.JS/src/Boot.Server.ts
javiercn: Not exactly sure what you mean here, but I did add a test that covers all possible cases:
* Only prerendered components.
* Only non-prerendered components.
* Mix of prerendered components + non-prerendered components.

src/Components/Server/ref/Microsoft.AspNetCore.Components.Server.netcoreapp3.0.cs
javiercn: That's fair. I would also hope that at some point this ends up being something on ComponentEndpointConventionBuilder instead of overloads.

src/Components/Server/src/CircuitOptions.cs
javiercn: Nope, this is cleanup.

src/Components/Browser.JS/src/Boot.Server.ts
javiercn: Fair enough

src/Components/Server/src/Circuits/CircuitHost.cs
javiercn: The way things work now:

Prerendered components or prerendered+non-prerendered
* Prerender -> Runs on Circuit Opened
* Connect -> Runs OnConnectionUp
* StartCircuit -> Renders additional components

Only non prerendered components
* StartCircuit runs OnCircuitOpenedAsyn -> OnConnectionUpAsync -> RendersAdditional components. 

OnConnectionUp always marks when the socket is opened and we're ready to perform JSInterop.

src/Components/Server/src/Circuits/CircuitIdFactory.cs
javiercn: That's something we can do in the future, but it's not necessary. The way it works is we add an Identity to the request principal with a claim "bcid" which is the circuit id and we verify that the current principal has a claim that matches the provided circuit id.

src/Components/Server/src/Circuits/CircuitIdFactory.cs
javiercn: Yeah, I also plan to do some cleanup here.

src/Components/Server/src/Circuits/CircuitRegistry.cs
javiercn: Yes, but in all effects is the string we use to identify the circuit

src/Components/Server/src/Circuits/DefaultCircuitFactory.cs
javiercn: Nope, just baggage from a previous implementation where it was done through an authentication handler.

src/Components/Server/src/DependencyInjection/ComponentServiceCollectionExtensions.cs
javiercn: We're protecting against the circuit id being used outside of the browser context, if you manage an XSS and get access to the request token you have everything you need to connect to the server and do stuff as you are effectively running in the users context. (Although you need to do it from within the page, which I'm not sure is viable). 

The cookie prevents against the circuit id being used from outside the browser context and across origins (as is same-site by default, although this is something likely we will allow configuring, as its required based on conversations I had with Chris)

src/Components/Server/src/Endpoints/CircuitMatcherPolicy.cs
javiercn: Assuming you have 50 open circuits on a client at the same time, `0.00000298023223876953125‬` those are your chances of running into a collision I believe. We can add a couple more and won't hurt, but I think its already within target.

src/Components/Server/src/Endpoints/CircuitMatcherPolicy.cs
javiercn: Claims need to be short, but I would accept `mabcid` (Microsoft AspNetCore Blazor Circuit ID) sounds good?

src/Components/test/E2ETest/ServerExecutionTests/ConnectionQualityTests.cs
javiercn: You're right, It was the only test in this class and then I ended up adding more.

src/Components/Browser.JS/src/Boot.Server.ts
SteveSandersonMS: Sounds like the coverage is good then. I made that comment before I got to the end of the PR where the E2E tests were.

src/Components/Server/src/DependencyInjection/ComponentServiceCollectionExtensions.cs
SteveSandersonMS: I understand - my comment was just about the use of the term `csrf` in the line I'm commenting on. It seems unrelated to what we're doing here.

src/Components/Server/src/Endpoints/CircuitMatcherPolicy.cs
SteveSandersonMS: Sure. If each user has 2 circuits, then the expected number of users until you get a clash is about 12 million. It is very unlikely for any given user, but people do win lotteries :)

src/Components/Server/src/Endpoints/CircuitMatcherPolicy.cs
SteveSandersonMS: Yeah, sounds much less likely to clash.

javiercn:
> I tried running this with the Azure SignalR service, and found it was working fine _only on the first connection per browser instance_. After connecting once, pressing F5 in the browser to reload would fail with this:

Haven't experienced this in the past.

javiercn:
> 1. **Can XSS code hijack a circuit using SignalR's `negotiate` endpoint?**
> 
> I see that the protocol to connect is that we first make a POST request to `_blazor/negotiate`, passing the `circuitId` on the querystring, plus obviously any cookies go with the request too. The response gives a URL and access token for connecting to the SignalR service.
> 
> So, what's to stop XSS code from also doing a POST request to the same endpoint from your origin? It will also be sent with the cookie that includes the data protection version of the circuit identifier. Then the XSS code can read the response and exfiltrate the URL and access token.
> 
> Isn't that enough for an attacker then to connect to the SignalR service with your access token and assume control over your circuit?

I don't think there's anything that fully protects you against XSS, but this limits the impact of an XSS in the sense that every action to be taken needs to be performed within the page (as you don't have access to the cookie) which means you cannot send the circuit id to a 3rd party server to connect to the app and party on the circuit, a compromised app is easier to spot (as there's gonna be network traffic) and the attack is much more unreliable as the browser window can be closed at any time.

The other data point is that the circuit id or request token in this case, is not enough information to connect to the circuit by itself, so it puts us in a much better place when it comes to it being logged or leaked somewhere.

How the access token is handled belongs to the signalr service. If its a single use token then multiple connections could be detected and the connection dropped all together.

XSS is probably the most powerful way to PWN a site, so its hard to fully protect against it unless the user uses something like CSP.

halter73:
> So, what's to stop XSS code from also doing a POST request to the same endpoint from your origin?

Is the question how we defend against an attacker who already found a way to inject JS into your site? I'm not sure there's much to do in that case.

eng/Dependencies.props
davidfowl: This is going to be constantly broken while we're still making changes. I'd suggest not doing this right now and wait until we stop making big API changes that break the azure signalr SDK.

javiercn:
Closing this as given that we are getting rid of stateful prerendering there's nothing to savage from the PR associated with this issue..

Issue #6381, opened 2019-01-04
Add name for EventId in logs


Kahbazi:
Can we have log event name for logs in all the projects? Lately it has been added for MVC (https://github.com/aspnet/Mvc/issues/6818)

I can submit a PR if it's ok.

Here's a list of projects so we can see which one already has the event name and which one doesn't.
    
- [x] Microsoft.AspNetCore.DataProtection
- [x] Microsoft.AspNetCore.Hosting
- [x] Microsoft.AspNetCore.Identity
- [x] Microsoft.AspNetCore.Server.HttpSys
- [x] Microsoft.Extensions.Identity.Core
- [x] Microsoft.AspNetCore.Server.Kestrel.Core
- [x] Microsoft.AspNetCore.HostFiltering
- [x] Microsoft.AspNetCore.Antiforgery
- [x] Microsoft.AspNetCore.Routing
- [x] Microsoft.AspNetCore.Cors
- [x] Microsoft.AspNetCore.Diagnostics
- [x] Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore
- [x] Microsoft.AspNetCore.Localization
- [x] Microsoft.AspNetCore.HttpsPolicy
- [x] Microsoft.AspNetCore.ResponseCaching
- [x] Microsoft.AspNetCore.ResponseCompression
- [x] Microsoft.AspNetCore.Rewrite
- [x] Microsoft.AspNetCore.Session
- [x] Microsoft.AspNetCore.StaticFiles
- [x] Microsoft.AspNetCore.Mvc.Core
- [x] Microsoft.AspNetCore.Mvc.Formatters.Json
- [x] Microsoft.AspNetCore.Authentication
- [x] Microsoft.AspNetCore.Authentication.Cookies
- [x] Microsoft.AspNetCore.Authentication.JwtBearer
- [x] Microsoft.AspNetCore.Authentication.OpenIdConnect
- [x] Microsoft.AspNetCore.Authentication.Twitter
- [x] Microsoft.AspNetCore.Authentication.WsFederation
- [x] Microsoft.AspNetCore.Authorization
- [x] Microsoft.AspNetCore.CookiePolicy
- [x] Microsoft.AspNetCore.Http.Connections
- [x] Microsoft.AspNetCore.Http.Connections.Client
- [x] Microsoft.AspNetCore.SignalR.Client.Core
- [x] Microsoft.AspNetCore.SignalR.Core
[...]

PR #11379, opened 2019-06-19
Added name for EventId in logs

mikeo52:
Added name for EventId in Hosting, DataProtection, EntityFrameworkCore. #6381


src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs
```diff
@@ -172,7 +172,7 @@ private void LogRequestStarting(HttpContext httpContext)
             // IsEnabled is checked in the caller, so if we are here just log
             _logger.Log(
                 logLevel: LogLevel.Information,
-                eventId: LoggerEventIds.RequestStarting,
+                eventId: new EventId(LoggerEventIds.RequestStarting, nameof(LoggerEventIds.RequestStarting)),
```
anurse: I would avoid `nameof` here. Even though we don't intent to change this, if we did ever change the name of the property it would break consumers. We started using `nameof` originally but decided it was better to just use string constants. The goal of `nameof` is to ensure that when you rename the target of the `nameof` expression, your reference is updated which is actually *not* what we want here.

analogrelay (commented):
Looks good, but we should remove the `nameof`s. Even though they reference `LoggerEventIds`, we still want to reduce the possibility of unintentionally changing these names without realizing.

src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs
mikeo52: I hadn't thought about it that way at first, but I can see your point.  This should be corrected in the latest commit.

analogrelay (approved):
Looks good. It's unfortunate that some of these have `0` as the numeric event ID, but changing that is a breaking change and I don't think we want to do that.

anurse:
Build failures look unrelated. I'll restart.

anurse:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=230970

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/2220
https://github.com/aspnet/AspNetCore-Internal/issues/2410
https://github.com/aspnet/AspNetCore-Internal/issues/2623
https://github.com/aspnet/AspNetCore-Internal/issues/2415
https://github.com/aspnet/AspNetCore-Internal/issues/2674

Kahbazi:
@anurse is there any chance that this could merge into asp.net core 3?

Kahbazi:
@mikeo52 Can you resolve the conflict?
I'm hoping this would reach the  3.0 version.

Tratcher:
FYI this is targeting the wrong branch for v3.x. It would need to be rebased and retargeted to `release/3.0`.

@Pilchie bar check for 3.x? Incremental diagnostic improvements.

anurse:
Oof, this got lost in some of the push to get 3.0 out the door. I'm going to respin the builds and see where we land. There is also a conflict to resolve.

anurse:
@mikeo52 are you up to try rebasing this on `master` and resolving the conflict so we can try to take this for 5.0 at least?

mikeo52:
@anurse yeah I updated it locally since there are just two spots where it's conflicting, but I'm having a lovely fun time merging because of the same error someone else reported below.  

https://github.com/aspnet/AspNetCore/issues/15125

mikeo52:
Okay that makes absolutely no sense to me, but whatever.  On this PR's merge conflicts screen I can edit the merge to my repo via the browser and resolve the conflicts, but if I use the git command line I constantly run into the other issue.

src/DataProtection/DataProtection/src/LoggingExtensions.cs
```diff
@@ -149,268 +149,268 @@ internal static class LoggingExtensions
...
                 logLevel: LogLevel.Error,
                 formatString: "An error occurred while reading the key ring.");
             _keyRingDoesNotContainValidDefaultKey = LoggerMessage.Define(
-                eventId: 49,
+                eventId: new EventId(49, "KeyRingDoesNotContainValidDefaultKey"),
                 logLevel: LogLevel.Error,
                 formatString: "The key ring does not contain a valid default key, and the key manager is configured with auto-generation of keys disabled.");
             _usingInmemoryRepository = LoggerMessage.Define(
-                eventId: 50,
+                eventId: new EventId(50, "UsingInmemoryRepository"),
```
anurse:
```suggestion
                eventId: new EventId(50, "UsingInMemoryRepository"),
```

"In Memory" is two words, so we should capitalize it. (I realize the logging method and field name may not use this capitalization but those aren't technically publicly-viewable :)).

src/DataProtection/DataProtection/src/LoggingExtensions.cs
```diff
@@ -149,268 +149,268 @@ internal static class LoggingExtensions
...
                 formatString: "Neither user profile nor HKLM registry available. Using an ephemeral key repository. Protected data will be unavailable when application exits.");
             _usingEphemeralFileSystemLocationInContainer = LoggerMessage.Define<string>(
-                eventId: 60,
+                eventId: new EventId(60, "UsingEphemeralFileSystemLocationInContainer"),
                 logLevel: LogLevel.Warning,
                 formatString: Resources.FileSystem_EphemeralKeysLocationInContainer);
 
             _usingRegistryAsKeyRepositoryWithDPAPI = LoggerMessage.Define<string>(
-                eventId: 0,
+                eventId: new EventId(0, "UsingRegistryAsKeyRepositoryWithDPAPI"),
```
anurse: Given that this is 5.0, I feel OK putting proper EventIds in place here.

For simplicity, I think it's OK to just use the same EventId sequence as the previous ones, even if these messages are used with different loggers. Thoughts @davidfowl ? It makes it easy to keep them consistent and EventIds are arbitrary anyway.

@mikeo52 can you continue the EventId number sequence with these last few events? (61, 62, 63...)

src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs
```diff
@@ -175,7 +175,7 @@ private void LogRequestStarting(HostingApplication.Context context)
 
             _logger.Log(
                 logLevel: LogLevel.Information,
-                eventId: LoggerEventIds.RequestStarting,
+                eventId: new EventId(LoggerEventIds.RequestStarting, "RequestStarting"),
```
anurse: I think you could go to `LoggerEventIds` and change them from ints to `EventId` structs (you'd also have to change from `const` to `static readonly` fields but it's an internal type so I think that's OK. Could you give that a look @mikeo52 ?

src/Hosting/Hosting/src/Internal/HostingLoggerExtensions.cs
```diff
@@ -22,15 +22,15 @@ public static IDisposable RequestScope(this ILogger logger, HttpContext httpCont
         public static void ApplicationError(this ILogger logger, Exception exception)
         {
             logger.ApplicationError(
-                eventId: LoggerEventIds.ApplicationStartupException,
+                eventId: new EventId(LoggerEventIds.ApplicationStartupException, "ApplicationStartupException"),
```
anurse: Same here, we should try to just replace the `LoggerEventIds` constants themselves.

analogrelay (changes requested):
A few small things. Looking good though. Thanks for resolving the conflict!

src/DataProtection/DataProtection/src/LoggingExtensions.cs
mikeo52: @anurse should be good to go with the latest commit.

src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs
mikeo52: @anurse I think I've got the change you were after made in this regard.  Take a look and let me know.  I'm having trouble building locally since my local setup's a little borked, so my apologies if the extra commits are making noise.

Tratcher (approved):

Tratcher:
Thanks

Kahbazi:
@Tratcher Is it possible to also merge this to 3.1?

Tratcher:
No, 3.1 is only taking critical patches now.

Kahbazi:
I see, thanks.

PR #11382, opened 2019-06-19
Pass options to HubConnectionContext

### LINKED ISSUE #11258: [API Review]: HubConnectionContext should take an options object
We already have 2 constructors and this release adds a 3rd:

https://github.com/aspnet/AspNetCore/blob/9662a5bf542654b1ed1f8007efe068bb7063afcf/api-diffs/api-diffs_Microsoft.AspNetCore.SignalR.md

Might make sense to pass an options object through.

BrennanConroy:
Fixes #11258

Reusing `HubOption` to pass values to `HubConnectionContext`.

### DIFF
--- a/src/SignalR/server/Core/ref/Microsoft.AspNetCore.SignalR.Core.netcoreapp3.0.cs
+++ b/src/SignalR/server/Core/ref/Microsoft.AspNetCore.SignalR.Core.netcoreapp3.0.cs
@@ -127,9 +127,7 @@ namespace Microsoft.AspNetCore.SignalR
     }
     public partial class HubConnectionContext
     {
-        public HubConnectionContext(Microsoft.AspNetCore.Connections.ConnectionContext connectionContext, System.TimeSpan keepAliveInterval, Microsoft.Extensions.Logging.ILoggerFactory loggerFactory) { }
-        public HubConnectionContext(Microsoft.AspNetCore.Connections.ConnectionContext connectionContext, System.TimeSpan keepAliveInterval, Microsoft.Extensions.Logging.ILoggerFactory loggerFactory, System.TimeSpan clientTimeoutInterval) { }
-        public HubConnectionContext(Microsoft.AspNetCore.Connections.ConnectionContext connectionContext, System.TimeSpan keepAliveInterval, Microsoft.Extensions.Logging.ILoggerFactory loggerFactory, System.TimeSpan clientTimeoutInterval, int streamBufferCapacity) { }
+        public HubConnectionContext(Microsoft.AspNetCore.Connections.ConnectionContext connectionContext, Microsoft.AspNetCore.SignalR.HubOptions hubOptions, Microsoft.Extensions.Logging.ILoggerFactory loggerFactory) { }
         public virtual System.Threading.CancellationToken ConnectionAborted { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
         public virtual string ConnectionId { get { throw null; } }
         public virtual Microsoft.AspNetCore.Http.Features.IFeatureCollection Features { get { throw null; } }
--- a/src/SignalR/server/Core/ref/Microsoft.AspNetCore.SignalR.Core.netcoreapp3.0.cs
+++ b/src/SignalR/server/Core/ref/Microsoft.AspNetCore.SignalR.Core.netcoreapp3.0.cs
@@ -127,7 +127,7 @@ namespace Microsoft.AspNetCore.SignalR
     }
     public partial class HubConnectionContext
     {
-        public HubConnectionContext(Microsoft.AspNetCore.Connections.ConnectionContext connectionContext, Microsoft.AspNetCore.SignalR.HubOptions hubOptions, Microsoft.Extensions.Logging.ILoggerFactory loggerFactory) { }
+        public HubConnectionContext(Microsoft.AspNetCore.Connections.ConnectionContext connectionContext, Microsoft.AspNetCore.SignalR.HubConnectionContextOptions contextOptions, Microsoft.Extensions.Logging.ILoggerFactory loggerFactory) { }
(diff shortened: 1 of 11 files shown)


src/SignalR/server/Core/src/HubConnectionContext.cs
```diff
@@ -48,37 +48,17 @@ public class HubConnectionContext
         /// Initializes a new instance of the <see cref="HubConnectionContext"/> class.
         /// </summary>
         /// <param name="connectionContext">The underlying <see cref="ConnectionContext"/>.</param>
-        /// <param name="keepAliveInterval">The keep alive interval. If no messages are sent by the server in this interval, a Ping message will be sent.</param>
         /// <param name="loggerFactory">The logger factory.</param>
-        public HubConnectionContext(ConnectionContext connectionContext, TimeSpan keepAliveInterval, ILoggerFactory loggerFactory)
```
BrennanConroy: Question: Should we keep the 2 old constructors so this isn't a breaking change?

src/SignalR/server/Core/src/HubConnectionHandler.cs
```diff
@@ -75,24 +75,27 @@ IServiceScopeFactory serviceScopeFactory
...
 
             Log.ConnectedStarting(_logger);
 
-            var connectionContext = new HubConnectionContext(connection, keepAlive, _loggerFactory, clientTimeout, streamBufferCapacity);
+            var connectionContext = new HubConnectionContext(connection, _loggerFactory, options);
 
-            var resolvedSupportedProtocols = (supportedProtocols as IReadOnlyList<string>) ?? supportedProtocols.ToList();
-            if (!await connectionContext.HandshakeAsync(handshakeTimeout, resolvedSupportedProtocols, _protocolResolver, _userIdProvider, _enableDetailedErrors))
+            var resolvedSupportedProtocols = (options.SupportedProtocols as IReadOnlyList<string>) ?? options.SupportedProtocols.ToList();
+            if (!await connectionContext.HandshakeAsync(options.HandshakeTimeout.Value, resolvedSupportedProtocols, _protocolResolver, _userIdProvider, _enableDetailedErrors))
```
mikaelm12: Why did we have to call value here on the timeout when we didn't before? They should be effectively the same, right?

BrennanConroy: `HubOption` uses nullable values and `HandshakeAsync` doesn't accept a nullable value. `HandshakeTimeout` is guaranteed to have a value at this point because we set a default up above if no value provided.

src/SignalR/server/Core/src/HubConnectionContext.cs
mikaelm12: Meh.

halter73: @anurse Do we need to make an announcement?

src/SignalR/server/Core/ref/Microsoft.AspNetCore.SignalR.Core.netcoreapp3.0.cs
```diff
@@ -127,9 +127,7 @@ public static partial class HubClientsExtensions
     }
     public partial class HubConnectionContext
     {
-        public HubConnectionContext(Microsoft.AspNetCore.Connections.ConnectionContext connectionContext, System.TimeSpan keepAliveInterval, Microsoft.Extensions.Logging.ILoggerFactory loggerFactory) { }
-        public HubConnectionContext(Microsoft.AspNetCore.Connections.ConnectionContext connectionContext, System.TimeSpan keepAliveInterval, Microsoft.Extensions.Logging.ILoggerFactory loggerFactory, System.TimeSpan clientTimeoutInterval) { }
-        public HubConnectionContext(Microsoft.AspNetCore.Connections.ConnectionContext connectionContext, System.TimeSpan keepAliveInterval, Microsoft.Extensions.Logging.ILoggerFactory loggerFactory, System.TimeSpan clientTimeoutInterval, int streamBufferCapacity) { }
+        public HubConnectionContext(Microsoft.AspNetCore.Connections.ConnectionContext connectionContext, Microsoft.AspNetCore.SignalR.HubOptions hubOptions, Microsoft.Extensions.Logging.ILoggerFactory loggerFactory) { }
```
halter73: Do we have a standard order for parameters? I.e. do we always try to keep logger parameters last and options parameters second to last?

halter73 (approved):

mikaelm12 (approved):

src/SignalR/server/Core/src/HubConnectionContext.cs
anurse: Sure do!

src/SignalR/server/Core/ref/Microsoft.AspNetCore.SignalR.Core.netcoreapp3.0.cs
BrennanConroy: No idea

davidfowl: Should this context be the hub options? Why isn't it its own context

BrennanConroy: I chose `HubOption` because it contains the settings we want to pass to `HubConnectionContext` and I didn't want to make a new public type. We can change that if you'd like.

davidfowl: Yes I think we should have a `HubConnectionContextOptions` as this will start to break down the minute we need to pass something that isn't exactly an option.

davidfowl (approved):

BrennanConroy:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=232403

aspnet-hello:

Issue #7750, opened 2019-02-20
IIS (out-of-proc) calls AuthenticateAsync for all requests at the start of the pipeline


fractionalJoe:
When implementing a ClaimsTransformation that attempts to access  HttpContext.Session, an error "Session has not been configured for this application or request." is raised by Kestrel. This is related to #3805 and [this code](https://github.com/aspnet/IISIntegration/blob/6e54256fca6f2ae25525bfbc0a7ca966e0d1780d/src/Microsoft.AspNetCore.Server.IISIntegration/IISMiddleware.cs#L133-L137) appears to be the culprit as it invokes authentication middleware before UseSession is called in the Startup.Configure method. I can find no way to check if Session has been been configured without a try/catch block.

### To Reproduce
1. Using IIS 10 and ASP.Net Core 2.2. App must be configured to run Out-of-process for this issue to occur.
2. I have created a repo @ https://github.com/fractionalJoe/TestIisIntegration that has a skeleton project that can be deployed to IIS to demonstrate the issue. It seemed like too much code to paste inline since it was multiple files. 
3. I enable stdout to get the following message. 
```
Hosting environment: Production
Content root path: C:\Users\*****\source\TestIisIntegration\bin\Release\netcoreapp2.2\win-x64\publish
Now listening on: http://127.0.0.1:5781
Application started. Press Ctrl+C to shut down.
info: Microsoft.AspNetCore.Hosting.Internal.WebHost[1]
      Request starting HTTP/1.1 GET http://localhost/  
info: Microsoft.AspNetCore.Mvc.Internal.ControllerActionInvoker[1]
      Route matched with {action = "Index", controller = "Home"}. Executing action TestIisIntegration.Controllers.HomeController.Index (TestIisIntegration)
info: Microsoft.AspNetCore.Authorization.DefaultAuthorizationService[2]
      Authorization failed.
info: Microsoft.AspNetCore.Mvc.Internal.ControllerActionInvoker[3]
      Authorization failed for the request at filter 'Microsoft.AspNetCore.Mvc.Authorization.AuthorizeFilter'.
info: Microsoft.AspNetCore.Mvc.ChallengeResult[1]
      Executing ChallengeResult with authentication schemes ().
info: Microsoft.AspNetCore.Mvc.Internal.ControllerActionInvoker[2]
      Executed action TestIisIntegration.Controllers.HomeController.Index (TestIisIntegration) in 43.0389ms
[...]

PR #11390, opened 2019-06-19
Avoid calling AuthenticateAsync for IIS out-of-proc

Tratcher:
#7750 out-of-proc eagerly calls AuthenticateAsync to initialize the WindowsIdentity and avoid leaking handles. However, AuthenticateAsync has side-effects like calling claims transformation, which users have complained about because they need to run some other middleware first (e.g Session).

Fix: Initialize the WindowsIdentity in the IIS middleware and then flow it through the feature collection to the auth handler. In-proc does it the same way.

Officially this is a breaking change because claims transformation will now only be called if you use the middleware or call AuthenticateAsync directly.

### DIFF
--- a/src/Servers/IIS/IISIntegration/src/IISMiddleware.cs
+++ b/src/Servers/IIS/IISIntegration/src/IISMiddleware.cs
@@ -3,10 +3,11 @@
 
 using System;
 using System.Diagnostics;
+using System.Globalization;
+using System.Security.Principal;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Authentication;
 using Microsoft.AspNetCore.Builder;
-using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Features;
 using Microsoft.Extensions.Hosting;
@@ -21,8 +22,10 @@ namespace Microsoft.AspNetCore.Server.IISIntegration
         private const string MSAspNetCoreClientCert = "MS-ASPNETCORE-CLIENTCERT";
         private const string MSAspNetCoreToken = "MS-ASPNETCORE-TOKEN";
         private const string MSAspNetCoreEvent = "MS-ASPNETCORE-EVENT";
+        private const string MSAspNetCoreWinAuthToken = "MS-ASPNETCORE-WINAUTHTOKEN";
         private const string ANCMShutdownEventHeaderValue = "shutdown";
         private static readonly PathString ANCMRequestPath = new PathString("/iisintegration");
+        private static readonly Func<object, Task> ClearUserDelegate = ClearUser;
 
         private readonly RequestDelegate _next;
         private readonly IISOptions _options;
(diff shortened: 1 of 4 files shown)


jkotalik (approved):
🌶 🔥

src/Servers/testassets/ServerComparison.TestSites/OneTransformPerRequest.cs
```diff
@@ -0,0 +1,32 @@
...
+        }
+
+        public IHttpContextAccessor ContextAccessor { get; }
+
+        public Task<ClaimsPrincipal> TransformAsync(ClaimsPrincipal principal)
+        {
+            var context = ContextAccessor.HttpContext;
+            if (context.Items["Transformed"] != null)
+            {
+                throw new InvalidOperationException("Transformation ran multiple times.");
```
HaoK: Guessing this used to throw prior to the changes in this PR?

HaoK (approved):
Assuming the test website changes are verifying the new behavior, looks good

src/Servers/testassets/ServerComparison.TestSites/OneTransformPerRequest.cs
Tratcher: Correct, only for out-of-proc.

src/Servers/IIS/IISIntegration/src/IISMiddleware.cs
```diff
@@ -147,5 +156,40 @@ public async Task Invoke(HttpContext httpContext)
...
+                && int.TryParse(tokenHeader, NumberStyles.HexNumber, CultureInfo.InvariantCulture, out var hexHandle))
+            {
+                // Always create the identity if the handle exists, we need to dispose it so it does not leak.
+                var handle = new IntPtr(hexHandle);
+                var winIdentity = new WindowsIdentity(handle, IISDefaults.AuthenticationScheme);
+
+                // WindowsIdentity just duplicated the handle so we need to close the original.
+                NativeMethods.CloseHandle(handle);
+
+                context.Response.OnCompleted(ClearUserDelegate, (context, winIdentity));
```
davidfowl: FYI: This allocates now. Can you shove that state into the HttpContext.Items?

benaadams: Or double register
```csharp
context.Response.OnCompleted(DisposeUserDelegate, winIdentity);
context.Response.OnCompleted(ClearUserDelegate, context);
```

davidfowl: I thought about that but didn't suggest because I was thinking it's another allocation. I guess it isn't because the list default size is 4.

benaadams: I assume the `.User` on HttpContext can't be trusted? (as it may have changed and the one that needs disposing lost?)

src/Servers/IIS/IISIntegration/src/IISMiddleware.cs
```diff
@@ -147,5 +156,39 @@ public async Task Invoke(HttpContext httpContext)
...
+
+            if (!StringValues.IsNullOrEmpty(tokenHeader)
+                && int.TryParse(tokenHeader, NumberStyles.HexNumber, CultureInfo.InvariantCulture, out var hexHandle))
+            {
+                // Always create the identity if the handle exists, we need to dispose it so it does not leak.
+                var handle = new IntPtr(hexHandle);
+                var winIdentity = new WindowsIdentity(handle, IISDefaults.AuthenticationScheme);
+
+                // WindowsIdentity just duplicated the handle so we need to close the original.
+                NativeMethods.CloseHandle(handle);
```
halter73: I know it's not new, but should this be in a finally?

halter73 (approved):

src/Servers/IIS/IISIntegration/src/IISMiddleware.cs
Tratcher: The most likely thing to throw would be for an invalid handle, so closing it wouldn't help much. As this dates back to 1.0 and we haven't seen any issues, I'm going to leave it 😁.

PR #11394, opened 2019-06-19
Remove copy race for IIS projects

jkotalik:
There is a race when native dlls are being copied to build samples/test projects due to the file being locked. This can happen when the native dlls are being written to on build finishing. I believe this will fix the race.


JunTaoLuo (approved):
We'll give it a shot! Thanks!

PR #11399, opened 2019-06-19
Add Invoke API with with no completion result

### LINKED ISSUE #11220: [SignalR][JavaClient] HubConnection.invoke() without return value
### Is your feature request related to a problem? Please describe.
I´m using the .NET Client and Java Client for SignalR.
I implemented some of my **server** methods without a explicit returnvalue like so:
```csharp
public async Task SendMessage(string user, string message)
```

The **.NET Client** calls these methods like so:
```csharp
await connection.InvokeAsync("SendMessage", "UserName", "Message");
```

The **Java Client** does not offer such a function to call a hubmethod without expecting an return value (but wait for the completition).
There is only this method available:
```java
HubConnection.invoke(Class<T> returnType, String method, Object... args)
```
which is not suitable for this purpose.

The java client offers a invoke function that does not expect a returnvalue but waits for completition.

mikaelm12:
Fixes: https://github.com/aspnet/AspNetCore/issues/11220
This is to add the invoke overload that doesn't expect result in the completion message
There were also some issues in other tests about ensuring that `onSuccess` had run

### DIFF
--- a/src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
+++ b/src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
@@ -601,7 +601,7 @@ public class HubConnection {
      *
      * @param method The name of the server method to invoke.
      * @param args The arguments used to invoke the server method.
-     * @return A Single that yields the return value when the invocation has completed.
+     * @return A Completable that indicates when the invocation has completed.
      */
     @SuppressWarnings("unchecked")
     public Completable invoke(String method, Object... args) {
--- a/src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
+++ b/src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
@@ -591,8 +591,40 @@ public class HubConnection {
                 this.streamMap.put(streamId, stream);
             }
         }
-        
+
         return params.toArray();
+    }
+
+
+    /**
+     * Invokes a hub method on the server using the specified method name and arguments.
+     *
+     * @param method The name of the server method to invoke.
+     * @param args The arguments used to invoke the server method.
+     * @return A Single that yields the return value when the invocation has completed.
+     */
+    @SuppressWarnings("unchecked")
+    public Completable invoke(String method, Object... args) {
+        if (hubConnectionState != HubConnectionState.CONNECTED) {
+            throw new RuntimeException("The 'invoke' method cannot be called if the connection is not active.");
+        }
+
+        String id = connectionState.getNextInvocationId();
+
+        CompletableSubject subject = CompletableSubject.create();
+        InvocationRequest irq = new InvocationRequest(null, id);
+        connectionState.addInvocation(irq);
+
+        Subject<Object> pendingCall = irq.getPendingCall();
+
+        pendingCall.subscribe(result -> {},
+                error -> subject.onError(error),
+                () -> subject.onComplete());
+
+        // Make sure the actual send is after setting up the callbacks otherwise there is a race
+        // where the map doesn't have the callbacks yet when the response is returned
+        sendInvocationMessage(method, args, id, false);
+        return subject;
     }
 
     /**
--- a/src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
+++ b/src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
@@ -617,7 +617,7 @@ public class HubConnection {
 
         Subject<Object> pendingCall = irq.getPendingCall();
 
-        pendingCall.subscribe(result -> {},
+        pendingCall.subscribe(result -> subject.onComplete(),
                 error -> subject.onError(error),
                 () -> subject.onComplete());
 
--- a/src/SignalR/clients/java/signalr/src/test/java/com/microsoft/signalr/HubConnectionTest.java
+++ b/src/SignalR/clients/java/signalr/src/test/java/com/microsoft/signalr/HubConnectionTest.java
@@ -14,6 +14,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 
+import io.reactivex.Completable;
 import org.junit.jupiter.api.Test;
 
 import io.reactivex.Observable;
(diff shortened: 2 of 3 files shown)


src/SignalR/clients/java/signalr/src/test/java/com/microsoft/signalr/HubConnectionTest.java
```diff
@@ -14,6 +14,7 @@
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 
+import io.reactivex.Completable;
```
BrennanConroy: nit: sort

src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
```diff
@@ -591,10 +591,42 @@ void launchStreams(List<String> streamIds) {
...
+
+        String id = connectionState.getNextInvocationId();
+
+        CompletableSubject subject = CompletableSubject.create();
+        InvocationRequest irq = new InvocationRequest(null, id);
+        connectionState.addInvocation(irq);
+
+        Subject<Object> pendingCall = irq.getPendingCall();
+
+        pendingCall.subscribe(result -> {},
```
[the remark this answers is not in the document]
BrennanConroy: And then probably write a test for this

src/SignalR/clients/java/signalr/src/test/java/com/microsoft/signalr/HubConnectionTest.java
```diff
@@ -946,15 +946,63 @@ public void invokeWaitsForCompletionMessage() {
...
+    public void invokeNoReturnValueWaitsForCompletion() {
+        MockTransport mockTransport = new MockTransport();
+        HubConnection hubConnection = TestUtils.createHubConnection("http://example.com", mockTransport);
+
+        hubConnection.start().timeout(1, TimeUnit.SECONDS).blockingAwait();
+
+        AtomicBoolean done = new AtomicBoolean();
+        Completable result = hubConnection.invoke("test", "message");
+        result.doOnComplete(() -> done.set(true)).subscribe();
+        result.subscribe();
```
BrennanConroy: Is this needed?

src/SignalR/clients/java/signalr/src/test/java/com/microsoft/signalr/HubConnectionTest.java
```diff
@@ -946,15 +946,63 @@ public void invokeWaitsForCompletionMessage() {
...
+    @Test
+    public void invokeNoReturnValueHandlesError() {
+        MockTransport mockTransport = new MockTransport();
+        HubConnection hubConnection = TestUtils.createHubConnection("http://example.com", mockTransport);
+
+        hubConnection.start().timeout(1, TimeUnit.SECONDS).blockingAwait();
+
+        AtomicBoolean done = new AtomicBoolean();
+        Completable result = hubConnection.invoke("test", "message");
+        result.doOnComplete(() -> done.set(true)).subscribe(() -> {}, (error) -> {});
```
BrennanConroy: nit: normal `subscribe` is probably fine

src/SignalR/clients/java/signalr/src/test/java/com/microsoft/signalr/HubConnectionTest.java
```diff
@@ -946,15 +946,63 @@ public void invokeWaitsForCompletionMessage() {
...
+        assertFalse(done.get());
+
+        mockTransport.receiveMessage("{\"type\":3,\"invocationId\":\"1\",\"error\":\"There was an error\"}" + RECORD_SEPARATOR);
+
+        result.timeout(1000, TimeUnit.MILLISECONDS).blockingGet();
+
+        AtomicReference<String> errorMessage = new AtomicReference<>();
+        result.doOnError(error -> {
+            errorMessage.set(error.getMessage());
+        }).subscribe(() -> {}, (error) -> {});
```
BrennanConroy: nit: normal `subscribe`

src/SignalR/clients/java/signalr/src/main/java/com/microsoft/signalr/HubConnection.java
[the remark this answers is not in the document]
mikaelm12: > This one should probably also complete the subject. 

Hmmm. Should it though. I guess for the case in which someone just doesn't care about the result. I wouldn't want to encourage that patterns of using this API for result returning completions.

src/SignalR/clients/java/signalr/src/test/java/com/microsoft/signalr/HubConnectionTest.java
mikaelm12: Nope. Left over from debugging

src/SignalR/clients/java/signalr/src/test/java/com/microsoft/signalr/HubConnectionTest.java
mikaelm12: We actually do need to define the `onError` callback on this subscription. It won't fail the test but then we get all those exceptions in the test logs

src/SignalR/clients/java/signalr/src/test/java/com/microsoft/signalr/HubConnectionTest.java
```diff
@@ -946,15 +946,100 @@ public void invokeWaitsForCompletionMessage() {
...
+
+        assertEquals("{\"type\":1,\"invocationId\":\"1\",\"target\":\"test\",\"arguments\":[\"message\"]}" + RECORD_SEPARATOR, mockTransport.getSentMessages()[1]);
+        assertFalse(done.get());
+
+        mockTransport.receiveMessage("{\"type\":3,\"invocationId\":\"1\",\"error\":\"There was an error\"}" + RECORD_SEPARATOR);
+
+        result.timeout(1000, TimeUnit.MILLISECONDS).blockingGet();
+
+        AtomicReference<String> errorMessage = new AtomicReference<>();
+        result.doOnError(error -> {
```
BrennanConroy: We should be able to clean this up by doing a normal subscribe and providing error and success callbacks instead of `doOnComplete(...).subscribe(...)` and `doOnError(...).subscribe(...)`

BrennanConroy (approved):

jaragones:
Quick question (if you don't mind), I'm using signalR on an Android project importing library like this "implementation 'com.microsoft.signalr:signalr:1.+'" as it is said in documentation, version that gets installed is 1.0.5 but it doesn't contain this fix (I can see that it was merged on master on June). Do we need to import this library in a different way?

BrennanConroy:
Use version 3.0.0

jaragones:
Thanks!

PR #11407, opened 2019-06-20
Fix Template test runs on macOS and Linux

JunTaoLuo:
This should fix the template test runs on macOS and Linux.

@javiercn


.azure/pipelines/ci.yml
```diff
@@ -431,6 +431,10 @@ jobs:
     - bash: "./eng/scripts/install-nginx-mac.sh"
       displayName: Installing Nginx
     afterBuild:
+    - bash: ./build.sh --ci --pack --no-build --no-restore --no-build-deps "/bl:artifacts/log/packages.pack.binlog"
```
JunTaoLuo: I don't like the fact that we need to run these steps separately, since if we were to move these tests to another configuration, we need to move these steps as well. However I'd rather unblock these tests first and investigate simplifying via a script or MSBuild magic separately.

javiercn (approved):
As long as it works

JunTaoLuo:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=231735

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/2314

javiercn:
@JunTaoLuo I can merge this for you if it gets late in your time. Just let me know.

JunTaoLuo:
I was going to wait for the other test but I realized it's irrelevant so I'll merge now.

Issue #11186, opened 2019-06-13
SiteExtensions do not build with Arcade


natemcmaster:
While attempting to adopt Arcade, I ran into issues getting the src/SiteExtensions/ projects to build. In the latest draft of my PR, I have disabled the build of these projects because they might require a bigger refactor than I could fit into #11122.

Source of the problems:
* The hosting startup targets need to run a restore during build. The packages it needs to restore against might not exist or might not be in the correct format yet (haven't been code signed). https://github.com/aspnet/AspNetCore/blob/874050f1ddfa80f5fbba7ac87eaaa442de7e0f71/src/SiteExtensions/Sdk/SiteExtension.targets#L75-L85 Leads to errors like this: 
![image](https://user-images.githubusercontent.com/2696087/59453867-b702d280-8dc5-11e9-9adb-fe8b5257aaf5.png)
 
* The targets attempt to restore from the artifacts/ folder. Arcade is currently working on a plan to forbid the use of the 'RestoreSources' property in favor of nuGet.config (so we can use private restore feeds) https://github.com/aspnet/AspNetCore/blob/874050f1ddfa80f5fbba7ac87eaaa442de7e0f71/src/SiteExtensions/Sdk/SiteExtension.targets#L17-L21
* Lots of race conditions in the `ComposeStore` target causes MSBuild errors/warnings about files locked by another process. External issue, nguerrera is working on fixing for 3.0. dotnet/sdk#1211, dotnet/sdk#1072

JunTaoLuo:
Follow up in #11534

PR #11409, opened 2019-06-20
Build SiteExtensions

JunTaoLuo:
Addresses #11186 

WIP since it doesn't work yet.

### DIFF
--- a/src/SiteExtensions/build.cmd
+++ b/src/SiteExtensions/build.cmd
@@ -1,0 +1,25 @@
+@ECHO OFF
+SET RepoRoot=%~dp0..\..
+
+ECHO Building Microsoft.AspNetCore.Runtime.SiteExtension
+CALL %RepoRoot%\build.cmd -projects %~dp0Runtime\Microsoft.AspNetCore.Runtime.SiteExtension.pkgproj /bl:artifacts/log/se-runtime.binlog %*
+
+IF %ERRORLEVEL% NEQ 0 (
+   EXIT /b %ErrorLevel%
+)
+
+ECHO Building LoggingBranch
+CALL %RepoRoot%\build.cmd -projects %~dp0LoggingBranch\LB.csproj /bl:artifacts/log/se-lb.binlog %*
+
+IF %ERRORLEVEL% NEQ 0 (
+   EXIT /b %ErrorLevel%
+)
+
+ECHO Building Microsoft.AspNetCore.AzureAppServices.SiteExtension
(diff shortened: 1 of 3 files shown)


JunTaoLuo:
Build worked locally, testing on CI now. The implementation is a little rough but I want to see if I can remove the need for a build script and use project dependencies instead. Also need to verify the produced site extensions and see if they are correct.

dougbu (approved):
Checks are passing and this looks good despite the need for a new script.

If you can't do it in this PR, please file a cleanup issue to expand the names of the `.binlog` files, they're too abbreviated for most readers and the names could be much longer.

dougbu (commented):
@JunTaoLuo if you're making changes, please expand the binlog filenames e.g. SiteExtensions-LoggingBranch-x64.binlog

JunTaoLuo:
Verifying the contents of the site extension packages. The only issue is the missing ancm folder in AspNetCoreRuntime.3.0.RID's content. Investigating.

jkotalik:
@JunTaoLuo let me know if you need help with this.

src/SiteExtensions/build.cmd
```diff
@@ -0,0 +1,27 @@
...
+)
+
+ECHO Building Microsoft.AspNetCore.AzureAppServices.SiteExtension
+CALL %RepoRoot%\build.cmd -forceCoreMsbuild -projects %~dp0LoggingAggregate\src\Microsoft.AspNetCore.AzureAppServices.SiteExtension\Microsoft.AspNetCore.AzureAppServices.SiteExtension.csproj /bl:artifacts/log/SiteExtensions-LoggingAggregate.binlog %*
+
+IF %ERRORLEVEL% NEQ 0 (
+   EXIT /b %ErrorLevel%
+)
+
+ECHO SiteExtensions successly built!
```
jkotalik:
```suggestion
ECHO SiteExtensions successfully built!
```

jkotalik (approved):
I don't want to block this as it required for preview7, but there are at least a few issues:
- Building site extensions takes 23 minutes
- Calling explicit build steps like this seems likely to break in the future.

I'd say we file a follow-up issue for this and unblock site extension builds for now.

Also, maybe we can consider making it so this only builds on internal ci?

JunTaoLuo:
Sure we can optimize the build later. I compared against some pre-arcade builds and it seems like the difference in build time comes from the SiteExtension builds. https://github.com/aspnet/AspNetCore/issues/11534

PR #11416, opened 2019-06-20
Simplify attribute splatting

rynowak:
Removes handling for IEnumerable<KVP<string, string>>. This isn't
something we really need for the main scenario and it MASSIVELY
complicates the codegen part of the feature. Requiring the dictionary to
be an object-valued dictionary should cover the cases we care about.


rynowak:
@SteveSandersonMS @pranavkm

pranavkm (approved):

PR #11420, opened 2019-06-20
Quarantine some tests

analogrelay:
We don't currently believe there's a product issue here and this is in the top-5 flaky.

(Do wait for a build, I need to verify that I didn't miss a namespace or something like that. I was having trouble getting my machine in a good state to build and decided to save that time by getting the PR out)

### DIFF
--- a/src/Servers/Kestrel/test/FunctionalTests/Http2/ShutdownTests.cs
+++ b/src/Servers/Kestrel/test/FunctionalTests/Http2/ShutdownTests.cs
@@ -102,6 +102,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.FunctionalTests.Http2
         }
 
         [ConditionalFact]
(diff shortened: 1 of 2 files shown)


davidfowl (approved):

jkotalik (approved):

JunTaoLuo (approved):

halter73 (approved):

src/Servers/Kestrel/test/FunctionalTests/Http2/ShutdownTests.cs
```diff
@@ -102,6 +102,7 @@ public async Task GracefulShutdownWaitsForRequestsToFinish()
         }
 
         [ConditionalFact]
+        [Flaky("https://github.com/aspnet/AspNetCore-Internal/issues/2667", FlakyOn.All)]
```
halter73: See https://github.com/aspnet/AspNetCore-Internal/issues/2667#issuecomment-504206902 for my theory on why this is flaky.

Issue #10530, opened 2019-05-24
Consolidate Windows Auth config


Tratcher:
There are now three  ways to enable Windows Authentication:
- The [Negotiate Auth handler](https://github.com/aspnet/AspNetCore.Docs/issues/12420#issuecomment-492446725)
- HttpSys
- IIS/Express (in proc and out of proc)

Every one of them is mutually exclusive. The Negotiate handler throws if you try to use it with HttpSys or IIS because it depends on Kestrel connection state features. This makes for a terrible user experience if you want to develop in one environment but deploy to another (see [Music Store](https://github.com/aspnet/AspNetCore/blob/49c01eefecf1dbd75e5536aa803d689390c6770a/src/MusicStore/samples/MusicStore/Program.cs#L30-L54)).

Proposal:
Make the Negotiate handler the primary mechanism, but have it detect the other scenarios and forward to the appropriate handler:
- IIS In-Proc - forward to IISServerDefaults.AuthenticationScheme "Windows"
- IIS out-of-proc - forward to IISDefaults.AuthenticationScheme "Windows"
- HttpSys - forward to HttpSysDefaults.AuthenticationScheme "Windows"
  - `Configure<HttpSysOptions>(options => options.Authentication.Schemes =
                            AuthenticationSchemes.Negotiate` (if not already set?)

You'd still have to enable Windows Auth in IIS, but the app could use the same portable code across servers.
```
            services.AddAuthentication(NegotiateDefaults.AuthenticationScheme)
                .AddNegotiate();
```

anurse:
We'll look at having the new handler just no-op when it detects that the server supports negotiate directly.

Gonna try this in 3.0 :).

Tratcher:
Verified with SDK 3.0.100-preview7-012742 for in-proc, out-of-proc, and kestrel.

PR #11421, opened 2019-06-20
Consolidate Windows Auth config

Tratcher:
#10530 Enables the Negotiate handler to detect when a server supports integrated auth and defer to it. This makes it easier to use one server for development and another for production.

I did look at using server features for this but there was no way to flow them to an auth handler, they're only exposed to the app in Startup.Configure (e.g. middleware).

The biggest downside of this design is that it doesn't produce an error until the first request.

### DIFF
--- a/src/Security/Authentication/Negotiate/src/NegotiateOptions.cs
+++ b/src/Security/Authentication/Negotiate/src/NegotiateOptions.cs
@@ -33,7 +33,11 @@ namespace Microsoft.AspNetCore.Authentication.Negotiate
         /// </summary>
         public bool PersistNtlmCredentials { get; set; } = true;
 
-        internal bool DefferToServer { get; set; }
+        /// <summary>
+        /// Indicates if integrated server Windows Auth is being used instead of this handler.
+        /// See <see cref="PostConfigureNegotiateOptions"/>.
+        /// </summary>
+        internal bool DeferToServer { get; set; }
 
         // For testing
         internal INegotiateStateFactory StateFactory { get; set; } = new ReflectedNegotiateStateFactory();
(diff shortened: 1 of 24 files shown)


src/Security/Authentication/Negotiate/src/NegotiateOptions.cs
```diff
@@ -33,6 +33,8 @@ public new NegotiateEvents Events
         /// </summary>
         public bool PersistNtlmCredentials { get; set; } = true;
 
+        internal bool DefferToServer { get; set; }
```
DylanDmitri: nit: Deffer -> Defer

src/Security/Authentication/Negotiate/src/PostConfigureNegotiateOptions.cs
```diff
@@ -0,0 +1,64 @@
...
+        public void PostConfigure(string name, NegotiateOptions options)
+        {
+            // If the server supports integrated authentication...
+            if (_serverAuth != null)
+            {
+                // And it's on...
+                if (_serverAuth.IsEnabled)
+                {
+                    // Forward to the server if something else wasn't already configured.
+                    if (options.ForwardDefault == null)
```
HaoK: Is post configure really what you want (as opposed to just doing the initialize of DeferToServer inside the handler ctor itself?  I guess this makes it easier to test if that's the main idea of putting it in the options

HaoK (approved):

src/Security/Authentication/Negotiate/src/PostConfigureNegotiateOptions.cs
Tratcher: *shrug* It's a nice place to consolidate that logic, and it did make it easier to test.

Issue #10583, opened 2019-05-28
ResponseCaching returns no headers in 304 NotModified response


mrjpeterj:
### Describe the bug
ResponseCaching returns no headers in 304 NotModified response

### To Reproduce
1. Setup a server that uses ResponseCaching
2. Make a request that responds with a Last-Modified and Expires header and see that it gets put into the ResponseCache
3. The Response will be a 200 with headers including Last-Modified, Expires, Cache-Control
3. Make the same request again, but add the If-Modified-Since header with a value the same as the previously returned Last-Modified time
4. The Response will be a 304 but it won't have any headers.

### Expected behavior
The 304 response should contain a number of headers as specified in RFC 7232

https://tools.ietf.org/html/rfc7232#section-4.1

```
   The server generating a 304 response MUST generate any of the
   following header fields that would have been sent in a 200 (OK)
   response to the same request: Cache-Control, Content-Location, Date,
   ETag, Expires, and Vary.
```

### Additional context

You can see this quite clearly in https://github.com/aspnet/AspNetCore/blob/master/src/Middleware/ResponseCaching/src/ResponseCachingMiddleware.cs#L156

mikaelm12:
cc @JunTaoLuo

anurse:
I can repro this and it does seem like it's contrary to the spec. @mrjpeterj is there a specific scenario that this affects or is this just about complying with the spec? I agree it is important to do so, but just trying to get an idea of the priority of this, if it's blocking a user scenario then it would be a higher priority.

mrjpeterj:
Well it did crop up in implementing something for my job, but it turned out that for that specific thing I could just turn off server side caching to work around the issue. (setting Cache-Control: private)

The endpoint is using a MemoryCache to return the result anyway so it still performs fine.

If I was desperate for it, then I could implement the fix myself and submit it.

anurse:
Ok, thanks for the context there! Like I said, I don't disagree it's a thing we should look at, but we likely won't get to it in 3.0 unless you (or someone else) is interested in submitting a PR. We'll review it again after 3.0 is done. Thanks for the report!

alefranz:
I will give it a go

alefranz:
This should have been fixed in #11428

anurse:
Indeed! Unfortunately "Addresses" doesn't close the issue out when the PR is closed :).

PR #11428, opened 2019-06-20
ResponseCaching: include required headers in 304

alefranz:
Include the required headers for 304 response according to RFC 7232
https://tools.ietf.org/html/rfc7232#section-4.1

> The server generating a 304 response MUST generate any of the
>  following header fields that would have been sent in a 200 (OK)
>  response to the same request: Cache-Control, Content-Location, Date,
>  ETag, Expires, and Vary.

Addresses #10583

/cc @anurse

### DIFF
--- a/src/Middleware/ResponseCaching/src/ResponseCachingMiddleware.cs
+++ b/src/Middleware/ResponseCaching/src/ResponseCachingMiddleware.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
@@ -19,6 +19,10 @@ namespace Microsoft.AspNetCore.ResponseCaching
     public class ResponseCachingMiddleware
     {
         private static readonly TimeSpan DefaultExpirationTimeSpan = TimeSpan.FromSeconds(10);
+
+        // see https://tools.ietf.org/html/rfc7232#section-4.1
+        private static readonly string[] HeadersToIncludeIn304 =
+            new[] { "Cache-Control", "Content-Location", "Date", "ETag", "Expires", "Vary" };
 
         private readonly RequestDelegate _next;
         private readonly ResponseCachingOptions _options;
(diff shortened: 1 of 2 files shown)


src/Middleware/ResponseCaching/src/ResponseCachingMiddleware.cs
```diff
@@ -157,6 +161,18 @@ internal async Task<bool> TryServeCachedResponseAsync(ResponseCachingContext con
                 {
                     _logger.NotModifiedServed();
                     context.HttpContext.Response.StatusCode = StatusCodes.Status304NotModified;
+
+                    if (context.CachedResponseHeaders != null)
+                    {
+                        for (var i = 0; i < HeadersToIncludeIn304.Length; i++)
```
DylanDmitri:
```suggestion
                        for (var key in HeadersToIncludeIn304)
```

src/Middleware/ResponseCaching/src/ResponseCachingMiddleware.cs
```diff
@@ -157,6 +161,18 @@ internal async Task<bool> TryServeCachedResponseAsync(ResponseCachingContext con
...
+                        {
+                            var key = HeadersToIncludeIn304[i];
```
DylanDmitri:
```suggestion
                            
```

src/Middleware/ResponseCaching/src/ResponseCachingMiddleware.cs
alefranz: you mean `foreach` I guess. I thought a `for` loop was more performant here but it looks like I was wrong.

halter73 (approved):

halter73:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=232903

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/2410

src/Middleware/ResponseCaching/src/ResponseCachingMiddleware.cs
anurse: `foreach` and `for` should be comparable when the type you're iterating over is "concrete" rather than an interface. For example `foreach`ing over a `string[]` or `List<string>` is *fast*, but for `IList<string>`, `IEnumerable<string>` using `for` is faster. Interface-dispatch makes the different ;).

alefranz:
I believe the failure is unrelated. let's see if I can manage to run it again

anurse:
@alefranz Don't worry about Helix, it isn't required to merge. As long as the failures aren't related, we can go ahead. @halter73 can you merge? (assuming you still stand by your approval 😄)

Issue #11312, opened 2019-06-18
Remaining pubternal APIs in the shared framework


davidfowl:
We still have to look at packages but I haven't done that yet: I want to make sure these were intentional and not forgotten. If they were intentional I'd like to understand why they weren't made public APIs

- [x] - [Microsoft.AspNetCore.Authentication.Internal](https://github.com/aspnet/AspNetCore/blob/9662a5bf542654b1ed1f8007efe068bb7063afcf/api-diffs/api-diffs_Microsoft.AspNetCore.Authentication.Internal.md) (@HaoK @Tratcher, @Pilchie)
- [x] - [Microsoft.AspNetCore.Builder.Internal](https://github.com/aspnet/AspNetCore/blob/9662a5bf542654b1ed1f8007efe068bb7063afcf/api-diffs/api-diffs_Microsoft.AspNetCore.Builder.Internal.md) (@Tratcher @davidfowl, @anurse)
- [x] - [Microsoft.AspNetCore.DataProtection.Internal](https://github.com/aspnet/AspNetCore/blob/9662a5bf542654b1ed1f8007efe068bb7063afcf/api-diffs/api-diffs_Microsoft.AspNetCore.DataProtection.Internal.md) (@natemcmaster @blowdart)
- [x] - [Microsoft.AspNetCore.DataProtection.Cng.Internal](https://github.com/aspnet/AspNetCore/blob/ab7f1caaf21630f500837224a3fd8a1fc53448d1/api-diffs/api-diffs_Microsoft.AspNetCore.DataProtection.Cng.Internal.md) (@blowdart )
- [x] - [Microsoft.AspNetCore.Hosting.Internal](https://github.com/aspnet/AspNetCore/blob/9662a5bf542654b1ed1f8007efe068bb7063afcf/api-diffs/api-diffs_Microsoft.AspNetCore.Hosting.Internal.md) (@Tratcher @davidfowl @anurse )
- [x] - [Microsoft.AspNetCore.Http.Internal](https://github.com/aspnet/AspNetCore/blob/9662a5bf542654b1ed1f8007efe068bb7063afcf/api-diffs/api-diffs_Microsoft.AspNetCore.Http.Internal.md) (@Tratcher @davidfowl @anurse)
[...]

Issue #2979, opened 2018-03-19
Move BindingAddress to public namespace


davidfowl:
I'm not sure why this was made pubternal instead of public and why we're even using pubternal APIs across assemblies. We should fix this and BindingAddress should be public.

PS: This will break https://github.com/aspnet/BasicMiddleware/blob/726035e3fb8c4ee31a32522c16ff73278b9b13e7/src/Microsoft.AspNetCore.HttpsPolicy/HttpsRedirectionMiddleware.cs#L124, I'm not sure where else we use it.

/cc @Tratcher @jkotalik

anurse:
We should actually just look at making it `internal`.

Tratcher:
By which we mean shared source.

anurse:
Yes, as a result of making it `internal` we'll need to shared-source it to "share" it properly.

davidfowl:
I'd like to internal all the things as part of preview7 (or make it public why it think it should be public)

PR #11429, opened 2019-06-20
Remove or make public pubternals for Http

Tratcher:
#11312, #2979 Removes Microsoft.AspNetCore.Builder.Internal and Microsoft.AspNetCore.Http.Internal by making some types fully public and others fully internal.
Public:
- BindingAddress #2979 - Used by Kestrel and HttpsRedirection
- BufferingHelper - Often used by customers.
- ApplicationBuilder - Used in many tests
- FormFile & FormFileCollection - Used in many tests, hard to replace.
- QueryCollection - Used by many tests

Internal:
- DefaultConnectionInfo/HttpRequest/HttpResponse/WebSocketManager/ItemsDictionary - Only used internally by DefaultHttpContext.
- RequestCookieCollection  - Used by some tests but replaceable. (MVC, @pranavkm)
- ResponseCookies


src/Http/Http/ref/Microsoft.AspNetCore.Http.netcoreapp3.0.cs
```diff
@@ -17,6 +17,25 @@ public partial class ApplicationBuilder : Microsoft.AspNetCore.Builder.IApplicat
...
+        public string Scheme { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
+        public string UnixPipePath { get { throw null; } }
+        public override bool Equals(object obj) { throw null; }
+        public override int GetHashCode() { throw null; }
+        public static Microsoft.AspNetCore.Http.BindingAddress Parse(string address) { throw null; }
+        public override string ToString() { throw null; }
+    }
+    public static partial class BufferingHelper
+    {
+        public static Microsoft.AspNetCore.Http.HttpRequest EnableRewind(this Microsoft.AspNetCore.Http.HttpRequest request, int bufferThreshold = 30720, long? bufferLimit = default(long?)) { throw null; }
```
davidfowl: Why do we expose this when we have EnableBuffering on the HTTP request?

src/Http/Http/ref/Microsoft.AspNetCore.Http.netcoreapp3.0.cs
```diff
@@ -17,6 +17,25 @@ public partial class ApplicationBuilder : Microsoft.AspNetCore.Builder.IApplicat
...
+        public string PathBase { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
+        public int Port { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
+        public string Scheme { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
+        public string UnixPipePath { get { throw null; } }
+        public override bool Equals(object obj) { throw null; }
+        public override int GetHashCode() { throw null; }
+        public static Microsoft.AspNetCore.Http.BindingAddress Parse(string address) { throw null; }
+        public override string ToString() { throw null; }
+    }
+    public static partial class BufferingHelper
```
davidfowl: I think this should be removed and we should just add the useful method to one of the existing extensions classes.

src/Http/Http/src/Internal/DefaultHttpRequest.cs
```diff
@@ -10,10 +10,13 @@
 using Microsoft.AspNetCore.Routing;
 using Microsoft.Net.Http.Headers;
 
-namespace Microsoft.AspNetCore.Http.Internal
+namespace Microsoft.AspNetCore.Http
 {
-    public sealed class DefaultHttpRequest : HttpRequest
+    internal sealed class DefaultHttpRequest : HttpRequest
```
davidfowl: I feel like we should leave these public?

halter73: While changing namespaces?

davidfowl: Yep.

src/Http/Http/src/Internal/RequestCookieCollection.cs
```diff
@@ -6,9 +6,9 @@
 using System.Collections.Generic;
 using Microsoft.Net.Http.Headers;
 
-namespace Microsoft.AspNetCore.Http.Internal
+namespace Microsoft.AspNetCore.Http
 {
-    public class RequestCookieCollection : IRequestCookieCollection
+    internal class RequestCookieCollection : IRequestCookieCollection
```
halter73: Are we trying to get rid of internal namespaces too? Or are we fine with leaving internal types in internal namespaces?

src/Http/Http/src/Internal/DefaultHttpRequest.cs
davidfowl: It feels odd to me to expose the DefaultHttpContext but not the rest of the types. Maybe I'm just being picky.

src/Http/Http/src/Internal/RequestCookieCollection.cs
davidfowl: We can leave the types in the internal namespace and treat it as public but only if we think the usage is *really* high that it would break a large number of mainstream use-cases if we did so. We can probably get some data from apisof.net as well.

src/Http/Http/src/Internal/DefaultHttpRequest.cs
Tratcher: Why? They aren't used anywhere, nor very useful in isolation.

src/Http/Http/src/Internal/RequestCookieCollection.cs
Tratcher: The internal namespace seems redundant.

halter73: I don't disagree. The only reason I think we didn't change namespaces before was to reduce the diff. I just like consistency. The good news is now we can change Kestrel's internal namespaces whenever we want.

Tratcher:
Found a few more in in Http.Abstractions that all became internal:
- HeaderSegment 
- HeaderSegmentCollection 
- ParsingHelpers

src/Http/Http/src/Internal/DefaultHttpRequest.cs
Tratcher: I found one usage in our entire repo and it was unnecessary. We can bring them back if we find a scenario for them.

davidfowl (approved):
When you write up the announcement for this make sure you mention some of the common alternatives to using these pubternal APIs

Tratcher:
Can we do one announcement for all of #11312? Doing one for every feature area gets noisy.

davidfowl:
Sure

src/Http/Http/src/Internal/DefaultHttpRequest.cs
glucaci: Why this was made internal? We are using it in our tests where for example have an extension method on HttpRequest and want to test it without to create an entire TestHost.

Tratcher: You can still make a DefaultHttpContext to get HttpRequest.

glucaci: Sure, thanks!

PR #11434, opened 2019-06-21
Use Enterprise Queue to reduce provisioning issues

JunTaoLuo:
Switch to enterprise queue for better provisioning. This should help with our "Abandoned builds" problem. e.g. https://dnceng.visualstudio.com/internal/_build/results?buildId=232662.


MattGal (approved):
LGTM. Note we know we have "Windows provisioning is slow" issues beyond just the .BT queue, this one just seems extra bad.  @ulisesh and I are hoping to address this soon.

dougbu (approved):
once you get tests passing&hellip;

PR #11438, opened 2019-06-21
For two-way bindings, enforce consistency between .NET model and DOM by patching old tree. Fixes #8204

### LINKED ISSUE #8204: Components bind-value-oninput can lose cursor position with fast typing

SteveSandersonMS:
This is the rather nontrivial solution to a multitude of related async-binding and diff-inconsistency issues. As of this change, two-way bindings have a stricter behavior whereby they:

 * Find out from the JS code what the latest state in the DOM is, and apply that to the old render tree
 * Consequently, produce a diff output that may be smaller (no need to tell the client to perform changes it has already performed, which fixes the laggy typing thing), or may be larger (maybe we need to tell the client to revert its change, which fixes the "moving checkboxes" case we had reported by a customer)

In most ways this is purely advantageous. Various cases where the behavior was bizarre or broken before are now fixed (e.g., binding a checkbox to a property whose value is fixed as `true` or `false` before didn't reflect that in the UI - now the checkbox will forcibly revert itself to "checked" or "unchecked" as appropriate"). Also it's much safer to revert invalid edits to number fields, because now the user can see exactly how their input is being interpreted - previously they had no way to know if their edit was being ignored because it was considered invalid.

However the extra strictness will make developers go through extra hoops in a few cases. The most notable one is textboxes bound to non-nullable `int`/`double/`etc that use `bind-value:event="oninput"`. These will no longer allow unparseable values such as being empty or just containing `-`, so developers will have to either:

 * Not use `bind-value:event="oninput"` for non-nullable numbers, since you don't want to prevent the invalid intermediate states. Just use the default change event which is "change".
 * Or, use a nullable version of your number, or even bind to a `string` so you can do whatever you want with the invalid intermediate states
 * Or, use `<InputNumber>` which already handles this nicely. In the future we might extend that to support `bind-value:event="oninput"`, and it will do what you expect because behind the scenes it's binding to a `string` and handling parsing separately so it can record validation errors.


SteveSandersonMS:
@dougbu This PR contains some of the commits from dotnet-maestro, because I'm waiting for https://github.com/aspnet/AspNetCore/pull/11307. As long as the build for this PR passes, I think it should be fine to merge those commits as part of this PR. Or if #11307 gets merged first, it will make no difference after being rebased.

SteveSandersonMS:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=233678

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/2707

dougbu:
@SteveSandersonMS I see you force pushed the branch and no dependency updates remain. Do you still need anything from me here?

SteveSandersonMS:
@dougbu No, nothing else needed. Unless you’re part of the BRT or whatever the current equivalent to that is, because there is still https://github.com/aspnet/AspNetCore-Internal/issues/2707

SteveSandersonMS:
@aspnet-hello Triage https://dev.azure.com/dnceng/public/_build/results?buildId=235994&view=logs&jobId=1bca1137-cedc-5e44-af9b-a27ff406bb70

src/Components/Components/src/Rendering/RenderTreeUpdater.cs
```diff
@@ -0,0 +1,113 @@
...
+
+namespace Microsoft.AspNetCore.Components.Rendering
+{
+    internal class RenderTreeUpdater
+    {
+        public static void UpdateToMatchClientState(RenderTreeBuilder renderTreeBuilder, int eventHandlerId, object newFieldValue)
+        {
+            // We only allow the client to supply string or bool currently, since those are the only kinds of
+            // values we output on attributes that go to the client
+            if (!(newFieldValue is string || newFieldValue is bool))
```
javiercn: In which case can this not be either an int or bool, wouldn't that signal that something went terribly wrong? (Like could a number get in here somehow?)

src/Components/Components/src/Rendering/RenderTreeUpdater.cs
```diff
@@ -0,0 +1,113 @@
...
+                    break;
+                }
+
+                if (attributeFrame.AttributeName == attributeName)
+                {
+                    // Found an existing attribute we can update
+                    attributeFrame = attributeFrame.WithAttributeValue(attributeValue);
+                    return;
+                }
+            }
```
javiercn: Is it ok to silently ignore the update if we can't find the attribute?

src/Components/Components/src/Rendering/Renderer.cs
```diff
@@ -205,11 +206,12 @@ private ComponentState AttachAndInitComponent(IComponent component, int parentCo
         /// </summary>
         /// <param name="eventHandlerId">The <see cref="RenderTreeFrame.AttributeEventHandlerId"/> value from the original event attribute.</param>
         /// <param name="eventArgs">Arguments to be passed to the event handler.</param>
+        /// <param name="fieldInfo">Information that the renderer can use to update the state of the existing render tree to match the UI.</param>
         /// <returns>
         /// A <see cref="Task"/> which will complete once all asynchronous processing related to the event
         /// has completed.
         /// </returns>
-        public virtual Task DispatchEventAsync(int eventHandlerId, UIEventArgs eventArgs)
+        public virtual Task DispatchEventAsync(int eventHandlerId, EventFieldInfo fieldInfo, UIEventArgs eventArgs)
```
javiercn: Is the expectation that this chain grows bigger than 1 when there are rapidly firing events?

src/Components/Components/src/Rendering/RenderTreeUpdater.cs
SteveSandersonMS: Good point.

I'm not aware of any valid way that the value here might be something other than `string`/`bool`. If it was, we could just throw. Maybe that would be clearer.

javiercn: Looking at your tests, it's fine if this simply returns as the value comes from the client and is untrusted

src/Components/Components/src/Rendering/Renderer.cs
SteveSandersonMS: The chain of replaced event handler IDs? If so, yes, it can grow beyond 1.

src/Components/Components/test/RendererTest.cs
```diff
@@ -3271,6 +3271,105 @@ public void DisposingRenderer_CapturesExceptionsFromAllRegisteredComponents()
...
+                    // The only edit is updating the event handler ID, since the test component
+                    // deliberately uses a capturing lambda. The whole point of this test is to
+                    // show that the diff does *not* update the BoundString value attribute.
+                    Assert.Equal(RenderTreeEditType.SetAttribute, edit.Type);
+                    var attributeFrame = latestBatch.ReferenceFrames[edit.ReferenceFrameIndex];
+                    AssertFrame.Attribute(attributeFrame, "ontestevent", typeof(Action<UIChangeEventArgs>));
+                    Assert.NotEqual(0, attributeFrame.AttributeEventHandlerId);
+                    Assert.NotEqual(eventHandlerId, attributeFrame.AttributeEventHandlerId);
+                });
+            }
```
javiercn: Would this for example be a case where there are multiple events fired with the old event handler (like typing fast or moving the cursor quick) and the render batches have not been produced yet?

SteveSandersonMS: Yes, exactly!

javiercn (approved):
Great set of changes.

I didn't get to follow all the details because its very low-level and likely requires some debugging to understand it fully, but the changes look solid and the test coverage is appropriate.

Great job!

src/Components/Components/ref/Microsoft.AspNetCore.Components.netstandard2.0.cs
```diff
@@ -688,6 +688,12 @@ namespace Microsoft.AspNetCore.Components.Rendering
         public int ComponentId { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
         public System.Collections.Generic.IEnumerable<string> Tokens { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
     }
+    public partial class EventFieldInfo
+    {
+        public EventFieldInfo() { }
+        public int ComponentId { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
+        public object FieldValue { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute]set { } }
```
pranavkm: Filed https://github.com/aspnet/AspNetCore/issues/11517 in case we have interest in doing the work here.

src/Components/Components/src/EventCallbackFactoryBinderExtensions.cs
```diff
@@ -687,13 +687,26 @@ static DateTime ConvertDateTime(object obj, string format)
                 {
                 }
 
-                // We only invoke the setter if the conversion didn't throw. This is valuable because it allows us to attempt
-                // to process invalid input but avoid dirtying the state of the component if can't be converted. Imagine if
-                // we assigned default(T) on failure - this would result in trouncing the user's typed in value.
+                // We only invoke the setter if the conversion didn't throw, or if the newly-entered value is empty.
+                // If the user entered some non-empty value we couldn't parse, we leave the state of the .NET field
```
pranavkm: 👍

src/Components/Components/src/Rendering/RenderTreeUpdater.cs
pranavkm: This might help a bit - https://github.com/aspnet/AspNetCore/issues/11517

pranavkm (approved):

PR #11460, opened 2019-06-22
Add JsonExtensionData to ProblemDetails

### LINKED ISSUE #6202: Figure out ProblemDetails + System.Text.Json
* [ ] - `ProblemDetails` relied on `JsonExtensionAttribute` to ensure that swagger document generators produced the correct JsonSchema. Since the attribute would no longer be available, we need  to figure out an alternate way to describe this to document generators.

pranavkm:
Fixes https://github.com/aspnet/AspNetCore/issues/6202

### DIFF
--- a/src/Mvc/Mvc.Core/src/Formatters/SystemTextJsonOutputFormatter.cs
+++ b/src/Mvc/Mvc.Core/src/Formatters/SystemTextJsonOutputFormatter.cs
@@ -59,7 +59,12 @@ namespace Microsoft.AspNetCore.Mvc.Formatters
             var writeStream = GetWriteStream(httpContext, selectedEncoding);
             try
             {
-                await JsonSerializer.WriteAsync(writeStream, context.Object, context.ObjectType, SerializerOptions);
+                // context.ObjectType reflects the declared model type when specified.
+                // For polymorphic scenarios where the user declares a return type, but returns a derived type,
+                // we want to serialize all the properties on the derived type. This keeps parity with
+                // the behavior you get when the user does not declare the return type and with Json.Net at least at the top level.
+                var objectType = context.Object?.GetType() ?? context.ObjectType;
+                await JsonSerializer.WriteAsync(writeStream, context.Object, objectType, SerializerOptions);
                 await writeStream.FlushAsync();
             }
             finally
--- a/src/Mvc/test/Mvc.FunctionalTests/ApiBehaviorTest.cs
+++ b/src/Mvc/test/Mvc.FunctionalTests/ApiBehaviorTest.cs
@@ -2,7 +2,6 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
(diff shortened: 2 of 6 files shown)


src/Mvc/Mvc.Core/src/Formatters/SystemTextJsonOutputFormatter.cs
```diff
@@ -59,7 +59,12 @@ public sealed override async Task WriteResponseBodyAsync(OutputFormatterWriteCon
             var writeStream = GetWriteStream(httpContext, selectedEncoding);
             try
             {
-                await JsonSerializer.WriteAsync(writeStream, context.Object, context.ObjectType, SerializerOptions);
+                // context.ObjectType reflects the declared model type when specified.
+                // For polymorphic scenarios where the user declares a return type, but returns a derived type,
+                // we want to serialize all the properties on the derived type. This keeps parity with
+                // the behavior you get when the user does not declare the return type and with Json.Net at least at the top level.
+                var objectType = context.Object?.GetType() ?? context.ObjectType;
```
pranavkm: What do you think? I considered leaving it as is, but you get a different outcome depending on whether you do `ActionResult<T>` versus returning an `ObjectResult`. Doing this felt like it was moderately better. Polymorphic serialization does not work for properties, but I think there's plans for a feature there.

src/Mvc/test/Mvc.FunctionalTests/ApiBehaviorTest.cs
```diff
@@ -364,4 +332,132 @@ public async Task SerializingValidationProblemDetails_WithExtensionData()
...
+                    var error = Assert.Single(kvp.Value);
+                    Assert.Equal("The field Zip must match the regular expression '\\d{5}'.", error);
+                }
+            );
+
+            Assert.Collection(
+                problemDetails.Extensions,
+                 kvp =>
+                 {
+                     Assert.Equal("extensions", kvp.Key);
```
pranavkm: This is kinda busted - https://github.com/dotnet/corefx/issues/38769. I've added a test to make sure the rest of Json is correct. If we get this in, the most we'd have to do to "react" is to fix up tests

src/Mvc/Mvc.Core/src/Formatters/SystemTextJsonOutputFormatter.cs
rynowak: I think this is the right thing to do given that JSON.net does it.

src/Mvc/test/Mvc.FunctionalTests/ApiBehaviorTest.cs
rynowak: Should we skip this test instead? I don't really want to encode wrong behavior in tests. It's better to just skip if possible.

Put an other way, I don't really want a test that partially validates the expectations.

rynowak (approved):

src/Mvc/test/Mvc.FunctionalTests/ApiBehaviorTest.cs
pranavkm: Yeah, seems reasonable. It's likely https://github.com/dotnet/corefx/issues/38769 will get fixed for preview7.

PR #11466, opened 2019-06-22
Sign Windows zips

JunTaoLuo:
Signing of the windows zip was broken during the arcade conversion. Signing.props is imported by the Arcade Sign.proj. However, it doesn't import Directory.Build.props which defines `InstallersOutputPath`. `ArtifactsDir` is  defined so let's use that instead.

I'll do a quick check of the other artifacts we expect to be signed.


JunTaoLuo:
cc @johnbeisner This should resolve the issue we spoke about. I hope to get this in over the weekend so you have a build to work with on Monday.

dougbu (commented):
Well, I would have said :ship::it: 😈

PR #11412, opened 2019-06-20
Remove connection adapters and move things to middleware

davidfowl:
- Remove connection adapters from the public API surface (pubternal) and replace the existing adapters with connection middleware.
- Updated the tests

Fixes #11402

PR #11474, opened 2019-06-23
PR feedback from #11412

davidfowl:
- Remove ExpectedMiddlewareCount since everything is middleware now
- Renamed everything adapter to middleware
- Added a regression test for an https scenario
- Don't send client certs for tests that don't expect it

PS: This was supposed to take 30 mins...


src/Servers/Kestrel/Core/test/AddressBinderTests.cs
```diff
@@ -149,5 +153,39 @@ public async Task FallbackToIPv4WhenIPv6AnyBindFails(string address)
...
+            var options = new KestrelServerOptions()
+            {
+                // This stops the dev cert from being loaded
+                IsDevCertLoaded = true,
+                ApplicationServices = services.BuildServiceProvider()
+            };
+
+            options.ConfigureEndpointDefaults(e =>
+            {
+                if (e.IPEndPoint.Port == 5001)
```
halter73: If you didn't have `IsDevCertLoaded = true,` above, this test would pass even without the custom ConfigureEndpointDefaults, right?

I think this test would be a little more convincing if it made port 5000 the TLS endpoint.

halter73:
Why don't you rename `HttpsConnectionAdapterTests` to `HttpsConnectionMiddlewareTests`? :trollface:

halter73 (approved):

halter73 (approved):

src/Servers/Kestrel/Core/test/AddressBinderTests.cs
davidfowl: The default binding only checks 5001 for IsTls

Issue #10110, opened 2019-05-09
In some cases, ApiExplorer does not flatten query-bound models correctly


domaindrivendev:
### Describe the bug
In some cases, ApiExplorer does not "flatten" the properties of a complex, query-bound model correctly and this has a downstream impact on Swagger generation libraries such as Swashbuckle.

### To Reproduce

1. Using ASP.NET Core 3.0.0-preview
2. Create an action with the following query-bound model
```csharp

[HttpGet]
public void Get([FromQuery]TestModel testModel)
{ }

...

public class TestModel
{
    public string Id { get; set; }
    public TestSubmodel SubModel {get; set; }
    public TestSubmodel SubModel2 { get; set; }
}

public class TestSubmodel
{
    public string SubId { get; set; }
    public TestSubSubModel SubSubModel { get; set; }
}

public class TestSubSubModel
{
    public string SubSubId { get; set; }
    public double SubSubDouble { get; set; }
}
```
3. Run the application and observe the corresponding `ApiDescription` that's surfaced by ApiExplorer. Note that the first property of type `TestSubmodel` is flattened into multiple `ApiParameterDescriptions` correctly BUT the second property, which is of the same type, is not.
```
ParameterDescriptions: [
    {
        Name: "Id",
        Type: "System.String"
    },
    {
        Name: "SubModel.SubId",
        Type: "System.String"
    },
    {
        Name: "SubModel.SubSubModel.SubSubId",
        Type: "System.String"
    },
    {
        Name: "SubModel.SubSubModel.SubSubDouble",
        Type: "System.Double"
    },
    {
        Name: "SubModel2.SubId",
        Type: "System.String"
    },
    {
        Name: "SubModel2.SubSubModel", 
        Type: "NetCore3.Controllers.TestSubSubModel"
    }
]
```
*Why was `SubModel2.SubSubModel` not flattened into two parameters in the same way `SubModel.SubSubModel` was?*

### Expected behavior
```
ParameterDescriptions: [
    {
        Name: "Id",
        Type: "System.String"
    },
    {
        Name: "SubModel.SubId",
        Type: "System.String"
    },
    {
        Name: "SubModel.SubSubModel.SubSubId",
        Type: "System.String"
    },
    {
        Name: "SubModel.SubSubModel.SubSubDouble",
        Type: "System.Double"
    },
    {
        Name: "SubModel2.SubId",
        Type: "System.String"
    },
    {
[...]

PR #11482, opened 2019-06-23
Api Explorer: multiple complex properties

alefranz:
Currently the API Explorer Visitor logic cycles detection is too greedy, so if the same complex type is referenced by two different properties, the second time is skipped.

This PR pop the node after it has been visited, so that while still stopping the visitor when detecting a cycle, it now allows to have the same model referenced by different properties.

Addresses #10110

_Apologies for the edits, I found a better solution after submitting the PR_


rynowak (approved):

rynowak:
This is great! Thanks.

Issue #10872, opened 2019-06-05
Adopt Bedrock Client abstractions in SignalR client


anurse:
Epic #10869

SignalR currently uses the "Bedrock" abstractions on the server side (`ConnectionContext`, etc.) but has a custom client-side abstraction (`IConnectionFactory`). Bedrock has evolved to support a client-side abstraction now so SignalR should update to use it (since it would be a breaking change).

The new client abstraction shape is:

```csharp
public interface IConnectionFactory
{
    ValueTask<ConnectionContext> ConnectAsync(System.Net.EndPoint endpoint, CancellationToken cancellationToken = default);
}
```

Some elements to this:
* Add `IConnectionFactory` (as above) somewhere ([Connections.Abstractions](https://github.com/aspnet/AspNetCore/tree/master/src/Servers/Connections.Abstractions/src)? or a new library?)
* Add `HttpEndpont` sub-class of `System.Net.Endpoint` (for use in the `IConnectionFactory.ConnectAsync` method).
* Configure the `TransferFormat` in the constructor instead of in the `ConnectAsync` method?
* Delete existing connection factory and change to use the new abstraction.

anurse:
TBH, this is another thing making me nervous :).

cc @davidfowl to review my assessment of the elements needed to finish this.

davidfowl:
This looks good. We need to rename the interface to something that says client.

anurse:
![A man standing in a mirror, we see his reflection. He raises his hand in a thumbs-up gesture and says "We did it"](https://www.bing.com/th?id=OGC.3b320d1565afaa59c6451a91399e5156&pid=1.7&rurl=https%3a%2f%2fmedia.giphy.com%2fmedia%2f3ornkf4SAQutQCpLgc%2fgiphy.gif&ehk=0DOxbah74IMCJkssihcvGw)

PR #11484, opened 2019-06-24
Adopt Bedrock client abstractions in SignalR client

halter73:
This add IConnectionFactory and HttpEndPoint to Connections.Abtractions and implements it with the SignalR .NET Client's HttpConnectionFactory.

So far I've been pretty conservative with these changes trying to break as few people as possible while still removing the old SignalR.Client.IConnectionFactory.

Going forward we might want to consider something like adding a HubConnection.StartAsync overload that takes an EndPoint.

Addresses #10872


src/SignalR/clients/csharp/Client/src/HttpConnectionFactory.cs
```diff
@@ -16,16 +18,23 @@ namespace Microsoft.AspNetCore.SignalR.Client
     /// </summary>
     public class HttpConnectionFactory : IConnectionFactory
```
davidfowl: Make this internal. I want to make this properly public in the right assembly

src/Servers/Connections.Abstractions/src/HttpEndPoint.cs
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Net;
+
+namespace Microsoft.AspNetCore.Connections
+{
+    public class HttpEndPoint : EndPoint
```
davidfowl: I wonder if this should be in abstractions or in the https.connections.common assembly.

src/SignalR/clients/csharp/Client.Core/src/HubConnectionBuilder.cs
```diff
@@ -48,6 +52,12 @@ public HubConnection Build()
                 throw new InvalidOperationException($"Cannot create {nameof(HubConnection)} instance. An {nameof(IConnectionFactory)} was not configured.");
             }
 
+            var endPoint = serviceProvider.GetService<EndPoint>();
```
davidfowl: I'm not a fan of having the EndPoint be directly in the container, it needs to hang off some option that gets passed to the HubConnection instead.

src/SignalR/clients/csharp/Client/src/HttpConnectionFactory.cs
```diff
@@ -2,10 +2,12 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Net;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Connections;
 using Microsoft.AspNetCore.Http.Connections.Client;
+using Microsoft.AspNetCore.SignalR.Protocol;
```
davidfowl: This layering is backwards but I'm ok with it if we make this type internal.

src/SignalR/clients/csharp/Client/src/HttpConnectionFactory.cs
```diff
@@ -36,17 +45,46 @@ public HttpConnectionFactory(IOptions<HttpConnectionOptions> options, ILoggerFac
...
+            {
+                throw new NotSupportedException($"The provided {nameof(EndPoint)} must be of type {nameof(HttpEndPoint)}.");
+            }
+
+            if (_httpConnectionOptions.Url != null && _httpConnectionOptions.Url != castedHttpEndPoint.Url)
+            {
+                throw new InvalidOperationException($"If {nameof(HttpConnectionOptions)}.{nameof(HttpConnectionOptions.Url)} was set, it must match the {nameof(HttpEndPoint)}.{nameof(HttpEndPoint.Url)} passed to {nameof(ConnectAsync)}.");
+            }
+
+            var shallowCopiedOptions = ShallowCopyHttpConnectionOptions(_httpConnectionOptions);
```
davidfowl: Comments as to why we copy here.

src/SignalR/clients/csharp/Client/src/HttpConnectionFactory.cs
```diff
@@ -36,17 +45,46 @@ public HttpConnectionFactory(IOptions<HttpConnectionOptions> options, ILoggerFac
...
+        public async ValueTask<ConnectionContext> ConnectAsync(EndPoint httpEndPoint, CancellationToken cancellationToken = default)
         {
-            var connection = new HttpConnection(_httpConnectionOptions, _loggerFactory);
+            if (httpEndPoint == null)
+            {
+                throw new ArgumentNullException(nameof(httpEndPoint));
+            }
+
+            var castedHttpEndPoint = httpEndPoint as HttpEndPoint;
+            if (castedHttpEndPoint == null)
```
davidfowl: Use C# 7.something:

```C#
if (endPointis HttpEndPoint httpEndPoint)
```

src/SignalR/clients/csharp/Client/src/HttpConnectionFactory.cs
```diff
@@ -36,17 +45,46 @@ public HttpConnectionFactory(IOptions<HttpConnectionOptions> options, ILoggerFac
...
+        /// <summary>
+        /// Creates a new connection to an <see cref="HttpEndPoint"/>.
+        /// </summary>
+        /// <param name="httpEndPoint">The <see cref="HttpEndPoint"/> to connect to.</param>
+        /// <param name="cancellationToken">The token to monitor for cancellation requests. The default value is <see cref="CancellationToken.None" />.</param>
+        /// <returns>
+        /// A <see cref="ValueTask{TResult}"/> that represents the asynchronous connect.
+        /// The <see cref="Task{TResult}.Result"/> property returns a <see cref="ConnectionContext"/> for the new connection.
+        /// </returns>
+        public async ValueTask<ConnectionContext> ConnectAsync(EndPoint httpEndPoint, CancellationToken cancellationToken = default)
```
davidfowl: Rename httpEndPoint to endPoint

src/SignalR/clients/csharp/Client/src/HttpConnectionFactory.cs
```diff
@@ -36,17 +45,46 @@ public HttpConnectionFactory(IOptions<HttpConnectionOptions> options, ILoggerFac
...
+
+            var castedHttpEndPoint = httpEndPoint as HttpEndPoint;
+            if (castedHttpEndPoint == null)
+            {
+                throw new NotSupportedException($"The provided {nameof(EndPoint)} must be of type {nameof(HttpEndPoint)}.");
+            }
+
+            if (_httpConnectionOptions.Url != null && _httpConnectionOptions.Url != castedHttpEndPoint.Url)
+            {
+                throw new InvalidOperationException($"If {nameof(HttpConnectionOptions)}.{nameof(HttpConnectionOptions.Url)} was set, it must match the {nameof(HttpEndPoint)}.{nameof(HttpEndPoint.Url)} passed to {nameof(ConnectAsync)}.");
```
davidfowl: How would this happen?

src/SignalR/perf/Microbenchmarks/HubConnectionStartBenchmark.cs
```diff
@@ -40,19 +40,13 @@ public void GlobalSetup()
             _pipe = new TestDuplexPipe();
 
             var hubConnectionBuilder = new HubConnectionBuilder();
-            var delegateConnectionFactory = new DelegateConnectionFactory(format =>
+            var delegateConnectionFactory = new DelegateConnectionFactory(endPoinnt =>
```
davidfowl: nit: endPoint

src/SignalR/clients/csharp/Client/src/HttpConnectionFactory.cs
halter73: This is the HttpConnectionFactory that you want to make internal, so does it matter? Only HttpEndPoints will work anyway.

src/SignalR/clients/csharp/Http.Connections.Client/src/HttpConnection.cs
```diff
@@ -166,30 +188,7 @@ internal HttpConnection(HttpConnectionOptions httpConnectionOptions, ILoggerFact
...
-        /// Starts the connection using the specified transfer format.
-        /// </summary>
-        /// <param name="transferFormat">The transfer format the connection should use.</param>
-        /// <param name="cancellationToken">The token to monitor for cancellation requests. The default value is <see cref="CancellationToken.None" />.</param>
-        /// <returns>A <see cref="Task"/> that represents the asynchronous start.</returns>
-        /// <remarks>
-        /// A connection cannot be restarted after it has stopped. To restart a connection
-        /// a new instance should be created using the same options.
-        /// </remarks>
-        public async Task StartAsync(TransferFormat transferFormat, CancellationToken cancellationToken = default)
```
davidfowl: Keep this constructor. It can override the one passed in the constructor. That would avoid breaking the Azure SignalR SDK:

https://github.com/Azure/azure-signalr/blob/7257fda52f219601f8efc34dd240728a9c4ea31c/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ConnectionFactory.cs#L32-L44

src/SignalR/samples/ClientSample/RawSample.cs
```diff
@@ -34,10 +34,12 @@ public static async Task<int> ExecuteAsync(string baseUrl)
             baseUrl = string.IsNullOrEmpty(baseUrl) ? "http://localhost:5000/chat" : baseUrl;
 
             Console.WriteLine($"Connecting to {baseUrl}...");
-            var connection = new HttpConnection(new Uri(baseUrl));
+
+            var connection = new HttpConnection(new HttpConnectionOptions() { Url = new Uri(baseUrl) }, TransferFormat.Text, loggerFactory: null);
```
davidfowl: Is this the simplest way to construct the HttpConnection now? Can we add some defaults (like the TransferFormat) so this sort of code doesn't break:

https://github.com/Azure/azure-signalr/blob/7257fda52f219601f8efc34dd240728a9c4ea31c/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ConnectionFactory.cs#L32-L44

src/SignalR/clients/csharp/Client.Core/src/HubConnectionBuilder.cs
halter73: I figured you might say that. Do you want to add a new HubConnectionOptions type with and EndPoint property?

I was thinking making a HubConnection.StartAsync take an EndPoint might be even better if we're willing to make that hard of a break

src/Servers/Connections.Abstractions/src/HttpEndPoint.cs
halter73: I followed the pattern you used for FileHandleEndPoint down to the namespace and lack of virtual properties.

src/SignalR/clients/csharp/Client/src/HttpConnectionFactory.cs
halter73: If the IConnectionFactory gets resolved directly from the ServiceProvider anyone can call ConnectAsync with any EndPoint.

I'd be OK with having it silently overrides the option though. I was thinking maybe we do that and log a warning.

src/SignalR/clients/csharp/Http.Connections.Client/src/HttpConnection.cs
halter73: Dang it. This isn't the constructor though. This change is easier if we can lock

src/SignalR/clients/csharp/Client/src/HttpConnectionFactory.cs
davidfowl: So you don't have to have a strange name like "castedHttpEndpoint".

src/SignalR/clients/csharp/Http.Connections.Client/src/HttpConnection.cs
halter73: We can probably keep that working. Do you wanted to make HttpConnectionFactory internal, but keep HttpConnection public?

halter73: Isn't the Azure SignalR SDK going to have to change their Connection factory to implement the new IConnectionFactory interface anyway? Might as well break to the cleaner API.

halter73: Isn't the Azure SignalR SDK going to have to change their Connection factory to implement the new IConnectionFactory interface anyway? Might as well break to the cleaner API.

davidfowl: They don't have to because they don't use the HubConnection.

src/SignalR/clients/csharp/Client.Core/src/HubConnectionBuilder.cs
davidfowl: That's a pretty awful API though. It would be more flexible though.

src/SignalR/clients/csharp/Http.Connections.Client/src/HttpConnection.cs
halter73: They must be implementing the old IConnectionFactory for a reason though. The only reason I could think of is so it could be used by HubConnection. Unless you plan on keeping two IConnectionFactory's around in 3.0 and having HubConnection support both, the Azure SignalR SDK is going to have to update their ConnectionFactory implementation, so I don't see the value in keeping this legacy StartAsync overload around for the Azure SignalR SDK specifically.

davidfowl: They’re not implementing the old factory. They actually have their own factory (it’s just a copy of the shape from SignalR). So none of this will affect them (other than maybe a potential type conflict). I know because I originally wrote that code

src/SignalR/clients/csharp/Client.Core/src/HubConnectionBuilder.cs
davidfowl: Actually what you have isn’t so bad it gets bad only when you add this to a bigger service collection.

src/Servers/Connections.Abstractions/src/HttpEndPoint.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        public HttpEndPoint(Uri url)
+        {
+            Url = url ?? throw new ArgumentNullException(nameof(url));
+        }
+
+        public Uri Url { get;  }
```
jkotalik:
```suggestion
        public Uri Url { get; }
```

src/Servers/Connections.Abstractions/src/IConnectionFactory.cs
```diff
@@ -0,0 +1,26 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Net;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.Connections
+{
+    /// <summary>
+    /// A factory abstraction for creating connections to an endpoint.
```
jkotalik: The wording is funky here. Add these to other places too.

A factory abstraction for creating connections to an endpoint.
```suggestion
    /// A factory abstraction for creating connections for an endpoint.
```

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -181,8 +185,8 @@ public partial class HubConnection
         /// <remarks>
         /// The <see cref="IServiceProvider"/> used to initialize the connection will be disposed when the connection is disposed.
         /// </remarks>
-        public HubConnection(IConnectionFactory connectionFactory, IHubProtocol protocol, IServiceProvider serviceProvider, ILoggerFactory loggerFactory, IRetryPolicy reconnectPolicy)
-            : this(connectionFactory, protocol, serviceProvider, loggerFactory)
+        public HubConnection(IConnectionFactory connectionFactory, IHubProtocol protocol, EndPoint endPoint, IServiceProvider serviceProvider, ILoggerFactory loggerFactory, IRetryPolicy reconnectPolicy)
```
jkotalik: I assume keeping the original constructors (to avoid the breaking change) is the entire point of this change.

src/SignalR/clients/csharp/Client/src/HttpConnectionFactory.cs
jkotalik: Why not move it to the right assembly now?

src/SignalR/clients/csharp/Client.Core/src/HubConnectionBuilder.cs
jkotalik:
```suggestion
            var endPoint = serviceProvider.GetService<EndPoint>() ??
                 throw new InvalidOperationException($"Cannot create {nameof(HubConnection)} instance. An {nameof(EndPoint)} was not configured.");
```

src/SignalR/clients/csharp/Client/src/HubConnectionBuilderHttpExtensions.cs
```diff
@@ -138,8 +141,20 @@ private static IHubConnectionBuilder WithUrlCore(this IHubConnectionBuilder hubC
...
 
+            // Add HttpConnectionOptionsDerivedHttpEndPoint so HubConnection can read the Url from HttpConnectionOptions
+            // without the Signal.Client.Core project taking a new dependency on Http.Connections.Client.
+            hubConnectionBuilder.Services.AddSingleton<EndPoint, HttpConnectionOptionsDerivedHttpEndPoint>();
+
             hubConnectionBuilder.Services.AddSingleton<IConnectionFactory, HttpConnectionFactory>();
             return hubConnectionBuilder;
         }
+
+        private class HttpConnectionOptionsDerivedHttpEndPoint : HttpEndPoint
```
jkotalik: This name is interesting... maybe `HttpEndPointWithConnectionOptions`?

src/SignalR/clients/csharp/Client/src/Microsoft.AspNetCore.SignalR.Client.csproj
```diff
@@ -11,4 +11,7 @@
     <Reference Include="Microsoft.AspNetCore.Http.Connections.Client" />
   </ItemGroup>
 
+  <ItemGroup>
+    <InternalsVisibleTo Include="Microsoft.AspNetCore.SignalR.Client.Tests" />
```
jkotalik: Wait this is a thing now? You don't need to put this in assembly attributes 😱

src/SignalR/clients/csharp/Client/test/FunctionalTests/HubProtocolVersionTests.cs
```diff
@@ -17,6 +18,7 @@
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Testing;
 using Microsoft.Extensions.Options;
+using Moq;
```
BrennanConroy: ?

src/SignalR/clients/csharp/Client/test/UnitTests/HttpConnectionFactoryTests.cs
```diff
@@ -17,16 +28,93 @@ public async Task ConnectionIsDisposedIfItFailsToStartAsync()
             var testHandler = new TestHttpMessageHandler(autoNegotiate: false, handleFirstPoll: false);
             testHandler.OnRequest((req, next, ct) => Task.FromException<HttpResponseMessage>(new Exception("BOOM")));
 
-            var factory = new HttpConnectionFactory(Options.Create(new HttpConnectionOptions() {
-                Url = new Uri("http://example.com"),
-                HttpMessageHandlerFactory = _ => testHandler
-            }), NullLoggerFactory.Instance);
+            var factory = new HttpConnectionFactory(
+                Mock.Of<IHubProtocol>(p => p.TransferFormat == TransferFormat.Text),
```
BrennanConroy: Please try to avoid Moq where possible. This should easily be replaceable by `new JsonHubProtocol()`

src/SignalR/clients/csharp/Client/src/HubConnectionBuilderHttpExtensions.cs
```diff
@@ -138,8 +141,20 @@ private static IHubConnectionBuilder WithUrlCore(this IHubConnectionBuilder hubC
...
+        {
+            public HttpConnectionOptionsDerivedHttpEndPoint(IOptions<HttpConnectionOptions> httpConnectionOptions)
```
BrennanConroy: Couldn't we put the transferformat on `HttpEndPoint` and resolve it here through the `IHubProtocol` and avoid the `IHubProtocol` stuff you do in `HttpConnectionFactory`?

src/SignalR/clients/csharp/Http.Connections.Client/src/HttpConnection.cs
```diff
@@ -212,7 +211,7 @@ private async Task StartAsyncCore(TransferFormat transferFormat, CancellationTok
 
                 Log.Starting(_logger);
 
-                await SelectAndStartTransport(transferFormat, cancellationToken);
+                await SelectAndStartTransport(_transferFormat, cancellationToken);
```
BrennanConroy: All public API needs `.ForceAsync()`

Also, you've removed the logger scope

src/Servers/Connections.Abstractions/src/IConnectionFactory.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// </summary>
+    public interface IConnectionFactory
+    {
+        /// <summary>
+        /// Creates a new connection to an endpoint
```
BrennanConroy: full stop

src/Servers/Connections.Abstractions/src/HttpEndPoint.cs
BrennanConroy: xml comments!

src/Servers/Connections.Abstractions/src/HttpEndPoint.cs
BrennanConroy: nit: extra space

src/SignalR/clients/csharp/Client/test/UnitTests/HttpConnectionFactoryTests.cs
halter73: I considered using `new JsonHubProtocol()` when I changed this, but decided to go with `Mock.Of<IHubProtocol>(p => p.TransferFormat == TransferFormat.Text)` because it's more self-documenting. It shows what the test is actually using the IHubProtocol for.

src/SignalR/clients/csharp/Client/src/HubConnectionBuilderHttpExtensions.cs
halter73: `HttpEndPoint` is not a solely SignalR concept. Other things that might want to connect to an `HttpEndPoint` probably don't care about SignalR's TransferFormat. The project bedrock API all deal with bytes/binary anyway.

We could create a `SignalREndPoint` that derives from `HttpEndPoint` to contain extra options, but at that point what options go into `HttpConnectionOptions` and what options go into `SignalREndPoint`?

src/SignalR/clients/csharp/Http.Connections.Client/src/HttpConnection.cs
halter73: I'm surprised no tests failed after removing the logger scope. It'd be nice if we tested the `ForceAsync()` stuff too.

src/SignalR/clients/csharp/Client/src/HubConnectionBuilderHttpExtensions.cs
halter73: Meh, it's a private class. That's why I went with the overly-descriptive name.

src/SignalR/clients/csharp/Client/src/HttpConnectionFactory.cs
halter73: I'm assuming you want this in the Http.Connections.Client assembly, right? The layering made no sense before, so I didn't want to move things around until I was sure I understood your vision. Especially since the move would technically be breaking.

src/SignalR/clients/csharp/Client/src/HubConnectionBuilderHttpExtensions.cs
davidfowl: > That's why I went with the overly-descriptive name.

False, you do this in general 😄

src/SignalR/clients/csharp/Client/src/HubConnectionBuilderHttpExtensions.cs
davidfowl: `TransferFormat` is in Connection.Abstractions as well as it would be fine to add it to the HttpEndPoint

src/SignalR/clients/csharp/Client/src/Microsoft.AspNetCore.SignalR.Client.csproj
anurse: It IS a thing! You still need to have the full public key for things that aren't built by the repo, but if it's built here you can just use the simple assembly name and the rest is automated!

src/Servers/Connections.Abstractions/src/IConnectionFactory.cs
```diff
@@ -0,0 +1,26 @@
...
+        /// </summary>
+        /// <param name="endPoint">The <see cref="EndPoint"/> to connect to.</param>
+        /// <param name="cancellationToken">The token to monitor for cancellation requests. The default value is <see cref="CancellationToken.None" />.</param>
+        /// <returns>
+        /// A <see cref="ValueTask{TResult}"/> that represents the asynchronous connect.
+        /// The <see cref="Task{TResult}.Result"/> property returns a <see cref="ConnectionContext"/> for the new connection.
```
anurse: yuck, let's not encourage use of `Task.Result`. Even if other docs do this, it's bad :). I would reword the whole thing in to:

```csharp
/// A <see cref="ValueTask{TResult}" /> that represents the asynchronous connect, yielding the <see cref="ConnectionContext" /> for the new connection when completed.
```

Sorry, can't make it a suggestion since it's a multi-line change.

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
anurse: Users _shouldn't_ be constructing `HubConnection`. If it weren't for DI this would be `internal`. I'm kinda 🤷‍♂ on breaking changes in this kind of thing.

halter73: Funnily enough, HubConnection has always been difficult/impossible to construct manually. Now that the HubConnection needs to specify an EndPoint that previously was just an implementation detail of IConnectionFactory, HubConnection has to break at least a little.

The main thing I was worried about was keeping the HubConnectionBuilder backwards-compatible.

src/SignalR/clients/csharp/Client.Core/src/HubConnectionBuilder.cs
anurse: I could see value in an `IOptions<HubConnectionOptions>` given the earlier discussions re constructor breaking changes... might be a safer approach.

src/SignalR/clients/csharp/Client/src/HubConnectionBuilderHttpExtensions.cs
halter73: But what does a Text TransferFormat mean to anyone other than SignalR? The ConnectionContext exposes Pipes which are all binary. The TransferFormat seems like a protocol concern, not a transport/endpoint concern.

davidfowl: When using the lower HttpConnection layer, it implies that the transport cannot do binary, e.g. server sent events.

src/SignalR/clients/csharp/Client/test/UnitTests/HttpConnectionFactoryTests.cs
anurse: I'm mildly torn betweeen my preference to avoid Moq as much as possible and the fact that I do agree with @halter73 's self-documenting argument. I guess I'd consider just making a `TestTextHubProtocol` concrete class (that throws `NotImplementedException` except in the `TransferFormat` property) but I could go either way.

src/Servers/Connections.Abstractions/src/IConnectionFactory.cs
halter73: Clearly I wasn't careful enough when copying the old IConnectionFactory doc comments. I should have at least put *Value*Task{TResult}.Result in there! jk

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -192,27 +196,22 @@ public HubConnection(IConnectionFactory connectionFactory, IHubProtocol protocol
...
-        }
-
-        /// <summary>
-        /// Initializes a new instance of the <see cref="HubConnection"/> class.
-        /// </summary>
-        /// <param name="connectionFactory">The <see cref="IConnectionFactory" /> used to create a connection each time <see cref="StartAsync" /> is called.</param>
-        /// <param name="protocol">The <see cref="IHubProtocol" /> used by the connection.</param>
-        /// <param name="loggerFactory">The logger factory.</param>
-        public HubConnection(IConnectionFactory connectionFactory, IHubProtocol protocol, ILoggerFactory loggerFactory)
+        public HubConnection(IConnectionFactory connectionFactory,
```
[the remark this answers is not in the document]
halter73: There's a first time for everything. Seriously though, I prefer this for DI-invoked ctors with a lot of arguments. In fact, it looks like we already do this in our HubConnectionHandler.

src/SignalR/clients/csharp/Client/src/HttpConnectionFactory.cs
```diff
@@ -16,16 +18,23 @@ namespace Microsoft.AspNetCore.SignalR.Client
...
         /// <summary>
         /// Initializes a new instance of the <see cref="HttpConnectionFactory"/> class.
         /// </summary>
+        /// <param name="hubProtocol">The <see cref="IHubProtocol"/> that will use the connections.</param>
         /// <param name="options">The connection options.</param>
         /// <param name="loggerFactory">The logger factory.</param>
-        public HttpConnectionFactory(IOptions<HttpConnectionOptions> options, ILoggerFactory loggerFactory)
+        public HttpConnectionFactory(IHubProtocol hubProtocol, IOptions<HttpConnectionOptions> options, ILoggerFactory loggerFactory)
         {
+            if (hubProtocol == null)
```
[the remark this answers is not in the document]
halter73: I would, but throw expressions don't look as nice once you start trying to pull out a specific property like IHubProtocol.TransferFormat and store that in the field instead of the entire argument. You end up needing a ternary expression or something. I assume this is the same reason we didn't use a throw expression for options and loggerFactory.

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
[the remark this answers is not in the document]
mikaelm12: I also prefer this for more than 3 args. I angling more for consistency.

src/SignalR/clients/csharp/Client/test/UnitTests/HttpConnectionFactoryTests.cs
halter73: I find the mock more self-documenting than the TestTextHubProtocol for something as trivial as this, because I don't have to scroll to TestTextHubProtocol to see what it's doing.

And when TestTextHubProtocol inevitably starts getting used by other tests that started life as a copy of this one, you won't be able to see the implementation for TestTextHubProtocol in the PR. And when those tests inevitably adds more requirement, you get more methods/properties that don't throw a NotImplementedException, so you don't know what's really being called by any given test even if you can find the TestTextHubProtocol implementation.

All of this for what should be a one-liner.

davidfowl: I'm over Moq

src/SignalR/clients/csharp/Client/src/HubConnectionBuilderHttpExtensions.cs
halter73: If we're going to put a TransferFormat on HttpEndPoint, shouldn't we add an Encoding as well? In SignalR's case, TransferFormat.Text implies that all bytes sent over the pipe is valid UTF-8, but what if a non-SignalR EndPoint only accepts ASCII or something else? Do you set TransferFormat.Text on the Endpoint, but use an options object for the encoding? The TransferFormat and encoding should be configured together IMHO.

Also, I looked at the long polling transport and it just completely ignores the TransferFormat. SSE validates the TransferFormat at the beginning of StartAsync, but f assumes TransferFormat.Text from there on out. If the code calling the ITransport simply wrote valid UTF-8 to the IDuplexPipe, everything would work fine with SSE without a need for ever passing in a TransferFormat.

So it's really only the WebSocket transport that needs to know up front whether you're sending binary or text data, and if it weren't for the F12 browser tools being crappy with binary WebSockets, we'd probably just force WebSockets to binary all the time, even for text-based protocols.

Since this is the .NET-client, maybe forcing WebSocket to use the binary message mode even for text protocols is more reasonable. I'd also be OK with adding a TransferFormat to a `WebSocketEndPoint` which could b derived from `HttpEndPoint`.

davidfowl: > If we're going to put a TransferFormat on HttpEndPoint, shouldn't we add an Encoding as well? In SignalR's case, TransferFormat.Text implies that all bytes sent over the pipe is valid UTF-8, but what if a non-SignalR EndPoint only accepts ASCII or something else? Do you set TransferFormat.Text on the Endpoint, but use an options object for the encoding? The TransferFormat and encoding should be configured together IMHO.

Sounds good add a TextEncoding property that should be used if TransferFormat is Text?

halter73: That raises the question of whether the Encoding is really a property of the HttpEndPoint. If Encoding *is* part and parcel of the HttpEndPoint, what configuration *isn't*. Why don't we consider the CloseTimeout part of the HttpEndPoint? Or IWebProxy

src/SignalR/clients/csharp/Client/test/UnitTests/HttpConnectionFactoryTests.cs
halter73: Are you suggesting I rewrite this? If so, can I at least get a reason? Even if it's something completely irrelevant like not working in environments that don't support IL emit.

src/SignalR/clients/csharp/Client/src/HubConnectionBuilderHttpExtensions.cs
davidfowl: This is similar to socket options for IPEndPoint, we had this discussion remember? Either we need to abandon Endpoint and have something that is more loose that allows passing binding specific information (like all of the useful properties you describe) or we say those come in via a side channel.

src/SignalR/clients/csharp/Client/test/UnitTests/HttpConnectionFactoryTests.cs
davidfowl: No I don't really care, we've been gradually moving away from Moq as a team. Leave it the way it is if you want but generally I've been trying to avoid Moq as well when the situation is trivial. As we move towards support more AOT scenarios Moq won't work well (or it'll work but be really slow). Though that's a future thing, nothing to worry about yet 😄

src/SignalR/clients/csharp/Client/src/HubConnectionBuilderHttpExtensions.cs
halter73: You can specify any EndPoint you want if you supply an IConnectionFactory that supports it. 
Just look at the sample project. It uses an IPEndPoint with HubConnection.

I'm fine with putting a bunch of SignalR specific stuff on an EndPoint type (though people will still likely get confused which options go where unless we put either all or none of the options non the EndPoint type). Just call it SignalREndPoint or something and move it out of the Connection.Abstractions assembly.

src/SignalR/clients/csharp/Client/src/HttpConnectionFactory.cs
halter73: I didn't move this to Http.Connections.Client yet, because it needs to be accessible from HubConnectionBuilderHttpExtensions.

halter73:
⬆️📅

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
jkotalik: Stupid question, can we make HubConnection internal?

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
```diff
@@ -452,7 +451,7 @@ private async Task StartAsyncCore(CancellationToken cancellationToken)
 
         private Task CloseAsync(ConnectionContext connection)
         {
-            return _connectionFactory.DisposeAsync(connection);
+            return connection.DisposeAsync().AsTask();
```
jkotalik: Can this return ValueTask?

src/SignalR/clients/csharp/Client/src/HttpConnectionFactory.cs
jkotalik: For my own knowledge, where do these options get modified? Feels like a bad pattern to be modifying options in general. Should just store internal state rather than needing to shallow copy.

src/SignalR/clients/csharp/Client.Core/src/HubConnection.cs
BrennanConroy: We can't make the class internal as it's the class users interact with to send and receive messages. We would like the constructors to be internal but like [Andrew said](https://github.com/aspnet/AspNetCore/pull/11484/files#r296904380) we can't because of DI.

BrennanConroy (approved):
Looking good

davidfowl (approved):

davidfowl:
👍

PR #11487, opened 2019-06-24
Added some doc comments to bedrock APIs

davidfowl:
- Also cleaned up some HttpContext doc comments

### DIFF
--- a/src/Http/Http.Abstractions/src/HttpRequest.cs
+++ b/src/Http/Http.Abstractions/src/HttpRequest.cs
@@ -101,12 +101,13 @@ namespace Microsoft.AspNetCore.Http
         /// <summary>
         /// Gets or sets the request body <see cref="Stream"/>.
         /// </summary>
-        /// <returns>The request body <see cref="Stream"/>.</returns>
+        /// <value>The request body <see cref="Stream"/>.</value>
         public abstract Stream Body { get; set; }
 
         /// <summary>
         /// Gets the request body <see cref="PipeReader"/>.
         /// </summary>
+        /// <value>The request body <see cref="PipeReader"/>.</value>
         public virtual PipeReader BodyReader { get => throw new NotImplementedException();  }
 
         /// <summary>
--- a/src/Http/Http.Abstractions/src/HttpRequest.cs
+++ b/src/Http/Http.Abstractions/src/HttpRequest.cs
@@ -99,13 +99,13 @@ namespace Microsoft.AspNetCore.Http
         public abstract string ContentType { get; set; }
 
         /// <summary>
-        /// Gets or sets the RequestBody Stream.
+        /// Gets or sets the request body <see cref="Stream"/>.
         /// </summary>
-        /// <returns>The RequestBody Stream.</returns>
+        /// <returns>The request body <see cref="Stream"/>.</returns>
         public abstract Stream Body { get; set; }
 
         /// <summary>
-        /// Gets the request body pipe <see cref="PipeReader"/>.
+        /// Gets the request body <see cref="PipeReader"/>.
         /// </summary>
         public virtual PipeReader BodyReader { get => throw new NotImplementedException();  }
 
--- a/src/Http/Http.Abstractions/src/HttpResponse.cs
+++ b/src/Http/Http.Abstractions/src/HttpResponse.cs
@@ -47,6 +47,7 @@ namespace Microsoft.AspNetCore.Http
         /// <summary>
         /// Gets the response body <see cref="PipeWriter"/>
         /// </summary>
+        /// <value>The response body <see cref="PipeWriter"/>.</value>
         public virtual PipeWriter BodyWriter { get => throw new NotImplementedException(); }
 
         /// <summary>
--- a/src/Http/Http.Abstractions/src/HttpResponse.cs
+++ b/src/Http/Http.Abstractions/src/HttpResponse.cs
@@ -45,7 +45,7 @@ namespace Microsoft.AspNetCore.Http
         public abstract Stream Body { get; set; }
 
         /// <summary>
-        /// Gets the response body pipe <see cref="PipeWriter"/>
+        /// Gets the response body <see cref="PipeWriter"/>
         /// </summary>
         public virtual PipeWriter BodyWriter { get => throw new NotImplementedException(); }
 
--- a/src/Servers/Connections.Abstractions/src/ConnectionContext.cs
+++ b/src/Servers/Connections.Abstractions/src/ConnectionContext.cs
@@ -12,22 +12,50 @@ using Microsoft.AspNetCore.Http.Features;
 
 namespace Microsoft.AspNetCore.Connections
 {
+    /// <summary>
+    /// Encapsulates all information about an individual connection irrespective of the underlying transport.
+    /// </summary>
     public abstract class ConnectionContext : IAsyncDisposable
     {
+        /// <summary>
+        /// Gets or sets a unique identifier to represent this connection in trace logs.
+        /// </summary>
         public abstract string ConnectionId { get; set; }
 
+        /// <summary>
+        /// Gets the collection of features provided by the server and middleware available on this connection.
+        /// </summary>
         public abstract IFeatureCollection Features { get; }
 
+        /// <summary>
+        /// Gets or sets a key/value collection that can be used to share data within the scope of this connection.
+        /// </summary>
         public abstract IDictionary<object, object> Items { get; set; }
 
+        /// <summary>
+        /// Gets or sets the <see cref="IDuplexPipe"/> that can be used to read and write to the this connection.
+        /// </summary>
         public abstract IDuplexPipe Transport { get; set; }
 
+        /// <summary>
+        /// Notifies when the client connection has closed.
+        /// </summary>
         public virtual CancellationToken ConnectionClosed { get; set; }
 
+        /// <summary>
+        /// Gets or sets the local endpoint for this connection.
+        /// </summary>
         public virtual EndPoint LocalEndPoint { get; set; }
 
+        /// <summary>
+        /// Gets or sets the remote endpoint for this connection.
+        /// </summary>
         public virtual EndPoint RemoteEndPoint { get; set; }
 
+        /// <summary>
+        /// Aborts the underlying connection.
+        /// </summary>
+        /// <param name="abortReason">The reason for aborting the connection.</param>
         public virtual void Abort(ConnectionAbortedException abortReason)
         {
             // We expect this to be overridden, but this helps maintain back compat
@@ -36,8 +64,15 @@ namespace Microsoft.AspNetCore.Connections
             Features.Get<IConnectionLifetimeFeature>()?.Abort();
         }
 
+        /// <summary>
+        /// Aborts the underlying connection.
+        /// </summary>
         public virtual void Abort() => Abort(new ConnectionAbortedException("The connection was aborted by the application via ConnectionContext.Abort()."));
 
+        /// <summary>
+        /// Releases resources for the underlying connection.
+        /// </summary>
+        /// <returns>A <see cref="ValueTask"/> that completes when resources have been released.</returns>
         public virtual ValueTask DisposeAsync()
         {
             return default;
--- a/src/Servers/Connections.Abstractions/src/ConnectionContext.cs
+++ b/src/Servers/Connections.Abstractions/src/ConnectionContext.cs
@@ -13,7 +13,7 @@ using Microsoft.AspNetCore.Http.Features;
 namespace Microsoft.AspNetCore.Connections
 {
     /// <summary>
-    /// Encapsulates all information about an individual connection irrespective of the underlying transport.
+    /// Encapsulates all information about an individual connection.
     /// </summary>
     public abstract class ConnectionContext : IAsyncDisposable
     {
@@ -33,12 +33,12 @@ namespace Microsoft.AspNetCore.Connections
         public abstract IDictionary<object, object> Items { get; set; }
 
         /// <summary>
-        /// Gets or sets the <see cref="IDuplexPipe"/> that can be used to read and write to the this connection.
+        /// Gets or sets the <see cref="IDuplexPipe"/> that can be used to read or write data on this connection.
         /// </summary>
         public abstract IDuplexPipe Transport { get; set; }
 
         /// <summary>
-        /// Notifies when the client connection has closed.
+        /// Triggered when the client connection is closed.
         /// </summary>
         public virtual CancellationToken ConnectionClosed { get; set; }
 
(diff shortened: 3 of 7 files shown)


src/Http/Http.Abstractions/src/HttpRequest.cs
```diff
@@ -99,13 +99,13 @@ public abstract class HttpRequest
         public abstract string ContentType { get; set; }
 
         /// <summary>
-        /// Gets or sets the RequestBody Stream.
+        /// Gets or sets the request body <see cref="Stream"/>.
         /// </summary>
-        /// <returns>The RequestBody Stream.</returns>
+        /// <returns>The request body <see cref="Stream"/>.</returns>
```
anurse: I think properties should be using `<value>` rather than `<returns>`. For example: https://docs.microsoft.com/en-us/dotnet/api/system.string.length?view=netframework-4.8#property-value

src/Http/Http.Abstractions/src/HttpRequest.cs
```diff
@@ -99,13 +99,13 @@ public abstract class HttpRequest
...
         public abstract Stream Body { get; set; }
 
         /// <summary>
-        /// Gets the request body pipe <see cref="PipeReader"/>.
+        /// Gets the request body <see cref="PipeReader"/>.
         /// </summary>
```
anurse:
```suggestion
        /// </summary>
        /// <value>The request body <see cref="PipeReader"/>.</value>
```

src/Http/Http.Abstractions/src/HttpResponse.cs
```diff
@@ -45,7 +45,7 @@ public abstract class HttpResponse
         public abstract Stream Body { get; set; }
 
         /// <summary>
-        /// Gets the response body pipe <see cref="PipeWriter"/>
+        /// Gets the response body <see cref="PipeWriter"/>
         /// </summary>
```
anurse:
```suggestion
        /// </summary>
        /// <value>The response body <see cref="PipeWriter"/>.</value>
```

src/Servers/Connections.Abstractions/src/ConnectionContext.cs
```diff
@@ -12,22 +12,50 @@
 
 namespace Microsoft.AspNetCore.Connections
 {
+    /// <summary>
+    /// Encapsulates all information about an individual connection irrespective of the underlying transport.
```
anurse:
```suggestion
    /// Encapsulates all information about an individual connection.
```

I don't think raising the transport is useful here and "irrespective" is one of those "trying to sound smart" words to me ;). Also, I believe there could be transport-specific data in this class, it's just that the class itself isn't transport-specific. I think it's clearer without that clause.

src/Servers/Connections.Abstractions/src/ConnectionContext.cs
```diff
@@ -12,22 +12,50 @@
...
+    /// </summary>
     public abstract class ConnectionContext : IAsyncDisposable
     {
+        /// <summary>
+        /// Gets or sets a unique identifier to represent this connection in trace logs.
+        /// </summary>
```
anurse: `<value>` elements for these properties?

src/Http/Http.Abstractions/src/HttpRequest.cs
anurse:
```suggestion
        /// <value>The request body <see cref="Stream"/>.</value>
```

src/Servers/Connections.Abstractions/src/ConnectionContext.cs
```diff
@@ -12,22 +12,50 @@
...
         public abstract IFeatureCollection Features { get; }
 
+        /// <summary>
+        /// Gets or sets a key/value collection that can be used to share data within the scope of this connection.
+        /// </summary>
         public abstract IDictionary<object, object> Items { get; set; }
 
+        /// <summary>
+        /// Gets or sets the <see cref="IDuplexPipe"/> that can be used to read and write to the this connection.
```
anurse:
```suggestion
        /// Gets or sets the <see cref="IDuplexPipe"/> that can be used to read or write data on this connection.
```

src/Servers/Connections.Abstractions/src/ConnectionContext.cs
```diff
@@ -12,22 +12,50 @@
...
+        /// </summary>
         public abstract IDuplexPipe Transport { get; set; }
 
+        /// <summary>
+        /// Notifies when the client connection has closed.
```
anurse:
```suggestion
        /// Triggered when the client connection is closed.
```

(Source for precedent: https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.hosting.iapplicationlifetime?view=aspnetcore-2.2)

src/Servers/Connections.Abstractions/src/ConnectionContext.cs
```diff
@@ -12,22 +12,50 @@
...
 
+        /// <summary>
+        /// Gets or sets the remote endpoint for this connection.
+        /// </summary>
         public virtual EndPoint RemoteEndPoint { get; set; }
 
+        /// <summary>
+        /// Aborts the underlying connection.
+        /// </summary>
+        /// <param name="abortReason">The reason for aborting the connection.</param>
```
anurse:
```suggestion
        /// <param name="abortReason">An optional <see cref="ConnectionAbortedException"/> describing the reason the connection is being terminated.</param>
```

I say optional because it seems to be completely ignored in the base class version...

src/Servers/Connections.Abstractions/src/ConnectionDelegate.cs
```diff
@@ -5,5 +5,10 @@
 
 namespace Microsoft.AspNetCore.Connections
 {
+    /// <summary>
+    /// A function that can process an connection.
```
anurse:
```suggestion
    /// A function that can process a connection.
```

Somebody copy-pasta'd from https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.requestdelegate?view=aspnetcore-2.2 ;)

src/Servers/Connections.Abstractions/src/ConnectionDelegate.cs
```diff
@@ -5,5 +5,10 @@
...
+    /// </summary>
+    /// <param name="connection">The new <see cref="ConnectionContext"/></param>
```
anurse:
```suggestion
    /// <param name="connection">A <see cref="ConnectionContext" /> representing the connection.</param>
```

src/Servers/Connections.Abstractions/src/ConnectionDelegate.cs
```diff
@@ -5,5 +5,10 @@
...
+    /// <returns>A <see cref="Task"/> that represents the connection lifetime. When the task completes, the connection is complete.</returns>
```
anurse:
```suggestion
    /// <returns>A <see cref="Task"/> that represents the connection lifetime. When the task completes, the connection will be closed.</returns>
```

anurse: Adds more clarity to what it means to "complete" the connection.

src/Servers/Connections.Abstractions/src/IConnectionBuilder.cs
```diff
@@ -5,12 +5,27 @@
 
 namespace Microsoft.AspNetCore.Connections
 {
+    /// <summary>
+    /// Defines an interface that provides the mechanisms to configure a connection pipeline.
```
anurse: This is some Andrew-approved copy-pasta 👍

src/Servers/Connections.Abstractions/src/IConnectionBuilder.cs
```diff
@@ -5,12 +5,27 @@
...
+    /// </summary>
     public interface IConnectionBuilder
     {
+        /// <summary>
+        /// Gets or sets the <see cref="IServiceProvider"/> that provides access to the application's service container.
```
jkotalik:
```suggestion
        /// Gets the <see cref="IServiceProvider"/> that provides access to the application's service container.
```

src/Servers/Connections.Abstractions/src/IConnectionListener.cs
```diff
@@ -8,12 +8,28 @@
 
 namespace Microsoft.AspNetCore.Connections
 {
+    /// <summary>
+    /// Defines an interface that represents a listener bound to a specific <see cref="EndPoint"/>.
+    /// </summary>
     public interface IConnectionListener : IAsyncDisposable
     {
+        /// <summary>
+        /// The end point that was bound. This may differ from the requested end point.
```
anurse: Inconsistent use of `end point` vs `endpoint`. Looks like [endpoint is a word](https://www.lexico.com/en/definition/endpoint) and that's what we've used in other places, so let's just be consistent.

```suggestion
        /// The endpoint that was bound. This may differ from the requested endpoint, such as when the caller requested that any free port be selected.
```

Good to include an example of why it would be different too.

jkotalik (approved):

src/Servers/Connections.Abstractions/src/IConnectionListener.cs
anurse: If you'd rather not provide the example (I'm not super worried either way):

```suggestion
        /// The endpoint that was bound. This may differ from the requested endpoint.
```

src/Servers/Connections.Abstractions/src/IConnectionListener.cs
```diff
@@ -8,12 +8,28 @@
...
+        /// </summary>
         EndPoint EndPoint { get; }
 
+        /// <summary>
+        /// Begins an asynchronous operation to accept an incoming connection.
+        /// </summary>
+        /// <param name="cancellationToken">The token to monitor for cancellation requests.</param>
+        /// <returns>A <see cref="ValueTask{ConnectionContext}"/> that represents the accepted connection.</returns>
```
anurse:
```suggestion
        /// <returns>A <see cref="ValueTask{ConnectionContext}"/> that completes when a connection is accepted, yielding the <see cref="ConnectionContext" /> representing the connection.</returns>
```

src/Servers/Connections.Abstractions/src/IConnectionListenerFactory.cs
```diff
@@ -10,8 +10,17 @@
...
 {
+    /// <summary>
+    /// Defines an interface that provides the mechanisms for binding to various types of <see cref="EndPoint"/>s.
+    /// </summary>
     public interface IConnectionListenerFactory
     {
+        /// <summary>
+        /// Creates an <see cref="IConnectionListener"/> bound to the specified <see cref="EndPoint"/>.
+        /// </summary>
+        /// <param name="endpoint">The endpoint to bind to.</param>
```
anurse:
```suggestion
        /// <param name="endpoint">The <see cref="EndPoint" /> to bind to.</param>
```

Since we're talking about a specific type here...

src/Servers/Connections.Abstractions/src/IConnectionListenerFactory.cs
```diff
@@ -10,8 +10,17 @@
...
+        /// <param name="cancellationToken">The token to monitor for cancellation requests.</param>
+        /// <returns>A <see cref="ValueTask{IConnectionListener}"/> that represents the bound listener.</returns>
```
anurse:
```suggestion
        /// <returns>A <see cref="ValueTask{IConnectionListener}"/> that completes when the listener has been bound, yielding a <see cref="IConnectionListener" /> representing the new listener.</returns>
```

analogrelay (commented):
P R O S E

👏 	👏 	👏 	for docs!

src/Http/Http.Abstractions/src/HttpRequest.cs
davidfowl: Yea I was surprised they all did `<returns>` as it doesn't even show up in intellisense.

davidfowl:
Why can't I add these suggestions to a batch?!

src/Servers/Connections.Abstractions/src/ConnectionContext.cs
davidfowl: We don't do it anywhere. We need to do a huge find replace and update all the properties. Also is it valuable to describe the gets part of gets and sets?

davidfowl:
Got it

src/Servers/Connections.Abstractions/src/ConnectionContext.cs
anurse: I don't think waiting for "huge find replace" operations that never happen is worth it. Even so, it doesn't preclude the value of just fixing them when we find them.

davidfowl: I don't see the "value" in adding them, can you point me some prior art?

anurse: `<value>Using this element instead of 'returns' makes it actually show up in API docs both in IntelliSense and docs.microsoft.com</value>` :trollface: 

But seriously, compare one of our properties:

[image](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.options.optionsmonitor-1.currentvalue?view=aspnetcore-3.0)

With one of the BCL ones:

[image](https://docs.microsoft.com/en-us/dotnet/api/system.net.websockets.clientwebsocketoptions.clientcertificates?view=netcore-3.0#System_Net_WebSockets_ClientWebSocketOptions_ClientCertificates)

If we think that it doesn't really matter to fill in the "Property Value" heading (because the summary is sufficiently explaining the property) then I'm **fine** with that, but the `<returns>` element does nothing.

So, I guess I'm kinda responding more to the other one (where you did put a `<returns>`). If you don't want to do it, let's just not do it, rather than half doing it (by using `<returns>`, which doesn't work, instead of `<value>`, which does).

davidfowl: > So, I guess I'm kinda responding more to the other one (where you did put a <returns>). If you don't want to do it, let's just not do it, rather than half doing it (by using <returns>, which doesn't work, instead of <value>, which does).

Sounds good to me.

src/Servers/Connections.Abstractions/src/IConnectionBuilder.cs
halter73: Seriously though, should we start all our interface summary with "Defines an interface"? Seems obvious/redundant.

davidfowl: I vote @halter73 should fix up our definitions across the board.

halter73: Do I get complete creative control? :trollface:

anurse: You get complete creative control to react to all my wordsmithing ;)

PR #11488, opened 2019-06-24
Simplified HttpConnection

davidfowl:
- Don't register callbacks until we have a request processor
- Remove the IConnectionTimeoutFeature from the feature collection, it was used by connection adapters but isn't needed anymore.
- Rename OnInputOrOutputCompleted to OnConnectionClosed

### DIFF
--- a/src/Servers/Kestrel/Core/src/Internal/HttpConnection.cs
+++ b/src/Servers/Kestrel/Core/src/Internal/HttpConnection.cs
@@ -3,8 +3,6 @@
 
 using System;
 using System.Diagnostics;
-using System.IO.Pipelines;
-using System.Net;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Connections;
 using Microsoft.AspNetCore.Connections.Features;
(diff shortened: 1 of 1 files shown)


src/Servers/Kestrel/Core/src/Internal/HttpConnection.cs
```diff
@@ -39,80 +37,77 @@ public HttpConnection(HttpConnectionContext context)
...
+
+                    // These features should never be null in Kestrel itself, if this middleware is ever refactored to run outside of kestrel,
+                    // we'll need to handle these missing.
+                    Debug.Assert(connectionHeartbeatFeature != null, nameof(IConnectionHeartbeatFeature) + " is missing!");
+                    Debug.Assert(connectionLifetimeNotificationFeature != null, nameof(IConnectionLifetimeNotificationFeature) + " is missing!");
+
+                    // Register the various callbacks once we're going to start processing requests
+
+                    // The heart beat for various timeouts
+                    connectionHeartbeatFeature?.OnHeartbeat(state => ((HttpConnection)state).Tick(), this);
```
jkotalik: Why can't we remove these debug asserts and nullables? Arguably it is better to throw an NRE than let the server continue to run if these are null.

jkotalik (approved):

src/Servers/Kestrel/Core/src/Internal/HttpConnection.cs
davidfowl: Throw an NRE? It just means you don't get timeouts. That would be another server implementation so it's not a huge deal ATM (we're not there yet).

davidfowl: Also, it's super unfortunate we didn't return `IDisposable` from this...

halter73:
19 seconds too late lol

halter73:
> Remove the IConnectionTimeoutFeature from the feature collection, it was used by connection adapters but isn't needed anymore.

So you'd rather have initialize a new timer per TLS connection in 3.0?

davidfowl:
> So you'd rather have initialize a new timer per TLS connection in 3.0?

Yes, rather than refactor the logic to use this timeout logic, absolutely.  FWIW, we got a speed bump with the new https middleware 😄

halter73:
> FWIW, we got a speed bump with the new https middleware

This would only affect the `Connection: close` HTTPS scenario though.

davidfowl:
> The extra timers would only affect the Connection: close HTTPS scenario though.

That scenario did regress but unrelated to the https change. I also doubt it would show up in a measurable way as there were optimizations done to short lived timers (for scenarios like this one).

Issue #11495, opened 2019-06-24
Derived type arguments are dispatched to hub methods as null


NickDarvey:
...but only when you include a synthetic parameter in your hub method's signature.

### Describe the bug
I'm invoking a hub method and sending a tree where each node might be a subclass of some root node class. The deserializer correctly deserializes each node into their subclass so the runtime type of the argument is not the exact type as in the method signature, but rather a derived one.

This works correctly _unless_ a synthetic parameter is included in the method, i.e.  a CancellationToken token parameter. If you include a CancellationToken, the derived arguments will be dispatched as nulls to the hub method.

### To Reproduce
1. Create a class relationship like:
   ```csharp
   public class ANode {}
   public class BNode : ANode {}
   ```
2. And a hub method like:
   ```csharp
   public async IAsyncEnumerable<string> DoThing(ANode param, [EnumeratorCancellation]CancellationToken token) { /* ... */ }
   ```
3. Invoke the hub method with a `BNode` (e.g. by using Json.NET's `TypeNameHandling = TypeNameHandling.All`)

### Expected behavior
Hub method receives a `BNode` as `param`.

### Actual behavior
Hub method receives `null` as `param`.

### Cause
When there are [synthetic arguments](https://github.com/aspnet/AspNetCore/blob/5b31a9540aa64dcf613ce423c25eb630b84b0fac/src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs#L254), the original arguments from the `HubInvocationMesssage` are looped through and matched against the `HubMethodDescriptor`'s parameters to rebuild the arguments with the added synthetic parameters. [It uses equality comparison to see if the argument fits the parameter](https://github.com/aspnet/AspNetCore/blob/5b31a9540aa64dcf613ce423c25eb630b84b0fac/src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs#L264), which won't match on derived types. In scenario above [the code path labelled 'This should never happen'](https://github.com/aspnet/AspNetCore/blob/5b31a9540aa64dcf613ce423c25eb630b84b0fac/src/SignalR/server/Core/src/Internal/DefaultHubDispatcher.cs#L288) does happen and the current index of the arguments is left null.

### Additional context
ASP.NET Core 3.0.0-preview6.19307.2

anurse:
Ooh, nice catch, and thanks for the corresponding PR.

[...]

PR #11496, opened 2019-06-24
Fixes dispatching of derived type arguments to hub methods

NickDarvey:
- Uses `IsAssignableFrom` instead of equality comparison for matching invocation arguments against method parameters
- Adds test to validate behaviour

Addresses #11495


src/SignalR/server/SignalR/test/Internal/DefaultHubDispatcherTests.cs
```diff
@@ -0,0 +1,143 @@
...
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+using Microsoft.Extensions.Options;
+using Xunit;
+
+namespace Microsoft.AspNetCore.SignalR.Tests.Internal
+{
+    public class DefaultHubDispatcherTests
+    {
+        private class MockHubConnectionContext<TValue> : HubConnectionContext
```
NickDarvey: The simplest way I could think to validate the values that get dispatched to the hub was via the HubConnectionContext, capturing the responses it receives. (The behaviour is buried a little deep inside the dispatcher.)

src/SignalR/server/SignalR/test/Internal/DefaultHubDispatcherTests.cs
```diff
@@ -0,0 +1,143 @@
...
+                    }
+                }
+
+                else throw new NotImplementedException();
+
+                return default;
+            }
+        }
+
+        private static DefaultHubDispatcher<THub> CreateDispatcher<THub>() where THub : Hub
```
NickDarvey: I based the instantiation of our target (the dispatcher) on the way DefaultHubDispatcherBenchmark does it. I believe this is the minimum we require to put it under test.

anurse: We have tests that do method dispatching in [HubConnectionHandlerTests.cs](https://github.com/aspnet/AspNetCore/blob/master/src/SignalR/server/SignalR/test/HubConnectionHandlerTests.cs) using a fake connection. I'd think these kinds of tests can go in that.

NickDarvey: I've moved this test into HubConnectionHandlerTests and copied the patterns used in the other tests of dispatching for streaming hubs.

I'm referring to this test case as supporting 'derived parameters' in hub methods, so I've prefixed the related artefacts (classes, interfaces, methods) with 'DerivedParameter'.
If there's another way you'd like to refer to this, let me know and I'll update the names.

anurse: Very nice! Thanks for doing that!

I think 'DerivedParameter' is a fine name. I'll let the team give a review and expect we'll be able to get this merged in.

BrennanConroy (approved):
Looks good, however I would like to avoid the usage of `TypeNameHandling.All`

src/SignalR/server/SignalR/test/HubConnectionHandlerTests.cs
```diff
@@ -3576,6 +3576,62 @@ public async Task CanPassStreamingParameterToStreamHubMethod()
...
+            {
+                var argument = new StreamingHub.DerivedParameterTestObject { Value = "test" };
+                var protocolOptions = new NewtonsoftJsonHubProtocolOptions
+                {
+                    PayloadSerializerSettings = new JsonSerializerSettings()
+                    {
+                        // The usage of TypeNameHandling.All is a security risk.
+                        // If you're implementing this in your own application instead use your own 'type' field and a custom JsonConverter
+                        // or ensure you're restricting to only known types with a custom SerializationBinder.
+                        // See https://github.com/aspnet/AspNetCore/issues/11495#issuecomment-505047422
```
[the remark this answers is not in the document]
anurse: Agreed, we have a pretty hard policy about not using this kind of stuff, even in tests. I'd be OK with the [KnownTypesBinder](https://www.newtonsoft.com/json/help/html/SerializeSerializationBinder.htm) approach here.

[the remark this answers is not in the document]
NickDarvey: I've added a `DerivedParameterKnownTypesBinder` which limits the types to only the test type.

src/SignalR/server/SignalR/test/HubConnectionHandlerTests.cs
```diff
@@ -3576,6 +3576,59 @@ public async Task CanPassStreamingParameterToStreamHubMethod()
...
+        {
+            using (StartVerifiableLog())
+            {
+                var argument = new StreamingHub.DerivedParameterTestObject { Value = "test" };
+                var protocolOptions = new NewtonsoftJsonHubProtocolOptions
+                {
+                    PayloadSerializerSettings = new JsonSerializerSettings()
+                    {
+                        TypeNameHandling = TypeNameHandling.All,
+                        SerializationBinder = StreamingHub.DerivedParameterKnownTypesBinder.Instance
```
halter73: I think we should keep the comment about TypeNameHandling.All being a potential security risk. We can just mention it's mitigated by providing a custom ISerializationBinder like you do here.

halter73 (approved):

src/SignalR/server/SignalR/test/HubConnectionHandlerTests.cs
NickDarvey: Done

anurse:
Thanks @NickDarvey !

PR #11504, opened 2019-06-24
[Fixes #11454] Fix encoding bug on prerendering

### LINKED ISSUE #11454: The Blazor circuitId can contain two consecutive dashes
The pre-rendered HTML of a server-side component contains a configuration string comment (`<!-- M.A.C.Component:{"circuitId":"...`) and the circuitId there can contain a double dash (`--`).
This is invalid HTML syntax: https://www.w3.org/TR/html51/syntax.html#sec-comments

Different browsers react differently. Chromium (for example Edge Dev) doesn't really care, but I have seen at least a warning in IE11.

### To Reproduce
Given how the circuitId is generated: Try often enough?

### Expected behavior
Valid HTML syntax

### Screenshots
image

javiercn:
Transforms the circuit id when it sees two dashes in a row and replaces them for underscores, as the HTML parser in some browsers produce warnings


src/Components/Server/src/Circuits/CircuitPrerenderer.cs
```diff
@@ -71,8 +76,14 @@ public async Task<ComponentPrerenderResult> PrerenderComponentAsync(ComponentPre
                 Prerendered = true
             });
 
+            var record = JsonSerializer.ToString(new PrerenderedComponentRecord(
+                    circuitHost.CircuitId.Replace("--", "__"), // We need to do this due to the fact that -- is not allowed within HTML comments and HTML doesn't encode '-'.
```
Tragetaschen: I was pretty sure, I saw `_` in the generated circuitId, so I looked it up:
https://github.com/aspnet/AspNetCore/blob/f35564ba0635da588214791b3662085bd30d243e/src/Shared/WebEncoders/WebEncoders.cs#L385

Tragetaschen: `CircuitIdFactory` -> `Microsoft.AspNetCore.WebUtilities.Base64UrlTextEncoder` -> `WebEncoders`

src/Components/Server/src/Circuits/CircuitPrerenderer.cs
```diff
@@ -71,8 +76,14 @@ public async Task<ComponentPrerenderResult> PrerenderComponentAsync(ComponentPre
                 Prerendered = true
             });
 
+            var record = JsonSerializer.ToString(new PrerenderedComponentRecord(
+                    circuitHost.CircuitId.Replace("--", ".."), // We need to do this due to the fact that -- is not allowed within HTML comments and HTML doesn't encode '-'.
```
pranavkm: Is the expectation that `CircuitId` would never have `..` in them? Or is that having existing '..` also change to dashes isn't problematic? Could we note that expectation as a comment here?

src/Components/Server/src/Circuits/CircuitPrerenderer.cs
```diff
@@ -164,6 +175,22 @@ private string GetFullBaseUri(HttpRequest request)
             return result;
         }
 
+        private struct PrerenderedComponentRecord
```
pranavkm: `readonly struct`

pranavkm (approved):

src/Components/Server/src/Circuits/CircuitPrerenderer.cs
javiercn: We will never have '.' on them because we base64url encode the id and the only non alphanumeric symbols are '-' and '_', dots are safe on the url, so we simply replace '-' with '.'. You could say its our "own custom base64 encoding scheme"

pranavkm: 👍  Could you make a note of that?

SteveSandersonMS (approved):

rynowak:
@javiercn - any reason why this shouldn't be merged?

javiercn:
There were some test failures I have to investigate.

PR #11505, opened 2019-06-24
Remove Hosting.Internal APIs

Tratcher:
#11312 Everything became internal except IStartupConfigureContainerFilter and IStartupConfigureServicesFilter which became public but obsolete. These are only used in testing infrastructure, and only for WebHost.

I did need to clean up a mess where the Internal HostingApplication.Context had leaked into TestHost.


src/Hosting/TestHost/test/ClientHandlerTests.cs
```diff
@@ -421,7 +420,7 @@ public Task ExceptionFromOnStartingWithErrorHandlerIsReported()
                 HttpCompletionOption.ResponseHeadersRead));
         }
 
-        private class DummyApplication : IHttpApplication<Context>
+        private class DummyApplication : ApplicationWrapper, IHttpApplication<HostingApplication.Context>
```
davidfowl: This is the last thing to clean up or is this a test so it's fine?

Tratcher: Just a test, I'm not worried about it.

davidfowl: Sure, it's pretty horrible though

davidfowl (approved):

src/Hosting/TestHost/test/ClientHandlerTests.cs
Tratcher: Fixed

davidfowl (approved):

halter73 (approved):

johnnyreilly:
When the `IStartupConfigureContainerFilter` is actually removed then this will create issues for Autofac users relying on this workaround: https://blog.johnnyreilly.com/2020/10/02/autofac-6-integration-tests-and-generic-hosting/

This is to do with `ConfigureTestContainer` not working with `GenericHost` - context here: https://github.com/dotnet/aspnetcore/issues/14907

PR #11512, opened 2019-06-24
Update refs

### LINKED ISSUE #10954: Missing completions for Router parameters
I'm not getting completions for some of the new templated parameters on the Router component, like [NotAuthorizedContent](https://github.com/aspnet/AspNetCore/blob/master/src/Components/Components/src/Routing/Router.cs#L46):

image

pranavkm:
Fixes https://github.com/aspnet/AspNetCore/issues/10954

### DIFF
--- a/src/Components/test/testassets/ComponentsApp.App/ComponentsApp.App.csproj
+++ b/src/Components/test/testassets/ComponentsApp.App/ComponentsApp.App.csproj
@@ -7,7 +7,7 @@
 
   <ItemGroup>
     <Reference Include="Microsoft.Extensions.DependencyInjection" />
-    <Reference Include="Microsoft.AspNetCore.Components" />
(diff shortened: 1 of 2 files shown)


src/Components/test/testassets/ComponentsApp.App/ComponentsApp.App.csproj
```diff
@@ -7,7 +7,7 @@
 
   <ItemGroup>
     <Reference Include="Microsoft.Extensions.DependencyInjection" />
-    <Reference Include="Microsoft.AspNetCore.Components" />
+    <ProjectReference Include="..\..\..\Components\ref\Microsoft.AspNetCore.Components.csproj" />
```
pranavkm: It's not much, but we can get the Component.App to start compiling against the ref assemblies. TestAssets is more valuable, but it's much more churn to do so since it references Blazor.

SteveSandersonMS (approved):

Issue #10734, opened 2019-06-01
~/ should not require async


NTaylorMullen:
Because of how ~/ is implemented today (as a TagHelper) it requires the method that contains it to be asynchronous. This means when utilizing `~/` in local functions, lambdas etc. you must mark those methods as `async Task`. Sadly, `~/` does not show up at design time as a TagHelper so you don't get errors for this scenario.

All the being said, we should consider moving ~/ off of the TagHelper infrastructure and make it a compiler intrinsic.

Filing this issue because https://github.com/aspnet/AspNetCore/issues/8630 did not cover the `~/` scenario.

NTaylorMullen:
So I've been investigating multiple ways to do this over the past several days all of which have hit various roadblocks all exposing different issues with fulfilling this issues core requirement. The core requirement for this feature is: Work at least as well as the `UrlResolutionTagHelper` works today.

To make that possible we need to support:
1. Plain HTML attributes, `<img src="~/some/path.png" />`
2. Mixed HTML attributes, `<img src="~/@somePath more stuff" />`
3. TagHelper bound attributes, (imagine there's another ImageTagHelper with `public string Src { get; set; }`), `<img src="~/some/path.png" />`
4. TagHelper unbound attributes  `<img src="~/some/path.png" />`

Less important but possible today:

5. HTML  attributes with statements, `<img src="~/@somePath @if (true) { @extra } " />`
6. Ignores non-standard `~/` attributes, `<randomTag something="~/path/to/somewhere" />`

So we're left with a few options:

### Option 1
Stop supporting all that `UrlResolutionTagHelper` does today. Only support 1, 3 and 4 from above. 

Pros: Lesser implementation/maintenance cost
Cons: Will make it more difficult to port some applications because of differing behavior from older versions of Razor.

### Option 2
Buffer the world. Everytime we see a `~/` build a new system to buffer values and then resolve their app relative URL after rendering.

Pros: Keeps consistency with previous versions of Razor
Cons: Lots of work, hard to maintain, not very performant.

### Option 3
Leave `~/` resolution async and have it show up in VS as a TagHelper to provide valid errors in markup blocks.

Pros: Easy to implement and maintain
Cons: TagHelpers still have to pay the cost to go async and any methods that contain a `~/` valid attribute must be async.

### Option 4
Do nothing
Pros:  lol
Cons: ¯\\_(ツ)_/¯

__________________________________

[...]

PR #11514, opened 2019-06-24
Make ~/ resolution show up at design time.

NTaylorMullen:
- This change results in proper errors for tags that utilize `~/`inside of local functions at design time.

Turns out my VS changes that I did for the last release 99% cover improper coloring of this scenario. The 1 case that it doesn't is when a user utilizes the `itemid` attribute on a non-standard HTML tag:

image

Addresses #10734

### DIFF
--- a/src/Mvc/Mvc.Analyzers/src/DiagnosticDescriptors.cs
+++ b/src/Mvc/Mvc.Analyzers/src/DiagnosticDescriptors.cs
@@ -59,7 +59,7 @@ namespace Microsoft.AspNetCore.Mvc.Analyzers
             new DiagnosticDescriptor(
                 "MVC1006",
                 "Methods containing TagHelpers must be async and return Task.",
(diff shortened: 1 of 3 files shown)


rynowak:
LOL EZ.

Is there anything we can do to improve the error message? Should we mention `~/` specifically in the error message? I could see a lot of confusion coming from this error message about tag helpers and saying "but there isn't a tag helper???"

NTaylorMullen:
> I could see a lot of confusion coming from this error message about tag helpers and saying "but there isn't a tag helper???"

Ya I could add an example to the analyzer error which utilizes ~/ as an example to make it extra clear.

NTaylorMullen:
@rynowak Updated error message to be: `The method contains a TagHelper and therefore must be async and return a Task. For instance, usage of ~/ typically results in a TagHelper and requires an async Task returning parent method.`

rynowak (approved):

PR #11516, opened 2019-06-24
Dylan/event source

DylanDmitri:
EventSource for the new RequestThrottling middleware. Exposes counters for:

 - `queue-length` (current length of queue)
 - `queue-duration` (rolling average for how long currently processing requests have had to wait)
 - `requests-rejected` (number of requests returned with 503-server-busy errors)

Emits a Warning event on rejected requests. 

Possibly want to change: 
 - change `requests-rejected` from a cumulative total to a rate


src/Middleware/RequestThrottling/src/RequestThrottlingEventSource.cs
```diff
@@ -0,0 +1,102 @@
...
+            _parent = Log
+        };
+
+        private PollingCounter _rejectedRequestsCounter;   // incrementing polling counter
+        private PollingCounter _queueLengthCounter;        // polling counter
+        private EventCounter _queueDuration;     // event counter average
+
+        internal long _rejectedRequests;
+        internal int _queueLength;
+
```
jkotalik:
```suggestion

```

src/Middleware/RequestThrottling/src/RequestThrottlingEventSource.cs
```diff
@@ -0,0 +1,102 @@
...
+            internal RequestThrottlingEventSource _parent;
+
+            public void Dispose()
+            {
+                Interlocked.Decrement(ref _parent._queueLength);
+
+                if (_parent.IsEnabled())
+                {
+                    var duration = _timer.IsActive ? _timer.GetElapsedTime().TotalMilliseconds : 0.0;
+                    _parent._queueDuration.WriteMetric(duration);
```
jkotalik: I'm not sure how this is calculating an average. Can you clarify why/how this works?

src/Middleware/RequestThrottling/src/RequestThrottlingEventSource.cs
```diff
@@ -0,0 +1,102 @@
...
+                {
+                    _timer = ValueStopwatch.StartNew(),
+                    _parent = this
+                };
+            }
+
+            return _cachedNonTimerResult;
+        }
+
+        internal struct QueueFrame : IDisposable
```
jkotalik: This pattern is interesting for creating your own event source. Is this the natural way? Did you find docs/ info on how to do this?

src/Middleware/RequestThrottling/src/RequestThrottlingEventSource.cs
```diff
@@ -0,0 +1,102 @@
...
+using System.Text;
+using System.Threading;
+using Microsoft.Extensions.Internal;
+
+namespace Microsoft.AspNetCore.RequestThrottling
+{
+    internal sealed class RequestThrottlingEventSource : EventSource
+    {
+        public static readonly RequestThrottlingEventSource Log = new RequestThrottlingEventSource();
+        private static QueueFrame _cachedNonTimerResult = new QueueFrame
```
jkotalik: nit: make this a property with a getter that caches it after use. That way, you don't always allocate the frame.

src/Middleware/RequestThrottling/src/RequestThrottlingEventSource.cs
```diff
@@ -0,0 +1,102 @@
...
+        {
+            _parent = Log
+        };
+
+        private PollingCounter _rejectedRequestsCounter;   // incrementing polling counter
```
jkotalik: nit: comments are a bit redundant. Consider dropping them.

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
```diff
@@ -20,8 +20,6 @@ public class RequestThrottlingMiddleware
         private readonly RequestDelegate _onRejected;
         private readonly ILogger _logger;
 
-        private int _queuedRequests;
```
jkotalik: As we chatted about earlier, I don't know if we should rely on EventSource to count the number of requests for the QueuedRequestCount. It looks like you need to instantiate the _queueCount to zero, which seems odd too.

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
```diff
@@ -49,19 +47,17 @@ public RequestThrottlingMiddleware(RequestDelegate next, ILoggerFactory loggerFa
         /// <returns>A <see cref="Task"/> that completes when the request leaves.</returns>
         public async Task Invoke(HttpContext context)
         {
-            Interlocked.Increment(ref _queuedRequests);
+            var waitInQueueTask = _queuePolicy.TryEnterAsync();
```
jkotalik: Nice little perf improvement here 🎈

Kahbazi: Based on @halter73 's [comment](https://github.com/aspnet/AspNetCore/pull/11293#discussion_r296336437), I changed `Task<bool>` to a resettable `ValueTask<bool>` to see the impacts, and if there's a plan to convert to `ValueTask<bool>` I think this would throw an exception, since you can't access the property of a `ValueTask` after it's resetted.

src/Middleware/RequestThrottling/src/RequestThrottlingEventSource.cs
anurse: `EventCounter`s automatically calculate averages. You just have have to write the individual values and it produces Max, Min, Avg and StdDev.

src/Middleware/RequestThrottling/src/RequestThrottlingEventSource.cs
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        // Used for testing
+        internal RequestThrottlingEventSource(string eventSourceName)
+            : base(eventSourceName)
+        {
+        }
+
+        [Event(1, Level = EventLevel.Warning)]
+        public void RequestRejected()
```
anurse: Should we include the request's `HttpContext.TraceIdentifier` value here? cc @halter73 @davidfowl ?

It's not relevant to the counter, which is just a number, but would be useful data if you were actually capturing the raw event. We don't really have access to logging scopes here, so we can't rely on that.

anurse: @DylanDmitri get @halter73 and/or @shirhatti to help you capture the raw events as well so you can see how that works and the data associated with them. You can use `dotnet-trace`, by enabling the `Microsoft.AspNetCore.RequestThrottling` provider, and view the data in PerfView:

1. Install using: `dotnet tool install --global dotnet-trace --version 1.0.4-preview6.19311.1`
1. Launch the app, get the process ID
1. Start collecting: `dotnet trace collect --providers Microsoft.AspNetCore.RequestThrottling -p [pid]`
1. Open `trace.netperf` in [PerfView](http://github.com/Microsoft/PerfView)

src/Middleware/RequestThrottling/src/RequestThrottlingEventSource.cs
anurse: There aren't really good patterns here yet. This seems reasonable. It's pretty common to want to do "duration" counters like this (with an `IDiposable` that lets you wrap a lexical scope representing the operation). SignalR uses a different pattern of:

```csharp
public ValueStopwatch ConnectionStart(string connectionId);
public void ConnectionStop(string connectionId, ValueStopwatch timer)
```

Which is basically the same behavior but requires the caller to track it. In that case though, SignalR doesn't have a lexical scope that defines the lifetime of a connection.

src/Middleware/RequestThrottling/src/RequestThrottlingEventSource.cs
davidfowl: We don't really have a good way to look at these events and their arguments, we're mainly using the EventSource as a means to an end (counters), so I wouldn't spend too much time on the structure of the events as we're inconsistent all over the place.

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
DylanDmitri: Expose the current queue length through an internal field in `IQueuePolicy`? The field would make testing better, but would be internal and only used in testing.

DylanDmitri: Maybe make a `TestQueuePolicy` which implements `IQueuePolicy` but has additional fields. Use this test policy in the tests. This is the "correct" approach but adds code and abstraction in the tests.

src/Middleware/RequestThrottling/src/RequestThrottlingEventSource.cs
DylanDmitri: There's also 
```C#
public async Task TimeWaitInQueue(Task<bool> waitInQueueTask)
```
Creates the stopwatch, awaits the task, writes the results. Feels bad from a layered architecture standpoint.

src/Middleware/RequestThrottling/src/RequestThrottlingEventSource.cs
DylanDmitri: I've ran the tool and watched the counters; they function as expected.

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
DylanDmitri: @Kahbazi The code in the middleware here is good, what would need to switch is the code in the individual policies. I hit some of problems you describe when I tried switching to `ValueTask`; specifically resetting a `ValueTask` to immediately use it for another request is a bad idea. One approach that I tried was having n+1 `ValueTasks` and resetting not the task to be immediately used, but the task that would be needed by the next request if the queue were still full. I wasn't able to get a satisfactory & reliable implementation, but may revisit if I have time later.

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
DylanDmitri: Already had a `TestQueuePolicy`, so added the field.

src/Middleware/RequestThrottling/sample/Startup.cs
```diff
@@ -22,10 +22,10 @@ public class Startup
         // For more information on how to configure your application, visit https://go.microsoft.com/fwlink/?LinkID=398940
         public void ConfigureServices(IServiceCollection services)
         {
-            services.AddTailDropQueue((options) =>
+            services.AddStackQueue((options) =>
             {
                 options.MaxConcurrentRequests = Math.Max(1, _config.GetValue<int>("maxCores"));
-                options.RequestQueueLimit = Math.Max(1, _config.GetValue<int>("maxQueue"));
+                options.RequestQueueLimit = Math.Max(5, _config.GetValue<int>("maxQueue"));
```
halter73: Why did you make this 5?

src/Middleware/RequestThrottling/src/RequestThrottlingEventSource.cs
```diff
@@ -0,0 +1,110 @@
...
+using System.Text;
+using System.Threading;
+using Microsoft.Extensions.Internal;
+
+namespace Microsoft.AspNetCore.RequestThrottling
+{
+    internal sealed class RequestThrottlingEventSource : EventSource
+    {
+        public static readonly RequestThrottlingEventSource Log = new RequestThrottlingEventSource();
+        private static QueueFrame _cachedNonTimerResult
```
halter73: Property names shouldn't start with `_`

```suggestion
        private static QueueFrame CachedNonTimerResult
```

src/Middleware/RequestThrottling/src/RequestThrottlingEventSource.cs
```diff
@@ -0,0 +1,110 @@
...
+        {
+            get
+            {
+                if (_queueFrameCache == null)
+                {
+                    _queueFrameCache = new QueueFrame { _parent = Log };
```
halter73: Since this is a static property this should probably be thread-safe. Usually that means using `Lazy<T>` or manually locking, but in this case I think we can just greedily initialize it like we do for the RequestThrottlingEventSource. The static initializers/ctor for the class shouldn't be run until the class is reference anyway.

src/Middleware/RequestThrottling/src/RequestThrottlingEventSource.cs
halter73: Actually, just change this to an eagerly-initialized static field.

src/Middleware/RequestThrottling/src/RequestThrottlingEventSource.cs
```diff
@@ -0,0 +1,110 @@
...
+                };
+            }
+
+            return _cachedNonTimerResult;
+        }
+
+        internal struct QueueFrame : IDisposable
+        {
+            internal ValueStopwatch _timer;
+            internal RequestThrottlingEventSource _parent;
```
halter73: Make these private and add a ctor.

src/Middleware/RequestThrottling/src/RequestThrottlingEventSource.cs
```diff
@@ -0,0 +1,110 @@
...
+
+            public void Dispose()
+            {
+                Interlocked.Decrement(ref _parent._queueLength);
+
+                if (_parent.IsEnabled())
+                {
+                    var duration = _timer.IsActive ? _timer.GetElapsedTime().TotalMilliseconds : 0.0;
```
halter73: In what cases wouldn't the timer be active here?

Also, if we're making QueueFrame a struct and using a ValueStopwatch, doesn't that imply we expect the request to complete without every going async and releasing the thread/stack? In that case, are milliseconds precise enough of a measurement?

src/Middleware/RequestThrottling/src/RequestThrottlingEventSource.cs
```diff
@@ -0,0 +1,110 @@
...
+                };
+
+                _queueLengthCounter ??= new PollingCounter("queue-length", this, () => _queueLength)
+                {
+                    DisplayName = "Queue Length",
+                };
+
+                _queueDuration ??= new EventCounter("queue-duration", this)
+                {
+                    DisplayName = "Average Queue Duration",
```
halter73:
```suggestion
                    DisplayName = "Average Time in Queue",
```

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
```diff
@@ -49,19 +47,17 @@ public RequestThrottlingMiddleware(RequestDelegate next, ILoggerFactory loggerFa
...
-            var success = false;
-            try
-            {
-                success = await _queuePolicy.TryEnterAsync();
-            }
-            finally
+            if (!waitInQueueTask.IsCompleted)
             {
-                Interlocked.Decrement(ref _queuedRequests);
+                using (RequestThrottlingEventSource.Log.QueueTimer())
```
halter73: If we're calculating "Average Queue Duration" or "Average Time in Queue" using this counter, shouldn't we count requests that don't need to wait on the queue as spending 0 time in the queue?

src/Middleware/RequestThrottling/sample/Startup.cs
DylanDmitri: This website is meant for people to run, and refresh the page with their fingers until they refresh too fast and it 503s. Setting the limit at 5 rather than 1 results in a more satisfying demo experience.

src/Middleware/RequestThrottling/src/RequestThrottlingEventSource.cs
DylanDmitri: Changed it back to eager-initialized, it will allocate the struct every time.

src/Middleware/RequestThrottling/src/RequestThrottlingEventSource.cs
DylanDmitri: I copy-pasted this from some other tests. I'm erring on the side of caution; I do not know if inactive timers are possible.

DylanDmitri: From benchmarking, I suspect the queues are only useful on longer running requests (5ms+). I'm open to better methods of measuring.

src/Middleware/RequestThrottling/src/RequestThrottlingMiddleware.cs
DylanDmitri: Nice that's a big oversight from me. Want to either
 - add a "queue skipped" event that reports duration 0
 - wrap more of the middleware invocation in the `using(QueueTimer)`

src/Middleware/RequestThrottling/sample/Startup.cs
halter73: Then add a config file to the sample and change it there. We shouldn't silently ignore what was configured.

src/Middleware/RequestThrottling/src/RequestThrottlingEventSource.cs
halter73: The funny thing is you're allocating at exactly the same time you were before, but now you're allocating less.

Since QueueFrame is a struct, the lazy initialization was even more pointless than usual. Both QueueFrame and `Nullable<QueueFrame>` are structs whos values are contained directly in the parent object instead of indirectly referenced. It just that `Nullable<QueueFrame>` is even bigger since it contains both QueueFrame and the HasValue bool.

By delaying the construction of the type, it only delayed a method call (i.e. the constructor) that initialized the space already reserved for the QueueFrame. This could maybe make sense if the constructor was doing something really expensive, but then I'd argue the expensive operation should probably be moved outside of the constructor.

src/Middleware/RequestThrottling/src/RequestThrottlingEventSource.cs
halter73: > I'm erring on the side of caution; I do not know if inactive timers are possible.

As a general rule, handling cases that you know cannot or should not happen in a graceful way, ends up creating more problems than it solves.

It causes future programmers to defensively code around the same impossible case which adds verbosity and untestable code, since it's literally impossible to hit the case even in a test.

If the assumptions change, and a previously-impossible case is now possible, that usually indicates a bug. In that case you want to fail as quickly as possible. You don't want never-before-tested code to suddenly start trying to handle it gracefully.

In that case, the far more cautious thing to do is to fail fast with an Trace.Assert failure or an exception. I'm starting to really dislike Debug.Asert though.

src/Middleware/RequestThrottling/sample/Startup.cs
DylanDmitri: oh yeah I forgot about the Math.Max

Kahbazi:
@DylanDmitri I don't know if it's intentional or not but `IQueuePolicy` location and namespace don't match.

src/Middleware/RequestThrottling/test/MiddlewareTests.cs
```diff
@@ -73,36 +74,62 @@ public async void RequestsBlockedIfQueueFull()
...
+            Assert.Equal(2, testQueue.QueuedRequests);
+        }
+
+        [Fact]
+        public void EventCountersTrackQueuedRequests()
+        {
+            var blocker = new TaskCompletionSource<bool>();
+
+            // requires
+            var testQueue = new TestPolicy(
```
jkotalik: I feel like these `testQueue` constructors aren't clear about what is being tested. Maybe it's the name of TestPolicy or something.  Also, maybe changing invoke to onInvoke?

src/Middleware/RequestThrottling/test/MiddlewareTests.cs
```diff
@@ -73,36 +74,62 @@ public async void RequestsBlockedIfQueueFull()
...
+
+            Assert.Equal(0, testQueue.QueuedRequests);
         }
 
         [Fact]
         public async Task CleanupHappensEvenIfNextErrors()
         {
             var flag = false;
 
+            var testQueue = new TestPolicy(
```
jkotalik: Same here. It's really hard to tell what's going on here.

src/Middleware/RequestThrottling/test/MiddlewareTests.cs
```diff
@@ -73,36 +74,62 @@ public async void RequestsBlockedIfQueueFull()
...
+
+            blocker.SetResult(true);
+
+            Assert.Equal(0, testQueue.QueuedRequests);
         }
 
         [Fact]
         public async Task CleanupHappensEvenIfNextErrors()
         {
             var flag = false;
```
jkotalik:
```suggestion
            var onExitCalled = false;
```

src/Middleware/RequestThrottling/test/MiddlewareTests.cs
```diff
@@ -111,6 +138,12 @@ public async void ExceptionThrownDuringOnRejected()
         {
             TaskCompletionSource<bool> tsc = new TaskCompletionSource<bool>();
 
+            var testQueue = new TestPolicy(
+                invoke: (state) =>
```
jkotalik: What is state here? Is it the queue?

src/Middleware/RequestThrottling/test/MiddlewareTests.cs
```diff
@@ -111,6 +138,12 @@ public async void ExceptionThrownDuringOnRejected()
         {
             TaskCompletionSource<bool> tsc = new TaskCompletionSource<bool>();
```
jkotalik:
```suggestion
            TaskCompletionSource<bool> tcs = new TaskCompletionSource<bool>();
```

jkotalik (approved):
Some super minor nits, mostly relating to clarity.

src/Middleware/RequestThrottling/ref/Microsoft.AspNetCore.RequestThrottling.netcoreapp3.0.cs
```diff
@@ -30,6 +24,11 @@ public partial class RequestThrottlingOptions
 }
 namespace Microsoft.AspNetCore.RequestThrottling.QueuePolicies
 {
+    public partial interface IQueuePolicy
```
halter73: I don't mind organizing the directory structure this way, but I think all the public types should live in the Microsoft.AspNetCore.RequestThrottling namespace. Having to "use" two namespaces just to use this feature would be pretty annoying.

src/Middleware/RequestThrottling/src/RequestThrottlingEventSource.cs
```diff
@@ -0,0 +1,109 @@
...
+using System.Text;
+using System.Threading;
+using Microsoft.Extensions.Internal;
+
+namespace Microsoft.AspNetCore.RequestThrottling
+{
+    internal sealed class RequestThrottlingEventSource : EventSource
+    {
+        public static readonly RequestThrottlingEventSource Log = new RequestThrottlingEventSource();
+        private static QueueFrame CachedNonTimerResult = new QueueFrame(null, Log);
```
halter73: nit: readonly

halter73 (approved):

src/Middleware/RequestThrottling/test/MiddlewareTests.cs
```diff
@@ -73,69 +65,118 @@ public async void RequestsBlockedIfQueueFull()
...
         {
-            var middleware = TestUtils.CreateTestMiddleware(
-                queue: TestStrategy.AlwaysBlock);
+            var testQueue = TestQueue.AlwaysBlock;
+            var middleware = TestUtils.CreateTestMiddleware(testQueue);
 
-            Assert.Equal(0, middleware.QueuedRequestCount);
+            Assert.Equal(0, testQueue.QueuedRequests);
 
             _ = middleware.Invoke(new DefaultHttpContext());
```
halter73: Assert these two invoke Tasks aren't completed too.

src/Middleware/RequestThrottling/ref/Microsoft.AspNetCore.RequestThrottling.netcoreapp3.0.cs
Tratcher: As a general rule do not create a new sub-namespace if it's only going to have a few types in it. (< 5?)

Issue #11005, opened 2019-06-07
Add logging for controller creation


WBarge:
I am seeing a slow down in the processing of a request before it is handed over to the controller.
if you look in the log you will see 1.6sec time elapse between "Execution plan of "result" filters" and "Attempting to bind parameter" (timestamp 2019-06-07 14:45:36.810 -07:00 in the log below)

My logic is taking 58.2462ms. per the logs which is ok but over all the request is taking 1759.5546ms

here is the log of the request
```
14:45:36.767 -07:00 [Information] Request starting HTTP/1.1 GET http://localhost:44385/api/Client/GetClientsForTimeKeeper?timeKeeperId=-1&searchStr=morgan&maxNumOfRecordsReturned=50  
===> 2019-06-07 14:45:36.768 -07:00 [Debug] Connection id ""0HLNBI59LIRA8"" received FIN.
===> 2019-06-07 14:45:36.773 -07:00 [Debug] Connection id ""0HLNBI59LIRA8"" disconnecting.
===> 2019-06-07 14:45:36.799 -07:00 [Debug] Request successfully matched the route with name 'null' and template '"api/Client/GetClientsForTimeKeeper"'
===> 2019-06-07 14:45:36.800 -07:00 [Debug] Connection id ""0HLNBI59LIRA8"" sending FIN because: ""The Socket transport's send loop completed gracefully.""
===> 2019-06-07 14:45:36.802 -07:00 [Information] Route matched with "{action = \"GetClientsForTimeKeeper\", controller = \"Client\"}". Executing controller action with signature "System.Threading.Tasks.Task`1[Microsoft.AspNetCore.Mvc.IActionResult] GetClientsForTimeKeeper(System.String, System.String, Int32)" on controller "Minisoft.Api.Controllers.API.ClientController" ("Minisoft.Api").
===> 2019-06-07 14:45:36.804 -07:00 [Debug] Connection id ""0HLNBI59LIRA8"" stopped.
===> 2019-06-07 14:45:36.805 -07:00 [Debug] Execution plan of "authorization" filters (in the following order): ["None"]
===> 2019-06-07 14:45:36.806 -07:00 [Debug] Execution plan of "resource" filters (in the following order): ["None"]
===> 2019-06-07 14:45:36.807 -07:00 [Debug] Execution plan of "action" filters (in the following order): ["None"]
[...]

PR #11526, opened 2019-06-24
 Mvc.Core logging for controller and page factories

alefranz:
This adds logging around controller and page creation.

- `ControllerActionInvoker` (and `PageActionInvoker`) are the best place to log it is in because, even if it doesn't allow to get more fine grained info, it work for any factory and would not required a closure on the factory delegate.
- Not sure which `EventId` ids I should use. What is the logic? Incremental numbers and reusing it for different log levels?
- I wasn't able to find existing tests around similar logging. Is it worth adding some?
- I've included a fix for the name of a test in the same area but not related to this change, let me know if I should move it to a separate PR.

Addresses #11005


src/Mvc/Mvc.Core/src/MvcCoreLoggerExtensions.cs
```diff
@@ -155,6 +158,16 @@ internal static class MvcCoreLoggerExtensions
...
         {
+            _controllerFactoryExecuting = LoggerMessage.Define<string, string>(
+                LogLevel.Debug,
+                new EventId(1, "ControllerFactoryExecuting"),
+                "Executing controller factory for controller {Controller} ({AssemblyName})");
+
+            _controllerFactoryExecuted = LoggerMessage.Define<string, string>(
+                LogLevel.Debug,
+                new EventId(2, "ControllerFactoryExecuted"),
+                "Executed controller factory for controller {Controller} ({AssemblyName})");
```
rynowak: re: your question about event ids, they need to be unique per logging source, which these are :+1:

src/Mvc/Mvc.Core/test/Controllers/ControllerFactoryProviderTest.cs
```diff
@@ -64,7 +64,7 @@ public void CreateControllerReleaser_InvokesIControllerFactory_IfItIsNotDefaultC
         }
 
         [Fact]
-        public void CreateControllerReleaser_UsesControllerActivatorAndPropertyActivator()
+        public void CreateControllerFactory_UsesControllerActivatorAndPropertyActivator()
```
rynowak: 👍

src/Mvc/Mvc.RazorPages/src/PageLoggerExtensions.cs
```diff
@@ -27,6 +32,26 @@ static PageLoggerExtensions()
         {
             // These numbers start at 101 intentionally to avoid conflict with the IDs used by ResourceInvoker.
 
+            _modelFactoryExecuting = LoggerMessage.Define<string, string>(
+                LogLevel.Debug,
+                new EventId(101, "ExecutingModelFactory"),
+               "Executing model factory for page {Page} ({AssemblyName})");
```
rynowak: I'd like the see the text of this message (and the method/field names) updated to use "page model factory" instead of "model factory" - that's the correct term, and it would be possible to confuse to the model term with some other concepts.

rynowak:
> I wasn't able to find existing tests around similar logging. Is it worth adding some?

It would be nice to add some tests for these if you can. You should be able to find *some* tests for the logging in resource invoker to use as a base. 

The risk is that the logging could get broken and throw, and we only find that out when someone actually needs it and turns it out. We've had this happen a few times 😢

alefranz:
@rynowak I've added tests.
If this approach looks good I can add tests for the rest of the logging in the `Invoker`s in a separate PR.

Thank you,
 Alessio

rynowak:
Thanks is great! Thanks. Up to you if you feel like sending a PR to add more tests.

rynowak (approved):

alefranz:
Can you confirm that the failure in the build is unrelated?
Thank you

rynowak:
Agreed, I retried the failing parts of the build.

alefranz:
Thanks! only Helix is failing now. Is it good to be merged?

rynowak:
Merged. Thanks!

PR #11528, opened 2019-06-25
Fixup ref assemblies again

### LINKED ISSUE #11240: Namespace of PageDisplay.cs
### Describe the bug
The PageDisplay.cs is situated in Layouts Folder (Microsoft\AspNetCore\Components\Layouts ) but its namespace is Microsoft.AspNetCore.Components (without Layouts). I create a custom router. If I point Microsoft.AspNetCore.Components.Layouts namespace then I have the compile error "error CS0103: The name 'PageDisplay' does not exist in the current context". When I point  Microsoft.AspNetCore.Components namespace I cannot compile the project at all because I have "CS0234	The type or namespace name 'PageDisplay' does not exist in the namespace 'Microsoft.AspNetCore.Components' (are you missing an assembly reference?)". What I do wrong?

pranavkm:
Fixes https://github.com/aspnet/AspNetCore/issues/11240

### DIFF
--- a/src/Components/Components/ref/Microsoft.AspNetCore.Components.netstandard2.0.Manual.cs
+++ b/src/Components/Components/ref/Microsoft.AspNetCore.Components.netstandard2.0.Manual.cs
@@ -102,6 +102,21 @@ namespace Microsoft.AspNetCore.Components
         public void Configure(Microsoft.AspNetCore.Components.RenderHandle renderHandle) { }
         public System.Threading.Tasks.Task SetParametersAsync(Microsoft.AspNetCore.Components.ParameterCollection parameters) { throw null; }
     }
+
+    public partial class PageDisplay : Microsoft.AspNetCore.Components.IComponent
+    {
+        public PageDisplay() { }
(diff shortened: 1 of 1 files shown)


danroth27 (approved):
LGTM

rynowak (approved):

CornedBee:
This is the important part, fixing the ref assembly. The PageDisplay.cs file is still in the wrong place in the directory hierarchy given its namespace.

SteveSandersonMS:
> The PageDisplay.cs file is still in the wrong place in the directory hierarchy given its namespace.

We don't always mirror the namespace hierarchy in the directory hierarchy. Sometimes there are constraints that make us want a particular namespace for the consumer that isn't also what we want from a file organization perspective.

However there are inconsistencies that aren't based on any reasoning. It doesn't really make sense that `PageDisplay` is in the root namespace for components, but `LayoutAttribute` isn't. These kinds of unintentional variations are part of what we intend to address with #4050.

SteveSandersonMS (approved):

PR #11533, opened 2019-06-25
Improve IIS UrlRewrite parsing

cnblogs-dudu:
This PR addresses two null reference exceptions.
*  If the `<rewrite>` tag is missing in the UrlRewite xml file, the following NullReferenceException is thrown.
```text
Application startup exception: System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.AspNetCore.Rewrite.IISUrlRewriteOptionsExtensions.AddIISUrlRewrite(RewriteOptions options, TextReader reader, Boolean alwaysUseManagedServerVariables) in C:\Dev\github\AspNetCore\src\Middleware\Rewrite\src\IISUrlRewriteOptionsExtensions.cs:line 64
```
* If a `rule` element contains a `conditions` without child `add` , the following NullReferenceException is thrown. If using  IIS UrlRewrite Module, the `conditions` is ignored and the `rule` works.
```text
System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.AspNetCore.Rewrite.Internal.IISUrlRewrite.ConditionEvaluator.Evaluate(ConditionCollection conditions, RewriteContext context, BackReferenceCollection backReferences) in C:\Dev\github\AspNetCore\src\Middleware\Rewrite\src\Internal\IISUrlRewrite\ConditionEvaluator.cs:line 46
```

### DIFF
--- a/src/Middleware/Rewrite/src/Internal/IISUrlRewrite/UrlRewriteFileParser.cs
+++ b/src/Middleware/Rewrite/src/Internal/IISUrlRewrite/UrlRewriteFileParser.cs
@@ -28,7 +28,7 @@ namespace Microsoft.AspNetCore.Rewrite.Internal.IISUrlRewrite
 
             if (xmlRoot == null)
             {
-                return null;
+                throw new InvalidUrlRewriteFormatException(new XElement(RewriteTags.Rewrite), "The root element '<rewrite>' is missing");
             }
 
             _inputParser = new InputParser(RewriteMapParser.Parse(xmlRoot), alwaysUseManagedServerVariables);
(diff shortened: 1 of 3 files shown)


jkotalik (approved):
Looks good to me, thanks!

src/Middleware/Rewrite/src/Internal/IISUrlRewrite/UrlRewriteFileParser.cs
```diff
@@ -113,9 +113,15 @@ private void ParseConditions(XElement conditions, UrlRewriteRuleBuilder builder,
 
             var grouping = ParseEnum(conditions, RewriteTags.LogicalGrouping, LogicalGrouping.MatchAll);
             var trackAllCaptures = ParseBool(conditions, RewriteTags.TrackAllCaptures, defaultValue: false);
+            var adds = conditions.Elements(RewriteTags.Add);
+            if (adds == null || adds.Count() == 0)
```
davidfowl: Any instead of Count

cnblogs-dudu: Resloved in #11540

Issue #10698, opened 2019-05-31
Authentication revalidation for server-side Blazor


mkArtakMSFT:

mkArtakMSFT:
Moving the client-side work out to: https://github.com/aspnet/AspNetCore/issues/10846

SteveSandersonMS:
Done in https://github.com/aspnet/AspNetCore/pull/11548

Yen:
Please let me know if I am wrong, but #11548 does not solve this issue for authentication, it solves it for identity based _authorization_.

We are currently using cookie authentication with microsoft AD to validate logins to internal applications where we do not require associated data from a db with identity. For the traditional MVC style system we would set `CookieAuthenticationOptions.ExpireTimeSpan` and be done with it, but how would we go about expiring a server side blazor session in a secure way outside of rolling our own hacks?

What we are currently using is a check in `OnGet` to see if the user is authenticated followed by a redirect to a login razor page from the `_Host.cshtml` file. This is fine for stopping people from loading any of the blazor pages using _Host as the root, but does nothing to stop an existing connection from being used.

Is there any way this system can be used with authentication _without_ identity as we do not want an entity framework store of any sort where not required, and if not, are there plans to have this scenario supported?

ADefWebserver:
@Yen Take a look at: [A Demonstration of Simple Server-side Blazor Cookie Authentication](http://lightswitchhelpwebsite.com/Blog/tabid/61/EntryId/4316/A-Demonstration-of-Simple-Server-side-Blazor-Cookie-Authentication.aspx)

Yen:
@ADefWebserver Thank you for your reference but I don't believe this is what I am looking for. I am sure it is possible to check for the authentication state by manually looking at the expiration dates of the cookies as they are given during the creation of the websocket instance. However this is not a complete solution, nor do I believe it is a response to this issue. In this case we are looking for a way to revalidate authentication on the existing websocket connection, without the requirement to validate each function call made from the blazor component.

SteveSandersonMS:
@Yen You're correct that the built-in system only addresses the scenario for identity-based authentication. It periodically checks whether the user's security stamp has changed, and if so, invalidates their current authentication state (which in turn causes authorization rules to be re-evaluated).

[...]

PR #11548, opened 2019-06-25
Add authentication revalidation logic to IndividualLocalAuth server-side Blazor template. Implements #10698

SteveSandersonMS:
Implements #10698

Following discussions with security and SignalR folks, we've decided we need to have some form of periodic revalidation for authentication in server-side Blazor. This is meaningful only in the Identity case, because that's what has the concept of security stamps and can re-query the DB to get updated state.

**Before broader code review, I'd first like to check with authentication/Identity experts that the logic here looks right.** Ping @HaoK. And @blowdart, is there anyone else you'd like to look at this?

The specific bit that authentication/Identity experts might most be interested in is the contents of `CheckIfAuthenticationStateIsValidAsync`. This gets called periodically (every 30 minutes by default) and is responsible for deciding whether to sign the user out. The rule is very simple: if your security stamp changed, we log you out. Then the user is forced to go through the login flow again, which refreshes everything about their `ClaimsPrincipal` (if they are still even able to log in).

So, @HaoK, can you let me know if you think the logic in `CheckIfAuthenticationStateIsValidAsync` accounts for everything necessary? For example, if a user got locked out or had their roles changed, then their security stamp will have changed, right?

If you need any further information, I'd be happy to have a brief meeting to explain any finer points of the mechanism. Let me know.

### DIFF
--- a/src/ProjectTemplates/Web.ProjectTemplates/content/RazorComponentsWeb-CSharp/Areas/Identity/RevalidatingAuthenticationStateProvider.cs
+++ b/src/ProjectTemplates/Web.ProjectTemplates/content/RazorComponentsWeb-CSharp/Areas/Identity/RevalidatingAuthenticationStateProvider.cs
@@ -1,0 +1,98 @@
+using Microsoft.AspNetCore.Components;
+using Microsoft.AspNetCore.Identity;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using System;
+using System.Security.Claims;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace RazorComponentsWeb_CSharp.Areas.Identity
+{
+    /// <summary>
+    /// An <see cref="AuthenticationStateProvider"/> service that revalidates the
+    /// authentication state at regular intervals. If a signed-in user's security
+    /// stamp changes, this revalidation mechanism will sign the user out.
+    /// </summary>
+    /// <typeparam name="TUser">The type encapsulating a user.</typeparam>
+    public class RevalidatingAuthenticationStateProvider<TUser>
+        : AuthenticationStateProvider, IDisposable where TUser: class
+    {
+        private readonly static TimeSpan RevalidationInterval = TimeSpan.FromMinutes(30);
+
+        private readonly CancellationTokenSource _loopCancellationTokenSource = new CancellationTokenSource();
+        private readonly IServiceScopeFactory _scopeFactory;
+        private readonly ILogger _logger;
+        private Task<AuthenticationState> _currentAuthenticationStateTask;
+
+        public RevalidatingAuthenticationStateProvider(
+            IServiceScopeFactory scopeFactory,
+            SignInManager<TUser> circuitScopeSignInManager,
+            ILogger<RevalidatingAuthenticationStateProvider<TUser>> logger)
+        {
+            var initialUser = circuitScopeSignInManager.Context.User;
+            _currentAuthenticationStateTask = Task.FromResult(new AuthenticationState(initialUser));
+            _scopeFactory = scopeFactory;
+            _logger = logger;
+
+            if (initialUser.Identity.IsAuthenticated)
+            {
+                _ = RevalidationLoop();
+            }
+        }
+
+        public override Task<AuthenticationState> GetAuthenticationStateAsync()
+            => _currentAuthenticationStateTask;
+
+        private async Task RevalidationLoop()
+        {
+            var cancellationToken = _loopCancellationTokenSource.Token;
+
+            while (!cancellationToken.IsCancellationRequested)
+            {
+                try
+                {
+                    await Task.Delay(RevalidationInterval, cancellationToken);
+                }
+                catch (TaskCanceledException)
+                {
+                    break;
+                }
+
+                var isValid = await CheckIfAuthenticationStateIsValidAsync();
+                if (!isValid)
+                {
+                    // Force sign-out. Also stop the revalidation loop, because the user can
+                    // only sign back in by starting a new connection.
+                    var anonymousUser = new ClaimsPrincipal();
+                    _currentAuthenticationStateTask = Task.FromResult(new AuthenticationState(anonymousUser));
+                    NotifyAuthenticationStateChanged(_currentAuthenticationStateTask);
+                    _loopCancellationTokenSource.Cancel();
+                }
+            }
+        }
+
+        private async Task<bool> CheckIfAuthenticationStateIsValidAsync()
+        {
+            try
+            {
+                // Get the sign-in manager from a new scope to ensure it fetches fresh data
+                using (var scope = _scopeFactory.CreateScope())
+                {
+                    var signInManager = scope.ServiceProvider.GetRequiredService<SignInManager<TUser>>();
+                    var authenticationState = await _currentAuthenticationStateTask;
+                    var validatedUser = await signInManager.ValidateSecurityStampAsync(authenticationState.User);
+                    return validatedUser != null;
+                }
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(ex, "An error occurred while revalidating authentication state");
+                return false;
+            }
+        }
+
+        void IDisposable.Dispose()
+            => _loopCancellationTokenSource.Cancel();
+    }
+}
--- a/src/ProjectTemplates/Web.ProjectTemplates/content/RazorComponentsWeb-CSharp/Areas/Identity/RevalidatingAuthenticationStateProvider.cs
+++ b/src/ProjectTemplates/Web.ProjectTemplates/content/RazorComponentsWeb-CSharp/Areas/Identity/RevalidatingAuthenticationStateProvider.cs
@@ -1,11 +1,11 @@
-using Microsoft.AspNetCore.Components;
-using Microsoft.AspNetCore.Identity;
-using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.Logging;
 using System;
 using System.Security.Claims;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Components;
+using Microsoft.AspNetCore.Identity;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
 
 namespace RazorComponentsWeb_CSharp.Areas.Identity
 {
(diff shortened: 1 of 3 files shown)


HaoK:
Looks good to me, the security stamp is exactly that, its meant to be used to sign out everywhere, that's what cookies does, and by checking it directly, you are getting all the same invalidations that would cause cookies to be invalidated.  Lockouts don't cause security stamps to roll I don't believe, but other than that yes.

SteveSandersonMS:
Thanks for the feedback, @HaoK!

> Lockouts don't cause security stamps to roll I don't believe, but other than that yes.

That means the logic here (calling `ValidateSecurityStampAsync`) is insufficient then, doesn't it? What's the correct way to know if a given identity user is current locked out?

Is there anything else we should check besides this?

HaoK:
I don't think you should deal with lockouts, just like lockouts don't invalidate all cookies (that would be a denial of service attack since you could just lock out a user and that would prevent all existing logged in users from ever accessing the site as well).

The normal login flow should fail if the user is locked out, but if the security stamp hasn't changed, that isn't something you should worry about.

@blowdart can confirm my understanding is correct

SteveSandersonMS:
Sorry Hao, I don't understand your points.

> you could just lock out a user and that would prevent all existing logged in users from ever accessing the site as well

Why would locking out one user affect other users?

> The normal login flow should fail if the user is locked out, but if the security stamp hasn't changed, that isn't something you should worry about.

The normal login flow isn't applicable - our goal is for the long-running Blazor circuit state to get updated periodically. If a user is marked as locked out, we'd expect them to stop being allowed to use the app (at least as of the next time we revalidate them). So I think we *do* need to account for this. If you still think I'm wrong please let me know and perhaps we could chat verbally.

Besides lock-out, is there anything else that would be a basis for thinking a user should no longer be allowed to interact with the app? I'm guessing no, but just checking for completeness.

HaoK:
What I meant was, lets say I know your email, but I don't know your password, if I just spam bad login attempts with your email which resulted in login attempts being blocked for a certain amount of time, today that doesn't automatically cause existing cookies to be invalidated.  I'm not clear exactly what the analogy is for blazor/circuits.  But I'd imagine most people are permanently logged into their email accounts, someone being able to kick you out by attempting bad logins seems bad.  Its only when you go and change your password, or update your authenticator key, or any other security sensitive action that causes the security stamp to change, that should result in your current sessions to become invalid.

That's what I was trying to explain, does that make sense?

SteveSandersonMS:
OK, thanks for clarifying. I was guessing that the purpose of the "lock out" feature was something more like "site owner decides they don't like a given user and wants to ban them" or "this employee just got fired, so make sure they can't access company data".

From your comment I guess the real purpose is rate-limiting password guess attempts, in which case I agree it should not be something that triggers automatic logout for existing authenticated sessions.

Based on this I will regard the existing logic in this PR as correct. Thanks for the info!

SteveSandersonMS:
@rynowak @javiercn:

Since @HaoK has now taken a look and we think the logic here matches identity's requirements, this is now ready for actual review. Also cc @blowdart in case he wants to see.

I know the build is failing due to me not updating template baselines yet.

HaoK:
Right, so if you want to revoke all existing sign ins like that permanently, you can disable/lockout the user and then call `UpdateSecurityStamp` which forces sign out everywhere

SteveSandersonMS:
Ping @rynowak @javiercn for review here. If either of you don't feel you're the right person to review this, do you have someone else in mind? I've already had @HaoK look at it.

javiercn:
@SteveSandersonMS Sorry for the delay, I've been busy putting out other fires. I'll try and take a look at it today.

src/ProjectTemplates/Web.ProjectTemplates/content/RazorComponentsWeb-CSharp/Areas/Identity/RevalidatingAuthenticationStateProvider.cs
```diff
@@ -0,0 +1,98 @@
+using Microsoft.AspNetCore.Components;
+using Microsoft.AspNetCore.Identity;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using System;
+using System.Security.Claims;
+using System.Threading;
+using System.Threading.Tasks;
```
rynowak: What sort order do the other template files use for these? We use "System first" for our code which is not the VS default.

src/ProjectTemplates/Web.ProjectTemplates/content/RazorComponentsWeb-CSharp/Areas/Identity/RevalidatingAuthenticationStateProvider.cs
```diff
@@ -0,0 +1,98 @@
...
+
+namespace RazorComponentsWeb_CSharp.Areas.Identity
+{
+    /// <summary>
+    /// An <see cref="AuthenticationStateProvider"/> service that revalidates the
+    /// authentication state at regular intervals. If a signed-in user's security
+    /// stamp changes, this revalidation mechanism will sign the user out.
+    /// </summary>
+    /// <typeparam name="TUser">The type encapsulating a user.</typeparam>
```
rynowak: Should we have a comment here that says something like: "this is an important security feature, do not modify it unless you have a good reason"?

src/ProjectTemplates/Web.ProjectTemplates/content/RazorComponentsWeb-CSharp/Areas/Identity/RevalidatingAuthenticationStateProvider.cs
```diff
@@ -0,0 +1,98 @@
...
+    public class RevalidatingAuthenticationStateProvider<TUser>
+        : AuthenticationStateProvider, IDisposable where TUser: class
```
rynowak: `TUser: class` -> `TUser : class`

src/ProjectTemplates/Web.ProjectTemplates/content/RazorComponentsWeb-CSharp/Areas/Identity/RevalidatingAuthenticationStateProvider.cs
```diff
@@ -0,0 +1,98 @@
...
+                }
+            }
+        }
+
+        private async Task<bool> CheckIfAuthenticationStateIsValidAsync()
+        {
+            try
+            {
+                // Get the sign-in manager from a new scope to ensure it fetches fresh data
+                using (var scope = _scopeFactory.CreateScope())
```
rynowak: It looks like there's a race condition here that I'm not sure how to mitigate. The `_scopeFactory` could be disposed in any order with `Dispose` being called on this class. (I'm assuming that `_scopeFactory`) will become electrified when the app services are disposed. 

If this happens the `ObjectDisposedException` is going be handled by the exception handler at L88, so we're probably fine. 

Just thought I'd mention.

src/ProjectTemplates/Web.ProjectTemplates/content/RazorComponentsWeb-CSharp/Areas/Identity/RevalidatingAuthenticationStateProvider.cs
```diff
@@ -0,0 +1,98 @@
...
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(ex, "An error occurred while revalidating authentication state");
+                return false;
+            }
+        }
+
+        void IDisposable.Dispose()
+            => _loopCancellationTokenSource.Cancel();
```
rynowak: I think it's safe to call `.Cancel()` multiple times, but I wanted to mention because it will definitely happen. If we see an error or a change in security stamp we cancel the token, then the user disconnects, then the circuit is disposed and calls cancel again.

src/ProjectTemplates/Web.ProjectTemplates/content/RazorComponentsWeb-CSharp/Areas/Identity/RevalidatingAuthenticationStateProvider.cs
SteveSandersonMS: Good point. It is "system first" elsewhere in the template. Fixed this.

src/ProjectTemplates/Web.ProjectTemplates/content/RazorComponentsWeb-CSharp/Areas/Identity/RevalidatingAuthenticationStateProvider.cs
SteveSandersonMS: I don't know, that seems pretty strange. Certain modifications would be totally reasonable (e.g., changing `RevalidationInterval` to a shorter or longer duration).

If you feel strongly, and want to suggest a particular phrasing, please let me know. Otherwise it seems more normal in our templates not to add scare-comments. For example we don't have similar comments about `app.UseHsts` or `app.UseHttpsRedirection`.

src/ProjectTemplates/Web.ProjectTemplates/content/RazorComponentsWeb-CSharp/Areas/Identity/RevalidatingAuthenticationStateProvider.cs
SteveSandersonMS: Done

src/ProjectTemplates/Web.ProjectTemplates/content/RazorComponentsWeb-CSharp/Areas/Identity/RevalidatingAuthenticationStateProvider.cs
SteveSandersonMS: You're right, but I also don't see an clear way of improving on this. We don't have a way to know for sure whether `_scopeFactory` was already disposed, or even if we did we couldn't lock it to prevent it being disposed in between that check and the usage of it or any services it supplies.

Like you've pointed out, the worst case is that it will log the exception, but the behavior will still be correct. Given this cycle only occurs every 30 mins by default, this should be extremely rare, if app developers ever see it at all.

If this becomes an issue in the future, I suspect we might need to introduce some event like "circuit will shut down soon" that things like this can listen to and proactively stop themselves.

src/ProjectTemplates/Web.ProjectTemplates/content/RazorComponentsWeb-CSharp/Areas/Identity/RevalidatingAuthenticationStateProvider.cs
SteveSandersonMS: My research also agrees with the conclusion that you can cancel multiple times. I think it's preferable to keep the code here simpler by leaving this as-is instead of introducing some locking or whatever we'd do to guarantee single cancel.

rynowak (approved):

Issue #9466, opened 2019-04-17
Provide an API to configure the TempDirectory used by ASP.NET Core


pranavkm:
ASP.NET Core currently uses an environment variable to drive the temp directory, falling back `Path.GetTempPath()`. See https://github.com/aspnet/AspNetCore/blob/56e2080ef7741dea9a5c24a8e0639b0921be2d22/src/Http/Http/src/Internal/BufferingHelper.cs#L18-L39. 

This path is currently in use by `FileBufferingReadStream` and as of https://github.com/aspnet/AspNetCore/pull/9015 change, by `FileBufferingWriteStream`. 

This is one of the few parts of the stack that is only configurable via environment variable. We should instead follow the regular pattern of having an option type (that can be configured via configuration) and use those in when constructing these streams and everywhere else ASP.NET Core may need to write to the temp directory.

pranavkm:
See https://github.com/aspnet/AspNetCore/pull/9015#discussion_r271938569 for disussions. cc @Tratcher

davidfowl:
Tricky as we don’t have a good way to flow this to the places that need this.

Tratcher:
Yes we do. https://github.com/aspnet/AspNetCore/pull/9015#discussion_r275858444

davidfowl:
That doesn't look like a solution. Doesn't that say pass an argument? Isn't this about defaults? What am I missing?

Tratcher:
The options would be consumed by the thing creating the buffer streams such as EnableRewind or MVC's output formatters.
https://github.com/aspnet/AspNetCore/blob/425c196cba530b161b120a57af8f1dd513b96f67/src/Http/Http/src/Internal/BufferingHelper.cs#L41-L51
There was also a proposal to bundle that configuration and stream creation into a service/factory so the individual call sites didn't need to deal with it.

anurse:
This seems reasonable but we don't have resources to address this in 3.0. The basic idea would be to resolve `IOptions<HttpBufferingOptions>` (where `HttpBufferingOptions` is a new type) in the `EnableRewind` method and use it to read this path instead of a global static.

This would be breaking since we'd be deleting (or even just not using) the existing static.

We could take a PR. It's not super complicated. But it will have to land in a major version because it's breaking.

Tratcher:
[...]

PR #11569, opened 2019-06-25
[Addresses #9466] API to configure TempDirectory used by HttpBuffering

cjaliaga:
Addresses #9466

**This is still not ready but before implementing it in all the callers I'd like to check if the logic looks right.** 

Currently the `HttpBufferingOptions` is the one configured in the DI, and `IOptions<HttpBufferingOptions>` resolved in the callers, such `EnableRewind` or the MVC formatters. 

For the creation of the Streams, a new Factory is added `(HttpFileBufferingFactory`) in `Microsoft.AspNetCore.Http` which gets in the ctor the options. This factory could be resolved instead of being manually created, in that case, what should be the right place to be registered into the Service Collection?

Another option would be making the factory more generic, just getting the `TempDirectoryPath` as string in the ctor instead of being tied to the `HttpBufferingOptions,` so it could be added to the `WebUtilities project`.

@Tratcher, @anurse could you please take a look and let me know what you think?


src/Http/Http/src/IFileBufferingStreamFactory.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.IO;
+using Microsoft.AspNetCore.WebUtilities;
+
+namespace Microsoft.AspNetCore.Http
+{
+    public interface IFileBufferingStreamFactory
```
Tratcher: This would likely go in the Http.Abstractions package.

src/Http/Http/src/IFileBufferingStreamFactory.cs
```diff
@@ -0,0 +1,15 @@
...
+    {
+        FileBufferingReadStream CreateReadStream(Stream inner, int bufferThreshold, long? bufferLimit = null);
```
Tratcher: Why can't this return Stream?

src/Http/Http/src/Features/FormFeature.cs
```diff
@@ -182,9 +184,11 @@ private async Task<IFormCollection> InnerReadFormAsync(CancellationToken cancell
                             var fileSection = new FileMultipartSection(section, contentDisposition);
 
                             // Enable buffering for the file if not already done for the full body
+                            var httpBufferingOptions = _request.HttpContext.RequestServices.GetService<IOptions<HttpBufferingOptions>>();
```
Tratcher:
```suggestion
                            var httpBufferingOptions = _request.HttpContext.RequestServices.GetRequiredService<IOptions<HttpBufferingOptions>>();
```

src/Http/Http/src/Internal/BufferingHelper.cs
```diff
@@ -21,15 +22,17 @@ public static HttpRequest EnableRewind(this HttpRequest request, int bufferThres
             var body = request.Body;
             if (!body.CanSeek)
             {
-                var fileStream = new FileBufferingReadStream(body, bufferThreshold, bufferLimit, AspNetCoreTempDirectory.TempDirectoryFactory);
+                var httpBufferingOptions = request.HttpContext.RequestServices.GetRequiredService<IOptions<HttpBufferingOptions>>();
+                var factory = new HttpFileBufferingStreamFactory(httpBufferingOptions);
```
Tratcher: Why isn't the factory being resolved from DI? Yes something would need to register it, probably Hosting with the rest of the default services.
https://github.com/aspnet/AspNetCore/blob/70b478136a702d2b6061ac9b159029189ad230d0/src/Hosting/Hosting/src/GenericHost/GenericWebHostBuilder.cs#L84
https://github.com/aspnet/AspNetCore/blob/70b478136a702d2b6061ac9b159029189ad230d0/src/Hosting/Hosting/src/WebHostBuilder.cs#L284

src/Http/Http/test/Features/FormFeatureTests.cs
```diff
@@ -625,5 +627,16 @@ private void CompareStreams(Stream streamA, Stream streamB)
                 loops++;
             }
         }
+
+        private static DefaultHttpContext GetDefaultHttpContext()
+        {
+            var httpContext = new DefaultHttpContext
+            {
+                RequestServices = new ServiceCollection()
+                .AddSingleton(Options.Create(new HttpBufferingOptions()))
```
Tratcher: We don't register individual options types, AddOptions should be sufficient, it will create the default HttpBufferingOptions.
https://github.com/aspnet/AspNetCore/blob/70b478136a702d2b6061ac9b159029189ad230d0/src/Hosting/Hosting/src/WebHostBuilder.cs#L286

src/Http/Http/src/HttpBufferingOptions.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.IO;
+
+namespace Microsoft.AspNetCore.Http
+{
+    public class HttpBufferingOptions
+    {
+        public string TempFileDirectory { get; set; } = Path.GetTempPath();
```
Tratcher: How do you intend to account for `ASPNETCORE_TEMP`?

One pattern you could use is to leave TempFileDirectory null here and use PostConfigure to set a default if nobody else does. PostConfigure could also read `ASPNETCORE_TEMP` from config. See https://github.com/aspnet/AspNetCore/blob/master/src/Security/Authentication/Cookies/src/PostConfigureCookieAuthenticationOptions.cs

src/Http/Http/src/IFileBufferingStreamFactory.cs
cjaliaga: The consumers are actually expecting a Stream, so it can return it. The only reason was because the consumers of the writer need the  `FileBufferingWriteStream` explicitly so I kept this one same way. Moving this to return `Stream` sounds better.

https://github.com/aspnet/AspNetCore/blob/6152fc37b7ab4d61ac17f4aa1c5109fd5ee8b2e8/src/Mvc/Mvc.Formatters.Xml/src/XmlDataContractSerializerOutputFormatter.cs#L291

Tratcher: > The only reason was because the consumers of the writer need the FileBufferingWriteStream explicitly

I expect we could remove that requirement. DrainBufferAsync looks like a fancy CopyToAsync.

src/Http/Http/src/Internal/BufferingHelper.cs
cjaliaga: Great, I thought about register and resolve the factory but didn't know where to register it. Hosting sounds good, thanks :)

I can register together the IPostConfiguration for the options and the factory, probably with an Extension method, like AddHttpBuffering()?

The scope I believe it should be a singleton, out of curiosity, why the `DefaultHttpContextFactory` in the Generic is a Singleton and in the other one is Transient? Should I follow same pattern here?

Tratcher: > I can register together the IPostConfiguration for the options and the factory, probably with an Extension method, like AddHttpBuffering()?

Sounds good.

> The scope I believe it should be a singleton, out of curiosity, why the DefaultHttpContextFactory in the Generic is a Singleton and in the other one is Transient? Should I follow same pattern here?

I don't recall, but follow what we do in Generic, it's more recent.

src/Http/Http/src/Features/FormFeature.cs
cjaliaga: If we are registering the factory I think here we should resolve it and pass it to the EnableRewind for the `MultipartSection` as an overload. We can let the original method creating the `FileBufferingReadStream` manually in case someone was using it, since it was public. Or just delete it given that we are keeping, at least for now, the use of the Environment Variable via the `PostConfigure`.

src/Http/Http/src/HttpBufferingOptions.cs
cjaliaga: PostConfigure checking for ` Environment.GetEnvironmentVariable("ASPNETCORE_TEMP") ??                              Path.GetTempPath()` when TempFileDirectory is null looks very clean. Will go this way 👍

src/Http/Http/src/IFileBufferingStreamFactory.cs
cjaliaga: > I expect we could remove that requirement. DrainBufferAsync looks like a fancy CopyToAsync.

By default `CopyTo `throws `NotSupportedException` when a stream cannot be read, like in this case. Could it be confusing if we allow to copy this stream but reading it is not allowed? I don't know, however, if this behavior is usually honored.

Tratcher: You're right, that came up in the design so we added DrainBufferAsync. I'd rather not have the concrete type exposed here. Maybe we can abstract it.
```
public interface IBufferedWriteStream
{
  Stream Buffer { get; }
  Task DrainBufferAsync(Stream destination);
}
```

cjaliaga: If I got it right, `FileBufferingWriteStream` would be implementing `IBufferedWriteStream` and returned from the factory instead of the concrete type.

Regarding of `Buffer` property, and also that we need the actual `Stream` in the consumers, what is the purpose of the property and how are we getting the `Stream`?

I have the feeling that property could be the answer since `IBufferedWriteStream` doesn't know it's a `Stream` (cannot extend the class), so TL;DR, could be the implementation this one and the way of getting it instead of casting on the consumers?

`Stream Buffer => this;`

src/Http/Http/src/IFileBufferingStreamFactory.cs
cjaliaga: I was thinking on putting `IBufferedWriteStream` on WebUtilities but in that case Http.Abstractions would have a dependency on it, is that OK?

src/Http/Http/src/IFileBufferingStreamFactory.cs
Tratcher: > `Stream Buffer => this;`

Exactly

src/Http/Http/src/IFileBufferingStreamFactory.cs
Tratcher: Hmm, tradeoffs. Let's not change the dependencies right now. That means IFileBufferingStreamFactory stays in Http and IBufferedWriteStream in WebUtilities.

src/Http/WebUtilities/src/IBufferedWriteStream.cs
```diff
@@ -0,0 +1,16 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNetCore.WebUtilities
+{
+    public interface IBufferedWriteStream : IAsyncDisposable, IDisposable
```
cjaliaga: Marking this as `IDisposable` helps in situations like this one:

https://github.com/aspnet/AspNetCore/blob/d6c629a005a859ca653cbb2ec96d4b6c385584cf/src/Mvc/Mvc.ViewFeatures/src/ViewComponentResultExecutor.cs#L157

And it also would simplify this:

https://github.com/aspnet/AspNetCore/blob/d6c629a005a859ca653cbb2ec96d4b6c385584cf/src/Mvc/Mvc.NewtonsoftJson/src/NewtonsoftJsonOutputFormatter.cs#L169

What do you think?

Tratcher: Looks good.

davidfowl:
I haven't been following this closely but I'm not a fan of adding more services to hosting to enable this to work. I strongly recommend looking for an alternative design. We've been slowly moving away from adding these "default services" to hosting (see https://github.com/aspnet/AspNetCore/issues/5944).

cjaliaga:
Would it make sense to add the required services with `AddHttpBuffering()` in the places that is used? Currently in Mvc:
-  Mvc.Formatters.Xml\src\XmlDataContractSerializerOutputFormatter.cs
-  Mvc.Formatters.Xml\src\XmlSerializerOutputFormatter.cs
-  Mvc.NewtonsoftJson\src\NewtonsoftJsonOutputFormatter.cs
-  Mvc.NewtonsoftJson\src\NewtonsoftJsonResultExecutor.cs
-  Mvc.ViewFeatures\src\ViewComponentResultExecutor.cs

For these ones, `Mvc.Core` or the individual component? 

It's also used in Http\src\Internal\BufferingHelper.cs, called from from FormFeature:

https://github.com/aspnet/AspNetCore/blob/a4a08a68f3a01b6f89a357de2cef67b6b823b348/src/Http/Http/src/Features/FormFeature.cs#L140

https://github.com/aspnet/AspNetCore/blob/a4a08a68f3a01b6f89a357de2cef67b6b823b348/src/Http/Http/src/Features/FormFeature.cs#L187-L191

but exposed publicly via `HttpRequest` extension methods in 

https://github.com/aspnet/AspNetCore/blob/a4a08a68f3a01b6f89a357de2cef67b6b823b348/src/Http/Http/src/Extensions/HttpRequestRewindExtensions.cs#L21-L24

So, in these cases, do you know of another place more specific than hosting that could work? Having a fallback in case is not registered (not sure if this would be clean)?

Tratcher:
No, features exposed directly from HttpContext need to work with pre-registered services. Adding it by default in Hosting should be adequate.

Tratcher:
@davidfowl I'll follow-up with you directly.

Tratcher:
Offline discussion:
- Options and services for things used by HttpContext are problematic.
- A temp dir is a more fundamental app level concern, we should move this to IHostEnvironment
https://github.com/aspnet/Extensions/blob/master/src/Hosting/Abstractions/src/IHostEnvironment.cs
or IWebHostEnvironment
https://github.com/aspnet/AspNetCore/blob/master/src/Hosting/Abstractions/src/IWebHostEnvironment.cs
then it would be configured through the host APIs, not via options and services.

@glennc to check if IHostEnvironment vs IWebHostEnvironment makes the most sense.

cjaliaga:
Thanks @Tratcher, I like it. This way it's not just the temp dir for HttpBuffering but a way for setting/getting the temp dir for everyone.

The consumers still need to create the Buffering Read and Write Stream, in this case if we use the host APIs to configure the TempDirectory, we check if the Directory exists on configuration and everything there, could we resolve the HostEnvironment and do this for the accessor?

`new FileBufferingWriteStream(tempFileDirectoryAccessor: () => _environment.TempDirectory);`

Or is still better and safer to pass a Func like this in case there's something in the middle breaking it?
https://github.com/aspnet/AspNetCore/blob/a4a08a68f3a01b6f89a357de2cef67b6b823b348/src/Http/Http/src/HttpFileBufferingStreamFactory.cs#L47-L52

Regarding the default path, are we still keeping the default behavior in the Streams for getting the accessor or deprecating/marking obsolete (this is the breaking-change itself)?

https://github.com/aspnet/AspNetCore/blob/a4a08a68f3a01b6f89a357de2cef67b6b823b348/src/Http/WebUtilities/src/FileBufferingWriteStream.cs#L58

Tratcher:
> Or is still better and safer to pass a Func like this in case there's something in the middle breaking it?

Yeah, reading and validating TempFileDirectory should be lazy because it may not be created until after startup, and could even be deleted later.

> Regarding the default path, are we still keeping the default behavior in the Streams for getting the accessor or deprecating/marking obsolete (this is the breaking-change itself)?

Yes, we'll keep any back-compat that's reasonably possible, but may decided to Obsolete(Warning) some of them.

cjaliaga:
> Yeah, reading and validating TempFileDirectory should be lazy because it may not be created until after startup, and could even be deleted later.

Makes sense. We could keep the factory then but just in WebUtilities this way:

```csharp
public class FileBufferingStreamFactory : IFileBufferingStreamFactory
{
     public FileBufferingStreamFactory(string tempPath)
     {
         _tempPath = tempPath;
         _getTempDirectory = () => _tempDirectory;
     }

     private string _tempDirectory
     {
          get
           {
               if (!Directory.Exists(_tempPath))
               {
                   throw new DirectoryNotFoundException(_tempPath);
               }

               return _tempPath;
           }
       }
}
```
And create it when needed resolving `HostEnvironment` to get TempDirectory. Or just keep creating the BufferingStreams in the consumers manually.

> @glennc to check if IHostEnvironment vs IWebHostEnvironment makes the most sense.

Glenn, could you please take a look to see in which one it suits better? Specially as the former would make it cross-repo.

cjaliaga:
@Tratcher I've started with this on a new branch (https://github.com/cjaliaga/AspNetCore/commit/8b9625a3585f8a7c19ae541fedd9db9fd6207dcf), adding it on `IWebHostEnvironment` for the time being.

Is it better to create a new PR with the new branch linking this one or to reset this one to master and do it here to keep the PR active?

anurse:
This got backburnered during the run up to 3.0. We'll try to look at this for 5.0 now!

Tratcher:
Let's close this in favor of #12555

PR #11576, opened 2019-06-26
Rename @aspnet/signalr to @microsoft/signalr

### LINKED ISSUE #11568: Rename '@aspnet/signalr' and '@aspnet/signalr-protocol-msgpack' to '@microsoft/...'
We have [the name reserved on NPM](https://www.npmjs.com/package/@microsoft/signalr) and want to do this rename to help clarify that the SignalR client can be used in scenarios with the Azure SignalR Service where there is no ASP.NET :). We need to rename our `package.json` files and update dependencies to do this.

@pranavkm @rynowak This shouldn't affect Blazor as you embed the client, right? We'll have to update the build logic to make sure it works (we'll do the needful) but there shouldn't be a runtime impact from my understanding

fyi @DamianEdwards @bradygaster @davidfowl

mikaelm12:
Fixes: https://github.com/aspnet/AspNetCore/issues/11568
Replacing all instances of `@aspnet` with `@microsoft`. Are there any other tricks here?
Follow up items are to update docs and publish preview 7 under the `@aspnet/signalr` name with a deprecated message

### DIFF
--- a/src/SignalR/clients/ts/FunctionalTests/web.config
+++ b/src/SignalR/clients/ts/FunctionalTests/web.config
@@ -7,8 +7,11 @@
     <handlers>
       <add name="aspNetCore" path="*" verb="*" modules="AspNetCoreModuleV2" resourceType="Unspecified" />
     </handlers>
-    <aspNetCore processPath="%LAUNCHER_PATH%" arguments="%LAUNCHER_ARGS%" stdoutLogEnabled="false" stdoutLogFile=".\logs\stdout" forwardWindowsAuthToken="false" startupTimeLimit="3600" requestTimeout="23:00:00">
-      <environmentVariables />
+    <aspNetCore processPath="%LAUNCHER_PATH%" arguments="%LAUNCHER_ARGS%" stdoutLogEnabled="false" stdoutLogFile=".\logs\stdout" forwardWindowsAuthToken="false" startupTimeLimit="3600" requestTimeout="23:00:00" hostingModel="InProcess">
+      <environmentVariables>
+        <environmentVariable name="COMPLUS_ForceENC" value="1" />
+        <environmentVariable name="ASPNETCORE_ENVIRONMENT" value="Development" />
+      </environmentVariables>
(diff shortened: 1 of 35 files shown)


anurse:
Don't see any blazor-related changes, I presume they weren't using the ID anywhere, just the files?

Also, speaking of docs, we should sync up on the JS API docs. They'll probably need to be updated.

analogrelay (approved):
Looks right to me, as long as the build passes and things are working, I'm good with this.

src/SignalR/clients/ts/FunctionalTests/web.config
```diff
@@ -7,8 +7,11 @@
     <handlers>
       <add name="aspNetCore" path="*" verb="*" modules="AspNetCoreModuleV2" resourceType="Unspecified" />
     </handlers>
-    <aspNetCore processPath="%LAUNCHER_PATH%" arguments="%LAUNCHER_ARGS%" stdoutLogEnabled="false" stdoutLogFile=".\logs\stdout" forwardWindowsAuthToken="false" startupTimeLimit="3600" requestTimeout="23:00:00">
```
mikaelm12: I'll revert this

mikaelm12:
> Looks right to me, as long as the build passes and things are working, 

Just need to do the needful for the Code check and everything should be green

src/SignalR/clients/ts/FunctionalTests/web.config
BrennanConroy: Forget?

mikaelm12: I was making the change as you were typing this lol

BrennanConroy (approved):

benaadams:
@anurse  This needs an announcement?

anurse:
Yep, it will have one!

anurse:
https://github.com/aspnet/Announcements/issues/372

mikaelm12:
Thanks @anurse!

PR #11580, opened 2019-06-26
Cleanup of components based on CaptureUnmatchedAttributes

### LINKED ISSUE #10713: Follow ups from add unmatched parameters support to our built-in components
While doing https://github.com/aspnet/AspNetCore/issues/5071 - I left two things untouched that seem like they need some discussion.

----

Should we remove the `Id` and `Class` parameters from `EditForm` and `InputBase<>`? 

Making these attributes *first class* makes it obvious that you can set them, and these components have some special handling for `class`. We need to make a design decision here on what this experience should be like.

----

Should we rewrite `NavLink` to use this new feature?

NavLink already *functionally* has this behaviour. There's a big comment in NavLink about how it was intentionally written at a low level of abstraction that scared me away from making this change. We could do this easily and it would simplify the codes, but I want to make sure I'm not throwing away something valuable.

rynowak:
Fixes: #10713

See commits for deeeets

### DIFF
--- a/src/Components/Components/src/Routing/NavLink.cs
+++ b/src/Components/Components/src/Routing/NavLink.cs
@@ -3,35 +3,21 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Components.RenderTree;
 
 namespace Microsoft.AspNetCore.Components.Routing
 {
-    // NOTE: This could be implemented in a more performant way by iterating through
-    // the ParameterCollection only once (instead of multiple TryGetValue calls), and
-    // avoiding allocating a dictionary in the case where there are no additional params.
-    // However the intention here is to get a sense of what more high-level coding patterns
-    // will exist and what APIs are needed to support them. Later in the project when we
-    // have more examples of components implemented in pure C# (not Razor) we could change
-    // this one to the more low-level perf-sensitive implementation.
-
     /// <summary>
     /// A component that renders an anchor tag, automatically toggling its 'active'
     /// class based on whether its 'href' matches the current URI.
     /// </summary>
-    public class NavLink : IComponent, IDisposable
+    public class NavLink : ComponentBase, IDisposable
     {
         private const string DefaultActiveClass = "active";
 
-        private RenderHandle _renderHandle;
         private bool _isActive;
-
-        private RenderFragment _childContent;
-        private string _cssClass;
         private string _hrefAbsolute;
-        private IReadOnlyDictionary<string, object> _allAttributes;
 
         /// <summary>
         /// Gets or sets the CSS class name applied to the NavLink when the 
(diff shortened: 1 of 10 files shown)


src/Components/Components/src/Routing/NavLink.cs
```diff
@@ -49,29 +53,38 @@ public class NavLink : IComponent, IDisposable
...
-            parameters.TryGetValue(RenderTreeBuilder.ChildContent, out _childContent);
-            parameters.TryGetValue("class", out _cssClass);
-            parameters.TryGetValue("href", out string href);
-            ActiveClass = parameters.GetValueOrDefault(nameof(ActiveClass), DefaultActiveClass);
-            Match = parameters.GetValueOrDefault(nameof(Match), NavLinkMatch.Prefix);
-            _allAttributes = parameters.ToDictionary();
-
-            // Update computed state and render
+            // Run the default parameter setting logic to populate our properties.
+            base.SetParametersAsync(parameters);
```
campersau: Why no await here or return the resulting task?

src/Components/Components/src/Routing/NavLink.cs
```diff
@@ -49,29 +53,38 @@ public class NavLink : IComponent, IDisposable
...
+            if (AdditionalAttributes != null && AdditionalAttributes.TryGetValue("href", out var obj))
+            {
+                href = Convert.ToString(obj);
+            }
+
             _hrefAbsolute = href == null ? null : UriHelper.ToAbsoluteUri(href).AbsoluteUri;
             _isActive = ShouldMatch(UriHelper.GetAbsoluteUri());
-            _renderHandle.Render(Render);
+
+            ActiveClass ??= DefaultActiveClass;
```
SteveSandersonMS: Rather than writing to `ActiveClass`, could we combine them on some other local (e.g., `var effectiveActiveClass = ActiveClass ?? DefaultActiveClass`)?

Because, although we don't really expect people to be reading back the values of parameters on component instances, the API does allow them to do so, and then someone might start depending on this behavior which is more of an implementation detail.

src/Components/Components/src/Routing/NavLink.cs
SteveSandersonMS: Or to make this simpler, why not override `OnParametersSetAsync` instead of `SetParametersAsync`? The logic here doesn't use the `ParameterCollection` directly. Also then you wouldn't have to call any `base` method.

SteveSandersonMS (approved):
Minor suggestions, but looks great! Especially nice to see those simplifications on `NavLink`.

src/Components/Components/src/Routing/NavLink.cs
rynowak: overriding `OnParametersSet` now

src/Components/Components/src/Routing/NavLink.cs
rynowak: I took this suggestion and just folded this expression into the one place where it's used.

rynowak:
I'm putting this PR on hold until functional tests are working again.

rynowak:
Rebased and ran these tests locally.

PR #11581, opened 2019-06-26
Add support for passing CultureInfo to binders

rynowak:
This change allows the binding infrastructure to accept a CultureInfo
everywhere we create a binder. The next step will be to add support to
the binding directive attribute to pass a culture.


src/Components/Components/src/EventCallbackFactoryBinderExtensions.cs
```diff
@@ -67,7 +67,7 @@ private static bool ConvertToIntCore(object obj, out int value)
                 return false;
             }
 
-            if (!int.TryParse(text, NumberStyles.Integer, CultureInfo.CurrentCulture, out var converted))
+            if (!int.TryParse(text, NumberStyles.Number, culture ?? CultureInfo.CurrentCulture, out var converted))
```
rynowak: changed this to number so that it will accept a wider range of stuff like a thousands separator.

src/Components/Components/src/EventCallbackFactoryBinderExtensions.cs
```diff
@@ -346,14 +346,16 @@ private static bool ConvertToNullableDateTimeCore(object obj, out DateTime? valu
...
         /// <param name="existingValue"></param>
+        /// <param name="culture"></param>
         /// <returns></returns>
         public static EventCallback<UIChangeEventArgs> CreateBinder(
             this EventCallbackFactory factory,
             object receiver,
             Action<string> setter,
-            string existingValue)
+            string existingValue,
+            CultureInfo culture = null)
```
rynowak: I made these methods accept a culture even where it doesn't do anything (for consistency). These are targeted by codegen + overload resolution, so the error experience won't be good if you specify a culture, but we don't accept one.

src/Components/Components/ref/Microsoft.AspNetCore.Components.netstandard2.0.cs
```diff
@@ -148,23 +148,23 @@ public sealed partial class EventCallbackFactory
...
+        public static Microsoft.AspNetCore.Components.EventCallback<Microsoft.AspNetCore.Components.UIChangeEventArgs> CreateBinder(this Microsoft.AspNetCore.Components.EventCallbackFactory factory, object receiver, System.Action<float?> setter, float? existingValue, System.Globalization.CultureInfo culture = null) { throw null; }
+        public static Microsoft.AspNetCore.Components.EventCallback<Microsoft.AspNetCore.Components.UIChangeEventArgs> CreateBinder(this Microsoft.AspNetCore.Components.EventCallbackFactory factory, object receiver, System.Action<float> setter, float existingValue, System.Globalization.CultureInfo culture = null) { throw null; }
+        public static Microsoft.AspNetCore.Components.EventCallback<Microsoft.AspNetCore.Components.UIChangeEventArgs> CreateBinder(this Microsoft.AspNetCore.Components.EventCallbackFactory factory, object receiver, System.Action<string> setter, string existingValue, System.Globalization.CultureInfo culture = null) { throw null; }
```
SteveSandersonMS: Is this just for completeness or consistency or similar? I wouldn't have thought it makes sense to talk about a culture when the thing you're binding is a `string`.

SteveSandersonMS: Same with `bool`

SteveSandersonMS: Ah, I see you addressed this question below already.

SteveSandersonMS (approved):
Looks excellent!

src/Components/Components/ref/Microsoft.AspNetCore.Components.netstandard2.0.cs
rynowak: Yeah I'm not sure that's there's something 100% sensible to do. In this case I feel like the culture being ignored is better than a really hard to understand compiler error

PR #11582, opened 2019-06-26
Fix runtime targeting pack to contain full paths to dlls

### LINKED ISSUE #11577: ASP.NET Runtime pack manifest should have relative path in package to files
In the RuntimeList.xml of the ASP.NET runtime packs, the `Path` attribute currently only includes the file name:

```xml
<File Type="Managed" Path="Microsoft.AspNetCore.Antiforgery.dll" AssemblyName="Microsoft.AspNetCore.Antiforgery" PublicKeyToken="adb9793829ddae60" AssemblyVersion="3.0.0.0" FileVersion="3.0.19.32405" />
```

For the SDK to resolve these files, the `Path` needs to be the full path relative to the root of the package, for example:

```xml
<File Type="Managed" Path="runtimes/win-x86/lib/netcoreapp3.0/Microsoft.AspNetCore.Antiforgery.dll" AssemblyName="Microsoft.AspNetCore.Antiforgery" PublicKeyToken="adb9793829ddae60" AssemblyVersion="3.0.0.0" FileVersion="3.0.19.32405" />
```

Note that the relative path will be different for native assets too.

We didn't catch this because the SDK was looking for the wrong filename for the manifest: [RuntimeListList.xml](https://github.com/dotnet/sdk/blob/d72a57f27e8d8d5a3bf8376599346d325a6a2491/src/Tasks/Microsoft.NET.Build.Tasks/ResolveRuntimePackAssets.cs#L71) instead of RuntimeList.xml. (Oops 🙃).  We can't fix the typo in the SDK until the ASP.NET runtime packs are fixed, otherwise the SDK will look for the runtime pack assets in the root of the runtime pack and fail to find them.

jkotalik:
Fixes https://github.com/aspnet/AspNetCore/issues/11577

Looks like the TargetPath isn't set for us like it is here: https://github.com/dotnet/core-setup/blob/master/src/pkg/projects/dir.targets#L98. For now, I just passed in the managed and native roots into the CreateFrameworkListFile task and prepended them. I can also do the target paths approach, but it's about the same complexity and I went with what was simpler for now. 

Here is an example of the two types of output:
```xml
<FileList Name="ASP.NET Core 3.0" TargetFrameworkIdentifier=".NETCoreApp" TargetFrameworkVersion="3.0" FrameworkName="Microsoft.AspNetCore.App">
  <File Type="Managed" Path="runtimes/win-x64/lib/netcoreapp3.0/Microsoft.AspNetCore.Antiforgery.dll" AssemblyName="Microsoft.AspNetCore.Antiforgery" PublicKeyToken="adb9793829ddae60" AssemblyVersion="42.42.42.42" FileVersion="42.42.42.42424" />
  <File Type="Native" Path="runtimes/win-x64/native/aspnetcorev2_inprocess.dll" FileVersion="13.0.19177.0" />

```

### DIFF
--- a/eng/tools/RepoTasks/CreateFrameworkListFile.cs
+++ b/eng/tools/RepoTasks/CreateFrameworkListFile.cs
@@ -23,6 +23,10 @@ namespace RepoTasks
         [Required]
         public string TargetFile { get; set; }
 
+        public string ManagedRoot { get; set; } = "";
+
+        public string NativeRoot { get; set; } = "";
+
         /// <summary>
         /// Extra attributes to place on the root node.
         ///
@@ -46,7 +50,6 @@ namespace RepoTasks
                 {
                     Item = item,
                     Filename = Path.GetFileName(item.ItemSpec),
-                    TargetPath = item.GetMetadata("TargetPath"),
                     AssemblyName = FileUtilities.GetAssemblyName(item.ItemSpec),
                     FileVersion = FileUtilities.GetFileVersion(item.ItemSpec),
                     IsNative = item.GetMetadata("IsNativeImage") == "true",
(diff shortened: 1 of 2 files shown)


dougbu (approved):
Looks fine as long as it satisfies @dsplaisted and the `AssemblyVersion="42.42.42.42"` bit in the description is because you were building locally i.e. without real versions

dsplaisted (approved):

JunTaoLuo (approved):

PR #11583, opened 2019-06-26
Don't complete the connection pipe in Http2FrameWriter

### LINKED ISSUE #11560: Network errors when refreshing an app repeatedly
I created a blazorserverside app using the latest sdk from core-sdk

dotnet new blazorserverside
dotnet run
Open chrome
Press F5 repeatedly at some point a request will fail and the UI will change.

A8mjiYKM8d

davidfowl:
- This leads to trunated data in some cases. Instead just yield the middleware so we can be sure no more user code is running (Http1OutputProducer does this as well). There are still cases where a misbeaving application that doesn't properly await writes gets cut off but that will be fixed in the SteamPipeWriter itself.
- Updated tests

Fixes #11560


davidfowl:
This PR doesn't fix the issue. Closing for now.

PR #11587, opened 2019-06-26
Fix default value for ResponseType in doc comments

thomaslevesque:
The comment says that the default value for `ResponseType` is `CodeIdToken` (hybrid flow), but the property is actually set to `IdToken` (implicit flow)


Tratcher (approved):

Tratcher:
Thanks

PR #11596, opened 2019-06-26
Revert "Move SPA flows to use code (#11227)"

javiercn:
This reverts commit 2287abcc3e00bf0d1350402cdddd00485a639cf2.

Found during validation that this actually breaks the SPA templates authorization flow. For some reason the functional tests for the SPA templates pass, so we should investigate why they aren't running or validating things correctly.


ryanbrandenburg (approved):
Assuming you just ran `git revert 2287abc` :shipit:.

PR #11601, opened 2019-06-26
Revert the output pipe in the DuplexStreamPipeAdapter

### LINKED ISSUE #11560: Network errors when refreshing an app repeatedly

davidfowl:
- Don't complete the connection pipe in Http2FrameWriter 
- Replace the output pipe only
- Updated tests

Fixes #11560

### DIFF
--- a/src/Servers/Kestrel/Core/src/Internal/Http2/Http2FrameWriter.cs
+++ b/src/Servers/Kestrel/Core/src/Internal/Http2/Http2FrameWriter.cs
@@ -89,7 +89,6 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http2
 
                 _completed = true;
                 _connectionOutputFlowControl.Abort();
-                _outputWriter.Complete();
             }
         }
 
--- a/src/Servers/Kestrel/Core/src/Middleware/Internal/DuplexPipeStreamAdapter.cs
+++ b/src/Servers/Kestrel/Core/src/Middleware/Internal/DuplexPipeStreamAdapter.cs
@@ -15,6 +15,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
     /// <typeparam name="TStream"></typeparam>
     internal class DuplexPipeStreamAdapter<TStream> : DuplexPipeStream, IDuplexPipe where TStream : Stream
     {
+        private readonly Pipe _output;
         private Task _outputTask;
 
         public DuplexPipeStreamAdapter(IDuplexPipe duplexPipe, Func<Stream, TStream> createStream) :
@@ -36,18 +37,20 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
                                                 useSynchronizationContext: false);
 
             Input = PipeReader.Create(Stream, readerOptions);
-            Output = new Pipe(outputOptions);
+
+            // We're using a pipe here because the HTTP/2 stack in Kestrel currently makes assumptions
+            // about when it is ok to write to the PipeWriter. This should be reverted back to PipeWriter.Create once
+            // those patterns are fixed.
+            _output = new Pipe(outputOptions);
         }
 
         public ILogger Log { get; set; }
 
         public TStream Stream { get; }
 
-        private Pipe Output { get; }
-
         public PipeReader Input { get; }
-        
-        PipeWriter IDuplexPipe.Output
+
+        public PipeWriter Output
         {
             get
             {
@@ -56,14 +59,14 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
                     _outputTask = WriteOutputAsync();
                 }
 
-                return Output.Writer;
+                return _output.Writer;
             }
         }
 
         public override async ValueTask DisposeAsync()
         {
             Input.Complete();
-            Output.Writer.Complete();
+            _output.Writer.Complete();
 
             if (_outputTask != null)
             {
@@ -79,7 +82,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
             {
                 while (true)
                 {
-                    var result = await Output.Reader.ReadAsync();
+                    var result = await _output.Reader.ReadAsync();
                     var buffer = result.Buffer;
 
                     try
@@ -90,6 +93,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
                             {
                                 break;
                             }
+
                             await Stream.FlushAsync();
                         }
                         else if (buffer.IsSingleSegment)
@@ -106,7 +110,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
                     }
                     finally
                     {
-                        Output.Reader.AdvanceTo(buffer.End);
+                        _output.Reader.AdvanceTo(buffer.End);
                     }
                 }
             }
@@ -116,7 +120,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
             }
             finally
             {
-                Output.Reader.Complete();
+                _output.Reader.Complete();
             }
         }
     }
--- a/src/Servers/Kestrel/Core/src/Middleware/Internal/DuplexPipeStreamAdapter.cs
+++ b/src/Servers/Kestrel/Core/src/Middleware/Internal/DuplexPipeStreamAdapter.cs
@@ -39,7 +39,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
             Output = new Pipe(outputOptions);
         }
 
-        public ILogger Log { get; private set; }
+        public ILogger Log { get; set; }
 
         public TStream Stream { get; }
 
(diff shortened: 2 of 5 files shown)


src/Servers/Kestrel/Core/src/Middleware/Internal/DuplexPipeStreamAdapter.cs
```diff
@@ -14,36 +15,110 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
...
+                                                readerScheduler: PipeScheduler.Inline,
+                                                writerScheduler: PipeScheduler.Inline,
+                                                pauseWriterThreshold: 1,
+                                                resumeWriterThreshold: 1,
+                                                minimumSegmentSize: writerOptions.MinimumBufferSize,
+                                                useSynchronizationContext: false);
+
             Input = PipeReader.Create(Stream, readerOptions);
-            Output = PipeWriter.Create(Stream, writerOptions);
+            Output = new Pipe(outputOptions);
```
jkotalik: This probably needs a comment on why we are doing this.

src/Servers/Kestrel/Core/src/Middleware/Internal/DuplexPipeStreamAdapter.cs
```diff
@@ -14,36 +15,110 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
...
         }
 
+        public ILogger Log { get; set; }
+
         public TStream Stream { get; }
 
+        private Pipe Output { get; }
```
jkotalik: nit: why inconsistencies on private and public for input/output.

src/Servers/Kestrel/Core/src/Middleware/Internal/DuplexPipeStreamAdapter.cs
```diff
@@ -14,36 +15,110 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
...
+            try
+            {
+                while (true)
+                {
+                    var result = await Output.Reader.ReadAsync();
+                    var buffer = result.Buffer;
+
+                    try
+                    {
+                        if (buffer.IsEmpty)
```
jkotalik: I think you should check result.IsCompleted even if the buffer isn't empty.

davidfowl: This code isn't new, why make tweaks to it at this point?

davidfowl: It also doesn't make sense to check IsCompleted for non empty buffers. That can result in truncation.

jkotalik (approved):

src/Servers/Kestrel/Core/src/Middleware/Internal/DuplexPipeStreamAdapter.cs
```diff
@@ -80,39 +80,7 @@ private async Task WriteOutputAsync()
...
-                                await Stream.WriteAsync(memory);
-                            }
-                        }
-                    }
-                    finally
-                    {
-                        _output.Reader.AdvanceTo(buffer.End);
-                    }
-                }
+                await _output.Reader.CopyToAsync(Stream);
```
jkotalik: 🌶

src/Servers/Kestrel/Core/src/Middleware/Internal/DuplexPipeStreamAdapter.cs
```diff
@@ -14,36 +15,114 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
...
+                    }
+                    finally
+                    {
+                        _output.Reader.AdvanceTo(buffer.End);
+                    }
+                }
+            }
+            catch (Exception ex)
+            {
+                Log?.LogError(0, ex, $"{GetType().Name}.{nameof(WriteOutputAsync)}");
```
halter73: Given that both LibuvConnection and SocketConnection will not complete their PipeReaders with an error unless it was completely unexpected, it's tempting to make this a critical log. That's unless we're worried about something like SslStream throwing an Exception in the middle of WriteAsync despite there being no transport issue.

Maybe that is something we should be worried about... I'm just extra worried about swallowing errors now.

src/Servers/Kestrel/Core/src/Middleware/Internal/DuplexPipeStreamAdapter.cs
```diff
@@ -14,36 +15,114 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal
...
+        public DuplexPipeStreamAdapter(IDuplexPipe duplexPipe, StreamPipeReaderOptions readerOptions, StreamPipeWriterOptions writerOptions, Func<Stream, TStream> createStream) :
+            base(duplexPipe.Input, duplexPipe.Output)
         {
             Stream = createStream(this);
+
+            var outputOptions = new PipeOptions(pool: writerOptions.Pool,
+                                                readerScheduler: PipeScheduler.Inline,
+                                                writerScheduler: PipeScheduler.Inline,
+                                                pauseWriterThreshold: 1,
+                                                resumeWriterThreshold: 1,
```
halter73: We should probably use the defaults here like we did before to reduce thrashing:

https://github.com/aspnet/AspNetCore/blob/release/3.0-preview6/src/Servers/Kestrel/Core/src/Internal/HttpConnection.cs#L67-L76

```c#
        internal PipeOptions AdaptedOutputPipeOptions => new PipeOptions
        (
            pool: MemoryPool,
            readerScheduler: PipeScheduler.Inline,
            writerScheduler: PipeScheduler.Inline,
            pauseWriterThreshold: _context.ServiceContext.ServerOptions.Limits.MaxResponseBufferSize ?? 0,
            resumeWriterThreshold: _context.ServiceContext.ServerOptions.Limits.MaxResponseBufferSize ?? 0,
            useSynchronizationContext: false,
            minimumSegmentSize: MemoryPool.GetMinimumSegmentSize()
        );
```

halter73 (approved):
Please fix the pauseWriterThreshold and resumeWriterThreshold before merging though.

src/Servers/Kestrel/Core/src/Internal/Http2/Http2FrameWriter.cs
```diff
@@ -89,7 +89,6 @@ public void Complete()
 
                 _completed = true;
                 _connectionOutputFlowControl.Abort();
-                _outputWriter.Complete();
```
halter73: 👍

halter73:
Maybe trying out the lower thresholds are OK for a preview. It does seem a little unecsary to try out new options now if we're removing it completely in the next preview though.

davidfowl:
> Maybe trying out the lower thresholds are OK for a preview. It does seem a little unecsary to try out new options now if we're removing it completely in the next preview though.

I'd rather leave it as it so its as passthrough as possible. Also the fact that it is temporary and the fact that those options aren't readily available (they need to be flowed from the UseHttps call) is just more churn. If you do feel strongly about changing those limits back to the Kestrel limits.

halter73:
> If you do feel strongly about changing those limits back to the Kestrel limits, I'll do it.

I'm fine with the way it is since it's temporary.

src/Servers/Kestrel/Core/src/Middleware/Internal/DuplexPipeStreamAdapter.cs
jkotalik: I'm going to leave it at error for now. I do agree that swallowing errors is a concern, but what else should we do. Debug.Assert? :trollface:

halter73: David already passed in a logger here. Critical logs are the new, much-improved Debug.Assert. Any critical log should cause the majority of our tests to fail.

davidfowl:
> I'm fine with the way it is since it's temporary. Otherwise, I'd at least ask for benchmarks.

Maybe it's reasonable to leave the defaults then.

davidfowl:
> I'm fine with the way it is since it's temporary. Otherwise, I'd at least ask for benchmarks.

I just realized though, you'll very likely never hit those limits since we always consume the entire buffer and the scheduling is inline. It means call to write will yield the await, write to the Stream then call Advance). It'll only matter when we're experiencing back pressure from the transport itself (which is the effect we want, complete passthrough).

halter73:
> I just realized though, you'll very likely never hit those limits since we always consume the entire buffer and the scheduling is inline. It means call to write will yield the await, write to the Stream then call Advance). It'll only matter when we're experiencing back pressure from the transport itself (which is the effect we want, complete passthrough).

This has always been my assumption. That's why I'm not *too* worried. I like validating my assumptions though.

PR #11602, opened 2019-06-26
Enforce selenium tests run on the expected platforms

javiercn:
Throw an exception in places where we conditionally execute selenium tests to ensure they run where we expect them to do.


SteveSandersonMS (approved):
Looks very useful!

PR #11603, opened 2019-06-26
Remove build Java from E2E testing

javiercn:
See if this actually causes the build to fail.


pranavkm:
@aspnet/build could you guys sign off? Not having our tests run is really problematic

dougbu (approved):
Guess this will work. I worry about the experience for local users who want to run _other_ e2e tests and don't have or want Java installed.

Suggest changing the `EnforcePrerequisites` error or the referenced readme to mention `/p:SeleniumE2ETestsSupported=false` or equivalent env variable.

JunTaoLuo (approved):
To unblock but this is not the ideal fix

PR #11606, opened 2019-06-26
Use task.IsCompletedSuccessfully rather than ReferenceEquals

benaadams:
As ASP.NET Core now targets .NET Core it can use `task.IsCompletedSuccessfully`; this is preferable to comparing to Task.CompletedTask as an async method uses a different completed task than Task.CompletedTask so it will catch more instances.

Also will make @stephentoub happier 😄

### DIFF
--- a/src/Mvc/Mvc.ViewFeatures/src/Buffers/PagedBufferedTextWriter.cs
+++ b/src/Mvc/Mvc.ViewFeatures/src/Buffers/PagedBufferedTextWriter.cs
@@ -111,7 +111,7 @@ namespace Microsoft.AspNetCore.Mvc.ViewFeatures.Buffers
 
             // FlushAsyncCore will return CompletedTask if nothing sync buffered
             // Fast-path and skip async state-machine if only a single async operation
-            return ReferenceEquals(flushTask, Task.CompletedTask) ? 
+            return flushTask.IsCompletedSuccessfully ? 
                 _inner.WriteAsync(value) :
                 WriteAsyncAwaited(flushTask, value);
         }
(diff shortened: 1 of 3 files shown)


stephentoub:
> so it will catch more instances

Including asynchronously completing operations that then complete fast enough that they're complete by the time the check happens.

davidfowl:
Much better,

stephentoub (approved):

rynowak (approved):
<3

halter73 (approved):

src/Mvc/Mvc.ViewFeatures/src/Buffers/PagedBufferedTextWriter.cs
```diff
@@ -111,7 +111,7 @@ public override Task WriteAsync(char value)
 
             // FlushAsyncCore will return CompletedTask if nothing sync buffered
```
IDisposable: Probably should delete these comments or update to something like
```
// If nothing sync buffered or only a single operation is already completed,
// flushTask.IsCompletedSuccessfully will be true, so fast-path and skip async state-machine
```

Same on https://github.com/aspnet/AspNetCore/pull/11606/files#diff-d6564431f116b915d3f4b3b4cb508867R129 and https://github.com/aspnet/AspNetCore/pull/11606/files#diff-d6564431f116b915d3f4b3b4cb508867R146

benaadams: Removed

benaadams:
Rebased

jkotalik (approved):

PR #11616, opened 2019-06-26
Generate VS nuget package for 2.1

jkotalik:
Fixes https://github.com/aspnet/AspNetCore/issues/11071. This will also be in 2.2 once we merge up.

### DIFF
--- a/src/Installers/Windows/SharedFramework/SharedFramework.wixproj
+++ b/src/Installers/Windows/SharedFramework/SharedFramework.wixproj
@@ -17,6 +17,7 @@
     <HarvestSource>$(SharedFrameworkHarvestRootPath)\$(Platform)\</HarvestSource>
     <DefineConstants>$(DefineConstants);AspNetCoreSharedFrameworkSource=$(HarvestSource)</DefineConstants>
     <NamespaceGuid>$(SharedFrameworkNamespaceGuid)</NamespaceGuid>
+    <ToolsetInstallerNuspecFile>$(RepoRoot)\src\Installers\Windows\SharedFramework\SharedFrameworkPackage.nuspec</ToolsetInstallerNuspecFile>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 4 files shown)


jkotalik:
Builds are here: http://aspnetci/viewLog.html?buildId=659155&buildTypeId=Releases_21Public_WindowsInstallers&tab=artifacts#!fk6iluux85. Only difference is that the Major and Minor versions are actually set 😄

src/Installers/Windows/SharedFramework/SharedFramework.wixproj
```diff
@@ -47,5 +49,24 @@
     </HarvestDirectory>
   </ItemGroup>
 
+  <Target Name="CreateSharedFrameworkNugetPackage" AfterTargets="CopyToArtifactsDirectory;Build">
+    <PropertyGroup>
+      <!--TODO this may not have en-US. it's confusing-->
```
dougbu: File a bug to track this if the TODO isn't resolved before you merge

dougbu (approved):
@anurse I suggest we at least let Shiproom know before merging this. I suspect we're done building 2.x for the next releases and that this will need to wait 'til August.

dougbu:
Note: I'm approving because the fix looks fine but do not merge without additional authorization

anurse:
Yes, the branches are ❌ Closed ❌ right now.

src/Installers/Windows/SharedFramework/SharedFrameworkPackage.nuspec
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
+    <metadata>
+        <id> VS.Redist.Common.AspNetCore.SharedFramework.$ARCH$.$MAJOR$.$MINOR$</id>
```
anurse:
```suggestion
        <id>VS.Redist.Common.AspNetCore.SharedFramework.$ARCH$.$MAJOR$.$MINOR$</id>
```

Leading space is *probably* not problematic but is weird.

src/Installers/Windows/SharedFramework/SharedFrameworkPackage.nuspec
```diff
@@ -0,0 +1,19 @@
...
+        <version>1.0.0</version>
+        <title> VS.Redist.Common.AspNetCore.SharedFramework.$ARCH$.$MAJOR$.$MINOR$</title>
```
anurse:
```suggestion
        <title>VS.Redist.Common.AspNetCore.SharedFramework.$ARCH$.$MAJOR$.$MINOR$</title>
```

analogrelay (approved):

jkotalik:
@anurse feel free to merge this whenever 😄

PR #11620, opened 2019-06-26
Remove LangVersion from Blazor server side

pranavkm:


pranavkm:
Not doing client side for now.

ryanbrandenburg (approved):
Yup, that's what it says on the tin.

PR #11634, opened 2019-06-27
Update E2ETest requirements

JunTaoLuo:
Updating the default behaviour.

- Do not enforce prerequisites by default
- Warn if prerequistes are not met unless explicitly enforced
- Run yarn install by default if Java is available

Addresses https://github.com/aspnet/AspNetCore/issues/11342.

### DIFF
--- a/src/Shared/E2ETesting/E2ETesting.props
+++ b/src/Shared/E2ETesting/E2ETesting.props
@@ -3,8 +3,8 @@
     <_DefaultProjectFilter>$(MSBuildProjectDirectory)\..\..</_DefaultProjectFilter>
     <DefaultItemExcludes>$(DefaultItemExcludes);node_modules\**</DefaultItemExcludes>
     <SeleniumProcessTrackingFolder Condition="'$(SeleniumProcessTrackingFolder)' == ''">$([MSBuild]::EnsureTrailingSlash('$(RepoRoot)'))artifacts\tmp\selenium\</SeleniumProcessTrackingFolder>
-    <SeleniumE2ETestsSupported Condition="'$(SeleniumE2ETestsSupported)' == '' and '$(TargetArchitecture)' != 'arm' and '$(OS)' == 'Windows_NT' and '$(BuildJava)' == 'true'">true</SeleniumE2ETestsSupported>
-    <EnforcePrerequisites Condition="'$(SeleniumE2ETestsSupported)' == 'true' and '$(EnforcePrerequisites)' == ''">true</EnforcePrerequisites>
+    <SeleniumE2ETestsSupported Condition="'$(SeleniumE2ETestsSupported)' == '' and '$(TargetArchitecture)' != 'arm' and '$(OS)' == 'Windows_NT'">true</SeleniumE2ETestsSupported>
+    <EnforcePrerequisites Condition="'$(SeleniumE2ETestsSupported)' == 'true' and '$(EnforcePrerequisites)' == ''">false</EnforcePrerequisites>
 
     <!-- WebDriver is not strong-named, so this test project cannot be strong named either. -->
     <SignAssembly>false</SignAssembly>
--- a/src/Shared/E2ETesting/E2ETesting.targets
+++ b/src/Shared/E2ETesting/E2ETesting.targets
@@ -4,14 +4,10 @@
 
   <!-- Ensuring that everything is ready before build -->
 
-  <Target Name="EnsureNodeJSRestored" Condition="'$(SeleniumE2ETestsSupported)' == 'true'" BeforeTargets="Build">
-    <Message Text="Running yarn install on $(MSBuildProjectFile)" Importance="High" />
-
-    <Message Condition="'$(EnforcePrerequisites)' == ''"
-      Importance="High"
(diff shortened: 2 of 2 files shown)


src/Shared/E2ETesting/E2ETesting.props
```diff
@@ -3,8 +3,8 @@
     <_DefaultProjectFilter>$(MSBuildProjectDirectory)\..\..</_DefaultProjectFilter>
     <DefaultItemExcludes>$(DefaultItemExcludes);node_modules\**</DefaultItemExcludes>
     <SeleniumProcessTrackingFolder Condition="'$(SeleniumProcessTrackingFolder)' == ''">$([MSBuild]::EnsureTrailingSlash('$(RepoRoot)'))artifacts\tmp\selenium\</SeleniumProcessTrackingFolder>
-    <SeleniumE2ETestsSupported Condition="'$(SeleniumE2ETestsSupported)' == '' and '$(TargetArchitecture)' != 'arm' and '$(OS)' == 'Windows_NT' and '$(BuildJava)' == 'true'">true</SeleniumE2ETestsSupported>
-    <EnforcePrerequisites Condition="'$(SeleniumE2ETestsSupported)' == 'true' and '$(EnforcePrerequisites)' == ''">true</EnforcePrerequisites>
+    <SeleniumE2ETestsSupported Condition="'$(SeleniumE2ETestsSupported)' == '' and '$(TargetArchitecture)' != 'arm' and '$(OS)' == 'Windows_NT'">true</SeleniumE2ETestsSupported>
+    <EnforcePrerequisites Condition="'$(SeleniumE2ETestsSupported)' == 'true' and '$(EnforcePrerequisites)' == ''">false</EnforcePrerequisites>
```
JunTaoLuo: I'm of the opinion that we can remove this property and just go with the default behaviour. We'll never enforce prereqs and you will need to install java for tests to work. The difference is that instead of errors at build time, you'll see errors during test instead.

src/Shared/E2ETesting/E2ETesting.targets
```diff
@@ -23,19 +19,46 @@
...
-    <Message Importance="High" Text="JAVA is available on the PATH" />
+    <Exec Command="java -version" IgnoreExitCode="true">
+      <Output TaskParameter="ExitCode" PropertyName="ErrorCode"/>
+    </Exec>
+
+    <Message
+      Condition="'$(ErrorCode)' == '0'"
+      Importance="High"
+      Text="JAVA is available on the PATH" />
+    <Message
```
JunTaoLuo: Could consider making this a warning instead, applies in a few places.

src/Shared/E2ETesting/E2ETesting.targets
```diff
@@ -23,19 +19,46 @@
 
   <!-- Running prerequisites -->
 
-  <Target Name="EnsurePrerequisites" Condition="'$(EnforcePrerequisites)' == 'true'" BeforeTargets="EnsureNodeJSRestored">
+  <Target Name="EnsurePrerequisites">
```
dougbu: Much preferred the previous approach of only executing this target when `$(EnforcePrerequisites)` is enabled. As-is, the property only slightly changes the messages but doesn't make a real difference. There's a message already about `$(EnforcePrerequsites)` being disabled in `EnsureNodeJSRestored`. And, doubling all the messages makes the target harder to read.

src/Shared/E2ETesting/E2ETesting.targets
dougbu: As I said above, get rid of the `'$(EnforcePrerequisites)' == 'false'` cases entirely. If a user passes `/p:EnforcePrerequisites=false` on the command line, assume they know Java is where it should be.

dougbu (changes requested):

javiercn:
I wasn't sure how to explain my thoughts properly, so I made a PR. I think this would be easier, wouldn't it? https://github.com/aspnet/AspNetCore/pull/11642

It makes sure the prerequisites are enforced when building from VS or on the CI in supported platforms and enforces them when building from the command-line for projects that have E2E tests.

Shouldn't that be enough?

dougbu:
> Shouldn't that be enough?

Perhaps. This PR does have a couple of advantages over #11642:
- enforces prerequisites by default when Selenium tests are requested
- runs Yarn installations more often
- doesn't require changes to build.* files
- after reverting some parts I don't see as helpful, simpler

javiercn:
This PR has a way bigger scope than the one I sent out.

> * enforces prerequisites by default when Selenium tests are requested

Mine does too.


> * runs Yarn installations more often

How is this an advantage? I would see it as a disadvantage. It only adds time and potential flakiness

> * doesn't require changes to build.* files

Don't think that is a problem at all. Specially given that it literally does it in two places.

> * After reverting some parts I don't see as helpful, simpler

I don't even know how to answer this because its not even clear what you are suggesting, but perceive this comment largely as a personal opinion, and I'm not sure I agree with it. Alternatively the changes I suggested:
* Don't fundamentally change the way things work.
* Are very limited in scope (have reduced risk).
* Achieve our goal, which is to run this in the correct set of environments.

I'm open to revisit how things work in this area in the future, but right now the priority is to make the least risky think that makes things work/run again as this blocks any candidate build.

Hope this helps.

JunTaoLuo:
Sure go for it @javiercn

Issue #11393, opened 2019-06-19
Address some rough edges in service reference feature


dougbu:
- [x] avoid NU1704 build warnings and other client / server incompatibility issues
- [x] change default `$(OpenApiGenerateDocuments)` to `false` in projects using .NET Core 2.0 or earlier
- [x] add inputs and outputs to `GenerateOpenApiDocuments` target (reducing how often it runs)

See #4923 for more detail on some of the above. Conversation at https://github.com/aspnet/AspNetCore/pull/10669#discussion_r289567963 also contains some details.

dougbu:
4e4ccdd92a91

PR #11638, opened 2019-06-27
Address some rough edges in service reference feature

dougbu:
- #11393
- avoid NU1702 warnings
- move defaults for `$(OpenApiGenerateDocuments)` and `$(OpenApiGenerateDocumentsOnBuild)` later, into .targets
  - do not generate documents for .NET Core 2.0 and earlier TFMs (feature is not supported there)
- add `Inputs` and `Outputs` to the `GenerateOpenApiDocuments` target, reducing redundant builds

nit: correct Microsoft.Extensions.ApiDescription.Server package version when building locally

### DIFF
--- a/src/Mvc/Extensions.ApiDescription.Server/src/buildMultiTargeting/Microsoft.Extensions.ApiDescription.Server.targets
+++ b/src/Mvc/Extensions.ApiDescription.Server/src/buildMultiTargeting/Microsoft.Extensions.ApiDescription.Server.targets
@@ -1,5 +1,11 @@
 ﻿<?xml version="1.0" encoding="utf-8" standalone="no"?>
 <Project>
+  <PropertyGroup>
+    <OpenApiGenerateDocuments Condition=" '$(OpenApiGenerateDocuments)' == '' ">true</OpenApiGenerateDocuments>
+    <OpenApiGenerateDocumentsOnBuild
+        Condition=" '$(OpenApiGenerateDocumentsOnBuild)' == '' ">$(OpenApiGenerateDocuments)</OpenApiGenerateDocumentsOnBuild>
+  </PropertyGroup>
+
   <ItemGroup Condition=" '$(OpenApiGenerateDocuments)' == 'true' ">
     <ProjectCapability Include="OpenApiGenerateDocuments" />
   </ItemGroup>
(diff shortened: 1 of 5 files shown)


dougbu:
@rynowak this fixes a few issues you raised 😺

dougbu:
FYI @mkArtakMSFT @glennc

dougbu:
> Passes all tests? What CI witchery is this?

None of the tests we've marked as flaky made a peep, even in Helix. Only guess I have is that a celestial convergence occurred while the build ran 😈

src/Mvc/Extensions.ApiDescription.Server/src/buildMultiTargeting/Microsoft.Extensions.ApiDescription.Server.targets
```diff
@@ -17,11 +23,27 @@
       DependsOnTargets="GenerateOpenApiDocuments" />
 
   <Target Name="OpenApiGetDocuments" Returns="@(_OpenApiProjectDocuments)">
+    <ItemGroup>
+      <_Temporary Remove="@(_Temporary)" />
+      <_Temporary Include="$(TargetFrameworks)" Exclude="netcoreapp1.0;netcoreapp1.1;netcoreapp2.0" />
```
rynowak: Will we need to maintain this list? Are these just the releases that are too early for us to support?

src/Mvc/Extensions.ApiDescription.Server/src/buildMultiTargeting/Microsoft.Extensions.ApiDescription.Server.targets
```diff
@@ -17,11 +23,27 @@
...
         RemoveProperties="RuntimeIdentifier">
       <Output TaskParameter="TargetOutputs" ItemName="_OpenApiProjectDocuments" />
     </MSBuild>
+
+    <ItemGroup>
+      <_Temporary Remove="@(_Temporary)" />
+    </ItemGroup>
+    <PropertyGroup>
+      <_Temporary />
+    </PropertyGroup>
```
rynowak: I'm not totally sure that all of this messing temporaries is really required. There are property functions to check if an item group is empty and such.

rynowak (approved):

src/Mvc/Extensions.ApiDescription.Server/src/buildMultiTargeting/Microsoft.Extensions.ApiDescription.Server.targets
dougbu: Yes, these are just releases the `dotnet-getdocument` tool doesn't support. The list won't change until and unless we make drop .NET Core 2.1 or 2.2 support in the tool. If we do something like that, changes here should be part of it.

src/Mvc/Extensions.ApiDescription.Server/src/buildMultiTargeting/Microsoft.Extensions.ApiDescription.Server.targets
dougbu: Not checking anything. This is just about removing properties and items that other targets may use i.e. clean up. Leaving as is because someone on the MSBuild team complimented me on this practice 😺

PR #11642, opened 2019-06-27
Enforce E2E test prerequisites when building individual solutions only

javiercn:
Addresses https://github.com/aspnet/AspNetCore/issues/11342


SteveSandersonMS:
This looks good to me. I suggest also cherry-picking cd5106f into this branch because the devserver-based tests will fail without it (or if they somehow pass without it, that's crazy and we need to investigate).

Also one of the E2E tests about benchmarks will fail until we merge https://github.com/aspnet/AspNetCore/pull/11649

src/Shared/E2ETesting/E2ETesting.props
```diff
@@ -3,8 +3,16 @@
     <_DefaultProjectFilter>$(MSBuildProjectDirectory)\..\..</_DefaultProjectFilter>
     <DefaultItemExcludes>$(DefaultItemExcludes);node_modules\**</DefaultItemExcludes>
     <SeleniumProcessTrackingFolder Condition="'$(SeleniumProcessTrackingFolder)' == ''">$([MSBuild]::EnsureTrailingSlash('$(RepoRoot)'))artifacts\tmp\selenium\</SeleniumProcessTrackingFolder>
-    <SeleniumE2ETestsSupported Condition="'$(SeleniumE2ETestsSupported)' == '' and '$(TargetArchitecture)' != 'arm' and '$(OS)' == 'Windows_NT' and '$(BuildJava)' == 'true'">true</SeleniumE2ETestsSupported>
-    <EnforcePrerequisites Condition="'$(SeleniumE2ETestsSupported)' == 'true' and '$(EnforcePrerequisites)' == ''">true</EnforcePrerequisites>
+    <SeleniumE2ETestsSupported Condition="'$(SeleniumE2ETestsSupported)' == '' and '$(TargetArchitecture)' != 'arm' and '$(OS)' == 'Windows_NT'">true</SeleniumE2ETestsSupported>
+
+    <!-- We want to enforce prerequisites when we build from the CI or within Visual Studio -->
+    <EnforcedE2EBuildEnvironment Condition="'$(CI)' == 'true' or '$(BuildingInsideVisualStudio)' == 'true'">true</EnforcedE2EBuildEnvironment>
```
dougbu: - `EnforceE2EBuildEnvironment` (no 'd')
- But, why is an additional variable needed? Suggest `EnforceE2ETestPrerequisites` is enough

src/Shared/E2ETesting/E2ETesting.props
```diff
@@ -3,8 +3,16 @@
...
+
+    <!-- If not explicitly set, enforce E2E test prerequisites when the platform is supported and we are building on the CI or visual studio -->
+    <EnforceE2ETestPrerequisites Condition="'$(SeleniumE2ETestsSupported)' == 'true' and $(EnforcedE2EBuildEnvironment) == 'true' and '$(EnforceE2ETestPrerequisites)' == ''">true</EnforceE2ETestPrerequisites>
+
+    <!-- Never enforce E2E test prerequisites if the platform is not supported -->
+    <EnforceE2ETestPrerequisites Condition="'$(SeleniumE2ETestsSupported)' == ''"></EnforceE2ETestPrerequisites>
```
dougbu: ` '$(SeleniumE2ETestsSupported)' != 'true' `

src/Shared/E2ETesting/E2ETesting.targets
```diff
@@ -7,7 +7,7 @@
   <Target Name="EnsureNodeJSRestored" Condition="'$(SeleniumE2ETestsSupported)' == 'true'" BeforeTargets="Build">
     <Message Text="Running yarn install on $(MSBuildProjectFile)" Importance="High" />
 
-    <Message Condition="'$(EnforcePrerequisites)' == ''"
+    <Message Condition="'$(EnforceE2ETestPrerequisites)' == ''"
```
dougbu: ` '$(EnforceE2ETestPrerequisites)' != 'true' `

dougbu (changes requested):

src/Shared/E2ETesting/E2ETesting.targets
javiercn: I'm not going to be changing '== ''' to '!= 'true'' this is a well stablished msbuild convention and is not the focus of this PR.

src/Shared/E2ETesting/E2ETesting.props
javiercn: Clarity. It has a d because it is qualifying the environment as a place where E2E test prerequisites should be enforced.

src/Shared/E2ETesting/E2ETesting.targets
dougbu: The supposed convention you mention is _wrong_ and isn't used in most of our infrastructure. The problem is it prevents use of the explicit `/p:EnforceE2ETestPrerequisites=false` on the command line: An empty property means use the default, not `false`.

JunTaoLuo: I agree. We mostly use `'$(EnforceE2ETestPrerequisites)' != 'true'` convention which respects command line options.

src/Shared/E2ETesting/E2ETesting.props
JunTaoLuo: Is running E2E tests possible from the command line?

javiercn: Prerequisites will be enforced when calling build.cmd and when running from VS. I'm not sure we can do anything about dotnet test. The tests will still run, even though no prerequisites will be enforced. (Which should display a warning).

src/Shared/E2ETesting/E2ETesting.targets
javiercn: Ok, updated.

src/ProjectTemplates/test/Helpers/Project.cs
```diff
@@ -196,6 +197,7 @@ internal AspNetProcess StartPublishedProjectAsync()
             var environment = new Dictionary<string, string>
             {
                 ["ASPNETCORE_URLS"] = _urls,
+                ["ASPNETCORE_Kestrel__EndpointDefaults__Protocols"] = "Http1"
```
Tratcher: Oh?

javiercn: To workaround the ongoing HTTP/2.0 bug

src/Hosting/Hosting/src/StaticWebAssets/StaticWebAssetsLoader.cs
```diff
@@ -21,10 +22,11 @@ public class StaticWebAssetsLoader
         /// <summary>
         /// Configure the <see cref="IWebHostEnvironment"/> to use static web assets.
         /// </summary>
-        /// <param name="environment"></param>
-        public static void UseStaticWebAssets(IWebHostEnvironment environment)
+        /// <param name="environment">The application <see cref="IWebHostEnvironment"/>.</param>
+        /// <param name="configuration">The host <see cref="IConfiguration"/>.</param>
+        public static void UseStaticWebAssets(IWebHostEnvironment environment, IConfiguration configuration)
```
javiercn: I had to add config as a parameter here as the blazor dev-server needs to load the manifest for client-side apps. We will check the host config for a staticwebassets key and use that path if available.

If there is an alternative way of doing this, I'm all ears.

src/Components/test/E2ETest/Infrastructure/ServerFixtures/DevHostServerFixture.cs
```diff
@@ -32,7 +38,29 @@ protected override IWebHost CreateWebHost()
                 args.Add(Environment);
             }
 
-            return DevHostServerProgram.BuildWebHost(args.ToArray());
+            return new FakeWebHost(DevHostServerProgram.BuildWebHost(args.ToArray()));
+        }
+
+        private class FakeWebHost : IWebHost
```
javiercn: Don't want to go through all the tests updating the signatures. I'll file a bug and do clean-up afterwards.

dougbu:
Questions though I suspect this is ready to go (once tests pass):
1. Have you confirmed the Selenium tests in all the places run on the CI now?
2. What requires 56336622fbeb and the `--applicationPath` changes in this PR? Seems fairly unrelated and difficult to review together w/ the simpler MSBuild / scripting changes. I also thought the test fixes were being done in separate PRs.

src/ProjectTemplates/test/Helpers/Project.cs
Tratcher: The Kestrel one? That's been reverted.

javiercn:
> 1. Have you confirmed the Selenium tests in all the places run on the CI now?

Yes



> 2\. What requires [5633662](https://github.com/aspnet/AspNetCore/commit/56336622fbebfb2c77ef767cc131327c38db7095) and the `--applicationPath` changes in this PR? Seems fairly unrelated and difficult to review together w/ the simpler MSBuild / scripting changes. I also thought the test fixes were being done in separate PRs.

Failing tests as part of fixing the E2E tests. I'll have @SteveSandersonMS take a look at that tomorrow. Don't worry about it.

src/ProjectTemplates/test/Helpers/Project.cs
javiercn: I'll file a bug and do cleanup afterwards. We don't want to delay this fix.

javiercn: Well, I won't file a bug because I'll just make a PR right after I merge this. I don't want to do anything that destabilizes these tests at this point.

src/Hosting/Hosting/src/StaticWebAssets/StaticWebAssetsLoader.cs
javiercn: @Tratcher FYI

src/Components/Blazor/DevServer/src/Server/Program.cs
```diff
@@ -18,12 +25,24 @@ public class Program
...
+            Host.CreateDefaultBuilder(args)
+                .ConfigureHostConfiguration(cb => {
+                    var applicationPath = args.SkipWhile(a => a != "--applicationpath").Skip(1).FirstOrDefault();
+                    var name = Path.ChangeExtension(applicationPath,".StaticWebAssets.xml");
+
+                    if (name != null)
+                    {
+                        cb.AddInMemoryCollection(new Dictionary<string, string>
+                        {
+                            [WebHostDefaults.StaticWebAssetsKey] = name
```
javiercn: @Tratcher This is where the new API gets used.

Blazor standalone apps run using the blazor dev-server so it needs to be able to point to the manifest without using the client assembly as it won't be in its dependency graph

src/Components/Blazor/DevServer/src/Server/Program.cs
```diff
@@ -18,12 +25,24 @@ public class Program
...
-                .UseConfiguration(new ConfigurationBuilder()
-                    .AddCommandLine(args)
-                    .Build())
-                .UseStartup<Startup>()
-                .Build();
+        public static IHost BuildWebHost(string[] args) =>
+            Host.CreateDefaultBuilder(args)
+                .ConfigureHostConfiguration(cb => {
+                    var applicationPath = args.SkipWhile(a => a != "--applicationpath").Skip(1).FirstOrDefault();
+                    var name = Path.ChangeExtension(applicationPath,".StaticWebAssets.xml");
```
javiercn: We compute the value like this today. We'll pass it in through MSBuild in the future.

src/Components/Blazor/DevServer/src/Server/Program.cs
Tratcher: Move the mitigation to launchsettings?

javiercn: We don't want that, as the whole point of this package is to setup a suitable host for your blazor standalone app (which is only files on disk).

BTW, don't worry about this a lot because this piece is not shipping for 3.0, but we want to fix this as customers have been complaining.

JunTaoLuo (approved):
Once tests pass.

src/Components/Blazor/DevServer/src/Server/Program.cs
SteveSandersonMS: Would be good to have a tracking issue for the cleanup work

src/Components/test/testassets/TestContentPackage/TestContentPackage.csproj
```diff
@@ -4,19 +4,9 @@
...
-    <!-- .js files will be referenced via <script> tags -->
-    <EmbeddedResource Include="content\**\*.js" LogicalName="blazor:js:%(RecursiveDir)%(Filename)%(Extension)" />
-
-    <!-- .css files will be referenced via <link rel='Stylesheet'> tags -->
-    <EmbeddedResource Include="content\**\*.css" LogicalName="blazor:css:%(RecursiveDir)%(Filename)%(Extension)" />
-
-    <!-- Any other files will be included in the 'dist' output but without any tags referencing them -->
-    <EmbeddedResource Include="content\**" Exclude="**\*.js;**\*.css" LogicalName="blazor:file:%(RecursiveDir)%(Filename)%(Extension)" />
-  </ItemGroup>
-
```
SteveSandersonMS: It's great to see this stuff getting removed!

src/Hosting/Abstractions/src/WebHostDefaults.cs
```diff
@@ -21,5 +21,6 @@ public static class WebHostDefaults
         public static readonly string SuppressStatusMessagesKey = "suppressStatusMessages";
 
         public static readonly string ShutdownTimeoutKey = "shutdownTimeoutSeconds";
+        public static readonly string StaticWebAssetsKey = "staticwebassets";
```
SteveSandersonMS: Tiny nit - others use camelcase:

```suggestion
        public static readonly string StaticWebAssetsKey = "staticWebAssets";
```

src/Hosting/Hosting/src/StaticWebAssets/StaticWebAssetsLoader.cs
SteveSandersonMS: This makes sense now you explained it to me offline. For anyone else, the point here is that we don't want to assume it's OK to load the application assembly at runtime just to find out where it is on disk. This applies in the dev server case, but could in other cases too such as automated tests. So it's a good change even besides how it fixes the Blazor dev server case.

src/Components/Blazor/DevServer/src/Server/Program.cs
```diff
@@ -18,12 +25,24 @@ public class Program
...
+
+                    if (name != null)
```
SteveSandersonMS: Nit: I don't think `name` can be null here anyway, because if `applicationPath` was null there would already have been an exception on L32.

SteveSandersonMS (approved):
Great!

src/Components/Blazor/DevServer/src/Server/Program.cs
SteveSandersonMS: Also, just to check, is there any chance we need to be checking whether `File.Exists(name)` before proceeding with the rest of the static assets code paths below? I'm assuming it's possible for no such manifest file to exist.

javiercn: The extensions in WebHost already check if the file exist, and noop otherwise. I wouldn't want to do anything catastrophic here like throwing an exception as I don't want to make a bug/change in this logic to make the server unusable.

We have an E2E test for it, so we'll know if things change.

PR #11645, opened 2019-06-27
Apply selection to options even within markup block or optgroup. Fixes #5616 #10648

### LINKED ISSUE #5616: Markup blocks don't run tryApplySpecialProperty logic
In the Blazor.Browser.JS code, after it inserts a markup block, should run the same logic that it would if it had constructed the DOM attribute-by-attribute. But it doesn't, which means that things like:

```
<select value=@someValue>
    <option value="a">Item 1</option>
    <option value="b">Item 2</option>
</select>
```

... don't pre-populate the dropdown with the correct initial selection if (for example) `someValue` equals `"a"`.

To fix this, we need to run a pass through the inserted subtree to apply the captured values.

### LINKED ISSUE #10648: Bind for selects does not set selected value when using optgroups
When you set the value of a property that is bound to a select the option with that value will not be selected if it is nested in an optgroup.

### To Reproduce
```
<select bind="@SelectValueWithoutOptGroup">
    @foreach (var i in Enumerable.Range(1, 3))
    {
        <option value="@i">@i</option>
    }
</select>

<select bind="@SelectValueWithOptGroup">
    @foreach (var i in Enumerable.Range(1, 3))
    {
        <optgroup label="@i">
            @foreach(var j in Enumerable.Range(1, 3))
            {
                <option value="@i-@j">@i-@j</option>
            }
        </optgroup>
    }
</select>

@functions {
    string SelectValueWithoutOptGroup { get; set; }
    string SelectValueWithOptGroup { get; set; }

    protected override async Task OnInitAsync()
    {
        SelectValueWithOptGroup = "2-3";
        SelectValueWithoutOptGroup = "2";
    }
}
```

### Expected behavior
Here only the select without any optgroups will have the option selected ("2" in my example) but the one with optgroups will still have the first option selected ("1-1").  I expected the option "2-3" to be selected in the select that has optgroups as well.

### Additional context
Include the output of `dotnet --info`
```
 Version:   3.0.100-preview5-011568
 Commit:    b487ff10aa

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.18362
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\3.0.100-preview5-011568\

Host (useful for support):
  Version: 3.0.0-preview5-27626-15
  Commit:  61f30f5a23

.NET Core SDKs installed:
  2.1.800-preview-009677 [C:\Program Files\dotnet\sdk]
  3.0.100-preview5-011568 [C:\Program Files\dotnet\sdk]

.NET Core runtimes installed:
  Microsoft.AspNetCore.All 2.1.11 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.App 2.1.11 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 3.0.0-preview5-19227-01 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 2.1.11 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 3.0.0-preview5-27626-15 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.WindowsDesktop.App 3.0.0-preview5-27626-15 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
```

SteveSandersonMS:
This is pretty simple - it's just adding a second case where we need to apply the value to a `<select>`.

While doing this I realised it would be almost free to also fix #10648 (add optgroup support) so I did that too.


rynowak (approved):
Looks right to me, but ideally @javiercn would look as well 😆

SteveSandersonMS:
Thanks @rynowak.

@javiercn is a bit busy currently. If we wait, this will inevitably end up with a merge conflict because it touches `blazor.*.js`, so I've merged now. Please feel free to look later though @javiercn!

src/Components/Browser.JS/src/Rendering/BrowserRenderer.ts
```diff
@@ -353,6 +366,18 @@ export class BrowserRenderer {
...
+  private findClosestAncestorSelectElement(element: Element | null) {
+    while (element) {
+      if (element instanceof HTMLSelectElement) {
+        return element;
+      } else {
+        element = element.parentElement;
+      }
+    }
+
+    return null;
```
javiercn: Is this the right approach? Wouldn't this silently fail? Should we at least log that we didn't find a select element? Am I being overly cautious here?

javiercn (commented):
:shipit:

src/Components/Browser.JS/src/Rendering/BrowserRenderer.ts
SteveSandersonMS: I think you are being overly cautious. A developer could choose to have an `<option>` on its own (not nested inside a `<select>`). It wouldn't do anything useful, but browsers tolerate it. I don't think we should be trying to create extra errors in these cases, so I'm happy with the behavior of just doing nothing with such `<option>` elements.

PR #11662, opened 2019-06-27
Support fewer data types in BlazorPackHubProtocol

pranavkm:


src/Components/Server/src/BlazorPack/BlazorPackHubProtocol.cs
```diff
@@ -445,10 +437,6 @@ private static object DeserializeObject(ref MessagePackReader reader, Type type,
                 {
                     return ReadString(ref reader, "argument");
                 }
-                else if (type == typeof(bool))
-                {
-                    return reader.ReadBoolean();
-                }
                 else if (type == typeof(int))
```
pranavkm: We don't actually need this, but we lose a lot of test coverage if we do so.

src/Components/Server/src/BlazorPack/BlazorPackHubProtocol.cs
```diff
@@ -457,17 +445,6 @@ private static object DeserializeObject(ref MessagePackReader reader, Type type,
                 {
                     return reader.ReadInt64();
                 }
-                else if (type == typeof(float))
-                {
-                    return reader.ReadSingle();
-                }
-                else if (type == typeof(byte[]))
```
pranavkm: Yaay

src/Components/Server/src/BlazorPack/BlazorPackHubProtocol.cs
```diff
@@ -420,18 +420,10 @@ private void SerializeArgument(ref MessagePackWriter writer, object argument)
...
 
-                case float floatValue:
-                    writer.Write(floatValue);
-                    break;
-
                 case byte[] byteArray:
                     writer.Write(byteArray);
                     break;
 
-                case Exception exception:
```
pranavkm: And gone

davidfowl:
- No description
- No issue linked

What's the reason for doing this?

rynowak:
I'm not sure why we'd do this other than removing exception (security)

pranavkm:
These are data types not required for Blazor to function. Removing dead code particularly when there are security implications (serialization \ deserialization) seems like a good thing, no?

src/Components/Server/src/BlazorPack/BlazorPackHubProtocol.cs
```diff
@@ -457,17 +445,6 @@ private static object DeserializeObject(ref MessagePackReader reader, Type type,
                 {
                     return reader.ReadInt64();
                 }
-                else if (type == typeof(float))
```
mikaelm12: Sort of unrelated, why don't you use `is` for these checks?

davidfowl: It's a `Type` instance itself not an instance of an object.

pranavkm: I thought `is` only works on instances, such as `someInstance is float`, not when comparing two `Type` instances. Did I get that wrong?

mikaelm12: Ah, gotcha

pranavkm:
*bump*

SteveSandersonMS (approved):
I'm happy with this. Thanks @pranavkm!

Before merging, could you check with Ryan whether he has strong opinions about removing the benign things like `float` etc? His comment suggests he's unconvinced this is the right thing to do.

For my purposes, I'm fine either way since we're not using these. I do lean towards removing stuff we're not using, except if this is going to come at significant maintenance cost if we're planning to "upgrade" to newer versions of MessagePack by copy/pasting code.

rynowak:
I think it's wierd seeing these tests diverge for something that's harmless. We saw a security concern raised because `Exception` was in the list of types. There's no concern about primitives.

pranavkm:
We already support fewer data types and consequently run a subset of the tests that SignalR's message pack does. Feels like a small extension to it

pranavkm:
🆙 📅

rynowak (approved):

PR #11672, opened 2019-06-28
Spike an IConnectionFactory for the sockets transport

davidfowl:
- Made the unix domain sockets test use it

I'd like to have more than one implementation of the `IConnectionFactory` to make sure the API is sound. This is one potential implementation.

### DIFF
--- a/src/Servers/Kestrel/Transport.Sockets/src/SharedSocketOptions.cs
+++ b/src/Servers/Kestrel/Transport.Sockets/src/SharedSocketOptions.cs
@@ -1,0 +1,30 @@
+using System;
+using System.Buffers;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets
+{
+    public class SharedSocketOptions
+    {
+        /// <summary>
+        /// Set to false to enable Nagle's algorithm for all connections.
+        /// </summary>
+        /// <remarks>
+        /// Defaults to true.
+        /// </remarks>
+        public bool NoDelay { get; set; } = true;
+
+        /// <summary>
+        /// The maximum size before the transport will stop proactively reading from the transport.
+        /// </summary>
+        public long? MaxReadBufferSize { get; set; } = 1024 * 1024;
+
+        /// <summary>
+        /// The maximum size the transport will buffer outside of OS buffers before pausing writes.
+        /// </summary>
+        public long? MaxWriteBufferSize { get; set; } = 64 * 1024;
+
+        internal Func<MemoryPool<byte>> MemoryPoolFactory { get; set; } = System.Buffers.SlabMemoryPoolFactory.Create;
+    }
+}
--- a/src/Servers/Kestrel/Transport.Sockets/src/SocketConnectionFactory.cs
+++ b/src/Servers/Kestrel/Transport.Sockets/src/SocketConnectionFactory.cs
@@ -1,0 +1,53 @@
+using System;
+using System.Collections.Generic;
+using System.IO.Pipelines;
+using System.Net;
+using System.Net.Sockets;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Connections;
+using Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets.Internal;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.AspNetCore.Server.Kestrel.Transport.Sockets
+{
+    /// <summary>
+    /// Defines a class for creating socket connections based on the specified endpoint.
+    /// </summary>
+    public class SocketConnectionFactory : IConnectionFactory
+    {
+        private readonly ILogger _logger;
+
+        /// <summary>
+        /// Creates the <see cref="SocketConnectionFactory"/>.
+        /// </summary>
+        /// <param name="loggerFactory">The logger factory</param>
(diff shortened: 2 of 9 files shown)


src/Servers/Kestrel/Transport.Sockets/src/SocketConnectionFactory.cs
```diff
@@ -0,0 +1,53 @@
...
+        /// </returns>
+        public async ValueTask<ConnectionContext> ConnectAsync(EndPoint endPoint, CancellationToken cancellationToken = default)
+        {
+            var protocolType = endPoint is UnixDomainSocketEndPoint ? ProtocolType.Unspecified : ProtocolType.Tcp;
+
+            var socket = new Socket(endPoint.AddressFamily, SocketType.Stream, protocolType);
+
+            await socket.ConnectAsync(endPoint);
+
+            var connection = new SocketConnection(socket, memoryPool: null, PipeScheduler.ThreadPool, new SocketsTrace(_logger));
```
davidfowl: The pool and scheduler might be better as constructor arguments to this factory.

halter73: I say you get ahead of it and create and options object (though not necessarily ioptions) before we add too many params to the constructor.

src/Shared/Pipelines/PipeReaderExtensions.cs
```diff
@@ -0,0 +1,107 @@
+﻿using System;
+using System.Buffers;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace System.IO.Pipelines
+{
+    public static class PipeReaderExtensions
```
halter73: Who's calling these extension methods?

src/Servers/Kestrel/Transport.Sockets/src/SocketConnectionFactory.cs
davidfowl: The test

davidfowl: Yep

src/Shared/Pipelines/PipeReaderExtensions.cs
davidfowl: Tests

davidfowl:
@halter73 give this another look.

src/Servers/Kestrel/Transport.Sockets/src/SharedSocketOptions.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// </summary>
+        /// <remarks>
+        /// Defaults to true.
+        /// </remarks>
+        public bool NoDelay { get; set; } = true;
+
+        /// <summary>
+        /// The maximum size before the transport will stop proactively reading from the transport.
+        /// </summary>
+        public long? MaxReadBufferSize { get; set; } = 1024 * 1024;
```
halter73: While this is a break from KestrelServerLimits, I think we should standardize on having null be the default value for nullable ints and longs so we can more reasonably change the defaults later. We can use long.MaxValue or -1 to represent unlimited if we feel that's necessary. This should be more consistent with PipeOptions, StreamPipeReaderOptions, and StreamPipeWriterOptions.

src/Servers/Kestrel/Transport.Sockets/src/SocketConnectionFactory.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        private readonly ILogger _logger;
+        private readonly SocketClientOptions _options;
+
+        /// <summary>
+        /// Creates the <see cref="SocketConnectionFactory"/>.
+        /// </summary>
+        /// <param name="options">The options for this transport</param>
+        /// <param name="loggerFactory">The logger factory</param>
+        public SocketConnectionFactory(IOptions<SocketClientOptions> options, ILoggerFactory loggerFactory)
```
halter73: Do we want a client API to take IOptions? Do we expect this normally to be retrieved from a DI container?

src/Servers/Kestrel/Transport.Sockets/src/SharedSocketOptions.cs
halter73: I forgot SocketTransportOptions still derives from this... I guess that means we shouldn't be changing the interpretation of null. It is an inconsistency with PipeOptions, but at leas this is more consistent with other Kestrel-related options I guess 😞

src/Servers/Kestrel/Transport.Sockets/src/SocketConnectionFactory.cs
davidfowl: Tough question but a good one. The only client API we have now is DI enabled.

davidfowl:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=277157

aspnet-hello:

src/Servers/Kestrel/Transport.Sockets/src/SocketConnectionFactory.cs
davidfowl: You know how we solve this? We leave it DI friendly and then add static factory methods for the non DI case.

`var factory = SocketConnectionFactory.Create();`

davidfowl:
@anurse  I want to get something like this in 3.1. Lets discuss next week on possibilities.

Issue #7370, opened 2019-02-07
Revisit making Response.BodyPipe.Complete() actually complete the PipeWriter for Http2


jkotalik:
In the Http1 and Http2 output producers, we currently make Complete effectively noop. For Http1 we can't call PipeWriter.Complete() as that is the connection pipe and there isn't a notion of a response pipe. For Http2, we CAN call Complete() as there is a Data pipe available. However, there are issue with creating a well formatted response if we do support Complete.

- Do we require users to call StartAsync() before calling Complete(). This will guarantee that headers are committed.
- How do we handle Trailers? ProduceEnd?

Tratcher:
This would prevent middleware from adding trailers.

Tratcher:
Isn't this superseded by CompleteAsync?

halter73:
> Isn't this superseded by CompleteAsync?

No. CompleteAsync is preferred b/c it allows the app to observe back pressure while also signalling response completion. Had it not been for the back pressure issue, there wouldn't be a need for both Response.BodyPipe.Complete and CompleteAsync.

Tratcher:
BodyPipe.Complete also stops working as soon as you use a middleware that replaces the response stream (e.g. response compression) and the pipe auto-replaces with an adapter. We're going to implement CompleteAsync there too. CompleteAsync will be more reliable for apps to call than BodyPipe.Complete.

Tratcher:
To that end, we should not have Complete close the response, only CompleteAsync, so devs always get a consistent experience with and without the adapters.

halter73:
> Disposing the response body has never completed the response either.

Exactly. We could have made disposing the response body  stream exhibit reasonable behavior if we didn't have to continue supporting the broken behavior from 1.0. Notice how no tests failed except for two that explicitly expected this not to throw. Without the albatross of backwards-compatibility, we can finally make a nice API.

The fact that middleware that replaces the response stream has to think about the CompleteAsync scenario isn't new at all. Like you said they have to replace CompleteAsync already, which as far as I'm aware nothing does yet, so you can fix both things together.

[...]

PR #11675, opened 2019-06-28
Don't throw Exceptions from PipeWriter APIs after RST_STREAM

halter73:
See https://github.com/dotnet/corefx/issues/38911#issuecomment-506574681

### DIFF
--- a/src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
+++ b/src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
@@ -32,7 +32,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
         private static readonly byte[] _bytesTransferEncodingChunked = Encoding.ASCII.GetBytes("\r\nTransfer-Encoding: chunked");
         private static readonly byte[] _bytesServer = Encoding.ASCII.GetBytes("\r\nServer: " + Constants.ServerName);
 
-        protected BodyControl bodyControl;
+        protected BodyControl _bodyControl;
         private Stack<KeyValuePair<Func<object, Task>, object>> _onStarting;
         private Stack<KeyValuePair<Func<object, Task>, object>> _onCompleted;
 
@@ -315,17 +315,17 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 
         public void InitializeBodyControl(MessageBody messageBody)
         {
-            if (bodyControl == null)
+            if (_bodyControl == null)
             {
-                bodyControl = new BodyControl(bodyControl: this, this);
+                _bodyControl = new BodyControl(bodyControl: this, this);
             }
 
-            (RequestBody, ResponseBody, RequestBodyPipeReader, ResponseBodyPipeWriter) = bodyControl.Start(messageBody);
+            (RequestBody, ResponseBody, RequestBodyPipeReader, ResponseBodyPipeWriter) = _bodyControl.Start(messageBody);
             _requestStreamInternal = RequestBody;
             _responseStreamInternal = ResponseBody;
         }
 
-        public void StopBodies() => bodyControl.Stop();
+        public Task StopBodiesAsync() => _bodyControl.StopAsync();
 
         // For testing
         internal void ResetState()
@@ -497,7 +497,7 @@ namespace Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http
 
         protected void PoisonRequestBodyStream(Exception abortReason)
         {
-            bodyControl?.Abort(abortReason);
+            _bodyControl?.Abort(abortReason);
         }
 
         // Prevents the RequestAborted token from firing for the duration of the request.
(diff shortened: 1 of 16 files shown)


halter73:
By fixing the tests for this PR, I'm pretty sure I also accidentally fixed #7370.

src/Servers/Kestrel/Core/src/Internal/Http/HttpResponsePipeWriter.cs
```diff
@@ -35,7 +35,7 @@ public override void CancelPendingFlush()
         public override void Complete(Exception exception = null)
         {
             ValidateState();
-            _pipeControl.Complete(exception);
+            _ = _pipeControl.CompleteAsync(exception);
```
davidfowl: This task needs to be tracked and awaited

halter73: Good catch. That was only ever meant to be temporary ofc.

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
```diff
@@ -1418,7 +1417,22 @@ public void Complete(Exception ex)
...
+            {
+                await FireOnStarting();
+            }
+
+            // Flush headers, body, trailers...
+            if (!HasResponseCompleted)
+            {
+                if (!VerifyResponseContentLength(out var lengthException))
+                {
+                    throw lengthException;
```
davidfowl: IS this the code we had before? This should be `ExceptionDispatchInfo.Throw(lengthException)`

Tratcher: VerifyResponseContentLength is a factory that creates an exception, it hasn't been thrown before. On other code paths the exception is only logged.

src/Servers/Kestrel/Core/src/Internal/Http2/Http2OutputProducer.cs
```diff
@@ -360,6 +410,25 @@ private async ValueTask<FlushResult> ProcessDataWrites()
             return flushResult;
         }
 
+        private Memory<byte> GetFakeMemory(int sizeHint)
```
jkotalik: Maybe we should consider a shared output producer base class as there are now quite a few methods that are "shared".

src/Servers/Kestrel/test/InMemory.FunctionalTests/ResponseTests.cs
```diff
@@ -3819,19 +3822,17 @@ public async Task ResponseGetMemoryAndStartAsyncAdvanceThrows()
...
+            {
                 httpContext.Response.BodyWriter.Complete();
-                for (var i = 0; i < 5; i++)
-                {
-                    var memory = httpContext.Response.BodyWriter.GetMemory(); // Shouldn't throw
-                    httpContext.Response.BodyWriter.Advance(memory.Length);
-                }
 
-                await Task.CompletedTask;
+                writeEx = Assert.Throws<InvalidOperationException>(() => httpContext.Response.BodyWriter.GetMemory());
```
jkotalik: This part confuses me. I thought we wanted to have the same behavior as streams where calling WriteAsync after Complete noops. However, I  think the difference is that the noop behavior only occurs when the connection was aborted or the streams go away. Hence why you added the fake memory to HTTP2?

jkotalik (approved):
Do we need an explicit test for not throwing an InvalidOperationException after abort?

src/Servers/Kestrel/Core/src/Internal/Http/HttpProtocol.cs
```diff
@@ -1418,7 +1417,22 @@ public void Complete(Exception ex)
...
+            // Finalize headers
+            if (!HasResponseStarted)
+            {
+                await FireOnStarting();
+            }
+
+            // Flush headers, body, trailers...
+            if (!HasResponseCompleted)
+            {
+                if (!VerifyResponseContentLength(out var lengthException))
```
Tratcher: Throwing here is problematic because besides surfacing in user code, it would also surface in StopAcceptingWritesAsync which previously never threw and doesn't have error handling.

src/Servers/Kestrel/Core/src/Internal/Http/HttpResponsePipeWriter.cs
```diff
@@ -35,7 +37,7 @@ public override void CancelPendingFlush()
         public override void Complete(Exception exception = null)
         {
             ValidateState();
-            _pipeControl.Complete(exception);
+            _completeTask = _pipeControl.CompleteAsync(exception);
```
Tratcher: This sets us up for an inconsistent / confusing end-to-end story. Let's go over that before continuing here.
https://github.com/aspnet/AspNetCore/issues/7370#issuecomment-506806184

Tratcher (changes requested):

src/Servers/Kestrel/test/InMemory.FunctionalTests/ResponseTests.cs
halter73: We want the same behavior as Streams where aborting http connection or http/2 stream makes writing no-op. If the application explicitly completes the PipeWriter, that a different story entirely.

Tratcher (approved):

src/Servers/Kestrel/Core/src/Internal/Http2/Http2OutputProducer.cs
halter73: There's definitely some potential for better reuse. It could be a nice refactoring/cleanup down the road. I don't think it's critical for this change as this only duplicates a little more than we already were, and the logic that was copied is quite simple.

PR #11683, opened 2019-06-28
Add xml comments to HubInvocationContext

BrennanConroy:
Let the word-smithing begin!

### DIFF
--- a/src/SignalR/server/Core/src/HubInvocationContext.cs
+++ b/src/SignalR/server/Core/src/HubInvocationContext.cs
@@ -14,8 +14,8 @@ namespace Microsoft.AspNetCore.SignalR
         /// <summary>
         /// Instantiates a new instance of the <see cref="HubInvocationContext"/> class.
         /// </summary>
-        /// <param name="context">Context for the active hub connection and caller.</param>
-        /// <param name="hubMethodName">The name of the hub method being invoked.</param>
+        /// <param name="context">Context for the active Hub connection and caller.</param>
+        /// <param name="hubMethodName">The name of the Hub method being invoked.</param>
         /// <param name="hubMethodArguments">The arguments provided by the client.</param>
         public HubInvocationContext(HubCallerContext context, string hubMethodName, object[] hubMethodArguments)
         {
@@ -25,12 +25,12 @@ namespace Microsoft.AspNetCore.SignalR
         }
 
         /// <summary>
-        /// Gets the context for the active hub connection and caller.
+        /// Gets the context for the active Hub connection and caller.
         /// </summary>
         public HubCallerContext Context { get; }
 
         /// <summary>
-        /// Gets the name of the hub method being invoked.
+        /// Gets the name of the Hub method being invoked.
         /// </summary>
         public string HubMethodName { get; }
 
--- a/src/SignalR/server/Core/src/HubInvocationContext.cs
+++ b/src/SignalR/server/Core/src/HubInvocationContext.cs
@@ -7,16 +7,16 @@ using Microsoft.AspNetCore.Authorization;
 namespace Microsoft.AspNetCore.SignalR
 {
     /// <summary>
-    /// A context passed to <see cref="IAuthorizationHandler"/>'s for custom Authorization on Hub methods.
+    /// Context for a Hub invocation.
     /// </summary>
     public class HubInvocationContext
     {
         /// <summary>
-        /// Instantiates a new <see cref="HubInvocationContext"/> instance.
+        /// Instantiates a new instance of the <see cref="HubInvocationContext"/> class.
         /// </summary>
-        /// <param name="context">The context that provides information about the hub connection.</param>
+        /// <param name="context">Context for the active hub connection and caller.</param>
         /// <param name="hubMethodName">The name of the hub method being invoked.</param>
-        /// <param name="hubMethodArguments">The arguments provided by the client to call the hub method.</param>
+        /// <param name="hubMethodArguments">The arguments provided by the client.</param>
         public HubInvocationContext(HubCallerContext context, string hubMethodName, object[] hubMethodArguments)
         {
             HubMethodName = hubMethodName;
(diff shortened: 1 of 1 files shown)


src/SignalR/server/Core/src/HubInvocationContext.cs
```diff
@@ -2,20 +2,41 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.Collections.Generic;
+using Microsoft.AspNetCore.Authorization;
 
 namespace Microsoft.AspNetCore.SignalR
 {
+    /// <summary>
+    /// A context passed to <see cref="IAuthorizationHandler"/>'s for custom Authorization on Hub methods.
```
anurse: I'd be more general here. It will also be the type we use for the pipeline when/if we do that.

```suggestion
    /// Context for a Hub invocation.
```

src/SignalR/server/Core/src/HubInvocationContext.cs
```diff
@@ -2,20 +2,41 @@
...
+    /// </summary>
     public class HubInvocationContext
     {
+        /// <summary>
+        /// Instantiates a new <see cref="HubInvocationContext"/> instance.
```
anurse:
```suggestion
        /// Instantiates a new instance of the <see cref="HubInvocationContext"/> class.
```

Is the standard boilerplate AFAICT (https://docs.microsoft.com/en-us/dotnet/api/system.exception.-ctor?view=netcore-2.2)

src/SignalR/server/Core/src/HubInvocationContext.cs
```diff
@@ -2,20 +2,41 @@
...
+        /// </summary>
+        /// <param name="context">The context that provides information about the hub connection.</param>
```
anurse:
```suggestion
        /// <param name="context">Context for the active hub connection and caller.</param>
```

src/SignalR/server/Core/src/HubInvocationContext.cs
```diff
@@ -2,20 +2,41 @@
...
+        /// <param name="hubMethodName">The name of the hub method being invoked.</param>
+        /// <param name="hubMethodArguments">The arguments provided by the client to call the hub method.</param>
```
anurse:
```suggestion
        /// <param name="hubMethodArguments">The arguments provided by the client.</param>
```

src/SignalR/server/Core/src/HubInvocationContext.cs
```diff
@@ -2,20 +2,41 @@
...
         public HubInvocationContext(HubCallerContext context, string hubMethodName, object[] hubMethodArguments)
         {
             HubMethodName = hubMethodName;
             HubMethodArguments = hubMethodArguments;
             Context = context;
         }
 
+        /// <summary>
+        /// The context that provides information about the hub connection.
```
anurse:
```suggestion
        /// Gets context for the active hub connection and caller.
```

src/SignalR/server/Core/src/HubInvocationContext.cs
```diff
@@ -2,20 +2,41 @@
...
+        /// </summary>
         public HubCallerContext Context { get; }
+
+        /// <summary>
+        /// The name of the hub method being invoked.
```
anurse:
```suggestion
        /// Gets the name of the hub method being invoked.
```

src/SignalR/server/Core/src/HubInvocationContext.cs
```diff
@@ -2,20 +2,41 @@
...
+        /// </summary>
         public string HubMethodName { get; }
+
+        /// <summary>
+        /// The arguments provided by the client to call the hub method.
```
anurse:
```suggestion
        /// Gets the arguments provided by the client.
```

analogrelay (approved):
SMITHED!

src/SignalR/server/Core/src/HubInvocationContext.cs
BrennanConroy: "Gets the"?

src/SignalR/server/Core/src/HubInvocationContext.cs
davidfowl: Yes

src/SignalR/server/Core/src/HubInvocationContext.cs
mikaelm12: I think that's what he meant there

src/SignalR/server/Core/src/HubInvocationContext.cs
```diff
@@ -2,20 +2,41 @@
...
 {
+    /// <summary>
+    /// Context for a Hub invocation.
+    /// </summary>
     public class HubInvocationContext
     {
+        /// <summary>
+        /// Instantiates a new instance of the <see cref="HubInvocationContext"/> class.
+        /// </summary>
+        /// <param name="context">Context for the active hub connection and caller.</param>
```
mikaelm12: nit:Should this be "hub connection" or `HubConnection`?

src/SignalR/server/Core/src/HubInvocationContext.cs
```diff
@@ -2,20 +2,41 @@
...
+        /// <param name="hubMethodArguments">The arguments provided by the client.</param>
         public HubInvocationContext(HubCallerContext context, string hubMethodName, object[] hubMethodArguments)
         {
             HubMethodName = hubMethodName;
             HubMethodArguments = hubMethodArguments;
             Context = context;
         }
 
+        /// <summary>
+        /// Gets the context for the active hub connection and caller.
```
mikaelm12: Same question here

src/SignalR/server/Core/src/HubInvocationContext.cs
```diff
@@ -2,20 +2,41 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System.Collections.Generic;
+using Microsoft.AspNetCore.Authorization;
 
 namespace Microsoft.AspNetCore.SignalR
 {
+    /// <summary>
+    /// Context for a Hub invocation.
```
mikaelm12: super nit: You capitalize Hub here but have it lower everywhere else.

mikaelm12 (approved):

src/SignalR/server/Core/src/HubInvocationContext.cs
anurse: Please to get context

src/SignalR/server/Core/src/HubInvocationContext.cs
BrennanConroy: I don't think it should be `HubConnection` as that is the client side type, on the server it's a "connection to a hub" i.e. "hub connection". But I could be convinced either way

src/SignalR/server/Core/src/HubInvocationContext.cs
BrennanConroy: :+1: capitalizing Hub is probably correct

halter73 (approved):

PR #11685, opened 2019-06-28
Add a diagnostic source event that fires when a route is matched

davidfowl:
- Usually more information becomes available about a request once route is matched. This event should allow diagnostic systems to enlighten the typical "begin request" metadata to include more information about the matched route and more importantly the selected endpoint and associated metadata.

cc @lmolkova @SergeyKanzhelev 

I'd expect open telemetry to use this to determine what type of request is being routed instead of the specific MVC event:

https://github.com/open-telemetry/opentelemetry-dotnet/blob/3538e05f0eed9df008d30e82a70bef11bd579bb5/src/OpenTelemetry.Collector.AspNetCore/Implementation/HttpInListener.cs#L108-L137

You could also imagine crawling through the endpoint metadata to light up more information about the request (like the gRPC method or controller method that got selected).

### DIFF
--- a/src/Http/Routing/src/EndpointRoutingMiddleware.cs
+++ b/src/Http/Routing/src/EndpointRoutingMiddleware.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Diagnostics;
 using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
@@ -13,9 +14,12 @@ namespace Microsoft.AspNetCore.Routing
 {
     internal sealed class EndpointRoutingMiddleware
     {
+        private const string DiagnosticsRouteMatchedKey = "Microsoft.AspNetCore.Routing.RouteMatched";
+
         private readonly MatcherFactory _matcherFactory;
         private readonly ILogger _logger;
         private readonly EndpointDataSource _endpointDataSource;
+        private readonly DiagnosticListener _diagnosticListener;
         private readonly RequestDelegate _next;
 
         private Task<Matcher> _initializationTask;
@@ -24,6 +28,7 @@ namespace Microsoft.AspNetCore.Routing
             MatcherFactory matcherFactory,
             ILogger<EndpointRoutingMiddleware> logger,
             IEndpointRouteBuilder endpointRouteBuilder,
+            DiagnosticListener diagnosticListener,
             RequestDelegate next)
         {
             if (matcherFactory == null)
@@ -48,6 +53,7 @@ namespace Microsoft.AspNetCore.Routing
 
             _matcherFactory = matcherFactory;
             _logger = logger;
+            _diagnosticListener = diagnosticListener;
             _next = next;
 
             _endpointDataSource = new CompositeEndpointDataSource(endpointRouteBuilder.DataSources);
@@ -106,6 +112,13 @@ namespace Microsoft.AspNetCore.Routing
             }
             else
             {
+                // Raise an event if the route matched
+                if (_diagnosticListener.IsEnabled() && _diagnosticListener.IsEnabled(DiagnosticsRouteMatchedKey))
+                {
+                    // We're just going to send the HttpContext since it has all of the relevant information
+                    _diagnosticListener.Write(DiagnosticsRouteMatchedKey, httpContext);
+                }
+
                 Log.MatchSuccess(_logger, endpoint);
             }
 
--- a/src/Http/Routing/src/EndpointRoutingMiddleware.cs
+++ b/src/Http/Routing/src/EndpointRoutingMiddleware.cs
@@ -14,7 +14,7 @@ namespace Microsoft.AspNetCore.Routing
 {
     internal sealed class EndpointRoutingMiddleware
     {
-        private const string DiagnosticsRouteMatchedKey = "Microsoft.AspNetCore.Routing.RouteMatched";
+        private const string DiagnosticsEndpointMatchedKey = "Microsoft.AspNetCore.Routing.EndpointMatched";
 
         private readonly MatcherFactory _matcherFactory;
         private readonly ILogger _logger;
@@ -31,30 +31,15 @@ namespace Microsoft.AspNetCore.Routing
             DiagnosticListener diagnosticListener,
             RequestDelegate next)
         {
-            if (matcherFactory == null)
-            {
-                throw new ArgumentNullException(nameof(matcherFactory));
-            }
-
-            if (logger == null)
-            {
-                throw new ArgumentNullException(nameof(logger));
-            }
-
             if (endpointRouteBuilder == null)
             {
                 throw new ArgumentNullException(nameof(endpointRouteBuilder));
             }
 
-            if (next == null)
-            {
-                throw new ArgumentNullException(nameof(next));
-            }
-
-            _matcherFactory = matcherFactory;
-            _logger = logger;
-            _diagnosticListener = diagnosticListener;
-            _next = next;
+            _matcherFactory = matcherFactory ?? throw new ArgumentNullException(nameof(matcherFactory));
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            _diagnosticListener = diagnosticListener ?? throw new ArgumentNullException(nameof(diagnosticListener));
+            _next = next ?? throw new ArgumentNullException(nameof(next));
 
             _endpointDataSource = new CompositeEndpointDataSource(endpointRouteBuilder.DataSources);
         }
@@ -113,10 +98,10 @@ namespace Microsoft.AspNetCore.Routing
             else
             {
                 // Raise an event if the route matched
-                if (_diagnosticListener.IsEnabled() && _diagnosticListener.IsEnabled(DiagnosticsRouteMatchedKey))
+                if (_diagnosticListener.IsEnabled() && _diagnosticListener.IsEnabled(DiagnosticsEndpointMatchedKey))
                 {
                     // We're just going to send the HttpContext since it has all of the relevant information
-                    _diagnosticListener.Write(DiagnosticsRouteMatchedKey, httpContext);
+                    _diagnosticListener.Write(DiagnosticsEndpointMatchedKey, httpContext);
                 }
 
                 Log.MatchSuccess(_logger, endpoint);
--- a/src/Http/Routing/test/UnitTests/EndpointRoutingMiddlewareTest.cs
+++ b/src/Http/Routing/test/UnitTests/EndpointRoutingMiddlewareTest.cs
@@ -67,11 +67,11 @@ namespace Microsoft.AspNetCore.Routing
             var loggerFactory = new TestLoggerFactory(sink, enabled: true);
             var listener = new DiagnosticListener("TestListener");
 
-            listener.Subscribe(new DelegateObserver(pair =>
+            using var subscription = listener.Subscribe(new DelegateObserver(pair =>
             {
                 eventFired = true;
 
-                Assert.Equal("Microsoft.AspNetCore.Routing.RouteMatched", pair.Key);
+                Assert.Equal("Microsoft.AspNetCore.Routing.EndpointMatched", pair.Key);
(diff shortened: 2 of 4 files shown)


src/Http/Routing/test/UnitTests/EndpointRoutingMiddlewareTest.cs
```diff
@@ -57,16 +59,26 @@ public async Task Invoke_OnCall_WritesToConfiguredLogger()
         {
             // Arrange
             var expectedMessage = "Request matched endpoint 'Test endpoint'";
+            bool eventFired = true;
```
campersau:
```suggestion
            bool eventFired = false;
```

davidfowl: Damn, I missed that.

rynowak:
>once route is matched

What's a route :trollface:

src/Http/Routing/src/EndpointRoutingMiddleware.cs
```diff
@@ -13,9 +14,12 @@ namespace Microsoft.AspNetCore.Routing
 {
     internal sealed class EndpointRoutingMiddleware
     {
+        private const string DiagnosticsRouteMatchedKey = "Microsoft.AspNetCore.Routing.RouteMatched";
```
rynowak: This doesn't seem create. This should be `EndpointMatched`.

src/Http/Routing/src/EndpointRoutingMiddleware.cs
```diff
@@ -24,6 +28,7 @@ internal sealed class EndpointRoutingMiddleware
             MatcherFactory matcherFactory,
             ILogger<EndpointRoutingMiddleware> logger,
             IEndpointRouteBuilder endpointRouteBuilder,
+            DiagnosticListener diagnosticListener,
```
rynowak: Add null check

src/Http/Routing/src/EndpointRoutingMiddleware.cs
```diff
@@ -106,6 +112,13 @@ private Task SetRoutingAndContinue(HttpContext httpContext)
             }
             else
             {
+                // Raise an event if the route matched
+                if (_diagnosticListener.IsEnabled() && _diagnosticListener.IsEnabled(DiagnosticsRouteMatchedKey))
+                {
+                    // We're just going to send the HttpContext since it has all of the relevant information
+                    _diagnosticListener.Write(DiagnosticsRouteMatchedKey, httpContext);
```
rynowak: >  // We're just going to send the HttpContext since it has all of the relevant information

How does that actually work? Does someone need the `HttpContext` in their code to do anything useful?

davidfowl: What do you mean? I think the answer is yes but I’m not sure what you’re asking.

src/Http/Routing/src/EndpointRoutingMiddleware.cs
davidfowl: For our tests?

davidfowl:
Man this broke a bunch of tests 😂

src/Http/Routing/src/EndpointRoutingMiddleware.cs
```diff
@@ -13,9 +14,12 @@ namespace Microsoft.AspNetCore.Routing
 {
     internal sealed class EndpointRoutingMiddleware
     {
+        private const string DiagnosticsRouteMatchedKey = "Microsoft.AspNetCore.Routing.RouteMatched";
+
         private readonly MatcherFactory _matcherFactory;
         private readonly ILogger _logger;
         private readonly EndpointDataSource _endpointDataSource;
+        private readonly DiagnosticListener _diagnosticListener;
```
JamesNK: I think this should be typed as `DiagnosticSource` instead of `DiagnosticListener`. We're only using the type to raise diagnostic events, not subscribe to them.

When implementing this in gRPC yesterday I looked for the asp.net core source to see what is done here, and there is a mix between using listener and source. It probably doesn't matter, but some consistency might be helpful.

src/Http/Routing/src/EndpointRoutingMiddleware.cs
JamesNK: > How does that actually work?

The person who subscribes to the event will get the HttpContext. They can use it to log additional information, e.g. when the event is raised they can get the endpoint from the HttpContext and write its name/metadata.

src/Http/Routing/test/UnitTests/EndpointRoutingMiddlewareTest.cs
```diff
@@ -159,19 +172,46 @@ private HttpContext CreateHttpContext()
         private EndpointRoutingMiddleware CreateMiddleware(
             Logger<EndpointRoutingMiddleware> logger = null,
             MatcherFactory matcherFactory = null,
+            DiagnosticListener listener = null,
             RequestDelegate next = null)
         {
             next ??= c => Task.CompletedTask;
             logger ??= new Logger<EndpointRoutingMiddleware>(NullLoggerFactory.Instance);
             matcherFactory ??= new TestMatcherFactory(true);
+            listener ??= new DiagnosticListener("Microsoft.AspNet");
```
JamesNK: nit: `Microsoft.AspNetCore` is the name at runtime.
```suggestion
            listener ??= new DiagnosticListener("Microsoft.AspNetCore");
```

src/Http/Routing/src/EndpointRoutingMiddleware.cs
davidfowl: Perf!! The listener has the faster IsEnabled check

src/Http/Routing/src/EndpointRoutingMiddleware.cs
rynowak: >The person who subscribes to the event will get the HttpContext. They can use it to log additional information, e.g. when the event is raised they can get the endpoint from the HttpContext and write its name/metadata.

I invented `DiagnosticSource` 😆 

What I mean is - do we now expect users to have access to the 3.0 `HttpContext` type when writing listener code? We used to pass a bunch of a few pieces of top-level data when we defined events because it's not always easy to retrieve data using reflection on the `HttpContext` if it's more complex than a property.

Has the landscape changed? Does ApplicationInsights just reference our stuff directly?

rynowak: The other point is that if we ever wanted to pass some other data here as this event, it wouldn't be possible without a breaking change.

src/Http/Routing/src/EndpointRoutingMiddleware.cs
JamesNK: Huh, ok. Why is it on the listener and not the source :\

src/Http/Routing/src/EndpointRoutingMiddleware.cs
rynowak: for throughness, we null check the other parameters.

src/Http/Routing/src/EndpointRoutingMiddleware.cs
JamesNK: > I invented DiagnosticSource

TIL 😋 

Won't passing an anonymous object end up requiring reflection anyway?

JamesNK:
@davidfowl Is this PR what you refer to here: https://github.com/grpc/grpc-dotnet/pull/341#discussion_r298665014

So would the idea be that a diagnostic framework could listen for this event, get the endpoint, then inspect the endpoint metadata to figure out that it is a gRPC/MVC/SignalR request?

src/Http/Routing/src/EndpointRoutingMiddleware.cs
rynowak: virtual + sealed override.

src/Http/Routing/src/EndpointRoutingMiddleware.cs
rynowak: That's why I'm asking. The usage patterns have shifted a bit since we invented it.

JamesNK: Looks like reflection on all the things: https://github.com/open-telemetry/opentelemetry-dotnet/blob/3538e05f0eed9df008d30e82a70bef11bd579bb5/src/OpenTelemetry.Collector.AspNetCore/Implementation/HttpInListener.cs

src/Http/Routing/test/UnitTests/EndpointRoutingMiddlewareTest.cs
```diff
@@ -57,16 +59,26 @@ public async Task Invoke_OnCall_WritesToConfiguredLogger()
...
             var expectedMessage = "Request matched endpoint 'Test endpoint'";
+            bool eventFired = false;
 
             var sink = new TestSink(
                 TestSink.EnableWithTypeName<EndpointRoutingMiddleware>,
                 TestSink.EnableWithTypeName<EndpointRoutingMiddleware>);
             var loggerFactory = new TestLoggerFactory(sink, enabled: true);
+            var listener = new DiagnosticListener("TestListener");
+
+            listener.Subscribe(new DelegateObserver(pair =>
```
lmolkova: dispose it, or it will hurt later :)

src/Http/Routing/src/EndpointRoutingMiddleware.cs
lmolkova: 1. Is it correct that this event will fire for both: WebApi and MVC?
2. How do we get route from HttpContext?
3. AppInsights uses https://github.com/aspnet/Mvc/blob/master/src/Microsoft.AspNetCore.Mvc.Core/MvcCoreDiagnosticSourceExtensions.cs events to know about routing today. Should we stop listening to those?

lmolkova: > What I mean is - do we now expect users to have access to the 3.0 HttpContext type when writing listener code? We used to pass a bunch of a few pieces of top-level data when we defined events because it's not always easy to retrieve data using reflection on the HttpContext if it's more complex than a property.
Has the landscape changed? Does ApplicationInsights just reference our stuff directly?

@davidfowl made it HttpContext only in https://github.com/aspnet/AspNetCore/pull/11020 for perf and made it backward compatible by adding `HttpContext.HttpContext`

AppInsights piece for AspNetCore depends on Http Abstractions and knows about HttpContext. It uses reflection to fetch it from anonymous payload, but we'll switch to type cast for 3.0

JamesNK: 1. It will fire for any framework that matches using endpoint routing: MVC, razor pages, gRPC, SignalR
2. Yes
3. It should still be used for legacy apps. You should be able to listen for both side-by-side. I can't imagine why an ASP.NET Core would use both.

lmolkova (approved):

src/Http/Routing/src/EndpointRoutingMiddleware.cs
davidfowl: Yes people will have the HttpContext.

davidfowl: > The other point is that if we ever wanted to pass some other data here as part of this event, it wouldn't be possible without a breaking change.

Yea, but I rather not allocate. Other values can be shuttled via the `HttpContext`.

davidfowl:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=244354

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/2705

src/Http/Routing/src/EndpointRoutingMiddleware.cs
davidfowl: > That's why I'm asking. The usage patterns have shifted a bit since we invented it.

Right, it's unrealistic to mirror the entire object graph though there are versioning concerns around this in general, the http context is a pretty good exchange type to depend on. To get more interesting information the consumer will also need to look at specific endpoint metadata.

rynowak (approved):

JamesNK (approved):

davidfowl:
Hmm preview7 or 8...

davidfowl:
7

PR #11702, opened 2019-06-29
Make HttpConnectionFactory public

halter73:
- Move HttpConnectionFactory to the Http.Connections.Client assembly


davidfowl (approved):
The options dependency only makes me a little sad

Issue #10348, opened 2019-05-17
Rename Blazor project templates


mkArtakMSFT:
In-box is called "Blazor Server App" / blazorserver
OOB is called "Blazor WebAssembly App" / blazorwasm
With an option to add a hosting ASP.NET Core project

Mike-EEE:
I would like to recommend having "WebSockets" prominently displayed somewhere for server-side mode, if not in the name itself, then in the description (if it's not already).  Having this in the name would certainly help alleviate confusion with the different modes and provide a clearer description of how the state is updated in the application.  At least, it would for me. 😁

It would also be great to see this used in general terminology and documentation as well, as personally, I think it's one of the coolest features to hit the .NET space in quite a while.  Yes, cooler than WebAssembly as it allows you to take a more traditional (and safer?) route for those who are more risk-averse in their development strategies.

isaac2004:
@danroth27 @DamianEdwards @mkArtakMSFT Is this official? I will take this one but want to confirm these are the names. Should we update the description in VS as well?

danroth27:
@isaac2004 Yes, these are the final template names. If you'd like to help out with the updates, that would be great, although it's a bit more work than just changing the names.

### ASP.NET Core hosted parameter

We need to decide on the parameter name for the Blazor WebAssembly App template to indicate that it should be ASP.NET Core hosted. I propose we go with `--hosted`. It should be false by default. 

### Blazor Server App description

I'm also in favor of updating the description of the Blazor Server App template. Here's the current description:

> A project template for creating a Blazor application that runs server-side inside an ASP.NET Core application. This template can be used for web applications with rich dynamic user interfaces (UIs).

Proposed new description:

> A project template for creating a Blazor server app that runs server-side inside an ASP.NET Core app and handles user interactions over a SignalR connection. This template can be used for web apps with rich dynamic user interfaces (UIs).

### Blazor WebAssembly App desription

We also need a new description for the Blazor WebAssembly App template that combines the descriptions of the existing templates. Here are the current template descriptions

Current blazor description:

> A project template for creating a Blazor application that runs on WebAssembly and is hosted on an ASP.NET Core server.

Current blazorhosted description:

> A project template for creating a Blazor application that runs on WebAssembly and is hosted on an ASP.NET Core server.

Proposed new blazorwasm description;

> A project template for creating a Blazor app that runs on WebAssembly. This template can be used for web apps with rich dynamic user interfaces (UIs).

### Updated template and group identities

[...]

PR #11707, opened 2019-06-29
Remove package guids from client-side Blazor templates

### LINKED ISSUE #11706: Client-side Blazor templates indistinguishable in VS 16.3 P1
The client-side Blazor templates currently show up with the same name and description in VS 16.3 P1:

blazor-templates

This is because the updated localization content for these templates already went into 16.3 before the work to rename and refactor the templates was done (see https://github.com/aspnet/AspNetCore/issues/10348).

danroth27:
Fixes #11706 

The client-side Blazor templates currently show up with the same name and description in VS 16.3 P1. This is because the updated localization content for these templates already went into 16.3 before the work to rename and refactor the templates was done (see #10348).

We worked with the web tools folks on a resolution and the recommended workaround for this issue is to remove the package GUIDs from the VS host files for these templates. This will cause the localized content in VS to not be used for these templates and instead the strings from the templates will be used. This was tested and verified with the help of @phenning. 

After the localized template content is moved to this repo (see https://github.com/aspnet/AspNetCore/pull/11040) we can reenable localization for these templates in VS. 

/cc @vijayrkn

### DIFF
--- a/src/Components/Blazor/Templates/src/content/BlazorHosted-CSharp/.template.config.src/vs-2017.3.host.json
+++ b/src/Components/Blazor/Templates/src/content/BlazorHosted-CSharp/.template.config.src/vs-2017.3.host.json
@@ -2,7 +2,6 @@
   "$schema": "http://json.schemastore.org/vs-2017.3.host",
   "name": {
     "text": "Blazor (ASP.NET Core hosted)",
-    "package": "{0CD94836-1526-4E85-87D3-FB5274C5AFC9}",
     "id": "1050"
   },
   "description": {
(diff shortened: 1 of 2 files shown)


mkArtakMSFT (approved):

joslat:
Hi @danroth27 , is there any guideline on how to do this? (removing the package GUIDs from the VS host files for the blazor client templates) 
This comes from not seeing the blazor client template, see the thread at https://github.com/aspnet/AspNetCore.Docs/issues/13531

cc @guardrex 
Extreme thanks to @SteveSandersonMS and everybody involved for creating Blazor! you rock!!

Issue #3752, opened 2018-10-29
Enable source-build for the ASP.NET Core shared framework


natemcmaster:
See also https://github.com/dotnet/source-build/issues/375.

The Microsoft.AspNetCore.App shared framework and many of dependencies are not currently supported in source-build. To enable source-build on this shared runtime, we expect the following changes will need to be made:

* Add a build flag to enable building only source-buildable assets. Not everything in the repo will support source-build
* Enabling compiling without any remote assets.
* Change our build scripts skip steps which require unavailable tools, such as NuGetPackageVerifier, xunit analyzers, etc.

## Requirements

The following must be available via source-build:
* everything in the Microsoft.AspNetCore.App shared framework for Linux (currently excluding Windows, which requires Visual Studio to build)
* all assemblies in the Microsoft.AspNetCore.App  targeting pack

What's not in source-build yet:
* test projects
* internal tools
* KoreBuild
* anything not in Microsoft.AspNetCore.App

natemcmaster:
I've looked closer at implementing this. Here are some thoughts

Notes:

* Repos to add to source-build, in order
  1. aspnet/Extensions
  2. aspnet/AspNetCore-Tooling
  3. aspnet/AspNetCore
* It's unclear how our contributors and devs will know if a particular change would break source-build. I cannot identify a simple way to test source-build support as a pull-request check.
* Dependency flow for Maestro and dependency flow for source-build are slightly different (see https://github.com/dotnet/source-build/tree/master/Documentation/auto-dependency-flow). I think it should be possible to make both work without having duplicate dependency-flow systems, but concerned it may be fragile unless we have a good way to test this continually.
* We will need a fix or workaround for https://github.com/dotnet/sdk/issues/2910 before we can add aspnet/AspNetCore. Our build currently the build assumes the SDK is in `$repoRoot/.dotnet/`, but according to the Repo API, we may need to allow for external location and version of dotnet
* https://github.com/aspnet/AspNetCore/issues/7280 - convert to Arcade. Done on 2/3 repos. This should be done first so we don't also need to convert KoreBuild to source-build
* The Repo API calls for supporting .rsp files as input to build.sh/cmd scripts (see https://github.com/dotnet/source-build/blob/master/Documentation/RepoApi.md#rsp-files). This may be challenging to implement in PowerShell/bash. We should look to see if we can copy what other repos have done.
[...]

Issue #7503, opened 2019-02-12
Microsoft.SourceLink does not work with Razor components


natemcmaster:
Projects which use source link (https://github.com/dotnet/sourcelink) and Razor components fail to build with errors like:

```
CSC : error CS0016: Could not write to output file 'C:\src\aspnet\AspNetCore\src\Components\Blazor\testassets\StandaloneApp\obj\Debug\netstandard2.0\StandaloneApp.sourcelink.json' 
-- 'Could not find file 'C:\src\aspnet\AspNetCore\src\Components\Blazor\testassets\StandaloneApp\obj\Debug\netstandard2.0\StandaloneApp.sourcelink.json'.' [C:\src\aspnet\AspNetCore
\src\Components\Blazor\testassets\StandaloneApp\StandaloneApp.csproj]
```

This happens because the target which generates the source link file (GenerateSourceLinkFile) uses `BeforeTargets="CoreCompile"`. Razor components run the `Csc` task as a part of the RazorComponentGenerate target before `CoreCompile`, so sourcelink.json does not exist yet, and Csc will fail.

### Workaround

Add this to your .csproj file
```xml
  <!-- Workaround for target dependency order in Microsoft.NET.Sdk.Razor and Microsoft.SourceLink.Common -->
  <Target Name="_EnsureSourceLinkHappensBeforeRazorComponentGeneration" BeforeTargets="PrepareForRazorComponentGenerate" DependsOnTargets="GenerateSourceLinkFile" />
```

mkArtakMSFT:
Thanks @natemcmaster.
@rynowak , can you please look into this? Feel free to reassign if you believe somebody will be able to handle this easily.

pranavkm:
This was addressed in SourceLink as part of https://github.com/dotnet/sourcelink/pull/272. It sets the `SourceLink` path late enough for us to not run in to it during the initial codegen. We could consider cleaning up the target to not pass this value through during the declaration compilation: https://github.com/aspnet/AspNetCore-Tooling/blob/master/src/Razor/src/Microsoft.NET.Sdk.Razor/build/netstandard2.0/Microsoft.NET.Sdk.Razor.Component.targets#L247, but it doesn't seem pertinent at this point.

PR #11717, opened 2019-06-30
Add support for source-build

JunTaoLuo:
Addresses #3752

### DIFF
--- a/.azure/pipelines/ci.yml
+++ b/.azure/pipelines/ci.yml
@@ -471,3 +471,50 @@ jobs:
     - name: Linux_Test_Results
       path: artifacts/TestResults/
       publishOnError: true
+
+# Source build
+- job: Source_Build
+  displayName: 'Test: Linux Source Build'
+  container: centos:7
+  pool:
+    vmImage: 'ubuntu-16.04'
+  variables:
+    DotNetCoreSdkDir: $(Agent.ToolsDirectory)/dotnet
+    DOTNET_SYSTEM_GLOBALIZATION_INVARIANT: true
+  steps:
+  - script: |
+      source eng/common/native/common-library.sh
+      mkdir -p $HOME/bin
+      GetFile https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64 $HOME/bin/jq
+      chmod +x $HOME/bin/jq
+      echo "##vso[task.prependpath]$HOME/bin"
+    displayName: Install jq
+  - task: UseDotNet@2
+    displayName: 'Use .NET Core sdk'
+    inputs:
+      packageType: sdk
+      # The SDK version selected here is intentionally supposed to use the latest release
+      # For the purpose of building Linux distros, we can't depend on features of the SDK
+      # which may not exist in pre-built versions of the SDK
+      version: 3.0.x
+      installationPath: $(DotNetCoreSdkDir)
+      includePreviewVersions: true
+  - script: ./eng/scripts/ci-source-build.sh --ci --configuration Release /p:BuildManaged=true
+    displayName: Run ci-source-build.sh
+  - task: PublishBuildArtifacts@1
+    displayName: Upload logs
+    condition: always()
+    continueOnError: true
+    inputs:
+      pathtoPublish: artifacts/log/
+      artifactName: Source_Build_Logs
+      artifactType: Container
+      parallel: true
+  - task: PublishBuildArtifacts@1
+    displayName: Upload package artifacts
+    condition: always()
+    inputs:
+      pathtoPublish: artifacts/packages/
+      artifactName: Source_Build_Packages
+      artifactType: Container
+      parallel: true
--- a/.azure/pipelines/ci.yml
+++ b/.azure/pipelines/ci.yml
@@ -512,7 +512,9 @@ jobs:
       parallel: true
   - task: PublishBuildArtifacts@1
     displayName: Upload package artifacts
-    condition: always()
+    # Only capture source build artifacts in PRs for the sake of inspecting
+    # changes that impact source-build. The artifacts from this build pipeline are never actually used.
+    condition: and(succeeded(), eq(variables['Build.Reason'], 'PullRequest'))
     inputs:
       pathtoPublish: artifacts/packages/
       artifactName: Source_Build_Packages
--- a/eng/AfterSolutionBuild.targets
+++ b/eng/AfterSolutionBuild.targets
@@ -17,7 +17,7 @@
       SharedFrameworkTargetFramework="netcoreapp$(AspNetCoreMajorVersion).$(AspNetCoreMinorVersion)" />
   </Target>
 
-  <Target Name="GenerateBuildAssetManifest" AfterTargets="Pack" Condition="'$(ContinuousIntegrationBuild)' == 'true'">
+  <Target Name="GenerateBuildAssetManifest" AfterTargets="Pack" Condition="'$(ContinuousIntegrationBuild)' == 'true' and '$(DotNetBuildFromSource)' != 'true'">
     <!-- Generate build manifests. These manifests are used by Maestro and the Build Asset Registry to flow dependencies to other repos. -->
     <MSBuild Projects="$(MSBuildThisFileDirectory)tools\Maestro\Maestro.csproj"
           Targets="Restore"
--- a/eng/Workarounds.targets
+++ b/eng/Workarounds.targets
@@ -21,9 +21,14 @@
   </ItemGroup>
 
   <!-- Workaround https://github.com/aspnet/AspNetCore/issues/7503. This chains GenerateSourceLinkFile before razor component targets run. -->
-  <Target Name="_EnsureSourceLinkHappensBeforeRazorComponentGeneration"
+  <!-- Workaround https://github.com/dotnet/source-build/issues/1112. Source link is currently disabled in source build so do not apply this worksaround. -->
+  <Target Condition="'$(DotNetBuildFromSource)' != 'true'"
+          Name="_EnsureSourceLinkHappensBeforeRazorComponentGeneration"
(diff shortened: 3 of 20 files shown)


eng/AfterSolutionBuild.targets
```diff
@@ -17,7 +17,7 @@
       SharedFrameworkTargetFramework="netcoreapp$(AspNetCoreMajorVersion).$(AspNetCoreMinorVersion)" />
   </Target>
 
-  <Target Name="GenerateBuildAssetManifest" AfterTargets="Pack" Condition="'$(ContinuousIntegrationBuild)' == 'true'">
+  <Target Name="GenerateBuildAssetManifest" AfterTargets="Pack" Condition="'$(ContinuousIntegrationBuild)' == 'true' and '$(DotNetBuildFromSource)' != 'true'">
```
JunTaoLuo: I think we should skip publishing source build artifacts to BAR.

JunTaoLuo:
@aspnet-hello triage https://dev.azure.com/dnceng/public/_build/results?buildId=253104&view=logs

aspnet-hello:

https://github.com/aspnet/AspNetCore-Internal/issues/2771

eng/AfterSolutionBuild.targets
dougbu: Agreed

.azure/pipelines/ci.yml
```diff
@@ -471,3 +471,50 @@ jobs:
...
+    condition: always()
+    continueOnError: true
+    inputs:
+      pathtoPublish: artifacts/log/
+      artifactName: Source_Build_Logs
+      artifactType: Container
+      parallel: true
+  - task: PublishBuildArtifacts@1
+    displayName: Upload package artifacts
+    condition: always()
```
dougbu: Never upload packages for builds from a fork. See for example the equivalent in Extensions:
```suggestion
    # Only capture source build artifacts in PRs for the sake of inspecting
    # changes that impact source-build. The artifacts from this build pipeline are never actually used.
    condition: and(succeeded(), eq(variables['Build.Reason'], 'PullRequest'))
```

eng/Workarounds.targets
```diff
@@ -21,10 +21,15 @@
   </ItemGroup>
 
   <!-- Workaround https://github.com/aspnet/AspNetCore/issues/7503. This chains GenerateSourceLinkFile before razor component targets run. -->
-  <Target Name="_EnsureSourceLinkHappensBeforeRazorComponentGeneration"
+  <!-- Workaround https://github.com/dotnet/source-build/issues/1112. Source link is currently disabled in source build so do not apply this worksaround. -->
```
dougbu: Should either restore the link to our tracking bug (https://github.com/aspnet/AspNetCore/issues/7503) or include it too.

Also, restore comments describing what the target does.

eng/WorkaroundsImported.targets
```diff
@@ -0,0 +1,5 @@
+<!-- Use this file to workaround issues. List the issue tracking the item to fix so we can remove the workaround when the issue is resolved. -->
+<Project>
+  <!-- Workaround https://github.com/dotnet/source-build/issues/1112. Source link is currently disabled in source build so define this dummy target which is required for pack. -->
+  <Target Name="InitializeSourceControlInformation"/>
```
dougbu: This is an empty target in Microsoft.Common.CurrentVersion.targets. Why does it need to be defined again here?

eng/scripts/ci-source-build.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
```
dougbu: Is this file identical to the one in Extensions?

src/Framework/ref/Microsoft.AspNetCore.App.Ref.csproj
```diff
@@ -76,6 +76,9 @@ This package is an internal implementation of the .NET Core SDK and is not meant
       _InstallTargetingPackIntoLocalDotNet;
       _CreateTargetingPackArchive;
     </BuildDependsOn>
+
+    <!-- No-op when in source build -->
+    <BuildDependsOn Condition="'$(IsTargetingPackBuilding)' != 'false' and '$(DotNetBuildFromSource)' == 'true'"/>
```
dougbu: Delete because this additional definition of the target should override the previous one. Not needed anyhow.

src/Identity/UI/src/Microsoft.AspNetCore.Identity.UI.csproj
```diff
@@ -40,7 +40,8 @@
     <UIFrameworkVersionMoniker Include="V4" />
   </ItemGroup>
 
-  <Target Name="BuildRazorViews" DependsOnTargets="Compile" BeforeTargets="Build">
+  <!-- Source build doesn't build this package -->
+  <Target Name="BuildRazorViews" DependsOnTargets="Compile" BeforeTargets="Build" Condition="'$(DotNetBuildFromSource)' != 'true'">
```
dougbu: Why is this necessary? This project isn't in Microsoft.AspNetCore.App

dougbu (changes requested):

eng/Workarounds.targets
JunTaoLuo: I didn't remove anything?

eng/WorkaroundsImported.targets
JunTaoLuo: It's normally a target defined by in source link. I see that in a few place in our code we have custom targets that depends on this target or reference it in Before/AfterTarget during pack. Since sourcelink is currently disabled, we need to define a this target so that those custom targets still work.

eng/scripts/ci-source-build.sh
JunTaoLuo: Not quite, I had to add a step to build the RepoTask.csproj.

src/Framework/ref/Microsoft.AspNetCore.App.Ref.csproj
JunTaoLuo: I don't quite follow? Removing this line breaks things. By default, this is not empty and there are targets that run.

src/Identity/UI/src/Microsoft.AspNetCore.Identity.UI.csproj
JunTaoLuo: So what happens is that when `ExcludeFromSourceBuild` is set is that noop'ing targets are imported and run during build: https://github.com/dotnet/arcade/blob/master/src/Microsoft.DotNet.Arcade.Sdk/tools/Empty.targets. However, the way that these targets are defined, they still run explicitly when these projects are built, even though the Build target itself noops.

.azure/pipelines/ci.yml
JunTaoLuo: Actually, how about I set it to just Build.Reason = PullRequest. I was debugging earlier and the succeed() condition meant I couldn't get the artifacts.

dougbu: That's fine while you're debugging but not once this is checked in. Look at all other artifacts published in this pipeline. Only artifacts/log files are ever published on error.

eng/Workarounds.targets
dougbu: Not sure why but I thought line 23 was removed.

:ok: just make it clear #7503 tracks dotnet/source-build#1112

src/Framework/ref/Microsoft.AspNetCore.App.Ref.csproj
dougbu: Is that coming in another iteration?

dougbu (approved):
Just a couple of outstanding points, including a few on earlier iterations

.azure/pipelines/ci.yml
dougbu: Oh, you've left `succeeded()` in. Alright&hellip;

JunTaoLuo: I made the comment after merging your suggestion, wanted to see what you think before I make the changes.

eng/Workarounds.targets
JunTaoLuo: I don't think it's quite the same. The issue described in #7503 is general to source link. dotnet/source-build#1112 is specific to source build.

src/Framework/ref/Microsoft.AspNetCore.App.Ref.csproj
JunTaoLuo: I removed the condition above but will keep this one.

dougbu:
Hurray:grey_exclamation:

sharpninja:
What is the expected happy path for building the repository from scratch on Ubuntu 18.04 in WSL 2?  I'm using this script:

```bash
#!/usr/bin/env bash

cd /mnt/e/GitHub/sharpninja

rm -r -f aspnetcore

git clone --recursive git@github.com:sharpninja/AspNetCore.git

git submodule update --init --recursive

cd aspnetcore

git remote add upstream https://github.com/aspnet/aspnetcore
git fetch upstream
git merge upstream/master

./restore.sh
./build.sh --all --restore -c Debug -v d
```
Unfortunately, it attempts to import `internal.aspnetcore.buildtasks/3.0.0-build-20190530.3` twice and bombs out on the second attempt.  I'm attaching a log file.
[aspnet-build.log](https://github.com/aspnet/AspNetCore/files/3477459/aspnet-build.log)

dougbu:
@sharpninja it sounds like you're having an issue with building this repo locally. If that's the case, commenting on a merged PR is not likely to get attention. Please file a new issue if my comments don't help.

Suggest leaving `--restore` out of the `./build.sh` command. In addition, execute `git submodule update --init --recursive` much later, after updating your local 'master' branch and just before `./restore.sh`.

In case your fork contains commits that are causing problems, I also suggest `git reset --hard upstream/master` instead of `git merge upstream/master`.

PR #11722, opened 2019-06-30
Improved HttpContext pooling

### LINKED ISSUE #11069: [Discussion] NullReferenceException at Microsoft.AspNetCore.Http.Internal.DefaultHttpRequest.get_Query()

benaadams:
Improved HttpContext pooling from https://github.com/aspnet/AspNetCore/pull/10319

Moves pooling from Kestrel to HttpContextFactory

Resolves #11069

/cc @davidfowl

### DIFF
--- a/src/Http/Http.Features/src/FeatureReferences.cs
+++ b/src/Http/Http.Features/src/FeatureReferences.cs
@@ -11,7 +11,7 @@ namespace Microsoft.AspNetCore.Http.Features
         public FeatureReferences(IFeatureCollection collection)
         {
             Collection = collection;
-            Cache = default(TCache);
+            Cache = default;
             Revision = collection.Revision;
         }
 
@@ -65,7 +65,7 @@ namespace Microsoft.AspNetCore.Http.Features
             Func<TState, TFeature> factory) where TFeature : class
         {
             var flush = false;
-            var revision = Collection.Revision;
+            var revision = Collection?.Revision ?? ContextDisposed();
             if (Revision != revision)
             {
                 // Clear cached value to force call to UpdateCached
@@ -108,5 +108,17 @@ namespace Microsoft.AspNetCore.Http.Features
 
         public TFeature Fetch<TFeature>(ref TFeature cached, Func<IFeatureCollection, TFeature> factory)
             where TFeature : class => Fetch(ref cached, Collection, factory);
+
+        private static int ContextDisposed()
+        {
+            ThrowContextDisposed();
+            return 0;
+        }
+
+        private static void ThrowContextDisposed()
+        {
+            // Can't use nameof as Microsoft.AspNetCore.Http.Abstractions is higher in the stack.
+            throw new ObjectDisposedException("HttpContext", "Request has finished and HttpContext disposed.");
+        }
     }
 }
--- a/src/Http/Http/src/DefaultHttpContext.cs
+++ b/src/Http/Http/src/DefaultHttpContext.cs
@@ -90,7 +90,7 @@ namespace Microsoft.AspNetCore.Http
         private IHttpRequestIdentifierFeature RequestIdentifierFeature =>
             _features.Fetch(ref _features.Cache.RequestIdentifier, _newHttpRequestIdentifierFeature);
 
-        public override IFeatureCollection Features => _features.Collection;
+        public override IFeatureCollection Features => _features.Collection ?? ContextDisposed();
 
         public override HttpRequest Request => _request;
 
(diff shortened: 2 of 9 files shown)


benaadams:
Rebased

benaadams:
Rebased

benaadams:
egh; something weird happened with the rebase

benaadams:
Fixed rebase

src/Http/Http.Features/src/FeatureReferences.cs
```diff
@@ -108,5 +108,17 @@ public void Initalize(IFeatureCollection collection, int revision)
 
         public TFeature Fetch<TFeature>(ref TFeature cached, Func<IFeatureCollection, TFeature> factory)
             where TFeature : class => Fetch(ref cached, Collection, factory);
+
+        private static int ContextDisposed()
```
dougbu: Curious: Why add these extra methods instead of changing their use above to
``` c#
if (Collection?.Revision == null)
{
    throw new ObjectDisposedException("HttpContext", "Request has finished and HttpContext disposed.");
}
```

src/Http/Http/src/DefaultHttpContext.cs
```diff
@@ -90,7 +90,7 @@ public void Uninitialize()
         private IHttpRequestIdentifierFeature RequestIdentifierFeature =>
             _features.Fetch(ref _features.Cache.RequestIdentifier, _newHttpRequestIdentifierFeature);
 
-        public override IFeatureCollection Features => _features.Collection;
+        public override IFeatureCollection Features => _features.Collection ?? ContextDisposed();
```
dougbu: Similarly, why not use a single body here instead of two methods?
```suggestion
        public override IFeatureCollection Features
        {
            get
            {
                if (_features.Collection == null)
                {
                    throw new ObjectDisposedException(nameof(HttpContext), $"Request has finished and {nameof(HttpContext)} disposed.");
                }

                return _features.Collection;
            }
        }
```

dougbu (commented):
@benaadams realize my review request was due to a bad rebase. But, a couple of curiosity / style questions&hellip;

src/Http/Http.Features/src/FeatureReferences.cs
benaadams: Inlining. 

`Load string, Load string, New Exception, Throw` is quite a chunky bit of code to inline to every caller

src/Http/Http/src/DefaultHttpContext.cs
benaadams: The Jit has special optimizations for a void method that only throws.

1. It recognises it as cold code (shouldn't be throwing exceptions in common paths), so generates the branches for it to be unpredicted (also does this with regular exception throws)
2. As it recognised it as cold code it won't inline it; whereas regular throws are not split out (Jit doesn't split methods); so the instruction cache is hotter (exception throws are quite big).
3. It doesn't generate the asm for it until the exception is thrown (faster Jitting).
4. As it recognises it as non-returning, it doesn't bother trying to sort out the stack after the call (popping regsiters back), meaning the code is hotter.
5. If the method is generic, moving the throws into a non-returning method in a non-generic class drastically reduces the asm size generated (doesn't do it for every type).
6. If the method inlines, moving the the throws into a non-returning method reduces the asm size generated as it doesn't regenerate it for every caller.

tl;dr makes the CPU instruction cache better; generics smaller; inlines smaller; faster Jitting

benaadams:
Superseded by https://github.com/aspnet/AspNetCore/pull/12385

benaadams:
Superseded by https://github.com/aspnet/AspNetCore/pull/12385

PR #11723, opened 2019-06-30
Don't capture HttpContext into the log objects

### LINKED ISSUE #10556: HostingRequestStartingLog shouldn't capture HttpRequest for its values

### LINKED ISSUE #10557: HostingRequestFinishedLog shouldn't capture HttpContext for its values

benaadams:
Don't capture HttpContext into the log objects from https://github.com/aspnet/AspNetCore/pull/10319

Resolves #10556
Resolves #10557

/cc @davidfowl

### DIFF
--- a/src/Hosting/Hosting/src/Internal/HostingRequestStartingLog.cs
+++ b/src/Hosting/Hosting/src/Internal/HostingRequestStartingLog.cs
@@ -13,7 +13,15 @@ namespace Microsoft.AspNetCore.Hosting
     {
         internal static readonly Func<object, Exception, string> Callback = (state, exception) => ((HostingRequestStartingLog)state).ToString();
 
-        private readonly HttpRequest _request;
+        public string Protocol { get; }
+        public string Method { get; }
+        public string ContentType { get; }
+        public long? ContentLength { get; }
+        public string Scheme { get; }
+        public string Host { get; }
+        public string PathBase { get; }
+        public string Path { get; }
+        public string QueryString { get; }
 
         private string _cachedToString;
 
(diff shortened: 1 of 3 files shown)


src/Hosting/Hosting/src/Internal/HostingRequestStartingLog.cs
```diff
@@ -51,7 +59,17 @@ internal class HostingRequestStartingLog : IReadOnlyList<KeyValuePair<string, ob
 
         public HostingRequestStartingLog(HttpContext httpContext)
         {
-            _request = httpContext.Request;
+            var request = httpContext.Request;
+
+            Protocol = request.Protocol;
+            Method = request.Method;
+            ContentType = !string.IsNullOrEmpty(request.ContentType) ? Uri.EscapeUriString(request.ContentType) : string.Empty;
```
davidfowl: Allocations!

davidfowl:
I think this change is fine but there's no issue with console and other loggers that store messages in a queue for further processing (at least the ones that we ship). They materialize the message before en-queuing.

benaadams:
>  there's no issue with console and other loggers that store messages

If you are happy with the current state, I'll close this as it does make `HostingRequestStartingLog` a bit bloaty

neuecc:
@davidfowl @benaadams 

The logger I'm developing has encountered this issue:
https://github.com/Cysharp/ZLogger

ZLogger adopts an architecture that consistently delays formatting and writes to an `IBufferWriter<byte>` to avoid allocating strings each time and for performance batching.
As a result, this logging, which captures the HttpContext, is causing problems:
https://github.com/Cysharp/ZLogger/issues/165

Looking at the PR, it seems that incorporating this would solve the issue.
Is there still a possibility of reconsidering and merging this?

Also, I think it's bad to capture the HttpContext.
I were aware that the presence of such a capture could cause problems, but I did not believe that there was anything in the standard that would do such a behavior.

PR #11730, opened 2019-07-01
MVC use concrete types for DiagnosticListener

benaadams:
Pattern: 

Under `Diagnostics` namespace

Event Type is same name as Event Name 

Type has a `EventName` string const so what to listen to is exposed without having to check the source (and is a const so can be used in `switch` and `Attribute`s).

Type has its data exposed as readonly properties.

Type implements `IReadOnlyList<KeyValuePair<string, object>>` so the data can be iterated over.

Type implements `IEnumerable<KeyValuePair<string, object>>` so the data can be enumerated over,

Interfaces are explicit so they don't interfere with reflection/serialization (i.e. `Count` being seen as a data item)

Each event has its own type so pattern matching can be used to differentiate events.

e.g.

```csharp
namespace Microsoft.AspNetCore.Mvc.Diagnostics
{
    public sealed partial class AfterAction : IEnumerable<KeyValuePair<string, object>>, IReadOnlyList<KeyValuePair<string, object>>
    {
        public const string EventName = "Microsoft.AspNetCore.Mvc.AfterAction";
        
        public AfterAction(ActionDescriptor actionDescriptor, HttpContext httpContext, RouteData routeData);
        
        public ActionDescriptor ActionDescriptor { get; }
        public HttpContext HttpContext { get; }
        public RouteData RouteData { get ; }

        int IReadOnlyCollection<KeyValuePair<string, object>>.Count { get; }
        KeyValuePair<string, object> IReadOnlyList<KeyValuePair<string, object>>.this[int index];

        IEnumerator<KeyValuePair<string, object>> IEnumerable<KeyValuePair<string, object>>.GetEnumerator(););
    }
}
```

/cc @rynowak @davidfowl @NickCraver


src/Mvc/Mvc.Razor/src/RazorPagePropertyActivator.cs
```diff
@@ -108,9 +108,9 @@ internal ViewDataDictionary CreateViewDataDictionary(ViewContext context)
             {
                 valueAccessor = valueAccessors.JsonHelperAccessor;
             }
-            else if (property.PropertyType == typeof(DiagnosticSource))
+            else if (property.PropertyType == typeof(DiagnosticListener))
```
davidfowl: This is breaking, it should support both.

benaadams: Returned

benaadams:
Are still anonymous diagnostics in other places like Hosting, but will see how this goes down first before seeing about those.

NickCraver:
This would be a huge improvement - constants as event names make both the adapter case and general usage (e.g. a `case` in an `IObserver`) much better. The strong typing is very welcomed for anyone taking the reference and makes the whole thing sane (over loads of reflection today).

If anyone's curious about context: https://github.com/dotnet/corefx/issues/18114 is the original issue pushing for this, but closed since the teams weren't going to make any more breaking changes. With 3.0, it seems like hopefully that's back on the table.

rynowak:
@benaadams - this is dope, I like the design of this.

rynowak:
@ajcvickers - is it work comparing the design of this with https://github.com/aspnet/EntityFrameworkCore/tree/9026770af72683ba3cf6c2a8125fa7a8d602703f/src/EFCore/Diagnostics ?

I notice that you're using the same payloads for logging and diagnostics source.

src/Http/Http/src/DefaultHttpContext.cs
```diff
@@ -159,7 +159,7 @@ public override ISession Session
 
         // This property exists because of backwards compatibility.
         // We send an anonymous object with an HttpContext property
-        // via DiagnosticSource in various events throughout the pipeline. Instead
+        // via DiagnosticListener in various events throughout the pipeline. Instead
```
rynowak: LOL, didn't know about this property. That's real spicy.

src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
```diff
@@ -42,14 +42,14 @@ public class DeveloperExceptionPageMiddleware
...
-        /// <param name="diagnosticSource"></param>
+        /// <param name="diagnosticListener"></param>
         /// <param name="filters"></param>
         public DeveloperExceptionPageMiddleware(
             RequestDelegate next,
             IOptions<DeveloperExceptionPageOptions> options,
             ILoggerFactory loggerFactory,
             IWebHostEnvironment hostingEnvironment,
-            DiagnosticSource diagnosticSource,
+            DiagnosticListener diagnosticListener,
```
rynowak: This is a breaking change that can easily be avoided.

src/Mvc/Mvc.Abstractions/ref/Microsoft.AspNetCore.Mvc.Abstractions.netcoreapp3.0.cs
```diff
@@ -209,6 +209,20 @@ public partial interface IAllowAnonymousFilter : Microsoft.AspNetCore.Mvc.Filter
     {
     }
 }
+namespace Microsoft.AspNetCore.Mvc.Diagnostics
+{
+    public abstract partial class MvcDiagnostic : System.Collections.Generic.IEnumerable<System.Collections.Generic.KeyValuePair<string, object>>, System.Collections.Generic.IReadOnlyCollection<System.Collections.Generic.KeyValuePair<string, object>>, System.Collections.Generic.IReadOnlyList<System.Collections.Generic.KeyValuePair<string, object>>, System.Collections.IEnumerable
```
rynowak: This should live in `Mvc.Core` there isn't really anything useful to do with the base class by itself. All of the implementation packages reference `Mvc.Core` 

I'd also prefer a name like `EventData` or `DiagnosticEventData` (something without MVC in the name).

src/Mvc/Mvc.Abstractions/src/Diagnostics/MvcDiagnostic.cs
```diff
@@ -0,0 +1,34 @@
...
+        int IReadOnlyCollection<KeyValuePair<string, object>>.Count => Count;
+        KeyValuePair<string, object> IReadOnlyList<KeyValuePair<string, object>>.this[int index] => this[index];
+
+        IEnumerator<KeyValuePair<string, object>> IEnumerable<KeyValuePair<string, object>>.GetEnumerator()
+            => GetEnumerator();
+
+        IEnumerator IEnumerable.GetEnumerator()
+            => GetEnumerator();
+
+        private IEnumerator<KeyValuePair<string, object>> GetEnumerator()
```
rynowak: Is there any value in a struct-enumerator here? For instance there's no problem with `GetEnumerator` as a public method.

src/Mvc/Mvc.Core/ref/Microsoft.AspNetCore.Mvc.Core.netcoreapp3.0.cs
```diff
@@ -1718,6 +1718,288 @@ public partial interface IAntiforgeryValidationFailedResult : Microsoft.AspNetCo
...
+        public Microsoft.AspNetCore.Http.HttpContext HttpContext { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
+        protected override System.Collections.Generic.KeyValuePair<string, object> this[int index] { get { throw null; } }
+        public Microsoft.AspNetCore.Routing.RouteData RouteData { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
+    }
+    public sealed partial class AfterActionMethod : Microsoft.AspNetCore.Mvc.Diagnostics.MvcDiagnostic
+    {
+        public const string EventName = "Microsoft.AspNetCore.Mvc.AfterActionMethod";
+        public AfterActionMethod(Microsoft.AspNetCore.Mvc.ActionContext actionContext, System.Collections.Generic.IDictionary<string, object> arguments, object controller, Microsoft.AspNetCore.Mvc.IActionResult result) { }
+        public Microsoft.AspNetCore.Mvc.ActionContext ActionContext { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
+        public System.Collections.Generic.IDictionary<string, object> Arguments { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
```
rynowak: Ideally this would be `IReadOnlyDictionary<>` We want to discourage folks from using these to modify data. Since this is all reflection based, it shouldn't be a breaking change - this is just declaring the shape of the API.

src/Mvc/Mvc.Core/ref/Microsoft.AspNetCore.Mvc.Core.netcoreapp3.0.cs
```diff
@@ -1718,6 +1718,288 @@ public partial interface IAntiforgeryValidationFailedResult : Microsoft.AspNetCo
...
+        public const string EventName = "Microsoft.AspNetCore.Mvc.AfterActionResult";
+        public AfterActionResult(Microsoft.AspNetCore.Mvc.ActionContext actionContext, Microsoft.AspNetCore.Mvc.IActionResult result) { }
+        public Microsoft.AspNetCore.Mvc.ActionContext ActionContext { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
+        protected override int Count { get { throw null; } }
+        protected override System.Collections.Generic.KeyValuePair<string, object> this[int index] { get { throw null; } }
+        public Microsoft.AspNetCore.Mvc.IActionResult Result { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
+    }
+    public sealed partial class AfterOnActionExecuted : Microsoft.AspNetCore.Mvc.Diagnostics.MvcDiagnostic
+    {
+        public const string EventName = "Microsoft.AspNetCore.Mvc.AfterOnActionExecuted";
```
rynowak: Ooops. This is a really unfortunate event name. I think we should call this type something like `ActionFilterOnActionExecuted` on so on for the other filter-related events.

src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
benaadams: Means lose the fast inlining `diagnosticListerner.IsEnabled()` test and have to use the slow virtual `diagnosticSource.IsEnabled(string))`

src/Mvc/Mvc.Core/ref/Microsoft.AspNetCore.Mvc.Core.netcoreapp3.0.cs
```diff
@@ -1718,6 +1718,288 @@ public partial interface IAntiforgeryValidationFailedResult : Microsoft.AspNetCo
...
+        public Microsoft.AspNetCore.Http.HttpContext HttpContext { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
+        protected override System.Collections.Generic.KeyValuePair<string, object> this[int index] { get { throw null; } }
+        public Microsoft.AspNetCore.Routing.RouteData RouteData { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
+    }
+    public sealed partial class BeforeActionMethod : Microsoft.AspNetCore.Mvc.Diagnostics.MvcDiagnostic
+    {
+        public const string EventName = "Microsoft.AspNetCore.Mvc.BeforeActionMethod";
+        public BeforeActionMethod(Microsoft.AspNetCore.Mvc.ActionContext actionContext, System.Collections.Generic.IDictionary<string, object> arguments, object controller) { }
+        public Microsoft.AspNetCore.Mvc.ActionContext ActionContext { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
+        public System.Collections.Generic.IDictionary<string, object> Arguments { [System.Runtime.CompilerServices.CompilerGeneratedAttribute]get { throw null; } }
```
rynowak: `IReadOnlyDictionary<>`

src/Mvc/Mvc.Core/src/Diagnostics/MvcDiagnostics.cs
```diff
@@ -0,0 +1,709 @@
...
+        protected override int Count => 3;
+
+        public BeforeAction(ActionDescriptor actionDescriptor, HttpContext httpContext, RouteData routeData)
+        {
+            ActionDescriptor = actionDescriptor;
+            HttpContext = httpContext;
+            RouteData = routeData;
+        }
+
+        protected override KeyValuePair<string, object> this[int index] => index switch
```
rynowak: PranavPoints++

src/Mvc/Mvc.Core/src/Diagnostics/MvcDiagnostics.cs
```diff
@@ -0,0 +1,709 @@
...
+    public sealed class BeforeAction : MvcDiagnostic
+    {
+        public const string EventName = EventNamespace + nameof(BeforeAction);
+        public ActionDescriptor ActionDescriptor { get; }
+        public HttpContext HttpContext { get; }
+        public RouteData RouteData { get; }
+
+        protected override int Count => 3;
+
+        public BeforeAction(ActionDescriptor actionDescriptor, HttpContext httpContext, RouteData routeData)
```
rynowak: We typically do:
- Fields
- Constructor
- Properties
- Methods

src/Mvc/Mvc.Core/src/Diagnostics/MvcDiagnostics.cs
```diff
@@ -0,0 +1,709 @@
...
+using System;
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc.Abstractions;
+using Microsoft.AspNetCore.Mvc.Filters;
+using Microsoft.AspNetCore.Routing;
+
+namespace Microsoft.AspNetCore.Mvc.Diagnostics
+{
+    public sealed class BeforeAction : MvcDiagnostic
```
rynowak: We typically do one class per-file - but I don't feel really strongly about it in this case.

ajcvickers:
@rynowak We don't use the same payloads, although we use the same infrastructure to create them. We generally create simple strings to go in `ILogger` messages, since controlling the formatting is hard for other objects, and use actual runtime instances directly in `DiagnosticListener` events. This avoids creating strings for diagnostic events, which often don't need them. However, we capture the ability to create the strings in the diagnostics payload so that calling `ToString` on it will dynamically create and return the string that we would have used for `ILogger`.

rynowak:
@ajcvickers - do you have any feedback on the design proposal [here](https://github.com/aspnet/AspNetCore/pull/11730#issue-293144112) based on EF's experience using nomimal types for `DiagnosticSource` in EF?

src/Mvc/Mvc.Abstractions/ref/Microsoft.AspNetCore.Mvc.Abstractions.netcoreapp3.0.cs
benaadams: `EventData` in the `Diagnostics` namespace sounds good

src/Mvc/Mvc.Abstractions/src/Diagnostics/MvcDiagnostic.cs
benaadams: Sure why not :)

benaadams:
@davidfowl lulled me into this with the suggestion MVC only had 5 events

> We don't have many events, we have 3 in hosting, 1 in each of the error middleware and a couple in MVC (maybe 5). https://twitter.com/davidfowl/status/1145450075780308992

I feel it was an underestimation...

src/Mvc/Mvc.Razor/src/Diagnostics/MvcDiagnostics.cs
```diff
@@ -0,0 +1,140 @@
...
+using Microsoft.AspNetCore.Mvc.Abstractions;
+using Microsoft.AspNetCore.Mvc.Razor;
+using Microsoft.AspNetCore.Mvc.Rendering;
+
+namespace Microsoft.AspNetCore.Mvc.Diagnostics
+{
+    public sealed class BeforeViewPage : MvcDiagnostic
+    {
+        public const string EventName = EventNamespace + 
+            "Razor." +
```
benaadams: Since we are changing some event names; though these ones have the additional "Razor." in their name?

src/Middleware/Diagnostics/src/DeveloperExceptionPage/DeveloperExceptionPageMiddleware.cs
benaadams: Suppose it should be enabled in production

benaadams:
Done everything but event renames; need some more clarity here

src/Mvc/Mvc.Core/ref/Microsoft.AspNetCore.Mvc.Core.netcoreapp3.0.cs
benaadams: Is that everything that has  `IFilterMetadata` property?

e.g. `AfterOnException` => `AfterFilterOnException`

rynowak:
> lulled me into this with the suggestion MVC only had 5 events

Don't trust fowler when he wants something from you :trollface:

src/Mvc/Mvc.Razor/src/Diagnostics/MvcDiagnostics.cs
rynowak: Wait which ones change? I don't think we want to change any event names. I made a comment before about the **class names**. Sorry if that was unclear.

src/Mvc/Mvc.Razor/src/Diagnostics/MvcDiagnostics.cs
```diff
@@ -0,0 +1,140 @@
...
+        {
+            0 => new KeyValuePair<string, object>(nameof(Page), Page),
+            1 => new KeyValuePair<string, object>(nameof(ViewContext), ViewContext),
+            2 => new KeyValuePair<string, object>(nameof(ActionDescriptor), ActionDescriptor),
+            3 => new KeyValuePair<string, object>(nameof(HttpContext), HttpContext),
+            _ => throw new IndexOutOfRangeException(nameof(index))
+        };
+    }
+
+    public sealed class BeginInstrumentationContext : EventData
```
rynowak: If possible, I'd prefer not to define these events. I don't think we actually fire them anymore.

src/Mvc/Mvc.Razor/src/RazorPage.cs
```diff
@@ -274,40 +274,11 @@ public override void EnsureRenderedBodyOrSections()
...
-                        httpContext = Context,
-                        path = Path,
-                        position = position,
-                        length = length,
-                        isLiteral = isLiteral,
-                    });
-            }
-        }
+        public override void BeginContext(int position, int length, bool isLiteral) 
+            => DiagnosticListener.BeginInstrumentationContext(Context, Path, position, length, isLiteral);
```
rynowak: @pranavkm @NTaylorMullen  - we no longer call this from the generated code right?

src/Mvc/Mvc.Razor/src/RazorPageActivator.cs
```diff
@@ -32,7 +32,7 @@ public class RazorPageActivator : IRazorPageActivator
             IModelMetadataProvider metadataProvider,
             IUrlHelperFactory urlHelperFactory,
             IJsonHelper jsonHelper,
-            DiagnosticSource diagnosticSource,
+            DiagnosticListener diagnosticListener,
```
rynowak: This is a breaking change that we should not make.

src/Mvc/Mvc.Razor/src/Diagnostics/MvcDiagnostics.cs
benaadams: https://github.com/aspnet/AspNetCore/pull/11730#discussion_r302622142 event names are `nameof(classname)` so the current event name drives the class name?

benaadams: https://github.com/aspnet/AspNetCore/pull/11730#discussion_r302622142 event names are `nameof(classname)` so the current event name drives the class name?

Weird if there is a mis-match; that you listen to one name and get different named events?

src/Mvc/Mvc.Core/ref/Microsoft.AspNetCore.Mvc.Core.netcoreapp3.0.cs
rynowak: I would call it `ExceptionFilterAfterOnException`. So the scheme is `<filter kind> [Before|After] Method`

src/Mvc/Mvc.Razor/src/Diagnostics/MvcDiagnostics.cs
rynowak: That's just the code you wanted to write 😆. I don't think it's a good idea for us to just change the event names and break people. 

I don't think it's weird if the class names add clarity, since the event names are now constants, you don't really see them.

The reason I brought this up is because of events like the following:

```txt
M.A.Mvc.BeforeAction (Before we begin processing for an action)
M.A.Mvc.BeforeOnActionExecution (Before an async action filter)
```

These seem like they are interchangeable but they are super different.

if you really feel like this naming thing is too much, just leave the names as they were originally and I'll do a pass after this is merged.

src/Mvc/Mvc.Razor/src/RazorPage.cs
pranavkm: We don't generate code that calls these methods anymore. Can we have these no-op instead?

NTaylorMullen: Yup, we don't gen code for this. Honestly, we should at the very least obsolete these methods for 3.0

benaadams: obsolete and make `sealed`; they are called from a couple methods in the code base, so I'll let you do it in another PR. Have nooped them.

src/Mvc/Mvc.Razor/src/Diagnostics/MvcDiagnostics.cs
benaadams: Removed

src/Mvc/Mvc.Razor/src/Diagnostics/MvcDiagnostics.cs
benaadams: I'll let you pick them up as I think you know what you want them named more :)

src/Mvc/Mvc.Razor/src/RazorPageActivator.cs
benaadams: :(

benaadams: Changed back

NickCraver:
I'm just curious around what constitutes a breaking change on these APIs w.r.t. event names.

Since the property names on the object are going from camelCase to proper C# Pascal case for properties, isn't it already breaking? Or is the latest binder reflection case-insensitive and that keeps working? I'm just confused on the event names being breaking when the property casing isn't..._if_ we're firing the old events but they just don't work, maybe it _would be_ better to change and break the event names? Just tossing that thought out there.

rynowak:
> Since the property names on the object are going from camelCase to proper C# Pascal case for properties, isn't it already breaking? Or is the latest binder reflection case-insensitive and that keeps working?

It depends on the consumer. The consumer that I'm aware of is app-insights which I think is still using the `Microsoft.Extensions.DiagnosticSourceAdapter` package - which is case-insensitive.

Technically speaking, top level entries are defined as case-insensitive, but what really matters to me is whether we break people that have real usage.

If this is a breaking change that impacts existing usage of diagnostic source, then I would lean towards using lowercase properties. This is too late in the development cycle to be cavalier about breaking changes.

@lmolkova - any insights here?

NickCraver:
Given the very few people using these (as far as I'm aware) outside diagnostic adapter, IMO we should change them now while making changes and have them reflect the casing that should be on a public type API. Even if it's breaking, as long as `DiagnosticAdapter` is case-insensitive.

benaadams:
Aside: Just to make note on my understanding of the status of this PR. I've made the requested changes and @rynowak is picking up class naming changes.

Also outstanding issue on Instrumentation apis https://github.com/aspnet/AspNetCore/pull/11730#discussion_r303040343 to be picked up separately?

rynowak:
There's still a breaking change here: https://github.com/aspnet/AspNetCore/pull/11730/files#diff-1f8dd5fa9cfd7a11a18340f90b1706d7R18

benaadams:
> There's still a breaking change here:

Surely not with the magic of DI and all that 😉 

Reverted

lmolkova:
I'll look more into this PR later today.

Regarding this:

> Microsoft.Extensions.DiagnosticSourceAdapter package - which is case-insensitive.

AppInsights is NOT using DiagnosticAdapter anymore and event handling is case-sensitive. We still have several work items to support ASP.NET Core 3.0 and can fit case-insensitive work if needed.


Also

> very few people using these (as far as I'm aware) outside diagnostic adapter

After we introduced Activities, it became hard and inefficient to use DiagnosticAdapter because of https://github.com/aspnet/EventNotification/issues/67 (check out this https://github.com/aspnet/Extensions/issues/1895 as an example I just found)

We also removed adapter because of perf - DiagnosticsSource.IsEnabled became twice as fast. And the ability to filter out events in IsEnabled is not supported by the adapter,

So, I'd argue DiagnosticAdapter is the default way is a good assumption.

rynowak:
> I'll look more into this PR later today.
> 
> Regarding this:
> 
> > Microsoft.Extensions.DiagnosticSourceAdapter package - which is case-insensitive.
> 
> AppInsights is NOT using DiagnosticAdapter anymore and event handling is case-sensitive. We still have several work items to support ASP.NET Core 3.0 and can fit case-insensitive work if needed.

Ok great. Is there somewhere you want me to file this?

> After we introduced Activities, it became hard and inefficient to use DiagnosticAdapter because of [aspnet/EventNotification#67](https://github.com/aspnet/EventNotification/issues/67) (check out this [aspnet/Extensions#1895](https://github.com/aspnet/Extensions/issues/1895) as an example I just found)
> 
> We also removed adapter because of perf - DiagnosticsSource.IsEnabled became twice as fast. And the ability to filter out events in IsEnabled (which brought more improvements) is not supported by the adapter.
> 
> So, I'd argue DiagnosticAdapter is the default way is not a good assumption.

Also great from my POV. I'd like to use deprecate the adapter and stop shipping it in a future release. It hasn't really proved to be essential for diagnostic source.

rynowak (approved):

PR #11748, opened 2019-07-01
Fix Blazor VSIX

SteveSandersonMS:
Making a PR to see if this works. I've changed the `PreCreateVsixContainer` condition to match @dougbu's suggestion.

### DIFF
--- a/src/Components/Blazor/BlazorExtension/src/Microsoft.VisualStudio.BlazorExtension.csproj
+++ b/src/Components/Blazor/BlazorExtension/src/Microsoft.VisualStudio.BlazorExtension.csproj
@@ -52,7 +52,7 @@
     <CopyOutputSymbolsToOutputDirectory>true</CopyOutputSymbolsToOutputDirectory>
   </PropertyGroup>
 
-  <Target Name="PreCreateVsixContainer" BeforeTargets="GetVsixSourceItems">
+  <Target Name="PreCreateVsixContainer" BeforeTargets="GetVsixSourceItems" Condition="'$(TargetOsName)' == 'win' AND '$(TargetArchitecture)' != 'arm'">
     <ItemGroup>
       <_TemplatePackage Include="$(ArtifactsShippingPackagesDir)Microsoft.AspNetCore.Blazor.Templates.*.nupkg" />
     </ItemGroup>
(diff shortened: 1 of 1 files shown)


src/Components/Blazor/BlazorExtension/src/Microsoft.VisualStudio.BlazorExtension.csproj
```diff
@@ -52,7 +52,7 @@
     <CopyOutputSymbolsToOutputDirectory>true</CopyOutputSymbolsToOutputDirectory>
   </PropertyGroup>
 
-  <Target Name="PreCreateVsixContainer" BeforeTargets="GetVsixSourceItems" Condition="'$(IsPackable)' == 'true'">
+  <Target Name="PreCreateVsixContainer" BeforeTargets="GetVsixSourceItems" Condition="'$(TargetOsName)' == 'win' AND '$(TargetArchitecture)' != 'arm'">
```
dougbu: Double-checked and the `'$(TargetOsName)' == 'win'` bit isn't necessary because this project is only built with full `msbuild` i.e. on Windows. But, it doesn't hurt.

dougbu (approved):
Recommend getting this in if tests pass and the created VSIX contains the **Right Stuff:tm:** i.e. treat my command as FYI

mkArtakMSFT:
@SteveSandersonMS we should verify this as soon as we can, so ideally we merge the fix before branching.

SteveSandersonMS:
@dougbu This should be merged before we branch, but the build has hung twice for unrelated reasons. I would recommend just merging it now, though don’t have the power to do so myself.

mkArtakMSFT:
I've force-merged this as @SteveSandersonMS has verified the VSIX contains the right stuff.

dougbu:
@SteveSandersonMS do the template tests cover the Blazor VSIX at all? Just want to be sure the test hang there really is unrelated to this change.
```
2019-07-01T16:34:14.6376151Z   Running tests: F:\workspace\_work\1\s\src\ProjectTemplates\test\bin\Release\netcoreapp3.0\ProjectTemplates.Tests.dll [netcoreapp3.0\|x64] |  
-- | --
  | 2019-07-01T18:55:43.7767086Z Attempting to cancel the build...
```

SteveSandersonMS:
The template tests don’t cover the Blazor VSIX.

PR #11752, opened 2019-07-01
Get rid of pubternal in routing

rynowak:
The only pubternal that remains is DfaGraphWriter - it's a good use case
for pubternal because it's something we want to expose, but we'd never
service and would willingly break the API in the future. It's also
really unlikely that anyone would build on top of it.

Contributes to #11312


jkotalik (approved):
💰

davidfowl (approved):

davidfowl:
See if we can get this into preview7

pranavkm (approved):

PR #11757, opened 2019-07-01
Fix race with CTS disposing

BrennanConroy:
Fixes https://github.com/aspnet/AspNetCore-Internal/issues/2293

`DisposeAsync` will dispose the CTS and then set it to null, however we could still try to cancel it between the dispose and the null resulting in ODE.

Couldn't think of a great way to do this without locking, but it shouldn't be a contested lock so it's fine.

### DIFF
--- a/src/SignalR/common/Http.Connections/src/Internal/HttpConnectionContext.cs
+++ b/src/SignalR/common/Http.Connections/src/Internal/HttpConnectionContext.cs
@@ -253,7 +253,7 @@ namespace Microsoft.AspNetCore.Http.Connections.Internal
 
                     if (TransportType == HttpTransportType.WebSockets)
                     {
-                        // The websocket transport will close the application output automatically when reading is cancelled
+                        // The websocket transport will close the application output automatically when reading is canceled
                         Cancellation?.Cancel();
                     }
                     else
@@ -440,6 +440,37 @@ namespace Microsoft.AspNetCore.Http.Connections.Internal
                 // Connection was disposed
                 nonClonedContext.Response.StatusCode = StatusCodes.Status404NotFound;
                 nonClonedContext.Response.ContentType = "text/plain";
+            }
+        }
+
+        internal async Task<bool> CancelPreviousPoll(HttpContext context)
+        {
+            CancellationTokenSource cts;
+            lock (_stateLock)
+            {
+                // Need to sync cts access with DisposeAsync as that will dispose the cts
+                cts = Status == HttpConnectionStatus.Disposed ? null : Cancellation;
+            }
+
+            using (cts)
+            {
+                // Cancel the previous request
+                cts?.Cancel();
+
+                try
+                {
+                    // Wait for the previous request to drain
+                    await PreviousPollTask;
+                }
+                catch (OperationCanceledException)
+                {
+                    // Previous poll canceled due to connection closing, close this poll too
+                    context.Response.ContentType = "text/plain";
+                    context.Response.StatusCode = StatusCodes.Status204NoContent;
+                    return false;
+                }
+
+                return true;
             }
         }
 
--- a/src/SignalR/common/Http.Connections/src/Internal/HttpConnectionContext.cs
+++ b/src/SignalR/common/Http.Connections/src/Internal/HttpConnectionContext.cs
@@ -449,7 +449,15 @@ namespace Microsoft.AspNetCore.Http.Connections.Internal
             lock (_stateLock)
             {
                 // Need to sync cts access with DisposeAsync as that will dispose the cts
-                cts = Status == HttpConnectionStatus.Disposed ? null : Cancellation;
+                if (Status == HttpConnectionStatus.Disposed)
+                {
+                    cts = null;
+                }
+                else
+                {
+                    cts = Cancellation;
+                    Cancellation = null;
+                }
             }
 
             using (cts)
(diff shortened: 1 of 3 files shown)


mikaelm12 (approved):
Let's see if this fixes the ODE

src/SignalR/common/Http.Connections/src/Internal/HttpConnectionContext.cs
```diff
@@ -439,6 +439,37 @@ private void FailActivationUnsynchronized(HttpContext nonClonedContext, ILogger
...
+            lock (_stateLock)
+            {
+                // Need to sync cts access with DisposeAsync as that will dispose the cts
+                cts = Status == HttpConnectionStatus.Disposed ? null : Cancellation;
+            }
+
+            using (cts)
+            {
+                // Cancel the previous request
+                cts?.Cancel();
```
halter73: If you don't keep _stateLock trough the call to Cancel, this could still ODE as this gets disposed by another thread calling DispoasAsync(). After the _stateLock is released by CancelPreviousPoll(), DispoasAsync() can acquire the lock, see the HttpConnectionStatus still isn't dispose and start disposing the Cancellation CTS before CancelPreviousPoll() ever calls Cancel() because the lock was released.

halter73 (changes requested):
We need to keep the _statLock until the HttpConnectionContext is put into a state where it's impossible to have the Cancellation CTS disposed from underneath us.

src/SignalR/common/Http.Connections/src/Internal/HttpConnectionContext.cs
BrennanConroy: Whoops, I knew that and meant to null it out here...

halter73: Cancellation can never get reinitialized, right?

halter73 (approved):

src/SignalR/common/Http.Connections/src/Internal/HttpConnectionContext.cs
BrennanConroy: It can, but it's in `TryActivateLongPollingConnection` which is inside the lock

halter73: Make sure `TryActivateLongPollingConnection` cannot recreate the CTS. It doesn't look like TryActivateLongPollingConnection checks whether Status is HttpConnectionStatus.Disposed currently.

BrennanConroy: We both didn't line up the curly braces correctly when looking at this :D
If the status != inactive it goes to the `FailActivationUnsynchronized` call which will set the correct response and return

src/SignalR/common/Http.Connections/test/HttpConnectionDispatcherTests.cs
```diff
@@ -1051,9 +1051,20 @@ public async Task RequestToActiveConnectionIdKillsPreviousConnectionLongPolling(
                 Assert.True(request1.IsCompleted);
 
                 request1 = dispatcher.ExecuteAsync(context1, options, app);
+                // Wait until the request has started internally
+                while (connection.TransportTask.IsCompleted)
```
BrennanConroy: I believe this test was flaky before (although no failures happened) it was just a very small race, but now that I added another statemachine the test became more flaky and this is my "solution" to make it consistent.

Before we dispatched 2 polls and assumed that one would start before the other, but they actually race each other once the first one goes async. I couldn't find a good place to inject our own sync points so am waiting on the `TransportTask` to change meaning the request has gotten far enough.

If someone has a better idea please help :)

BrennanConroy: Might want to add a maximum wait here...

BrennanConroy:
Ping, I modified this a little since when it was approved.

Had to do some ugly test hackery to try to make the test do what it claimed to be doing.

src/SignalR/common/Http.Connections/test/HttpConnectionDispatcherTests.cs
```diff
@@ -1051,9 +1051,24 @@ public async Task RequestToActiveConnectionIdKillsPreviousConnectionLongPolling(
...
 
                 request1 = dispatcher.ExecuteAsync(context1, options, app);
+                var count = 0;
+                // Wait until the request has started internally
+                while (connection.TransportTask.IsCompleted && count < 50)
+                {
+                    count++;
+                    await Task.Delay(1);
+                }
+                var transportTask = connection.TransportTask;
```
halter73: Where is this var used?

src/SignalR/common/Http.Connections/test/HttpConnectionDispatcherTests.cs
```diff
@@ -1051,9 +1051,24 @@ public async Task RequestToActiveConnectionIdKillsPreviousConnectionLongPolling(
                 Assert.True(request1.IsCompleted);
 
                 request1 = dispatcher.ExecuteAsync(context1, options, app);
+                var count = 0;
+                // Wait until the request has started internally
+                while (connection.TransportTask.IsCompleted && count < 50)
+                {
+                    count++;
+                    await Task.Delay(1);
```
halter73: Shouldn't the test fail if the while loop ends because the count was exceeded?

anurse:
@BrennanConroy do you want to try and get this in today for preview 2? Assuming not (because of the milestone) I've labeled this "NO MERGE" so we hold off on merging until the branch is open for preview 3 changes.

BrennanConroy:
I don't want to rush it, but it would be nice to fix in 3.1. Fixes most LongPolling tests from being flaky

anurse:
Cool. I believe we stay in M2 approval for Preview 3 so let's bring it up then. Especially if it reduces test flakiness :).

halter73 (approved):

anurse:
Branches are open again for preview 3 work, but this PR does still need @Pilchie approval before merging.

Pilchie:
Approved for 3.1 Preview 3.  Thanks @BrennanConroy.

Issue #10505, opened 2019-05-23
Move JSON helpers for HttpClient in Blazor out of .Net Core 3.0


mkArtakMSFT:
This should be an OOB experimental package.

Wider issue: https://github.com/aspnet/AspNetCore/issues/10397

PR #11767, opened 2019-07-01
Move Blazor HttpClient functionality out of sfx

### LINKED ISSUE #11756: Blazor API Review: HttpClient functionality
# Blazor API Review: HttpClient functionality

Blazor currently has extension methods on `HttpClient` for using the new System.Text.Json serializer. Ultimately we want this functionality to be part of CoreFx, and not something that the Blazor team control.

As part of 3.0 we should move this functionality to another package that will keep its *preview* branding until similar convenience APIs are available from CoreFx. This avoids any naming conflicts within the actual shipping product.

Once the functionality moves to CoreFx, we can encourage users to find it there and stop shipping our package, but it won't break anyone's app or components. Shipped code that that depends on this package will continue working (not a binary breaking change because its already a separate package).

I'm proposing **Microsoft.AspNetCore.Blazor.HttpClient** as the package/assembly name. 

This name will not be used as a namespace since all of the types in the package are in existing namespaces.

## APIs

There are no planned changes to the primary API as part of this change - just a package split. Since these APIs remain in preview they are not our focus at this time.

```C#
    public static partial class HttpClientJsonExtensions
    {
        public static System.Threading.Tasks.Task<T> GetJsonAsync<T>(this System.Net.Http.HttpClient httpClient, string requestUri) { throw null; }
        public static System.Threading.Tasks.Task PostJsonAsync(this System.Net.Http.HttpClient httpClient, string requestUri, object content) { throw null; }
        public static System.Threading.Tasks.Task<T> PostJsonAsync<T>(this System.Net.Http.HttpClient httpClient, string requestUri, object content) { throw null; }
        public static System.Threading.Tasks.Task PutJsonAsync(this System.Net.Http.HttpClient httpClient, string requestUri, object content) { throw null; }
        public static System.Threading.Tasks.Task<T> PutJsonAsync<T>(this System.Net.Http.HttpClient httpClient, string requestUri, object content) { throw null; }
        public static System.Threading.Tasks.Task SendJsonAsync(this System.Net.Http.HttpClient httpClient, System.Net.Http.HttpMethod method, string requestUri, object content) { throw null; }
        public static System.Threading.Tasks.Task<T> SendJsonAsync<T>(this System.Net.Http.HttpClient httpClient, System.Net.Http.HttpMethod method, string requestUri, object content) { throw null; }
    }
```

## Additional Details

We need to decide what to do with `JsonSerializerOptionsProvider`. This class provides **internal** access to a `JsonSerializerOptions` object that's used by several of our components. This is internal currently and used **inside** the methods that are moving so we need to make a decision.

```C#
using System.Text.Json;

{
    internal static class JsonSerializerOptionsProvider
    {
        public static readonly JsonSerializerOptions Options = new JsonSerializerOptions
        {
            PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
            PropertyNameCaseInsensitive = true,
        };
    }
}
```

It seems like these settings are used everywhere Blazor wants to interop with javascript (but not via JS interop). For instance when using `InvokeUnmarshalled`. I think we could safely make this type shared-source with few drawbacks since it's not mutated and not exposed.

----

We need to decide how users acquire this package. There are a few options (can be combined):
- Just publish the package and update docs
- Make `M.A.Blazor` depend on new package
- Add new package to templates (what about RCL? that's an *RTW* template)

I think it's most important that we get an answer for the RCL question. Given that we have HTTP in our client-side template, we need to show the correct way to do it, which means it's either a dependency of those templates or of `M.A.Blazor`.

rynowak:
Fixes: #11756 #10505

See description in #11756


src/Components/Components/src/Microsoft.AspNetCore.Components.csproj
```diff
@@ -9,6 +9,10 @@
     <RazorLangVersion>3.0</RazorLangVersion>
   </PropertyGroup>
 
+    <ItemGroup>
```
dougbu: All Ryans everywhere space things strangely.

dougbu (commented):
FWIW this looks fine to me.

Just curious: Why do we need to ship this class separately, is there an issue tracking this change, and does it need to happen in preview 7?

rynowak:
> Just curious: Why do we need to ship this class separately, is there an issue tracking this change, and does it need to happen in preview 7?

The issue is linked in the description.

pranavkm (approved):

src/Components/Components/src/Microsoft.AspNetCore.Components.csproj
rynowak: S U R R E A L F O R M A T T I N G

